PR #4216, opened 2020-09-25
Bug 1859174: Upgrade Terraform provider OpenStack

pierreprinetti:
Also upgrade the Gophercloud dependency and Gophercloud-utils.


data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -22,77 +22,58 @@ spec:
         description: InstallConfig is the configuration for an OpenShift install.
         properties:
           additionalTrustBundle:
-            description: AdditionalTrustBundle is a PEM-encoded X.509 certificate
-              bundle that will be added to the nodes' trusted certificate store.
+            description: AdditionalTrustBundle is a PEM-encoded X.509 certificate bundle that will be added to the nodes' trusted certificate store.
```
abhinavdahiya, line 25: hmm.. I can't figure out why these changes from multi-line to single string are happening here?

pierreprinetti: I'm not sure either, but if I don't run `./hack/verify-codegen.sh`, then [it fails in the CI](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/4216/pull-ci-openshift-installer-master-verify-codegen/1310860004277882880).

pierreprinetti: Possibly a side effect of [bumping `gopkg.in/yaml` to v2.3.0](https://github.com/openshift/installer/pull/4216/commits/b246fd4baeeb16b256d7a1d2f2738827bad09d23#diff-37aff102a57d3d7b797f152915a6dc16R98).

pierreprinetti:
infra issues, should be resolved now.
/retest

PR #4226, opened 2020-09-29
Bug 1881532: openstack: Only list external networks in prompt

pierreprinetti:
Non-external networks are not accepted and would make the Installer
error.


mandre (commented):
/approve
/lgtm
/hold until tests pass

PR #4227, opened 2020-09-29
Bug 1884558: do not use local cacert path in generated clouds.yaml

Fedosin:
When we generate an OpenShift cred secret with clouds.yaml, we keep the local cacert path there. In other words, path to cacert on the user's machine.

This is wrong, because this value is useless after the installation. To avoid this we need to use the predefined const value, like we do with cloud config:  https://github.com/openshift/installer/blob/4aac205a59a9102ddf1ae329b3c4c6f0dff7e330/pkg/asset/manifests/openstack/cloudproviderconfig.go#L21

/label platform/openstack

### DIFF
--- a/pkg/asset/manifests/openshift.go
+++ b/pkg/asset/manifests/openshift.go
@@ -129,6 +129,12 @@ func (o *Openshift) Generate(dependencies asset.Parents) error {
 		if err != nil {
 			return err
 		}
+
+		// We need to replace the local cacert path with one that is used in OpenShift
+		if cloud.CACertFile != "" {
+			cloud.CACertFile = "/etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem"
+		}
+
 		clouds := make(map[string]map[string]*clientconfig.Cloud)
 		clouds["clouds"] = map[string]*clientconfig.Cloud{
 			osmachine.CloudName: cloud,


mandre (commented):
/lgtm
/approve

pkg/asset/manifests/openshift.go
```diff
@@ -129,6 +129,12 @@ func (o *Openshift) Generate(dependencies asset.Parents) error {
 		if err != nil {
 			return err
 		}
+
+		// We need to replace the local cacert path with one that is used in OpenShift
+		if cloud.CACertFile != "" {
+			cloud.CACertFile = "/etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem"
```
wking, line 135: Will this break [Terraform variable generation][1]?  I'd be happier with a new asset, or asset property, that read the local CA cert file from the user-configured location into the asset store, so we didn't have to read it in dynamically durring Terraform variable generation.  Then we wouldn't have to worry about whether clobbering the value here would have an impact on that behavior or not.

[1]: https://github.com/openshift/installer/blob/7a772518015fc14b48426344e8b3800b16b50d15/pkg/asset/cluster/tfvars.go#L384-L385

abhinavdahiya:
just so that i can understand the context.
- the installer takes the user credentials and puts them in the cluster so that cred minter can make new ones for operators.
- the openstack creation of this secret takes the user cloud-configuration as-is, it has credentials and other configuration like the ca bundle etc. what else?

In my opinion the secret should only contain credentials and nothing else. All other configuration about how to communicate with the cloud apis should be extracted from actual objects like the infrastructures.config.openshift.io .

So is it possible to restrict this secret to creds only, and if not now can we plan to make that deprecation/move.
And I would prefer we set the ca bundle field to empty always because no one should use this field for configuring.

Fedosin:
@abhinavdahiya hello! you got it right - we use `clouds.yaml` file in the installer to provision basic resources, but we also copy its contents to "openstack-credentials" secret to allow various OpenShift components to interact with OpenStack.

The issue is that since we copy contents unmodified, we still have local user's `cacert` file path in the secret, for instance "/home/john/my-ca-bundle.pem". Apparently this value doesn't make sense for OpenShift because we always use the constant path "/etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem" for all components that communicate with OpenStack. To avoid this we have to introduce hacks: https://github.com/openshift/csi-driver-manila-operator/blob/master/pkg/controllers/secret/secretsync.go#L149-L152
It would be much better if we could set correct value there initially, during manifest generation in the installer, rather than updating it manually in the components.

In the future release we will consider keeping only auth parameters (login/password/project) in clouds.yaml and moving unrelated parts, like cacert path, to `infrastructures.config.openshift.io`, as you ask.

abhinavdahiya:
> 
> In the future release we will consider keeping only auth parameters (login/password/project) in clouds.yaml and moving unrelated parts, like cacert path, to `infrastructures.config.openshift.io`, as you ask.

I think we should be setting this field to empty instead of this _default_ which based on it's contents is valid only for static pod resources on the node and not general pods trying to communicate with openstack api.

regardless if you think we can come back and fix later, I would recommend you create a JIRA story for the shiftstack backlog, hopefully targeted for 4.7, and link here. At that point I'm more than happy to approve this change for the time being.

Fedosin:
@abhinavdahiya agree, the path looks confusing, but it has been historically formed... 

Fyi, we use it in general pods like Manila CSI operator in the example above (it is a deployment), and we can't set it empty because otherwise custom cert won't be used by the driver: https://github.com/openshift/csi-driver-manila-operator/blob/master/pkg/controllers/secret/secretsync.go#L149-L152

mandre:
> I think we should be setting this field to empty instead of this _default_ which based on it's contents is valid only for static pod resources on the node and not general pods trying to communicate with openstack api.

The other pods communicating with the openstack API generally consume the `cloud.conf` file that already implements the same logic: https://github.com/openshift/installer/blob/4aac205/pkg/asset/manifests/openstack/cloudproviderconfig.go#L20-L22

Perhaps the path is not the best and can be confusing for general pods but this is what we have now. We expect pods that need to talk to the openstack API to [mount the `/etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem` file](https://github.com/openshift/enhancements/blob/master/enhancements/security/openstack-self-signed-certificates.md#summary). Would be happy to change the well known path to the cloud provider cert if there is a better alternative.

Manila CSI is special in the sense that it consumes the `clouds.yaml` file and not `cloud.conf`. As Mike said, we can't set the value to empty otherwise Manila CSI wouldn't use the necessary CA cert to talk to openstack API.

mandre:
> regardless if you think we can come back and fix later, I would recommend you create a JIRA story for the shiftstack backlog, hopefully targeted for 4.7, and link here. At that point I'm more than happy to approve this change for the time being.

Hi @abhinavdahiya, I added https://issues.redhat.com/browse/OSASINFRA-2096 to our backlog to rework the openstack secrets in CCO based on your advice. Thanks.

mandre:
/retitle Bug 1884558: do not use local cacert path in generated clouds.yaml

@abhinavdahiya just to let you know there is no rush in getting this merged in 4.6 as it's not impacting functionality. The consumer of the secret (Manila CSI) already sets the cacert path to the expected value in 4.6
This patch should at least remove the local path from the `clouds.yaml` until we'll rework our secrets in 4.7.

abhinavdahiya:
> > regardless if you think we can come back and fix later, I would recommend you create a JIRA story for the shiftstack backlog, hopefully targeted for 4.7, and link here. At that point I'm more than happy to approve this change for the time being.
> 
> Hi @abhinavdahiya, I added https://issues.redhat.com/browse/OSASINFRA-2096 to our backlog to rework the openstack secrets in CCO based on your advice. Thanks.

/approve

PR #4231, opened 2020-10-01
Bug 1876815: OpenStack: unset OS_CLOUD

Fedosin:
We should unset OS_CLOUD env variable during cloudinfo and session generation, and cluster destruction. We have to do it because the real cloud name is defined by user in the install-config. OS_CLOUD has more priority, so the user-defined value will be ignored if OS_CLOUD contains something.

/label platform/openstack

### DIFF
--- a/pkg/asset/installconfig/openstack/session.go
+++ b/pkg/asset/installconfig/openstack/session.go
@@ -2,6 +2,7 @@
 package openstack
 
 import (
+	"os"
 	"sync"
 
 	"github.com/pkg/errors"
@@ -21,6 +22,12 @@ type Session struct {
 // GetSession returns an OpenStack session for a given cloud name in clouds.yaml.
 func GetSession(cloudName string) (*Session, error) {
 	opts := defaultClientOpts(cloudName)
+
+	// We should unset OS_CLOUD env variable here, because the real cloud name was
+	// defined on the previous step. OS_CLOUD has more priority, so the value from
+	// "opts" variable will be ignored if OS_CLOUD contains something.
+	os.Unsetenv("OS_CLOUD")
+
 	cloudConfig, err := clientconfig.GetCloudFromYAML(opts)
 	if err != nil {
 		return nil, err
--- a/pkg/asset/installconfig/openstack/validation/cloudinfo.go
+++ b/pkg/asset/installconfig/openstack/validation/cloudinfo.go
@@ -1,6 +1,7 @@
 package validation
 
 import (
+	"os"
 	"strings"
 
 	"github.com/gophercloud/gophercloud"
@@ -50,6 +51,11 @@ func GetCloudInfo(ic *types.InstallConfig) (*CloudInfo, error) {
 	}
 
 	opts := &clientconfig.ClientOpts{Cloud: ic.OpenStack.Cloud}
+
+	// We should unset OS_CLOUD env variable here, because the real cloud name was
+	// defined on the previous step. OS_CLOUD has more priority, so the value from
+	// "opts" variable will be ignored if OS_CLOUD contains something.
+	os.Unsetenv("OS_CLOUD")
 
 	ci.clients.networkClient, err = clientconfig.NewServiceClient("network", opts)
 	if err != nil {
--- a/pkg/destroy/openstack/openstack.go
+++ b/pkg/destroy/openstack/openstack.go
@@ -1,6 +1,7 @@
 package openstack
 
 import (
+	"os"
 	"strings"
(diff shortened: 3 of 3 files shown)


iamemilio:
I am hesitant to approve un-setting variables in the user's environment. Could we just make a point of this in the docs? 
I feel like we should seek abhinav's approval on this.

pkg/asset/installconfig/openstack/session.go
```diff
@@ -21,6 +22,12 @@ type Session struct {
 // GetSession returns an OpenStack session for a given cloud name in clouds.yaml.
 func GetSession(cloudName string) (*Session, error) {
 	opts := defaultClientOpts(cloudName)
+
+	// We should unset OS_CLOUD env variable here, because the real cloud name was
+	// defined on the previous step. OS_CLOUD has more priority, so the value from
+	// "opts" variable will be ignored if OS_CLOUD contains something.
+	os.Unsetenv("OS_CLOUD")
```
abhinavdahiya, line 29: what! why is the library taking env as priority over what's actually being provided to it explicitly?

Also I think you will have to remember to make this change everytime you use the library function, and there is high chance for forgetting this. Can you modify the invocation of the GetCloudFromYAML such that it either changes the order of priority or stop using the env.

unsetting the env is no really a good behavior, it is racey and can have unintended effect on the user environement.

Fedosin:
@iamemilio what do you mean by un-setting variables in the user's environment? I think we do not modify user's environment, but just change the installer process env.

pkg/asset/installconfig/openstack/session.go
Fedosin: The gophercloud code that confirms this behavior: https://github.com/openshift/installer/blob/aaff33ec09d42d527a424a723e144449382c6e9e/vendor/github.com/gophercloud/utils/openstack/clientconfig/requests.go#L717-L733

Fedosin: As you can notice, "cloud" overwriting happens before the invocation of `GetCloudFromYAML`, so its modification won't help here.

abhinavdahiya: But the function you linked above is not something you are actually invoking here.. [NewServiceClient](https://github.com/openshift/installer/blob/aaff33ec09d42d527a424a723e144449382c6e9e/vendor/github.com/gophercloud/utils/openstack/clientconfig/requests.go#L709) and you are directly calling GetCloudFromYAML next after the unset.

So why are we unsetting here again?

abhinavdahiya: looking at the function for region https://github.com/openshift/installer/blob/aaff33ec09d42d527a424a723e144449382c6e9e/vendor/github.com/gophercloud/utils/openstack/clientconfig/requests.go#L756-L772

here the precedence order is opt > env . which is same as endpoint but different from the cloud order?

Fedosin: GetSession is called in several places in the code (to generate tfvars, for instance), so I decided to unset the variable here to reduce duplication of os.Unsetenv("OS_CLOUD") later. 

Also please notice, that Terraform's OpenStack Provider also uses gophercloud library, and this behavior is maintained there as well.

abhinavdahiya: > Also please notice, that Terraform's OpenStack Provider also uses gophercloud library, and this behavior is maintained there as well.

can you link to code that unsets the env. also can we open an issue with gophercloud to fix their order, this currently is just odd imo.

Fedosin: I looked at the code, they've already created a separate auth implementation in the same repo: https://github.com/gophercloud/utils/tree/master/terraform/auth
So, Terraform is safe.

iamemilio:
> @iamemilio what do you mean by un-setting variables in the user's environment? I think we do not modify user's environment, but just change the installer process env.

That is what I meant by the user's environment, the environment that is running the installer. IMHO, I think we would be better off fixing this with documentation for the time being, and then opening an issue in gophercloud for a long run solution.

pierreprinetti:
/approve
/lgtm

I agree that Gophercloud's behaviour is surprising to say the least.

However, given Gophercloud's behaviour, I don't see any short-term alternative that would end up being any cleaner than this.

Also, even if Gophercloud accepted to change its order of preference, it would require a major version bump that we don't necessarily want to wait for.

Holding for a second LGTM, since there is no clear consensus yet.
/hold

PR #4237, opened 2020-10-02
Bug 1884435: vsphere - add delay if resolv.conf is not available; wait for dhcp

jcpowermac:
Randomly the resolver will not be configured with 127.0.0.1 causing
bootstrap failure, adding status `down` to see if it improves.


jcpowermac:
/bugzilla refresh

/test e2e-vsphere

I am unsure this will fix the random not applying the loopback to `/etc/resolv.conf`

jcpowermac:
I have tested this a couple times and haven't seen the failure. Not saying this 100% fixes it though

jcpowermac:
Changed to use:
https://github.com/openshift/machine-config-operator/pull/2030

mtnbikenc (approved):

PR #4238, opened 2020-10-02
Bug 1884691: types: allow manual cred mode for gcp and azure

staebler:
Allow the user to set `credentialsMode` to "Manual" when installing on GCP and Azure. The cloud-credential-operator has support for the manual mode on those platforms.


dgoodwin:
I tested this, QE tested it, and Docs documented it. This happened as our README was out of date in the CCO repo showing unknown for these platforms, as I had not tested them at this time.

Investigating if we can beat code freeze today. 
/cc @sdodson @abhinavdahiya

pkg/types/validation/installconfig_test.go
```diff
@@ -962,16 +962,6 @@ func TestValidateInstallConfig(t *testing.T) {
...
-		{
-			name: "unsupported manual cloud credentials mode",
-			installConfig: func() *types.InstallConfig {
-				c := validInstallConfig()
-				c.Platform = types.Platform{GCP: validGCPPlatform()}
-				c.CredentialsMode = types.ManualCredentialsMode
-				return c
-			}(),
-			expectedError: `^credentialsMode: Unsupported value: "Manual": supported values: "Mint", "Passthrough"$`,
-		},
```
dgoodwin, line 974: Could this test stay if changed to a platform that isn't AWS, Azure, or GCP

staebler: It could, but it is already covered by the next test case below it.

dgoodwin:
Please hold until first 4.6.z, I've discovered the QE test plans did not go through this exact method, rather the old ConfigMap, and thus we're not 100% sure this is working until the test plans are updated and run.

pkg/types/validation/installconfig_test.go
dgoodwin: Ack

abhinavdahiya:
/hold cancel

as we are now open for 4.7

PR #4242, opened 2020-10-05
log output: logrus doesn't preserve newlines, so explicitly render on…

deads2k:
… separate lines

https://github.com/sirupsen/logrus/issues/608 indicates this is intentional, but
human readable conditions often use newlines to separate output for humans.
Workaround the problem for clusteroperator output.  The overall error from the installer
would also benefit, so those lines are tagged with TODOs.

### DIFF
--- a/cmd/openshift-install/create.go
+++ b/cmd/openshift-install/create.go
@@ -131,6 +131,7 @@ var (
 					if err2 := logClusterOperatorConditions(ctx, config); err2 != nil {
 						logrus.Error("Attempted to gather ClusterOperator status after installation failure: ", err2)
 					}
+					// TODO this error would be better rendered with newlines intact instead of quoted.
 					logrus.Fatal(err)
 				}
 				timer.StopTimer(timer.TotalTimeElapsed)
--- a/cmd/openshift-install/gather.go
+++ b/cmd/openshift-install/gather.go
@@ -280,9 +280,21 @@ func logClusterOperatorConditions(ctx context.Context, config *rest.Config) erro
 				continue
 			}
 			if condition.Type == configv1.OperatorDegraded {
-				logrus.Errorf("Cluster operator %s %s is %s with %s: %s", operator.ObjectMeta.Name, condition.Type, condition.Status, condition.Reason, condition.Message)
+				// many operators have messages split by newlines. Per condition docs, these are human readable, so messages are what we will show:
+				// condition docs: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1408
+				// logrus doesn't render newlines, because of this we will split manually
+				// logrus no-newlines: https://github.com/sirupsen/logrus/issues/608
+				// Per condition docs, reasons are intended to be machine readable, not human readable.  We will not display reasons
+				// condition docs: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1397
+				messages := strings.Split(condition.Message, "\n")
+				for _, message := range messages{
+					logrus.Errorf("Cluster operator %s %s is %s: %s", operator.ObjectMeta.Name, condition.Type, condition.Status, message)
+				}
 			} else {
-				logrus.Infof("Cluster operator %s %s is %s with %s: %s", operator.ObjectMeta.Name, condition.Type, condition.Status, condition.Reason, condition.Message)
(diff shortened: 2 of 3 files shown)


deads2k:
oh that's much nicer: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/4242/pull-ci-openshift-installer-master-e2e-libvirt/1313115770103795712

abhinavdahiya:
When a user is running the installer in the terminal, it will automatically render the new line right? 

This seems only useful when the output is being directed to non-iteractive-terminal locations where logrus is switching to field based `level="debug message="...""` like here https://github.com/openshift/installer/blob/b2199d8ee9c6611476f59ce6deb08f96a8bd5bae/cmd/openshift-install/main.go#L97


Secondly, if we do decide to split on new line, why prefix the operator sections again for each new line instead of printing it as-is like intented?

abhinavdahiya:
Also, if possible can we re-use https://github.com/openshift/installer/tree/master/pkg/lineprinter to split the message and invoke per new line?

deads2k:
> When a user is running the installer in the terminal, it will automatically render the new line right?
> 
> This seems only useful when the output is being directed to non-iteractive-terminal locations where logrus is switching to field based `level="debug message="...""` like here

We (openshift) don't auto parse these today.  Which means when they go wrong, they should be optimized for humans to review, while leaving it possible for machine analysis if we which.



> Secondly, if we do decide to split on new line, why prefix the operator sections again for each new line instead of printing it as-is like intented?

It allows future machine analysis and later presentation to have the data available too.  Trying to satisfy both use cases.



> Also, if possible can we re-use https://github.com/openshift/installer/tree/master/pkg/lineprinter to split the message and invoke per new line?

I have no particular preference if it works.  Do you prefer it?

abhinavdahiya:
> > Also, if possible can we re-use https://github.com/openshift/installer/tree/master/pkg/lineprinter to split the message and invoke per new line?
> 
> I have no particular preference if it works. Do you prefer it?

I would prefer to re use that impl for consistency.

cmd/openshift-install/create.go
```diff
@@ -131,6 +131,7 @@ var (
 					if err2 := logClusterOperatorConditions(ctx, config); err2 != nil {
 						logrus.Error("Attempted to gather ClusterOperator status after installation failure: ", err2)
 					}
+					// TODO this error would be better rendered with newlines intact instead of quoted.
```
abhinavdahiya, line 134: This error is not quoted for interactive-terminal users. So this TODO is not accurate.

cmd/openshift-install/waitfor.go
```diff
@@ -88,6 +88,7 @@ func newWaitForInstallCompleteCmd() *cobra.Command {
 					logrus.Error("Attempted to gather ClusterOperator status after wait failure: ", err2)
 				}
 
+				// TODO this error would be better rendered with newlines intact instead of quoted.
```
abhinavdahiya, line 91: same as https://github.com/openshift/installer/pull/4242/files#r500472594

cmd/openshift-install/gather.go
```diff
@@ -280,9 +280,21 @@ func logClusterOperatorConditions(ctx context.Context, config *rest.Config) erro
...
 			}
 			if condition.Type == configv1.OperatorDegraded {
-				logrus.Errorf("Cluster operator %s %s is %s with %s: %s", operator.ObjectMeta.Name, condition.Type, condition.Status, condition.Reason, condition.Message)
+				// many operators have messages split by newlines. Per condition docs, these are human readable, so messages are what we will show:
+				// condition docs: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1408
+				// logrus doesn't render newlines, because of this we will split manually
+				// logrus no-newlines: https://github.com/sirupsen/logrus/issues/608
+				// Per condition docs, reasons are intended to be machine readable, not human readable.  We will not display reasons
+				// condition docs: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L1397
+				messages := strings.Split(condition.Message, "\n")
```
deads2k, line 289: like this?

```go
				errPrinter := &lineprinter.LinePrinter{Print: (&lineprinter.Trimmer{WrappedPrint: logrus.Error}).Print}
				defer errPrinter.Close()
				errPrinter.Write([]byte(condition.Message))
```

are you sure?

deads2k:
replaced by a better impl in https://github.com/openshift/installer/pull/4282

PR #4244, opened 2020-10-06
OpenStack: Rename lbFloatingIP to apiFloatingIP

Fedosin:
/label platform/openstack

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -998,6 +998,9 @@ spec:
               openstack:
                 description: OpenStack is the configuration used when installing on OpenStack.
                 properties:
+                  apiFloatingIP:
+                    description: APIFloatingIP is the IP address of an available floating IP in your OpenStack cluster to associate with the OpenShift API load balancer.
+                    type: string
                   apiVIP:
                     description: 'APIVIP is the static IP on the nodes subnet that the api port for openshift will be assigned Default: will be set to the 5 on the first entry in the machineNetwork CIDR'
                     type: string
@@ -1062,7 +1065,7 @@ spec:
                     description: 'IngressVIP is the static IP on the nodes subnet that the apps port for openshift will be assigned Default: will be set to the 7 on the first entry in the machineNewtwork CIDR'
                     type: string
                   lbFloatingIP:
-                    description: LbFloatingIP is the IP address of an available floating IP in your OpenStack cluster to associate with the OpenShift load balancer.
+                    description: 'LbFloatingIP is the IP address of an available floating IP in your OpenStack cluster to associate with the OpenShift load balancer. Deprecated: this value has been renamed to apiFloatingIP.'
                     type: string
                   machinesSubnet:
                     description: MachinesSubnet is the UUIDv4 of an openstack subnet. This subnet will be used by all nodes created by the installer. By setting this, the installer will no longer create a network and subnet. The subnet and network specified in MachinesSubnet will not be deleted or modified by the installer.
--- a/docs/user/openstack/customization.md
+++ b/docs/user/openstack/customization.md
@@ -23,8 +23,8 @@ Beyond the [platform-agnostic `install-config.yaml` properties](../customization
 * `computeFlavor` (required string): The OpenStack flavor to use for compute and control-plane machines.
     This is currently required, but has lower precedence than [the `type` property](#machine-pools) on [the `compute` and `controlPlane` machine-pools](../customization.md#platform-customization).
 * `externalDNS` (optional list of strings): The IP addresses of DNS servers to be used for the DNS resolution of all instances in the cluster
-* `externalNetwork` (optional string): Name of external network the installer will use to provide access to the cluster. If defined, a floating ip from this network will be created and associated with the bootstrap node to facilitate debugging and connection to the bootstrap node during installation. The lbFloatingIP property is a floating ip address selected from this network.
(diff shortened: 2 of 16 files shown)


mandre (commented):
/lgtm

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -998,6 +998,9 @@ spec:
               openstack:
                 description: OpenStack is the configuration used when installing on OpenStack.
                 properties:
+                  apiFloatingIP:
+                    description: APIFloatingIP is the IP address of an available floating IP in your OpenStack cluster to associate with the OpenShift API.
```
wking, line 1002: nit: `with the OpenShift API.` -> `with the OpenShift API load balancer.`?

docs/user/openstack/customization.md
```diff
@@ -71,7 +71,7 @@ platform:
     externalDNS:
       - "8.8.8.8"
       - "192.168.1.12"
-    lbFloatingIP: 128.0.0.1
+    apiFloatingIP: 128.0.0.1
```
wking, line 74: nit: re-alphabetize, now that you start with `a`, here and below?

pkg/types/conversion/installconfig.go
```diff
@@ -26,19 +27,21 @@ func ConvertInstallConfig(config *types.InstallConfig) error {
...
 	}
 
 	config.APIVersion = types.InstallConfigVersion
 	return nil
 }
 
-// ConvertNetworking upconverts deprecated fields in networking
-func ConvertNetworking(config *types.InstallConfig) {
+// convertNetworking upconverts deprecated fields in networking
+func convertNetworking(config *types.InstallConfig) {
```
wking, line 44: I dunno why this was public before, but making it private seems like an orthogonal change that should go in its own PR.

pkg/types/conversion/installconfig.go
```diff
@@ -79,11 +82,19 @@ func ConvertNetworking(config *types.InstallConfig) {
...
+func convertBaremetal(config *types.InstallConfig) {
 	if config.Platform.BareMetal.DeprecatedProvisioningDHCPExternal == true && config.Platform.BareMetal.ProvisioningNetwork == "" {
 		config.Platform.BareMetal.ProvisioningNetwork = baremetal.UnmanagedProvisioningNetwork
 	}
 }
+
+// convertOpenStack upconverts deprecated fields in the OpenStack platform.
+func convertOpenStack(config *types.InstallConfig) {
+	// LbFloatingIP has been renamed to APIFloatingIP
+	if config.Platform.OpenStack.DeprecatedLbFloatingIP != "" && config.Platform.OpenStack.APIFloatingIP == "" {
```
wking, line 97: Seems like it should be a fatal error to have both set.  Also, probably worth warn-level logging when you do shuffle the value over, so folks using the old value are more likely to hear about the deprecation and update their generators.

pkg/types/conversion/installconfig.go
```diff
@@ -79,11 +82,19 @@ func ConvertNetworking(config *types.InstallConfig) {
...
+		config.Platform.OpenStack.APIFloatingIP = config.Platform.OpenStack.DeprecatedLbFloatingIP
```
wking, line 98: I'd expect to clear the old property when you shuffle the value over.

pkg/types/conversion/installconfig.go
Fedosin: this will go in a separate PR for all deprecated values. we have a dedicated jira card for this.

data/data/install.openshift.io_installconfigs.yaml
Fedosin: done

pkg/types/conversion/installconfig.go
Fedosin: I think we should keep it to show a warning message later, during install config validation.

pkg/types/conversion/installconfig.go
Fedosin: okay, I'll propose another PR to fix this.

pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -106,13 +106,13 @@ func validatePlatformFlavor(p *openstack.Platform, ci *CloudInfo, fldPath *field
 }
 
 func validateFloatingIPs(p *openstack.Platform, ci *CloudInfo, fldPath *field.Path) (allErrs field.ErrorList) {
```
abhinavdahiya, line 108: can we add a validation that makes sure people don't set both with different values.

pkg/types/conversion/installconfig.go
abhinavdahiya: > I'd expect to clear the old property when you shuffle the value over.

Don't reset user set values, if a user set the deprecated value, we can use that to set the new values but do not modify the value set by the user.

abhinavdahiya:
> https://github.com/openshift/installer/pull/4244/files#r501164253

LGTM, only one change requested.

pkg/asset/installconfig/openstack/validation/platform.go
Fedosin: I added a check that returns an error if both values are set regardless of whether they are equal or not. I.e. "you cannot specify lbFloatingIP and apiFloatingIP in the install config together".

pierreprinetti (approved):

pierreprinetti:
/hold
Do you want to squash your commits?

Fedosin:
/hold cancel

nope :)

PR #4245, opened 2020-10-06
cmd/openshift-install/create: Use library-go's IngressURI helper

wking:
Using the helper from openshift/library-go#911, so we:

* No longer hard-code `https://`.  It's calculated based on the route's TLS config.
* No longer assume that the available ingress host will match the route's `spec.host`.
* No longer assume that the route is admitted.  Now we check the ingress conditions to confirm `Admitted=True`.

Generated with:

```console
$ emacs cmd/openshift-install/create.go # everything in the diff
$ go get github.com/openshift/library-go@f360b9835cc8815eccc26cb52dafe05d3cbb3e93
go: github.com/openshift/library-go f360b9835cc8815eccc26cb52dafe05d3cbb3e93 => v0.0.0-20201006230840-f360b9835cc8
go: downloading github.com/openshift/library-go v0.0.0-20201006230840-f360b9835cc8
$ go mod tidy
...
github.com/openshift/installer/cmd/openshift-install imports
github.com/openshift/client-go/config/clientset/versioned imports
k8s.io/client-go/discovery imports
github.com/googleapis/gnostic/OpenAPIv2: module github.com/googleapis/gnostic@latest found (v0.5.1), but does not contain package github.com/googleapis/gnostic/OpenAPIv2
$ emacs go.mod  # set gnostic replacement
$ go mod tidy
$ go mod vendor
$ git add -A go.* vendor
```

using:

```console
$ go version
go version go1.15.2 linux/amd64
```

The gnostic replacement avoids the casing change that landed in googleapis/gnostic@896953e6749863beec38e27029c804e88c3144b8.  We'll want to drop the replacement-pin once we bump k8s.io/client-go/discovery to a version that uses the lowercased package.

### DIFF
--- a/cmd/openshift-install/create.go
+++ b/cmd/openshift-install/create.go
@@ -440,32 +440,21 @@ func waitForConsole(ctx context.Context, config *rest.Config) (string, error) {
 	silenceRemaining := logDownsample
 	timer.StartTimer("Console")
 	wait.Until(func() {
-		consoleRoutes, err := rc.RouteV1().Routes(consoleNamespace).List(ctx, metav1.ListOptions{})
-		if err == nil && len(consoleRoutes.Items) > 0 {
-			for _, route := range consoleRoutes.Items {
-				logrus.Debugf("Route found in openshift-console namespace: %s", route.Name)
-				if route.Name == consoleRouteName {
-					if uri, _, err2 := routeapihelpers.IngressURI(&route, ""); err2 == nil {
-						url = uri.String()
-						logrus.Debug("OpenShift console route is admitted")
-						cancel()
-					} else {
-						err = err2
-					}
-					break
-				}
+		route, err := rc.RouteV1().Routes(consoleNamespace).Get(ctx, consoleRouteName, metav1.GetOptions{})
+		if err == nil {
+			logrus.Debugf("Route found in openshift-console namespace: %s", consoleRouteName)
+			if uri, _, err2 := routeapihelpers.IngressURI(route, ""); err2 == nil {
+				url = uri.String()
+				logrus.Debug("OpenShift console route is admitted")
+				cancel()
+			} else {
+				err = err2
 			}
 		}
 		if err != nil {
 			silenceRemaining--
 			if silenceRemaining == 0 {
 				logrus.Debugf("Still waiting for the console route: %v", err)
-				silenceRemaining = logDownsample
-			}
-		} else if len(consoleRoutes.Items) == 0 {
-			silenceRemaining--
-			if silenceRemaining == 0 {
-				logrus.Debug("Still waiting for the console route...")
 				silenceRemaining = logDownsample
 			}
 		}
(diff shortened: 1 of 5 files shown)


wking:
/retest

Prow was temporarily out of GitHub tokens, but should have them back now.

vendor/modules.txt
```diff
@@ -2048,9 +2051,11 @@ k8s.io/client-go/util/retry
 # k8s.io/klog v1.0.0
 ## explicit
 k8s.io/klog
+# k8s.io/klog/v2 v2.3.0
+k8s.io/klog/v2
```
abhinavdahiya, line 2055: We currently try to silent any klog outputs because the clients see https://github.com/openshift/installer/blob/88a611e858a414910e69edeaacdfa695569ce7f9/cmd/openshift-install/main.go#L27-L32

Adding klog/v2 could be reverting that change. We need to make sure no log output ever comes out of klog.

cmd/openshift-install/create.go
```diff
@@ -444,12 +444,18 @@ func waitForConsole(ctx context.Context, config *rest.Config) (string, error) {
...
+						err = err2
+					}
+					break
 				}
 			}
-			logrus.Debug("OpenShift console route is created")
-			cancel()
-		} else if err != nil {
+		}
+		if err != nil {
```
abhinavdahiya, line 458: any specific reason to make this change from on conditions block to 2 different ones?

wking: so that the existing `err` handling code can catch both the "list failed" and "route not admitted" failure modes.

vendor/modules.txt
wking: makes sense.  I'm happy to idle this PR until the klog/v2 pivot happens, since that seems like an orthogonal change that the installer should pick up in a targeted PR.

abhinavdahiya:
@wking i think you can now rebase as https://github.com/openshift/installer/pull/4266 bumped the library-go to the latest version

wking:
Rebased around s #4266 with 6ea5e9cbc -> ad3f7cda8

vendor/modules.txt
wking: https://github.com/openshift/installer/pull/4245#issuecomment-714814915

cmd/openshift-install/create.go
```diff
@@ -445,12 +445,18 @@ func waitForConsole(ctx context.Context, config *rest.Config) (string, error) {
 			for _, route := range consoleRoutes.Items {
```
abhinavdahiya, line 445: qq: so we know the route's name today, but still we use List and range over the items to find out route resource, that seems like an overkill. @wking wdyt changing this to Get + check for not found ??

wking: Done with ad3f7cda8 -> eb343b19d.

staebler (approved):
/lgtm

PR #4248, opened 2020-10-07
Bug 1882191: Add GODEBUG=x509ignoreCN=0 to systemd DefaultEnvironment

sdodson:
Go 1.15 no longer falls back to the common name for hostname validation
requiring that a DNS SAN exist. We need more time to deprecate this so
revert back to previous behavior for the time being.

### DIFF
--- a/data/data/bootstrap/files/etc/systemd/system.conf.d/10-default-env-godebug.conf.template
+++ b/data/data/bootstrap/files/etc/systemd/system.conf.d/10-default-env-godebug.conf.template
@@ -1,0 +1,2 @@
+[Manager]
+DefaultEnvironment=GODEBUG=x509ignoreCN=0


sdodson:
Hmm, this works for systemd initiated processes but doesn't cover anything that's part of an interactive shell, is that another item we need to consider here or will everything we care about be managed by systemd?

abhinavdahiya:
> Hmm, this works for systemd initiated processes but doesn't cover anything that's part of an interactive shell, is that another item we need to consider here or will everything we care about be managed by systemd?

I think this uses the same mechanics as the proxy setup for bootstrap host, https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/etc/systemd/system.conf.d/10-default-env.conf.template which affects all the clients that would be pulling content from outside entities. So this change should be good enough.

PR #4254, opened 2020-10-08
Bug 1874248: types/vsphere/validation: ensure vcenter is all lower case

abhinavdahiya:
the vCenter hostname must be all lower case because otherwise the cluster components cannot accurately connect with vCenter.


mtnbikenc (approved):
/lgtm

PR #4257, opened 2020-10-09
Bug 1882649: enable image import again

Fedosin:
In 4.4 we disabled image import mechanism because it was not stable enough on the testing environment. Now we expect that it works better, so we can enable image import uploading by default.
But, to prevent possible failures, the installer falls back to legacy uploading if image import fails to upload data.

### DIFF
--- a/pkg/tfvars/openstack/openstack.go
+++ b/pkg/tfvars/openstack/openstack.go
@@ -104,10 +104,26 @@ func TFVars(masterConfigs []*v1alpha1.OpenstackProviderSpec, cloud string, exter
 			return nil, errors.Errorf("Unsupported URL scheme: '%v'", url.Scheme)
 		}
 
-		err = uploadBaseImage(cloud, localFilePath, imageName, infraID)
+		useImageImport, err := isImageImportSupported(cloud)
 		if err != nil {
 			return nil, err
 		}
+
+		err = uploadBaseImage(cloud, localFilePath, imageName, infraID, useImageImport)
+		if err != nil {
+			// If we tried to upload the data with the image import mechanism and failed,
+			// let's disable it and try legacy uploading.
+			if useImageImport {
+
+				err = uploadBaseImage(cloud, localFilePath, imageName, infraID, false)
+				if err != nil {
+					return nil, err
+				}
+			} else {
+				return nil, err
+			}
+		}
+
 	} else {
 		// Not a URL -> use baseImage value as an overridden Glance image name.
 		// Need to check if this image exists and there are no other images with this name.
--- a/pkg/tfvars/openstack/rhcos_image.go
+++ b/pkg/tfvars/openstack/rhcos_image.go
@@ -15,7 +15,7 @@ import (
 )
 
 // uploadBaseImage creates a new image in Glance and uploads the RHCOS image there
-func uploadBaseImage(cloud string, localFilePath string, imageName string, clusterID string) error {
+func uploadBaseImage(cloud string, localFilePath string, imageName string, clusterID string, useImageImport bool) error {
 	logrus.Debugln("Creating a Glance image for RHCOS...")
 
 	f, err := os.Open(localFilePath)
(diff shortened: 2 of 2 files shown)


pkg/tfvars/openstack/openstack.go
```diff
@@ -104,10 +104,26 @@ func TFVars(masterConfigs []*v1alpha1.OpenstackProviderSpec, cloud string, exter
...
-		err = uploadBaseImage(cloud, localFilePath, imageName, infraID)
+		useImageImport, err := isImageImportSupported(cloud)
 		if err != nil {
 			return nil, err
 		}
+
+		err = uploadBaseImage(cloud, localFilePath, imageName, infraID, useImageImport)
+		if err != nil {
+			// If we tried to upload the data with the image import mechanism and failed,
+			// let's disable it and try legacy uploading.
```
adduarte, line 115: remove comment, its clear from code

pkg/tfvars/openstack/openstack.go
```diff
@@ -104,10 +104,26 @@ func TFVars(masterConfigs []*v1alpha1.OpenstackProviderSpec, cloud string, exter
...
+			if useImageImport {
+
+				err = uploadBaseImage(cloud, localFilePath, imageName, infraID, false)
+				if err != nil {
+					return nil, err
+				}
+			} else {
+				return nil, err
```
adduarte, line 123: we loose the first error. i.e. if the first time it fails but it was due to something not related to useImageImport=True, then we'll loose the first error. It could make it hard to debug. any way to annotate the first error the second. Or at least log it?

pkg/tfvars/openstack/rhcos_image.go
```diff
@@ -110,6 +73,52 @@ func uploadBaseImage(cloud string, localFilePath string, imageName string, clust
 	return nil
 }
 
+func importImage(imageName, imageID string, conn *gophercloud.ServiceClient, f *os.File) error {
+	logrus.Debugf("Using Image Import API to upload RHCOS to the image %q with ID %q", imageName, imageID)
```
adduarte, line 77: nit, no need to say RHCOS , this is a function to upload an image. In the future it could be used by something else besides trying to upload and RHCOS image.

pkg/tfvars/openstack/rhcos_image.go
```diff
@@ -110,6 +73,52 @@ func uploadBaseImage(cloud string, localFilePath string, imageName string, clust
...
+	stageRes := imagedata.Stage(conn, imageID, f)
+	if stageRes.Err != nil {
+		logrus.Debugf("Image uploading failed")
+		return stageRes.Err
+	}
+	logrus.Debugf("The data was uploaded")
```
adduarte, line 83: I think you mean, "image uploaded"  or what is data? 
Just take out the log message an only report a failure I think.

pkg/tfvars/openstack/rhcos_image.go
```diff
@@ -110,6 +73,52 @@ func uploadBaseImage(cloud string, localFilePath string, imageName string, clust
...
+	}
+	importRes := imageimport.Create(conn, imageID, co)
+	if importRes.Err != nil {
+		return importRes.Err
+	}
+	logrus.Debugf("Image import started")
+
+	// Image import is an asynchronous operation, so we have to wait until the image becomes "active"
+	const numRetries = 5000
+	const timeSleepSeconds = 15
```
adduarte, line 97: It is never a good sign when you have to define a "sleep" variable. 
Any other way to do this without it.

pkg/tfvars/openstack/rhcos_image.go
```diff
@@ -110,6 +73,52 @@ func uploadBaseImage(cloud string, localFilePath string, imageName string, clust
 	return nil
 }
 
+func importImage(imageName, imageID string, conn *gophercloud.ServiceClient, f *os.File) error {
```
adduarte, line 76: "imageName" is not used for anything except writing debug messages. 
It would be best to get rid of it all together. Since this function is a generic importImage just displaying the imaged ID in debug messages is good enough because the calling function will probably log what image it s trying to download. Even better just return error !=nill if a failure.

adduarte (commented):
looks good,  remove the debug logs to make functions more generic. In a production environment is not really necessary to add all the debug messages. The "err" returned will probably have all necessary info.

PR #4258, opened 2020-10-09
pkg/asset/installconfig/aws: Validate install-config instance type

mtnbikenc:
Ensure the user provided instance type meets minimum resource requirements.

Example error:
```
FATAL failed to fetch Metadata: failed to load asset "Install Config": [controlPlane.platform.aws.type: Invalid value: "m5.large": Instance type m5.large does not meet minimum controlPlane resource requirements of 4 vCPUs, controlPlane.platform.aws.type: Invalid value: "m5.large": Instance type m5.large does not meet minimum controlPlane resource requirements of 16384 MiB Memory]
```

### DIFF
--- a/pkg/asset/installconfig/aws/instancetypes.go
+++ b/pkg/asset/installconfig/aws/instancetypes.go
@@ -1,0 +1,38 @@
+package aws
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/pkg/errors"
+)
+
+// InstanceType holds metadata for an instance type.
+type InstanceType struct {
+	DefaultVCpus int64
+	MemInMiB     int64
+}
+
+// instanceTypes retrieves a list of instance types for the given region.
+func instanceTypes(ctx context.Context, session *session.Session, region string) (map[string]InstanceType, error) {
+	types := map[string]InstanceType{}
+
+	client := ec2.New(session, aws.NewConfig().WithRegion(region))
+	if err := client.DescribeInstanceTypesPagesWithContext(ctx,
+		&ec2.DescribeInstanceTypesInput{},
+		func(page *ec2.DescribeInstanceTypesOutput, lastPage bool) bool {
+			for _, info := range page.InstanceTypes {
+				types[*info.InstanceType] = InstanceType{
+					DefaultVCpus: aws.Int64Value(info.VCpuInfo.DefaultVCpus),
+					MemInMiB:     aws.Int64Value(info.MemoryInfo.SizeInMiB),
+				}
+			}
+			return !lastPage
+		}); err != nil {
+		return nil, errors.Wrap(err, "fetching instance types")
+	}
+
+	return types, nil
+}
--- a/pkg/asset/installconfig/aws/validation.go
+++ b/pkg/asset/installconfig/aws/validation.go
@@ -16,6 +16,21 @@ import (
 	awstypes "github.com/openshift/installer/pkg/types/aws"
 )
 
+type resourceRequirements struct {
+	minimumVCpus  int64
+	minimumMemory int64
+}
+
+var controlPlaneReq = resourceRequirements{
+	minimumVCpus:  4,
+	minimumMemory: 16384,
+}
+
+var computeReq = resourceRequirements{
+	minimumVCpus:  2,
+	minimumMemory: 8192,
+}
+
 // Validate executes platform-specific validation.
 func Validate(ctx context.Context, meta *Metadata, config *types.InstallConfig) error {
 	allErrs := field.ErrorList{}
@@ -26,12 +41,12 @@ func Validate(ctx context.Context, meta *Metadata, config *types.InstallConfig) 
 	allErrs = append(allErrs, validatePlatform(ctx, meta, field.NewPath("platform", "aws"), config.Platform.AWS, config.Networking, config.Publish)...)
 
 	if config.ControlPlane != nil && config.ControlPlane.Platform.AWS != nil {
-		allErrs = append(allErrs, validateMachinePool(ctx, meta, field.NewPath("controlPlane", "platform", "aws"), config.Platform.AWS, config.ControlPlane.Platform.AWS)...)
+		allErrs = append(allErrs, validateMachinePool(ctx, meta, field.NewPath("controlPlane", "platform", "aws"), config.Platform.AWS, config.ControlPlane.Platform.AWS, controlPlaneReq)...)
 	}
 	for idx, compute := range config.Compute {
 		fldPath := field.NewPath("compute").Index(idx)
 		if compute.Platform.AWS != nil {
-			allErrs = append(allErrs, validateMachinePool(ctx, meta, fldPath.Child("platform", "aws"), config.Platform.AWS, compute.Platform.AWS)...)
+			allErrs = append(allErrs, validateMachinePool(ctx, meta, fldPath.Child("platform", "aws"), config.Platform.AWS, compute.Platform.AWS, computeReq)...)
 		}
 	}
 	return allErrs.ToAggregate()
@@ -51,7 +66,7 @@ func validatePlatform(ctx context.Context, meta *Metadata, fldPath *field.Path, 
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("serviceEndpoints"), platform.ServiceEndpoints, err.Error()))
 	}
 	if platform.DefaultMachinePlatform != nil {
-		allErrs = append(allErrs, validateMachinePool(ctx, meta, fldPath.Child("defaultMachinePlatform"), platform, platform.DefaultMachinePlatform)...)
+		allErrs = append(allErrs, validateMachinePool(ctx, meta, fldPath.Child("defaultMachinePlatform"), platform, platform.DefaultMachinePlatform, controlPlaneReq)...)
 	}
 	return allErrs
 }
@@ -104,7 +119,7 @@ func validateSubnets(ctx context.Context, meta *Metadata, fldPath *field.Path, s
 	return allErrs
 }
 
-func validateMachinePool(ctx context.Context, meta *Metadata, fldPath *field.Path, platform *awstypes.Platform, pool *awstypes.MachinePool) field.ErrorList {
+func validateMachinePool(ctx context.Context, meta *Metadata, fldPath *field.Path, platform *awstypes.Platform, pool *awstypes.MachinePool, req resourceRequirements) field.ErrorList {
 	allErrs := field.ErrorList{}
 	if len(pool.Zones) > 0 {
 		availableZones := sets.String{}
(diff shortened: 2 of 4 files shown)


pkg/asset/installconfig/aws/instancetypes.go
```diff
@@ -0,0 +1,60 @@
...
+	minimumComputeVCpus       int64 = 2
+	minimumComputeMemory      int64 = 8192
+)
+
+// FooValidateInstanceType returns information of the all the instance types available for a region.
+// It returns a map of instance type name to it's information.
+func validateInstanceType(ctx context.Context, sess *session.Session, region string, InstanceType string, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	spew.Dump(sess)
```
mtnbikenc, line 27: (*session.Session)(<nil>)

mtnbikenc:
https://issues.redhat.com/browse/CORS-1547

pkg/asset/installconfig/aws/validation.go
```diff
@@ -129,6 +129,10 @@ func validateMachinePool(ctx context.Context, meta *Metadata, fldPath *field.Pat
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("zones"), pool.Zones, errMsg))
 		}
 	}
+	if pool.InstanceType != "" {
+		session, _ := meta.Session(ctx)
```
abhinavdahiya, line 133: So currently this validation package uses [`meta *Metadata`](https://github.com/openshift/installer/pull/4258/files#diff-784272713154e11b21c479afd5c0ad47R107) to get information from the AWS API..

see https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/aws/metadata.go#L19-L20 being used in validation here https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/aws/validation.go#L75

We can add a private function to load the instance types on the metadata maybe all for the time being, and then we can use those to also unit tests the validation code.

I think you can use the public / private subnet handling for example.

pkg/asset/installconfig/aws/instancetypes.go
```diff
@@ -0,0 +1,60 @@
...
+	// Minimum resource requirements
+	minimumControlPlaneVCpus  int64 = 4
+	minimumControlPlaneMemory int64 = 16384
+	minimumComputeVCpus       int64 = 2
+	minimumComputeMemory      int64 = 8192
+)
+
+// validateInstanceType validates a provided instance type meets minimum resource requirements.
+// It returns any errors from validation.
+func validateInstanceType(ctx context.Context, sess *session.Session, region string, InstanceType string, fldPath *field.Path) field.ErrorList {
```
abhinavdahiya, line 24:
```suggestion
func validateInstanceType(ctx context.Context, sess *session.Session, region string, instanceType string, fldPath *field.Path) field.ErrorList {
```

pkg/asset/installconfig/aws/instancetypes.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return append(allErrs, field.InternalError(fldPath, err))
+	}
+
+	// need to check if this is for compute or controlplane
+	poolType := fldPath.Root().String()
+	switch poolType {
+	case "compute":
+		if *resp.InstanceTypes[0].VCpuInfo.DefaultVCpus < minimumComputeVCpus {
+			errMsg := fmt.Sprintf("Instance type %s does not meet minimum %s resource requirements of %d vCPUs", InstanceType, poolType, minimumComputeVCpus)
```
abhinavdahiya, line 39: In Go I think we prefer the error messages to not start with capital letters as it doesn't read well when wrapped.

pkg/asset/installconfig/aws/validation.go
mtnbikenc: I had considered using metadata but I wasn't sure if extending the metadata was a good idea.  I'll rework this to use a function to query all instance types then do the validation.

mtnbikenc: Updated.

pkg/asset/installconfig/aws/instancetypes.go
```diff
@@ -0,0 +1,60 @@
...
+	client := ec2.New(sess, aws.NewConfig().WithRegion(region))
+	resp, err := client.DescribeInstanceTypesWithContext(ctx, &ec2.DescribeInstanceTypesInput{InstanceTypes: []*string{&InstanceType}})
+	if err != nil {
+		return append(allErrs, field.InternalError(fldPath, err))
+	}
+
+	// need to check if this is for compute or controlplane
+	poolType := fldPath.Root().String()
+	switch poolType {
+	case "compute":
```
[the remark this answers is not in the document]
mtnbikenc: Updated to pass in the constraint.

pkg/asset/installconfig/aws/instancetypes.go
mtnbikenc: Updated.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -51,7 +66,7 @@ func validatePlatform(ctx context.Context, meta *Metadata, fldPath *field.Path,
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("serviceEndpoints"), platform.ServiceEndpoints, err.Error()))
 	}
 	if platform.DefaultMachinePlatform != nil {
-		allErrs = append(allErrs, validateMachinePool(ctx, meta, fldPath.Child("defaultMachinePlatform"), platform, platform.DefaultMachinePlatform)...)
+		allErrs = append(allErrs, validateMachinePool(ctx, meta, fldPath.Child("defaultMachinePlatform"), platform, platform.DefaultMachinePlatform, computeReq)...)
```
mtnbikenc, line 69: For `DefaultMachinePlatform` I was thinking this was compute machine pools.

pkg/asset/installconfig/aws/instancetypes.go
```diff
@@ -0,0 +1,60 @@
...
+
+const (
+	// Minimum resource requirements
+	minimumControlPlaneVCpus  int64 = 4
+	minimumControlPlaneMemory int64 = 16384
+	minimumComputeVCpus       int64 = 2
+	minimumComputeMemory      int64 = 8192
+)
+
+// validateInstanceType validates a provided instance type meets minimum resource requirements.
```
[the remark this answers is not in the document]
mtnbikenc: Added unit tests.

pkg/asset/installconfig/aws/validation.go
abhinavdahiya: so currently the installer uses 3 places to find the instance size,
- internal default for the region
- platform.defaultMachinePool
- corresponding machinePool

So if default machine pool was set to compute requirements and no control plane machinepool was set, we will allow the user to use the instance type only sufficient for compute.

I think setting it to control plane requirements (which is larger) means that it will satisfy our requirement more accurately.

pkg/asset/installconfig/aws/validation_test.go
```diff
@@ -168,6 +185,41 @@ func TestValidate(t *testing.T) {
 		}(),
 		availZones:     validAvailZones(),
 		privateSubnets: validPrivateSubnets(),
+	}, {
+		name: "valid instance types",
+		installConfig: func() *types.InstallConfig {
+			c := validInstallConfig()
+			c.Platform.AWS = &aws.Platform{Region: "us-east-1"}
```
abhinavdahiya, line 192: We already set the region in validInstallConfig to this value https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/aws/validation_test.go#L28 so I don;t think we need to set it again..

pkg/asset/installconfig/aws/validation_test.go
```diff
@@ -168,6 +185,41 @@ func TestValidate(t *testing.T) {
...
+			c.ControlPlane.Platform.AWS.InstanceType = "t2.small"
+			c.Compute[0].Platform.AWS.InstanceType = "m5.large"
+			return c
+		}(),
+		availZones:    validAvailZones(),
+		instanceTypes: validInstanceTypes(),
+		expectErr:     `^\[controlPlane\.platform\.aws\.type: Invalid value: \"t2.small\": instance type does not meet minimum resource requirements of 4 vCPUs, controlPlane\.platform\.aws\.type: Invalid value: \"t2.small\": instance type does not meet minimum resource requirements of 16384 MiB Memory\]$`,
```
abhinavdahiya, line 210: I think you can use https://stackoverflow.com/a/30776921 to increase the readability of the regex match here.

pkg/asset/installconfig/aws/validation_test.go
```diff
@@ -168,6 +185,41 @@ func TestValidate(t *testing.T) {
...
+		installConfig: func() *types.InstallConfig {
+			c := validInstallConfig()
+			c.Platform.AWS = &aws.Platform{Region: "us-east-1"}
+			c.ControlPlane.Platform.AWS.InstanceType = "m5.xlarge"
+			c.Compute[0].Platform.AWS.InstanceType = "t2.small"
+			return c
+		}(),
+		availZones:    validAvailZones(),
+		instanceTypes: validInstanceTypes(),
+		expectErr:     `^\[compute\[0\]\.platform\.aws\.type: Invalid value: \"t2.small\": instance type does not meet minimum resource requirements of 2 vCPUs, compute\[0\]\.platform\.aws\.type: Invalid value: \"t2.small\": instance type does not meet minimum resource requirements of 8192 MiB Memory\]$`,
```
abhinavdahiya, line 222: would like to see one testcase for defaultMachinePool too.

pkg/asset/installconfig/aws/validation_test.go
mtnbikenc: I either had to set this to override the entire aws section from validInstallConfig or set set `c.Platform.AWS.Subnets = nil` because it would fail on the predefined subnets in validInstallConfig.  Should I be doing it a different way?

pkg/asset/installconfig/aws/validation_test.go
mtnbikenc: That works well.  Should I update the rest of the err strings with the same pattern?

abhinavdahiya: if you like in a separate commit, but i wouldn't require that in this PR.

pkg/asset/installconfig/aws/instancetypes.go
```diff
@@ -0,0 +1,34 @@
...
+// InstanceType holds metadata for an instance type.
+type InstanceType struct {
+	DefaultVCpus int64
+	MemInMiB     int64
+}
+
+// instanceTypes retrieves a list of instance types for the given region.
+func instanceTypes(ctx context.Context, session *session.Session, region string) (map[string]InstanceType, error) {
+	client := ec2.New(session, aws.NewConfig().WithRegion(region))
+	resp, err := client.DescribeInstanceTypesWithContext(ctx, &ec2.DescribeInstanceTypesInput{})
```
abhinavdahiya, line 21: this came to light from a slack conversation. we should be using https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.DescribeInstanceTypesPagesWithContext  to make sure we list all the instance types.

pkg/asset/installconfig/aws/instancetypes.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "fetching instance types")
+	}
+
+	types := make(map[string]InstanceType, len(resp.InstanceTypes))
+	for _, item := range resp.InstanceTypes {
+		types[*item.InstanceType] = InstanceType{
+			DefaultVCpus: *item.VCpuInfo.DefaultVCpus,
```
abhinavdahiya, line 29:
```suggestion
			DefaultVCpus: https://docs.aws.amazon.com/sdk-for-go/api/aws/#Int64Value(*item.VCpuInfo.DefaultVCpus),
```

mtnbikenc: Could you explain what this does?

pkg/asset/installconfig/aws/validation_test.go
```diff
@@ -168,6 +185,53 @@ func TestValidate(t *testing.T) {
 		}(),
 		availZones:     validAvailZones(),
 		privateSubnets: validPrivateSubnets(),
+	}, {
+		name: "valid instance types",
+		installConfig: func() *types.InstallConfig {
+			c := validInstallConfig()
+			c.Platform.AWS = &aws.Platform{Region: "us-east-1"}
+			c.Platform.AWS.DefaultMachinePlatform.InstanceType = "m5.xlarge"
```
mtnbikenc, line 193: @abhinavdahiya  This is causing a panic in unit tests and I can't figure out what I'm doing wrong.

pkg/asset/installconfig/aws/instancetypes.go
abhinavdahiya:
```
func Int64Value
func Int64Value(v *int64) int64

Int64Value returns the value of the int64 pointer passed in or 0 if the pointer is nil. 
```

since you are referencing a pointer int64, this function makes sure it checks for nil and sets that to 0 instead of panic

pkg/asset/installconfig/aws/validation_test.go
mtnbikenc: Pushed an update that works, but I want to understand why.

mtnbikenc: I think i figured it out.  For compute and controlplane, &aws.MachinePool was already instantiated in valdiInstallConfig.  It is not the case for DefaultmachinePlatform, so direct assignment of InstanceType was failing.

pkg/asset/installconfig/aws/instancetypes.go
mtnbikenc: Converted to use Pages.

pkg/asset/installconfig/aws/validation_test.go
mtnbikenc: Added defaultMachinePool to the valid instance types test.

pkg/asset/installconfig/aws/validation.go
mtnbikenc: Updated to compare defaultMachinePool against control plane requirements.

PR #4259, opened 2020-10-11
install-complete: provide troubleshooting info when operators fail

jstuever:
This change provides a link to the documentation when operators are
reporting failures during install-complete. It also gives instructions
to reinvoke with wait-for install-complete once all issues have been
resolved.

### DIFF
--- a/cmd/openshift-install/create.go
+++ b/cmd/openshift-install/create.go
@@ -131,6 +131,7 @@ var (
 					if err2 := logClusterOperatorConditions(ctx, config); err2 != nil {
 						logrus.Error("Attempted to gather ClusterOperator status after installation failure: ", err2)
 					}
+					logTroubleshootingLink()
 					logrus.Fatal(err)
 				}
 				timer.StopTimer(timer.TotalTimeElapsed)
@@ -508,4 +509,11 @@ func waitForInstallComplete(ctx context.Context, config *rest.Config, directory 
 	}
 
 	return logComplete(rootOpts.dir, consoleURL)
+}
+
+func logTroubleshootingLink() {
+	logrus.Error("Cluster initialization failed because one or more operators are not functioning properly.")
+	logrus.Error("The cluster should be accessible for troublsehooting as detailed in the documentation linked below.")
+	logrus.Error("The 'wait-for install-complete' subcommand can then be used to continue the installation.")
+	logrus.Error("https://docs.openshift.com/container-platform/latest/support/troubleshooting/troubleshooting-installations.html")
 }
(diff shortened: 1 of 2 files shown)


jstuever:
Looks like this:

```
ERROR Cluster initialization failed because one or more operators are not functioning properly. The cluster should be accessible for troub
lsehooting as detailed in the following documentation. The 'wait-for install-complete' subcommand can then be used to continue the installation.

https://docs.openshift.com/container-platform/latest/support/troubleshooting/troubleshooting-installations.html
  
FATAL failed to initialize the cluster: Some cluster operators are still updating: authentication, console, image-registry, ingress, kube-storage-
version-migrator, monitoring 

```

cmd/openshift-install/gather.go
```diff
@@ -270,22 +270,34 @@ func logClusterOperatorConditions(ctx context.Context, config *rest.Config) erro
...
+				errorCount++
 			}
 		}
 	}
+	if(errorCount > 0){
+		logrus.Error(
+			"Cluster initialization failed because one or more operators are not functioning properly." +
+			" The cluster should be accessible for troublsehooting as detailed in the following documentation." +
+			" The 'wait-for install-complete' subcommand can then be used to continue the installation.\n\n" +
+			"https://docs.openshift.com/container-platform/latest/support/troubleshooting/troubleshooting-installations.html\n\n")
```
abhinavdahiya, line 299: I think we should just put this error [here](https://github.com/openshift/installer/blob/b894e4c734366194b4773ca4bbcf90a249c2a7e4/cmd/openshift-install/create.go#L408) instead. When install-complete fails due to timeout this error response is valid for almost all cases.

abhinavdahiya: secondly, adding `\n` at the end will force the output to look like

```
ERROR The 'wait-for install-complete' subcommand can then be used to continue the installation.

https://docs.openshift.com/container-platform/latest/support/troubleshooting/troubleshooting-installations.html

```

when it should look like
```
ERROR The 'wait-for install-complete' subcommand can then be used to continue the installation.
ERROR https://docs.openshift.com/container-platform/latest/support/troubleshooting/troubleshooting-installations.html
```

which keeps the UX consistent and also looks better when there are other INFO/ERROR lines around. it.
since logrus doesn't automatically do this https://github.com/sirupsen/logrus/issues/608 let's just call logrus.Error for each newline you want.

jstuever:
Latest modifications look like this....
```
ERROR Cluster initialization failed because one or more operators are not functioning properly. 
ERROR The cluster should be accessible for troublsehooting as detailed in the documentation linked below. 
ERROR The 'wait-for install-complete' subcommand can then be used to continue the installation. 
ERROR https://docs.openshift.com/container-platform/latest/support/troubleshooting/troubleshooting-installations.html 
FATAL failed to initialize the cluster: Some cluster operators are still updating: authentication, console, image-registry, ingress, kube-storage-version-migrator, monitoring 
```

PR #4262, opened 2020-10-12
Bug 1887525: baremetal: Wait for master-bmh-update script to succeed

zaneb:
Allow users to debug failures in the master-bmh-update script.

We are not done with the bootstrap host until the `master-bmh-update.sh`
script has successfully completed, adding the IP address data needed by
the master nodes, and enabling reconciliation on the master Host
objects. To ensure the bootstrap cannot go away before this is done,
delay signalling that bootstrapping is complete until it is finished.

Correct the name of the unit in the log gathering script, so that users
can see the output without logging in to the bootstrap host.

### DIFF
--- a/data/data/bootstrap/baremetal/systemd/units/master-bmh-update.service
+++ b/data/data/bootstrap/baremetal/systemd/units/master-bmh-update.service
@@ -2,6 +2,7 @@
 Description=Update master BareMetalHosts with introspection data
 Wants=bootkube.service
 After=release-image.service
+Before=progress.service
 
 [Service]
 Type=oneshot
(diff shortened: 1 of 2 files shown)


data/data/bootstrap/baremetal/systemd/units/master-bmh-update.service
```diff
@@ -2,6 +2,7 @@
 Description=Update master BareMetalHosts with introspection data
 Wants=bootkube.service
 After=release-image.service
+Before=progress.service
```
abhinavdahiya, line 5: progress service is supposed to be our one place that defines `done`, adding this to this service would make figuring that out a little more complex.

If we have a better way to define this in a more concise way, i would love that. but for now i wouldn't block this change.

zaneb: Thanks. Theoretically we could put `After=master-bmh-update.service` in the progess service instead - systemd treats ordering completely separately from dependencies, so it wouldn't care on platforms where this service doesn't exist. I didn't go with that initially because it seemed like it might cause more confusion on non-baremetal platforms than it solves on baremetal :) Completely open to switching it over if you prefer that though.

PR #4265, opened 2020-10-12
Bug 1882844: vsphere destroy: delete tag category created by installer

jstuever:
Prior to this change, the installer created a tag cagetory in vsphere
but did not delete it on destroy. This change updates destroy to delete
the category after deleting the tags.


patrickdillon:
This is causing problems for QE in earlier versions
/cherry-pick release-4.6
/cherry-pick release-4.5

PR #4266, opened 2020-10-13
Bump K8s.io version to v0.19.0

nirarg:


.gitignore
```diff
@@ -1 +1,3 @@
 /bin/
+*.swp
+*dev.log
```
abhinavdahiya, line 3: I don't think the installer repo needs to add user-specific files to git ignore.

cmd/openshift-install/main.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh/terminal"
-	"k8s.io/klog"
+	klogv2 "k8s.io/klog/v2"
```
abhinavdahiya, line 14: so the [vendor update](https://www.youtube.com/watch?v=KR0g-1hnQPA) doesn't reflect that klog was dropped or not used anymore by anyone.

so how can ensure no  logs end up into stderr from both versions

nirarg: I can use here both klog versions, do you think this is better approach?

.gitignore
nirarg: I agree about the dev.log, will drop it
Regarding the *.swp, this is temporary file used by the vi/vim editor and might be accidentally added to commits. as this type file never should be added, isn't it good enough reason to add it (and maybe also other editor's specific files)

cmd/openshift-install/main.go
abhinavdahiya: > I can use here both klog versions, do you think this is better approach?

i think that would be safer for now.

.gitignore
abhinavdahiya: I think people should use https://stackoverflow.com/a/1753078 to add any editor ignore files. Forcing repos to include ignore files for all editors is too much of a slippery slope imo, so i would personally not add anything here.

nirarg: Thanks @abhinavdahiya 
I removed this change

cmd/openshift-install/main.go
nirarg: Thanks @abhinavdahiya 
I changed it, please review

cmd/openshift-install/main.go
```diff
@@ -28,8 +29,10 @@ func main() {
 	// to log anything.
 	var fs flag.FlagSet
 	klog.InitFlags(&fs)
+	klogv2.InitFlags(&fs)
```
abhinavdahiya, line 32: I think using the same flag set for both v1 and v2 is causing this failure https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/4266/pull-ci-openshift-installer-master-e2e-aws/1316299561756528640#1:build-log.txt%3A129

PR #4269, opened 2020-10-14
Bug 1878900: openstack: Fix error messages in flavor validation

pierreprinetti:
Flavor validation was reporting issues with wrong `install-config.yaml`
property names.

/label platform/openstack


mandre (approved):
/approve
/lgtm

PR #3772, opened 2020-06-18
destroy: return aws resources that could not be deleted

staebler:
The `RunWithContext` method has been modified to return a slice of ARNs that could not be destroyed. Only the ARNs of the first-level resources will be returned. For example, when deleting a VPC, the uninstaller will first delete other resources that use the VPC. Any of those resources that are blocked from being deleted but are not tagged for the cluster will not show up in the list of blocked resources. However, the first-level VPC will show up as blocked in this case.
    
This will be used by Hive to expose to the user the resources that cannot be deleted so that the user can take action on those resources.
    
https://issues.redhat.com/browse/CO-973

PR #4270, opened 2020-10-14
destroy: return aws resources that could not be deleted

staebler:
The RunWithContext method has been modified to return a slice of ARNs that could not be destroyed. Only the ARNs of the first-level resources will be returned. For example, when deleting a VPC, the uninstaller will first delete other resources that use the VPC. Any of those resources that are blocked from being deleted but are not tagged for the cluster will not show up in the list of blocked resources. However, the first-level VPC will show up as blocked in this case.

This will be used by Hive to expose to the user the resources that cannot be deleted so that the user can take action on those resources.

https://issues.redhat.com/browse/CO-973


staebler:
This PR is a re-creation of https://github.com/openshift/installer/pull/3772. The original PR was having problems with the e2e-aws test.

PR #4273, opened 2020-10-15
Bug 1886553: GCP - Increase worker NAT min ports

alexanderConstantinescu:
Our worker nodes NAT instances are configured on GCP to allow for 512 ports per VM instance.
Master nodes have the same setting set to 7168 ports. However worker nodes usually run the load
(i.e pods) which instantiate a huge amount of open connections, we thus need to increase this
value to avoid NAT packet drops on GCP when running high load. This has been a big cause of
connection timeout problems on GCP clusters.

Slack channel for incident which took 2 weeks to solve: https://coreos.slack.com/archives/C01BQ9UR5J9
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1886553

Signed-off-by: Alexander Constantinescu <aconstan@redhat.com>

### DIFF
--- a/data/data/gcp/network/network.tf
+++ b/data/data/gcp/network/network.tf
@@ -50,7 +50,7 @@ resource "google_compute_router_nat" "worker_nat" {
   name                               = "${var.cluster_id}-nat-worker"
   router                             = google_compute_router.router[0].name
   nat_ip_allocate_option             = "AUTO_ONLY"
-  min_ports_per_vm                   = 512
+  min_ports_per_vm                   = 4096
   source_subnetwork_ip_ranges_to_nat = "LIST_OF_SUBNETWORKS"
 
   subnetwork {


sdodson:
@jstuever Does this affect the number of VMs that can be behind this NAT?

stevekuznetsov:
This mitigated DPTP-1582

sdodson:
> @jstuever Does this affect the number of VMs that can be behind this NAT?

Answering my own question, the NAT is setup to autoscale based on the min ports and VMS so this shouldn't be a problem other than allocation of additional public ips. We should probably attempt to monitor where a busy cluster ends up as far as utilization goes before we jump directly to 7k but that may not be worth the effort.

abhinavdahiya:
Personally I think we should let the CI cluster that needs this excessive external traffic to update their nat gateway to support more ports per vm.
 Otherwise this will force increased number of ips per cluster for each new VM that is added to the cluster for our users. With 7000 ports per vm,  we will need 1 additional public per 10 VMs .. which is an overkill. 

This is better docs with some kcs article that explains how you can improve where traffic nat in case you were not using pre existing VPC.

/hold

alexanderConstantinescu:
> Personally I think we should let the CI cluster that needs this excessive external traffic to update their nat gateway to support more ports per vm.

How would they know though? Their networking will "silently" fail / drop packets due to this, they will open a support ticket to us and we will probably need 2 weeks more to figure out that this has happened again (probably engaging Google's GCP support in the process too)...

> Otherwise this will force increased number of ips per cluster for each new VM that is added to the cluster for our users. With 7000 ports per vm, we will need 1 additional public per 10 VMs .. which is an overkill.

Why is that overkill? What harm does that do? 

> Making 512 connections parallel to one destination on internet by one VM seems like enough for most usecases.

"enough for most usecases", which are those? And how have those been measured? 

What I am trying to say is that the effect of not having this patch is that a customer might open a up a support ticket to us in the future with a vague description such as "OCP is dropping packets"/"OCP network performance is terrible" which will be incredibly difficult to weave out. The opposite is that we will use additional public IPs if the cluster scales out, something which is completely transparent to everyone since the cluster DNS name would still resolve to all of those addresses.

I think it's better to integrate this as to avoid such a problem and have the performance and scale test team evaluate which value is best for heavy traffic load on GCP OCP clusters, as we mentioned in the channel: https://coreos.slack.com/archives/C01BQ9UR5J9/p1602773245045600?thread_ts=1602703490.019500&cid=C01BQ9UR5J9

data/data/gcp/network/network.tf
```diff
@@ -50,9 +53,12 @@ resource "google_compute_router_nat" "worker_nat" {
   name                               = "${var.cluster_id}-nat-worker"
   router                             = google_compute_router.router[0].name
   nat_ip_allocate_option             = "AUTO_ONLY"
-  min_ports_per_vm                   = 512
+  min_ports_per_vm                   = 7168
```
alvaroaleman, line 56: If this setting is identical for master/worker now, is there still value in having two nat gateways?

alexanderConstantinescu: Good question. I would say stick to keeping two separate ones. Once a proper investigation is done into all of this we might realize we need to separate values. But then again, I am always in favor of the philosophy "fix it when broken/do it when needed", but in this case I am not sure. Maybe someone from installer has a stronger opinion?

sdodson:
This value defines the number of connections per VM to a 3-tuple consisting of a destination IP address, a destination port, and an IP protocol (such as TCP or UDP), see https://cloud.google.com/nat/docs/ports-and-addresses#ports-and-connections.

512 connections to a such a 3-tuple is really really high in my opinion which is why we've probably never heard of this, likely due to the fact that our oc discovery caching is failing and the nature of our tests. Real world behavior will rarely see this exceptionally high connection churn to a single 3-tuple, right?

I'm fine with scaling this up but I think we should do so in a manner that leaves things unchanged for the 95% of clusters which are 16 nodes and smaller. Can we go with 4096 until we have data that shows a different value is necessary? Even if the data is as shaky as build02 was failing at < 16 nodes it's something.

alexanderConstantinescu:
> 512 connections to a such a 3-tuple is really really high in my opinion which is why we've probably never heard of this, likely due to the fact that our oc discovery caching is failing and the nature of our tests. Real world behavior will rarely see this exceptionally high connection churn to a single 3-tuple, right?

I am not sure. If a company has a large OCP cluster deployed on GCP and it needs to talk to one API endpoint (database service / LB / gateway) it might just generate some traffic spikes of that sort. 

> I'm fine with scaling this up but I think we should do so in a manner that leaves things unchanged for the 95% of clusters which are 16 nodes and smaller. Can we go with 4096 until we have data that shows a different value is necessary? Even if the data is as shaky as build02 was failing at < 16 nodes it's something.

OK, I'll update the value to 4096 and lets take it from there.

sdodson:
@abhinavdahiya How do you feel now that it's at 4k and we won't have to scale anything for 95% of clusters that are < 16 workers?

data/data/gcp/network/network.tf
```diff
@@ -37,7 +37,10 @@ resource "google_compute_router_nat" "master_nat" {
   nat_ip_allocate_option             = "AUTO_ONLY"
   min_ports_per_vm                   = 7168
   source_subnetwork_ip_ranges_to_nat = "LIST_OF_SUBNETWORKS"
-
+  log_config {
+    enable = true
+    filter = ERRORS_ONLY
+  }
```
abhinavdahiya, line 43: we don't enable logging for various other networking components and I don't think we should enable these for all the customers. the users should be able to enable this for their clusters whenever they want.

abhinavdahiya:
> @abhinavdahiya How do you feel now that it's at 4k and we won't have to scale anything for 95% of clusters that are < 16 workers?

sure, that's more acceptable.

keeping the hold since b213acf0a6661273d2c895c66a21e75dcbad5fb1 is not best for default setup as enabling logging for various components is not something we do today for any resource in any platform and i think the users should be able to make this change as and when they require. don't really know how the cost structure is for gcp logging

sdodson:
/retitle Bug 1886553: GCP - Increase worker NAT min ports

sdodson:
/lgtm
Now that the logging has been dropped

PR #4274, opened 2020-10-15
vSphere: Update dockerfile removing certificate

jcpowermac:
This vSphere environment is being decommissioned and VMC
uses valid certificates for vCenter so this is no longer required.

This is required for UPI CI testing to continue to function
(and whatever else tests that use `upi-installer`)


jcpowermac:
This one no longer uses `upi-installer` image but just in case
/test e2e-vsphere

/test e2e-vsphere-upi

PR #3964, opened 2020-07-27
OpenStack: add clusterOSImage validations

Fedosin:
This commit adds validations that custom Glance image specified by `clusterOSImage` config parameter exists and has Active status.

PR #4277, opened 2020-10-16
OpenStack: remove legacy image validation

Fedosin:
Previously we validate overriden Glance images during generation of tfvars variables. This is not a correct place to do it, so we
implemented the check in the "validation" module:
https://github.com/openshift/installer/pull/3964

To prevent code duplication we should remove the legacy code.


mandre (approved):
/approve
/lgtm

PR #4279, opened 2020-10-19
Bug 1882649: Determine Glance disk format based on file extension

Fedosin:
Now we always use "qcow2" disk format for Glance images. This is not correct if a user wants to upload a raw format image (with "raw" file extension).
This commit checks the extension and allows to set disk format accordingly.


mandre (approved):
/approve
/lgtm
/hold until the test pass

PR #4282, opened 2020-10-19
cmd: split newlines in logrus stderr output to new entry

abhinavdahiya:
Currently the logrus hook runs formatter once for each entry, which causes newlines to be escaped see sirupsen/logrus#608
Because of this multiline messages end up looking like

```go
logrus.Info(`some message
same mesage multiline 1`)
logrus.Info("next meesage")
```

```
LEVEL some message
same message multiline 1
LEVEL next messsage
```

With this change the hook will format each entry's message split on newlines so that,

```go
logrus.Info(`some message
same mesage multiline 1`)
logrus.Info("next meesage")
```

```
LEVEL some message
LEVEL same message multiline 1
LEVEL next messsage
```

This will help 2 cases,
- af9b49c like informational messages that are require multiple lines to be more user-friendly.
- meesages from cluster operators on install-complete failure see #4242

/cc @deads2k @jstuever

### DIFF
--- a/cmd/openshift-install/main.go
+++ b/cmd/openshift-install/main.go
@@ -95,7 +95,7 @@ func runRootCmd(cmd *cobra.Command, args []string) {
 		level = logrus.InfoLevel
 	}
 
-	logrus.AddHook(newFileHook(os.Stderr, level, &logrus.TextFormatter{
+	logrus.AddHook(newFileHookWithNewlineTruncate(os.Stderr, level, &logrus.TextFormatter{
 		// Setting ForceColors is necessary because logrus.TextFormatter determines
 		// whether or not to enable colors by looking at the output of the logger.
 		// In this case, the output is ioutil.Discard, which is not a terminal.
@@ -104,6 +104,7 @@ func runRootCmd(cmd *cobra.Command, args []string) {
 		ForceColors:            terminal.IsTerminal(int(os.Stderr.Fd())),
 		DisableTimestamp:       true,
 		DisableLevelTruncation: true,
+		DisableQuote:           true,
 	}))
 
 	if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/openshift-install/main.go
```diff
@@ -104,6 +104,7 @@ func runRootCmd(cmd *cobra.Command, args []string) {
 		ForceColors:            terminal.IsTerminal(int(os.Stderr.Fd())),
 		DisableTimestamp:       true,
 		DisableLevelTruncation: true,
+		DisableQuote:           true,
```
deads2k, line 107: this eliminates the escaping logic they've been providing?  I like the change, but I recognize that this could change machine parsability in a way to splitting newlines does not, right?

deads2k:
logs look good to me

```
level=error msg=Cluster initialization failed because one or more operators are not functioning properly.
level=error msg=The cluster should be accessible for troublsehooting as detailed in the documentation linked below,
level=error msg=https://docs.openshift.com/container-platform/latest/support/troubleshooting/troubleshooting-installations.html
level=error msg=The 'wait-for install-complete' subcommand can then be used to continue the installation
level=fatal msg=failed to initialize the cluster: Working towards 4.7.0-0.ci.test-2020-10-19-204836-ci-op-1wpj6qwz: 9% complete
level=debug msg=OpenShift Installer 4.7.0-0.ci.test-2020-10-19-204836-ci-op-1wpj6qwz

```

/lgtm
/hold

holding for my question (I think it's fine to make the change, just want to make sure its intentional) and  in case @abhinavdahiya has other reviewers he'd like

cmd/openshift-install/main.go
abhinavdahiya: For users with interactive terminal this has no effect.
For non-interactive outputs like CI pods, this will no quote the message and yes I think this will make the message show up on multiple lines instead of one line.
but the other change will already have split up the message on newline so this only makes the message look cleaner with no unnecessary escaping.

PR #4285, opened 2020-10-20
Bug 1889267: oVirt, increase terraform template and release image timeout to 20m

Gal-Zaidman:
We noticed that sometime the installation fails on template creation or release image upload just because of the 10m timeout.
We increase the time out to 20m allowing resources to finish

Signed-off-by: Gal-Zaidman <gzaidman@redhat.com>

### DIFF
--- a/data/data/ovirt/template/main.tf
+++ b/data/data/ovirt/template/main.tf
@@ -38,6 +38,9 @@ resource "ovirt_image_transfer" "releaseimage" {
   source_url        = var.openstack_base_image_local_file_path
   storage_domain_id = var.ovirt_storage_domain_id
   sparse            = true
+  timeouts {
+    create = "20m"
+  }
 }
 
 resource "ovirt_vm" "tmp_import_vm" {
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -100,6 +100,7 @@ require (
 	k8s.io/apiextensions-apiserver v0.19.0
 	k8s.io/apimachinery v0.19.0
 	k8s.io/client-go v12.0.0+incompatible
+	k8s.io/klog v1.0.0
```
abhinavdahiya, line 103: why this explicit dep?

vendor/github.com/ovirt/terraform-provider-ovirt/ovirt/resource_ovirt_image_transfer.go
```diff
@@ -38,7 +38,7 @@ func resourceOvirtImageTransfer() *schema.Resource {
 		},
 
 		Timeouts: &schema.ResourceTimeout{
-			Create: schema.DefaultTimeout(10 * time.Minute),
+			Create: schema.DefaultTimeout(20 * time.Minute),
```
abhinavdahiya, line 41: can you use timeout section in the terraform files to override this value?

Gal-Zaidman: I wasn't aware that is possible, but now that you pointed it I see that the timeout block gives us that ability.
I'll fix and revert the patch to the terraform provider.
Thanks for the catch

go.mod
Gal-Zaidman: removed

sandrobonazzola (approved):
/lgtm

mtnbikenc (approved):
/lgtm

PR #4287, opened 2020-10-21
bug 1870728: pkg/asset/ignition/bootstrap: Display warning if certificates are expired

jhixson74:
When loading bootstrap.ign, check if there are expired certificates and display a warning if so.

https://bugzilla.redhat.com/show_bug.cgi?id=1870728

### DIFF
--- a/pkg/asset/ignition/bootstrap/bootstrap.go
+++ b/pkg/asset/ignition/bootstrap/bootstrap.go
@@ -2,7 +2,9 @@ package bootstrap
 
 import (
 	"bytes"
+	"crypto/x509"
 	"encoding/json"
+	"encoding/pem"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -11,6 +13,7 @@ import (
 	"path/filepath"
 	"strings"
 	"text/template"
+	"time"
 
 	"github.com/containers/image/pkg/sysregistriesv2"
 	ignutil "github.com/coreos/ignition/v2/config/util"
@@ -18,6 +21,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	"github.com/vincent-petithory/dataurl"
 
 	"github.com/openshift/installer/data"
 	"github.com/openshift/installer/pkg/asset"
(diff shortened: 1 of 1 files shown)


rvanderp3:
> When loading bootstrap.ign, check if there are expired certificates and display a warning if so. I'm not coupling this to the bug until I get verification this is going the right direction ;-)
> 
> https://bugzilla.redhat.com/show_bug.cgi?id=1870728

Yes, this is what I was looking for in the BZ but there may be other thoughts on the implementation.  I like this a lot more than the solution I had put together in the PR attached to the BZ.

mtnbikenc (approved):

mtnbikenc:
Just noticed a small golint error that needs to be addressed.

mtnbikenc (changes requested):
Just noticed a small golint issue that needs to be addressed.
/lgtm cancel

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -557,5 +561,41 @@ func (a *Bootstrap) Load(f asset.FileFetcher) (found bool, err error) {
...
 	a.File, a.Config = file, config
+	warnIfCertificatesExpired(a.Config)
 	return true, nil
 }
+
+// warnIfCertificatesExpired checks for expired certificates and warns if so
+func warnIfCertificatesExpired(config *igntypes.Config) {
+	expiredCerts := 0
+	for _, file := range config.Storage.Files {
+		if strings.HasSuffix(file.Path, ".crt") && file.Contents.Source != nil {
```
abhinavdahiya, line 572: https://golang.org/pkg/path/filepath/#Ext

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -557,5 +561,41 @@ func (a *Bootstrap) Load(f asset.FileFetcher) (found bool, err error) {
...
+			// ^data:text/plain;charset=utf-8;base64,LS0.+==
+			i := strings.Index(*file.Contents.Source, ",LS0")
+			if i == -1 {
+				continue
+			}
```
abhinavdahiya, line 577: https://github.com/vincent-petithory/dataurl I think ignition uses this..

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -557,5 +561,41 @@ func (a *Bootstrap) Load(f asset.FileFetcher) (found bool, err error) {
...
+			block, _ := pem.Decode(data)
+			if block != nil {
+				cert, err := x509.ParseCertificate(block.Bytes)
+				if err == nil {
+					if time.Now().UTC().After(cert.NotAfter) {
+						logrus.Warnf("Bootstrap Ignition-Config Certificate %s expired at %s.", path.Base(file.Path), cert.NotAfter.Format(time.RFC3339))
+						expiredCerts++
+					}
+				} else {
+					logrus.Debugf("Unable to parse certificate. %s", err.Error())
```
abhinavdahiya, line 592: some information on which certificate would be helpful in the logs

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -557,5 +561,41 @@ func (a *Bootstrap) Load(f asset.FileFetcher) (found bool, err error) {
...
+			i := strings.Index(*file.Contents.Source, ",LS0")
+			if i == -1 {
+				continue
+			}
+			data, err := base64.StdEncoding.DecodeString((*file.Contents.Source)[i+1:])
+			if err != nil {
+				logrus.Debugf("Unable to decode certificate. %s", err.Error())
+				continue
+			}
+			block, _ := pem.Decode(data)
```
abhinavdahiya, line 583: I think this assumes there will only be one certificate in the file??

pkg/asset/ignition/bootstrap/bootstrap.go
jhixson74: Fixed.

pkg/asset/ignition/bootstrap/bootstrap.go
jhixson74: Fixed.

pkg/asset/ignition/bootstrap/bootstrap.go
jhixson74: Is there ever a case where there is a certificate chain in bootstrap.ign? If not, this works as intended. Every certificate in bootstrap.ign is checked for expiration.

pkg/asset/ignition/bootstrap/bootstrap.go
jhixson74: Fixed.

PR #4289, opened 2020-10-21
Bug 1887863: Patch Flavor Not Found validation for OpenStack Install Config

iamemilio:
Fixes an issue where we expected cloudInfo to not have a
key for the flavorName if it was not found. In the code and the unit tests, cloudInfo's flavor map now has a pointer value
instead of a struct. This allows us to set it to nil when the flavor is not found in openstack.

/label platform/openstack

### DIFF
--- a/pkg/asset/installconfig/openstack/validation/cloudinfo.go
+++ b/pkg/asset/installconfig/openstack/validation/cloudinfo.go
@@ -93,13 +93,6 @@ func (ci *CloudInfo) collectInfo(ic *types.InstallConfig) error {
 		return errors.Wrap(err, "failed to fetch external network info")
 	}
 
-	if flavorName := ic.OpenStack.FlavorName; flavorName != "" {
-		ci.Flavors[flavorName], err = ci.getFlavor(flavorName)
-		if err != nil {
-			return errors.Wrap(err, "failed to fetch platform flavor info")
-		}
-	}
-
 	// Fetch the image info if the user provided a Glance image name
 	imagePtr := ic.OpenStack.ClusterOSImage
 	if imagePtr != "" {
@@ -111,11 +104,21 @@ func (ci *CloudInfo) collectInfo(ic *types.InstallConfig) error {
 		}
 	}
 
+	// Get flavor info
+	if flavorName := ic.OpenStack.FlavorName; flavorName != "" {
+		ci.Flavors[flavorName], err = ci.getFlavor(flavorName)
+		if err != nil {
+			return err
+		}
+	}
+
 	if ic.ControlPlane != nil && ic.ControlPlane.Platform.OpenStack != nil {
 		if flavorName := ic.ControlPlane.Platform.OpenStack.FlavorName; flavorName != "" {
-			ci.Flavors[flavorName], err = ci.getFlavor(flavorName)
-			if err != nil {
-				return err
+			if _, seen := ci.Flavors[flavorName]; !seen {
+				ci.Flavors[flavorName], err = ci.getFlavor(flavorName)
+				if err != nil {
+					return err
+				}
 			}
 		}
 	}
@@ -169,11 +172,15 @@ func (ci *CloudInfo) getSubnet(subnetID string) (*subnets.Subnet, error) {
 	return subnet, nil
 }
 
+func isNotFoundError(err error) bool {
+	var errNotFound *gophercloud.ErrResourceNotFound
+	return errors.As(err, &errNotFound) || strings.Contains(err.Error(), "Resource not found")
+}
+
 func (ci *CloudInfo) getFlavor(flavorName string) (Flavor, error) {
 	flavorID, err := flavorutils.IDFromName(ci.clients.computeClient, flavorName)
 	if err != nil {
-		var gerr *gophercloud.ErrResourceNotFound
-		if errors.As(err, &gerr) {
+		if isNotFoundError(err) {
 			return Flavor{}, nil
 		}
 		return Flavor{}, err
@@ -187,10 +194,9 @@ func (ci *CloudInfo) getFlavor(flavorName string) (Flavor, error) {
 	var baremetal bool
 	{
 		const baremetalProperty = "baremetal"
-		var errNotFound *gophercloud.ErrResourceNotFound
 
 		m, err := flavors.GetExtraSpec(ci.clients.computeClient, flavorID, baremetalProperty).Extract()
-		if err != nil && !errors.As(err, &errNotFound) && !strings.Contains(err.Error(), "Resource not found") {
+		if err != nil && !isNotFoundError(err) {
 			return Flavor{}, err
 		}
 
--- a/pkg/asset/installconfig/openstack/validation/machinepool.go
+++ b/pkg/asset/installconfig/openstack/validation/machinepool.go
@@ -41,17 +41,10 @@ func ValidateMachinePool(p *openstack.MachinePool, ci *CloudInfo, controlPlane b
 		}
 	}
 
-	if p.FlavorName != "" {
-		flavor, ok := ci.Flavors[p.FlavorName]
-		if ok {
-			if controlPlane {
-				allErrs = append(allErrs, validateFlavor(flavor, ctrlPlaneFlavorMinimums, fldPath.Child("type"))...)
-			} else {
-				allErrs = append(allErrs, validateFlavor(flavor, computeFlavorMinimums, fldPath.Child("type"))...)
-			}
-		} else {
-			allErrs = append(allErrs, field.NotFound(fldPath.Child("type"), p.FlavorName))
-		}
+	if controlPlane {
+		allErrs = append(allErrs, validateFlavor(p.FlavorName, ci, ctrlPlaneFlavorMinimums, fldPath.Child("type"))...)
+	} else {
+		allErrs = append(allErrs, validateFlavor(p.FlavorName, ci, computeFlavorMinimums, fldPath.Child("type"))...)
 	}
 
 	allErrs = append(allErrs, validateZones(p.Zones, ci.Zones, fldPath.Child("zones"))...)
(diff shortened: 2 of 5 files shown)


pkg/asset/installconfig/openstack/validation/platform_test.go
```diff
@@ -121,7 +121,7 @@ func TestOpenStackPlatformValidation(t *testing.T) {
 			cloudInfo:      validPlatformCloudInfo(),
 			networking:     validNetworking(),
 			expectedError:  true,
-			expectedErrMsg: `platform.openstack.computeFlavor: Invalid value: "invalid-control-plane-flavor": Flavor did not meet the following minimum requirements: Must have minimum of 16 GB RAM, had 8 GB; Must have minimum of 4 VCPUs, had 2; Must have minimum of 25 GB Disk, had 20 GB`,
+			expectedErrMsg: `platform.openstack.flavorName: Invalid value: "invalid-control-plane-flavor": Flavor did not meet the following minimum requirements: Must have minimum of 16 GB RAM, had 8 GB; Must have minimum of 4 VCPUs, had 2; Must have minimum of 25 GB Disk, had 20 GB`,
```
Fedosin, line 124: the option is called `computeFlavor`: https://github.com/openshift/installer/blob/master/data/data/install.openshift.io_installconfigs.yaml#L1013

No need to rename it.

pkg/asset/installconfig/openstack/validation/machinepool_test.go
```diff
@@ -137,9 +137,13 @@ func TestOpenStackMachinepoolValidation(t *testing.T) {
...
 			}(),
-			cloudInfo:      validMpoolCloudInfo(),
+			cloudInfo: func() *CloudInfo {
+				ci := validMpoolCloudInfo()
+				ci.Flavors[notExistFlavor] = nil
+				return ci
+			}(),
 			expectedError:  true,
-			expectedErrMsg: "controlPlane.platform.openstack.type: Not found: \"non-existant-flavor\"",
+			expectedErrMsg: "controlPlane.platform.openstack.flavorName: Not found: \"non-existant-flavor\"",
```
Fedosin, line 146: it is called "type", not "flavorName": https://github.com/openshift/installer/blob/master/data/data/install.openshift.io_installconfigs.yaml#L428

let's keep it as it was before

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -41,17 +41,10 @@ func ValidateMachinePool(p *openstack.MachinePool, ci *CloudInfo, controlPlane b
...
-			if controlPlane {
-				allErrs = append(allErrs, validateFlavor(flavor, ctrlPlaneFlavorMinimums, fldPath.Child("type"))...)
-			} else {
-				allErrs = append(allErrs, validateFlavor(flavor, computeFlavorMinimums, fldPath.Child("type"))...)
-			}
-		} else {
-			allErrs = append(allErrs, field.NotFound(fldPath.Child("type"), p.FlavorName))
-		}
+	if controlPlane {
+		allErrs = append(allErrs, validateFlavor(p.FlavorName, ci, ctrlPlaneFlavorMinimums, fldPath.Child("flavorName"))...)
```
Fedosin, line 45: fldPath.Child("type")

Fedosin (changes requested):

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -105,11 +98,23 @@ func validUUIDv4(s string) bool {
 	return true
 }
 
-func validateFlavor(flavor Flavor, req flavorRequirements, fldPath *field.Path) field.ErrorList {
+// validate flavor checks to make sure that a given flavor exists and meets the minimum requrement to run a cluster
+// this function does not validate proper install config usage
+// NOTE: Because we append the result of this function to an error list, it cannot be nil
+func validateFlavor(flavorName string, ci *CloudInfo, req flavorRequirements, fldPath *field.Path) field.ErrorList {
+	if flavorName == "" {
```
adduarte, line 105: could this possibly hide a problem?  The flavorName being passed int should never be empty. 
But here if it is, it is silently overlooked.  Maybe fill the errorList? or let it fall through 
The next lookup "ci.Flavors[flavorName]  will catch it

iamemilio: That should be handled by the static usage validations. There are cases where it is valid for flavor name to be empty

pkg/asset/installconfig/openstack/validation/platform_test.go
iamemilio: ack to all name changes

pkg/asset/installconfig/openstack/validation/machinepool.go
iamemilio: The purpose of this block of code is just to make sure that if a flavor is provided, that flavor is valid

pierreprinetti:
> The error message was reporting errors on field "type" instead of "flavorName"

My patch #4269 changes the error message to mention "type" (and not "flavorName"), because [that seems to be the name of the JSON property in MachinePool](https://github.com/openshift/installer/blob/31dba4362530c3afaacf20e3a7971a10c1d9f288/pkg/types/openstack/machinepool.go#L8).

> cloudInfo's flavor map now has a pointer value instead of a struct

Can you help me understand why a pointer is necessary? The "comma OK" idiom is how you would check if a value was in the map. See this example: https://play.golang.org/p/Ubj2NqhG9pg

---

I am not sure what case you want to address with this patch; perhaps you could create a bug in Bugzilla that describes what happens vs what you expected to happen?

/hold

iamemilio:
https://issues.redhat.com/browse/OSASINFRA-2133

iamemilio:
The pointer lets us set the entire value to nil in the Flavors map when openstack fails to find a given flavor, rather than setting it to Flavor{}. This lets us more easily indicate that it was not found. The way the library is built, it is not optimal for us to exclude lookups that failed from the Flavors map, since we use it as an optomization to check which openstack queries we have already run to avoid duplicating calls to the openstack api.

iamemilio:
NOTE: needs squash before merge

pierreprinetti:
I think that the problem is here:
https://github.com/openshift/installer/blob/ac6f233c352843294abf9719fe5659527cdc5bf6/pkg/asset/installconfig/openstack/validation/cloudinfo.go#L175-L178

If you just remove these 3 lines, the validation errors with a nice message when the flavor is not found by name.

@Fedosin Do you perhaps remember why we are silently ignoring a FlavorNotFound error?

iamemilio:
We try to defer errors caused by user error in the install config to a section of code that can append it to the error list so that the customer gets all of their install config validation errors in a list.

Fedosin:
@pierreprinetti during cloud info generation we check only critical errors that cause the installer to stop immediately (auth problems, 5xx errors). Here we do not consider ErrResourceNotFound critical to allow the validator to check other install config fields later, and return a full list of issues.

iamemilio:
/assign @Fedosin 

Going to assign mike too since he is knowledgeable on this code

pierreprinetti:
/assign

Turning a map of structs into a map of pointers to pass the message that the flavor might not have been found is not a good idea.

If you don't want to directly add the notFound error to the list of errors returned to the user, then maybe you can embed it into the Flavor struct.

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -105,11 +98,25 @@ func validUUIDv4(s string) bool {
...
+		return field.ErrorList{}
+	}
```
staebler, line 107: The note is not accurate. It is perfectly fine to unpack a nil slice.
```suggestion
func validateFlavor(flavorName string, ci *CloudInfo, req flavorRequirements, fldPath *field.Path) field.ErrorList {
	if flavorName == "" {
		return nil
	}
```

pkg/asset/installconfig/openstack/validation/cloudinfo.go
```diff
@@ -169,28 +176,37 @@ func (ci *CloudInfo) getSubnet(subnetID string) (*subnets.Subnet, error) {
 	return subnet, nil
 }
 
+func isNotFoundError(err error) bool {
+	var errNotFound *gophercloud.ErrResourceNotFound
+	if errors.As(err, &errNotFound) || strings.Contains(err.Error(), "Resource not found") {
+		return true
+	}
+	return false
```
staebler, line 184:
```suggestion
	return errors.As(err, &errNotFound) || strings.Contains(err.Error(), "Resource not found")
```

pkg/asset/installconfig/openstack/validation/cloudinfo.go
```diff
@@ -45,6 +45,7 @@ type clients struct {
 type Flavor struct {
 	*flavors.Flavor
 	Baremetal bool
+	userError error // userError contains a non-critical error that is cause by incorrect usage of the flavor field
```
staebler, line 48: It is not clear to me why the flavor is a special case that requires storing `userError`. For example, if the OS image is not found, then validation will yield a generic "NotFound" error. This is accomplished by checking that `OSImage` in the cloud info is nil. Can we not similarly check if `ci.Flavors[flavorName].Flavor == nil` to determine if the flavor was not found?

pkg/asset/installconfig/openstack/validation/machinepool.go
iamemilio: ack, will fix

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -105,11 +98,25 @@ func validUUIDv4(s string) bool {
...
+
+	flavor, ok := ci.Flavors[flavorName]
+	if !ok || flavor.userError != nil {
+		if isNotFoundError(flavor.userError) {
+			return field.ErrorList{field.NotFound(fldPath, flavorName)}
+		}
+	}
```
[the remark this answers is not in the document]
iamemilio: yeah the check for ok doesnt do anything actually

iamemilio:
/retest 

I think I took everyones points into consideration for the latest draft. PTAL

pkg/asset/installconfig/openstack/validation/cloudinfo.go
iamemilio: didnt know that! will fix

Fedosin (approved):
/lgtm
/approve

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -124,7 +128,7 @@ func validateFlavor(flavor Flavor, req flavorRequirements, fldPath *field.Path)
 	}
 
 	if len(errs) == 0 {
-		return nil
+		return field.ErrorList{}
```
pierreprinetti, line 131: nit: This can stay `nil` :)

pierreprinetti (approved):
I like how this last version of the patch makes it consistent with the other resource checks (without hurting my feelings with unnecessary additional pointers :) )

I will /lgtm again if you plan to squash. Thanks!

pkg/asset/installconfig/openstack/validation/machinepool.go
iamemilio: ack, ill fix it

iamemilio:
/hold cancel

Squashed and ready for final review

PR #4296, opened 2020-10-22
WIP: installconfig: aws: allow role based credentials

sjenning:
@dgoodwin @joelddiaz @abhinavdahiya @derekwaynecarr @brenton 

I was able to generate an RSA keypair, create an OIDC provider from that, create a JWT signed by the keypair such that it can assume a high powered role for installation.

However the current credential chain does not include the `AssumeRoleProvider`.

Current result is:
```
$ aws/openshift-install create cluster --dir=aws
? The AWS access key ID to use for installation (this is not your username).
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_access-keys.html
? AWS Access Key ID 
FATAL failed to fetch Terraform Variables: failed to fetch dependency of "Terraform Variables": failed to fetch dependency of "Bootstrap Ignition Config": failed to fetch dependency of "Master Machines": failed to generate asset "Platform Credentials Check": creating AWS session: interrupt 

$ echo $AWS_ROLE_ARN 
arn:aws:iam::269733383066:role/sjenning-admin-installer-role

$ echo $AWS_WEB_IDENTITY_TOKEN_FILE 
/home/sjennings/pod-identity/token
```

This PR adds it.

Granted, even once this merges, I imagine it will fail to generate the `aws-creds` secret as some point.  Just the first step.

### DIFF
--- a/pkg/asset/installconfig/aws/session.go
+++ b/pkg/asset/installconfig/aws/session.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/credentials/stscreds"
 	"github.com/aws/aws-sdk-go/aws/defaults"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/request"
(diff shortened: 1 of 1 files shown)


sjenning:
meh, apparently `AssumeRoleProvider` doesn't initialize the same way

chemapolo:
Hi,

I've had the same issue with release 4.5, i don't know if it will help but at least it worked for me.

I've forked the repo at: https://github.com/chemapolo/installer/tree/chema/debug

and I've modified just session.go 

https://github.com/chemapolo/installer/blob/chema/debug/pkg/asset/installconfig/aws/session.go

and commented the credentials part, just because of the fact that following AWS SDK doc adding:

$AWS_CONFIG_FILE
$AWS_DEFAULT_PROFILE
$AWS_SDK_LOAD_CONFIG=1 
$AWS_SHARED_CREDENTIALS_FILE

vars to environment it works perfectly. let's say that adding the credentials to the session was removing the remaining part of the defined profile @config files.

Just keeping ssn := session.Must(session.NewSessionWithOptions(options)) it has all the required info to be able to assume properly the role.

Hope this helps. Regards

PR #4300, opened 2020-10-23
Add `openshift-install ignition convert`

cgwalters:
The transition to Ignition Spec 3 with 4.6 creates a
discontinuity.  Some users want to update their bootimages,
e.g. for a cluster originally provisioned as 4.4 but upgraded
in place to 4.6, it should be possible to directly use RHCOS 4.6
bootimages for new workers.

In some cases in fact, this could be *required* for things like
adding a node with newer hardware.

The main stumbling block here is the pointer ignition config.
Since `openshift-install` already includes Ignition bits, let's
add translation capability here using
https://github.com/coreos/ign-converter
the same as the MCO uses.

xref https://github.com/openshift/enhancements/pull/492#issuecomment-713096391
xref https://bugzilla.redhat.com/show_bug.cgi?id=1884750


cgwalters:
We've hit this when porting some of our UPI tests [as well as the proxy test](https://github.com/openshift/release/pull/11750) - that one we ported to FCOS, but if we'd had this PR we could have kept the spec2 configs for a while and used this to upconvert.

staebler:
My gut reaction to this is that it is not an appropriate command for `openshift-install`. The purview of `openshift-install` is installing clusters and destroying clusters: It is not supporting day-2 administration of clusters. As such, `openshift-install` is not a good home for tooling to support adding new nodes to an existing cluster. Although I do not like the idea of adding more to its hodgepodge of commands, I feel that `oc` is stronger candidate for this command than `openshift-install`.

cgwalters:
>  It is not supporting day-2 administration of clusters.

Agree.  However, today we have a model where admins often need to download `openshift-install` and the RHCOS bootimages at the same time, so they will have both.  Plus `openshift-install` *generated* these pointer configs, so it's somewhat natural to have it support converting them.

Plus `openshift-install` already vendors Ignition code.

>  I feel that oc is stronger candidate for this command than openshift-install.

Yeah, that was my original proposal as well...I dunno.  I flip-flopped based on the above but yeah I am OK with doing `oc`.

cgwalters:
OK moved to https://github.com/openshift/oc/pull/628

PR #4304, opened 2020-10-26
WIP: Add feature gate manifest generation

Fedosin:


jsafrane:
I don't see this mentioned in https://github.com/openshift/enhancements/pull/463
@Danil-Grigorev can you please add it there?

pkg/types/installconfig.go
```diff
@@ -139,6 +140,10 @@ type InstallConfig struct {
 	// GCP: "Mint", "Passthrough", "Manual"
 	// +optional
 	CredentialsMode CredentialsMode `json:"credentialsMode,omitempty"`
+
+	// FeatureGateSelection allows to enable additional unsupported features.
```
staebler, line 144: I am not keen on putting configuration for unsupported features into the InstallConfig. Is there a compelling reason for why users that want to use feature gates cannot create the manifests themselves?

PR #4312, opened 2020-10-27
bootkube.sh: remove deprecated etcd-member

hexfusion:
we copied tls certs to etcd-member path to allow for quorum-guard to function. As we have migrated quorum-guard to cluster-etcd-namespace this is no longer needed. also removing the deprecated etcd-discovery-domain flag.


hexfusion:
was not thinking straight about this one. :)

PR #4315, opened 2020-10-27
aws: ensure users set ami id for c2s regions

staebler:
As with US Govcloud and China partitions, C2S partitions do not allow copying AMI from commercial region us-east-1. Therefore the users need to provide the AMI for control plane and compute.

https://issues.redhat.com/browse/CORS-1582


mtnbikenc (approved):
/lgtm

PR #4316, opened 2020-10-27
aws: block cluster destroy in c2s region

staebler:
The C2S partition does not have resourcetaggingapi. Consequently, the aws destroyer is not able to find resources to clean up when the cluster is installed in that partition. Rather than making the user wait through numerous timeouts and backoffs, fail the destroy command early when the region used for the cluster is the C2S region.

https://issues.redhat.com/browse/CORS-1581


mtnbikenc (approved):
/lgtm

PR #4317, opened 2020-10-27
bug 1891702: pkg/asset/ignition/bootstrap: exit loop if pem.Decode() fails

jhixson74:
https://bugzilla.redhat.com/show_bug.cgi?id=1891702

### DIFF
--- a/pkg/asset/ignition/bootstrap/bootstrap.go
+++ b/pkg/asset/ignition/bootstrap/bootstrap.go
@@ -579,23 +579,22 @@ func warnIfCertificatesExpired(config *igntypes.Config) {
 			data := decoded.Data
 			for {
 				block, rest := pem.Decode(data)
-				if block != nil {
-					cert, err := x509.ParseCertificate(block.Bytes)
-					if err == nil {
-						if time.Now().UTC().After(cert.NotAfter) {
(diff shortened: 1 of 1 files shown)


pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -579,7 +579,9 @@ func warnIfCertificatesExpired(config *igntypes.Config) {
 			data := decoded.Data
 			for {
 				block, rest := pem.Decode(data)
-				if block != nil {
+				if block == nil {
```
staebler, line 582: This section has become somewhat convoluted with all of the breaks. Can we simplify the `for` loop?
```	
for len(data) > 0 {
	block, rest := pem.Decode(data)
	if block == nil {
		break
	}
	data = rest
	// check cert
}
```

staebler: We actually don't even need the check for the length of the data either, since an empty data will return a nil block from `pem.Decode`.

jhixson74: I've simplified this.

staebler (approved):
/lgtm

PR #4320, opened 2020-10-28
Bug 1888378: [Azure][Destroy] Check if resource group exists

suhanime:
Azure clubs all resources for a cluster within a resource group. The
pattern for delete is to first delete all the resources and then delete
the resource group.

However, when attempting delete, we should first check if the
resource group exists and then try to delete the resources. This is to
ensure delete exits gracefully when resource group wasn't created and/or
deleted before attempting delete.

### DIFF
--- a/pkg/destroy/azure/azure.go
+++ b/pkg/destroy/azure/azure.go
@@ -112,6 +112,17 @@ func (o *ClusterUninstaller) Run() error {
 	waitCtx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
+	// check if resource group exists before attempting delete
+	resourceGroup, err := o.resourceGroupsClient.Get(waitCtx, o.ResourceGroupName)
+	if err != nil {
+		if resourceGroup.Response.StatusCode == http.StatusNotFound {
+			o.Logger.Debug(errors.Wrap(err, "resource group does not exist, nothing to delete"))
+			return nil
+		}
+		o.Logger.Error(err)
+		return err
+	}
+
 	wait.UntilWithContext(
 		waitCtx,
 		func(ctx context.Context) {


pkg/destroy/azure/azure.go
```diff
@@ -112,6 +112,17 @@ func (o *ClusterUninstaller) Run() error {
 	waitCtx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
+	// check if resource group exists before attempting delete
+	resourceGroup, err := o.resourceGroupsClient.Get(waitCtx, o.ResourceGroupName)
+	if err != nil {
+		if resourceGroup.Response.StatusCode == http.StatusNotFound {
+			o.Logger.Debug(errors.Wrap(err, "resource group does not exist, nothing to delete"))
+			return nil
```
jhixson74, line 120: This should probably return err, other than that, LGTM.

jhixson74 (changes requested):

abhinavdahiya:
Don't check for existence of a resource group and decide if any delete need to happen.

There are ans can be resources in other resource groups that need to be cleaned up, like the aad applications. 

Check for no work needed to be done in the error message. For example if we try to delete a resource group and it returns does not exist, treat that error as success.

/hold

suhanime:
> Don't check for existence of a resource group and decide if any delete need to happen.
> 
> There are ans can be resources in other resource groups that need to be cleaned up, like the aad applications.
> 
> Check for no work needed to be done in the error message. For example if we try to delete a resource group and it returns does not exist, treat that error as success.
> 
> /hold

You make a good point.
The current code is trying to delete the public records (like dns zones) first before deleting the resource group. The error in the bug can be traced to when the code tries to list the resources within the resource group, making the assumption that it already exists. To not break the logic, I have moved the code to "deletePublicRecords" func. Code is set to return if it cannot find a resource group simply because the rest of the logic in the method depends upon its existence.

PR #4321, opened 2020-10-28
Deprecate computeFlavor in OpenStack platform

Fedosin:
/label platform/openstack


Fedosin:
/hold

we need to fix flavor validation first: https://github.com/openshift/installer/pull/4289

pkg/types/openstack/platform.go
```diff
@@ -22,7 +22,10 @@ type Platform struct {
 	ExternalNetwork string `json:"externalNetwork,omitempty"`
 
 	// FlavorName is the name of the compute flavor to use for instances in this cluster.
-	FlavorName string `json:"computeFlavor"`
+	// Deprecated: this value is not used by the installer, use type in the defaultMachinePlatform section
+	// to define the flavor.
```
adduarte, line 26: "type" or "flavorName", in func "convertOpenStack" it reads 
config.Platform.OpenStack.DefaultMachinePlatform.FlavorName = config.Platform.OpenStack.FlavorName

pkg/asset/machines/worker.go
```diff
@@ -106,10 +106,9 @@ func defaultGCPMachinePoolPlatform() gcptypes.MachinePool {
 	}
 }
 
-func defaultOpenStackMachinePoolPlatform(flavor string) openstacktypes.MachinePool {
+func defaultOpenStackMachinePoolPlatform() openstacktypes.MachinePool {
 	return openstacktypes.MachinePool{
-		FlavorName: flavor,
-		Zones:      []string{""},
+		Zones: []string{""},
 	}
```
adduarte, line 112: gratutious change between Zones: and [], but is just a nit.

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -1011,7 +1011,7 @@ spec:
                     description: ClusterOSImage is either a URL with `http(s)` or `file` scheme to override the default OS image for cluster nodes, or an existing Glance image name.
                     type: string
                   computeFlavor:
-                    description: FlavorName is the name of the compute flavor to use for instances in this cluster.
+                    description: 'FlavorName is the name of the compute flavor to use for instances in this cluster. Deprecated: this value is not used by the installer, use type in the defaultMachinePlatform section to define the flavor.'
```
adduarte, line 1014: I think you mean "flavorName". in installconfig.go, it reads: "config.Platform.OpenStack.DefaultMachinePlatform.FlavorName = config.Platform.OpenStack.FlavorName"

adduarte (commented):
check the word "type"

pkg/asset/machines/worker.go
mdbooth: I believe go fmt would have made this change automatically.

pkg/types/openstack/platform.go
```diff
@@ -22,7 +22,10 @@ type Platform struct {
 	ExternalNetwork string `json:"externalNetwork,omitempty"`
 
 	// FlavorName is the name of the compute flavor to use for instances in this cluster.
-	FlavorName string `json:"computeFlavor"`
+	// Deprecated: this value is not used by the installer, use type in the defaultMachinePlatform section
```
staebler, line 25: We need to continue to honor any values set in the field. That is the contract of deprecation. The installer cannot just ignore the field.

Fedosin: we continue to honor any values set in this field. to do so I implemented the conversion and related tests: https://github.com/openshift/installer/pull/4321/files#diff-2bfaee7ebe030f992df86bdd39b047cfb12132e99ff453d2ddf4dc97c5d8530fR109-R122

staebler: OK. Good. Can we change the comment to not state that the field is ignored?

EmilienM:
Did you keep the platform flavor validation on purpose?
I would disable it so the flavor is only checked at the machine pool validation level.

EmilienM (changes requested):
See my comment, I would disable the flavor validation at the platform level, and only keep the validation at the machine pool level.

pkg/types/openstack/platform.go
Fedosin: sure! I removed the part where it says that this value is not used by the installer

data/data/install.openshift.io_installconfigs.yaml
Fedosin: done

pkg/types/openstack/platform.go
Fedosin: done

Fedosin:
> Did you keep the platform flavor validation on purpose?
> I would disable it so the flavor is only checked at the machine pool validation level.

yep, because of https://github.com/openshift/installer/pull/4321#issuecomment-718120016

Now that PR is merged and I can remove useless validations and finaly rename the field to DeprecatedFlavorName

iamemilio:
/lgtm

I think this is in shape to merge

pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -23,9 +23,6 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
 	// validate the externalNetwork
 	allErrs = append(allErrs, validateExternalNetwork(p, ci, fldPath)...)
 
-	// validate platform flavor
-	allErrs = append(allErrs, validateFlavor(p.FlavorName, ci, ctrlPlaneFlavorMinimums, fldPath.Child("computeFlavor"), true)...)
```
EmilienM, line 27: FYI I talked to @iamemilio today and we might re-introduce that validation somewhere, for backward compatibility.
Since we still support that parameter, I think we still need to validate it, but we haven't figured out how yet; waiting on this PR to merge first to re-iterate on that topic.

Therefore, I have no strong opinion if whether or not we should remove that validation in that PR or just leave it here and we'll deal with it later with Emilio.

staebler: Validation is done after the up-conversion. There is no need to validate the deprecated field.

EmilienM: this will be done into a different PR: https://github.com/openshift/installer/pull/4383 for easier backport to 4.6.

Fedosin:
/hold

we need to fix https://github.com/openshift/installer/pull/4383 first

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -1098,7 +1098,7 @@ spec:
                     description: ClusterOSImage is either a URL with `http(s)` or `file` scheme to override the default OS image for cluster nodes, or an existing Glance image name.
                     type: string
                   computeFlavor:
```
mandre, line 1100: Can you remove this property from [docs/user/openstack/customization.md](https://github.com/openshift/installer/blob/master/docs/user/openstack/customization.md) too? It's currently marked as required...

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -1098,7 +1098,7 @@ spec:
...
-                    description: FlavorName is the name of the compute flavor to use for instances in this cluster.
+                    description: 'DeprecatedFlavorName is the name of the compute flavor to use for instances in this cluster. Deprecated: use FlavorName in DefaultMachinePlatform to define default compute flavor.'
```
mandre, line 1101: The term `compute` can be confusing, I suggest removing it. 
```suggestion
                    description: 'DeprecatedFlavorName is the name of the flavor to use for instances in this cluster. Deprecated: use FlavorName in DefaultMachinePlatform to define default flavor.'
```

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -1098,7 +1098,7 @@ spec:
...
                     type: string
                   defaultMachinePlatform:
```
mandre, line 1103: I just noticed that defaultMachinePlatform isn't documented in docs/user/openstack/customization.md. We'll need to fix that too.

data/data/install.openshift.io_installconfigs.yaml
iamemilio: make sure @maxwelldb  is aware

maxwelldb: Is the short of this that `flavor` in `platform.openstack.defaultMachinePlatform` replaces `computeFlavor` for 4.7+?

iamemilio: platform.openstack.computeFlavor is now depricated, but still supported.
If you want to set a compute flavor to be used by default for all machine sets, you will now need to set it in platform.openstack.defaultMachinePlatform.flavorName. Otherwise, you should set the flavor in each machine pool

data/data/install.openshift.io_installconfigs.yaml
Fedosin: done

data/data/install.openshift.io_installconfigs.yaml
Fedosin: done!

mandre (changes requested):
Can mark https://github.com/openshift/installer/pull/4321/files#diff-8d13bb62fd33c601f7f64e3a9c861445f547962a0079d20613daab0c07dd73b8R23-R24 as deprecated?

pkg/asset/installconfig/openstack/validation/cloudinfo.go
```diff
@@ -105,13 +105,6 @@ func (ci *CloudInfo) collectInfo(ic *types.InstallConfig) error {
 	}
 
 	// Get flavor info
-	if flavorName := ic.OpenStack.FlavorName; flavorName != "" {
```
staebler, line 108: Should this be replaced with a fetch of the flavor from the DefaultMachinePlatform?

pkg/types/conversion/installconfig.go
```diff
@@ -106,5 +106,19 @@ func convertOpenStack(config *types.InstallConfig) error {
 		}
 	}
 
+	// computeFlavor has been deprecated in favor of type in defaultMachinePlatform.
+	if config.Platform.OpenStack.DeprecatedFlavorName != "" {
+		if config.Platform.OpenStack.DefaultMachinePlatform == nil {
+			config.Platform.OpenStack.DefaultMachinePlatform = &openstack.MachinePool{}
+		}
+
+		if config.Platform.OpenStack.DefaultMachinePlatform.FlavorName != "" {
```
staebler, line 115: [nit] Should we support the user setting both, so long as they are equivalent?

pkg/asset/installconfig/openstack/validation/cloudinfo.go
Fedosin: oh, sure... thanks for noticing this!

Fedosin: I started validating DefaultMachinePlatform here  https://github.com/openshift/installer/pull/4321/files#diff-99d5288a13cd46e2e89659b42a327f77fcb0c806f651085b9458843e767cf767R32-R34
Flavor will be checked as a part of this validation.

docs/user/openstack/customization.md
```diff
@@ -20,7 +20,7 @@ Beyond the [platform-agnostic `install-config.yaml` properties](../customization
 ## Cluster-scoped properties
 
 * `cloud` (required string): The name of the OpenStack cloud to use from `clouds.yaml`.
-* `computeFlavor` (required string): The OpenStack flavor to use for compute and control-plane machines.
+* `computeFlavor` (deprecated string): The OpenStack flavor to use for compute and control-plane machines.
     This is currently required, but has lower precedence than [the `type` property](#machine-pools) on [the `compute` and `controlPlane` machine-pools](../customization.md#platform-customization).
```
mandre, line 24:
```suggestion
    Deprecated: use `type` in DefaultMachinePlatform to define default flavor.
```

Fedosin: yes

mandre (commented):
/lgtm
/retest

staebler (commented):
/approve

PR #4322, opened 2020-10-28
Revert "[Azure][Destroy] Check if resource group exists"

staebler:
Reverts openshift/installer#4320

I concur with the sentiment expressed by @abhinavdahiya [here](https://github.com/openshift/installer/pull/4320#issuecomment-718153041)


staebler:
/cc @jhixson74 
since you gave the offending PR a lgtm

PR #4323, opened 2020-10-29
Bug 1891543: openstack: consider volumes for storage requirements checks

EmilienM:
It is possible to boot a server without using ephemeral storage, with
boot from volume:
https://docs.openstack.org/cinder/latest/admin/blockstorage-boot-from-volume.html

This patch will disable flavor storage check if rootVolume is used in
the Machine Pool and checks that the volume is at least 25Gb sized.

https://bugzilla.redhat.com/show_bug.cgi?id=1891543

### DIFF
--- a/docs/user/openstack/README.md
+++ b/docs/user/openstack/README.md
@@ -82,19 +82,19 @@ Once you configure the quota for your project, please ensure that the user for t
 
 ### Master Nodes
 
-The default deployment stands up 3 master nodes, which is the minimum amount required for a cluster. For each master node you stand up, you will need 1 instance, and 1 port available in your quota. They should be assigned a flavor with at least 16 GB RAM, 4 vCPUs, and 25 GB Disk. It is theoretically possible to run with a smaller flavor, but be aware that if it takes too long to stand up services, or certain essential services crash, the installer could time out, leading to a failed install.
+The default deployment stands up 3 master nodes, which is the minimum amount required for a cluster. For each master node you stand up, you will need 1 instance, and 1 port available in your quota. They should be assigned a flavor with at least 16 GB RAM, 4 vCPUs, and 25 GB Disk (or Root Volume). It is theoretically possible to run with a smaller flavor, but be aware that if it takes too long to stand up services, or certain essential services crash, the installer could time out, leading to a failed install.
 
 The Master Nodes are placed in a single Server Group with "soft anti-affinity" policy; the machines will therefore be creted on separate hosts when possible.
 
 ### Worker Nodes
 
-The default deployment stands up 3 worker nodes. Worker nodes host the applications you run on OpenShift. The flavor assigned to the worker nodes should have at least 2 vCPUs, 8 GB RAM and 25 GB Disk. However, if you are experiencing `Out Of Memory` issues, or your installs are timing out, try increasing the size of your flavor to match the master nodes: 4 vCPUs and 16 GB RAM.
+The default deployment stands up 3 worker nodes. Worker nodes host the applications you run on OpenShift. The flavor assigned to the worker nodes should have at least 2 vCPUs, 8 GB RAM and 25 GB Disk (or Root Volume). However, if you are experiencing `Out Of Memory` issues, or your installs are timing out, try increasing the size of your flavor to match the master nodes: 4 vCPUs and 16 GB RAM.
 
 See the [OpenShift documentation](https://docs.openshift.com/container-platform/4.4/architecture/control-plane.html#defining-workers_control-plane) for more information on the worker nodes.
 
 ### Bootstrap Node
 
-The bootstrap node is a temporary node that is responsible for standing up the control plane on the masters. Only one bootstrap node will be stood up and it will be deprovisioned once the production control plane is ready. To do so, you need 1 instance, and 1 port. We recommend a flavor with a minimum of 16 GB RAM, 4 vCPUs, and 25 GB Disk.
+The bootstrap node is a temporary node that is responsible for standing up the control plane on the masters. Only one bootstrap node will be stood up and it will be deprovisioned once the production control plane is ready. To do so, you need 1 instance, and 1 port. We recommend a flavor with a minimum of 16 GB RAM, 4 vCPUs, and 25 GB Disk (or Root Volume).
 
 ### Image Registry Requirements
 
--- a/pkg/asset/installconfig/openstack/validation/machinepool.go
+++ b/pkg/asset/installconfig/openstack/validation/machinepool.go
@@ -14,16 +14,20 @@ type flavorRequirements struct {
 	RAM, VCPUs, Disk int
 }
 
+const (
+	minimumStorage = 25
+)
+
 var (
 	ctrlPlaneFlavorMinimums = flavorRequirements{
 		RAM:   16,
 		VCPUs: 4,
-		Disk:  25,
+		Disk:  minimumStorage,
 	}
 	computeFlavorMinimums = flavorRequirements{
 		RAM:   8,
 		VCPUs: 2,
-		Disk:  25,
+		Disk:  minimumStorage,
 	}
 )
 
@@ -31,20 +35,24 @@ var (
 func ValidateMachinePool(p *openstack.MachinePool, ci *CloudInfo, controlPlane bool, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
+	var checkStorageFlavor bool
 	// Validate Root Volumes
 	if p.RootVolume != nil {
 		if p.RootVolume.Type == "" {
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume").Child("type"), p.RootVolume.Type, "Volume type must be specified to use root volumes"))
 		}
-		if p.RootVolume.Size <= 0 {
-			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume").Child("size"), p.RootVolume.Size, "Volume size must be greater than zero to use root volumes"))
+		if p.RootVolume.Size < minimumStorage {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume").Child("size"), p.RootVolume.Size, fmt.Sprintf("Volume size must be greater than %d to use root volumes, had %d", minimumStorage, p.RootVolume.Size)))
 		}
+	} else {
+		// Not using root volume, so must check flavor
+		checkStorageFlavor = true
 	}
 
 	if controlPlane {
-		allErrs = append(allErrs, validateFlavor(p.FlavorName, ci, ctrlPlaneFlavorMinimums, fldPath.Child("type"))...)
+		allErrs = append(allErrs, validateFlavor(p.FlavorName, ci, ctrlPlaneFlavorMinimums, fldPath.Child("type"), checkStorageFlavor)...)
 	} else {
-		allErrs = append(allErrs, validateFlavor(p.FlavorName, ci, computeFlavorMinimums, fldPath.Child("type"))...)
+		allErrs = append(allErrs, validateFlavor(p.FlavorName, ci, computeFlavorMinimums, fldPath.Child("type"), checkStorageFlavor)...)
 	}
 
 	allErrs = append(allErrs, validateZones(p.Zones, ci.Zones, fldPath.Child("zones"))...)
@@ -100,7 +108,7 @@ func validUUIDv4(s string) bool {
 
 // validate flavor checks to make sure that a given flavor exists and meets the minimum requrement to run a cluster
 // this function does not validate proper install config usage
-func validateFlavor(flavorName string, ci *CloudInfo, req flavorRequirements, fldPath *field.Path) field.ErrorList {
+func validateFlavor(flavorName string, ci *CloudInfo, req flavorRequirements, fldPath *field.Path, storage bool) field.ErrorList {
 	if flavorName == "" {
 		return nil
 	}
@@ -123,7 +131,7 @@ func validateFlavor(flavorName string, ci *CloudInfo, req flavorRequirements, fl
 	if flavor.VCPUs < req.VCPUs {
 		errs = append(errs, fmt.Sprintf("Must have minimum of %d VCPUs, had %d", req.VCPUs, flavor.VCPUs))
 	}
-	if flavor.Disk < req.Disk {
+	if flavor.Disk < req.Disk && storage {
 		errs = append(errs, fmt.Sprintf("Must have minimum of %d GB Disk, had %d GB", req.Disk, flavor.Disk))
 	}
 
--- a/pkg/asset/installconfig/openstack/validation/machinepool_test.go
+++ b/pkg/asset/installconfig/openstack/validation/machinepool_test.go
@@ -22,12 +22,38 @@ const (
 	invalidCtrlPlaneFlavor = "invalid-control-plane-flavor"
 
 	baremetalFlavor = "baremetal-flavor"
+
+	volumeType      = "performance"
+	volumeSmallSize = 10
+	volumeLargeSize = 25
 )
 
 func validMachinePool() *openstack.MachinePool {
 	return &openstack.MachinePool{
 		FlavorName: validCtrlPlaneFlavor,
 		Zones:      []string{""},
+	}
+}
+
+func invalidMachinePoolSmallVolume() *openstack.MachinePool {
+	return &openstack.MachinePool{
+		FlavorName: validCtrlPlaneFlavor,
+		Zones:      []string{""},
+		RootVolume: &openstack.RootVolume{
+			Type: volumeType,
+			Size: volumeSmallSize,
+		},
+	}
+}
+
+func validMachinePoolLargeVolume() *openstack.MachinePool {
+	return &openstack.MachinePool{
+		FlavorName: validCtrlPlaneFlavor,
+		Zones:      []string{""},
+		RootVolume: &openstack.RootVolume{
+			Type: volumeType,
+			Size: volumeLargeSize,
+		},
 	}
 }
 
@@ -190,6 +216,30 @@ func TestOpenStackMachinepoolValidation(t *testing.T) {
 			mpool: func() *openstack.MachinePool {
 				mp := validMachinePool()
 				mp.FlavorName = baremetalFlavor
+				return mp
+			}(),
+			cloudInfo:      validMpoolCloudInfo(),
+			expectedError:  false,
+			expectedErrMsg: "",
+		},
+		{
+			name:         "volume too small",
+			controlPlane: false,
+			mpool: func() *openstack.MachinePool {
+				mp := invalidMachinePoolSmallVolume()
+				mp.FlavorName = invalidCtrlPlaneFlavor
+				return mp
+			}(),
+			cloudInfo:      validMpoolCloudInfo(),
+			expectedError:  true,
+			expectedErrMsg: "Volume size must be greater than 25 to use root volumes, had 10",
+		},
+		{
+			name:         "volume big enough",
+			controlPlane: false,
+			mpool: func() *openstack.MachinePool {
+				mp := validMachinePoolLargeVolume()
+				mp.FlavorName = invalidCtrlPlaneFlavor
 				return mp
 			}(),
 			cloudInfo:      validMpoolCloudInfo(),
--- a/docs/user/openstack/customization.md
+++ b/docs/user/openstack/customization.md
@@ -40,7 +40,7 @@ Beyond the [platform-agnostic `install-config.yaml` properties](../customization
 * `additionalSecurityGroupIDs` (optional list of strings): IDs of additional security groups for machines.
 * `type` (optional string): The OpenStack flavor name for machines in the pool.
 * `rootVolume` (optional object): Defines the root volume for instances in the machine pool. The instances use ephemeral disks if not set.
-  * `size` (required integer): Size of the root volume in GB.
+  * `size` (required integer): Size of the root volume in GB. Must be set to at least 25.
   * `type` (required string): The volume pool to create the volume from.
 * `zones` (optional list of strings): The names of the availability zones you want to install your nodes on. If unset, the installer will use your default compute zone.
 
--- a/pkg/asset/installconfig/openstack/validation/platform.go
+++ b/pkg/asset/installconfig/openstack/validation/platform.go
@@ -24,7 +24,7 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
 	allErrs = append(allErrs, validateExternalNetwork(p, ci, fldPath)...)
 
 	// validate platform flavor
-	allErrs = append(allErrs, validateFlavor(p.FlavorName, ci, ctrlPlaneFlavorMinimums, fldPath.Child("computeFlavor"))...)
+	allErrs = append(allErrs, validateFlavor(p.FlavorName, ci, ctrlPlaneFlavorMinimums, fldPath.Child("computeFlavor"), true)...)
 
 	// validate floating ips
 	allErrs = append(allErrs, validateFloatingIPs(p, ci, fldPath)...)


EmilienM:
Here is a quick demo of the results:

[asciicast](https://asciinema.org/a/J8ngtxu4J1mVHvvC7hiiIxaLO)

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -119,8 +119,13 @@ func validateFlavor(flavor Flavor, req flavorRequirements, fldPath *field.Path)
 	if flavor.VCPUs < req.VCPUs {
 		errs = append(errs, fmt.Sprintf("Must have minimum of %d VCPUs, had %d", req.VCPUs, flavor.VCPUs))
 	}
-	if flavor.Disk < req.Disk {
-		errs = append(errs, fmt.Sprintf("Must have minimum of %d GB Disk, had %d GB", req.Disk, flavor.Disk))
+	// Here we tolerate the minimum storage requirements to come from
+	// Disk or Ephemeral.
+	// The root disk is an ephemeral disk that the base image is copied into
+	// whereas an ephemeral disk offers to the machine a local disk storage
+	// linked to the lifecycle of a VM instance.
```
adduarte, line 126: trhough explanation, but I don't think we need to leave the comment in the code.  The code below is pretty clear, and adding the note just creates the risk of the comment becoming stale. Anyone changing the code can easily forget to edit the comment.

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -119,8 +119,13 @@ func validateFlavor(flavor Flavor, req flavorRequirements, fldPath *field.Path)
...
+	if flavor.Disk < req.Storage && flavor.Ephemeral < req.Storage {
+		errs = append(errs, fmt.Sprintf("Must have minimum of %d GB Disk, had %d GB for root disk and %d GB for ephemeral", req.Storage, flavor.Disk, flavor.Ephemeral))
```
adduarte, line 128: Maybe rework the error message. right now it doesn't read well.  also is it "and" GB for ephemereal or "or"?. The code says "or" (the inverse of &&)

pkg/asset/installconfig/openstack/validation/machinepool_test.go
```diff
@@ -13,65 +13,72 @@ import (
...
 				Flavor: &flavors.Flavor{
-					Name:  baremetalFlavor,
-					RAM:   8,  // too low
-					Disk:  10, // too low
-					VCPUs: 2,  // too low
+					Name:      baremetalFlavor,
+					RAM:       8,  // too low
+					Disk:      10, // too low
+					VCPUs:     2,  // too low
+					Ephemeral: 0,
```
adduarte, line 81: you don't technically need this.  Ephemeral is an int defined in the flavor.Flavor struct,  it will be initialized to 0 if you don't provide it, when isntantiating  a flavors.Flavor struct.

adduarte (commented):
just a couple of comments

pierreprinetti (commented):
@EmilienM 
I think that you are not addressing the problem reported in the bug.

To my understanding, the bug report says:

---

**Do not check the flavor disk size when I set OCP nodes to boot from volume.** When I [pass a volume](https://github.com/openshift/installer/blob/master/docs/user/openstack/customization.md#machine-pools) in an `install-config.yaml` [machine-pool](https://github.com/openshift/installer/blob/b0c6ae554ed2755d0012fc28839e0060d6cdf722/pkg/types/openstack/machinepool.go#L13), the disk capacity of the flavor is irrelevant, and the volume size should be checked instead.

---

The interesting part though, is that you are addressing the case of the ephemeral disks which we haven't considered so far. This might be worth some investigation; however, I suspect that we would need to write some new code to be able to automatically use ephemeral storage (as in "OS-FLV-EXT-DATA:ephemeral") for our purposes (ie: is it automatically mounted? To which mountpoint?).

ppeereb1:
I'm also running into this when bootstrapping a new 4.6.1 cluster. Unfortunately our cloud has both ephemeral and root disk size to 0.
```
paul@paul-ThinkPad-X1-Carbon-6th:~$ openstack flavor show m1.4large
+----------------------------+--------------------------------------+
| Field                      | Value                                |
+----------------------------+--------------------------------------+
| OS-FLV-DISABLED:disabled   | False                                |
| OS-FLV-EXT-DATA:ephemeral  | 0                                    |
| access_project_ids         | None                                 |
| disk                       | 0                                    |
| id                         | 6d14ea74-f17d-43d2-84a9-fbdfa12561a4 |
| name                       | m1.4large                            |
| os-flavor-access:is_public | True                                 |
| properties                 | ostype='generic'                     |
| ram                        | 16384                                |
| rxtx_factor                | 1.0                                  |
| swap                       |                                      |
| vcpus                      | 4                                    |
+----------------------------+--------------------------------------+
````

They are expecting you're using a cinder volume for booting a vm. You can create a separate root volume when you specify the rootVolume parameter in install-config.yaml: https://github.com/openshift/installer/blob/master/docs/user/openstack/customization.md#custom-machine-pools 

It would be nice if the disk check is dropped when specifying a rootVolume parameter in the install-config.yaml

**Comment by EmilienM** [comment]
[...]
> They are expecting you're using a cinder volume for booting a vm. You can create a separate root volume when you specify the rootVolume parameter in install-config.yaml: https://github.com/openshift/installer/blob/master/docs/user/openstack/customization.md#custom-machine-pools
> 
> It would be nice if the disk check is dropped when specifying a rootVolume parameter in the install-config.yaml

Ok, I see what the problem is now. Thanks for taking the time to explain. I'll separate my original commit from the next one I'll submit to fix your use case, and submit it via a single PR as it's the same theme.

File: pkg/asset/installconfig/openstack/validation/machinepool.go (hunk 1 above)
Comment (EmilienM) [reply pkg/asset/installconfig/openstack/validation/machinepool.go:126]: Fair enough, I'll follow your judgement and remove the comment.

File: pkg/asset/installconfig/openstack/validation/machinepool.go (hunk 2 above)
Comment (EmilienM) [reply pkg/asset/installconfig/openstack/validation/machinepool.go:128]: I think the message is fine:

```
Must have minimum of 25 GB Disk, had 0 GB for root disk and 15 GB for ephemeral
```
In case of not meeting the requirements, we present what the flavor offers. How would you reword it?

File: pkg/asset/installconfig/openstack/validation/machinepool_test.go (hunk 3 above)
Comment (EmilienM) [reply pkg/asset/installconfig/openstack/validation/machinepool_test.go:81]: ok, I'll remove them.

File: pkg/asset/installconfig/openstack/validation/machinepool.go [hunk 4]
```diff
@@ -119,8 +126,8 @@ func validateFlavor(flavor Flavor, req flavorRequirements, fldPath *field.Path)
 	if flavor.VCPUs < req.VCPUs {
 		errs = append(errs, fmt.Sprintf("Must have minimum of %d VCPUs, had %d", req.VCPUs, flavor.VCPUs))
 	}
-	if flavor.Disk < req.Disk {
-		errs = append(errs, fmt.Sprintf("Must have minimum of %d GB Disk, had %d GB", req.Disk, flavor.Disk))
+	if flavor.Disk < req.Storage && flavor.Ephemeral < req.Storage {
```
EmilienM, line 129: I need to change the condition here as well, I'll fix it today. Please don't review that line yet.

pkg/asset/installconfig/openstack/validation/machinepool.go
EmilienM: @adduarte I improved the message, please let me know if it's ok now. Thanks

EmilienM:
> The interesting part though, is that you are addressing the case of the ephemeral disks which we haven't considered so far. This might be worth some investigation; however, I suspect that we would need to write some new code to be able to automatically use ephemeral storage (as in "OS-FLV-EXT-DATA:ephemeral") for our purposes (ie: is it automatically mounted? To which mountpoint?).

We don't need to check for the OS-FLV-EXT-DATA:ephemeral, the Flavor already provides the Ephemeral parameter: https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/flavors/requests.go#L130

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -33,21 +35,25 @@ func ValidateMachinePool(p *openstack.MachinePool, ci *CloudInfo, controlPlane b
...
 	// Validate Root Volumes
 	if p.RootVolume != nil {
+		// No need to check the storage in the Flavor since RootVolume is used
+		checkStorageFlavor = false
 		if p.RootVolume.Type == "" {
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume").Child("type"), p.RootVolume.Type, "Volume type must be specified to use root volumes"))
 		}
-		if p.RootVolume.Size <= 0 {
-			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume").Child("size"), p.RootVolume.Size, "Volume size must be greater than zero to use root volumes"))
+		if p.RootVolume.Size <= minimumStorage {
```
pierreprinetti, line 43: why `<=` instead of `<`? Isn't it OK to boot from a volume with size exactly 25GB?

(also below, "at least" instead of "greater than")

pierreprinetti:
>We don't need to check for the OS-FLV-EXT-DATA:ephemeral, the Flavor already provides the Ephemeral parameter

True, Gophercloud parses that data.

My concern around ephemeral storage is due to the fact that I don't know how it works from the client perspective. Depending on how it works, our installer might require changes in order to use it. For example, I can suppose that the ephemeral disk is treated as a separate device on the VM.

---

Let's take the example of a flavor with `rootSize == 10` and `OS-FLV-EXT-DATA:ephemeral == 100`. No volume attached.

Before this patch, this configuration would be refused. We only proceed with installation when we can use the root disk. Instead, with this patch, we accept this configuration in the pre-flight validation.

My doubts (_because I don't know how Nova ephemeral storage works_):

* **Is the ephemeral storage automatically attached to the VM, or does it have to be explicitly requested upon machine creation?**

If the latter is true, then we have to change our Terraform code to explicitly request that storage upon machine creation. Otherwise, we are failing to install on the 10GB root disk.

* **Is the ephemeral disk automatically mounted in the VM, or do we have to mount it?**

Again: if mounting is required, we need to add the `mount` command in the machine's Ignition file probably. And make sure that every service in the machine stores data under the mountpoint, because the root disk might be full or readonly.

---

What I am saying is that if we accept `OS-FLV-EXT-DATA:ephemeral` as the source of storage, we must be sure we support using it. Perhaps you know the answer to my two questions already?

iamemilio:
To clarify, this issue is about validating that the installation has at least a minimum of 25 GB Storage for the root user. That can either be ephemeral storage or a root volume. The ephemeral storage is defined by the flavor the user selected, so we already have that information. We don't offer customers the option to attach additional ephemeral storage, so I wouldn't worry about that. The root volume is defined in the [openstack machine pool spec](https://github.com/openshift/installer/blob/6b2861cd4bead8e8e290b51fab8187d3a0d1abb5/pkg/types/openstack/machinepool.go#L13). So you may have to make a gophercloud query to GET that and check its size. [gophercloud volume docs](https://godoc.org/github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes). Ephemeral disk is always mounted to the instance, since it is defined in the flavor. Your goal is to check if the root disk || the ephemeral disk >= 25.

pkg/asset/installconfig/openstack/validation/machinepool.go
EmilienM: that's what happen when coding late !

docs/user/openstack/README.md
```diff
@@ -25,13 +25,19 @@ In addition, it covers the installation with the default CNI (OpenShiftSDN), as
     - [Initial Setup](#initial-setup)
     - [API Access](#api-access)
       - [Using Floating IPs](#using-floating-ips)
+        - [Create API and Ingress Floating IP Addresses](#create-api-and-ingress-floating-ip-addresses)
+        - [Create API and Ingress DNS Records](#create-api-and-ingress-dns-records)
+        - [External API Access](#external-api-access)
+        - [External Ingress (apps) Access](#external-ingress-apps-access)
```
EmilienM, line 31: oops it sounds like my IDE made that change for me. I'll revert back.

iamemilio, line 31: It looks like your IDE added some entries to the Table of contents. Personally, I dont mind this, but I have been asked to remove these in past reviews.

docs/user/openstack/README.md
```diff
@@ -25,13 +25,19 @@ In addition, it covers the installation with the default CNI (OpenShiftSDN), as
...
       - [Without Floating IPs](#without-floating-ips)
     - [Running a Deployment](#running-a-deployment)
     - [Current Expected Behavior](#current-expected-behavior)
     - [Checking Cluster Status](#checking-cluster-status)
     - [Destroying The Cluster](#destroying-the-cluster)
   - [Post Install Operations](#post-install-operations)
     - [Adding a MachineSet](#adding-a-machineset)
+      - [Using a Server Group](#using-a-server-group)
+      - [Setting Nova Availability Zones](#setting-nova-availability-zones)
```
iamemilio, line 40: see ^

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -33,21 +35,25 @@ func ValidateMachinePool(p *openstack.MachinePool, ci *CloudInfo, controlPlane b
...
 	// Validate Root Volumes
 	if p.RootVolume != nil {
+		// No need to check the storage in the Flavor since RootVolume is used
+		checkStorageFlavor = false
 		if p.RootVolume.Type == "" {
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume").Child("type"), p.RootVolume.Type, "Volume type must be specified to use root volumes"))
 		}
-		if p.RootVolume.Size <= 0 {
-			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume").Child("size"), p.RootVolume.Size, "Volume size must be greater than zero to use root volumes"))
+		if p.RootVolume.Size < minimumStorage {
```
iamemilio, line 43:
```suggestion
		if p.RootVolume.Size <= minimumStorage {
```

EmilienM: I think we want to authorize 25GB?

pkg/asset/installconfig/openstack/validation/machinepool_test.go
```diff
@@ -31,6 +35,28 @@ func validMachinePool() *openstack.MachinePool {
 	}
 }
 
+func validMachinePoolSmallVolume() *openstack.MachinePool {
```
iamemilio, line 38:
```suggestion
func invalidMachinePoolSmallVolume() *openstack.MachinePool {
```
This case is meant to cause an error, so it should be invalid

pkg/asset/installconfig/openstack/validation/machinepool.go
EmilienM: I can remove that comment, since I didn't touch that code at the end.

docs/user/openstack/README.md
EmilienM: yeah I submitted another PR: https://github.com/openshift/installer/pull/4328

pkg/asset/installconfig/openstack/validation/machinepool.go
iamemilio: Yeah 25 GB is the minimum acceptable storage size

iamemilio: Oh, my mistake

iamemilio: This is fine

iamemilio:
Looks good, great work!
/lgtm


/cc @maxwelldb 
This may require some minor updates to the downstream docs

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -15,15 +15,17 @@ type flavorRequirements struct {
 }
 
 var (
+	checkStorageFlavor      bool
```
pierreprinetti, line 18: Please move the declaration of `checkStorageFlavor` in the context of `ValidateMachinePool` or we'll run into data races if we make those checks concurrent.

pierreprinetti (approved):
Thanks!

pierreprinetti:
/lgtm
/hold

Feel free to `/hold cancel` when CI is happy!

/honk

mandre (commented):
/lgtm
/retest

ppeereb1:
Also if flavor does not exists I get a nil pointer dereference, maybe there should be check build in for this aswelll?
```
STDERR:
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x2208910]
goroutine 1 [running]:
github.com/openshift/installer/pkg/asset/installconfig/openstack/validation.validatePlatformFlavor(0xc00075c540, 0xc000f851d0, 0xc000db4510, 0x0, 0x0, 0x0)
	/go/src/github.com/openshift/installer/pkg/asset/installconfig/openstack/validation/platform.go:83 +0xb0
github.com/openshift/installer/pkg/asset/installconfig/openstack/validation.ValidatePlatform(0xc00075c540, 0xc0007c4240, 0xc000f851d0, 0x0, 0x0, 0x0)
	/go/src/github.com/openshift/installer/pkg/asset/installconfig/openstack/validation/platform.go:25 +0x1ed
github.com/openshift/installer/pkg/asset/installconfig/openstack.Validate(0xc0005a9200, 0x0, 0x0)
	/go/src/github.com/openshift/installer/pkg/asset/installconfig/openstack/validate.go:19 +0x8e
github.com/openshift/installer/pkg/asset/installconfig.(*InstallConfig).platformValidation(0xc000739860, 0x0, 0x0)
	/go/src/github.com/openshift/installer/pkg/asset/installconfig/installconfig.go:188 +0x87
github.com/openshift/installer/pkg/asset/installconfig.(*InstallConfig).finish(0xc000739860, 0xd4f01cb, 0x13, 0x1142, 0xd216b00)
	/go/src/github.com/openshift/installer/pkg/asset/installconfig/installconfig.go:148 +0x234
github.com/openshift/installer/pkg/asset/installconfig.(*InstallConfig).Load(0xc000739860, 0xe7c97e0, 0xc000767180, 0xe8be2a0, 0xc000739860, 0xc000739860)
	/go/src/github.com/openshift/installer/pkg/asset/installconfig/installconfig.go:125 +0x1f0
github.com/openshift/installer/pkg/asset/store.(*storeImpl).load(0xc000781110, 0xe7f3f20, 0xc0007394c0, 0xc00003dc2c, 0x4, 0xc00003dc2c, 0x4, 0x0)
	/go/src/github.com/openshift/installer/pkg/asset/store/store.go:264 +0x455
github.com/openshift/installer/pkg/asset/store.(*storeImpl).load(0xc000781110, 0xe7f3ee0, 0xc000739440, 0xd4838c4, 0x2, 0xd4838c4, 0x2, 0x0)
	/go/src/github.com/openshift/installer/pkg/asset/store/store.go:247 +0x2d7
github.com/openshift/installer/pkg/asset/store.(*storeImpl).load(0xc000781110, 0xe7f4260, 0x154c3500, 0x0, 0x0, 0x2, 0x2, 0xe740e60)
	/go/src/github.com/openshift/installer/pkg/asset/store/store.go:247 +0x2d7
github.com/openshift/installer/pkg/asset/store.(*storeImpl).fetch(0xc000781110, 0xe7f4260, 0x154c3500, 0x0, 0x0, 0x40b4e5, 0xbf8cde0)
	/go/src/github.com/openshift/installer/pkg/asset/store/store.go:201 +0xa3a
github.com/openshift/installer/pkg/asset/store.(*storeImpl).Fetch(0xc000781110, 0xe7f4260, 0x154c3500, 0x1547f8e0, 0x4, 0x4, 0x1100000000000000, 0xed7320de1)
	/go/src/github.com/openshift/installer/pkg/asset/store/store.go:77 +0x4b
main.runTargetCmd.func1(0x7ffff3e1fe06, 0x30, 0xc0007393a0, 0xc000780f00)
	/go/src/github.com/openshift/installer/cmd/openshift-install/create.go:172 +0x135
main.runTargetCmd.func2(0x1548b280, 0xc000739160, 0x0, 0x2)
	/go/src/github.com/openshift/installer/cmd/openshift-install/create.go:199 +0xb5
github.com/spf13/cobra.(*Command).execute(0x1548b280, 0xc000739100, 0x2, 0x2, 0x1548b280, 0xc000739100)
	/go/src/github.com/openshift/installer/vendor/github.com/spf13/cobra/command.go:844 +0x2c2
github.com/spf13/cobra.(*Command).ExecuteC(0xc000524000, 0xc000df5e58, 0x1, 0x1)
	/go/src/github.com/openshift/installer/vendor/github.com/spf13/cobra/command.go:945 +0x336
github.com/spf13/cobra.(*Command).Execute(...)
	/go/src/github.com/openshift/installer/vendor/github.com/spf13/cobra/command.go:885
main.installerMain()
	/go/src/github.com/openshift/installer/cmd/openshift-install/main.go:63 +0x2b8
main.main()
	/go/src/github.com/openshift/installer/cmd/openshift-install/main.go:43 +0xcf
```

EmilienM:
> Also if flavor does not exists I get a nil pointer dereference, maybe there should be check build in for this aswelll?
> 
> ```
> STDERR:
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x2208910]

This sounds like a bug that should be fixed in a separated PR. Could you please file a BZ? I'll investigate that one and take care of it.

ppeereb1:
> > Also if flavor does not exists I get a nil pointer dereference, maybe there should be check build in for this aswelll?
> > ```
> > STDERR:
> > panic: runtime error: invalid memory address or nil pointer dereference
> > [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x2208910]
> > ```
> 
> This sounds like a bug that should be fixed in a separated PR. Could you please file a BZ? I'll investigate that one and take care of it.

Agree, created a seperate issue for it: https://github.com/openshift/installer/issues/4333

mandre:
> > Also if flavor does not exists I get a nil pointer dereference, maybe there should be check build in for this aswelll?
> > ```
> > STDERR:
> > panic: runtime error: invalid memory address or nil pointer dereference
> > [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x2208910]
> > ```
> 
> This sounds like a bug that should be fixed in a separated PR. Could you please file a BZ? I'll investigate that one and take care of it.

PR at https://github.com/openshift/installer/pull/4289 should address this issue (with the [related bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1887863)).

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -15,39 +15,45 @@ type flavorRequirements struct {
 }
 
 var (
+	minimumStorage          = 25
```
mdbooth, line 18: nit: this could be a const.

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -15,39 +15,45 @@ type flavorRequirements struct {
...
-		Disk:  25,
+		Disk:  minimumStorage,
 	}
 )
 
 // ValidateMachinePool checks that the specified machine pool is valid.
 func ValidateMachinePool(p *openstack.MachinePool, ci *CloudInfo, controlPlane bool, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
+	var checkStorageFlavor bool
```
mdbooth, line 35: This is both trivial and highly subjective! Writing this only for the thought process because we're both relatively new to golang :) I'm not asking for a change.

checkStorageFlavor is defined to be false here, so the assignment to false on L39 is technically redundant. However, it's explicitness is nice. I might have written:

```
checkStorageFlavor := false
if p.RootVolume != nil {
  ...
} else {
  // Not using root volume, so must check flavor
  checkStorageVolume = true
}
```

mdbooth (approved):
Incidentally, a flavor with a root disk size of 0 would also work by default, but that's a weird edge case which we should continue to ignore:

https://docs.openstack.org/nova/latest/user/flavors.html

mandre (commented):
/lgtm

pierreprinetti:
/hold cancel
:tada:

EmilienM:
We need https://github.com/openshift/installer/pull/4344 to land before backporting this one

PR #4325, opened 2020-10-29
Bug 1888378: Ignore error if resource group already deleted 

suhanime:
In Azure, while trying to delete public and private DNS zones,
ignore error if resource group was already deleted. This will
ensure that if the resource group wasn't created, or if it was
deleted before attempting destroy of cluster, it doesn't get
stuck in a loop.

### DIFF
--- a/pkg/destroy/azure/azure.go
+++ b/pkg/destroy/azure/azure.go
@@ -205,6 +205,10 @@ func deletePublicRecords(ctx context.Context, dnsClient dns.ZonesClient, records
 
 	zonesPage, err := dnsClient.ListByResourceGroup(ctx, rgName, to.Int32Ptr(100))
 	if err != nil {
+		if zonesPage.Response().IsHTTPStatus(http.StatusNotFound) {
+			logger.Debug("already deleted")
+			return utilerrors.NewAggregate(errs)
+		}
 		errs = append(errs, errors.Wrap(err, "failed to list dns zone"))
 		if isAuthError(err) {
 			return err
@@ -232,6 +236,10 @@ func deletePublicRecords(ctx context.Context, dnsClient dns.ZonesClient, records
 
 	privateZonesPage, err := privateDNSClient.ListByResourceGroup(ctx, rgName, to.Int32Ptr(100))
 	if err != nil {
+		if privateZonesPage.Response().IsHTTPStatus(http.StatusNotFound) {
+			logger.Debug("already deleted")
+			return utilerrors.NewAggregate(errs)
+		}
 		errs = append(errs, errors.Wrap(err, "failed to list private dns zone"))
 		if isAuthError(err) {
 			return err


suhanime:
/cc @abhinavdahiya 
In response to your [comment](https://github.com/openshift/installer/pull/4320#issuecomment-718153041) I have made the relevant changes so it does not break the logic. Simply checking for existence of resource group before listing the resources should ensure the bug is resolved

pkg/destroy/azure/azure.go
```diff
@@ -193,10 +193,22 @@ func (o *ClusterUninstaller) Run() error {
...
 
+	// check if resource group exists
+	resourceGroup, err := rgClient.Get(ctx, rgName)
+	if err != nil {
+		if resourceGroup.Response.StatusCode == http.StatusNotFound {
+			logger.Debug("resource group already deleted, no records to delete")
+			return nil
+		}
+		return err
+	}
```
abhinavdahiya, line 210: instead of making another call, check for no-work error here https://github.com/openshift/installer/pull/4325/files#diff-a2fdf21d9b56bf657a0251ad81376851d27e5b801510337c8f5c9d2cd6495885R216-R221

suhanime: I think the link points to this commit diff.. Can you tell me where to check?

suhanime: Addressed.
There is a type mismatch to use `wasNotFound` here, but I found an inbuilt method that essentially does the same thing

pkg/destroy/azure/azure.go
```diff
@@ -205,7 +205,10 @@ func deletePublicRecords(ctx context.Context, dnsClient dns.ZonesClient, records
 
 	zonesPage, err := dnsClient.ListByResourceGroup(ctx, rgName, to.Int32Ptr(100))
 	if err != nil {
-		errs = append(errs, errors.Wrap(err, "failed to list dns zone"))
+		// ignore error if resource group already deleted
+		if !zonesPage.Response().IsHTTPStatus(http.StatusNotFound) {
+			errs = append(errs, errors.Wrap(err, "failed to list dns zone"))
+		}
```
abhinavdahiya, line 211: for consistency i think we should continue to reuse existing way like https://github.com/openshift/installer/blob/fa6a6b12dbf32441e978ac895cdc4cbd3ea99562/pkg/destroy/azure/azure.go#L379-L381

suhanime: It makes sense to log and return here - so addressed.

pkg/destroy/azure/azure.go
```diff
@@ -205,6 +205,10 @@ func deletePublicRecords(ctx context.Context, dnsClient dns.ZonesClient, records
 
 	zonesPage, err := dnsClient.ListByResourceGroup(ctx, rgName, to.Int32Ptr(100))
 	if err != nil {
+		if zonesPage.Response().IsHTTPStatus(http.StatusNotFound) {
+			logger.Debug("already deleted")
+			return nil
```
staebler, line 210:
```suggestion
			return utilerrors.NewAggregate(errs)
```

pkg/destroy/azure/azure.go
```diff
@@ -232,6 +236,10 @@ func deletePublicRecords(ctx context.Context, dnsClient dns.ZonesClient, records
 
 	privateZonesPage, err := privateDNSClient.ListByResourceGroup(ctx, rgName, to.Int32Ptr(100))
 	if err != nil {
+		if privateZonesPage.Response().IsHTTPStatus(http.StatusNotFound) {
+			logger.Debug("already deleted")
+			return nil
```
staebler, line 241:
```suggestion
			return utilerrors.NewAggregate(errs)
```
otherwise any errors collected earlier in the function will be swallowed

pkg/destroy/azure/azure.go
```diff
@@ -205,6 +205,10 @@ func deletePublicRecords(ctx context.Context, dnsClient dns.ZonesClient, records
 
 	zonesPage, err := dnsClient.ListByResourceGroup(ctx, rgName, to.Int32Ptr(100))
 	if err != nil {
+		if zonesPage.Response().IsHTTPStatus(http.StatusNotFound) {
+			logger.Debug("already deleted")
+			return nil
+		}
 		errs = append(errs, errors.Wrap(err, "failed to list dns zone"))
 		if isAuthError(err) {
 			return err
```
staebler, line 214: On a side note, this should also be changed, but not within the scope of your PR.
```suggestion
			return utilerrors.NewAggregate(errs)
```

pkg/destroy/azure/azure.go
suhanime: Nice catch - addressed

staebler (approved):
/lgtm

PR #2186, opened 2019-08-08
Bug 1737660: data/azure/master: use ReadOnly caching for OSDisk

abhinavdahiya:
In search for find the performance bottleneck for etcd on control-plane host, I contacted the Microsoft Support with the FIO results. They mentioned that the Disks for the control-plane were using Read/Write caches and for Premium SSDs certain applications can benefit from switching to Read-only cache.

`Azure Red Hat Openshift` deployments already use Read-only cache for etcd data disks [1][1]

Switching to read-only has provided definite speed ups. This is still not comparable to AWS/GCP but I think this is definitely required as we optimize more.

[1]: https://github.com/openshift/openshift-azure/blob/ced59421d8f8122315d5b649ba89c43b52c5e183/pkg/arm/v5/resources.go#L475

PR #4326, opened 2020-10-30
use 'None' caching for OSDisk

seasonny:
use 'None' caching for OSDisk

Changed Host Caching to ‘None’, throughput will increase to 192 MB/s, 
it will reduce the chance of throttling

### DIFF
--- a/data/data/azure/master/master.tf
+++ b/data/data/azure/master/master.tf
@@ -104,7 +104,7 @@ resource "azurerm_linux_virtual_machine" "master" {
 
   os_disk {
     name                 = "${var.cluster_id}-master-${count.index}_OSDisk" # os disk name needs to match cluster-api convention
-    caching              = "ReadOnly"
+    caching              = "None"
     storage_account_type = var.os_volume_type
     disk_size_gb         = var.os_volume_size
   }


seasonny:
We set “Host Caching” to Read-Only on master nodes(Standard_D8s_v3) for better IOPS and low latency, here we can find some discussion[1], however, we met a throttling event a few days ago, this event cause ‘etcdHighNumberOfLeaderChanges’ issue, when this alerting was fired the throughput was around 140 MB/s and IOPS is around 1000, but the max cached is 128 MB/s on this type of VM, so it was throttled by Azure, according to Azure’s recommendations, they say it might be changed Host Caching to ‘None’, throughput will increase to 192 MB/s, but the IOPS will drop from 16000 to 12800.

[1] https://github.com/openshift/installer/pull/2186
[2] https://docs.microsoft.com/en-us/azure/virtual-machines/dv3-dsv3-series#dsv3-series

hexfusion:
I think it's worthwhile to review it was technically never tested. Unfortunate that it has a tradeoff on IOPS.

kwoodson:
I am currently benchmarking these settings and came across this mention in the [documentation](https://docs.microsoft.com/en-us/azure/virtual-machines/premium-storage-performance#disk-caching):
```
Currently, None is only supported on data disks. It is not supported on OS disks. If you set None on an OS disk it will override this internally and set it to ReadOnly.
```
I am reaching out to MSFT to verify this is the case. If so, then we can close as this is not supported on the OS disk.

kwoodson:
Confirmed through MSFT that it is not possible to set `None` on the OS disk.

PR #4329, opened 2020-10-30
asset/installconfig/gcp: Validate install-config instance types

jstuever:
Ensure the user provided instance types meets minimum resource
requirements.

### DIFF
--- a/pkg/asset/installconfig/gcp/client.go
+++ b/pkg/asset/installconfig/gcp/client.go
@@ -19,6 +19,7 @@ import (
 // API represents the calls made to the API.
 type API interface {
 	GetNetwork(ctx context.Context, network, project string) (*compute.Network, error)
+	GetMachineTypes(ctx context.Context, project, filter string) (map[string]*compute.MachineType, error)
 	GetPublicDomains(ctx context.Context, project string) ([]string, error)
 	GetPublicDNSZone(ctx context.Context, project, baseDomain string) (*dns.ManagedZone, error)
 	GetSubnetworks(ctx context.Context, network, project, region string) ([]*compute.Subnetwork, error)
@@ -46,6 +47,36 @@ func NewClient(ctx context.Context) (*Client, error) {
 		ssn: ssn,
 	}
 	return client, nil
+}
+
+// GetMachineTypes uses the GCP Compute Service API to get a list of machine types from a project.
+func (c *Client) GetMachineTypes(ctx context.Context, project, filter string) (map[string]*compute.MachineType, error) {
+	types := map[string]*compute.MachineType{}
+
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+
+	svc, err := c.getComputeService(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	req := svc.MachineTypes.AggregatedList(project)
+	if filter != "" {
+		req = req.Filter(filter)
+	}
+
+	if err := req.Pages(ctx, func(page *compute.MachineTypeAggregatedList) error {
+		for _, scopedList := range page.Items {
+			for _, item := range scopedList.MachineTypes {
+				types[item.Name] = item
+			}
+		}
+		return nil
+	}); err != nil {
+		return nil, errors.Wrapf(err, "failed to get machine types from project %s", project)
+	}
+	return types, nil
 }
 
 // GetNetwork uses the GCP Compute Service API to get a network by name from a project.
--- a/pkg/asset/installconfig/gcp/validation.go
+++ b/pkg/asset/installconfig/gcp/validation.go
@@ -17,14 +17,89 @@ import (
 	"github.com/openshift/installer/pkg/types"
 )
 
+type resourceRequirements struct {
+	minimumVCpus  int64
+	minimumMemory int64
+}
+
+var controlPlaneReq = resourceRequirements{
+	minimumVCpus:  4,
+	minimumMemory: 15360,
+}
+
+var computeReq = resourceRequirements{
+	minimumVCpus:  2,
+	minimumMemory: 7680,
+}
+
 // Validate executes platform-specific validation.
 func Validate(client API, ic *types.InstallConfig) error {
 	allErrs := field.ErrorList{}
 
 	allErrs = append(allErrs, validateProject(client, ic, field.NewPath("platform").Child("gcp"))...)
 	allErrs = append(allErrs, validateNetworks(client, ic, field.NewPath("platform").Child("gcp"))...)
+	allErrs = append(allErrs, validateInstanceTypes(client, ic)...)
 
 	return allErrs.ToAggregate()
+}
+
+// ValidateInstanceType ensures the instance type has sufficient Vcpu and Memory.
+func ValidateInstanceType(client API, ic *types.InstallConfig, fieldPath *field.Path, instanceType string, req resourceRequirements) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if instanceType == "" {
+		return nil
+	}
+
+	filter := fmt.Sprintf("name = %s", instanceType)
+	instanceTypes, err := client.GetMachineTypes(context.TODO(), ic.GCP.ProjectID, filter)
+	if err != nil {
+		return append(allErrs, field.InternalError(fieldPath, err))
+	}
+
+	if typeMeta, ok := instanceTypes[instanceType]; ok {
+		if typeMeta.GuestCpus < req.minimumVCpus {
+			errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d vCPUs", req.minimumVCpus)
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+		}
+		if typeMeta.MemoryMb < req.minimumMemory {
+			errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d MB Memory", req.minimumMemory)
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+		}
+	} else {
+		errMsg := fmt.Sprintf("instance type %s not found", instanceType)
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+	}
+
+	return allErrs
+}
+
+// validateInstanceTypes checks that the user-provided instance types are valid.
+func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList {
+	allErrs := field.ErrorList{}
+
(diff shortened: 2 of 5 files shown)


pkg/asset/installconfig/gcp/client.go
```diff
@@ -48,6 +49,36 @@ func NewClient(ctx context.Context) (*Client, error) {
 	return client, nil
 }
 
+// GetMachineTypes uses the GCP Compute Service API to get a list of machine types from a project
```
patrickdillon, line 52: nit: missing period

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -17,16 +17,69 @@ import (
...
+		allErrs = append(allErrs, validateInstanceType(client, ic, field.NewPath("controlPlane", "platform", "gcp"), ic.ControlPlane.Platform.GCP.InstanceType, controlPlaneReq)...)
+	}
+
+	for idx, compute := range ic.Compute {
+		fieldPath := field.NewPath("compute").Index(idx)
+		if compute.Platform.GCP != nil && compute.Platform.GCP.InstanceType != "" {
+			allErrs = append(allErrs, validateInstanceType(client, ic, fieldPath.Child("platform", "gcp"), compute.Platform.GCP.InstanceType, computeReq)...)
+		}
+	}
+
```
patrickdillon, line 52: nit: for consistency/readability I think you could move this into a function e.g. `validateInstanceTypes` so it would follow the pattern of `validateProject` and `validateNetworks`

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -17,16 +17,69 @@ import (
...
 	return allErrs.ToAggregate()
 }
 
+// ValidateInstanceType ensures the macyhine type has sufficient Vcpu and Memory
```
patrickdillon, line 56: nit: case doesn't match and missing period.

patrickdillon (commented):
Looks good to me.

/approve

I left some minor nits. I am happy to `/lgtm` but want to give a moment in case any one else wants to take a look.

patrickdillon:
Needs mocks generated: https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#generating-test-mocks

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -17,16 +17,69 @@ import (
...
+func validateInstanceType(client API, ic *types.InstallConfig, fieldPath *field.Path, instanceType string, req resourceRequirements) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if instanceType != "" {
```
mtnbikenc, line 60: Is this needed since the check for "" is done on line 42 above?

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -17,16 +17,69 @@ import (
...
+		if err != nil {
+			return append(allErrs, field.InternalError(fieldPath, err))
+		}
+		if typeMeta, ok := instanceTypes[instanceType]; ok {
+			if typeMeta.GuestCpus < req.minimumVCpus {
+				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d vCPUs", req.minimumVCpus)
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("instancetype"), instanceType, errMsg))
+			}
+			if typeMeta.MemoryMb < req.minimumMemory {
+				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d MiB Memory", req.minimumMemory)
```
staebler, line 72: Will this lead to any confusion for the user if we are reporting the memory requirements as MiB whereas GCP reports it as MB? I don't think that I often see MiB used in the context of memory.

pkg/asset/installconfig/gcp/validation_test.go
```diff
@@ -122,6 +127,12 @@ func TestGCPInstallConfigValidation(t *testing.T) {
 			expectedError:  true,
 			expectedErrMsg: "computeSubnet: Invalid value.*controlPlaneSubnet: Invalid value",
 		},
+		{
+			name:           "Invalid both machine types",
```
staebler, line 131: Shouldn't this test case need an expected call to `GetMachineTypes` added to the GCP client mock?

pkg/asset/installconfig/gcp/client.go
```diff
@@ -48,6 +49,36 @@ func NewClient(ctx context.Context) (*Client, error) {
...
+
+	req := svc.MachineTypes.AggregatedList(project)
+	if filter != "" {
+		req = req.Filter(filter)
+	}
+
+	if err := req.Pages(ctx, func(page *compute.MachineTypeAggregatedList) error {
+		for _, scopedList := range page.Items {
+			for _, item := range scopedList.MachineTypes {
+				types[item.Name] = item
```
bogdando, line 72: shouldn't this nested lookup exit earlier whenever possible?

staebler: @bogdando What are you suggesting as criteria for existing early? The expectation is that the code will add **all** of the machine types to the `types` map.

jstuever: No, this is designed to allow multiple machine types to be retrieved. It is true that it will experience several of the same type (one for each zone, in each region). However, if we exited here then we would miss all of the other types.

pkg/asset/installconfig/gcp/validation.go
abhinavdahiya, line 52: missing a check for the instance type in platform's defaultMachinePool similar to https://github.com/openshift/installer/blob/a409de073b472944be458283e69ef831cbdfd1c9/pkg/asset/installconfig/aws/validation.go#L68-L70

pkg/asset/installconfig/gcp/client.go
jstuever: Added https://github.com/openshift/installer/pull/4329#event-3949892939

pkg/asset/installconfig/gcp/validation.go
jstuever: Added https://github.com/openshift/installer/pull/4329#event-3949892939

pkg/asset/installconfig/gcp/validation.go
jstuever: Reworded this to instance type to be consistent.

pkg/asset/installconfig/gcp/validation.go
jstuever: I think this is needed as the function is now public (ValidateInstanceType) for consistency and can therefore be called from other code where instanceType might be empty. I did flatten the code as requested. https://github.com/openshift/installer/pull/4329#event-3949892939

pkg/asset/installconfig/gcp/validation.go
jstuever: Probably so, was brought over from AWS code doing the same. I reworded to MB for clarity. https://github.com/openshift/installer/pull/4329#event-3949892939

pkg/asset/installconfig/gcp/validation_test.go
jstuever: Yes, I forgot to generate the mock. https://github.com/openshift/installer/pull/4329#event-3949892939

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -17,16 +17,85 @@ import (
...
+
+	return allErrs
+}
+
+// validateInstanceTypes checks that the user-provided instance types are valid.
+func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if ic.Platform.GCP.DefaultMachinePlatform != nil && ic.Platform.GCP.DefaultMachinePlatform.InstanceType != "" {
+		allErrs = append(allErrs, ValidateInstanceType(client, ic, field.NewPath("platform", "gcp", "defaultMachinePlatform"), ic.Platform.GCP.DefaultMachinePlatform.InstanceType, controlPlaneReq)...)
```
staebler, line 82: This does not quite seem like the right check here. The default instance type does not **need** to work for the control plane machines. The user could specify a default instance type and then explicitly specify a different instance type for the control plane machines. I think that the check should be that the instance type that is going to be **used** for the control plane machines and the compute machines meets the requirements.

But maybe that is overkill considering
(1) the AWS validation works the way that you have it now and
(2) the default machine platform is only really helpful for the user if they are going to use it for both the control plane machines and the compute machines.

pkg/asset/installconfig/gcp/validation_test.go
```diff
@@ -122,6 +149,30 @@ func TestGCPInstallConfigValidation(t *testing.T) {
...
+			edits:          editFunctions{invalidateControlPlaneMachineTypes},
+			expectedError:  true,
+			expectedErrMsg: "[controlPlane.platform.gcp.type: Invalid value: \"n1-standard-1\": instance type does not meet minimum resource requirements of 4 vCPUs, controlPlane.platform.gcp.type: Invalid value: \"n1-standard-1\": instance type does not meet minimum resource requirements of 15360 MB Memory]",
+		},
+		{
+			name:           "Invalid compute machine types",
+			edits:          editFunctions{invalidateComputeMachineTypes},
+			expectedError:  true,
+			expectedErrMsg: "[compute[0].platform.gcp.type: Invalid value: \"n1-standard-1\": instance type does not meet minimum resource requirements of 2 vCPUs, compute[0].platform.gcp.type: Invalid value: \"n1-standard-1\": instance type does not meet minimum resource requirements of 7680 MB Memory]",
+		},
```
staebler, line 175: Can you add a test case where the instance type selected is not a valid instance type?

pkg/asset/installconfig/gcp/validation.go
jstuever: Added... https://github.com/openshift/installer/commit/8e4f255c510b9dc898c8fd90c297c0c3cea4965c

pkg/asset/installconfig/gcp/validation_test.go
jstuever: Done... https://github.com/openshift/installer/commit/0d74861bd7fa7cb28cde3f79fa63dc43704b69b2

pkg/asset/installconfig/gcp/validation.go
jstuever: Reworked to consider control plane. https://github.com/openshift/installer/commit/0d74861bd7fa7cb28cde3f79fa63dc43704b69b2

staebler (approved):
/lgtm
/approve

staebler:
/test e2e-gcp

Holding to make sure that gcp tests pass.
/hold

mtnbikenc:
e2e-gcp test passed
/hold cancel

PR #4331, opened 2020-10-30
Bug 1866925: display Azure destroy auth error

patrickdillon:
This appends Azure auth errors to the error list, so destroy is properly terminated and an error shown to the user. Without this, the metadata.json is deleted (along with other artifacts) and no error is shown to the user.

With this change, users (with incorrect credentials) will see:
```
$ ./openshift-install destroy cluster --dir azure-delete/cluster/
INFO Credentials loaded from file "/home/padillon/.azure/osServicePrincipal.json" 
FATAL Failed to destroy cluster: [unable to authenticate when deleting public DNS records: azure.BearerAuthorizer#WithAuthorization: Failed to refresh the Token for request to [...]
```
Without this change, users (at standard log level)  see only:
```
$ ./openshift-install destroy cluster --dir azure-delete/cluster/
INFO Credentials loaded from file "/home/padillon/.azure/osServicePrincipal.json" 
INFO Time elapsed: 1s    
```
(and more importantly, as noted before, the artifacts are deleted)

cc @jhixson74

### DIFF
--- a/pkg/destroy/azure/azure.go
+++ b/pkg/destroy/azure/azure.go
@@ -121,6 +121,7 @@ func (o *ClusterUninstaller) Run() error {
 				o.Logger.Debug(err)
 				if isAuthError(err) {
 					cancel()
+					errs = append(errs, errors.Wrap(err, "unable to authenticate when deleting public DNS records"))
 				}
 				return
 			}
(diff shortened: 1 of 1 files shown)


staebler (commented):
Not anything that should be done with your changes here, but this is a lot more complicated than it needs to be. We should be using `wait.PollUntil` instead of `wait.UntilWithContext`. Then we could return the error from the condition function rather than cancelling the context.

/lgtm
/approve

PR #4334, opened 2020-11-03
Bug 1855351: Handle CTRL+C in GCP survey

patrickdillon:
Corrects oversight where GCP survey was not handling CTRL+C.


staebler (approved):
/lgtm
/approve

PR #4335, opened 2020-11-03
aws: support custom trust bundle for c2s regions

staebler:
When installing to an AWS C2S region, the user needs to be able to specify additional trust bundles to be used when using the AWS API. These trust bundles need to be made available to the operators in the cluster that use the AWS API.

With these changes, the additional trust bundle specified in the install config will be included in the "ca-bundle.pem" key of the cloud-provider-config ConfigMap for AWS C2S installs.

Also, the additional trust bundle will be made available to the machine-config-operator running on the bootstrap node in the /opt/openshift/tls/cloud-ca-cert.pem file.

https://issues.redhat.com/browse/CORS-1584


mtnbikenc (approved):
/lgtm

PR #4336, opened 2020-11-03
Document how to deploy cluster with BM and VM workers

adduarte:


docs/user/openstack/deploy_baremetal_workers.md
```diff
@@ -177,6 +178,44 @@ Plane and Compute nodes. The cluster is deployed to a preexisting network.
     If the installer times out waiting for the bare-metal workers to complete booting,
     restart the installation using [the appropriate *wait-for* command](#considerations-when-deploying-bare-metal-workers).
 
+deploying-cluster-with-mixture-of-vm-and-bm-workers
```
EmilienM, line 181: I don't think you need that line.

EmilienM (changes requested):
One leftover, see inline

docs/user/openstack/deploy_baremetal_workers.md
adduarte: no I don't

pierreprinetti:
/approve
/lgtm
:+1:

pierreprinetti:
/hold
Please add a reference to a BZ or Jira card so that the change doesn't escape QE and dowstream-docs

PR #4337, opened 2020-11-03
Adding kubevirt-installer image

bardielle:
My team is working on a new OCP IPI to enable installation of nested OCP (i.e. OCP +Openshift Virtualization (KubeVirt) is the platform to host the nodes of the nested cluster).
You can find more information here: openshift/enhancements#417

As part of that project we want to add a new workflow in the ci "e2e-kubevirt", so the first step us adding a kubevirt-installer image build that does nothing.

### DIFF
--- a/images/kubevirt/Dockerfile.ci
+++ b/images/kubevirt/Dockerfile.ci
@@ -1,0 +1,18 @@
+# This Dockerfile is used by CI to test using OpenShift Installer against an KubeVirt platform.
+# It builds an image containing the openshift-installer command but does nothing.
+# This is the first step in running end-to-end tests on kubevirt-installer before merging it.
+
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7 AS builder
+WORKDIR /go/src/github.com/openshift/installer
+COPY . .
+RUN hack/build.sh
+
+FROM registry.svc.ci.openshift.org/ocp/4.7:base
+
+COPY --from=builder /go/src/github.com/openshift/installer/bin/openshift-install /bin/openshift-install
+
+RUN mkdir /output && chown 1000:1000 /output
+USER 1000:1000
+ENV PATH /bin
+ENV HOME /output
+WORKDIR /output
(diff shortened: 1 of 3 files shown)


images/kubevirt/OWNERS
```diff
@@ -0,0 +1,15 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+# This file just uses aliases defined in OWNERS_ALIASES.
```
staebler, line 2: This comment is not true. However, assuming that there will be more work to do in other parts of the installer repo, you probably want to make it true by specifying the kubevirt approvers and reviewers in the OWNERS_ALIASES file rather than here.

images/kubevirt/Dockerfile.ci
```diff
@@ -0,0 +1,20 @@
+# This Dockerfile is used by CI to test using OpenShift Installer against an KubeVirt cloud.
+# It builds an image containing the kubevirt-install command.
```
staebler, line 2: This comment is not true.

images/kubevirt/Dockerfile.ci
```diff
@@ -0,0 +1,20 @@
...
+
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7 AS builder
+WORKDIR /go/src/github.com/openshift/installer
+COPY . .
+RUN hack/build.sh
+
+FROM registry.svc.ci.openshift.org/origin/4.7:cli AS cli
+FROM registry.svc.ci.openshift.org/ocp/4.7:base
+
+COPY --from=cli /usr/bin/oc /bin/oc
```
staebler, line 12: Out of curiosity, what will you be using `oc` for in the CI image? That seems to be the only difference between this image and the standard installer CI image at the moment. Will there be more changes to come?

bardielle: I removed the oc part
We will change it in the future

staebler: Hmm. Now we have a new CI image is the same as the standard installer CI image. Can we wait on this until there are actual differences in the images?

bardielle: You're right 
I'm closing that PR for now

PR #4338, opened 2020-11-03
Wrap errors after OpenStack creds validation

Fedosin:
This commit wraps an error after creds validation to comply with the other platforms.

/label platform/openstack


mandre:
Could you please link to a BZ so that we can more easily backport if needed?

Fedosin:
@mandre this is not a bugfix. this is just a code cleanup - we return the err anyway in the end of the function.

mandre (commented):
/lgtm

PR #4342, opened 2020-11-04
Use 'kubernetes.io/tls' instead of SecretTypeTLS which is not being substituted

jhadvig:
/assign @wking


hexfusion:
LGTM

cc @staebler

staebler (approved):
/approve
/lgtm

PR #4347, opened 2020-11-04
data/manifests/bootkube/cvo-overrides: Bump default to stable-4.7

wking:
Like f8a0f7b069 (#3848), but for 4.7 (now that release-4.6 is no longer being fast-forwarded to track master).


staebler (approved):
/lgtm
/approve

PR #4352, opened 2020-11-05
Bug 1895099: Create empty vSphere infra PlatformStatus in UPI

patrickdillon:
This creates an empty infrastructure.Status.PlatformStatus.VSphere object during UPI installs. Prior to this change, the object has been nil during UPI installs. **The object is optional,** but a recent change to MCO templating[0] has a dependency on the object.

[0]: https://github.com/openshift/machine-config-operator/pull/2079

cc @jcpowermac

### DIFF
--- a/pkg/asset/manifests/infrastructure.go
+++ b/pkg/asset/manifests/infrastructure.go
@@ -147,11 +147,10 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
 		}
 	case vsphere.Name:
 		config.Spec.PlatformSpec.Type = configv1.VSpherePlatformType
+		config.Status.PlatformStatus.VSphere = &configv1.VSpherePlatformStatus{}
 		if installConfig.Config.VSphere.APIVIP != "" {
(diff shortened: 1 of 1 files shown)


jcpowermac:
Makes sense to me. I will let tests run and see if upi is green.

patrickdillon:
@jcpowermac righly pointed out (on slack) this will not fix upgrades, not sure if this PR has merit on its own, i will look at an analogue mco fix

patrickdillon:
/hold 

Hm it looks like these changes might just result in a mangled config file where the IP addresses are missing but the rest of the config file is present: https://github.com/openshift/machine-config-operator/blob/master/templates/common/on-prem/files/NetworkManager-resolv-prepender.yaml

Before the mco changes, we wouldn't populate any of the config file: https://github.com/openshift/machine-config-operator/pull/2079/files#diff-88b9eeb5fb253707d42dfa34ac2143dda76d949209de64c032a8cd11a0d97c29

patrickdillon:
/close
in favor of openshift/machine-config-operator#2207

PR #4354, opened 2020-11-06
pkg: validation for previously existing ovirt configuration

janoszen:
This PR fixes an issue when running the installer on ovirt with an already existing, but invalid `ovirt-config.yaml`. Previously, the installer did not allow the user to re-enter the credentials in this case. This patch fixes that.

## Previous behavior

```
? Platform ovirt
FATAL failed to fetch Install Config: failed to fetch dependency of "Install Config": failed to fetch dependency of "Base Domain": failed to generate asset "Platform": The Password must not be empty
```

## New behavior

```
? Platform ovirt
{{WARNING found invalid existing ovirt configuration (Error during SSO authentication access_denied : Cannot authenticate user 'admin@N/A': No valid profile found in credentials..) }}
? Engine FQDN[:PORT] [? for help]
```

### DIFF
--- a/pkg/asset/installconfig/ovirt/config.go
+++ b/pkg/asset/installconfig/ovirt/config.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"path/filepath"
 
+	ovirtsdk "github.com/ovirt/go-ovirt"
+	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
 )
 
@@ -84,4 +86,18 @@ func (c *Config) Save() error {
 		return err
 	}
 	return ioutil.WriteFile(path, out, 0600)
+}
+
+// getValidatedConnection will create a connection and validate it before returning.
+func (c *Config) getValidatedConnection() (*ovirtsdk.Connection, error) {
+	connection, err := getConnection(*c)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to build configuration for ovirt connection validation")
+	}
+
+	if err := connection.Test(); err != nil {
+		_ = connection.Close()
+		return nil, err
+	}
+	return connection, nil
 }
--- a/pkg/asset/installconfig/ovirt/ovirt.go
+++ b/pkg/asset/installconfig/ovirt/ovirt.go
@@ -9,7 +9,7 @@ import (
 	"github.com/openshift/installer/pkg/types/ovirt"
 )
 
-const PlatformValidationMaxTries = 3
+const platformValidationMaxTries = 3
 
 // Platform collects ovirt-specific configuration.
 func Platform() (*ovirt.Platform, error) {
@@ -19,7 +19,7 @@ func Platform() (*ovirt.Platform, error) {
 
 	// Fetch config from file
 	ovirtConfig, err := NewConfig()
-	for tries := 0; tries < PlatformValidationMaxTries; tries++ {
+	for tries := 0; tries < platformValidationMaxTries; tries++ {
 		if err == nil {
 			// If no error happened previously (loading file or configuration), validate the connection.
 			c, err = ovirtConfig.getValidatedConnection()
--- a/pkg/asset/installconfig/ovirt/ovirt.go
+++ b/pkg/asset/installconfig/ovirt/ovirt.go
@@ -17,27 +17,21 @@ func Platform() (*ovirt.Platform, error) {
 
 	var c *ovirtsdk4.Connection
 
-	// Fetch config from file
 	ovirtConfig, err := NewConfig()
 	for tries := 0; tries < platformValidationMaxTries; tries++ {
-		if err == nil {
-			// If no error happened previously (loading file or configuration), validate the connection.
-			c, err = ovirtConfig.getValidatedConnection()
+		if err != nil {
+			ovirtConfig, err = engineSetup()
 			if err != nil {
-				// If validation failed log and drop into reconfig below.
-				logrus.Error(errors.Wrap(err, "failed to validate oVirt configuration"))
-			} else {
-				// If connection is valid, break
-				break
+				logrus.Error(errors.Wrap(err, "oVirt configuration failed"))
 			}
 		}
 
-		if err != nil {
-			// If a previous error happened (validation or loading from file), rerun the setup.
-			ovirtConfig, err = engineSetup()
+		if err == nil {
+			c, err = ovirtConfig.getValidatedConnection()
 			if err != nil {
-				// If validation failed log and loop back to engineSetup()
-				logrus.Error(errors.Wrap(err, "oVirt configuration failed"))
+				logrus.Error(errors.Wrap(err, "failed to validate oVirt configuration"))
+			} else {
+				break
 			}
 		}
 	}
--- a/pkg/asset/installconfig/ovirt/ovirt.go
+++ b/pkg/asset/installconfig/ovirt/ovirt.go
@@ -3,40 +3,52 @@ package ovirt
 import (
 	ovirtsdk4 "github.com/ovirt/go-ovirt"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"gopkg.in/AlecAivazis/survey.v1"
 
 	"github.com/openshift/installer/pkg/types/ovirt"
 )
 
+const PlatformValidationMaxTries = 3
+
 // Platform collects ovirt-specific configuration.
 func Platform() (*ovirt.Platform, error) {
 	p := ovirt.Platform{}
 
+	var c *ovirtsdk4.Connection
+
+	// Fetch config from file
 	ovirtConfig, err := NewConfig()
-	if err != nil {
-		ovirtConfig, err = engineSetup()
+	for tries := 0; tries < PlatformValidationMaxTries; tries++ {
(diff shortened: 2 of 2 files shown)


staebler (commented):
It is not appropriate to do this check on every call to `NewConfig`. If behavior around creds checking needs to be improved or reported to the user better, look at https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/ovirt/ovirt.go and https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/platformcredscheck.go.

dougsland:
looks like WIP, please use WIP in the subject or set the patch as Draft.

pkg/asset/installconfig/ovirt/config.go
```diff
@@ -85,3 +87,33 @@ func (c *Config) Save() error {
 	}
 	return ioutil.WriteFile(path, out, 0600)
 }
+
+// Validate will validate if the provided certificate and credentials are valid. It will also issue log messages if the
+// credentials are not valid or the connection could otherwise not be established.
+func (c *Config) Validate() bool {
```
staebler, line 93: Rather than logging in the `Validate` function, return an error and let the caller decide what to do with the error.

pkg/asset/installconfig/ovirt/ovirt.go
```diff
@@ -12,7 +12,7 @@ func Platform() (*ovirt.Platform, error) {
 	p := ovirt.Platform{}
 
 	ovirtConfig, err := NewConfig()
-	if err != nil {
+	if err != nil || !ovirtConfig.Validate() {
```
staebler, line 15: It does not appear that `ovirtConfig.Validate` is any different than the test that is performed a few lines down from here.

janoszen: Hey @staebler thank you very much for your help with this issue. I'm going to push a proposed fix, but since I don't have a fully working ovirt setup yet please allow me until next week to test it properly.

janoszen:
I have verified this PR against the latest 4.7 images up to the point that the bootstrap cluster comes up and starts installing. From my point of view it's good to go for a review.

pkg/asset/installconfig/ovirt/config.go
```diff
@@ -85,3 +88,16 @@ func (c *Config) Save() error {
 	}
 	return ioutil.WriteFile(path, out, 0600)
 }
+
+// getValidatedConnection will create a connection and validate it before returning.
+func (c *Config) getValidatedConnection() (*ovirtsdk.Connection, error) {
+	connection, err := getConnection(*c)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build configuration for ovirt connection validation (%v)", err)
```
staebler, line 96: Use `Wrap` from "github.com/pkg/errors".
```suggestion
		return nil, errors.Wrap(err, "failed to build configuration for ovirt connection validation)
```

pkg/asset/installconfig/ovirt/ovirt.go
```diff
@@ -1,41 +1,39 @@
 package ovirt
 
 import (
-	ovirtsdk4 "github.com/ovirt/go-ovirt"
+	"github.com/sirupsen/logrus"
 	"gopkg.in/AlecAivazis/survey.v1"
 
 	"github.com/openshift/installer/pkg/types/ovirt"
+	ovirtsdk "github.com/ovirt/go-ovirt"
```
staebler, line 8: Keep this import in the section that it was in.

pkg/asset/installconfig/ovirt/config.go
```diff
@@ -2,10 +2,13 @@ package ovirt
 
 import (
 	"crypto/x509"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 
+	ovirtsdk "github.com/ovirt/go-ovirt"
+
```
staebler, line 11: Keep all imports that are not standard imports, kube imports, or openshift imports in the same section.
```suggestion
```

pkg/asset/installconfig/ovirt/ovirt.go
```diff
@@ -1,41 +1,39 @@
...
 // Platform collects ovirt-specific configuration.
 func Platform() (*ovirt.Platform, error) {
 	p := ovirt.Platform{}
 
+	var c *ovirtsdk.Connection
 	ovirtConfig, err := NewConfig()
+	if err == nil {
+		c, err = ovirtConfig.getValidatedConnection()
+		if err != nil {
+			logrus.Errorf("failed to validate existing oVirt configuration (%v)", err)
```
staebler, line 20:
```suggestion
			logrus.Errorf("failed to validate existing oVirt configuration: %v", err)
```
or
```suggestion
			logrus.Error(errors.Wrap(err, "failed to validate existing oVirt configuration"))
```

pkg/asset/installconfig/ovirt/config.go
```diff
@@ -85,3 +88,16 @@ func (c *Config) Save() error {
...
+// getValidatedConnection will create a connection and validate it before returning.
+func (c *Config) getValidatedConnection() (*ovirtsdk.Connection, error) {
+	connection, err := getConnection(*c)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build configuration for ovirt connection validation (%v)", err)
+	}
+
+	if err := connection.Test(); err != nil {
+		return nil, err
```
staebler, line 100: The connection is not closed in this code path.

pkg/asset/installconfig/ovirt/ovirt.go
```diff
@@ -1,41 +1,39 @@
...
+	ovirtsdk "github.com/ovirt/go-ovirt"
 )
 
 // Platform collects ovirt-specific configuration.
 func Platform() (*ovirt.Platform, error) {
 	p := ovirt.Platform{}
 
+	var c *ovirtsdk.Connection
 	ovirtConfig, err := NewConfig()
+	if err == nil {
```
Gal-Zaidman, line 17: Why just ask once? I mean if we are giving the user the ability to ask fix credentials...
Something like 

```
for i := 0; i < MAX_RETRY; i++ {
		if err != nil {
			ovirtConfig, err = engineSetup()
			if err != nil {
				return nil, err
			}
		}
		c, err = ovirtConfig.getValidatedConnection()
		if err != nil {
			logrus.Warnf("failed to validate oVirt configuration (%v)", err)
		} else {
			break
		}
	}
```

janoszen: @Gal-Zaidman how's this? (See latest push.)

pkg/asset/installconfig/ovirt/ovirt.go
```diff
@@ -2,40 +2,53 @@ package ovirt
...
 	p := ovirt.Platform{}
 
+	var c *ovirtsdk4.Connection
+
+	// Fetch config from file
 	ovirtConfig, err := NewConfig()
-	if err != nil {
-		ovirtConfig, err = engineSetup()
+	for tries := 0; tries < PlatformValidationMaxTries; tries++ {
+		if err == nil {
```
Gal-Zaidman, line 23: You need to switch the order of the if's, meaning:
```
 	ovirtConfig, err := NewConfig()
	for tries := 0; tries < PlatformValidationMaxTries; tries++ {
		if err != nil {
                    ..........
                }
                if err == nil {
                    ..........
                }
```

That way you always get the same number of tries and validations

pkg/asset/installconfig/ovirt/ovirt.go
```diff
@@ -2,40 +2,53 @@ package ovirt
...
+			// If no error happened previously (loading file or configuration), validate the connection.
```
Gal-Zaidman, line 24: you can remove that comment, no need to make it to verbus

pkg/asset/installconfig/ovirt/ovirt.go
```diff
@@ -2,40 +2,53 @@ package ovirt
...
+			c, err = ovirtConfig.getValidatedConnection()
+			if err != nil {
+				// If validation failed log and drop into reconfig below.
```
Gal-Zaidman, line 27: you can remove that comment, no need to make it to verbus

pkg/asset/installconfig/ovirt/ovirt.go
```diff
@@ -2,40 +2,53 @@ package ovirt
...
+				logrus.Error(errors.Wrap(err, "failed to validate oVirt configuration"))
+			} else {
+				// If connection is valid, break
```
Gal-Zaidman, line 30: you can remove that comment, no need to make it to verbus

pkg/asset/installconfig/ovirt/ovirt.go
```diff
@@ -2,40 +2,53 @@ package ovirt
...
+				break
+			}
+		}
+
 		if err != nil {
-			return nil, err
+			// If a previous error happened (validation or loading from file), rerun the setup.
+			ovirtConfig, err = engineSetup()
+			if err != nil {
+				// If validation failed log and loop back to engineSetup()
```
Gal-Zaidman, line 39: you can remove that comment, no need to make it to verbus

PR #4355, opened 2020-11-06
doc/openstack/UPI: improvements

EmilienM:
- Update package names for Fedora, now python3
- Replace the usage of swiftclient by openstackclient
  for container/object creates (which uses clouds.yaml)
- Fix logs alignment


mandre (approved):
/lgtm
/approve

PR #4362, opened 2020-11-10
doc/openstack: update quota reqs for kuryr

EmilienM:
Update the Neutron quotas now required when Kuryr is used at the
networking plugin for an OCP cluster.

The numbers were taken from:
https://docs.openshift.com/container-platform/4.6/installing/installing_openstack/installing-openstack-user-kuryr.html#installation-osp-default-kuryr-deployment_installing-openstack-user-kuryr

and confirmed with the Kuryr team.

It updates the table and the command that updates the quotas so the
reader won't be confused nor set the wrong quotas, which lead to a
failed deployment of that type: http://paste.openstack.org/show/799839/

### DIFF
--- a/docs/user/openstack/kuryr.md
+++ b/docs/user/openstack/kuryr.md
@@ -29,12 +29,12 @@ or
 When using Kuryr SDN, as the pods, services, namespaces, network policies, etc., are using resources from the OpenStack Quota, the minimum requirements are higher. Kuryr also has some additional requirements on top of what a default install requires. At a bare minimum you need the following quota to run a *default* cluster:
 
 * Floating IPs: 3 (plus the expected number of services of LoadBalancer type)
-* Security Groups: 100 (1 needed per network policy)
-* Security Group Rules: 500
+* Security Groups: 250 (1 needed per Service and per NetworkPolicy)
+* Security Group Rules: 1000
 * Routers: 1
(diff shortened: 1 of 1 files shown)


MaysaMacedo:
Thanks!

/lgtm

docs/user/openstack/kuryr.md
```diff
@@ -29,15 +29,15 @@ or
...
-* Subnets: 100 (1 needed per namespace)
-* Networks: 100 (1 needed per namespace)
-* Ports: 1000
+* Subnets: 250 (1 needed per namespace)
+* Networks: 250 (1 needed per namespace)
+* Ports: 1500
 * RAM: 112 GB
 * vCPUs: 28
-* Volume Storage: 175 GB
+* Volume Storage: 275 GB
```
mandre, line 40: Please leave 175 GB, as it is on the README for the [generic install on OpenStack](https://github.com/openshift/installer/tree/master/docs/user/openstack#openstack-requirements). The extra 100 GB is for cases where swift isn't available, so perhaps we should also add:
```
Depending on the type of image registry backend either 1 Swift container or an additional 100 GB volume.
```

EmilienM: done

mandre (approved):
/lgtm
/approve

PR #4367, opened 2020-11-11
data/data/aws/route53: use CNAME for us-iso-east-1 region

jstuever:
We cannot create ALIAS records to the NLBs in C2S cloud partition, which includes `us-iso-east-1`. So we switch to using CNAME for our api records in that region.


staebler:
https://issues.redhat.com/browse/CORS-1583

staebler (approved):
/lgtm
/approve

PR #4372, opened 2020-11-12
Add an automatically generated Ignition provisioning token

cgwalters:
Part of implementing: https://github.com/openshift/enhancements/pull/443

The installer generates a random token (~password) and injects
it into the Ignition pointer configuration and as a secret into the
cluster.

The MCO will check it.


cgwalters:
Hmm...not sure what's up with that e2e-aws run.  I can't find a must-gather or extracted cluster logs in the artifacts?
/test e2e-aws

pkg/asset/cluster/cluster.go
```diff
@@ -51,6 +52,7 @@ func (c *Cluster) Dependencies() []asset.Asset {
 		&quota.PlatformQuotaCheck{},
 		&TerraformVariables{},
 		&password.KubeadminPassword{},
+		&ignition.ProvisioningToken{},
```
staebler, line 55: Why does the Cluster asset need this dependency?

pkg/asset/ignition/machine/node.go
```diff
@@ -37,16 +37,19 @@ func pointerIgnitionConfig(installConfig *types.InstallConfig, rootCA []byte, ro
 			ignitionHost = net.JoinHostPort(installConfig.VSphere.APIVIP, "22623")
 		}
 	}
+	queryvals := url.Values{}
+	queryvals.Add("token", token)
```
staebler, line 41: Move this inside the function where the URL is created.

pkg/asset/ignition/provisioningtoken.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// Generate the token
+func (a *ProvisioningToken) Generate(asset.Parents) error {
+	b := make([]byte, tokenLen)
+	n, err := rand.Read(b)
+	if err != nil {
+		return err
+	}
+	if n != tokenLen {
```
staebler, line 36: This check is not necessary. It is already guaranteed by the `rand.Read` function.

pkg/asset/ignition/machine/node.go
cgwalters: This is that function, right?

pkg/asset/ignition/provisioningtoken.go
cgwalters: Yeah, excess of paranoia there.  Removed.

pkg/asset/ignition/machine/node.go
cgwalters: Oh sorry misread.  Fixed.

cgwalters:
Hmm I'm confused the contexts say "Job failed" yet it looks like e.g. the gofmt and unit contexts succeeded?

staebler (commented):
/lgtm

/hold
on approval of the enhancement

PR #4373, opened 2020-11-12
Support user provided service-account-signing-key and issuer

abhinavdahiya:
ref: https://issues.redhat.com/browse/CO-1266

 bootkube.sh: update kap-o render to include flag for authentication object file

`--cluster-auth-file=` flag added in openshift/cluster-kube-apiserver-operator#1006
the kap-o render handles the case where the file does not exist even when set so that installer script does not have
to handle the conditional when the authentication file does not exist (since it is not created by installer by default
and only added by user when using STS+OIDC)

tls: add asset to load user-provided bound-sa keys

Allow users to provide the service account signing keys at
- $INSTALL_DIR/tls/bound-service-account-signing.key
The installer reads the private RSA key and extracts the public key creates the corresponding bound-service-account-signing.pub
file for the bootstrap host.

These files are traferred to the bootstrap host so that kap-o can read these files
and configure the bootstrap-kube-apiserver and in-cluster kube-apiserver with this key.

kap-o render PR openshift/cluster-kube-apiserver-operator#1006 adds the functionality
to read these files from /opt/openshift/tls/<> and setup the bootkube-kube-apiserver to use the key on bootstrap host
and also create a secret with these files so in-cluster kap-o can configure the actual kube-apiserver to also use these
keys.


abhinavdahiya:
/hold
for https://github.com/openshift/cluster-kube-apiserver-operator/pull/1006

pkg/asset/tls/boundsasigningkey.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Name returns a human friendly name for the asset.
+func (*BoundSASigningKey) Name() string {
+	return "User-provided Service Account Signing key"
+}
+
+// Dependencies returns all of the dependencies directly needed to generate
+// the asset.
+func (*BoundSASigningKey) Dependencies() []asset.Asset {
+	return []asset.Asset{}
```
staebler, line 28:
```suggestion
	return nil
```

pkg/asset/tls/boundsasigningkey.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// Generate generates the CloudProviderConfig.
+func (*BoundSASigningKey) Generate(dependencies asset.Parents) error { return nil }
+
+// Files returns the files generated by the asset.
+func (sk *BoundSASigningKey) Files() []*asset.File {
+	if len(sk.FileList) == 0 {
+		return []*asset.File{}
+	}
```
staebler, line 38: Why is this needed? Can't the function return `sk.FileList` even when it is empty or nil?

pkg/asset/tls/boundsasigningkey.go
```diff
@@ -0,0 +1,59 @@
...
+	return sk.FileList
+}
+
+// Load reads the private and public key from the disk.
+// If only one of the private or public key are provided, the loads fails with error.
```
staebler, line 43: This comment does not match the actual behavior of the function. The actual behavior is that, if one of the files does not exist, then the other file is ignored.

pkg/asset/tls/boundsasigningkey.go
```diff
@@ -0,0 +1,59 @@
...
+func (sk *BoundSASigningKey) Load(f asset.FileFetcher) (bool, error) {
+	keyFile, err := f.FetchByName(filepath.Join(tlsDir, "bound-service-account-signing-key.key"))
```
staebler, line 45: The error from the fetch gets swallowed if the error is something other than NotExist.

pkg/asset/tls/boundsasigningkey.go
abhinavdahiya: :+1:

pkg/asset/tls/boundsasigningkey.go
abhinavdahiya: :+1:

pkg/asset/tls/boundsasigningkey.go
abhinavdahiya: fixed to match the documentation in eb2f689

pkg/asset/tls/boundsasigningkey.go
abhinavdahiya: fixed the error handling in eb2f689

staebler (commented):
/lgtm
/approve

marun:
@abhinavdahiya Is it necessary to require that the public key be supplied given that rsa allows the public key to be trivially discovered from the private key? If so, there needs to be a step (either installer or ckao render) that checks that the public key and private key are related. In either case, the key material should be parsed to ensure its validity to avoid installing with a configuration that might potentially result in a kube apiserver that won't start.

abhinavdahiya:
> @abhinavdahiya Is it necessary that the public key be supplied given that rsa allows the public key to be trivially discovered from the private key? If so, there needs to be a step that checks that the public key and private key are related. In either case, the key material should be parsed to ensure its validity to avoid installing with a configuration that might potentially result in a kube apiserver that won't start. I'm not sure who - installer or ckao render - should be responsible for validating the key material.

@marun Updated https://github.com/openshift/installer/compare/eb2f6898acb4a750f81cb2d321916d0049c5c956..c0ab2f4a0a5f4ac6c54d981b642867a97c2e8115 to allow users to provide only the RSA key. The installer should now ensure that the key provided is RSA and then it will extract the public key from it and make that available for the bootstrap host.

kap-o render should be able to depend on the installer to provide the right/validated key-pair . Validating in installer imo made more sense because any validation at kap-o render stage is more difficult to fix (remove all infra resources etc.)

pkg/asset/tls/boundsasigningkey.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	rsaKey, err := PemToPrivateKey(keyFile.Data)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to load rsa.PrivateKey from the file")
+	}
+	pubData, err := PublicKeyToPem(&rsaKey.PublicKey)
+	if err != nil {
+		return false, errors.Wrap(err, "failed to extract public key from the key")
+	}
+	sk.FileList = []*asset.File{keyFile, {Filename: filepath.Join(tlsDir, "bound-service-account-signing-key.pub"), Data: pubData}}
```
marun, line 58: (No action required) DRY: Maybe define the filename once rather than both here and on line 43.

marun (commented):
Good improvements, thank you! LGTM

staebler (commented):
/lgtm

abhinavdahiya:
/hold cancel

Dependency PR in KAP-O merged

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -141,7 +141,8 @@ then
 		--asset-input-dir=/assets/tls \
 		--asset-output-dir=/assets/kube-apiserver-bootstrap \
 		--config-output-file=/assets/kube-apiserver-bootstrap/config \
-		--cluster-config-file=/assets/manifests/cluster-network-02-config.yml
+		--cluster-config-file=/assets/manifests/cluster-network-02-config.yml \
+		--cluster-auth-file=/assets/manifests/cluster-authentication-02-config.yaml
```
deads2k, line 145: trying to work out where this comes from.  Is this filename `cluster-authentication-02-config.yaml` actually produced by the cluster-config-operator?

Issue #4374, opened 2020-11-13
[Openstack UPI] Openstack server group error occurs due to os-api-version


gyohuangxin:
# Version

```console
$ openshift-install version
./openshift-install 4.6.1
built from commit ebdbda57fc18d3b73e69f0f2cc499ddfca7e6593
release image quay.io/openshift-release-dev/ocp-release@sha256:d14e2a166a2889633850f4bb472e6803c6408622316ffe63d21a68f068eb4bc9
```

# Platform:

Openstack (os api version: 2.72)

UPI

# What happened?

Refer to: https://github.com/openshift/installer/blob/9be316bc89f560653e3ad07cde7a716bf938cf6d/upi/openstack/control-plane.yaml#L45

If it's the second time to do Openstack UPI on an Openstack which OS_COMPUTE_API_VERSION=2.72, server group error will occur as below.
`The server has either erred or is incapable of performing the requested operation. (HTTP 500) (Request-ID: req-ecc9a588-fe38-4ac2-89da-88ccd48bb045)`

# What you expected to happen?

Enter text here.

# How to reproduce it (as minimally and precisely as possible)?

```console
$ your-commands-here
```

# Anything else we need to know?

Enter text here.

# References

- enter text here.

mandre:
/label platform/openstack

Hi @gyohuangxin, thanks for the report. Just to make sure I understand, with OpenStack Stein (micro-version 2.72), you get the above error when re-running the `control-plane.yaml` playbook?

gyohuangxin:
> /label platform/openstack
> 
> Hi @gyohuangxin, thanks for the report. Just to make sure I understand, with OpenStack Stein (micro-version 2.72), you get the above error when re-running the `control-plane.yaml` playbook?

Yes, failed to re-run `openstack server group list -f json -c ID -c Name` , because the previous record has the key `policies`. Hence, we used `openstack --os-compute-api-version=2.15 server group list -f json -c ID -c Name`.

PR #4375, opened 2020-11-13
Bug 1910067: openstack UPI: Set the API microversion when operating openstack server groups

gyohuangxin:
Fix https://github.com/openshift/installer/issues/4374

### DIFF
--- a/upi/openstack/control-plane.yaml
+++ b/upi/openstack/control-plane.yaml
@@ -44,7 +44,8 @@
 
   - name: 'List the Server groups'
     command:
-      cmd: "openstack server group list -f json -c ID -c Name"
+      # os-compute-api-version 2.15 or higher is required for the 'soft-anti-affinity' policy
+      cmd: "openstack --os-compute-api-version=2.15 server group list -f json -c ID -c Name"
     register: server_group_list
 
   - name: 'Parse the Server group ID from existing'
@@ -57,6 +58,7 @@
 
   - name: 'Create the Control Plane server group'
     command:
+      # os-compute-api-version 2.15 or higher is required for the 'soft-anti-affinity' policy
       cmd: "openstack --os-compute-api-version=2.15 server group create -f json -c id --policy=soft-anti-affinity {{ os_cp_server_group_name }}"
     register: server_group_created
     when:
--- a/upi/openstack/down-control-plane.yaml
+++ b/upi/openstack/down-control-plane.yaml
@@ -16,10 +16,26 @@
       state: absent
     with_indexed_items: "{{ [os_cp_server_name] * os_cp_nodes_number }}"
 
+  - name: 'List the Server groups'
+    command:
+      # os-compute-api-version 2.15 or higher is required for the 'soft-anti-affinity' policy
+      cmd: "openstack --os-compute-api-version=2.15 server group list -f json -c ID -c Name"
+    register: server_group_list
+
+  - name: 'Parse the Server group ID from existing'
+    set_fact:
+      server_group_id: "{{ (server_group_list.stdout | from_json | json_query(list_query) | first).ID }}"
+    vars:
+      list_query: "[?Name=='{{ os_cp_server_group_name }}']"
+    when:
+    - "os_cp_server_group_name|string in server_group_list.stdout"
+
   - name: 'Remove the Control Plane server group'
-    os_server_group:
-      name: "{{ os_cp_server_group_name }}"
-      state: absent
+    command:
+      # os-compute-api-version 2.15 or higher is required for the 'soft-anti-affinity' policy
+      cmd: "openstack --os-compute-api-version=2.15 server group delete {{ server_group_id }}"
+    when:
+    - server_group_id is defined
 
   - name: 'List the Compute trunks'
     command:


pierreprinetti:
/hold
/retest

Thank you for the patch. Do you mind waiting until #4379 is merged, to see if Ansible-Galaxy provides us with an OpenStack module that handles the microversion better?

gyohuangxin:
> /hold
> /retest
> 
> Thank you for the patch. Do you mind waiting until #4379 is merged, to see if Ansible-Galaxy provides us with an OpenStack module that handles the microversion better?

Of course, thank you.

pierreprinetti:
@gyohuangxin 
Do you still experience the version-mismatch issue [using `ansible-galaxy` to fetch the modules](https://github.com/openshift/installer/pull/4379/files#diff-6048f620b6429cf39395b4c267287d0d9b4ac9da05a2fa17cebc5f2381bba5c3R155-R157)?

gyohuangxin:
> @gyohuangxin
> Do you still experience the version-mismatch issue [using `ansible-galaxy` to fetch the modules](https://github.com/openshift/installer/pull/4379/files#diff-6048f620b6429cf39395b4c267287d0d9b4ac9da05a2fa17cebc5f2381bba5c3R155-R157)?

Hi, this issue still occurs after using this modules.

```
TASK [List the Server groups] ********************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["openstack", "server", "group", "list", "-f", "json", "-c", "ID", "-c", "Name"], "delta": "0:00:03.759616", "end": "2020-11-24 02:44:07.057232", "msg": "non-zero return code", "rc": 1, "start": "2020-11-24 02:44:03.297616", "stderr": "The server has either erred or is incapable of performing the requested operation. (HTTP 500) (Request-ID: req-5b4774ba-d681-4543-b954-01a30eab8cee)", "stderr_lines": ["The server has either erred or is incapable of performing the requested operation. (HTTP 500) (Request-ID: req-5b4774ba-d681-4543-b954-01a30eab8cee)"], "stdout": "", "stdout_lines": []}
```

Still need specify`--os-compute-api-version` to list server group.
```
# openstack --os-compute-api-version=2.15 server group list
+--------------------------------------+------------------------+--------------------+
| ID                                   | Name                   | Policies           |
+--------------------------------------+------------------------+--------------------+
| 78be7fc6-9036-4b8a-b4db-85496509cc7d | ocp461kvm-fpkfm-master | soft-anti-affinity |
+--------------------------------------+------------------------+--------------------+
```

mdbooth (changes requested):
I don't believe we should do this when the bug is in the client library. We should instead assert an up to date version of the client library.

However, even if we do this for expediency, we *must* include a comment explaining that:

* this is a hack
* there is no underlying requirement for the api version specified
* the conditions in which the hack can be removed

Without the above at a minimum, I believe we should not add this hack. And to reiterate, I would prefer not to add it at all. We should address the library dependency issue instead.

upi/openstack/control-plane.yaml
```diff
@@ -45,7 +45,7 @@
 
   - name: 'List the Server groups'
     command:
-      cmd: "openstack server group list -f json -c ID -c Name"
+      cmd: "openstack --os-compute-api-version=2.15 server group list -f json -c ID -c Name"
```
mdbooth, line 48: Why 2.15? Presumably to avoid version 2.16, but that was released in 2016 and doesn't appear to affect the output of the server_groups api as far as I can tell. If we have a dependency on version 2.15, or are in some way incompatible with version 2.16 we should clearly document here what it is.

adduarte: The problem is that the library is not installed by us. This problem is caused because the user is executing ansible with an openstack-sdk that requires the library api version to be specified. 
See https://bugzilla.redhat.com/show_bug.cgi?id=1891816

mdbooth:
To reiterate, this fix appears to work by coincidence by switching to a very old microversion whose api response coincidentally happens not to trigger a bug in openstacksdk. I believe we should fix this by:

1. Ensuring that openstacksdk is updated with the fix in RHEL.
2. Documenting the openstackdsk version requirement for anybody not using RHEL.

If we can clearly document here why the above isn't possible, I would want to see us:

1. Pick a microversion which relates to the problem we're trying to workaround (2.15 has been chosen arbitrarily afaict)
2. Add a clear comment which:
  - Explains why the microversion chosen was picked.
  - Links to the openstacksdk bug which is the reason for the workaround.

This will allow somebody in the future to remove the workaround with confidence when it is no longer required.

upi/openstack/control-plane.yaml
pierreprinetti: @mdbooth Actually, it's 2.15 because [the official docs for the latest version](https://docs.openstack.org/python-openstackclient/latest/cli/command-objects/server-group.html) advise so when treating with "soft" anti-affinity policies.

pierreprinetti:
/retitle Bug 1910067: openstack UPI: Set the API microversion when operating openstack server groups

mdbooth:
I remain strongly opposed to merging this.

upi/openstack/control-plane.yaml
mdbooth: That's a minimum. My comment about the 'fix' being coincidental stands because we were previously using latest, which is also >=2.15.

mdbooth (changes requested):
I don't think we should merge this. We should concentrate on delivering the fix to openstacksdk rather than adding technical debt to openshift for a temporary issue which is already fixed upstream.

upi/openstack/control-plane.yaml
pierreprinetti: I have no reason to think that bug 1891816 is related here.

If I stick to the idea that they're separate, it makes sense.

Think of this: the OP says that only the _second time_ they run it, it fails.

* `list` succeeds
* `list` fails the second time, because there's already a server group with a "*soft*" policy

In the end: the docs tell us to set a microversion. We don't set the microversion. We get a report that the thing fails, and it doesn't fail any more with a microversion.

The behaviour is not about the SDK (at least not directly), but about `python-openstackclient`.

pierreprinetti:
@mdbooth As a result to our discussion out-of-band, I propose to set `--os-compute-api-version=2.15` for all calls involving server groups. This means exactly the one in "List the Server group", and the one in "Remove the Control Plane server group", just like this PR proposes. However, setting the server group should be properly documented.

@gyohuangxin Can you please comment in all three places (this means including the call in "Create the Control Plane server group") a comment that sounds like:

> os-compute-api-version 2.15 or higher is required for the 'soft-anti-affinity' policy

Would that work for the both of you?

gyohuangxin:
Thank you for all of your reviews!  I'm glad to see it. @pierreprinetti I have updated these comments, thanks!

upi/openstack/down-control-plane.yaml
```diff
@@ -17,9 +17,9 @@
     with_indexed_items: "{{ [os_cp_server_name] * os_cp_nodes_number }}"
 
   - name: 'Remove the Control Plane server group'
-    os_server_group:
-      name: "{{ os_cp_server_group_name }}"
-      state: absent
+    command:
+      # os-compute-api-version 2.15 or higher is required for the 'soft-anti-affinity' policy
+      cmd: "openstack --os-compute-api-version=2.15 server group delete {{ os_cp_server_group_name }}"
```
pierreprinetti, line 22: One last note: this is going to fail if the server group does not exist. Have you experienced errors with the Ansible module? If so, I am afraid that we might re-do the list-parse-use dance we do in the `up` counterpart

gyohuangxin: Yes, it will fail if the server group does not exist.
```
$ ansible-playbook -i inventory.yaml down-control-plane.yaml 

PLAY [all] ************************************************************************************************

TASK [Compute resource names] *****************************************************************************
ok: [localhost]
ok: [bastion]

PLAY [localhost] ******************************************************************************************

TASK [Remove the Control Plane servers] *******************************************************************
ok: [localhost] => (item=[0, 'openshift47-cbtmc-master'])
ok: [localhost] => (item=[1, 'openshift47-cbtmc-master'])
ok: [localhost] => (item=[2, 'openshift47-cbtmc-master'])

TASK [Remove the Control Plane server group] **************************************************************
fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["openstack", "--os-compute-api-version=2.15", "server", "group", "delete", "openshift47-cbtmc-master"], "delta": "0:00:04.377825", "end": "2021-01-24 01:03:52.192066", "msg": "non-zero return code", "rc": 1, "start": "2021-01-24 01:03:47.814241", "stderr": "WARNING: yacc table file version is out of date\nNo servergroup with a name or ID of 'openshift47-cbtmc-master' exists.\n1 of 1 server groups failed to delete.", "stderr_lines": ["WARNING: yacc table file version is out of date", "No servergroup with a name or ID of 'openshift47-cbtmc-master' exists.", "1 of 1 server groups failed to delete."], "stdout": "", "stdout_lines": []}

PLAY RECAP ************************************************************************************************
bastion                    : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
localhost                  : ok=2    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   

```

pierreprinetti:
@gyohuangxin 
Please incorporate the same workflow for the `down-` playbook, to make it idempotent.

I tested this to work: https://github.com/openshift/installer/commit/63740d0a04d86f78dc2167c6c2b186d669d515ae

gyohuangxin:
@pierreprinetti Thanks, has  incorporated it.

mandre (commented):
/lgtm

gyohuangxin:
@pierreprinetti @mandre I think this two failed tests are not related to this pr, could you pls help to merge?

gyohuangxin:
@pierreprinetti @mandre I think this two failed tests are not related to this pr, could you pls help to merge?

gyohuangxin:
@pierreprinetti @mandre I think this two failed tests are not related to this pr, could you pls help to merge?

mdbooth:
/hold cancel

After patient explanation by Pierre I realised I was conflating this PR with another one, and the situations are quite different. In this case we are increasing a microversion to the minimum which supports the required feature, which is fine. My apologies for the noise.

pierreprinetti:
Thank you @gyohuangxin!

gyohuangxin:
Thank all of you and above discussion!

PR #4376, opened 2020-11-13
Cleanup: make conversion functions private

Fedosin:
Now some functions in "conversion" module are public but they are used only inside the module, so it is reasonable to make them
private.

Also this commit fixes typos in deprecated properties descriptions.


Fedosin:
For the context: https://github.com/openshift/installer/pull/4244#discussion_r501112976

staebler (approved):
/lgtm
/approve

PR #4383, opened 2020-11-16
Bug 1891543: openstack: remove platform flavor validation

EmilienM:
Flavors are validated in machine pools already, therefore it is
redundant to also check computeFlavor at the platform level.

This patch removes the computeFlavor validation at the platform
level, since this flavor would already be checked during machine
pool validation.

The installer generates machine pools for both masters and workers.
If the flavor isn't found at the machine pool level, it'll take
platform.computeFlavor, so we don't loose validation coverage.

https://bugzilla.redhat.com/show_bug.cgi?id=1891543

Related: #4323


Fedosin (approved):
/lgtm
/approve
/hold

mandre (approved):
/lgtm
/approve

iamemilio:
/hold cancel

passes e2e-openstack, lgtm

PR #4385, opened 2020-11-16
Bug 1898238: Validate the the API and Ingress FIPs are not the same

iamemilio:
Patches  [OSASINFRA-2178](https://issues.redhat.com/browse/OSASINFRA-2178) where the user is able to pass the same floating IP to the API and Ingress FIP in the install config.

/label platform/openstack


Fedosin (approved):
/lgtm
/approve
/hold

EmilienM (approved):

iamemilio:
We passed e2e openstack, so its safe to remove the hold

/hold cancel

PR #4386, opened 2020-11-16
Bug 1898194: installconfig/gcp/validation: handle custom machine types

jstuever:
Prior to this change, the validation on gcp instance types would return
an error that the type did not exist. This change adds the logic to
properly handle custom machine types while validating their CPU and
Memeory values.

### DIFF
--- a/pkg/asset/installconfig/gcp/client.go
+++ b/pkg/asset/installconfig/gcp/client.go
@@ -19,12 +19,13 @@ import (
 // API represents the calls made to the API.
 type API interface {
 	GetNetwork(ctx context.Context, network, project string) (*compute.Network, error)
-	GetMachineTypes(ctx context.Context, project, filter string) (map[string]*compute.MachineType, error)
+	GetMachineType(ctx context.Context, project, zone, machineType string) (*compute.MachineType, error)
 	GetPublicDomains(ctx context.Context, project string) ([]string, error)
 	GetPublicDNSZone(ctx context.Context, project, baseDomain string) (*dns.ManagedZone, error)
 	GetSubnetworks(ctx context.Context, network, project, region string) ([]*compute.Subnetwork, error)
 	GetProjects(ctx context.Context) (map[string]string, error)
 	GetRecordSets(ctx context.Context, project, zone string) ([]*dns.ResourceRecordSet, error)
+	GetZones(ctx context.Context, project, filter string) ([]*compute.Zone, error)
 	GetEnabledServices(ctx context.Context, project string) ([]string, error)
 }
 
@@ -49,10 +50,8 @@ func NewClient(ctx context.Context) (*Client, error) {
 	return client, nil
 }
 
-// GetMachineTypes uses the GCP Compute Service API to get a list of machine types from a project.
-func (c *Client) GetMachineTypes(ctx context.Context, project, filter string) (map[string]*compute.MachineType, error) {
-	types := map[string]*compute.MachineType{}
-
+// GetMachineType uses the GCP Compute Service API to get the specified machine type.
+func (c *Client) GetMachineType(ctx context.Context, project, zone, machineType string) (*compute.MachineType, error) {
 	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
 	defer cancel()
 
@@ -61,22 +60,12 @@ func (c *Client) GetMachineTypes(ctx context.Context, project, filter string) (m
 		return nil, err
 	}
 
-	req := svc.MachineTypes.AggregatedList(project)
-	if filter != "" {
-		req = req.Filter(filter)
+	req, err := svc.MachineTypes.Get(project, zone, machineType).Context(ctx).Do()
+	if err != nil {
+		return nil, err
 	}
 
-	if err := req.Pages(ctx, func(page *compute.MachineTypeAggregatedList) error {
-		for _, scopedList := range page.Items {
-			for _, item := range scopedList.MachineTypes {
-				types[item.Name] = item
-			}
-		}
-		return nil
-	}); err != nil {
-		return nil, errors.Wrapf(err, "failed to get machine types from project %s", project)
-	}
-	return types, nil
+	return req, nil
 }
 
 // GetNetwork uses the GCP Compute Service API to get a network by name from a project.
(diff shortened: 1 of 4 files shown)


pkg/asset/installconfig/gcp/validation.go
```diff
@@ -66,8 +67,36 @@ func ValidateInstanceType(client API, ic *types.InstallConfig, fieldPath *field.
 			errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d MB Memory", req.minimumMemory)
 			allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
 		}
+	} else if index := strings.Index(instanceType, "custom-"); index != -1 {
+
+		// Ensure the CPU and MEMORY are defined in custom instance type names
+		if customConfig := strings.Split(instanceType[index+len("custom-"):], "-"); len(customConfig) == 2 {
```
staebler, line 73: How would you feel about using a regular expression rather than `strings.Index` and `strings.Split`? The regular expression makes it a bit more concise and (arguably) conveys the expected format for custom instance types more clearly. I am fine if you want to keep it as is, too.
```
regexp.MustCompile(`^(\w+-)?custom-(\d+)-(\d+)$`)
```

staebler (commented):
Could you add some unit tests for custom instance types?

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -66,8 +67,36 @@ func ValidateInstanceType(client API, ic *types.InstallConfig, fieldPath *field.
...
+				errMsg := "instance type not found"
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+			} else if customCpus < req.minimumVCpus {
+				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d vCPUs", req.minimumVCpus)
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+			}
+
+			// Ensure the custom Memory is a valid integer
+			if customMemory, err := strconv.ParseInt(customConfig[1], 10, 0); err != nil {
+				errMsg := "instance type not found"
```
patrickdillon, line 86: similar suggestion as for CPUs. With the current error message we would have the message "instance type not found" repeated twice for an invalid instance type like "custom-x-x"

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -66,8 +67,36 @@ func ValidateInstanceType(client API, ic *types.InstallConfig, fieldPath *field.
...
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+			} else if customMemory < req.minimumMemory {
+				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d MB Memory", req.minimumMemory)
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+			}
+
+		} else {
+			errMsg := "instance type not found"
```
patrickdillon, line 94: errMsg := "invalid custom instance type"

pkg/asset/installconfig/gcp/validation.go
abhinavdahiya: I think we should maybe request the instance type from GCP api instead of this string-extraction. ServiceDelivery previously mentioned how these can change.

pkg/asset/installconfig/gcp/validation.go
jstuever: No longer in current code.

pkg/asset/installconfig/gcp/validation.go
jstuever: Switched to using the GCP API to reduce future maintenance on this code.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -44,30 +44,20 @@ func Validate(client API, ic *types.InstallConfig) error {
...
-	if instanceType == "" {
-		return nil
-	}
-
-	filter := fmt.Sprintf("name = %s", instanceType)
-	instanceTypes, err := client.GetMachineTypes(context.TODO(), ic.GCP.ProjectID, filter)
+	typeMeta, err := client.GetMachineType(context.TODO(), project, zone, instanceType)
 	if err != nil {
-		return append(allErrs, field.InternalError(fieldPath, err))
+		return append(allErrs, field.InternalError(fieldPath.Child("type"), err))
```
staebler, line 52: If the error here is a not-found error, then the field error should be Invalid instead of InternalError.

pkg/asset/installconfig/gcp/client.go
```diff
@@ -233,6 +222,35 @@ func (c *Client) GetProjects(ctx context.Context) (map[string]string, error) {
 	return projects, nil
 }
 
+// GetZones uses the GCP Compute Service API to get a list of machine types from a project.
+func (c *Client) GetZones(ctx context.Context, project, filter string) (map[string]*compute.Zone, error) {
```
staebler, line 226: I think we should just return a slice of zones rather than going through the effort of making a map. This makes it easier on the only current caller, too.
```suggestion
func (c *Client) GetZones(ctx context.Context, project, filter string) ([]*compute.Zone, error) {
```

staebler, line 226: There is some heavy overlap between this new function and the existing `AvailabilityZones` function. Can we avoid this duplication?

https://github.com/openshift/installer/blob/692bdd1255f7a3076b637f5bb77fdc4a0e71f12a/pkg/asset/machines/gcp/zones.go#L16

staebler (commented):
I like this approach better. Nice!

pkg/asset/installconfig/gcp/client.go
jstuever: @staebler I was thinking the same. I'm assuming we don't want to simply import the machines/gcp package here. However, I envision this being something in /pkg/client/gcp or similar so that it can be shared across assets (such as destroy). This feels more like an RFE or tech debt. Thoughts?

staebler: Oh, yeah. I was thinking that the gcp client was in a more accessible package, not tucked away in the installconfig assets package. That is certainly beyond the scope of what this BZ fix should cover. Would you mind creating an Jira card that we can throw in the backlog to cover making the cloud clients more accessible?

pkg/asset/installconfig/gcp/client.go
```diff
@@ -233,6 +222,35 @@ func (c *Client) GetProjects(ctx context.Context) (map[string]string, error) {
 	return projects, nil
 }
 
+// GetZones uses the GCP Compute Service API to get a list of machine types from a project.
```
patrickdillon, line 225: copypasta

pkg/asset/installconfig/gcp/client.go
jstuever: https://issues.redhat.com/browse/CORS-1598

pkg/asset/installconfig/gcp/client.go
jstuever: Fixed: https://github.com/openshift/installer/commit/60a57050d1738c13289e883ac5f32cf338d2ea0f

pkg/asset/installconfig/gcp/client.go
jstuever: Done: https://github.com/openshift/installer/commit/60a57050d1738c13289e883ac5f32cf338d2ea0f

pkg/asset/installconfig/gcp/validation.go
jstuever: Fixed: https://github.com/openshift/installer/commit/60a57050d1738c13289e883ac5f32cf338d2ea0f

staebler (approved):
/lgtm
/approve

jstuever:
/hold cancel
e2e-gcp passed.

PR #4387, opened 2020-11-17
Bug 1895874: Use oVirt Engine certificate verified by the user for installation

janoszen:
## Abstract

This pull request changes the oVirt/RVH installer to use the certificate downloaded from the oVirt engine and verified by the user. It eliminates the need for the user to add the certificate to the trust root on their system.

This PR is based on #4354 and fixes [Bugzilla#1895874](https://bugzilla.redhat.com/show_bug.cgi?id=1895874).

## Behavior before this change

> $ openshift-install create install-config
> ? Platform ovirt
> ? Engine FQDN[:PORT] engine.example.com
> INFO Loaded the following PEM file:
> INFO    Version: 3
> INFO    Signature Algorithm: SHA256-RSA
> INFO    Serial Number: 4096
> INFO    Issuer: CN=engine.example.com.39197,O=example.com,C=US
> INFO    Validity:
> INFO            Not Before: 2020-08-11 13:50:26 +0000 UTC
> INFO            Not After: 2030-08-10 13:50:26 +0000 UTC
> INFO    Subject: CN=engine.example.com.39197,O=example.com,C=US
> ? Would you like to use the above certificate to connect to Engine?  Yes
> ? Engine username admin
> ? Engine password [Press Ctrl+C to switch username, ? for help] ***********
> X Sorry, your reply was invalid: failed to connect to Engine platform Post "https://engine.example.com/ovirt-engine/sso/oauth/token": x509: certificate signed by unknown authority

## Behavior after this change

The installation no longer fails.

## Verification

I have verified the installation process against the 4.7.0-0.ci-2020-11-17-102916 image up to the point that the bootstrap node comes up and starts installing the cluster.

### DIFF
--- a/pkg/asset/installconfig/ovirt/client.go
+++ b/pkg/asset/installconfig/ovirt/client.go
@@ -15,6 +15,7 @@ func getConnection(ovirtConfig Config) (*ovirtsdk.Connection, error) {
 		Username(ovirtConfig.Username).
 		Password(ovirtConfig.Password).
 		CAFile(ovirtConfig.CAFile).
+		CACert([]byte(ovirtConfig.CABundle)).
 		Insecure(ovirtConfig.Insecure).
 		Build()
 	if err != nil {
--- a/pkg/asset/installconfig/ovirt/config.go
+++ b/pkg/asset/installconfig/ovirt/config.go
@@ -2,7 +2,6 @@ package ovirt
 
 import (
 	"crypto/x509"
-	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
--- a/pkg/asset/installconfig/ovirt/config.go
+++ b/pkg/asset/installconfig/ovirt/config.go
@@ -2,6 +2,7 @@ package ovirt
 
 import (
 	"crypto/x509"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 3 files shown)


pkg/asset/installconfig/ovirt/ovirt.go
```diff
@@ -1,41 +1,39 @@
...
 )
 
 // Platform collects ovirt-specific configuration.
 func Platform() (*ovirt.Platform, error) {
 	p := ovirt.Platform{}
 
+	var c *ovirtsdk.Connection
 	ovirtConfig, err := NewConfig()
+	if err == nil {
+		c, err = ovirtConfig.getValidatedConnection()
```
Gal-Zaidman, line 18: I don't understand why we need to call ovirtConfig.getValidatedConnection twice.
Why not just replace the "c, err := ovirtsdk4.NewConnectionBuilder()...." with "c, err = ovirtConfig.getValidatedConnection()" ?

If I understand this, then:
NewConfig() is the flow if using the install-config and engineSetup() is the flow of the survey, so if the NewConfig has been successful then we can get a connection without the need to fall back to the engineSetup config

janoszen: This is from PR #4354. That one is not merged yet, but it doesn't make sense to do this change until that one is done.

janoszen: The reason for getting the connection twice is the setup process: the first validates the config from the ovirt config. If that fails we re-query the user for details.

Gal-Zaidman: Oh now I see that this PR depends on 4354... sorry missed that part.... and that explains the logic here. Lets resolve this conversation and continue the discussion in the appropriate place

janoszen:
@Gal-Zaidman can we take care of this next?

Gal-Zaidman:
/lgtm

please remove the WIP label

PR #4389, opened 2020-11-17
OSASINFRA-902: Update OpenStack Custom External LB and DNS Docs

iamemilio:
The underlying network architecture has changed a lot since these docs
were initially written. We want to make sure that these docs are accurate
and up to date so that users with complex networking use cases like workers
on a custom subnet and baremetal workers are able to manage their ingress/egress
traffic as needed. More up to date examples and reference information has been added.

We have chosen to omit sections outlining how to replace the internal lb and dns
services since they were inaccurate. We are targeting an upcoming release to handle
these features better given the complexity of our current networking architecture.

### DIFF
--- a/docs/user/openstack/README.md
+++ b/docs/user/openstack/README.md
@@ -38,7 +38,7 @@ In addition, it covers the installation with the default CNI (OpenShiftSDN), as 
     - [Adding a MachineSet](#adding-a-machineset)
       - [Using a Server Group](#using-a-server-group)
       - [Setting Nova Availability Zones](#setting-nova-availability-zones)
-    - [Using an External Load Balancer](#using-an-external-load-balancer)
+    - [Using a Custom External Load Balancer](#using-a-custom-external-load-balancer)
     - [Refreshing a CA Certificate](#refreshing-a-ca-certificate)
   - [Reporting Issues](#reporting-issues)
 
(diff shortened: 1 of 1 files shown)


maxwelldb:
@iamemilio Would you like suggestions on this, or is the review request more of an FYI only thing?

iamemilio:
Yes please!

docs/user/openstack/README.md
```diff
@@ -526,84 +526,110 @@ property of the MachineSet.
...
   "platform": "linux/amd64"
 }
 ```
 
-Another useful thing to check is that the ignition configurations are only available from within the deployment. The following command should only succeed from a node in the OpenShift cluster:
+Note: the versions may be different, but as long as you get a json payload response, it worked correctly.
+
+#### Verifying Apps Reachable
+
+An easy way to check that the workers are load balanced correctly is to try to access the OpenShift console. This can be done from your web browser. If you don't have access to a web browser, then you can try to query it with the following curl command:
```
maxwelldb, line 609:
```suggestion
The simplest way to verify that apps are reachable is to open the OpenShift console in a web browser. 

If you don't have access to a web browser, query the console by using the following `curl` command:
```

maxwelldb (approved):
Made a number of comments. Feel free to accept/trash/interrogate however you like.

docs/user/openstack/README.md
```diff
@@ -526,84 +526,108 @@ property of the MachineSet.
...
+        server <cluster-name>-worker-2 192.168.1.253:80 check
 ```
 
-You could also specify a specific IP address with /32 if you wish.
+#### DNS Lookups
 
-You can verify the operation of the load balancer now if you wish, using the curl commands given below.
-
-Now the DNS entry for `api.<cluster name>.<base domain>` needs to be updated to point to the new load balancer:
+To ensure that your API and apps are accessible through your load balancer, [create or update DNS entries](#Create API and Ingress DNS Records) for them. To use your new load balancing service for external traffic, make sure the IP address for these DNS entries is the IP address your load balancer is reachable at.
```
mandre, line 572: The anchor should be #create-api-and-ingress-dns-records, otherwise the link doesn't render. The sentence ends bizarrely too.
```suggestion
To ensure that your API and apps are accessible through your load balancer, [create or update DNS entries](#create-api-and-ingress-dns-records) for them. To use your new load balancing service for external traffic, make sure the IP address for these DNS entries is the IP address your load balancer:
```

iamemilio:
removing the hold, reviewers can lgtm again when ready

/hold cancel

PR #4395, opened 2020-11-18
WIP: [fcos] Azure TF fixup

LorbusChris:
The source_uri field on the rhcos_image resource was renamed to source
in a prior change on the fcos branch. The source field name however
might have special functionality and would thus not be usable here.

### DIFF
--- a/data/data/azure/main.tf
+++ b/data/data/azure/main.tf
@@ -183,8 +183,8 @@ resource "azurerm_storage_blob" "rhcos_image" {
   storage_account_name   = azurerm_storage_account.cluster.name
   storage_container_name = azurerm_storage_container.vhd.name
   type                   = "Page"
-  source                 = var.azure_image_url
-  metadata               = map("source", var.azure_image_url)
+  source_uri             = var.azure_image_url
+  metadata               = map("source_uri", var.azure_image_url)
 }
 
 resource "azurerm_image" "cluster" {


data/data/azure/main.tf
```diff
@@ -183,8 +183,8 @@ resource "azurerm_storage_blob" "rhcos_image" {
   storage_account_name   = azurerm_storage_account.cluster.name
   storage_container_name = azurerm_storage_container.vhd.name
   type                   = "Page"
-  source                 = var.azure_image_url
-  metadata               = map("source", var.azure_image_url)
+  source_uri             = var.azure_image_url
```
staebler, line 186: The commit [1] that changed to `source` indicates that the image was going to be on the local filesystem. Is this no longer true? Or was it never true?

[1] https://github.com/openshift/installer/commit/12292c7314e65b1b38abfb3088bc6c711b47563f

LorbusChris: Thank you for linking the original commit!
I don't think it was ever true and judging from the [error a user reported](https://github.com/openshift/okd/issues/371#issue-736430937) the source field contains a URL and not a local file path.
That said, I don't feel super confident about whether or not this change will fix the reported issue either.
Unfortunately we currently don't have Azure e2e testing for OKD.

@vrutkovs (who's currently on PTO), might have some more insights.
I'm also wondering whether we could add the Azure e2e tests even though FCOS images aren't being uploaded to that platform yet (could it not just pull the image from the Fedora servers?)

staebler: I am not very familiar with the OKD installer, but I don't see anything that would change `azure_image_url` from the value stored in fcos.json [1]. That URI is certainly not an absolute path to a file on the local filesystem.

[1] https://github.com/openshift/installer/blob/fcos/data/data/fcos.json#L66

vrutkovs:
This needs to be conditional, so that we could merge this in master effortlessly

data/data/azure/main.tf
vrutkovs: >I'm also wondering whether we could add the Azure e2e tests

Sure thing, but the installer need to run in Azure VM, fetch the URL locally and use local path in `azure_image_url` as Matthew suggested. We had this issue in 4.4/4.5 as well - with no FCOS uploaded to Azure we can't use installer flow and have to hack our way through it

mjudeikis: Just thinking out loud, not for this PR. So this is 3rd iteration of How we "publish" RHCOS image for our different versions of the product. I think we might be better off to refactor this in a way to fit all use cases. Will drop a separate email for this to @staebler :)

ARO for azure is using a published image from Azure Marketplace so we have an Installer carry-on patch with something under lines: 
```
if platform.Image != nil {
		if platform.Image.ResourceID != "" {
			spec.Image = azureprovider.Image{
				ResourceID: platform.Image.ResourceID,
			}
		} else {
			spec.Image = azureprovider.Image{
				Publisher: platform.Image.Publisher,
				Offer:     platform.Image.Offer,
				SKU:       platform.Image.SKU,
				Version:   platform.Image.Version,
			}
		}
	}
```

LorbusChris:
#4398 contains the proper fix.
If anybody can come up with a solution that can merge into master, that'd be even better :)

PR #4399, opened 2020-11-20
libvirt: allow customization of the pool path

zregvart:
When creating the storage pool path, that path needs to be on a file
system big enough to accommodate virtual machine images spun up by the
installer. In some cases the default `/var/lib/libvirt/openshift-images`
could reside on a file system that if filled might cause issues, for
example when `/var` resides on the root file system, as it does by
default on Fedora Linux.

This allows customizing the libvirt storage pool directory path,
allowing the user to choose where to store the virtual machine images.

The default is left to `/var/lib/libvirt/openshift-images`.

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -1078,6 +1078,10 @@ spec:
                         description: The interface make used for the network. Default is tt0.
                         type: string
                     type: object
+                  poolPath:
+                    default: /var/lib/libvirt/openshift-images
+                    description: Directory path for the libvirt storage pool
+                    type: string
                 type: object
               none:
                 description: None is the empty configuration used when installing on an unsupported platform.
--- a/data/data/libvirt/main.tf
+++ b/data/data/libvirt/main.tf
@@ -5,7 +5,7 @@ provider "libvirt" {
 resource "libvirt_pool" "storage_pool" {
   name = var.cluster_id
   type = "dir"
-  path = "/var/lib/libvirt/openshift-images/${var.cluster_id}"
+  path = "${var.libvirt_pool_path}/${var.cluster_id}"
 }
 
 module "volume" {
@@ -126,4 +126,3 @@ data "libvirt_network_dns_host_template" "masters_int" {
   ip       = var.libvirt_master_ips[count.index]
   hostname = "api-int.${var.cluster_domain}"
 }
-
--- a/data/data/libvirt/variables-libvirt.tf
+++ b/data/data/libvirt/variables-libvirt.tf
@@ -58,3 +58,9 @@ variable "libvirt_master_size" {
   description = "Size of the volume in bytes"
   default     = "17179869184"
 }
+
+variable "libvirt_pool_path" {
+  type        = string
+  description = "Directory path for the libvirt storage pool"
+  default     = "/var/lib/libvirt/openshift-images"
+}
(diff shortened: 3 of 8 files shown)


patrickdillon:
/ok-to-test 
if you're adding new fields to install config you will probably need to generate new `explain` docs: https://github.com/openshift/installer/blob/master/docs/dev/explain.md

pkg/asset/installconfig/libvirt/libvirt.go
```diff
@@ -22,13 +22,27 @@ func Platform() (*libvirt.Platform, error) {
 			},
 			Validate: survey.ComposeValidators(survey.Required, uriValidator),
 		},
-	}, &uri)
-	if err != nil {
+	}, &uri); err != nil {
+		return nil, errors.Wrap(err, "failed UserInput")
+	}
+
+	var poolPath string
```
staebler, line 29: Do not add this as a question for the user. This does not meet the bar for inclusion in the guided-install path.

vendor/github.com/openshift/cluster-api-provider-libvirt/pkg/apis/libvirtproviderconfig/v1beta1/libvirtmachineproviderconfig_types.go
```diff
@@ -33,6 +33,7 @@ type LibvirtMachineProviderConfig struct {
 	NetworkUUID              string     `json:"networkUUID"`
 	Autostart                bool       `json:"autostart"`
 	URI                      string     `json:"uri"`
+	PoolPath                 string     `json:"poolPath"`
```
staebler, line 36: You cannot add this field. This is vendored code. The machine-api-operator is not responsible for creating libvirt pools. As such, this is not be an appropriate field to add to the machine spec, anyway.

zregvart:
Thanks for the review @staebler, can you have another look?

cfergeau:
> When creating the storage pool path, that path needs to be on a file
> system big enough to accommodate virtual machine images spun up by the
> installer. In some cases the default `/var/lib/libvirt/openshift-images`
> could reside on a file system that if filled might cause issues, for
> example when `/var` resides on the root file system, as it does by
> default on Fedora Linux.
> 
> This allows customizing the libvirt storage pool directory path,
> allowing the user to choose where to store the virtual machine images.
> 
> The default is left to `/var/lib/libvirt/openshift-images`.

Just wanted to note that this issue can easily be worked around with `mount --bind /some/where /var/lib/libvirt/openshift-images` (or the opposite way, I always forget the right order for these args).

zregvart:
@cfergeau true, or via a symbolic link. I still think it would make sense to have the path configurable.

staebler:
> @cfergeau true, or via a symbolic link. I still think it would make sense to have the path configurable.

I'm inclined towards this not meeting the bar for inclusion in the install config, given that there are other ways around it that do not seem too onerous.

zregvart:
@cfergeau and @staebler I fully respect your decisions here, though I would like to see that bar you keep mentioning documented so in the future someone else proposing a change would have the ability to see where that bar actually is. Feel free to close this if you don't wish to accept this.

staebler:
I will let @praveenkumar and @cfergeau weigh in on whether they want to carry the maintenance burden of this additional field. I don't use libvirt enough to have a good feel for how many users would make use of this customization or how cumbersome the workarounds are.

PR #4400, opened 2020-11-20
WIP: Bug 1857945: add confirmation for using insecure mode to ovirt

janoszen:
This change adds a confirmation question to enable insecure mode when decoding the certificate from the oVirt engine has failed. This fixes [Bugzilla bug #1857945](https://bugzilla.redhat.com/show_bug.cgi?id=1857945).

## User experience

```
? Platform ovirt
? Engine FQDN[:PORT] vm-10-208.lab.eng.tlv2.redhat.com
INFO    Version: 3
INFO    Serial Number: 4096
INFO    Issuer: CN=ovirt-engine.example.com.39197,O=example.com,C=US
INFO    Validity:
INFO            Not Before: 2020-08-11 13:50:26 +0000 UTC
INFO            Not After: 2030-08-10 13:50:26 +0000 UTC
INFO    Subject: CN=ovirt-engine.example.com.39197,O=example.com,C=US
? Would you like to use the above certificate to connect to Engine?  No
? Would you like to import another PEM bundle? No
? Failed to load engine certificate. Do you want to continue in insecure mode? No
FATAL failed to fetch Install Config: failed to fetch dependency of "Install Config": failed to fetch dependency
of "Base Domain": failed to generate asset "Platform": could not load engine certificate, not proceeding in
insecure mode
```

**Note:** retries are implemented in #4354

### DIFF
--- a/pkg/asset/installconfig/ovirt/credentials.go
+++ b/pkg/asset/installconfig/ovirt/credentials.go
@@ -316,18 +316,19 @@ func engineSetup() (Config, error) {
 		answer, err := askQuestionTrueOrFalse(
 			"Would you like to continue?",
 			"By not using a trusted CA, insecure connections can "+
-				"cause man-in-the-middle attacks among many others.")
+				"cause man-in-the-middle attacks among others.")
 		if err != nil || !answer {
 			return engineConfig, err
 		}
 	} else {
 		err = showPEM(httpResource.saveFilePath)
 		if err != nil {
+			logrus.Warning("could not decode oVirt engine certificate", err)
 			engineConfig.Insecure = true
 		} else {
 			answer, err := askQuestionTrueOrFalse(
 				"Would you like to use the above certificate to connect to Engine? ",
-				"Certificate to connecto with Engine. Make sure this cert CA is trusted locally.")
+				"Certificate to connect to with Engine. Make sure this cert CA is trusted locally.")
 			if err != nil {
 				return engineConfig, err
 			}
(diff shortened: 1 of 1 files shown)


staebler:
I am a bit late to this BZ, but I am not in favor of asking this question of the user. We should not be making it easy for the user to use insecure communication. I would much rather see us fail the installation rather than setting the configuration to allow insecure communication. If the user **really** wants to use insecure communication, then they should have to set up the configuration themselves.

staebler:
In addition to not wanting to make it easy to use insecure communication, there should be a high bar for adding prompts for the user. The intention of the guided installation is to provide a very small set of prompts limited to items that are required for all or almost all typical installations. The insecure prompt does not meet this bar.

janoszen:
> In addition to not wanting to make it easy to use insecure communication, there should be a high bar for adding prompts for the user. The intention of the guided installation is to provide a very small set of prompts limited to items that are required for all or almost all typical installations. The insecure prompt does not meet this bar.

The behavior so far was to turn on insecure mode and only inform the user, not ask them, if the certificate download failed or the PEM cannot be decoded. If insecure mode was not enabled the user had to trust the certificate on a system level or add the certificate to `~/.ovirt/ovirt-config.yaml` (undocumented AFAIK). #4387 adds the ability for the user to save the certificate to `ovirt-config.yaml`, which could potentially be a path forward to removing insecure mode entirely. I would not remove insecure mode before that patch has been merged.

janoszen:
Added a BZ for removing insecure mode: FYI, added a bug for removing insecure mode: https://bugzilla.redhat.com/show_bug.cgi?id=1900138

janoszen:
Closing this PR in favor of #4404. /cc @staebler

PR #4401, opened 2020-11-21
OpenStack: Adds clusterOSImageProperties to install-config.yaml

adduarte:
ClusterOSImageProperty is a list of properties to be added to the cluster OS glance image metadata. 
example usage
platform:
  openstack:
      clusterOSImageProperties:
           hw_qemu_guest_agent: yes
           hw_scsi_model: virtio-scsi

implementes [OSASINFRA-1980](https://issues.redhat.com/browse/OSASINFRA-1980)

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -1097,6 +1097,11 @@ spec:
                   clusterOSImage:
                     description: ClusterOSImage is either a URL with `http(s)` or `file` scheme to override the default OS image for cluster nodes, or an existing Glance image name.
                     type: string
+                  clusterOSImageProperties:
+                    additionalProperties:
+                      type: string
+                    description: 'ClusterOSImageProperties is a list of properties to be added to the metadata of the uploaded Glance ClusterOSImage. Default: the default is to not set any properties.'
+                    type: object
                   computeFlavor:
                     description: FlavorName is the name of the compute flavor to use for instances in this cluster.
                     type: string
--- a/docs/user/openstack/customization.md
+++ b/docs/user/openstack/customization.md
@@ -30,6 +30,7 @@ Beyond the [platform-agnostic `install-config.yaml` properties](../customization
 * `region` (deprecated string): The OpenStack region where the cluster will be created. Currently this value is not used by the installer.
 * `trunkSupport` (deprecated string): Whether OpenStack ports can be trunked (`1` for true or `0` for false)
 * `clusterOSImage` (optional string): Either a URL with `http(s)` or `file` scheme to override the default OS image for cluster nodes or an existing Glance image name.
+* `clusterOSImageProperties` (optional list of strings): a list of properties to be added to the installer-uploaded ClusterOSImage in Glance. The default is to not set any properties. `clusterOSImageProperties` is ignored when `clusterOSImage` points to an existing image in Glance.
 * `apiVIP` (optional string): An IP address on the machineNetwork that will be assigned to the API VIP. Be aware that the `10` and `11` of the machineNetwork will be taken by neutron dhcp by default, and wont be available.
 * `ingressVIP` (optional string): An IP address on the machineNetwork that will be assigned to the ingress VIP. Be aware that the `10` and `11` of the machineNetwork will be taken by neutron dhcp by default, and wont be available.
 * `machinesSubnet` (optional string): the UUID of an OpenStack subnet to install the nodes of the cluster onto. For more information on how to install with a custom subnet, see the [custom subnets](#custom-subnets) section of the docs.
--- a/pkg/tfvars/openstack/rhcos_image.go
+++ b/pkg/tfvars/openstack/rhcos_image.go
@@ -16,7 +16,7 @@ import (
 )
 
 // uploadBaseImage creates a new image in Glance and uploads the RHCOS image there
-func uploadBaseImage(cloud string, localFilePath string, imageName string, clusterID string) error {
+func uploadBaseImage(cloud string, localFilePath string, imageName string, clusterID string, imageProperties map[string]string) error {
 	logrus.Debugln("Creating a Glance image for RHCOS...")
 
 	f, err := os.Open(localFilePath)
@@ -46,6 +46,7 @@ func uploadBaseImage(cloud string, localFilePath string, imageName string, clust
 		ContainerFormat: "bare",
 		DiskFormat:      diskFormat,
 		Tags:            []string{fmt.Sprintf("openshiftClusterID=%s", clusterID)},
+		Properties:      imageProperties,
 		// TODO(mfedosin): add Description when gophercloud supports it.
 	}
 
--- a/pkg/types/openstack/platform.go
+++ b/pkg/types/openstack/platform.go
@@ -62,6 +62,11 @@ type Platform struct {
 	// +optional
 	ClusterOSImage string `json:"clusterOSImage,omitempty"`
 
+	// ClusterOSImageProperties is a list of properties to be added to the metadata of the uploaded Glance ClusterOSImage.
+	// Default: the default is to not set any properties.
+	// +optional
+	ClusterOSImageProperties map[string]string `json:"clusterOSImageProperties,omitempty"`
+
 	// APIVIP is the static IP on the nodes subnet that the api port for openshift will be assigned
 	// Default: will be set to the 5 on the first entry in the machineNetwork CIDR
 	// +optional
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -411,6 +411,7 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			installConfig.Config.Platform.OpenStack.APIVIP,
 			installConfig.Config.Platform.OpenStack.IngressVIP,
 			string(*rhcosImage),
+			installConfig.Config.Platform.OpenStack.ClusterOSImageProperties,
 			clusterID.InfraID,
 			caCert,
 			bootstrapIgn,
--- a/pkg/tfvars/openstack/openstack.go
+++ b/pkg/tfvars/openstack/openstack.go
@@ -45,7 +45,7 @@ type config struct {
 }
 
 // TFVars generates OpenStack-specific Terraform variables.
-func TFVars(masterConfigs []*v1alpha1.OpenstackProviderSpec, cloud string, externalNetwork string, externalDNS []string, apiFloatingIP string, ingressFloatingIP string, apiVIP string, ingressVIP string, baseImage string, infraID string, userCA string, bootstrapIgn string, mpool *types_openstack.MachinePool, machinesSubnet string) ([]byte, error) {
+func TFVars(masterConfigs []*v1alpha1.OpenstackProviderSpec, cloud string, externalNetwork string, externalDNS []string, apiFloatingIP string, ingressFloatingIP string, apiVIP string, ingressVIP string, baseImage string, baseImageProperties map[string]string, infraID string, userCA string, bootstrapIgn string, mpool *types_openstack.MachinePool, machinesSubnet string) ([]byte, error) {
 	zones := []string{}
 	seen := map[string]bool{}
 	for _, config := range masterConfigs {
@@ -103,7 +103,7 @@ func TFVars(masterConfigs []*v1alpha1.OpenstackProviderSpec, cloud string, exter
 			return nil, errors.Errorf("Unsupported URL scheme: '%v'", url.Scheme)
 		}
 
-		err = uploadBaseImage(cloud, localFilePath, imageName, infraID)
+		err = uploadBaseImage(cloud, localFilePath, imageName, infraID, baseImageProperties)
 		if err != nil {
 			return nil, err
 		}


iamemilio:
Why are we adding this explicitly to the installer? This is was not intended to be customer facing. Is there a Jira card or Bugzilla that you can link to explain these changes?

iamemilio:
@adduarte what is the goal of these code changes? Is there a BZ or Jira issue associated with this?

iamemilio:
References: OSASINFRA-1980

abhinavdahiya:
i would say instead of a general API that allows users to add _any_ property to the image is to restrict what users can ask the installer to do to very specific use-case. And if the users want more control asking them to take control and giving the installer the image directly.

mandre:
/retest

@abhinavdahiya It's not clear to me if you're suggesting to implement validation for `clusterOSImageProperties` and reject properties that haven't been vetted by the developers, or instead ask to not expose the image properties and rather have new options per use case, for example `enableQEMUGestAgent`, or `useVirtioSCSI`.

adduarte:
@abhinavdahiya  and @mandre, with respect to adding a more specific command. 
I originally thought it would be a good way forward to move that way, but then the problem is that  the properties being passed along to glance is just metadata composed of strings. 
This means that if we use something like enableQEMUGestAgent, or useVirtioSCSI, the problem is that if the required string we have to pass into glance changes (because of openstack version being used) then we will break. 

The moment we use something like enableQEMUGestAgent, or useVirtioSCSI we will lock the installer into a very specific string. That could cause problems down the line if openstack changes the requirements. 

After all that said though, we can easily implement "enableQemuGestAgent" and "userVirtioSCSI", I am just not sure we need to make the feature that complicated at this time.
To enable the guest agent the user will either use: 

- platform: 
        - clusterOSImageProperties:
           - "qemu-agent = yes"
           - "virtioscsi=yes"

or

- platform: 
       - enableQemuGestAgent
       - useVritioSCSI


I don't see the value at this time to use the second method.  And actually the second method suggests that the installer will not only add the correct glance properties but also download the appropriate rhcos image. And at this time we don't have the functionality to do that.

adduarte:
@mandre and @abhinavdahiya , 
Not sure if we can use something like "enableQEMUGestAgent", as that would imply that the installer will enable the qemu guest agent everywhere, including rhel workers, which is not the case. The installer cannot enable the qemu guest agent there. 

This patch was left very general to be able to configure which properties (metadata) is added to the coreos image being uploaded. 
Also, if we use soemthing like enableQEMUGestAgent we end up locking the string that is to be given to glance. This could cause problems if openstack changes that string from one version to another.  in other words if openstack changes the string that is required in the metadata to enable qemu agent support, then we won't be able to support multiple versions of openstack, because we hardcoded the string. 

We are faced with two options: 
(this patch)
- Platform:
    coreOSImageProperty:
      - "hw_qemu_agent=yes"
      - "use_virtio_scsci=yes"

or 

- Platform: 
      - enableQEMUGestAgent: yes
      - enableVirtioSCSCI: yes

besides the extra typing, I don't see much value to using the former vs the later. but I can push a pull request implementing it to see what the differences would be so we can decide.

adduarte:
created https://github.com/openshift/installer/pull/4433 which implements using UseHwQEMUAgentForClusterOSImage, so we can compare and decide which one we want to use.

adduarte:
@mandre and @abhinavdahiya the latest changes reduce the acutal number of lines and I belive gives us the necessary control where we can implement a  validation control where we can limit the properties we allow the user to set if required (in a follow up pr)

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -1097,6 +1097,11 @@ spec:
                   clusterOSImage:
                     description: ClusterOSImage is either a URL with `http(s)` or `file` scheme to override the default OS image for cluster nodes, or an existing Glance image name.
                     type: string
+                  clusterOSImageProperties:
+                    additionalProperties:
+                      type: string
+                    description: ClusterOSImageProperties is a list of properties to be added to the metadata of the uploaded Glance ClusterOSImage. in the install-config.yaml. hw_qemu_agent=yes would look like
```
mandre, line 1103:
```suggestion
                    description: ClusterOSImageProperties is a list of properties to be added to the metadata of the uploaded Glance ClusterOSImage.
```

pkg/types/openstack/platform.go
```diff
@@ -62,6 +62,11 @@ type Platform struct {
 	// +optional
 	ClusterOSImage string `json:"clusterOSImage,omitempty"`
 
+	// ClusterOSImageProperties is a list of properties to be added to the metadata of the uploaded Glance ClusterOSImage.
+	// in the install-config.yaml. hw_qemu_agent=yes would look like
```
mandre, line 66: Depending on whether it's possible or not to have different defaults for OKD / OCP, where we can assume the hypervisor is libvirt/kvm. 
```suggestion
	// Default: the default is to not set any properties.
```

data/data/install.openshift.io_installconfigs.yaml
iamemilio: Because this is not platform specific, I think that we should remove any reference to Glance or OpenStack in this. It should just say it will be added to the image metadata.

iamemilio:
We will need to be very about how we document the usage of this feature, and what we do and do not support.

adduarte:
Just making it explicit
/lgtm
/approve

adduarte:
for reference: The user input model is based on similar use by other platforms: https://github.com/openshift/installer/blob/master/pkg/types/aws/platform.go#L33

mandre (commented):
/approve
/lgtm

iamemilio:
needs approval from @abhinavdahiya

abhinavdahiya:
> needs approval from @abhinavdahiya

/assign @staebler

pkg/types/openstack/platform.go
```diff
@@ -62,6 +62,11 @@ type Platform struct {
 	// +optional
 	ClusterOSImage string `json:"clusterOSImage,omitempty"`
 
+	// ClusterOSImageProperties is a list of properties to be added to the metadata of the uploaded Glance ClusterOSImage.
+	// Default: the default is to not set any properties.
+	// +optional
+	ClusterOSImageProperties map[string]string `json:"clusterOSImageProperties,omitempty"`
```
staebler, line 68: The description should include all of the properties and their values that we accept. At which point, we may as well specify them as separate fields rather than in a map.

staebler (commented):
I am in agreement with Abhinav on this that it does not seem appropriate to pass any arbitrary image property. If the user wants to use the installer to upload the image, then they get the limited access to the limited properties that we want to support because there are good uses cases for them that cover many users.

pkg/tfvars/openstack/rhcos_image.go
```diff
@@ -15,10 +15,36 @@ import (
...
+		}
+	}
+}
+
 // uploadBaseImage creates a new image in Glance and uploads the RHCOS image there
-func uploadBaseImage(cloud string, localFilePath string, imageName string, clusterID string) error {
+func uploadBaseImage(cloud string, localFilePath string, imageName string, clusterID string, imageProperties map[string]string) error {
 	logrus.Debugln("Creating a Glance image for RHCOS...")
 
+	validateImageProperties(imageProperties)
```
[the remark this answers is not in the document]
mandre: We didn't want to stop the installation if the user passed an untested property but instead print a warning. With that in mind, it seemed fine to validate right before the image upload.

I've now changed it to the install-config validation as requested.

pkg/tfvars/openstack/rhcos_image.go
```diff
@@ -15,10 +15,36 @@ import (
...
+		if validValues, ok := validProperties[property]; ok {
+			for _, allowedValue := range validValues {
+				if value == allowedValue {
+					found = true
+					break
+				}
+			}
+		}
+		if !found {
+			logrus.Warningln("'"+property+"="+value+"'", "is not a supported image property")
```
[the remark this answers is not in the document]
mandre: Our rational was that it is already possible to pass an existing glance image with any properties, and this PR was about allowing this feature in the "preferred" installation flow where the installer manages the image.

Also we didn't want to prevent the user from setting any [valid image properties](https://docs.openstack.org/glance/latest/admin/useful-image-properties.html), because they may be appropriate in their environment. Our message was "use this property at your own risks".

I've changed the validation as requested and we now recommend to use `clusterOSImage` pointing to an existing image in Glance if they want more control over the properties.

pkg/tfvars/openstack/rhcos_image.go
```diff
@@ -15,10 +15,36 @@ import (
...
+		found := false
+		if validValues, ok := validProperties[property]; ok {
+			for _, allowedValue := range validValues {
+				if value == allowedValue {
+					found = true
+					break
+				}
+			}
+		}
+		if !found {
```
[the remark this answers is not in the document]
mandre: Good suggestion. This is now easier to follow too.

pkg/types/openstack/platform.go
mandre: Done.

staebler: So, now we are in a situation where there are only specific fields allowed in the map. This should not be a map but rather defined fields.
```suggestion
	ClusterOSImageProperties *ClusterOSImageProperties `json:"clusterOSImageProperties,omitempty"`
```
```
type ClusterOSImageProperties struct{
	QEMUGuestAgent QEMUGuestAgent `json:"qemuGuestAgent,omitempty"`
	DiskBus DiskBus `json:"diskBus,omitempty"`
	SCSIModel SCSIModel `json:"scsiModel,omitempty"`
}

// +kubebuilder:validation:Enum="";Enabled;Disabled
type QEMUGuestAgent string

const (
	EnabledQEMUGuestAgent QEMUGuestAgent = "Enabled"
	DisabledQEMUGuestAgent QEMUGuestAgent = "Disabled"
)

...
```

mandre: I'd prefer we just expose glance images properties because this is the commonly accepted way to tune your VMs and these are already documented in openstack. Coming up with our own names is not good for discoverability once we start to support more of these properties.

adduarte: There is presedence for this elsewhere in the installer: 
https://github.com/openshift/installer/blob/master/pkg/types/aws/platform.go#L33

aws use of "userTags"

adduarte: One obstacle of using something like 

type ClusterOSImageProperties struct{
	QEMUGuestAgent QEMUGuestAgent `json:"qemuGuestAgent,omitempty"`
	DiskBus DiskBus `json:"diskBus,omitempty"`
	SCSIModel SCSIModel `json:"scsiModel,omitempty"`
}

Is that using enable/disable boolean to implement the UI, is that it becomes exponetially more dificult, becuse DiskBus, and SCSIModel are not booleans. They are not a disable/enable. They are fields that take arbritray values which are also mutually exclusive
so in reality it becomes like this: 

type ClusterOSImageProperties struct{
	QEMUGuestAgent QEMUGuestAgent `json:"qemuGuestAgent,omitempty"`
	DiskBus DiskBusSCSI `json:"diskBusSCSCI,omitempty"`
        DiskBus DiskBusVirtIO `json:"diskBusVirtIO,omitempty"`

	SCSIModel SCSIModel `json:"scsiModel,omitempty"`
}

And DiskBUsSCSI is now exclusive of DiskBUSVirtIO, (we have to code logic to make this so)
and they dependon the alue of SCSIMOdel, 
and it becomes much more complex to code and use and more likely to introduce bugs and user errors.

And I still don't see the actual value of obfuscating the options from the user.

adduarte: Perhaps the correct approach is to modify the validation. 
The real desire is to allow the user to provide any values they want, 
and as a precaution we inform the user if the field they provided is not in the list of supported (by openshift) options. 
But we do not really want to prevent the user, but our current validatio framework does not allow for warnings, only for errors.

pkg/types/openstack/platform.go
```diff
@@ -62,6 +62,15 @@ type Platform struct {
 	// +optional
 	ClusterOSImage string `json:"clusterOSImage,omitempty"`
 
+	// ClusterOSImageProperties is a list of properties to be added to the metadata of the uploaded Glance ClusterOSImage.
+	// Valid properties are:
+	// 	- hw_qemu_guest_agent: yes/no
+	// 	- hw_disk_bus: scsi/virtio
+	// 	- hw_scsi_model: virtio-scsi
```
staebler, line 69: Everything is valid now.
```suggestion
```

docs/user/openstack/customization.md
```diff
@@ -30,6 +30,7 @@ Beyond the [platform-agnostic `install-config.yaml` properties](../customization
 * `region` (deprecated string): The OpenStack region where the cluster will be created. Currently this value is not used by the installer.
 * `trunkSupport` (deprecated string): Whether OpenStack ports can be trunked (`1` for true or `0` for false)
 * `clusterOSImage` (optional string): Either a URL with `http(s)` or `file` scheme to override the default OS image for cluster nodes or an existing Glance image name.
+* `clusterOSImageProperties` (optional list of strings): a list of properties to be added to the installer-uploaded ClusterOSImage in Glance. Valid properties are `hw_qemu_guest_agent: yes/no`, `hw_disk_bus: scsi/virtio`, and `hw_scsi_model: virtio-scsi`. The default is to not set any properties. For more control over the properties, use `clusterOSImage` pointing to an existing image in Glance. `clusterOSImageProperties` is ignored when `clusterOSImage` points to an existing image in Glance.
```
staebler, line 33:
```suggestion
* `clusterOSImageProperties` (optional list of strings): a list of properties to be added to the installer-uploaded ClusterOSImage in Glance. The default is to not set any properties. For more control over the properties, use `clusterOSImage` pointing to an existing image in Glance. `clusterOSImageProperties` is ignored when `clusterOSImage` points to an existing image in Glance.
```

mandre:
@staebler looks like I pushed at the same time you commented :smile: 
I agree with your comment and will remove references to supported values.

staebler (commented):
/approve

PR #4403, opened 2020-11-23
Mgmt 2794 sno bootstrap inplace

eranco74:
This is the code I got while exploring options for [installing Single Node Openshift (SNO) without an auxiliary bootstrap node](https://issues.redhat.com/browse/MGMT-2794).

In essence, you can use the bootstrap ignition generated with this installer to spinup SNO.
It will setup bootstrap on the node you are installing and then pivot the node to become a master node.

How to and more details are [here](https://docs.google.com/document/d/17ANid5OmjPd2MWhhPxCv-Lw3_3jvfC6S5GBK5Y-CxxM/edit?usp=sharing)

### DIFF
--- a/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
+++ b/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
@@ -314,7 +314,7 @@ fi
 
 # Wait for the etcd cluster to come up.
 until bootkube_podman_run \
-		--rm \
+        --rm \
 		--name etcdctl \
 		--env ETCDCTL_API=3 \
 		--volume /opt/openshift/tls:/opt/openshift/tls:ro,z \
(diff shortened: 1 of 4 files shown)


eranco74:
Initial POC for installing Single node OpenShift without auxiliary bootstrap node.
Please take a look @staebler @eparis @walters @deads2k @marun @hexfusion @crawford.

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -265,6 +266,7 @@ func (a *Bootstrap) getTemplateData(installConfig *types.InstallConfig, releaseI
 		Registries:            registries,
 		BootImage:             string(*rhcosImage),
 		PlatformData:          platformData,
+		SingleNode:            *installConfig.ControlPlane.Replicas == 1,
```
hexfusion, line 269: So regardless of single-node deployment type all of the changes gated by `SingleNode` are true? CRC, libvirt, etc.

data/data/bootstrap/files/usr/local/bin/patch.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash -x
+
+function patchit {
```
hexfusion, line 3: Could naming be more explicit to usage? Maybe patch_etcd-quorum-guard

data/data/bootstrap/files/usr/local/bin/patch.sh
```diff
@@ -0,0 +1,36 @@
...
+)" --type=merge || return 1
+
+    # scale down etcd-quorum-guard
+    oc scale --kubeconfig ./auth/kubeconfig-loopback --replicas=1 deployment/etcd-quorum-guard -n openshift-etcd || return 1
+
+    return 0
+}
+
+while ! patchit; do
+    echo "Waiting to try again..."
```
hexfusion, line 32: for debug, it would be helpful if the message had more details

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -325,14 +325,126 @@ until bootkube_podman_run \
...
+    # Rename the manifests files
+    ls bootstrap-manifests/ | xargs -I {} mv bootstrap-manifests/{} bootstrap-manifests/bootstrap-{}
+
+    # Rename the temporary control plane containers
+    sed -i 's/name: cluster-version-operator/name: bootstrap-cluster-version-operator/g' bootstrap-manifests/bootstrap-bootstrap-pod.yaml
+    sed -i 's/name: cloud-credential-operator/name: bootstrap-cloud-credential-operator/g' bootstrap-manifests/bootstrap-cloud-credential-operator-pod.yaml
+    sed -i 's/name: kube-apiserver/name: bootstrap-kube-apiserver/g' bootstrap-manifests/bootstrap-kube-apiserver-pod.yaml
+    sed -i 's/name: kube-controller-manager/name: bootstrap-kube-controller-manager/g' bootstrap-manifests/bootstrap-kube-controller-manager-pod.yaml
+    sed -i 's/name: kube-controller-manager/name: bootstrap-kube-controller-manager/g' bootstrap-manifests/bootstrap-kube-controller-manager-pod.yaml
+    sed -i 's/name: kube-scheduler/name: bootstrap-kube-scheduler/g' bootstrap-manifests/bootstrap-kube-scheduler-pod.yaml
```
sttts, line 383: I don't think we want these hacks.

/hold

data/data/bootstrap/files/usr/local/bin/patch.sh
hexfusion: if we omit quorum-guard with annotation will this function fail? I believe that was the intent behind "Add single-node-developer Cluster Profile" enhancement.[1]

[1]https://github.com/openshift/enhancements/pull/482

sttts:
I assume this PR does this patching of rendered resources only as a proof-of-concept. This is not acceptable in production code we have to maintain. These changes must go into the render commands of the respective operators, e.g. via some `--single-node=true` flags in order to have visibility for the owners of these components.

data/data/bootstrap/files/usr/local/bin/patch.sh
eranco74: Yes, this patch is just for allowing a single node installation and isn't related to the in-place-bootstrapping.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
eranco74: of course. this is just a POC.
I think it will be great if the control plane operators will render the actual control plane for the cluster and make this operation and the static pods swap obsolete.

eranco74:
> I assume this PR does this patching of rendered resources only as a proof-of-concept. This is not acceptable in production code we have to maintain. These changes must go into the render commands of the respective operators, e.g. via some `--single-node=true` flags in order to have visibility for the owners of these components.

I agree, render seems to be the place to make the changes.
I published it as draft with `/hold` just to show the POC.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
sttts: 👍

pkg/asset/ignition/bootstrap/bootstrap.go
eranco74: Unsure I understand the question.
In case SingleNode == true, bootkube.service.template will render a different bootkube.service.
This bootkube service will be packed into the bootstrap ignition.
When the user ignite an RHCoS image with this the bootkube service will execute the bootstrap logic and will "pivot" the node to become a master node and cleanup the bootstrap garbage, so we end up with a Single Node Openshift.

hexfusion: I just mean this all works with other single node usages such as CRC right? The base condition for the logic being used is master replica 1. I was just pointing out it could affect others.

eranco74:
Closing this one. I'll create a new PR that runs the bootstrap in live-ISO.

PR #4404, opened 2020-11-23
Bug 1900138: Removed support for insecure mode for oVirt/RHV installation

janoszen:
[This change](https://bugzilla.redhat.com/show_bug.cgi?id=1900138) removes support for the insecure mode from the oVirt installer. Previously, when no certificate could be obtained from the oVirt engine the installer would proceed without certificate verification. This PR is based on #4387 and **depends on [OCPRHV-426](https://issues.redhat.com/browse/OCPRHV-426)**.

Due to recent improvements this is no longer a valid use case and is being deprecated. The user is instead presented with a message explaining the situation and linking to the to-be-written documentation. If the user wants to use insecure mode they have to create a file named `~/.ovirt/ovirt-config.yaml` with the following contents before running the installer:

```yaml
ovirt_url: https://ovirt.example.com/ovirt-engine/api
ovirt_fqdn: ovirt.example.com
ovirt_pem_url: ""
ovirt_username: admin@internal
ovirt_password: super-secret-password
ovirt_insecure: true
```

## User experience

```
$ bin/openshift-install create install-config
? Platform ovirt
? Engine FQDN[:PORT] vm-10-208.lab.eng.tlv2.redhat.com
INFO    Version: 3
INFO    Serial Number: 4096
INFO    Issuer: CN=ovirt.example.com.39197,O=example.com,C=US
INFO    Validity:
INFO            Not Before: 2020-08-11 13:50:26 +0000 UTC
INFO            Not After: 2030-08-10 13:50:26 +0000 UTC
INFO    Subject: CN=ovirt.example.com.39197,O=example.com,C=US
? Would you like to use the above certificate to connect to Engine?  No
? Would you like to import another PEM bundle? No
ERROR ****************************************************************************
ERROR * Could not configure secure communication to the oVirt engine.            *
ERROR * As of 4.7 insecure mode for oVirt is no longer supported from the        *
ERROR * installer. Please see ____LINK____ for details how to configure insecure *
ERROR * mode manually.                                                           *
ERROR ****************************************************************************
ERROR oVirt configuration failed: cannot detect Engine CA cert imported in the system.
? Engine FQDN[:PORT]
```

### DIFF
--- a/docs/user/ovirt/install_ipi.md
+++ b/docs/user/ovirt/install_ipi.md
@@ -226,3 +226,24 @@ $ oc get nodes
 ```
 
 [Bare Metal IPI Networking Infrastructure]: https://github.com/openshift/installer/blob/master/docs/design/baremetal/networking-infrastructure.md
+
+#### Installing OpenShift on RHV/oVirt in *insecure* mode
+
+<!-- Do not change this title as it is used in the code to point users to the right place -->
+
+Starting OpenShift 4.7 we are sunsetting the “insecure” option from the OpenShift Installer. Starting with this version, the installer only supports installation methods from the user interface that lead to using verified certificates.
+
+This change also means that setting up the CA certificate for RHV is no longer required before running the installer. The installer will ask you for confirmation about the certificate and store the CA certificate for use during the installation.
+
+Should you, nevertheless, require an installation without certificate verification you can create a file named ovirt-config.yaml in the .ovirt directory in your home directory (~/.ovirt/ovirt-config.yaml) before running the installer with the following content:
+
+```yaml
+ovirt_url: https://ovirt.example.com/ovirt-engine/api
+ovirt_fqdn: ovirt.example.com
+ovirt_pem_url: ""
+ovirt_username: admin@internal
+ovirt_password: super-secret-password
+ovirt_insecure: true
+```
+
+Please note that this option is **not recommended** as it will allow a potential attacker to perform a Man-in-the-Middle attack and capture sensitive credentials on the network.
--- a/pkg/asset/installconfig/ovirt/credentials.go
+++ b/pkg/asset/installconfig/ovirt/credentials.go
@@ -366,9 +366,19 @@ func engineSetup() (Config, error) {
 	}
 
 	if engineConfig.Insecure {
-		logrus.Warning(
-			"cannot detect Engine CA cert imported in the system. ",
-			"Communication with the Engine will be insecure.")
+		logrus.Error(
+			"****************************************************************************\n",
+			"* Could not configure secure communication to the oVirt engine.            *\n",
+			"* As of 4.7 insecure mode for oVirt is no longer supported in the          *\n",
+			"* installer. Please see the help article titled \"Installing OpenShift on   *\n",
+			"* RHV/oVirt in insecure mode\" for details how to configure insecure mode   *\n",
+			"* manually.                                                                *\n",
+			"****************************************************************************",
+		)
+		return engineConfig,
+			errors.New(
+				"cannot detect engine ca cert imported in the system",
+			)
 	}
 	return askCredentials(engineConfig)
 }


janoszen:
@Gal-Zaidman this one's next. We may need QA to look at this before we merge.

Gal-Zaidman:
> We may need QA to look at this before we merge and we still need text for it.

Can you also look at the commits I think something funky happened there

janoszen:
@Gal-Zaidman thanks, resolved the issue with the commits. QA got back to us and it looks good to them too, now we are just waiting for docs.

eslutsky (commented):
we need to make sure that LINK is persistent as it will be hardcoded in the code

janoszen:
@eslutsky the docs bug is here: https://issues.redhat.com/browse/OCPRHV-426

eslutsky:
> @eslutsky the docs bug is here: https://issues.redhat.com/browse/OCPRHV-426

thanks , replied in the bug

janoszen:
Link issue has been resolved in the latest update.

dougsland:
Can we start review this one? I see WIP in the title.

dougsland:
The patch looks good but  I would exit the loop in case user reach the level of insecure as it's a critical thing and they would need to fix their env.

janoszen:
@dougsland not necessarily, they might just land in that branch by mistake. Very few people will actually want to use it without any verification, I can't even think of a single reason why you would want that.

dougsland:
> @dougsland not necessarily, they might just land in that branch by mistake. 
> Very few people will actually want to use it without any verification,
> I can't even think of a single reason why you would want that.

If you believe it's a dead code/scenario, probably consider removing the entire condition.
However, I believe we cannot guarantee the scenario for every single user out there.

```
ERROR **************************************************************************** 
ERROR * Could not configure secure communication to the oVirt engine.            * 
ERROR * As of 4.7 insecure mode for oVirt is no longer supported in the          * 
ERROR * installer. Please see the help article titled "Installing OpenShift on   * 
ERROR * RHV/oVirt in insecure mode" for details how to configure insecure mode   * 
ERROR * manually.                                                                * 
ERROR **************************************************************************** 
ERROR oVirt configuration failed: cannot detect engine ca cert imported in the system 
```

From my understanding: 
This is a case where IPI installer detected an unsupported scenario and provided an error message to users. 
On top of that, it's nothing something to solve in a blink of eyes (replaces/add certs) in the server with real users
 depending on it (as might require maint window). 

If you believe 100% users will achieve this code part as a mistake flow and want to continue the installer alive I would replace error with info message type.

peterclauterbach:
> From my understanding:
> This is a case where IPI installer detected an unsupported scenario and provided an error message to users.
> On top of that, it's nothing something to solve in a blink of eyes (replaces/add certs) in the server with real users
> depending on it (as might require maint window).
> 
> If you believe 100% users will achieve this code part as a mistake flow and want to continue the installer alive I would replace error with info message type.

I think there is a case where you've made a genuine mistake in the certificates (like point to an expired one) , we should give you a chance to fix it, instead of failing immediately. The install will loop on this 3 times, and then fail.  This seem like a good compromise of being forgiving, but not letting the user do something dangerous.

dougsland:
> > From my understanding:
> > This is a case where IPI installer detected an unsupported scenario and provided an error message to users.
> > On top of that, it's nothing something to solve in a blink of eyes (replaces/add certs) in the server with real users
> > depending on it (as might require maint window).
> > If you believe 100% users will achieve this code part as a mistake flow and want to continue the installer alive I would replace error with info message type.
> 
> I think there is a case where you've made a genuine mistake in the certificates (like point to an expired one) , we should give
> you a chance to fix it, instead of failing immediately. The install will loop on this 3 times, and then fail. This seem like a good
> compromise of being forgiving, but not letting the user do something dangerous.

Moving forward as PM requested.

PR #4405, opened 2020-11-23
Bug 1786314: OpenStack: support self-signed certificates in the installer

Fedosin:
Recently TLS support was implemented in gophercloud/utils[1], so, by updating the library we allow the installer to work with self-signed certificates without any code modifications.

[1] https://github.com/gophercloud/utils/issues/131

/label platform/openstack

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	github.com/google/martian v2.1.1-0.20190517191504-25dcb96d9e51+incompatible // indirect
 	github.com/google/uuid v1.1.1
 	github.com/gophercloud/gophercloud v0.12.1-0.20200821143728-362eb785d617
-	github.com/gophercloud/utils v0.0.0-20200918191848-da0e919a012a
+	github.com/gophercloud/utils v0.0.0-20201203161420-f41c1768a042
 	github.com/h2non/filetype v1.0.12
 	github.com/hashicorp/go-azure-helpers v0.10.0
 	github.com/hashicorp/go-plugin v1.2.2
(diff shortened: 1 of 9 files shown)


Fedosin:
`curl: (6) Could not resolve host: github.com` :(

pkg/utils/openstack/OWNERS
```diff
@@ -0,0 +1,7 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
```
abhinavdahiya, line 1: Please no utils package. This package has a very high chance of becoming collections of various different types of packages and causes long term confusion. Please try and use something that is more specific that doesn't have `utils` as directory in path.

pkg/auth/openstack/serviceclient.go
```diff
@@ -0,0 +1,91 @@
...
+	"io/ioutil"
+	"net/http"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/utils/env"
+	"github.com/gophercloud/utils/openstack/clientconfig"
+	"github.com/pkg/errors"
+)
+
+// NewServiceClient is a wrapper function around gophercloud/utils that enables support of self-sighed
```
staebler, line 16:
```suggestion
// NewServiceClient is a wrapper function around gophercloud/utils that enables support of self-signed
```

mandre:
/hold
We're exploring another way of doing this by adding the support for cacert directly in gophercloud.
https://github.com/gophercloud/utils/pull/131

staebler (commented):
/approve

EmilienM:
I think we need to have https://github.com/openshift/installer/pull/4457 merged first, then rebase this one and it'll be good.

EmilienM:
https://github.com/openshift/installer/pull/4457 has landed, please rebase on top of master and this should be ready to go.

Fedosin:
This change is included in https://github.com/openshift/installer/pull/4457

PR #4406, opened 2020-11-23
[WIP] docs/dev/bootstrap: specify logging dirs

patrickdillon:
Specifies directories where static pods should write logs.

@abhinavdahiya is this what you had in mind for https://issues.redhat.com/browse/CORS-1544?


abhinavdahiya:
this looks like what i had in mind. Can we also make sure we are capturing logs from both the locations @patrickdillon ?

mtnbikenc (commented):
Is this still needed?

PR #4407, opened 2020-11-23
vSphere Destroy: add log messages for tag destruction

patrickdillon:
Prints Info message when tag and tag categories are successfully destroyed.

/test e2e-vsphere

### DIFF
--- a/pkg/destroy/vsphere/vsphere.go
+++ b/pkg/destroy/vsphere/vsphere.go
@@ -213,16 +213,20 @@ func (o *ClusterUninstaller) Run() error {
 	}
 
 	o.Logger.Debug("Delete tag")
+	tagLogger := o.Logger.WithField("Tag", o.InfraID)
 	if err = deleteTag(context.TODO(), o.RestClient, o.InfraID); err != nil {
-		o.Logger.Errorln(err)
+		tagLogger.Errorln(err)
 		return err
 	}
+	tagLogger.Info("Destroyed")
 
 	o.Logger.Debug("Delete tag category")
+	tcLogger := o.Logger.WithField("TagCategory", "openshift-"+o.InfraID)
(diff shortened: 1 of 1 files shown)


pkg/destroy/vsphere/vsphere.go
```diff
@@ -217,12 +217,14 @@ func (o *ClusterUninstaller) Run() error {
...
 		return err
 	}
+	o.Logger.WithField("Tag", o.InfraID).Info("Destroyed")
 
 	o.Logger.Debug("Delete tag category")
 	if err = deleteTagCategory(context.TODO(), o.RestClient, "openshift-"+o.InfraID); err != nil {
 		o.Logger.Errorln(err)
 		return err
 	}
+	o.Logger.WithField("Tag Category", "openshift-"+o.InfraID).Info("Destroyed")
```
staebler, line 227: It is a bit easier to read log entries when the log fields do not have spaces.
```suggestion
	o.Logger.WithField("TagCategory", "openshift-"+o.InfraID).Info("Destroyed")
```

pkg/destroy/vsphere/vsphere.go
```diff
@@ -217,12 +217,14 @@ func (o *ClusterUninstaller) Run() error {
 		o.Logger.Errorln(err)
 		return err
 	}
+	o.Logger.WithField("Tag", o.InfraID).Info("Destroyed")
```
staebler, line 220: For consistency, if we are going to add the field to the "Destroyed" log event, we should add it to the other log events for the action. Specifically, use the logger returned by the call to `WithField` for the "Delete tag" log event and the error log event as well.

patrickdillon: Ok. I think the new changes are in line with this suggestion.

staebler (commented):
/lgtm
/approve

PR #4410, opened 2020-11-24
OpenStack: add cinder client to the ci image

Fedosin:
To support CSI tests we need OpenStack Cinder client to be installed[1] in the CI image.

[1] https://github.com/openshift/release/pull/13317#issuecomment-732866652


mandre (changes requested):
`python3-cinderclient` is already pulled as a dependency of `python3-openstackclient`, see for example https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_installer/4405/pull-ci-openshift-installer-master-e2e-openstack/1330934412740661248/artifacts/build-logs/openstack-installer.log.

Your problem is likely that you're not using the openstack image for your CI job.

PR #4411, opened 2020-11-24
Bug 1899853: openstack: CP nodes port to use addtional SGs

pierreprinetti:
Assign the user-defined additional security groups to the Control Plane
port.

Fixes [Bug 1899853](https://bugzilla.redhat.com/show_bug.cgi?id=1899853)

### DIFF
--- a/data/data/openstack/main.tf
+++ b/data/data/openstack/main.tf
@@ -81,6 +81,7 @@ module "topology" {
   octavia_support     = var.openstack_octavia_support
   machines_subnet_id  = var.openstack_machines_subnet_id
   machines_network_id = var.openstack_machines_network_id
+  master_extra_sg_ids = var.openstack_master_extra_sg_ids
 }
 
 data "openstack_images_image_v2" "base_image" {
--- a/data/data/openstack/topology/private-network.tf
+++ b/data/data/openstack/topology/private-network.tf
@@ -41,10 +41,13 @@ resource "openstack_networking_port_v2" "masters" {
   name  = "${var.cluster_id}-master-port-${count.index}"
   count = var.masters_count
 
-  admin_state_up     = "true"
-  network_id         = local.nodes_network_id
-  security_group_ids = [openstack_networking_secgroup_v2.master.id]
-  tags               = ["openshiftClusterID=${var.cluster_id}"]
+  admin_state_up = "true"
+  network_id     = local.nodes_network_id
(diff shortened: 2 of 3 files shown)


Fedosin (changes requested):
Your patch leads to sg duplication here https://github.com/openshift/installer/blob/master/data/data/openstack/main.tf#L55-L58 because sgs from `topology` already contain `openstack_master_extra_sg_ids`

pierreprinetti:
As far as I can tell, `module.topology.master_sg_id` comes from the [output](https://github.com/openshift/installer/blob/5ce7e599e68ece532f2d52ec2b37edf75b2b5b6f/data/data/openstack/topology/outputs.tf#L2) of the [stock SG creation](https://github.com/openshift/installer/blob/5ce7e599e68ece532f2d52ec2b37edf75b2b5b6f/data/data/openstack/topology/sg-master.tf#L1-L4) only.

So if I'm correct, there's no duplication. @Fedosin can you please check and possibly disprove me please?

mandre:
It looks like we also need to add extra security groups to `api_port` and `ingress_port`, isn't it?

mandre:
> As far as I can tell, `module.topology.master_sg_id` comes from the [output](https://github.com/openshift/installer/blob/5ce7e599e68ece532f2d52ec2b37edf75b2b5b6f/data/data/openstack/topology/outputs.tf#L2) of the [stock SG creation](https://github.com/openshift/installer/blob/5ce7e599e68ece532f2d52ec2b37edf75b2b5b6f/data/data/openstack/topology/sg-master.tf#L1-L4) only.
> 
> So if I'm correct, there's no duplication... however my Terraform is still a bit uncertain!
> 
> @Fedosin can you please check and possibly disprove me please?

[master_sg_ids](https://github.com/openshift/installer/blob/master/data/data/openstack/main.tf#L55-L58) is only used in the `master` module creating the instances, and not in the `topology` one creating the ports on the subnets.
We should pass the same `master_sg_ids` variable in the `topology` module, and I think also a `worker_sg_ids` one for the `ingress_port`.

mandre (approved):
/approve
/lgtm

The patch fixes exactly the issue that is reported in the BZ so I do not see any reason not to approve. It would be nice to know if we also need to apply the additional security groups to the VIP ports. Otherwise what is the point of adding security groups at all to `api_port` and `ingress_port`?

PR #4308, opened 2020-10-26
Single node v2

eranco74:
/hold

PR #4412, opened 2020-11-24
Igal/single node v2 on eran

tsorya:
/hold

### DIFF
--- a/pkg/asset/tls/tls.go
+++ b/pkg/asset/tls/tls.go
@@ -24,6 +24,9 @@ const (
 	// ValidityOneDay sets the validity of a cert to 24 hours.
 	ValidityOneDay = time.Hour * 24
 
+	// ValidityOneMinute sets the validity of a cert to 1 month.
+	ValidityOneMonth = ValidityOneDay * 30
+
 	// ValidityOneYear sets the validity of a cert to 1 year.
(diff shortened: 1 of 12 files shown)


cfergeau:
What is this about? Does this need early reviews?

gbraad:
Hi, can you explain what this PR tries to solve and what this aligns with? There is an effort to create a Single Node Cluster at https://github.com/code-ready/snc/ which is used by CRC. Also, for resource tuning, effort is being made with the Single Node Profile enhancement, and is work on by the same team. Would appreciate some back story or more info

pkg/asset/tls/tls.go
```diff
@@ -24,6 +24,9 @@ const (
 	// ValidityOneDay sets the validity of a cert to 24 hours.
 	ValidityOneDay = time.Hour * 24
 
+	// ValidityOneMinute sets the validity of a cert to 1 month.
+	ValidityOneMonth = ValidityOneDay * 30
```
gbraad, line 28: Same comment as here: https://github.com/openshift/installer/pull/4308#discussion_r530360087

This is a no-go situation. Initial certificates are deemed to be only 24hrs, after which they rotate to the expected 30 days validity.

staebler:
This seems to be obsolete. Feel free to re-open if I am mistaken.
/close

PR #4419, opened 2020-11-25
asset/installconfig/Azure: Validate install-config instance types

jstuever:
Ensure the user provided instance types meets minimum resource
requirements.

### DIFF
--- a/pkg/asset/installconfig/azure/client.go
+++ b/pkg/asset/installconfig/azure/client.go
@@ -22,6 +22,7 @@ type API interface {
 	GetControlPlaneSubnet(ctx context.Context, resourceGroupName, virtualNetwork, subnet string) (*aznetwork.Subnet, error)
 	ListLocations(ctx context.Context) (*[]azsubs.Location, error)
 	GetResourcesProvider(ctx context.Context, resourceProviderNamespace string) (*azres.Provider, error)
+	GetVirtualMachineSku(ctx context.Context, name, region string) (*azsku.ResourceSku, error)
 	GetDiskSkus(ctx context.Context, region string) ([]azsku.ResourceSku, error)
 	GetGroup(ctx context.Context, groupName string) (*azres.Group, error)
 	ListResourceIDsByGroup(ctx context.Context, groupName string) ([]string, error)
@@ -210,4 +211,35 @@ func (c *Client) ListResourceIDsByGroup(ctx context.Context, groupName string) (
 		}
 	}
 	return res, nil
+}
+
+// GetVirtualMachineSku retrieves the resource SKU of a specified virtual machine SKU in the specified region.
+func (c *Client) GetVirtualMachineSku(ctx context.Context, name, region string) (*azsku.ResourceSku, error) {
+	client := azsku.NewResourceSkusClientWithBaseURI(c.ssn.Environment.ResourceManagerEndpoint, c.ssn.Credentials.SubscriptionID)
+	client.Authorizer = c.ssn.Authorizer
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	for page, err := client.List(ctx); page.NotDone(); err = page.NextWithContext(ctx) {
+		if err != nil {
+			return nil, errors.Wrap(err, "error fetching SKU pages")
+		}
+		for _, sku := range page.Values() {
+			// Filter out resources that are not virtualMachines
+			if !strings.EqualFold("virtualMachines", *sku.ResourceType) {
+				continue
+			}
+			// Filter out resources that do not match the provided name
+			if !strings.EqualFold(name, *sku.Name) {
+				continue
+			}
+			// Return the resource from the provided region
+			for _, location := range to.StringSlice(sku.Locations) {
+				if strings.EqualFold(location, region) {
+					return &sku, nil
+				}
+			}
+		}
+	}
+	return nil, nil
 }
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net"
 	"sort"
+	"strconv"
 	"strings"
 
 	azdns "github.com/Azure/azure-sdk-for-go/profiles/latest/dns/mgmt/dns"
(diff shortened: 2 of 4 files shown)


jstuever:
https://issues.redhat.com/browse/CORS-1549

pkg/asset/installconfig/azure/client.go
```diff
@@ -211,3 +212,60 @@ func (c *Client) ListResourceIDsByGroup(ctx context.Context, groupName string) (
...
+				if strings.EqualFold(location, region) {
+					results = append(results, sku)
+				}
+			}
+		}
+	}
+
+	if len(results) == 0 {
+		return nil, errors.Errorf("no SKUs for specified subscription in region %s", region)
+	}
```
staebler, line 267: Why return an error rather than just returning an empty slice and letting the caller decide how to handle there being no SKUs?

pkg/asset/installconfig/azure/client.go
```diff
@@ -211,3 +212,60 @@ func (c *Client) ListResourceIDsByGroup(ctx context.Context, groupName string) (
 	}
 	return res, nil
 }
+
+// GetVirtualMachineSku retrieves the restource SKU of a specified virtual machine in the given region.
+func (c *Client) GetVirtualMachineSku(ctx context.Context, name, region string) (azsku.ResourceSku, error) {
+	results, err := c.GetResourceSkus(ctx, "virtualMachines", region)
+	if err != nil {
+		return azsku.ResourceSku{}, err
```
staebler, line 220: Can we change the return type to `*azsku.ResoureSku` so that we can return `nil` instead of an empty struct?

pkg/asset/installconfig/azure/client.go
```diff
@@ -211,3 +212,60 @@ func (c *Client) ListResourceIDsByGroup(ctx context.Context, groupName string) (
...
+			return sku, nil
+		}
+	}
+
+	return azsku.ResourceSku{}, errors.Errorf("SKU not found for specified subscription in region %s", region)
+
+}
+
+// GetResourceSkus retrieves the resource SKUs of a specified type for a given region.
+func (c *Client) GetResourceSkus(ctx context.Context, resourceType, region string) ([]azsku.ResourceSku, error) {
```
staebler, line 234: I don't think that we have any need for this function outside of getting a specific SKU, so let's forego returning **all** of the SKUs in the region. Returning all of the SKUs in a slice exposes us to crashes in the (unlikely) event that the server responds with more SKUs than we can handle.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -17,15 +18,95 @@ import (
...
+	allErrs := field.ErrorList{}
+
+	typeMeta, err := client.GetVirtualMachineSku(context.TODO(), instanceType, region)
+	if err != nil {
+		return append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, err.Error()))
+	}
+
+	for _, capability := range *typeMeta.Capabilities {
+
+		if strings.EqualFold(*capability.Name, "vCPUs") {
```
staebler, line 57: Are we willing to accept a SKU that does not have the "vCPUs" or "MemoryGB" capability?

pkg/asset/installconfig/azure/client.go
```diff
@@ -211,3 +212,60 @@ func (c *Client) ListResourceIDsByGroup(ctx context.Context, groupName string) (
...
+	client := azsku.NewResourceSkusClientWithBaseURI(c.ssn.Environment.ResourceManagerEndpoint, c.ssn.Credentials.SubscriptionID)
+	client.Authorizer = c.ssn.Authorizer
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	var results []azsku.ResourceSku
+
+	for page, err := client.List(ctx); page.NotDone(); err = page.NextWithContext(ctx) {
```
staebler, line 242: It is unfortunate that the Azure client does not easily support adding the location filter. In the subscription that I am using, there are 7757 virtual machine SKUs that all need to be fetched to validate the instance type.

jstuever: I totally agree.

pkg/asset/installconfig/azure/validation.go
jstuever: Currently, if the capability is not present, then we won't show any error. We filter out any SKU that is not of type "virtualMachines". I am unable to find anything in the docs where this type of SKU is guaranteed to have these capabilities specified (or not).

pkg/asset/installconfig/azure/client.go
```diff
@@ -211,3 +212,60 @@ func (c *Client) ListResourceIDsByGroup(ctx context.Context, groupName string) (
 	}
 	return res, nil
 }
+
+// GetVirtualMachineSku retrieves the restource SKU of a specified virtual machine in the given region.
```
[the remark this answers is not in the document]
jstuever: Fixed in https://github.com/openshift/installer/commit/91402a22c47edd67c4b9802c4f2e47bfce3b3f39

pkg/asset/installconfig/azure/client.go
jstuever: Fixed in https://github.com/openshift/installer/commit/91402a22c47edd67c4b9802c4f2e47bfce3b3f39

pkg/asset/installconfig/azure/client.go
jstuever: Fixed in https://github.com/openshift/installer/commit/91402a22c47edd67c4b9802c4f2e47bfce3b3f39

pkg/asset/installconfig/azure/client.go
jstuever: Fixed in https://github.com/openshift/installer/commit/91402a22c47edd67c4b9802c4f2e47bfce3b3f39

pkg/asset/installconfig/azure/validation.go
jstuever: I think we should move forward as-is. We can always make changes if we find cu issues later on.

mjudeikis:
overall looks good. Is this in any is aligned with machine-api to do the same when creating new sets? we seen that in most cases user workaround this by creating a cluster and changing compute types post creation :/

staebler (approved):
/lgtm
/approve

PR #4421, opened 2020-11-26
hack/verify-vendor.sh: Fail early if error

pierreprinetti:
Prior to this patch, errors in vendoring or verification were ignored.

With this patch, `verify-vendor` fails if the dependency files are
corrupted.

### DIFF
--- a/hack/verify-vendor.sh
+++ b/hack/verify-vendor.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 if [ "$IS_CONTAINER" != "" ]; then
+  set -euxo pipefail
   go mod vendor
   go mod verify
   git diff --exit-code


hack/verify-vendor.sh
```diff
@@ -3,6 +3,7 @@
 if [ "$IS_CONTAINER" != "" ]; then
   go mod vendor
   go mod verify
+  go mod tidy
```
bogdando, line 6: isn't the verify should be the last one?

pierreprinetti: Right; thank you. I have eventually removed the call to `go mod tidy`, focusing on a more pressing need for `set -e` which I aim at merging with minimum controversy (which `tidy` could generate).

I will certainly consider your comment when I try to introduce `tidy` again.

pierreprinetti:
and indeed:

```
+ go mod vendor
warning: ignoring symlink /go/src/github.com/openshift/installer/pkg/asset/store/data
go: github.com/openshift/machine-api-operator@v0.2.1-0.20200429102619-d36974451290 requires
	github.com/openshift/library-go@v0.0.0-20200324092245-db2a8546af81 requires
	bitbucket.org/ww/goautoneg@v0.0.0-20120707110453-75cd24fc2f2c: reading https://api.bitbucket.org/2.0/repositories/ww/goautoneg?fields=scm: 404 Not Found
```

Fedosin (approved):
/lgtm

mtnbikenc (approved):

staebler (commented):
These changes look good. But we cannot merge this until we actually fix the modules issues.

pierreprinetti:
/retest

Should succeed after https://github.com/openshift/installer/commit/20567f74d657046eebabb3c9edccd4c9d6690d83

pierreprinetti:
Indeed! The test is successful. I think we're ready for your review @staebler

PR #4426, opened 2020-11-27
Bug 1813949: ignore local env variables when we create a service client

Fedosin:
This commit explicitly disables reading auth data from env variables by setting an invalid EnvPrefix. By doing this, we make sure that the data from clouds.yaml is enough to authenticate.

After this change we don't have to unset OS_CLOUD env variable explicitly anymore.

### DIFF
--- a/pkg/asset/installconfig/openstack/session.go
+++ b/pkg/asset/installconfig/openstack/session.go
@@ -2,9 +2,9 @@
 package openstack
 
 import (
-	"os"
 	"sync"
 
+	openstackdefaults "github.com/openshift/installer/pkg/types/openstack/defaults"
 	"github.com/pkg/errors"
 
 	"github.com/ghodss/yaml"
@@ -21,12 +21,8 @@ type Session struct {
 
 // GetSession returns an OpenStack session for a given cloud name in clouds.yaml.
 func GetSession(cloudName string) (*Session, error) {
-	opts := defaultClientOpts(cloudName)
-
-	// We should unset OS_CLOUD env variable here, because the real cloud name was
-	// defined on the previous step. OS_CLOUD has more priority, so the value from
-	// "opts" variable will be ignored if OS_CLOUD contains something.
-	os.Unsetenv("OS_CLOUD")
+	opts := openstackdefaults.DefaultClientOpts(cloudName)
+	opts.YAMLOpts = new(yamlLoadOpts)
 
 	cloudConfig, err := clientconfig.GetCloudFromYAML(opts)
 	if err != nil {
@@ -35,13 +31,6 @@ func GetSession(cloudName string) (*Session, error) {
 	return &Session{
 		CloudConfig: cloudConfig,
 	}, nil
-}
-
-func defaultClientOpts(cloudName string) *clientconfig.ClientOpts {
-	opts := new(clientconfig.ClientOpts)
-	opts.Cloud = cloudName
-	opts.YAMLOpts = new(yamlLoadOpts)
-	return opts
 }
 
 type yamlLoadOpts struct{}
--- a/pkg/asset/installconfig/openstack/validation/cloudinfo.go
+++ b/pkg/asset/installconfig/openstack/validation/cloudinfo.go
@@ -2,7 +2,6 @@ package validation
 
 import (
 	"net/url"
-	"os"
 	"strings"
 
 	"github.com/gophercloud/gophercloud"
@@ -24,6 +23,7 @@ import (
 
 	"github.com/openshift/installer/pkg/quota"
 	"github.com/openshift/installer/pkg/types"
+	openstackdefaults "github.com/openshift/installer/pkg/types/openstack/defaults"
 )
 
 // CloudInfo caches data fetched from the user's openstack cloud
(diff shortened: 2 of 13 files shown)


mdbooth:
We need some library code for this. We're already doing the os.Unsetenv("OS_CLOUD") dance in 4 places:

- https://github.com/Fedosin/installer/blob/validate_clouds_yaml/pkg/asset/installconfig/openstack/openstack.go#L49-L52
- https://github.com/Fedosin/installer/blob/validate_clouds_yaml/pkg/asset/installconfig/openstack/session.go#L26-L29
- https://github.com/Fedosin/installer/blob/validate_clouds_yaml/pkg/asset/installconfig/openstack/validation/cloudinfo.go#L65-L68
- https://github.com/Fedosin/installer/blob/validate_clouds_yaml/pkg/asset/installconfig/openstack/validation/cloudinfo.go#L65-L68

To avoid inconsistent behaviour we're going to want the new hack in all those places, too. We should really pull this out somewhere. This would also allow us to unit test it.

pierreprinetti:
Is this a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1813949 ?

pkg/asset/installconfig/openstack/validation/cloudinfo.go
```diff
@@ -57,6 +57,11 @@ func GetCloudInfo(ic *types.InstallConfig) (*CloudInfo, error) {
 
 	opts := &clientconfig.ClientOpts{Cloud: ic.OpenStack.Cloud}
 
+	// We explicitly disable reading auth data from env variables by setting an invalid EnvPrefix.
+	// By doing this, we make sure that the data from clouds.yaml is enough to authenticate.
+	// For more information: https://github.com/gophercloud/utils/blob/8677e053dcf1f05d0fa0a616094aace04690eb94/openstack/clientconfig/requests.go#L508
+	opts.EnvPrefix = "NO_ENV_VARIABLES_"
```
EmilienM, line 63: This is a nice workaround but I think this needs to be applied to other places as well.

pierreprinetti:
Can we apply this to all the steps (install-config generation, manifest creation, ignition... etc), and remove our various calls to `unset OS_CLOUD`?

mandre (commented):
/lgtm

pkg/asset/installconfig/openstack/validvaluesfetcher.go
```diff
@@ -28,9 +28,7 @@ func getCloudNames() ([]string, error) {
 // getExternalNetworkNames interrogates OpenStack to get the external network
 // names.
 func getExternalNetworkNames(cloud string) ([]string, error) {
-	conn, err := clientconfig.NewServiceClient("network", &clientconfig.ClientOpts{
-		Cloud: cloud,
-	})
+	conn, err := clientconfig.NewServiceClient("network", defaultValidValuesFetcherClientOpts(cloud))
```
mdbooth, line 31: Related, but not necessarily in this patch, I would refactor all of these functions to take a ProviderClient instead of a cloud name. The reason is you pull the code managing authentication options into a single place, and also you only hit keystone once rather than every time you create a new service client.

pkg/asset/installconfig/openstack/session.go
```diff
@@ -23,11 +22,6 @@ type Session struct {
 func GetSession(cloudName string) (*Session, error) {
```
mdbooth, line 22: I would prefer to see every case where we instantiate an openstack client call this method.

pkg/asset/installconfig/openstack/validvaluesfetcher.go
```diff
@@ -131,3 +125,13 @@ func getFloatingIPNames(cloud string, floatingNetworkName string) ([]string, err
...
 	return floatingIPNames, nil
 }
+
+func defaultValidValuesFetcherClientOpts(cloudName string) *clientconfig.ClientOpts {
+	opts := new(clientconfig.ClientOpts)
+	opts.Cloud = cloudName
+	// We explicitly disable reading auth data from env variables by setting an invalid EnvPrefix.
+	// By doing this, we make sure that the data from clouds.yaml is enough to authenticate.
+	// For more information: https://github.com/gophercloud/utils/blob/8677e053dcf1f05d0fa0a616094aace04690eb94/openstack/clientconfig/requests.go#L508
+	opts.EnvPrefix = "NO_ENV_VARIABLES_"
```
mdbooth, line 135: This is duplicating what we're already doing with defaultClientOpts(). I'd prefer we just called that.

EmilienM (changes requested):
please rebase on top of master as the PR that revendored gophercloud+deps just landed.

pierreprinetti:
/approve
/lgtm
/hold for CI

Note for self:
For Terraform, we already set OS_CLOUD explicitly here:

https://github.com/openshift/installer/blob/4bd58eb3d5a82058175d86f23ac6401aa70393a6/data/data/openstack/main.tf#L4

and an explicit value is supposed to override the env var default:

https://github.com/hashicorp/terraform-provider-openstack/blob/fc6a9169ef533cbaf9f914d44db92a36c4e0b8a3/openstack/provider.go#L223-L228

mandre (approved):
/lgtm

mandre (commented):
/approve

staebler (commented):
/approve

PR #4432, opened 2020-11-30
openstack: check quotas before creating cluster

EmilienM:
Note: openshift-installer already check quotas for AWS and GCP.

1) Calculate the quota constraints based on the InstallConfig.
   For both ControlPlane and workers, get the flavors and
   create a list of quota resources that will be needed to
   successfully deploy the cluster.
   Note: for now, only instances, CPUs and RAM are supported.
   In the future, we'll add networking and storage resources.

2) Fetch project quotas by using OpenStack API (via gophercloud)
   and create the constraints (for instances, CPUs and RAM only
   for now).
   The Quota constraints will be stored in the CloudInfo struct,
   for caching so we avoid multiple calls to the OpenStack APIs
   to get quotas.

3) The logging is improved when there is no region name for
   a cloud provider.

Fixes: [OSASINFRA-1141](https://issues.redhat.com/browse/OSASINFRA-1141)

Co-Authored-By: Matthew Booth <mbooth@redhat.com>

### DIFF
--- a/pkg/asset/installconfig/openstack/validation/cloudinfo.go
+++ b/pkg/asset/installconfig/openstack/validation/cloudinfo.go
@@ -7,7 +7,10 @@ import (
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/availabilityzones"
+	computequotasets "github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/quotasets"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
+	tokensv2 "github.com/gophercloud/gophercloud/openstack/identity/v2/tokens"
+	tokensv3 "github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
 	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
@@ -17,7 +20,9 @@ import (
 	imageutils "github.com/gophercloud/utils/openstack/imageservice/v2/images"
 	networkutils "github.com/gophercloud/utils/openstack/networking/v2/networks"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 
+	"github.com/openshift/installer/pkg/quota"
 	"github.com/openshift/installer/pkg/types"
 )
 
@@ -30,6 +35,7 @@ type CloudInfo struct {
 	MachinesSubnet  *subnets.Subnet
 	OSImage         *images.Image
 	Zones           []string
+	Quotas          []quota.Quota
 
 	clients *clients
 }
@@ -45,6 +51,13 @@ type clients struct {
 type Flavor struct {
 	*flavors.Flavor
 	Baremetal bool
+}
+
+// record stores the data from quota limits and usages.
+type record struct {
+	Service string
+	Name    string
+	Value   int64
 }
 
 // GetCloudInfo fetches and caches metadata from openstack
@@ -77,7 +90,7 @@ func GetCloudInfo(ic *types.InstallConfig) (*CloudInfo, error) {
 		return nil, errors.Wrap(err, "failed to create an image client")
 	}
 
-	err = ci.collectInfo(ic)
+	err = ci.collectInfo(ic, opts)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to generate OpenStack cloud info")
 	}
@@ -85,7 +98,7 @@ func GetCloudInfo(ic *types.InstallConfig) (*CloudInfo, error) {
 	return &ci, nil
 }
 
-func (ci *CloudInfo) collectInfo(ic *types.InstallConfig) error {
+func (ci *CloudInfo) collectInfo(ic *types.InstallConfig, opts *clientconfig.ClientOpts) error {
 	var err error
 
 	ci.ExternalNetwork, err = ci.getNetwork(ic.OpenStack.ExternalNetwork)
@@ -154,6 +167,15 @@ func (ci *CloudInfo) collectInfo(ic *types.InstallConfig) error {
 	ci.Zones, err = ci.getZones()
 	if err != nil {
 		return err
+	}
+
+	ci.Quotas, err = loadQuotas(opts)
+	if isUnauthorized(err) {
+		logrus.Warnf("Missing permissions to fetch Quotas and therefore will skip checking them: %v", err)
+		return nil
+	}
+	if err != nil {
+		return errors.Wrap(err, "failed to load Quota")
 	}
 
 	return nil
@@ -296,4 +318,115 @@ func (ci *CloudInfo) getZones() ([]string, error) {
 	}
 
 	return zones, nil
+}
+
+// loadLimits loads the consumer quota metric.
+func loadLimits(opts *clientconfig.ClientOpts) ([]record, error) {
+	var limits []record
+
+	projectID, err := getProjectID(opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get keystone project ID")
+	}
+
+	computeRecords, err := getComputeLimits(opts, projectID)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get compute quota records")
+	}
+	for _, r := range computeRecords {
+		limits = append(limits, r)
+	}
+
+	return limits, nil
+}
+
+func getComputeLimits(opts *clientconfig.ClientOpts, projectID string) ([]record, error) {
+	computeClient, err := clientconfig.NewServiceClient("compute", opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to connect against OpenStack Comute v2 API")
+	}
+	qs, err := computequotasets.GetDetail(computeClient, projectID).Extract()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get QuotaSets from OpenStack Compute API")
+	}
+
+	var records []record
+	addRecord := func(name string, quota computequotasets.QuotaDetail) {
+		records = append(records, record{
+			Service: "compute",
+			Name:    name,
+			Value:   int64(quota.Limit - quota.InUse - quota.Reserved),
+		})
+	}
+	addRecord("Cores", qs.Cores)
+	addRecord("Instances", qs.Instances)
+	addRecord("RAM", qs.RAM)
+
+	return records, nil
+}
+
+func getProjectID(opts *clientconfig.ClientOpts) (string, error) {
+	keystoneClient, err := clientconfig.NewServiceClient("identity", opts)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to conect against OpenStack Keystone API")
+	}
+	authResult := keystoneClient.GetAuthResult()
+	if authResult == nil {
+		return "", errors.Errorf("Client did not use openstack.Authenticate()")
+	}
+
+	switch authResult.(type) {
+	case tokensv2.CreateResult:
+		// Gophercloud has support for v2, but keystone has deprecated
+		// and it's not even documented.
+		return "", errors.Errorf("Extracting project ID using the keystone v2 API is not supported")
+
+	case tokensv3.CreateResult:
+		v3Result := authResult.(tokensv3.CreateResult)
+		project, err := v3Result.ExtractProject()
+		if err != nil {
+			return "", errors.Wrap(err, "Extracting project from v3 authResult")
+		} else if project == nil {
+			return "", errors.Errorf("Token is not scoped to a project")
+		}
+		return project.ID, nil
+
+	default:
+		return "", errors.Errorf("Unsupported AuthResult type: %T", authResult)
+	}
+}
+
+// loadQuotas loads the quota information for a project and provided services. It provides information
+// about the usage and limit for each resource quota.
+func loadQuotas(opts *clientconfig.ClientOpts) ([]quota.Quota, error) {
+	records, err := loadLimits(opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to load quota limits")
+	}
+	return newQuota(records), nil
+}
+
+func newQuota(limits []record) []quota.Quota {
+	var ret []quota.Quota
+	for _, limit := range limits {
+		q := quota.Quota{
+			Service: limit.Service,
+			Name:    limit.Name,
+			// Since limit.Value contains the actual
+			// available resources, we can set InUse to 0.
+			InUse: 0,
+			Limit: limit.Value,
+		}
+		ret = append(ret, q)
+	}
+	return ret
+}
+
+// isUnauthorized checks if the error is unauthorized (http code 403)
+func isUnauthorized(err error) bool {
+	if err == nil {
+		return false
+	}
+	var gErr gophercloud.ErrDefault403
+	return errors.As(err, &gErr)
 }
--- a/pkg/asset/quota/openstack/openstack.go
+++ b/pkg/asset/quota/openstack/openstack.go
@@ -1,0 +1,86 @@
+package openstack
+
+import (
+	machineapi "github.com/openshift/cluster-api/pkg/apis/machine/v1beta1"
+	"github.com/openshift/installer/pkg/asset/installconfig/openstack/validation"
+	"github.com/openshift/installer/pkg/quota"
+	openstackprovider "sigs.k8s.io/cluster-api-provider-openstack/pkg/apis/openstackproviderconfig/v1alpha1"
+)
+
+// Constraints returns a list of quota constraints based on the InstallConfig.
+// These constraints can be used to check if there is enough quota for creating a cluster
+// for the install config.
+func Constraints(ci *validation.CloudInfo, controlPlanes []machineapi.Machine, computes []machineapi.MachineSet) []quota.Constraint {
+	ctrplConfigs := make([]*openstackprovider.OpenstackProviderSpec, len(controlPlanes))
+	for i, m := range controlPlanes {
+		ctrplConfigs[i] = m.Spec.ProviderSpec.Value.Object.(*openstackprovider.OpenstackProviderSpec)
+	}
+	computeReplicas := make([]int64, len(computes))
+	computeConfigs := make([]*openstackprovider.OpenstackProviderSpec, len(computes))
+	for i, m := range computes {
+		computeReplicas[i] = int64(*m.Spec.Replicas)
+		computeConfigs[i] = m.Spec.Template.Spec.ProviderSpec.Value.Object.(*openstackprovider.OpenstackProviderSpec)
+	}
+
+	var ret []quota.Constraint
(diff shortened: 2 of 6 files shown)


abhinavdahiya:
quotas are more related to _counts_ of the resources like do you have x number of instances of a type etc.
CPU, RAM seem more like does this instance type work for the CPU/RAM requirements for openshift components 16G for control plane and 8G for workers. this probably should be a validation of the flavor chosen instead of quota check?

EmilienM:
> quotas are more related to _counts_ of the resources like do you have x number of instances of a type etc.
> CPU, RAM seem more like does this instance type work for the CPU/RAM requirements for openshift components 16G for control plane and 8G for workers. this probably should be a validation of the flavor chosen instead of quota check?

Flavors in OpenStack aren't predictable like they are with Amazon, so what we do here is that we compute the amount of resources that are needed by the flavors being used and compare with what the cloud quotas allow us to create in the cloud. This is similar with what AWS & GCP providers do.

As far as your suggestion for flavor validation, this is a different issue where we check that the flavors used are well sized so masters/workers can be installed. We already do flavor validation here:
https://github.com/openshift/installer/blob/36cf196e4bb097967ed6d0355f448eaf3c9763af/pkg/asset/installconfig/openstack/validation/machinepool.go#L21-L32

Thanks in advance for your review.

pkg/asset/quota/openstack/openstack.go
```diff
@@ -0,0 +1,119 @@
...
+			quotas[i].Count += quotas[j].Count
+		} else {
+			i++
+			if i != j {
+				quotas[i] = quotas[j]
+			}
+		}
+	}
+	return quotas[:i+1]
+}
```
staebler, line 85: Using a temporary map would make this easier to follow.
```suggestion
func aggregate(quotas []quota.Constraint) []quota.Constraint {
	counts := map[string]int64{}
	for _, q := range quotas {
		counts[q.Name] = counts[q.Name] + q.Count
	}
	aggregatedQuotas := make([]quota.Constraint, 0, len(counts))
	for n, c := range counts {
		aggregatedQuotas = append(aggregatedQuotas, quota.Constraint{Name: n, Count: c})
	}
	return aggregatedQuotas
}
```

pkg/asset/quota/openstack/openstack.go
```diff
@@ -0,0 +1,119 @@
...
+
+//constraintGenerator generates a list of constraints.
+type constraintGenerator func() []quota.Constraint
+
+func machineFlavorCoresToQuota(f string, config *types.InstallConfig) quota.Constraint {
+	ci, _ := validation.GetCloudInfo(config)
+	flavor, _ := ci.Flavors[f]
+	var ret quota.Constraint
+	ret = generateConstraint("Cores", int64(flavor.VCPUs))
+	return ret
```
staebler, line 95:
```suggestion
	return generateConstraint("Cores", int64(flavor.VCPUs))
```

pkg/asset/quota/openstack/openstack.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func generateConstraint(name string, count int64) quota.Constraint {
+	var ret quota.Constraint
+	ret = quota.Constraint{
+		Name:  name,
+		Count: count,
+	}
+	return ret
```
staebler, line 118:
```suggestion
	return quota.Constraint{
		Name:  name,
		Count: count,
	}
```

pkg/asset/quota/openstack/openstack.go
```diff
@@ -0,0 +1,119 @@
...
+	ret = generateConstraint("Cores", int64(flavor.VCPUs))
+	return ret
+}
+
+func machineFlavorRAMToQuota(f string, config *types.InstallConfig) quota.Constraint {
+	ci, _ := validation.GetCloudInfo(config)
+	flavor, _ := ci.Flavors[f]
+	var ret quota.Constraint
+	ret = generateConstraint("RAM", int64(flavor.RAM))
+	return ret
```
staebler, line 103:
```suggestion
	return generateConstraint("RAM", int64(flavor.RAM))
```

pkg/asset/quota/openstack/openstack.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func instanceConstraint(count int64) quota.Constraint {
+	var ret quota.Constraint
+	ret = generateConstraint("Instances", count)
+	return ret
```
staebler, line 109:
```suggestion
	return generateConstraint("Instances", count)
```

pkg/asset/quota/openstack/openstack.go
```diff
@@ -0,0 +1,119 @@
...
+	for i, w := range computes {
+		computeReplicas[i] = int64(*w.Spec.Replicas)
+		computeConfigs[i] = w.Spec.Template.Spec.ProviderSpec.Value.Object.(*openstackprovider.OpenstackProviderSpec)
+	}
+
+	var ret []quota.Constraint
+	for _, gen := range []constraintGenerator{
+		controlPlane(config, ctrplConfigs),
+		compute(config, computeReplicas, computeConfigs),
+		others,
```
staebler, line 32: Why do we need this, since it never yields any constraints?

pkg/asset/quota/openstack/openstack.go
```diff
@@ -0,0 +1,119 @@
...
+	} {
+		ret = append(ret, gen()...)
+	}
```
staebler, line 35: I think we're being too cute here. Change `controlPlane` and `compute` to return a slice of constraints instead of a function.
```suggestion
	ret := controlPlane(config, ctrplConfigs)
	ret = append(ret, compute(config, computeReplicas, computeConfigs)...)
```

pkg/asset/quota/quota.go
```diff
@@ -135,10 +155,16 @@ func (a *PlatformQuotaCheck) Name() string {
...
 	for _, report := range reports {
 		switch report.Result {
 		case quota.NotAvailable:
-			notavailable = append(notavailable, fmt.Sprintf("%s is not available in %s because %s", report.For.Name, report.For.Region, report.Message))
+			if report.For.Region != "" {
+				regionMessage = "in " + report.For.Region
+			} else {
+				regionMessage = ""
+			}
+			notavailable = append(notavailable, fmt.Sprintf("%s is not available %s because %s", report.For.Name, regionMessage, report.Message))
```
staebler, line 167: If there is no region, we end up with 2 spaces between "available" and "because"
```suggestion
			if report.For.Region != "" {
				regionMessage = " in " + report.For.Region
			} else {
				regionMessage = ""
			}
			notavailable = append(notavailable, fmt.Sprintf("%s is not available%s because %s", report.For.Name, regionMessage, report.Message))
```

pkg/asset/quota/quota.go
```diff
@@ -162,10 +188,16 @@ func summarizeFailingReport(reports []quota.ConstraintReport) error {
...
 	for _, report := range reports {
 		switch report.Result {
 		case quota.AvailableButLow:
-			low = append(low, fmt.Sprintf("%s (%s)", report.For.Name, report.For.Region))
+			if report.For.Region != "" {
+				regionMessage = "(" + report.For.Region + ")"
+			} else {
+				regionMessage = ""
+			}
+			low = append(low, fmt.Sprintf("%s %s", report.For.Name, regionMessage))
```
staebler, line 200:
```suggestion
			if report.For.Region != "" {
				regionMessage = " (" + report.For.Region + ")"
			} else {
				regionMessage = ""
			}
			low = append(low, fmt.Sprintf("%s%s", report.For.Name, regionMessage))
```

pkg/asset/quota/openstack/openstack.go
```diff
@@ -0,0 +1,119 @@
...
+	return quotas[:i+1]
+}
+
+//constraintGenerator generates a list of constraints.
+type constraintGenerator func() []quota.Constraint
+
+func machineFlavorCoresToQuota(f string, config *types.InstallConfig) quota.Constraint {
+	ci, _ := validation.GetCloudInfo(config)
+	flavor, _ := ci.Flavors[f]
```
staebler, line 92:
```suggestion
	flavor := ci.Flavors[f]
```

pkg/asset/quota/openstack/openstack.go
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}
+	return quotas[:i+1]
+}
+
+//constraintGenerator generates a list of constraints.
+type constraintGenerator func() []quota.Constraint
+
+func machineFlavorCoresToQuota(f string, config *types.InstallConfig) quota.Constraint {
+	ci, _ := validation.GetCloudInfo(config)
```
staebler, line 91: We cannot ignore the error returned here.

pkg/asset/quota/openstack/openstack.go
EmilienM: I followed what was done for GCP/AWS:

https://github.com/openshift/installer/blob/47b4c1a046ee43bc500ff19c98a231a754cc2a70/pkg/asset/quota/gcp/gcp.go#L16-L43

https://github.com/openshift/installer/blob/47b4c1a046ee43bc500ff19c98a231a754cc2a70/pkg/asset/quota/aws/aws.go#L14-L39

But if you want me to change it I'm happy to do it.

pkg/asset/quota/openstack/openstack.go
EmilienM: ack, I'll take that suggestion, thanks.

pkg/asset/quota/openstack/openstack.go
EmilienM: Noted; thanks for the suggestion, I'll change the patch.

Note that I re-used a function from AWS quotas:

https://github.com/openshift/installer/blob/47b4c1a046ee43bc500ff19c98a231a754cc2a70/pkg/asset/quota/aws/aws.go#L41-L58

pkg/asset/quota/openstack/openstack.go
EmilienM: I agree, a mistake on my side. I'll fix this.

EmilienM (commented):
thanks @staebler for your insightful review, I'll update the patch.

EmilienM:
Fun fact: CI is now failing because the not enough quotas are available :-)

pkg/asset/quota/openstack/openstack.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return aggregatedQuotas
+}
+
+//constraintGenerator generates a list of constraints.
+type constraintGenerator func() []quota.Constraint
+
+func machineFlavorCoresToQuota(f string, config *types.InstallConfig) quota.Constraint {
+	var ret quota.Constraint
+	ci, err := validation.GetCloudInfo(config)
```
mandre, line 69: The `GetCloudInfo` function is quite expensive and you call it twice for each machine. How about passing a reference to the `CloudInfo` object to the `machineFlavorCoresToQuota` and `machineFlavorRAMToQuota` functions instead?

pkg/asset/quota/quota.go
```diff
@@ -118,7 +122,23 @@ func (a *PlatformQuotaCheck) Generate(dependencies asset.Parents) error {
 			return summarizeFailingReport(reports)
 		}
 		summarizeReport(reports)
-	case azure.Name, baremetal.Name, libvirt.Name, none.Name, openstack.Name, ovirt.Name, vsphere.Name:
+	case typesopenstack.Name:
+		opts := &openstackclientconfig.ClientOpts{Cloud: ic.Config.Platform.OpenStack.Cloud}
+		os.Unsetenv("OS_CLOUD")
```
mandre, line 127: I really wish we could wrap that unset OS_CLOUD somehow.

pkg/asset/quota/quota.go
```diff
@@ -118,7 +122,23 @@ func (a *PlatformQuotaCheck) Generate(dependencies asset.Parents) error {
...
+		q, err := quotaopenstack.Load(opts)
```
mandre, line 128: Hmm, we probably want to store the quota info in the `CloudInfo` stuct.

Then we could call `GetCloudInfo` here. This would take care of unsetting the OS_CLOUD variable for us and make it easier to deal with self-signed certificates.

pkg/quota/openstack/limits.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	for _, r := range computeRecords {
+		limits = append(limits, r)
+	}
+
+	return limits, nil
+}
+
+func getComputeLimits(pClient *gophercloud.ProviderClient, eo gophercloud.EndpointOpts, projectID string) ([]record, error) {
+	computeClient, err := openstack.NewComputeV2(pClient, eo)
```
mandre, line 52: If you need to create an openstack client, try to use the `NewServiceClient()` function instead as it will make it easier to [support self-signed certificates](https://github.com/openshift/installer/pull/4405).

pkg/asset/quota/openstack/openstack.go
```diff
@@ -0,0 +1,96 @@
...
+// These constraints can be used to check if there is enough quota for creating a cluster
+// for the install config.
+func Constraints(config *types.InstallConfig, controlPlanes []machineapi.Machine, computes []machineapi.MachineSet) []quota.Constraint {
+	ctrplConfigs := make([]*openstackprovider.OpenstackProviderSpec, len(controlPlanes))
+	for i, m := range controlPlanes {
+		ctrplConfigs[i] = m.Spec.ProviderSpec.Value.Object.(*openstackprovider.OpenstackProviderSpec)
+	}
+	computeReplicas := make([]int64, len(computes))
+	computeConfigs := make([]*openstackprovider.OpenstackProviderSpec, len(computes))
+	for i, w := range computes {
```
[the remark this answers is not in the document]
EmilienM: I'll do like controlPlanes and use `m` for `Machine`. Let me know if you don't like it.

pkg/asset/quota/quota.go
EmilienM: Yeah I know, this was mentioned here: https://github.com/openshift/installer/pull/4426#issuecomment-734891653

I propose that we leave it here for now, and submit a separated PR to address this valid issue.

pkg/quota/openstack/limits.go
EmilienM: ack, and just to make sure the rest is fine, can you confirm that we don't need to call `NewServiceClient()` for:

https://github.com/openshift/installer/blob/9665db1d1783ff540e9631501ea39e54490eb265/pkg/quota/openstack/limits.go#L77

pkg/asset/quota/quota.go
EmilienM: ack, looking at it now.

pkg/asset/quota/quota.go
EmilienM: nevermind, I managed to get rid of it when using CloudInfo struct. Thanks!

pkg/asset/quota/quota.go
```diff
@@ -118,7 +120,17 @@ func (a *PlatformQuotaCheck) Generate(dependencies asset.Parents) error {
 			return summarizeFailingReport(reports)
 		}
 		summarizeReport(reports)
-	case azure.Name, baremetal.Name, libvirt.Name, none.Name, openstack.Name, ovirt.Name, vsphere.Name:
+	case typesopenstack.Name:
+		ci, err := openstackvalidation.GetCloudInfo(ic.Config)
+		if err != nil {
+			return errors.Wrap(err, "failed to get cloud info")
+		}
+		reports, err := quota.Check(ci.Quotas, openstack.Constraints(ic.Config, masters, workers))
```
mandre, line 128: You can pass a reference to the CloudInfo object to `Constraints()` now, in order to not call `GetCloudInfo()` from the [pkg/asset/quota/openstack/openstack.go](https://github.com/openshift/installer/pull/4432/files#diff-db1151cdf97e6679829445032d2f0815774f20dfd956c48d2863321979f098fe) file.

pkg/quota/openstack/limits.go
```diff
@@ -0,0 +1,92 @@
+package openstack
```
mandre, line 1: I would even go further and move all of this into `pkg/asset/installconfig/openstack/validation/cloudinfo.go`.

pkg/asset/quota/openstack/openstack.go
EmilienM: yeah I did it and the Platform Quota Check is around 3 times faster from my initial testing, thanks!

pkg/quota/openstack/limits.go
EmilienM: ack, I'm looking at it now.

pkg/asset/quota/quota.go
EmilienM: ack, I'm looking at it now.

pkg/asset/installconfig/openstack/validation/cloudinfo.go
```diff
@@ -297,3 +319,114 @@ func (ci *CloudInfo) getZones() ([]string, error) {
...
+		return project.ID, nil
+
+	default:
+		return "", errors.Errorf("Unsupported AuthResult type: %T", authResult)
+	}
+}
+
+// LoadQuotas loads the quota information for a project and provided services. It provides information
+// about the usage and limit for each resource quota..
+func LoadQuotas(opts *clientconfig.ClientOpts) ([]quota.Quota, error) {
```
mandre, line 401: This function doesn't have to be public.
```suggestion
func loadQuotas(opts *clientconfig.ClientOpts) ([]quota.Quota, error) {
```

pkg/asset/installconfig/openstack/validation/cloudinfo.go
```diff
@@ -297,3 +319,114 @@ func (ci *CloudInfo) getZones() ([]string, error) {
...
+			InUse: 0,
+			Limit: limit.Value,
+		}
+		ret = append(ret, q)
+	}
+	return ret
+}
+
+// IsUnauthorized checks if the error is unauthorized (http code 403)
+func IsUnauthorized(err error) bool {
```
mandre, line 426: Same here.
```suggestion
func isUnauthorized(err error) bool {
```

mandre (commented):
/lgtm

@staebler PTAL for approval.

staebler (commented):
/approve

staebler (commented):
/lgtm

pkg/asset/quota/quota.go
```diff
@@ -119,7 +121,17 @@ func (a *PlatformQuotaCheck) Generate(dependencies asset.Parents) error {
...
+		ci, err := openstackvalidation.GetCloudInfo(ic.Config)
+		if err != nil {
+			return errors.Wrap(err, "failed to get cloud info")
+		}
+		reports, err := quota.Check(ci.Quotas, openstack.Constraints(ci, masters, workers))
+		if err != nil {
+			return summarizeFailingReport(reports)
+		}
+		summarizeReport(reports)
+	case azure.Name, baremetal.Name, libvirt.Name, none.Name, ovirt.Name, vsphere.Name, kubevirt.Name::
```
staebler, line 134:
```suggestion
	case azure.Name, baremetal.Name, libvirt.Name, none.Name, ovirt.Name, vsphere.Name, kubevirt.Name:
```

iamemilio:
openstack passing, will remove hold

/hold cancel

PR #4433, opened 2020-11-30
Adds platform.openstack.useHwQEMUAgentForClusterOSImage

adduarte:
useHwQEMUAgentForClusterOSImage causes "hw_qemu_guest_agent=yes"
to be added to the properties metadata of the Glance image uploaded
for clusterOSImage

### DIFF
--- a/pkg/types/openstack/platform.go
+++ b/pkg/types/openstack/platform.go
@@ -62,6 +62,10 @@ type Platform struct {
 	// +optional
 	ClusterOSImage string `json:"clusterOSImage,omitempty"`
 
+	// UseHwQEMUAgentForClusterOSImage indicates that the property "hw_qemu_guest_agent=yes" should be added to the Glance ClusterOSImage
+	// +optional
+	UseHwQEMUAgentForClusterOSImage bool `json:"useHwQEMUAgentForClusterOSImage,omitempty`
+
 	// APIVIP is the static IP on the nodes subnet that the api port for openshift will be assigned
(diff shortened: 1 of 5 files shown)


pkg/types/openstack/platform.go
```diff
@@ -62,6 +62,10 @@ type Platform struct {
 	// +optional
 	ClusterOSImage string `json:"clusterOSImage,omitempty"`
 
+	// UseHwQEMUAgentForClusterOSImage indicates that the property "hw_qemu_guest_agent=yes" should be added to the Glance ClusterOSImage
+	// +optional
+	UseHwQEMUAgentForClusterOSImage bool `json:"useHwQEMUAgentForClusterOSImage,omitempty`
```
abhinavdahiya, line 67:
```suggestion
	UseHWQEMUAgentForClusterOSImage bool `json:"useHwQEMUAgentForClusterOSImage,omitempty`
```

also maybe create a *ClusterOSImageProperties struct if you think there will be more properties. and enforce that it be nil when clusterOSImage is set to pre existing glance image.

also needs more clearer documentation when this field is used and not, like is user gave us glance image we probably can't change or add these properties.

abhinavdahiya (commented):
would like to see some documentation or enhancement or link to RFE or something that provides the background on why we need to support this and how important it is.

adduarte:
This was just an example, the changes will be incorporated into https://github.com/openshift/installer/pull/4401

PR #4438, opened 2020-12-01
WIP: OCVRHV-45: Allow preallocated disk

janoszen:
This change adds the option to use full clones instead of thin provisioning for the master nodes. Full provisioning increases the performance at the cost of up-front disk usage. The default is now on, but can be turned off by the user in the survey.

The install config can be set to the old behavior (thin provisioning) by setting the `ovirt_master_clone` option to `false`.

**Jira issue:** https://issues.redhat.com/browse/OCPRHV-45

## User experience

```
$ bin/openshift-install create install-config
? Platform ovirt
? Cluster ocp-on-rhv
? Storage domain vserver-spider-nfs-lif1
? Network ovn-vmnet-11
? Internal API virtual IP 192.168.211.30
? Ingress virtual IP 192.168.211.32
? Thick provisioning improves master node performance, but makes the installation process slower
and will consume the entire disk up-front on the oVirt cluster.
? Use thick disk provisioning for master nodes? (Y/n)
```

## Disabling thick provisioning from the install-config

Please add the following option to your YAML:

```yaml
platform:
  ovirt:
    ovirt_master_clone: false
```

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -1226,6 +1226,9 @@ spec:
                   ovirt_cluster_id:
                     description: The target cluster under which all VMs will run
                     type: string
+                  ovirt_master_clone:
+                    description: MasterClone indicates that the disk of the master node should be cloned (thick provisioning)
+                    type: boolean
                   ovirt_network_name:
                     description: NetworkName is the target network of all the network interfaces of the nodes. When no ovirt_network_name is provided it defaults to `ovirtmgmt` network, which is a default network for every ovirt cluster.
                     type: string
@@ -1239,6 +1242,7 @@ spec:
                 - api_vip
                 - ingress_vip
                 - ovirt_cluster_id
+                - ovirt_master_clone
                 - ovirt_storage_domain_id
                 type: object
               vsphere:
--- a/pkg/asset/installconfig/ovirt/ovirt.go
+++ b/pkg/asset/installconfig/ovirt/ovirt.go
@@ -92,5 +92,19 @@ func Platform() (*ovirt.Platform, error) {
 		return nil, errors.Wrap(err, "failed UserInput")
 	}
 
+	err = survey.Ask([]*survey.Question{
+		{
+			Prompt: &survey.Confirm{
+				Message: "Use thick disk provisioning for master nodes?",
+				Help:    "Thick provisioning improves master node performance, but makes the installation process slower and will consume the entire disk up-front on the oVirt cluster.",
+				Default: true,
+			},
+			Validate: survey.ComposeValidators(survey.Required),
+		},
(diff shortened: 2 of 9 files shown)


data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -1226,6 +1226,9 @@ spec:
                   ovirt_cluster_id:
                     description: The target cluster under which all VMs will run
                     type: string
+                  ovirt_master_clone:
+                    description: MasterClone indicates that the disk of the master node should be cloned (thick provisioning)
```
michalskrivanek, line 1230: what's the reason for naming it this way? shouldn't it say something more tangible..like "use_disk_thick_provisioning" or "use_preallocated_disks" or something like that?

michalskrivanek: maybe still keep "master" but eventually i would want this option to apply to workers too, even if the code for that is elsewhere. 
not necessarily now

janoszen: The oVirt API calls it clone and I believed it best to stay in convention. Adding an additional name for the same thing would increase potential confusion.

michalskrivanek: on a second thought seeing the right comments throughout the code it's fine as is for me too, if it's better to align the parameter name with go-ovirt

michalskrivanek: yeah. it's an unfortunate naming. this parameter is confusing everyone in oVirt to this day:/

pkg/asset/installconfig/ovirt/ovirt.go
```diff
@@ -92,5 +92,19 @@ func Platform() (*ovirt.Platform, error) {
 		return nil, errors.Wrap(err, "failed UserInput")
 	}
 
+	err = survey.Ask([]*survey.Question{
+		{
+			Prompt: &survey.Confirm{
+				Message: "Use thick disk provisioning for master nodes?",
+				Help:    "Thick provisioning improves master node performance, but makes the installation process slower and will consume the entire disk up-front on the oVirt cluster.",
```
michalskrivanek, line 99: thinking a bit more about it, I wouldn't want to claim that it's slower. Some aspects are slower for sure, the initial master VMs creation. But it's most likely not that significant considering the overall deployment time. And on the other hand the operations are supposedly a tiny bit faster so it easily could be that the rest of the procedure offsets for it. 
I would rephrase and leave out "makes the installation process slower and"

eslutsky (approved):

janoszen:
Heads up, looks like the `clone` flag doesn't do what we think it does. It clones the base image, but still leaves the disk on thin provisioning. This means that the `clone` flag won't help us here.

PR #4439, opened 2020-12-01
Create SR-IOV UPI Docs

iamemilio:
Adds docs for manually creating worker nodes with SR-IOV NICs attached. These
docs walk users through what basic components are needed and why, as well as
providing examples using the openstack command line client.

Fixes: [OSASINFRA-2159](https://issues.redhat.com/browse/OSASINFRA-2159)

### DIFF
--- a/docs/user/openstack/install_upi.md
+++ b/docs/user/openstack/install_upi.md
@@ -58,6 +58,7 @@ of this method of installation.
     - [Compute Nodes Trunks (Kuryr SDN)](#compute-nodes-trunks-kuryr-sdn)
     - [Approve the worker CSRs](#approve-the-worker-csrs)
     - [Wait for the OpenShift Installation to Complete](#wait-for-the-openshift-installation-to-complete)
+    - [Compute Nodes with SR-IOV NICs](#compute-nodes-with-sr-iov-nics)
   - [Destroy the OpenShift Cluster](#destroy-the-openshift-cluster)
 
 ## Prerequisites
@@ -312,7 +313,7 @@ open(path, "w").write(yaml.dump(data, default_flow_style=False))'
 Next, we need to correct the `apiVIP` and `ingressVIP` values.
 
 The following script will clear the values from the `install-config.yaml` file so that the installer will pick
-the 5th and 7th IP addresses in the new range, 192.0.2.5 and 192.0.2.7. 
+the 5th and 7th IP addresses in the new range, 192.0.2.5 and 192.0.2.7.
 
 ```sh
 $ python -c 'import yaml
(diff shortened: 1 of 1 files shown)


docs/user/openstack/install_upi.md
```diff
@@ -885,6 +886,34 @@ The workers need no ignition override.
 
 If `os_networking_type` is set to `Kuryr` in the Ansible inventory, the playbook creates the Trunks for Kuryr to plug the containers into the OpenStack SDN.
 
+### Compute Nodes with SR-IOV NICs
```
maxwelldb, line 889: Would it make sense to add:

* A definition for "SR-IOV" before using the abbreviation.
* Circumstances where you would want to pay attention to this section. 
* Any differences between clusters that rely on SR-IOV and those that don't.

docs/user/openstack/install_upi.md
```diff
@@ -885,6 +886,34 @@ The workers need no ignition override.
...
+openstack port os_port_worker_0 --network <infraID>-network --security-group <infraID>-worker --fixed-ip subnet=<infraID>-nodes,ip-address=<a fixed IP> --allowed-address ip-address=<infraID>-ingress-port
+```
+
+The SR-IOV port(s) must be created explicitly by the user and passed as a NIC during instance creation, otherwise the `vnic-type` will not be `direct` and it will not work.
+
+```sh
+openstack port create radio_port --vnic-type direct --network radio --fixed-ip subnet=radio,ip-address=<a fixed ip> --tag=radio --disable-port-security
+```
+
+When you create your instance, make sure that the SR-IOV port and the ocp port you created for it are added as a NICs.
```
maxwelldb, line 911:
```suggestion
When you create your instance, make sure that the SR-IOV port and the ocp port you created for it are added as NICs.
```

ocp -> OCP?

atyronesmith (approved):

docs/user/openstack/install_upi.md
```diff
@@ -885,6 +886,34 @@ The workers need no ignition override.
...
+openstack subnet create radio --network radio --subnet-range <your CIDR range> --dhcp
+```
+
+Your compute nodes will need to have two types of ports for this feature to work. One port needs to connect the node to your OpenShift network so that it can join the cluster and communicate with the other nodes. The other type of port is for your SR-IOV traffic. The OpenShift networking port should be created the same way we normally create ports for compute nodes.
+
+```sh
+openstack port os_port_worker_0 --network <infraID>-network --security-group <infraID>-worker --fixed-ip subnet=<infraID>-nodes,ip-address=<a fixed IP> --allowed-address ip-address=<infraID>-ingress-port
+```
+
+The SR-IOV port(s) must be created explicitly by the user and passed as a NIC during instance creation, otherwise the `vnic-type` will not be `direct` and it will not work.
```
adduarte, line 905: Also worth adding that a good check after everything is done is that the vnic-type is indeed set to direct.

docs/user/openstack/install_upi.md
```diff
@@ -885,6 +886,34 @@ The workers need no ignition override.
...
+
+```sh
+openstack port create radio_port --vnic-type direct --network radio --fixed-ip subnet=radio,ip-address=<a fixed ip> --tag=radio --disable-port-security
+```
+
+When you create your instance, make sure that the SR-IOV port and the OCP port you created for it are added as NICs.
```
adduarte, line 911: Maybe move this info a few lines above, so it is clear that both nics need to be added to  instance on first creation.

adduarte (approved):
This looks good to me. The suggestions are not necessary to merge/approve

PR #4440, opened 2020-12-01
Bug 1903277: destroy/aws: remove rules from default security group

staebler:
Even though the destroyer cannot directly delete the default security group, the destroyer still needs to remove the ingress and egress rules from the default security group. There could be rules on the default security group that source another security group, which would block the destroyer from deleting that other security group.

https://bugzilla.redhat.com/show_bug.cgi?id=1903277

### DIFF
--- a/pkg/destroy/aws/aws.go
+++ b/pkg/destroy/aws/aws.go
@@ -1096,11 +1096,6 @@ func deleteEC2SecurityGroup(ctx context.Context, client *ec2.EC2, id string, log
 }
 
 func deleteEC2SecurityGroupObject(ctx context.Context, client *ec2.EC2, group *ec2.SecurityGroup, logger logrus.FieldLogger) error {
-	if group.GroupName != nil && *group.GroupName == "default" {
-		logger.Debug("Skipping default security group")
-		return nil
-	}
-
 	if len(group.IpPermissions) > 0 {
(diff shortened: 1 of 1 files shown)


wking (approved):
/lgtm
/approve

I'm not convinced this is what was happening in this particular stuck-security-group issue, but I agree that the default group referencing a separate security group is theoretically possible.

wking:
@staebler found an example of this, where the default security group had [`IpPermissions.UserIdGroupPairs`][1] entries mentioning other security groups in the VPC.  So they were allowing instances and other things in the other security groups to access instances and other things in the default security group.  We're not yet sure what created those entries, since we don't expect to be actually using the default security group for anything.

[1]: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_IpPermission.html#API_IpPermission_Contents

staebler:
Note that for shared VPC, the rules will not be deleted from the default security group. The destroyer does not delete the shared VPC, so the destroyer does not consider the default security group. If we diagnose that there is a legitimate for creating the ingress rule in the default security group, then we will need to add more code to handle the clean up in the case of shared VPC. In the meantime, I would like to push forward the changes as they are in this PR in order to stop the leaking of resources in CI.

staebler:
It took 1.5 hours to acquire the aws-quota-slice lease. The deprovision had 33 minutes left in the allotted 4 hours and did not complete in time.

/test e2e-aws

PR #4409, opened 2020-11-24
Set default release image to 4.7

pierreprinetti:
Align the default release image with the current development version.

PR #4442, opened 2020-12-02
Set default release image to 4.6

pierreprinetti:
Align the default release image with the current release version.

This is the 4.6 counterpart of the [patch](https://github.com/openshift/installer/pull/4409) on `master`.

Please let me know if this patch needs a BZ, or feel free to directly close it if it is not worth the effort!

### DIFF
--- a/pkg/asset/releaseimage/default.go
+++ b/pkg/asset/releaseimage/default.go
@@ -20,7 +20,7 @@ import (
 
 var (
 	// defaultReleaseImageOriginal is the value served when defaultReleaseImagePadded is unmodified.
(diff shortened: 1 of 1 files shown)


staebler (commented):
This change looks fine. But I'm not sure it meets the bar for inclusion in 4.6. At a minimum, it would need a BZ. That BZ should describe the impact and whom it affects.

pierreprinetti:
> I'm not sure it meets the bar for inclusion in 4.6.

Perfect, I’ve only filed this for consistency. I’ll reopen if I directly face issues with the skewed version. Thank you!

PR #4443, opened 2020-12-02
docs: update docs/design/resource_dep.svg

hardys:
I missed updating this in #4413 which introduced a new
asset

Generated via:
bin/openshift-install graph | dot -Tsvg >docs/design/resource_dep.svg


staebler (approved):
/lgtm
/approve

PR #4447, opened 2020-12-03
cloudproviderconfig: Refactor openstack

pierreprinetti:
Move OpenStack platform-specific code in the platform package to
increase ownership and leave a clean API footprint on the orchestrating
logic.

Implements [OSASINFRA-2195](https://issues.redhat.com/browse/OSASINFRA-2195)

### DIFF
--- a/pkg/asset/manifests/cloudproviderconfig.go
+++ b/pkg/asset/manifests/cloudproviderconfig.go
@@ -2,7 +2,6 @@ package manifests
 
 import (
 	"fmt"
-	"io/ioutil"
 	"path/filepath"
 
 	"github.com/ghodss/yaml"
@@ -13,7 +12,6 @@ import (
 
 	"github.com/openshift/installer/pkg/asset"
 	"github.com/openshift/installer/pkg/asset/installconfig"
-	icopenstack "github.com/openshift/installer/pkg/asset/installconfig/openstack"
 	"github.com/openshift/installer/pkg/asset/manifests/azure"
 	gcpmanifests "github.com/openshift/installer/pkg/asset/manifests/gcp"
 	openstackmanifests "github.com/openshift/installer/pkg/asset/manifests/openstack"
@@ -93,22 +91,17 @@ func (cpc *CloudProviderConfig) Generate(dependencies asset.Parents) error {
 			return nil
 		}
 		cm.Data[cloudProviderConfigCABundleDataKey] = trustBundle
+
 	case openstacktypes.Name:
-		cloud, err := icopenstack.GetSession(installConfig.Config.Platform.OpenStack.Cloud)
+		cloudProviderConfigData, cloudProviderConfigCABundleData, err := openstackmanifests.GenerateCloudProviderConfig(*installConfig.Config)
 		if err != nil {
-			return errors.Wrap(err, "failed to get cloud config for openstack")
+			return errors.Wrap(err, "failed to generate OpenStack provider config")
+		}
+		cm.Data[cloudProviderConfigDataKey] = cloudProviderConfigData
+		if cloudProviderConfigCABundleData != "" {
+			cm.Data[cloudProviderConfigCABundleDataKey] = cloudProviderConfigCABundleData
 		}
 
-		cm.Data[cloudProviderConfigDataKey] = openstackmanifests.CloudProviderConfig(cloud.CloudConfig)
-
-		// Get the ca-cert-bundle key if there is a value for cacert in clouds.yaml
-		if caPath := cloud.CloudConfig.CACertFile; caPath != "" {
-			caFile, err := ioutil.ReadFile(caPath)
-			if err != nil {
-				return errors.Wrap(err, "failed to read clouds.yaml ca-cert from disk")
-			}
-			cm.Data[cloudProviderConfigCABundleDataKey] = string(caFile)
-		}
 	case azuretypes.Name:
 		session, err := installConfig.Azure.Session()
 		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/asset/manifests/cloudproviderconfig.go
```diff
@@ -93,22 +91,17 @@ func (cpc *CloudProviderConfig) Generate(dependencies asset.Parents) error {
 			return nil
 		}
 		cm.Data[cloudProviderConfigCABundleDataKey] = trustBundle
+
 	case openstacktypes.Name:
-		cloud, err := icopenstack.GetSession(installConfig.Config.Platform.OpenStack.Cloud)
+		cloudProviderConfigData, cloudProviderConfigCABundleData, err := openstackmanifests.GenerateCloudProviderConfig(*installConfig.Config)
```
staebler, line 96: Why not just pass the pointer to the install config rather than dereferencing it here?

pkg/asset/manifests/openstack/cloudproviderconfig.go
```diff
@@ -1,29 +1,25 @@
...
-	}
-
-	if cloud.CACertFile != "" {
-		res += "ca-file = /etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem\n"
-	}
-
-	return res
+// Error represents a failure while generating OpenStack provider
+// configuration.
+type Error struct {
```
staebler, line 15: Why do we need this error wrapper? Can we use `Wrap` from `github.com/pkg/errors`?

staebler (commented):
+1

pkg/asset/manifests/openstack/cloudproviderconfig.go
pierreprinetti: My rationale is twofold:
1. switching to the pure-stdlib `errors.Unwrapper` interface
2. removing one external dependency in a couple lines of code

pkg/asset/manifests/cloudproviderconfig.go
pierreprinetti: By passing a copy of the struct instead of a pointer, I make it clear that we expect `GenerateCloudProviderConfig` to make no modification to `installConfig.Config`. I see that as "documentation in code".

Another advantage of dereferencing early is that I don't carry around the itchy `nil` possibility any deeper.

Also I have a personal adversity for passing pointers around, to which I would stick unless there's more pressing business need?

pkg/asset/manifests/openstack/cloudproviderconfig.go
pierreprinetti: Happy to revert to Dave Cheney’s package if it serves consistency

pkg/asset/manifests/cloudproviderconfig.go
pierreprinetti: Note that this is a cosmetic change I am not wedded to

staebler: I think you are going to be hard-pressed holding that line in go, but I have no objection to you doing it here.

pkg/asset/manifests/openstack/cloudproviderconfig.go
staebler: 1) The error returned from `Wrap` already implements the `Unwrapper` interface.
2) The installer already has that dependency.
3) I could see that if it was a meaningful type that provided some more information about the type of the error or a way to get information specific to the error. I personally don't see that value in this case.

With that said, this is code that you are going to own. So if you feel that the value of the type outweighs the maintentance burden, then have at it.

staebler (commented):
/approve
/lgtm

pierreprinetti:
@staebler Thank you for the thoughtful review!

petr-muller:
/skip

`ci/prow/e2e-crc` that job is likely doomed and not needed for merge, let's stop the bot going all crazy on it

pkg/asset/manifests/openstack/cloudproviderconfig.go
```diff
@@ -75,3 +71,35 @@ func CloudProviderConfigSecret(cloud *clientconfig.Cloud) ([]byte, error) {
...
+`
+	if regionName := cloudConfig.RegionName; regionName != "" {
+		cloudProviderConfigData += "region = " + regionName + "\n"
+	}
+
+	if caCertFile := cloudConfig.CACertFile; caCertFile != "" {
+		cloudProviderConfigData += "ca-file = /etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem\n"
+		caFile, err := ioutil.ReadFile(caCertFile)
+		if err != nil {
+			return "", "", Error{err, "failed to read clouds.yaml ca-cert from disk"}
```
mdbooth, line 88: Suggestion:

```
return "", "", fmt.Errorf("failed to read clouds.yaml ca-cert from disk: %w", err)
```

I believe this also implements Unwrap?

pierreprinetti: Nice, I had completely forgot about that thing. Thank you! This is definitely going in the next revision

staebler: Count me as confused. How is using `fmt.Errorf` acceptable when `errors.Wrap` is not? They both satisfy your rationale (1). Neither of them satisfy your rationale (3). And `errors.Wrap` is used extensively throughout the installer codebase, whereas `fmt.Errorf` is not.

pierreprinetti: @mdbooth’s solution has the interesting property of addressing your concern (type maintenance burden), all while not breaking my most cherished point 2 (not an external dependency). I can give up 3 if nobody cares about the “sentinel” pattern :)

PR #4456, opened 2020-12-04
baremetal: Add cluster admin kubeconfig into a secret

kirankt:
The baremetal platform is pivoting away from using pointer ignition config. It will be using a combination of full rendered ignition config fetched from the MCS (to deploy masters) [1][2]  and the use of the rendered machineconfig objects for adding workers [3]. 

This PR saves the admin kubeconfig into a machineconfig asset. This is required for the baremetal machine controller to save the contents of the admin kubeconfig on the workers.

[1] https://github.com/openshift/installer/pull/4427
[2] https://github.com/openshift/installer/pull/4413
[3] https://github.com/openshift/cluster-api-provider-baremetal/pull/127


hardys:
Some additional context - it seems that the kubeconfig file is appended to the MCS config ref https://github.com/openshift/machine-config-operator/blob/master/pkg/server/cluster_server.go#L116

This means it's missing when we consume the rendered config via the merged MachineConfig object in openshift/cluster-api-provider-baremetal#127 because that doesn't access via the MCS (due to the network rules discussed previously ref https://github.com/openshift/machine-config-operator/issues/1690)

pkg/asset/machines/worker.go
```diff
@@ -173,6 +174,7 @@ func (w *Worker) Dependencies() []asset.Asset {
 		// it is put in the dependencies but not fetched in Generate
 		&installconfig.PlatformCredsCheck{},
 		&installconfig.InstallConfig{},
+		&kubeconfig.AdminClient{},
```
staebler, line 177: This is going to have a side-effect of creating the kube certs during manifest generation. I don't know that that is **necessarily** a problem. But if I recall correctly, we've tried to delay the cert creation until the ignition step so that the user does not accidentally use expired or soon-to-be expired certs from a `create manifests` done previously.

staebler (commented):
Are we OK that anyone who has RBAC permissions to get MachineConfigs can get the admin kubeconfig?

pkg/asset/machines/worker.go
kirankt: +1.

I am also in agreement with you and wary about the admin cert/key in the machineconfigs. This seems like using a sledgehammer on a tiny nail not to mention super insecure.

I was wondering if there is an alternative we should pursue. Basically the MCO outputs two different rendered configs. Once by the Machine config server (mcs), which includes contents of /etc/kubernetes/kubeconfig and the other is the machineconfig object which omits this info. Since we cannot access the mcs (blocked on purpose within the pod network), we wanted to rely on the machineconfig objects, which is not complete.

kirankt: I think I'm going to rework this to send the kubeconfig as a secret to the operator-machine-api namespace.

hardys:
This seems reasonable, but I'd like feedback from the MCO team to understand why the MCO generates the kubeconfig on the fly, and loads the cert/secret from disk ref https://github.com/openshift/machine-config-operator/blob/master/pkg/server/cluster_server.go#L116

I'm not clear if that's a historical thing or a deliberate choice, e.g to ensure the admin kubeconfig and related secret isn't accessible directly via the k8s API?

The other alternative would be to generate the kubeconfig in the cluster-api-provider-baremetal similar to the MCS, but I'm not clear if the following cert/secret is sufficient for that?

```
$ oc get pod machine-api-controllers-c85c69c75-45nnx -n openshift-machine-api -o json | jq '.spec.containers[1]|.name,.volumeMounts' | more
"machine-controller"
[
  {
    "mountPath": "/etc/pki/ca-trust/extracted/pem",
    "name": "trusted-ca",
    "readOnly": true
  },
  {
    "mountPath": "/var/run/secrets/kubernetes.io/serviceaccount",
    "name": "machine-api-controllers-token-gqzjr",
    "readOnly": true
  }
]
```

I don't see the `/var/run/secrets/kubernetes.io/serviceaccount` mount in the MAO code, does anyone know where that comes from, and if it's equivalent to the secret used in the MCS?

All that said, I'm fine with solving this in the installer, provided we can get sufficient feedback re the security aspects.

kirankt:
> This seems reasonable, but I'd like feedback from the MCO team to understand why the MCO generates the kubeconfig on the fly, and loads the cert/secret from disk ref https://github.com/openshift/machine-config-operator/blob/master/pkg/server/cluster_server.go#L116
> 
> I'm not clear if that's a historical thing or a deliberate choice, e.g to ensure the admin kubeconfig and related secret isn't accessible directly via the k8s API?
> 
> The other alternative would be to generate the kubeconfig in the cluster-api-provider-baremetal similar to the MCS, but I'm not clear if the following cert/secret is sufficient for that?
> 
> ```
> $ oc get pod machine-api-controllers-c85c69c75-45nnx -n openshift-machine-api -o json | jq '.spec.containers[1]|.name,.volumeMounts' | more
> "machine-controller"
> [
>   {
>     "mountPath": "/etc/pki/ca-trust/extracted/pem",
>     "name": "trusted-ca",
>     "readOnly": true
>   },
>   {
>     "mountPath": "/var/run/secrets/kubernetes.io/serviceaccount",
>     "name": "machine-api-controllers-token-gqzjr",
>     "readOnly": true
>   }
> ]
> ```
> 
> I don't see the `/var/run/secrets/kubernetes.io/serviceaccount` mount in the MAO code, does anyone know where that comes from, and if it's equivalent to the secret used in the MCS?
> 
> All that said, I'm fine with solving this in the installer, provided we can get sufficient feedback re the security aspects.

I did a little bit of digging and noticed that the contents in MCO comes from a secret in the opeshift-machine-config-operator namespace called node-bootstrapper-secret, which has the contents of the kublet's kubeconfig data.

I surmise that the decision made by MCO (to defer adding the contents of the secret in the MCS) is that this kubeconfig data is secure and only needs to exist in the master nodes, where the MCS daemonset runs. This also explains why there was an explicit firewall rule in the workers nodes to block access from the pods to MCS.

In any case, this PR now creates a similar secret asset in the openshift-machine-api namespace, which the baremetal controller now [1] can read and append to the rendered machineconfig much like what MCS does...

[1] https://github.com/openshift/cluster-api-provider-baremetal/pull/127

kirankt:
@staebler . Can you PTAL at this now? I've made the kubeconfig a secret in the openshift-machine-api namespace, which we can access from the baremetal machine-controller.

staebler:
> I did a little bit of digging and noticed that the contents in MCO comes from a secret in the `opeshift-machine-config-operator` namespace called `node-bootstrapper-token`, which has the contents of the kublet's kubeconfig data.

This secret is managed by kubernetes and is associated with the `node-bootstrapper` service account. Can the machine-api-controller generate a kubeconfig using a service account token in the same way? I am very leery of laying down the admin kubeconfig in a secret. It is also not clear to me yet why machines need the admin kubeconfig secret at all--particular worker machines.

staebler:
> > I did a little bit of digging and noticed that the contents in MCO comes from a secret in the `opeshift-machine-config-operator` namespace called `node-bootstrapper-token`, which has the contents of the kublet's kubeconfig data.
> 
> This secret is managed by kubernetes and is associated with the `node-bootstrapper` service account. Can the machine-api-controller generate a kubeconfig using a service account token in the same way? I am very leery of laying down the admin kubeconfig in a secret. It is also not clear to me yet why machines need the admin kubeconfig secret at all--particular worker machines.

Looking at https://github.com/openshift/cluster-api-provider-baremetal/pull/127/commits/5391f7e061192463f7c49b49c3ca35468d8da11a#diff-2635a4094a1b9403011c9044be0b7adff3ed8d6bcec5cb76f215e95f33e04699R605, it seems to me that the kubeconfig is going to be used by the kubelet on the machines. This should almost certainly **not** be the admin kubeconfig.

kirankt:
> > I did a little bit of digging and noticed that the contents in MCO comes from a secret in the `opeshift-machine-config-operator` namespace called `node-bootstrapper-token`, which has the contents of the kublet's kubeconfig data.
> 
> This secret is managed by kubernetes and is associated with the `node-bootstrapper` service account. Can the machine-api-controller generate a kubeconfig using a service account token in the same way? I am very leery of laying down the admin kubeconfig in a secret. It is also not clear to me yet why machines need the admin kubeconfig secret at all--particular worker machines.

Hi @staebler . Yes. This secret is the kubelet secret *NOT* the admin secret as originally planned. Sorry, if I wasn't clear. I believe the node-bootstrapper-token one lives in the openshift-machine-controller-operator namespace and is the exact same thing as this kubelet kubeconfig which this PR creates in the openshift-machine-api namespace. Thanks.

staebler:
> > > I did a little bit of digging and noticed that the contents in MCO comes from a secret in the `opeshift-machine-config-operator` namespace called `node-bootstrapper-token`, which has the contents of the kublet's kubeconfig data.
> > 
> > 
> > This secret is managed by kubernetes and is associated with the `node-bootstrapper` service account. Can the machine-api-controller generate a kubeconfig using a service account token in the same way? I am very leery of laying down the admin kubeconfig in a secret. It is also not clear to me yet why machines need the admin kubeconfig secret at all--particular worker machines.
> 
> Hi @staebler . Yes. This secret is the kubelet secret _NOT_ the admin secret as originally planned. Sorry, if I wasn't clear. I believe the node-bootstrapper-token one lives in the openshift-machine-controller-operator namespace and is the exact same thing as this kubelet kubeconfig which this PR creates in the openshift-machine-api namespace. Thanks.
> 
> Here is the proof: smile
> 
> https://github.com/openshift/installer/blob/6d3609f79c9168335ba376b16a0a3bd68e086acf/pkg/asset/manifests/openshift.go#L81

Oh yeah, I missed that it was using the kubelet kubeconfig. It is still not clear to me why this needs to be laid down by the installer instead of handling with a service account token by the machine-api-operator.

kirankt:
Closing this PR as we need explore iso-based userdata installs.

PR #4067, opened 2020-08-18
Bug 1813012: Remove unused etcd discovery domain

retroflexer:
etcd has not been using DNS entries since 4.4 and EtcdDiscoveryDomain has been a redundant flag. This PR attempts to get rid of the unused flag.

This is an attempt to get https://github.com/openshift/installer/pull/4024 split into smaller PRs to understand the build errors.

PR #4465, opened 2020-12-09
Bug 1901034: manifests: restore setting of etcdDiscoveryDomain in infrastructure

staebler:
With https://github.com/openshift/installer/pull/4067, the etcdDiscoveryDomain field is no longer set in the infrastructure.config.openshift.io manifest. However, the cluster-network-operator relies on that field when syncing the status of proxy.config.openshift.io [1].

[1] https://github.com/openshift/cluster-network-operator/blob/c23495cf6e6ffeffc0290c85ee4608102f7b47d1/pkg/util/proxyconfig/no_proxy.go#L113

### DIFF
--- a/pkg/asset/manifests/infrastructure.go
+++ b/pkg/asset/manifests/infrastructure.go
@@ -76,6 +76,7 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
 			InfrastructureName:   clusterID.InfraID,
 			APIServerURL:         getAPIServerURL(installConfig.Config),
 			APIServerInternalURL: getInternalAPIServerURL(installConfig.Config),
+			EtcdDiscoveryDomain:  getEtcdDiscoveryDomain(installConfig.Config),
 			PlatformStatus:       &configv1.PlatformStatus{},
 		},
 	}
--- a/pkg/asset/manifests/utils.go
+++ b/pkg/asset/manifests/utils.go
@@ -40,3 +40,7 @@ func getAPIServerURL(ic *types.InstallConfig) string {
 func getInternalAPIServerURL(ic *types.InstallConfig) string {
 	return fmt.Sprintf("https://api-int.%s:6443", ic.ClusterDomain())
 }
+
+func getEtcdDiscoveryDomain(ic *types.InstallConfig) string {
+	return ic.ClusterDomain()
+}


staebler:
@retroflexer Do you have any qualms about adding this back?

staebler:
In the e2e-proxy run, the proxy.config.openshift.io resource is keeping the correct value for the `status.noProxy` field now. Notice that the etcd-* entries are using the FQDN.
```
    noProxy: .cluster.local,.svc,.us-east-2.compute.internal,10.0.0.0/16,10.128.0.0/14,127.0.0.1,169.254.169.254,172.30.0.0/16,api-int.ci-op-407xfgbt-bb067.origin-ci-int-aws.dev.rhcloud.com,etcd-0.ci-op-407xfgbt-bb067.origin-ci-int-aws.dev.rhcloud.com,etcd-1.ci-op-407xfgbt-bb067.origin-ci-int-aws.dev.rhcloud.com,etcd-2.ci-op-407xfgbt-bb067.origin-ci-int-aws.dev.rhcloud.com,localhost
```

retroflexer:
Please don't restore setting the etcdDiscoveryDomain back. The etcd component doesn't use DNS anymore and, therefore should not require etcdDiscoveryDomain at all.

If you look at the DNS cluster config it includes the basedomain for the cluster, https://github.com/openshift/api/blob/e9036946d01a899fa1eacaddaf79310f47dff04f/config/v1/types_dns.go#L24-L31

retroflexer:
MCO also was using etcdDiscoveryDomain for getting the value of basedomain, but they changed it to use the basedomain, which is a better approach to getting the domain name. I suggest you modify your proxy to get the domain name in the similar manner (it should not be called etcdHost :-)) 

https://github.com/openshift/machine-config-operator/pull/2025

staebler:
> Please don't restore setting the etcdDiscoveryDomain back. The etcd component doesn't use DNS anymore and, therefore should not require etcdDiscoveryDomain at all.
> 
> If you look at the DNS cluster config it includes the basedomain for the cluster, https://github.com/openshift/api/blob/e9036946d01a899fa1eacaddaf79310f47dff04f/config/v1/types_dns.go#L24-L31

Closing in favor of https://github.com/openshift/cluster-network-operator/pull/910.

/close

PR #4469, opened 2020-12-10
Bug 1902059: Bound SA keys: add default key in case user did not supply it

stlaz:
/cc @sttts @tnozicka


abhinavdahiya:
I would personally like to see this key being generated in the operator when none is provided. Keeping the keys generated in installer to minimum.

stlaz:
The interfaces of the installer appear to be made exactly for this purpose, is there a reason to move the logic to the operator?

stlaz:
/close
closing for now in favour of https://github.com/openshift/cluster-kube-apiserver-operator/pull/1019 which generates the keypair inside the render command

PR #4470, opened 2020-12-10
Bug 1906459: openstack: fix quota checks when they're unlimited

EmilienM:
In OpenStack, if a quota resource is limited to "-1", it means that it's
unlimited. It's the case of a lot of OpenStack clouds, were quotas
aren't defined for all projects and by default "-1" is used.

This patch will ignore a quota record if its limit is set to "-1".

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1906459

### DIFF
--- a/pkg/quota/quota.go
+++ b/pkg/quota/quota.go
@@ -15,6 +15,8 @@ type Quota struct {
 
 	InUse int64
 	Limit int64
+
+	Unlimited bool
 }
 
 // Constraint defines a check against availablity
@@ -75,6 +77,13 @@ func Check(quotas []Quota, checks []Constraint) ([]ConstraintReport, error) {
 		if !ok {
 			report.Result = Unknown
 			report.Message = "No matching quota found for the constraint"
+			reports = append(reports, report)
+			continue
+		}
+
+		if matched.Unlimited {
+			report.Result = Available
+			report.Message = "Unlimited quota found for the constraint"
 			reports = append(reports, report)
 			continue
 		}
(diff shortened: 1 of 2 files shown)


mdbooth (commented):
I've tested and it fixes the issue. Is there unit test coverage somewhere we can add this case to?

mandre (commented):
/lgtm
Thanks for the quick fix.

EmilienM:
> I've tested and it fixes the issue. Is there unit test coverage somewhere we can add this case to?

no unit tests for cloudinfo nor quotas but I agree it would be nice to have at some point.

mandre:
/assign @staebler 
We need your approval for the [pkg/quota/quota.go](https://github.com/openshift/installer/pull/4470/files#diff-432be77457b307682c5eca44fd31cf75a477571ed2913b35d62e8e08392f702f) change, PTAL.

pkg/quota/quota.go
```diff
@@ -79,6 +79,14 @@ func Check(quotas []Quota, checks []Constraint) ([]ConstraintReport, error) {
 			continue
 		}
 
+		// -1 means unlimited in OpenStack
+		if matched.Limit == -1 {
```
staebler, line 83: Let's not carry forward the smell of using -1 to mean unlimited into code that we control. Instead, add a field to `Quota` that will indicate whether the quota is unlimited.

EmilienM: yeah I thought about it and wasn't sure what direction would be the best. I'll do that instead, thanks for your feedback!

staebler (commented):
/approve

PR #4474, opened 2020-12-10
Bug 1906517: OpenStack: Skip collecting info for empty subnet IDs

mandre:
Some clouds may return a generic 404 error instead of an
`ErrResourceNotFound` error, causing the installer to fail with:

    FATAL failed to fetch Install Config: failed to generate asset "Install Config": failed to generate OpenStack cloud info: failed to fetch machine subnet info: Resource not found

We need to not gather subnet information unless a machine subnet was
provided.

### DIFF
--- a/pkg/asset/installconfig/openstack/validation/cloudinfo.go
+++ b/pkg/asset/installconfig/openstack/validation/cloudinfo.go
@@ -186,6 +186,9 @@ func (ci *CloudInfo) collectInfo(ic *types.InstallConfig, opts *clientconfig.Cli
 }
 
 func (ci *CloudInfo) getSubnet(subnetID string) (*subnets.Subnet, error) {
+	if subnetID == "" {
+		return nil, nil
+	}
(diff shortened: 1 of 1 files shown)


pierreprinetti (approved):
What do you think about checking for the zero value one level down the stack, on line 156?

```Go
if ic.OpenStack.MachinesSubnet != "" {
	ci.MachinesSubnet, err = ci.getSubnet(ic.OpenStack.MachinesSubnet)
	if err != nil {
		return errors.Wrap(err, "failed to fetch machine subnet info")
	}
}
```

It would make it more explicit that you're leaving `ci.MachinesSubnet` unchanged, IMHO.

However, that's just a nit, and LGTM otherwise

mandre:
> What do you think about checking for the zero value one level down the stack, on line 156?
> It would make it more explicit that you're leaving `ci.MachinesSubnet` unchanged, IMHO.

I considered it, but I have a slight preference for returning early on invalid values rather than having this logic implemented in the calling function. Plus we're already using this pattern in `ci.getFloatingIP()` so why not being consistent.

/hold cancel

PR #4477, opened 2020-12-11
Bug 1890228: pkg/destroy/aws: Pass destroy if HostedZone does not exist

mtnbikenc:
HostedZone may already be removed and causes destroy to get stuck.


staebler (commented):
/lgtm
/approve

PR #4480, opened 2020-12-14
Bug 1907628: Consistently Set Primary Subnet OpenStack

iamemilio:
The primary subnet was only set in the machine set when the user was
using the bring your own network feature, and this inconsistent usage
was confusing. This changes things so that it's always set.

/label platform/openstack

### DIFF
--- a/pkg/asset/machines/openstack/machines.go
+++ b/pkg/asset/machines/openstack/machines.go
@@ -140,6 +140,11 @@ func generateProvider(clusterID string, platform *openstack.Platform, mpool *ope
 		})
 	}
 
+	primarySubnet := platform.MachinesSubnet
+	if primarySubnet == "" {
+		primarySubnet = fmt.Sprintf("%s-openshift", clusterID)
+	}
+
 	spec := openstackprovider.OpenstackProviderSpec{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: openstackprovider.SchemeGroupVersion.String(),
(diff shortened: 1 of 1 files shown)


pkg/asset/machines/openstack/machines.go
```diff
@@ -140,6 +140,11 @@ func generateProvider(clusterID string, platform *openstack.Platform, mpool *ope
 		})
 	}
 
+	primarySubnet := platform.MachinesSubnet
+	if primarySubnet == "" {
+		primarySubnet = fmt.Sprintf("%s-openshift", clusterID)
```
mandre, line 145: IIRC, `platform.MachinesSubnet` is the ID of the machine subnet, so using a name here would make `PrimarySubnet` attribute inconsistent.
Can we retrieve the subnet ID and pass that to the provider spec?

iamemilio: Ok, i just checked. Actually This is all broken, and we cant do it this way :|

iamemilio: https://github.com/openshift/cluster-api-provider-openstack/blob/7a86c52029dae191c1aeff96bd18208069172e44/pkg/cloud/openstack/machine/actuator.go#L527

iamemilio:
Primary Subnet cannot be set this way in the installer because the Machine Network has not been created by the installer yet, so its ID cant be referenced. Instead, we should enforce usage of primary subnet the same way we use machinesSubnet in the installer: 
as an override to identify which network the kubernetes endpoints get published on when installing with multiple networks. If there is only one network that is being installed with, it is implicitly the primary network and does not need to be indicated.

PR #4481, opened 2020-12-14
openstack: add networking quota checks

EmilienM:
We now check the quotas for:

* Ports, with a base number + growing for each machine and its number of
attached networks.
* Routers
* Subnets
* Networks
* Security Groups
* Security Group Rules

We make the difference between requirements when Kuryr is the networking
backend or not, since the quotas aren't the same.

Note: this PR also includes vendoring for both gophercloud and
      terraform-openstack, which is required to get quota details for networking
      resources.

JIRA: [OSASINFRA-2006](https://issues.redhat.com/browse/OSASINFRA-2006)
/label platform/openstack

### DIFF
--- a/pkg/asset/installconfig/openstack/validation/cloudinfo.go
+++ b/pkg/asset/installconfig/openstack/validation/cloudinfo.go
@@ -41,9 +41,10 @@ type CloudInfo struct {
 }
 
 type clients struct {
-	networkClient *gophercloud.ServiceClient
-	computeClient *gophercloud.ServiceClient
-	imageClient   *gophercloud.ServiceClient
+	networkClient  *gophercloud.ServiceClient
+	computeClient  *gophercloud.ServiceClient
+	imageClient    *gophercloud.ServiceClient
+	identityClient *gophercloud.ServiceClient
 }
 
 // Flavor embeds information from the Gophercloud Flavor struct and adds
@@ -90,6 +91,11 @@ func GetCloudInfo(ic *types.InstallConfig) (*CloudInfo, error) {
 	ci.clients.imageClient, err = clientconfig.NewServiceClient("image", opts)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to create an image client")
+	}
+
+	ci.clients.identityClient, err = clientconfig.NewServiceClient("identity", opts)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create an identity client")
 	}
 
 	err = ci.collectInfo(ic, opts)
@@ -175,7 +181,7 @@ func (ci *CloudInfo) collectInfo(ic *types.InstallConfig, opts *clientconfig.Cli
 		return err
 	}
 
-	ci.Quotas, err = loadQuotas(opts)
+	ci.Quotas, err = loadQuotas(ci)
 	if err != nil {
 		if isUnauthorized(err) {
 			logrus.Warnf("Missing permissions to fetch Quotas and therefore will skip checking them: %v", err)
@@ -326,15 +332,15 @@ func (ci *CloudInfo) getZones() ([]string, error) {
 }
 
 // loadLimits loads the consumer quota metric.
-func loadLimits(opts *clientconfig.ClientOpts) ([]record, error) {
+func loadLimits(ci *CloudInfo) ([]record, error) {
 	var limits []record
 
-	projectID, err := getProjectID(opts)
+	projectID, err := getProjectID(ci)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get keystone project ID")
 	}
 
-	computeRecords, err := getComputeLimits(opts, projectID)
+	computeRecords, err := getComputeLimits(ci, projectID)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get compute quota records")
 	}
@@ -342,7 +348,7 @@ func loadLimits(opts *clientconfig.ClientOpts) ([]record, error) {
 		limits = append(limits, r)
 	}
 
-	networkRecords, err := getNetworkLimits(opts, projectID)
+	networkRecords, err := getNetworkLimits(ci, projectID)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get network quota records")
 	}
@@ -353,12 +359,8 @@ func loadLimits(opts *clientconfig.ClientOpts) ([]record, error) {
 	return limits, nil
 }
 
-func getComputeLimits(opts *clientconfig.ClientOpts, projectID string) ([]record, error) {
-	computeClient, err := clientconfig.NewServiceClient("compute", opts)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to connect against OpenStack Compute v2 API")
-	}
-	qs, err := computequotasets.GetDetail(computeClient, projectID).Extract()
+func getComputeLimits(ci *CloudInfo, projectID string) ([]record, error) {
+	qs, err := computequotasets.GetDetail(ci.clients.computeClient, projectID).Extract()
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get QuotaSets from OpenStack Compute API")
 	}
@@ -383,12 +385,8 @@ func getComputeLimits(opts *clientconfig.ClientOpts, projectID string) ([]record
 	return records, nil
 }
 
-func getNetworkLimits(opts *clientconfig.ClientOpts, projectID string) ([]record, error) {
-	networkClient, err := clientconfig.NewServiceClient("network", opts)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to connect against OpenStack Network v2 API")
-	}
-	qs, err := networkquotasets.GetDetail(networkClient, projectID).Extract()
+func getNetworkLimits(ci *CloudInfo, projectID string) ([]record, error) {
+	qs, err := networkquotasets.GetDetail(ci.clients.networkClient, projectID).Extract()
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get QuotaSets from OpenStack Network API")
 	}
@@ -416,12 +414,8 @@ func getNetworkLimits(opts *clientconfig.ClientOpts, projectID string) ([]record
 	return records, nil
 }
 
-func getProjectID(opts *clientconfig.ClientOpts) (string, error) {
-	keystoneClient, err := clientconfig.NewServiceClient("identity", opts)
-	if err != nil {
-		return "", errors.Wrap(err, "failed to conect against OpenStack Keystone API")
-	}
-	authResult := keystoneClient.GetAuthResult()
+func getProjectID(ci *CloudInfo) (string, error) {
+	authResult := ci.clients.identityClient.GetAuthResult()
 	if authResult == nil {
 		return "", errors.Errorf("Client did not use openstack.Authenticate()")
 	}
@@ -449,8 +443,8 @@ func getProjectID(opts *clientconfig.ClientOpts) (string, error) {
 
 // loadQuotas loads the quota information for a project and provided services. It provides information
 // about the usage and limit for each resource quota.
-func loadQuotas(opts *clientconfig.ClientOpts) ([]quota.Quota, error) {
-	records, err := loadLimits(opts)
+func loadQuotas(ci *CloudInfo) ([]quota.Quota, error) {
+	records, err := loadLimits(ci)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to load quota limits")
 	}
--- a/pkg/asset/installconfig/openstack/validation/cloudinfo.go
+++ b/pkg/asset/installconfig/openstack/validation/cloudinfo.go
@@ -10,6 +10,7 @@ import (
 	tokensv3 "github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
 	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
+	networkquotasets "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/quotas"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
 	"github.com/gophercloud/utils/openstack/clientconfig"
@@ -341,13 +342,21 @@ func loadLimits(opts *clientconfig.ClientOpts) ([]record, error) {
 		limits = append(limits, r)
 	}
 
+	networkRecords, err := getNetworkLimits(opts, projectID)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get network quota records")
+	}
+	for _, r := range networkRecords {
+		limits = append(limits, r)
+	}
+
 	return limits, nil
 }
 
 func getComputeLimits(opts *clientconfig.ClientOpts, projectID string) ([]record, error) {
 	computeClient, err := clientconfig.NewServiceClient("compute", opts)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to connect against OpenStack Comute v2 API")
+		return nil, errors.Wrap(err, "failed to connect against OpenStack Compute v2 API")
 	}
 	qs, err := computequotasets.GetDetail(computeClient, projectID).Extract()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/asset/installconfig/openstack/validation/cloudinfo.go
```diff
@@ -374,6 +383,39 @@ func getComputeLimits(opts *clientconfig.ClientOpts, projectID string) ([]record
...
+		return nil, errors.Wrap(err, "failed to connect against OpenStack Network v2 API")
+	}
+	qs, err := networkquotasets.GetDetail(networkClient, projectID).Extract()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get QuotaSets from OpenStack Network API")
+	}
+
+	var records []record
+	addRecord := func(name string, quota networkquotasets.QuotaDetail) {
+		qval := int64(quota.Limit - quota.Used - quota.Reserved)
```
iamemilio, line 398: shouldnt limit - used - reserved always == 0?

pkg/asset/quota/openstack/openstack.go
```diff
@@ -7,10 +7,28 @@ import (
 	openstackprovider "sigs.k8s.io/cluster-api-provider-openstack/pkg/apis/openstackproviderconfig/v1alpha1"
 )
 
+const (
+	// These numbers should reflect what is documented here:
+	// https://github.com/openshift/installer/tree/master/docs/user/openstack
+	// https://github.com/openshift/installer/blob/master/docs/user/openstack/kuryr.md
+	// Number or ports here don't include the constraints needed for each machine, which are calculated later
+	portsConstraintKuryr                  = 15
```
iamemilio, line 15: What is the difference between subnetsConstraintKuryr and subnetsConstraintWithKuryr? This is pretty confusing

iamemilio:
And if you want to make the  function calls less complex, pack the constant values into a struct

pkg/asset/quota/openstack/openstack.go
```diff
@@ -25,14 +43,36 @@ func Constraints(ci *validation.CloudInfo, controlPlanes []machineapi.Machine, c
...
+	ret = append(ret, generateConstraint("SecurityGroup", securityGroup))
+	return append(ret, generateConstraint("SecurityGroupRule", securityGroupRule))
+}
+
 func controlPlane(ci *validation.CloudInfo, machines []*openstackprovider.OpenstackProviderSpec) []quota.Constraint {
 	var ret []quota.Constraint
 	for _, m := range machines {
 		flavorInfo := ci.Flavors[m.Flavor]
 		ret = append(ret, machineFlavorCoresToQuota(flavorInfo, ci), machineFlavorRAMToQuota(flavorInfo, ci))
+		ret = append(ret, generateConstraint("Port", int64(len(m.Networks))))
```
pierreprinetti, line 75: Question (on the previous line 74, really): can it be that if you have 3 ControlPlane Machines defined with 16GB RAM each, then you are setting a "16GB" requirement three times, instead of setting a requirement of 16*3GB?

pkg/asset/quota/openstack/openstack.go
```diff
@@ -47,6 +87,7 @@ func compute(ci *validation.CloudInfo, replicas []int64, machines []*openstackpr
 		ramConstraint := machineFlavorRAMToQuota(flavorInfo, ci)
 		ramConstraint.Count = ramConstraint.Count * replicas[idx]
 		ret = append(ret, instanceConstraint(int64(replicas[idx])), coresConstraint, ramConstraint)
+		ret = append(ret, generateConstraint("Port", int64(len(m.Networks))))
```
pierreprinetti, line 90: On line 89you might be setting N times a requirement for the resources, instead of adding requirements from multiple MachineSets

pierreprinetti (commented):
A couple nits on Network, plus what might be a bug in the previous Compute quota check

pkg/asset/quota/openstack/openstack.go
pierreprinetti: Can it be that `portsConstraintKuryr` is _without Kuryr_ while `portsConstraintWithKuryr` is _with_ Kuryr? :laughing:

pkg/asset/installconfig/openstack/validation/cloudinfo.go
```diff
@@ -374,6 +383,39 @@ func getComputeLimits(opts *clientconfig.ClientOpts, projectID string) ([]record
 	return records, nil
 }
 
+func getNetworkLimits(opts *clientconfig.ClientOpts, projectID string) ([]record, error) {
+	networkClient, err := clientconfig.NewServiceClient("network", opts)
```
mdbooth, line 387: Note that this API hits keystone every time. If you get an AuthenticatedClient and use it to instantiate a service client it doesn't have to hit keystone. That's something we should look at generally, though.

mdbooth (commented):
I haven't finished reviewing, but if I don't hit send I'll risk forgetting to!

pkg/asset/quota/openstack/openstack.go
mdbooth: I think this is the safest thing to do here as there are 3 separate Machine profiles. It may seem safe to assume that they're all the same, but the performance penalty of handling them separately is inconsequential. This is simple and avoids an assumption.

pkg/asset/quota/openstack/openstack.go
mdbooth: I think it's slightly clearer in my refactor #4483 how these are added up. These are run through aggregate() below before being returned.

pkg/asset/quota/openstack/openstack.go
EmilienM: right, I was being asked to renamed it. Happy to rename it again if needed.

pkg/asset/installconfig/openstack/validation/cloudinfo.go
```diff
@@ -341,13 +342,21 @@ func loadLimits(opts *clientconfig.ClientOpts) ([]record, error) {
...
+	}
+
 	return limits, nil
 }
 
 func getComputeLimits(opts *clientconfig.ClientOpts, projectID string) ([]record, error) {
 	computeClient, err := clientconfig.NewServiceClient("compute", opts)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to connect against OpenStack Comute v2 API")
+		return nil, errors.Wrap(err, "failed to connect against OpenStack Compute v2 API")
```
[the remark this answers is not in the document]
EmilienM: > And if you want to make the function calls less complex, pack the constant values into a struct

I thought about it and you're right, I'll do it.

pkg/asset/installconfig/openstack/validation/cloudinfo.go
EmilienM: Don't we call ```getNetworkLimits``` only once?

mdbooth: Probably, but we hit Keystone every time we create a service client. i.e. We hit it once for Nova and once again for Neutron. If we used AuthenticatedClient (or whatever it's called) we'd only hit it once, and reuse that token for all service clients.

pkg/asset/quota/openstack/openstack.go
EmilienM: sorry it was a typo, I'll fix it.

pkg/asset/quota/openstack/openstack.go
```diff
@@ -25,14 +43,36 @@ func Constraints(ci *validation.CloudInfo, controlPlanes []machineapi.Machine, c
...
+		ret = append(ret, generateNetworkConstraints(int64(portsConstraintWithKuryr), int64(subnetsConstraintWithKuryr), int64(networksConstraintWithKuryr), int64(securityGroupsConstraintWithKuryr), int64(securityGroupRulesConstraintWithKuryr))...)
+	} else {
+		ret = append(ret, generateNetworkConstraints(int64(portsConstraintKuryr), int64(subnetsConstraintKuryr), int64(networksConstraintKuryr), int64(securityGroupsConstraintKuryr), int64(securityGroupRulesConstraintKuryr))...)
+	}
+	return ret
+}
+
+func generateNetworkConstraints(port int64, subnet int64, network int64, securityGroup int64, securityGroupRule int64) []quota.Constraint {
+	var ret []quota.Constraint
+	ret = append(ret, generateConstraint("Port", port))
```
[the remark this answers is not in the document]
EmilienM: done, thanks a lot for this suggestion!

pkg/asset/quota/openstack/openstack.go
```diff
@@ -25,14 +43,36 @@ func Constraints(ci *validation.CloudInfo, controlPlanes []machineapi.Machine, c
...
+	ret = append(ret, generateConstraint("Subnet", subnet))
+	ret = append(ret, generateConstraint("Network", network))
+	ret = append(ret, generateConstraint("SecurityGroup", securityGroup))
+	return append(ret, generateConstraint("SecurityGroupRule", securityGroupRule))
```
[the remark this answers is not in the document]
EmilienM: done thanks!

pkg/asset/quota/openstack/openstack.go
EmilienM: I agree with @mdbooth and it was my reasoning when we implementing this.

pkg/asset/installconfig/openstack/validation/cloudinfo.go
EmilienM: I agree with you. However I was told to use ``NewServiceClient`` because it's an abstraction from the actual client function, e.g. ```openstack.NewNetworkV2```.
By using ``clientconfig.NewServiceClient("network", opts)``, we don't care if the default for Network client switches to NewNetworkV3, it'll be transparent for us.

I'll defer to @mandre on this one, I'm happy to call ```openstack.NewNetworkV2``` (and change compute as well).

pkg/asset/quota/openstack/openstack.go
EmilienM: > On line 89 you might be setting N times a requirement for the resources, instead of adding up requirements from multiple MachineSets

nice catch, thanks.

EmilienM: > I think it's slightly clearer in my refactor #4483 how these are added up. These are run through aggregate() below before being returned.

ack, happy to rebase if yours merges before mine!

pkg/asset/installconfig/openstack/validation/cloudinfo.go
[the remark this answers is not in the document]
mdbooth: Don't do it! 🤣

It's fine. Liberating, even!

pkg/asset/quota/openstack/openstack.go
```diff
@@ -25,14 +41,37 @@ func Constraints(ci *validation.CloudInfo, controlPlanes []machineapi.Machine, c
 	var ret []quota.Constraint
 	ret = controlPlane(ci, ctrplConfigs)
 	ret = append(ret, compute(ci, computeReplicas, computeConfigs)...)
+	ret = append(ret, getNetworkConstraints(networkType)...)
 	return aggregate(ret)
 }
 
+func getNetworkConstraints(networkType string) (ret []quota.Constraint) {
```
pierreprinetti, line 48: nit: you don't need a named return value because you declare it again in the `return` statements

pkg/asset/quota/openstack/openstack.go
pierreprinetti: I agree with the no-assumptions rationale. My concern was that we weren't adding up the RAM, disk, and vCPU requirements (I am commenting code that is not touched in this PR).

pkg/asset/installconfig/openstack/validation/cloudinfo.go
mandre: Agreed we should re-use the same `AuthenticatedClient` for all service clients. This is however out of scope for this change.

For now please make `loadQuotas()` a method of `CloudInfo` and use `ci.clients.networkClient`.

pkg/asset/quota/openstack/openstack.go
EmilienM: I'll fix that, yeah it's a leftover. Thanks!

pkg/asset/installconfig/openstack/validation/cloudinfo.go
EmilienM: ack, will close that comment then and we'll fix it separately.

EmilienM: and ack, I'll make `loadQuotas()` a method of `CloudInfo` and use `ci.clients.networkClient`.

EmilienM: > Agreed we should re-use the same `AuthenticatedClient` for all service clients. This is however out of scope for this change.
> 
> For now please make `loadQuotas()` a method of `CloudInfo` and use `ci.clients.networkClient`.

Please take a look at https://github.com/gophercloud/utils/blob/master/openstack/clientconfig/requests.go#L748-L851
I actually thing it's good to use `NewServiceClient()` since there is a bunch of logic for variables and SSL stuff.

pkg/asset/quota/openstack/openstack.go
EmilienM: @pierreprinetti to follow-up on our conversation, today I confirmed that the existing code here works fine.
I reduced my quotas on purpose to see what error message I would get when trying to deploy 3 master with a flavor of 4 Cores and 16GB of RAM + 3 workers with a flavor of 4 Cores and 8GB of RAM:

```FATAL failed to fetch Cluster: failed to fetch dependency of "Cluster": failed to generate asset "Platform Quota Check": error(MissingQuota): Cores is not available because the required number of resources (24) is more than the limit of 2```

The required number of resources is right (3x4 + 3x4) for the Cores.
Same for RAM:

```FATAL failed to fetch Cluster: failed to fetch dependency of "Cluster": failed to generate asset "Platform Quota Check": error(MissingQuota): RAM is not available because the required number of resources (73728) is more than the limit of 2```

3 x 16 + 3 x 8 = 72GB

Comment (pierreprinetti) [reply pkg/asset/quota/openstack/openstack.go:74]: Perfect! Thanks

**Comment by pierreprinetti** [comment]
/approve
/lgtm
:rocket:

File: pkg/asset/installconfig/openstack/validation/cloudinfo.go (hunk 5 above)
Comment (mandre) [reply pkg/asset/installconfig/openstack/validation/cloudinfo.go:390]: We're using `NewServiceClient()` to allocate `ci.clients.networkClient`:
https://github.com/openshift/installer/blob/338310d4f7a3383054cb12c9dd19df258c313640/pkg/asset/installconfig/openstack/validation/cloudinfo.go#L79
So it's there to use freely if we make `loadQuotas()`, `getNetworkLimits()`, etc,  methods of `CloudInfo`.

Comment (EmilienM) [reply pkg/asset/installconfig/openstack/validation/cloudinfo.go:390]: ok I see what you mean now. I'll take a look and update accordingly.

Comment (EmilienM) [reply pkg/asset/installconfig/openstack/validation/cloudinfo.go]: done, and I fixed computed too, in another commit!

File: pkg/asset/quota/openstack/openstack.go [hunk 10]
```diff
@@ -25,14 +41,37 @@ func Constraints(ci *validation.CloudInfo, controlPlanes []machineapi.Machine, c
...
+		}
+	}
+	return []quota.Constraint{
+		{Name: "Port", Count: networkConstraints.ports},
+		{Name: "Router", Count: networkConstraints.routers},
+		{Name: "Subnet", Count: networkConstraints.subnets},
+		{Name: "Network", Count: networkConstraints.networks},
+		{Name: "SecurityGroup", Count: networkConstraints.securityGroups},
+		{Name: "SecurityGroupRule", Count: networkConstraints.securityGroupRules},
+	}
```
staebler, line 66: nit: I think this makes it a bit easier to read and a bit more maintainable as you do not need to copy the translation logic.
```suggestion
	nc := networkConstraints
	if networkType == "kuryr" {
		nc = networkConstraintsWithKuryr
	}
	return []quota.Constraint{
		{Name: "Port", Count: nc.ports},
		{Name: "Router", Count: nc.routers},
		{Name: "Subnet", Count: nc.subnets},
		{Name: "Network", Count: nc.networks},
		{Name: "SecurityGroup", Count: nc.securityGroups},
		{Name: "SecurityGroupRule", Count: nc.securityGroupRules},
	}
```

Although, it is not immediately clear to me why we need the `networkConstraint` type at all. Why not just have the type of the `networkConstraints` and `networkConstraintsWithKuryr` variables be `quota.Constraint`. You could have a function that takes in as parameters the six fields of `networkConstraint` and returns a `quota.Constraint`.

```
var networkConstraint = buildNetworkConstraint(15, 1, 1, 1, 3, 60)
var networkConstraintWithKuryr = buildNetworkConstraint(1500, 1, 250, 250, 250, 1000)

func buildNetworkConstraint(ports, routers, subnets, networks, securityGroups, securityGroupRules int64) quota.Constraint {
	return []quota.Constraint{
		{Name: "Port", Count: ports},
		{Name: "Router", Count: routers},
		{Name: "Subnet", Count: subnets},
		{Name: "Network", Count: networks},
		{Name: "SecurityGroup", Count: securityGroups},
		{Name: "SecurityGroupRule", Count: securityGroupRules},
	}
}

func getNetworkConstraints(networkType string) []quota.Constraint {
	if networkType == "kuryr" {
		return []quota.Constraint{networkConstraintsWithKuryr}
	}
	return []quota.Constraint{networkConstraints}
}
```

staebler (commented):
/approve

pkg/asset/quota/openstack/openstack.go
EmilienM: I'll apply your last suggestion, without the type. Thanks!

EmilienM:
For the record, we hit this: https://bugs.launchpad.net/neutron/+bug/1918565

pierreprinetti:
> For the record, we hit this: https://bugs.launchpad.net/neutron/+bug/1918565

(internally referenced as https://bugzilla.redhat.com/show_bug.cgi?id=1937460 )

PR #4482, opened 2020-12-15
Single Node deployment with bootstrap-in-place

eranco74:
## single-node deployment with bootstrap-in-place
Add new `create single-node-ignition-config` command to the installer to create
 `bootstrap-in-place-for-live-iso.ign` Ignition config.
This new target will not output master.ign and worker.ign.

This Ignition config will have a different `bootkube.sh` from the
default bootstrap Ignition. In addition to the standard rendering
logic, the modified script will:

1. Start `cluster-bootstrap` without required pods by setting `--required-pods=''`
2. Run `cluster-bootstrap` with the `--bootstrap-in-place` option.
3. Fetch the master Ignition and combine it with the original Ignition
   config, the control plane static pod manifests, the required
   kubernetes resources, and the bootstrap etcd database snapshot to
   create a new Ignition config for the host.
4. At this point bootkube finish and a new service named wrrite-to-disk start.
   This service will execute `coreos-installer install` to write the rhcos image
   and the rendered master Ignition to disk and reboot the node.

## Log gathering on bootstrap-in-place.
Support collecting logs from bootstrap-in-place cluster with, for example, `openshift-install gather bootstrap --bootstrap 192.168.126.10 --master 192.168.126.10 --key id_rsa`
### Pre-pivot - gather bootstrap only
The command works as usual before the pivot occurs because the `/usr/local/bin/installer-gather.sh` script is present and works as expected without any changes.
### Gathering bootstrap logs
Before reboot, bootstrap will gather from itself using `/usr/local/bin/installer-gather.sh`. This script also tries to gather from the masters in the cluster, but since there are none, it will only gather about itself.

After the gathering is complete, the bundle is added to the master ignition - thus making the bootstrap logs available from the master after reboot.

### Gathering master logs
Typically, in non-BiP scenarios, masters logs are gathered using the bootstrap node as a proxy by remotely running, via `ssh`, the `/usr/local/bin/installer-gather.sh` script present on the bootstrap node. That script in turn detects all masters (or receives a list of masters via the `--master` command-line arguments), then for each master it `scp`s a second script called `/usr/local/bin/installer-masters-gather.sh` to that master. It then runs that script remotely, on the master, using `ssh`. When the script is finished running, it `scp`s the resulting files back to the bootstrap node, adding them to the bootstrap log bundle, with each master logs appearing in its own named-folder inside the `control-plane` directory of the bootstrap node log bundle.

In the BiP scenario, we created a new script called `installer-master-bootstrap-in-place-gather.sh`. This script is copied to the master (using ignition) to the same location where the bootstrap node usually has `installer-gather.sh`. i.e. this script masquerades as `installer-gather.sh`. We also copy the `/usr/local/bin/installer-masters-gather.sh`.

The `installer-master-bootstrap-in-place-gather.sh`, masquerading as `installer-gather.sh`, gets remotely called by the `openshift-install gather` command that believes its collecting logs from a bootstrap node. The script however behaves slightly differently - instead of collecting bootstrap logs and then remotely running `/usr/local/bin/installer-masters-gather.sh` on all master nodes, it collects the bootstrap logs from the bundle copied to via the master-ignition, and it collect master logs by running `/usr/local/bin/installer-masters-gather.sh` directly on itself instead of remotely on other masters (which don't exist. All master IP addresses passed to it are completely ignored). The final archiving of all the logs into the home directory is done exactly in the same manner as `/usr/local/bin/installer-gather.sh`.

The end result is a log bundle that looks pretty much the same as non-bootstrap-in-place scenarios.

### Small caveat
As you can see - `openshift-install gather bootstrap --bootstrap 192.168.126.10 --master 192.168.126.10 --key id_rsa` - the command requires you to specify at least one master and one bootstrap node. In our case, you just pass the single node IP as both master and bootstrap. This causes a small "bug" when the node is pre-pivot - it collects both bootstrap logs on itself but also "remotely" collects master logs from itself via SSH. This causes a duplication of log files when the gather command is ran pre-pivot.


data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -336,8 +336,64 @@ do
...
+    mv /etc/kubernetes/manifests/etcd-member-pod.yaml /etc/kubernetes
+
+    until ! crictl ps | grep etcd
+    do
+        echo "Waiting for etcd to go down"
+        sleep 10
+    done
+
+    echo "Getting master Ignition from MCS"
+    # Get the master ignition for MCS
```
romfreiman, line 365: from MCS

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -336,8 +336,64 @@ do
...
+    curl -k  -H "Accept:'application/vnd.coreos.ignition+json;version=3.1.0, */*;q=0.1'" \
+    https://localhost:22623/config/master -o /opt/openshift/master.ign
+
+    echo "Updating master Ignition"
```
romfreiman, line 369: update where?

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -336,8 +336,64 @@ do
...
+        bootstrap-in-place --asset-dir=/assets --ignition-path=/assets/master.ign
+
+    touch update-master-ignition.done
+fi
+
+if [ ! -f write-to-disk.done ]
+then
+    echo "Writing RHCOS + master Ignition to disk"
+    #TODO: which disk to use?
+    coreos-installer install --insecure -i /opt/openshift/master.ign /dev/vda
```
romfreiman, line 386: I assume it should be parametrized? Env variable? Otherwise, how would we use it from assisted installer?

romfreiman:
Relevant cluster-bootstrap pr is: https://github.com/openshift/cluster-bootstrap/pull/46

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
romfreiman: To allow AssistedInstaller to pick up the flow, and be able to select disk, inject network customization, and other improvements, I assume we'll need some flag saying --bootstrap-only, meaning the bootstrap phase will finish with master.ign generated and the AI will continue from that point

staebler (commented):
I don't have any major qualms with this approach.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
eranco74: The AssistedInsatller can create these files to skip these operations:
/opt/openshift/write-to-disk.done
/opt/openshift/reboot.done

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -278,6 +279,7 @@ func (a *Bootstrap) getTemplateData(installConfig *types.InstallConfig, releaseI
 		BootImage:             string(*rhcosImage),
 		PlatformData:          platformData,
 		ClusterProfile:        clusterProfile,
+		SingleNode:            *installConfig.ControlPlane.Replicas == 1,
```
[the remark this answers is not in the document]
eranco74: I think not.
We have a few options:

1. In the case of a single node control plane replica the create ignition config command will create a bootstrap-in-place-for-live-iso.ign in addition to the Ignition configs it's generating today.
2. Check environment variable (EXPIRAMENTAL_BOOTSTRAP_INPLACE) in addition to the number of control plane node.
3. Check environment variable + create a new config (bootstrap-in-place-for-live-iso.ign)

hack/after_reboot.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash -x
```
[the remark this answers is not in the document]
eranco74: Sorry, deleted.

sno_manifest.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: machineconfiguration.openshift.io/v1
```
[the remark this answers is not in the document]
eranco74: Sorry about that.
this file should be converted to a manifest that is added in case of bootstrap-in-place.
Updating the PR

pkg/asset/ignition/bootstrap/bootstrap.go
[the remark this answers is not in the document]
eranco74: Added env var

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -345,10 +345,20 @@ then
         --volume "$PWD:/assets:z" \
         --volume /etc/kubernetes:/etc/kubernetes:z \
         "${CLUSTER_BOOTSTRAP_IMAGE}" \
+{{if .BootstrapInPlace}}
+        start --tear-down-early=false --asset-dir=/assets --required-pods=""  --assets-create-timeout=5m
```
romfreiman, line 349: remind me what tear-down-early is about?

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -345,10 +345,20 @@ then
         --volume "$PWD:/assets:z" \
         --volume /etc/kubernetes:/etc/kubernetes:z \
         "${CLUSTER_BOOTSTRAP_IMAGE}" \
+{{if .BootstrapInPlace}}
```
romfreiman, line 348: SingleNodeBootstrapInPlace maybe?

data/data/bootstrap/files/usr/local/bin/bootstrap-in-place.sh
```diff
@@ -0,0 +1,51 @@
...
+        echo "Waiting for etcd to go down"
+        sleep 10
+    done
+
+    echo "Creating master ignition and writing it to disk"
+    # Get the master ignition from MCS
+    curl -k  -H "Accept:'application/vnd.coreos.ignition+json;version=3.1.0, */*;q=0.1'" \
+    https://localhost:22623/config/master -o /opt/openshift/master.ign
+
+    echo "Updating master ignition"
```
romfreiman, line 20: with.... so next person who gonna look into it will understand the magic that is happening

data/data/bootstrap/files/usr/local/bin/bootstrap-in-place.sh
```diff
@@ -0,0 +1,51 @@
...
+    bootkube_podman_run \
+        --rm \
+        --privileged \
+        --volume "/var/lib/etcd:/var/lib/etcd" \
+        --volume "$PWD:/assets:z" \
+        --volume "/etc/kubernetes:/etc/kubernetes" \
+        quay.io/eranco74/cluster-bootstrap:bip \
```
romfreiman, line 27: why isnt it CLUSTER_BOOTSTRAP_IMAGE

data/data/bootstrap/files/usr/local/bin/bootstrap-in-place.sh
```diff
@@ -0,0 +1,51 @@
...
+    #TODO: update to the proper image once it's supported
+
+    touch create-static-pods-data.done
+fi
+
+if [ ! -f coreos-installer.done ]
+then
+    # Write image + ignition to disk
+    echo "Getting installation disk"
+    INSTALL_DISK=$(lsblk | grep disk | awk 'NR==1{print $1}')
```
romfreiman, line 38: is it possible we get an empty value here? For example, if u spawn a vm with live iso without attaching another volume?

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -266,6 +268,19 @@ func (a *Bootstrap) getTemplateData(installConfig *types.InstallConfig, releaseI
 		logrus.Warnf("Found override for Cluster Profile: %q", cp)
 		clusterProfile = cp
 	}
+	bootstrapInPlace := false
+    if bootstrapInPlaceEnv := os.Getenv("OPENSHIFT_INSTALL_EXPERIMENTAL_BOOTSTRAP_IN_PLACE"); bootstrapInPlaceEnv != "" {
```
romfreiman, line 272: aligment?

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
eranco74: tear-down-early will take down the control plane once required pods are running on the master nodes.
Not relevant for the bootstrap-in-place use case

data/data/bootstrap/files/usr/local/bin/bootstrap-in-place.sh
eranco74: I guess it's possible.
I'll add an ENV VAR for allowing the user to set it

crawford (commented):
Please update the scripts to use long flags where available.

data/data/bootstrap/bootstrap-in-place/usr/local/bin/bootstrap-in-place.sh
```diff
@@ -0,0 +1,61 @@
...
+    "${CLUSTER_BOOTSTRAP_IMAGE}" \
+    bootstrap-in-place --asset-dir=/assets --ignition-path=/assets/master.ign
+
+  touch master-ignition.done
+fi
+
+if [ ! -f coreos-installer.done ]; then
+  # Write image + ignition to disk
+  echo "Getting installation disk"
+  INSTALL_DISK="$(lsblk | grep disk | awk 'NR==1{print $1}')"
```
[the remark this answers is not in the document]
eranco74: This is the default, I'll add an option for the user to specify the installation disk in another PR

data/data/bootstrap/bootstrap-in-place/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -0,0 +1,49 @@
...
+        sleep 30
+    done
+}
+function restart_kubelet {
+  echo "Restarting kubelet"
+  while grep  bootstrap-kube-apiserver /etc/kubernetes/manifests/kube-apiserver-pod.yaml ; do
+    echo "Waiting for kube-apiserver to apply the new static pod configuration"
+    sleep 10
+  done
+  systemctl daemon-reload
```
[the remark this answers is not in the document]
eranco74: It's not required.

data/data/bootstrap/bootstrap-in-place/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -0,0 +1,49 @@
...
+  needed_to_approve=false
+  until [ "$(oc get nodes | grep master | grep -v NotReady | grep -c Ready)" -eq 1 ]; do
+      needed_to_approve=true
+      echo "Approving csrs ..."
+     oc get csr -o go-template='{{range .items}}{{if not .status}}{{.metadata.name}}{{"\n"}}{{end}}{{end}}' | xargs oc adm certificate approve &> /dev/null || true
+     sleep 30
+    done
+  # Restart kubelet only if node was added
+  if $needed_to_approve ; then
+    sleep 60
```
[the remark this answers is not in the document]
eranco74: Hmmm, seems like a leftover from the time I didn't wait for the kub-eapiserver.
Removing

data/data/bootstrap/bootstrap-in-place/usr/local/bin/bootstrap-in-place.sh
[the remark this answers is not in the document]
eranco74: Decided to remove the default.
If the user didn't specify the coreos-installer args env we will not execute the coreos-installer command.

cmd/openshift-install/create.go
```diff
@@ -79,6 +79,16 @@ var (
 		},
 		assets: targetassets.IgnitionConfigs,
 	}
+	singleNodeignitionConfigsTarget = target{
+		name: "Single Node Ignition Config",
+		command: &cobra.Command{
+			Use:   "single-node-ignition-config",
+			Short: "Generates the bootstrap-in-place-for-live-iso Ignition Config asset",
+			// FIXME: add longer descriptions for our commands with examples for better UX.
```
tsorya, line 87: lets fix?

data/data/bootstrap/bootstrap-in-place/usr/local/bin/installer-master-bootstrap-in-place-gather.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
romfreiman, line 1: @staebler @crawford PTAL

jstuever (changes requested):
I have some questions about the implications of these changes as well as some nitpicky style changes.

data/data/bootstrap/bootstrap-in-place/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -0,0 +1,68 @@
...
+      needed_to_approve=true
+      echo "Approving csrs ..."
+     oc get csr -o go-template='{{range .items}}{{if not .status}}{{.metadata.name}}{{"\n"}}{{end}}{{end}}' | xargs oc adm certificate approve &> /dev/null || true
+     sleep 30
+  done
+  # Restart kubelet only if node was added
+  if $needed_to_approve ; then
+    restart_kubelet
+  fi
+}
```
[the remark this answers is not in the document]
romfreiman: It's done only once. And the behavior is similar to how it's currently work in the installer AFAIK
https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/approve-csr.sh#L8

data/data/bootstrap/bootstrap-in-place/usr/local/bin/installer-master-bootstrap-in-place-gather.sh
romfreiman: https://docs.google.com/presentation/d/1nEOaXePYcRRF-6U0X14s0BzuFYfpMzqmSVP7dX5rFqg/edit#slide=id.g3d946fe0a1_0_0

pkg/asset/ignition/bootstrap/bootstrap-in-place.go
```diff
@@ -0,0 +1,40 @@
...
+	if *installConfig.ControlPlane.Replicas != 1 {
+		return errors.Errorf("Bootstrap in place require single control plane replica, current value: %d", *installConfig.ControlPlane.Replicas)
+	}
+	logrus.Warnf("Creating single node bootstrap in place configuration")
+	return nil
+}
+
+// getCoreosInstallerArgs checks for bootstrap in place coreos installer args env
+func getCoreosInstallerArgs() string {
+	coreosInstallerEnv := os.Getenv("OPENSHIFT_INSTALL_EXPERIMENTAL_BOOTSTRAP_IN_PLACE_COREOS_INSTALLER_ARGS")
```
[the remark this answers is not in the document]
romfreiman: It was intentionally done this way. It's not a GA   feature yet, and there was explicit guidelines not to add new values to the install config. @crawford @staebler WDYT?

data/data/bootstrap/bootstrap-in-place/usr/local/bin/installer-master-bootstrap-in-place-gather.sh
eranco74: This will be called by the installer-gather.
This script will be used as the installer-gather post-reboot, see [here](https://github.com/openshift/installer/pull/4482/files#diff-d09d8f9e83a054002d5344223d496781ea603da7c52706dfcf33debf8ceb1df3R47)

data/data/bootstrap/bootstrap-in-place/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
+set -euoE pipefail ## -E option will cause functions to inherit trap
+
+export KUBECONFIG=/etc/kubernetes/bootstrap-secrets/kubeconfig
+
+function wait_for_api {
+  until oc get csr &> /dev/null
+  do
+      echo "Waiting for api ..."
```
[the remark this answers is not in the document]
eranco74: The benifit is that we can see in the log that it keeps waiting for it and didn't crash / hang

data/data/bootstrap/bootstrap-in-place/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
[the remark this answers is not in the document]
eranco74: @jstuever you are right, `needed_to_approve` is misleading.
I was trying to avoid restarting kubelet for no reason.
Yet, since this script should run only once I think I can remove this "optimization"

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -345,10 +358,17 @@ then
         --volume "$PWD:/assets:z" \
         --volume /etc/kubernetes:/etc/kubernetes:z \
         "${CLUSTER_BOOTSTRAP_IMAGE}" \
+{{- if .SingleNode.BootstrapInPlace}}
+        start --tear-down-early=false --asset-dir=/assets --required-pods=""  --assets-create-timeout=5m
```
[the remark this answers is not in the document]
eranco74: First, yes bootstrap.done means a different thing for bootstrap-in-place.
I think this is the relevant part of the enhancement - https://github.com/openshift/enhancements/pull/565/files#diff-f323c28b80ad16321f0d3b7775dd2aeba57289409b516b6aaae0362fe3337ebfR312

pkg/asset/ignition/bootstrap/bootstrap-in-place.go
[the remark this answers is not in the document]
jstuever: It is in the enhancement this way, so as long as that merges....
https://github.com/openshift/enhancements/pull/565/files#diff-f323c28b80ad16321f0d3b7775dd2aeba57289409b516b6aaae0362fe3337ebfR157

jstuever:
/hold
for enhancement to merge https://github.com/openshift/enhancements/pull/565

data/data/bootstrap/bootstrap-in-place/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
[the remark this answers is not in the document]
jstuever: I can accept that, as long as we all are aware that the trade-off is disk usage, which can compound in places like CI.

cmd/openshift-install/create.go
```diff
@@ -79,6 +79,16 @@ var (
 		},
 		assets: targetassets.IgnitionConfigs,
 	}
+	singleNodeignitionConfigTarget = target{
```
jstuever, line 82: Capital `Ignition` in `singleNodeignitionConfigTarget`

cmd/openshift-install/create.go
```diff
@@ -141,7 +151,7 @@ var (
 		assets: targetassets.Cluster,
 	}
 
-	targets = []target{installConfigTarget, manifestsTarget, ignitionConfigsTarget, clusterTarget}
+	targets = []target{installConfigTarget, manifestsTarget, ignitionConfigsTarget, clusterTarget, singleNodeignitionConfigsTarget}
```
jstuever, line 154: I believe `singleNodeignitionConfigTarget` needs to match what was used on line 82?

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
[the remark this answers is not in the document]
jstuever: I can accept that as long as it merges this way in the enhancement.

data/data/bootstrap/bootstrap-in-place/usr/local/bin/installer-master-bootstrap-in-place-gather.sh
jstuever: Thanks for pointing out where it is dropped on disk, that clarifies a lot. I'm concerned about the installer-bootstrap-in-place-gather.sh diverging from installer-gather.sh and us not gathering updated information from the cluster post-reboot. For example, one section[1] in installer-gather.sh does `Gathering cluster resources ...` Because this only ran on the bootstrap host pre-reboot, it is what will land in the log bundle and may be very confusing to someone trying to diagnose. Are we assuming that if this script can connect using the oc cli, then the user should instead be using `oc adm must-gather`?

https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/installer-gather.sh#L82

data/data/bootstrap/bootstrap-in-place/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
```
jstuever, line 1: How is the output of this script being captured in the `gather bootstrap` log bundle?

data/data/bootstrap/bootstrap-in-place/usr/local/bin/installer-master-bootstrap-in-place-gather.sh
omertuc: With regards to what information it is collecting - it is not diverging.

For getting bootstrap logs, it simply uses the bootstrap log bundle that was collected before reboot by the original `installer-gather.sh` pre-pivot, passed on to it via the master ignition.

For getting master logs, it simply calls `installer-masters-gather.sh` on itself - which is exactly what the originals `installer-gather.sh` script runs on remote masters.

Calling `openshift-installer gather bootstrap` after the bootstrap node is taken down in non-single-node deployments is not possible. In single node deployments, we added that "ability" that was not there before. The user can still call `oc adm must-gather`, but if the user also wants information about the bootstrap process, they still have the option to call `openshift-installer gather bootstrap`.

data/data/bootstrap/bootstrap-in-place/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
eranco74: Good catch, should be added.

data/data/bootstrap/bootstrap-in-place/usr/local/bin/installer-master-bootstrap-in-place-gather.sh
jstuever: I'm not concerned about the `installer-masters-gather.sh` portion of this. The log bundle will clearly contain the logs from the host from a control plane perspective. I am concerned that the various `oc get`[1] (and similar) portions of the `installer-gather.sh` script are only executed while the host is configured as a pre-reboot bootstrap host. This point-in-time snapshot of the API does not contain an up-to-date picture of the cluster status at the time the `openshift-installer gather bootstrap` command is executed. There will surely be changes to the cluster post-reboot. This unnecessarily complicates the troubleshooting process by removing a piece of the puzzle.

To this same point, perhaps the gather scripts should be removed from the host once the cluster passes the requirements where a non-single-node installation would tear down the bootstrap node. A "day 2" `gather bootstrap` log bundle is going to be equally confusing to troubleshoot. This will reinforce the use of `must-gather` at this point.

[1] https://github.com/openshift/installer/blob/95855a1b77e1a6904a91e4b4a1f931d194f659f2/data/data/bootstrap/files/usr/local/bin/installer-gather.sh#L82-L111

eranco74: Agree, it might be a bit misleading.
We can place the pre-reboot logs under bootstrap-pre-reboot directory.
We can also add the `oc get` commands to installer-master-bootstrap-in-place-gather.sh
That been said, the bootstrap log bandle and the installer-master-bootstrap-in-place-gather.sh should be deleted upon successful installation.

pkg/asset/ignition/bootstrap/bootstrap-in-place.go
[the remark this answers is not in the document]
eranco74: Removed this env, the user will have to ssh to complete the installation manually, or [edit the ignition](https://github.com/eranco74/bootstrap-in-place-poc/pull/9).

data/data/bootstrap/bootstrap-in-place/usr/local/bin/installer-master-bootstrap-in-place-gather.sh
omertuc: Added [PR](https://github.com/eranco74/installer/pull/11) in response to this discussion, @jstuever and @eranco74 please review

jstuever: That PR might be our solution. However, before that I'm interested if we can simply run the original `installer-gather.sh` on the node, which should run the `installer-masters-gather.sh` on itself. We could then update`installer-gather.sh` to override specific subfolders (bootstrap, rendered-assets, unit-assets) when the BOOTSTRAP_PHASE_LOG_BUNDLE_ARCHIVE_PATH file exists and then dump everything else into a new subfolder. Alternatively, we can instead just dump everything in this file into the new subfolder.

omertuc: That makes sense and sounds like it will simplify things, I'll take a look

omertuc: Got rid of the PR mentioned above.

Created this instead https://github.com/eranco74/installer/pull/12

Is that what you had in mind?

data/data/bootstrap/bootstrap-in-place/opt/openshift/bootstrap-in-place/master-update.fcc
```diff
@@ -0,0 +1,66 @@
...
+      contents:
+        local: bin/installer-masters-gather.sh
+      mode: 0555
+    - path: /usr/local/bin/installer-gather.sh
+      contents:
+        local: bin/installer-master-bootstrap-in-place-gather.sh
+      mode: 0555
+systemd:
+  units:
+    - name: bootstrap-in-place-post-reboot.service
```
jstuever, line 53: @eranco74 I believe the bootkube service no longer exists at this point. Does it make sense to land `bootstrap-in-place-post-reboot.service` as `bootkube` instead? The script could then echo a note about being a BiP workflow and it would show up at the top of the bootkube service logs. This would also remove the necessity of collecting this new service logs in the `installer-gather.sh`.

eranco74: Hmmm, I must say that I like it :smile: ,
This service serves the same purpose as bootkube.
Debugging is more straight forward (don't need to remember the new service name)
I'm just a bit concerned that it might be confusing, and less clear if the node is after reboot or before the reboot.
@romfreiman, @tsorya @omertuc your thoughts on this?

romfreiman: I guess it can work. No strong objection from my side. Just have explicit logging there.

pkg/types/installconfig.go
```diff
@@ -141,6 +141,10 @@ type InstallConfig struct {
 	// GCP: "Mint", "Passthrough", "Manual"
 	// +optional
 	CredentialsMode CredentialsMode `json:"credentialsMode,omitempty"`
+
+	// BootstrapInPlace is the configuration for installing single node
+	// with bootstrap in place installation.
+	BootstrapInPlace *BootstrapInPlace `json:"bootstrapInPlace"`
```
jstuever, line 147: This is not a required field, and should be omitempty. This should also fix the unit tests.
```diff
-	BootstrapInPlace *BootstrapInPlace `json:"bootstrapInPlace"`
+	BootstrapInPlace *BootstrapInPlace `json:"bootstrapInPlace,omitempty"`
```

pkg/asset/manifests/operators_test.go
```diff
@@ -60,6 +60,7 @@ func TestRedactedInstallConfig(t *testing.T) {
 	}
 	expectedConfig := createInstallConfig()
 	expectedYaml := `baseDomain: test-domain
+bootstrapInPlace: null
```
jstuever, line 63: Why is this needed?

jstuever (changes requested):
This is getting really close IMHO. I left some comments that should help resolve the unit test failure. Also quite a few nit picky changes (sorry).

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -0,0 +1,69 @@
...
+  do
+    echo "Approving csrs ..."
+    oc get csr -o go-template='{{range .items}}{{if not .status}}{{.metadata.name}}{{"\n"}}{{end}}{{end}}' | xargs --no-run-if-empty oc adm certificate approve &> /dev/null || true
+    sleep 30
+  done
+}
+
+function wait_for_cvo {
+  echo "Waiting for cvo"
+  until [ "$(oc get clusterversion -o jsonpath='{.items[0].status.conditions[?(@.type=="Available")].status}')" == "True" ];
```
wking, line 46: What is the motivation for waiting on the cluster-version operator declaring the install complete?  On typical installs, we lose the ability to run `openshift-install gather bootstrap ...` after bootstrap-teardown.  This post-reboot script is the bootstrap-in-place analog of post-bootstrap-teardown, right?  Can't we rely on must-gather and other day-2 tooling at that point, without having to preserve assets to support `openshift-install gather bootstrap ...`?

data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/bootstrap-in-place.sh
```diff
@@ -0,0 +1,53 @@
...
+
+bootkube_podman_run() {
+  # we run all commands in the host-network to prevent IP conflicts with
+  # end-user infrastructure.
+  podman run --quiet --net=host "${@}"
+}
+
+if [ ! -f stop-etcd.done ]; then
+  echo "Stop etcd static pod by moving the manifest"
+  mv /etc/kubernetes/manifests/etcd-member-pod.yaml /etc/kubernetes
```
wking, line 16: Seems like there is a perma-fail possibility here where we successfully `mv` here but time out or otherwise hiccup while waiting for etcd to go down.  Then, because we died before the `stop-etcd.done` touch, we come into this block again, try to `mv` again, and fail, and continue to fail on the `mv` in all future unit/script invocations.  Can we:

```sh
test -f /etc/kubernetes/manifests/etcd-member-pod.yaml && mv /etc/kubernetes/manifests/etcd-member-pod.yaml /etc/kubernetes
```

or:

```sh
mv /etc/kubernetes/manifests/etcd-member-pod.yaml /etc/kubernetes || echo "already moved etcd-member-pod.yaml
```

or something?

data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/bootstrap-in-place.sh
```diff
@@ -0,0 +1,53 @@
...
+    sleep 10
+  done
+
+  touch stop-etcd.done
+fi
+
+if [ ! -f master-ignition.done ]; then
+  echo "Creating master ignition and writing it to disk"
+  # Get the master ignition from MCS
+  curl --header "Accept:'application/vnd.coreos.ignition+json;version=3.1.0 ;q=0.1'" \
```
wking, line 29: No need for the `;q=0.1`, is there?  [RFC 7231][1] has the weight coming in via an optional `accept-params`.  I also don't see the RFC asking for your `'` quoting.  So I'd expect:

```sh
curl --header 'Accept:application/vnd.coreos.ignition+json;version=3.1.0' \
  http://localhost:22624/config/master -o /opt/openshift/original-master.ign
```

Am I just missing something?

[1]: https://tools.ietf.org/html/rfc7231#section-5.3.2

data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/install-to-disk.sh.template
```diff
@@ -0,0 +1,24 @@
...
+  echo "Executing coreos-installer with the following options: install -i /opt/openshift/master.ign {{.BootstrapInPlace.InstallationDisk}}"
+  coreos-installer install -i /opt/openshift/master.ign {{.BootstrapInPlace.InstallationDisk}}
++
+  touch coreos-installer.done
+fi
+
+if [ ! -f reboot.done ]; then
+  echo "Going to reboot"
+  shutdown -r +1 "Bootstrap completed, server is going to reboot."
+  touch reboot.done
```
wking, line 22: Does this `reboot.done` survive the reboot?  If not, do we want to drop it and the `if` guard here, so if for some reason the shutdown sticks and the unit/script gets restarted, it will call `shutdown` again?

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -345,10 +358,17 @@ then
         --volume "$PWD:/assets:z" \
         --volume /etc/kubernetes:/etc/kubernetes:z \
         "${CLUSTER_BOOTSTRAP_IMAGE}" \
+{{- if .BootstrapInPlace}}
+        start --tear-down-early=false --asset-dir=/assets --required-pods=""  --assets-create-timeout=5m
```
wking, line 362: nit: doubled space before `--assets-create-timeout=5m`

data/data/bootstrap/files/usr/local/bin/installer-gather.sh
```diff
@@ -57,8 +57,13 @@ find "${ARTIFACTS}/rendered-assets" -name "*kubeconfig*" -print0 | xargs -0 rm
 find "${ARTIFACTS}/rendered-assets" -name "*.key" -print0 | xargs -0 rm
 find "${ARTIFACTS}/rendered-assets" -name ".kube" -print0 | xargs -0 rm -rf
 
-
 # Collect cluster data
+GATHER_KUBECONFIG="/opt/openshift/auth/kubeconfig"
+if [[ -f "/etc/kubernetes/bootstrap-secrets/kubeconfig" ]]; then
+    echo "Using /etc/kubernetes/bootstrap-secrets/kubeconfig as KUBECONFIG"
```
wking, line 63: nit: you can DRY this up slightly by shifting this `echo` down one line and using:

```sh
echo "Using ${GATHER_KUBECONFIG} as KUBECONFIG"
```

data/data/bootstrap/files/usr/local/bin/installer-gather.sh
```diff
@@ -71,53 +76,79 @@ function queue() {
...
+    echo "Including existing bootstrap bundle ${LOG_BUNDLE_BOOTSTRAP_ARCHIVE_NAME}"
+    tar -xzf ${LOG_BUNDLE_BOOTSTRAP_ARCHIVE_NAME}
+    mkdir -p "${ARTIFACTS}/bootstrap-log-bundle"
+    cp -r "${LOG_BUNDLE_BOOTSTRAP_NAME}"/* "${ARTIFACTS}/bootstrap-log-bundle/"
+fi
+
 echo "Gather remote logs"
 export MASTERS=()
-if [ "$#" -ne 0 ]; then
+if [[ -f ${LOG_BUNDLE_BOOTSTRAP_ARCHIVE_NAME} ]]; then
```
wking, line 144: is it worth switching on this dynamically at script-run-time, instead of making this `installer-gather.sh.template` and switching on `{{- if .BootstrapInPlace}}` at script-creation-time?

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -34,6 +34,16 @@ spec:
             description: BaseDomain is the base domain to which the cluster should
               belong.
             type: string
+          bootstrapInPlace:
+            description: BootstrapInPlace is the configuration for installing single
```
wking, line 38: nit "installing single node" -> "installing a single node"

pkg/asset/ignition/bootstrap/bootstrap_in_place.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"github.com/pkg/errors"
+
+	"github.com/openshift/installer/pkg/types"
+)
+
+// verifyBootstrapInPlace validate the number of control plane replica is one and that installation disk is set
+func verifyBootstrapInPlace(installConfig *types.InstallConfig) error {
+	if *installConfig.ControlPlane.Replicas != 1 {
+		return errors.Errorf("bootstrap in place require single control plane replica, current value: %d", *installConfig.ControlPlane.Replicas)
```
wking, line 12: nit: "bootstrap in place requires a single control plane replica" ("require single" -> "requires a single")

pkg/asset/ignition/bootstrap/bootstrap_in_place.go
```diff
@@ -0,0 +1,21 @@
...
+	}
+	if installConfig.BootstrapInPlace == nil {
+		return errors.Errorf("Missing bootstrap in place configuration")
```
wking, line 15: nit: [error strings should not be capitalized][1].

[1]: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

data/data/bootstrap/files/usr/local/bin/installer-gather.sh
eranco74: The same script is executed on the bootstrap prior to the reboot (where the log bundle doesn't exists) and after the reboot (where we need it to add the bootstrap log bundle).
This is why we are switching on run-time.

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
eranco74: On typical installs, the bootstrap goes down only once we have a working control plane on the master nodes.

We can replace `wait_for_cvo` with waiting for the same `required-pods` cluster-bootstrap waits for:
"openshift-kube-apiserver/kube-apiserver,openshift-kube-scheduler/openshift-kube-scheduler,openshift-kube-controller-manager/kube-controller-manager,openshift-cluster-version/cluster-version-operator"

Yet I don't think there is much value in deleting the bootstrap logs early.

data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/bootstrap-in-place.sh
eranco74: You are right.

data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/install-to-disk.sh.template
eranco74: It will not survive the reboot.
I added this just to mark that the script executed the reboot command.
I agree there is little value to it and I'll just remove it.

jstuever:
/hold cancel
Enhancement doc has merged.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -317,24 +337,17 @@ then
...
-		--cacert=/opt/openshift/tls/etcd-ca-bundle.crt \
-		--cert=/opt/openshift/tls/etcd-client.crt \
-		--key=/opt/openshift/tls/etcd-client.key \
-		--endpoints="${ETCD_ENDPOINTS}" \
-		endpoint health
-do
-	echo "etcdctl failed. Retrying in 5 seconds..."
-	sleep 5
-done
+# in case of single node, if we removed etcd, there is no point to wait for it on restart
```
dhellmann, line 340: Would it be better to have the logic of the script stay the same in all forms, but to generate settings in variables that are different when we're using the bootstrap-in-place approach? That would benefit someone debugging multiple clusters of different types, since the script content would always be the same.

For example, if we put something like this early in the file

```
{{- if .BootstrapInPlace }}
BOOTSTRAP_INPLACE=true
{{ else }}
BOOTSTRAP_INPLACE=false
{{ end -}}
```

then, in the places where it matters, the shell script can test that variable.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -345,10 +358,17 @@ then
         --volume "$PWD:/assets:z" \
         --volume /etc/kubernetes:/etc/kubernetes:z \
         "${CLUSTER_BOOTSTRAP_IMAGE}" \
+{{- if .BootstrapInPlace}}
```
dhellmann, line 361: This section could use `.BootstrapInPlace` to set a variable like `REQUIRED_PODS` and then use that variable in the command line below.

What is `assets-create-timeout` if we are not using the bootstrap-in-place approach?

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -317,24 +337,17 @@ then
...
-		--key=/opt/openshift/tls/etcd-client.key \
-		--endpoints="${ETCD_ENDPOINTS}" \
-		endpoint health
-do
-	echo "etcdctl failed. Retrying in 5 seconds..."
-	sleep 5
-done
+# in case of single node, if we removed etcd, there is no point to wait for it on restart
+{{- if .BootstrapInPlace}}
+if [ ! -f stop-etcd.done ]
```
dhellmann, line 342: Is there any reason this test should only be run when bootstrapping in place?

pkg/asset/ignition/bootstrap/bootstrap_in_place.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/openshift/installer/pkg/types"
+)
+
+// verifyBootstrapInPlace validate the number of control plane replica is one and that installation disk is set
+func verifyBootstrapInPlace(installConfig *types.InstallConfig) error {
+	if *installConfig.ControlPlane.Replicas != 1 {
+		return errors.Errorf("bootstrap in place requires a single control plane replica, current value: %d", *installConfig.ControlPlane.Replicas)
+	}
+	if installConfig.BootstrapInPlace == nil {
+		return errors.Errorf("missing bootstrap in place configuration")
```
dhellmann, line 15: In other places in the validation code the validation framework is used to indicate exactly what field name is missing or has a bad value. Can that framework be used here, or do we need to update the error messages to be more specific? (See `pkg/types/baremetal/validation/platform.go` for an example)

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
eranco74: It's how long cluster-bootstrap should wait for all manifests to be created.
In a regular installation, it might take some time because some manifests might require an operator to start running and create the CRD in order to apply the CR.
In bootstrap-in-place deployment, no operator will start running so there is no reason to wait that long.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
eranco74: Yes, because the bootstrap-in-place.sh takes etcd down once cluster-bootstrap is done.
So in case bootstrap-in-place.sh failed or timed out bootkube unit gets restarted.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
eranco74: We can definitely add this variable or a comment saying `###USING BOOTSTRAP IN PLACE CONFIGURATION`.
 Just note that this is a template and we are switching on {{- if .BootstrapInPlace}} at script-creation-time and not run-time.

dhellmann: Right, I get that this is a template. My point is that we shouldn't be changing the _logic_ of the script at compile time, we should only change the inputs. Changing the logic means that someone looking at the raw input file has to run it through a golang template translator in their head to understand which lines should appear on a single-node vs. multi-node deployment for debugging, instead of being able to follow the shell logic after seeing that the template simply sets a variable near the top of the file.

Maybe the other reviewers are less concerned with that new style of change in this file, though, since no one else has mentioned it.

staebler: For what it's worth, I made a similar point on one of my earlier [reviews](https://github.com/staebler). I am in favor of preferring script conditional logic over template conditional logic.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
eranco74: This used to be an optimization while single node installation required patches to CVO.
I think we can go back to use the default.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
eranco74: OK, I don't have a preference for one over the other, I'll update the script.

pkg/asset/ignition/bootstrap/bootstrap_in_place.go
```diff
@@ -0,0 +1,21 @@
+package bootstrap
+
+import (
+	"github.com/pkg/errors"
+
+	"github.com/openshift/installer/pkg/types"
+)
+
+// verifyBootstrapInPlace validate the number of control plane replica is one and that installation disk is set
+func verifyBootstrapInPlace(installConfig *types.InstallConfig) error {
```
dhellmann, line 10: This comment got lost when the last update was made:

In other places in the validation code the validation framework is used to indicate exactly what field name is missing or has a bad value. Can that framework be used here, or do we need to update the error messages to be more specific? (See pkg/types/baremetal/validation/platform.go for an example)

eranco74: I'm not sure that I understand,
I don't see a difference between these error messages and the errors in  pkg/types/baremetal/validation/platform.go.
Are you suggesting to move pkg/asset/ignition/bootstrap/bootstrap_in_place.go to pkg/types/none/validation/ ?

dhellmann: Eran and I talked about this and I'm fine with using another PR to touch up the error messages to be clearer about exactly which field has the bad value.

dhellmann (approved):

pkg/asset/ignition/bootstrap/bootstrap_in_place.go
dhellmann: @eranco74 I don't have permission to resolve threads, so feel free to close this one when you have a minute.

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
jstuever: waiting for the same required-pods would further reduce the differences in this workflow.

data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/bootstrap-in-place.sh
```diff
@@ -0,0 +1,53 @@
...
+  echo "Creating master ignition and writing it to disk"
+  # Get the master ignition from MCS
+  curl --header 'Accept:application/vnd.coreos.ignition+json;version=3.1.0' \
+    http://localhost:22624/config/master -o /opt/openshift/original-master.ign
+
+  GATHER_ID="bootstrap"
+  GATHER_TAR_FILE=log-bundle-${GATHER_ID}.tar.gz
+
+  echo "Gathering installer bootstrap logs"
+  TAR_FILE=${GATHER_TAR_FILE} /usr/local/bin/installer-gather.sh --id $GATHER_ID
```
jstuever, line 36: nit: `${GATHER_ID}`

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -317,25 +343,25 @@ then
...
+if [ "$BOOTSTRAP_INPLACE" = true ]
+then
+    if [ ! -f stop-etcd.done ]
+    then
+        # Wait for the etcd cluster to come up.
+        wait_for_etcd_cluster
+    fi
+else
+    wait_for_etcd_cluster
+fi
```
jstuever, line 356: nit: this could be flattened... if bootstrap_inpace = false or ! -f stop-etcd.done; then wait_for_etcd_cluster; fi

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -345,23 +371,28 @@ then
         --volume "$PWD:/assets:z" \
         --volume /etc/kubernetes:/etc/kubernetes:z \
         "${CLUSTER_BOOTSTRAP_IMAGE}" \
-        start --tear-down-early=false --asset-dir=/assets --required-pods="openshift-kube-apiserver/kube-apiserver,openshift-kube-scheduler/openshift-kube-scheduler,openshift-kube-controller-manager/kube-controller-manager,openshift-cluster-version/cluster-version-operator"
+        start --tear-down-early=false --asset-dir=/assets --required-pods="$REQUIRED_PODS"
```
jstuever, line 374: nit: `${REQUIRED_PODS}`

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -345,23 +371,28 @@ then
...
-			wait-for-ceo \
-			--kubeconfig /assets/auth/kubeconfig
+    . /usr/local/bin/bootstrap-in-place.sh "${CLUSTER_BOOTSTRAP_IMAGE}"
+else
+    rm --force /etc/kubernetes/manifests/machineconfigoperator-bootstrap-pod.yaml
+
+    if [ ! -z "$CLUSTER_ETCD_OPERATOR_IMAGE" ]
+    then
+        echo "Waiting for CEO to finish..."
+        bootkube_podman_run \
```
jstuever, line 388: nit: this could be flattened out of the if/then starting on 379... if bootstrap_inplace=false and ! -z cluster_edcd_operator_image; then...

data/data/bootstrap/files/usr/local/bin/installer-gather.sh
```diff
@@ -71,53 +76,80 @@ function queue() {
...
+    # Instead of running installer-masters-gather.sh on remote masters, run it on the current node
+    MASTER_GATHER_ID="master-${GATHER_ID}"
+    MASTER_ARTIFACTS="/tmp/artifacts-${MASTER_GATHER_ID}"
+    mkdir -p "${ARTIFACTS}/control-plane/master"
+    sudo /usr/local/bin/installer-masters-gather.sh --id "${MASTER_GATHER_ID}" </dev/null
+    cp -r "$MASTER_ARTIFACTS"/* "${ARTIFACTS}/control-plane/master/"
```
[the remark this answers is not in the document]
jstuever: nit: `${MASTER_ARTIFACTS}`

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -317,25 +343,25 @@ then
...
+    wait_for_etcd_cluster
+fi
 
+
+if [ "$BOOTSTRAP_INPLACE" = true ]
+then
+    REQUIRED_PODS=""
+else
+    REQUIRED_PODS="openshift-kube-apiserver/kube-apiserver,openshift-kube-scheduler/openshift-kube-scheduler,openshift-kube-controller-manager/kube-controller-manager,openshift-cluster-version/cluster-version-operator"
+fi
```
jstuever, line 364: nit: could be flattened...
```
REQUIRED_PODS="openshift-kube-apiserver/kube-apiserver,openshift-kube-scheduler/openshift-kube-scheduler,openshift-kube-controller-manager/kube-controller-manager,openshift-cluster-version/cluster-version-operator"
if [ "${BOOTSTRAP_INPLACE}" = true ]
then 
    REQUIRED_PODS=""
fi
```

pkg/asset/installconfig/azure/mock/azureclient_generated.go
```diff
@@ -38,21 +38,6 @@ func (m *MockAPI) EXPECT() *MockAPIMockRecorder {
 	return m.recorder
 }
 
-// GetVirtualNetwork mocks base method.
```
jstuever, line 41: Were these supposed to change, or was this the result of a prior `./hack/go-genmock.sh`?
- `pkg/asset/installconfig/azure/mock/azureclient_generated.go`
- `pkg/asset/installconfig/gcp/mock/gcpclient_generated.go`
- `pkg/asset/installconfig/kubevirt/mock/client_generated.go`

jstuever (changes requested):
It looks like there were some unrelated go-genmock.sh changes. Other than that, a lot of nickpick due to recent changes.

romfreiman:
@jstuever if it looks fine, can you please approve and Ill lgtm tomorrow once @eranco74 will fix all the nits?

pkg/asset/ignition/bootstrap/bootstrap_in_place.go
```diff
@@ -0,0 +1,23 @@
...
+	if installConfig.BootstrapInPlace == nil {
+		return field.Required(field.NewPath("BootstrapInPlace"), "missing BootstrapInPlace configuration")
+	}
+	if *installConfig.ControlPlane.Replicas != 1 {
+		errorList = append(errorList, field.Invalid(field.NewPath("ControlPlane").Child("Replicas"), installConfig.ControlPlane.Replicas,
+			"bootstrap in place requires a single ControlPlane replica"))
+	}
+	if installConfig.BootstrapInPlace.InstallationDisk == "" {
+		errorList = append(errorList, field.Invalid(field.NewPath("BootstrapInPlace").Child("InstallationDisk"), installConfig.BootstrapInPlace.InstallationDisk,
+			"no InstallationDisk configured for BootstrapInPlace, please set this value to be the target disk drive for the installation"))
```
staebler, line 20:
```suggestion
		errorList = append(errorList, field.Required(field.NewPath("bootstrapInPlace", "installationDisk"), installConfig.BootstrapInPlace.InstallationDisk,
			"installationDisk must be set to the target disk drive for the installation"))
```

pkg/asset/targets/targets.go
```diff
@@ -60,6 +60,14 @@ var (
 		&cluster.Metadata{},
 	}
 
+	// SingleNodeIgnitionConfig is the bootstrap-in-place ignition-config targeted assets.
+	SingleNodeIgnitionConfig = []asset.WritableAsset{
+		&kubeconfig.AdminClient{},
+		&password.KubeadminPassword{},
+		&bootstrap.Bootstrap{BootstrapInPlace: true},
```
staebler, line 67: This is not valid. You need to have a separate asset type. You can have that separate asset type compose `bootstrap.Bootstrap`.

The asset store differentiates between asset by their types. So if you use the same type for both bootstrap ignition assets, the asset store will not be able to tell them apart. The consequence of this is that if the user were to run `openshift-install create ignition-configs` followed by `openshift-install create single-node-ignition-config`, then the resulting ignition config would not have the bootstrap-in-place bits.

pkg/asset/ignition/bootstrap/bootstrap_in_place.go
```diff
@@ -0,0 +1,23 @@
...
+func verifyBootstrapInPlace(installConfig *types.InstallConfig) error {
+	errorList := field.ErrorList{}
+	if installConfig.BootstrapInPlace == nil {
+		return field.Required(field.NewPath("BootstrapInPlace"), "missing BootstrapInPlace configuration")
+	}
+	if *installConfig.ControlPlane.Replicas != 1 {
+		errorList = append(errorList, field.Invalid(field.NewPath("ControlPlane").Child("Replicas"), installConfig.ControlPlane.Replicas,
+			"bootstrap in place requires a single ControlPlane replica"))
+	}
+	if installConfig.BootstrapInPlace.InstallationDisk == "" {
```
staebler, line 18: Is there any other validation that we may want to perform on the installation disk?

pkg/asset/installconfig/azure/mock/azureclient_generated.go
eranco74: They wherre not supposed to change due to my changes but for some reason they did.
I guess someone forgot to generate them after updating instalconfig.go

data/data/bootstrap/files/usr/local/bin/installer-gather.sh
```diff
@@ -71,53 +79,79 @@ function queue() {
...
+# An example for a scenario where this happens is when we're running inside a single-node
+# bootstrap-in-place deployment post-pivot master node, rather than a typical bootstrap node.
+# In that case, the bootstrap node collects and bundles logs pre-reboot and transfers that bundle
+# via an ignition file to this post-pivot master node.
+LOG_BUNDLE_BOOTSTRAP_NAME="log-bundle-bootstrap"
+LOG_BUNDLE_BOOTSTRAP_ARCHIVE_NAME="/var/log/${LOG_BUNDLE_BOOTSTRAP_NAME}.tar.gz"
+
+if [[ -f ${LOG_BUNDLE_BOOTSTRAP_ARCHIVE_NAME} ]]; then
+    echo "Including existing bootstrap bundle ${LOG_BUNDLE_BOOTSTRAP_ARCHIVE_NAME}"
+    tar -xzf ${LOG_BUNDLE_BOOTSTRAP_ARCHIVE_NAME}
```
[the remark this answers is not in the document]
omertuc: > Do we need to extract the archive or can we just add the archive as is to the artifacts?
Choosing one over the other is rather arbitrary. I lean towards extracting it because it's more convenient for the user. Also "double compression" i.e. `.tar.gz` inside `.tar.gz` feels wrong.

> What is the working directory where these extracted files are going to be left?
`~`
I'll change it so it gets extracted directly into the bundle instead

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
eranco74: This will change the behavior of non bootstrap-in-place installation.
`rm --force /etc/kubernetes/manifests/machineconfigoperator-bootstrap-pod.yaml`
Doesn't depend on CLUSTER_ETCD_OPERATOR_IMAGE env.

Also, the machineconfigoperator-bootstrap-pod. is required for bootstrap-in-place.sh so it shouldn't be removed in case "$BOOTSTRAP_INPLACE" = true

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
eranco74: I agree it will reduce the differences., but I'm not sure it will work.
In a regular installation, etcd will keep running and the static pods resources and secrets will still exist under /etc/kubernetes.
Also, I think that if we really want it we should add support for this wait option in cluster-bootstrap. meaning it should just wait without setting up a temporary control plane. 
Trying to replicate the cluster-bootstrap code in bash seems wrong.  
 So I suggest keeping the current implementation and have another task for checking if it's possible to do the cleanup earlier, if it's possible we can add the option to cluster-bootstrap and use it.

pkg/asset/ignition/bootstrap/bootstrap_in_place.go
```diff
@@ -8,16 +8,20 @@ import (
 // verifyBootstrapInPlace validate the number of control plane replica is one and that installation disk is set
 func verifyBootstrapInPlace(installConfig *types.InstallConfig) error {
 	errorList := field.ErrorList{}
-	if installConfig.BootstrapInPlace == nil {
-		return field.Required(field.NewPath("BootstrapInPlace"), "missing BootstrapInPlace configuration")
+	if installConfig.ControlPlane.Replicas == nil {
+		errorList = append(errorList, field.Invalid(field.NewPath("controlPlane", "replicas"), installConfig.ControlPlane.Replicas,
```
staebler, line 12:
```suggestion
	errorList = append(errorList, field.Required(field.NewPath("controlPlane", "replicas"),
```

pkg/asset/ignition/bootstrap/bootstrap_in_place.go
```diff
@@ -8,16 +8,20 @@ import (
...
 	}
-	if installConfig.BootstrapInPlace.InstallationDisk == "" {
-		errorList = append(errorList, field.Invalid(field.NewPath("BootstrapInPlace").Child("InstallationDisk"), installConfig.BootstrapInPlace.InstallationDisk,
-			"no InstallationDisk configured for BootstrapInPlace, please set this value to be the target disk drive for the installation"))
+	if installConfig.BootstrapInPlace == nil {
+		errorList = append(errorList, field.Required(field.NewPath("bootstrapInPlace"), "bootstrapInPlace is required when creating a single node bootstrap-in-place ignition"))
+	}
+	else if installConfig.BootstrapInPlace.InstallationDisk == "" {
+		errorList = append(errorList, field.Required(field.NewPath("bootstrapInPlace", "installationDisk"),
+			"installationDisk must be set the target disk drive for the installation"))
```
staebler, line 24:
```suggestion
			"installationDisk must be set to the target disk drive for the installation"))
```

pkg/asset/ignition/bootstrap/bootstrap_in_place.go
```diff
@@ -0,0 +1,27 @@
...
+			"bootstrap in place requires ControlPlane.Replicas configuration"))
+	}
+	if *installConfig.ControlPlane.Replicas != 1 {
+		errorList = append(errorList, field.Invalid(field.NewPath("controlPlane", "replicas"), installConfig.ControlPlane.Replicas,
+			"bootstrap in place requires a single ControlPlane replica"))
+	}
+	if installConfig.BootstrapInPlace == nil {
+		errorList = append(errorList, field.Required(field.NewPath("bootstrapInPlace"), "bootstrapInPlace is required when creating a single node bootstrap-in-place ignition"))
+	}
+	else if installConfig.BootstrapInPlace.InstallationDisk == "" {
```
staebler, line 22:
```suggestion
	} else if installConfig.BootstrapInPlace.InstallationDisk == "" {
```

data/data/bootstrap/files/usr/local/bin/installer-gather.sh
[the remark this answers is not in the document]
staebler: Sounds good. I prefer extracting it, too.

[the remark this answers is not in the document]
omertuc: Fixed broken link, the diff is actually https://github.com/eranco74/installer/pull/19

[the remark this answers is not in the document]
eranco74: Fixed

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
jstuever: Allow me to clarify.... keep:
```
if [ "$BOOTSTRAP_INPLACE" = true ]
then 
    . /usr/local/bin/bootstrap-in-place.sh "${CLUSTER_BOOTSTRAP_IMAGE}"
else
    rm --force /etc/kubernetes/manifests/machineconfigoperator-bootstrap-pod.yaml
fi
```
and the flatten the rest:
```
if [ ! -z "$CLUSTER_ETCD_OPERATOR_IMAGE" ] && [ ! "$BOOTSTRAP_INPLACE" = true ]
then
        echo "Waiting for CEO to finish..."
```

pkg/asset/installconfig/azure/mock/azureclient_generated.go
jstuever: Are they required to change for your PR to merge? I'd prefer to pull them out of this PR if not.

staebler: I agree. Let's leave them out. We'll put up a separate PR with these changes.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
staebler: I don't like the first part of that. Combining the call to `bootstrap-in-place.sh` with the remocal of the MCO pod makes it seem like they are more closely related than they are.

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -278,6 +280,7 @@ func (a *Bootstrap) getTemplateData(installConfig *types.InstallConfig, releaseI
 		BootImage:             string(*rhcosImage),
 		PlatformData:          platformData,
 		ClusterProfile:        clusterProfile,
+		BootstrapInPlace:      installConfig.BootstrapInPlace,
```
staebler, line 283: This makes it so that if the user includes the `bootstrapInPlace` field in the install config, then the bootstrap ignition created be calling `create ignition-configs` or `create cluster` will not work. The `bootkube.sh` will be set up for bootstrap-in-place.

pkg/asset/ignition/bootstrap/bootstrap_in_place.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	err = a.Bootstrap.addStorageFiles("/", "bootstrap/bootstrap-in-place/files", templateData)
+	if err != nil {
+		return err
+	}
+	err = a.Bootstrap.addSystemdUnits("bootstrap/bootstrap-in-place/systemd/units", templateData)
+	if err != nil {
+		return err
+	}
+	a.File = &asset.File{
```
staebler, line 76: You need to populate `a.Config` as part of the generate as well.

pkg/asset/ignition/bootstrap/bootstrap_in_place.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		return false, err
+	}
+
+	config := &igntypes.Config{}
+	if err := json.Unmarshal(file.Data, config); err != nil {
+		return false, errors.Wrapf(err, "failed to unmarshal %s", bootstrapIgnFilename)
+	}
+
+	a.File, a.Config = file, config
```
staebler, line 98: The `Load` needs to populate `a.Bootstrap` as well, or you need to make `a.Bootstrap` private. Otherwise, the asset store will always consider the asset dirty.

The asset store has a JSON copy of the asset. When the store loads the asset from disk, the store compares the asset loaded against the JSON copy of the asset. If they are different, then the asset store considers the asset dirty and all of the assets that depend on the dirty asset to be dirty as well. In the case of the bootstrap-in-place asset, that does not have consequences since there are no other assets that depend upon it, but it should still be fixed.

staebler (commented):
Can you try to improve on the changes included in each commit? There should not be intermediate changes in one commit with fixes in another commit. For example, the second commit should not have changes to the `cmd/openshift-install/create.go` file. All of the changes to that file should be in the first commit.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
jstuever: Are you talking about switching that to:
```
if [ "$BOOTSTRAP_INPLACE" = true ]
then 
    . /usr/local/bin/bootstrap-in-place.sh "${CLUSTER_BOOTSTRAP_IMAGE}"
fi

if [ ! "$BOOTSTRAP_INPLACE" = true ]
then 
    rm --force /etc/kubernetes/manifests/machineconfigoperator-bootstrap-pod.yaml
fi
```

pkg/asset/ignition/bootstrap/bootstrap.go
eranco74: That is correct.
To be honest I think that once we added the bootstrapInPlace field to the install config there is no longer reason to have the `create single-node-ignition-config` command.
I'll send a PR to the enhancement proposing that change.
How do you suggest to mitigate that?, seems that I should add the BootstrapInPlace bool back to `Bootstrap`

staebler (commented):
/approve

romfreiman:
booooom!!!!!

staebler:
/lgtm cancel

```
Feb 11 22:04:34 ip-10-0-38-130 bootkube.sh[2178]: /usr/local/bin/bootkube.sh: line 348: syntax error near unexpected token `fi'
```

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -317,24 +343,18 @@ then
...
+if [ ! -f stop-etcd.done ]
+then
+	# Wait for the etcd cluster to come up.
+	wait_for_etcd_cluster
+fi
+
+REQUIRED_PODS="openshift-kube-apiserver/kube-apiserver,openshift-kube-scheduler/openshift-kube-scheduler,openshift-kube-controller-manager/kube-controller-manager,openshift-cluster-version/cluster-version-operator"
+if [ "$BOOTSTRAP_INPLACE" = true ]
+    REQUIRED_PODS=""
+then
```
staebler, line 356:
```suggestion
then
    REQUIRED_PODS=""
```

staebler:
```
 2021/02/12 05:00:32 Pod e2e-aws-ipi-install-install succeeded after 33m0s
```
:man_dancing:

PR #4483, opened 2020-12-15
Bug 1907822: Don't panic on bad data in quota validations

mdbooth:
Fixes https://bugzilla.redhat.com/1907822

### DIFF
--- a/pkg/asset/quota/openstack/openstack.go
+++ b/pkg/asset/quota/openstack/openstack.go
@@ -1,6 +1,9 @@
 package openstack
 
 import (
+	"github.com/sirupsen/logrus"
+
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
 	"github.com/openshift/installer/pkg/asset/installconfig/openstack/validation"
 	"github.com/openshift/installer/pkg/quota"
 	machineapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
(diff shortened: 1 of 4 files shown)


pkg/asset/quota/openstack/openstack.go
```diff
@@ -11,44 +14,90 @@ import (
...
 	}
-	computeReplicas := make([]int64, len(computes))
-	computeConfigs := make([]*openstackprovider.OpenstackProviderSpec, len(computes))
-	for i, m := range computes {
-		computeReplicas[i] = int64(*m.Spec.Replicas)
-		computeConfigs[i] = m.Spec.Template.Spec.ProviderSpec.Value.Object.(*openstackprovider.OpenstackProviderSpec)
+
+	osps, ok := spec.Value.Object.(*openstackprovider.OpenstackProviderSpec)
+	if !ok {
+		logrus.Warnf("ProviderSpec of unexpected type %T", spec.Value.Object)
```
pierreprinetti, line 41: Honestly I'd be fine with panicking here: if the type assertion fails, then the panic makes it very clear that there is a programming error somewhere.

pkg/asset/quota/openstack/openstack.go
```diff
@@ -11,44 +14,90 @@ import (
...
+	}
+	constraints = append(constraints, instanceConstraint(int64(len(computes))))
+
+	return aggregate(constraints)
+}
+
+func getOpenstackProviderSpec(spec *machineapi.ProviderSpec) *openstackprovider.OpenstackProviderSpec {
+	// An invalid ProviderSpec would presumably be a programming error here.
+	if spec.Value == nil {
+		logrus.Warnf("Empty ProviderSpec")
```
pierreprinetti, line 35: A warning makes sense here, now that the platform-level flavor property is deprecated :+1:

pkg/asset/quota/openstack/openstack.go
```diff
@@ -11,44 +14,90 @@ import (
...
-	for _, m := range machines {
-		flavorInfo := ci.Flavors[m.Flavor]
-		ret = append(ret, machineFlavorCoresToQuota(flavorInfo, ci), machineFlavorRAMToQuota(flavorInfo, ci))
+func getFlavor(ci *validation.CloudInfo, flavorName string) *flavors.Flavor {
+	if flavorInfo, ok := ci.Flavors[flavorName]; ok {
+		return flavorInfo.Flavor
 	}
-	ret = append(ret, instanceConstraint(int64(len(machines))))
-	return ret
+	return nil
```
pierreprinetti, line 52: Are you trying to fix Go? Please don't :laughing:

pkg/asset/quota/openstack/openstack.go
```diff
@@ -11,44 +14,90 @@ import (
...
-		ramConstraint.Count = ramConstraint.Count * replicas[idx]
-		ret = append(ret, instanceConstraint(int64(replicas[idx])), coresConstraint, ramConstraint)
+func machineConstraints(ci *validation.CloudInfo, machine *machineapi.Machine) []quota.Constraint {
+	osps := getOpenstackProviderSpec(&machine.Spec.ProviderSpec)
+	if osps == nil {
+		logrus.Warnf("Skipping quota validation for Machine %s: Invalid ProviderSpec", machine.Name)
+		return nil
+	}
+
+	flavor := getFlavor(ci, osps.Flavor)
```
pierreprinetti, line 62: Please use Go's `, ok` syntax, this function call is making the code harder to follow

pkg/asset/quota/openstack/openstack.go
```diff
@@ -11,44 +14,90 @@ import (
...
+
+	// This will result in a separate validation failure
+	if flavor == nil {
+		logrus.Warnf("Skipping quota validation for Machine %s: Flavor '%s' is not valid",
+			machine.Name, osps.Flavor)
+		return nil
```
pierreprinetti, line 68: this block goes into the `!ok` arm of the standard Go map lookup:

```Go
var flavor *flavors.Flavor
if flavor, ok := ci.Flavors[flavorName]; !ok {
	logrus.Warnf("Skipping quota validation for Machine %s: Flavor '%s' is not valid",
	machine.Name, osps.Flavor)
	return nil
}
return []quota.Constraint{machineFlavorCoresToQuota(flavor), machineFlavorRAMToQuota(flavor)}
```

pkg/asset/quota/openstack/openstack.go
```diff
@@ -11,44 +14,90 @@ import (
...
+
+		logrus.Warnf("Skipping quota validation for MachineSet %s due to unspecified replica count", ms.Name)
+		return nil
+	}
+
+	flavor := getFlavor(ci, osps.Flavor)
+	if flavor == nil {
+		logrus.Warnf("Skipping quota validation for MachineSet %s: Flavor '%s' is not valid", ms.Name, osps.Flavor)
+		return nil
+	}
```
pierreprinetti, line 93:
```Go
var flavor *flavors.Flavor
if flavor, ok := ci.Flavors[osps.Flavor]; !ok {
	logrus.Warnf("Skipping quota validation for MachineSet %s: Flavor '%s' is not valid", ms.Name, osps.Flavor)
	return nil
}
```

pkg/asset/quota/openstack/openstack.go
```diff
@@ -11,44 +14,90 @@ import (
...
-		ctrplConfigs[i] = m.Spec.ProviderSpec.Value.Object.(*openstackprovider.OpenstackProviderSpec)
+	var constraints []quota.Constraint
+
+	for i := 0; i < len(controlPlanes); i++ {
+		constraints = append(constraints, machineConstraints(ci, &controlPlanes[i])...)
+	}
+	constraints = append(constraints, instanceConstraint(int64(len(controlPlanes))))
+
+	for i := 0; i < len(computes); i++ {
+		constraints = append(constraints, machineSetConstraints(ci, &computes[i])...)
```
pierreprinetti, line 25: Please help me understand.

Let's imagine the case of 3 MachineSets with one replica each, and each replica has a flavor that needs 16GB.

I am figuring the resulting `[]quota.Constraint`:
```Go
[]quota.Constraint{
  {RAM, 16},
  {RAM, 16},
  {RAM, 16},
}
```

So the checker will check three times that we have 16 GB total. Instead, I expected  `[]quota.Constraint` to look like this:

```Go
[]quota.Constraint{
  {RAM, 16*3},
}
```

What am I missing?

mdbooth: You're missing the call to aggregate(constraints) on line 29 below. The aggregate call was already there, and I think it simplifies the code a bit. Basically it adds up all constraints against the same resource.

pkg/asset/quota/openstack/openstack.go
```diff
@@ -11,44 +14,90 @@ import (
...
+	}
+	constraints = append(constraints, instanceConstraint(int64(len(computes))))
+
+	return aggregate(constraints)
```
mdbooth, line 29: 👆 This adds everything up.

pkg/asset/quota/openstack/openstack.go
mdbooth: Panic with a specific error message (i.e. turn this into Errorf), or just remove this checking entirely and leave it to fail on its own?

pkg/asset/quota/openstack/openstack.go
mdbooth: If that's more idiomatic, then sure. How about using an else rather than defining the variable outside of the scope of the if statement? They're both a bit ugly, tbh.

pkg/asset/quota/openstack/openstack.go
mdbooth: Ack.

pkg/asset/quota/openstack/openstack.go
pierreprinetti: I’d really leave it to fail on its own, ie the good ol’ type assertion panic.

mdbooth: Agree. I've left the spec.Value check in so the 2 different errors can be distinguished, but I've also turned it into a panic and removed further error checking around it.

pkg/asset/quota/openstack/openstack.go
pierreprinetti: Defining the variable in the outer scope lets you apply early returns, which is another fashionable way to show off your Go orthodoxy

pkg/asset/quota/openstack/openstack.go
pierreprinetti: Let me unroll this comment.

IIUC, there are two valid cases:
* `spec.Value` is either `nil`; OR
* `spec.Value.Object` is of type `*openstackprovider.OpenstackProviderSpec`.

As a consequence:
* if `spec.Value` is nil, we accept (and warn) because the properties could be defined in the deprecated properties of the `platform` section;
* if `spec.Value.Object` is not OpenStack, we panic.

In other words: I agree with the code that I commented here (`logrus.Warnf`).

Is that any clearer, and agreeable?

pkg/asset/quota/openstack/openstack.go
mdbooth: This is subtle, but on closer inspection I remember why I did this: flavorInfo.Flavor is also a pointer so therefore may *also* be nil. i.e. checking `ok` here is insufficient to ensure you don't have a nil pointer.

Honestly I think this is an unnecessary landmine in this datastructure, so I'm going to try to change that in another small patch to see what you think.

pierreprinetti (approved):

pkg/asset/quota/openstack/openstack.go
```diff
@@ -11,44 +14,79 @@ import (
 // These constraints can be used to check if there is enough quota for creating a cluster
 // for the install config.
 func Constraints(ci *validation.CloudInfo, controlPlanes []machineapi.Machine, computes []machineapi.MachineSet) []quota.Constraint {
-	ctrplConfigs := make([]*openstackprovider.OpenstackProviderSpec, len(controlPlanes))
-	for i, m := range controlPlanes {
-		ctrplConfigs[i] = m.Spec.ProviderSpec.Value.Object.(*openstackprovider.OpenstackProviderSpec)
+	var constraints []quota.Constraint
+
+	for i := 0; i < len(controlPlanes); i++ {
```
iamemilio, line 19:
```suggestion
	for _, controlPlaneMachine := range controlPlanes {
```
why not use the range here?

pkg/asset/quota/openstack/openstack.go
```diff
@@ -11,44 +14,79 @@ import (
...
+		constraints = append(constraints, machineConstraints(ci, &controlPlanes[i])...)
 	}
-	computeReplicas := make([]int64, len(computes))
-	computeConfigs := make([]*openstackprovider.OpenstackProviderSpec, len(computes))
-	for i, m := range computes {
-		computeReplicas[i] = int64(*m.Spec.Replicas)
-		computeConfigs[i] = m.Spec.Template.Spec.ProviderSpec.Value.Object.(*openstackprovider.OpenstackProviderSpec)
+	constraints = append(constraints, instanceConstraint(int64(len(controlPlanes))))
+
+	for i := 0; i < len(computes); i++ {
```
iamemilio, line 24: ^

iamemilio:
/lgtm

Lets add a manifest test for this case

PR #4350, opened 2020-11-05
Add KubeVirt platform as infrastructure for Openshift installation

nirarg:
This PR is part of task done to add KubeVirt platform as infra provider for Openshift.

The change introduced in this PR purpose is to provide a way to install Openshift on KubeVirt infrastructure

For more information:
* The enhancement PR:
https://github.com/openshift/enhancements/pull/417
* The Jira Epic: [OCPCNV-7](https://issues.redhat.com/browse/OCPCNV-7)

Related PRs/changes done as part of this task:
1. Created `openshift/cluster-api-provider-kubevirt` repository, hosts an implementation of a provider for KubeVirt for the OpenShift machine-api: https://github.com/openshift/cluster-api-provider-kubevirt
2. `openshift/machine-api-operator`: https://github.com/openshift/machine-api-operator/pull/716#issuecomment-721567660
3. `openshift/cloud-credential-operator`: https://github.com/openshift/cloud-credential-operator/pull/260/
4. `openshift/machine-config-operator`: https://github.com/openshift/machine-config-operator/pull/2098
5. `openshift/api`: https://github.com/openshift/api/pull/734

Additional job is done to have tests and CI coverage (work in progress in github.com/openshift/release repository)

PR #4484, opened 2020-12-15
Bug 1907947: cloud-creds-secret creation with current context info only

nirarg:


nirarg:
Hi @staebler ,
This PR is related to this comment: https://github.com/openshift/installer/pull/4350#discussion_r535833106

pkg/asset/installconfig/kubevirt/client.go
```diff
@@ -59,9 +60,37 @@ var (
...
+	for k, v := range rawConfig.Contexts {
+		if k != rawConfig.CurrentContext {
+			delete(rawConfig.Contexts, k)
+		} else {
+			currentContextValue = v
+		}
+	}
+	if currentContextValue == nil {
+		return nil, fmt.Errorf("currentContext is not included in rawConfig.Contexts")
+	}
```
staebler, line 81:
```suggestion
	currentContextValue := rawConfig.Contexts[rawConfig.CurrentContext]
	if currentContextValue == nil {
		return nil, fmt.Errorf("currentContext is not included in rawConfig.Contexts")
	}
	rawConfig.Contexts = map[string]*clientcmdapi.Context{
		rawConfig.CurrentContext: currentContextValue,
	}
```

pkg/asset/installconfig/kubevirt/client.go
```diff
@@ -59,9 +60,37 @@ var (
...
+	for k := range rawConfig.Clusters {
+		if k != currentContextValue.Cluster {
+			delete(rawConfig.Clusters, k)
+		}
+	}
```
staebler, line 86:
```suggestion
	if v, ok := rawConfig.Clusters[currentContextValue.Cluster]; ok {
		rawConfig.Clusters = map[string]*clientcmdapi.Cluster{
			currentContextValue.Cluster: v
		}
	}
```
Do a similar thing for AuthInfos, as well.

staebler (commented):
Looks good. Just some nits.

/approve

pkg/asset/installconfig/kubevirt/client.go
nirarg: Changed

pkg/asset/installconfig/kubevirt/client.go
nirarg: Changed

PR #4486, opened 2020-12-16
Bug 1786314: bump gophercloud/utils

Fedosin:
This patch introduces recent optimizations for tls config generation.
1. Now the library ignores the fact the CA certificate can be broken
(if it can't parse it for some reason, it just leaves an empty string
without a notice). From now we raise an error in this case.
2. It tries to prevent some obvious formatting errors in the certificate
like leading and trailing spaces, additional new lines etc.


mandre (approved):
/approve
/lgtm
/hold until tests pass

mandre (commented):
/approve
/hold cancel

mandre:
Hi Matt, could you take a look?
/assign @staebler

PR #4487, opened 2020-12-16
Bug 1908280: update docs for Cinder AZs

Fedosin:
This commit documents the correct value that should be set to enable ignore-volume-az option in the cloud provider.


mandre (approved):
/lgtm
/approve

PR #4492, opened 2020-12-17
Bug 1903055: OpenStack: move platform validations to cluster asset

Fedosin:
Now we perform validations of install-config asset right after its generation. Unfortunately, later this asset is modified during "machines" generation. It leads to the fact that even if the validation succeeds, installation may fail after that.

To prevent this we move the validation to "cluster" asset, because there are no install-config modifications after that.

Also this commit removes defaultMachinePlatform validation, because it's not used for the cluster provisioning and just contains default values for the real machine pools.

### DIFF
--- a/pkg/asset/installconfig/installconfig.go
+++ b/pkg/asset/installconfig/installconfig.go
@@ -14,7 +14,6 @@ import (
 	icazure "github.com/openshift/installer/pkg/asset/installconfig/azure"
 	icgcp "github.com/openshift/installer/pkg/asset/installconfig/gcp"
 	ickubevirt "github.com/openshift/installer/pkg/asset/installconfig/kubevirt"
-	icopenstack "github.com/openshift/installer/pkg/asset/installconfig/openstack"
 	icovirt "github.com/openshift/installer/pkg/asset/installconfig/ovirt"
 	icvsphere "github.com/openshift/installer/pkg/asset/installconfig/vsphere"
 	"github.com/openshift/installer/pkg/types"
@@ -191,9 +190,6 @@ func (a *InstallConfig) platformValidation() error {
 	}
 	if a.Config.Platform.Ovirt != nil {
 		return icovirt.Validate(a.Config)
-	}
-	if a.Config.Platform.OpenStack != nil {
-		return icopenstack.Validate(a.Config)
 	}
 	if a.Config.Platform.Kubevirt != nil {
 		client, err := ickubevirt.NewClient()
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -382,6 +382,11 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			Data:     data,
 		})
 	case openstack.Name:
+		err = openstackconfig.Validate(installConfig.Config)
+		if err != nil {
+			return err
+		}
+
 		cloud, err := openstackconfig.GetSession(installConfig.Config.Platform.OpenStack.Cloud)
 		if err != nil {
 			return errors.Wrap(err, "failed to get cloud config for openstack")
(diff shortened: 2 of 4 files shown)


pierreprinetti:
@Fedosin 

Can you please be more specific as to:
* what parameters are changed in the "machines" step that impact validation
* what is the actual problem that pushed you to make this change

That information might help supporting your case! Thank you.

pkg/asset/installconfig/installconfig.go
```diff
@@ -192,9 +191,6 @@ func (a *InstallConfig) platformValidation() error {
 	if a.Config.Platform.Ovirt != nil {
 		return icovirt.Validate(a.Config)
 	}
-	if a.Config.Platform.OpenStack != nil {
-		return icopenstack.Validate(a.Config)
-	}
```
patrickdillon, line 197: I don't think you want to remove this. Install config validation should be done when creating assets like manifests or ignition configs. Try `create manifests` with this change and see if your validation is catching.

patrickdillon (commented):
Can you explain more about how machines is causing this failure?

Fedosin:
@patrickdillon yeah, it happens here https://github.com/openshift/installer/blob/master/pkg/asset/machines/master.go#L239 when apply defaultMachinePlatform values

If you compare install-config asset before and after "machines" they can be different.

pkg/asset/installconfig/installconfig.go
Fedosin: I agree that logically the validation should be here, but it returns incorrect results because the asset will be modified later.
Therefore I see two solutions here:

1. Move the validation to "cluster" asset where it works properly (this PR).
2. Stop modifying install-config in different places.

patrickdillon:
> @patrickdillon yeah, it happens here https://github.com/openshift/installer/blob/master/pkg/asset/machines/master.go#L239 when apply defaultMachinePlatform values
> 
> If you compare install-config asset before and after "machines" they can be different.

I see your point: the machines asset is modifying its dependent install config asset. That definitely smells to me, but I would need to look at it more closely. I wonder if @staebler could provide some more insight. 

On the other hand, I don't see how that is posing a practical problem/relates to solving this particular bug. The values written by machines should be valid, so they shouldn't need more validation. Do you need the values from the machines to validate the rest of the install config? If you could give an example with the validation problem and the `flavor` value that would help.

staebler:
Couple points here.
1. The line in question is not mutating the install config. It is mutating the `mpool` variable using the values from the install config. If the `Set` function is modifying the install config, then that is an issue with the `Set` function.
2. Any mutations to the install config by another asset's `Generate` function will only mutate the local copy of that install config that is limited to the scope of that `Generate` function. It will not mutate the install config asset held in the asset store.

staebler:
@Fedosin Would you mind sharing the .openshift_install_state.json before and after running `openshift-install create ignition-configs` so that I can get a better feel for exactly what parts of the install config are mutating?

patrickdillon:
> Couple points here.
> 
> 1. The line in question is not mutating the install config. It is mutating the `mpool` variable using the values from the install config. If the `Set` function is modifying the install config, then that is an issue with the `Set` function.
> 2. Any mutations to the install config by another asset's `Generate` function will only mutate the local copy of that install config that is limited to the scope of that `Generate` function. It will not mutate the install config asset held in the asset store.

`mpool` is a reassignment of a machinepool in the installconfig. These are pointers so it seems possible the values in the store could be modified. I have never looked at the asset store implementation to see if/how it handles pointers.

staebler:
The problem is really with https://github.com/openshift/installer/blob/master/pkg/asset/machines/master.go#L241, where `pool` points to the `ControlPlane` `MachinePool` from the install config. I see where the asset fetched from the asset store is not a deep copy of the asset, so there is an implicit contract that the caller that fetches the dependency will not modify the asset. We should fix the master and worker assets so that they do not violate that contract.

PR #4493, opened 2020-12-17
Bug 1882022: installconfig: enable vsphere folders for datacenters and clusters.

jstuever:
This change modifies the vsphere prompt behavior for both datacenter and
cluster inputs. When these items exist within one or more folders in
the vsphere infrastructure, the folder names will be added as a prefix
to the item in the list of choices. As a result, this folder information
will now be saved into the install-config.yaml and used to deploy the
cluster.

### DIFF
--- a/pkg/asset/installconfig/vsphere/vsphere.go
+++ b/pkg/asset/installconfig/vsphere/vsphere.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


jstuever:
Example result (backward compatible)

Vcenter/foo/datacenter-1/bar/cluster-1...
platforms:
  vsphere:
    cluster: bar/cluster-1
    datacenter: foo/datacenter-1

Vcenter/datacenter-1/cluster-1...
platforms:
  vsphere:
    cluster: cluster-1
    datacenter: datacenter-1

pkg/asset/installconfig/vsphere/vsphere.go
```diff
@@ -217,13 +218,17 @@ func getCluster(ctx context.Context, path string, finder *find.Finder, client *v
...
+		name = strings.TrimLeft(name, "/host/")
+		logrus.Infof("Defaulting to only available cluster: %s", name)
+		return name, nil
 	}
 
 	var clusterChoices []string
 	for _, c := range clusters {
-		clusterChoices = append(clusterChoices, c.Name())
+		name := strings.TrimLeft(c.InventoryPath, path)
+		name = strings.TrimLeft(name, "/host/")
```
patrickdillon, line 230: nit: I think `[TrimPrefix](https://golang.org/pkg/strings/#TrimPrefix)` might be more appropriate here (and elsewhere), and the trim could be consolidated into a single call.

patrickdillon (approved):
Prior to this change, the survey displayed/selected the names of the datacenters and clusters and this changes to display/select the path. The path is needed in the case of nested resources.

/approve

This LGTM. I left one nitpicky comment so I'll give a chance to respond or let @jcpowermac or @rvanderp3  take a look before I give `/lgtm`

pkg/asset/installconfig/vsphere/vsphere.go
staebler: This **must** use `TrimPrefix` instead. Otherwise, you will end up trimming more than you intend to. For example, if the name is something like `/host/hello`, it will trim down to `ello` instead of `hello`.

https://play.golang.org/p/Xf8oto1noG7

jstuever: That explains a lot! Thanks. Fixed in https://github.com/openshift/installer/commit/0afab1b4bcac64773a266242d563cc7bb94fde50

staebler (approved):
/approve

Issue #1007, opened 2019-01-07
libvirt: Unable to access web console


rhopp:
# Version

```console
$ openshift-install version
v0.9.0-master
```
(compiled from master)

# Platform (aws|libvirt|openstack):

libvirt

# What happened?
I'm trying to install openshift 4 using this installer. It seems, that everything was OK. I've done all the steps described in [here](https://github.com/openshift/installer/blob/master/docs/dev/libvirt-howto.md). Installation was ok, I was able to login using `oc` with credentials from the installation output, but I'm not able to access web console.

Looking at `openshift-console` project, everything seems ok:
<details>
<summary>OUTPUT</summary>

```
╭─rhopp@dhcp-10-40-4-106 ~/go/src/github.com/openshift/installer  ‹master*› 
╰─$ oc project openshift-console
Already on project "openshift-console" on server "https://test1-api.tt.testing:6443".
╭─rhopp@dhcp-10-40-4-106 ~/go/src/github.com/openshift/installer  ‹master*› 
╰─$ oc get all
NAME                                     READY     STATUS    RESTARTS   AGE
pod/console-operator-79b8b8cb8d-cgpfn    1/1       Running   1          1h
pod/openshift-console-6ddfcc76b5-2kmpx   1/1       Running   0          1h
pod/openshift-console-6ddfcc76b5-sp5zm   1/1       Running   0          1h
pod/openshift-console-6ddfcc76b5-z52hq   1/1       Running   0          1h

NAME              TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)   AGE
service/console   ClusterIP   172.30.198.57   <none>        443/TCP   1h

NAME                                DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/console-operator    1         1         1            1           1h
deployment.apps/openshift-console   3         3         3            3           1h

NAME                                           DESIRED   CURRENT   READY     AGE
replicaset.apps/console-operator-79b8b8cb8d    1         1         1         1h
replicaset.apps/openshift-console-6ddfcc76b5   3         3         3         1h

NAME                               HOST/PORT                                         PATH      SERVICES   PORT      TERMINATION          WILDCARD
route.route.openshift.io/console   console-openshift-console.apps.test1.tt.testing             console    https     reencrypt/Redirect   None
```
</details>

The pods are running, service and route are up, but accessing https://console-openshift-console.apps.test1.tt.testing in browser says it couldn't resolve IP address.

[...]

PR #4494, opened 2020-12-18
Bug 1912571: libvirt: support setting network dnsmasq options through the install config

Prashanth684:
Since libvirt 5.6.0, there is an option to pass in dnsmasq options through the libvirt network [1]. This addresses the following problems:

- eliminate the need for hacking routes in the cluster (the workaround mentioned in [3]) so that libvirt's dnsmasq does not manage the domain (and so the requests from inside the cluster will go up the chain to the host itself).
- eliminate the hacky workaround used in the multi-arch CI automation to inject `*.apps` entries in the libvirt network that point to a single worker node [2]. Instead of waiting for the libvirt networks to come up and update entries, we can set this before the installation itself through the install config.
- another issue this solves - with the above mentioned workaround, having multiple worker nodes becomes problematic when running upgrade tests. Having the route to just one worker node would fail the upgrade when that worker node is down. With this change, we could now point to the .1 address and have a load balancer forward traffic to any worker node.

With this change, the option can be specified through the install config yaml in the network section as pairs of option name and values. An example:
 ```
    platform:
      libvirt:
        network:
          dnsmasqOptions:
          - name: "address"
            value: "/.apps.tt.testing/192.168.126.51"
          if: tt0
 ```
The terraform provider supports rendering these options through a datasource and injecting them into the network xml. Since this config is optional, not specifying it will continue to work as before without issues.

[1] https://libvirt.org/formatnetwork.html#elementsNamespaces
[2] https://github.com/openshift/release/blob/master/ci-operator/templates/openshift/installer/cluster-launch-installer-remote-libvirt-e2e.yaml#L532-L554
[3] https://github.com/openshift/installer/issues/1007


data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -1479,6 +1479,13 @@ spec:
                   network:
                     description: Network
                     properties:
+                      dnsmasqoptions:
```
EmilienM, line 1482: don't we want to name it ```dnsmasqOptions``` ?

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -1479,6 +1479,13 @@ spec:
...
+                        additionalProperties:
+                          type: string
+                        description: DnsmasqOptions is the dnsmasq options to be used
```
EmilienM, line 1485: wrongly named here

pkg/tfvars/libvirt/libvirt.go
```diff
@@ -18,29 +18,42 @@ import (
...
+	BootstrapIP     string            `json:"libvirt_bootstrap_ip,omitempty"`
+	MasterMemory    string            `json:"libvirt_master_memory,omitempty"`
+	MasterVcpu      string            `json:"libvirt_master_vcpu,omitempty"`
+	BootstrapMemory int               `json:"libvirt_bootstrap_memory,omitempty"`
+	MasterDiskSize  string            `json:"libvirt_master_size,omitempty"`
+	DnsmasqOptions  map[string]string `json:"libvirt_dnsmasq_options,omitempty"`
+}
+
+// TFVarsSources contains the parameters to be converted into Terraform variables
+type TFVarsSources struct {
```
EmilienM, line 34: :+1: for moving that to a struct, it makes it easily readable.

EmilienM:
small nit wrt naming, otherwise lgtm.

data/data/install.openshift.io_installconfigs.yaml
Prashanth684: yes correct. thanks for that catch. will update

data/data/libvirt/variables-libvirt.tf
```diff
@@ -58,3 +58,9 @@ variable "libvirt_master_size" {
   description = "Size of the volume in bytes"
   default     = "17179869184"
 }
+
+variable "libvirt_dnsmasq_options" {
+  type        = map(string)
+  description = "Dnsamasq options to be applied to the libvirt network"
```
cfergeau, line 64: `Dnsmasq` (extra 'a' in the description)

data/data/libvirt/variables-libvirt.tf
```diff
@@ -61,6 +61,6 @@ variable "libvirt_master_size" {
 
 variable "libvirt_dnsmasq_options" {
   type        = map(string)
-  description = "Dnsamasq options to be applied to the libvirt network"
+  description = "Dnsmasq options to be applied to the libvirt network"
```
cfergeau, line 64: Can this be squashed in `libvirt: add support for dnsmasq options in the terraform files` rather than being changed in ` libvirt: support setting network dnsmasq options through the install … ` ?

cfergeau (approved):
Looks good to me apart from a few minor comments. Not very familiar with the terraform plumbing, but since this code has been tested, it's fine :)

data/data/libvirt/variables-libvirt.tf
Prashanth684: yeah that would be cleaner..i was just being lazy :) . will do.

pkg/types/libvirt/platform.go
```diff
@@ -33,4 +33,11 @@ type Network struct {
 	// +kubebuilder:default="tt0"
 	// +optional
 	IfName string `json:"if,omitempty"`
+
+	// DnsmasqOptions is the dnsmasq options to be used when installing on
+	// libvirt. This is useful to force domains to a specific IP or to a
+	// loadbalancer.
+	//
+	// +optional
+	DnsmasqOptions map[string]string `json:"dnsmasqOptions,omitempty"`
```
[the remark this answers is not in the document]
Prashanth684: oh yes..will change the commit message thanks.

staebler, line 42: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

staebler (commented):
This is a feature, not a bug. Please wait until the master branch opens.

Prashanth684:
> This is a feature, not a bug. Please wait until the master branch opens.

Ok. that makes sense. I'll remove the ref to the BZ and close it. thanks.

pkg/types/libvirt/platform.go
Prashanth684: so reading the comments in the link, it looks like we prefer named subobjects over maps for better ux and readability. would something like this suffice? 

```
platform:
     libvirt:
       network:
         dnsmasqOptions:
         - optionName: "address"
           optionValue: "/.apps.tt.testing/192.168.126.51"
         if: tt0
```

staebler: Yes, although the "option" part of `optionName` and `optionValue` are a bit redundant with `dnsmasqOptions`. How would you feel about simply `name` and `value`?

Prashanth684: makes sense, respun with the changes.

andymcc:
LGTM - one thing we should probably add is a note somewhere (probably the libvirt-dev install docs) to say that in order to use this setting you need libvirt version > 5.6.0

Prashanth684:
> LGTM - one thing we should probably add is a note somewhere (probably the libvirt-dev install docs) to say that in order to use this setting you need libvirt version > 5.6.0

ah yes thanks @andymcc for pointing out. i added a section in the docs specifying the way to specify the dnsmasq option through the install config for libvirt 5.6.0+

Prashanth684:
tagging it back to Bug 1912571 even though it's not a bug per se just to keep track of it.

Prashanth684:
@staebler now that master is open, any objections to approving this ?

staebler:
> @staebler now that master is open, any objections to approving this ?

Nope.

/approve

PR #4496, opened 2020-12-21
Bug 1908171: fix Terraform issue with GCP custom machine types

patrickdillon:
Similar to previous bug (see below) but in this case for custom machine types. In the previous bz, the config generated by Terraform did not match the resource that was ultimately created. So when apply is run a second time for bootstrap destroy, an error occurs when TF tries to switch the resource back to match the config.

Similar: fdefcca25cb12fc241198b0688f421411c9a580a

The error message states:
```
"Error: Changing the machine_type, min_cpu_platform, service_account, or enable display on a started instance requires stopping it.
```
We could potentially add `service_account` and `enable display` (however that is handled in TF) to prevent this error from happening in those cases, but I chose to opt for the minimal change and wait and see if those become a problem.


jstuever (approved):
/lgtm

staebler (commented):
/approve

PR #4497, opened 2020-12-21
Add controlPlaneTopology to infrastructure manifest

ravidbro:
Implements enhancement https://github.com/openshift/enhancements/pull/555

Depends on https://github.com/openshift/api/pull/827

Signed-off-by: Ravid Brown <ravid@redhat.com>


pkg/asset/manifests/infrastructure.go
```diff
@@ -80,6 +80,12 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
 		},
 	}
 
+	if *installConfig.Config.ControlPlane.Replicas == 1 {
```
dhellmann, line 83: Both `ControlPlane` and `Replicas` are pointers. Is it possible for either to be `nil` at this point in the code?

In the enhancement, @hexfusion [points out](https://github.com/openshift/enhancements/pull/560/files#r553001390) that we may want this check to be `< 3` instead of `== 1`. I didn't find any validation that prevents a user from setting replicas to 2, for example. Maybe we're guarded in a way that I just didn't find?

ravidbro: Did we decide what the desired behavior of the cluster in the case of 2 control nodes is?
Are we expecting all the operators to behave the same in the case of 1 node or 2 nodes?
according to https://github.com/openshift/enhancements/pull/555#discussion_r553113665 if we will change the enum to something like 'SingleNode' then it explicitly one and not two, 
We need to get to a conclusion in the enhancement before updating the code here.

pkg/asset/manifests/infrastructure.go
```diff
@@ -80,6 +80,27 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
...
+
+	numOfWorkers := int64(0)
+	for _, mp := range installConfig.Config.Compute {
+		if mp.Replicas != nil {
+			numOfWorkers += *mp.Replicas
+		}
+	}
+	switch numOfWorkers {
+	case 0:
+		config.Status.InfrastructureTopology = config.Status.ControlPlaneTopology
```
wking, line 97: Is this:

> We expect the "infrastructure" workload to be schedulable on the control-plane nodes in this compact cluster.

?  Because in that case it may be worth switching on `schedulers.config.openshift.io`'s `mastersSchedulable`.

wking: Although maybe in the `mastersSchedulable` case, operators that would usually schedule their operands should look at `mastersSchedulable`, realize they were being allowed onto the control-plane machines, and, based on [the API docs][1] consume `controlPlaneTopology` instead of `infrastructureTopology`?

[1]: https://github.com/openshift/api/pull/827/files#diff-07b264a49084976b670fb699badaca1795027d6ea732a99226a5388104f6174fR88-R9

romfreiman: It's already happens. If you install only masters nodes, there are schedualable

staebler: https://github.com/openshift/installer/blob/master/pkg/asset/manifests/scheduler.go#L64-L74

staebler (commented):
/approve

pkg/asset/manifests/infrastructure.go
wking: Right.  My concern is that you are assuming that the `scheduler.go` code is going to keep doing what it's doing today.  I'd rather have this code look at `Scheduler.Spec.MastersSchedulable`, so it didn't have to make assumptions about `scheduler.go`.  And I'm still not sure if `infrastructureTopology` matters at all, or who it controls, if `mastersSchedulable` is `true`.

staebler: That's not a bad suggestion. It would be nice to de-duplicate the logic. I think that I would rather have a common function that the two assets could each call to determine whether the masters should be considered schedulable as opposed to having the Infrastructure asset carry a dependency on the Scheduler asset, though.

staebler:
The OpenStack tests are failing due to rotated creds. The change in this PR do not affect the success of th eopenshift-manifests test.

/skip

PR #4344, opened 2020-11-04
Bug 1894678: Cherry-Pick: Patch Flavor Not Found validation for OpenStack Install Config

iamemilio:
Flavors are always added to the Flavors map, however if they are not found
the flavor pointer is nil. Corrected the code logic to account for this.

Reference: https://github.com/openshift/installer/pull/4289
JIRA: [OCPBUGSM-20190](https://issues.redhat.com/browse/OCPBUGSM-20190)

PR #4505, opened 2020-12-22
Bug 1894678: Patch Flavor Not Found validation for OpenStack Install Config

pierreprinetti:
Flavors are always added to the Flavors map, however if they are not found
the flavor pointer is nil. Corrected the code logic to account for this.

Bug 1894678
Backports https://github.com/openshift/installer/pull/4289

### DIFF
--- a/pkg/asset/installconfig/openstack/validation/cloudinfo.go
+++ b/pkg/asset/installconfig/openstack/validation/cloudinfo.go
@@ -77,17 +77,21 @@ func (ci *CloudInfo) collectInfo(ic *types.InstallConfig) error {
 		return errors.Wrap(err, "failed to fetch external network info")
 	}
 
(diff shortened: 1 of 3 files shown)


pierreprinetti:
This is a rebase of https://github.com/openshift/installer/pull/4344 on top of the current `release-4.6` branch.
/cc iamemilio

markmc:
(patch manager) it seems this is related to #4383 and resolving an issue with flavor validation that is impacting a number of real-world users

EmilienM:
> (patch manager) it seems this is related to #4383 and resolving an issue with flavor validation that is impacting a number of real-world users

thanks Mark, FYI this is related to https://bugzilla.redhat.com/show_bug.cgi?id=1899161 and we'll have another backport (last one) after this PR is merged.

PR #4508, opened 2020-12-22
Bug 1899161: openstack: consider volumes for storage requirements checks

EmilienM:
It is possible to boot a server without using ephemeral storage, with
boot from volume:
https://docs.openstack.org/cinder/latest/admin/blockstorage-boot-from-volume.html

This patch will disable flavor storage check if rootVolume is used in
the Machine Pool and checks that the volume is at least 25Gb sized.

https://bugzilla.redhat.com/show_bug.cgi?id=1899161

(cherry picked from commit 1b400b48f8face099ff9eeed92266ee32edf106e)

### DIFF
--- a/docs/user/openstack/README.md
+++ b/docs/user/openstack/README.md
@@ -73,19 +73,19 @@ openstack quota set --secgroups 8 --secgroup-rules 100 <project>`
 
 ### Master Nodes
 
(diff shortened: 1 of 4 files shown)


pierreprinetti (approved):

russellb:
(patch manager) 4.6 blocker identified, so pulling unmerged patches from the queue

ashcrow:
https://github.com/openshift/installer/pull/4323

PR #4516, opened 2020-12-31
Bug 1911819: Add kubevirt user documentation

nirarg:


staebler:
@nirarg This PR broke the yaml-lint check. Can you fix up the `docs/user/kubevirt/install-config.yaml` file so that it passes `./hack/yaml-lint.sh`?

PR #4517, opened 2021-01-04
WIP: Moved oVirt installconfig to externalprovider

janoszen:
This is the first PR of many as detailed in openshift/enhancements/pull/566. This PR is WIP and intended as a basis for discussion and will be amended later.

### DIFF
--- a/pkg/externalprovider/provider/base.go
+++ b/pkg/externalprovider/provider/base.go
@@ -4,4 +4,7 @@ package provider
 type ExternalProvider interface {
 	// Name returns the name of the external provider.
 	Name() string
+
+	// InstallConfigExternalProvider contains the installconfig-specific steps.
+	InstallConfigExternalProvider
 }
--- a/pkg/externalprovider/provider/base.go
+++ b/pkg/externalprovider/provider/base.go
@@ -1,0 +1,7 @@
+package provider
+
+// ExternalProvider is a provider that is not maintained by the core installer team.
+type ExternalProvider interface {
+	// Name returns the name of the external provider.
+	Name() string
+}
--- a/pkg/externalprovider/provider/installconfig.go
+++ b/pkg/externalprovider/provider/installconfig.go
@@ -27,4 +27,12 @@ type InstallConfigExternalProvider interface {
 		AWS *aws.Metadata,
 		Azure *icazure.Metadata,
 	) error
+
+	// PlatformPermsCheck validates the platform permissions.
+	PlatformPermsCheck(
+		Config *types.InstallConfig,
+		File *asset.File,
+		AWS *aws.Metadata,
+		Azure *icazure.Metadata,
+	) error
 }
--- a/pkg/externalprovider/provider/installconfig.go
+++ b/pkg/externalprovider/provider/installconfig.go
@@ -35,4 +35,12 @@ type InstallConfigExternalProvider interface {
 		AWS *aws.Metadata,
 		Azure *icazure.Metadata,
 	) error
+
+	// PlatformProvisionCheck validates the if provisioning can commence on the platform.
+	PlatformProvisionCheck(
+		Config *types.InstallConfig,
+		File *asset.File,
+		AWS *aws.Metadata,
+		Azure *icazure.Metadata,
+	) error
 }
--- a/pkg/externalprovider/provider/installconfig.go
+++ b/pkg/externalprovider/provider/installconfig.go
@@ -19,4 +19,12 @@ type InstallConfigExternalProvider interface {
 		AWS *aws.Metadata,
 		Azure *icazure.Metadata,
 	) error
+
+	// PlatformCredsCheck validates the platform credentials.
+	PlatformCredsCheck(
+		Config *types.InstallConfig,
+		File *asset.File,
+		AWS *aws.Metadata,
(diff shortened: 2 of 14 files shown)


Gal-Zaidman (commented):
Code looks good

pkg/externalprovider/provider/installconfig.go
```diff
@@ -0,0 +1,11 @@
+package provider
+
+import (
+	"github.com/openshift/installer/pkg/types"
+)
+
+// InstallConfigExternalProvider describes the methods required for the installconfig asset.
+type InstallConfigExternalProvider interface {
```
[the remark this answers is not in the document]
janoszen: I'm up for ideas instead of "External Provider", I'd like to wait for some feedback from the installer team on this.

pkg/externalprovider/simpleapi.go
```diff
@@ -0,0 +1,6 @@
+package externalprovider
+
+var providerRegistry = NewRegistry()
+
```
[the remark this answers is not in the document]
janoszen: Ideally, at some point this API should be removed, but that's up in the air.

pkg/externalprovider/simpleapi.go
```diff
@@ -19,3 +22,20 @@ func AddToInstallConfigPlatform(
 	provider := providerRegistry.MustGet(string(ProviderName))
 	return provider.AddToInstallConfigPlatform(p)
 }
+
+// ValidateInstallConfig validates the install config.
+func ValidateInstallConfig(
+	ProviderName Name,
+	Config *types.InstallConfig,
+	File *asset.File,
```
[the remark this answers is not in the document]
janoszen: The API is designed to 1:1 replace the current switch-cases with minimal changes to the code. Adding parameters to existing APIs is harder than adding new API calls.

pkg/externalprovider/provider/base.go
```diff
@@ -0,0 +1,10 @@
+package provider
+
+// ExternalProvider is a provider that is not maintained by the core installer team.
```
patrickdillon, line 3: Perfect example of [Conway's Law](https://en.wikipedia.org/wiki/Conway%27s_law)

I don't think the providers should be grouped based on installer team/non-installer team.

pkg/externalprovider/provider/installconfig.go
```diff
@@ -0,0 +1,46 @@
...
+		AWS *aws.Metadata,
+		Azure *icazure.Metadata,
+	) error
+
+	// PlatformCredsCheck validates the platform credentials.
+	PlatformCredsCheck(
+		Config *types.InstallConfig,
+		File *asset.File,
+		AWS *aws.Metadata,
+		Azure *icazure.Metadata,
```
patrickdillon, line 28: It would make more sense to just have this signature match the dependencies of the asset being genereated:
https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/platformcredscheck.go#L37

oVirt should never have a function that takes `*aws.Metadata` as a parameter.

pkg/externalprovider/provider/base.go
janoszen: The API is designed to be consumable by all providers, but I'm pragmatic about the adoption.

pkg/externalprovider/provider/installconfig.go
janoszen: That is true, but the API is designed to be adopted by all providers, so I added all parameters that are used by any provider at this time. Adding new methods is easier than adding new parameters to existing methods.

patrickdillon: IMHO my suggestion would make it easier and more ergonomic for this API to be adopted by all providers. Also, I meant this as a general suggestion. The one function I picked here is just an example, but this should be applied to all platform interface functions related to an asset. By mapping the parameters to the asset dependencies you ensure that every platform would have all needed resources. It would then be the responsibility of the provider to extract those resources from the dependencies, e.g. oVirt would have *aws.Metadata available but could ignore it

janoszen: Ideally, we would pass on the install config directly, but if we do that we have a circular dependency because the IC structure is in the same package as the code needing to call the new API.

patrickdillon: Moving the interface to the types package would resolve the dependency right? IMHO it is better if asset package does not depend on platforms.

patrickdillon: Ah, I see how the circular dependency is a problem now.  I think there should be a way around the circular dependency by having a common interface. The asset package should not really need to depend on the external providers, and I think that is a benefit of the approach. I think moving the definition of the interface to the asset package could resolve this, but it is a bit too much for me to map it all out now.

janoszen: The main issue is this piece of code:

```go
func init() {
	providerRegistry.Register(ovirt.NewOvirtProvider())
}
```

If we can call this from somewhere that is not in the installconfig flow (e.g. some main.go) then that would solve the whole problem as the registry and the API interface could be split into two packages resolving the circular dependency.  However, I don't know what would be a good place to add this.

PR #4519, opened 2021-01-04
Bug 1911819: Fix docs/user/kubevirt/install-config.yaml to pass ./hack/yaml-lint.sh

nirarg:


nirarg:
@staebler this is the fix for `hack/yaml-lint.sh`

staebler:
Please add `Bug 1911819` to the title of the PR.

PR #4522, opened 2021-01-06
Bug 1909587: openstack: Create Control plane nodes sequentially

pierreprinetti:
This patch addresses the OpenStack issue where the Compute module won't
honour the soft-anti-affinity policy when servers are created
concurrently.

### DIFF
--- a/data/data/openstack/masters/main.tf
+++ b/data/data/openstack/masters/main.tf
@@ -40,21 +40,37 @@ resource "openstack_compute_servergroup_v2" "master_group" {
   policies = ["soft-anti-affinity"]
 }
 
-resource "openstack_compute_instance_v2" "master_conf" {
-  name = "${var.cluster_id}-master-${count.index}"
-  count = var.instance_count
+# The master servers are created in three separate resource definition blocks,
+# rather than a single block with a "count" meta-property, because they need to
+# be created sequentially rather than concurrently by Terraform.
+#
+# The reason why they need to be created one at a time is that OpenStack's
+# Compute module is currently unable to honour the "soft-anti-affinity" policy
+# when the servers are created concurrently.
+#
+# We chose to unroll the loop into three instances, because three is the
+# minimum number of required Control plane nodes, as stated in the
+# documentation[1].
+#
+# The expectation is that machine-api-operator will take care of creating any
+# other requested nodes as soon as the deployment is effective, and that a
+# similar workaround is applied for day-2 operations.
+#
+# [1]: https://github.com/openshift/installer/tree/master/docs/user/openstack#master-nodes
+resource "openstack_compute_instance_v2" "master_conf_0" {
+  name = "${var.cluster_id}-master-0"
 
   flavor_id = data.openstack_compute_flavor_v2.masters_flavor.id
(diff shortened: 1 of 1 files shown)


pierreprinetti:
/title WIP: Bug 1909587: openstack: Create Control plane nodes sequentially

This does not apparently work.

mdbooth (approved):
/lgtm

data/data/openstack/masters/main.tf
```diff
@@ -40,21 +40,84 @@ resource "openstack_compute_servergroup_v2" "master_group" {
...
 
-resource "openstack_compute_instance_v2" "master_conf" {
-  name = "${var.cluster_id}-master-${count.index}"
-  count = var.instance_count
+# The master servers are created in three separate resource definition blocks,
+# rather than a single block with a "count" meta-property, because they need to
+# be created sequentially rather than concurrently by Terraform.
+#
+# The reason why they need to be created one at a time is that OpenStack's
+# Compute module is currently unable to honour the "soft-anti-affinity" policy
```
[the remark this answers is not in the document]
pierreprinetti: HM please

mandre:
At first glance this looks OK because we require a minimum of 3 control plane nodes. However we need to confirm that machine API is able to pick the remaining machines and provision the additional nodes in case we set a replica count > 3.
@pierreprinetti Could you share the results of your testing?

pierreprinetti:
@mandre 
Tested with:
```
$ openshift-install version
openshift-install-linux-4.7.0-0.nightly-2021-01-10-070949 4.7.0-0.nightly-2021-01-10-070949
built from commit 69f0bbc18e8c6b1a6e278c54efa2def9b210033a
release image quay.io/openshift-release-dev/ocp-release-nightly@sha256:62768f2e8bb5850c469c1d15f6cd4337bd0f3f7ecb560b009c61e6d36a134fe8
```

The cluster correctly comes up with 5 Control plane nodes when `replicas` is set to 5 in `install-config`.

@staebler We are hardcoding 3 masters in Terraform, to work around a long-standing bad behaviour of OpenStack's Compute node. I believe it would be good to have a couple eyeballs from the Installer team on this.

data/data/openstack/masters/main.tf
```diff
@@ -40,21 +40,84 @@ resource "openstack_compute_servergroup_v2" "master_group" {
...
+# when the servers are created concurrently.
+#
+# We chose to unroll the loop into three instances, because three is the
+# minimum number of required Control plane nodes, as stated in the
+# documentation[1].
+#
+# The expectation is that machine-api-operator will take care of creating any
```
staebler, line 55: The machine-api-operator is not involved in creating control plane nodes. Adding or removing control plane nodes after installation requires manual intervention by the user. However, there are plans to change this in the future.

staebler (commented):
This is acceptable in the short term where there is a strict requirement for 3 master nodes. That requirement will be relaxed in a future release (maybe 4.9, 4.10). We need to come up with a better solution at some point.

data/data/openstack/masters/main.tf
pierreprinetti: cf. https://issues.redhat.com/browse/OSASINFRA-2262

PR #4525, opened 2021-01-07
add script to patch single-node libvirt deployments

eranco74:
Just testing, no need to review


eranco74:
Works in cluster bot:
`launch https://github.com/openshift/installer/pull/4525 aws,single-node`

Your cluster is ready, it will be shut down automatically in ~131 minutes.

PR #4528, opened 2021-01-07
Bug 1908583: Set same additional networks on Bootstrap as Control Plane

iamemilio:
The bootstrap node was only being attached to the private network created by the installer,
causing deployments with more advanced networking topologies that made use of the additional
networking feature to fail. The bootstrap needs to be attached to the same networks as the
control plane nodes in order to guarantee the install will succeed.

Fixes Bug 1908583

/label platform/openstack


Fedosin (approved):
/lgtm
/approve
/hold

mandre:
I would have loved to see this change reflected in the user documentation.
See the first note at https://github.com/openshift/installer/blob/master/docs/user/openstack/customization.md#machine-pools

PR #4530, opened 2021-01-08
Bug 1908468: azure/validation: some memory values are float and not int

jstuever:
This change fixes an issue where the memory values for azure instances were assumed to be integer while some of them were actually floating point values.

### DIFF
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -60,20 +60,20 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 	for _, capability := range *typeMeta.Capabilities {
 
 		if strings.EqualFold(*capability.Name, "vCPUs") {
-			cpus, err := strconv.ParseInt(*capability.Value, 10, 0)
+			cpus, err := strconv.ParseFloat(*capability.Value, 0)
 			if err != nil {
 				return append(allErrs, field.InternalError(fieldPath, err))
 			}
-			if cpus < req.minimumVCpus {
+			if cpus < float64(req.minimumVCpus) {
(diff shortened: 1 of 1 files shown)


pkg/asset/installconfig/azure/validation.go
```diff
@@ -60,20 +60,20 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 	for _, capability := range *typeMeta.Capabilities {
 
 		if strings.EqualFold(*capability.Name, "vCPUs") {
-			cpus, err := strconv.ParseInt(*capability.Value, 10, 0)
+			cpus, err := strconv.ParseFloat(*capability.Value, 0)
 			if err != nil {
 				return append(allErrs, field.InternalError(fieldPath, err))
 			}
-			if cpus < req.minimumVCpus {
+			if cpus < float64(req.minimumVCpus) {
```
staebler, line 67: Should we change `minimumVCpus` and `minimumMemory` to a float rather than casting at comparison time?

jstuever: I debated the same. My concerns were 1) trying to keep the code as similar as possible between the different platforms, and 2) that the output[1] for invalid CPU/Memory would be different showing an int in some platforms and a float in others. I'm happy to use defined floats instead if you think that is more appropriate.

[1] https://github.com/openshift/installer/blob/62703abc863809af7b5bb7a6419b541ac63a3560/pkg/asset/installconfig/azure/validation.go#L77

staebler: Sounds good. Let's leave it as integers.

staebler (approved):
/lgtm
/approve

PR #2270, opened 2019-08-26
Bug 1749816: OpenStack: Validates cluster name length and document known issues

iamemilio:
OpenStack is shipping with some known issues for this release. We are choosing to address them with preventative validation in the installer, and with documentation so that customers don't mistakenly deploy buggy clusters. 

https://bugzilla.redhat.com/show_bug.cgi?id=1749816

PR #4537, opened 2021-01-12
Bug 1915460: ovirt: validate cluster name during install

dougsland:
Similar to OpenStack platform, we require to
validate cluster name size to avoid DNS resolution
issues in the deploy.

See also: https://github.com/openshift/installer/pull/2270/commits

Signed-off-by: Douglas Schilling Landgraf <dougsland@redhat.com>

### DIFF
--- a/pkg/asset/installconfig/clustername.go
+++ b/pkg/asset/installconfig/clustername.go
@@ -36,6 +36,12 @@ func (a *clusterName) Generate(parents asset.Parents) error {
 			return validate.ClusterName1035(ans.(string))
 		})
 	}
+	if platform.Ovirt != nil {
+		// FIX-ME: As soon bz#1915122 get resolved remove the limitation of 14 chars for the clustername
+		validator = survey.ComposeValidators(validator, func(ans interface{}) error {
+			return validate.ClusterNameMaxLength(ans.(string), 14)
+		})
+	}
 	validator = survey.ComposeValidators(validator, func(ans interface{}) error {
 		installConfig := &types.InstallConfig{BaseDomain: bd.BaseDomain}
 		installConfig.ObjectMeta.Name = ans.(string)
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -70,6 +70,10 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
 	if c.Platform.GCP != nil || c.Platform.Azure != nil {
 		nameErr = validate.ClusterName1035(c.ObjectMeta.Name)
 	}
+	if c.Platform.Ovirt != nil {
+		// FIX-ME: As soon bz#1915122 get resolved remove the limitation of 14 chars for the clustername
+		nameErr = validate.ClusterNameMaxLength(c.ObjectMeta.Name, 14)
+	}
 	if nameErr != nil {
 		allErrs = append(allErrs, field.Invalid(field.NewPath("metadata", "name"), c.ObjectMeta.Name, nameErr.Error()))
 	}
--- a/pkg/validate/validate.go
+++ b/pkg/validate/validate.go
@@ -115,13 +115,23 @@ func ClusterName1035(v string) error {
 	return ClusterName(v)
 }
 
+// ClusterNameMaxLength validates if the string provided length is
+// greater than maxlen argument.
+func ClusterNameMaxLength(v string, maxlen int) error {
+	if len(v) > maxlen {
+		return errors.New(validation.MaxLenError(maxlen))
+	}
+	return nil
+}
+
 // ClusterName checks if the given string is a valid name for a cluster and returns an error if not.
 // The max length of the DNS label is `DNS1123LabelMaxLength + 9` because the public DNS zones have records
 // `api.clustername`, `*.apps.clustername`, and *.apps is rendered as the nine-character \052.apps in DNS records.
 func ClusterName(v string) error {
-	maxlen := validation.DNS1123LabelMaxLength - 9
-	if len(v) > maxlen {
-		return errors.New(validation.MaxLenError(maxlen))
+	const maxlen = validation.DNS1123LabelMaxLength - 9
+	err := ClusterNameMaxLength(v, maxlen)
+	if err != nil {
+		return err
 	}
 	return validateSubdomain(v)
 }


dougsland:
doesnt seems related to the patch
/retest

pkg/asset/installconfig/ovirt/validation.go
```diff
@@ -24,9 +24,11 @@ func Validate(ic *types.InstallConfig) error {
 			"validation requires a Engine platform configuration").Error())
 	}
 
+	testConfig := types.InstallConfig{}
+	testConfig.ObjectMeta.Name = "test"
 	allErrs = append(
 		allErrs,
-		validation.ValidatePlatform(ic.Platform.Ovirt, ovirtPlatformPath)...)
+		validation.ValidatePlatform(ic.Platform.Ovirt, ovirtPlatformPath, &testConfig)...)
```
staebler, line 31: Don't you want to validate the **actual** install config here?
```suggestion
		validation.ValidatePlatform(ic.Platform.Ovirt, ovirtPlatformPath, ic)...)
```

Outside of the scope of this PR, this validation does not even need to be done here. This duplicates validation that has already been done [here](https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/installconfig.go#L149).

pkg/types/ovirt/validation/platform.go
```diff
@@ -1,14 +1,18 @@
...
 import (
+	"fmt"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
+	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/ovirt"
 	"github.com/openshift/installer/pkg/validate"
 )
 
+const maxCharsName int = 14
```
staebler, line 12:
```suggestion
const maxCharsName = 14
```

pkg/asset/installconfig/ovirt/validation.go
dougsland: oh okay.

staebler: This has been marked as resolved, but the code is still using a test InstallConfig instead of the real InstallConfig.

pkg/types/ovirt/validation/platform.go
```diff
@@ -1,14 +1,18 @@
...
 import (
+	"fmt"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
+	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/ovirt"
 	"github.com/openshift/installer/pkg/validate"
 )
 
+const maxCharsName = 14
```
staebler, line 12: Can you add a comment explaining why the limit is 14 characters?

pkg/types/ovirt/validation/platform.go
```diff
@@ -30,5 +34,9 @@ func ValidatePlatform(p *ovirt.Platform, fldPath *field.Path) field.ErrorList {
 	if p.DefaultMachinePlatform != nil {
 		allErrs = append(allErrs, ValidateMachinePool(p.DefaultMachinePlatform, fldPath.Child("defaultMachinePlatform"))...)
 	}
+	if len(c.ObjectMeta.Name) > maxCharsName {
```
staebler, line 37: Looking back over this again, I am struck that this is not the correct place to check this. The `ValidatePlatform` function is meant to validate the `.platform.ovirt` field of the install config. It is not the ideal place to validate other fields. Might it be better to add a call to a new `ovirtvalidation.ValidateClusterName` function in [this](https://github.com/openshift/installer/blob/master/pkg/types/validation/installconfig.go#L69-L72) area?

staebler: I do see that the OpenStack validation has set the precedence for doing it the way that you have done it here. I still suggest that we don't copy that pattern here, though.

dougsland:
@staebler totally agreed. Here a new try.

pkg/types/validation/installconfig.go
```diff
@@ -67,12 +67,20 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
...
-	if c.Platform.GCP != nil || c.Platform.Azure != nil {
+	switch {
+	case c.Platform.GCP != nil:
+	case c.Platform.Azure != nil:
+	case c.Platform.Ovirt != nil:
 		nameErr = validate.ClusterName1035(c.ObjectMeta.Name)
+		if nameErr != nil {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("metadata", "name"), c.ObjectMeta.Name, nameErr.Error()))
+		}
+		nameErr = validate.ClusterNameMaxLong(c.ObjectMeta.Name)
```
staebler, line 78: The 14-character limit should not be enforced for GCP or Azure.

pkg/validate/validate.go
```diff
@@ -107,25 +107,63 @@ func ImagePullSecret(secret string) error {
 // the ClusterName function to confirm compliance with the rest. The resulting name will
 // therefore match RFC1035 with the exception of possible periods ".", which will be
 // translated into dashes "-" in the InfraID before being used to create cloud resources.
-func ClusterName1035(v string) error {
-	re := regexp.MustCompile("^[a-z]")
-	if !re.MatchString(v) {
-		return errors.New("cluster name must begin with a lower-case letter")
+// The optional param m is the max length for the string which the platform support.
+func ClusterName1035(v string, m ...int) error {
```
staebler, line 111: Do not use a variadic function as a way of overloading the function name. Instead, add a new function `ClusterName1035WithMaxLength` and have the existing `ClusterName1035` call the new function.

```
func ClusterName1035(v string) error {
  const maxlen = validation.DNS1123LabelMaxLength - 9
  ClusterName1035WithMaxLength(v, maxlen)
}
```

pkg/validate/validate.go
```diff
@@ -107,25 +107,63 @@ func ImagePullSecret(secret string) error {
...
+	}
+
 	if len(v) > maxlen {
 		return errors.New(validation.MaxLenError(maxlen))
 	}
 	return validateSubdomain(v)
 }
 
+// ClusterStartName checks the cluster name starts with a lower-case letter "^[a-z]"
+func ClusterStartName(v string) error {
```
staebler, line 159: This function is only called internally to the package, is short, and is only called in one place. It does not warrant extracting into a function.

pkg/asset/installconfig/clustername.go
```diff
@@ -36,6 +36,12 @@ func (a *clusterName) Generate(parents asset.Parents) error {
 			return validate.ClusterName1035(ans.(string))
 		})
 	}
+	if platform.Ovirt != nil {
+		/* FIX-ME: As soon bz#1915122 get resolved remove the limitation of 14 chars for the clustername */
```
staebler, line 40: Do not use /**/ comment blocks.
```suggestion
		// FIX-ME: As soon bz#1915122 get resolved remove the limitation of 14 chars for the clustername
```

pkg/types/validation/installconfig.go
```diff
@@ -70,6 +70,10 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
 	if c.Platform.GCP != nil || c.Platform.Azure != nil {
 		nameErr = validate.ClusterName1035(c.ObjectMeta.Name)
 	}
+	if c.Platform.Ovirt != nil {
+		/* FIX-ME: As soon bz#1915122 get resolved remove the limitation of 14 chars for the clustername */
```
staebler, line 74:
```suggestion
		// FIX-ME: As soon bz#1915122 get resolved remove the limitation of 14 chars for the clustername
```

pkg/validate/validate.go
```diff
@@ -107,25 +107,63 @@ func ImagePullSecret(secret string) error {
...
+		if err != nil {
+			return err
+		}
+	} else {
+		err := ClusterName(v)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
staebler, line 131: We should get rid of the variadic function, but for future reference, you do not need to have this conditional logic.
```suggestion
	return ClusterName(v, m...)
```

pkg/validate/validate.go
```diff
@@ -115,13 +115,23 @@ func ClusterName1035(v string) error {
 	return ClusterName(v)
 }
 
+// ClusterName1035WithMaxLength validates if the string provided length is
+// greater than maxlen argument.
+func ClusterName1035WithMaxLength(v string, maxlen int) error {
```
staebler, line 120: This function is not doing RFC-1035 validation. Change the name to reflect what the function is actually doing.

staebler (commented):
/lgtm
/approve

PR #4551, opened 2021-01-14
Bug 1915998: Set Additional Control Plane Security Groups on Bootstrap Node

iamemilio:
The bootstrap node in the installer does not have any way to set additional
networks and security groups. The installer currently adds additional control
plane networks to the bootrap node, and should do the same for security groups.
Since the bootstrap is an ephemeral node that has the same networking requirements
as the master nodes, this logically makes sense and makes the user experience more
consistent.

Fixes: Bug 1915998

/label platform/openstack

### DIFF
--- a/docs/user/openstack/customization.md
+++ b/docs/user/openstack/customization.md
@@ -45,7 +45,7 @@ Beyond the [platform-agnostic `install-config.yaml` properties](../customization
   * `type` (required string): The volume pool to create the volume from.
 * `zones` (optional list of strings): The names of the availability zones you want to install your nodes on. If unset, the installer will use your default compute zone.
 
-**NOTE:** The bootstrap node follows the `type` and `rootVolume` parameters from the `controlPlane` machine pool.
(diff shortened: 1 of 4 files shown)


docs/user/openstack/customization.md
```diff
@@ -192,7 +192,9 @@ controlPlane:
       - fa806b2f-ac49-4bce-b9db-124bc64209bf
 ```
 
-**NOTE:** Allowed address pairs won't be created for the additional networks.
+**NOTES:**
```
mandre, line 195: Can you also update the note in the "Machine Pools" section, on line 48 of this file?

https://github.com/openshift/installer/pull/4551/files#diff-8d13bb62fd33c601f7f64e3a9c861445f547962a0079d20613daab0c07dd73b8R48

mandre (changes requested):

mandre (approved):
/lgtm
/approve

PR #4556, opened 2021-01-15
docs/aws-iam: improving the concepts for Account root user and IAM user

mtulio:
Documentation improvement:

There are a few difference of AWS account root user and AWS IAM user, so my suggestions are:
- improve the initial description based on official documentation
- remove the "secondary" as **AWS account root user** is not AWS IAM user.


jstuever:
/assign @staebler 
Sweet... looks like no tests required since it's docs only.

PR #660, opened 2018-11-13
pkg/destroy/libvirt: Use prefix-based deletion

wking:
To avoid wiping out the caller's whole libvirt environment, regardless of whether it was associated with our cluster or not.  Using cluster-name prefixes still makes me a bit jumpy, so I've added warnings to both the environment-variable and asset-prompt docs warning libvirt users to pick something sufficiently unique.

There are also a number of other libvirt-destroy pivots in separate commits.  I can pull any or all of those out into separate pull requests if that makes review easier.  The only other change which seems user-visible is 926b351, which converts our looping goroutine deletion to a single pass.

Fixes #354.
Fixes #647.
Fixes #656.

PR #4563, opened 2021-01-18
Bug 1916692: OpenStack: Delete leftover LBs when destroying cluster

mandre:
If the user destroyed a cluster without removing all the associated
service LBs, the `destroy` command would fail to remove the network
and loop until it hits the timeout.

The destroy command now looks if there are any leftover LBs where its
`VipNetworkID` matches the network ID and deprovisions it.  We filter on
services LBs created by the openstack cloud provider, matching the
`Kubernetes external service` string in the description [1], to ensure
we're not destroying a user-created resource by mistake.

[1] https://github.com/openshift/kubernetes/blob/442a69c/staging/src/k8s.io/legacy-cloud-providers/openstack/openstack_loadbalancer.go#L446

### DIFF
--- a/pkg/destroy/openstack/openstack.go
+++ b/pkg/destroy/openstack/openstack.go
@@ -646,6 +646,57 @@ func getRouterByPort(client *gophercloud.ServiceClient, allPorts []ports.Port) (
 	return "", nil
 }
 
+func deleteLeftoverLoadBalancers(opts *clientconfig.ClientOpts, logger logrus.FieldLogger, networkID string) {
+	conn, err := clientconfig.NewServiceClient("load-balancer", opts)
+	if err != nil {
+		// Ignore the error if Octavia is not available for the cloud
+		var gerr *gophercloud.ErrEndpointNotFound
+		if errors.As(err, &gerr) {
+			logger.Debug("Skip load balancer deletion because Octavia endpoint is not found")
+			return
+		}
+		logger.Error(err)
+		return
+	}
+
+	listOpts := loadbalancers.ListOpts{
+		VipNetworkID: networkID,
+	}
+	allPages, err := loadbalancers.List(conn, listOpts).AllPages()
+	if err != nil {
+		logger.Error(err)
+		return
+	}
+
+	allLoadBalancers, err := loadbalancers.ExtractLoadBalancers(allPages)
+	if err != nil {
+		logger.Error(err)
+		return
+	}
+	deleteOpts := loadbalancers.DeleteOpts{
+		Cascade: true,
+	}
+	for _, loadbalancer := range allLoadBalancers {
+		if !strings.HasPrefix(loadbalancer.Description, "Kubernetes external service") {
+			logger.Debugf("Not deleting LoadBalancer %q with description %q", loadbalancer.ID, loadbalancer.Description)
+			continue
+		}
+		logger.Debugf("Deleting LoadBalancer %q", loadbalancer.ID)
+		err = loadbalancers.Delete(conn, loadbalancer.ID, deleteOpts).ExtractErr()
+		if err != nil {
+			// Ignore the error if the load balancer cannot be found
+			var gerr gophercloud.ErrDefault404
+			if !errors.As(err, &gerr) {
+				// This can fail when the load balancer is still in use so return/retry
+				logger.Debugf("Deleting load balancer %q failed: %v", loadbalancer.ID, err)
+				return
+			}
+			logger.Debugf("Cannot find load balancer %q. It's probably already been deleted.", loadbalancer.ID)
+		}
+	}
+	return
+}
+
 func isClusterSubnet(subnets []subnets.Subnet, subnetID string) bool {
 	for _, subnet := range subnets {
 		if subnet.ID == subnetID {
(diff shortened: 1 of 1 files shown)


pkg/destroy/openstack/openstack.go
```diff
@@ -653,6 +653,57 @@ func getRouterByPort(client *gophercloud.ServiceClient, allPorts []ports.Port) (
...
+	allLoadBalancers, err := loadbalancers.ExtractLoadBalancers(allPages)
+	if err != nil {
+		logger.Error(err)
+		return false, nil
+	}
+	deleteOpts := loadbalancers.DeleteOpts{
+		Cascade: true,
+	}
+	for _, loadbalancer := range allLoadBalancers {
+		if !strings.HasPrefix(loadbalancer.Description, "Kubernetes external service") {
```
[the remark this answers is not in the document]
mandre: We can't. The in-tree cloud provider [doesn't tag the LB resources](https://github.com/openshift/kubernetes/blob/442a69c/staging/src/k8s.io/legacy-cloud-providers/openstack/openstack_loadbalancer.go#L444-L449).

pkg/destroy/openstack/openstack.go
```diff
@@ -738,6 +789,8 @@ func deleteNetworks(opts *clientconfig.ClientOpts, filter Filter, logger logrus.
 			if !errors.As(err, &gerr) {
 				// This can fail when network is still in use
 				logger.Debugf("Deleting Network %q failed: %v", network.ID, err)
+				// Try to delete eventual leftover load balancers
+				deleteLeftoverLoadBalancers(opts, logger, network.ID)
```
Fedosin, line 793: deleteLeftoverLoadBalancers returns bool and error. are you going to handle it?

pkg/destroy/openstack/openstack.go
```diff
@@ -653,6 +653,57 @@ func getRouterByPort(client *gophercloud.ServiceClient, allPorts []ports.Port) (
...
+		}
+		logger.Error(err)
+		return false, nil
+	}
+
+	listOpts := loadbalancers.ListOpts{
+		VipNetworkID: networkID,
+	}
+	allPages, err := loadbalancers.List(conn, listOpts).AllPages()
+	if err != nil {
```
Fedosin, line 673: I suggest to handle ErrDefault403 errors separately  to prevent situations when a user doesn't have permissions to list load balancers

pkg/destroy/openstack/openstack.go
mandre: I wasn't planning to. I don't want to change the return value of `deleteNetworks()` based on the return of `deleteLeftoverLoadBalancers()`, since this would only be triggered when `deleteNetworks()` has failed already.

IMO, the destroy module needs a good refactoring that I don't want to commit to right now.

pkg/destroy/openstack/openstack.go
mandre: Hmm, this is a pattern we're using all over the place in this file, and if we wanted to treat 403 differently I think we should do it in a separate change.
How do you suggest we should handle a ErrDefault403 ?

pkg/destroy/openstack/openstack.go
```diff
@@ -653,6 +653,57 @@ func getRouterByPort(client *gophercloud.ServiceClient, allPorts []ports.Port) (
 	return "", nil
 }
 
+func deleteLeftoverLoadBalancers(opts *clientconfig.ClientOpts, logger logrus.FieldLogger, networkID string) (bool, error) {
+	conn, err := clientconfig.NewServiceClient("load-balancer", opts)
+	if err != nil {
+		// Ignore the error if Octavia is not available for the cloud
+		var gerr *gophercloud.ErrEndpointNotFound
```
bogdando, line 660: here you use it as a pointer type, and below in the similar clause, as a refular type. This doesn't look consistent?

pkg/destroy/openstack/openstack.go
```diff
@@ -653,6 +653,57 @@ func getRouterByPort(client *gophercloud.ServiceClient, allPorts []ports.Port) (
...
+	for _, loadbalancer := range allLoadBalancers {
+		if !strings.HasPrefix(loadbalancer.Description, "Kubernetes external service") {
+			logger.Debugf("Not deleting LoadBalancer %q with description %q", loadbalancer.ID, loadbalancer.Description)
+			continue
+		}
+		logger.Debugf("Deleting LoadBalancer %q", loadbalancer.ID)
+		err = loadbalancers.Delete(conn, loadbalancer.ID, deleteOpts).ExtractErr()
+		if err != nil {
+			// Ignore the error if the load balancer cannot be found
+			var gerr gophercloud.ErrDefault404
```
bogdando, line 695: please compare to the line #660

Fedosin: Actually this is correct... This is how errors in gophercloud are organized

pkg/destroy/openstack/openstack.go
Fedosin:
```
if err != nil {
    var gerr gophercloud.ErrDefault403
    if !errors.As(err, &gerr) {
        logger.Debugf("It's forbidden to list load balancers")
        return true, nil
    }
    logger.Error(err)
    return false, nil
}
```

pkg/destroy/openstack/openstack.go
Fedosin: I mean you created a new function `deleteLeftoverLoadBalancers` and it returns two parameters that you ignore. I think we need to either change the signature of the function, or handle the returned values.

mandre: OK, I changed the signature. PTAL.

Fedosin (approved):
/approve
/lgtm

PR #4564, opened 2021-01-18
add systemd-resolved section to libvirt docs

kwiesmueller:
This PR adds the steps required to setup DNS on a systemd-resolved based OS when using the libvirt provider.

Thanks to @cfergeau for the assistance resolving this.

### DIFF
--- a/docs/dev/libvirt/README.md
+++ b/docs/dev/libvirt/README.md
@@ -227,9 +227,25 @@ NOTE: When the firewall rules are no longer needed, `sudo firewall-cmd --reload`
 will remove the changes made as they were not permanently added. For persistence,
 add `--permanent` to the `firewall-cmd` commands and run them a second time.
 
-### Set up NetworkManager DNS overlay
+### Set up DNS
 
 This step allows installer and users to resolve cluster-internal hostnames from your host.
+
+#### Using systemd-resolved
+
+If your system is using `systemd-resolved` (the default since Fedora 33), then you can setup DNS without using `dnsmasq`.
+
+Just point `resovectl` to use `192.168.126.1` for your `baseDomain`. Replace `baseDomain` in the example accordingly.
+
+```sh
+sudo resolvectl dns tt0 192.168.126.1
+sudo resolvectl domain tt0 ~<baseDomain>
+```
+
+Your cluster might run on a different network interface. To find which one it is, either check your `install-config.yaml`
+or run `nmcli` and find the interface that's running the `192.168.126.0/24` net.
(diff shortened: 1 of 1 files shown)


kwiesmueller:
/hold
I just recreated my cluster and the change was gone.
Might really be required as a hook.

cfergeau:
> /hold
> I just recreated my cluster and the change was gone.
> Might really be required as a hook.

The systemd-resolved changes go away when the corresponding libvirt network is destroyed or undefined, and do not come back automatically when it's restarted or recreated. Maybe this is what you saw?

kwiesmueller:
Yes, that's what I saw.
I suppose a hook can reapply them when it comes back?

cfergeau:
> Yes, that's what I saw.
> I suppose a hook can reapply them when it comes back?

A libvirt hook should be usable for that purpose https://www.libvirt.org/hooks.html

cfergeau:
I've attempted to consolidate https://github.com/openshift/installer/pull/4929/ and this PR in https://github.com/openshift/installer/pull/5132 in order to finally update the libvirt documentation for resolvectl.

cfergeau:
> > Yes, that's what I saw.
> > I suppose a hook can reapply them when it comes back?
> 
> A libvirt hook should be usable for that purpose https://www.libvirt.org/hooks.html

I've had success with the hook below. The cluster name needs to be adjusted to match the one you are creating, but apart from this it should be generic. The hook file must be executable to be used by libvirt.

```
$ cat /etc/libvirt/hooks/network

#!/bin/bash

# Change this to the name of the cluster which will be created.
cluster_name="tt"

network_name=$1
operation=$2

if ! [[ "$network_name" =~ ^$cluster_name-.* ]]; then
        # not a network created by openshift-install
        exit 0
fi

net_xml=$(cat)
interface=$(echo $net_xml|xmllint --xpath 'string(/hookData/network/bridge/@name)' -)
domain=$(echo $net_xml | xmllint --xpath 'string(/hookData/network/domain/@name)' -)
ip=$(echo $net_xml | xmllint --xpath 'string(/hookData/network/ip/@address)' -)

if [[ -z "$interface" ]]; then
        exit 1
fi
if [[ -z "$domain" ]]; then
        exit 1
fi
if [[ -z "$ip" ]]; then
        exit 1
fi

case $operation in
        "started")
                /usr/bin/resolvectl domain "$interface" "$ip" ~"$domain"
                /usr/bin/resolvectl dns "$interface" "$ip"
                ;;
esac
```

PR #4566, opened 2021-01-18
Bug 1893117: terraform/vsphereprivate: ignore hosts in maintenance mode

jstuever:
This change modifies the vsphereprivate terraform provider to ignore
all hosts in maintenance mode when importing the rhcos ova. As a result,
it will no longer throw an error when selecting a host in maintenance
mode to upload the ova to.

### DIFF
--- a/pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go
+++ b/pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go
@@ -175,18 +175,22 @@ func findImportOvaParams(client *vim25.Client, datacenter, cluster, datastore, n
 	if err != nil {
 		return nil, err
 	}
-	foundDatastore := false
-	foundNetwork := false
 	var hostSystemManagedObject mo.HostSystem
 
 	// Confirm that the network and datastore that was provided is
 	// available for use on the HostSystem we will import the
 	// OVA to.
 	for _, hostObj := range hosts {
-		hostObj.Properties(ctx, hostObj.Reference(), []string{"network", "datastore"}, &hostSystemManagedObject)
-
+		foundDatastore := false
+		foundNetwork := false
+		hostObj.Properties(ctx, hostObj.Reference(), []string{"network", "datastore", "runtime"}, &hostSystemManagedObject)
 		if err != nil {
 			return nil, err
+		}
+
+		// Skip all hosts that are in maintenance mode.
+		if hostSystemManagedObject.Runtime.InMaintenanceMode {
+			continue
 		}
 		for _, dsMoRef := range hostSystemManagedObject.Datastore {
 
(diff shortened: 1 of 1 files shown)


pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go
```diff
@@ -183,11 +183,17 @@ func findImportOvaParams(client *vim25.Client, datacenter, cluster, datastore, n
 	// available for use on the HostSystem we will import the
 	// OVA to.
 	for _, hostObj := range hosts {
-		hostObj.Properties(ctx, hostObj.Reference(), []string{"network", "datastore"}, &hostSystemManagedObject)
-
+		foundDatastore = false
+		foundNetwork = false
```
staebler, line 187: Nice catch here.

I think there needs to be some work on the error messages on lines 214-219. Let's say that there are two hosts. The first host contains the datastore but not the network. The second host is in maintenance mode. At the end of the for loop, both `foundDatastore` and `foundNetwork` will be false. The error returned will indicate that a host with the datastore was not found, which is not true. I think we can combine the two errors into something like the following.

"failed to find a host in the cluster that contains the provided datacenter and network"

If we are not distinguishing in the error message between missing the datacenter or missing the network, then we can keep the `foundDatastore` and `foundNetwork` in the scope of the for loop. We can just return from the function when the host is found.

pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go
```diff
@@ -183,11 +183,17 @@ func findImportOvaParams(client *vim25.Client, datacenter, cluster, datastore, n
...
+		hostObj.Properties(ctx, hostObj.Reference(), []string{"network", "datastore", "runtime"}, &hostSystemManagedObject)
 		if err != nil {
 			return nil, err
 		}
+
+		// Skip all hosts that are in maintenance mode.
+		if hostSystemManagedObject.Runtime.InMaintenanceMode {
+			break
```
staebler, line 195: This should be `continue`, right?

pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go
jstuever: This is a good point. I had not really considered the logic of the error messages themselves. Resolved with https://github.com/openshift/installer/commit/ceedd7da83a3b2b99d2ece59723871ef5d5656e9

pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go
jstuever: It absolutely should be... what was I thinking? https://github.com/openshift/installer/commit/ceedd7da83a3b2b99d2ece59723871ef5d5656e9

staebler (commented):
/lgtm
/approve

staebler:
> /test e2e-azure
> /hold for azure e2e tests

Oops. Meant vsphere not azure.
/test e2e-vsphere

jstuever:
/hold cancel
e2e-vsphere passed.

PR #4572, opened 2021-01-20
Revert "[fcos] Update FCOS to 33.20210104.3.0"

vrutkovs:
Reverts openshift/installer#4567.

Apparently this breaks DNS resolution on vSphere IPI.


vrutkovs:
/skip

Not sure why vSphere still fails

PR #4575, opened 2021-01-21
Bug 1916938: tolerate equal APIFloatingIP and LbFloatingIP

Fedosin:
In 4.7 LbFloatingIP was deprecated and renamed to APIFloatingIP.
But if we regenerate a config from 4.6 version, it doesn't unset the deprecated LbFloatingIP and both values are present in the new config, which leads to an error.

This commit allows to set both parameters in the config, but only if their values are equal.

/label platform/openstack


staebler (commented):
/lgtm
/approve

PR #4576, opened 2021-01-21
Stream metadata prep

cgwalters:
cache: Create an internal "urlWithIntegrity" API, use better filename

Prep for openshift/os#477
Clarify internally that this sha256 is uncompressed; the stream
will also have the compressed sha256 which we should support
verifying too (as well as the size).

Since we're now parsing the URL better, use the uncompressed
filename as the target name.  This makes it much *much* more
obvious to users how to manually clean up old images in
the cache.  We can do this because for both FCOS/RHCOS
the filenames contain versions and are immutable - we will
never go back and change an image.

### DIFF
--- a/pkg/tfvars/internal/cache/cache.go
+++ b/pkg/tfvars/internal/cache/cache.go
@@ -172,11 +172,11 @@ func cacheFile(reader io.Reader, filePath string, sha256Checksum string) (err er
 	return os.Rename(tempPath, filePath)
 }
 
-// DownloadFile obtains a file from a given URL, puts it in the cache folder, defined by dataType parameter,
+// downloadFile obtains a file from a given URL, puts it in the cache folder, defined by dataType parameter,
 // and returns the local file path.
 // If the query string contains sha256 parameter (i.e. https://example.com/data.bin?sha256=098a5a...),
 // then the downloaded data checksum will be compared with the provided value.
-func DownloadFile(baseURL string, dataType string) (string, error) {
+func downloadFile(baseURL string, dataType string) (string, error) {
 	// Convert the given URL into a file name using md5 algorithm
 	fileName := fmt.Sprintf("%x", md5.Sum([]byte(baseURL)))
 
@@ -232,5 +232,5 @@ func DownloadFile(baseURL string, dataType string) (string, error) {
 func DownloadImageFile(baseURL string) (string, error) {
 	logrus.Infof("Obtaining RHCOS image file from '%v'", baseURL)
 
-	return DownloadFile(baseURL, imageDataType)
+	return downloadFile(baseURL, imageDataType)
 }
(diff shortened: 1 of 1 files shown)


cgwalters:
(I tested this locally w/libvirt btw, works fine)

cgwalters:
(This can wait for 4.8)

cgwalters:
OK, 4.7 is branched; ping for review?
/bugzilla refresh

pkg/tfvars/internal/cache/cache.go
```diff
@@ -232,5 +231,17 @@ func DownloadFile(baseURL string, dataType string) (string, error) {
...
-	return DownloadFile(baseURL, imageDataType)
+	var u urlWithIntegrity
+	parsedURL, err := url.ParseRequestURI(baseURL)
+	if err != nil {
+		return "", err
+	}
+	q := parsedURL.Query()
+	if uncompressedSHA256, ok := q["sha256"]; ok {
+		u.uncompressedSHA256 = uncompressedSHA256[0]
+		delete(q, "sha256")
```
staebler, line 242: If the intention is to remove the query value from `parsedURL`, this is not going to have that effect.
https://play.golang.org/p/vISRRGvej-w

Also, I think it is preferred to use the `Del` function.
```suggestion
		q.Del("sha256")
```

cgwalters: Also needed to re-set the `RawQuery`.  Both fixed.

pkg/tfvars/internal/cache/cache.go
```diff
@@ -232,5 +231,18 @@ func DownloadFile(baseURL string, dataType string) (string, error) {
...
+	var u urlWithIntegrity
+	parsedURL, err := url.ParseRequestURI(baseURL)
+	if err != nil {
+		return "", err
+	}
+	q := parsedURL.Query()
+	if uncompressedSHA256, ok := q["sha256"]; ok {
+		u.uncompressedSHA256 = uncompressedSHA256[0]
+		q.Del("sha256")
+		parsedURL.RawQuery = q.Encode()
```
staebler, line 243: I don't know if this is a bug in go, but this is not preserving an encoded space.
https://play.golang.org/p/Teo3kCdijPi

staebler: Nevermind. This is my poor understanding of URL encoding for query fragments. Apparently '+' is equivalent to space for a query value. And if you actually want '+' you need to use %2B.

cgwalters:
ovirt is a flake, nothing related to this PR
/override ci/prow/e2e-ovirt

staebler:
> ovirt is a flake, nothing related to this PR
> /override ci/prow/e2e-ovirt

/override ci/prow/e2e-ovirt

PR #4583, opened 2021-01-22
Bug 1919072: AWS cloudformation change protocol from ESP to 50

markdgray:
Signed-off-by: Mark Gray <mark.d.gray@redhat.com>


markdgray:
@staebler

PR #4584, opened 2021-01-22
[FCOS] vsphere bootstrap: support systemd-resolved in NM prepender

vrutkovs:
FCOS has switched to systemd-resolved to control nameservers, so DNS_IP should be injected correctly


vrutkovs:
/skip

vSphere needs MCD fix to proceed

PR #4586, opened 2021-01-22
updating ignition file (getting started) link

nladha09:


mtnbikenc (changes requested):
Thanks for your submission.  The link is broken but should be updated on line 26, the `[ignition]` reference-style link.

PR #3912, opened 2020-07-17
Bug 1858418: ovirt_template: Increase ovirt_templates search criteria

lewisdenny:
Current search criteria will miss a template called Blank1 and fail the install.

PR #4587, opened 2021-01-23
Bug 1858418: [RFE] oVirt: Add Template to an openshift-installer survey

zerodayz:
Closes: BZ#1858418

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -1704,6 +1704,11 @@ spec:
                     description: The target storage domain under which all VM disk
                       would be created.
                     type: string
+                  ovirt_template_name:
+                    description: TemplateName is the target oVirt template When no
+                      ovirt_template_name is provided it defaults to `Blank` template,
+                      which is a default template for every ovirt cluster.
+                    type: string
                   vnicProfileID:
                     description: VNICProfileID defines the VNIC profile ID to use
                       the the VM network interfaces. When no vnicProfileID is provided
--- a/data/data/ovirt/main.tf
+++ b/data/data/ovirt/main.tf
@@ -14,6 +14,7 @@ module "template" {
   openstack_base_image_local_file_path = var.openstack_base_image_local_file_path
   ovirt_network_name                   = var.ovirt_network_name
   ovirt_vnic_profile_id                = var.ovirt_vnic_profile_id
+  ovirt_template_name                  = var.ovirt_template_name
 }
 
 module "bootstrap" {
--- a/data/data/ovirt/template/variables.tf
+++ b/data/data/ovirt/template/variables.tf
@@ -43,3 +43,8 @@ variable "ovirt_vnic_profile_id" {
   type        = string
   description = "The ID of the vNIC profile of Logical Network."
 }
+
+variable "ovirt_template_name" {
+  type        = string
+  description = "The name of the Template for the selected Cluster."
+}
--- a/data/data/ovirt/variables-ovirt.tf
+++ b/data/data/ovirt/variables-ovirt.tf
@@ -86,3 +86,8 @@ variable "ovirt_master_instance_type_id" {
   type        = string
   description = "master VM instance type ID"
 }
+
+variable "ovirt_template_name" {
+  type        = string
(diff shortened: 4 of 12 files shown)


zerodayz:
@dougsland @gekorob @Gal-Zaidman  You guys worked with @ldenny-rhn on  https://github.com/openshift/installer/pull/3912 I decided to take this further by following your suggestions, please review if you are happy with PR. There might still be missing integration with install-config tho. But this covers the first part which is defining the template to use from oVirt available templates and defaults to Blank if none.

zerodayz:
I will need help to integrate this with `main.tf` or other terraform files. Feel free to add your branch and compare to mine. I will make sure those changes are added to this PR afterwards. Thank you for your reviews!

Gal-Zaidman:
Hi @zerodayz first of all thanks for the contribution, I have to say the referenced PR[1] is almost nostalgic to me haha.

So looking at your PR what you are doing here is ask the User if he wants to use a preinstalled template for the installation,
There are 2 different problems with that, and I'll elaborate:
**1. This will not solve Bug#1858418:**
It took me a long time to understand that on the referenced PR[1] as well, I much like you I thought that PR[1]/Bug#1858418 was for providing a more convenient way to use a different template in the installation, but as you can see in my comment[2] this is not the case.
On oVirt a VM is created from a base template called "Blank" by default, regardless of openshift, that is just how oVirt works.
That template comes with every oVirt installation and you normally shouldn't change its name or something.
In our installer code we relay on that convention and search for the base template by its normal name "Blank" to let the installer create the name.
The bug#1858418 is for times when the User changed the name of the template in the oVirt engine prior to the installation and broke that convention, this will cause the installer to fail with a weird terraform message.
Now the user shouldn't break that convention, and on regular customers that have a RHV env and know what they are doing its hard to imagine that they will change the name, but as you know... expect the unexpected.
What we can do to solve bug#1858418:
1. Go with the approach on search parm like[1] but use the first result - I'm not a fan of that since it is only a very minor improvement, it only handles cases when the user changed the name to blank something.
2. Let the user override the base template in the install-config - not a fan since it is an over complication in my opinion
3. Change the logic to use the blank template default ID which is "00000000-0000-0000-0000-000000000000" - my personal favorite :)

**2. Asking the user to use a preexisting template can cause problems**
This template will be used for all the machines creation in the cluster, and generally shouldn't be managed by the user, it can lead to problems that will be very hard to debug and solve, and we never had a user/customer that wanted to use a preexisting template.
So we don't want to make it easy for the user - that is why the only way to achieve that is to override the terraform var, meaning we don't even want to add it as a customization.
The general flow is that the user should let the installer create the template

[1] https://github.com/openshift/installer/pull/3912
[2] https://github.com/openshift/installer/pull/3912#issuecomment-662435512

zerodayz:
Well intention of this was to show users all available templates and let them choose one, so it doesn't end up in an err. Hence I thought it would solve the problem. Since Blank1 would be an option and Blank wouldn't. This can be changed in a way that if only one template is returned default to it.

If we do not want to allow users to choose the template due to other issues

Well then it doesn't make sense to have this PR.

This PR was also aiming to add this as an option to install-config.

But yeah if we don't want allow any customisation then it doesn't make sense and we can close this PR.

Thanks for looking at it:-)

Gal-Zaidman:
Thanks for the PR and I will try to implement the default ID way.... I think it should make it much more simplr then the current logic...
Honestly I forgot about this bug

zerodayz:
Also I am not RHV person so I might need to learn more about the Blank template 😀. Thought its something like Template for VM. But then it makes sense that OCP installer makes the template for user. Is that the reason it uses Blank? Thanks!

zerodayz:
Maybe it should contact the API of oVirt and make sure the template is there. And if not create one? What do you think. I could change the PR in that direction. Makes happy everyone? @Gal-Zaidman

zerodayz:
@Gal-Zaidman Any other template will use same ID? Then it makes sense. Thank you for working on it.

zerodayz:
@Gal-Zaidman If you are going with 0000s. Do you also check amount of resources in Blank template? I wonder if we should add the check that the template will work for OCP.

zerodayz:
Closing this PR as the fix will be implemented in a different way.

PR #4589, opened 2021-01-25
Bug 1896218: remove GCP role bindings before service accounts

patrickdillon:
This updates GCP destroy code to remove IAM roles before deleting service accounts.

Service accounts are deleted one at a time but IAM roles are deleted by rewriting the entire policy. We are seeing conflicts (with two clusters in the same project) when one cluster is rewriting the IAM policy as another cluster deletes its service accounts. There is a conflict because the state of service accounts in the newly written policy is out-of-date.  By deleting the role bindings first, we decouple the service accounts from the IAM policy and avoid the conflicts.


patrickdillon:
/assign jstuever

Thanks for offering to check this out.

PR #4592, opened 2021-01-26
data/bootstrap: Check if release image architecture matches host architecture

Prashanth684:
This is a simple sanity check to see if release image architecture matches the bootstrap node's architecture. This is useful in cases of user error where an incorrect release image is specified and we error right away rather than downloading the associated images and failing in say crio when downloading the incorrect pause image.

### DIFF
--- a/data/data/bootstrap/files/usr/local/bin/release-image-download.sh.template
+++ b/data/data/bootstrap/files/usr/local/bin/release-image-download.sh.template
@@ -26,3 +26,17 @@ do
         echo "Pull failed. Retrying $RELEASE_IMAGE..."
     fi
 done
+
+# Sanity check the image metadata to see if the arches match
+image_arch=$(podman inspect $RELEASE_IMAGE --format {{ "{{.Architecture}}" }})
+host_arch=$(uname -m)
+case $host_arch in
+    "x86_64")  host_arch="amd64"   ;;
+    "aarch64") host_arch="arm64"   ;; # not used, just for completeness
+esac
+
+if [[ "$image_arch" != "$host_arch" ]]; then
+    record_service_stage_failure
+    echo "ERROR: release image arch $image_arch does not match host arch $host_arch"
+    exit 1
+fi


data/data/bootstrap/files/usr/local/bin/release-image-download.sh.template
```diff
@@ -17,3 +17,16 @@ do
     echo "Pull failed. Retrying $RELEASE_IMAGE..."
 done
 
+# Sanity check the image metadata to see if the arches match
+image_arch=$(podman inspect $RELEASE_IMAGE --format {{ "{{.Architecture}}" }})
+host_arch=$(uname -m)
+case $host_arch in
```
staebler, line 23: Are the `ppc64le` and `s390x` architectures not included here because they do not need any mangling?

data/data/bootstrap/files/usr/local/bin/release-image-download.sh.template
```diff
@@ -17,3 +17,16 @@ do
...
+    "x86_64")  host_arch="amd64"   ;;
+    "aarch64") host_arch="arm64"   ;; # not used, just for completeness
+esac
+
+if [[ "$image_arch" != "$host_arch" ]]; then
+    echo "ERROR: release image arch $image_arch does not match host arch $host_arch"
+    exit 1
```
staebler, line 30: A failing release-image.service won't prevent other services from starting. So is the purpose of this just to have a failing unit in the bootstrap gather?

staebler (commented):
We are very close to getting the boot image included in the release image (see https://github.com/openshift/installer/pull/4760). Will these changes still be helpful at that point?

> we error right away rather than downloading the associated images and failing in
say crio when downloading the incorrect pause image.

The user would still have to wait the entire 20 minutes for the installer to time out waiting for the bootstrap control plane to be reachable.

data/data/bootstrap/files/usr/local/bin/release-image-download.sh.template
Prashanth684: correct. `x86_64` and `aarch64` map to `amd64` and `arm64` resp. it is a cause of confusion in many places.

data/data/bootstrap/files/usr/local/bin/release-image-download.sh.template
Prashanth684: essential services like crio and kubelet won't start if `release-image.service` throws an error. so in that sense, we have caught this error early and let the user know that there is an arch mismatch in the specified release image.

Prashanth684:
> We are very close to getting the boot image included in the release image (see #4760). Will these changes still be helpful at that point?
> 
> > we error right away rather than downloading the associated images and failing in
> > say crio when downloading the incorrect pause image.
> 
> The user would still have to wait the entire 20 minutes for the installer to time out waiting for the bootstrap control plane to be reachable.

that is true unfortunately. This cannot be done outside of the bootstrap process early on when the installer runs as we would need to download the release image to check the arch metadata. these changes will be help a little in that the error will be obvious on looking at the bootstrap logs rather than seeing a crio error and trying to figure out what the issue could be.

Prashanth684:
> We are very close to getting the boot image included in the release image (see #4760). Will these changes still be helpful at that point?
> 
> > we error right away rather than downloading the associated images and failing in
> > say crio when downloading the incorrect pause image.
> 
> The user would still have to wait the entire 20 minutes for the installer to time out waiting for the bootstrap control plane to be reachable.

Answering the first part of the question which i missed - yes these changes will be useful irrespective of the boot image changes, because this sanity checks the Openshift release image (which can be overridden through the OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE and which is done in local testing and multi-arch CI as well) to see if it matches the target architecture.

Prashanth684:
rebased to latest

data/data/bootstrap/files/usr/local/bin/release-image-download.sh.template
staebler: A release-image.service that fails will not prevent crio.service or kubelet.service from starting.

Prashanth684: ah..right ..yeah. but it does prevent crio-configure.service from starting which in turn disables bootkube.service when there is an error, so in that sense , this error is visible right away after running ` journalctl -b -f -u release-image.service -u bootkube.service`

staebler: No, my point is that it *does not* prevent any of those services from starting.

staebler: > No, my point is that it **does not** prevent any of those services from starting.

Ah, my bad. You are correct. Sorry for the noise. I looked at crio.service and kubelet.service but not bootkube.service.

staebler (approved):
With https://github.com/openshift/installer/pull/4751 as well, then the architecture mismatch will be displayed to the user in the terminal when the installation fails.

/lgtm

staebler:
These changes have no effect on upgrades.

staebler:
The libvirt job has a successful install.
The bootstrap machine is failing to fetch its ignition config in the openstack job. That is prevalent in other runs of the job and is not related to the changes in this PR.
/hold cancel

LorbusChris:
@staebler could you override `e2e-aws-upgrade` again? It doesn't seem to have worked the first time

PR #4595, opened 2021-01-27
Bug 1895917: ovirt - check if username contains user and domain

dougsland:
For login users require to provide user AND domain.
This patch adds such validation.

Signed-off-by: Douglas Schilling Landgraf <dougsland@redhat.com>

### DIFF
--- a/pkg/asset/installconfig/ovirt/credentials.go
+++ b/pkg/asset/installconfig/ovirt/credentials.go
@@ -156,6 +156,12 @@ func askUsername(c *Config) error {
 		return err
 	}
 
+	if !strings.Contains(c.Username, "@") {
+		logrus.Info("Please use the format: username@domain. Example: admin@internal")
+		askUsername(c)
+
+	}
+
 	return nil
 }
 


pkg/asset/installconfig/ovirt/credentials.go
```diff
@@ -156,6 +156,12 @@ func askUsername(c *Config) error {
 		return err
 	}
 
+	if !strings.Contains(c.Username, "@") {
+		logrus.Info("Please use the format: username@domain. Example: admin@internal")
+		askUsername(c)
```
Gal-Zaidman, line 161: In that case if the user will enter a bad cred again then it will pass, I think you can add it in a loop....

janoszen: I don't think a loop is needed, the username query is already in a loop. Just exit out with an error and the main credential loop should take care of it.

dougsland: Hi guys, thanks for the review. It's a recursion. The main loop takes care. Is this address the question?

```
? Engine username admin
INFO Please use the format: username@domain. Example: admin@internal 
? Engine username admin
INFO Please use the format: username@domain. Example: admin@internal 
? Engine username admin
INFO Please use the format: username@domain. Example: admin@internal 
? Engine username admin
INFO Please use the format: username@domain. Example: admin@internal 
? Engine username admin
INFO Please use the format: username@domain. Example: admin@internal 
? Engine username admin
INFO Please use the format: username@domain. Example: admin@internal 
? Engine username admin
INFO Please use the format: username@domain. Example: admin@internal 
? Engine username admin@internal
? Engine password [Press Ctrl+C to switch username, ? for help] 
```

Gal-Zaidman: yep missed that :)

Gal-Zaidman: I must say I really really hate recursion (with the exception of a tree structs), I find it easy to miss and don't encourage unless its the obvious path .
Why not use a while loop here (for {} in go) and exit when the condition is met?
I don't want to use the external loop since the external loop is for login attempts...

@janoszen what is your opinion on the recursion call here? is it just me ?

janoszen: There should not be a recursion here. Simply return an error and be done with it. The main loop takes care of it.

Gal-Zaidman: Nope, look at:

https://github.com/openshift/installer/blob/0473d9536f79686921f894360e49996c1e6c08ca/pkg/asset/installconfig/ovirt/credentials.go#L190-L193

If he returns with an error then he is out...

Regardless the main loop takes care of login attempts so if he will change the main loop to handle that then a login attempt will be wasted on a bad input which is not what we want as well...

I agree with the first part "There should not be a recursion here" ;)

dougsland: > I must say I really really hate recursion (with the exception of a tree structs), I find it easy to miss and don't
> encourage unless its the obvious path .
> Why not use a while loop here (for {} in go) and exit when the condition is met?

I cannot see a chain of loops solve this problem better than a simple recursion.

dougsland: @Gal-Zaidman @janoszen to be very honest, I believe this is not a bug in OCP on RHV code (installer). IMO, it should be opened a bug in ovirt-engine code that generate the sdk code to add such validation.  The current code looks pretty good. Just need to change the error message in higher layers.

```
? Engine FQDN[:PORT] engine.medogz.com
DEBUG engine FQDN: engine.medogz.com               
DEBUG Engine URL: https://engine.medogz.com/ovirt-engine/api 
DEBUG checking URL response... urlAddr: https://engine.medogz.com/ovirt-engine/api skipVerify: true 
DEBUG PEM URL: https://engine.medogz.com/ovirt-engine/services/pki-resource?resource=ca-certificate&format=X509-PEM-CA 
INFO    Version: 3                                  
INFO    Serial Number: 4096                         
INFO    Issuer: CN=engine.medogz.home.52705,O=medogz.home,C=US 
INFO    Validity:                                   
INFO            Not Before: 2019-11-05 21:55:57 +0000 UTC  
INFO            Not After: 2029-11-03 21:55:57 +0000 UTC   
INFO    Subject: CN=engine.medogz.home.52705,O=medogz.home,C=US 
? Would you like to use the above certificate to connect to Engine?  Yes
DEBUG loaded  into the system pool:                
DEBUG login attempts available: 3                  
? Engine username admin
? Engine password [Press Ctrl+C to switch username, ? for help] ******
X Sorry, your reply was invalid: failed to connect to Engine platform Failed to validate the connection, 'Error during SSO authentication access_denied : Cannot authenticate user 'admin@N/A': No valid profile found in credentials..'
```

Please let me know your opinion before closing this PR and the Bug.

janoszen: I would agree.

Gal-Zaidman: @dougsland I agree in general but if we where to solve it on RHV it be a problem:
- Will take much longer to get in due to RHV release cycle
- We really really really try to avoid requiring a specic version of RHV, just think of the amount of comments we got saying that we will support only 4.4 and above.... so solving this in RHV means documenting it in OCP docs or KB that we recommend a specific version and if you will not use it then... that is just a bigger headache in my opinion.

You can open a bug on RHV side but I think we should still have the validation on our side

dougsland: @Gal-Zaidman I don't think it worth adding such validation via chain of loops or recursion as the OCP on RHV users usually are already RHV admins. They are already familiar with the user@domain schema. On top of that, in the installer we suggest admin@internal. If users still try to use just the username it will show an error message which also show in the hosted engine deploys.  Without mention that our documentation is pretty clear about username@domain. https://docs.openshift.com/container-platform/4.6/installing/installing_rhv/installing-rhv-default.html

dougsland: Of course, if you still think we should have this validation in the installer please let me know I can restore this patch.

Probably a good thread to have PM Peter and @sandrobonazzola

Gal-Zaidman: I don't have any strong feeling about the issue and it can go either way in my opinion...
The only thing that I care about is, if we decide that this is a bug and we want a fix, its better to have it in our code then patch ovirt for the reason I mentioned above

PR #4596, opened 2021-01-27
aws: support more auth options in manual mode

abhinavdahiya:
Currently the installer only supports `static credentials` from
AWS config/credentials file or environment for authenticating with AWS.

For the default credential modes (mint/passthrough) this restriction makes
sense because the installer transfers those credentials to the cluster to mint/
passthrough creds to all the components.

For manual mode this restriction can be removed as the credentials are used only
by the installer and installer's internal components(terraform). This is especially
important for allowing users to use STS Webhook identity work, where the users want the
installer to assume a certain role using specific token specified by env variables or,
 the aws config file.

The new credentials loading,
- loads the previously used env and static credentials provider first,
- when not provided it uses the AWS SDK's default internal logic to load of credentials
  from all kinds of supported methods

Now, when the installer tried to create the secret for moving creds to secret, it checks if
credentials are static. If the credentials are not static, these are only allowed in manual
credentials modes like,

```
FATAL failed to fetch Openshift Manifests: failed to generate asset "Openshift Manifests": AWS credentials provided by AssumeRoleProvider are not valid for default credentials mode
```

installer does not configure the terraform aws provider with specific credentials and since
it also has the same behaviour as described above there should be no problem using these temporary
credentials for terraform.

xref: https://issues.redhat.com/browse/CCO-29

### DIFF
--- a/pkg/asset/installconfig/aws/session.go
+++ b/pkg/asset/installconfig/aws/session.go
@@ -3,9 +3,11 @@ package aws
 import (
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/defaults"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
@@ -20,15 +22,11 @@ import (
 	"github.com/openshift/installer/pkg/version"
 )
 
-const (
-	sharedCredentialsProviderName = "SharedCredentialsProvider"
-	envProviderName               = "EnvProvider"
-)
-
 var (
 	onceLoggers = map[string]*sync.Once{
-		sharedCredentialsProviderName: new(sync.Once),
-		envProviderName:               new(sync.Once),
+		credentials.SharedCredsProviderName: new(sync.Once),
+		credentials.EnvProviderName:         new(sync.Once),
+		"credentialsFromSession":            new(sync.Once),
 	}
 )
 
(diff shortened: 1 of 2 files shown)


joelddiaz (commented):
:+1: this worked for me by setting AWS_ROLE_ARN and AWS_WEB_IDENTITY_TOKEN_FILE. I did have an empty .aws/config file and I had to move .aws/credentials out of the way (otherwise the installer kept trying to use that).
But it did work once I could get the installer to show:
`INFO Credentials loaded from the AWS config using "WebIdentityCredentials" provider `

abhinavdahiya:
/cc @staebler @patrickdillon
for review as the master is now open for 4.8

pkg/asset/installconfig/aws/session.go
```diff
@@ -60,48 +58,115 @@ func GetSession() (*session.Session, error) { return GetSessionWithOptions() }
...
 	}
-	if err == credentials.ErrNoValidProvidersFoundInChain {
-		err = getCredentials()
+
+	creds := credentials.NewChainCredentials(providers)
+	credsValue, err := creds.Get()
+	if err != nil && errCodeEquals(err, "NoCredentialProviders") {
+		// getCredentialsFromSession returns credentials derived from a session. A
+		// session uses the AWS SDK Go chain of providers so may use a provider (e.g.,
+		// STS) which provider temporary credentials.
```
staebler, line 99:
```suggestion
		// STS) which provides temporary credentials.
```

pkg/asset/installconfig/aws/session.go
```diff
@@ -60,48 +58,115 @@ func GetSession() (*session.Session, error) { return GetSessionWithOptions() }
...
+		creds, err = getCredentialsFromSession(options)
 		if err != nil {
 			return nil, err
 		}
+		return creds, nil
```
staebler, line 104:
```suggestion
		return getCredentialsFromSession(options)
```

pkg/asset/installconfig/aws/session.go
```diff
@@ -60,48 +58,115 @@ func GetSession() (*session.Session, error) { return GetSessionWithOptions() }
...
-		case envProviderName:
-			onceLoggers[envProviderName].Do(func() {
-				logrus.Info("Credentials loaded from default AWS environment variables")
-			})
-		}
 	}
-	if err == credentials.ErrNoValidProvidersFoundInChain {
-		err = getCredentials()
+
+	creds := credentials.NewChainCredentials(providers)
```
staebler, line 94: Can you explain why we need to attempt to get the credentials from the custom chain first? My understanding is that the env provider and the shared creds provider are both providers used in the default credentials chain. Is it that we are changing the priority of the providers?

pkg/asset/manifests/openshift.go
```diff
@@ -81,13 +81,22 @@ func (o *Openshift) Generate(dependencies asset.Parents) error {
...
-		}))
+		ssn, err := installConfig.AWS.Session(context.TODO())
+		if err != nil {
+			return err
+		}
 		creds, err := ssn.Config.Credentials.Get()
 		if err != nil {
 			return err
 		}
+		if !installconfigaws.IsStaticCredentials(creds) {
```
staebler, line 92: I think these checks should be part of the platform validation. They should yield field errors saying that the credentialsMode field is not valid rather than errors saying that the credentials are not valid.

https://github.com/openshift/installer/blob/cdf74784ee90e21f0961fa379e4917c895fafd62/pkg/asset/installconfig/aws/validation.go#L55

pkg/asset/installconfig/aws/session.go
abhinavdahiya: Because that is the existing behavior today, and i don't intent to change that for supporting new methods.

if there is consensus that default code path and priority makes sense, taking that on in separate follow up PR would be better.

pkg/asset/manifests/openshift.go
abhinavdahiya: the requirement for a certain credential source is not mandated by the credentialMode but the aws secret manifest, and therefore i think it's best to check in the credentials manifests asset that i can't move forward with this configuration.
And i disagree the invalid input is credentialMode rather it is the credential source.

staebler: It is two sides of the same coin.
1) Given that the creds are not static, then the cred mode must be manual.
1) Given that the cred mode is not manual, then the creds must be static.
So, I suppose the question is, given a misconfiguration, which is the user likely to change--the cred mode or the creds? If your contention is that the user is more likely to switch to different creds, then it makes sense to point them at the creds as being invalid. If the user is more likely to switch the cred mode, however, I think it would be better to point the user at the field in the install config that they need to change.

abhinavdahiya: > It is two sides of the same coin.
> 
>     1. Given that the creds are not static, then the cred mode must be manual.
> 
>     2. Given that the cred mode is not manual, then the creds must be static.
> 
> So, I suppose the question is, given a misconfiguration, which is the user likely to change--the cred mode or the creds? If your contention is that the user is more likely to switch to different creds, then it makes sense to point them at the creds as being invalid. If the user is more likely to switch the cred mode, however, I think it would be better to point the user at the field in the install config that they need to change.

most users are guided better onto the safe path of user static creds and the defaut creds mode. Manual creds mode has dragons all the way down for now. So i rather have users think that they needs to fix their credentials instead of updating the credentialMode to manual.

pkg/asset/installconfig/aws/session.go
abhinavdahiya: fixed with dd7ab0197

pkg/asset/installconfig/aws/session.go
abhinavdahiya: fixed with dd7ab0197

abhinavdahiya:
@patrickdillon @staebler updated to fix some minor nits. PTAL :ty:

staebler (commented):
/lgtm
/approve

PR #4598, opened 2021-01-28
Bug 1921578: OpenStack: disable quota validations

Fedosin:
OpenStack cloud may return incorrect current resource usage, which will cause the validation to fail. Since we can't rely on the output, it's better to stop validating quotas for this platform.

/label platform/openstack


bogdando:
could the quotas checking be disabled without being removed? there could be also an option to check it, if user wants.

Fedosin:
yeah, the main issue is addressed in https://github.com/openshift/installer/pull/4599, so I'm going to close this PR

PR #4601, opened 2021-01-28
Bug 1896218: remove GCP role bindings before service accounts

patrickdillon:
Service accounts are deleted one at a time but IAM roles are deleted by rewriting the entire policy. We are seeing conflicts (with two clusters in the same project) when one cluster is rewriting the IAM policy as another cluster deletes its service accounts. There is a conflict because the state of service accounts in the newly written policy is out-of-date.  By deleting the role bindings first, we decouple the service accounts from the IAM policy and avoid the conflicts.

Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1896218


patrickdillon:
/test e2e-gcp
To get a few runs under our belt.

jstuever:
/hold
We discussed doing this slightly differently.

jstuever:
/close
in favor of https://github.com/openshift/installer/pull/4602

PR #4603, opened 2021-01-29
Bug 1909978: update documentation for ignore-volume-az usage

Fedosin:
Now if we just set "ignore-volume-az = yes" in the config, users can't create PVs with the standard storage class. To workaround this issue users have to create another storage class with the `availability` parameter.

This commit documents how to do it.

/label platform/openstack


docs/user/openstack/known-issues.md
```diff
@@ -61,15 +61,30 @@ OpenShift does not currently support Cinder availability zones. When attaching a
...
+allowVolumeExpansion: true
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: sc-availability
+provisioner: kubernetes.io/cinder
+parameters:
+  availability: nova
+reclaimPolicy: Delete
+volumeBindingMode: Immediate
```
mandre, line 85: The default `standard` storage class has `volumeBindingMode: WaitForFirstConsumer`. Is there a need to change it to `Immediate` ?

Fedosin: It comes from the example in the BZ. I agree, it is reasonable to keep it consistent with the `standard` SC.

mandre (commented):
/approve
/lgtm

PR #4606, opened 2021-02-01
Bug 1923038: OpenStack: cache cloud info

Fedosin:
Now Installer loads cloud info for OpenStack platform twice: for base validations and for quota checking. This is not reasonable, because we just double the number of requests to OpenStack cloud.

This commit caches cloud info, which allows to reuse what we get from the first time.

/label platform/openstack

### DIFF
--- a/pkg/asset/installconfig/openstack/validation/cloudinfo.go
+++ b/pkg/asset/installconfig/openstack/validation/cloudinfo.go
@@ -59,10 +59,17 @@ type record struct {
 	Value   int64
 }
 
+var ci *CloudInfo
+
 // GetCloudInfo fetches and caches metadata from openstack
 func GetCloudInfo(ic *types.InstallConfig) (*CloudInfo, error) {
 	var err error
-	ci := CloudInfo{
+
+	if ci != nil {
+		return ci, nil
+	}
+
+	ci = &CloudInfo{
 		clients: &clients{},
 		Flavors: map[string]Flavor{},
 	}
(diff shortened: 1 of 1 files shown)


pkg/asset/installconfig/openstack/validation/cloudinfo.go
```diff
@@ -59,10 +59,17 @@ type record struct {
 	Value   int64
 }
 
+var ci *CloudInfo
+
 // GetCloudInfo fetches and caches metadata from openstack
 func GetCloudInfo(ic *types.InstallConfig) (*CloudInfo, error) {
```
bogdando, line 65: nit: the ic param doesn't look used?

mdbooth (commented):
There's actually a larger problem here: each of the NewServiceClient calls on lines 79, 84, and 89 hits keystone again. Gophercloud gives us an API (something like AuthenticatedClient IIRC) which avoids doing that. We could reduce ourselves to hitting keystone just once.

Also agree with @mandre that we should hold this low prio fix until the next release window.

Fedosin:
`NewServiceClient` function always tries to [authenticate](https://github.com/gophercloud/utils/blob/master/openstack/clientconfig/requests.go#L815) and there is no way to disable it. We can start using  AuthenticatedClient API, but in this case we'll have to reimplement the service client function and call it `NewServiceClientWithCustomAuthenticatedClient`, which is not a goal of this PR.

iamemilio:
based on the description, isn't the ci pointer supposed to be accessed from the quota asset? I don't see how its doing that here

Fedosin:
@iamemilio do you mean this? https://github.com/openshift/installer/blob/master/pkg/asset/quota/quota.go#L125

PR #4608, opened 2021-02-02
Bug 1924140: fix a typo in an env variable name

Fedosin:
It should be called OPENSHIFT_INSTALL_SKIP_PREFLIGHT_VALIDATIONS.

### DIFF
--- a/hack/openstack/test-manifests.sh
+++ b/hack/openstack/test-manifests.sh
@@ -14,7 +14,7 @@ declare \
 
 # Let install-config describe a configuration that is incompatible with the
 # target CI infrastructure
-export OPENSHFIT_INSTALL_SKIP_PREFLIGHT_VALIDATIONS=1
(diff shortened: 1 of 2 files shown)


mandre (commented):
/lgtm
/approve

pierreprinetti:
/hold

I don't think we should be merging anything this low-priority in this particular moment.

Please consider closing this and the BZ

pierreprinetti:
/hold cancel

Thanks for the patience, I needed to check that `manifest-test` was OK; it was! Thanks Mike.

I am still a bit surprised by this change, because in my understanding:
* if the flag is an unsupported, dev-only flag, then I don't see the value of the (backwards-incompatible) fixing;
* if the flag is used by our customers, then it's part of the API and we should only make backwards-compatible changes (e.g. by ADDING the typo-free version instead of changing).

That being said, given the (presumably very small) reach of this flag, I agree that fixing a typo is probably going to have a more positive than negative impact.

Thanks!

PR #4612, opened 2021-02-03
Bug 1924670: openstack/validation: enforce control plane size

pierreprinetti:
This is a cherry-pick of commit fb94d1cd4d713ee79efa672af71b52a067cb2e36

This is a follow up to b6e3088d, which hard codes the size of the
control plane to three nodes. We have a broad policy within OpenShift to
only support three-node control planes, but that's there's nothing
technically preventing a user from using a different size. In the case
of IPI on OpenStack, however, there is a technical restriction, so this
explicitly validates that.

### DIFF
--- a/pkg/asset/installconfig/openstack/validate.go
+++ b/pkg/asset/installconfig/openstack/validate.go
@@ -31,3 +31,11 @@ func Validate(ic *types.InstallConfig) error {
 
 	return allErrs.ToAggregate()
 }
+
+// ValidateForProvisioning validates that the install config is valid for provisioning the cluster.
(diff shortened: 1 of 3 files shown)


pierreprinetti:
https://github.com/openshift/installer/pull/4585

Fedosin (approved):
/lgtm

mdbooth:
I still don't like it. I think it's confusing and unnecessary.

knobunc:
[As patch-manager] Pulling the cherry-pick-approval since the PR didn't make the merge window... sorry.  Please resolve the CI problems so it can make next week's window.

PR #4614, opened 2021-02-03
manifests: migrate CRDs to use apiextensions/v1

staebler:
The `networks.operator.openshift.io` CRD created by the manifests is using the deprecated apiextensions.k8s.io/v1alpha1 version.
These changes update to apiextensions.k8s.io/v1. The CRD is not using any fields that have been removed in v1, so no changes are needed other than changing the API version.


staebler:
```
Feb 04 00:26:53 ip-10-0-1-43 bootkube.sh[2194]: "cluster-network-01-crd.yml": failed to create customresourcedefinitions.v1.apiextensions.k8s.io/networks.operator.openshift.io -n : CustomResourceDefinition.apiextensions.k8s.io "networks.operator.openshift.io" is invalid: spec.versions[0].schema.openAPIV3Schema: Required value: schemas are required
```

There is more work that needs to be done to update the CRD.

PR #4617, opened 2021-02-04
Bug 1909978: update ignore-volume-az documentation

Fedosin:
Now in the documentation we assume that the Cinder AZ is always called "nova". This commit replaces the hardcoded "nova" string with "<Cinder_AZ>", which then will be changed by users to the real Cinder AZ name.

/label platform/openstack


mandre (commented):
/lgtm
/approve

PR #4623, opened 2021-02-06
update iptables call with sudo

sallyom:
quick fix to add sudo to iptables command in libvirt docs


cfergeau (approved):

cgwalters (approved):
Fine by me, prefixing everything with `sudo` is definitely the Fedora-derivative zeitgeist.
/lgtm

PR #4624, opened 2021-02-08
Add support for passwordHash for bootstrap

hamzy:
There is a use case in debugging an initial deployment where
the networking does not work.  If you have access to the VM's
console, then setting the passwordHash value for the core
user allows you to access the VM for debugging.

### DIFF
--- a/pkg/asset/ignition/bootstrap/bootstrap.go
+++ b/pkg/asset/ignition/bootstrap/bootstrap.go
@@ -194,12 +194,26 @@ func (a *Bootstrap) Generate(dependencies asset.Parents) error {
 
 	a.addParentFiles(dependencies)
 
+	var core igntypes.PasswdUser
+
+	if installConfig.Config.PasswordHash == "" {
+		core = igntypes.PasswdUser{Name: "core",
+			SSHAuthorizedKeys: []igntypes.SSHAuthorizedKey{
+				igntypes.SSHAuthorizedKey(installConfig.Config.SSHKey),
+				igntypes.SSHAuthorizedKey(string(bootstrapSSHKeyPair.Public())),
+			}}
+	} else {
+		core = igntypes.PasswdUser{Name: "core",
+			PasswordHash: &installConfig.Config.PasswordHash,
+			SSHAuthorizedKeys: []igntypes.SSHAuthorizedKey{
+				igntypes.SSHAuthorizedKey(installConfig.Config.SSHKey),
+				igntypes.SSHAuthorizedKey(string(bootstrapSSHKeyPair.Public())),
+			}}
+	}
+
 	a.Config.Passwd.Users = append(
 		a.Config.Passwd.Users,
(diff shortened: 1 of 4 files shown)


staebler:
@hamzy Thanks for the PR. Unfortunately, this is not something that I feel is useful to enough install scenarios to warrant adding it to the install config. Also, I do not want to give the impression that this is something that a typical user should be using. You can always manually add the password hash to the bootstrap.ign after running `create ignition-configs`.

hamzy:
@staebler I've spent a while writing python code to manually add the password hash to the bootstrap.ign after running `create ignition-configs`.  The problem is that `create cluster` will nuke my changes and recreate the bootstrap.ign file.  I see the following message from the installer:
"Discarding the Bootstrap Ignition Config that was provided in the target directory because its dependencies are dirty and it needs to be regenerated"

staebler:
You should definitely be able to modify the bootstrap.ign file. If that is not working, then I would be interested in more details. Are you making any other changes to assets besides the bootstrap.ign?

staebler:
> I've spent a while writing python code to manually add the password hash to the bootstrap.ign

`jq` can be a friend to you in this regard.
```
<bootstrap.ign jq '.passwd.users[0].passwordHash="myhash"' > bootstrap.new.ign && mv bootstrap.new.ign bootstrap.ign
```

hamzy:
FYI, for future searches.  I was modifying both the master.ign and bootstrap.ign files.

The following process seems to be working for me at the moment, thanks @staebler!

```
# Generate manifests
${OPENSHIFT_INSTALL} --dir=initial create manifests --log-level=debug

# Generate the ignition configs
${OPENSHIFT_INSTALL} --dir=initial create ignition-configs --log-level=debug

# Create a ramdisk for the etcd storage.  This helps with disk latency
cp initial/master.ign initial/master.ign.old
python3 -c 'import json, sys; j = json.load(sys.stdin); j[u"systemd"] = {}; j[u"systemd"][u"units"] = [{u"contents": "[Unit]\nDescription=Mount etcd as a ramdisk\nBefore=local-fs.target\n[Mount]\n What=none\nWhere=/var/lib/etcd\nType=tmpfs\nOptions=size=2G\n[Install]\nWantedBy=local-fs.target", u"enabled": True, u"name":u"var-lib-etcd.mount"}]; json.dump(j, sys.stdout)' < initial/master.ign.old > initial/master.ign.new
cp initial/master.ign.new initial/master.ign

# Regenerate ignition configs
${OPENSHIFT_INSTALL} --dir=initial create ignition-configs --log-level=debug

# Add a passwordHash for the core user
cp initial/bootstrap.ign initial/bootstrap.ign.old
< initial/bootstrap.ign jq '.passwd.users[0].passwordHash="hash"' > initial/bootstrap.new.ign && cp initial/bootstrap.new.ign initial/bootstrap.ign

# Regenerate the ignition configs
${OPENSHIFT_INSTALL} --dir=initial create ignition-configs --log-level=debug

# And away we go...
${OPENSHIFT_INSTALL} --dir=initial create cluster --log-level=debug
```

PR #4632, opened 2021-02-09
Bug 1873649: Validate noProxy input and add prefix for proxy urls

rna-afk:
The installer currently trims the spaces for the noProxy URLs
provided by the user but the Cluster Network Operator does not which
causes the machine config operator to fail with invalid proxy urls.
Adding a validation check to reject the input from the user if there
are spaces in the noProxy input.

Also added validation checks for the HTTPProxy and the HTTPSProxy
similar to the cluster network operator for the same.

### DIFF
--- a/pkg/asset/manifests/proxy.go
+++ b/pkg/asset/manifests/proxy.go
@@ -65,7 +65,7 @@ func (p *Proxy) Generate(dependencies asset.Parents) error {
 		p.Config.Spec = configv1.ProxySpec{
 			HTTPProxy:  installConfig.Config.Proxy.HTTPProxy,
 			HTTPSProxy: installConfig.Config.Proxy.HTTPSProxy,
-			NoProxy:    trimSpaceNoProxy(installConfig.Config.Proxy.NoProxy),
+			NoProxy:    installConfig.Config.Proxy.NoProxy,
 		}
 
 		if installConfig.Config.AdditionalTrustBundle != "" {
@@ -86,7 +86,7 @@ func (p *Proxy) Generate(dependencies asset.Parents) error {
 		p.Config.Status = configv1.ProxyStatus{
 			HTTPProxy:  installConfig.Config.Proxy.HTTPProxy,
 			HTTPSProxy: installConfig.Config.Proxy.HTTPSProxy,
-			NoProxy:    trimSpaceNoProxy(noProxy),
+			NoProxy:    noProxy,
 		}
 	}
 
@@ -184,13 +184,4 @@ func (p *Proxy) Files() []*asset.File {
 // Load loads the already-rendered files back from disk.
 func (p *Proxy) Load(f asset.FileFetcher) (bool, error) {
 	return false, nil
-}
-
-// trimSpaceNoProxy removes the space from comma separated items.
-func trimSpaceNoProxy(noProxy string) string {
-	split := strings.Split(noProxy, ",")
-	for idx, v := range split {
-		split[idx] = strings.TrimSpace(v)
-	}
-	return strings.Join(split, ",")
 }
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -3,6 +3,7 @@ package validation
 import (
 	"fmt"
 	"net"
+	"net/url"
 	"os"
 	"sort"
 	"strings"
@@ -462,22 +463,22 @@ func validateProxy(p *types.Proxy, fldPath *field.Path) field.ErrorList {
 		allErrs = append(allErrs, field.Required(fldPath, "must include httpProxy or httpsProxy"))
 	}
 	if p.HTTPProxy != "" {
-		if err := validate.URI(p.HTTPProxy); err != nil {
-			allErrs = append(allErrs, field.Invalid(fldPath.Child("httpProxy"), p.HTTPProxy, err.Error()))
-		}
+		allErrs = append(allErrs, validateURI(p.HTTPProxy, fldPath.Child("httpProxy"), []string{"http"})...)
 	}
 	if p.HTTPSProxy != "" {
-		if err := validate.URI(p.HTTPSProxy); err != nil {
-			allErrs = append(allErrs, field.Invalid(fldPath.Child("httpsProxy"), p.HTTPSProxy, err.Error()))
-		}
+		allErrs = append(allErrs, validateURI(p.HTTPSProxy, fldPath.Child("httpsProxy"), []string{"http", "https"})...)
 	}
 	if p.NoProxy != "" && p.NoProxy != "*" {
+		if strings.Contains(p.NoProxy, " ") {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("noProxy"), p.NoProxy, fmt.Sprintf("noProxy must not have spaces")))
+		}
 		for idx, v := range strings.Split(p.NoProxy, ",") {
 			v = strings.TrimSpace(v)
 			errDomain := validate.NoProxyDomainName(v)
 			_, _, errCIDR := net.ParseCIDR(v)
 			if errDomain != nil && errCIDR != nil {
-				allErrs = append(allErrs, field.Invalid(fldPath.Child("noProxy").Index(idx), v, "must be a CIDR or domain, without wildcard characters"))
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("noProxy"), p.NoProxy, fmt.Sprintf(
+					"each element of noProxy must be a CIDR or domain without wildcard characters, which is violated by element %d %q", idx, v)))
 			}
 		}
 	}
@@ -554,4 +555,19 @@ func validateCloudCredentialsMode(mode types.CredentialsMode, fldPath *field.Pat
 		allErrs = append(allErrs, field.Invalid(fldPath, mode, fmt.Sprintf("cannot be set when using the %q platform", platform)))
 	}
 	return allErrs
+}
+
+// validateURI checks if the given url is of the right format. It also checks if the scheme of the uri
+// provided is within the list of accepted schema provided as part of the input.
+func validateURI(uri string, fldPath *field.Path, schemes []string) field.ErrorList {
+	parsed, err := url.ParseRequestURI(uri)
+	if err != nil {
+		return field.ErrorList{field.Invalid(fldPath, uri, err.Error())}
+	}
+	for _, scheme := range schemes {
+		if scheme == parsed.Scheme {
+			return nil
+		}
+	}
+	return field.ErrorList{field.NotSupported(fldPath, parsed.Scheme, schemes)}
 }
--- a/pkg/types/validation/installconfig_test.go
+++ b/pkg/types/validation/installconfig_test.go
@@ -42,7 +42,7 @@ func validInstallConfig() *types.InstallConfig {
 		Proxy: &types.Proxy{
 			HTTPProxy:  "http://user:password@127.0.0.1:8080",
 			HTTPSProxy: "https://user:password@127.0.0.1:8080",
-			NoProxy:    "valid-proxy.com, 172.30.0.0/16",
+			NoProxy:    "valid-proxy.com,172.30.0.0/16",
 		},
 	}
 }
@@ -694,7 +694,25 @@ func TestValidateInstallConfig(t *testing.T) {
 				c.Proxy.HTTPProxy = "http://bad%20uri"
 				return c
 			}(),
-			expectedError: `^proxy.httpProxy: Invalid value: "http://bad%20uri": parse .*: invalid URL escape "%20"$`,
+			expectedError: `^proxy.httpProxy: Invalid value: "http://bad%20uri": parse "http://bad%20uri": invalid URL escape "%20"$`,
+		},
+		{
+			name: "invalid HTTPProxy Schema missing",
+			installConfig: func() *types.InstallConfig {
+				c := validInstallConfig()
+				c.Proxy.HTTPProxy = "http//baduri"
+				return c
+			}(),
+			expectedError: `^proxy.httpProxy: Invalid value: "http//baduri": parse "http//baduri": invalid URI for request$`,
+		},
+		{
+			name: "invalid HTTPProxy Schema different schema",
+			installConfig: func() *types.InstallConfig {
+				c := validInstallConfig()
+				c.Proxy.HTTPProxy = "ftp://baduri"
+				return c
+			}(),
+			expectedError: `^proxy.httpProxy: Unsupported value: "ftp": supported values: "http"$`,
 		},
 		{
 			name: "invalid HTTPSProxy",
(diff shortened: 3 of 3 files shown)


pkg/asset/manifests/proxy.go
```diff
@@ -62,10 +62,27 @@ func (p *Proxy) Generate(dependencies asset.Parents) error {
 	}
 
 	if installConfig.Config.Proxy != nil {
+		noProxy := installConfig.Config.Proxy.NoProxy
```
staebler, line 65: Validation should be done in the types validation.
https://github.com/openshift/installer/blob/2952715dcf96a2529047aafd5f047db213c43474/pkg/types/validation/installconfig.go#L48

pkg/asset/manifests/proxy.go
```diff
@@ -62,10 +62,27 @@ func (p *Proxy) Generate(dependencies asset.Parents) error {
...
+		if noProxy != "" {
+			if strings.Contains(noProxy, " ") {
+				return errors.New(fmt.Sprintf("noProxy name %s must not have spaces", noProxy))
+			}
+		}
+
+		httpProxy := installConfig.Config.Proxy.HTTPProxy
+		if !strings.HasPrefix(httpProxy, "http://") {
+			installConfig.Config.Proxy.HTTPProxy = fmt.Sprintf("%s%s", "http://", httpProxy)
```
staebler, line 74: I don't think that we should add the scheme. For one thing, the https proxy does not need to use the https scheme. We should require the user to include the scheme.

I recommend that we do the same validation being done by the cluster-network-operator, with the exception of rejecting spaces in the noProxy rather than trimming them.
https://github.com/openshift/cluster-network-operator/blob/29c7414ae742621bcbcdc4d86a513f7927880dce/pkg/controller/proxyconfig/validation.go#L38

pkg/types/validation/installconfig.go
```diff
@@ -555,3 +582,26 @@ func validateCloudCredentialsMode(mode types.CredentialsMode, fldPath *field.Pat
...
+
+// validateURI checks if the given url is of the right format.
+func validateURI(uri string, fldPath *field.Path) (string, field.ErrorList) {
+	allErrs := field.ErrorList{}
+	parsed, err := url.ParseRequestURI(uri)
+	if err != nil {
+		return "", append(allErrs, field.Invalid(fldPath, uri, err.Error()))
+	}
+	if !parsed.IsAbs() {
+		return "", append(allErrs, field.Invalid(fldPath, uri, fmt.Sprintf("failed validating URI, no scheme for URI")))
```
staebler, line 594: I would add this error to `allErrs` and continue rather than returning here. Ultimately, I think that `validateURI` should take a list of acceptable schemes and validate the scheme rather than returning the scheme. The `validateURI` function should return all errors rather than just the first error encountered.

pkg/types/validation/installconfig.go
```diff
@@ -555,3 +582,26 @@ func validateCloudCredentialsMode(mode types.CredentialsMode, fldPath *field.Pat
...
 }
+
+// validateURI checks if the given url is of the right format.
+func validateURI(uri string, fldPath *field.Path) (string, field.ErrorList) {
+	allErrs := field.ErrorList{}
+	parsed, err := url.ParseRequestURI(uri)
+	if err != nil {
+		return "", append(allErrs, field.Invalid(fldPath, uri, err.Error()))
+	}
+	if !parsed.IsAbs() {
```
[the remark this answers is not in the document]
patrickdillon: > Understanding that this is copied over from the cluster-network-operator, this check is not necessary. Since the URI is being parsed as a request URI, the URI must already be an absolute URI.



I'm not sure I follow. I thought this was just user input from the install config.

pkg/types/validation/installconfig.go
```diff
@@ -555,3 +582,26 @@ func validateCloudCredentialsMode(mode types.CredentialsMode, fldPath *field.Pat
...
+	if port := parsed.Port(); len(port) != 0 {
+		intPort, err := strconv.Atoi(port)
+		if err != nil {
+			return "", append(allErrs, field.Invalid(fldPath, uri, fmt.Sprintf("failed converting port to integer for URI %q: %v", uri, err)))
+		}
+		invalidPorts := validation.IsValidPortNum(intPort)
+		if invalidPorts != nil {
+			return "", append(allErrs, field.Invalid(fldPath, uri, fmt.Sprintf("invalid port number %d", intPort)))
+		}
+	}
```
patrickdillon, line 605: `url.ParseRequestURI` already handles validating the port so you can simplify this and use the underlying implementation: https://play.golang.org/p/g0kuARdvlyZ

patrickdillon (commented):
I think this is missing the functionality from [comment 1](https://bugzilla.redhat.com/show_bug.cgi?id=1873649#c1):
 > We should also check that these seeings don't overlap with other network settings like the service network.

Maybe @staebler or @sferich888 could spell this out a bit more.

pkg/types/validation/installconfig.go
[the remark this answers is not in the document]
staebler: I mean that the code here is copied from cluster-network-operator. The check of `IsAbs` comes from there. If we want to stay as a pristine copy of that code, we can keep the `IsAbs` check, but that check is not necessary.

staebler:
> I think this is missing the functionality from [comment 1](https://bugzilla.redhat.com/show_bug.cgi?id=1873649#c1):
> 
> > We should also check that these seeings don't overlap with other network settings like the service network.
> 
> Maybe @staebler or @sferich888 could spell this out a bit more.

I don't know what is or is not allowed here with respect to other network settings. I'll have to defer to @sferich888 for more information.

pkg/types/validation/installconfig.go
```diff
@@ -50,6 +51,24 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
 	if c.TypeMeta.APIVersion == "" {
 		return field.ErrorList{field.Required(field.NewPath("apiVersion"), "install-config version required")}
 	}
+	if c.Proxy != nil {
+		if c.Proxy.NoProxy != "" && strings.Contains(c.Proxy.NoProxy, " ") {
+			return field.ErrorList{field.Invalid(field.NewPath("noProxy"), c.Proxy.NoProxy, fmt.Sprintf("noProxy must not have spaces"))}
```
staebler, line 56: Don't return when an invalid field is found. Add the error to the list and continue looking for more validation errors.
```suggestion
			allErrs = append(allErrs, field.Invalid(field.NewPath("noProxy"), c.Proxy.NoProxy, fmt.Sprintf("noProxy must not have spaces")))
```

pkg/types/validation/installconfig.go
```diff
@@ -50,6 +51,24 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
...
+		}
+		if c.Proxy.HTTPProxy != "" {
+			err := validateURI(c.Proxy.HTTPProxy, field.NewPath("httpProxy"), "HTTPProxy", []string{"http"})
+			if err != nil {
+				return err
+			}
```
staebler, line 62:
```suggestion
			allErrs = append(allErrs, validateURI(c.Proxy.HTTPProxy, field.NewPath("httpProxy"), "HTTPProxy", []string{"http"})...)
```

pkg/types/validation/installconfig.go
```diff
@@ -555,3 +574,18 @@ func validateCloudCredentialsMode(mode types.CredentialsMode, fldPath *field.Pat
...
+	parsed, err := url.ParseRequestURI(uri)
+	if err != nil {
+		return field.ErrorList{field.Invalid(fldPath, uri, err.Error())}
+	}
+	for _, scheme := range schemes {
+		if scheme == parsed.Scheme {
+			return nil
+		}
+	}
+	return field.ErrorList{field.Invalid(fldPath, parsed.Scheme, fmt.Sprintf("%q is an invalid scheme for %s %s", parsed.Scheme, proxyType, uri))}
```
staebler, line 590: Use `NotSupported` when there is a list of valid options.
```suggestion
	return field.ErrorList{field.NotSupported(fldPath, parsed.Scheme, schemes)}
```
Note, that this also means you do not need the `proxyType` parameter any more.

pkg/types/validation/installconfig.go
rna-afk: Was fixing this. When there's no errors I return nil and that actually gets added to the list which was confusing so I added the if clause.

pkg/types/validation/installconfig.go
```diff
@@ -462,22 +463,25 @@ func validateProxy(p *types.Proxy, fldPath *field.Path) field.ErrorList {
 		allErrs = append(allErrs, field.Required(fldPath, "must include httpProxy or httpsProxy"))
 	}
 	if p.HTTPProxy != "" {
-		if err := validate.URI(p.HTTPProxy); err != nil {
-			allErrs = append(allErrs, field.Invalid(fldPath.Child("httpProxy"), p.HTTPProxy, err.Error()))
+		if err := validateURI(p.HTTPProxy, fldPath.Child("httpProxy"), []string{"http"}); err != nil {
+			allErrs = append(allErrs, err...)
 		}
```
staebler, line 468: The `if` is not necessary.
```suggestion
		allErrs = append(allErrs, validateURI(p.HTTPProxy, fldPath.Child("httpProxy"), []string{"http"})...)
```

pkg/types/validation/installconfig.go
```diff
@@ -462,22 +463,25 @@ func validateProxy(p *types.Proxy, fldPath *field.Path) field.ErrorList {
...
-				allErrs = append(allErrs, field.Invalid(fldPath.Child("noProxy").Index(idx), v, "must be a CIDR or domain, without wildcard characters"))
+		if strings.Contains(p.NoProxy, " ") {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("noProxy"), p.NoProxy, fmt.Sprintf("noProxy must not have spaces")))
+		} else {
+			for idx, v := range strings.Split(p.NoProxy, ",") {
+				errDomain := validate.NoProxyDomainName(v)
+				_, _, errCIDR := net.ParseCIDR(v)
+				if errDomain != nil && errCIDR != nil {
+					allErrs = append(allErrs, field.Invalid(fldPath.Child("noProxy").Index(idx), v, "must be a CIDR or domain, without wildcard characters"))
+				}
```
staebler, line 484: I would consider continuing the validation of the individual elements even if there are spaces.
```suggestion
		for idx, v := range strings.Split(p.NoProxy, ",") {
			if trimmedV := strings.Trim(v); v != trimmedV {
				allErrs = append(allErrs, field.Invalid(fldPath.Child("noProxy"), p.NoProxy, fmt.Sprintf("noProxy must not have spaces, which are present in element %d %q", idx, v)))
				v = trimmedV
			}
			errDomain := validate.NoProxyDomainName(v)
			_, _, errCIDR := net.ParseCIDR(v)
			if errDomain != nil && errCIDR != nil {
				allErrs = append(allErrs, field.Invalid(fldPath.Child("noProxy").Index(idx), v, "must be a CIDR or domain, without wildcard characters"))
		}
```

pkg/types/validation/installconfig.go
```diff
@@ -462,22 +463,25 @@ func validateProxy(p *types.Proxy, fldPath *field.Path) field.ErrorList {
...
 	}
 	if p.NoProxy != "" && p.NoProxy != "*" {
-		for idx, v := range strings.Split(p.NoProxy, ",") {
-			v = strings.TrimSpace(v)
-			errDomain := validate.NoProxyDomainName(v)
-			_, _, errCIDR := net.ParseCIDR(v)
-			if errDomain != nil && errCIDR != nil {
-				allErrs = append(allErrs, field.Invalid(fldPath.Child("noProxy").Index(idx), v, "must be a CIDR or domain, without wildcard characters"))
+		if strings.Contains(p.NoProxy, " ") {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("noProxy"), p.NoProxy, fmt.Sprintf("noProxy must not have spaces")))
```
staebler, line 477: The field path is not correct.
```suggestion
			allErrs = append(allErrs, field.Invalid(fldPath.Child("noProxy"), p.NoProxy, fmt.Sprintf("noProxy must not have spaces")))
```

pkg/types/validation/installconfig.go
```diff
@@ -462,22 +463,25 @@ func validateProxy(p *types.Proxy, fldPath *field.Path) field.ErrorList {
...
+		} else {
+			for idx, v := range strings.Split(p.NoProxy, ",") {
+				errDomain := validate.NoProxyDomainName(v)
+				_, _, errCIDR := net.ParseCIDR(v)
+				if errDomain != nil && errCIDR != nil {
+					allErrs = append(allErrs, field.Invalid(fldPath.Child("noProxy").Index(idx), v, "must be a CIDR or domain, without wildcard characters"))
```
staebler, line 483: You cannot use `Index` here. The `noProxy` field is not a list: It is a string.
```suggestion
					allErrs = append(allErrs, field.Invalid(fldPath.Child("noProxy"), p.NoProxy, fmt.Sprintf("each element of noProxy must be a CIDR or domain without wildcard characters, which is violated by element %d %q", idx, v)))
```

rna-afk: I think the user can pass in multiple noProxy values with comma separation and this code splits the noProxy string into a list and validates each of the links provided. I did not change this part but let me know if you want me to change it.

pkg/types/validation/installconfig.go
rna-afk: I feel checking for spaces once outside the for loop would be better.

pkg/types/validation/installconfig.go
rna-afk: validateURI returns nil if there are no errors and it gets appended to allErrs. This causes https://github.com/openshift/installer/blob/e0761e06048b2f80176f2b0573453b7ca912663e/pkg/asset/installconfig/installconfig.go#L149 to be true because err != nil turns false when the array has value. I'm guessing that is why the if clause was present before this change too. Let me know if there is a better solution.

staebler: The difference between what you have here and the code linked is the `ToAggregate` call, which converts the slice of errors into a single error. With the code that you have, if `validateURI` returns `nil`, then nothing will be appended.

staebler: https://play.golang.org/p/4LRc3kSqgzG

rna-afk: Yeah got it, thanks! I'll return an empty list just like the other functions instead.

pkg/types/validation/installconfig.go
```diff
@@ -555,3 +556,18 @@ func validateCloudCredentialsMode(mode types.CredentialsMode, fldPath *field.Pat
...
+// validateURI checks if the given url is of the right format. It also checks if the scheme of the uri
+// provided is within the list of accepted schema provided as part of the input.
+func validateURI(uri string, fldPath *field.Path, schemes []string) field.ErrorList {
+	parsed, err := url.ParseRequestURI(uri)
+	if err != nil {
+		return field.ErrorList{field.Invalid(fldPath, uri, err.Error())}
+	}
+	for _, scheme := range schemes {
+		if scheme == parsed.Scheme {
+			return field.ErrorList{}
```
staebler, line 569:
```suggestion
			return nil
```

pkg/types/validation/installconfig.go
staebler: Keep it returning `nil`. There is no need to create an empty list.

patrickdillon:
> > I think this is missing the functionality from [comment 1](https://bugzilla.redhat.com/show_bug.cgi?id=1873649#c1):
> > > We should also check that these seeings don't overlap with other network settings like the service network.
> > 
> > 
> > Maybe @staebler or @sferich888 could spell this out a bit more.
> 
> I don't know what is or is not allowed here with respect to other network settings. I'll have to defer to @sferich888 for more information.

@danehans may also be able to give some insight

staebler (commented):
/lgtm
/approve

PR #4636, opened 2021-02-09
Bug 1880758: Replace description to a more meaningful one

rna-afk:
The current description is filled by default "Managed by Terraform"
message that may confuse users. Changing the message to
"Created by Openshift Installer" message to make it more meaningful.


data/data/config.tf
```diff
@@ -123,3 +123,14 @@ Should the cluster be created with ipv6 networking.
...
 }
+
+variable "default_description" {
+  type = string
+
+  description = <<EOF
+Default description that should show for the components
+EOF
+
+default = "Created by Openshift Installer"
```
staebler, line 134: This needs leading spaces. And the s in OpenShift should be capitalized.
```suggestion
  default = "Created by OpenShift Installer"
```

staebler:
There are more terraform formatting changes needed, too. Run `./hack/tf-fmt.sh` to see what needs to be changed.

The following can be used to make the changes.
```
podman run --rm \
 --env IS_CONTAINER=TRUE \
 --volume "${PWD}:${PWD}:z" \
 --workdir "${PWD}" \
 terraform -write=true -recursive data/data
```

rna-afk:
Was testing my tf-lint plugin for vscode. Doesn't work so I'll try that.

mandre (commented):
/retest

It looks like openstack relies on the `default_description` set in `data/data/config.tf` while other platforms have a more specialized `default_description`, for example in `data/data/aws/bootstrap/variables.tf`. Any particular reason for this?

mandre (changes requested):
Well, the [openstack failure](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/4636/pull-ci-openshift-installer-master-e2e-openstack/1361953129632894976) is legit:

```
level=error msg=Error: Reference to undeclared input variable
level=error
level=error msg=  on ../tmp/openshift-install-637007395/bootstrap/main.tf line 33, in resource "openstack_blockstorage_volume_v3" "bootstrap_volume":
level=error msg=  33:   description = "${var.default_description}"
level=error
level=error msg=An input variable with the name "default_description" has not been declared.
level=error msg=This variable can be declared with a variable "default_description" {} block. 
```

rna-afk:
> It looks like openstack relies on the default_description set in data/data/config.tf while other platforms have a more specialized default_description, for example in data/data/aws/bootstrap/variables.tf. Any particular reason for this?

I did try to keep it in the data/data/config.tf file but some individual modules could not pick it up from there. It looks like the modules/resources with the source tag set are not able to pick it up from the main file so I had to also put it in the individual files. Would be happy to hear if there is a better solution :-)

data/data/config.tf
```diff
@@ -123,3 +123,14 @@ Should the cluster be created with ipv6 networking.
 EOF
 
 }
+
+variable "default_description" {
```
staebler, line 127: This variable is declared in the root module but is not passed down to the other modules. Consequently, if the user does set this variable, it will only apply to resources created by the root module.

One of two changes should be made.
1) Make `default_description` a local value in each module rather than an input variable.
2) Pass `default_description` from the root module to each nested module. Do not provide a default value of the variable in the nested module to enforce that the caller of the module must supply the variable.

My preference is (1) as I do not think that we need to make `default_description` variable.

rna-afk: I was hoping to create a global variable so in the future it would be easy to change the value but now since this is not working the way I intended, I think (1) would be easier. I'll make the changes.

staebler: Terraform does not have a concept of global variables.

data/data/aws/bootstrap/main.tf
```diff
@@ -154,7 +154,8 @@ resource "aws_lb_target_group_attachment" "bootstrap" {
 }
 
 resource "aws_security_group" "bootstrap" {
-  vpc_id = var.vpc_id
+  vpc_id      = var.vpc_id
+  description = "Created by OpenShift Installer"
```
staebler, line 158: I was not suggesting that you copy this string everywhere. I was suggesting that you use a [local value](https://www.terraform.io/docs/language/values/locals.html).
```
locals {
  description = "Created by OpenShift Installer"
}
```

```suggestion
  description = local.description
```

rna-afk: ah ok sorry new to terraform

data/data/aws/vpc/sg-master.tf
```diff
@@ -1,5 +1,9 @@
+locals {
+  description = "Created By OpenShift Installer"
```
staebler, line 2: Move this to `data/data/aws/vpc/common.tf` rather than repeating it in sg-master.tf and sg-worker.tf. Local values are visible to everything in the module and not just to the file in which they are declared.

Do the same in `data/data/gcp/network` and `data/data/openstack/topology`.

data/data/aws/iam/variables.tf
```diff
@@ -7,4 +7,3 @@ variable "tags" {
   default     = {}
   description = "AWS tags to be applied to created resources."
 }
-
```
staebler, line 10: Superfluous change.

data/data/azure/vnet/nsg.tf
```diff
@@ -1,3 +1,6 @@
+locals {
+  description = "Created By OpenShift Installer"
+}
```
staebler, line 3: Separate with an empty line. Repeat for the other files where you added the `locals` block.
```suggestion
}

```

data/data/openstack/bootstrap/variables.tf
```diff
@@ -83,3 +83,4 @@ variable "additional_network_ids" {
   type = list(string)
   description = "IDs of additional networks for the bootstrap node."
 }
+
```
staebler, line 86: Superfluous change.

data/data/vsphere/bootstrap/variables.tf
```diff
@@ -45,4 +45,4 @@ variable "thin_disk" {
 
 variable "scrub_disk" {
   type = bool
-}
```
staebler, line 48: Superfluous change.

data/data/vsphere/variables-vsphere.tf
```diff
@@ -76,4 +76,3 @@ variable "vsphere_control_plane_num_cpus" {
 variable "vsphere_control_plane_cores_per_socket" {
   type = number
 }
-
```
staebler, line 79: Superfluous change.

rna-afk: Just wanted to check where the code fails. I'll fix all these fmt errors when I'm a little more confident the code would work. Sorry about the mess :-)

staebler:
Installs on aws, azure, openstack, and gcp have all succeeded.
/hold cancel

PR #4637, opened 2021-02-09
bug 1924701: Fix FIP detach from Router

MaysaMacedo:
During a FIP less installation VMs that were not
created by the installer can get a FIP detached when
the cluster in being destroyed. This commit fixes the
issue by detaching only FIPs that were created by
installer or Kuryr during a FIP less installation.

### DIFF
--- a/pkg/destroy/openstack/openstack.go
+++ b/pkg/destroy/openstack/openstack.go
@@ -403,24 +403,8 @@ func deleteSecurityGroups(opts *clientconfig.ClientOpts, filter Filter, logger l
 	return len(allGroups) == 0, nil
 }
 
-func updateFips(routerID string, opts *clientconfig.ClientOpts, filter Filter, logger logrus.FieldLogger) error {
-	// If a user provisioned floating ip was used, it needs to be dissociated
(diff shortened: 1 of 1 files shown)


dulek:
/retest

OpenStack test failure is related to authentication failure which doesn't seem related.

dulek:
@MaysaMacedo: Do I understand correctly that the point is that this will now remove all the FIPs required to remove the router and all the FIPs that are tagged? I have hard time trying to grasp this.

MaysaMacedo:
> nd correctly that the point is that this will now remove all the FIPs required to remove the router and all the FIPs that are tagged? I have hard time trying to grasp this.

@dulek it detaches all FIPs that were created by Kuryr (meaning that are tagged https://github.com/openstack/kuryr-kubernetes/blob/master/kuryr_kubernetes/controller/drivers/public_ip.py#L133) when installing on byo with/without fipless where a custom router is present, otherwise the router clean up would hang.
When no byo/fipless is in place a router is created by the installer and the FIPs detachment should continue to be handled as it was before.

PR #4638, opened 2021-02-09
machines/openstack: re-enable external clientOpts

abhinavdahiya:
In 05453ef0d the MachineSets function was updated to remove the option
to specify external clientOpts. This clientOpts is used by hive to generate
the machinesets and ensuring the correct client authentication is used.

using the default client opts as forced in 05453ef0d only allows configuring the
openstack client using the default location on the host, which doesn't sit well
with something like hive which needs to manage the machinesets for many different
clients from single location.

Reverting the change in 05453ef0d to allow machinesets to accept external clientOpts
and use the default opts when it is nil.

xref: https://issues.redhat.com/browse/HIVE-1372

/assign @Fedosin

### DIFF
--- a/pkg/asset/machines/openstack/machines.go
+++ b/pkg/asset/machines/openstack/machines.go
@@ -41,7 +41,7 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 
 	mpool := pool.Platform.OpenStack
 	platform := config.Platform.OpenStack
-	trunkSupport, err := checkNetworkExtensionAvailability(platform.Cloud, "trunk")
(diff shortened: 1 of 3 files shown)


Fedosin (approved):
/lgtm

Fedosin:
Openstack CI may be broken atm...

abhinavdahiya:
/hold cancel

the ci jobs are green
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/4638/pull-ci-openshift-installer-master-e2e-openstack/1359225675671670784
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/4638/pull-ci-openshift-installer-master-openstack-manifests/1359557576244596736

PR #4640, opened 2021-02-10
Allow to define custom featuresets for developers

Fedosin:
To test new unsupported features developers need a mechanism to enable a custom feature set at the time of installation. It's also important for creating CI jobs for these features, as developers need automated testing for their work.

This PR allows OpenShift installer to generate custom FeatureGate manifests that will be applied during during bootstrap process.


staebler:
In general, environment variables should not be used to customize the install behavior. Customizations should be done in the install config. In this case, I do not think that feature-gate support is useful enough to our users to warrant inclusion in the install config. A user that want to enable a feature gate should add the manifest manually.

mandre (commented):
The developer docs on how to enable feature gates is useful, we shouldn't however introduce new environment variables to expose the knob as @staebler pointed out. There's already a way for developers to inject custom manifests at install time, see for instance the [AWS docs for advanced network configuration](https://docs.openshift.com/container-platform/4.6/installing/installing_aws/installing-aws-network-customizations.html#modifying-nwoperator-config-startup_installing-aws-network-customizations).

IMO we should turn this PR into a docs only patch.

cdoan1:
@staebler @mandre it is possible this issue comes from trying to provision a cluster using `hive`, where the a manual interaction is not available? Is it possible to enabling the `IPv6DualStackNoUpgrade` featureSet when in the install-config.yaml that we pass to openshift-installer, in a similar way we define cluster wide proxy, or ICSP? If this PR was merged, I would just need to set the environment variable on the hive provisioning job?

staebler:
> @staebler @mandre it is possible this issue comes from trying to provision a cluster using `hive`, where the a manual interaction is not available? Is it possible to enabling the `IPv6DualStackNoUpgrade` featureSet when in the install-config.yaml that we pass to openshift-installer, in a similar way we define cluster wide proxy, or ICSP? If this PR was merged, I would just need to set the environment variable on the hive provisioning job?

Even when using Hive, you should be using a custom manifest. Hive has the capability of pulling custom manifests from a configmap.

elmiko (commented):
this looks really nice @Fedosin! big +1 from me
/lgtm

elmiko:
given the other comments here, i will retract my lgtm. but i still think this is really useful for testing and users who are interested in trying out tech preview stuff.
/lgtm cancel

PR #4641, opened 2021-02-10
Fix OWNERS for openstack

Fedosin:
Now OWNERS is a symbolic link and looks like github doesn't accept it. This commit makes OWNERS a regular text file.

/label platform/openstack


staebler:
This is just a change to an owner file. It does not affect ovirt behavior.

/override ci/prow/e2e-ovirt

PR #4648, opened 2021-02-12
WIP: POC for cgroupsv2

rphillips:


staebler:
@rphillips Can you provide some more context for this (maybe a Jira card or a BZ) and some justification for why this should be a field in the install config?

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -44,6 +44,9 @@ spec:
             required:
             - installationDisk
             type: object
+          cgroupsV2:
```
cgwalters, line 47: I think to start we can omit this from the install config and just leave it as something where if you want to turn it on you provide the additional manifests.  The installer doesn't need to be involved IMO.

(The ugly exception here is the bootstrap node, but hopefully it works as v1 to start)

See also https://github.com/openshift/ci-chat-bot/issues/112 where I argue that we should have a documented set of example "day 1 manifests" in a git repo like github.com/openshift/ci-cluster-variants or something.

cgwalters: TL;DR I don't think we need this patch - *except* if the bootstrap node is relevant, and hopefully it isn't.

(I'm not arguing strongly against this either...perhaps instead this manifests as `env OPENSHIFT_INSTALL_CGROUPS_V2=1` or something for convenience.  But it's more of a developer convenience instead of something we should put in front of customers I think until we enable it by default.  (Right?))

rphillips:
@staebler @cgwalters The use case I am thinking of is that all machines within the cluster should be cgroupsv2 enabled. That means if a new Machine Config Pool is created then it should inherit that it is a cgroupsv2 cluster.

rphillips:
This PR is just a placeholder to instantiate a cgroupsv2 cluster with clusterbot. We will need a longer term PR within MCO or Installer to turn it on. This isn't on anyones jira cards at the moment until we iron out some of the bugs with cgroupsv2+kubelet+crio.

cgwalters:
> That means if a new Machine Config Pool is created then it should inherit that it is a cgroupsv2 cluster.

All custom pools inherit from worker: https://github.com/openshift/machine-config-operator/blob/master/docs/custom-pools.md
So a MC object that enables cgroups v2 will apply to all workers.

staebler:
> This PR is just a placeholder to instantiate a cgroupsv2 cluster with clusterbot. We will need a longer term PR within MCO or Installer to turn it on. This isn't on anyones jira cards at the moment until we iron out some of the bugs with cgroupsv2+kubelet+crio.

Please note that in the PR description so that others don't waste time looking at the PR.

wewang58:
Hi @rphillips  can you help to rebase your pr, so i can use your pr to test a cgroupv2 related bug, thanks.

PR #4649, opened 2021-02-12
Add docs about issues on UPI with Kuryr

MaysaMacedo:
Installations with Kuryr can timeout because of
recent changes on the way the nodes' subnet is retrieved.
This commit adds a section to the know issues describing
why the issue occour and how to avoid it.

### DIFF
--- a/docs/user/openstack/known-issues.md
+++ b/docs/user/openstack/known-issues.md
@@ -112,3 +112,20 @@ When anti-affinity is a requirement, it is advised to rely on the more stable
 control plane replicas only. Add more nodes as a day-2 operation by assigning
 them a new Server group with "anti-affinity" policy. Use the `ServerGroupID`
 property of the Machine ProviderSpec.
+
+## Requirement to create Control Plane Machines manifests (Kuryr SDN)
+
+Installations with Kuryr SDN can timeout due to changes in the way Kuryr detects
+the OpenStack Subnet used by the cluster's nodes. Kuryr relied on the Network of
+the cluster's nodes Subnet having a specific tag, but the tag was removed for IPI
+Installations causing the need to discover it from the OpenShift Machine objects,
+which the creation is removed on one of the UPI steps. Until the fix for
+[the issue][bugzilla-upi] is available, as a workaround, only the compute machine
+manifests should be removed in the [Remove machines and machinesets][manifests-removal]
+section of the UPI guide. The command to run is:
+
+```console
+$ rm -f openshift/99_openshift-cluster-api_worker-machineset-*.yaml
+```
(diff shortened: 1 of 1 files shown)


MaysaMacedo:
/hold

Need to verify if this would only be needed for multiple subnets support.

EmilienM (commented):
/lgtm

docs/user/openstack/known-issues.md
```diff
@@ -112,3 +112,19 @@ When anti-affinity is a requirement, it is advised to rely on the more stable
 control plane replicas only. Add more nodes as a day-2 operation by assigning
 them a new Server group with "anti-affinity" policy. Use the `ServerGroupID`
 property of the Machine ProviderSpec.
+
+## Requirement to create Control Plane Machines manifests (Kuryr SDN)
+
+Installations with Kuryr SDN can timeout due to changes in the way Kuryr detects
+the OpenStack Subnet used by the cluster's nodes. Kuryr relied on the Network of
+the cluster's nodes Subnet having a specific tag, but the tag was removed causing
+the need to the discover it from the OpenShift Machine objects, which the creation
```
mandre, line 121:
```suggestion
the need to discover it from the OpenShift Machine objects, which the creation
```

docs/user/openstack/known-issues.md
```diff
@@ -112,3 +112,19 @@ When anti-affinity is a requirement, it is advised to rely on the more stable
...
+is removed on one of the UPI steps. As a workaround, only the control plane machine
+manifests should **not** be removed in the [Remove machines and machinesets][manifests-removal]
```
mandre, line 123: This is hard to read, I suggest you reverse this to point what should be removed, instead of what should not.
```suggestion
is removed on one of the UPI steps. As a workaround, only the compute machine
manifests should be removed in the [Remove machines and machinesets][manifests-removal]
```

mandre (commented):
It would be nice to link to a BZ if you have one so that we know if the issue is fixed or not and can update the doc.

docs/user/openstack/known-issues.md
dulek: +1!

dulek (commented):
Shouldn't we add a note that this is to be fixed and link the 4.7 bugzilla (https://bugzilla.redhat.com/show_bug.cgi?id=1929168)?

dulek:
/lgtm

Ah, typo.

MaysaMacedo:
@EmilienM can you take another look at this one?

Even though the fix is already merged, it might be good to highlight it in the docs.

mandre (commented):
/approve

PR #4650, opened 2021-02-14
Validate HotplugVolumes feature gate in Kubevirt

ydayagi:
In Kubevirt platform the CSI driver requires this feature gate

https://issues.redhat.com/browse/CNV-9820

### DIFF
--- a/pkg/asset/installconfig/kubevirt/client.go
+++ b/pkg/asset/installconfig/kubevirt/client.go
@@ -39,6 +39,7 @@ type Client interface {
 	GetStorageClass(ctx context.Context, name string) (*storagev1.StorageClass, error)
 	GetNetworkAttachmentDefinition(ctx context.Context, name string, namespace string) (*unstructured.Unstructured, error)
 	CreateSelfSubjectAccessReview(ctx context.Context, reviewObj *authv1.SelfSubjectAccessReview) (*authv1.SelfSubjectAccessReview, error)
+	GetHyperConverged(ctx context.Context, name string, namespace string) (*unstructured.Unstructured, error)
 }
 
 type client struct {
@@ -207,6 +208,16 @@ func (c *client) GetNetworkAttachmentDefinition(ctx context.Context, name string
 func (c *client) CreateSelfSubjectAccessReview(ctx context.Context, reviewObj *authv1.SelfSubjectAccessReview) (*authv1.SelfSubjectAccessReview, error) {
 	return c.kubernetesClient.AuthorizationV1().SelfSubjectAccessReviews().Create(ctx, reviewObj, metav1.CreateOptions{})
 }
+
+func (c *client) GetHyperConverged(ctx context.Context, name string, namespace string) (*unstructured.Unstructured, error) {
+	resource := schema.GroupVersionResource{
+		Group:    "hco.kubevirt.io",
+		Version:  "v1beta1",
+		Resource: "hyperconvergeds",
+	}
+	return c.getResource(ctx, namespace, name, resource)
+}
+
 func (c *client) createResource(ctx context.Context, obj interface{}, namespace string, resource schema.GroupVersionResource) error {
 	resultMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
 	if err != nil {
--- a/pkg/asset/installconfig/kubevirt/validation.go
+++ b/pkg/asset/installconfig/kubevirt/validation.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/openshift/installer/pkg/types"
 	authv1 "k8s.io/api/authorization/v1"
+	unstructured "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
@@ -53,6 +54,25 @@ func ValidatePermissions(client Client, ic *types.InstallConfig) error {
 		}
 
 		return fmt.Errorf("the user is missing the following permissions: %s", strings.Join(notAllowed, ", "))
+	}
+
+	return nil
+}
+
+// ValidateForProvisioning is called by PlatformProvisionCheck
+func ValidateForProvisioning(client Client) error {
+	hcUnstructured, err := client.GetHyperConverged(context.Background(), "kubevirt-hyperconverged", "openshift-cnv")
+	if err != nil {
+		return fmt.Errorf("failed to get resource openshift-cnv/kubevirt-hyperconverged, with error: %v", err)
+	}
+
+	enabled, found, err := unstructured.NestedBool(hcUnstructured.Object, "spec", "featureGates", "hotplugVolumes")
+	if err != nil {
+		return fmt.Errorf("failed to read boolean value 'spec.featureGates.hotplugVolumes' from resource openshift-cnv/kubevirt-hyperconverged, with error: %v", err)
+	}
+
+	if !found || !enabled {
+		return fmt.Errorf("feature gate hotplugVolumes is either missing or not set to true. Review resource openshift-cnv/kubevirt-hyperconverged. Follow Kubevirt CSI driver documentation for setting the feature gate (https://github.com/openshift/kubevirt-csi-driver)")
 	}
 
 	return nil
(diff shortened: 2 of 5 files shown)


pkg/asset/installconfig/kubevirt/client.go
```diff
@@ -202,6 +203,15 @@ func (c *client) GetNetworkAttachmentDefinition(ctx context.Context, name string
 	return c.getResource(ctx, namespace, name, nadRes)
 }
 
+func (c *client) GetHyperConverged(ctx context.Context, name string, namespace string) (*unstructured.Unstructured, error) {
```
nirarg, line 206: This call need aditional permission "Get hyperconvergeds"
Is this madatory?
If yes, need to add relevant permission to the role example: https://github.com/openshift/installer/blob/8fdf359d8edbc0ad52a0fe3adaf801c56f28327a/docs/user/kubevirt/role.yaml#L2
If not, the validation shouldn't fail in case of "Not Authorized error"

pkg/asset/installconfig/kubevirt/client.go
```diff
@@ -202,6 +203,15 @@ func (c *client) GetNetworkAttachmentDefinition(ctx context.Context, name string
...
+	nadRes := schema.GroupVersionResource{
```
nirarg, line 207: nadRes is wrong parameter name (hcoRes)

pkg/asset/installconfig/kubevirt/validation.go
```diff
@@ -94,3 +96,49 @@ func validateNamespace(namespace string, client Client, fieldPath *field.Path) f
 
 	return allErrs
 }
+
+func validateHcoFeatureGates(client Client, fieldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	hcUnstructured, err := client.GetHyperConverged(context.Background(), "kubevirt-hyperconverged", "openshift-cnv")
```
nirarg, line 103: Is it possible that the client would return the "HyperConverged" type in order to prevent the following parsing?
(https://github.com/kubevirt/hyperconverged-cluster-operator/blob/314e87d50b35be7f611964bbcabf16adb9713cf0/pkg/apis/hco/v1beta1/hyperconverged_types.go#L187)

pkg/asset/installconfig/kubevirt/validation.go
```diff
@@ -17,6 +18,7 @@ func Validate(ic *types.InstallConfig, client Client) error {
 	allErrs = append(allErrs, validateNamespace(kubevirtPlatform.Namespace, client, fldPath.Child("namespace"))...)
 	allErrs = append(allErrs, validateStorageClassExistsInInfraCluster(kubevirtPlatform.StorageClass, client, fldPath.Child("storageClass"))...)
 	allErrs = append(allErrs, validateNetworkAttachmentDefinitionExistsInNamespace(kubevirtPlatform.NetworkName, kubevirtPlatform.Namespace, client, fldPath.Child("networkName"))...)
+	allErrs = append(allErrs, validateHcoFeatureGates(client, fldPath.Child("featureGates"))...)
```
nirarg, line 21: The field path extension should be done using valid field name from the platform fields:
https://github.com/openshift/installer/blob/8fdf359d8edbc0ad52a0fe3adaf801c56f28327a/pkg/types/kubevirt/platform.go#L4
Need to consider which field is the appropriate one to perform this validation for

pkg/asset/installconfig/kubevirt/validation.go
ydayagi: I prefer not to use it. I do not see any benefit from using the type. On the contrary. It leads to many dependencies. This means also a lot to add to go.mod.
This also means that any changes in the type will require changes in our code.
In my professional/personal opinion, using the whole type when u need just a subset of it is leads to more trouble than necessary. Extracting just the desired piece of of data makes the code resilient to changes and future versions.

pkg/asset/installconfig/kubevirt/client.go
ydayagi: silly me. will fix

pkg/asset/installconfig/kubevirt/client.go
ydayagi: what do u mean by additional permission? by whom? What role? The team told me that the credentials we use are the credentials used by the user that executes the installer.

pkg/asset/installconfig/kubevirt/validation.go
staebler: I concur with @nirarg. You cannot use a field path of `featureGates` here because the install config does not have that field. The reason that there is a question about what the field name should be is because this is not the right place to be validating whether the feature gate is enabled. This `Validate` function is intended to validate the install config in ways that require access to the platform on which the install is being performed. Checking for whether feature gates are enabled or not is not a validation on the install config. I suggest that the feature-gate validation be moved to the [Platform Provision Check asset](https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/platformprovisioncheck.go).

pkg/asset/installconfig/kubevirt/validation.go
```diff
@@ -94,3 +96,49 @@ func validateNamespace(namespace string, client Client, fieldPath *field.Path) f
...
+			field.InternalError(
+				fieldPath,
+				err,
+			),
+		)
+
+		return allErrs
+	}
+
+	enabled, ok := gojsonq.New().FromString(string(bytes)).Find("spec.featureGates.HotplugVolumes").(bool)
```
staebler, line 129: Rather than introducing a new module, use the methods in the the https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/unstructured package. This has the added benefit of also not having to re-marshal the data, too.

```suggestion
	enabled, exists, err := unstructured.NestedBool(hcUnstructured, "spec", "featureGates", "HotplugVolumes")
```

staebler (commented):
I suggest that you do not need the `gojsonq` module. But if you do keep it, then please move the vendor changes into a separate commit.

pkg/asset/installconfig/kubevirt/validation.go
ydayagi: Thanks. That's exactly what I was looking for.

pkg/asset/installconfig/kubevirt/client.go
ydayagi: @nirarg as we discussed, the permissions validation and example will be handled in separate task.
I opened [Validate user permissions during tenant install](https://issues.redhat.com/browse/CNV-10142)

pkg/asset/installconfig/platformprovisioncheck.go
```diff
@@ -8,6 +8,7 @@ import (
 	azconfig "github.com/openshift/installer/pkg/asset/installconfig/azure"
 	bmconfig "github.com/openshift/installer/pkg/asset/installconfig/baremetal"
 	gcpconfig "github.com/openshift/installer/pkg/asset/installconfig/gcp"
+	ickubevirt "github.com/openshift/installer/pkg/asset/installconfig/kubevirt"
```
nirarg, line 11: Please keep the naming convention like other platforms   (kubevirtconfig / kvconfig)

pkg/asset/installconfig/kubevirt/validation.go
```diff
@@ -21,6 +22,25 @@ func Validate(ic *types.InstallConfig, client Client) error {
...
+// ValidateForProvisioning is called by PlatformProvisionCheck
+func ValidateForProvisioning(client Client) error {
+	hcUnstructured, err := client.GetHyperConverged(context.Background(), "kubevirt-hyperconverged", "openshift-cnv")
+	if err != nil {
+		return fmt.Errorf("failed to get resource openshift-cnv/kubevirt-hyperconverged, with error: %v", err)
+	}
+
+	enabled, found, err := unstructured.NestedBool(hcUnstructured.Object, "spec", "featureGates", "HotplugVolumes")
+	if err != nil {
+		return err
```
nirarg, line 34: Maybe need to add here readable relevant description to the error (would be displayed to user)

pkg/asset/installconfig/platformprovisioncheck.go
ydayagi: done

pkg/asset/installconfig/kubevirt/validation.go
ydayagi: An error here is something not expected. It happens if JSON is malformed or some other internal error.

ydayagi: done

staebler (commented):
/approve

nirarg:
/hold
need to validate e2e-kubevirt passed before the merge

ydayagi:
We put the PR on hold for now.
We depend on a feature in CNV 2.6.
After release of CNV 2.6 we will continue with this PR.

PR #4653, opened 2021-02-14
Bug 1928537: bump ignition to v3_2

yuvalk:
my attempt at fixing https://bugzilla.redhat.com/show_bug.cgi?id=1928537


staebler (commented):
These changes look good. You are just missing a similar change in the new `pkg/asset/ignition/bootstrap/bootstrap_in_place.go` file.

yuvalk:
> These changes look good. You are just missing a similar change in the new `pkg/asset/ignition/bootstrap/bootstrap_in_place.go` file.

fixed, though it is weird as I was using:
`git grep -l 'v3_1' | xargs sed -i 's/v3_1/v3_2/g'

yuvalk:
I've re-added v3_1 as it is required for the terraform-provider
but I guess an update there is imminent too ?

cgwalters:
A fundamental issue here is that by bumping the baseline Ignition version, we're now disallowing older RHCOS versions (e.g. 4.6) from reading configs generated by the new installer, *even in cases where new features aren't used*.  In other words, it's clearly not going to work to use the new Ignition Tang+LUKS for 4.6 disk images, but it shouldn't break if that feature isn't used.

This issue actually mirrors a MCO bug - see https://github.com/coreos/ign-converter/issues/22 for some work on teaching Ignition how to downgrade to the earliest version it can.

yuvalk:
> A fundamental issue here is that by bumping the baseline Ignition version, we're now disallowing older RHCOS versions (e.g. 4.6) from reading configs generated by the new installer, _even in cases where new features aren't used_. In other words, it's clearly not going to work to use the new Ignition Tang+LUKS for 4.6 disk images, but it shouldn't break if that feature isn't used.
> 
> This issue actually mirrors a MCO bug - see [coreos/ign-converter#22](https://github.com/coreos/ign-converter/issues/22) for some work on teaching Ignition how to downgrade to the earliest version it can.

the problem is that ignition itself is not fully backward compatible
https://github.com/coreos/ignition/blob/main/docs/migrating-configs.md#upgrading-configs

so if installer is not using latest ignition, there are features in rhcos that cant be used, NBDE for example.
ie - this broke NBDE support in 4.7.

cgwalters:
> the problem is that ignition itself is not fully backward compatible

It is.

> so if installer is not using latest ignition, there are features in rhcos that cant be used, NBDE for example.

Please re-read my comment - I am basically saying that openshift/installer should be using the same ign-converter PR to downgrade to the minimal Ignition version it can, for the same reasons as the MCO.  This would allow using NBDE, but avoid breaking use of 4.6 bootimages if NBDE is *not* used.

kikisdeliveryservice:
As per Colin's comments above, is there any further action intend on this?

PR #4655, opened 2021-02-15
aws/session.go: include another internal provider name for static creds

abhinavdahiya:
AWS SDK session uses another const for loading credentials from the ennvironment
see [1]. This is same as the env provider and should be treated like one.

[1]: https://github.com/aws/aws-sdk-go/blob/v1.37.10/aws/session/env_config.go#L16

/assign @staebler 
/cc @patrickdillon

### DIFF
--- a/pkg/asset/installconfig/aws/session.go
+++ b/pkg/asset/installconfig/aws/session.go
@@ -142,7 +142,7 @@ func getCredentialsFromSession(options session.Options) (*credentials.Credential
 // static credentials safe for installer to transfer to cluster for use as-is.
 func IsStaticCredentials(credsValue credentials.Value) bool {
 	switch credsValue.ProviderName {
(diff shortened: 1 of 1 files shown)


staebler:
With the installer from master, I get the following error when I have AWS_SECRET_ACCESS_KEY and AWS_ACCESS_KEY_ID set.
```
$ ./bin/openshift-install create ignition-configs --dir install-dir
INFO Credentials loaded from default AWS environment variables 
FATAL failed to fetch Bootstrap Ignition Config: failed to fetch dependency of "Bootstrap Ignition Config": failed to generate asset "Openshift Manifests": AWS credentials provided by EnvConfigCredentials are not valid for default credentials mode 
```
With the changes from this PR, the command completes successfully.
```
$ ./bin/openshift-install create ignition-configs --dir install-dir
INFO Credentials loaded from default AWS environment variables 
INFO Ignition-Configs created in: install-dir and install-dir/auth 
```

/lgtm
/approve

PR #4656, opened 2021-02-15
Bug 1917931: Fix GCP openssl not found error

rna-afk:
With the current code, yum install pyopenssl is used to install
the openssl package that gcloud sdk needs to signurl which has
stopped working because gcloud is not able to pick up the package.

Added a code to install pip of a version less than 21 (pip support
for python 2 ends with pip version 21) and installed pyopenssl
which is the recommended way to install openssl for gcloud.

### DIFF
--- a/images/installer/Dockerfile.upi.ci
+++ b/images/installer/Dockerfile.upi.ci
@@ -3,14 +3,14 @@
 # It also contains the `upi` directory that contains various terraform and cloud formation templates that are used to create infrastructure resources.
 
 
-FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.14 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.8 AS builder
 WORKDIR /go/src/github.com/openshift/installer
 COPY . .
 RUN hack/build.sh
 
-FROM registry.svc.ci.openshift.org/ocp/4.6:cli as cli
+FROM registry.ci.openshift.org/ocp/4.8:cli as cli
 
-FROM registry.svc.ci.openshift.org/ocp/4.1:base
+FROM registry.ci.openshift.org/ocp/4.8:base
 COPY --from=cli /usr/bin/oc /bin/oc
 COPY --from=builder /go/src/github.com/openshift/installer/bin/openshift-install /bin/openshift-install
 COPY --from=builder /go/src/github.com/openshift/installer/upi /var/lib/openshift-install/upi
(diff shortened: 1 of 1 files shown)


staebler:
```
STEP 19: RUN pip install --user pyopenssl
/bin/sh: pip: command not found 
```

staebler:
The pip script is being installed to /output/.local/bin, but the PATH env is set to /bin.

rna-afk:
> The pip script is being installed to /output/.local/bin, but the PATH env is set to /bin.

Yeah was thinking of this but now I'm thinking if gsutil would also face this problem.

patrickdillon:
Not related to the BZ at hand, but do these need to be updated?? 
https://github.com/openshift/installer/pull/4656/files#diff-b6825bbac221d8a688bd04adf109e93e31ddbc78198db9a1859a59f793272e3fL11-L13

tsze-redhat:
Is there a BZ associated with this that needs to be verified?
The linked BZ (191793 doesn't look like an installer related bug.
Thanks.

staebler:
> Is there a BZ associated with this that needs to be verified?
> The linked BZ (191793 doesn't look like an installer related bug.
> Thanks.

Fixed. Thanks.

/bugzilla refresh

staebler:
> Is there a BZ associated with this that needs to be verified?
> The linked BZ (191793 doesn't look like an installer related bug.
> Thanks.

@tsze-redhat Sorry that I glossed over this when I originally saw your comment, but the BZ affects CI only so should not need QE verification.

staebler:
I guess this buys us at least a handful of months.
```
STEP 19: RUN python -m pip install pyopenssl
DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. pip 21.0 will drop support for Python 2.7 in January 2021. More details about Python 2 support in pip can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support pip 21.0 will remove support for this functionality.
```

patrickdillon:
> I guess this buys us at least a handful of months.
> 
> ```
> STEP 19: RUN python -m pip install pyopenssl
> DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. pip 21.0 will drop support for Python 2.7 in January 2021. More details about Python 2 support in pip can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support pip 21.0 will remove support for this functionality.
> ```

This seems like there should be some upgra

rna-afk:
Python 2.7 is the latest version in the Python2 family so I guess the upgrade would be from python2 to 3. Pip has stopped support for Python2 family as of pip version 21 which is why I did the 'pip<21' condition. I wasn't sure python 3 is supported with gcloud because when I looked for issues in gcloud, python2 was always mentioned but I think the latest gcloud version should support python3.

staebler:
In the latest e2e-gcp-upi test, the bootstrap machine is created and runs successfully. That is sufficient for verifying that the changes in this PR are valid.

/approve
/lgtm

staebler:

These changes are limited to the UPI CI image. That image is not used by the e2e-aws-upgrade job.

staebler:
The e2e-aws-upi job has not passed since August. When it does get passed setting up the infrastructure, it fails waiting for the bootstrap machine to set up the control plane. That is the same failure that we see in this PR, so it is unlikely that the changes in this PR are contributing to the situation. We need to figure out what is broken about the e2e-aws-upi tests, but we do not need to hold up this PR for that investigation.

/hold cancel

PR #4657, opened 2021-02-15
Bug 1928948: openstack: Allow to skip master replica validation

pierreprinetti:
With this patch, when the environment variable
OPENSHIFT_INSTALL_SKIP_PREFLIGHT_VALIDATIONS is set to "1", the
Installer skips the validation of the number of Control plane replicas.

### DIFF
--- a/pkg/asset/installconfig/openstack/validate.go
+++ b/pkg/asset/installconfig/openstack/validate.go
@@ -42,6 +42,11 @@ func Validate(ic *types.InstallConfig) error {
 
 // ValidateForProvisioning validates that the install config is valid for provisioning the cluster.
 func ValidateForProvisioning(ic *types.InstallConfig) error {
+	if skip := os.Getenv("OPENSHIFT_INSTALL_SKIP_PREFLIGHT_VALIDATIONS"); skip == "1" {
+		logrus.Warnf("OVERRIDE: pre-flight validation disabled.")
+		return nil
+	}
+
 	if ic.ControlPlane.Replicas != nil && *ic.ControlPlane.Replicas != 3 {
 		return field.Invalid(field.NewPath("controlPlane", "replicas"), ic.ControlPlane.Replicas, "control plane must be exactly three nodes when provisioning on OpenStack")
 	}


EmilienM (commented):
/lgtm

staebler:
The control plane replica check is not a pre-flight validation. It is a validation on the install config that has no dependence on the platform provider.

pierreprinetti:
@staebler 
I am confused by both statements. I'll try to be schematic:
1. the control plane replica check prevents installing; "pre-flight" is a loose term but I interpret it as "before installing", which AFAICS applies to this case
2. the validation is indeed [platform-specific](https://github.com/openshift/installer/pull/4585)

Can you help me understand what is your suggested course of action?

staebler:
My understanding of the purpose of the environment variable is that it is useful in CI where you either do not have access to a real OpenStack system or the OpenStack system does not support certain configurations that you want to test. Maybe I am missing how skipping this check helps with your CI tests.

From https://github.com/openshift/installer/pull/4452,

> Skipping pre-flight validations is essential to testing manifest
> generation when the exotic settings under test are not available in the
> CI infrastructure.

There are two types of validation-- static checks and pre-flight checks. A static check validates the install config in ways that do not require access to the platform provider (the OpenStack system in this case). A pre-flight check validates in ways that do require access to the platform provider. The validation of the number of control plane replicas is a static check, not a pre-flight check. If you are going to skip this validation, then you should also skip [the static validation in the installconfig asset](https://github.com/openshift/installer/blob/bd9db106a4cfe0fd2b2d594a8ff77352f6720ca9/pkg/asset/installconfig/installconfig.go#L149), which I do not believe is the behavior that we want.

pierreprinetti:
The purpose of this change is to enable @akrzos to test single-node clusters.

I am not going to pretend I know what's the best way around the validation.

staebler:
> The purpose of this change is to enable @akrzos to test single-node clusters.
> 
> I am not going to pretend I know what's the best way around the validation.

How can you perform a single-node installation when the terraform always creates 3 control-plane nodes?

staebler:
@pierreprinetti Does https://github.com/openshift/installer/pull/4672 obviate the need for these changes? Can we close this PR?

PR #4011, opened 2020-08-05
Packet fixes and improvements

mulbc:
* Fix boot issues with `c2.medium` instance types
* Make it easier to chose facility
* Add usage files to .gitignore

PR #4658, opened 2021-02-15
upi/metal: add more configuration options for bare metal

jhixson74:
Allow facilities, plan and hardware_reservation_id to be configured

Replace packet provider with new Equinix metal provider. The packet provider is now deprecated and now longer being worked on. All bugfixes now occur in the metal provider.

See https://bugzilla.redhat.com/show_bug.cgi?id=1925343

### DIFF
--- a/upi/metal/config.tf
+++ b/upi/metal/config.tf
@@ -167,41 +167,41 @@ EOF
 
 # ================PACKET=====================
 
-variable "packet_project_id" {
+variable "metal_project_id" {
   type = string
 
   description = <<EOF
-The Project ID for Packet.net where servers will be deployed.
+The Project ID for Equinix where servers will be deployed.
 EOF
 
 }
 
-variable "packet_plan" {
+variable "metal_plan" {
   type    = string
   default = "c1.small.x86"
 
   description = <<EOF
-The Packet.Net device plan slug.
+The Equinix device plan slug.
 EOF
 
 }
 
-variable "packet_facility" {
+variable "metal_facility" {
   type    = string
   default = "any"
 
   description = <<EOF
-The Packet.Net facilities code to be used.
+The Equinix facilities code to be used.
 EOF
 
 }
 
-variable "packet_hardware_reservation_id" {
+variable "metal_hardware_reservation_id" {
   type    = string
   default = ""
 
   description = <<EOF
-The UUID of the hardware reservation where you want this device deployed on Packet.net.
+The UUID of the hardware reservation where you want this device deployed on Equinix.
 EOF
 
 }
--- a/upi/metal/config.tf
+++ b/upi/metal/config.tf
@@ -133,7 +133,7 @@ EOF
 }
 
 variable "pxe_kernel_args" {
-  type = string
+  type    = string
   default = ""
 
   description = <<EOF
@@ -172,6 +172,36 @@ variable "packet_project_id" {
 
   description = <<EOF
 The Project ID for Packet.net where servers will be deployed.
+EOF
+
+}
+
+variable "packet_plan" {
+  type    = string
+  default = "c1.small.x86"
+
+  description = <<EOF
+The Packet.Net device plan slug.
+EOF
+
+}
+
+variable "packet_facility" {
+  type    = string
+  default = "any"
+
+  description = <<EOF
+The Packet.Net facilities code to be used.
+EOF
+
+}
+
+variable "packet_hardware_reservation_id" {
+  type    = string
+  default = ""
+
+  description = <<EOF
+The UUID of the hardware reservation where you want this device deployed on Packet.net.
 EOF
 
 }
--- a/upi/metal/main.tf
+++ b/upi/metal/main.tf
@@ -91,37 +91,37 @@ resource "matchbox_group" "worker" {
   }
 }
 
-# ================PACKET=====================
+# ================METAL=====================
 
-provider "packet" {
-  version = "3.2.0"
+provider "metal" {
+  version = "1.0.0"
 }
 
-resource "packet_device" "masters" {
+resource "metal_device" "masters" {
   count                   = var.master_count
   hostname                = "master-${count.index}.${var.cluster_domain}"
-  plan                    = var.packet_plan
-  facilities              = [var.packet_facility]
+  plan                    = var.metal_plan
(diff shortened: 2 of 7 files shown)


upi/metal/main.tf
```diff
@@ -143,6 +145,9 @@ module "bootstrap" {
 
   packet_facility   = "any"
```
staebler, line 146:
```suggestion
  packet_facility   = var.packet_facility
```

upi/metal/config.tf
```diff
@@ -176,6 +176,33 @@ EOF
 
 }
 
+variable "packet_plan" {
```
staebler, line 179: Let's put a default here what we were using previously, "c1.small.x86". That way if there are current users of the UPI terraform, they can continue to use their flow without any changes.

upi/metal/config.tf
```diff
@@ -176,6 +176,33 @@ EOF
...
+  type = string
+
+  description = <<EOF
+The Packet.Net device plan slug.
+EOF
+
+}
+
+variable "packet_facility" {
```
staebler, line 188: Default to "any"

upi/metal/config.tf
```diff
@@ -176,6 +176,33 @@ EOF
...
+  type = string
+
+  description = <<EOF
+The Packet.Net facilities code(s) to be used.
+EOF
+
+}
+
+variable "packet_hardware_reservation_id" {
```
staebler, line 197: Default to "".

upi/metal/terraform.tfvars.example
```diff
@@ -42,6 +42,15 @@ matchbox_trusted_ca_cert = "matchbox/tls/ca.crt"
 // The Project ID for Packet.net where servers will be deployed.
 packet_project_id = ""
 
+// The Packet.Net device plan slug.
+packet_plan = ""
```
staebler, line 46: Set defaults for `packet_plan` and `packet_facility` here, as well.

upi/metal/terraform.tfvars.example
```diff
@@ -42,6 +42,15 @@ matchbox_trusted_ca_cert = "matchbox/tls/ca.crt"
...
+
+// The Packet.Net facilities code(s) to be used.
+packet_facility = ""
+
+// The UUID of the hardware reservation where you want this device deployed on Packet.net.
```
staebler, line 51: It may be worth calling out here and in config.tf that the value "next_available" can be used instead of specifying a specific UUID.

upi/metal/main.tf
```diff
@@ -100,27 +100,29 @@ locals {
 }
```
staebler, line 100: Remove the local `packet_facility` variable, as it is unused (and has been unused).

upi/metal/config.tf
jhixson74: Fixed.

upi/metal/config.tf
jhixson74: Fixed.

upi/metal/terraform.tfvars.example
jhixson74: Fixed.

upi/metal/terraform.tfvars.example
jhixson74: Fixed.

upi/metal/main.tf
jhixson74: Fixed.

upi/metal/config.tf
jhixson74: Fixed.

upi/metal/config.tf
```diff
@@ -176,6 +176,36 @@ EOF
...
+EOF
+
+}
+
+variable "packet_facility" {
+  type    = string
+  default = "any"
+
+  description = <<EOF
+The Packet.Net facilities code(s) to be used.
```
staebler, line 194: nit: this can only be one code
```suggestion
The Packet.Net facilities code to be used.
```

upi/metal/terraform.tfvars.example
```diff
@@ -42,6 +42,16 @@ matchbox_trusted_ca_cert = "matchbox/tls/ca.crt"
...
 
+// The Packet.Net device plan slug.
+packet_plan = "c1.small.x86"
+
+// The Packet.Net facilities code(s) to be used.
+packet_facility = "any"
+
+// The UUID of the hardware reservation where you want this device deployed on Packet.net.
+// The value "next-available" can also be used instead of a specific UUID. 
+packet_hardware_reservation_id = "next-available"
```
staebler, line 53: This should be "" to preserve the current default behavior of not using reserved instances.

staebler (commented):
Minor things that we can fix up in a follow on. Let's get this in to fix the issue in CI.

upi/metal/config.tf
jhixson74: Fixed.

upi/metal/terraform.tfvars.example
jhixson74: Fixed.

staebler:
Now that https://github.com/openshift/release/pull/15880 has merged, I see the latest e2e-metal job has created c3.small.x86 machines on reserved instances in the ny5 datacenter, which is all what we expect.

jhixson74:
It looks like we need to remove pxe_os_image_url from the CI test.

staebler:
> It looks like we need to remove pxe_os_image_url from the CI test.

That is just a warning. It is not why the test failed.

The bootstrap machine is not getting some of the new customization variables.

Interestingly, terraform ended up creating two worker-2 machines and no worker-0 machines.

upi/metal/main.tf
staebler: This change did not end up getting made.

staebler:
There is something not right. There are multiples instances created. For example, looking at the equinox console, there is two each of master-0, master-1, master-2, and worker-1.

staebler:
@jhixson74 I wonder if any of the changes in https://github.com/openshift/installer/pull/4011 would help with the boot failures that we are seeing?

staebler:
Taking over this PR with https://github.com/openshift/installer/pull/4938.
/close

PR #4659, opened 2021-02-16
Bug 1873079: openstack: restrict SSH to machineNetwork CIDR

EmilienM:
TL;DR: SSH shouldn't be exposed to the public by default and if an
operator explicitly wants it, they should do it by modifying the
security group on day 2.

Reasons to not open SSH to all networks:

* SSH to an OCP cluster should be done only by advanced operators and
  shouldn't be encouraged, therefore not open by default that easily. Instead, an experienced operator should make configuration changes via `machineconfig` objects.

* Operators who know what they do should reach the nodes from a secure
  network (e.g. provider network or tenant network, ie internal). Not
  from a public network (e.g. Internet or any public faced network).

* Other cloud providers don't allow SSH from 0.0.0.0/0, we shouldn't be
  the one doing it.

* Running `oc debug` doesn't require SSH to be open from the client.

### DIFF
--- a/data/data/openstack/topology/sg-worker.tf
+++ b/data/data/openstack/topology/sg-worker.tf
@@ -17,13 +17,12 @@ resource "openstack_networking_secgroup_rule_v2" "worker_ingress_icmp" {
 }
 
 resource "openstack_networking_secgroup_rule_v2" "worker_ingress_ssh" {
-  direction      = "ingress"
-  ethertype      = "IPv4"
-  protocol       = "tcp"
-  port_range_min = 22
-  port_range_max = 22
-  # FIXME(mandre) AWS only allows SSH from cidr_block
-  remote_ip_prefix  = "0.0.0.0/0"
+  direction         = "ingress"
+  ethertype         = "IPv4"
+  protocol          = "tcp"
+  port_range_min    = 22
+  port_range_max    = 22
+  remote_ip_prefix  = var.cidr_block
   security_group_id = openstack_networking_secgroup_v2.worker.id
 }
 
--- a/data/data/openstack/topology/sg-master.tf
+++ b/data/data/openstack/topology/sg-master.tf
@@ -27,13 +27,12 @@ resource "openstack_networking_secgroup_rule_v2" "master_ingress_icmp" {
 }
 
 resource "openstack_networking_secgroup_rule_v2" "master_ingress_ssh" {
-  direction      = "ingress"
-  ethertype      = "IPv4"
-  protocol       = "tcp"
-  port_range_min = 22
-  port_range_max = 22
-  # FIXME(mandre) AWS only allows SSH from cidr_block
-  remote_ip_prefix  = "0.0.0.0/0"
+  direction         = "ingress"
+  ethertype         = "IPv4"
+  protocol          = "tcp"
+  port_range_min    = 22
+  port_range_max    = 22
+  remote_ip_prefix  = var.cidr_block
   security_group_id = openstack_networking_secgroup_v2.master.id
 }
 
--- a/upi/openstack/security-groups.yaml
+++ b/upi/openstack/security-groups.yaml
@@ -43,6 +43,7 @@
     os_security_group_rule:
       security_group: "{{ os_sg_master }}"
       protocol: tcp
+      remote_ip_prefix: "{{ os_subnet_range }}"
       port_range_min: 22
       port_range_max: 22
 
@@ -197,6 +198,7 @@
     os_security_group_rule:
       security_group: "{{ os_sg_worker }}"
       protocol: tcp
+      remote_ip_prefix: "{{ os_subnet_range }}"
       port_range_min: 22
       port_range_max: 22
 


mandre (commented):
We'll also have to do it for UPI too.

Also there are a couple more FIXMEs in that file (and possibly other divergences between openstack and aws's security groups), we should take action and address them as new PRs if possible:

```
# FIXME(mandre) AWS only allows API port from cidr_block
# FIXME(mandre) AWS only allows ICMP from cidr_block
# TODO(mandre) Explicitely enable egress
```

mdbooth:
@adduarte raised a good point on Slack that we've allowed this for some time and therefore it's possible customers have workflows which assume it. I can think of 3 reasonable uses for ssh. Can we come up with good alternatives:

1. Interactive shell access
2. Non-interactive command execution
3. Port forwarding

The doc text for `oc debug` suggests it should do cover both 1 and 2. However, I'll note that I can't shell to a node with `oc debug node/<node>` [as described in the docs](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.6/html/nodes/working-with-nodes) in my cluster.

[3] is a can of worms. Quite possibly a development workflow, for eg. I've used this myself for exposing a non-HTTP service during development.

It's also possible we'd get away with just switching it off, but it's a risk.

mandre:
>     3. Port forwarding
>
> [3] is a can of worms. Quite possibly a development workflow, for eg. I've used this myself for exposing a non-HTTP service during development.

For port forwarding we have the `oc port-forward` command so I think we should be covered.

EmilienM:
[...]
> The doc text for `oc debug` suggests it should do cover both 1 and 2. However, I'll note that I can't shell to a node with `oc debug node/<node>` [as described in the docs](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.6/html/nodes/working-with-nodes) in my cluster.

It works for me:

```
$ oc debug nodes/foch-ocp-zwtcg-master-0 --as-root                            
Creating debug namespace/openshift-debug-node-8mf4w ...
Starting pod/foch-ocp-zwtcg-master-0-debug ...
To use host binaries, run `chroot /host`
Pod IP: 192.168.25.165
If you don't see a command prompt, try pressing enter.
sh-4.4
```

```
$ ssh core@192.168.25.165
Connection refused
```

```
[stack@openstack] $ openstack security group rule show a4e9d50e-2161-408d-8057-d68ad2cd9cfb
+-------------------+--------------------------------------+
| Field             | Value                                |
+-------------------+--------------------------------------+
| created_at        | 2021-02-15T23:48:08Z                 |
| description       |                                      |
| direction         | ingress                              |
| ether_type        | IPv4                                 |
| id                | a4e9d50e-2161-408d-8057-d68ad2cd9cfb |
| name              | None                                 |
| port_range_max    | 22                                   |
| port_range_min    | 22                                   |
| project_id        | fa5725768fe04cdbb9f3e05a337e2898     |
| protocol          | tcp                                  |
| remote_group_id   | None                                 |
| remote_ip_prefix  | 192.168.25.0/24                      |
| revision_number   | 0                                    |
| security_group_id | 71af347b-c8a9-45d7-89c0-6d912dd87a4b |
| tags              | []                                   |
| updated_at        | 2021-02-15T23:48:08Z                 |
+-------------------+--------------------------------------+
```

> It's also possible we'd get away with just switching it off, but it's a risk.

Right it has a risk but it also improves the security, I think it's worth the risk if it's well documented and tested.

data/data/openstack/topology/sg-worker.tf
```diff
@@ -17,13 +17,12 @@ resource "openstack_networking_secgroup_rule_v2" "worker_ingress_icmp" {
...
 
 resource "openstack_networking_secgroup_rule_v2" "worker_ingress_ssh" {
-  direction      = "ingress"
-  ethertype      = "IPv4"
-  protocol       = "tcp"
-  port_range_min = 22
-  port_range_max = 22
-  # FIXME(mandre) AWS only allows SSH from cidr_block
-  remote_ip_prefix  = "0.0.0.0/0"
+  direction        = "ingress"
```
mandre, line 20: The missing space makes [ci/prow/tf-fmt angry](https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_installer/4659/pull-ci-openshift-installer-master-tf-fmt/1361671440692154368/build-log.txt):
```suggestion
  direction         = "ingress"
```

cgwalters:
I'm a fan of this.  Since it will only apply to newly provisioned clusters that seems sane - probably many customers don't read the release notes (this will definitely need to be there) but as long as we rank high in google/customer portal search for "openstack openshift ssh" that would help.  Or maybe at least just "openshift ssh" leads you to a doc page that describes the per-platform defaults around ssh.

adduarte:
I think merging this for master and 4.8  releases should be no problem since it will only affect new deployed clusters, which should have no user workflow stablished. 
We should be careful of backporting to older releases, but I think we seem ok. 
There is the "go gather" case we have to verify, I am not sure if the ci tests cover this enough.

staebler:
> There is the "go gather" case we have to verify, I am not sure if the ci tests cover this enough.

By the "go gather" do you mean the `openshift-install gather bootstrap` and `oc adm must-gather`. If so, then neither of those need external ssh to the masters.

adduarte:
awsome! so we are pretty safe to merge I think. 
Although, we might want to revisit at some point how much more we close down. 
This patch confines ssh to the cidr_block of the machineNetwork. that means machines inside that network can still reach ssh ports of the control-plane and workers. 
which is what we actually do in the UPI deployment example: 
https://github.com/openshift/installer/blob/master/upi/openstack/security-groups.yaml#L42

mandre (commented):
/lgtm
/approve

PR #4660, opened 2021-02-16
vSphere: mark imported virtual machines as template

jcpowermac:
From internal forums it was suggested that marking
the imported RHCOS virtual machines as a template
would be better than leaving it as a virtual machine.

This PR adds to the end of the `resourceVSpherePrivateImportOvaCreate`
fuction:

- From `info.Entity` get the VirtualMachine
- If the VirtualMachine exists, mark as a template

### DIFF
--- a/pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go
+++ b/pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go
@@ -363,6 +363,18 @@ func resourceVSpherePrivateImportOvaCreate(d *schema.ResourceData, meta interfac
 
 	log.Printf("[DEBUG] %s: ova import complete", d.Get("name").(string))
 
+	vm := object.NewVirtualMachine(client, info.Entity)
+	if vm == nil {
+		return fmt.Errorf("error VirtualMachine not found, managed object id: %s", d.Id())
+	}
+	log.Printf("[DEBUG] %s: mark as template", vm.Name())
+
+	err = vm.MarkAsTemplate(ctx)
+	if err != nil {
+		return errors.Errorf("failed to mark vm as template: %s", err)
+	}
+	log.Printf("[DEBUG] %s: mark as template complete", vm.Name())
+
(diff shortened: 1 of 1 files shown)


jcpowermac:
Seems to work as expected:
image

jcpowermac:
@jstuever can you take a look?

jcpowermac:
The e2e-vsphere openshift-test failure is long past where this change would have caused an issue.

jcpowermac:
Thats weird

/remove-label do-not-merge/work-in-progress

staebler:
/override ci/prow/e2e-ovirt

These changes have no effect on the e2e-ovirt tests.

mtulio:
@jcpowermac Just for curious: what should be the expected behavior of [`machine-api`](https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/vsphere/reconciler.go#L509-L515) when creating a new machine from a "VM Template" unless a "VM base" on current code base after marking the imported RHCOS VM as template? I don't know that lib to test if the current [**"Clone to VM"**](https://github.com/openshift/machine-api-operator/blob/master/pkg/controller/vsphere/reconciler.go#L652) will work similar to a desired option **"New VM From This Template..."** on vSphere.

lobziik:
@mtulio - I just checked manually on 4.7 with template, machine-api cloning behaviour did not change.

mtulio:
@lobziik good to know, thanks! 👍🏼  I will update the KCS to recovery referencing this PR as a good point to recovery the template VM, as a alternative to re-import the OVA.

jcpowermac:
@mtulio can you send the KCS to @rvanderp3 and I?
The only difference in a template is:

NAME | TYPE | VALUE
-- | -- | --
template | boolean | true

jcpowermac:
Under `VirtualMachineConfigInfo`
https://vdc-repo.vmware.com/vmwb-repository/dcr-public/723e7f8b-4f21-448b-a830-5f22fd931b01/5a8257bd-7f41-4423-9a73-03307535bd42/doc/vim.vm.ConfigInfo.html

PR #4662, opened 2021-02-16
use vendored mockgen for generating mock clients

staebler:
Rather than building an image to run mockgen and having that image always get the latest verion of mockgen, use the mockgen from the vendor directory. This has a couple benefits.
 1) We do not get changes to the client mocks that pop up in random PRs when a new version of mockgen is released.
 2) Contributors that do not have access to the CI registry can generate the mock clients.


pkg/dependencymagent/doc.go
```diff
@@ -0,0 +1,11 @@
+// +build tools
```
abhinavdahiya, line 1: any reason already existing setup for tools doesn't work here https://github.com/openshift/installer/blob/master/tools.go

staebler: Nope. I just did not realize it existed. I will move to that. Thanks.

staebler:
1f8f0ff6e...b788ca718
* Moved dependency of mockgen to existing tools.go instead of adding a new dependencymagnet folder.

jstuever:
/lgtm
This solves the issue I was experiencing as well.

PR #4666, opened 2021-02-17
Make copies of machine pools in machines asset

Fedosin:
Now, during generation of machines, we modify machine pools in the install config. It's not correct because by contract all assets are immutable after generation.

To prevent this, this commit creates copies of existing machine pools and all changes are made to it, leaving the original values intact.

### DIFF
--- a/pkg/asset/machines/master.go
+++ b/pkg/asset/machines/master.go
@@ -147,7 +147,7 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 
 	ic := installConfig.Config
 
-	pool := ic.ControlPlane
+	pool := *ic.ControlPlane
 	var err error
 	machines := []machineapi.Machine{}
 	switch ic.Platform.Name() {
@@ -199,7 +199,7 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 			clusterID.InfraID,
 			installConfig.Config.Platform.AWS.Region,
 			subnets,
-			pool,
+			&pool,
 			"master",
 			"master-user-data",
 			installConfig.Config.Platform.AWS.UserTags,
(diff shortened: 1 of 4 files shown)


staebler:
@Fedosin Before you go too much further with this, I think we should have a discussion about what problem you are trying to resolve with these changes. If you are trying to resolve the issue where the machine assets are mutating the install config asset, then I think the better solution would be to fix the machine assets so that they do not mutate. Is there some other reason compelling these changes beyond that?

EmilienM:
A commit message and a PR description would be nice to know the context a little bit.

pkg/asset/machines/master.go
```diff
@@ -147,7 +147,7 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 
 	ic := installConfig.Config
 
-	pool := ic.ControlPlane
+	pool := ic.ControlPlane.Copy()
```
staebler, line 150: I think we can get away with doing a shallow copy. None of the non-platform fields are changed and the platform fields are overwritten.
```suggestion
	pool := *ic.ControlPlane
```

pkg/tfvars/openstack/openstack.go
```diff
@@ -152,12 +152,12 @@ func TFVars(masterConfigs []*v1alpha1.OpenstackProviderSpec, cloud string, exter
 	}
 
 	cfg.AdditionalNetworkIDs = []string{}
-	if mpool.AdditionalNetworkIDs != nil {
+	if mpool != nil && mpool.AdditionalNetworkIDs != nil {
```
staebler, line 155: The check that `AdditionalNetworkIDs` is not nil is unnecessary.
```suggestion
	if mpool != nil {
```

pkg/tfvars/openstack/openstack.go
```diff
@@ -152,12 +152,12 @@ func TFVars(masterConfigs []*v1alpha1.OpenstackProviderSpec, cloud string, exter
...
 		cfg.AdditionalNetworkIDs = append(cfg.AdditionalNetworkIDs, mpool.AdditionalNetworkIDs...)
 	}
 
 	cfg.AdditionalSecurityGroupIDs = []string{}
-	if mpool.AdditionalSecurityGroupIDs != nil {
+	if mpool != nil && mpool.AdditionalSecurityGroupIDs != nil {
```
staebler, line 160:
```suggestion
	if mpool != nil {
```

staebler (commented):
/lgtm
/approve
/hold for tests from other platforms

staebler (commented):
/lgtm

PR #4668, opened 2021-02-18
Bump bootstrap in place MCS master ignition to v3_2

eranco74:
A recent [PR](https://github.com/openshift/installer/pull/4653) bumped the ignition generated by the installer to v3.2.
This PR aligns the master ignition that bootstrap in place is getting from the machine config server to the same version.
This is mainly relevant for assisted-installer that might want to merge user custom Ignition (that should match the ignition version generated by the installer) with this ignition config.

### DIFF
--- a/data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/bootstrap-in-place.sh
+++ b/data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/bootstrap-in-place.sh
@@ -26,7 +26,7 @@ fi
 if [ ! -f master-ignition.done ]; then
   echo "Creating master ignition and writing it to disk"
   # Get the master ignition from MCS
-  curl --header 'Accept:application/vnd.coreos.ignition+json;version=3.1.0' \
+  curl --header 'Accept:application/vnd.coreos.ignition+json;version=3.2.0' \
     http://localhost:22624/config/master -o /opt/openshift/original-master.ign
 
   GATHER_ID="bootstrap"


romfreiman:
@staebler @eranco74 

live-iso is expected to fail till this one will be merged:
https://github.com/openshift/assisted-test-infra/pull/540

For the same reason :man_shrugging:

staebler:
> @staebler @eranco74
> 
> live-iso is expected to fail till this one will be merged:
> [openshift/assisted-test-infra#540](https://github.com/openshift/assisted-test-infra/pull/540)
> 
> For the same reason man_shrugging

Can you help me find in the artifacts for the job where I can see the error causing the install to fail? Or is the job still not capturing adequate logs to show the failure?

If we wait for https://github.com/openshift/assisted-test-infra/pull/540 to merge, can we expect the job to pass for this PR? Or are they dependent such that both need to merge before the job will pass?

romfreiman:
I assume it should pass but I don't know for sure.
regarding logs, currently liveiso does not even boot since it gets 3.2.0
ignition on 4.6 rhcos so nothing cannot be done.
lets see what happens once test-infra merged. I'm interested as well.
I would not block on it at this point anyhow. but whatever you prefer.

romfreiman:
ok, after another thought, I don't understand how test infra can pass
without this one being merged.
so it would be great if you can merge once the required tests are done.

Thanks

staebler:
I have a couple of concerns.
1) Why do we have a test on the installer repo that is dependent upon a user (assisted-installer)? The e2e tests that verify the behavior of single-node bootstrap-in-place should not be running the assisted-installer.
2) Why are the tests not using the OS image appropriate for the release image rather than some hard-coded OS image? How do we plan to test different branches of the installer, which each may use different OS images?

romfreiman:
Regarding your concerns:
1. It's not assisted installer, but a test infra that we use also there. It knows how to take an iso, and boot vms from it. With ipv4, ipv6, customizations, and more. That's why we're using it. We discussed it in the past. 
2. Good point. It's a bug that we should fix. Although, usually, it would've not matter, since the only thing that it will do is cause another reboot. But yes, it should be fixed.

staebler:
>     1. It's not assisted installer, but a test infra that we use also there. It knows how to take an iso, and boot vms from it. With ipv4, ipv6, customizations, and more. That's why we're using it. We discussed it in the past.

By now you should know that you need to discuss things with me multiple times for them to sink in. What specifically have we discussed in the past?

I remember discussing that the installer CI jobs will not use the assisted installer.

I do not recall discussing that the installer CI jobs will use the assisted-test-infra repo to perform the installation. I am not strictly opposed to that. It did came as a shock to me to see the use of a repo where the first line of the README states "[t]his project deploys the OpenShift Assisted Installer in Minikube [...]". It does concern me slightly that this is another installation mechanism that the installer team will need to have at least a passing understanding of in order to maintain the bootstrap-in-place functionality in the future. At this point, looking at the build logs and artifacts collected, I am ignorant about what the errors may have been that caused the test to fail.

In any event, this discussion (while hopefully fruitful) is not really germane to this PR.

/lgtm

staebler:
/override ci/prow/e2e-ovirt

The e2e-ovirt job is broken. The changes in this PR do not effect that job..

PR #4670, opened 2021-02-18
proxy: Check if proxies are in cluster and service networks

rna-afk:
Adding a validation to check if the provided httpProxy and
httpsProxy are not a part of the service or cluster networks
to prevent customers from trying to proxy into those networks.

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -106,7 +106,7 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
 		allErrs = append(allErrs, field.Invalid(field.NewPath("pullSecret"), c.PullSecret, err.Error()))
 	}
 	if c.Proxy != nil {
-		allErrs = append(allErrs, validateProxy(c.Proxy, field.NewPath("proxy"))...)
+		allErrs = append(allErrs, validateProxy(c.Proxy, c, field.NewPath("proxy"))...)
 	}
 	allErrs = append(allErrs, validateImageContentSources(c.ImageContentSources, field.NewPath("imageContentSources"))...)
 	if _, ok := validPublishingStrategies[c.Publish]; !ok {
@@ -456,7 +456,7 @@ func validatePlatform(platform *types.Platform, fldPath *field.Path, network *ty
 	return allErrs
 }
 
-func validateProxy(p *types.Proxy, fldPath *field.Path) field.ErrorList {
+func validateProxy(p *types.Proxy, c *types.InstallConfig, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	if p.HTTPProxy == "" && p.HTTPSProxy == "" {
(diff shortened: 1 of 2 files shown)


pkg/types/validation/installconfig.go
```diff
@@ -456,17 +456,24 @@ func validatePlatform(platform *types.Platform, fldPath *field.Path, network *ty
 	return allErrs
 }
 
-func validateProxy(p *types.Proxy, fldPath *field.Path) field.ErrorList {
+func validateProxy(c *types.InstallConfig, fldPath *field.Path) field.ErrorList {
```
staebler, line 459: Pass in networking as a separate field, to enforce the concept that the proxy is being validated and not the entire install config.
```suggestion
func validateProxy(p *types.Proxy, networking *types.Networking, fldPath *field.Path) field.ErrorList {
```

pkg/types/validation/installconfig.go
```diff
@@ -571,3 +578,27 @@ func validateURI(uri string, fldPath *field.Path, schemes []string) field.ErrorL
...
 	return field.ErrorList{field.NotSupported(fldPath, parsed.Scheme, schemes)}
 }
+
+// validateIPProxy checks if the given proxy string is an IP and if so checks the service and
+// cluster networks and returns error if the IP belongs in them. Returns nil if the proxy is
+// not an IP address.
+func validateIPProxy(proxy string, n *types.Networking, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	proxyIP := net.ParseIP(proxy)
```
staebler, line 588: You need to get the IP from only the host.
```suggestion
	parsed, err := url.ParseRequestedURI(proxy)
	if err != nil {
		return allErrs
	}
	proxyIP = net.ParseIP(parsed.Host)
```

pkg/types/validation/installconfig.go
```diff
@@ -571,3 +578,27 @@ func validateURI(uri string, fldPath *field.Path, schemes []string) field.ErrorL
...
+	if proxyIP == nil {
+		return nil
+	}
+	for _, network := range n.ClusterNetwork {
+		if network.CIDR.Contains(proxyIP) {
+			allErrs = append(allErrs, field.Invalid(fldPath, proxy, "proxy value is part of the cluster networks"))
```
staebler, line 594: If the CIDRs overlap, then the proxy IP could be contained in multiple CIDRs. This would result in the same error duplicated. The CIDRs are not allowed to overlap and should be caught by other validation, but this validation runs at the same time.

staebler (commented):
Looks good. Could you add test cases for a proxy with an IP host, where one test case overlaps and fails validation and one test case does not overlap and succeeds validation?

staebler (commented):
/lgtm
/approve

staebler:
The install in the e2e-aws-proxy job was successful, which is sufficient for the changes in this PR.

PR #4672, opened 2021-02-19
openstack: allow for fewer than 3 masters

staebler:
Modify the terraform for IPI OpenStack so that it only creates at most the number of control planes requested by the user. In other words, if the user requests only one control plane node, then the terraform will only create master-0 and not master-1 nor master-2.

Relax the provsioning validation that enforces that the number of control plane replicas must be exactly 3 to allow for the control plane replicas to be at most 3.


pierreprinetti:
Nice solution for the use case.

/approve
/lgtm

staebler:
/override ci/prow/e2e-ovirt

Broken test not affected by changes in this PR.

PR #4674, opened 2021-02-21
Bug 1931217: add affinity groups to oVirt installer

Gal-Zaidman:
We want the installer to have the following behavior:

Default:
Create 2 unique affinity groups with a hard negative affinity rule:
- ovirt11-****-controlplane - contain the masters only and have a higher priority(5).
- ovirt11-****-compute - contain the compute only, priority(3).

Customizations:
1. No Affinity groups:
To not create any AffinityGroups the user needs to have an empty field on the install config for the platform Like:
```yaml
...
platform:
  ovirt:
    api_vip: xxx.xxx.xxx.xxx
    ingress_vip: xxx.xxx.xxx.xxx
    ovirt_cluster_id: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
    ovirt_storage_domain_id: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
    affinityGroups: []
...
```

2. Customize created affinity Groups:
The user can define the affinity groups that the installer will create by editing the affinityGroups field, for example:
```yaml
platform:
  ovirt:
    api_vip: xxx.xxx.xxx.xxx
    ingress_vip: xxx.xxx.xxx.xxx
    ovirt_cluster_id: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
    ovirt_storage_domain_id: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
    affinityGroups:
    - description: test1
      enforcing: true
      name: test1
      priority: 3
    - description: test2
        host
      enforcing: true
      name: test2
      priority: 5
...
```

3. Decide which machine pools will be added to the affinity groups:
The user can specify on the machine pool the affinity groups that those machines will join on creation, by editing the affinityGroupsNames field, those groups need to be created by the installer or existing in the ovirt cluster.
```yaml
...
compute:
- name: worker
  platform:
    ovirt:
      affinityGroupsNames:
      - test1
...
  replicas: 2
controlPlane:
...
  name: master
  platform:
    ovirt:
      affinityGroupsNames:
      - test1
      - test2
...
  replicas: 3
...

### DIFF
--- a/data/data/ovirt/affinity_group/main.tf
+++ b/data/data/ovirt/affinity_group/main.tf
@@ -1,0 +1,12 @@
+resource "ovirt_affinity_group" "affinity_groups" {
+  count        = length(var.ovirt_affinity_groups)
+  name         = var.ovirt_affinity_groups[count.index]["name"]
+  description  = var.ovirt_affinity_groups[count.index]["description"]
+  cluster_id   = var.ovirt_cluster_id
+  priority     = var.ovirt_affinity_groups[count.index]["priority"]
+  vm_positive  = false
+  vm_enforcing = var.ovirt_affinity_groups[count.index]["enforcing"]
+  lifecycle {
+    ignore_changes = [vm_list]
+  }
+}
--- a/data/data/ovirt/affinity_group/outputs.tf
+++ b/data/data/ovirt/affinity_group/outputs.tf
@@ -1,0 +1,3 @@
+output "ovirt_affinity_group_count" {
+  value = length(ovirt_affinity_group.affinity_groups)
+}
--- a/data/data/ovirt/affinity_group/variables.tf
+++ b/data/data/ovirt/affinity_group/variables.tf
@@ -1,0 +1,11 @@
+variable "ovirt_cluster_id" {
+  type        = string
+  description = "The ID of Cluster"
+}
+
+variable "ovirt_affinity_groups" {
+  type        = list(object({ name = string, priority = number, description = string, enforcing = string }))
+  description = "Control plane affinity groups that will be created."
+  default     = []
+}
+
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -553,6 +553,7 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			string(*rhcosImage),
 			clusterID.InfraID,
 			masters[0].Spec.ProviderSpec.Value.Object.(*ovirtprovider.OvirtMachineProviderSpec),
+			installConfig.Config.Platform.Ovirt.AffinityGroups,
 		)
 		if err != nil {
 			return errors.Wrapf(err, "failed to get %s Terraform variables", platform)
--- a/pkg/asset/installconfig/ovirt/client.go
+++ b/pkg/asset/installconfig/ovirt/client.go
@@ -69,3 +69,30 @@ func FetchVNICProfileByClusterNetwork(con *ovirtsdk.Connection, clusterID string
 	}
 	return nil, fmt.Errorf("there are no vNic profiles for the given cluster ID %s and network name %s", clusterID, networkName)
 }
+
+// GetClusterName returns the name of the ovirt cluster with the specified Cluster ID
+func GetClusterName(con *ovirtsdk.Connection, clusterID string) (string, error) {
+	clusterResponse, err := con.SystemService().ClustersService().ClusterService(clusterID).Get().Send()
+	if err != nil {
+		return "", fmt.Errorf("failed to fetch cluster %s (%w)", clusterID, err)
+	}
+	cluster, ok := clusterResponse.Cluster()
+	if !ok {
+		return "", fmt.Errorf("failed to find cluster with id %s", clusterID)
+	}
+	return cluster.MustName(), nil
+}
+
+// FindHostsInCluster returns a list of the hosts that exists in the specified cluster
+func FindHostsInCluster(con *ovirtsdk.Connection, cName string) ([]*ovirtsdk.Host, error) {
+	res, err := con.SystemService().HostsService().
+		List().Search(fmt.Sprintf("cluster=%s", cName)).Send()
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch hosts for cluster %s (%w)", cName, err)
+	}
+	hosts, ok := res.Hosts()
+	if !ok {
+		return nil, fmt.Errorf("failed to find hosts in cluster %s", cName)
+	}
+	return hosts.Slice(), nil
+}
--- a/pkg/asset/installconfig/ovirt/validation.go
+++ b/pkg/asset/installconfig/ovirt/validation.go
@@ -5,6 +5,7 @@ import (
 
 	ovirtsdk "github.com/ovirt/go-ovirt"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"gopkg.in/AlecAivazis/survey.v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
@@ -39,27 +40,108 @@ func Validate(ic *types.InstallConfig) error {
 			allErrs,
 			field.Invalid(ovirtPlatformPath.Child("vnicProfileID"), ic.Ovirt.VNICProfileID, err.Error()))
 	}
-
 	if ic.ControlPlane != nil && ic.ControlPlane.Platform.Ovirt != nil {
 		allErrs = append(
 			allErrs,
-			validateMachinePool(con, field.NewPath("controlPlane", "platform", "ovirt"), ic.ControlPlane.Platform.Ovirt)...)
+			validateMachinePool(con, field.NewPath("controlPlane", "platform", "ovirt"), ic.ControlPlane.Platform.Ovirt, *ic.Ovirt)...)
 	}
 	for idx, compute := range ic.Compute {
 		fldPath := field.NewPath("compute").Index(idx)
 		if compute.Platform.Ovirt != nil {
 			allErrs = append(
 				allErrs,
-				validateMachinePool(con, fldPath.Child("platform", "ovirt"), compute.Platform.Ovirt)...)
+				validateMachinePool(con, fldPath.Child("platform", "ovirt"), compute.Platform.Ovirt, *ic.Ovirt)...)
 		}
 	}
+
+	allErrs = append(
+		allErrs,
+		validateAffinityGroups(ic, ovirtPlatformPath.Child("affinityGroups"), con)...)
 
 	return allErrs.ToAggregate()
 }
 
-func validateMachinePool(con *ovirtsdk.Connection, child *field.Path, pool *ovirt.MachinePool) field.ErrorList {
+func validateMachinePool(con *ovirtsdk.Connection, child *field.Path, pool *ovirt.MachinePool, platform ovirt.Platform) field.ErrorList {
 	allErrs := field.ErrorList{}
 	allErrs = append(allErrs, validateInstanceTypeID(con, child, pool)...)
+	allErrs = append(allErrs, validateMachineAffinityGroups(con, child, pool, platform)...)
+
+	return allErrs
+}
+
+// validateAffinityGroups validates that the affinity group definitions on all machinePools are valid
+// - Affinity group contains valid fields
+// - Affinity group doesn't already exist in the cluster
+// - oVirt cluster has sufficient resources to fulfil the affinity group constraints
+func validateAffinityGroups(ic *types.InstallConfig, fldPath *field.Path, con *ovirtsdk.Connection) field.ErrorList {
+	allErrs := field.ErrorList{}
+	allErrs = append(allErrs, validateExistingAffinityGroup(con, *ic.Ovirt, fldPath)...)
+	allErrs = append(allErrs, validateClusterResources(con, ic, fldPath)...)
+	return allErrs
+}
+
+// validateExistingAffinityGroup checks that there is no affinity group with the same name in the cluster
+func validateExistingAffinityGroup(con *ovirtsdk.Connection, platform ovirt.Platform, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	res, err := con.SystemService().ClustersService().
+		ClusterService(platform.ClusterID).AffinityGroupsService().List().Send()
+	if err != nil {
+		return append(
+			allErrs,
+			field.InternalError(
+				fldPath,
+				errors.Errorf("failed listing affinity groups for cluster %v", platform.ClusterID)))
+	}
+	for _, ag := range res.MustGroups().Slice() {
+		for _, agNew := range platform.AffinityGroups {
+			if ag.MustName() == agNew.Name {
+				allErrs = append(
+					allErrs,
+					field.Invalid(
+						fldPath,
+						ag,
+						fmt.Sprintf("affinity group %v already exist in cluster %v", agNew.Name, platform.ClusterID)))
+			}
+		}
+	}
+	return allErrs
+}
+
+func validateClusterResources(con *ovirtsdk.Connection, ic *types.InstallConfig, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	mAgReplicas := make(map[string]int)
+	for _, agn := range ic.ControlPlane.Platform.Ovirt.AffinityGroupsNames {
+		mAgReplicas[agn] = mAgReplicas[agn] + int(*ic.ControlPlane.Replicas)
+	}
+	for _, compute := range ic.Compute {
+		for _, agn := range compute.Platform.Ovirt.AffinityGroupsNames {
+			mAgReplicas[agn] = mAgReplicas[agn] + int(*compute.Replicas)
+		}
+	}
+
+	clusterName, err := GetClusterName(con, ic.Ovirt.ClusterID)
+	if err != nil {
+		return append(allErrs, field.InternalError(fldPath, err))
+	}
+	hosts, err := FindHostsInCluster(con, clusterName)
+	if err != nil {
+		return append(allErrs, field.InternalError(fldPath, err))
+	}
+	for _, ag := range ic.Ovirt.AffinityGroups {
+		if _, found := mAgReplicas[ag.Name]; found {
+			if len(hosts) < mAgReplicas[ag.Name] {
+				msg := fmt.Sprintf("Affinity Group %v cannot be fulfilled, oVirt cluster doesn't"+
+					"have enough hosts: found %v hosts but %v replicas assigned to affinity group",
+					ag.Name, len(hosts), mAgReplicas[ag.Name])
+				if ag.Enforcing {
+					allErrs = append(allErrs, field.Invalid(fldPath, ag, msg))
+				} else {
+					logrus.Warning(msg)
+				}
+			}
+		}
+	}
 	return allErrs
 }
 
@@ -69,6 +151,41 @@ func validateInstanceTypeID(con *ovirtsdk.Connection, child *field.Path, machine
 		_, err := con.SystemService().InstanceTypesService().InstanceTypeService(machinePool.InstanceTypeID).Get().Send()
 		if err != nil {
 			allErrs = append(allErrs, field.NotFound(child.Child("instanceTypeID"), machinePool.InstanceTypeID))
+		}
+	}
+	return allErrs
+}
+
+// validateMachineAffinityGroups checks that the affinity groups on the machine object exist in the oVirt cluster
+// or created by the installer.
+func validateMachineAffinityGroups(con *ovirtsdk.Connection, child *field.Path, machinePool *ovirt.MachinePool, platform ovirt.Platform) field.ErrorList {
+	allErrs := field.ErrorList{}
+	existingAG := make(map[string]int)
+
+	res, err := con.SystemService().ClustersService().
+		ClusterService(platform.ClusterID).AffinityGroupsService().List().Send()
+	if err != nil {
+		return append(
+			allErrs,
+			field.InternalError(
+				child,
+				errors.Errorf("failed listing affinity groups for cluster %v", platform.ClusterID)))
+	}
+	for _, ag := range res.MustGroups().Slice() {
+		existingAG[ag.MustName()] = 0
+	}
+	// add affinity groups the installer creates
+	for _, ag := range platform.AffinityGroups {
+		existingAG[ag.Name] = 0
+	}
+	for _, ag := range machinePool.AffinityGroupsNames {
+		if _, ok := existingAG[ag]; !ok {
+			allErrs = append(
+				allErrs,
+				field.Invalid(
+					child.Child("affinityGroupsNames"), ag,
+					fmt.Sprintf(
+						"Affinity Group %v doesn't exist in oVirt cluster or created by the installer", ag)))
 		}
 	}
 	return allErrs
--- a/pkg/asset/machines/ovirt/machines.go
+++ b/pkg/asset/machines/ovirt/machines.go
@@ -4,6 +4,7 @@ package ovirt
 import (
 	"fmt"
 
+	ovirtprovider "github.com/openshift/cluster-api-provider-ovirt/pkg/apis/ovirtprovider/v1beta1"
 	machineapi "github.com/openshift/machine-api-operator/pkg/apis/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -11,8 +12,6 @@ import (
 
 	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/ovirt"
-
-	ovirtprovider "github.com/openshift/cluster-api-provider-ovirt/pkg/apis/ovirtprovider/v1beta1"
 )
 
 // Machines returns a list of machines for a machinepool.
@@ -29,7 +28,7 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 	if pool.Replicas != nil {
 		total = *pool.Replicas
 	}
-	provider := provider(platform, pool, userDataSecret, osImage)
+	provider := provider(platform, pool, userDataSecret, clusterID, osImage)
 	var machines []machineapi.Machine
 	for idx := int64(0); idx < total; idx++ {
 		machine := machineapi.Machine{
@@ -59,7 +58,7 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 	return machines, nil
 }
 
-func provider(platform *ovirt.Platform, pool *types.MachinePool, userDataSecret string, osImage string) *ovirtprovider.OvirtMachineProviderSpec {
+func provider(platform *ovirt.Platform, pool *types.MachinePool, userDataSecret string, clusterID string, osImage string) *ovirtprovider.OvirtMachineProviderSpec {
 	spec := ovirtprovider.OvirtMachineProviderSpec{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "ovirtproviderconfig.machine.openshift.io/v1beta1",
@@ -73,6 +72,20 @@ func provider(platform *ovirt.Platform, pool *types.MachinePool, userDataSecret 
 		MemoryMB:          pool.Platform.Ovirt.MemoryMB,
 		VMType:            string(pool.Platform.Ovirt.VMType),
 	}
+	uniqueNewAG := make(map[string]ovirt.AffinityGroup)
+	for _, ag := range platform.AffinityGroups {
+		uniqueNewAG[ag.Name] = ag
+	}
+	ags := make([]string, len(pool.Platform.Ovirt.AffinityGroupsNames))
+	for i, agName := range pool.Platform.Ovirt.AffinityGroupsNames {
+		if _, ok := uniqueNewAG[agName]; ok {
+			// add the cluster name only if the affinity group is created by the installer
+			ags[i] = clusterID + "-" + agName
+		} else {
+			ags[i] = agName
+		}
+	}
+	spec.AffinityGroupsNames = ags
 	if pool.Platform.Ovirt.CPU != nil {
 		spec.CPU = &ovirtprovider.CPU{
 			Cores:   pool.Platform.Ovirt.CPU.Cores,
--- a/pkg/destroy/ovirt/destroyer.go
+++ b/pkg/destroy/ovirt/destroyer.go
@@ -2,6 +2,7 @@ package ovirt
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -42,6 +43,9 @@ func (uninstaller *ClusterUninstaller) Run() error {
 	}
 	if err := uninstaller.removeTemplate(con); err != nil {
 		uninstaller.Logger.Errorf("Failed to remove template: %s", err)
+	}
+	if err := uninstaller.removeAffinityGroups(con); err != nil {
+		uninstaller.Logger.Errorf("Failed to removing Affinity Groups: %s", err)
 	}
 
 	return nil
@@ -138,6 +142,25 @@ func (uninstaller *ClusterUninstaller) removeTemplate(con *ovirtsdk.Connection) 
 				if err != nil {
 					return err
 				}
+			}
+		}
+	}
+	return nil
+}
+
+func (uninstaller *ClusterUninstaller) removeAffinityGroups(con *ovirtsdk.Connection) error {
+	cID := uninstaller.Metadata.Ovirt.ClusterID
+	affinityGroupService := con.SystemService().ClustersService().ClusterService(cID).AffinityGroupsService()
+	res, err := affinityGroupService.List().Send()
+	if err != nil {
+		return err
+	}
+	for _, ag := range res.MustGroups().Slice() {
+		if strings.HasPrefix(ag.MustName(), fmt.Sprintf("%s-", uninstaller.Metadata.InfraID)) {
+			uninstaller.Logger.Infof("Removing AffinityGroup %s", ag.MustName())
+			_, err := affinityGroupService.GroupService(ag.MustId()).Remove().Send()
+			if err != nil {
+				uninstaller.Logger.Errorf("failed to remove AffinityGroup: %s", err)
 			}
 		}
 	}
--- a/pkg/tfvars/ovirt/ovirt.go
+++ b/pkg/tfvars/ovirt/ovirt.go
@@ -3,11 +3,13 @@ package ovirt
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/openshift/cluster-api-provider-ovirt/pkg/apis/ovirtprovider/v1beta1"
 
 	"github.com/openshift/installer/pkg/rhcos"
 	"github.com/openshift/installer/pkg/tfvars/internal/cache"
+	"github.com/openshift/installer/pkg/types/ovirt"
 )
 
 // Auth is the collection of credentials that will be used by terraform.
@@ -22,18 +24,20 @@ type Auth struct {
 
 type config struct {
 	Auth                   `json:",inline"`
-	ClusterID              string `json:"ovirt_cluster_id"`
-	StorageDomainID        string `json:"ovirt_storage_domain_id"`
-	NetworkName            string `json:"ovirt_network_name,omitempty"`
-	VNICProfileID          string `json:"ovirt_vnic_profile_id,omitempty"`
-	BaseImageName          string `json:"openstack_base_image_name,omitempty"`
-	BaseImageLocalFilePath string `json:"openstack_base_image_local_file_path,omitempty"`
-	MasterInstanceTypeID   string `json:"ovirt_master_instance_type_id"`
-	MasterVMType           string `json:"ovirt_master_vm_type,omitempty"`
-	MasterMemory           int32  `json:"ovirt_master_memory"`
-	MasterCores            int32  `json:"ovirt_master_cores"`
-	MasterSockets          int32  `json:"ovirt_master_sockets"`
-	MasterOsDiskGB         int64  `json:"ovirt_master_os_disk_gb"`
+	ClusterID              string                   `json:"ovirt_cluster_id"`
+	StorageDomainID        string                   `json:"ovirt_storage_domain_id"`
+	NetworkName            string                   `json:"ovirt_network_name,omitempty"`
+	VNICProfileID          string                   `json:"ovirt_vnic_profile_id,omitempty"`
+	AffinityGroups         []map[string]interface{} `json:"ovirt_affinity_groups,omitempty"`
+	BaseImageName          string                   `json:"openstack_base_image_name,omitempty"`
+	BaseImageLocalFilePath string                   `json:"openstack_base_image_local_file_path,omitempty"`
+	MasterInstanceTypeID   string                   `json:"ovirt_master_instance_type_id"`
+	MasterVMType           string                   `json:"ovirt_master_vm_type,omitempty"`
+	MasterMemory           int32                    `json:"ovirt_master_memory"`
+	MasterCores            int32                    `json:"ovirt_master_cores"`
+	MasterSockets          int32                    `json:"ovirt_master_sockets"`
+	MasterOsDiskGB         int64                    `json:"ovirt_master_os_disk_gb"`
+	MasterAffinityGroups   []string                 `json:"ovirt_master_affinity_groups"`
 }
 
 // TFVars generates ovirt-specific Terraform variables.
@@ -45,8 +49,9 @@ func TFVars(
 	vnicProfileID string,
 	baseImage string,
 	infraID string,
-	masterSpec *v1beta1.OvirtMachineProviderSpec) ([]byte, error) {
-
+	masterSpec *v1beta1.OvirtMachineProviderSpec,
+	affinityGroups []ovirt.AffinityGroup,
+) ([]byte, error) {
 	cfg := config{
 		Auth:                 auth,
 		ClusterID:            clusterID,
@@ -58,6 +63,7 @@ func TFVars(
 		MasterVMType:         masterSpec.VMType,
 		MasterOsDiskGB:       masterSpec.OSDisk.SizeGB,
 		MasterMemory:         masterSpec.MemoryMB,
+		MasterAffinityGroups: masterSpec.AffinityGroupsNames,
 	}
 	if masterSpec.CPU != nil {
 		cfg.MasterCores = masterSpec.CPU.Cores
@@ -73,6 +79,21 @@ func TFVars(
 		}
 		cfg.BaseImageLocalFilePath = imageFilePath
 	}
-
+	if len(affinityGroups) > 0 {
+		cfg.AffinityGroups = handleAffinityGroups(affinityGroups, infraID)
+	}
 	return json.MarshalIndent(cfg, "", "  ")
+}
+
+func handleAffinityGroups(ags []ovirt.AffinityGroup, infraID string) []map[string]interface{} {
+	tfAffinityGroups := make([]map[string]interface{}, len(ags))
+	for i, ag := range ags {
+		tfAffinityGroups[i] = map[string]interface{}{
+			"name":        fmt.Sprintf("%s-%s", infraID, ag.Name),
+			"priority":    ag.Priority,
+			"description": ag.Description,
+			"enforcing":   ag.Enforcing,
+		}
+	}
+	return tfAffinityGroups
 }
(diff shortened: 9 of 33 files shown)

### CODE REVIEW

**Comment by Gal-Zaidman** [comment]
/hold

Need https://github.com/openshift/cluster-api-provider-ovirt/pull/80 to work

File: data/data/ovirt/affinity_group/main.tf [hunk 1]
```diff
@@ -0,0 +1,18 @@
+resource "ovirt_affinity_group" "cluster_masters_affinity_group" {
+  count        = length(var.ovirt_control_plane_affinity_groups)
+  name         = var.ovirt_control_plane_affinity_groups[count.index]["name"]
+  cluster_id   = var.ovirt_cluster_id
+  priority     = var.ovirt_control_plane_affinity_groups[count.index]["priority"]
+  vm_positive  = false
+  vm_enforcing = false
```
janoszen, line 7: `vm_enforcing=false` means that the masters MAY end up on the same physical host if the user does not have enough hosts instead of failing. This can lead to inadvertently having a single point of failure.

This option should be customizable and I would strongly argue for setting the default value to `true` to stay in line with all Kubernetes deployment recommendations that I know of.

data/data/ovirt/affinity_group/main.tf
```diff
@@ -0,0 +1,18 @@
...
+  vm_list      = var.masters_ids
+}
+
+resource "ovirt_affinity_group" "cluster_affinity_group" {
+  count        = length(var.ovirt_compute_affinity_groups)
+  name         = var.ovirt_compute_affinity_groups[count.index]["name"]
+  cluster_id   = var.ovirt_cluster_id
+  priority     = var.ovirt_compute_affinity_groups[count.index]["priority"]
+  vm_positive  = false
+  vm_enforcing = false
```
janoszen, line 17: This option should be customizable. Furthermore, there should be a customizable description with the following text:

> AG for spreading the load out to multiple hosts without hard enforcement.

janoszen (changes requested):

pkg/tfvars/ovirt/ovirt.go
```diff
@@ -71,6 +78,50 @@ func TFVars(
...
-
+	if len(controlPlaneAffinityGroups) > 0 {
+		cfg.ControlPlaneAffinityGroups = createControlPlaneAffinityGroups(controlPlaneAffinityGroups, infraID)
+	}
+	cfg.ComputeAffinityGroups = createComputeAffinityGroups(
+		computeMachines, controlPlaneAffinityGroups, infraID)
 	return json.MarshalIndent(cfg, "", "  ")
 }
+
+func createControlPlaneAffinityGroups(masterAffiniryGroups []*ovirt.AffinityGroup, infraID string) []map[string]interface{} {
```
janoszen, line 89: Typo!

pkg/tfvars/ovirt/ovirt.go
```diff
@@ -71,6 +78,50 @@ func TFVars(
...
+	var agMap []map[string]interface{}
+	for _, ag := range masterAffiniryGroups {
+		if ag.Name != "" {
```
janoszen, line 92: Can we please fail if the user doesn't input a name for the AG?

pkg/tfvars/ovirt/ovirt.go
```diff
@@ -71,6 +78,50 @@ func TFVars(
...
+	return agMap
+}
+
+func createComputeAffinityGroups(computeMachines []*ovirt.MachinePool, masterAffiniryGroups []*ovirt.AffinityGroup, infraID string) []map[string]interface{} {
+	var agMap []map[string]interface{}
+	uniqeAGName := make(map[string]*ovirt.AffinityGroup)
+
+	for _, pool := range computeMachines {
+		for _, poolAG := range pool.AffinityGroups {
+			if poolAG.Name != "" {
```
janoszen, line 108: Let's fail if the user doesn't enter an AG name.

janoszen (changes requested):

pkg/tfvars/ovirt/ovirt.go
```diff
@@ -3,6 +3,9 @@ package ovirt
 
 import (
 	"encoding/json"
+	"github.com/openshift/installer/pkg/types/ovirt"
+	ovirtdefaults "github.com/openshift/installer/pkg/types/ovirt/defaults"
```
staebler, line 7: Wrong location for these imports.

pkg/tfvars/ovirt/ovirt.go
staebler: Failing here is good, but I would hope that this would have already been caught by validation.

pkg/tfvars/ovirt/ovirt.go
```diff
@@ -71,6 +78,50 @@ func TFVars(
...
+				"priority": float64(ag.Priority),
+			})
+		}
+	}
+	return agMap
+}
+
+func createComputeAffinityGroups(computeMachines []*ovirt.MachinePool, masterAffiniryGroups []*ovirt.AffinityGroup, infraID string) []map[string]interface{} {
+	var agMap []map[string]interface{}
+	uniqeAGName := make(map[string]*ovirt.AffinityGroup)
```
staebler, line 104:
```suggestion
	uniqueAGName := make(map[string]*ovirt.AffinityGroup)
```

pkg/tfvars/ovirt/ovirt.go
```diff
@@ -71,6 +78,50 @@ func TFVars(
...
+
+	for _, pool := range computeMachines {
+		for _, poolAG := range pool.AffinityGroups {
+			if poolAG.Name != "" {
+				uniqeAGName[poolAG.Name] = poolAG
+			}
+		}
+	}
+	// If the affinity group exist on master list don't create it twice
```
staebler, line 113: What does the `vm_list` field of the `ovirt_affinity_group` terraform resource indicate? If an affinity group is included in both the masters and compute affinity groups in the install config, then the resulting affinity group will have a `vm_list` that only includes the masters.

pkg/types/ovirt/defaults/machinepools.go
```diff
@@ -0,0 +1,42 @@
+package defaults
+
+import (
+	"github.com/openshift/installer/pkg/types"
+	"github.com/openshift/installer/pkg/types/ovirt"
+)
+
+const (
+	//DAffinyGroupBase is a holder which will be replaced with the cluster name when creating the affinity group
+	DAffinyGroupBase                 = "CLUSTERNAME"
```
staebler, line 10: I strongly discourage using this as a means of coming up with an affinity group that includes the cluster name. Instead, I recommend adding the cluster name to the beginning or end of the name provided by the user. At a minimum, if you go the route of using a string to be replaced, then you need to document that around the type and add validation that enforces the existing of the string exactly once in the name.

pkg/types/ovirt/defaults/machinepools.go
```diff
@@ -0,0 +1,42 @@
...
+func setMachinePool(p *types.MachinePool) {
+	if p.Platform.Ovirt == nil {
+		p.Platform.Ovirt = &ovirt.MachinePool{}
+	}
+}
+
+// SetControlPlaneDefaults sets the defaults for the ControlPlane Machines.
+func SetControlPlaneDefaults(p *types.MachinePool) {
+	setMachinePool(p)
+	if p.Platform.Ovirt.AffinityGroups == nil {
```
staebler, line 26:
```suggestion
	if len(p.Platform.Ovirt.AffinityGroups) == 0 {
```

pkg/types/ovirt/defaults/machinepools.go
```diff
@@ -0,0 +1,42 @@
...
+			{Name: DAffinyGroupBase + dControlPlaneAffinyGroupPostfix, Priority: dControlPlaneAffinyGroupPriority},
+			{Name: DAffinyGroupBase + dComputeAffinyGroupPostfix, Priority: dComputeAffinyGroupPriority},
+		}
+	}
+}
+
+// SetComputeDefaults sets the defaults for the Compute Machines.
+func SetComputeDefaults(p *types.MachinePool) {
+	setMachinePool(p)
+	if p.Platform.Ovirt.AffinityGroups == nil {
```
staebler, line 37:
```suggestion
	if len(p.Platform.Ovirt.AffinityGroups) == 0 {
```

pkg/tfvars/ovirt/ovirt.go
staebler, line 113: How are we going to deal with the affinity group name used for both control plane and compute but using different priorities for each? Is that going to be caught by validation?

pkg/types/ovirt/machinepool.go
```diff
@@ -43,6 +47,12 @@ type CPU struct {
 	Cores int32 `json:"cores"`
 }
 
+// AffinityGroup defines the affinity group that the installer will create
+type AffinityGroup struct {
+	Name     string `json:"name"`
+	Priority int    `json:"priority,omitempty"`
```
staebler, line 53: You should not have an `int` type that is optional. If you want it to be optional, then the type should be `*int`. Are you sure that you even want it to be optional? I don't see code anywhere that sets the default priority.

pkg/types/ovirt/machinepool.go
```diff
@@ -90,4 +100,8 @@ func (p *MachinePool) Set(required *MachinePool) {
 	if required.OSDisk != nil {
 		p.OSDisk = required.OSDisk
 	}
+
+	if required.AffinityGroups != nil {
```
staebler, line 104:
```suggestion
	if len(required.AffinityGroups) > 0 {
```

data/data/ovirt/affinity_group/main.tf
```diff
@@ -0,0 +1,18 @@
...
+  count        = length(var.ovirt_control_plane_affinity_groups)
+  name         = var.ovirt_control_plane_affinity_groups[count.index]["name"]
+  cluster_id   = var.ovirt_cluster_id
+  priority     = var.ovirt_control_plane_affinity_groups[count.index]["priority"]
+  vm_positive  = false
+  vm_enforcing = false
+  vm_list      = var.masters_ids
+}
+
+resource "ovirt_affinity_group" "cluster_affinity_group" {
```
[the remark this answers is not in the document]
Gal-Zaidman: Initially this affinity group should've contain all the machines in the cluster, but we decided to make some changes to the affinity group layout and now it will be just for the compute nodes so I will change it :)

data/data/ovirt/affinity_group/main.tf
Gal-Zaidman: Making this configurable and default to true

data/data/ovirt/affinity_group/main.tf
Gal-Zaidman: +1

data/data/ovirt/affinity_group/variables.tf
```diff
@@ -0,0 +1,22 @@
+variable "ovirt_cluster_id" {
+  type        = string
+  description = "The ID of Cluster"
+}
+
+variable "ovirt_control_plane_affinity_groups" {
+  type        = list(map(string))
+  description = "Control plane of the masters affinity groups that will be created."
```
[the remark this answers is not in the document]
Gal-Zaidman: Done thanks :)

data/data/ovirt/affinity_group/variables.tf
```diff
@@ -0,0 +1,22 @@
...
+variable "ovirt_compute_affinity_groups" {
+  type        = list(map(string))
+  description = "Compute affinity groups that will be created."
+  default     = []
+}
+
+
+variable "masters_ids" {
+  type        = list(string)
+  description = "Name of the Affinity Group that will be created."
```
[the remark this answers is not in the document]
Gal-Zaidman: Fix that to describe the resource thanks for catching that!

pkg/asset/cluster/tfvars.go
```diff
@@ -494,6 +494,10 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			return err
 		}
 
+		workersInstallConfig := make([]*ovirt.MachinePool, 0)
```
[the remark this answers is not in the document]
Gal-Zaidman: +1

pkg/asset/machines/ovirt/machines.go
```diff
@@ -73,6 +75,13 @@ func provider(platform *ovirt.Platform, pool *types.MachinePool, userDataSecret
 		MemoryMB:          pool.Platform.Ovirt.MemoryMB,
 		VMType:            string(pool.Platform.Ovirt.VMType),
 	}
+	ags := make([]string, 0)
```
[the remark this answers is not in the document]
Gal-Zaidman: +1

pkg/tfvars/ovirt/ovirt.go
Gal-Zaidman: Done

pkg/tfvars/ovirt/ovirt.go
Gal-Zaidman: done

pkg/types/ovirt/defaults/machinepools.go
janoszen: @Gal-Zaidman any thoughts on this?

pkg/tfvars/ovirt/ovirt.go
Gal-Zaidman: Currently, it is not, but I'll add that to the validations

pkg/tfvars/ovirt/ovirt.go
Gal-Zaidman: done

pkg/tfvars/ovirt/ovirt.go
Gal-Zaidman: will be handled by validations

pkg/tfvars/ovirt/ovirt.go
Gal-Zaidman: vm_list fields are VMs that will be added to the affinity group.
This field will only exists in the `control_plain_affinity_group` because the masters are created by terraform so machine API create logic will not be called for them.
In case a affinity group exist in the masters and workers then the masters will be added to it by terraform with the vm_list field and the workers will be added by the cluster API provider create logic

Gal-Zaidman: I will add this to the validations

pkg/types/ovirt/defaults/machinepools.go
Gal-Zaidman: Well currently we use the replace to add the cluster name at the beginning.
The reason I used the place holder is that if a user inputs a custom name then it will not be replaced and if the default configuration is applied then the string will be replaced.
Instead, I can add some if to check if the affinity group name is different then the default affinity group name...

pkg/types/ovirt/defaults/machinepools.go
Gal-Zaidman: I'm ok with having 0 size, in that case no affinity groups will be created, that is the customization that will disable the affinity groups

pkg/types/ovirt/defaults/machinepools.go
Gal-Zaidman: I'm ok with having 0 size, in that case no affinity groups will be created, that is the customization that will disable the affinity groups

pkg/tfvars/ovirt/ovirt.go
staebler: Makes sense. Thanks for the explanation.

pkg/types/ovirt/defaults/machinepools.go
staebler: Oh, so you are *not* expecting the user to include "CLUSTERNAME" somewhere in the affinity group name. That was not clear to me. How is the destroyer going to clean up affinity groups with user-supplied names then?

pkg/types/ovirt/defaults/machinepools.go
staebler: A roundtrip marshaling and unmarshaling of the install config is going to lose the difference between a missing `AffinityGroups` field and an `AffinityGroups` field set to an empty list.

Take for example an install-config.yaml with the following `controlPlane` section.
```
controlPlane:
  platform:
    ovirt:
      affinityGroups: []
```

Now, run `openshift-install create install-config`. The `controlPlane` section will become the following.
```
controlPlane:
  architecture: amd64
  hyperthreading: Enabled
  name: master
  platform:
    ovirt: {}
  replicas: 3
```

Now, run `openshift-install create install-config` again. The `controlPlane` section will become the following.
```
controlPlane:
  architecture: amd64
  hyperthreading: Enabled
  name: master
  platform:
    ovirt:
      affinityGroups:
      - name: CLUSTERNAME-controlplane
        priority: 5
      - name: CLUSTERNAME-cluster
        priority: 3
  replicas: 3
```

pkg/types/ovirt/defaults/machinepools.go
Gal-Zaidman: I think you are right on second thought, there is no harm in always adding the cluster name to a new affinity group that the installer creates, it will simplify the logic and improve the destroy logic :)
Change that part, thanks for the feedback!

pkg/types/ovirt/machinepool.go
Gal-Zaidman: You are right, if a user decided that he wants to use customizations for affinity groups he should probably add all the fields for the Affinity Group, which will force him to think about it

pkg/types/ovirt/defaults/machinepools.go
Gal-Zaidman: Ok I see I guess we should handle that case, and when a user enters an empty array we shouldn't change the yaml and keep the empty array.
Great catch

janoszen:
@Gal-Zaidman the blocking bug is now merged.

pkg/asset/installconfig/ovirt/validation.go
```diff
@@ -63,6 +66,129 @@ func validateMachinePool(con *ovirtsdk.Connection, child *field.Path, pool *ovir
...
+	}
+	for _, compute := range ic.Compute {
+		for _, agn := range compute.Platform.Ovirt.AffinityGroupsNames {
+			mAgReplicas[agn] = mAgReplicas[agn] + int(*compute.Replicas)
+		}
+	}
+
+	clusterName, err := GetClusterName(con, ic.Ovirt.ClusterID)
+	if err != nil {
+		allErrs = append(allErrs, field.InternalError(fldPath, err))
```
staebler, line 169: Do you want to return here? The rest of the validation is worthless if we don't have a cluster name.

pkg/asset/installconfig/ovirt/validation.go
```diff
@@ -63,6 +66,129 @@ func validateMachinePool(con *ovirtsdk.Connection, child *field.Path, pool *ovir
...
+	}
+	hosts, err := FindHostsInCluster(con, clusterName)
+	if err != nil {
+		allErrs = append(allErrs, field.InternalError(fldPath, err))
```
staebler, line 173: Return here, too?

pkg/asset/installconfig/ovirt/validation.go
```diff
@@ -63,6 +66,129 @@ func validateMachinePool(con *ovirtsdk.Connection, child *field.Path, pool *ovir
...
+	}
+	for _, ag := range ic.Ovirt.AffinityGroups {
+		if _, found := mAgReplicas[ag.Name]; found {
```
staebler, line 176: Is there validation somewhere to ensure that the affinity group selected for a machine pool either is included in the affinity groups created by the installer or already exists in the oVirt cluster?

pkg/types/ovirt/defaults/machinepools.go
```diff
@@ -0,0 +1,34 @@
...
+				mp.Platform.Ovirt.AffinityGroupsNames = []string{agName}
+			}
+		}
+	}
+}
+
+// SetControlPlaneDefaults sets the defaults for the ControlPlane Machines.
+func SetControlPlaneDefaults(p *ovirt.Platform, mp *types.MachinePool) {
+	setMachinePool(mp)
+	setDefaultAffinityGroups(p, mp, DefaultControlPlaneAffinityGroupName)
```
staebler, line 27: It is surprising behavior to me that the afinity group of a machine pool gets set if and only if there is an affinity group specified with an exact name. At a minimum, this needs to be well documented in the `affinityGroups` field description in the machine pool type.

pkg/asset/machines/ovirt/machines.go
```diff
@@ -9,10 +9,9 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
+	ovirtprovider "github.com/openshift/cluster-api-provider-ovirt/pkg/apis/ovirtprovider/v1beta1"
```
staebler, line 12: Move this import into the previous section, with other non-installer imports.

pkg/types/ovirt/defaults/machinepools.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func setMachinePool(p *types.MachinePool) {
+	if p.Platform.Ovirt == nil {
+		p.Platform.Ovirt = &ovirt.MachinePool{}
+	}
+}
+
+func setDefaultAffinityGroups(p *ovirt.Platform, mp *types.MachinePool, agName string) {
+	if mp.Platform.Ovirt.AffinityGroupsNames == nil {
```
staebler, line 15: This field is `omitempty`, so we need to treat nil and an empty slice identically.
```suggestion
	if len(mp.Platform.Ovirt.AffinityGroupsNames) == 0 {
```

pkg/tfvars/ovirt/ovirt.go
```diff
@@ -2,12 +2,14 @@
 package ovirt
 
 import (
-	"encoding/json"
+	"fmt"
 
+	"encoding/json"
```
staebler, line 7: Keep this import where it was.

pkg/asset/machines/ovirt/machines.go
```diff
@@ -73,6 +72,20 @@ func provider(platform *ovirt.Platform, pool *types.MachinePool, userDataSecret
...
+	}
+	ags := make([]string, 0, len(pool.Platform.Ovirt.AffinityGroupsNames))
+	for _, agName := range pool.Platform.Ovirt.AffinityGroupsNames {
+		if _, ok := uniqueNewAG[agName]; ok {
+			// add the cluster name only if the affinity group is created by the installer
+			ags = append(ags, clusterID+"-"+agName)
+		} else {
+			ags = append(ags, agName)
+		}
+	}
```
staebler, line 87:
```suggestion

	ags := make([]string, len(pool.Platform.Ovirt.AffinityGroupsNames))
	for i, agName := range pool.Platform.Ovirt.AffinityGroupsNames {
		if _, ok := uniqueNewAG[agName]; ok {
			// add the cluster name only if the affinity group is created by the installer
			ags[i] = clusterID+"-"+agName
		} else {
			ags[i] = agName
		}
	}
```

pkg/tfvars/ovirt/ovirt.go
```diff
@@ -73,6 +79,21 @@ func TFVars(
...
 		cfg.BaseImageLocalFilePath = imageFilePath
 	}
-
+	if len(affinityGroups) > 0 {
+		cfg.AffinityGroups = handleAffinityGroups(affinityGroups, infraID)
+	}
 	return json.MarshalIndent(cfg, "", "  ")
 }
+
+func handleAffinityGroups(ags []*ovirt.AffinityGroup, infraID string) []map[string]interface{} {
```
staebler, line 88: This is not a very descriptive name for this function.

pkg/tfvars/ovirt/ovirt.go
```diff
@@ -73,6 +79,21 @@ func TFVars(
...
+	var agMap []map[string]interface{}
+	for _, ag := range ags {
+		agMap = append(agMap, map[string]interface{}{
+			"name":        fmt.Sprintf("%s-%s", infraID, ag.Name),
+			"priority":    float64(ag.Priority),
+			"description": ag.Description,
+			"enforcing":   ag.Enforcing,
+		})
+	}
```
staebler, line 97:
```suggestion
	agMap := make([]map[string]interface{}, len(ags))
	for i, ag := range ags {
		agMap[i] = map[string]interface{}{
			"name":        fmt.Sprintf("%s-%s", infraID, ag.Name),
			"priority":    float64(ag.Priority),
			"description": ag.Description,
			"enforcing":   ag.Enforcing,
		}
	}
```

pkg/tfvars/ovirt/ovirt.go
```diff
@@ -73,6 +79,21 @@ func TFVars(
...
 	}
-
+	if len(affinityGroups) > 0 {
+		cfg.AffinityGroups = handleAffinityGroups(affinityGroups, infraID)
+	}
 	return json.MarshalIndent(cfg, "", "  ")
 }
+
+func handleAffinityGroups(ags []*ovirt.AffinityGroup, infraID string) []map[string]interface{} {
+	var agMap []map[string]interface{}
```
staebler, line 89: This is a confusing name for this variable, since it is not a map but is instead a slice.

pkg/tfvars/ovirt/ovirt.go
```diff
@@ -73,6 +79,21 @@ func TFVars(
...
+	for _, ag := range ags {
+		agMap = append(agMap, map[string]interface{}{
+			"name":        fmt.Sprintf("%s-%s", infraID, ag.Name),
+			"priority":    float64(ag.Priority),
```
staebler, line 93: Why does this need to be converted to a float? Is it necessary for the way that the value gets marshaled?

pkg/destroy/ovirt/destroyer.go
```diff
@@ -144,6 +148,25 @@ func (uninstaller *ClusterUninstaller) removeTemplate(con *ovirtsdk.Connection)
...
 
+func (uninstaller *ClusterUninstaller) removeAffinityGroups(con *ovirtsdk.Connection) error {
+	cID := uninstaller.Metadata.Ovirt.ClusterID
+	affinityGroupService := con.SystemService().ClustersService().ClusterService(cID).AffinityGroupsService()
+	res, err := affinityGroupService.List().Send()
+	if err != nil {
+		return err
+	}
+	for _, ag := range res.MustGroups().Slice() {
+		if strings.Contains(ag.MustName(), uninstaller.Metadata.InfraID) {
```
staebler, line 159: Don't we only want to remove the affinity group if it has the prefix of "{INFRA_ID}-"?
```suggestion
		if strings.HasPrefix(ag.MustName(), fmt.Sprintf("%s-", uninstaller.Metadata.InfraID)) {
```

pkg/destroy/ovirt/destroyer.go
```diff
@@ -144,6 +148,25 @@ func (uninstaller *ClusterUninstaller) removeTemplate(con *ovirtsdk.Connection)
...
+			uninstaller.Logger.Infof("Removing AffinityGroup %s", ag.MustName())
+			_, err := affinityGroupService.GroupService(ag.MustId()).Remove().Send()
+			if err != nil {
+				uninstaller.Logger.Errorf("failed to remove AffinityGroup: %s", err)
```
staebler, line 163: Don't we also want to ultimately return an error from `removeAffinityGroups` when we fail to remove an affinity group, so that we keep trying?

But I see that we are not retrying for other resources. I think that is a serious deficiency that needs to be addressed for the ovirt destroyer, in general.

pkg/types/ovirt/platform.go
```diff
@@ -33,4 +33,16 @@ type Platform struct {
...
+	// +optional
+	AffinityGroups []*AffinityGroup `json:"affinityGroups"`
+}
+
+// AffinityGroup defines the affinity group that the installer will create
+type AffinityGroup struct {
+	Name        string `json:"name"`
+	Priority    int    `json:"priority"`
+	Description string `json:"description"`
+	Enforcing   bool   `json:"enforcing"`
```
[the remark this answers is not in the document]
Gal-Zaidman: I think I prefer keeping it a bool since the ovirt API expects a bool, and if we will just an enum is will just be deferent then ovirt plus make the code a bit harder to read :/

pkg/types/ovirt/platform.go
```diff
@@ -33,4 +33,16 @@ type Platform struct {
...
 	DefaultMachinePlatform *MachinePool `json:"defaultMachinePlatform,omitempty"`
+
+	// AffinityGroupsNames contains the names of the RHV affinity groups that the machines will join
+	// +optional
+	AffinityGroups []*AffinityGroup `json:"affinityGroups"`
+}
+
+// AffinityGroup defines the affinity group that the installer will create
+type AffinityGroup struct {
+	Name        string `json:"name"`
```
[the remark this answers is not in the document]
Gal-Zaidman: Done

pkg/asset/installconfig/ovirt/validation.go
```diff
@@ -63,6 +66,129 @@ func validateMachinePool(con *ovirtsdk.Connection, child *field.Path, pool *ovir
...
+	allErrs := field.ErrorList{}
+	allErrs = append(allErrs, validateAffinityGroupFields(*ic.Ovirt, fldPath)...)
+	allErrs = append(allErrs, validateExistingAffinityGroup(con, *ic.Ovirt, fldPath)...)
+	allErrs = append(allErrs, validateAffinityGroupDuplicate(ic.Ovirt.AffinityGroups)...)
+	allErrs = append(allErrs, validateClusterResources(con, ic, fldPath)...)
+
+	return allErrs
+}
+
+func validateAffinityGroupFields(platform ovirt.Platform, fldPath *field.Path) field.ErrorList {
```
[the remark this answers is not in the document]
Gal-Zaidman: Added optional on description, and since enforcing is its zero value is false so no need for validation on it

pkg/asset/installconfig/ovirt/validation.go
```diff
@@ -63,6 +66,129 @@ func validateMachinePool(con *ovirtsdk.Connection, child *field.Path, pool *ovir
...
+	allErrs := field.ErrorList{}
+	for _, ag := range platform.AffinityGroups {
+		if ag.Name == "" {
+			allErrs = append(
+				allErrs,
+				field.Invalid(fldPath, ag,
```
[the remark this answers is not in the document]
Gal-Zaidman: Done

pkg/asset/installconfig/ovirt/validation.go
```diff
@@ -63,6 +66,129 @@ func validateMachinePool(con *ovirtsdk.Connection, child *field.Path, pool *ovir
...
+	allErrs = append(allErrs, validateAffinityGroupDuplicate(ic.Ovirt.AffinityGroups)...)
+	allErrs = append(allErrs, validateClusterResources(con, ic, fldPath)...)
+
+	return allErrs
+}
+
+func validateAffinityGroupFields(platform ovirt.Platform, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	for _, ag := range platform.AffinityGroups {
+		if ag.Name == "" {
```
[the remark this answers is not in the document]
Gal-Zaidman: Done

pkg/tfvars/ovirt/ovirt.go
Gal-Zaidman: Done

pkg/tfvars/ovirt/ovirt.go
Gal-Zaidman: Yes the oVirt API expects a float so the terraform provider expects a float

pkg/destroy/ovirt/destroyer.go
Gal-Zaidman: Well, it is a unique string so contains felt safe enough but sure.
Done

pkg/destroy/ovirt/destroyer.go
Gal-Zaidman: Opened a new bug for that https://bugzilla.redhat.com/show_bug.cgi?id=1947424 Thanks!

pkg/tfvars/ovirt/ovirt.go
staebler: Terraform does not distinguish between whole numbers and fractional values.

pkg/tfvars/ovirt/ovirt.go
Gal-Zaidman: Done

Gal-Zaidman:
@staebler sorry this took a while to be ready, we needed to make some changes to the terraform provider to do this effort right.
Can you please re-review this PR, there are no real changes from your last review beside the fixes you requested :)

pkg/types/ovirt/platform.go
```diff
@@ -33,4 +33,16 @@ type Platform struct {
 	// Default will set the image field to the latest RHCOS image.
 	// +optional
 	DefaultMachinePlatform *MachinePool `json:"defaultMachinePlatform,omitempty"`
+
+	// AffinityGroupsNames contains the names of the RHV affinity groups that the machines will join
```
staebler, line 37: This is not the correct name of the field.
Field descriptions should end in periods.
```
	// AffinityGroups is the list RHV affinity groups that the machines will join.
```

pkg/types/ovirt/defaults/platform.go
```diff
@@ -7,9 +7,39 @@ import (
 // DefaultNetworkName is the default network name to use in a cluster
 const DefaultNetworkName = "ovirtmgmt"
 
+// DefaultControlPlaneAffinityGroupName is the default affinity group name for the control plain VMs
```
staebler, line 10:
```suggestion
// DefaultControlPlaneAffinityGroupName is the default affinity group name for the control plane VMs
```

pkg/types/ovirt/defaults/platform.go
```diff
@@ -7,9 +7,39 @@ import (
...
+const DefaultControlPlaneAffinityGroupName = "controlplane"
+
+// DefaultComputeAffinityGroupName is the default affinity group name for the compute VMs
+const DefaultComputeAffinityGroupName = "compute"
+
+func defaultControlPlaneAffinityGroup() *ovirt.AffinityGroup {
+	return &ovirt.AffinityGroup{
+		Name:        DefaultControlPlaneAffinityGroupName,
+		Priority:    5,
+		Description: "AffinityGroup for spreading each control plane machines to a different host",
```
staebler, line 20:
```suggestion
		Description: "AffinityGroup for spreading each control plane machine to a different host",
```

pkg/types/ovirt/platform.go
```diff
@@ -33,4 +33,16 @@ type Platform struct {
...
+	// +optional
+	AffinityGroups []*AffinityGroup `json:"affinityGroups"`
```
staebler, line 39: This does not need to be a slice of pointers. The extra indirection adds unnecessary complexity.
```
	AffinityGroups []AffinityGroup `json:"affinityGroups"`
```

pkg/types/ovirt/platform.go
```diff
@@ -41,8 +41,14 @@ type Platform struct {
...
 type AffinityGroup struct {
-	Name        string `json:"name"`
-	Priority    int    `json:"priority"`
+	// Name name of the affinity group
+	Name string `json:"name"`
+	// Priority of the affinity group, needs to be between 1 (lowest) - 5 (highest)
+	Priority int `json:"priority"`
+	// Description of the affinity group
+	// +optional
 	Description string `json:"description"`
```
staebler, line 50:
```suggestion
	Description string `json:"description,omitempty"`
```

pkg/asset/installconfig/ovirt/validation.go
```diff
@@ -2,9 +2,9 @@ package ovirt
 
 import (
 	"fmt"
-
```
staebler, line 5: This blank line still needs to be put back.

staebler (commented):
What is the plan for the commits? Are you planning to squash them down? If not, then the commit messages need some work. At a minimum they should make it clear that the commit is specific to oVirt. The commit messages should each stand on their own and not require the context of the entire PR to make it clear what is being done in the commit.
See https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#commit-message-format

pkg/types/ovirt/platform.go
Gal-Zaidman: Done

pkg/types/ovirt/defaults/platform.go
Gal-Zaidman: Done

pkg/asset/installconfig/ovirt/validation.go
Gal-Zaidman: Done

pkg/types/ovirt/platform.go
Gal-Zaidman: Done

pkg/types/ovirt/defaults/platform.go
Gal-Zaidman: Done

pkg/types/ovirt/platform.go
Gal-Zaidman: Done

Gal-Zaidman:
> What is the plan for the commits? Are you planning to squash them down? If not, then the commit messages need some work. At a minimum they should make it clear that the commit is specific to oVirt. The commit messages should each stand on their own and not require the context of the entire PR to make it clear what is being done in the commit.
> See https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#commit-message-format

I squashed some commits together and added description for each commit

pkg/asset/installconfig/ovirt/validation.go
```diff
@@ -63,6 +66,129 @@ func validateMachinePool(con *ovirtsdk.Connection, child *field.Path, pool *ovir
...
+	res, err := con.SystemService().ClustersService().
+		ClusterService(platform.ClusterID).AffinityGroupsService().List().Send()
+	if err != nil {
+		allErrs = append(
+			allErrs,
+			field.InternalError(
+				fldPath,
+				errors.Errorf("failed listing affinity groups for cluster %v", platform.ClusterID)))
+	}
+	for _, ag := range res.MustGroups().Slice() {
```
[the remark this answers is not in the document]
Gal-Zaidman: Yes it would, thanks for the catch

pkg/asset/installconfig/ovirt/validation.go
Gal-Zaidman: Yes you are correct, fixed

pkg/asset/machines/ovirt/machines.go
Gal-Zaidman: Done

pkg/asset/machines/ovirt/machines.go
Gal-Zaidman: Done

pkg/tfvars/ovirt/ovirt.go
Gal-Zaidman: Done

pkg/types/ovirt/defaults/machinepools.go
Gal-Zaidman: Done

pkg/types/ovirt/defaults/machinepools.go
Gal-Zaidman: Well I'm not sure I understood you, but this is only for the defaults.
The user can specify anything he wants basically

pkg/asset/installconfig/ovirt/validation.go
Gal-Zaidman: Added one

data/data/ovirt/affinity_group/main.tf
```diff
@@ -0,0 +1,9 @@
+resource "ovirt_affinity_group" "affinity_groups" {
```
janoszen, line 1: Shouldn't the `vm_list` be put in the `ignore_changes` lifecycle attribute here? After all, the PR that removes this attribute is not merged and there are some valid discussions around this.

data/data/ovirt/affinity_group/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "ovirt_affinity_group" {
```
janoszen, line 1: Nitpick: this could be named `ovirt_affinity_group_count`

pkg/asset/installconfig/ovirt/client.go
```diff
@@ -69,3 +69,30 @@ func FetchVNICProfileByClusterNetwork(con *ovirtsdk.Connection, clusterID string
 	}
 	return nil, fmt.Errorf("there are no vNic profiles for the given cluster ID %s and network name %s", clusterID, networkName)
 }
+
+// GetClusterName returns the name of the ovirt cluster with the specified Cluster ID
+func GetClusterName(con *ovirtsdk.Connection, clusterID string) (string, error) {
+	clusterResponse, err := con.SystemService().ClustersService().ClusterService(clusterID).Get().Send()
+	if err != nil {
+		return "", err
```
janoszen, line 77: Please add context to the error here, e.g. `fmt.Errorf("failed to fetch cluster %s (%w)", clusterID, err)`

pkg/asset/installconfig/ovirt/client.go
```diff
@@ -69,3 +69,30 @@ func FetchVNICProfileByClusterNetwork(con *ovirtsdk.Connection, clusterID string
...
+	}
+	return cluster.MustName(), nil
+}
+
+// FindHostsInCluster returns a list of the hosts that exists in the specified cluster
+func FindHostsInCluster(con *ovirtsdk.Connection, cName string) ([]*ovirtsdk.Host, error) {
+	res, err := con.SystemService().HostsService().
+		List().Search(fmt.Sprintf("cluster=%s", cName)).Send()
+	if err != nil {
+		return nil, err
```
janoszen, line 91: Please add context to the error here, like above.

pkg/types/ovirt/machinepool.go
```diff
@@ -24,6 +24,11 @@ type MachinePool struct {
 	// +kubebuilder:validation:Enum="";desktop;server;high_performance
 	// +optional
 	VMType VMType `json:"vmType,omitempty"`
+
+	// AffinityGroupsNames contains the name of the OpenShift cluster affinity groups
```
janoszen, line 28: Suggestion:

> AffinityGroupNames contains a list of oVirt affinity group names that will be used to separate the OpenShift cluster nodes from each other to separate physical machines. If the affinity group is not created in the install config it is assumed that the AG already exists in oVirt.

pkg/types/ovirt/platform.go
```diff
@@ -33,4 +33,22 @@ type Platform struct {
 	// Default will set the image field to the latest RHCOS image.
 	// +optional
 	DefaultMachinePlatform *MachinePool `json:"defaultMachinePlatform,omitempty"`
+
+	// AffinityGroups contains the RHV affinity groups that the installer will create.
```
janoszen, line 37: Suggestion:

> AffinityGroups contains a list of affinity groups that will be created by the installer. If an affinity group is not listed here, but specified in AffinityGroupNames it is assumed that the affinity group already exists in oVirt.

janoszen (changes requested):

data/data/ovirt/affinity_group/main.tf
Gal-Zaidman: Well, we are not passing it so it is never changed.
The problem in the past was the dependency that we had between the bootstrap and masters that caused the affinity groups to get reevaluated on destroy but we removed that thanks to moving the lock to the affinity group function in the VM resource, so no problem now.

data/data/ovirt/affinity_group/outputs.tf
Gal-Zaidman: Done

pkg/asset/installconfig/ovirt/client.go
Gal-Zaidman: Done

pkg/asset/installconfig/ovirt/client.go
Gal-Zaidman: Done

pkg/types/ovirt/machinepool.go
Gal-Zaidman: Changed it to:
```
AffinityGroupsNames contains a list of oVirt affinity group names that the newly created machines will join.
The affinity groups should exist on the oVirt cluster or created by the OpenShift installer.
```
I don't want to say for what the affinity group will be used for :)

pkg/types/ovirt/platform.go
Gal-Zaidman: I don't want to add a reference to the machine field here, I think that saying it on the machine field is more than enough

data/data/ovirt/affinity_group/main.tf
Gal-Zaidman: Done

janoszen (approved):

janoszen (approved):

PR #4678, opened 2021-02-22
Bug 1868755: vsphereprivate: tf plugin to no longer error if no network found.

jstuever:
Prior to this change, the vsphereprivate Terraform plugin would fatally
error if the specified network was not found on any hosts when importing
the ova. Because the error was a common pointer reference error, it was
difficult to determine what caused the error. This change adds a more
clear error when this situation happens.


staebler (commented):
/lgtm
/approve

jstuever:
/hold cancel
e2e-vsphere passed.

PR #4679, opened 2021-02-23
OpenStack: Stop using deprecated interpolation syntax

mandre:
This syntax was deprecated in terraform 0.11. Let's use the new syntax
to prevent a deprecation warning:

    Warning: Interpolation-only expressions are deprecated


Fedosin (approved):
/approve
/lgtm

PR #4681, opened 2021-02-23
cmd/openshift-install: fix dropped error

alrs:
This fixes a dropped error variable in `cmd/openshift-install`.

### DIFF
--- a/cmd/openshift-install/gather.go
+++ b/cmd/openshift-install/gather.go
@@ -249,6 +249,9 @@ func extractHostAddresses(config *types.InstallConfig, tfstate *terraform.State)
 			return bootstrap, port, masters, err
 		}
 		masters, err = gatherkubevirt.ControlPlaneIPs(tfstate)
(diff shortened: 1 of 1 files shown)


cmd/openshift-install/gather.go
```diff
@@ -249,6 +249,9 @@ func extractHostAddresses(config *types.InstallConfig, tfstate *terraform.State)
 			return bootstrap, port, masters, err
 		}
 		masters, err = gatherkubevirt.ControlPlaneIPs(tfstate)
+		if err != nil {
+			return bootstrap, port, masters, err
```
staebler, line 253: This should not be returning an error. The control planes may not be available, but we still want to gather from the bootstrap. We can log this error like it is done for other platforms, but we should not return the error.

alrs: I've changed it.

staebler (commented):
/lgtm
/approve

PR #4682, opened 2021-02-23
Enable re-use of existing private hosted zone in AWS

mihail-milev:
This modification allows by modifying the cluster-dns-02-config.yml manifest
file to ask the installer to re-use an already existing private AWS
hosted zone in Route53.

In the data/data/aws/route53/base.tf file a new "data" section was
added and based on the value of a TF variable either the "data" section
or the "ressource" section is being used. If the TF variable is set
to Golang's string zero value ("") then a new hosted zone will be
created (default behavior), otherwise, if not string zero, then TF will
try to use the hosted zone, which ID corresponds to the one specified
in the TF variable.

In order to activate the mechanism:
1) specify "publish: Internal" inside the install-config.yaml
2) generate the manifests
3) modify manifests/cluster-dns-02-config.yml by:
3.1) removing the "tags" section from "privateZone"
3.2) adding "id" string value inside the "privateZone". Set the value
     of this string to the hosted zone ID ("Z0123QQQZXX..."), you
     wish to reuse
4) execute "openshift-install create cluster ..."

This is exactly what is described in this file:
docs/user/aws/install_upi.md
under the chapter "Identify the internal DNS zone"

This mechanism could be extended for other cloud providers, e.g.
Azure or GCP, but I don't have the resources to do it.

### DIFF
--- a/data/data/aws/main.tf
+++ b/data/data/aws/main.tf
@@ -89,6 +89,7 @@ module "dns" {
   vpc_id                   = module.vpc.vpc_id
   region                   = var.aws_region
   publish_strategy         = var.aws_publish_strategy
+  zone_id                  = var.aws_private_zone_id
 }
 
 module "vpc" {
@@ -128,4 +129,3 @@ resource "aws_ami_copy" "imported" {
     local.tags,
   )
 }
-
--- a/data/data/aws/route53/base.tf
+++ b/data/data/aws/route53/base.tf
@@ -5,6 +5,8 @@ locals {
   // So publish_strategy serves an coordinated proxy for that decision.
   public_endpoints = var.publish_strategy == "External" ? true : false
 
+  use_existing_private_zone = var.zone_id == "" ? false : true
+
   use_cname = contains(["us-gov-west-1", "us-gov-east-1", "us-iso-east-1"], var.region)
   use_alias = ! local.use_cname
 }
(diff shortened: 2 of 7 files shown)


staebler:
Thank you for your PR. Unfortunately, this is not a change that we will be accepting.

We deliberately do not provide all possible configurations in the installer-provisioned installation flow. The intention is for the options available to be curated to a small set that is useful for a large share of our users or difficult to work around in other ways.

mihail-milev:
> Thank you for your PR. Unfortunately, this is not a change that we will be accepting.
> 
> We deliberately do not provide all possible configurations in the installer-provisioned installation flow. The intention is for the options available to be curated to a small set that is useful for a large share of our users or difficult to work around in other ways.

Hello,

I do understand the urge to keep it as simple as possible and to make it easy for the users to deploy a cluster. That is the reason, why I also kept the default behavior of the installer - just execute it like you are doing it until now, and it will work as it worked until now.

But there are some "power users" over there, which deploy > 10 clusters a day and there is the need to do customization sometimes. Most of the time there are restrictions (like e.g. not allowed to create new hosted zones) and every time "workarounds" need to be done. That is not oriented towards the users - that is against the users.

What would be the "other ways" to solve this problem?

staebler:
> But there are some "power users" over there, which deploy > 10 clusters a day and there is the need to do customization sometimes. Most of the time there are restrictions (like e.g. not allowed to create new hosted zones) and every time "workarounds" need to be done. That is not oriented towards the users - that is against the users.
> 
> What would be the "other ways" to solve this problem?

The "other ways" to solve this problem is to take the user-provisioned infrastructure approach. It is precisely for those "power users" with special requirements that we offer that approach, where the user has more control over what is and is not provisioned. You could start with and modify to your needs the CloudFormation from https://github.com/openshift/installer/tree/master/upi/aws/cloudformation or even the terraform that the installer uses to do an installer-provisioned infrastructure installation. Or you can use whatever other tools you prefer to provision the infrastructure.

staebler:
Bring-your-own private hosted zone is being implemented as a configurable field in the install config with https://github.com/openshift/installer/pull/4772.
/close

PR #4683, opened 2021-02-23
baremetal: Allow IPA kernel console to be configured

derekhiggins:
Add a new baremetal platform config option "provisioningConsole"
this can be used to pass a console to ironic to be used by the
kernel running IPA. A value of "ttyS0" would result in IPA
running on a virtual CI environment which can then be logged.

Makes use of the recently added IRONIC_KERNEL_PARAMS  parameter
openshift/ironic-image#151


derekhiggins:
/label platform/baremetal

Only covers IPA running on master nodes, would need to pass this param to the cluster-baremetal-operator to add support for the worker nodes.

PR #4687, opened 2021-02-24
Bug 1894774: openstack: Document setting affinity for workers

pierreprinetti:


docs/user/openstack/affinity.md
```diff
@@ -0,0 +1,136 @@
...
+
+In OpenStack Compute (Nova), scheduler policies are set using Server groups.
+
+The first step is then to create the Server group.
+
+There are four possible policies:
+* affinity
+* soft-affinity
+* anti-affinity
+* soft-anti-affinity
```
EmilienM, line 25: I think it depends on how nova is configured, see https://docs.openstack.org/nova/latest/admin/configuration/schedulers.html and the `enabled_filters` flag in nova.conf.

I checked in my OSP environment and I have these filters enabled by default:
`enabled_filters=AvailabilityZoneFilter,ComputeFilter,ComputeCapabilitiesFilter,ImagePropertiesFilter,ServerGroupAntiAffinityFilter,ServerGroupAffinityFilter`
I know that operators can change that during their deployments of OpenStack and it makes me wonder if it can have an impact on what policies can be used here when creating a server group.
If that would be the case, maybe we can just add a note here saying that it's up to the user to check if whether or not the filter is enabled in Nova.

cc @mdbooth

EmilienM: Nevermind, actually only the ServerGroupAffinityFilter needs to be enabled (default), but still an operator could deploy Nova and disable ServerGroupAffinityFilter (for some reason), and therefore it wouldn't be possible to create server groups with these policies, AFIU.

pierreprinetti: Changed the wording to reflect that not all policies might be available in all clusters.

pierreprinetti:
Thank you @EmilienM, I have addressed your comment. Can you please take a look again?

docs/user/openstack/affinity.md
```diff
@@ -0,0 +1,136 @@
...
+
+The first step is then to create the Server group.
+
+Depending on the OpenStack configuration, there are a maximum of four possible policies:
+* affinity
+* soft-affinity
+* anti-affinity
+* soft-anti-affinity
+
+These options are best described in the [OpenStack API reference][openstack-compute-api-docs].
```
mdbooth, line 27: Is this a placeholder, or is there some docs wizardry to turn this into a real link?

Edit: it's sourcery. See bottom of document.

docs/user/openstack/affinity.md
```diff
@@ -0,0 +1,136 @@
...
+
+The "create manifests" command creates several files, among which the worker
+MachineSet definition in
+`openshift/99_openshift-cluster-api_worker-machineset-0.yaml`.
+
+We must add a new property to the MachineSet: `serverGroupID`. That property lives under `spec.template.spec.providerSpec.value`.
+
+**Use the Server group id from step 1.**
+
+I have used the example UUID `aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee` in the template below.
```
mdbooth, line 63: Stylistic note: you've also used <infrastructure_ID>, <node_role>, <nova_flavor>, etc. You should probably do with with <server_group_uuid>, and perhaps call it out here because it's important.

docs/user/openstack/affinity.md
```diff
@@ -0,0 +1,136 @@
...
+          availabilityZone: <optional_openstack_availability_zone>
+```
+
+### Step 4: Complete the installation
+
+After saving the MachineSet file in place, use `openshift-install` to complete
+the installation.
+
+Note: running the following command has the side effect of deleting both
+`openshift` and `manifests` directories from the current working directory.
```
mdbooth, line 126: Aside: just the fact that we need to point this out to reduce user surprise is a massive smell 🙄

docs/user/openstack/affinity.md
```diff
@@ -0,0 +1,136 @@
...
+
+```shell
+openshift-install create cluster
+```
+
+The modified `MachineSet` will be used by the Installer to create the Compute
+nodes; the `serverGroupID` property make OpenShift create the Compute nodes
+within that OpenStack Server Group.
+
+[openstack-compute-api-docs]: https://docs.openstack.org/api-ref/compute/?expanded=create-server-group-detail#server-groups-os-server-groups
```
mdbooth, line 136: 🤯

mdbooth (approved):
/lgtm

Stylistic note if you want to address it.

docs/user/openstack/affinity.md
mandre: @pierreprinetti perhaps you already know about this but your modifications are not deleted, they're saved under another form in the `.openshift_install_state.json` file. You can re-generate the `openshift` and `manifests` directories re-running the `openshift-install create manifests` command if you like.

Same thing with your `install-config.yaml`, you can retrieve it re-running the `openshift-install create install-config` command.

pierreprinetti: WAT

PR #4688, opened 2021-02-24
Bug 1921169: enforce credentials modes in GCP terraform [WIP]

patrickdillon:
Work in progress. I need to test and flesh out the documentation/description.

### DIFF
--- a/data/data/config.tf
+++ b/data/data/config.tf
@@ -123,3 +123,12 @@ Should the cluster be created with ipv6 networking.
 EOF
 
 }
+
+variable "credentials_mode" {
+  type = string
+  default = null
+
+  description = <<EOF
+Is the cluster in Mint, Passthrough, or Manual credential mode.
+EOF
+}
--- a/data/data/gcp/iam/main.tf
+++ b/data/data/gcp/iam/main.tf
@@ -1,14 +1,58 @@
+locals {
+  # Mint Mode is default so assume Mint unless explicitly set otherwise.
+  mint_mode = var.credentials_mode != "Passthrough" && var.credentials_mode != "Manual"
+}
+
 resource "google_service_account" "worker-node-sa" {
-  account_id   = "${var.cluster_id}-w"
+  count        = local.mint_mode ? 1 : 0
+  account_id   = split("@", var.worker_sa_email)[0]
   display_name = "${var.cluster_id}-worker-node"
 }
 
 resource "google_project_iam_member" "worker-compute-viewer" {
+  count  = local.mint_mode ? 1 : 0
   role   = "roles/compute.viewer"
-  member = "serviceAccount:${google_service_account.worker-node-sa.email}"
+  member = "serviceAccount:${google_service_account.worker-node-sa[0].email}"
 }
 
 resource "google_project_iam_member" "worker-storage-admin" {
+  count  = local.mint_mode ? 1 : 0
   role   = "roles/storage.admin"
-  member = "serviceAccount:${google_service_account.worker-node-sa.email}"
+  member = "serviceAccount:${google_service_account.worker-node-sa[0].email}"
+}
+
+resource "google_service_account" "master-node-sa" {
+  count        = local.mint_mode ? 1 : 0
+  account_id   = split("@", var.master_sa_email)[0]
+  display_name = "${var.cluster_id}-master-node"
+}
+
+resource "google_project_iam_member" "master-compute-admin" {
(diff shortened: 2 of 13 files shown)


data/data/config.tf
```diff
@@ -123,3 +123,12 @@ Should the cluster be created with ipv6 networking.
 EOF
 
 }
+
+variable "credentials_mode" {
+  type = string
+  default = null
```
patrickdillon, line 129: Another idea I considered is having three boolean values to represent mint, manual, passthrough.

pkg/asset/machines/gcp/machines.go
```diff
@@ -148,3 +154,42 @@ func getNetworks(platform *gcp.Platform, clusterID, role string) (string, string
...
+		Scopes: []string{"https://www.googleapis.com/auth/cloud-platform"},
+	}
+	switch cm {
+	case types.PassthroughCredentialsMode:
+		var err error
+		if sa.Email, err = getServiceAccountEmail(); err != nil {
+			return nil, errors.Wrap(err, "failed to get service account email for passthrough mode")
+		}
+	case types.ManualCredentialsMode:
+		// sa.Email is empty
```
patrickdillon, line 173: We will need to have a separate doc fix to demonstrate how to edit the manifests for manual mode. As this stands, I think install will fail without user intervention. Perhaps I could add something like a platform provision check to make sure the sa.Email value is present.

data/data/gcp/iam/main.tf
```diff
@@ -1,14 +1,58 @@
+locals {
+  # Mint Mode is default so assume Mint unless explicitly set otherwise.
+  mint_mode = var.credentials_mode != "Passthrough" && var.credentials_mode != "Manual"
+}
+
 resource "google_service_account" "worker-node-sa" {
-  account_id   = "${var.cluster_id}-w"
+  count        = local.mint_mode ? 1 : 0
```
staebler, line 7: Would it be cleaner to put a count on the module in main.tf rather than on every resource in the module?

data/data/gcp/iam/main.tf
```diff
@@ -1,14 +1,58 @@
...
   role   = "roles/compute.viewer"
-  member = "serviceAccount:${google_service_account.worker-node-sa.email}"
+  member = "serviceAccount:${google_service_account.worker-node-sa[0].email}"
 }
 
 resource "google_project_iam_member" "worker-storage-admin" {
+  count  = local.mint_mode ? 1 : 0
   role   = "roles/storage.admin"
-  member = "serviceAccount:${google_service_account.worker-node-sa.email}"
+  member = "serviceAccount:${google_service_account.worker-node-sa[0].email}"
```
staebler, line 21: Can we use locals for the master and worker member values?

data/data/gcp/iam/outputs.tf
```diff
@@ -0,0 +1,5 @@
+# This simply passes the variable value through this module,
+# but enforces a dependency in the master module.
```
staebler, line 2: It may be worth waiting until we upgrade our terraform version and then just using module dependencies directly.

pkg/asset/machines/gcp/machines.go
staebler: I don't think that we can ask the user to make manual modifications to manifests in order to use a supported install config. If we are going to go the route of requiring the user to supply the service account, then we need to provide that configuration in the install config.

However, both approaches break compatibility with existing install configs, which is not acceptable.

staebler (commented):
I was not paying deep enough attention to this before to understand exactly what the plan was here. Unfortunately, I don't think that we can take this approach. I think that credentials mode is not exactly related to instance service accounts. If we do want to stop creating service accounts based on credentials mode, then we need to do the same for all of our cloud platforms to keep the behavior and expectations the same.

patrickdillon:
> I was not paying deep enough attention to this before to understand exactly what the plan was here. Unfortunately, I don't think that we can take this approach. I think that credentials mode is not exactly related to instance service accounts. If we do want to stop creating service accounts based on credentials mode, then we need to do the same for all of our cloud platforms to keep the behavior and expectations the same.

After looking at this more closely, I don't really think there is a bug to solve here. I had a misunderstanding that service accounts were credentials, but it is really the _service account key_ that seems to be the credential. My concern was that organizations would not want to hand out credentials with the ability to create service accounts, but apparently only creating service account keys would be the concern. So I will close this and the underlying bz.

/close

PR #4689, opened 2021-02-24
WIP openstack: make security groups optional

### LINKED ISSUE #4555: make openstack security rules optional
IBM's PowerVC (more info: https://www.ibm.com/products/powervc) is built on top of Openstack, and exposes Openstack APIs. We're hoping to make some tweaks to the installer for openstack that would allow us to use this against PowerVC. In experimenting with this, there are a few PowerVC limitations we've come across. I've been asked to open this issue for the security rules.

I would like to make them optional, as the only way to use them with PowerVC is to set up Novalink (more info: https://www.ibm.com/support/knowledgecenter/en/SSXK2N_1.4.0/com.ibm.powervc.standard.help.doc/powervc_sdn.html), which not all customers will have configured. 

I've done some initial hacking (mostly gutting), and can run the installer without creating any security rules. One concern was with DHCP server conflicts, but since network isolation (I believe), seems to be accomplished using tagged VLANs on the SEA, we may not need to worry about that.

Any other thoughts or concerns are welcome.

/cc @EricArrakis @iamemilio

clnperez:
Some openstack-based offerings, such as PowerVC, do not support security
 groups. Allow users to create a cluster without them. The default is to create
 the security groups (disable = false).
    
 Fixes #4555
  
### DIFF
--- a/data/data/openstack/main.tf
+++ b/data/data/openstack/main.tf
@@ -40,7 +40,10 @@ module "bootstrap" {
   cluster_domain          = var.cluster_domain
   nodes_subnet_id         = module.topology.nodes_subnet_id
   private_network_id      = module.topology.private_network_id
-  master_sg_id            = var.openstack_disable_sg ? null : module.topology.master_sg_id
+  master_sg_ids           = var.openstack_disable_sg ? [] : concat(
+    var.openstack_master_extra_sg_ids,
+    [module.topology.master_sg_id],
+  )
   bootstrap_shim_ignition = var.openstack_bootstrap_shim_ignition
   master_port_ids         = module.topology.master_port_ids
   root_volume_size        = var.openstack_master_root_volume_size
--- a/data/data/openstack/main.tf
+++ b/data/data/openstack/main.tf
@@ -56,13 +56,12 @@ module "masters" {
   source = "./masters"
 
   base_image_id   = data.openstack_images_image_v2.base_image.id
-  openstack_disable_sg      = var.openstack_disable_sg
   cluster_id      = var.cluster_id
   flavor_name     = var.openstack_master_flavor_name
   instance_count  = var.master_count
   master_port_ids = module.topology.master_port_ids
   user_data_ign   = var.ignition_master
-  master_sg_ids   = var.openstack_disable_sg ? null : concat(
+  master_sg_ids   = module.topology.master_sg_id == null ? [] : concat(
     var.openstack_master_extra_sg_ids,
     [module.topology.master_sg_id],
   )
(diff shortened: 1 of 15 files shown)


data/data/openstack/main.tf
```diff
@@ -70,22 +72,23 @@ module "masters" {
...
+  api_floating_ip       = var.openstack_api_floating_ip
+  ingress_floating_ip   = var.openstack_ingress_floating_ip
+  api_int_ip            = var.openstack_api_int_ip
+  ingress_ip            = var.openstack_ingress_ip
+  external_dns          = var.openstack_external_dns
+  trunk_support         = var.openstack_trunk_support
+  octavia_support       = var.openstack_octavia_support
+  machines_subnet_id    = var.openstack_machines_subnet_id
+  machines_network_id   = var.openstack_machines_network_id
+  master_extra_sg_ids   = var.openstack_disable_sg ? null : var.openstack_master_extra_sg_ids
```
staebler, line 91: Do we need to enforce that null is used for the extra security group IDs? With `openstack_disable_sg` being set to true, `master_extra_sg_ids` will not be used in the module anyway.

data/data/openstack/masters/variables.tf
```diff
@@ -8,6 +8,10 @@ variable "cluster_id" {
   description = "The identifier for the cluster."
 }
 
+variable "openstack_disable_sg" {
```
staebler, line 11: This variable is unused.

pkg/types/openstack/platform.go
```diff
@@ -84,4 +84,8 @@ type Platform struct {
 	// The subnet and network specified in MachinesSubnet will not be deleted or modified by the installer.
 	// +optional
 	MachinesSubnet string `json:"machinesSubnet,omitempty"`
+
+	// DisableSecurityGroups will allow clusters to be created without the use of OpenStack Security Groups. Some Openstack
+	// configurations do not support security groups. This is generally not recommended where security groups are supported.
+	DisableSecurityGroups bool `json:"disableSecurityGroups,omitempty"`
```
staebler, line 90: Is there a way for the installer to determine whether the OpenStack configuration supports security groups without requiring the user to specify this?

clnperez: I wondered about that as well, and am not sure. If someone knows, this would be a good add. Maybe just in a follow-up patch.

clnperez:
Thanks @staebler for the review. I pushed changes addressing your comments to a new commit (will squash these later). The openstack-related jobs here did all pass on the previous run, so, that is encouraging.

pkg/types/openstack/platform.go
staebler: Note that what I am suggesting is that if there is a way for the installer to determine this, then we should not even have it as an option for the user. In other words, we would eliminate this field.

data/data/openstack/masters/main.tf
```diff
@@ -63,7 +63,7 @@ resource "openstack_compute_instance_v2" "master_conf_0" {
 
   flavor_id = data.openstack_compute_flavor_v2.masters_flavor.id
   image_id = var.root_volume_size == null ? var.base_image_id : null
-  security_groups = var.master_sg_ids
+  security_groups = var.master_sg_ids == null ? [] : var.master_sg_ids
```
staebler, line 66: This should not be needed. (1) The resource should be able to handle null. (2) The variable will never be null: It is an empty list when security groups are disabled.

clnperez:
Closing for now, as priorities have swizzled

PR #4693, opened 2021-02-25
gather-bootstrap: Gather logs before bootstrap shuts down

rna-afk:
The bootkube tries to keep the bootstrap control-plane on for
30 minutes before it shuts down due to a timeout. In which case,
the bootkube tries to re-run the control-plane to gather the logs.
The re-run is not guaranteed by the api team and hence there is
a possibility the logs for the control-plane will not be gathered.

Adding a check to trap the error message from the run bootstrap
command in the bootkube to run installer-gather script to gather
the logs in case of any error before the resources get deleted.
The logs will be stored in a folder with a default name temporarily.
Then the size of the openapi.json.gz file will be checked to see
if there was any extra information was logged from the control-
plane and if there was not a change, the tar file and the folder
will be removed as the bootstrap control-plane did not come up.
If it has been changed, the folder is moved from temporary to
permanent folder (removing the temp part of the name).

Also adding a mechanism to pull the created logs with the default
folder name to get the latest information about the control-plane.

### DIFF
--- a/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
+++ b/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
@@ -4,6 +4,7 @@ set -euoE pipefail ## -E option will cause functions to inherit trap
 . /usr/local/bin/bootstrap-service-record.sh
 
 . /usr/local/bin/release-image.sh
+. /usr/local/bin/bootstrap-cluster-gather.sh
 
 mkdir --parents /etc/kubernetes/{manifests,bootstrap-configs,bootstrap-manifests}
 
@@ -377,19 +378,28 @@ then
 fi
 
 echo "Starting cluster-bootstrap..."
-
-if [ ! -f cb-bootstrap.done ]
-then
-    record_service_stage_start "cb-bootstrap"
-    bootkube_podman_run \
+run_cluster_bootstrap() {
+	record_service_stage_start "cb-bootstrap"
+	bootkube_podman_run \
         --rm \
         --volume "$PWD:/assets:z" \
         --volume /etc/kubernetes:/etc/kubernetes:z \
         "${CLUSTER_BOOTSTRAP_IMAGE}" \
         start --tear-down-early=false --asset-dir=/assets --required-pods="${REQUIRED_PODS}"
-
-    touch cb-bootstrap.done
-    record_service_stage_success
+}
+    
+if [ ! -f cb-bootstrap.done ]
+then
+    if run_cluster_bootstrap
+    then
+        touch cb-bootstrap.done
+        record_service_stage_success
+    else
+        ret=$?
+        set +u
+        cluster_bootstrap_gather
+        exit $ret
+    fi
 fi
 
 if [ "$BOOTSTRAP_INPLACE" = true ]
--- a/data/data/bootstrap/files/usr/local/bin/bootstrap-cluster-gather.sh
+++ b/data/data/bootstrap/files/usr/local/bin/bootstrap-cluster-gather.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+
+API_SERVER_ARTIFACTS_DIR="/tmp/artifacts-api-server-temp"
+function queue() {
+    local TARGET="${ARTIFACTS_TEMP}/${1}"
+    shift
+    # shellcheck disable=SC2155
+    local LIVE="$(jobs | wc -l)"
+    while [[ "${LIVE}" -ge 45 ]]; do
+        sleep 1
+        LIVE="$(jobs | wc -l)"
+    done
+    if [[ -n "${FILTER-}" ]]; then
+        # shellcheck disable=SC2024
+        sudo KUBECONFIG="${GATHER_KUBECONFIG}" "${@}" | "${FILTER}" >"${TARGET}" &
+    else
+        # shellcheck disable=SC2024
+        sudo KUBECONFIG="${GATHER_KUBECONFIG}" "${@}" >"${TARGET}" &
+    fi
+}
+
+function cluster_bootstrap_gather() {
+    GATHER_KUBECONFIG="/opt/openshift/auth/kubeconfig"
+
+    ALTERNATIVE_KUBECONFIG="/etc/kubernetes/bootstrap-secrets/kubeconfig"
+    if [[ -f ${ALTERNATIVE_KUBECONFIG} ]]; then
+        GATHER_KUBECONFIG=${ALTERNATIVE_KUBECONFIG}
+    fi
+
+    echo "Using ${GATHER_KUBECONFIG} as KUBECONFIG"
+
+    ARTIFACTS_TEMP="$(mktemp -d)"
+
+    mkdir -p "${ARTIFACTS_TEMP}/resources"
+
+    echo "Gathering cluster resources ..."
+    queue resources/nodes.list oc --request-timeout=5s get nodes -o jsonpath --template '{range .items[*]}{.metadata.name}{"\n"}{end}'
+    queue resources/masters.list oc --request-timeout=5s get nodes -o jsonpath -l 'node-role.kubernetes.io/master' --template '{range .items[*]}{.metadata.name}{"\n"}{end}'
+    # ShellCheck doesn't realize that $ns is for the Go template, not something we're trying to expand in the shell
+    # shellcheck disable=2016
+    queue resources/containers oc --request-timeout=5s get pods --all-namespaces --template '{{ range .items }}{{ $name := .metadata.name }}{{ $ns := .metadata.namespace }}{{ range .spec.containers }}-n {{ $ns }} {{ $name }} -c {{ .name }}{{ "\n" }}{{ end }}{{ range .spec.initContainers }}-n {{ $ns }} {{ $name }} -c {{ .name }}{{ "\n" }}{{ end }}{{ end }}'
+    queue resources/api-pods oc --request-timeout=5s get pods -l apiserver=true --all-namespaces --template '{{ range .items }}-n {{ .metadata.namespace }} {{ .metadata.name }}{{ "\n" }}{{ end }}'
+
+    queue resources/apiservices.json oc --request-timeout=5s get apiservices -o json
+    queue resources/clusteroperators.json oc --request-timeout=5s get clusteroperators -o json
+    queue resources/clusterversion.json oc --request-timeout=5s get clusterversion -o json
+    queue resources/configmaps.json oc --request-timeout=5s get configmaps --all-namespaces -o json
+    queue resources/csr.json oc --request-timeout=5s get csr -o json
+    queue resources/endpoints.json oc --request-timeout=5s get endpoints --all-namespaces -o json
+    queue resources/events.json oc --request-timeout=5s get events --all-namespaces -o json
+    queue resources/kubeapiserver.json oc --request-timeout=5s get kubeapiserver -o json
+    queue resources/kubecontrollermanager.json oc --request-timeout=5s get kubecontrollermanager -o json
+    queue resources/machineconfigpools.json oc --request-timeout=5s get machineconfigpools -o json
+    queue resources/machineconfigs.json oc --request-timeout=5s get machineconfigs -o json
+    queue resources/namespaces.json oc --request-timeout=5s get namespaces -o json
+    queue resources/nodes.json oc --request-timeout=5s get nodes -o json
+    queue resources/openshiftapiserver.json oc --request-timeout=5s get openshiftapiserver -o json
+    queue resources/pods.json oc --request-timeout=5s get pods --all-namespaces -o json
+    queue resources/rolebindings.json oc --request-timeout=5s get rolebindings --all-namespaces -o json
+    queue resources/roles.json oc --request-timeout=5s get roles --all-namespaces -o json
+    # this just lists names and number of keys
+    queue resources/secrets-names.txt oc --request-timeout=5s get secrets --all-namespaces
+    # this adds annotations, but strips out the SA tokens and dockercfg secrets which are noisy and may contain secrets in the annotations
+    queue resources/secrets-names-with-annotations.txt oc --request-timeout=5s get secrets --all-namespaces -o=custom-columns=NAMESPACE:.metadata.namespace,NAME:.metadata.name,TYPE:.type,ANNOTATIONS:.metadata.annotations | grep -v -- '-token-' | grep -v -- '-dockercfg-'
+    queue resources/services.json oc --request-timeout=5s get services --all-namespaces -o json
+
+    FILTER=gzip queue resources/openapi.json.gz oc --request-timeout=5s get --raw /openapi/v2
+
+    echo "Waiting for logs ..."
+    wait
+
+    if (( $(stat -c%s "${ARTIFACTS_TEMP}/resources/openapi.json.gz") <= 20 ))
+    then
+        rm -rf "${ARTIFACTS_TEMP}"
+    else
+        rm -rf "${API_SERVER_ARTIFACTS_DIR}"
+        mkdir -p "${API_SERVER_ARTIFACTS_DIR}"
+        mv "${ARTIFACTS_TEMP}/resources" "${API_SERVER_ARTIFACTS_DIR}"
+    fi
+}
--- a/data/data/bootstrap/files/usr/local/bin/installer-gather.sh
+++ b/data/data/bootstrap/files/usr/local/bin/installer-gather.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# shellcheck disable=SC1091
+. /usr/local/bin/bootstrap-cluster-gather.sh
+
 if test "x${1}" = 'x--id'
 then
 	GATHER_ID="${2}"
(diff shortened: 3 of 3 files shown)


cmd/openshift-install/gather.go
```diff
@@ -162,6 +162,10 @@ func logGatherBootstrap(bootstrap string, port int, masters []string, directory
 	if err := ssh.PullFileTo(client, fmt.Sprintf("/home/core/log-bundle-%s.tar.gz", gatherID), file); err != nil {
 		return errors.Wrap(err, "failed to pull log file from remote")
 	}
+	apiServerFiles := filepath.Join(directory, fmt.Sprintf("log-bundle-api-server-%s.tar.gz", gatherID))
```
staebler, line 165: Let's not pull two separate bundles. That is going to be confusing to the user. Let's find a way to include the most recent results from the api server in the other log bundle.

1. When the cluster-bootstrap pods exits with an error code, gather the cluster resources. As the last part of gathering resources, check if the openapi.json.gz file is of sufficient size. If it is, then store the cluster resources in a known location.
2. When installer-gather.sh is executed, do the same gathering of cluster resources, including moving to a known location if appropriate. Then, copy the cluster resources from the known location to the bundle location, so that they will be included in the bundle.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -368,15 +368,32 @@ fi
...
         --volume "$PWD:/assets:z" \
         --volume /etc/kubernetes:/etc/kubernetes:z \
         "${CLUSTER_BOOTSTRAP_IMAGE}" \
         start --tear-down-early=false --asset-dir=/assets --required-pods="${REQUIRED_PODS}"
+}
+
+gather_bootstrap_on_failure() {
+	/usr/local/bin/installer-gather.sh --id 'api-server-temp'
+	gunzip -k /tmp/artifacts-api-server-temp/resources/openapi.json.gz
+	if [ ! -s '/tmp/artifacts-api-server-temp/resources/openapi.json' ]
```
staebler, line 383: Let's avoid doing a full unzip here.
```suggestion
	if (( $(stat -c%s '/tmp/artifacts-api-server-temp/resources/openapi.json.gz' <= 20 ))
```

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -368,15 +368,32 @@ fi
...
+		rm $HOME/log-bundle-api-server-temp.tar.gz
+	else
+		mv $HOME/log-bundle-api-server-temp.tar.gz $HOME/log-bundle-api-server.tar.gz
+	fi
+}
 
+
+if [ ! -f cb-bootstrap.done ]
+then
+	trap 'gather_bootstrap_on_failure' ERR
```
staebler, line 395: Setting the trap means that gather_bootstrap_on_failure will run when there are errors in subsequent steps of the script. That is not terrible, but it is not quite what we want. Can we capture the result of the podman run instead of trapping?

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -368,15 +368,32 @@ fi
...
+	bootkube_podman_run \
         --rm \
         --volume "$PWD:/assets:z" \
         --volume /etc/kubernetes:/etc/kubernetes:z \
         "${CLUSTER_BOOTSTRAP_IMAGE}" \
         start --tear-down-early=false --asset-dir=/assets --required-pods="${REQUIRED_PODS}"
+}
+
+gather_bootstrap_on_failure() {
+	/usr/local/bin/installer-gather.sh --id 'api-server-temp'
```
staebler, line 381: Don't do the entire gathering. Only do the cluster resource gathering.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
rna-afk: I did try podman logs/history (which I feel might be hard to parse) and tried to reset traps (didn't work).  Any other ideas?

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -368,15 +370,30 @@ fi
...
+}
 
+
+if [ ! -f cb-bootstrap.done ]
+then
+	if ! run_cluster_bootstrap
+	then
+		gather_bootstrap_on_failure
+	fi
     touch cb-bootstrap.done
```
staebler, line 397: Make sure that we exit when `run_cluster_bootstrap` fails.
```suggestion
	if run_cluster_bootstrap
	then
		touch cb-bootstrap.done
	else
		exit $?
	fi
```

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -368,15 +370,30 @@ fi
...
+	bootkube_podman_run \
         --rm \
         --volume "$PWD:/assets:z" \
         --volume /etc/kubernetes:/etc/kubernetes:z \
         "${CLUSTER_BOOTSTRAP_IMAGE}" \
         start --tear-down-early=false --asset-dir=/assets --required-pods="${REQUIRED_PODS}"
+}
+
+gather_bootstrap_on_failure() {
+	/usr/local/bin/bootstrap-cluster-gather.sh
```
staebler, line 383: Source a script with the function that you want to call and call that function rather than calling another script directly here.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -368,16 +368,34 @@ fi
...
         start --tear-down-early=false --asset-dir=/assets --required-pods="${REQUIRED_PODS}"
+}
 
-    touch cb-bootstrap.done
+gather_bootstrap_on_failure() {
+	source ./bootstrap-cluster-gather.sh
+	cluster_bootstrap_gather
+	if (( $(stat -c%s '/tmp/artifacts-api-server-temp/resources/openapi.json.gz') <= 20 ))
+	then
+		rm -rf /tmp/artifacts-api-server-temp
```
staebler, line 385: This is not the correct behavior. Imagine that the first time through bootkube.sh, the api server artifacts were collected successfully. The second time through bootkube.sh, the control plane did not start, and the api server artifacts were not collected successfully. On the second time through, the code will delete the artifacts gathered, leaving us with nothing. There is also the problem of having the second time through write into the same directory where the artifacts were already collected, polluting the capture with existing files.

1. Gather in a temp location.
2. If the gather was successful, move to a permanent location.
3. In install-gather.sh, after doing a gather and maybe move again, move the artifacts from the permanent location to the gathered bundle.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -3,6 +3,8 @@ set -euoE pipefail ## -E option will cause functions to inherit trap
 
 . /usr/local/bin/release-image.sh
 
+source ./bootstrap-cluster-gather.sh
```
staebler, line 6: For consistency with other sources in scripts used for bootstrapping,
```suggestion
. /usr/local/bin/bootstrap-cluster-gather.sh
```

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -368,16 +370,36 @@ fi
...
+}
 
-    touch cb-bootstrap.done
+gather_bootstrap_on_failure() {
+	cluster_bootstrap_gather
+	if (( $(stat -c%s '/tmp/artifacts-api-server-temp/resources/openapi.json.gz') <= 20 ))
+	then
+		rm -rf "/tmp/artifacts-api-server-temp"
+	else
+		mv "/tmp/artifacts-api-server-temp/resources/*" "${HOME}/artifacts-api-server-temp/resources"
```
staebler, line 388:
```suggestion
		rm -rf "${HOME}/artifacts-api-server-temp/resources"
		mv "/tmp/artifacts-api-server-temp/resources" "${HOME}/artifacts-api-server-temp"
```

data/data/bootstrap/files/usr/local/bin/installer-gather.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+source ./bootstrap-cluster-gather.sh
```
staebler, line 3:
```suggestion
. /usr/local/bin/bootstrap-cluster-gather.sh
```

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -368,16 +370,36 @@ fi
...
         --volume "$PWD:/assets:z" \
         --volume /etc/kubernetes:/etc/kubernetes:z \
         "${CLUSTER_BOOTSTRAP_IMAGE}" \
         start --tear-down-early=false --asset-dir=/assets --required-pods="${REQUIRED_PODS}"
+}
 
-    touch cb-bootstrap.done
+gather_bootstrap_on_failure() {
+	cluster_bootstrap_gather
+	if (( $(stat -c%s '/tmp/artifacts-api-server-temp/resources/openapi.json.gz') <= 20 ))
```
staebler, line 384: Move this logic into `cluster_bootstrap_gather`. The same logic should happen regardless of whether `cluster_bootstrap_gather` is run after a failed cluster bootstrap or as part of responding to a bootstrap gather request.

data/data/bootstrap/files/usr/local/bin/installer-gather.sh
```diff
@@ -166,6 +168,12 @@ do
     scp -o PreferredAuthentications=publickey -o StrictHostKeyChecking=false -o UserKnownHostsFile=/dev/null -r -q "core@[${master}]:/tmp/artifacts-${GATHER_ID}/*" "${ARTIFACTS}/control-plane/${master}/"
 done
 
+cluster_bootstrap_gather
+if [ -d "${HOME}/artifacts-api-server-temp" ]
+then
+    mv "${HOME}/artifacts-api-server-temp/resources/*" "${ARTIFACTS}/resources"
+fi
```
staebler, line 175: For this to work, the code to move from the api server artifacts from tmp to HOME has to be moved to `cluster_bootstrap_gather` (which we should do).

Do not do a wildcard move. Move the entire directory instead.
```suggestion
if [ -d "${HOME}/artifacts-api-server-temp" ]
then
    mv "${HOME}/artifacts-api-server-temp/resources" "${ARTIFACTS}"
fi
```

staebler (commented):
This is getting really close.

data/data/bootstrap/files/usr/local/bin/bootstrap-cluster-gather.sh
```diff
@@ -0,0 +1,77 @@
...
+    echo "Waiting for logs ..."
+    wait
+
+    if (( $(stat -c%s '/tmp/artifacts-api-server-temp/resources/openapi.json.gz') <= 20 ))
+	then
+		rm -rf "/tmp/artifacts-api-server-temp"
+	else
+		rm -rf "${HOME}/artifacts-api-server-temp/resources"
+		mv "/tmp/artifacts-api-server-temp/resources" "${HOME}/artifacts-api-server-temp"
+	fi
```
staebler, line 76:
```suggestion
    if (( $(stat -c%s '${ARTIFACTS}/resources/openapi.json.gz') <= 20 ))
	then
		rm -rf "${ARTIFACTS}"
	else
		rm -rf "${HOME}/artifacts-api-server-temp/resources"
		mv "${ARTIFACTS}/resources" "${HOME}/artifacts-api-server-temp"
	fi
```

data/data/bootstrap/files/usr/local/bin/bootstrap-cluster-gather.sh
```diff
@@ -0,0 +1,77 @@
...
+
+    echo "Waiting for logs ..."
+    wait
+
+    if (( $(stat -c%s '/tmp/artifacts-api-server-temp/resources/openapi.json.gz') <= 20 ))
+	then
+		rm -rf "/tmp/artifacts-api-server-temp"
+	else
+		rm -rf "${HOME}/artifacts-api-server-temp/resources"
+		mv "/tmp/artifacts-api-server-temp/resources" "${HOME}/artifacts-api-server-temp"
```
staebler, line 75: We need a `mkdir` here, too, right?
```suggestion
		mkdir -p "${HOME}/artifacts-api-server-temp"
		mv "/tmp/artifacts-api-server-temp/resources" "${HOME}/artifacts-api-server-temp"
```

data/data/bootstrap/files/usr/local/bin/installer-gather.sh
```diff
@@ -166,6 +168,12 @@ do
     scp -o PreferredAuthentications=publickey -o StrictHostKeyChecking=false -o UserKnownHostsFile=/dev/null -r -q "core@[${master}]:/tmp/artifacts-${GATHER_ID}/*" "${ARTIFACTS}/control-plane/${master}/"
 done
 
+cluster_bootstrap_gather
+if [ -d "${HOME}/artifacts-api-server-temp" ]
```
staebler, line 172: This should actually check for the resources directory, otherwise we could get a failure on the `mv`.
```suggestion
if [ -d "${HOME}/artifacts-api-server-temp/resources" ]
```

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -368,16 +370,30 @@ fi
...
+run_cluster_bootstrap() {
+	bootkube_podman_run \
         --rm \
         --volume "$PWD:/assets:z" \
         --volume /etc/kubernetes:/etc/kubernetes:z \
         "${CLUSTER_BOOTSTRAP_IMAGE}" \
         start --tear-down-early=false --asset-dir=/assets --required-pods="${REQUIRED_PODS}"
+}
+
+gather_bootstrap_on_failure() {
```
staebler, line 382: Is this function useful any more? Or can we call `cluster_bootstrap_gather` directly instead?

rna-afk: Yeah you're right. Maybe I need to slow down the changes and think. Sorry :-)

data/data/bootstrap/files/usr/local/bin/bootstrap-cluster-gather.sh
```diff
@@ -0,0 +1,77 @@
...
+    echo "Waiting for logs ..."
+    wait
+
+    if (( $(stat -c%s "${ARTIFACTS}/resources/openapi.json.gz") <= 20 ))
+	then
+		rm -rf "${ARTIFACTS}"
+	else
+        mkdir -p "${HOME}/artifacts-api-server-temp"
+		mv "${ARTIFACTS}/resources" "${HOME}/artifacts-api-server-temp"
+	fi
```
staebler, line 76: The rest of this file is using spaces instead of tabs, so let's stay consistent.
```suggestion
    if (( $(stat -c%s "${ARTIFACTS}/resources/openapi.json.gz") <= 20 ))
    then
        rm -rf "${ARTIFACTS}"
    else
        mkdir -p "${HOME}/artifacts-api-server-temp"
        mv "${ARTIFACTS}/resources" "${HOME}/artifacts-api-server-temp"
    fi
```

data/data/bootstrap/files/usr/local/bin/installer-gather.sh
```diff
@@ -166,6 +169,12 @@ do
     scp -o PreferredAuthentications=publickey -o StrictHostKeyChecking=false -o UserKnownHostsFile=/dev/null -r -q "core@[${master}]:/tmp/artifacts-${GATHER_ID}/*" "${ARTIFACTS}/control-plane/${master}/"
 done
 
+cluster_bootstrap_gather
```
staebler, line 172: Just noticed that the gathering of cluster resources has not been removed from earlier in this file. This call to `cluster_bootstrap_gather` should be moved up to replace the current resource gathering.

data/data/bootstrap/files/usr/local/bin/installer-gather.sh
```diff
@@ -85,44 +88,12 @@ function queue() {
...
-# this adds annotations, but strips out the SA tokens and dockercfg secrets which are noisy and may contain secrets in the annotations
-queue resources/secrets-names-with-annotations.txt oc --request-timeout=5s get secrets --all-namespaces -o=custom-columns=NAMESPACE:.metadata.namespace,NAME:.metadata.name,TYPE:.type,ANNOTATIONS:.metadata.annotations | grep -v -- '-token-' | grep -v -- '-dockercfg-'
-queue resources/services.json oc --request-timeout=5s get services --all-namespaces -o json
-
-FILTER=gzip queue resources/openapi.json.gz oc --request-timeout=5s get --raw /openapi/v2
-
-echo "Waiting for logs ..."
-wait
-
+mkdir -p "${ARTIFACTS}/control-plane"
```
staebler, line 91: Move this down to where the directory is actually used.

staebler (commented):
Just one last nit.

staebler (commented):
This looks good.

@rna-afk Can you run through an install that will fail waiting to cluster-bootstrap and show me that the cluster resources are still getting collected correctly? You should be able to add a malformed manifest during installation to provoke the failure.

rna-afk:
After ironing out a few small errors that showed up during cluster installation, the gather happens as soon as the cluster bootstrap command fails and gets collected in the /home/core directory and is being used as a fallback in case installer-gather fails to query the resources.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -377,19 +378,28 @@ then
...
-    touch cb-bootstrap.done
-    record_service_stage_success
+}
+    
+if [ ! -f cb-bootstrap.done ]
+then
+    if run_cluster_bootstrap
+    then
+        touch cb-bootstrap.done
+		record_service_stage_success
```
staebler, line 396:
```suggestion
        record_service_stage_success
```

data/data/bootstrap/files/usr/local/bin/bootstrap-cluster-gather.sh
```diff
@@ -0,0 +1,83 @@
...
+function queue() {
+    local TARGET="${ARTIFACTS_TEMP}/${1}"
+    shift
+    # shellcheck disable=SC2155
+    local LIVE="$(jobs | wc -l)"
+    while [[ "${LIVE}" -ge 45 ]]; do
+        sleep 1
+        LIVE="$(jobs | wc -l)"
+    done
+    # echo "${@}"
```
staebler, line 12:
```suggestion
```

data/data/bootstrap/files/usr/local/bin/bootstrap-cluster-gather.sh
```diff
@@ -0,0 +1,83 @@
...
+    GATHER_KUBECONFIG="/opt/openshift/auth/kubeconfig"
+
+    ALTERNATIVE_KUBECONFIG="/etc/kubernetes/bootstrap-secrets/kubeconfig"
+    if [[ -f ${ALTERNATIVE_KUBECONFIG} ]]; then
+        GATHER_KUBECONFIG=${ALTERNATIVE_KUBECONFIG}
+    fi
+
+    echo "Using ${GATHER_KUBECONFIG} as KUBECONFIG"
+
+    ARTIFACTS_TEMP="/tmp/artifacts-api-server-temp-$(date +%s%N)"
```
staebler, line 32: Let's use a real temp directory instead of relying on timestamp.
```suggestion
    ARTIFACTS_TEMP="$(mktemp -d)"
```

data/data/bootstrap/files/usr/local/bin/bootstrap-cluster-gather.sh
```diff
@@ -0,0 +1,83 @@
...
+    wait
+
+    if (( $(stat -c%s "${ARTIFACTS_TEMP}/resources/openapi.json.gz") <= 20 ))
+    then
+        rm -rf "${ARTIFACTS_TEMP}"
+    else
+        if [ -d "/home/core/artifacts-api-server-temp" ]
+        then
+            rm -rf "/home/core/artifacts-api-server-temp"
+        fi
```
staebler, line 79: Did you find that you needed to check for the existence of the directory first? Since we are using the `-f` flag of `rm`, I would expect it to be OK to call `rm` even if the directory does not exist.
```suggestion
        rm -rf "/home/core/artifacts-api-server-temp"
```

data/data/bootstrap/files/usr/local/bin/bootstrap-cluster-gather.sh
```diff
@@ -0,0 +1,83 @@
...
+        mkdir -p "/home/core/artifacts-api-server-temp"
+        mv "${ARTIFACTS_TEMP}/resources" "/home/core/artifacts-api-server-temp"
```
staebler, line 81: Use `${HOME}` instead of `/home/core`.

rna-afk: bootkube runs as root and ${HOME} is / for some odd reason so I had to do this. It creates the folder there with the values but user core has no permissions for that folder.

data/data/bootstrap/files/usr/local/bin/bootstrap-cluster-gather.sh
rna-afk: Yeah this was overkill. I'll remove it.

data/data/bootstrap/files/usr/local/bin/bootstrap-cluster-gather.sh
```diff
@@ -0,0 +1,79 @@
...
+    echo "Waiting for logs ..."
+    wait
+
+    if (( $(stat -c%s "${ARTIFACTS_TEMP}/resources/openapi.json.gz") <= 20 ))
+    then
+        rm -rf "${ARTIFACTS_TEMP}"
+    else
+        rm -rf "/home/core/artifacts-api-server-temp"
+        mkdir -p "/home/core/artifacts-api-server-temp"
+        mv "${ARTIFACTS_TEMP}/resources" "/home/core/artifacts-api-server-temp"
```
staebler, line 77: In case this got missed from the last review,
```suggestion
        rm -rf "${HOME}/artifacts-api-server-temp"
        mkdir -p "${HOME}/artifacts-api-server-temp"
        mv "${ARTIFACTS_TEMP}/resources" "${HOME}/artifacts-api-server-temp"
```

staebler: Sorry. I did not notice that you commented on this in the last review. I will discuss it there.

data/data/bootstrap/files/usr/local/bin/bootstrap-cluster-gather.sh
staebler: Let's use `/tmp` instead then.

data/data/bootstrap/files/usr/local/bin/bootstrap-cluster-gather.sh
```diff
@@ -0,0 +1,79 @@
...
+    FILTER=gzip queue resources/openapi.json.gz oc --request-timeout=5s get --raw /openapi/v2
+
+    echo "Waiting for logs ..."
+    wait
+
+    if (( $(stat -c%s "${ARTIFACTS_TEMP}/resources/openapi.json.gz") <= 20 ))
+    then
+        rm -rf "${ARTIFACTS_TEMP}"
+    else
+        rm -rf "/tmp/artifacts-api-server-temp"
```
staebler, line 75: Can you create a variable for `/tmp/artifacts-api-server-temp` so that we can keep it defined in one place?

staebler (commented):
Sorry, one more change.

data/data/bootstrap/files/usr/local/bin/bootstrap-cluster-gather.sh
```diff
@@ -0,0 +1,80 @@
...
+
+    ALTERNATIVE_KUBECONFIG="/etc/kubernetes/bootstrap-secrets/kubeconfig"
+    if [[ -f ${ALTERNATIVE_KUBECONFIG} ]]; then
+        GATHER_KUBECONFIG=${ALTERNATIVE_KUBECONFIG}
+    fi
+
+    echo "Using ${GATHER_KUBECONFIG} as KUBECONFIG"
+
+    ARTIFACTS_TEMP="$(mktemp -d)"
+    ARTIFACTS_DIR="/tmp/artifacts-api-server-temp"
```
staebler, line 32: Move this to the top of this file, out of the function. Also, rename this to something like `API_SERVER_ARTIFACTS_DIR`.

staebler (approved):
/lgtm

staebler:
These changes have no effect on upgrades.

PR #4694, opened 2021-02-25
Bug 1932789: Fix proxy validation when specified with port.

rna-afk:
When the port number is specified, the validation that whether
the given proxy value overlaps with the cluster or service
networks is not done due to the port number not being a valid
value for the parseIP function used for converting string to
a valid IP.

Adding a fix to make sure that the port number part of the code
is removed from the given input proxy value before it is converted
to a vaild IP.


staebler (commented):
/lgtm
/approve

PR #4695, opened 2021-02-25
Bug 1903055: Set default values to machine pools before validation

Fedosin:
Now we validate just the user input, but in fact we additionally set default values there if they are not provided directly. It leads
to incorrect validation of the install config.

This commit sets the default values to machine pools before validation to make them just like what we use to build a cluster.

/label platform/openstack

### DIFF
--- a/pkg/asset/installconfig/openstack/validate.go
+++ b/pkg/asset/installconfig/openstack/validate.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/openshift/installer/pkg/asset/installconfig/openstack/validation"
 	"github.com/openshift/installer/pkg/types"
+	"github.com/openshift/installer/pkg/types/openstack"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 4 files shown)


habibahma (commented):
Readme

pkg/asset/installconfig/openstack/validate.go
```diff
@@ -47,3 +53,9 @@ func ValidateForProvisioning(ic *types.InstallConfig) error {
 	}
 	return nil
 }
+
+func defaultOpenStackMachinePoolPlatform() openstack.MachinePool {
+	return openstack.MachinePool{
+		Zones: []string{""},
```
mandre, line 59: This is really an awful default. Please remind me why we do this?

Fedosin: @mandre Because of this https://github.com/openshift/installer/blob/master/pkg/asset/machines/worker.go#L112-L116

mandre: That doesn't explain why the default Zones is a slice with an empty element. I'd expect it to be an empty slice but surely I'm missing something.

mandre: This default is bothering me but changing it isn't in scope for this PR.

According to @iamemilio, its a slice with an empty string because passing an empty string to zones is how you set it to use Nova default.

mandre (commented):
/approve

PR #4696, opened 2021-02-25
gcp: Remove support for Licenses, only use public image

cgwalters:
gcp: Ignore Licenses: vmx, only use public image

The history here is basically that originally for GCP, we
cloned a disk image, just like how Azure works today.

Then later, GCP cracked down on the allowable frequency of image
creation and told us to use public bootable images, so we started
uploading a public RHCOS base image.

But we originally did that without the nested virt license, so
the KubeVirt team added a PR to clone the image and add the
nested virt license.

And then finally, in 4.6 we fixed the CoreOS public image
to add the nested virt license by default (which also, GCP
recommended we do):
https://github.com/coreos/coreos-assembler/pull/1618/commits/73320cf633e362f96de5f9d9aabb02b8647d4156

So there's no need for any of this image cloning or license
stuff anymore.  For compatibility (with e.g. the KubeVirt CI)
we silently ignore the `enable-vmx` licenses because it's
implicitly enabled by default anyways.  Copying the image
to enable it is a pessimization in fact!

Other licenses are a fatal error - I am not aware of
any that anyone would want.  And anyone who wants to
do something truly custom can use
the image override variable anyways; they don't *have*
to use our Terraform code.  Or they can hack the installer
code.

Prep for using the stream metadata in
https://github.com/openshift/installer/pull/4582

### DIFF
--- a/pkg/types/gcp/platform.go
+++ b/pkg/types/gcp/platform.go
@@ -30,11 +30,7 @@ type Platform struct {
 	// +optional
 	ComputeSubnet string `json:"computeSubnet,omitempty"`
 
-	// Licenses is a list of licenses to apply to the compute images
-	// The value should a list of strings (https URLs only) representing the license keys.
-	// When set, this will cause the installer to copy the image into user's project.
-	// This option is incompatible with any mechanism that makes use of pre-built images
-	// such as the current env OPENSHIFT_INSTALL_OS_IMAGE_OVERRIDE
+	// Licenses is no longer used and should be either empty or contain the nested virt license.
 	// +optional
 	Licenses []string `json:"licenses,omitempty"`
 }
--- a/data/data/gcp/main.tf
+++ b/data/data/gcp/main.tf
@@ -5,7 +5,7 @@ locals {
   worker_subnet_cidr = cidrsubnet(var.machine_v4_cidrs[0], 3, 1) #worker subnet is a smaller subnet within the vnet. i.e from /21 to /24
   public_endpoints   = var.gcp_publish_strategy == "External" ? true : false
 
-  gcp_image = var.gcp_preexisting_image ? var.gcp_image : google_compute_image.cluster[0].self_link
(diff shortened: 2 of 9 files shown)


cgwalters:
A simple way to look at this is GCP now always works how we do AWS, we just boot the AMI references, we don't clone them.

pkg/types/gcp/platform.go
```diff
@@ -30,11 +30,7 @@ type Platform struct {
 	// +optional
 	ComputeSubnet string `json:"computeSubnet,omitempty"`
 
-	// Licenses is a list of licenses to apply to the compute images
-	// The value should a list of strings (https URLs only) representing the license keys.
-	// When set, this will cause the installer to copy the image into user's project.
-	// This option is incompatible with any mechanism that makes use of pre-built images
-	// such as the current env OPENSHIFT_INSTALL_OS_IMAGE_OVERRIDE
+	// Licenses support has been removed and must be empty.
```
staebler, line 33: This is a breaking change that we cannot make. We can deprecate the field, but we need to continue to support it.

cgwalters: Compromise: we can detect the case where the only thing listed in the licenses is the nested virt, and silently ignore it because we do that by default.

cgwalters: Updated now with that compromise!  cc https://github.com/amorenoz from https://github.com/openshift/installer/pull/3430

staebler: If we had had that restriction to the nested virt license initially, this would be great. Unfortunately, we cannot add a restriction now.

cgwalters: I totally understand the PoV that this API was exposed and hence we need to support it ~forever.  If you feel strongly about it, feel free to close this PR.

But again I'm pretty sure no actual customers or CI systems are using anything except nested virt.

PR #4699, opened 2021-03-01
installconfig/aws: validate connectivity to service endpoints.

jstuever:
This change ensures any provided endpoints can be resolved via
DNS and connected to while parsing the install config.

https://issues.redhat.com/browse/CORS-1546

### DIFF
--- a/pkg/asset/installconfig/aws/validation.go
+++ b/pkg/asset/installconfig/aws/validation.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"net/url"
 	"sort"
 
 	"github.com/aws/aws-sdk-go/aws/endpoints"
@@ -62,9 +63,7 @@ func validatePlatform(ctx context.Context, meta *Metadata, fldPath *field.Path, 
 	if len(platform.Subnets) > 0 {
 		allErrs = append(allErrs, validateSubnets(ctx, meta, fldPath.Child("subnets"), platform.Subnets, networking, publish)...)
 	}
-	if err := validateServiceEndpoints(fldPath.Child("serviceEndpoints"), platform.Region, platform.ServiceEndpoints); err != nil {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("serviceEndpoints"), platform.ServiceEndpoints, err.Error()))
-	}
+	allErrs = append(allErrs, validateServiceEndpoints(fldPath.Child("serviceEndpoints"), platform.Region, platform.ServiceEndpoints)...)
 	if platform.DefaultMachinePlatform != nil {
 		allErrs = append(allErrs, validateMachinePool(ctx, meta, fldPath.Child("defaultMachinePlatform"), platform, platform.DefaultMachinePlatform, controlPlaneReq)...)
 	}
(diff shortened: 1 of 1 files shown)


pkg/asset/installconfig/aws/validation.go
```diff
@@ -211,12 +212,39 @@ func validateDuplicateSubnetZones(fldPath *field.Path, subnets map[string]Subnet
...
+		URL, err := url.Parse(service.URL)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		port := URL.Port()
+		// If the port was not specified, use the Scheme or default to http.
+		if port == "" {
+			if URL.Scheme != "" {
+				port = URL.Scheme
```
staebler, line 231: This is neat that you can use the scheme as the port in Dial.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -211,12 +212,39 @@ func validateDuplicateSubnetZones(fldPath *field.Path, subnets map[string]Subnet
 }
 
 func validateServiceEndpoints(fldPath *field.Path, region string, services []awstypes.ServiceEndpoint) error {
+	var errs []error
+	// For each provided service endpoint, verify we can resolve and connect with net.Dial.
+	for _, service := range services {
+		// Ignore e2e.local from unit tests.
+		if service.URL == "e2e.local" {
```
staebler, line 219: Can we run a server in the unit tests that can be used as a fake service endpoint? That would have a couple benefits.
1) We won't have to have extra code in the production code to deal with special test cases.
2) We can add test cases that verify that the service endpoint connectivity validation is working correctly.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -211,12 +212,39 @@ func validateDuplicateSubnetZones(fldPath *field.Path, subnets map[string]Subnet
...
+			continue
+		}
+		URL, err := url.Parse(service.URL)
+		if err != nil {
+			errs = append(errs, err)
```
staebler, line 224: The `validateServiceEndpoints` function should be returning a `field.ErrorList` so that each error returned can be specific to the field that failed validation.
For example, this error should point to the url field in the specific item of the serviceEndpoints list (where the field path to that list is passed into the function).
```suggestion
			errs = append(errs, field.Invalid(fldPath.Index(i).Child("url), service.URL, err.Error()))
```

patrickdillon:
I'm not as familiar with AWS as other platforms: do we need to resolve the endpoints before creating the cluster for things like validation, defaults, etc? 

My initial reaction is we should move this check so that it only happens before cluster provisioning, e.g. move it here: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/platformprovisioncheck.go but perhaps we need th endpoints for pre-provisioning logic...

staebler:
> I'm not as familiar with AWS as other platforms: do we need to resolve the endpoints before creating the cluster for things like validation, defaults, etc?
> 
> My initial reaction is we should move this check so that it only happens before cluster provisioning, e.g. move it here: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/platformprovisioncheck.go but perhaps we need the endpoints for creating assets other than the cluster.

My understanding of the purpose of platformprovisioncheck is to validate things conditions that must be met in order to provision for an IPI install. The endpoints are used for IPI and UPI, so I don't think it is appropriate to have the endpoint validation in platformprovisioncheck.

patrickdillon:
It looks like the endpoints are needed for at least the `PreferredInstanceType` function when creating the machine manifests, so I agree this makes sense to put them here.

pkg/asset/installconfig/aws/validation.go
jstuever: This is proving a difficult task. As we discussed today, running a server would be great if we were testing if net.Dial works as expected, but does little to ensure the code here is doing what we expect. Furthermore, we really don't want to be doing external connections for things like DNS. When looking to use gomock, I found it is actually a lot more code because we would need to pass an object into the platform validation function for platforms in order to mock the net functionality. Probably not what we would want either. I'm advocating to continue ignoring net.dial validations for a specified URL, be it `e2e.local` or something more generic such as `localhost`.

patrickdillon: I think in this case we should be using [`httptest`](https://golang.org/pkg/net/http/httptest/) rather than `gomock`. You can pass the URLs for the test servers in the `ServiceEndpoint`s and assert that the behavior of the function matches the responses of test servers, i.e. returns errors or passes as expected.

staebler: Let's put it in as is. We can improve up on later if we have a good idea on how to do it easily. As you told me earlier today, the worst that happens with this logic in the function is that we do not fail an installation early that incorrectly uses "e2e.local" as a service endpoint. I can live with that.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -211,12 +212,34 @@ func validateDuplicateSubnetZones(fldPath *field.Path, subnets map[string]Subnet
 }
 
 func validateServiceEndpoints(fldPath *field.Path, region string, services []awstypes.ServiceEndpoint) error {
```
staebler, line 214: Make the return `field.ErrorList` so that each error can be added individually rather than aggregated in this function only to be aggregated again later.
```suggestion
func validateServiceEndpoints(fldPath *field.Path, region string, services []awstypes.ServiceEndpoint) field.ErrorList {
```
This means also changing the local error about failing to find an endpoint for a service to a `field.Invalid` error in this function rather than having the caller put the error into a `field.Invalid` error.

jstuever: Great idea. I hadn't even considered changing the overall function workflow. Done.

staebler (commented):
/lgtm
/approve

PR #4705, opened 2021-03-03
OpenStack: limit icmp traffic to control-plane

adduarte:
Limits icmp traffic destined to the control-plane
to only traffic sourced from ip addresses in the
machineSubnet CIDR


EmilienM:
What about UPI?

PR #4707, opened 2021-03-03
OpenStack: Availability zones for root volumes

Fedosin:
This PR allows users to specify custom availability zones for their root volumes.

Implements: https://issues.redhat.com/browse/OSASINFRA-2226
Enhancement proposal: https://github.com/openshift/enhancements/pull/691

### DIFF
--- a/data/data/openstack/masters/main.tf
+++ b/data/data/openstack/masters/main.tf
@@ -38,6 +38,8 @@ resource "openstack_blockstorage_volume_v3" "master_volume" {
   size = var.root_volume_size
   volume_type = var.root_volume_type
   image_id = var.base_image_id
+
+  availability_zone = var.root_volume_zones[count.index % length(var.root_volume_zones)]
 }
 
 resource "openstack_compute_servergroup_v2" "master_group" {
--- a/pkg/asset/machines/openstack/machines.go
+++ b/pkg/asset/machines/openstack/machines.go
@@ -46,6 +46,11 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 		return nil, err
 	}
 
+	volumeAZs := openstackdefaults.DefaultRootVolumeAZ()
+	if mpool.RootVolume != nil && len(mpool.RootVolume.Zones) != 0 {
+		volumeAZs = mpool.RootVolume.Zones
+	}
+
 	total := int64(1)
 	if pool.Replicas != nil {
 		total = *pool.Replicas
@@ -66,6 +71,7 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 				role,
 				userDataSecret,
 				trunkSupport,
+				volumeAZs[int(idx)%len(volumeAZs)],
 			)
 			if err != nil {
 				return nil, err
@@ -105,7 +111,7 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 	return machines, nil
 }
 
-func generateProvider(clusterID string, platform *openstack.Platform, mpool *openstack.MachinePool, osImage string, az string, role, userDataSecret string, trunkSupport bool) (*openstackprovider.OpenstackProviderSpec, error) {
+func generateProvider(clusterID string, platform *openstack.Platform, mpool *openstack.MachinePool, osImage string, az string, role, userDataSecret string, trunkSupport bool, rootVolumeAZ string) (*openstackprovider.OpenstackProviderSpec, error) {
 	var networks []openstackprovider.NetworkParam
 	if platform.MachinesSubnet != "" {
 		networks = []openstackprovider.NetworkParam{{
@@ -169,6 +175,7 @@ func generateProvider(clusterID string, platform *openstack.Platform, mpool *ope
 			SourceType: "image",
 			SourceUUID: osImage,
 			VolumeType: mpool.RootVolume.Type,
+			Zone:       rootVolumeAZ,
 		}
 	} else {
 		spec.Image = osImage
--- a/pkg/tfvars/openstack/openstack.go
+++ b/pkg/tfvars/openstack/openstack.go
@@ -23,30 +23,31 @@ import (
 )
 
 type config struct {
-	BaseImageName              string   `json:"openstack_base_image_name,omitempty"`
-	ExternalNetwork            string   `json:"openstack_external_network,omitempty"`
-	Cloud                      string   `json:"openstack_credentials_cloud,omitempty"`
-	FlavorName                 string   `json:"openstack_master_flavor_name,omitempty"`
-	APIFloatingIP              string   `json:"openstack_api_floating_ip,omitempty"`
-	IngressFloatingIP          string   `json:"openstack_ingress_floating_ip,omitempty"`
-	APIVIP                     string   `json:"openstack_api_int_ip,omitempty"`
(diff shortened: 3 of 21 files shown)


pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -44,6 +44,9 @@ func ValidateMachinePool(p *openstack.MachinePool, ci *CloudInfo, controlPlane b
 		if p.RootVolume.Size < minimumStorage {
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume").Child("size"), p.RootVolume.Size, fmt.Sprintf("Volume size must be greater than %d to use root volumes, had %d", minimumStorage, p.RootVolume.Size)))
 		}
+		if p.RootVolume.Zones != nil {
+			allErrs = append(allErrs, validateZones(p.RootVolume.Zones, ci.Zones, fldPath.Child("rootVolume").Child("zones"))...)
```
mandre, line 48: It seems like `ci.Zones` only contains [compute AZs](https://github.com/openshift/installer/blob/5a12073/pkg/asset/installconfig/openstack/validation/cloudinfo.go#L322), so unless your cinder AZs match your compute AZs, this validation will fail.

It would be nice also to differentiate nova AZs from cinder AZs.

mandre (commented):
Just a small comment about validation that's going to fail if your compute and cinder AZs don't match, otherwise lgtm.

pkg/asset/installconfig/openstack/validation/machinepool.go
Fedosin: right... I think we need to remove this validation, because currently gophercloud doesn't support listing volume availability zones.

mandre (commented):
/lgtm
/assign @staebler 
This PR revendors CAPO and add a new `zones` parameter under `rootVolume` for openstack machine pools, PTAL.

pkg/asset/machines/openstack/machines.go
```diff
@@ -46,6 +46,11 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 		return nil, err
 	}
 
+	volumeAZs := []string{""}
+	if mpool.RootVolume != nil && mpool.RootVolume.Zones != nil {
```
staebler, line 50: I think that you want to use the zones from the root volume only if the slice is non-empty. Otherwise, if it is an empty non-nil slice, you will get a panic later when you try to use the length of `volumeAZs` as a modulus later. I think, following through how the pool gets defined and set, that the zones will not be an empty non-nil slice, but it would still be better to have the correct check to be resilient to future changes with how the pool is set up.
```suggestion
	if mpool.RootVolume != nil && len(mpool.RootVolume.Zones) != 0 {
```

data/data/openstack/masters/main.tf
```diff
@@ -30,14 +30,40 @@ data "ignition_config" "master_ignition_config" {
   ]
 }
 
-resource "openstack_blockstorage_volume_v3" "master_volume" {
-  name = "${var.cluster_id}-master-${count.index}"
+resource "openstack_blockstorage_volume_v3" "master_volume_0" {
```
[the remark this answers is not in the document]
Fedosin: done

pkg/asset/machines/openstack/machines.go
```diff
@@ -46,6 +46,11 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 		return nil, err
 	}
 
+	volumeAZs := []string{""}
```
staebler, line 49: Can we add the default `[]string{""}` value for `RootVolume.Zones` to `defaultOpenStackMachinePoolPlatform` so that we don't have to have this code here?

pkg/tfvars/openstack/openstack.go
```diff
@@ -68,6 +69,10 @@ func TFVars(masterConfigs []*v1alpha1.OpenstackProviderSpec, cloud string, exter
 		MasterAvailabilityZones: zones,
 	}
 
+	if mpool != nil && mpool.RootVolume != nil {
+		cfg.MasterRootVolumeAvalabilityZones = mpool.RootVolume.Zones
```
staebler, line 73: The `mpool` passed into this function is `ControlPlane.Platform.OpenStack`. It does not consider whether the zones may be set in `Openstack.DefaultMachinePlatform`.

staebler (commented):
Please improve on the commit messages. Between the commit messages being sparse and the PR description being empty, we are missing a lot of color around these changes.

pkg/asset/machines/openstack/machines.go
Fedosin: I think we can't... `RootVolume` is an optional parameter for machine pool, and it's not enabled by default (in other words it is `nil` in `defaultOpenStackMachinePoolPlatform`). If we add `[]string{""}` value for `RootVolume.Zones`, it means that the default would be:

```go
openstacktypes.MachinePool{
		Zones: []string{""},
		RootVolume: &openstacktypes.RootVolume{
			Zones: []string{""},
		},
	}
```

And we always create the RootVolume even if it's not required.

Fedosin: Instead, I added the default value in "github.com/openshift/installer/pkg/types/openstack/defaults" which imo is more appropriate.

pkg/tfvars/openstack/openstack.go
Fedosin: yeah, now it's fixed

pkg/types/openstack/defaults/machinepool.go
```diff
@@ -0,0 +1,6 @@
+package defaults
+
+// SetRootVolumeAZDefault sets the defaults for Root Volume availability zones.
+func SetRootVolumeAZDefault() []string {
```
staebler, line 4: This function does not **Set** anything. It only returns a value.

pkg/asset/machines/openstack/machines.go
staebler: Ah, that makes sense about not being able to set it in `defaultOpenStackMachinePoolPlatform`. It would be nice to be able to set the default value when the install config defaults are set then, but I won't hold up this PR for that.

pkg/tfvars/openstack/openstack.go
```diff
@@ -56,16 +57,21 @@ func TFVars(masterConfigs []*v1alpha1.OpenstackProviderSpec, cloud string, exter
...
+		IngressFloatingIP:                 ingressFloatingIP,
+		APIVIP:                            apiVIP,
+		IngressVIP:                        ingressVIP,
+		ExternalDNS:                       externalDNS,
+		MachinesSubnet:                    machinesSubnet,
+		MasterAvailabilityZones:           zones,
+		MasterRootVolumesAvalabilityZones: []string{""},
+	}
+
+	if mpool != nil && mpool.RootVolume != nil && mpool.RootVolume.Zones != nil {
```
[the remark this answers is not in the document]
mandre: Agreed, I was also confused by this `mpool.RootVolume.Zones != nil` check at first and thought it could cause a bug. Looks like it's fine and we respect the user provided mpool after all, but this code it still confusing and not necessary.

staebler (commented):
This looks good.
Please improve the commit messages. At a minimum, make it clear that the changes are for OpenStack.

PR #4708, opened 2021-03-03
vsphere: Validate vcenter user input to match RFC standards

rna-afk:
Vcenter input must always only be the hostname yet it is possible
for the user to provide other types of input like URLs. URLs would
cause an error to occur only during the time of cluster creation
as connection to the vcenter would succeed.

Adding a validation check for the user input for the vcenter value
so that it conforms to the RFC-1035 standard and forces the user
to re-enter the value rather than throw an error during cluster
creation.

### DIFF
--- a/pkg/asset/installconfig/vsphere/vsphere.go
+++ b/pkg/asset/installconfig/vsphere/vsphere.go
@@ -102,9 +102,11 @@ func getClients() (*vCenterClient, error) {
 		{
 			Prompt: &survey.Input{
 				Message: "vCenter",
-				Help:    "The hostname of the vCenter to be used for installation.",
+				Help:    "The domain name or IP address of the vCenter to be used for installation.",
 			},
-			Validate: survey.Required,
+			Validate: survey.ComposeValidators(survey.Required, func(ans interface{}) error {
+				return validate.Host(ans.(string))
+			}),
 		},
 	}, &vcenter); err != nil {
 		return nil, errors.Wrap(err, "failed UserInput")
--- a/pkg/types/vsphere/validation/platform.go
+++ b/pkg/types/vsphere/validation/platform.go
@@ -29,8 +29,10 @@ func ValidatePlatform(p *vsphere.Platform, fldPath *field.Path) field.ErrorList 
 		allErrs = append(allErrs, field.Required(fldPath.Child("defaultDatastore"), "must specify the default datastore"))
 	}
 
-	if strings.ToLower(p.VCenter) != p.VCenter {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("vCenter"), p.VCenter, "must be all lower case"))
+	if len(p.VCenter) != 0 {
+		if err := validate.Host(p.VCenter); err != nil {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("vCenter"), p.VCenter, "must be the domain name or IP address of the vCenter"))
+		}
 	}
 
 	// If all VIPs are empty, skip IP validation.  All VIPs are required to be defined together.
--- a/pkg/validate/validate.go
+++ b/pkg/validate/validate.go
@@ -238,3 +238,16 @@ func UUID(val string) error {
 	_, err := uuid.Parse(val)
 	return err
 }
+
+// Host validates that a given string is a valid URI host.
+func Host(v string) error {
+	proxyIP := net.ParseIP(v)
+	if proxyIP != nil {
+		return nil
+	}
+	re := regexp.MustCompile("^[a-z]")
+	if !re.MatchString(v) {
+		return errors.New("domain name must begin with a lower-case letter")
+	}
+	return validateSubdomain(v)
+}
--- a/pkg/validate/validate_test.go
+++ b/pkg/validate/validate_test.go
@@ -49,6 +49,86 @@ func TestClusterName(t *testing.T) {
 	}
 }
 
+func TestClusterName1035(t *testing.T) {
+	maxSizeName := "a" + strings.Repeat("123456789.", 5) + "123"
+
+	cases := []struct {
+		name        string
+		clusterName string
+		valid       bool
+	}{
+		{"empty", "", false},
+		{"only whitespace", " ", false},
+		{"single lowercase", "a", true},
+		{"single uppercase", "A", false},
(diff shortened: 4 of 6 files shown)


patrickdillon:
/cc jstuever

This validation is only being applied to survey. It should also be applied to the install config.

pkg/asset/installconfig/vsphere/vsphere.go
```diff
@@ -104,7 +104,9 @@ func getClients() (*vCenterClient, error) {
 				Message: "vCenter",
 				Help:    "The hostname of the vCenter to be used for installation.",
 			},
-			Validate: survey.Required,
+			Validate: survey.ComposeValidators(survey.Required, func(ans interface{}) error {
+				return validate.ClusterName1035(ans.(string))
```
staebler, line 108: This check is not quite what we need.
1) It is valid for the vCenter hostname to be an IP address.
1) We should not validate the length of the hostname.

pkg/types/vsphere/validation/platform.go
```diff
@@ -29,8 +29,11 @@ func ValidatePlatform(p *vsphere.Platform, fldPath *field.Path) field.ErrorList
 		allErrs = append(allErrs, field.Required(fldPath.Child("defaultDatastore"), "must specify the default datastore"))
 	}
 
-	if strings.ToLower(p.VCenter) != p.VCenter {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("vCenter"), p.VCenter, "must be all lower case"))
+	if len(p.VCenter) != 0 {
+		if err := validate.ClusterName1035(p.VCenter); err != nil {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("vCenter"), p.VCenter, "must conform to RFC 1035 standard. "+
```
staebler, line 34: I don't think that we need to show this much detail in the error here. The user is not picking a hostname: The user is inputting what should be an existing hostname. This error should say something simpler like
```
must be the hostname of the vCenter
```

pkg/validate/validate_test.go
```diff
@@ -49,6 +49,47 @@ func TestClusterName(t *testing.T) {
 	}
 }
 
+func TestClusterName1035(t *testing.T) {
```
staebler, line 52: Thanks for adding this test?

rna-afk: Was using that function and it didn't have a test so thought I'd add it.

pkg/validate/validate.go
```diff
@@ -238,3 +238,16 @@ func UUID(val string) error {
 	_, err := uuid.Parse(val)
 	return err
 }
+
+// VCenter validates that a given string is either a valid IP or a hostname.
+func VCenter(v string) error {
```
staebler, line 243: Let's not make this specific to vSphere.
```suggestion
// Host validates that a given string is a valid URI host.
func Host(v string) error {
```

pkg/types/vsphere/validation/platform.go
```diff
@@ -29,8 +29,10 @@ func ValidatePlatform(p *vsphere.Platform, fldPath *field.Path) field.ErrorList
 		allErrs = append(allErrs, field.Required(fldPath.Child("defaultDatastore"), "must specify the default datastore"))
 	}
 
-	if strings.ToLower(p.VCenter) != p.VCenter {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("vCenter"), p.VCenter, "must be all lower case"))
+	if len(p.VCenter) != 0 {
+		if err := validate.VCenter(p.VCenter); err != nil {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("vCenter"), p.VCenter, "must be the hostname of the vCenter"))
```
staebler, line 34: After reading a bit more on the RFCs, the term hostname is not quite inclusive enough here.
```suggestion
			allErrs = append(allErrs, field.Invalid(fldPath.Child("vCenter"), p.VCenter, "must be the hostname or IP address of the vCenter"))
```

Although in the `vCenter` field description, we use domain name instead of hostname. We should probably stay consistent.
```suggestion
			allErrs = append(allErrs, field.Invalid(fldPath.Child("vCenter"), p.VCenter, "must be the domain name or IP address of the vCenter"))
```

pkg/asset/installconfig/vsphere/vsphere.go
```diff
@@ -104,7 +104,9 @@ func getClients() (*vCenterClient, error) {
 				Message: "vCenter",
 				Help:    "The hostname of the vCenter to be used for installation.",
```
staebler, line 105:
```suggestion
				Help:    "The hostname or IP address of the vCenter to be used for installation.",
```

pkg/asset/installconfig/vsphere/vsphere.go
```diff
@@ -102,9 +102,11 @@ func getClients() (*vCenterClient, error) {
 		{
 			Prompt: &survey.Input{
 				Message: "vCenter",
-				Help:    "The hostname of the vCenter to be used for installation.",
+				Help:    "The hostname or IP address of the vCenter to be used for installation.",
```
staebler, line 105: For consistency,
```suggestion
				Help:    "The domain name or IP address of the vCenter to be used for installation.",
```

pkg/validate/validate.go
```diff
@@ -238,3 +238,16 @@ func UUID(val string) error {
 	_, err := uuid.Parse(val)
 	return err
 }
+
+// Host validates that a given string is either a valid URI host.
```
staebler, line 242:
```suggestion
// Host validates that a given string is a valid URI host.
```

pkg/validate/validate.go
```diff
@@ -238,3 +238,16 @@ func UUID(val string) error {
...
+func Host(v string) error {
+	proxyIP := net.ParseIP(v)
+	if proxyIP != nil {
+		return nil
+	}
+	re := regexp.MustCompile("^[a-z]")
+	if !re.MatchString(v) {
+		return errors.New("cluster name must begin with a lower-case letter")
```
staebler, line 250:
```suggestion
		return errors.New("domain name must begin with a lower-case letter")
```

staebler (commented):
Looks good. Just two small nits.

PR #4719, opened 2021-03-07
Bug 1936223: [ovirt] Fix typo in connect to Engine

zerodayz:
This is to fix the typos in ovirt OCP installer in sentences such as:

'connect to Engine API' -> 'connect to the Engine API'

### DIFF
--- a/pkg/asset/installconfig/ovirt/credentials.go
+++ b/pkg/asset/installconfig/ovirt/credentials.go
@@ -108,7 +108,7 @@ func (c *clientHTTP) checkURLResponse() error {
 	return nil
 }
 
-// askPassword will ask the password to connect to Engine API.
+// askPassword will ask the password to connect to the Engine API.
 // The password provided will be added in the Config struct.
 // If an error happens, it will ask again username for users.
 func askPassword(c *Config) error {
@@ -138,7 +138,7 @@ func askPassword(c *Config) error {
 	return nil
 }
 
-// askUsername will ask username to connect to Engine API.
+// askUsername will ask username to connect to the Engine API.
 // The username provided will be added in the Config struct.
 // Returns Config and error if failure.
 func askUsername(c *Config) error {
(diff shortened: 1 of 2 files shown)


pkg/asset/installconfig/ovirt/credentials.go
```diff
@@ -343,7 +343,7 @@ func engineSetup() (Config, error) {
 			} else {
 				answer, err = askQuestionTrueOrFalse(
 					"Would you like to import another PEM bundle?",
-					"Users are able to use it's own PEM bundle to connect to Engine API")
+					"Users are able to use it's own PEM bundle to connect to the Engine API")
```
stoobie, line 346: Who is the "it" in this sentence? If it's the user, I suggest: "You can use your own PEM bundle..."

stoobie (changes requested):

pkg/asset/installconfig/ovirt/credentials.go
staebler: I agree with @stoobie that the sentence structure could use some work. If, however, we keep it as is, then let's at least fix the `it's`.
```suggestion
					"Users are able to use its own PEM bundle to connect to the Engine API")
```

zerodayz: Changing to: 'You can use your own PEM bundle to connect to the Engine API'

zerodayz:
@stoobie @staebler  Updated according to your suggestions.

zerodayz:
@stoobie @staebler Can you please review the last changes? Thank you!

staebler (commented):
Looks fine to me. But one of the ovirt owners should review this.
/cc @Gal-Zaidman @janoszen

janoszen (approved):

janoszen:
Thanks for this @zerodayz

PR #4720, opened 2021-03-08
kubevirt platfrom validate permissions

ydayagi:
Added validation of user permissions for platform kubevirt

https://issues.redhat.com/browse/CNV-10142

### DIFF
--- a/pkg/asset/installconfig/kubevirt/validation.go
+++ b/pkg/asset/installconfig/kubevirt/validation.go
@@ -2,9 +2,11 @@ package kubevirt
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/openshift/installer/pkg/types"
+	authv1 "k8s.io/api/authorization/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/asset/installconfig/kubevirt/validation.go
```diff
@@ -21,6 +23,42 @@ func Validate(ic *types.InstallConfig, client Client) error {
...
+
+	// Put all missing permissions in one error message
+	if len(notAllowedObjs) > 0 {
+		errMsg := "The user is missing the following permissions: "
+
+		for _, obj := range notAllowedObjs {
+			errMsg += fmt.Sprintf("%+v, ", *obj.Spec.ResourceAttributes)
+		}
+
+		return errors.New(errMsg)
```
staebler, line 56: Use `strings.Join` so that you don't end up with a trailing comma.
```suggestion
		var notAllowed []string
		for _, obj := range notAllowedObjs {
			notAllowed = append(notAllowed, fmt.Sprintf("%+v", *obj.Spec.ResourceAttributes))
		}
		
		return fmt.Errorf("The user is missing the following permissions: %s", strings.Join(notAllowed, ", "))
```

staebler (commented):
Looks good. Just one small nit.

PR #4726, opened 2021-03-09
Bug 1931115: Azure: Check Azure disk Instance Type for PremiumIO Capabilities

rna-afk:
If the Azure disktype is Premium_LRS, a check must be made to see
if the given InstanceType has the PremiumIO capabilities which when
not present causes creation of clusters to fail. Adding the check
which iterates through all the capabilities of the given instance
type disk to find the PremiumIO, check if it is set to True and
return error otherwise.

### DIFF
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/openshift/installer/pkg/types"
 	aztypes "github.com/openshift/installer/pkg/types/azure"
+	"github.com/openshift/installer/pkg/types/azure/defaults"
 )
 
 type resourceRequirements struct {
@@ -44,7 +45,7 @@ func Validate(client API, ic *types.InstallConfig) error {
 }
 
 // ValidateInstanceType ensures the instance type has sufficient Vcpu and Memory.
-func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceType string, req resourceRequirements) field.ErrorList {
+func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceType string, diskType string, req resourceRequirements) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	typeMeta, err := client.GetVirtualMachineSku(context.TODO(), instanceType, region)
@@ -77,6 +78,11 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d GB Memory", req.minimumMemory)
 				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
 			}
+		} else if diskType == "Premium_LRS" && strings.EqualFold(*capability.Name, "PremiumIO") {
+			if strings.EqualFold(*capability.Value, "False") {
+				errMsg := fmt.Sprintf("PremiumIO not supported for instance type %s", instanceType)
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("osDisk", "diskType"), diskType, errMsg))
+			}
 		}
 	}
 
@@ -87,25 +93,49 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	// Default requirements need to be sufficient to support Control Plane instances.
-	defaultInstanceReq := controlPlaneReq
+	defaultDiskType := aztypes.DefaultDiskType
+	defaultInstanceType := ""
 
-	if ic.ControlPlane != nil && ic.ControlPlane.Platform.Azure != nil && ic.ControlPlane.Platform.Azure.InstanceType != "" {
-		// Default requirements can be relaxed when the controlPlane type is set explicitly.
-		defaultInstanceReq = computeReq
-
-		allErrs = append(allErrs, ValidateInstanceType(client, field.NewPath("controlPlane", "platform", "azure"), ic.Azure.Region, ic.ControlPlane.Platform.Azure.InstanceType, controlPlaneReq)...)
+	if ic.Platform.Azure.DefaultMachinePlatform != nil && ic.Platform.Azure.DefaultMachinePlatform.OSDisk.DiskType != "" {
+		defaultDiskType = ic.Platform.Azure.DefaultMachinePlatform.OSDisk.DiskType
+	}
+	if ic.Platform.Azure.DefaultMachinePlatform != nil && ic.Platform.Azure.DefaultMachinePlatform.InstanceType != "" {
+		defaultInstanceType = ic.Platform.Azure.DefaultMachinePlatform.InstanceType
 	}
 
-	if ic.Platform.Azure.DefaultMachinePlatform != nil && ic.Platform.Azure.DefaultMachinePlatform.InstanceType != "" {
-		allErrs = append(allErrs, ValidateInstanceType(client, field.NewPath("platform", "azure", "defaultMachinePlatform"), ic.Azure.Region, ic.Platform.Azure.DefaultMachinePlatform.InstanceType, defaultInstanceReq)...)
+	if ic.ControlPlane != nil && ic.ControlPlane.Platform.Azure != nil {
+		diskType := ic.ControlPlane.Platform.Azure.OSDisk.DiskType
+		instanceType := ic.ControlPlane.Platform.Azure.InstanceType
+
+		if diskType == "" {
+			diskType = defaultDiskType
+		}
+		if instanceType == "" {
+			instanceType = defaultInstanceType
+		}
+		if instanceType == "" {
+			instanceType = defaults.ControlPlaneInstanceType(ic.Azure.Region)
+		}
+		allErrs = append(allErrs, ValidateInstanceType(client, field.NewPath("controlPlane", "platform", "azure"), ic.Azure.Region, instanceType, diskType, controlPlaneReq)...)
 	}
 
 	for idx, compute := range ic.Compute {
 		fieldPath := field.NewPath("compute").Index(idx)
-		if compute.Platform.Azure != nil && compute.Platform.Azure.InstanceType != "" {
+		if compute.Platform.Azure != nil {
+			diskType := compute.Platform.Azure.OSDisk.DiskType
+			instanceType := compute.Platform.Azure.InstanceType
+
+			if diskType == "" {
+				diskType = defaultDiskType
+			}
+			if instanceType == "" {
+				instanceType = defaultInstanceType
+			}
+			if instanceType == "" {
+				instanceType = defaults.ComputeInstanceType(ic.Azure.Region)
+			}
 			allErrs = append(allErrs, ValidateInstanceType(client, fieldPath.Child("platform", "azure"),
-				ic.Azure.Region, compute.Platform.Azure.InstanceType, computeReq)...)
+				ic.Azure.Region, instanceType, diskType, computeReq)...)
 		}
 	}
 
--- a/pkg/asset/installconfig/azure/validation_test.go
+++ b/pkg/asset/installconfig/azure/validation_test.go
@@ -37,9 +37,13 @@ var (
 	validResourceSkuRegions        = "southeastasia"
 
 	instanceTypeSku = []*azsku.ResourceSku{
-		{Name: to.StringPtr("Standard_A1_v2"), Capabilities: &[]azsku.ResourceSkuCapabilities{{Name: to.StringPtr("vCPUs"), Value: to.StringPtr("1")}, {Name: to.StringPtr("MemoryGB"), Value: to.StringPtr("2")}}},
-		{Name: to.StringPtr("Standard_D2_v4"), Capabilities: &[]azsku.ResourceSkuCapabilities{{Name: to.StringPtr("vCPUs"), Value: to.StringPtr("2")}, {Name: to.StringPtr("MemoryGB"), Value: to.StringPtr("8")}}},
-		{Name: to.StringPtr("Standard_D4_v4"), Capabilities: &[]azsku.ResourceSkuCapabilities{{Name: to.StringPtr("vCPUs"), Value: to.StringPtr("4")}, {Name: to.StringPtr("MemoryGB"), Value: to.StringPtr("16")}}},
+		{Name: to.StringPtr("Standard_D4s_v3"), Capabilities: &[]azsku.ResourceSkuCapabilities{{Name: to.StringPtr("vCPUs"), Value: to.StringPtr("4")}, {Name: to.StringPtr("MemoryGB"), Value: to.StringPtr("16")}, {Name: to.StringPtr("PremiumIO"), Value: to.StringPtr("True")}}},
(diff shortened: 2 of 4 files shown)


pkg/asset/installconfig/azure/validation.go
```diff
@@ -94,18 +98,18 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
 		// Default requirements can be relaxed when the controlPlane type is set explicitly.
 		defaultInstanceReq = computeReq
 
-		allErrs = append(allErrs, ValidateInstanceType(client, field.NewPath("controlPlane", "platform", "azure"), ic.Azure.Region, ic.ControlPlane.Platform.Azure.InstanceType, controlPlaneReq)...)
+		allErrs = append(allErrs, ValidateInstanceType(client, field.NewPath("controlPlane", "platform", "azure"), ic.Azure.Region, ic.ControlPlane.Platform.Azure.InstanceType, ic.ControlPlane.Platform.Azure.OSDisk.DiskType, controlPlaneReq)...)
 	}
 
 	if ic.Platform.Azure.DefaultMachinePlatform != nil && ic.Platform.Azure.DefaultMachinePlatform.InstanceType != "" {
```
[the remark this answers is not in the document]
rna-afk: What kind of validations are you thinking of? I think most of the diskType validations are in the machinepool.go file and was thinking let those validations take over while diskType=="Premium_LRS" would be enough (the value is not nil). Could check the az account.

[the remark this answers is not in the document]
staebler: Not validations. If the disk type is not defined in the individual machine pool, then it comes from the default machine pool. Your new code is always assuming that the disk type will be defined in the individual machine pool.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -77,6 +77,11 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d GB Memory", req.minimumMemory)
 				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
 			}
+		} else if diskType == "Premium_LRS" && strings.EqualFold(*capability.Name, "PremiumIO") {
+			if strings.EqualFold(*capability.Value, "False") {
+				errMsg := fmt.Sprintf("PremiumIO not supported for disk type with instance type %s", instanceType)
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("diskType"), diskType, errMsg))
```
staebler, line 83:
```suggestion
				allErrs = append(allErrs, field.Invalid(fieldPath.Child("osDisk", "diskType"), diskType, errMsg))
```

pkg/asset/installconfig/azure/validation.go
```diff
@@ -89,23 +94,35 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
 
 	// Default requirements need to be sufficient to support Control Plane instances.
 	defaultInstanceReq := controlPlaneReq
+	defaultDiskType := ""
```
staebler, line 97:
```suggestion
	defaultDiskType := "Premium_LRS"
```
As ugly as it is, the default is actually defined with the machines assets.
https://github.com/openshift/installer/blob/aa9ff594a8d582acf687303f18345ccc39b106b4/pkg/asset/machines/worker.go#L97

Let's define the default in a constant in `pkg/types/azure/machinepool.go`.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -89,23 +94,35 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+
+	if ic.Platform.Azure.DefaultMachinePlatform != nil {
+		defaultDiskType = ic.Platform.Azure.DefaultMachinePlatform.OSDisk.DiskType
+	}
 
 	if ic.ControlPlane != nil && ic.ControlPlane.Platform.Azure != nil && ic.ControlPlane.Platform.Azure.InstanceType != "" {
```
staebler, line 103: This conditional is still not quite right. The check for `ValidateInstanceType` needs to happen even if the instance type is not set explicitly on the control plane machine. That is because the user could set the disk type explicitly for the control plane machine without setting the instance type.

My suggestion is to do the following.
1. Determine the actual instance type used for the control plane machines and the compute machines in the same manner in which you determine the actual disk type used.
1. Get rid of the validation on the default machine platform.

mjudeikis:
Looks good from ARO perspective as a feature :) nit picking leaving for you :)

pkg/asset/installconfig/azure/validation.go
rna-afk: Isn't it possible for the user to set the DefaultMachinePlatform.InstanceType in the install-config file?

staebler: Yes. And that was covered by https://github.com/openshift/installer/blob/f85551fa5b0e94fb6083bec91e642bfddd3c5e55/pkg/asset/installconfig/azure/validation.go#L100-L102.

staebler: But that is not sufficient any more, since we need to consider both the instance type and the disk type from the default machine platform.

rna-afk: > Get rid of the validation on the default machine platform.
I think I got confused with this point. I understand that I need to figure out the actual instance type used but could to tell me which validation to get rid of here?

staebler: Get rid of the code that I linked two comments up.

rna-afk: > But that is not sufficient any more, since we need to consider both the instance type and the disk type from the default machine platform.

Was thinking since the installer sets default values to both instance type and disk type, if user sets either of these, we could still validate with the installer defaults. This would apply to control-plane, compute and defaultMachinePlatform.

tsze-redhat:
Do you guys have any combo that will pass / fail this check (that the machine type specified supports Premium_LRS)?

rna-afk:
Checked az vm list-skus --size D4 -l centralus command
Standard_D48_v4 seems to not have PremiumIO in centralus so it would fail.
Standard_D48s_v4 seems to have PremiumIO in centralus so it would pass.

staebler:
There are some details in the bug, which should be linked to this PR.

/retitle Bug 1931115: Azure: Check Azure disk Instance Type for PremiumIO Capabilities

pkg/asset/installconfig/azure/validation.go
```diff
@@ -77,6 +77,11 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d GB Memory", req.minimumMemory)
 				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
 			}
+		} else if diskType == "Premium_LRS" && strings.EqualFold(*capability.Name, "PremiumIO") {
+			if strings.EqualFold(*capability.Value, "False") {
+				errMsg := fmt.Sprintf("PremiumIO not supported for disk type with instance type %s", instanceType)
```
staebler, line 82:
```suggestion
				errMsg := fmt.Sprintf("PremiumIO not supported for instance type %s", instanceType)
```

pkg/asset/installconfig/azure/validation.go
```diff
@@ -88,24 +93,43 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
 	allErrs := field.ErrorList{}
 
 	// Default requirements need to be sufficient to support Control Plane instances.
```
staebler, line 95: This comment does not make sense any more. Remove it.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -88,24 +93,43 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+		defaultDiskType = ic.Platform.Azure.DefaultMachinePlatform.OSDisk.DiskType
 	}
-
 	if ic.Platform.Azure.DefaultMachinePlatform != nil && ic.Platform.Azure.DefaultMachinePlatform.InstanceType != "" {
-		allErrs = append(allErrs, ValidateInstanceType(client, field.NewPath("platform", "azure", "defaultMachinePlatform"), ic.Azure.Region, ic.Platform.Azure.DefaultMachinePlatform.InstanceType, defaultInstanceReq)...)
+		defaultInstanceType = ic.Platform.Azure.DefaultMachinePlatform.InstanceType
+	}
+
+	if ic.ControlPlane != nil && ic.ControlPlane.Platform.Azure != nil {
+		// Default requirements can be relaxed when the controlPlane type is set explicitly.
```
staebler, line 107: Likewise, this comment should be removed.

pkg/asset/installconfig/azure/validation_test.go
```diff
@@ -193,11 +201,6 @@ func TestAzureInstallConfigValidation(t *testing.T) {
 			edits:    editFunctions{validInstanceTypes},
 			errorMsg: "",
 		},
-		{
-			name:     "Invalid default machine type",
```
staebler, line 197: Keep this test case.

pkg/asset/installconfig/azure/validation_test.go
```diff
@@ -208,11 +211,6 @@ func TestAzureInstallConfigValidation(t *testing.T) {
 			edits:    editFunctions{invalidateComputeInstanceTypes},
 			errorMsg: `\[compute\[0\].platform.azure.type: Invalid value: "Standard_A1_v2": instance type does not meet minimum resource requirements of 2 vCPUs, compute\[0\].platform.azure.type: Invalid value: "Standard_A1_v2": instance type does not meet minimum resource requirements of 8 GB Memory\]`,
 		},
-		{
-			name:     "Undefined default instance types",
```
staebler, line 212: Keep this test case.

pkg/types/azure/machinepool.go
```diff
@@ -36,6 +36,12 @@ type OSDisk struct {
 	DiskType string `json:"diskType"`
 }
 
+// DefaultDiskType holds the default Azure disk type used by the VMs.
+const DefaultDiskType string = "Premium_LRS"
+
+//DefaultInstanceType holds the default Azure instance type used by VMs.
+const DefaultInstanceType string = "Standard_D4s_v3"
```
staebler, line 43: I steered you wrong on this constant. You should be calling these functions instead.
https://github.com/openshift/installer/blob/eb47ed1dea8b4ac393e58d1b6bd19d7edb113908/pkg/types/azure/defaults/machines.go#L19
https://github.com/openshift/installer/blob/eb47ed1dea8b4ac393e58d1b6bd19d7edb113908/pkg/types/azure/defaults/machines.go#L27

pkg/types/azure/machinepool.go
```diff
@@ -36,6 +36,12 @@ type OSDisk struct {
 	DiskType string `json:"diskType"`
 }
 
+// DefaultDiskType holds the default Azure disk type used by the VMs.
+const DefaultDiskType string = "Premium_LRS"
```
staebler, line 40: Use this constant here, too.
https://github.com/openshift/installer/blob/eb47ed1dea8b4ac393e58d1b6bd19d7edb113908/pkg/asset/machines/worker.go#L97

rna-afk: Tried it and it gives me import cycle error. Any other suggestions?

staebler: That does not make any sense. The `pkg/types/azure` package does not import any other installer packages. Can you be specific about (1) what you tried and (2) what the error was?

rna-afk: Ah got it. Was trying to avoid this default variable by directly using it in the validation file. I'll make the change thanks.

rna-afk:
```
const DefaultDiskType string = machines.DefaultAzureMachinePoolPlatform().OSDisk.DiskType

package github.com/openshift/installer/cmd/openshift-install
	imports github.com/openshift/installer/cmd/openshift-install/migrate/azure
	imports github.com/openshift/installer/pkg/migrate/azure
	imports github.com/openshift/installer/pkg/asset/installconfig/azure
	imports github.com/openshift/installer/pkg/types
	imports github.com/openshift/installer/pkg/types/azure
	imports github.com/openshift/installer/pkg/asset/machines
	imports github.com/openshift/installer/pkg/asset/ignition/machine
	imports github.com/openshift/installer/pkg/asset/installconfig
	imports github.com/openshift/installer/pkg/asset/installconfig/azure: import cycle not allowed

```

staebler: Yep, you've got it backwards. Keep the default defined in `pkg/asset/machines/worker.go`. Use the default in `DefaultAzureMachinePoolPlatform`.

rna-afk: ah makes sense

pkg/asset/installconfig/azure/validation.go
```diff
@@ -87,25 +93,46 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
...
+		controlPlaneInstanceType := defaultInstanceType
+
+		if ic.ControlPlane.Platform.Azure.OSDisk.DiskType != "" {
+			controlPlaneDiskType = ic.ControlPlane.Platform.Azure.OSDisk.DiskType
+		}
+		if ic.ControlPlane.Platform.Azure.InstanceType != "" {
+			controlPlaneInstanceType = ic.ControlPlane.Platform.Azure.InstanceType
+		} else if controlPlaneInstanceType == "" {
+			controlPlaneInstanceType = defaults.ControlPlaneInstanceType(ic.Azure.Region)
+		}
```
staebler, line 117: Re-arrange this to make it more clear what the precedence order is. I recommend using a similar structure for the disk type, too.
```suggestion
		controlPlaneInstanceType := ic.ControlPlane.Platform.Azure.InstanceType
		if controlPlaneInstanceType == "" {
			controlPlaneInstanceType = defaultInstanceType
		}
		if controlPlaneInstanceType == "" {
			controlPlaneInstanceType = defaults.ControlPlaneInstanceType(ic.Azure.Region)
		}
```

pkg/asset/installconfig/azure/validation.go
```diff
@@ -87,25 +93,46 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
...
 	}
-
 	if ic.Platform.Azure.DefaultMachinePlatform != nil && ic.Platform.Azure.DefaultMachinePlatform.InstanceType != "" {
-		allErrs = append(allErrs, ValidateInstanceType(client, field.NewPath("platform", "azure", "defaultMachinePlatform"), ic.Azure.Region, ic.Platform.Azure.DefaultMachinePlatform.InstanceType, defaultInstanceReq)...)
+		defaultInstanceType = ic.Platform.Azure.DefaultMachinePlatform.InstanceType
+	}
+
+	if ic.ControlPlane != nil && ic.ControlPlane.Platform.Azure != nil {
+		controlPlaneDiskType := defaultDiskType
+		controlPlaneInstanceType := defaultInstanceType
```
staebler, line 108: Change the names of these variables to `diskType` and `instanceType`. The `controlPlane` part is not necessary are it can easily be implied from the context of the code block. Do the same thing for the compute, too.

staebler (commented):
Looks good. Just a couple small nits.

staebler (commented):
/lgtm
/approve

staebler:
These changes are only for Azure clusters.

PR #4727, opened 2021-03-09
pkg/asset/cluster/aws: fix dropped error

alrs:
This fixes a dropped error in `pkg/asset/cluster/aws`.


staebler (commented):
/ok-to-test
/lgtm
/approve

PR #4729, opened 2021-03-09
Bug 1981941: Upgrade terraform to 0.14.6

jhixson74:
This updates the installer to use Terraform 0.14.6. 

Terraform 0.13.x updated the on-disk layout of the plugins directory, how to access offline plugins, and a lock file used for plugin versions. plugin_versions_generate.go has been added which generates plugin_versions.go. All terraform plugin files the installer uses have been updated to use the plugin version. All terraform config files have been updated to use the local providers. A terraform.rc file has been added to specify where the local providers are. 

A temporary vendor path for terraform is in this PR that allows the location of the new plugin version lock file to be overridden. Once I can get everything green it will be added to our local terraform fork. 

https://issues.redhat.com/browse/CORS-1520

https://www.terraform.io/docs/language/providers/requirements.html
https://www.terraform.io/docs/language/dependency-lock.html


pkg/asset/cluster/cluster.go
```diff
@@ -70,7 +71,13 @@ func (c *Cluster) Generate(parents asset.Parents) (err error) {
 	if err != nil {
 		return errors.Wrap(err, "failed to create temp dir for terraform execution")
 	}
-	defer os.RemoveAll(tmpDir)
+
+	tfKeep := strings.ToLower(os.Getenv("OPENSHIFT_INSTALL_KEEP_TERRAFORM"))
```
staebler, line 75: Which terraform files are you looking to keep?

vendor/github.com/hashicorp/terraform/command/meta_dependencies.go
```diff
@@ -27,6 +28,25 @@ import (
 // that the root module directory is always the current working directory.
 const dependencyLockFilename = ".terraform.lock.hcl"
 
+// getDependencyLockFileName gets the dependency lock file name. If the
+// environment variable TERRAFORM_LOCK_FILE_PATH is set, the lock file is
+// appended to it.
+func getDependencyLockFilename() string {
```
staebler: Are we going to carry this change in our terraform fork?

pkg/terraform/exec/plugins/aws.go
```diff
@@ -11,5 +11,9 @@ func init() {
 			ProviderFunc: aws.Provider,
 		})
 	}
-	KnownPlugins["terraform-provider-aws"] = exec
+	KnownPlugins["terraform-provider-aws"] = &TFPlugin{
+		Name:    "aws",
+		Exec:    exec,
+		Version: GetAwsVersion(),
```
staebler, line 17: It is not clear to me whether it is important that we use the real version here, since we are not setting a version requirement in the `required_providers` block in the tf files. What happens if we just set all of the versions to v1.0.0 without prejudice?

pkg/terraform/exec/plugins/plugin_versions_generate.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+// getTerraformPluginVersions gathers the terraform provider version numbers
+func getTerraformPluginVersions(file *modfile.File, plugins *map[string]*TFPluginInfo) {
+	hre := regexp.MustCompile(`([a-fA-F0-9]{12})$`)
+	ore := regexp.MustCompile(`[[:space:]]+|\^[[:digit:]]+`)
+	pre := regexp.MustCompile(`terraform-provider-[[:alnum:]]+/v[[:digit:]]+`)
+	tre := regexp.MustCompile(`[[:digit:]]{14}-([a-fA-F0-9]{12})$`)
+	ure := regexp.MustCompile(`^(.+terraform-provider-[[:alnum:]]+)`)
+	vre := regexp.MustCompile(`([0-9\.]+)`)
```
staebler, line 96: Same concern here as above, where it is hard to discern what these variables are used for by their names.

pkg/terraform/exec/plugins/plugin_versions_generate.go
```diff
@@ -0,0 +1,249 @@
...
+	if err != nil {
+		log.Fatal("failed to get absolute path of output file")
+	}
+
+	// Read and parse go module into file object
+	mod, err := ioutil.ReadFile(goMod)
+	if err != nil {
+		log.Fatal(err)
+	}
+	file, err := modfile.Parse(goMod, mod, nil)
```
staebler, line 215: I would feel better about using `go list` rather than parsing the go.mod file. The `go list` is the source of truth for the modules.

```
go list -m -json all | jq 'select(.Path | contains("terraform-providers"))'
```

pkg/terraform/exec/plugins/plugin_versions_generate.go
```diff
@@ -0,0 +1,249 @@
...
+		var mod *module.Version = &pi.Module
+
+		// XXX: we assume https checkouts
+		url := fmt.Sprintf("https://%s", mod.Path)
+
+		// If this is a v2 or up  module, chop off the end
+		if pre.Match([]byte(mod.Path)) {
+			path := ure.FindString(mod.Path)
+			url = fmt.Sprintf("https://%s", path)
+		}
```
staebler, line 109: I think we can combine the regular expressions together so we only have to do one match rather than two separate matches, and I think it makes the regex more expressive.

`^(.+terraform-provider-[[:alnum:]]+)(/v[[:digit:]]+)?`

```suggestion
		submatches := newRegex.FindStringSubmatch(mod.Path)
		url = fmt.Sprintf("https://%s", submatches[1])
```

pkg/terraform/exec/plugins/plugin_versions_generate.go
```diff
@@ -0,0 +1,249 @@
...
+
+		// git clone if directory doesn't exist, else git pull
```
staebler, line 111: I don't quite follow why we need to get the repo and get the tags. Why is the version declared for the module not sufficient? Even if it is a psuedo version, it still starts with a semver.

staebler (commented):
This looks really good. I have some concerns about how we are going about determining the version for the providers, but the overall course of this is spectacular.

go.mod
```diff
@@ -4,27 +4,26 @@ go 1.14
...
 	github.com/c4milo/gotoolkit v0.0.0-20190525173301-67483a18c17a // indirect
 	github.com/clarketm/json v1.14.1
 	github.com/containers/image v3.0.2+incompatible
 	github.com/coreos/ignition/v2 v2.9.0
 	github.com/dmacvicar/terraform-provider-libvirt v0.6.4-0.20201216193629-2b60d7626ff8
 	github.com/fatih/color v1.10.0 // indirect
-	github.com/frankban/quicktest v1.7.2 // indirect
 	github.com/ghodss/yaml v1.0.1-0.20190212211648-25d852aebe32
+	github.com/go-git/go-git/v5 v5.2.0
+	github.com/go-logr/logr v0.3.0 // indirect
```
staebler, line 26: I am seeing this module as unused. It is removed when I run `go mod tidy`.

staebler (commented):
I also get a one-line change in `vendor/github.com/hashicorp/terraform-plugin-test/v2/helper.go` when I run `go mod vendor`.

pkg/terraform/exec/plugins/plugin_versions_generate.go
```diff
@@ -0,0 +1,249 @@
...
+`
+}
+
+// TODO: local plugins (azureprivatedns, vsphereprivate)
+
+// getTerraformPluginInfo finds terraform providers used in go.mod, does any
+// necessary replacements and saves into our plugin map.
+func getTerraformPluginInfo(file *modfile.File, plugins *map[string]*TFPluginInfo) {
+	tre := regexp.MustCompile(`^[[:alnum:]][[:alnum:]\.\-/]+terraform-provider`)
+	nre := regexp.MustCompile(`(terraform-provider-([[:alnum:]]+))`)
```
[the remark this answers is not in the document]
jhixson74: I have updated the code and simplified the regular expressions.

pkg/terraform/exec/plugins/plugin_versions_generate.go
```diff
@@ -0,0 +1,249 @@
...
+func Get{{ .Name }}Version() string {
+	return "{{ .Version }}"
+}
+
+        {{- end }}
+{{- end}}
+`
+}
+
+// TODO: local plugins (azureprivatedns, vsphereprivate)
```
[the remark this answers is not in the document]
jhixson74: Left over cruft, removed.

pkg/terraform/exec/plugins/plugin_versions_generate.go
```diff
@@ -0,0 +1,249 @@
...
+
+func getTemplate() string {
+	return `// Code generated by plugin_versions_generate.go; DO NOT EDIT.
+
+package {{ .Pkg }}
+
+{{ with .TFPluginInfo }}
+        {{- range . }}
+
+func Get{{ .Name }}Version() string {
```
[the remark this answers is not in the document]
jhixson74: Originally, these were constants. Then I needed the github path. I managed to work this without the github path (for now). To be future proof, my preference is to leave these as functions so we can return more information. However, if you insist, I'll gladly change to constants ;-)

pkg/terraform/exec/plugins/plugin_versions_generate.go
```diff
@@ -0,0 +1,249 @@
...
+
+// getTerraformPluginInfo finds terraform providers used in go.mod, does any
+// necessary replacements and saves into our plugin map.
+func getTerraformPluginInfo(file *modfile.File, plugins *map[string]*TFPluginInfo) {
+	tre := regexp.MustCompile(`^[[:alnum:]][[:alnum:]\.\-/]+terraform-provider`)
+	nre := regexp.MustCompile(`(terraform-provider-([[:alnum:]]+))`)
+
+	// Find all terraform providers
+	for _, r := range file.Require {
+		if tre.Match([]byte(r.Mod.Path)) {
```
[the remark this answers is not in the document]
jhixson74: I've updated to use MatchString for all cases, thanks for the heads up on this.

pkg/asset/cluster/cluster.go
jhixson74: This particular commit is only for CI debugging. When this is ready to be merged, this commit will be removed.

pkg/terraform/exec/plugins/aws.go
jhixson74: If we set the version to 1.0.0, terraform will bomb out stating that version is wrong.

pkg/terraform/exec/plugins/plugin_versions_generate.go
jhixson74: I've simplified all the regular expressions and commented on what they do.

pkg/terraform/exec/plugins/plugin_versions_generate.go
jhixson74: I am using the golang module parsing module, which is what golang uses to parse modules ;-) I am not a big fan of parsing output, since it tends to change over time and break things silently. My preference is to use the module, since it has a stable API.

pkg/terraform/exec/plugins/plugin_versions_generate.go
jhixson74: I have simplified all of the regular expressions and minimized them, I think we are good here now.

go.mod
jhixson74: When I run go mod vendor, that line fails for me. I am not sure why and I didn't dive into it too deeply, but I did commit a fix for it. I'll investigate more, but for now this works.

pkg/terraform/exec/plugins/plugin_versions_generate.go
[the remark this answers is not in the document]
staebler: No, by all means, if you have a reason why they are better as functions, then let's leave them as functions.

pkg/terraform/exec/plugins/aws.go
staebler: That's a wonderful reason for setting the real version. :)

pkg/terraform/exec/plugins/plugin_versions_generate.go
staebler: Sounds good. I was mostly concerned about doing the parsing in a bespoke way. But if we are using the same methods that are used under the covers by `go mod`, then I am satisfied.

pkg/terraform/exec/plugins/plugin_versions_generate.go
staebler: Following up on this, I do notice that the semver in the psuedo version is not the same as the semver that terraform expects. I assume this is because of the dastardly v2 missing in many of the provider modules.

vendor/github.com/hashicorp/terraform/command/meta_dependencies.go
staebler: Ah, I missed this. I've merged that and created the `v0.14.6-openshift-1` tag.

pkg/terraform/exec/plugins/plugin_versions_generate.go
```diff
@@ -0,0 +1,249 @@
...
+// necessary replacements and saves into our plugin map.
+func getTerraformPluginInfo(file *modfile.File, plugins *map[string]*TFPluginInfo) {
+
+	// terraform provider and name regex
+	re := regexp.MustCompile(`^[[:alnum:]][[:alnum:]\.\-/]+(terraform-provider-([[:alnum:]]+))`)
+
+	// Find all terraform providers
+	for _, r := range file.Require {
+		matches := re.FindStringSubmatch(r.Mod.Path)
+		if len(matches) == 3 {
```
staebler, line 62: I don't see a compelling reason to check for a length match. The result will either be nil, or all of the submatches will be populated. If the slice is not nil and the length does not match the expected length, then that is a coding error. We should not have code that explicitly and silently skips coding errors.
```suggestion
		if matches != nil {
```

pkg/terraform/terraform.go
```diff
@@ -126,8 +126,9 @@ func unpackAndInit(dir string, platform string) (err error) {
 
 	os.Setenv("TF_CLI_CONFIG_FILE", filepath.Join(dir, "terraform.rc"))
 
-	// XXX: This is only here for debugging CI
+	// XXX: These are only here for debugging CI
 	os.Setenv("TF_LOG", "trace")
+	os.Setenv("TERRAFORM_LOCK_FILE_PATH", dir)
```
staebler, line 131: Rather than setting this env variable, can we change directories into the temp directory where all of the terraform files are and then change back to the previous working directory when terraform is done? That would keep us from having to carry the patch supporting this environment variable. It would also mean that we don't need to set the `OverrideDataDir` field in the terraform command meta.

staebler (commented):
This is working pretty well for me in my tests.
My main concerns are with how we are getting the versions for the providers. I haven't run into a scenario yet where the "correct" version is actually needed. It is a lot of code to take on for determining the versions, and I am hesitant to carry that tech burden. I like that we are automating the generation of the versions if they are indeed needed, but I am concerned about the cost of that automation. In your experience before creating the automation, was it onerous to figure out the version to use when bumping a provider version?

pkg/terraform/exec/plugins/aws.go
staebler: And, in case the (unauthenticated) is of concern, even when we set the "correct" version, it is still unauthenticated.
```
time="2021-03-18T21:45:56Z" level=debug msg="- Installed openshift/local/aws v3.1.0 (unauthenticated)"
```

pkg/terraform/terraform.go
staebler: Here are the changes that I made to test changing the directory, if you are interested in doing it this way.
https://github.com/staebler/installer/commit/5917fc333385c82612e51cbc83908dc7001fe614

pkg/terraform/exec/plugins/plugin_versions_generate.go
```diff
@@ -0,0 +1,249 @@
...
+{{- end}}
+`
+}
+
+// getTerraformPluginInfo finds terraform providers used in go.mod, does any
+// necessary replacements and saves into our plugin map.
+func getTerraformPluginInfo(file *modfile.File, plugins *map[string]*TFPluginInfo) {
+
+	// terraform provider and name regex
+	re := regexp.MustCompile(`^[[:alnum:]][[:alnum:]\.\-/]+(terraform-provider-([[:alnum:]]+))`)
```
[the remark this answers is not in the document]
jhixson74: I've updated this to use your recommended regex. While it may not seem too obvious, my regular expression did indeed ensure the slash was there before terraform-provider- ;-) But I agree, yours is not as restrictive.

pkg/terraform/exec/plugins/plugin_versions_generate.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+
+		// Extract version out of tag
+		pi.Version = vre.FindString(pi.Tag)
+	}
+}
+
+// getLocalTerraformPluginVersions gathers the terraform provider version
+// numbers for local plugins
+func getLocalTerraformPluginVersions(file *modfile.File, plugins *map[string]*TFPluginInfo) {
```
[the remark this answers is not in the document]
jhixson74: I have updated these functions to take and return a map and not use a pointer. Just out of curiosity, why is this better form than how I did it? The functions were manipulating the plugins variable. If this was C, this would be good form ;-) I am still new to golang, so this has go me curious why it's not good to do this.

pkg/terraform/exec/plugins/plugin_versions_generate.go
```diff
@@ -0,0 +1,249 @@
...
+			here, err := os.Getwd()
+			if err != nil {
+				log.Fatal(err)
+			}
+			err = os.Chdir(tmpDir)
+			if err != nil {
+				log.Fatal(err)
+			}
+			// XXX: HACK: fork git process until go-git supports name-rev
+			out, err := exec.Command("git", "name-rev", "--tags", "--name-only", hash.String()).Output()
```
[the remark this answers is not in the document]
jhixson74: All that happens here is anything extra after the tag, such as ^2 or ~3 is chopped off. The tag is saved and the versions extracted from it.

pkg/terraform/exec/plugins/plugin_versions_generate.go
jhixson74: Thanks for the tip, and this is good to know. I've updated all instances where I was counting the regular expression matches.

pkg/terraform/exec/plugins/plugin_versions_generate.go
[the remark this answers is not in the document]
staebler: Yours **permitted** a slash but did not **enforce** a slash.
https://play.golang.org/p/-SUxddgH7gv

pkg/terraform/exec/plugins/plugin_versions_generate.go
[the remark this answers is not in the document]
staebler: My main reasoning is that it is uncommon to see dereferences of a map, so it makes me have to think about what is going on. There is likely little performance difference between the two, so I would prefer to go with the one that is easy to read and rationalize about. For me it is easier to rationalize about a read-only input parameter and an output parameter, similar to `append`.

One major difference between C and GO is that C has a deterministic memory model. In Go, we have to deal with garbage collection. When you pass around a pointer, you make the garbage collection have to follow the pointer through more places to determine whether the block of memory is still in use. That is one reason why it may you may get better performance from copying over passing by pointer. That same thinking would also lend itself to considering `TFPluginInfo` over `*TFPluginInfo` for the value of the map.

pkg/terraform/exec/plugins/plugin_versions_generate.go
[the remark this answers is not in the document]
staebler: Right. I get that. My question is more around how we are determining what the correct version is, when it could be a tag from an earlier commit or a later commit.

pkg/terraform/terraform.go
```diff
@@ -119,8 +125,13 @@ func unpackAndInit(dir string, platform string) (err error) {
 	defer lpDebug.Close()
 	defer lpError.Close()
 
+	os.Setenv("TF_CLI_CONFIG_FILE", filepath.Join(dir, "terraform.rc"))
+
+	// XXX: This is only here for debugging CI
+	os.Setenv("TF_LOG", "trace")
```
staebler, line 131: We lost setting the `TERRAFORM_LOCK_FILE_PATH` environment variable.

jhixson74: Doh!

staebler:
@jhixson74 Do you have an idea for how we can resolve the module incompatibilities that you are currently working around by manual editing `vendor/github.com/hashicorp/terraform-plugin-test/v2/helper.go`?
```
diff --git a/vendor/github.com/hashicorp/terraform-plugin-test/v2/helper.go b/vendor/github.com/hashicorp/terraform-plugin-test/v2/helper.go 
index 87a15b3..a2cc470 100644                
--- a/vendor/github.com/hashicorp/terraform-plugin-test/v2/helper.go 
+++ b/vendor/github.com/hashicorp/terraform-plugin-test/v2/helper.go 
@@ -145,7 +145,7 @@ func symlinkAuxiliaryProviders(pluginDir string) error { 
  		if filenameExt == ".zip" {                
 			_, err = os.Stat(path)                   
 			if os.IsNotExist(err) {                  
-				zipDecompressor.Decompress(path, filepath.Join(auxiliaryProviderDir, filename), false, 0000) 
+				zipDecompressor.Decompress(path, filepath.Join(auxiliaryProviderDir, filename), false) 
 			} else if err != nil {                   
 				return fmt.Errorf("Unexpected error: %s", err) 
 			}
```

staebler:
The bootstrap destroy for ovirt is deleting more than it should be.
```
time="2021-04-08T02:55:44Z" level=debug msg="module.masters.*****_vm.master[0]: Still destroying... [id=fe3c788b-b42f-41ce-b415-b397e843bca4, 10s elapsed]"
time="2021-04-08T02:55:44Z" level=debug msg="module.masters.*****_vm.master[2]: Still destroying... [id=53895b87-fc60-4a09-80c1-cbcdcccdd645, 10s elapsed]"
time="2021-04-08T02:55:44Z" level=debug msg="module.masters.*****_vm.master[1]: Still destroying... [id=6b3b37fc-3476-4e77-a9ac-34cd7bb9b7ba, 10s elapsed]"
```

staebler:
The gcp bootstrap destroy is deleting too much, too.
```
time="2021-04-08T02:57:16Z" level=debug msg="2021/04/08 02:57:16 [TRACE] Completed graph transform *terraform.TransitiveReductionTransformer with new graph:"
time="2021-04-08T02:57:16Z" level=debug msg="  meta.count-boundary (EachMode fixup) - *terraform.NodeCountBoundary"
time="2021-04-08T02:57:16Z" level=debug msg="    module.bootstrap.data.google_storage_object_signed_url.ignition_url (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.bootstrap.data.ignition_config.redirect (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.bootstrap.google_compute_address.bootstrap (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.bootstrap.google_compute_firewall.bootstrap_ingress_ssh (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.bootstrap.google_storage_bucket.ignition (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="  module.bootstrap.data.google_storage_object_signed_url.ignition_url (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.network.google_compute_region_backend_service.api_internal (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.network.google_compute_target_pool.api[0] (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="  module.bootstrap.data.ignition_config.redirect (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.network.google_compute_region_backend_service.api_internal (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.network.google_compute_target_pool.api[0] (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    provider[\"openshift/local/ignition\"] - *terraform.NodeApplyableProvider"
time="2021-04-08T02:57:16Z" level=debug msg="  module.bootstrap.google_compute_address.bootstrap (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.network.google_compute_region_backend_service.api_internal (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.network.google_compute_target_pool.api[0] (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="  module.bootstrap.google_compute_firewall.bootstrap_ingress_ssh (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    provider[\"openshift/local/google\"] - *terraform.NodeApplyableProvider"
time="2021-04-08T02:57:16Z" level=debug msg="  module.bootstrap.google_storage_bucket.ignition (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.bootstrap.google_storage_bucket_object.ignition (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.network.google_compute_region_backend_service.api_internal (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.network.google_compute_target_pool.api[0] (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="  module.bootstrap.google_storage_bucket_object.ignition (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    provider[\"openshift/local/google\"] - *terraform.NodeApplyableProvider"
time="2021-04-08T02:57:16Z" level=debug msg="  module.dns.google_dns_managed_zone.int (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.dns.google_dns_record_set.api_external_internal_zone (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.dns.google_dns_record_set.api_internal (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="  module.dns.google_dns_record_set.api_external[0] (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.dns.google_dns_managed_zone.int (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="  module.dns.google_dns_record_set.api_external_internal_zone (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    provider[\"openshift/local/google\"] - *terraform.NodeApplyableProvider"
time="2021-04-08T02:57:16Z" level=debug msg="  module.dns.google_dns_record_set.api_internal (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    provider[\"openshift/local/google\"] - *terraform.NodeApplyableProvider"
time="2021-04-08T02:57:16Z" level=debug msg="  module.network.google_compute_forwarding_rule.api[0] (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.dns.google_dns_record_set.api_external[0] (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="  module.network.google_compute_forwarding_rule.api_internal (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.dns.google_dns_record_set.api_external_internal_zone (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.dns.google_dns_record_set.api_internal (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="  module.network.google_compute_region_backend_service.api_internal (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.network.google_compute_forwarding_rule.api_internal (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="  module.network.google_compute_target_pool.api[0] (destroy) - *terraform.NodeDestroyResourceInstance"
time="2021-04-08T02:57:16Z" level=debug msg="    module.network.google_compute_forwarding_rule.api[0] (destroy) - *terraform.NodeDestroyResourceInstance"
```

staebler:
I think that the behavior of `terraform destroy -target` has changed.
```
The -target flag, instead of affecting "dependencies" will instead also destroy any resources that depend on the target(s) specified.
```
Taking the gcp bootstrap destroy as an example, targeting a destroy of the bootstrap module also causes the `module.network.google_compute_target_pool.api[0]` resource to be destroyed, since that resource depends on outputs from the bootstrap module. That then cascades to other resources, deleting much more than intended.

We should look into bringing the changes that @rna-afk has done in https://github.com/openshift/installer/pull/4822 into this PR.

go.mod
jhixson74: I forced terraform-plugin-test to v2.2.0 to fix this.

mtnbikenc:
/retitle "Bug 1981941: Upgrade terraform to 0.14.6"

mtnbikenc:
/retitle Bug 1981941: Upgrade terraform to 0.14.6

staebler:
Tests are passing to the extent expected.
Let's wait for https://github.com/openshift/installer/pull/5484 to merge so that we can get a good signal for ovirt and openstack.
/approve

PR #4732, opened 2021-03-11
WIP: Update AWS CloudFormation templates to closely match what an IPI installation creates

jaredhocutt:
Update the CloudFormation templates to create the infrastructure in a way that matches how the IPI install would create the infrastructure. By having these match by default, this gives the user a common starting place before they make any customizations that are required during their UPI install.

The bulk of these changes are related to ensuring that objects created are named and tagged in the way that OpenShift expects them to be. This will enable things such as MachineSets to work out of the box (and even at install time for the default set of worker nodes).

These changes also include a bit of formatting update to provide consistency across the template files and for readability.

### DIFF
--- a/upi/aws/cloudformation/01_vpc.yaml
+++ b/upi/aws/cloudformation/01_vpc.yaml
@@ -14,7 +14,8 @@ Parameters:
     Type: String
 
   VpcCidr:
-    AllowedPattern: ^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/(1[6-9]|2[0-4]))$
+    AllowedPattern: >-
+      ^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/(1[6-9]|2[0-4]))$
     ConstraintDescription: CIDR block parameter must be in the form x.x.x.x/16-24.
     Default: 10.0.0.0/16
     Description: CIDR block for VPC.
@@ -45,15 +46,15 @@ Parameters:
 Metadata:
   AWS::CloudFormation::Interface:
     ParameterGroups:
-      - Label:
-          default: "Network Configuration"
-        Parameters:
-          - VpcCidr
-          - SubnetBits
-      - Label:
-          default: "Availability Zones"
-        Parameters:
-          - AvailabilityZoneCount
+    - Label:
+        default: "Network Configuration"
+      Parameters:
+      - VpcCidr
+      - SubnetBits
+    - Label:
+        default: "Availability Zones"
+      Parameters:
+      - AvailabilityZoneCount
     ParameterLabels:
       AvailabilityZoneCount:
         default: "Availability Zone Count"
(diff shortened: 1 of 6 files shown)


jaredhocutt:
While this is still a work in progress, one of the tests that failed is the `ci/prow/e2e-aws-upi` test with the error:

```
An error occurred (ValidationError) when calling the CreateStack operation: Parameters: [InfrastructureName] must have values
```

This is understandable and expected as I have added a new parameter to `upi/aws/cloudformation/01_vpc.yaml` named `InfrastructureName`.

If I'm looking in the correct place, I think the update to provide the value for this parameter to this test would need to happen here:

https://github.com/openshift/release/blob/2454d8e56cec0e3f2f2a445bc6a35f9686cfa7e9/ci-operator/templates/openshift/installer/cluster-launch-installer-upi-e2e.yaml#L1344

This is my first time submitting a PR for this project, so I'm not sure what the correct way to handle this is.

staebler:
Thanks for the PR, @jaredhocutt. It is going to take some time to work through all of these changes, but I appreciate the effort that you have put into converging the infrastructure created in a UPI install with that created in an IPI install.

staebler:
> <snip>
> This is understandable and expected as I have added a new parameter to `upi/aws/cloudformation/01_vpc.yaml` named `InfrastructureName`.
> <snip>
> This is my first time submitting a PR for this project, so I'm not sure what the correct way to handle this is.

Yes, you will need to make a change in the release repo before this can merge. Depending upon how tolerant CloudFormations is to being passed in arguments that it does not know about, you may even have to do a few coordinated PRs between the installer and release repo to get each one happy about the new argument.

jaredhocutt:
The original work that I did on these templates were in the OpenShift 4.5 timeframe. So before starting this work, I compared them with changes that had been made in the meantime in this repo and ensured that any changes were pulled in. That's what exists in this PR as of now.

I have manually verified in my own environment that these CloudFormation templates are still correct and valid for OpenShift 4.7 (as it's the latest GA version at this moment). It did this by using the process outlined in the `ci/prow/e2e-aws-upi` test, which as mentioned above, will need to be updated to pass in an additional parameter to the CloudFormation template for the VPC creation.

Unfortunately as expected, I did verify that you cannot pass a parameter to a CloudFormation template that does not exist in the template. So as @staebler mentioned, this is going to require a bit of orchestration with the https://github.com/openshift/release repository. I have a few ideas on how to handle this, but if anyone has guidance on how this has been handled in the past, I'd love to match the precedence already set.

Have a great weekend!

jstuever:
UPI clusters are, by definition, not owned by the cluster. All the changes suggesting such need to be removed.

jstuever:
> UPI resources are, by definition, not owned by the cluster. All the changes suggesting such need to be removed.

We discussed this in stand-up today and we need to discuss this further to determine if UPI resources should be considered "owned" or not.

jaredhocutt:
> We discussed this in stand-up today and we need to discuss this further to determine if UPI resources should be considered "owned" or not.

If possible, I'd love to be part of this conversation to better understand what it actually means to be `owned` vs `shared`.

Additionally, I'd like to understand why a UPI install where you create a VPC specifically for the cluster wouldn't be considered `owned` vs the use case where an existing VPC is used that has other workloads in it, in which case it would not be owned.

afflom:
This is a huge deficiency with the current UPI CF templates. These missing tags just inhibited us from converting an internally published  LB cluster to a externally published LB. Once we manually added these tags to the subnets, we were able to easily convert to external publishing. 

This PR is crucial. Please merge.

PR #4733, opened 2021-03-11
Bug 1941759: Make failure to fetch cloud info non-fatal

mandre:
It is possible that under some circumstance we fail to gather the
necessary information to validate the underlying cloud meets the
requirements. Recently, this was a neutron bug [1] that caused it to
return an unexpected variable type.

Our inability to perform pre-flight validation shouldn't prevent the
installer from running.

This commit makes the installer more tolerant to failures to run
validations, for OpenStack platform only. Legitimate validation
failures still block the installation.

[1] https://bugs.launchpad.net/neutron/+bug/1918565

### DIFF
--- a/pkg/asset/quota/quota.go
+++ b/pkg/asset/quota/quota.go
@@ -126,6 +126,10 @@ func (a *PlatformQuotaCheck) Generate(dependencies asset.Parents) error {
 		if err != nil {
 			return errors.Wrap(err, "failed to get cloud info")
 		}
+		if ci == nil {
+			logrus.Warnf("Empty OpenStack cloud info and therefore will skip checking quota validation.")
+			return nil
+		}
 		reports, err := quota.Check(ci.Quotas, openstack.Constraints(ci, masters, workers, ic.Config.NetworkType))
 		if err != nil {
 			return summarizeFailingReport(reports)
--- a/pkg/asset/installconfig/openstack/validate.go
+++ b/pkg/asset/installconfig/openstack/validate.go
@@ -21,6 +21,10 @@ func Validate(ic *types.InstallConfig) error {
 	if err != nil {
 		return err
 	}
+	if ci == nil {
+		logrus.Warnf("Empty OpenStack cloud info and therefore will skip pre-flight validation.")
+		return nil
+	}
 
 	allErrs := field.ErrorList{}
 
--- a/pkg/asset/installconfig/openstack/validation/cloudinfo.go
+++ b/pkg/asset/installconfig/openstack/validation/cloudinfo.go
@@ -100,7 +100,8 @@ func GetCloudInfo(ic *types.InstallConfig) (*CloudInfo, error) {
 
 	err = ci.collectInfo(ic, opts)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to generate OpenStack cloud info")
+		logrus.Warnf("Failed to generate OpenStack cloud info: %v", err)
+		return nil, nil
 	}
 
 	return ci, nil


pierreprinetti:
In the prospect of landing [the Gophercloud workaround](https://github.com/gophercloud/gophercloud/pull/2126), do you think we should turn this patch into a more general short-circuit for when quota validation fail for whatever reason?

mandre:
> In the prospect of landing [the Gophercloud workaround](https://github.com/gophercloud/gophercloud/pull/2126), do you think we should turn this patch into a more general short-circuit for when quota validation fail for whatever reason?

Sure, do you want me to update the comment, or were you thinking of something else?

pierreprinetti:
I was actually thinking about turning the whole collectInfo error as a warning here:

https://github.com/openshift/installer/blob/78c0350c6293be0159fd857012189e289216a099/pkg/asset/installconfig/openstack/validation/cloudinfo.go#L103

What do you think?

pierreprinetti:
/lgtm
/hold

This patch LGTM as it correctly resolves the issue. I'll let you choose whether you want to change the error catching or to directly cancel the hold.

pierreprinetti:
/hold cancel
/lgtm

Looks good to me; I defer to the code owners for the approval. Thank you!

pkg/asset/quota/quota.go
```diff
@@ -126,6 +126,10 @@ func (a *PlatformQuotaCheck) Generate(dependencies asset.Parents) error {
 		if err != nil {
 			return errors.Wrap(err, "failed to get cloud info")
```
staebler, line 127: What is an example of an error that we do not want the installer to tolerate?

EmilienM: https://prow.ci.openshift.org/view/gcs/origin-ci-test/logs/release-openshift-ocp-installer-e2e-openstack-4.8/1369468189917843456

 ```level=fatal msg=failed to fetch Cluster: failed to fetch dependency of "Cluster": failed to generate asset "Platform Quota Check": failed to get cloud info: failed to generate OpenStack cloud info: failed to load Quota: failed to load quota limits: failed to get network quota records: failed to get QuotaSets from OpenStack Network API: json: cannot unmarshal string into Go struct field QuotaDetail.quota.router.reserved of type int ```

EmilienM:
It seems like https://github.com/gophercloud/gophercloud/pull/2126 is closed to land, I think we don't need this PR if we can quickly revendor gophercloud here.

EmilienM:
@mandre should we go with https://github.com/openshift/installer/pull/4741 ?

pkg/asset/quota/quota.go
staebler: Isn't that an example of an error that we **do** want the installer to tolerate? I am looking for whether there are indeed errors that can come from `GetCloudInfo` that we want to cause the installation to fail. For example, if there is an error creating one of the service clients, do we want to fail installation then?

mandre: Right, we may also want to tolerate errors creating clients in which case `GetCloudInfo()` won't error at all, however I'm not sure it's going to be that useful: if we fail to create a client in `GetCloudInfo()` we're likely get the same error later on during installation as it would indicate an issue with the underlying cloud. This is probably something we should error on the earliest.

mandre:
> @mandre should we go with #4741 ?

What happens on the next neutron or nova bug? We should go with both IMO.

EmilienM:
OK Ignore my previous comment, I now agree with you.
/lgtm

pierreprinetti:
/retitle Bug 1941759: Make failure to fetch cloud info non-fatal

mandre:
/assign @staebler 
Unless you have other comments on this PR I think it's ready to merge.

PR #4740, opened 2021-03-11
Bug 1934123: generate glance url considering the region

Fedosin:
Now we ignore region when we generate the glance url with ignition config. This commit updates it by considering the region name.

/label platform/openstack


mandre (commented):
/lgtm
/hold for tests

PR #4743, opened 2021-03-12
doc/openstack: add a note about rootVolume

EmilienM:
Add a note about rootVolume and the fact that we want the Glance image
to be in QCOW2 format and not RAW to avoid conversion issues.


mdbooth:
/hold until we confirm this bug in Cinder

This doesn't sound correct to me, so I suspect there may be more going on. I wouldn't be confident enough to document it.

EmilienM:
> /hold until we confirm this bug in Cinder
> 
> This doesn't sound correct to me, so I suspect there may be more going on. I wouldn't be confident enough to document it.

i agree. I'll do more investigation.

PR #4744, opened 2021-03-12
[Libvirt][doc] Add system-resolved section for dns overlay

praveenkumar:
With fedora 33 systemd-resolved is by default enabled and dns changes
should be done using `resolved` instead using networkmanager's dnsmasq
plugin. This patch add the document around it.


cfergeau:
Hmm this is a duplicate of #4564 which was put on hold.

PR #4748, opened 2021-03-12
azure: divide machine CIDR into 2 networks for subnets

staebler:
Instead of dividing the IPv4 machine CIDR into 8 networks and only using 2 of those 8, divide into only 2 networks. Likewise,
for the IPv6 machine CIDR divide into only 2 networks instead of 2^16 networks.

https://issues.redhat.com/browse/CORS-1659

### DIFF
--- a/data/data/azure/vnet/common.tf
+++ b/data/data/azure/vnet/common.tf
@@ -26,10 +26,10 @@ data "azurerm_virtual_network" "preexisting_virtual_network" {
 
 // Only reference data sources which are guaranteed to exist at any time (above) in this locals{} block
 locals {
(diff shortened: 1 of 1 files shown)


staebler:
Etienne, could you take a look at this for me?
/cc @e-tienne

e-tienne (commented):
Nit: can we update the ipv4 comment to match the Azure config? "from /16 to /17"

Also, I am not sure if this change is really needed for ipv6.

staebler:
> Nit: can we update the ipv4 comment to match the Azure config? "from /16 to /17"
> 
> Also, I am not sure if this change is really needed for ipv6.

Done. I agree that the IPv6 change is not needed as the subnets are large enough. We do not support IPv6 on Azure at the moment anyway, so I will leave it as is.

staebler:
These changes only affect azure installations.

PR #4742, opened 2021-03-11
bootstrap: record progress of services

staebler:
Each OpenShift service running on the bootstrap machine will now create a json file in /var/log/openshift/ that contains an array of entries detailing the progress that the service has made.

The entries included in the json file are the following.
* Service start
* Service end, with result and error details
* Service stage start
* Service stage end, with result and error details

The json files in /var/log/openshift will be collected by the bootstrap gather in /bootstrap/services/ for evaluation by the
installer for improved failure reporting to the user. The evaluation is left for follow-on work.

https://issues.redhat.com/browse/CORS-1542
https://issues.redhat.com/browse/CORS-1543

PR #4751, opened 2021-03-14
cmd: diagnose problems downloading release image

staebler:
When an installation fails during bootstrapping, analyze the bootstrap gather bundle for problems with downloading the release image. This analysis uses data from the new service entries files.

https://issues.redhat.com/browse/CORS-1533

Sample output when pull secret is not valid.
```
$ ./bin/openshift-install gather bootstrap
INFO Pulling debug logs from the bootstrap machine
INFO Bootstrap gather logs captured here "log-bundle-20210313180900.tar.gz"
ERROR The bootstrap machine failed to download the release image
INFO Error: unable to pull registry.ci.openshift.org/ocp/release:4.8.0-0.nightly-2021-03-12-142603: Error initializing source docker://registry.ci.openshift.org/ocp/release:4.8.0-0.nightly-2021-03-12-142603: Error reading manifest 4.8.0-0.nightly-2021-03-12-142603 in registry.ci.openshift.org/ocp/release: unauthorized: authentication required
INFO Pull failed. Retrying registry.ci.openshift.org/ocp/release:4.8.0-0.nightly-2021-03-12-142603...
INFO Error: unable to pull registry.ci.openshift.org/ocp/release:4.8.0-0.nightly-2021-03-12-142603: Error initializing source docker://registry.ci.openshift.org/ocp/release:4.8.0-0.nightly-2021-03-12-142603: Error reading manifest 4.8.0-0.nightly-2021-03-12-142603 in registry.ci.openshift.org/ocp/release: unauthorized: authentication required
```

Also, add an `analyze` command to analyze an existing bootstrap gather bundle.

### DIFF
--- a/cmd/openshift-install/analyze.go
+++ b/cmd/openshift-install/analyze.go
@@ -1,0 +1,61 @@
+package main
+
+import (
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/openshift/installer/pkg/gather/service"
+)
+
+var (
+	analyzeOpts struct {
+		gatherBundle string
+	}
+)
+
+func newAnalyzeCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "analyze",
+		Short: "Analyze debugging data for a given installation failure",
+		Long: `Analyze debugging data for a given installation failure.
+
+This command helps users to analyze the reasons for an installation that failed while bootstrapping.`,
+		Args: cobra.ExactArgs(0),
+		Run: func(_ *cobra.Command, _ []string) {
+			gatherBundle := analyzeOpts.gatherBundle
+			if gatherBundle == "" {
+				var err error
+				gatherBundle, err = getGatherBundleFromAssetsDirectory()
+				if err != nil {
+					logrus.Fatal(err)
+				}
+			}
+			if !filepath.IsAbs(gatherBundle) {
+				gatherBundle = filepath.Join(rootOpts.dir, gatherBundle)
+			}
+			if err := service.AnalyzeGatherBundle(gatherBundle); err != nil {
+				logrus.Fatal(err)
+			}
+		},
+	}
+	cmd.PersistentFlags().StringVar(&analyzeOpts.gatherBundle, "file", "", "Filename of the bootstrap gather bundle; either absolute or relative to the assets directory")
+	return cmd
+}
+
+func getGatherBundleFromAssetsDirectory() (string, error) {
+	matches, err := filepath.Glob(filepath.Join(rootOpts.dir, "log-bundle-*.tar.gz"))
+	if err != nil {
+		return "", errors.Wrap(err, "could not find gather bundles in assets directory")
+	}
+	switch len(matches) {
+	case 0:
+		return "", errors.New("no bootstrap gather bundles found in assets directory")
+	case 1:
+		return matches[0], nil
+	default:
+		return "", errors.New("multiple bootstrap gather bundles found in assets directory; select specific gather bundle by using the --file flag")
+	}
+}
--- a/pkg/gather/service/analyze.go
+++ b/pkg/gather/service/analyze.go
@@ -1,0 +1,133 @@
+package service
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"encoding/json"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+// AnalyzeGatherBundle will analyze the bootstrap gather bundle at the specified path.
+// Analysis will be logged.
+// Returns an error if there was a problem reading the bundle.
+func AnalyzeGatherBundle(bundlePath string) error {
+	// open the bundle file for reading
+	bundleFile, err := os.Open(bundlePath)
+	if err != nil {
+		return errors.Wrap(err, "could not open the gather bundle")
+	}
+	defer bundleFile.Close()
+	return analyzeGatherBundle(bundleFile)
+}
+
+func analyzeGatherBundle(bundleFile io.Reader) error {
+	// decompress the bundle
+	uncompressedStream, err := gzip.NewReader(bundleFile)
+	if err != nil {
+		return errors.Wrap(err, "could not decompress the gather bundle")
+	}
+	defer uncompressedStream.Close()
+
+	// read through the tar for relevant files
+	tarReader := tar.NewReader(uncompressedStream)
+	var releaseImageAnalysis *analysis
+	for {
+		header, err := tarReader.Next()
+		if err == io.EOF {
+			break
+		}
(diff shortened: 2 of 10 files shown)


staebler:
#4742 has merged
/hold cancel

cmd/openshift-install/analyze.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return "", errors.Wrap(err, "could not find gather bundles in assets directory")
+	}
+	switch len(matches) {
+	case 0:
+		return "", errors.New("no bootstrap gather bundles found in assets directory")
+	case 1:
+		return matches[0], nil
+	default:
+		return "", errors.New("multiple bootstrap gather bundles found in assets directory; select specific gather bundle by using the --fileq flag")
```
wking, line 59: Not actually clear to me how `analyzeOpts.gatherBundle` gets set, but `--fileq` feels like a typo.

staebler: It is a typo. Thanks.

staebler: Fixed with cb719764f...d2ea19b20 change.
For reference, the `analyzeOpts.gatherBundle` field gets set [here](https://github.com/openshift/installer/pull/4751/files#diff-5e2dedbe784e334cf9d4047f4cec6c1d4907cd67bc7f2d6f7dd8d8dcf43ae0aaR44).

pkg/gather/service/analyze.go
```diff
@@ -0,0 +1,133 @@
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return errors.Wrap(err, "encountered an error reading from the gather bundle")
+		}
+		if header.Typeflag != tar.TypeReg {
+			continue
+		}
+		filenameParts := strings.SplitN(header.Name, "/", 2)
```
patrickdillon, line 50: This seems more like a path than parts of a filename. 

Also a comment might help here about why we're splitting into 2. I think we're dropping the arbitrary log bundle filename.

pkg/gather/service/analyze.go
```diff
@@ -0,0 +1,133 @@
...
+	if releaseImageAnalysis != nil && releaseImageAnalysis.starts > 0 {
+		if !releaseImageAnalysis.successful {
+			logrus.Error("The bootstrap machine failed to download the release image")
+			for _, l := range strings.Split(releaseImageAnalysis.lastError, "\n") {
+				logrus.Info(l)
+			}
+		}
+	} else {
+		logrus.Error("The bootstrap machine did not execute the release-image.service systemd unit")
+	}
```
patrickdillon, line 75: Despite the comment and recognition that we are limiting the scope right now to analyzing only a single service, I think this is too coupled to release image analysis and it would be a good idea to establish the pattern for future analysis.

I suggest we create a map:
```go
var analysisFuncs map[string]func(io.Reader)
```
where the key is the filename and the value is a func that accepts the file, calls `analyzeService` and handles the result.

pkg/gather/service/analyze.go
```diff
@@ -0,0 +1,133 @@
...
+		if err := decoder.Decode(entry); err != nil {
+			return nil, errors.Wrap(err, "could not decode an entry in the service entries file")
+		}
+
+		// record a new start of the service
+		if entry.Phase == ServiceStart {
+			a.starts++
+		}
+
+		// the service is only considered considered successful if the very last entry is either the service ending
```
patrickdillon, line 117:
```suggestion
		// the service is only considered successful if the last entry is either the service ending
```

pkg/gather/service/analyze.go
wking: I kind of like this recommendation, but I'm also concerned about it biasing towards "generate many recommendations".  Sometimes generating many recommendations might be useful, but it also feels like most cases we want to try to boil this down to a single, most-important recommendation.  I dunno how to structure that, vs. the possibly-random tar ordering, so I'm fine punting to follow-up work.

pkg/gather/service/analyze.go
```diff
@@ -47,36 +53,51 @@ func analyzeGatherBundle(bundleFile io.Reader) error {
...
 		}
-		// we only care about the release-image.service for now. in the future, we will look at other services, too.
-		if filenameParts[1] == "bootstrap/services/release-image.json" {
-			var err error
-			releaseImageAnalysis, err = analyzeService(tarReader)
-			if err != nil {
-				logrus.Infof("Could not analyze the release-image.service: %v", err)
-			}
-			break
+		serviceName := serviceEntriesFileSubmatch[1]
```
patrickdillon, line 61: Why are we using the second match (or submatch--whatever the right term is)? 

Perhaps it is a safe assumption that there are at least two matches to this regex pattern but I think it would be a good idea to check the length as well when checking for nil above.

pkg/gather/service/analyze.go
```diff
@@ -47,36 +53,51 @@ func analyzeGatherBundle(bundleFile io.Reader) error {
...
+		name  string
+		check func(analysis) bool
+	}{
+		{name: "release-image", check: checkReleaseImageDownload},
+	}
+	for _, check := range analysisChecks {
+		a := serviceAnalyses[check.name]
+		if a.starts == 0 {
+			logrus.Errorf("The bootstrap machine did not execute the %s.service systemd unit", check.name)
+			break
```
patrickdillon, line 82: once we add more checks then we want to remove this `break`, right?

pkg/gather/service/analyze.go
```diff
@@ -47,36 +53,51 @@ func analyzeGatherBundle(bundleFile io.Reader) error {
...
+		}
+		if !check.check(a) {
+			break
```
patrickdillon, line 85: This `break` too. IIUC these seem like optimizations based on the limited number of checks, which is fine for the time being.

pkg/gather/service/analyze.go
staebler: The length of the submatches is determined by the regex. If the length is less than 2, that is a coding error not a runtime error. Throwing a panic is the right thing to do in the face of a coding error.

staebler: And the second submatch is the match of `([^.]+)` in the overall `^[^\/]+\/bootstrap\/services\/([^.]+)\.json$` expression.

pkg/gather/service/analyze.go
staebler: No. If the higher-priority service check determines that the service did not start, then we want to stop looking for errors in lower-priority service checks. For example, if we determine that the release-image service did not start, then we don't want to print errors that we find in the bootkube service, because any errors in the bootkube service can be traced back to the release-image service failing.

pkg/gather/service/analyze.go
staebler: Same response as above, where this break is what we want long-term.

pkg/gather/service/analyze.go
patrickdillon: I missed that the check priority was determined by the order in the slice. This makes sense now.

PR #3372, opened 2020-03-30
vSphere: Add IPI-specific validation. 

patrickdillon:
vSphere IPI requires fields in the install-config that are optional in UPI, specifically `cluster`, `network`, and the VIPs. In order to not break previous working IPI installs https://github.com/openshift/installer/commit/7e823377edd97177935f08e98fa3bb80a548591d performs this IPI-specific validation when creating tfvars. 

Jira: [CORS-1354](https://issues.redhat.com/browse/CORS-1354)
Related: openshift/release#7989

PR #4754, opened 2021-03-15
Bug 1918469: Check if VIP IPs overlap with machine CIDR provided during vsphere installation

rna-afk:
If the machineCIDR is specified by the user, the VIP IPs provided
must overlap with the CIDRs provided and the installation fails
if they do not.

Adding a validation check to see if the CIDR is provided and if so,
checks to see if the IPs are within any of the machine CIDRs provided.

### DIFF
--- a/pkg/types/vsphere/validation/platform.go
+++ b/pkg/types/vsphere/validation/platform.go
@@ -2,16 +2,18 @@ package validation
 
 import (
 	"fmt"
+	"net"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
+	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/vsphere"
 	"github.com/openshift/installer/pkg/validate"
 )
 
 // ValidatePlatform checks that the specified platform is valid.
-func ValidatePlatform(p *vsphere.Platform, fldPath *field.Path) field.ErrorList {
+func ValidatePlatform(p *vsphere.Platform, network *types.Networking, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 	if len(p.VCenter) == 0 {
 		allErrs = append(allErrs, field.Required(fldPath.Child("vCenter"), "must specify the name of the vCenter"))
@@ -37,7 +39,7 @@ func ValidatePlatform(p *vsphere.Platform, fldPath *field.Path) field.ErrorList 
 
 	// If all VIPs are empty, skip IP validation.  All VIPs are required to be defined together.
 	if strings.Join([]string{p.APIVIP, p.IngressVIP}, "") != "" {
-		allErrs = append(allErrs, validateVIPs(p, fldPath)...)
+		allErrs = append(allErrs, validateVIPs(p, network, fldPath)...)
 	}
 
 	// folder is optional, but if provided should pass validation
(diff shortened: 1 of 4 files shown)


pkg/types/vsphere/validation/platform.go
```diff
@@ -38,6 +40,24 @@ func ValidatePlatform(p *vsphere.Platform, fldPath *field.Path) field.ErrorList
 	// If all VIPs are empty, skip IP validation.  All VIPs are required to be defined together.
 	if strings.Join([]string{p.APIVIP, p.IngressVIP}, "") != "" {
 		allErrs = append(allErrs, validateVIPs(p, fldPath)...)
+		if network != nil {
```
staebler, line 43: Networking must never be nil. So rather than ignore a nil, at least add an error for it.

pkg/types/vsphere/validation/platform.go
```diff
@@ -38,6 +40,24 @@ func ValidatePlatform(p *vsphere.Platform, fldPath *field.Path) field.ErrorList
...
+			for _, machine := range network.MachineNetwork {
+				if machine.CIDR.Contains(net.ParseIP(p.APIVIP)) {
+					apiVIPOverlap = true
+				}
+				if machine.CIDR.Contains(net.ParseIP(p.IngressVIP)) {
+					ingressVIPOverlap = true
+				}
+			}
+			if !apiVIPOverlap {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("apiVIP"), p.APIVIP, "must overlap with the machine CIDRs provided"))
```
staebler, line 55: Overlap is not the right term. Overlap is used for sets. What you want to say is that the machine CIDR must contain the VIP.
```suggestion
				allErrs = append(allErrs, field.Invalid(fldPath.Child("apiVIP"), p.APIVIP, "must be contained within one of the machine networks"))
```

pkg/types/vsphere/validation/platform.go
```diff
@@ -38,6 +40,24 @@ func ValidatePlatform(p *vsphere.Platform, fldPath *field.Path) field.ErrorList
 	// If all VIPs are empty, skip IP validation.  All VIPs are required to be defined together.
 	if strings.Join([]string{p.APIVIP, p.IngressVIP}, "") != "" {
 		allErrs = append(allErrs, validateVIPs(p, fldPath)...)
+		if network != nil {
+			apiVIPOverlap := false
+			ingressVIPOverlap := false
+			for _, machine := range network.MachineNetwork {
+				if machine.CIDR.Contains(net.ParseIP(p.APIVIP)) {
```
staebler, line 47: Do all of this contain checking in `validateVIPs`. That function is already parsing the string into an IP, albeit within the call to `validate.IP`.

This makes it a bit more complicated for the call to `validateVIPs` from `ValidateForProvisioning`. But I would argue that `ValidateForProvisioning` should not be doing full validation of the VIPs. It should only be validating that the VIPs have been set. If the VIPs are set, then the full validation would have been done doing the initial validation of the install config. If the VIPs are not set, then `validateVIPs` will fail and will not have anything more to validate beyond that the VIPs are missing anyway. What do you think, @patrickdillon (https://github.com/openshift/installer/pull/3372)?

staebler (commented):
Update the PR title and the commit message to reflect that this is for vSphere.

pkg/types/vsphere/validation/platform.go
patrickdillon: > This makes it a bit more complicated for the call to `validateVIPs` from `ValidateForProvisioning`. But I would argue that `ValidateForProvisioning` should not be doing full validation of the VIPs. It should only be validating that the VIPs have been set. If the VIPs are set, then the full validation would have been done doing the initial validation of the install config. If the VIPs are not set, then `validateVIPs` will fail and will not have anything more to validate beyond that the VIPs are missing anyway. What do you think, @patrickdillon (#3372)?

I agree we should move this logic to `validateVIPs`. And if I understand correctly, I am fine to refactor `ValidateForProvisioning` so it only checks for the presence of the required VIPs fields and does not do the full validation.

To frame the discussion a bit more for @rna-afk vSphere UPI was released before IPI. When UPI was released VIPs were not required in the platform, but they are required for IPI. We cannot require VIPs in the platform as it would break backwards compatibility with UPI. Therefore we added the check here in `ValidateForProvisioning` to distinguish the validation only run for IPI. 

If I understand @staebler's concern correctly, we want to avoid a slippery slope of allowing validation creeping in here, where it does not really belong.

pkg/types/vsphere/validation/platform.go
```diff
@@ -38,6 +40,24 @@ func ValidatePlatform(p *vsphere.Platform, fldPath *field.Path) field.ErrorList
 	// If all VIPs are empty, skip IP validation.  All VIPs are required to be defined together.
 	if strings.Join([]string{p.APIVIP, p.IngressVIP}, "") != "" {
 		allErrs = append(allErrs, validateVIPs(p, fldPath)...)
+		if network != nil {
+			apiVIPOverlap := false
+			ingressVIPOverlap := false
+			for _, machine := range network.MachineNetwork {
```
patrickdillon, line 46: Performance is not going to be an issue here (MachineNetwork will be a very small array) so I would opt for readability rather than optimizing performance. I would suggest losing the boolean flags like `apiVIPOverlap` and creating a helper function that returns bool if the MachineNetwork contains the VIP

pkg/types/vsphere/validation/platform.go
```diff
@@ -60,12 +62,24 @@ func ValidateForProvisioning(p *vsphere.Platform, fldPath *field.Path) field.Err
 		allErrs = append(allErrs, field.Required(fldPath.Child("network"), "must specify the network"))
 	}
 
-	allErrs = append(allErrs, validateVIPs(p, fldPath)...)
+	if strings.Join([]string{p.APIVIP, p.IngressVIP}, "") != "" {
+		allErrs = append(allErrs, validateVIPs(p, network, fldPath)...)
+	}
```
staebler, line 67: This check **should not** be done here. In `ValidateForProvisioning`, the two VIPs are required. The suggestion that I was trying to make was to only validate that both VIPs were provided here rather than doing all of the other validation for the VIPs, since the other validation will necessarily have already been done.
```suggestion
	if p.APIVIP == "" {
		// add error
	}
	if p.IngressVIP == "" {
		// add error
	}
```

pkg/types/vsphere/validation/platform.go
```diff
@@ -60,30 +62,63 @@ func ValidateForProvisioning(p *vsphere.Platform, fldPath *field.Path) field.Err
...
+
+	if len(p.IngressVIP) == 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("ingressVIP"), "must specify a VIP for Ingress"))
+	}
 	return allErrs
 }
 
+// validateIPinCIDR checks if the given ip is within the machine CIDRs.
+func validateIPinCIDR(vipIP net.IP, machineNetwork []types.MachineNetworkEntry) bool {
```
patrickdillon, line 76: `validateIPinCIDR` -> `ipInNetwork`
nit: comment could be be slightly improved `ipInNetwork returns true if the given..."

pkg/types/vsphere/validation/platform.go
```diff
@@ -49,7 +51,7 @@ func ValidatePlatform(p *vsphere.Platform, fldPath *field.Path) field.ErrorList
 }
 
 // ValidateForProvisioning checks that the specified platform is valid.
-func ValidateForProvisioning(p *vsphere.Platform, fldPath *field.Path) field.ErrorList {
+func ValidateForProvisioning(p *vsphere.Platform, network *types.Networking, fldPath *field.Path) field.ErrorList {
```
staebler, line 54: `network` is unused now.
```suggestion
func ValidateForProvisioning(p *vsphere.Platform, fldPath *field.Path) field.ErrorList {
```

pkg/types/vsphere/validation/platform.go
```diff
@@ -60,30 +62,63 @@ func ValidateForProvisioning(p *vsphere.Platform, fldPath *field.Path) field.Err
...
+		allErrs = append(allErrs, field.Required(fldPath.Child("apiVIP"), "must specify a VIP for API"))
+	}
+
+	if len(p.IngressVIP) == 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("ingressVIP"), "must specify a VIP for Ingress"))
+	}
 	return allErrs
 }
 
+// ipInNetwork return true if the given ip is within the machine CIDRs.
```
staebler, line 75:
```suggestion
// ipInNetwork return true if the given ip is within one of the machine networks.
```

pkg/types/vsphere/validation/platform.go
```diff
@@ -60,30 +62,63 @@ func ValidateForProvisioning(p *vsphere.Platform, fldPath *field.Path) field.Err
...
+		ip := net.ParseIP(p.APIVIP)
+		if ip != nil {
+			if !ipInNetwork(ip, network.MachineNetwork) {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("apiVIP"), p.APIVIP, "must be contained within one of the machine networks"))
+			}
+		} else {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("apiVIP"), p.APIVIP, "not a valid IP address"))
+		}
+	} else {
+		allErrs = append(allErrs, field.Required(fldPath.Child("apiVIP"), "must specify a VIP for API"))
```
staebler, line 103:
```suggestion
		allErrs = append(allErrs, field.Required(fldPath.Child("apiVIP"), "must specify both API and Ingress VIPs when specifying either"))
```

pkg/types/vsphere/validation/platform.go
```diff
@@ -60,30 +62,63 @@ func ValidateForProvisioning(p *vsphere.Platform, fldPath *field.Path) field.Err
...
+		ip := net.ParseIP(p.IngressVIP)
+		if ip != nil {
+			if !ipInNetwork(ip, network.MachineNetwork) {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("ingressVIP"), p.IngressVIP, "must be contained within one of the machine networks"))
+			}
+		} else {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("ingressVIP"), p.IngressVIP, "not a valid IP address"))
+		}
+	} else {
 		allErrs = append(allErrs, field.Required(fldPath.Child("ingressVIP"), "must specify a VIP for Ingress"))
```
staebler, line 116:
```suggestion
		allErrs = append(allErrs, field.Required(fldPath.Child("ingressVIP"),  "must specify both API and Ingress VIPs when specifying either"))
```

staebler (commented):
Sorry to keep piling on with the nits.

staebler (commented):
Looks good. Just need to make the unit test expectations match the reality of the recent changes.

staebler:
vsphere e2e jobs passed.
/hold cancel

staebler:
These changes are only for vSphere.

PR #4757, opened 2021-03-16
Support configurable libvirt disk size

marun:
As per https://github.com/hashicorp/terraform/issues/3287 terraform requires that libvirt_master_size be specified as a number denominated in bytes rather than the canonical bytes (e.g. 1024M) returned by kube's Quantity.String() method. This change ensures that overriding the libvirt master volumeSize in generated install manifests will result in valid terraform configuration. This allows a libvirt SNO cluster to run more than just the control plane (i.e. making it more useful for developers).

/cc @romfreiman @praveenkumar @staebler

### DIFF
--- a/pkg/tfvars/libvirt/libvirt.go
+++ b/pkg/tfvars/libvirt/libvirt.go
@@ -83,7 +83,14 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
 	}
 
 	if sources.MasterConfig.Volume.VolumeSize != nil {
-		cfg.MasterDiskSize = sources.MasterConfig.Volume.VolumeSize.String()
+		// As per https://github.com/hashicorp/terraform/issues/3287 the
+		// master disk size needs to be specified as the number of bytes.
+		diskSizeInBytes, converted := sources.MasterConfig.Volume.VolumeSize.AsInt64()
+		if !converted {
+			msgTemplate := "failed to convert master disk size to bytes: %s"
+			return nil, fmt.Errorf(msgTemplate, sources.MasterConfig.Volume.VolumeSize)
+		}
+		cfg.MasterDiskSize = fmt.Sprintf("%d", diskSizeInBytes)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/tfvars/libvirt/libvirt.go
```diff
@@ -83,7 +83,14 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
 	}
 
 	if sources.MasterConfig.Volume.VolumeSize != nil {
-		cfg.MasterDiskSize = sources.MasterConfig.Volume.VolumeSize.String()
+		// As per https://github.com/hashicorp/terraform/issues/3287 the
+		// master disk size needs to be specified as the number of bytes.
+		diskSizeInBytes, converted := sources.MasterConfig.Volume.VolumeSize.AsInt64()
```
praveenkumar, line 88: So this is just a validation to make sure if user try to change master disk size it should be number of bytes instead of using units for measures, We are using this in snc but always used bytes https://github.com/code-ready/snc/blob/master/snc.sh#L139 so validation should be good 👍🏼

cfergeau:
Makes sense,
/lgtm
/approve

pkg/tfvars/libvirt/libvirt.go
staebler: This is not just validation, right? This is conversion to a form that terraform can handle. The user can still use, for example, 1024M in the Machine and this code will convert that value to 1048576 (or whatever the converted value is) for consumption by terraform.

marun: Correct, it's ensuring that bytes are output for terraform since that's all it understands. In the generated manifests it's still possible to specify whatever is supported by kube's Quantity type.

praveenkumar: Ah my bad, I thought we are trying to first convert whatever user provide and then use this as validation but this is really good to see it support quantity type.

PR #4760, opened 2021-03-16
  Inject CoreOS stream metadata as configmap via CVO manifest

cgwalters:
Split out from #4582

This copies the bits from https://github.com/cgwalters/rhel-coreos-bootimages
which builds a ConfigMap out of the stream metadata and injects
it into the cluster.

We have an `installer` image in the release image today; this adds
the "is an operator" label, even though it's not really an
operator.  We just want the CVO to inject the manifest.

Among other important semantics, this will ensure that in-place
cluster upgrades that have new pinned CoreOS stream data will
have this configmap updated.

### DIFF
--- a/hack/build-coreos-manifest.go
+++ b/hack/build-coreos-manifest.go
@@ -1,0 +1,73 @@
+// We want to keep the stream metadata stored "directly"
+// in git so it's easy to read and validate.  This build
+// script is invoked as part of the container build to
+// inject the data into a ConfigMap that will be installed
+// via CVO manifests into the target cluster, and maintained
+// across upgrades.
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	streamJSON = "data/data/rhcos-stream.json"
+	dest       = "bin/manifests/coreos-bootimages.json"
+)
+
+func run() error {
+	bootimages, err := ioutil.ReadFile(streamJSON)
+	if err != nil {
+		return err
+	}
+
+	cm := &corev1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: corev1.SchemeGroupVersion.String(),
+			Kind:       "ConfigMap",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "openshift-machine-config-operator",
+			Name:      "coreos-bootimages",
+			Annotations: map[string]string{
+				"include.release.openshift.io/ibm-cloud-managed":              "true",
+				"include.release.openshift.io/self-managed-high-availability": "true",
+				"include.release.openshift.io/single-node-developer":          "true",
+			},
+		},
+		Data: map[string]string{
+			"releaseVersion": "0.0.1-snapshot",
+			"stream":         string(bootimages),
+		},
+	}
+
+	b, err := json.Marshal(cm)
+	if err != nil {
+		return err
+	}
+
+	if err := os.MkdirAll(filepath.Dir(dest), 0755); err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(dest, b, 0644)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func main() {
+	if err := run(); err != nil {
+		fmt.Fprintf(os.Stderr, "%s\n", err)
+		os.Exit(1)
+	}
+}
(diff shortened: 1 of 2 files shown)


cgwalters:
OK so I believe that basically what this commit needs to do is what's in https://github.com/cgwalters/rhel-coreos-bootimages - basically.  This would be the first time that the `installer` image in the release image is installing manifests.  Does that make sense to everyone?

staebler:
> OK so I believe that basically what this commit needs to do is what's in https://github.com/cgwalters/rhel-coreos-bootimages - basically. This would be the first time that the `installer` image in the release image is installing manifests. Does that make sense to everyone?

My understanding is that we need the ConfigMap added under the manifests directory in the installer image that is included in the release.

The installer image may also need the `io.openshift.release.operator=true` label for CVO to consider its manifests.

staebler:
> Per discussion, the real fix here will probably require adding minimal CVO bits to the installer image.

Did you need the temporary fix in order to unblock something? Or can we do the real fix from the start?

cgwalters:
> Did you need the temporary fix in order to unblock something? Or can we do the real fix from the start?

I think this latest code will work - it was pretty easy to copy over the approach from https://github.com/cgwalters/rhel-coreos-bootimages .

If it does work we'll see it in the CI artifacts.

cgwalters:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_installer/4760/pull-ci-openshift-installer-master-e2e-aws/1371905428375474176/artifacts/e2e-aws/gather-extra/artifacts/oc_cmds/configmaps has 
`openshift-machine-config-operator                  coreos-bootimages                                                 2      64m`

:tada:

manifests/coreos-bootimages.json
```diff
@@ -0,0 +1,16 @@
+{
+    "apiVersion": "v1",
+    "kind": "ConfigMap",
+    "metadata": {
+        "name": "coreos-bootimages",
+        "namespace": "openshift-machine-config-operator",
+        "annotations": {
+            "include.release.openshift.io/ibm-cloud-managed": "true",
+            "include.release.openshift.io/self-managed-high-availability": "true",
+            "include.release.openshift.io/single-node-developer": "true"
```
staebler, line 10: Would you happen to have a reference for what these annotations signify?

manifests/coreos-bootimages.json
```diff
@@ -0,0 +1,16 @@
...
+        }
+    },
+    "data": {
+        "releaseVersion": "0.0.1-snapshot"
```
staebler, line 14: Is this something that is automatically updated by the CVO or by the release tooling?

hack/build-coreos-manifest.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+const (
+	configmapBase = "manifests/coreos-bootimages.json"
+	streamJSON    = "data/data/rhcos-stream.json"
+	dest          = "bin/manifests/coreos-bootimages.json"
+)
+
+func run() error {
+	var cm map[string]interface{}
```
staebler, line 19: Is there some reason why you don't want to have the type here be corev1.ConfigMap? If it were, then we would get type safety later when filling out the `data` and not have to do any casting.

staebler (commented):
Looks great. I just have a couple questions for things that are not clear to me, since operator manifests are a new experience for me.

manifests/coreos-bootimages.json
cgwalters: See
https://github.com/openshift/enhancements/blob/master/enhancements/update/cluster-profiles.md

Admittedly I cargo culted these from the MCO; we want this data in all of them I think.  It's not directly relevant for e.g. `single-node-developer` but it also doesn't do anything and it's tiny.

manifests/coreos-bootimages.json
cgwalters: It's replaced by `oc adm release new` yep.

hack/build-coreos-manifest.go
cgwalters: In the original repo I wasn't vendoring anything, it was basically just a git repo with a JSON file, so adding all the kube stuff to `vendor/` felt like overkill.

But here, we're already vendoring, so indeed - fixed!

cgwalters:
Any further thoughts on this?

cgwalters:
https://github.com/openshift/origin/pull/25993 is a quick sanity test for this that can merge after this does.

staebler:
> Any further thoughts on this?

Sorry, I lost track of this. I'll take another look at this PR and the dependent PR early next week. From the last time I looked at it, it looked generally good, save for maybe a few nits.

cgwalters:
Rebased :surfer:

images/installer/Dockerfile.ci
```diff
@@ -5,13 +5,17 @@ FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.8 AS b
 WORKDIR /go/src/github.com/openshift/installer
 COPY . .
 RUN hack/build.sh
+RUN go run -mod=vendor hack/build-coreos-manifest.go
```
staebler, line 8: Is `-mod=vendor` necessary here since we are using go 1.15?

staebler (approved):
Small nit that we can leave.
/lgtm

PR #4763, opened 2021-03-17
aws: ensure default resource created for vpcs are tagged

staebler:
When an AWS VPC is created, a default network ACL, route table, and security group are created for the VPC. The installer
has not been tagging those resources. This commit adds to code in the terraform to add all of the tags to those three resources after the VPC is created.

Since the network acl is now being tagged, the AWS destroyer needs to know how to handle network acls. This commit adds
code to delete network acls, skipping the default one since it cannot be deleted. Ultimately, there are no additional deletions performed by the destroyer as the network acl that is newly tagged is the default for the VPC and so is skipped.

https://issues.redhat.com/browse/CORS-1657

### DIFF
--- a/data/data/aws/vpc/vpc.tf
+++ b/data/data/aws/vpc/vpc.tf
@@ -46,3 +46,27 @@ resource "aws_vpc_dhcp_options_association" "main" {
   vpc_id          = data.aws_vpc.cluster_vpc.id
   dhcp_options_id = aws_vpc_dhcp_options.main[0].id
 }
+
+resource "aws_ec2_tag" "new_vpc_default_network_acl" {
+  for_each = var.vpc == null ? var.tags : {}
+
+  resource_id = aws_vpc.new_vpc[0].default_network_acl_id
+  key         = each.key
+  value       = each.value
+}
+
+resource "aws_ec2_tag" "new_vpc_default_route_table" {
+  for_each = var.vpc == null ? var.tags : {}
+
+  resource_id = aws_vpc.new_vpc[0].default_route_table_id
+  key         = each.key
+  value       = each.value
+}
+
+resource "aws_ec2_tag" "new_vpc_default_security_group" {
+  for_each = var.vpc == null ? var.tags : {}
+
+  resource_id = aws_vpc.new_vpc[0].default_security_group_id
+  key         = each.key
+  value       = each.value
+}
--- a/pkg/destroy/aws/aws.go
+++ b/pkg/destroy/aws/aws.go
@@ -753,6 +753,8 @@ func deleteEC2(ctx context.Context, session *session.Session, arn arn.ARN, logge
 		return deleteEC2SecurityGroup(ctx, client, id, logger)
 	case "snapshot":
 		return deleteEC2Snapshot(ctx, client, id, logger)
+	case "network-acl":
+		return deleteEC2NetworkACL(ctx, client, id, logger)
 	case "network-interface":
 		return deleteEC2NetworkInterface(ctx, client, id, logger)
 	case "subnet":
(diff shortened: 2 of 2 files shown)


staebler:
Add a user tag of "test-key:test-value".
```
$ aws ec2 describe-network-acls --query "NetworkAcls[*].Tags" --filter Name=vpc-id,Values=vpc-0fd57ce9fc695ebd5 Name=default,Values=true
test-key	test-value
kubernetes.io/cluster/mstaeble-994j6	owned

$ aws ec2 describe-route-tables --query "RouteTables[*].Tags" --filter Name=vpc-id,Values=vpc-0fd57ce9fc695ebd5 Name=association.main,Values=true
Name	mstaeble-994j6-public
test-key	test-value
kubernetes.io/cluster/mstaeble-994j6	owned

$ aws ec2 describe-security-groups --query "SecurityGroups[*].Tags" --filter Name=vpc-id,Values=vpc-0fd57ce9fc695ebd5 Name=group-name,Values=default
kubernetes.io/cluster/mstaeble-994j6	owned
test-key	test-value
```

patrickdillon:
/lgtm

nit: github complains about no newline

staebler:
/lgtm cancel

There is a problem with the destroyer not being able to delete the network acl, which is now tagged. The destroyer does complete, but there is an warning.

```
WARNING unrecognized EC2 resource type network-acl    arn=arn:aws:ec2:us-east-1:269733383066:network-acl/acl-0189235b4715fd7a9
```

staebler:
9cbd79daf...077d4639c add code in aws destroyer to handle deleting network acls

abhinavdahiya:
> When an AWS VPC is created, a default network ACL, route table, and security group are created for the VPC.

Are these resources created by the installer or AWS internally for the VPC?

Also can i delete a VPC that only has these default resources in it?

personally, if AWS is creating those resources, we shouldn't be tagging them with owned tag.

staebler:
> > When an AWS VPC is created, a default network ACL, route table, and security group are created for the VPC.
> 
> Are these resources created by the installer or AWS internally for the VPC?

The latter

> 
> Also can i delete a VPC that only has these default resources in it?

Yes

> 
> personally, if AWS is creating those resources, we shouldn't be tagging them with owned tag.

That is a fair point. What is your opinion on adding the user-provided tags to the resources? I also think that we have a different understanding of what the owned tag imputes to the resource. My understanding, derived from observation, is that the tag indicates that a resource's lifetime should be tied to the lifetime of the cluster. With that understanding, it is reasonable to tag the resources that AWS is indirectly creating.

abhinavdahiya:
> > > When an AWS VPC is created, a default network ACL, route table, and security group are created for the VPC.
> > 
> > 
> > Are these resources created by the installer or AWS internally for the VPC?
> 
> The latter
> 
> > Also can i delete a VPC that only has these default resources in it?
> 
> Yes
> 
> > personally, if AWS is creating those resources, we shouldn't be tagging them with owned tag.
> 
> That is a fair point. What is your opinion on adding the user-provided tags to the resources? 

for me use-provided tags have an expectation that if you create/manage a resource, add these additional tags. if we don't add owned tag to that resource, we shouldn't add the user-tag to that resource.

> I also think that we have a different understanding of what the owned tag imputes to the resource. My understanding, derived from observation, is that the tag indicates that a resource's lifetime should be tied to the lifetime of the cluster. With that understanding, it is reasonable to tag the resources that AWS is indirectly creating.

For me owned tag is only for resources we touch i.e create/manage . all transitive resources that are created by AWS or any other service for our owned resources should stay untouched. This makes it clear that we are not managing that resource and the management burden is on someone else here AWS.

an example of this is network-interfaces created for LB we create. we tag and mange the LB, we don't manage the network-interfaces, those are managed by AWS.

staebler:
/close
Closing this as we do not want to tag resources that AWS itself is creating.

PR #4766, opened 2021-03-17
Fix ipv6/dual-stack validation with Azure

danwinship:
The installer considers Azure to be a supported platform for IPv6 and dual-stack because of our early plans for testing, but single-stack IPv6 on Azure never worked, and dual-stack on Azure has never been tested and is not supported in 4.8.

### DIFF
--- a/data/data/azure/main.tf
+++ b/data/data/azure/main.tf
@@ -45,9 +45,8 @@ module "bootstrap" {
   private                = module.vnet.private
   outbound_udr           = var.azure_outbound_user_defined_routing
 
-  use_ipv4                  = var.use_ipv4 || var.azure_emulate_single_stack_ipv6
-  use_ipv6                  = var.use_ipv6
-  emulate_single_stack_ipv6 = var.azure_emulate_single_stack_ipv6
+  use_ipv4 = var.use_ipv4
+  use_ipv6 = var.use_ipv6
 }
 
 module "vnet" {
@@ -67,9 +66,8 @@ module "vnet" {
   private                     = var.azure_private
   outbound_udr                = var.azure_outbound_user_defined_routing
 
-  use_ipv4                  = var.use_ipv4 || var.azure_emulate_single_stack_ipv6
-  use_ipv6                  = var.use_ipv6
-  emulate_single_stack_ipv6 = var.azure_emulate_single_stack_ipv6
+  use_ipv4 = var.use_ipv4
+  use_ipv6 = var.use_ipv6
 }
 
 module "master" {
@@ -94,9 +92,8 @@ module "master" {
   private                = module.vnet.private
   outbound_udr           = var.azure_outbound_user_defined_routing
 
-  use_ipv4                  = var.use_ipv4 || var.azure_emulate_single_stack_ipv6
-  use_ipv6                  = var.use_ipv6
-  emulate_single_stack_ipv6 = var.azure_emulate_single_stack_ipv6
+  use_ipv4 = var.use_ipv4
+  use_ipv6 = var.use_ipv6
 }
 
 module "dns" {
@@ -113,9 +110,8 @@ module "dns" {
   base_domain_resource_group_name = var.azure_base_domain_resource_group_name
   private                         = module.vnet.private
 
-  use_ipv4                  = var.use_ipv4 || var.azure_emulate_single_stack_ipv6
-  use_ipv6                  = var.use_ipv6
-  emulate_single_stack_ipv6 = var.azure_emulate_single_stack_ipv6
+  use_ipv4 = var.use_ipv4
+  use_ipv6 = var.use_ipv6
 }
 
 resource "random_string" "storage_suffix" {
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"net"
 	"net/url"
-	"os"
 	"sort"
 	"strings"
 
@@ -221,12 +220,6 @@ func validateNetworkingIPVersion(n *types.Networking, p *types.Platform) field.E
 		switch {
 		case p.BareMetal != nil:
 		case p.None != nil:
-
-		case p.Azure != nil && os.Getenv("OPENSHIFT_INSTALL_AZURE_EMULATE_SINGLESTACK_IPV6") == "true":
-			if !presence["machineNetwork"].IPv4 || !presence["machineNetwork"].IPv6 {
-				allErrs = append(allErrs, field.Invalid(field.NewPath("networking"), "IPv6", "OPENSHIFT_INSTALL_AZURE_EMULATE_SINGLESTACK_IPV6 requires both IPv4 and IPv6 machineNetwork values"))
-			}
-
 		default:
 			allErrs = append(allErrs, field.Invalid(field.NewPath("networking"), "IPv6", "single-stack IPv6 is not supported for this platform"))
 		}
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -4,7 +4,9 @@ import (
 	"fmt"
 	"net"
 	"net/url"
+	"os"
 	"sort"
+	"strconv"
 	"strings"
 
 	dockerref "github.com/containers/image/docker/reference"
@@ -194,8 +196,10 @@ func validateNetworkingIPVersion(n *types.Networking, p *types.Platform) field.E
 			allErrs = append(allErrs, field.Invalid(field.NewPath("networking", "serviceNetwork"), strings.Join(ipnetworksToStrings(n.ServiceNetwork), ", "), "when installing dual-stack IPv4/IPv6 you must provide two service networks, one for each IP address type"))
 		}
 
+		experimentalDualStackEnabled, _ := strconv.ParseBool(os.Getenv("OPENSHIFT_INSTALL_EXPERIMENTAL_DUAL_STACK"))
 		switch {
-		case p.Azure != nil:
+		case p.Azure != nil && experimentalDualStackEnabled:
+			logrus.Warnf("Using experimental Azure dual-stack support")
 		case p.BareMetal != nil:
 		case p.None != nil:
 		default:
--- a/data/data/azure/bootstrap/variables.tf
+++ b/data/data/azure/bootstrap/variables.tf
@@ -89,11 +89,6 @@ variable "use_ipv6" {
   description = "This value determines if this is cluster should use IPv6 networking."
 }
 
-variable "emulate_single_stack_ipv6" {
-  type        = bool
-  description = "This determines whether a dual-stack cluster is configured to emulate single-stack IPv6."
-}
-
 variable "outbound_udr" {
   type    = bool
   default = false
--- a/data/data/azure/dns/dns.tf
+++ b/data/data/azure/dns/dns.tf
@@ -18,8 +18,6 @@ resource "azureprivatedns_zone_virtual_network_link" "network" {
 }
 
 resource "azureprivatedns_a_record" "apiint_internal" {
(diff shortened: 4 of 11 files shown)


pkg/types/validation/installconfig.go
```diff
@@ -195,13 +195,15 @@ func validateNetworkingIPVersion(n *types.Networking, p *types.Platform) field.E
 			allErrs = append(allErrs, field.Invalid(field.NewPath("networking", "serviceNetwork"), strings.Join(ipnetworksToStrings(n.ServiceNetwork), ", "), "when installing dual-stack IPv4/IPv6 you must provide two service networks, one for each IP address type"))
 		}
 
-		switch {
-		case p.Azure != nil:
-		case p.BareMetal != nil:
-		case p.None != nil:
-		default:
-			allErrs = append(allErrs, field.Invalid(field.NewPath("networking"), "DualStack", "dual-stack IPv4/IPv6 is not supported for this platform, specify only one type of address"))
+		if os.Getenv("OPENSHIFT_INSTALL_SKIP_IP_FAMILY_PLATFORM_CHECK") != "true" {
```
staebler, line 198: There is a high bar for adding supported environment variables. Unless there is an expectation that this will be used in CI, the few people that would make use of this can build their own installer.

aojea: it will be used in the CI, the goal is to have an environment that is able to run dual-stack e2e conformance tests, and Azure seems a good candidate, they are already running e2e test upstream in dual-stack https://testgrid.k8s.io/sig-network-dualstack-azure-e2e#dualstack-azure-e2e-master-iptables so we know that that platform supports it.

As a funny note, it removes `OPENSHIFT_INSTALL_AZURE_EMULATE_SINGLESTACK_IPV6` :)

staebler: This doesn't make sense to me. Either we are supporting dual stack or not. How are we going to run conformance tests for dual stack using a configuration that we do not support? If we *can* pass conformance test, why are we not supporting it?

aojea: well, to be honest, I just want to be able to use the installer to deploy a dual-stack cluster, if I don't have to use an environment variable to override the validation, much better :smile: , right now the intsaller works for me, it is able to deploy a dual stack cluster in azure, there are some bugs in other components and maybe we'll find some minor things, but overall it works.
I let others decide on the best way to implement it and discuss the "supportability" topic

danwinship: @staebler We officially support dual-stack on bare metal for customers. We would like to allow-but-not-"support" dual-stack on Azure for developers because it's simpler/cheaper than bare metal and should hopefully mostly work. We do not want to officially support dual-stack for customers on Azure in 4.8 because we haven't really put much thought or testing into it yet. (We might officially support it in 4.9.)

My worry is that if we don't block it in the installer, some customer will see "OCP 4.8 supports dual-stack on bare metal" and "OCP 4.8 is based on kubernetes 1.21" and "kubernetes 1.21 supports dual-stack on Azure", and conclude "Therefore OCP 4.8 supports dual-stack on Azure", and then try to install it and have it successfully install but then turn out to be subtly broken later. So the environment variable is to make it so developers/CI can install dual-stack on Azure in 4.8, but customers can't (without at least making it really really obvious that they're doing something totally unsupported).

I made the environment variable platform-agnostic because we may eventually think about dual-stack on AWS/vSphere/OpenStack, etc, too. Although now that I think about it more, if we wanted to do that, we'd probably need code/template changes to the installer as well, so the people testing that out would have to build their own installer, so maybe the environment variable doesn't really help? (Though, again, we might eventually end up in a situation where, eg OCP 4.9 has partially-working vSphere dual-stack support, so we want to allow developers to hack on it, but not let customers install it...)

So... I could leave it as is (with `OPENSHIFT_INSTALL_SKIP_IP_FAMILY_PLATFORM_CHECK`) or change that to be Azure-specific (`OPENSHIFT_INSTALL_ALLOW_DUAL_STACK_AZURE`?) and then if we eventually officially support dual-stack-on-Azure then that environment variable could just go away, but also, we might add ones for other platforms instead (`OPENSHIFT_INSTALL_ALLOW_DUAL_STACK_VSPHERE`, etc).

Or we could allow anyone to bring up a dual-stack Azure cluster, and just rely on customers reading the docs to know they shouldn't do that (hahahaha).

staebler: I am 100% in favor of blocking the user from using dual-stack in Azure if we are not supporting it. My quibble is only with the environment variable. We either enable a user to do dual-stack or we do not. There is no disabled unless you find this environment variable mode.

danwinship: So you agree we should block customers from using dual-stack in Azure... and you say we can't have an environment variable... so are you saying then that there is no way we can allow developers to enable dual-stack in Azure but not customers?

staebler: Maybe I am missing something here. Is it burdensome to require a developer that wants to use an unsupported feature to have to build their own installer? I can appreciate that it is easier for said developer to enable the feature via an environment variable. But that pushes the burden of maintenance from the developer wanting to test the feature to the installer team. Everyone wants their own environment variable for their own unique use case, and the installer team cannot support that at scale.

aojea: it is hard to create automation if you have to build it ...

staebler: Help me understand what we are dealing with here, because I clearly do not have the context.

Are there existing informing CI jobs that will start to fail if we do not have the environment variable?

Are there CI jobs in another repo that does support IPv6 on Azure even though the OpenShift as a whole does not?

If we want to support IPv6 on Azure in 4.9, can we wait until the master branch opens for 4.9 to enable support for the feature in the installer, without needing an environment variable? Or is there some reason why you need access to IPv6 on Azure in 4.8?

aojea: So, dual-stack is going to be a thing for baremetal in 4.8 and it is going beta in Kubernetes in 1.21 (that is 4.8 for Openshift) Azure/Microsoft is leading with Redhat the dual-stack Kubernetes effort, and Kubernetes have CI upstream for dualstack in Azure, so we know Azure supports Kubernetes dual-stack.

I don't know exactly how Dual-stack baremetal CI works, but they use series of bash scripts that uses libvirt to create an environment in a super big machine and, until this week, it was running only a small subset of the tests.

In parallel to the baremetal efforts, I've checked the feasibility of using Azure directly and, surprisingly, it worked at first time with only 2 modifications. The CI jobs that I'm setting up for Azure are optional and for self consume, until someone says this has to be "official" supported, but when that happens we'll have everything working and tested.

staebler: > In parallel to the baremetal efforts, I've checked the feasibility of using Azure directly and, surprisingly, it worked at first time with only 2 modifications. The CI jobs that I'm setting up for Azure are optional and for self consume, until someone says this has to be "official" supported, but when that happens we'll have everything working and tested.

Do you expect to run these optional tests on PRs to ensure that dual-stack continues to work? If it works now, what is keeping us from adding official support, or at least having a target release where we add official support? In the absence of official support, would we feel comfortable telling a customer that we have experimental support? Can a customer try out dual-stack on Azure with the understanding that support is experimental?

danwinship: There are currently no users and no jobs using dual-stack on Azure in OCP, because dual-stack-on-Azure-in-OCP doesn't currently work and never did. However, we are now a few PRs away from having it finally work.

At that point, we want to create `e2e-azure-ovn-dualstack` jobs in a few repos, because the current `e2e-metal-ipi-ovn-dualstack` can't run the full e2e suite because some of the tests don't work on bare metal, but would run correctly on Azure.

Additionally, we'd like random OCP developers to be able to bring up dual stack clusters more easily, so that, eg, when we find a dual-stack-related bug in the console operator, we can assign it to them to fix, rather than having to fix it ourselves (which we mostly have to do now, because most OCP developers do not have any access to bare metal clusters). So we'd like random OCP developers to be able to bring up dual-stack Azure clusters just like they can any other cluster (more or less), and especially, we'd like cluster-bot to be able to bring up dual-stack Azure clusters.

staebler: OK. I can get behind these reasons for why we want the environment variable. Thank you for walking me to this. Can we rename the environment variable `OPENSHIFT_INSTALL_EXPERIMENTAL_DUAL_STACK` and check that specifically for Azure?

pkg/types/validation/installconfig.go
```diff
@@ -196,7 +196,7 @@ func validateNetworkingIPVersion(n *types.Networking, p *types.Platform) field.E
 		}
 
 		switch {
-		case p.Azure != nil:
+		case p.Azure != nil && os.Getenv("OPENSHIFT_INSTALL_EXPERIMENTAL_DUAL_STACK") == "true":
```
staebler, line 199: 1. Make the environment variable a bit more friendly by supporting more "true" values.
2. Add a warning that an experimental feature is being used.
```
experimentalDualStackEnabled, _ := strconv.ParseBool(os.Getenv("OPENSHIFT_INSTALL_EXPERIMENTAL_DUAL_STACK"))
switch {
case p.Azure != nil && experimentalDualStackEnabled:
	fmt.Warn("Using experimental dual-stack support")
```

danwinship:
updated.
/hold
for a bit of testing...

data/data/azure/main.tf
```diff
@@ -45,9 +45,8 @@ module "bootstrap" {
   private                = module.vnet.private
   outbound_udr           = var.azure_outbound_user_defined_routing
 
-  use_ipv4                  = var.use_ipv4 || var.azure_emulate_single_stack_ipv6
+  use_ipv4                  = var.use_ipv4
```
staebler, line 48: The extra whitespace before the `=` needs to be remove to appease the format checkers.
You can update the formatting with the following command.
```
podman run --rm \
 --env IS_CONTAINER=TRUE \
 --volume "${PWD}:${PWD}:z" \
 --workdir "${PWD}" \
 terraform fmt -write=true -recursive data/data
```

staebler (commented):
Looks good to me, other than some minor formatting.

data/data/azure/main.tf
danwinship: fixed

danwinship:
/hold cancel
dual-stack cluster comes up partially-Degraded... it's possible I missed one of Antonio's other pending PRs in my build, but anyway, the installer part is fine:

```
danw@p50:cluster> ./openshift-install create cluster
FATAL failed to fetch Metadata: failed to load asset "Install Config": invalid "install-config.yaml" file: networking: Invalid value: "DualStack": dual-stack IPv4/IPv6 is not supported for this platform, specify only one type of address 
danw@p50:cluster> OPENSHIFT_INSTALL_EXPERIMENTAL_DUAL_STACK=true ./openshift-install create cluster
WARNING Using experimental Azure dual-stack support  
INFO Credentials loaded from file "/home/danw/.azure/osServicePrincipal.json" 
...
FATAL Bootstrap failed to complete                 
danw@p50:cluster> oc get nodes -o yaml
...
  status:
    addresses:
    - address: dwinship-w9699-master-0
      type: Hostname
    - address: 10.0.0.5
      type: InternalIP
    - address: fc00::5
      type: InternalIP
...
```

/test e2e-azure
(looks like it's been flaky lately so I don't think the failure here is because of this PR. I tested that I could successfully install a single-stack Azure cluster with this installer.)

staebler:
> /test e2e-azure
> 
> (looks like it's been flaky lately so I don't think the failure here is because of this PR. I tested that I could successfully install a single-stack Azure cluster with this installer.)

The install succeeded on the e2e-azure run. That is good enough for me for these changes.
/approve

aojea:
lovely :clap:

PR #4772, opened 2021-03-19
aws: support for bring-your-own hosted zone

staebler:
Add the `.aws.hostedZone` field to the install config to support the user supplying an existing hosted zone for the internal private hosted zone for the cluster. This can only be used when the user is also supplying their own VPC. The hosted zone must already be associated with the user-provided VPC.

Add validation in the "Platform Provisioning Check" asset for the user-provided internal hosted zone. The validation checks that
(1) the hosted zone is associated with the user-provided VPC and that (2) the hosted zone does not contain any record sets for subdomains of the cluster's domain.

The latter of these checks is meant to provide a modicum of protection against the user accidentally trying to install again a cluster that they have already installed. When it comes time to destroy the second failed installation, the destroyer will not be able to tell that the records sets in the hosted zone are actually being used by a different cluster.

When the user provides the private hosted zone to use for the cluster, the destroyer still needs to delete the recordsets for the cluster when the cluster is destroyed. There is no way to tag recordsets, so we must rely on tagging the hosted zone as shared. When the destroyer encounters a hosted zone tagged as shared by the cluster, the destroyer will delete all recordsets in that hosted zone that are strict subdomains of the cluster's domain. The cluster domain is added to the AWS cluster metadata so that it is available to the destroyer.

https://issues.redhat.com/browse/CORS-1666

### DIFF
--- a/pkg/asset/installconfig/aws/validation.go
+++ b/pkg/asset/installconfig/aws/validation.go
@@ -6,8 +6,12 @@ import (
 	"net"
 	"net/url"
 	"sort"
+	"strings"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/route53"
 	"github.com/pkg/errors"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -261,4 +265,88 @@ var requiredServices = []string{
 	"s3",
 	"sts",
 	"tagging",
+}
+
+// ValidateForProvisioning validates if the install config is valid for provisioning the cluster.
+func ValidateForProvisioning(session *session.Session, ic *types.InstallConfig, metadata *Metadata) error {
+	allErrs := field.ErrorList{}
+	allErrs = append(allErrs, validateExistingHostedZone(session, ic, metadata)...)
+	return allErrs.ToAggregate()
+}
+
+func validateExistingHostedZone(session *session.Session, ic *types.InstallConfig, metadata *Metadata) field.ErrorList {
+	if ic.AWS.HostedZone == "" {
+		return nil
+	}
+
+	// validate that the hosted zone exists
+	hostedZonePath := field.NewPath("aws", "hostedZone")
+	client := route53.New(session)
+	zone, err := client.GetHostedZone(&route53.GetHostedZoneInput{Id: aws.String(ic.AWS.HostedZone)})
+	if err != nil {
+		return field.ErrorList{
+			field.Invalid(hostedZonePath, ic.AWS.HostedZone, "cannot find hosted zone"),
+		}
+	}
+
+	allErrs := field.ErrorList{}
+
+	// validate that the hosted zone is associated with the VPC containing the existing subnets for the cluster
+	vpcID, err := metadata.VPC(context.TODO())
+	if err == nil {
+		if !isHostedZoneAssociatedWithVPC(zone, vpcID) {
+			allErrs = append(allErrs, field.Invalid(hostedZonePath, ic.AWS.HostedZone, "hosted zone is not associated with the VPC"))
+		}
+	} else {
+		allErrs = append(allErrs, field.Invalid(hostedZonePath, ic.AWS.HostedZone, "no VPC found"))
+	}
+
+	// validate that the hosted zone does not already have any record sets for the cluster domain
+	dottedClusterDomain := ic.ClusterDomain() + "."
+	var problematicRecords []string
+	if err := client.ListResourceRecordSetsPages(
+		&route53.ListResourceRecordSetsInput{HostedZoneId: zone.HostedZone.Id},
+		func(out *route53.ListResourceRecordSetsOutput, lastPage bool) bool {
+			for _, recordSet := range out.ResourceRecordSets {
+				name := aws.StringValue(recordSet.Name)
+				// skip record sets that are not sub-domains of the cluster domain. Such record sets may exist for
+				// hosted zones that are used for other clusters or other purposes.
+				if !strings.HasSuffix(name, dottedClusterDomain) {
+					continue
+				}
+				// skip record sets that are the cluster domain. Record sets for the cluster domain are fine. If the
+				// hosted zone has the name of the cluster domain, then there will be NS and SOA record sets for the
+				// cluster domain.
+				if len(name) == len(dottedClusterDomain) {
+					continue
+				}
+				problematicRecords = append(problematicRecords, fmt.Sprintf("%s (%s)", name, aws.StringValue(recordSet.Type)))
+			}
+			return !lastPage
+		},
+	); err != nil {
+		allErrs = append(allErrs, field.InternalError(hostedZonePath,
+			errors.Wrapf(err, "could not list record sets for hosted zone %q", ic.AWS.HostedZone)))
+	}
+	if len(problematicRecords) > 0 {
+		detail := fmt.Sprintf(
+			"hosted zone already has record sets for the domain of the cluster: [%s]",
+			strings.Join(problematicRecords, ", "),
+		)
+		allErrs = append(allErrs, field.Invalid(hostedZonePath, ic.AWS.HostedZone, detail))
+	}
+
+	return allErrs
+}
+
+func isHostedZoneAssociatedWithVPC(hostedZone *route53.GetHostedZoneOutput, vpcID string) bool {
+	if vpcID == "" {
+		return false
+	}
+	for _, vpc := range hostedZone.VPCs {
+		if aws.StringValue(vpc.VPCId) == vpcID {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/destroy/aws/aws.go
+++ b/pkg/destroy/aws/aws.go
@@ -59,10 +59,11 @@ type ClusterUninstaller struct {
 	//   }
 	//
 	// will match resources with (a:b and c:d) or d:e.
-	Filters   []Filter // filter(s) we will be searching for
-	Logger    logrus.FieldLogger
-	Region    string
-	ClusterID string
+	Filters       []Filter // filter(s) we will be searching for
+	Logger        logrus.FieldLogger
+	Region        string
+	ClusterID     string
+	ClusterDomain string
 
 	// Session is the AWS session to be used for deletion.  If nil, a
 	// new session will be created based on the usual credential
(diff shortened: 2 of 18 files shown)


staebler:
5659a4fc1...34aef8216
* rebased to changes from https://github.com/openshift/installer/pull/4790

patrickdillon (commented):
I have reviewed the first three commits and will take a closer look at the destroy code soon.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -262,3 +266,81 @@ var requiredServices = []string{
...
+	var problematicRecords []string
+	if err := client.ListResourceRecordSetsPages(
+		&route53.ListResourceRecordSetsInput{HostedZoneId: zone.HostedZone.Id},
+		func(out *route53.ListResourceRecordSetsOutput, lastPage bool) bool {
+			for _, recordSet := range out.ResourceRecordSets {
+				name := aws.StringValue(recordSet.Name)
+				if !strings.HasSuffix(name, dottedClusterDomain) {
+					continue
+				}
+				if len(name) == len(dottedClusterDomain) {
```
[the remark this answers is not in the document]
staebler: Yes, every hosted zone has two records which are NS and SOA records with the same domain as the hosted zone. I will add a comment about this.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -262,3 +266,81 @@ var requiredServices = []string{
...
+					continue
+				}
+				problematicRecords = append(problematicRecords, fmt.Sprintf("%s (%s)", name, aws.StringValue(recordSet.Type)))
+			}
+			return !lastPage
+		},
+	); err != nil {
+		allErrs = append(allErrs, field.Invalid(hostedZonePath, ic.AWS.HostedZone, "could not list record sets for hosted zone"))
```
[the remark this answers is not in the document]
staebler: Yep, that makes sense.

pkg/destroy/aws/shared.go
```diff
@@ -0,0 +1,233 @@
...
+
+	var lastError error
+	err = client.ListResourceRecordSetsPagesWithContext(
+		ctx,
+		&route53.ListResourceRecordSetsInput{HostedZoneId: aws.String(id)},
+		func(results *route53.ListResourceRecordSetsOutput, lastPage bool) bool {
+			for _, recordSet := range results.ResourceRecordSets {
+				// skip record sets that are not part of the cluster
+				name := aws.StringValue(recordSet.Name)
+				if !strings.HasSuffix(name, dottedClusterDomain) {
```
patrickdillon, line 166: Is the name of the record set dotted or is it the the value of the record that is dotted? When I look in the GUI it looks like the name isn't dotted but the value is.

pkg/destroy/aws/shared.go
```diff
@@ -0,0 +1,233 @@
...
+					"recordset",
+					fmt.Sprintf("%s (%s)", aws.StringValue(recordSet.Name), aws.StringValue(recordSet.Type)),
+				)
+				// delete any matching record sets in the public hosted zone
+				if publicZoneID != "" {
+					if err := deleteMatchingRecordSetInPublicZone(ctx, client, publicZoneID, recordSet, logger); err != nil {
+						if lastError != nil {
+							logger.Debug(lastError)
+						}
+						lastError = errors.Wrapf(err, "deleting record set matching %#+v from public zone %s", recordSet, publicZoneID)
```
patrickdillon, line 182: same comment as above about `%#+v` verb

pkg/destroy/aws/shared.go
```diff
@@ -0,0 +1,233 @@
...
+						continue
+					}
+					recordSetLogger.Debug("Deleted from public zone")
+				}
+				// delete the record set
+				if err := deleteRoute53RecordSet(ctx, client, id, recordSet, logger); err != nil {
+					if lastError != nil {
+						logger.Debug(lastError)
+					}
+					lastError = errors.Wrapf(err, "deleting record set %#+v from zone %s", recordSet, id)
```
patrickdillon, line 194: same comment as above about `%#+v` verb

pkg/destroy/aws/shared.go
patrickdillon, line 182: Should we add in the error message for the public record that we skipped deleting the private record?

patrickdillon (commented):
This looks mostly sane to me with a few nits and a couple of questions, especially regarding the trailing dot. But it looks good. I may try to take another closer look as I don't have a lot of background in AWS destroy.

pkg/destroy/aws/aws.go
```diff
@@ -1862,14 +1863,17 @@ func deleteRoute53(ctx context.Context, session *session.Session, arn arn.ARN, l
...
-				if sharedEntry, ok := sharedEntries[key]; ok {
-					err := deleteRoute53RecordSet(ctx, client, sharedZoneID, sharedEntry, logger.WithField("public zone", sharedZoneID))
+				if publicEntry, ok := publicEntries[key]; ok {
+					err := deleteRoute53RecordSet(ctx, client, publicZoneID, publicEntry, logger.WithField("public zone", publicZoneID))
 					if err != nil {
 						if lastError != nil {
 							logger.Debug(lastError)
 						}
-						lastError = errors.Wrapf(err, "deleting public zone %s", sharedZoneID)
+						lastError = errors.Wrapf(err, "deleting record set %#+v from public zone %s", publicEntry, publicZoneID)
```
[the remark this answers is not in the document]
staebler: I agree. This was copied from https://github.com/openshift/installer/blob/6d778f911e79afad8ba2ff4301eda5b5cf4d8e9e/pkg/destroy/aws/aws.go#L1880. I will remove it here.

pkg/destroy/aws/shared.go
```diff
@@ -0,0 +1,233 @@
...
+						return err
+					}
+				} else {
+					o.Logger.Warnf("Ignoring non-owned cluster key %s: %s for shared-tag removal", key, value)
+				}
+			}
+		}
+	}
+
+	return nil
```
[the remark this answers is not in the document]
staebler: Yes, this is a bit of code that always takes me too long to understand every time that I look at it. I don't know that we want to rely on the `clusterID` being set correctly, although I can't think of a good reason why it wouldn't be.

Maybe it would be sufficient to have a function that gets the cluster tag keys from the filter.
```
func clusterOwnedKeys(filters []Filter) []string {
	var keys[] string
	for _, filter := range filters {
		for key, value := range filter {
			if !strings.HasPrefix(key, "kubernetes.io/cluster/") {
				continue
			}
			if value != "owned" {
				continue
			}
			keys = append(keys, key)
		}
	}
	return keys
}
```
And then the `removeSharedTags` function could be the following.
```
func removeSharedTags(ctx context.Context, tagClients []*resourcegroupstaggingapi.ResourceGroupsTaggingAPI, filters []Filter, logger logrus.FieldLogger) error {
	for _, key := range clusterOwnedKeys(filter) {
		if err := removeSharedTag(ctx, tagClients, key, logger); err != nil {
			return err
		}
	}
	return nil
}
```

pkg/destroy/aws/shared.go
```diff
@@ -0,0 +1,233 @@
...
+	if err != nil {
+		return err
+	}
+	logger = logger.WithField("id", id)
+
+	switch resourceType {
+	case "hostedzone":
+		return o.cleanSharedHostedZone(ctx, client, id, logger)
+	default:
+		logger.Debugf("nothing to clean for shared %s resource", resourceType)
```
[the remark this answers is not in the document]
staebler: It should be capitalized.

pkg/destroy/aws/shared.go
staebler: The name of the record set is dotted.
For example,
```
        {
            "Name": "api-int.ewolinetz3.devcluster.openshift.com.",
            "Type": "A",
            "AliasTarget": {
                "HostedZoneId": "ZLMOA37VPKANP",
                "DNSName": "ewolinetz3-rpcfk-int-42decb5de56b9b0b.elb.us-east-2.amazonaws.com.",
                "EvaluateTargetHealth": false
            }
        },
```

pkg/destroy/aws/shared.go
staebler: No. The destroyer should try again later to delete the records.

staebler:
@patrickdillon I have addressed your feedback.

patrickdillon:
I'm having a hard time reasoning through this destroy code in the context of avoiding deleting non-cluster records. I may also be hitting some limits in my knowledge.

I need to revisit the code & will do that immediately but we're in a bit of a time crunch so I want to write this out to aid the process/discussion. 

What about this scenario (I think this would pass pre-provision check, but let me know if something else makes this invalid):

We have a shared hosted zone: `shared.devcluster.openshift.com`

We install a cluster with the cluster domain `shared.devcluster.openshift.com`

We install a second cluster with the cluster domain `foo.shared.devcluster.openshift.com`

We delete the first cluster with cluster domain `shared.devcluster.openshift.com` doesn't that delete the records for the second cluster?

If so, I am thinking we should lean toward deleting whitelisted records  in this shared scenario (e.g. only delete `api` + `clusterdomain`, etc) I'm sure you've considered this and I imagine the downside is that IF the cluster creates non-whitelisted records they will be leaked, but that's not the case ATM right?

staebler:
For posterity, since this was discussed outside of this PR,

> I'm having a hard time reasoning through this destroy code in the context of avoiding deleting non-cluster records. I may also be hitting some limits in my knowledge.
> 
> I need to revisit the code & will do that immediately but we're in a bit of a time crunch so I want to write this out to aid the process/discussion.
> 
> What about this scenario (I think this would pass pre-provision check, but let me know if something else makes this invalid):
> 
> We have a shared hosted zone: `shared.devcluster.openshift.com`
> 
> We install a cluster with the cluster domain `shared.devcluster.openshift.com`
> 
> We install a second cluster with the cluster domain `foo.shared.devcluster.openshift.com`
> 
> We delete the first cluster with cluster domain `shared.devcluster.openshift.com` doesn't that delete the records for the second cluster?

Yes that will delete the records for the second cluster. The contention here is that it is a mis-configuration to have the second cluster use a cluster domain that is a sub domain of the first cluster. The first cluster owns its entire cluster domain. We cannot check for this on pre-provision because there is no way to (reliably) tell if a cluster with a parent domain is in the hosted zone already.

> 
> If so, I am thinking we should lean toward deleting whitelisted records in this shared scenario (e.g. only delete `api` + `clusterdomain`, etc) I'm sure you've considered this and I imagine the downside is that IF the cluster creates non-whitelisted records they will be leaked, but that's not the case ATM right?

A whitelist is problematic because record sets can be created by in-cluster components with names that the installer does not know about. The installer only creates the api and api-int record sets. For example, the *.apps record set is created in-cluster.

PR #4778, opened 2021-03-22
    Bug 1934557: RHCOS boot image bump for LUKS fixes

miabbott:
This boot image bump brings in fixes for the following BZs on the
RHCOS side:
    
BZ#1934174 - rootfs too small when enabling NBDE
BZ#1940704 - prjquota is dropped from rootflags if rootfs is reprovisioned
    
It also brings in support for the new ap-northeast-3 region in AWS.
    
BZ#1939661 - support new AWS region ap-northeast-3

### DIFF
--- a/data/data/rhcos-amd64.json
+++ b/data/data/rhcos-amd64.json
@@ -1,163 +1,166 @@
 {
     "amis": {
         "af-south-1": {
-            "hvm": "ami-01dd4193042351d0e"
+            "hvm": "ami-0b5c3b497fe4bc2ea"
         },
         "ap-east-1": {
-            "hvm": "ami-06320ea2b751acf5d"
+            "hvm": "ami-0a6976bcfac70580e"
         },
         "ap-northeast-1": {
-            "hvm": "ami-05368806cdd6277c7"
+            "hvm": "ami-06cbc55d814a971e4"
         },
         "ap-northeast-2": {
-            "hvm": "ami-0b2f40b281ab6903b"
+            "hvm": "ami-045e0e1127cf9d4ff"
+        },
+        "ap-northeast-3": {
+            "hvm": "ami-0afb77eabb999b7bc"
         },
         "ap-south-1": {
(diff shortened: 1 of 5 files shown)


jlebon:
I've verified that this version of RHCOS fixes the `prjquota` and LUKS growfs bugs on x86_64 (this time on NVMe, which triggered the `lsblk` bug).

/lgtm

miabbott:
`e2e-aws-upgrade`, `e2e-aws-workers-rhel7`, `e2e-openstack` appear to have been failing well before this PR

`e2e-crc` was a bit flaky before this PR, though it looks like at least the bootstrap node is getting created

cgwalters:
/hold
Let's rebase this on https://github.com/openshift/installer/pull/4582

miabbott:
> /hold
> Let's rebase this on #4582

While I'm not opposed technically, I do want to land this bump so that we can verify the 4.8 BZ and unblock the process for getting the 4.7.z bump done to land the fixes for NBDE, prjquota, and new AWS region in Osaka.

Can we remove the hold here to unblock the log jam?

cgwalters:
I've rolled in the changes here to https://github.com/openshift/installer/pull/4582

staebler:
> > /hold
> > Let's rebase this on #4582
> 
> While I'm not opposed technically, I do want to land this bump so that we can verify the 4.8 BZ and unblock the process for getting the 4.7.z bump done to land the fixes for NBDE, prjquota, and new AWS region in Osaka.
> 
> Can we remove the hold here to unblock the log jam?

For the purpose of cherry-picking to 4.7, it would be easier if this PR merges before #4582.

cgwalters:
Note we can't actually cherry pick because there are separate RHCOS builds per OpenShift version.  So no matter what we do, the process will need to be bypassed.

But sure, it's fine if this merges first.

cgwalters:
Since it looks like upgrades are going to be on the floor, and bootimage bumps can't affect that (in any way I can think of) let's override?
/hold cancel

PR #4785, opened 2021-03-23
images: Add Dockerfile.upi.ci.rhel8

cgwalters:
Fork the current rhel7 Dockerfile and port it to rhel8.  This
makes it easier to use both python2 and python3 at the same time,
which we need for gcloud (python2) and azure (python3).

We need to land this, then try changing CI to actually use
it in a rehearsal job in openshift/release.

### DIFF
--- a/images/installer/Dockerfile.upi.ci.rhel8
+++ b/images/installer/Dockerfile.upi.ci.rhel8
@@ -1,0 +1,68 @@
+# NOTE: Forked temporarily from Dockerfile.upi.ci to port to rhel8
+# This Dockerfile is used by CI to test UPI platforms for OpenShift Installer
+# It builds an image containing binaries like jq, terraform, awscli, oc, etc. to allow bringing up UPI infrastructure.
+# It also contains the `upi` directory that contains various terraform and cloud formation templates that are used to create infrastructure resources.
+
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.8 AS builder
+WORKDIR /go/src/github.com/openshift/installer
+COPY . .
+RUN hack/build.sh
+
+FROM registry.ci.openshift.org/ocp/4.8:cli as cli
+
+FROM registry.ci.openshift.org/ocp/4.8:base
+COPY --from=cli /usr/bin/oc /bin/oc
+COPY --from=builder /go/src/github.com/openshift/installer/bin/openshift-install /bin/openshift-install
+COPY --from=builder /go/src/github.com/openshift/installer/upi /var/lib/openshift-install/upi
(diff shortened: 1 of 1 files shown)


cgwalters:
So the proposal here is to land this PR and then do a separate PR to openshift/release which switches to it (and starts ignoring the current `Dockerfile.upi.ci`) and also fixes the `from:` there to use rhel8.

Alternatively, we could try powering through this:
- Overwrite the `Dockerfile.upi.ci` here with this and force merge
- Quickly merge the fix to openshift/release
- See if there's any fallout that needs fixing and iterate on that in this repo

images/installer/Dockerfile.upi.ci.rhel8
```diff
@@ -0,0 +1,68 @@
...
+      gettext \
+      google-cloud-sdk \
+      gzip \
+      jq \
+      unzip \
+      openssh-clients \
+      openssl \
+      python3-pyOpenSSL \
+      python2-pyyaml \
+      python3-pyyaml \
```
cgwalters, line 35: Note we can easily get both python2+3 pyyaml now too; there's a *lot* of stuff in openshift/release still using `python` (aka `python2`).

images/installer/Dockerfile.upi.ci.rhel8
```diff
@@ -0,0 +1,68 @@
...
+      util-linux && \
+    yum clean all && \
+    rm -rf /var/cache/yum/* && \
+    chmod g+w /etc/passwd
+
+# Not packaged for Python 2, but required by gcloud.  See https://cloud.google.com/sdk/crypto
+RUN pip-2 install pyopenssl
+ENV CLOUDSDK_PYTHON=/usr/bin/python
```
staebler, line 43: This may not be needed any more.

staebler (commented):
Long term I would like to separate out this image into smaller images that are built as supplemental CI images. But the changes in this PR seem reasonable as a short-term solution. Or I suppose even long-term we would still want to have the UPI image built on rhel8 rather than rhel7.

cgwalters:
Yeah, we could perhaps just do that now too - something like a new github.com/openshift/installer-upi-ci repo?  Or just stick this Dockerfile inline into openshift/release ?

> Or I suppose even long-term we would still want to have the UPI image built on rhel8 rather than rhel7.

Definitely, no reason to drag rhel7 around longer.

staebler:
> Yeah, we could perhaps just do that now too - something like a new github.com/openshift/installer-upi-ci repo? Or just stick this Dockerfile inline into openshift/release ?

I was thinking about putting the images in https://github.com/openshift/release/tree/master/clusters/app.ci/supplemental-ci-images, which I just learned about today.

staebler (approved):
/lgtm

cgwalters:
PR to use this in https://github.com/openshift/release/pull/17110

PR #4789, opened 2021-03-24
azure: changes the way we get credentials

m1kola:
Change in the function signatures allows a caller to provide credentials.

This is useful when we want to avoid a dependency on a file system or asking for manual input. This can happen in managed OpenShift services, where some packages from the installer are being used in the context of the webserver.

### DIFF
--- a/pkg/asset/installconfig/azure/metadata.go
+++ b/pkg/asset/installconfig/azure/metadata.go
@@ -22,12 +22,31 @@ type Metadata struct {
 	// ARMEndpoint indicates the resource management API endpoint used by AzureStack.
 	ARMEndpoint string `json:"armEndpoint,omitempty"`
 
+	// Credentials hold prepopulated Azure credentials.
+	// At the moment the installer doesn't use it and reads credentials
+	// from the file system, but external consumers of the package can
+	// provide credentials. This is useful when we run the installer
+	// as a service (Azure Red Hat OpenShift, for example): in this case
+	// we do not want to rely on the filesystem or user input as we
+	// serve multiple users with different credentials via a web server.
+	Credentials *Credentials `json:"credentials,omitempty"`
+
 	mutex sync.Mutex
 }
 
 // NewMetadata initializes a new Metadata object.
 func NewMetadata(cloudName typesazure.CloudEnvironment, armEndpoint string) *Metadata {
-	return &Metadata{CloudName: cloudName, ARMEndpoint: armEndpoint}
+	return NewMetadataWithCredentials(cloudName, armEndpoint, nil)
+}
+
+// NewMetadataWithCredentials initializes a new Metadata object
+// with prepopulated Azure credentials.
+func NewMetadataWithCredentials(cloudName typesazure.CloudEnvironment, armEndpoint string, credentials *Credentials) *Metadata {
+	return &Metadata{
+		CloudName:   cloudName,
+		ARMEndpoint: armEndpoint,
+		Credentials: credentials,
+	}
 }
 
 // Session holds an Azure session which can be used for Azure API calls
@@ -42,7 +61,7 @@ func (m *Metadata) Session() (*Session, error) {
 func (m *Metadata) unlockedSession() (*Session, error) {
 	if m.session == nil {
 		var err error
-		m.session, err = GetSession(m.CloudName, m.ARMEndpoint)
+		m.session, err = GetSessionWithCredentials(m.CloudName, m.ARMEndpoint, m.Credentials)
 		if err != nil {
 			return nil, errors.Wrap(err, "creating Azure session")
 		}
--- a/pkg/asset/installconfig/azure/session.go
+++ b/pkg/asset/installconfig/azure/session.go
@@ -43,29 +43,47 @@ type Credentials struct {
 // GetSession returns an azure session by using credentials found in ~/.azure/osServicePrincipal.json
 // and, if no creds are found, asks for them and stores them on disk in a config file
 func GetSession(cloudName azure.CloudEnvironment, armEndpoint string) (*Session, error) {
-	authFile := defaultAuthFilePath
-	if f := os.Getenv(azureAuthEnv); len(f) > 0 {
-		authFile = f
-	}
-	return newSessionFromFile(cloudName, authFile, armEndpoint)
+	return GetSessionWithCredentials(cloudName, armEndpoint, nil)
 }
 
-func newSessionFromFile(cloudName azure.CloudEnvironment, authFilePath, armEndpoint string) (*Session, error) {
-	// NewAuthorizerFromFileWithResource uses `auth.GetSettingsFromFile`, which uses the `azureAuthEnv` to fetch the auth credentials.
-	// therefore setting the local env here to authFilePath allows NewAuthorizerFromFileWithResource to load credentials.
-	os.Setenv(azureAuthEnv, authFilePath)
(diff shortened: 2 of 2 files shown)


pkg/asset/installconfig/azure/session.go
```diff
@@ -77,39 +115,22 @@ func newSessionFromFile(authFilePath string, cloudName azure.CloudEnvironment) (
 		return nil, errors.Wrap(err, "failed to get settings from file")
 	}
 
-	authSettings.Values[auth.ActiveDirectoryEndpoint] = env.ActiveDirectoryEndpoint
```
m1kola, line 80: I believe we no longer need this line as we explicitly set endpoint in `auth.ClientCredentialsConfig` now.

pkg/asset/installconfig/azure/metadata.go
```diff
@@ -19,12 +19,24 @@ type Metadata struct {
...
 	CloudName typesazure.CloudEnvironment `json:"cloudName,omitempty"`
 
+	// Credentials hold prepopulated Azure credentials.
+	// At the moment the installer doesn't use it and reads credentials
+	// from the file system, but external consumers of the package can
+	// provide credentials. This is useful when we run the installer
+	// as a service (Azure Red Hat OpenShift, for example): in this case
+	// we do not want to rely on the filesystem or user input as we
+	// serve multiple users with different credentials via a web server.
+	Credentials *Credentials `json:"credentials,omitempty"`
```
m1kola, line 29: For the installer this field will be omited during marshaling as the installer doesn't use it. 

Wondering if we can avoid marshaling, but my previous implementation (like [this one with an interface](https://github.com/openshift/installer/commit/de541bf88a6bda8a294a944f788f2e5d7266277f)) failed.

Have I missed any implications of doing this?

m1kola:
Some tests are flaky, but I think this is ready for review.

/assign @staebler

pkg/asset/installconfig/installconfig.go
```diff
@@ -144,7 +144,7 @@ func (a *InstallConfig) finish(filename string) error {
 		a.AWS = aws.NewMetadata(a.Config.Platform.AWS.Region, a.Config.Platform.AWS.Subnets, a.Config.AWS.ServiceEndpoints)
 	}
 	if a.Config.Azure != nil {
-		a.Azure = icazure.NewMetadata(a.Config.Azure.CloudName)
+		a.Azure = icazure.NewMetadata(a.Config.Azure.CloudName, nil)
```
staebler, line 147: Can you give some more details about how ARO will call `NewMetadata`? Will ARO replace the metadata set here? Or does ARO not load the install config asset?

pkg/asset/installconfig/azure/metadata.go
```diff
@@ -19,12 +19,24 @@ type Metadata struct {
...
+
 	mutex sync.Mutex
 }
 
 // NewMetadata initializes a new Metadata object.
-func NewMetadata(cloudName typesazure.CloudEnvironment) *Metadata {
-	return &Metadata{CloudName: cloudName}
+func NewMetadata(cloudName typesazure.CloudEnvironment, credentials *Credentials) *Metadata {
```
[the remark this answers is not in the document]
m1kola: I did not go for this initially because I was a bit afraid that it might look redundant when someone looks here without taking external usage into account. But I hope the code comment around credentials and the fact that this function is exported will deter future us from refactoring and removing `NewMetadataWithCredentials` :)

m1kola (commented):
@staebler updated. Please take another look

pkg/asset/installconfig/azure/metadata.go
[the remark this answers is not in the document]
staebler: I share your concern. However, I would be just as concerned about someone removing the `credentials` field.

pkg/asset/installconfig/installconfig.go
staebler: That is helpful. Thanks you.
Can you help me track down which calls in the ARO eventually lead to calls to `installConfig.Azure.Session`?

staebler: I think I found it. It is through code such as https://github.com/Azure/ARO-RP/blob/58fd2f4b79b6d0a71fd2575db0cf77e18b722328/pkg/cluster/deploystorage.go#L179-L188. Right?

staebler: Can we drill down into why you need the credentials in the installconfig metadata?
Are you using the metadata as a store for the credentials? Or are you using it to just so that the installer can have the credentials when generating assets?
Do you need the credentials marshaled? Or will the credentials only be used in the initial resolve of the graph and not later after reloading the graph?

m1kola: @staebler

> I think I found it. It is through code such as https://github.com/Azure/ARO-RP/blob/58fd2f4b79b6d0a71fd2575db0cf77e18b722328/pkg/cluster/deploystorage.go#L179-L188. Right?

You are right.

> Are you using the metadata as a store for the credentials? Or are you using it to just so that the installer can have the credentials when generating assets?

We only need creds so that the installer can generate assets.

> Do you need the credentials marshaled? Or will the credentials only be used in the initial resolve of the graph and not later after reloading the graph?

I believe we don't need them marshaled. I tried to avoid marshaling of the credentials in the earlier version of this PR ([here](https://github.com/openshift/installer/commit/52d40af64bc6acf13531de178cca33e40010f033#diff-da950a4ddabbede621d9d3b1058bb34f8931c89179306ee88a0e4d76a4cf0b13R48-R50) is one of the older commits). The earlier version works for ARO, but not for the installer because the installer reloads the graph and panics after unmarshalling because there will be no `credGetter` [here](https://github.com/openshift/installer/commit/52d40af64bc6acf13531de178cca33e40010f033#diff-6b7f83e796e074b2f55bd6a9b666d65d140fe7477bd98b78bcc51210980213ddR47).

In [our fork of the installer](https://github.com/openshift/installer/commit/afbdc9c2bec45e0b268e94ee31623712d93205eb#diff-6b7f83e796e074b2f55bd6a9b666d65d140fe7477bd98b78bcc51210980213ddR16) you can see that we we pass creds into the Metadata and do not have them marshalled as the field is not exported. It works for ARO, but in our fork backward compatibility with upstream installer is not the top priority :) I tried to rework this bit in a way that doesn't hurt the installer and doens't require gating code with an ARO-specific build tag.

I feel like you are asking these questions because you have an alternative aproach in mind. Please let me know if that's the caseю

staebler: I do not have an alternative approach in mind, but I am trying to understand the problem more clearly so that I can consider whether there may be alternative approaches that are less likely to be accidentally removed in the future.

staebler (commented):
Just one nit. Otherwise, it looks fine.

m1kola:
Rebased on top of the `master` and changed the code accrodingly to @staebler's suggestion.

staebler (approved):
/lgtm
/hold for azure e2e tests
/test e2e-azure

staebler:
e2e-azure and e2e-azure-shared-vpc both succeeded.
e2e-azure-upi failed, but that is not related to these changes.
/hold cancel

m1kola:
Rebased on top of the `master` branch to fix conflicts with https://github.com/openshift/installer/pull/5024.

/test e2e-azure
/test e2e-azure-upi
/test e2e-azure-shared-vpc

PR #4790, opened 2021-03-25
Byo IAM roles for IPI install

smrowley:


pkg/asset/cluster/aws/aws.go
```diff
@@ -29,48 +30,100 @@ func Metadata(clusterID, infraID string, config *types.InstallConfig) *awstypes.
 // PreTerraform performs any infrastructure initialization which must
 // happen before Terraform creates the remaining infrastructure.
 func PreTerraform(ctx context.Context, clusterID string, installConfig *installconfig.InstallConfig) error {
-	if len(installConfig.Config.Platform.AWS.Subnets) == 0 {
-		return nil
+	if len(installConfig.Config.Platform.AWS.Subnets) != 0 {
```
staebler, line 33: Rather than have a big if block, let's have `PreTerraform` call two functions: One function will tag the subnets, and the other function will tag the iam roles. The subnet-tagging function can return early when there are no user-provided subnets.

pkg/asset/cluster/aws/aws.go
```diff
@@ -29,48 +30,100 @@ func Metadata(clusterID, infraID string, config *types.InstallConfig) *awstypes.
...
 		return err
 	}
-
-	tags := []*ec2.Tag{
+	ec2Tags := []*ec2.Tag{
 		{
 			Key:   aws.String(fmt.Sprintf("kubernetes.io/cluster/%s", clusterID)),
 			Value: aws.String("shared"),
 		},
 	}
```
staebler, line 93: Let's use a function to get the tag key and value rather than duplicate the logic for the ec2 tags and the iam tags.
```suggestion
	key, value := sharedTag(clusterID)
	ec2Tags := []*ec2.Tag{{Key: &key, Value: &value}}
```

pkg/types/aws/machinepool.go
```diff
@@ -24,6 +24,11 @@ type MachinePool struct {
 	//
 	// +optional
 	EC2RootVolume `json:"rootVolume"`
+
+	// The name of a user-created IAM Role to use. This takes the place of an installer-created role.
```
staebler, line 28: The description should start with the name of the field. And some other suggestions on this description to keep it consistent with other BYO fields.
```suggestion
	// IAMRole is the name of the IAM Role to use for the instance profile of the machine. Leave unset to have the installer create the IAM Role on your behalf.
```

pkg/types/aws/machinepool.go
```diff
@@ -24,6 +24,11 @@ type MachinePool struct {
...
+	//
+	// +optional
+	IAMRoleName string `json:"iamRoleName,omitempty"`
```
staebler, line 31: The name part seems unnecessary.
```suggestion
	IAMRole string `json:"iamRole,omitempty"`
```

pkg/types/aws/machinepool.go
```diff
@@ -24,6 +24,11 @@ type MachinePool struct {
 	//
 	// +optional
 	EC2RootVolume `json:"rootVolume"`
+
+	// The name of a user-created IAM Role to use. This takes the place of an installer-created role.
+	//
```
staebler, line 29: We don't need this whitespace.
```suggestion
```

pkg/types/installconfig.go
```diff
@@ -15,14 +15,16 @@ import (
 	"github.com/openshift/installer/pkg/types/openstack"
 	"github.com/openshift/installer/pkg/types/ovirt"
 	"github.com/openshift/installer/pkg/types/vsphere"
+	"github.com/pkg/errors"
```
staebler, line 18: This should go in a new section between the golang imports and the installer imports. But I don't think you need this import anyway as you can use `fmt.Errorf` instead of `errors.Errorf`.

pkg/types/installconfig.go
```diff
@@ -15,14 +15,16 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
 	// InstallConfigVersion is the version supported by this package.
 	// If you bump this, you must also update the list of convertable values in
 	// pkg/types/conversion/installconfig.go
-	InstallConfigVersion = "v1"
+	InstallConfigVersion  = "v1"
+	WorkerMachinePoolName = "worker"
```
staebler, line 27: Does this need to be public?
```suggestion
	workerMachinePoolName = "worker"
```

pkg/types/installconfig.go
```diff
@@ -351,3 +353,13 @@ type BootstrapInPlace struct {
...
 }
+
+func (installConfig *InstallConfig) WorkerMachinePool() (*MachinePool, error) {
+	for _, machinePool := range installConfig.Compute {
+		if machinePool.Name == WorkerMachinePoolName {
+			return &machinePool, nil
+		}
+	}
+
+	return nil, errors.Errorf("failed to retrieve MachinePool with name '%s'", WorkerMachinePoolName)
```
staebler, line 364:
```suggestion
	return nil, fmt.Errorf("failed to retrieve MachinePool with name %q", WorkerMachinePoolName)
```

pkg/types/installconfig.go
```diff
@@ -351,3 +353,13 @@ type BootstrapInPlace struct {
 	// InstallationDisk is the target disk drive for coreos-installer
 	InstallationDisk string `json:"installationDisk"`
 }
+
+func (installConfig *InstallConfig) WorkerMachinePool() (*MachinePool, error) {
```
staebler, line 357: Stay consistent with the receiver name used elsewhere.
```suggestion
func (c *InstallConfig) WorkerMachinePool() (*MachinePool, error) {
```

staebler, line 357: I think that I would rather this function not return an error. It is sufficient to return a nil `*MachinePool` when the worker cannot be found. Although we do not support it currently, it is not inconceivable that at some point in the future we would allow the user to explicitly omit the worker pool.

pkg/types/aws/machinepool.go
```diff
@@ -25,10 +25,9 @@ type MachinePool struct {
 	// +optional
 	EC2RootVolume `json:"rootVolume"`
 
-	// The name of a user-created IAM Role to use. This takes the place of an installer-created role.
-	//
+	// IAMRole is the name of the IAM Role to use for the instance profile of the machine. Leave unset to have the installer create the IAM Role on your behalf.
```
staebler, line 28: Wrap this line.
```suggestion
	// IAMRole is the name of the IAM Role to use for the instance profile of the machine.
	// Leave unset to have the installer create the IAM Role on your behalf.
```

.gitignore
```diff
@@ -1,3 +1 @@
-/bin/
-
-.vscode
\ No newline at end of file
+/bin/
```
staebler, line 1: Add back in the newline that was there before, so that there are no changes to this file.

pkg/asset/cluster/aws/aws.go
```diff
@@ -99,27 +85,43 @@ func tagSubnetEC2Instances(ctx context.Context, clusterID string, installConfig
 	return nil
 }
 
-func tagIamRoles(ctx context.Context, clusterID string, installConfig *installconfig.InstallConfig, iamRoleNames []*string) error {
+func tagIamRoles(ctx context.Context, clusterID string, installConfig *installconfig.InstallConfig) error {
+	workerMachinePool := installConfig.Config.WorkerMachinePool()
+
+	iamRoleNames := make([]*string, 0)
```
staebler, line 91:
```suggestion
	var iamRoleNames []*string
```

data/data/aws/main.tf
```diff
@@ -66,12 +67,14 @@ module "masters" {
...
   publish_strategy         = var.aws_publish_strategy
+  iam_role_name            = var.aws_master_iam_role_name
 }
 
 module "iam" {
   source = "./iam"
 
-  cluster_id = var.cluster_id
+  cluster_id      = var.cluster_id
+  worker_iam_role_name   = var.aws_worker_iam_role_name
```
staebler, line 77: These need to be aligned on the `=`.
Run `./hack/tf-fmt.sh` to see the formatting violations.
You can also run the following to make the format changes.
```
podman run --rm \
 --env IS_CONTAINER=TRUE \
 --volume "${PWD}:${PWD}:z" \
 --workdir "${PWD}" \
 terraform fmt -write=true -recursive data/data
```

data/data/aws/iam/main.tf
```diff
@@ -45,9 +46,9 @@ resource "aws_iam_role_policy" "worker_policy" {
   // This list should not be updated any further without an operator owning migrating changes here for existing
   // clusters.
   // Please see: docs/dev/aws/iam_permissions.md
-
```
staebler, line 48: Don't remove this newline.

data/data/aws/master/main.tf
```diff
@@ -51,9 +52,9 @@ resource "aws_iam_role_policy" "master_policy" {
   // This list should not be updated any further without an operator owning migrating changes here for existing
   // clusters.
   // Please see: docs/dev/aws/iam_permissions.md
-
```
staebler, line 54: Don't remove this newline.

pkg/explain/printer_test.go
```diff
@@ -217,7 +217,7 @@ VERSION:  v1
 
 RESOURCE: <object>
   InstallConfig is the configuration for an OpenShift install.
-`,
+		`,
```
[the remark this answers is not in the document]
staebler: There are still gratuitous changes in these tests. Are you finding that these changes are necessary for the tests to pass for some reason?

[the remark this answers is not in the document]
smrowley: Yes, the test is expecting exact match on tabbing. I could add a trim if that is preferred.

staebler (approved):
/lgtm

PR #4799, opened 2021-03-28
WIP: IPI installation to Azure Stack

staebler:
* Add AzureStack as a valid value for the azure cloudName field.
* Add the `platform.azure.armEndpoint` field for specifying the ARM endpoint when installing on Azure Stack.
* Add validation specific to Azure Stack.
* Add the github.com/terraform-providers/terraform-provider-azurestack module. This is replaced with a fork that bumps the azurerm provider used and fixes a bug the an invalid ConflictsWith in the availability set resource.
* Add terraform for provisioning a cluster using the azurestack provider.

https://issues.redhat.com/browse/CORS-1665

This builds on https://github.com/openshift/installer/pull/4729

/cc @jhixson74 @patrickdillon


patrickdillon (commented):
Rather than `/data/data/azure/main-stack.tf` and the exclusion logic couldn't we just write the tf configs in `/data/data/azurestack`? At a quick glance it doesn't look like azurestack is reusing any of the azure modules; but perhaps there is a benefit or need?

staebler:
> Rather than `/data/data/azure/main-stack.tf` and the exclusion logic couldn't we just write the tf configs in `/data/data/azurestack`? At a quick glance it doesn't look like azurestack is reusing any of the azure modules; but perhaps there is a benefit or need?

That is how I originally wrote it. But I had used different variables names (like `azurestack_region`). I didn't like having to duplicate all the go code for building out the variables. And I didn't like the prospect of keeping those in sync. Now that I am using the same terraform variable names for both, I think that you are correct that it would be better to put the azure stack in `/data/data/azurestack`. I could symlink the terraform variables file from `/data/data/azure` to `/data/data/azurestack`.

staebler:
b833f34d2...503ea329f
* Move terraform files for Azure Stack to `data/data/azurestack`.

ausil:
what exactly does arm mean in this case? I am guessing not the arm architecture?

patrickdillon:
> what exactly does arm mean in this case? I am guessing not the arm architecture?

I believe it stands for azure resource manager

ausil:
> > what exactly does arm mean in this case? I am guessing not the arm architecture?
> 
> I believe it stands for azure resource manager

okay, thanks, I thought so, I only just learned it existed. This is likely to go cause confusion with the ARM architecture.

PR #4801, opened 2021-03-29
Bug 1944268: aws: allow use of unknown regions in known partitions

staebler:
Allow the user to specify a region that is not known to the instatller but does match the regex of a known partition without requiring the user to specify custom service endpoints. This is needed to support new regions that are added to known partitions.

Without this, it is not possible to use an unknown region is the standard partition. Terraform will not create the IAM roles correctly when specifying custom endpoints for a region in the standard partition because the request will try to create the role in the cluster's region rather than in us-east-1.

This seems like the behavior that was desired when the check for whether custom service endpoints are needed was added in https://github.com/openshift/installer/commit/cf79a75308faf38e11f758457ca40748b797e56a.

In addition to relaxing the service endpoint validation, the validation around whether an AMI must be specified has been tightened. The user must specify an AMI when the RHCOS stream does not contain an AMI for the region, rather than when the ASK SDK knows about the region.

### DIFF
--- a/pkg/asset/installconfig/aws/validation.go
+++ b/pkg/asset/installconfig/aws/validation.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
+	"github.com/openshift/installer/pkg/rhcos"
 	"github.com/openshift/installer/pkg/types"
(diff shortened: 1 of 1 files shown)


jstuever (approved):
/lgtm
/approve

jstuever:
/retitle Bug 1944268: aws: allow use of unknown regions in known partitions

PR #4802, opened 2021-03-30
Workload partitioning

lack:
This series of commits:
- Adds github.com/cri-o/cpuset
- Adds a new install-config.yaml 'workloadSettings' section (optional) as outlined in https://github.com/openshift/enhancements/pull/703, which uses the cpuset library to validate the input
- Uses the new 'workloadSettings' section to generate MachineConfig objects that configure both Cri-o and Kubelet on installed hosts to use the specified cpuset for management workloads.


pkg/types/validation/installconfig.go
```diff
@@ -609,3 +611,29 @@ func validateIPProxy(proxy string, n *types.Networking, fldPath *field.Path) fie
...
+		if partitionNames[p.Name] {
+			allErrs = append(allErrs, field.Duplicate(partitionFldPath.Child("name"), p.Name))
+		}
+		partitionNames[p.Name] = true
+		if p.CPUIds == "" {
+			allErrs = append(allErrs, field.Invalid(partitionFldPath.Child("cpuIDs"), p.CPUIds, "cannot be empty"))
+		} else if cpuset, err := cpuset.Parse(p.CPUIds); err != nil || cpuset.IsEmpty() {
+			// Re-check if the parsed cpuset is empty, since the parser may interpret some errors as the empty set.
+			allErrs = append(allErrs, field.Invalid(partitionFldPath.Child("cpuIDs"), p.CPUIds, "could not parse the cpuset"))
+		}
```
staebler, line 636: Do we need to check for empty twice? If the `IsEmpty` check sufficient?
```suggestion
		swtich cpuset, err := cpuset.Parse(p.CPUIds); {
		case err != nil:
			// invalid error
		case cpuset.IsEmpty():
			// required error
		}
```

pkg/types/validation/installconfig.go
```diff
@@ -609,3 +611,29 @@ func validateIPProxy(proxy string, n *types.Networking, fldPath *field.Path) fie
...
+		if p.Name != types.ManagementWorkloadPartition {
+			allErrs = append(allErrs, field.NotSupported(partitionFldPath.Child("name"), p.Name, []string{string(types.ManagementWorkloadPartition)}))
+		}
+		if partitionNames[p.Name] {
+			allErrs = append(allErrs, field.Duplicate(partitionFldPath.Child("name"), p.Name))
+		}
+		partitionNames[p.Name] = true
+		if p.CPUIds == "" {
+			allErrs = append(allErrs, field.Invalid(partitionFldPath.Child("cpuIDs"), p.CPUIds, "cannot be empty"))
+		} else if cpuset, err := cpuset.Parse(p.CPUIds); err != nil || cpuset.IsEmpty() {
```
staebler, line 633: The `cpuset` variable name hides the `cpuset` package name.

pkg/types/validation/installconfig_test.go
```diff
@@ -203,12 +203,23 @@ func validOvirtPlatform() *ovirt.Platform {
...
+func validWorkloadSettings() *types.InstallConfig {
+	c := validInstallConfig()
+	c.Workload = append(c.Workload, types.WorkloadPartition{
+		Name:   types.ManagementWorkloadPartition,
+		CPUIds: "0-1",
+	})
+	return c
+}
+
+type testCase struct {
```
staebler, line 215: Don't add this. Keep the test case type anonymous.

pkg/asset/machines/machineconfig/workload-partitioning.go
```diff
@@ -0,0 +1,127 @@
...
+// annotation_prefix = "io.openshift.workload.management"
+// resources = { "cpu" = "", "cpuset" = "0-1", }
+func CrioWorkloadDropinContents(name, cpuset string) (string, error) {
+	/*
+		[crio.runtime.workloads.management]
+		label = "management.workload.openshift.io/cores"
+		annotation_prefix = "io.openshift.workload.management"
+		resources = { "cpu" = "", "cpuset" = "0-1", }
+	*/
+	crioIni := ini.Empty()
```
mrunalp, line 31: The format we use is toml. CRI-O uses "github.com/BurntSushi/toml" to read it.

pkg/types/validation/installconfig_test.go
```diff
@@ -1208,6 +1219,58 @@ func TestValidateInstallConfig(t *testing.T) {
...
+			}(),
+			expectedError: `^workloadSettings\[0\].cpuIDs: Invalid value: "": cannot be empty`,
+		},
+	}
+	for _, id := range []string{"0", "0-1", "3,5,7", "3,5-8,11", "2-3,5-6,8-9"} {
+		cases = append(cases, validCpuidTestcase(id))
+	}
+	for _, id := range []string{"nodigits", "-7", "10-", ",2", "4,", "5,,6", "8--10", "2,4,bad", "5,6-,8", "2-3,-6,8-9", "10-6", "0,5-2,7"} {
+		cases = append(cases, invalidCpuidTestcase(id))
 	}
```
staebler, line 1274: These cases are testing the `cpuset.Parse` function, which is not the unit under test.

pkg/asset/machines/machineconfig/workload-partitioning.go
```diff
@@ -0,0 +1,127 @@
...
+	"encoding/json"
+	"fmt"
+
+	igntypes "github.com/coreos/ignition/v2/config/v3_2/types"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+	ini "gopkg.in/ini.v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/installer/pkg/asset/ignition"
+	"github.com/pkg/errors"
```
staebler, line 14: Move this import to the previous block.

pkg/asset/machines/machineconfig/workload-partitioning.go
```diff
@@ -0,0 +1,127 @@
...
+// label = "management.workload.openshift.io/cores"
+// annotation_prefix = "io.openshift.workload.management"
+// resources = { "cpu" = "", "cpuset" = "0-1", }
+func CrioWorkloadDropinContents(name, cpuset string) (string, error) {
+	/*
+		[crio.runtime.workloads.management]
+		label = "management.workload.openshift.io/cores"
+		annotation_prefix = "io.openshift.workload.management"
+		resources = { "cpu" = "", "cpuset" = "0-1", }
+	*/
```
staebler, line 30: Remove this. It is the same example in the description of the function.

pkg/asset/machines/machineconfig/workload-partitioning.go
```diff
@@ -0,0 +1,127 @@
...
+	crioIni := ini.Empty()
+	section := crioIni.Section(fmt.Sprintf("crio.runtime.workloads.%s", name))
+	err := section.ReflectFrom(&crioWorkloadCfg{
+		Label:            fmt.Sprintf(`"%s.workload.openshift.io/cores"`, name),
+		AnnotationPrefix: fmt.Sprintf(`"io.openshift.workload.%s"`, name),
+		Resources:        fmt.Sprintf(`{ "cpu" = "", "cpuset" = "%s", }`, cpuset),
+	})
+	if err != nil {
+		return "", errors.Wrap(err, "Could not reflect structure to INI")
```
staebler, line 39: Errors should start with lowercase.
```suggestion
		return "", errors.Wrap(err, "could not reflect structure to INI")
```

pkg/asset/machines/machineconfig/workload-partitioning.go
```diff
@@ -0,0 +1,127 @@
...
+	return crioBuf.String(), nil
+}
+
+type crioWorkloadCfg struct {
+	Label            string `ini:"label"`
+	AnnotationPrefix string `ini:"annotation_prefix"`
+	Resources        string `ini:"resources"`
+}
+
+// KubeletWorkloadDropinContents generates te content expected by Kubelet for workload partitioning
```
staebler, line 55:
```suggestion
// KubeletWorkloadDropinContents generates the content expected by Kubelet for workload partitioning
```

pkg/asset/machines/machineconfig/workload-partitioning.go
```diff
@@ -0,0 +1,127 @@
...
+		Ignition: igntypes.Ignition{
+			Version: igntypes.MaxVersion.String(),
+		},
+		Storage: igntypes.Storage{
+			Files: []igntypes.File{
+				ignition.FileFromString(
+					fmt.Sprintf("/etc/crio/crio.conf.d/01-%s-workload", name),
+					"root", 0644, crioCfg),
+				ignition.FileFromString(
+					"/etc/kubernetes/workload-pinning",
```
staebler, line 102: Does this filename not need specificity for the workload? Like `/etc/kubernetes/workload-management-pinning`.

pkg/asset/machines/machineconfig/workload-partitioning.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, errors.Wrap(err, "Could not convert to raw extension")
+	}
+
+	return &mcfgv1.MachineConfig{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: mcfgv1.SchemeGroupVersion.String(),
+			Kind:       "MachineConfig",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("02-%s-workload-partitioning", role),
```
staebler, line 118: It seems like this name should be specific to the workload as well.

staebler, line 118: What is the significance of the "02-" at the start of the machineconfig name?

pkg/asset/machines/machineconfig/workload-partitioning_test.go
```diff
@@ -0,0 +1,53 @@
...
+		assert.Equal(t, err, nil, "No err")
+		assert.Equal(t, expectedKubeletCfg, kubeletCfg)
+
+		result, err := ForWorkloadPartitioning(name, cpuset, role)
+		assert.Equal(t, err, nil, "No err")
+		cfg := igntypes.Config{}
+		err = json.Unmarshal(result.Spec.Config.Raw, &cfg)
+		assert.Equal(t, err, nil, "No err")
+		files := cfg.Storage.Files
+		assert.Equal(t, 2, len(files), "Two files in the machineconfig")
```
staebler, line 47:
```suggestion
		assert.Len(t, files, 2, "Two files in the machineconfig")
```

pkg/asset/machines/master.go
```diff
@@ -402,6 +402,14 @@ func (m *Master) Generate(dependencies asset.Parents) error {
...
 		machineConfigs = append(machineConfigs, ignFIPS)
 	}
+	if len(ic.Workload) > 0 {
+		w := ic.Workload[0]
+		ignWorkload, err := machineconfig.ForWorkloadPartitioning(string(w.Name), w.CPUIds, "master")
+		if err != nil {
+			return errors.Wrap(err, "failed to create ignition for workload partitioning for master machines")
+		}
+		machineConfigs = append(machineConfigs, ignWorkload)
+	}
```
staebler, line 412: Do this for all workload partitions now so that we don't have to change this later when we support more.
```suggestion
	for _, w := range ic.Workload {
		ignWorkload, err := machineconfig.ForWorkloadPartitioning(string(w.Name), w.CPUIds, "master")
		if err != nil {
			return errors.Wrap(err, "failed to create ignition for workload partitioning for master machines")
		}
		machineConfigs = append(machineConfigs, ignWorkload)
	}
```

pkg/asset/machines/master.go
```diff
@@ -402,6 +402,14 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 		}
 		machineConfigs = append(machineConfigs, ignFIPS)
 	}
+	if len(ic.Workload) > 0 {
+		w := ic.Workload[0]
+		ignWorkload, err := machineconfig.ForWorkloadPartitioning(string(w.Name), w.CPUIds, "master")
+		if err != nil {
+			return errors.Wrap(err, "failed to create ignition for workload partitioning for master machines")
```
staebler, line 409: Add context for which partition had problems.
```suggestion
			return errors.Wrapf(err, "failed to create ignition for workload partitioning %q for master machines", w.Name)
```

pkg/asset/machines/master_test.go
```diff
@@ -10,16 +11,57 @@ import (
 	"github.com/openshift/installer/pkg/asset"
 	"github.com/openshift/installer/pkg/asset/ignition/machine"
 	"github.com/openshift/installer/pkg/asset/installconfig"
+	"github.com/openshift/installer/pkg/asset/machines/machineconfig"
 	"github.com/openshift/installer/pkg/asset/rhcos"
 	"github.com/openshift/installer/pkg/types"
 	awstypes "github.com/openshift/installer/pkg/types/aws"
+	"github.com/vincent-petithory/dataurl"
```
staebler, line 18: Move this to the previous block of imports.

pkg/asset/machines/master_test.go
```diff
@@ -10,16 +11,57 @@ import (
...
+  osImageURL: ""
+`, role, dataurl.EncodeBytes([]byte(expectedCrioCfg)), dataurl.EncodeBytes([]byte(expectedKubeletCfg)))
+}
+
 func TestMasterGenerateMachineConfigs(t *testing.T) {
 	cases := []struct {
 		name                  string
 		key                   string
 		hyperthreading        types.HyperthreadingMode
+		workloadCpuset        string
```
staebler, line 64: Change this to `[]types.WorkloadPartition` and insert it directly into the InstallConfig when creating it.

pkg/asset/machines/machineconfig/workload-partitioning.go
lack: Looked a lot like ini, but sure, I can switch over :)

pkg/asset/machines/worker.go
```diff
@@ -223,6 +223,14 @@ func (w *Worker) Generate(dependencies asset.Parents) error {
 			}
 			machineConfigs = append(machineConfigs, ignFIPS)
 		}
+		if len(ic.Workload) > 0 {
+			w := ic.Workload[0]
+			ignWorkload, err := machineconfig.ForWorkloadPartitioning(string(w.Name), w.CPUIds, "worker")
```
staebler, line 228: This gives me pause. I've asked a question in the enhancement PR about whether the same cpuset should be used for master and workers. See https://github.com/openshift/enhancements/pull/703#pullrequestreview-624640278.

pkg/types/installconfig.go
```diff
@@ -351,3 +360,33 @@ type BootstrapInPlace struct {
...
+//
+// The first release only supports a single workload partition whose name must
+// be "management".
+// +kubebuilder:validation:Enum="";management
+type WorkloadPartitionName string
+
+const (
+	// ManagementWorkloadPartition pins all infrastructure pods to the
+	// defined cpuset.
+	ManagementWorkloadPartition WorkloadPartitionName = "management"
```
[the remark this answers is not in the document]
lack: That's an interesting thing... The explicit value `management` is what the name should be because then it will really match what the crio and kubelet configs get.  So do we really want the user input to not match what ends up in the configs?

pkg/asset/machines/master.go
lack: Can do.

pkg/asset/machines/worker.go
lack: Given that this will only be used for SNO in 4.8 it probably doesn't matter, but I agree it's worth designing this right out of the gate.

pkg/types/installconfig.go
```diff
@@ -145,6 +145,15 @@ type InstallConfig struct {
...
 	BootstrapInPlace *BootstrapInPlace `json:"bootstrapInPlace,omitempty"`
+
+	// WorkloadSettings defines a set of workload partitions, which will
+	// assign a set of workloads to a specific set of CPUs.
+	//
+	// The first release only supports a single workload partition whose
+	// name must be "management", which will pin all infrastructure pods to
+	// the defined cpuset.
+	// +optional
+	Workload []WorkloadPartition `json:"workloadSettings,omitempty"`
```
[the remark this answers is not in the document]
lack: Not as described today in the enhancement.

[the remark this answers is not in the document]
staebler: I don't want to get locked in with having to add another field at the top level if we need to. I think I'd rather add one more level here now.

pkg/types/installconfig.go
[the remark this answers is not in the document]
staebler: That makes sense. This isn't really an enum but rather a name with limited values.

pkg/asset/machines/machineconfig/workload-partitioning.go
lack: @mrunalp : I have a pending change that switches over to `github.com/BurtSushi/toml` and the encoder wants to create a file with this contents... Is it okay?  It looks a bit different than I was expecting:
```
["crio.runtime.workloads.master"]
label = "master.workload.openshift.io/cores"
annotation_prefix = "io.openshift.workload.master"
resources = "{ \"cpu\" = \"\", \"cpuset\" = \"0-1\", }"
```

pkg/asset/machines/master.go
lack: Turns out it works better to combine all of the partitions into a single machineconfig, since bot the CRI-O and Kubelet config file drop-ins are designed to handle multiple partitions in each, so I did it that way.  Push with the changes coming soon.

pkg/asset/machines/worker.go
lack: For this first release it probably doesn't matter, and if we want different pools in the future, perhaps it's not too hard to eventually also allow the same "WorkloadSettings" in install-config.yaml to be added under the `compute` and `controlPlane` in addition to it being globally available at the top level?

pkg/asset/machines/machineconfig/workload-partitioning.go
lack: @mrunalp : A slack conversation with @haircommander revealed that the format produced by `github.com/BurntSushi/toml` doesn't work, but the format produced in this current PR by `gopkg.in/ini.v1` works just fine:
```
[crio.runtime.workloads.management]
label             = "management.workload.openshift.io/cores"
annotation_prefix = "io.openshift.workload.management"
resources         = { "cpu" = "", "cpuset" = "0-1", }
```
So I think it's best to stick with what I've got here - `gopkg.in/ini.v1` is already used elsewhere in this repo anyway.

pkg/types/machinepools.go
```diff
@@ -125,3 +132,29 @@ func (p *MachinePoolPlatform) Name() string {
...
+	// The name of the workload
+	Name WorkloadName `json:"name"`
+
+	// CPUIds is a CPU set specifier for the CPUs to add to the isolated
+	// set for this workload, using the same cpuset syntax used elsewhere
+	// in kubernetes. Asking for explicit IDs instead of simply a count
+	// gives the user control in situations where they need specific CPUs
+	// to be available because they're close to accelerators, NICs, or
+	// other special hardware.  For example, "1,5,9" or "0-4" or
+	// "0-1,52-53" +kubebuilder:validation:Pattern=[0-9][-,0-9]*
```
staebler, line 158:
```suggestion
	// "0-1,52-53"
	// +kubebuilder:validation:Pattern=[0-9][-,0-9]*
```

pkg/asset/machines/worker_test.go
```diff
@@ -249,3 +256,59 @@ func TestComputeIsNotModified(t *testing.T) {
 		t.Fatalf("compute in the install config has been modified")
 	}
 }
+
+func TestWorkloadPartitioning(t *testing.T) {
```
staebler, line 260: What is this supposed to be testing? I don't see any mention of workloads in the function.

pkg/types/validation/installconfig.go
```diff
@@ -395,6 +395,9 @@ func validateCompute(platform *types.Platform, control *types.MachinePool, pools
 			allErrs = append(allErrs, field.Invalid(poolFldPath.Child("architecture"), p.Architecture, "heteregeneous multi-arch is not supported; compute pool architecture must match control plane"))
 		}
 		allErrs = append(allErrs, ValidateMachinePool(platform, &p, poolFldPath)...)
+		if len(p.Workloads) > 0 {
+			allErrs = append(allErrs, field.Invalid(poolFldPath.Child("workloads"), p.Workloads, "workloads can only be specified for controlPlane pools"))
```
staebler, line 399:
```suggestion
			allErrs = append(allErrs, field.Invalid(poolFldPath.Child("workloads"), p.Workloads, "workloads can not be specified for compute pools"))
```

pkg/asset/machines/machineconfig/workload-partitioning.go
```diff
@@ -0,0 +1,122 @@
...
+	return string(kubeletCfg), nil
+}
+
+type kubeletWorkloadEntry struct {
+	Cpuset string `json:"cpuset"`
+}
+
+// ForWorkloads creates the MachineConfig that configures Cri-O and
+// Kubelet with the workload partition settings from install-config.yaml
+// 'workloadSettings'
```
staebler, line 77: What is this line meant to indicate?

pkg/asset/machines/worker.go
```diff
@@ -223,6 +223,13 @@ func (w *Worker) Generate(dependencies asset.Parents) error {
 			}
 			machineConfigs = append(machineConfigs, ignFIPS)
 		}
+		if len(pool.Workloads) > 0 {
```
staebler, line 226: Don't put this code in here yet, since we don't support it.

pkg/types/validation/installconfig.go
```diff
@@ -395,6 +395,9 @@ func validateCompute(platform *types.Platform, control *types.MachinePool, pools
 			allErrs = append(allErrs, field.Invalid(poolFldPath.Child("architecture"), p.Architecture, "heteregeneous multi-arch is not supported; compute pool architecture must match control plane"))
 		}
 		allErrs = append(allErrs, ValidateMachinePool(platform, &p, poolFldPath)...)
+		if len(p.Workloads) > 0 {
```
staebler, line 398: Can we do this check in `ValidateMachinePool` instead so that we don't report potential errors for the specific workloads in addition to reporting that workloads as a whole are not allowed? You could either pass in the type of machine pool being validated or rely on the name of the machine pool to distinguish.

staebler (commented):
This looks good to me overall. I have some nits. And there are some nits that I think may have been missed from the previous review (from github helpfully hiding them). I took the liberty of resolving the ones of those that are obsolete.

pkg/types/machinepools.go
lack: Good catch, not sure how that formatting got broken.

pkg/asset/machines/worker_test.go
lack: Another odd editing error, removing.

pkg/asset/machines/machineconfig/workload-partitioning.go
lack: A left-over from the bulk rename.  Fixing now.

pkg/asset/machines/worker.go
lack: Good point.

pkg/types/validation/installconfig.go
lack: Good idea!  Moved.

pkg/asset/machines/machineconfig/workload-partitioning.go
lack: No, this is a single file which contains all the workloads.

pkg/types/validation/installconfig_test.go
lack: This was a hold-over from an earlier invocation that used a regex for validation instead of the cpuset module like the current implementation does.  I left it in as proof that the other library does what we want and external change there doesn't break is in the future.  But I suppose the unit test is there in vendors/ if we need it.

Gone.

lack:
> This looks good to me overall. I have some nits. And there are some nits that I think may have been missed from the previous review (from github helpfully hiding them). I took the liberty of resolving the ones of those that are obsolete.

Latest push addresses all of those nits.  Thanks for all your time on this!

pkg/asset/machines/machineconfig/workload-partitioning.go
lack: MCO creates `01-master-container-runtime` and `01-master-kubelet` so it felt like this, which modifies both CRI-O and Kubelet would be +1 from there.  Though strictly speaking I don't think it matters.

lack:
After some discussion today with @smarterclayton we're shelving this installer price of the feature for now.  Users opting in to this feature will manually create manifests (or use external tooling to do so.

PR #4803, opened 2021-03-31
Bug 1929136: OpenStack: document Manila share mounting

Fedosin:


docs/user/openstack/known-issues.md
```diff
@@ -87,6 +87,20 @@ volumeBindingMode: WaitForFirstConsumer
 
 [nova-az-setting]: ../openstack#setting-nova-availability-zones
 
+## Mounting Manila shares to pods
```
mandre, line 90: It bothers me a bit that we're documenting this as a known issue: it is not an issue but instruction on how to connect your pods to the manila network.
I think we should split this into its own "Deploying with Manila" file.

tombarron: Agree, and thanks for getting this documented!

mdbooth (changes requested):
This is a bug in the manila operator, imho. It should do this automatically if required.

I'm fine documenting this as a workaround, but I think we need to open an issue against the manila operator to track it. The issue should link to this documentation with a note to remove it when the issue is resolved. Additionally, this PR should contain a link to the issue.

tombarron (approved):
Thanks for documenting this critical bit of information, Mike!  We really need to get it in front of our customers as several of them have gotten stuck at this point.

docs/user/openstack/connect_nodes_to_manila_network.md
```diff
@@ -0,0 +1,35 @@
+# Connecting nodes to a dedicated Manila network
+
+If you are going to use OpenStack Manila on your cluster, you should be aware that the service usually uses a standalone network for its shares, and by default the installer does not connect nodes to this network. It causes problems with mounting shares on an already deployed cluster, as there is no route to the network:
```
maxwelldb, line 3: >  the service usually uses a standalone network for its shares

Is it worth enumerating situations where this is not the case?

tombarron: Here is the deal.  CephFS via NFS back end (Red Hat supplied) will always need this extra network connection.  So will some vendor storage back ends.  But vendor backends support a mode that we in Maniila call DHSS=True where the NFS server presents itself directly on the tenant network (private network where the manila client resides).  In these cases no extra network is required.

@maxwelldb as you can see I don't have a crisp, succinct way to enumerate those situations but maybe you can craft something worth adding here ...

mandre, line 3: Not sure we need to detail the different cases where Manila uses an additional network. Perhaps we could phase it like:

```suggestion
Depending on your OpenStack platform configuration, it is very likely that Manila uses a dedicated network for its shares. When that is the case, you need to attach your OpenShift compute nodes to this network otherwise pods might fail to mount the shares, as shown in the following example:
```

Also add something like `Consult with your OpenStack administrator to know what network Manila exposes its shares on.` after the example error message.

mandre (commented):
@mdbooth Are you suggesting that Manila CSI operator should modify the worker VMs to connect to the manila network? Even if it had the information on which network Manila exposes its shares, it doesn't seem to be the responsibility of Manila operator to change the definition of the VMs. More the responsibility of the Machine API Operator, which is what we're documenting here.

docs/user/openstack/connect_nodes_to_manila_network.md
mandre: Please see also the second part of my comment.

mandre (commented):
/lgtm
/approve

PR #4804, opened 2021-03-31
Bug 1945236: add proxy params to bootstrap ignition

EmilienM:
Add Proxy config to the bootstrap ignition, so in the case of
a disconnected install using a local proxy, ignition will be able to go
through the proxy.

### DIFF
--- a/pkg/tfvars/openstack/bootstrap_ignition.go
+++ b/pkg/tfvars/openstack/bootstrap_ignition.go
@@ -16,6 +16,7 @@ import (
 	"github.com/vincent-petithory/dataurl"
 
 	"github.com/openshift/installer/pkg/asset/ignition"
+	"github.com/openshift/installer/pkg/types"
 	openstackdefaults "github.com/openshift/installer/pkg/types/openstack/defaults"
 )
 
@@ -86,7 +87,7 @@ func parseCertificateBundle(userCA []byte) ([]igntypes.Resource, error) {
 
 // generateIgnitionShim is used to generate an ignition file that contains a user ca bundle
 // in its Security section.
-func generateIgnitionShim(userCA string, clusterID string, bootstrapConfigURL string, tokenID string) (string, error) {
+func generateIgnitionShim(userCA string, clusterID string, bootstrapConfigURL string, tokenID string, proxy *types.Proxy) (string, error) {
 	fileMode := 420
 	bootstrapHTTPResponseHeaders := 120
 
@@ -152,6 +153,7 @@ func generateIgnitionShim(userCA string, clusterID string, bootstrapConfigURL st
 					},
 				},
 			},
+			Proxy: ignitionProxy(proxy),
 		},
 		Storage: igntypes.Storage{
 			Files: []igntypes.File{
@@ -188,4 +190,25 @@ func getAuthToken(cloud string) (string, error) {
 	}
 
 	return token, nil
+}
+
+func ignitionProxy(proxy *types.Proxy) igntypes.Proxy {
+	var ignProxy igntypes.Proxy
+	if proxy == nil {
+		return ignProxy
+	}
+	if httpProxy := proxy.HTTPProxy; httpProxy != "" {
+		ignProxy.HTTPProxy = &httpProxy
+	}
+	if httpsProxy := proxy.HTTPSProxy; httpsProxy != "" {
+		ignProxy.HTTPSProxy = &httpsProxy
+	}
+	ignProxy.NoProxy = make([]igntypes.NoProxyItem, 0, len(proxy.NoProxy))
+	if noProxy := proxy.NoProxy; noProxy != "" {
+		noProxySplit := strings.Split(noProxy, ",")
+		for _, p := range noProxySplit {
+			ignProxy.NoProxy = append(ignProxy.NoProxy, igntypes.NoProxyItem(p))
+		}
+	}
+	return ignProxy
 }
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -466,6 +466,7 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			bootstrapIgn,
 			installConfig.Config.ControlPlane.Platform.OpenStack,
 			installConfig.Config.Platform.OpenStack.MachinesSubnet,
+			installConfig.Config.Proxy,
 		)
 		if err != nil {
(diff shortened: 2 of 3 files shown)


staebler (commented):
/approve

pkg/tfvars/openstack/bootstrap_ignition.go
```diff
@@ -137,6 +138,20 @@ func generateIgnitionShim(userCA string, clusterID string, bootstrapConfigURL st
 		},
 	}
 
+	var httpProxy, httpsProxy *string
+	if proxy.HTTPProxy != "" {
+		httpProxy = &proxy.HTTPProxy
+	}
+	if proxy.HTTPSProxy != "" {
+		httpsProxy = &proxy.HTTPSProxy
+	}
```
iamemilio, line 147: If `httpProxy` and `httpsProxy` can be *string type, then there is actually no reason to check for nil. You might as well just set them to the pointer in the proxy object pointer. This is because the [zero value for a pointer in GO is always nil](https://yourbasic.org/golang/default-zero-value/). However, what does need to be checked here is whether the proxy pointer itself is nil, since you de-reference its values by checking `proxy.HTTPProxy != ""`.

pkg/tfvars/openstack/bootstrap_ignition.go
```diff
@@ -137,6 +138,20 @@ func generateIgnitionShim(userCA string, clusterID string, bootstrapConfigURL st
...
+
+	noProxySplitted := strings.Split(proxy.NoProxy, ",")
```
iamemilio, line 149:
```suggestion
	noProxySplit := strings.Split(proxy.NoProxy, ",")
```
variable grammer nit

pkg/tfvars/openstack/bootstrap_ignition.go
iamemilio: This is why the code panics on line 142 when the proxy is unset (nil)
```log
INFO[2021-04-07T17:10:24Z] panic: runtime error: invalid memory address or nil pointer dereference 
INFO[2021-04-07T17:10:24Z] [signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xacf7fec] 
INFO[2021-04-07T17:10:24Z]                                              
INFO[2021-04-07T17:10:24Z] goroutine 1 [running]:                       
INFO[2021-04-07T17:10:24Z] github.com/openshift/installer/pkg/tfvars/openstack.generateIgnitionShim(0x0, 0x0, 0xc0007dab20, 0x14, 0xc000738360, 0x58, 0xc0012763c0, 0xb7, 0x0, 0x0, ...) 
INFO[2021-04-07T17:10:24Z] 	/go/src/github.com/openshift/installer/pkg/tfvars/openstack/bootstrap_ignition.go:142 +0x46c 
```

staebler: The check for not empty **is** necessary. There is a distinction between a nil pointer and a pointer to an empty string. A pointer to an empty string will not be omitted from the marshaled ignition config.

staebler:
> /hold still WIP and also we'll have to fix aws, which also use go to build ignition config, so potentially users want proxy.

This is not needed for AWS. The CA bundles added to the bootstrap ignition shim are needed to access the AWS API when fetching the full ignition config from the S3 bucket. That does not require going through a proxy.

/unlabel platform/aws

pkg/tfvars/openstack/bootstrap_ignition.go
```diff
@@ -137,6 +138,20 @@ func generateIgnitionShim(userCA string, clusterID string, bootstrapConfigURL st
...
+	noProxy := make([]igntypes.NoProxyItem, 0, len(proxy.NoProxy))
+	for _, p := range noProxySplitted {
+		noProxy = append(noProxy, igntypes.NoProxyItem(p))
+	}
```
staebler, line 153:
```suggestion
	noProxy := make([]igntypes.NoProxyItem, len(proxy.NoProxy))
	for i, p := range noProxySplitted {
		noProxy[i] = igntypes.NoProxyItem(p)
	}
```

staebler:
> /hold something fails when proxy vars are unset

We still need to address https://github.com/openshift/installer/pull/4804#discussion_r608856509, right?

EmilienM:
> > /hold something fails when proxy vars are unset
> 
> We still need to address [#4804 (comment)](https://github.com/openshift/installer/pull/4804#discussion_r608856509), right?

Yes my bad, I forgot to fix it. It should be good now, I tested it against my OSP cloud and actual proxies.

pkg/tfvars/openstack/bootstrap_ignition.go
```diff
@@ -137,6 +138,27 @@ func generateIgnitionShim(userCA string, clusterID string, bootstrapConfigURL st
 		},
 	}
 
+	var httpProxy, httpsProxy *string
+	Proxy := types.Proxy{}
+	if proxy != nil {
+		Proxy = *proxy
+	}
```
staebler, line 145: Don't use variable names that start with capital letters. We don't even need a new variable at all.
```suggestion
	if proxy == nil {
		proxy = &types.Proxy{}
	}
```
Although this seems like a bit of a convoluted way to get around the extra indentation of an `if` block.
```
if proxy != nil {
	if proxy.HTTPProxy != "" {
		httpProxy = &proxy.HTTPProxy
	}
	...
}
```
How would you feel about a different approach where we have a `ignitionProxy` function?
```
func ignitionProxy(proxy *types.Proxy) igntypes.Proxy {
	var ignProxy igntypes.Proxy
	if proxy == nil {
		return ignProxy
	}
	if p := proxy.HTTPProxy; p != "" {
		proxy.HTTPProxy = &p
	}
	...
}
```
Then when we are building the ignition config, we simply call the function.
```
			Proxy: ignitionProxy(proxy),
```

EmilienM: sounds good. I'll give it a try tomorrow, thanks!

EmilienM: done

staebler (commented):
Looks good to me. I'll leave it up to someone on the OpenStack team to give it a lgtm.
/approve

mandre (commented):
/lgtm
e2e-openstack [already succeeded](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/4804/pull-ci-openshift-installer-master-e2e-openstack/1380031623713525760)

PR #4805, opened 2021-03-31
Bug 1878655: aws installconfig: endpoint validation should be before others

jstuever:
This change moves the AWS end point validation to happen before
other validations. This will ensure the endpoints are validated before
being used by other validations, where they may cause undesired long
timeouts.

https://issues.redhat.com/browse/CORS-1546


staebler (commented):
Even when the service endpoints validation fails, the rest of the validation will still run? Won't we still have the delay with this change?

jstuever:
> Even when the service endpoints validation fails, the rest of the validation will still run. Won't we still have the delay with this change?

I believe you are right. Need to return here instead of continuing. Will fix.

staebler (commented):
This needs a rebase, but otherwise LGTM.

jstuever:
Rebased... not sure why the bot pulled that label, but thanks for pointing it out.

staebler (approved):
/lgtm

PR #4809, opened 2021-04-01
Bug 1945659: remove ovirt_cafile from ovirt-credentials secret

Gal-Zaidman:
When the user tried to use an ovirt.config with a ca_file the ca_file path is written to the secret but this will not be the path of the file in the created machines.
This problem will cause connection errors in the machine object that tries to find the file that doesn't exist.
Instead, we should read the file content and write it as the ovirt_ca_bundle in case the ovirt_ca_bundle is not set on the ovirt.config file.

### DIFF
--- a/pkg/asset/manifests/openshift.go
+++ b/pkg/asset/manifests/openshift.go
@@ -3,8 +3,10 @@ package manifests
 import (
 	"context"
 	"encoding/base64"
+	"io/ioutil"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/ghodss/yaml"
 	"github.com/gophercloud/utils/openstack/clientconfig"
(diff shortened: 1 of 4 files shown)


Gal-Zaidman:
Added WIP since I haven't tested the change yet

pkg/asset/manifests/openshift.go
```diff
@@ -3,8 +3,10 @@ package manifests
 import (
 	"context"
 	"encoding/base64"
+	"io/ioutil"
```
dougsland, line 6: For new code, I would use `os` library instead of iotuil as it's deprecated in recent golang version
```
Deprecation of io/ioutil
The io/ioutil package has turned out to be a poorly defined and hard to understand collection of things. All functionality provided by the package has been moved to other packages. The io/ioutil package remains and will continue to work as before, but we encourage new code to use the new definitions in the io and os packages. Here is a list of the new locations of the names exported by io/ioutil:

Discard => io.Discard
NopCloser => io.NopCloser
ReadAll => io.ReadAll
ReadDir => os.ReadDir (note: returns a slice of os.DirEntry rather than a slice of fs.FileInfo)
ReadFile => os.ReadFile
TempDir => os.MkdirTemp
TempFile => os.CreateTemp
WriteFile => os.WriteFile
```

dougsland (changes requested):
The code looks sane but require changes IMO.

pkg/asset/manifests/openshift.go
Gal-Zaidman: Thanks for the note!
The problem is that the installer is 1.14 golang, so os doesn't have that function on the installer golang version :/
I guess once the installer will move to 1.16 this will need to be fixed everywhere

janoszen (approved):

PR #4810, opened 2021-04-01
 WIP: Bug 1945674: default to insecure false on user questions

Gal-Zaidman:
When the user uses questions most of our defaults are insecure true.
Which should never be the case.

### DIFF
--- a/pkg/asset/installconfig/ovirt/credentials.go
+++ b/pkg/asset/installconfig/ovirt/credentials.go
@@ -274,8 +274,8 @@ func engineSetup() (Config, error) {
 	}
 	logrus.Debug("engine FQDN: ", engineConfig.FQDN)
 
-	// By default, we set Insecure true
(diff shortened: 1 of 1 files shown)


dougsland:
looks sane from what I can see.

janoszen (approved):

janoszen:
@Gal-Zaidman it seems we dropped the ball on this one, can we pick it up?

sadasu:
@Gal-Zaidman is there still interest in this work?

sadasu:
@dougsland, @Gal-Zaidman how to we proceed with this?

dougsland:
@sadasu this patch seems WIP more than 1 year. Probably good candidate to be deprecated.

sadasu:
/close
Please see comment https://github.com/openshift/installer/pull/4810#issuecomment-1202831803

Issue #4797, opened 2021-03-26
coreos-bootimages releaseVersion not substituted


cgwalters:
Just launched `4.8.0-0.nightly-2021-03-26-123311` and I noticed that 
```
$ oc -n openshift-machine-config-operator get -o yaml configmap/coreos-bootimages | head
apiVersion: v1
data:
  releaseVersion: 0.0.1-snapshot
  stream: |
    {
      "stream": "rhcos-4.8",
      "metadata": {
        "last-modified": "2021-03-24T13:05:37Z"
      },
      "architectures": {
...
```

Not sure yet why `releaseVersion` isn't substituted.  It's the same approach as [this key in the MCO](https://github.com/openshift/machine-config-operator/blob/041b9665b013422ad1459e0f2f8c765aee073089/install/0000_80_machine-config-operator_05_osimageurl.yaml#L11).

Not urgent but would be good to fix because it will allow an operator to wait until the configmap synchronizes with the expected cluster version.

wking:
The `0.0.1-snapshot` replacement is something that [`oc adm release new ...` does to incoming manifests][1].  But #4760 should be injecting the new files as a CVO manifest.  So checking a recent nightly:

```console
$ oc adm release extract --to manifests registry.ci.openshift.org/ocp/release:4.8.0-0.nightly-2021-04-01-213116
$ wc manifests/0000_50_installer_coreos-bootimages.json
    0   678 24490 manifests/0000_50_installer_coreos-bootimages.json
$ cut -b-80 manifests/0000_50_installer_coreos-bootimages.json
{"kind":"ConfigMap","apiVersion":"v1","metadata":{"name":"coreos-bootimages","na
```

So it's a single line of compact JSON, and lacking `oc`'s required leading `([\W]|^)`.  Probably better to write the file as YAML, instead of dumping it as JSON?  That would give you the leading whitespace.

[1]: https://github.com/openshift/oc/blob/bdd8f89997ce224cb6f186716c18bee872d3b318/pkg/cli/admin/release/image_mapper.go#L328-L334

staebler:
Interesting. So it doesn't match the regex because it is wrapped in quotes. Seems easy enough to fix.

wking:
For reasons I don't understand, #4814 doesn't seem to have fixed this.  I've filed [OCPBUGS-19824](https://issues.redhat.com/browse/OCPBUGS-19824) to track.

PR #4814, opened 2021-04-02
hack: use yaml for coreos manifest

staebler:
Marshal the coreos-bootimages ConfigMap included in the installer manifests as yaml instead of json. This allows `oc` to replace
the `0.0.1-snapshot` value of the `releaseVersion`. The quotes around the value when marshalled as json were preventing the value from matching the regular expression that `oc` uses.

/fixes https://github.com/openshift/installer/issues/4797

### DIFF
--- a/hack/build-coreos-manifest.go
+++ b/hack/build-coreos-manifest.go
@@ -7,19 +7,19 @@
 package main
 
 import (
-	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 
+	"github.com/ghodss/yaml"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
 	streamJSON = "data/data/rhcos-stream.json"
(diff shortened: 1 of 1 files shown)


cgwalters:
OK.  Based on the issue it seems like we could also pretty print it?  

One downside of this is that the file won't be directly consumable from by our tools that process stream metadata today (they expect JSON) but that can be fixed relatively easily.  It's just `plume cosa2stream` that would need updating, and actually in the hopefully near future we'll be generating this metadata outside of the installer and all that we need to do is copy/paste it in here anyways.

staebler:
> OK. Based on the issue it seems like we could also pretty print it?
> 
> One downside of this is that the file won't be directly consumable from by our tools that process stream metadata today (they expect JSON) but that can be fixed relatively easily. It's just `plume cosa2stream` that would need updating, and actually in the hopefully near future we'll be generating this metadata outside of the installer and all that we need to do is copy/paste it in here anyways.

The value of the `stream` data is still json.

cgwalters:
Ah right duh, thanks!
Don't quote me on this but
/lgtm

Issue #4816, opened 2021-04-05
AWS install fails if controlPlane.Platform.AWS.IAMRole is missing from install-config


sallyom:
# Version

```console
$ openshift-install version
openshift-install 4.8.0-0.ci-2021-04-03-201542
built from commit af24457b32af35787a0ec84c3b2037877ef27a33
release image registry.ci.openshift.org/ocp/release@sha256:2bc95d8ed5de089aa88175132e641c894e00824ea30f47138f7bfdec490416b3
```

# Platform:
aws

# What happened?

The install-config created from `openshift-install create install-config` does not include or ask for a controlPlane.Platform.AWS.IAMRole.
When launching with the install-config, I'm getting a nil pointer dereference panic at this line:
https://github.com/openshift/installer/blob/master/pkg/asset/cluster/aws/aws.go#L93

# How to reproduce it (as minimally and precisely as possible)?
download latest CI-release and:

```console
$ openshift-install create install-config --dir testdir
$ openshift-install create cluster --dir testdir
```

staebler:
https://bugzilla.redhat.com/show_bug.cgi?id=1946119

patrickdillon:
/close
Fix has merged

PR #4817, opened 2021-04-05
update append iamRoleNames only if Platform.AWS not nil

sallyom:
I noticed this issue: https://github.com/openshift/installer/issues/4816 and this will resolve it, unless the fix should be to require the ControlPlane.Platform type to be populated in the install-config. 

Signed-off-by: Sally O'Malley <somalley@redhat.com>


staebler:
https://github.com/openshift/installer/pull/4818 fixes the same issue, and does so for the `.compute[].platform` as well.
/close

PR #4818, opened 2021-04-05
Bug 1946119: aws: support setting IAM role in the default AWS machine platform

staebler:
If the IAM role is set in the default AWS machine platform but not in the machine platform for a specific machine pool, then the IAM role from the default AWS machine platform should be used for that machine pool.

These changes also fix a bug where if the AWS machine platform were omitted from a particular machine pool, the installer would panic when trying to find IAM roles to tag.

https://bugzilla.redhat.com/show_bug.cgi?id=1946119

### DIFF
--- a/pkg/asset/cluster/aws/aws.go
+++ b/pkg/asset/cluster/aws/aws.go
@@ -86,16 +86,22 @@ func tagSubnetEC2Instances(ctx context.Context, clusterID string, installConfig 
 }
 
 func tagIamRoles(ctx context.Context, clusterID string, installConfig *installconfig.InstallConfig) error {
-	workerMachinePool := installConfig.Config.WorkerMachinePool()
-
 	var iamRoleNames []*string
-
-	if installConfig.Config.ControlPlane.Platform.AWS.IAMRole != "" {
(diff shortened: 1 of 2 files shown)


pkg/asset/cluster/aws/aws.go
```diff
@@ -125,3 +131,12 @@ func tagIamRoles(ctx context.Context, clusterID string, installConfig *installco
 func sharedTag(clusterID string) (string, string) {
 	return fmt.Sprintf("kubernetes.io/cluster/%s", clusterID), "shared"
 }
+
+func iamRoleForMachinePool(defaultAWS *awstypes.MachinePool, mp *types.MachinePool) string {
```
patrickdillon, line 135: I don't see this being used anywhere. Am I missing the use or for future use? or did it sneak in?

staebler: Left in from other iterations of this solution. I will remove. Thanks.

patrickdillon:
close #4816

patrickdillon:
oh to link the issue the keyword (e. g. fixes) must be kn the description or commit message

PR #4811, opened 2021-04-01
Test GCP versioned binary

e-tienne:
Testing to see if this can help unblock the GCP issue with yum.

PR #4819, opened 2021-04-05
Bug 1946307: gcp: install google cloud sdk with yum the recommended way

e-tienne:
Google recommends to use disable repository GPG key
checking in the yum repo configuration by setting
repo_gpgcheck=0. Yum repositories do not usually
use GPG keys for repository validation.
Instead, the https endpoint is trusted.


patrickdillon:
I think it would probably be better to explicitly revert 7ecdbcb2aa1e5d231a77d7db2fab172d07866d57 from #4811 in this PR. That is, this PR would be two commits: first the `git revert 7ecdbcb2aa1e5d231a77d7db2fab172d07866d57` and then setting `repo_gpgcheck=0`

staebler:
Images build. The `gcloud` tool is available. The gcp upi tests fail on signurl, but that was an already existing issue.

/lgtm
/approve

PR #4822, opened 2021-04-06
[WIP] terraform: Consolidate bootstrap resources with module count feature 

rna-afk:
Terraform 0.13 has a module count feature that can be set to instruct
terraform to create the exact number of copies of the module in the
cloud provider. This feature can be used to control the creation and
deletion of the bootstrap resources needed for cluster.

Adding the count field in all the bootstrap modules of all cloud
providers in order to make it easy to destroy bootstrap after
nodes have been created. This field would also make it easy to
destroy the resources without having to check for the cloud
platform and have a separate destroy logic for each of the platforms.


staebler:
This builds on https://github.com/openshift/installer/pull/4729
/hold

data/data/aws/main.tf
```diff
@@ -40,8 +40,10 @@ module "bootstrap" {
   vpc_security_group_ids   = [module.vpc.master_sg_id]
   volume_kms_key_id        = var.aws_master_root_volume_kms_key_id
   publish_strategy         = var.aws_publish_strategy
+  iam_role_name            = var.aws_master_iam_role_name
```
staebler, line 43: Remove this. This is coming from a commit that, while in master, is not included in this PR. Alternatively, you can rebase all master onto this branch.

data/data/config.tf
```diff
@@ -20,6 +20,18 @@ EOF
 
 }
 
+variable "bootstrap_count" {
```
staebler, line 23: Make this variable a bool and rename it to something like `bootstrapping`.

data/data/config.tf
```diff
@@ -20,6 +20,18 @@ EOF
...
+  type = string
+
+  default = "1"
+
+  description = <<EOF
+The number of bootstrap nodes to be created.
+This applies only to cloud platforms.
```
staebler, line 30: Is this statement accurate? Why does it only apply to cloud platforms?

pkg/destroy/bootstrap/bootstrap.go
```diff
@@ -36,13 +36,7 @@ func Destroy(dir string) (err error) {
...
 		return errors.Wrap(err, "failed to create temporary directory for Terraform execution")
 	}
-
-	tfKeep := strings.ToLower(os.Getenv("OPENSHIFT_INSTALL_KEEP_TERRAFORM"))
-	if tfKeep == "true" {
-		defer os.Rename(tempDir, fmt.Sprintf("%s-bootstrap", tempDir))
-	} else {
-		defer os.RemoveAll(tempDir)
-	}
+	defer os.RemoveAll(tempDir)
```
staebler: Remove this change.

data/data/config.tf
```diff
@@ -20,6 +20,18 @@ EOF
 
 }
 
+variable "bootstrap_count" {
+  type = string
+
+  default = "1"
```
staebler, line 26: Don't provide a default. Require that this variable be explicitly set.

staebler (commented):
We also want to get rid of the bespoke platform code [here](https://github.com/openshift/installer/blob/f1af1d2be4cd686ea2b8c027cb35caad316dd1d1/pkg/destroy/bootstrap/bootstrap.go#L57-L86).

data/data/config.tf
```diff
@@ -20,6 +20,16 @@ EOF
 
 }
 
+variable "bootstrapping" {
```
staebler, line 23: How are we setting this to "true" for the initial provisioning? Should we set the default value here?

staebler: Note that UPI installations do not use any of these terraform files.

staebler: So the alternative to setting the default here, is to set the bootstrapping variable explicitly at https://github.com/openshift/installer/blob/5c07d6ebfc793212418f7bc1ef840eec58579e8c/pkg/tfvars/tfvars.go#L40-L52.

staebler:
This looks good to me.

data/data/ovirt/template/main.tf
```diff
@@ -33,7 +33,7 @@ locals {
 
 // upload the disk if we don't have an existing template
 resource "ovirt_image_transfer" "releaseimage" {
-  count             = length(local.existing_id) == 0 ? 1 : 0
+  count             = length(local.existing_id) == 0 && var.bootstrapping ? 1 : 0
```
staebler, line 36: You need to pass in the `bootstrapping` variable to this module.

PR #4823, opened 2021-04-07
WIP: AzureStack Update Cloud Provider Config and Session Handling

patrickdillon:
Builds on #4799 so my work is in the last two commits

First commit rewires session handling to accept the ARM Endpoint (NOTE: this does not handle DNS migration, suggestions welcome. Do we need the migration in ash?)

Second adds Azure-stack specific logic for generating the cloud provider config, which is kind of hard to read:
```
$ cat ash-cp/manifests/cloud-provider-config.yaml 
apiVersion: v1
data:
  config: "{\n\t\"cloud\": \"AzureStackCloud\",\n\t\"tenantId\": \"<REDACTED>\",\n\t\"aadClientId\":
    \"<REDACTED>\",\n\t\"aadClientSecret\": \"<REDACTED>",\n\t\"aadClientCertPath\":
    \"\",\n\t\"aadClientCertPassword\": \"\",\n\t\"useManagedIdentityExtension\":
    false,\n\t\"userAssignedIdentityID\": \"\",\n\t\"subscriptionId\": \"<REDACTED>\",\n\t\"resourceGroup\":
    \"padillon-t6lbg-rg\",\n\t\"location\": \"ppe3\",\n\t\"vnetName\": \"padillon-t6lbg-vnet\",\n\t\"vnetResourceGroup\":
    \"padillon-t6lbg-rg\",\n\t\"subnetName\": \"padillon-t6lbg-worker-subnet\",\n\t\"securityGroupName\":
    \"padillon-t6lbg-nsg\",\n\t\"routeTableName\": \"padillon-t6lbg-node-routetable\",\n\t\"primaryAvailabilitySetName\":
    \"\",\n\t\"vmType\": \"\",\n\t\"primaryScaleSetName\": \"\",\n\t\"cloudProviderBackoff\":
    true,\n\t\"cloudProviderBackoffRetries\": 0,\n\t\"cloudProviderBackoffExponent\":
    0,\n\t\"cloudProviderBackoffDuration\": 6,\n\t\"cloudProviderBackoffJitter\":
    0,\n\t\"cloudProviderRateLimit\": false,\n\t\"cloudProviderRateLimitQPS\": 0,\n\t\"cloudProviderRateLimitBucket\":
    0,\n\t\"cloudProviderRateLimitQPSWrite\": 0,\n\t\"cloudProviderRateLimitBucketWrite\":
    0,\n\t\"useInstanceMetadata\": false,\n\t\"loadBalancerSku\": \"basic\",\n\t\"excludeMasterFromStandardLB\":
    null,\n\t\"disableOutboundSNAT\": null,\n\t\"maximumLoadBalancerRuleCount\": 0\n}\n"
kind: ConfigMap
metadata:
  creationTimestamp: null
  name: cloud-provider-config
  namespace: openshift-config
```
I will work on adding the endpoints file next.


staebler:
> NOTE: this does not handle DNS migration, suggestions welcome. Do we need the migration in ash?

The DNS migration is not necessary for ash. That was to migrate clusters created before we used private DNS.

pkg/asset/installconfig/azure/session.go
```diff
@@ -42,27 +42,23 @@ type Credentials struct {
...
 	var env azureenv.Environment
 	var err error
 	switch cloudName {
 	case azure.StackCloud:
-		envFilepath := os.Getenv(azureenv.EnvironmentFilepathName)
-		if envFilepath == "" {
-			return nil, errors.Errorf("must set the %q environment variable for the Azure Stack cloud", azureenv.EnvironmentFilepathName)
-		}
-		env, err = azureenv.EnvironmentFromFile(envFilepath)
+		env, err = azureenv.EnvironmentFromURL(armEndpoint)
```
staebler, line 61: Oh, that is much easier. Nice.

pkg/migrate/azure/privatedns.go
```diff
@@ -484,7 +484,7 @@ func (client *privateDNSClient) migrateLegacyZone(legacyDNSZone *legacyDNSZone,
 
 // Migrate does a migration from a legacy zone to a private zone
 func Migrate(cloudName azure.CloudEnvironment, resourceGroup string, migrateZone string, virtualNetwork string, vnetResourceGroup string, link bool) error {
-	session, err := azconfig.GetSession(cloudName)
+	session, err := azconfig.GetSession(cloudName, "") //TODO pass in ARMEndpoint (or determine if needed)
```
staebler, line 487: The ARM endpoint is not needed.

patrickdillon:
> > NOTE: this does not handle DNS migration, suggestions welcome. Do we need the migration in ash?
> 
> The DNS migration is not necessary for ash. That was to migrate clusters created before we used private DNS.

So if we want to stick with this solution we can simply use an empty string in the DNS migration, which solves that problem in this PR.

pkg/asset/manifests/azure/cloudproviderconfig.go
```diff
@@ -24,19 +26,32 @@ type CloudProviderConfig struct {
 // JSON generates the cloud provider json config for the azure platform.
 // managed resource names are matching the convention defined by capz
 func (params CloudProviderConfig) JSON() (string, error) {
+	azureStack := params.CloudName.Name() == string(azure.StackCloud)
```
staebler, line 29: Does this need to be this complicated?
```suggestion
	azureStack := params.CloudName == azure.StackCloud
```

pkg/asset/manifests/azure/cloudproviderconfig.go
```diff
@@ -51,10 +66,10 @@ func (params CloudProviderConfig) JSON() (string, error) {
 		CloudProviderBackoff:         true,
 		CloudProviderBackoffDuration: 6,
 
-		UseInstanceMetadata: true,
+		UseInstanceMetadata: !azureStack,
```
staebler, line 69: I think I would prefer to keep all of the changes specific to Azure Stack in one `if` block. I find that makes it a bit easier to see exactly what the differences are for Azure Stack.
```
config := config{
   // same code we have in master now
}
if params.CloudName == azure.StackCloud {
	config.ac.UseManagedIdentityExtension = false
	config.ac.AADClientID = params.ClientID
	config.ac.AADClientSecret = params.ClientSecret
	config.UseInstanceMetadata = false
	config.LoadBalancerSku = "basic"
}
```

patrickdillon:
Maybe we should treat Azure credentials for different clouds differently. We save them all in `osServicePrincipal.json` but perhaps they should be able to coexist. e.g. `osServicePrincipalStack.json`

staebler:
> Maybe we should treat Azure credentials for different clouds differently. We save them all in `osServicePrincipal.json` but perhaps they should be able to coexist. e.g. `osServicePrincipalStack.json`

We don't do that for other clouds. I would not expect a user to use the same system to install one cluster to public Azure and another cluster to Azure stack.

pkg/asset/manifests/cloudproviderconfig.go
```diff
@@ -36,6 +37,7 @@ var (
 const (
 	cloudProviderConfigDataKey         = "config"
 	cloudProviderConfigCABundleDataKey = "ca-bundle.pem"
+	cloudProviderEndpointsKey          = "endpoints"
```
patrickdillon, line 40: I wonder if it would be better to set the key to `azurestackcloud.json`

pkg/asset/machines/worker.go
```diff
@@ -295,7 +295,10 @@ func (w *Worker) Generate(dependencies asset.Parents) error {
 			}
 		case azuretypes.Name:
 			mpool := defaultAzureMachinePoolPlatform()
-			mpool.InstanceType = azuredefaults.ComputeInstanceType(installConfig.Config.Platform.Azure.Region)
+			mpool.InstanceType = azuredefaults.ControlPlaneInstanceType(
```
staebler, line 298: Should this be using `ComputeInstanceType`?

patrickdillon: yes will fix it now.

patrickdillon:
needs rebase

PR #4126, opened 2020-09-02
Bug 1871712: pkg/destroy/aws: Detach attached policies during destroy

mtnbikenc:
Managed policies could be attached after initial installation. In order
to delete the IAM roles created during install, these custom managed
polices should be detached allowing the role to be deleted.

PR #4825, opened 2021-04-07
Bug 1938131: pkg/asset/installconfig/aws: Add iam permission for destorying clusters

mtnbikenc:
iam:ListAttachedRolePolicies

Permission required since #4126


pkg/asset/installconfig/aws/permissions.go
```diff
@@ -118,6 +118,7 @@ var permissions = map[PermissionGroup][]string{
 		"iam:GetRole",
 		"iam:GetRolePolicy",
 		"iam:GetUser",
+		"iam:ListAttachedRolePolicies",
```
staebler, line 121: This should go in `PermissionDeleteBase` instead of `PermissionCreateBase`.

mtnbikenc: Updated.

staebler (approved):
/lgtm

PR #4828, opened 2021-04-07
Bug 1921627: Shorten instance group suffix to ig

rna-afk:
GCP has a size restriction of 63 for the instance group name which
is mostly taken up by the suffix -instance-group that is being added
to make sure the resources have unique name. Reducing the size
of the suffix from -instance-group to -ig would help in restricting
the size of the name and would also help in keeping the names
unique.

### DIFF
--- a/docs/user/gcp/install_upi.md
+++ b/docs/user/gcp/install_upi.md
@@ -549,13 +549,13 @@ Manager, so we must add the bootstrap node manually.
 ### Add bootstrap instance to internal load balancer instance group
 
 ```sh
-gcloud compute instance-groups unmanaged add-instances ${INFRA_ID}-bootstrap-instance-group --zone=${ZONE_0} --instances=${INFRA_ID}-bootstrap
+gcloud compute instance-groups unmanaged add-instances ${INFRA_ID}-bootstrap-ig --zone=${ZONE_0} --instances=${INFRA_ID}-bootstrap
 ```
 
 ### Add bootstrap instance group to the internal load balancer backend service
 
 ```sh
-gcloud compute backend-services add-backend ${INFRA_ID}-api-internal-backend-service --region=${REGION} --instance-group=${INFRA_ID}-bootstrap-instance-group --instance-group-zone=${ZONE_0}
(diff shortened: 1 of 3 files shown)


patrickdillon:
You will want a corresponding PR for CI. The CI script follows the UPI docs and, I believe, utilizes these Python scripts. So this is probably actually breaking CI--hard to tell as it looks like the job is broken.

staebler:
> You will want a corresponding PR for CI. The CI script follows the UPI docs and, I believe, utilizes these Python scripts. So this is probably actually breaking CI--hard to tell as it looks like the job is broken.

@e-tienne is looking at fixing the e2e-gcp-upi tests. So let's wait for some progress on that effort, since the changes in this PR are not strictly blocking anything.
/hold

staebler:
The e2e-gcp-upi tests should succeed now.
/test e2e-gcp-upi

rna-afk:
e2e-gcp-upi is failing due to a hardcoded value here. It is trying to check for the instance group with the old suffix in the name.
https://github.com/openshift/release/blob/da2660ba2ae538ac611e34e1d7119d8f815f1f28/ci-operator/step-registry/upi/install/gcp/upi-install-gcp-commands.sh#L342

jstuever:
This will need be configured first in CI, which must be backwards compatible. For this, I would test if the name in the template is using the new or old scheme, and run different commands based on that.

https://github.com/openshift/release/blob/master/ci-operator/step-registry/upi/install/gcp/upi-install-gcp-commands.sh

jstuever (approved):
/approve

rna-afk:
Corresponding fix for the e2e test is here 
https://github.com/openshift/release/pull/18302
/unhold

tsze-redhat:
Looks like a whole bunch of additional work will need to be done to complement this:
- user doc updates (e.g. https://docs.openshift.com/container-platform/4.8/installing/installing_gcp/installing-gcp-user-infra.html)
- our automation scripts

How do we communicate with customers that their existing automation may not work after the change?

staebler:
CI changes have merged.
/test e2e-gcp

staebler:
The e2e-gcp-upi test was successful.
/lgtm
/approve

PR #4831, opened 2021-04-08
Revert "Dockerfile.upi.ci: Drop pip+pyopenssl installs"

e-tienne:
This reverts commit 9aeadb0ccef278538bf007afd0c0b405722c7ce9.

- pyopenssl is still required to use signurl with gcloud
- we still end up using python2 with the image used by prow
- the ssl certificate for files.pythonhosted.org appears to work now


e-tienne:
/retitle Revert "Dockerfile.upi.ci: Drop pip+pyopenssl installs

staebler (approved):
/lgtm

PR #4833, opened 2021-04-08
doc/openstack: add a note about image registry in AZ

EmilienM:
Add a note that if we deploy a cluster in an OpenStack AZ, it's
suggested to use Cinder backend if the service is available in this AZ,
since Cinder is topology aware.
Swift isn't deployed in AZs usually, so traffic would have to go
through the link between sites, which isn't optimal in real world.

### DIFF
--- a/docs/user/openstack/README.md
+++ b/docs/user/openstack/README.md
@@ -112,6 +112,14 @@ openstack role add --user <user> --project <project> swiftoperator
 
 If Swift is not available, the [PVC](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims) storage is used as the backend. For this purpose, a persistent volume of 100 GB will be created in Cinder and mounted to the image registry pod during the installation.
 
+**Note:** If you are deploying a cluster in an Availability Zone where Swift isn't available but where Cinder is, it is recommended to deploy the Image Registry with Cinder backend. It will try to schedule the volume into the same AZ as the Nova zone where the PVC is located; otherwise it'll pick the default availability zone. If needed, the Image registry can be moved to another availability zone by a day 2 operation.
+
+If you want to force Cinder to be used as a backend for the Image Registry, you need to remove the `swiftoperator` permissions. As an OpenStack administrator:
+
+```sh
+openstack role remove --user <user> --project <project> swiftoperator
+```
+
 **Note:** Since Cinder supports only [ReadWriteOnce](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) access mode, it's not possible to have more than one replica of the image registry pod.
 
 ### Disk Requirements


docs/user/openstack/README.md
```diff
@@ -104,6 +104,8 @@ The bootstrap node is a temporary node that is responsible for standing up the c
 
 If Swift is available in the cloud where the installation is being performed, it is used as the default backend for the OpenShift image registry. At the time of installation only an empty container is created without loading any data. Later on, for the system to work properly, you need to have enough free space to store the container images.
 
+**Note:** If you are deploying a cluster in an Availability Zone where Swift isn't available, but where Cinder is, it is highly suggested to deploy the Image Registry in Cinder, which is aware of the topology and will schedule the volumes into the right zone.
```
mandre, line 107: Can we move the note below, after we've stated that Cinder is the fallback when Swift is not available? I think where it's placed now it breaks the flow. For example the line below says "In this case" which could now refer to when swift is available or not, depending if you read the note.

I would also perhaps add a blurb about how to explicitly chose cinder (remove the `swiftoperator` permission).

Also, have we confirmed that the image-registry volume is always created in the same zone as the nodes, when only one zone is specified? If not we should probably add instruction to move the image-registry on day2. Doesn't have to be in this PR.

mandre (commented):
@maxwelldb, FYI, a small note stating that cinder can be a better choice for the image-registry in certain situations (mostly for edge deployments).

@EricArrakis this could actually be one of this few cases where it is beneficial for the installer that it is deploying at the edge. That being said, I still think we do not (yet) want to bake that logic in the installer.

docs/user/openstack/README.md
EmilienM: ack, will do

EmilienM: I modified the content a bit but I want a review from @Fedosin to confirm that the Image Registry will always request a volume in the right AZ, with proper technical explanation.

EmilienM: I got it from @gnufied:
```
a pod with pvc is scheduled in nova zone - us-east-1c, then volume is also created with same AZ. it is expected that cinder and nova zones match at that point
```

mandre (commented):
/lgtm
/approve

maxwelldb:
@mandre 4.8+ or all supported versions? 

Additionally, is there BZ for this?

mandre:
> @mandre 4.8+ or all supported versions?

That would be 4.8+ since it assumes support for cinder AZs.

> Additionally, is there BZ for this?

Sorry, there's not. We should have created one.

maxwelldb:
@mandre TY. And to be clear, since I see "it is recommended" in this PR, is the content in this note still fully supported according to QE?

mandre:
> @mandre TY. And to be clear, since I see "it is recommended" in this PR, is the content in this note still fully supported according to QE?

No, to my understanding we're not going to support a deployment in an AZ using swift for the image registry when swift isn't available in the same AZ. 
@EmilienM to keep me honest.

PR #4838, opened 2021-04-09
WIP: build upi image without specific pip version

e-tienne:
"RUN easy_install 'pip<21'" is failing.

Let's just install pip without specifying the version and see if our tests fail.


e-tienne:
/close

confirmed failure with pip 21

PR #4843, opened 2021-04-12
azure: don't use managed identity on ARO

m1kola:
* Adds a new `aro` build tag which is going to be used by ARO when vendoring the installer
* Disables managed identity for ARO


staebler:
This comes in timely for us, @m1kola. We need this same logic for Azure Stack, which does not have support for user-provided managed identities.

m1kola:
@staebler Is there anything I should take into account while working on this PR? Maybe you want me to change something in it (like naming)?

Or should I go ahead and do what we need for ARO and later you refactor the codebase to accommodate Azure Stack?

staebler:
> @staebler Is there anything I should take into account while working on this PR? Maybe you want me to change something in it (like naming)?
> 
> Or should I go ahead and do what we need for ARO and later you refactor the codebase to accommodate Azure Stack?

I haven't dug into the PR yet to offer any opinion on naming and such. My inclination is to have you do what you need to ARO and we based the Azure Stack work around that, making any tweaks that we need.

m1kola:
> My inclination is to have you do what you need to ARO and we based the Azure Stack work around that, making any tweaks that we need.

Sounds good to me. 

---

Will look into CI failures next (tomorrow hopefully). Doesn't look like they are related to PR.

/retest

pkg/asset/manifests/cloudproviderconfig.go
```diff
@@ -30,7 +32,10 @@ import (
 )
 
 var (
-	cloudProviderConfigFileName = filepath.Join(manifestDir, "cloud-provider-config.yaml")
+	cloudProviderConfigFileName         = filepath.Join(manifestDir, "cloud-provider-config.yaml")
+	aroCloudProviderRoleFileName        = filepath.Join(manifestDir, "aro-cloud-provider-secret-reader-role.yaml")
+	aroCloudProviderRoleBindingFileName = filepath.Join(manifestDir, "aro-cloud-provider-secret-reader-rolebinding.yaml")
+	aroCloudProviderSecretFileName      = filepath.Join(manifestDir, "aro-cloud-provider-secret.yaml")
```
patrickdillon, line 38: Cluster secrets and the secret readers are already created here: https://github.com/openshift/installer/blob/master/pkg/asset/manifests/openshift.go
with the templates here:
https://github.com/openshift/installer/tree/master/data/data/manifests/openshift

Wouldn't it make more sense to move this logic there?

patrickdillon (commented):
I took an initial look and left some general comments. 

As @staebler already mentioned, we need the logic for providing the clientID & secret in the cloud provider config for Azure Stack Hub support, so this PR is very helpful. 

As my comment says, we already have logic for creating the secret and secret reader, so we should try to move the logic in this PR out of the cloud provider.

Have you considered any of the changes that would be required from the Terraform code, for example where we create user-assigned roles: https://github.com/openshift/installer/blob/master/data/data/azure/main.tf#L151-L170?

m1kola (commented):
@patrickdillon thanks for the review. I'm looking into updating the PR. I left some responses to your comments below.

> Have you considered any of the changes that would be required from the Terraform code, for example where we create user-assigned roles: https://github.com/openshift/installer/blob/master/data/data/azure/main.tf#L151-L170?

On ARO we do not use terraform code, so my plan is to introduce the changes we need for ARO and try not to break rest of the installer in the process :)

pkg/asset/manifests/cloudproviderconfig.go
patrickdillon: Apologies, but my first link was incorrect. The logic for generating those templates is here: https://github.com/openshift/installer/tree/master/pkg/asset/templates/content/openshift
 
I do think we should use this pattern of generating the resources from a template, but I wonder if it makes more sense to keep that logic in the cloud provider config as you originally had it or to move it to `templates`. @staebler what do you think?

pkg/asset/manifests/openshift.go
```diff
@@ -309,3 +349,81 @@ func (o *Openshift) Load(f asset.FileFetcher) (bool, error) {
...
+}
+
+func aroRoleBinding(icazure.Credentials) ([]byte, error) {
+	return yaml.Marshal(&rbacv1.RoleBinding{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "RoleBinding",
+			APIVersion: "rbac.authorization.k8s.io/v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "aro-cloud-provider-secret-read",
```
patrickdillon, line 381: Should this name match the name from the [docs](https://kubernetes-sigs.github.io/cloud-provider-azure/install/configs/):

`name: system:azure-cloud-provider-secret-getter`

pkg/asset/manifests/openshift.go
```diff
@@ -309,3 +349,81 @@ func (o *Openshift) Load(f asset.FileFetcher) (bool, error) {
...
+			{
+				Kind:      "ServiceAccount",
+				Name:      "azure-cloud-provider",
+				Namespace: "kube-system",
+			},
+		},
+		RoleRef: rbacv1.RoleRef{
+			APIGroup: "rbac.authorization.k8s.io",
+			Kind:     "Role",
+			Name:     "aro-cloud-provider-secret-reader",
```
patrickdillon, line 394: similar comment as above. Also, keep in mind with `aro` specific naming that we will be using this logic for azure stack so it would be good to make it more general

patrickdillon (commented):
I am on PTO  today so just took a quick look. I can try to help with this more tomrrow or later this week.

m1kola:
@patrickdillon @staebler please take another look.

I updated naming and replaced yaml mrashaling with templates as requested. I left generation in `pkg/asset/manifests/openshift.go` for now, but I'm happy to move it back to `pkg/asset/manifests/cloudproviderconfig.go` as in the original version of this PR, if you think it is more appropriate place for it.

pkg/asset/manifests/openshift.go
```diff
@@ -69,6 +69,9 @@ func (o *Openshift) Dependencies() []asset.Asset {
 		&openshift.PrivateClusterOutbound{},
 		&openshift.BaremetalConfig{},
 		new(rhcos.Image),
+		&openshift.AzureCloudProviderSecret{},
```
m1kola, line 72: @patrickdillon an issue with the appraoch of using templates is this: we have to declare templates as dependencies and store will call `Generate()` on them even if we don't need them (e.g. non-ARO builds).

What do you think? Is it something we should worry about?

We can probably use build tags and create two extra files in this package:

1. One with `// +build aro` where we include them as dependencies
2. One with `// +build !aro` where we do not include them

But I don't want to rely on build tags too much in this case:

1. It will make the code harder to follow
2. It will probably be unnecessary once you implement Azure Stack Hub support

patrickdillon: If we're going with the template approach, I was thinking to just add them to the existing templates. Here's a simplified example: https://github.com/openshift/installer/compare/master...patrickdillon:azure-merge-cloud-config We would need to add a boolean field to `AzureCredsSecretData` to switch this behavior on or off depending on ASH/ARO. 

Also, not sure if it is valid to have the values of the cloud provider config secret be base64 encoded. I assume for the existing code, the cloud credential operator knows to base64 decode the values, but I assume that is up to the client. I don't see anything in the azure docs about this. If we don't know, we can test and see.

I would like to get @staebler's feedback about whether this is the right direction with templates. To summarize the context: In the case of ARO & ASH we need to create a secret (with the creds) , role, and role binding. I see a few options:

1. Use the template approach, like in my [example above](https://github.com/openshift/installer/compare/master...patrickdillon:azure-merge-cloud-config )
2. Generate all 3 resources in the cloud provider config, which was the original state of this PR
3. Some combination of the above
4. Generate the resource manifests somewhere else entirely

I'm pretty neutral on these options with a slight leaning toward option 1.

m1kola: @patrickdillon thanks for the feedback. Replied to some of the comments.

> If we're going with the template approach, I was thinking to just add them to the existing templates. Here's a simplified example: [master...patrickdillon:azure-merge-cloud-config](https://github.com/openshift/installer/compare/master...patrickdillon:azure-merge-cloud-config) We would need to add a boolean field to `AzureCredsSecretData` to switch this behavior on or off depending on ASH/ARO.

I was thinking about reusing `data/data/manifests/openshift/cloud-creds-secret.yaml.template`, but it is already overloaded with logic and I didn't want to make it worse. Happy to do this if you think it is a better appraoch.

> Also, not sure if it is valid to have the values of the cloud provider config secret be base64 encoded. I assume for the existing code, the cloud credential operator knows to base64 decode the values, but I assume that is up to the client. I don't see anything in the azure docs about this. If we don't know, we can test and see.

We must encode values into base64 manually here because:
* We add them into [`Secret`](https://kubernetes.io/docs/concepts/configuration/secret/#overview-of-secrets)'s `data`:
  > The values for all keys in the `data` field have to be base64-encoded strings.
* We use templates. Normaly when marshaling into json/yaml from a go struct it will be done implicitly due to the fact that `data` is defined as follows:

  ```go
  Data map[string][]byte `json:"data,omitempty" protobuf:"bytes,2,rep,name=data"`
  ```

  And [`[]byte` in go marshaled](https://golang.org/pkg/encoding/json/#Marshal) as base64-encoded string.

  > Array and slice values encode as JSON arrays, except that []byte encodes as a base64-encoded string, and a nil slice encodes as the null JSON value. 


On the receiving end, the cloud provider doesn't have to decode the value explicilty [here](https://github.com/openshift/origin/blob/90c050f5afb4c52ace82b15e126efe98fa798d88/vendor/k8s.io/legacy-cloud-providers/azure/azure_config.go#L77): go will do it behind the scenes.

patrickdillon: > I was thinking about reusing data/data/manifests/openshift/cloud-creds-secret.yaml.template, but it is already overloaded with logic and I didn't want to make it worse. Happy to do this if you think it is a better appraoch.

As I look at this more closely I do think this current approach is on the right track. I think the approach should very closely follow the pattern here: 
https://github.com/openshift/installer/blob/master/pkg/asset/manifests/openshift.go#L253-L259

> we have to declare templates as dependencies and store will call Generate() on them even if we don't need them (e.g. non-ARO builds).

So yes this is fine to have a dependency generated even when it is not used. The above example is only for private clusters, but is generated for all clusters. But you should consolidate all of these Azure Cloud Provider dependencies into a single asset. Note that the one asset could have many templates; so you could either have a separate template for each resource or combine the three resources in a single template. 

If this is unclear or your have questions let me know. I think this is the right drection!

Also, thanks for the detailed, awesome explanation of why secret values are b64 encoded. I didn't know that!

m1kola: @patrickdillon I left 3 template files, but consolidated them into one assset. Now the naming is little bit tricky: I called it `AzureCloudProviderSecret`, but it contains the role, the role binding as well as the secret which might be a little bit confusing. Suggestions are welcome :)

I'm also happy to put all yamls into one file, if you like. Having in separate templates/manifests is a tiny bit better for visibility, IMO. But I have no strong prefrences.

pkg/asset/manifests/openshift.go
```diff
@@ -258,6 +259,38 @@ func (o *Openshift) Generate(dependencies asset.Parents) error {
 		assetData["99_private-cluster-outbound-service.yaml"] = applyTemplateData(privateClusterOutbound.Files()[0].Data, templateData)
 	}
 
+	if installConfig.Config.Azure.IsARO() {
```
patrickdillon, line 262: I think we need to check `installConfig.Config.Azure` for nil.

pkg/asset/manifests/openshift.go
patrickdillon: I think this is looking good. Have you tested this? I hope to test it soon with Azure Stack.

> I called it AzureCloudProviderSecret... Suggestions are welcome :)

Maybe `AzureMergedCloudConfig`?

pkg/asset/manifests/openshift.go
m1kola: Good catch! Updated.

pkg/asset/manifests/openshift.go
m1kola: @patrickdillon

> Have you tested this?

I did `openshift-install create manifests` with `aro` build tag and manifests look good. I alos did it with a binary compiled without `aro` build tag and relevant manifests were not present in the output directory, as expected.

And I'm pretty confident that it will work with ARO because I tested it with older version of this PR (we did not change outputs significantly in recent changes). I'll ARO integration tomorrow, but it should not block this PR.


> Maybe `AzureMergedCloudConfig`?

I was wonder if we can better reflect the fact that it is not only secret, but role and role binding as well.
Let's probably keep it as `AzureCloudProviderSecret` as it consists of resource name `azure-cloud-provider` and resource type `secret`. Resource name is [forced on us by cloud provider](https://kubernetes-sigs.github.io/cloud-provider-azure/install/configs/#setting-azure-cloud-provider-from-kubernetes-secrets).

patrickdillon (commented):
> /retest

It looks like the relevant and required tests are passing so no need to `/retest` the rest of the tests unless you are concerned about a particular test.

patrickdillon:
> > /retest
> 
> It looks like the relevant and required tests are passing so no need to `/retest` the rest of the tests unless you are concerned about a particular test.

It makes sense to test the azure upi and shared box but looking at the failures in those tests the cluster is installing successfully and the test failures look like flakes to me.

m1kola:
@patrickdillon thanks for looking into these failuers. I also think these failures are just flakes. I just want to have as much green jobs as possible: gives me more confidence that I didn't break something unobvious in an unfamiliar repo :)

pkg/types/azure/platform_aro.go
```diff
@@ -0,0 +1,7 @@
+// +build aro
```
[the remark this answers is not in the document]
m1kola: I decided to split the build tag bit into a separate PR: https://github.com/openshift/installer/pull/4874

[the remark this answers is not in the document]
m1kola: Clarification: I'm happy for the build tag to be merged as part of this PR. But if there is still feedback related to managed identity - I suggest that we merge the build tag as part of #4874 as I have few more things that depend on it.

pkg/asset/manifests/openshift.go
```diff
@@ -258,6 +259,38 @@ func (o *Openshift) Generate(dependencies asset.Parents) error {
 		assetData["99_private-cluster-outbound-service.yaml"] = applyTemplateData(privateClusterOutbound.Files()[0].Data, templateData)
 	}
 
+	if installConfig.Config.Azure != nil && installConfig.Config.Azure.IsARO() {
```
patrickdillon, line 262: nit: it looks like the other conditionals are switching on `platform` so perhaps for consistency it might be better to do:

```go
if platform == azuretypes.Name && installConfig.Config.Azure.IsARO()
```

pkg/asset/manifests/openshift.go
staebler: > I'm also happy to put all yamls into one file, if you like. Having in separate templates/manifests is a tiny bit better for visibility, IMO. But I have no strong prefrences.

Leave it as one file for each resource. The CVO has (or had) issues with multiple resources in a single manifest when there are problems applying some of the resources.

staebler: > We must encode values into base64 manually here because:
>
>    We add them into Secret's data:

Use `stringData` instead of `data`. Then you don't need to do the base-64 encoding. The API server will do the base-64 encoding for you.

pkg/asset/machines/azure/machines.go
```diff
@@ -127,7 +127,13 @@ func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string
 		ResourceGroup:        rg,
 		NetworkResourceGroup: networkResourceGroup,
 		PublicLoadBalancer:   publicLB,
-	}, nil
+	}
+
+	if platform.IsARO() {
+		spec.ManagedIdentity = ""
```
staebler, line 133: Use a `managedIdentity` variable to populate the `ManagedIdentity` field, rather than clearing out the field after creating the spec.

pkg/asset/manifests/azure/cloudproviderconfig.go
```diff
@@ -56,6 +57,11 @@ func (params CloudProviderConfig) JSON() (string, error) {
 		// https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-tcp-reset
 		LoadBalancerSku: "standard",
 	}
+
+	if params.ARO {
+		config.authConfig.UseManagedIdentityExtension = false
```
staebler, line 62: Use a `useManagedIdentity` variable to populate the `UseManagedIdentityExtension` field.

staebler (commented):
This looks reasonable to me.

pkg/asset/manifests/openshift.go
m1kola: I didn't know that `stringData` merges into `data`.

@staebler I updated the PR to use `stringData`. Note that everything else in `pkg/asset/manifests/openshift.go` explicitly encodes into base64. I'm happy to revert back to explicit encoding for consistency, if you wish.

m1kola:
@staebler I updated the PR to address your feedback. Pleaes take another look.

---

/test e2e-azure
/test e2e-azure-upi
/test e2e-azure-shared-vpc

m1kola:
Quote issues

/retest

m1kola:
Quote issues on Azure

/retest

m1kola:
Looks like flakes in e2e-azure. Will try to retest.

/retest

PR #4846, opened 2021-04-13
Bug 1948923: aws: improve ami validation

staebler:
Improve the validation around whether a custom AMI has been provided by the user in the regions where one is required.

The install config is valid if any of the following are satisifed.
* The RHCOS stream metadata has an AMI for the region.
* The region is in the standard AWS partition. In this caes, the AMI can be copied from us-east-1.
* The custom AMI has been specified at the platform level.
* The custom AMI has been specified in the default machine platform.
* A custom AMI has been specified for each machine pool that has replicas.

### DIFF
--- a/pkg/asset/installconfig/aws/validation.go
+++ b/pkg/asset/installconfig/aws/validation.go
@@ -44,6 +44,7 @@ func Validate(ctx context.Context, meta *Metadata, config *types.InstallConfig) 
 	if config.Platform.AWS == nil {
 		return errors.New(field.Required(field.NewPath("platform", "aws"), "AWS validation requires an AWS platform configuration").Error())
 	}
+	allErrs = append(allErrs, validateAMI(ctx, config)...)
 	allErrs = append(allErrs, validatePlatform(ctx, meta, field.NewPath("platform", "aws"), config.Platform.AWS, config.Networking, config.Publish)...)
 
 	if config.ControlPlane != nil && config.ControlPlane.Platform.AWS != nil {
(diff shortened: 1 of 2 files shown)


pkg/asset/installconfig/aws/validation.go
```diff
@@ -75,6 +72,53 @@ func validatePlatform(ctx context.Context, meta *Metadata, fldPath *field.Path,
...
+	controlPlaneHasAMISpecified := false
+	if config.ControlPlane != nil && config.ControlPlane.Platform.AWS != nil {
+		controlPlaneHasAMISpecified = config.ControlPlane.Platform.AWS.AMIID != ""
+	}
+	computesHaveAMISpecified := true
+	for _, c := range config.Compute {
+		if c.Replicas != nil && *c.Replicas == 0 {
+			continue
+		}
+		if c.Platform.AWS != nil && c.Platform.AWS.AMIID == "" {
```
jstuever, line 103: Should this be `c.Platform.AWS == nil ||`?

pkg/asset/installconfig/aws/validation_test.go
```diff
@@ -472,7 +566,9 @@ func TestValidate(t *testing.T) {
 			if test.expectErr == "" {
 				assert.NoError(t, err)
 			} else {
-				assert.Regexp(t, test.expectErr, err.Error())
+				if assert.Error(t, err) {
+					assert.Regexp(t, test.expectErr, err.Error())
+				}
```
jstuever, line 571: Can we flatten this... `} else if {`

jstuever (changes requested):
I believe you have logic issue that needs to be sorted.

pkg/asset/installconfig/aws/validation.go
staebler: No, I don't think so. The test here is to find if there is a compute machine pool which explicitly has an AMI specified. If `c.Platform.AWS` is nil, then the compute machine pool does not have an AMI specified.

pkg/asset/installconfig/aws/validation_test.go
staebler: I would prefer not to. The current layout makes a distinction between the codepath where we *are not* expecting an error and the codepath where we *are* expecting an error. If we flatten to an `else if`, that makes the two codepaths are instead (1) we are not expecting an error and (2) we are expecting an error and we actually got an error.
```
if test.expectErr == "" {
   // assert that we did not get an error
} else {
   // assert that we got an error and that the error matches what we expected
}
```
versus
```
if test.expectErr == "" {
   // assert that we did not get an error
} else if assert.Error(t, err) {
   // assert that the error that we got matches what we expected
}
```

pkg/asset/installconfig/aws/validation.go
staebler: Looking at this again, you are correct. The logic should be as you stated instead of what I have here. I will make the change.

jstuever (approved):
/lgtm
/approve

PR #4848, opened 2021-04-13
Bug 1957951: AWS: Periodically send shut down requests for stuck EC2 instances

rna-afk:
During AWS destroy cluster sequence, the instances are requested
to be shut down but they do not get terminated for a long time.
The installer ignores all the EC2 instances that are in "shut-down"
status as they take a long time to complete but sometimes these
instances are stuck at the same status for a long time.

A fresh termination request has a more better chance at destroying
these stuck instances and hence changing the code to send the
termination request even for the instances in "shut-down" status.

### DIFF
--- a/pkg/destroy/aws/aws.go
+++ b/pkg/destroy/aws/aws.go
@@ -185,18 +185,24 @@ func (o *ClusterUninstaller) RunWithContext(ctx context.Context) ([]string, erro
 	// running on the cluster creating new resources while we are attempting to delete resources, which could leak
 	// the new resources.
 	ec2Client := ec2.New(awsSession)
+	lastTerminateTime := time.Now()
 	err = wait.PollImmediateUntil(
 		time.Second*10,
 		func() (done bool, err error) {
-			instancesToDelete, err := o.findEC2Instances(ctx, ec2Client, deleted)
+			instancesRunning, instancesNotTerminated, err := o.findEC2Instances(ctx, ec2Client, deleted)
 			if err != nil {
 				o.Logger.WithError(err).Info("error while finding EC2 instances to delete")
 				if err := ctx.Err(); err != nil {
 					return false, err
 				}
 			}
-			if len(instancesToDelete) == 0 && err == nil {
+			if len(instancesNotTerminated) == 0 && len(instancesRunning) == 0 && err == nil {
 				return true, nil
+			}
+			instancesToDelete := instancesRunning
+			if time.Since(lastTerminateTime) > 10*time.Minute {
+				instancesToDelete = instancesNotTerminated
+				lastTerminateTime = time.Now()
 			}
 			newlyDeleted, err := o.deleteResources(ctx, awsSession, instancesToDelete, tracker)
 			// Delete from the resources-to-delete set so that the current state of the resources to delete can be
@@ -259,18 +265,21 @@ func (o *ClusterUninstaller) RunWithContext(ctx context.Context) ([]string, erro
 }
 
 // findEC2Instances returns the EC2 instances with tags that satisfy the filters.
+// returns two lists, first one is the list of all resources that are not terminated and are not in shutdown
+// stage and the second list is the list of resources that are not terminated.
 //   deleted - the resources that have already been deleted. Any resources specified in this set will be ignored.
-func (o *ClusterUninstaller) findEC2Instances(ctx context.Context, ec2Client *ec2.EC2, deleted sets.String) ([]string, error) {
+func (o *ClusterUninstaller) findEC2Instances(ctx context.Context, ec2Client *ec2.EC2, deleted sets.String) ([]string, []string, error) {
 	if ec2Client.Config.Region == nil {
-		return nil, errors.New("EC2 client does not have region configured")
+		return nil, nil, errors.New("EC2 client does not have region configured")
 	}
 
 	partition, ok := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), *ec2Client.Config.Region)
 	if !ok {
-		return nil, errors.Errorf("no partition found for region %q", *ec2Client.Config.Region)
+		return nil, nil, errors.Errorf("no partition found for region %q", *ec2Client.Config.Region)
 	}
 
-	var resources []string
+	var resourcesRunning []string
+	var resourcesNotTerminated []string
 	for _, filter := range o.Filters {
 		o.Logger.Debugf("search for instances by tag matching %#+v", filter)
 		instanceFilters := make([]*ec2.Filter, 0, len(filter))
@@ -301,9 +310,12 @@ func (o *ClusterUninstaller) findEC2Instances(ctx context.Context, ec2Client *ec
 								instanceLogger.Info("Terminated")
 								deleted.Insert(arn)
 							}
-						} else {
-							resources = append(resources, arn)
+							continue
 						}
+						if *instance.State.Name != "shutting-down" {
+							resourcesRunning = append(resourcesRunning, arn)
+						}
+						resourcesNotTerminated = append(resourcesNotTerminated, arn)
 					}
 				}
 				return !lastPage
@@ -312,10 +324,10 @@ func (o *ClusterUninstaller) findEC2Instances(ctx context.Context, ec2Client *ec
 		if err != nil {
 			err = errors.Wrap(err, "get ec2 instances")
 			o.Logger.Info(err)
-			return resources, err
+			return resourcesRunning, resourcesNotTerminated, err
 		}
 	}
-	return resources, nil
+	return resourcesRunning, resourcesNotTerminated, nil
 }
 
 // findResourcesToDelete returns the resources that should be deleted.
@@ -868,8 +880,8 @@ func terminateEC2Instance(ctx context.Context, ec2Client *ec2.EC2, iamClient *ia
 }
 
 func terminateEC2InstanceByInstance(ctx context.Context, ec2Client *ec2.EC2, iamClient *iam.IAM, instance *ec2.Instance, logger logrus.FieldLogger) error {
-	// Skip 'shutting-down' and 'terminated' instances since they take a while to get cleaned up
-	if instance.State == nil || *instance.State.Name == "shutting-down" || *instance.State.Name == "terminated" {
+	// Ignore instances that are already terminated
+	if instance.State == nil || *instance.State.Name == "terminated" {
 		return nil
 	}
 


pkg/destroy/aws/aws.go
```diff
@@ -218,7 +218,7 @@ func (o *ClusterUninstaller) RunWithContext(ctx context.Context) ([]string, erro
 
 	// Delete the rest of the resources.
 	err = wait.PollImmediateUntil(
-		time.Second*10,
+		time.Minute*10,
```
staebler, line 221: We do not want to change how often we attempt to delete resources. This will make the destroy experience significantly worse.

pkg/destroy/aws/aws.go
```diff
@@ -868,8 +868,8 @@ func terminateEC2Instance(ctx context.Context, ec2Client *ec2.EC2, iamClient *ia
 }
 
 func terminateEC2InstanceByInstance(ctx context.Context, ec2Client *ec2.EC2, iamClient *iam.IAM, instance *ec2.Instance, logger logrus.FieldLogger) error {
-	// Skip 'shutting-down' and 'terminated' instances since they take a while to get cleaned up
-	if instance.State == nil || *instance.State.Name == "shutting-down" || *instance.State.Name == "terminated" {
+	// Skip 'terminated' instances since they take a while to get cleaned up
+	if instance.State == nil || *instance.State.Name == "terminated" {
```
staebler, line 872: This will re-send the terminate request an instance every 10 seconds while it is shutting down. Can we figure out some way to throttle the terminate requests?

pkg/destroy/aws/aws.go
```diff
@@ -868,9 +869,16 @@ func terminateEC2Instance(ctx context.Context, ec2Client *ec2.EC2, iamClient *ia
...
-		return nil
+	// Skip 'terminated' instances since they take a while to get cleaned up
+	if instance.State == nil {
+		if *instance.State.Name == "terminated" {
+			return nil
+		} else if *instance.State.Name == "shutting-down" {
+			if time.Now().Before(timer) {
+				return nil
+			}
+			timer = time.Now().Add(4 * time.Second)
```
staebler, line 880: Since the same timer is used for all instances, the first instance will starve out all of the other instances. In other words, the terminate request will be re-sent for the first instance only until it is terminated.

rna-afk: Ah yes of course! Thanks

pkg/destroy/aws/aws.go
```diff
@@ -33,7 +33,9 @@ import (
 )
 
 var (
-	exists = struct{}{}
+	exists      = struct{}{}
+	currentTime = time.Now()
```
staebler, line 37: This variable is never set anywhere but here.

pkg/destroy/aws/aws.go
```diff
@@ -868,9 +874,15 @@ func terminateEC2Instance(ctx context.Context, ec2Client *ec2.EC2, iamClient *ia
 }
 
 func terminateEC2InstanceByInstance(ctx context.Context, ec2Client *ec2.EC2, iamClient *iam.IAM, instance *ec2.Instance, logger logrus.FieldLogger) error {
-	// Skip 'shutting-down' and 'terminated' instances since they take a while to get cleaned up
-	if instance.State == nil || *instance.State.Name == "shutting-down" || *instance.State.Name == "terminated" {
-		return nil
+	// Skip 'terminated' instances since they take a while to get cleaned up
+	if instance.State == nil {
```
staebler, line 878:
```suggestion
	if instance.State != nil {
```

staebler (commented):
How about this? Instead of keeping a global last-request time, let's keep a time locally in the poll function. Let's change `findEC2Instances` so that it returns another slice for the instances that are not yet shutting down. When it has been long enough since the last terminate requests have been made, attempt to delete all of the instances. If it has not been long enough, only attempt to delete the instances that are not yet shutting down.

pkg/destroy/aws/aws.go
rna-afk: I am also updating the variable at L203?

staebler: No, you are creating a new `currentTime` variable at that line.

rna-afk: Got it now. Thanks

pkg/destroy/aws/aws.go
```diff
@@ -185,19 +185,23 @@ func (o *ClusterUninstaller) RunWithContext(ctx context.Context) ([]string, erro
...
 				if err := ctx.Err(); err != nil {
 					return false, err
 				}
 			}
-			if len(instancesToDelete) == 0 && err == nil {
+			if len(instancesToDelete) == 0 && len(instancesShuttingDown) == 0 && err == nil {
 				return true, nil
 			}
+			if time.Now().After(timer.Add(5 * time.Second)) {
+				instancesToDelete = append(instancesToDelete, instancesShuttingDown...)
```
jstuever, line 203: Doesn't the timer need to be refreshed here? Otherwise it will repeat every 10 seconds after the initial timer expires?

pkg/destroy/aws/aws.go
```diff
@@ -260,17 +264,18 @@ func (o *ClusterUninstaller) RunWithContext(ctx context.Context) ([]string, erro
 
 // findEC2Instances returns the EC2 instances with tags that satisfy the filters.
 //   deleted - the resources that have already been deleted. Any resources specified in this set will be ignored.
-func (o *ClusterUninstaller) findEC2Instances(ctx context.Context, ec2Client *ec2.EC2, deleted sets.String) ([]string, error) {
+func (o *ClusterUninstaller) findEC2Instances(ctx context.Context, ec2Client *ec2.EC2, deleted sets.String) ([]string, []string, error) {
```
staebler, line 267: Add to the description of this function details about what the different return values are, since it is not obvious from the function signature.

pkg/destroy/aws/aws.go
```diff
@@ -185,19 +185,23 @@ func (o *ClusterUninstaller) RunWithContext(ctx context.Context) ([]string, erro
 	// running on the cluster creating new resources while we are attempting to delete resources, which could leak
 	// the new resources.
 	ec2Client := ec2.New(awsSession)
+	timer := time.Now()
```
staebler, line 188: Improve upon the name of this variable.
1) It is not a timer.
2) The name has no context about what it is used for.

pkg/destroy/aws/aws.go
```diff
@@ -185,19 +185,23 @@ func (o *ClusterUninstaller) RunWithContext(ctx context.Context) ([]string, erro
...
 	err = wait.PollImmediateUntil(
 		time.Second*10,
 		func() (done bool, err error) {
-			instancesToDelete, err := o.findEC2Instances(ctx, ec2Client, deleted)
+			instancesToDelete, instancesShuttingDown, err := o.findEC2Instances(ctx, ec2Client, deleted)
```
staebler, line 192: Without any change in complexity in the `findEC2Instances` function, you can return (1) the instances that are not terminated and (2) the instances that are not terminated and not shutting down. Then, here you do not have to append one slice to another and can instead select which slice to use.
For example,
```
instancesNotTerminated, instancesRunning, err := o.findEC2Instances(ctx, ec2Client, deleted)
instancesToDelete := instancesRunning
if long enough since last re-terminate {
  instancesToDelete = instancesNotTerminated
}
```

pkg/destroy/aws/aws.go
```diff
@@ -185,19 +185,23 @@ func (o *ClusterUninstaller) RunWithContext(ctx context.Context) ([]string, erro
...
 				o.Logger.WithError(err).Info("error while finding EC2 instances to delete")
 				if err := ctx.Err(); err != nil {
 					return false, err
 				}
 			}
-			if len(instancesToDelete) == 0 && err == nil {
+			if len(instancesToDelete) == 0 && len(instancesShuttingDown) == 0 && err == nil {
 				return true, nil
 			}
+			if time.Now().After(timer.Add(5 * time.Second)) {
```
staebler, line 202:
```suggestion
			if time.Since(timer) > 5 * time.Second {
```

pkg/destroy/aws/aws.go
```diff
@@ -185,19 +185,25 @@ func (o *ClusterUninstaller) RunWithContext(ctx context.Context) ([]string, erro
 	// running on the cluster creating new resources while we are attempting to delete resources, which could leak
 	// the new resources.
 	ec2Client := ec2.New(awsSession)
+	shutdownTimeout := time.Now()
```
staebler, line 188:
```suggestion
	lastTerminateTime := time.Now()
```

pkg/destroy/aws/aws.go
```diff
@@ -868,8 +878,8 @@ func terminateEC2Instance(ctx context.Context, ec2Client *ec2.EC2, iamClient *ia
 }
 
 func terminateEC2InstanceByInstance(ctx context.Context, ec2Client *ec2.EC2, iamClient *iam.IAM, instance *ec2.Instance, logger logrus.FieldLogger) error {
-	// Skip 'shutting-down' and 'terminated' instances since they take a while to get cleaned up
-	if instance.State == nil || *instance.State.Name == "shutting-down" || *instance.State.Name == "terminated" {
+	// Skip 'terminated' instances since they take a while to get cleaned up
```
staebler, line 881: Change this comment to reflect that terminated is the desired state.
```suggestion
	// Ignore instances that are already terminated
```

pkg/destroy/aws/aws.go
```diff
@@ -868,8 +878,8 @@ func terminateEC2Instance(ctx context.Context, ec2Client *ec2.EC2, iamClient *ia
...
+	if instance.State == nil && *instance.State.Name == "terminated" {
```
staebler, line 882: This logic is not correct. Keep the logic as it was, where instances without state are ignored, since those instances are ignored in `findEC2Instances`.
```suggestion
	if instance.State == nil || *instance.State.Name == "terminated" {
```

pkg/destroy/aws/aws.go
```diff
@@ -301,9 +310,10 @@ func (o *ClusterUninstaller) findEC2Instances(ctx context.Context, ec2Client *ec
 								instanceLogger.Info("Terminated")
 								deleted.Insert(arn)
 							}
-						} else {
-							resources = append(resources, arn)
+						} else if *instance.State.Name != "shutting-down" {
+							resourcesRunning = append(resourcesRunning, arn)
 						}
+						resourcesNotTerminated = append(resourcesNotTerminated, arn)
```
staebler, line 316: You are adding all instances to `resourcesNotTerminated`, even instances that are already terminated.
```
if terminated
  add to delete
  continue for loop
if not shutting down
  add to resources running
add to resources not terminated
```
or
```
switch state
case terminated
  add to delete
case shutting down
  add to resources running
  fallthrough
default
  add to resources not terminated
```

pkg/destroy/aws/aws.go
```diff
@@ -185,19 +185,25 @@ func (o *ClusterUninstaller) RunWithContext(ctx context.Context) ([]string, erro
...
 				}
 			}
-			if len(instancesToDelete) == 0 && err == nil {
+			if len(instancesNotTerminated) == 0 && len(instancesRunning) == 0 && err == nil {
 				return true, nil
 			}
+			instancesToDelete := instancesRunning
+			if time.Since(shutdownTimeout) > 5*time.Second {
+				instancesToDelete = instancesNotTerminated
+				shutdownTimeout = time.Now().Add(5 * time.Second)
```
staebler, line 205: You don't want to add 5 seconds. You are checking that it has been more than 5 seconds since `shutdownTimeout` above, so adding 5 seconds here effectively makes the check that it has been more than 10 seconds since the last mass terminate.
```suggestion
				shutdownTimeout = time.Now()
```

pkg/destroy/aws/aws.go
```diff
@@ -185,19 +185,25 @@ func (o *ClusterUninstaller) RunWithContext(ctx context.Context) ([]string, erro
...
 				if err := ctx.Err(); err != nil {
 					return false, err
 				}
 			}
-			if len(instancesToDelete) == 0 && err == nil {
+			if len(instancesNotTerminated) == 0 && len(instancesRunning) == 0 && err == nil {
 				return true, nil
 			}
+			instancesToDelete := instancesRunning
+			if time.Since(shutdownTimeout) > 5*time.Second {
```
staebler, line 203: I assume that 5 seconds is just for your initial testing and that the actual delay between mass terminates will be much longer.

pkg/destroy/aws/aws.go
rna-afk: If I don't add 5 seconds here, it would keep trying to mass terminate after 5 seconds endlessly. I was hoping to request just one more mass terminate in the 10 second timeout allotted for the destroy instances.

rna-afk: oh wait no nvm I'l fix this

pkg/destroy/aws/aws.go
staebler: This is still adding terminated instances to `resourcesNotTerminated`.

pkg/destroy/aws/aws.go
```diff
@@ -185,19 +185,25 @@ func (o *ClusterUninstaller) RunWithContext(ctx context.Context) ([]string, erro
...
 				if err := ctx.Err(); err != nil {
 					return false, err
 				}
 			}
-			if len(instancesToDelete) == 0 && err == nil {
+			if len(instancesNotTerminated) == 0 && len(instancesRunning) == 0 && err == nil {
 				return true, nil
 			}
+			instancesToDelete := instancesRunning
+			if time.Since(lastTerminateTime) > 9*time.Second {
```
staebler, line 203: Do you have an idea what you want to use as the period here? 9 seconds is effectively "always".

rna-afk: Hmm doesn't this wait 9 seconds until it requests for another fresh termination? Since the timeout to delete these running instances is 10 seconds, I was hoping to wait for 9 seconds before trying to request a fresh termination request.

staebler: Yes, it waits 9 seconds. That wait time is much too short. It should be something like 10 minutes instead.

There is no "timeout to delete [those] running instances". There is a check on running instances every 10 seconds. We want to keep that check as is, but add a repeat of the terminate requests every 10 minutes for instances that are still shutting down.

pkg/destroy/aws/aws.go
```diff
@@ -301,9 +310,11 @@ func (o *ClusterUninstaller) findEC2Instances(ctx context.Context, ec2Client *ec
 								instanceLogger.Info("Terminated")
 								deleted.Insert(arn)
 							}
-						} else {
-							resources = append(resources, arn)
+							continue
+						} else if *instance.State.Name != "shutting-down" {
```
staebler, line 314:
```suggestion
						}
						if *instance.State.Name != "shutting-down" {
```

staebler (commented):
Looks good. Just one nit.

staebler (approved):
/lgtm

PR #4849, opened 2021-04-14
doc/gcp: remove no-longer-needed etcd records

shellyyang1989:
Removing etcd DNS records related stuff, which is no longer needed in 4.4 and above.


staebler (commented):
/lgtm
/approve
/hold for master branch to open

PR #4850, opened 2021-04-14
Bug 1950007: UPI image: use get-pip instead of easy_install

e-tienne:
easy_install is still failing. Let's make sure to use the
documented way to install an older pip version.

https://pip.pypa.io/en/stable/installing/


e-tienne:
/retitle UPI image: use get-pip instead of easy_install

e-tienne:
/retitle Bug 1950007: UPI image: use get-pip instead of easy_install

PR #4851, opened 2021-04-15
Bug 1949923: OpenStack: validate root volume availability zones

Fedosin:
This commit adds next validations for OpenStack Root Volume AZs:
* All specified AZs exist and are enabled in the cloud;
* There must be either just one common volume AZ or the number of compute and volume AZs must be equal.


pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -44,6 +44,12 @@ func ValidateMachinePool(p *openstack.MachinePool, ci *CloudInfo, controlPlane b
 		if p.RootVolume.Size < minimumStorage {
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume").Child("size"), p.RootVolume.Size, fmt.Sprintf("Volume size must be greater than %d to use root volumes, had %d", minimumStorage, p.RootVolume.Size)))
 		}
+
+		allErrs = append(allErrs, validateZones(p.RootVolume.Zones, ci.VolumeZones, fldPath.Child("rootVolume").Child("zones"))...)
+		if len(p.RootVolume.Zones) != 1 && len(p.RootVolume.Zones) != len(p.Zones) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume").Child("zones"), p.RootVolume.Zones, "There must be either just one common volume availability zone or the number of compute and volume availability zones must be equal"))
```
staebler, line 50:
```suggestion
			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume", "zones"), p.RootVolume.Zones, "there must be either just one common volume availability zone or the number of compute and volume availability zones must be equal"))
```

staebler (commented):
/approve

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -44,6 +44,12 @@ func ValidateMachinePool(p *openstack.MachinePool, ci *CloudInfo, controlPlane b
 		if p.RootVolume.Size < minimumStorage {
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume").Child("size"), p.RootVolume.Size, fmt.Sprintf("Volume size must be greater than %d to use root volumes, had %d", minimumStorage, p.RootVolume.Size)))
 		}
+
+		allErrs = append(allErrs, validateZones(p.RootVolume.Zones, ci.VolumeZones, fldPath.Child("rootVolume").Child("zones"))...)
+		if len(p.RootVolume.Zones) != 1 && len(p.RootVolume.Zones) != len(p.Zones) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume", "zones"), p.RootVolume.Zones, "there must be either just one common volume availability zone or the number of compute and volume availability zones must be equal"))
```
mandre, line 50: Nit, I find this error message slightly difficult to understand if you don't know the implementation of it. Below suggestion is a bit clearer I think:
```suggestion
			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume", "zones"), p.RootVolume.Zones, "there must be either just one volume availability zone common to all nodes or the number of compute and volume availability zones must be equal"))
```

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -44,6 +44,12 @@ func ValidateMachinePool(p *openstack.MachinePool, ci *CloudInfo, controlPlane b
 		if p.RootVolume.Size < minimumStorage {
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume").Child("size"), p.RootVolume.Size, fmt.Sprintf("Volume size must be greater than %d to use root volumes, had %d", minimumStorage, p.RootVolume.Size)))
 		}
+
+		allErrs = append(allErrs, validateZones(p.RootVolume.Zones, ci.VolumeZones, fldPath.Child("rootVolume").Child("zones"))...)
```
mandre, line 48: Should we consider it an error if `len(p.RootVolume.Zones)` is greater than the number of replicas? My guess is we don't really care since it's not affecting how the installer works.

mandre (commented):
Something else that comes to mind: the installer should probably set "ignore-volume-az" in the cloud provider config when it finds that the compute and volume zones do not match.

pkg/asset/installconfig/openstack/validation/machinepool.go
Fedosin: it's a good point, but in this case we need to validate compute zones too. with can do it in another pr.

Fedosin:
@mandre This PR implements root volume validations, and `ignore-volume-az` is for cinder PVs, which is different. We can add this warning in a separate PR.

PR #4852, opened 2021-04-15
Bug 1958216: libvirt: Allow duplicate dnsmasq options

Prashanth684:
A corner case was discovered where if a sepcific option is specified multiple times(for example the
"address" option could be specified multiple times for different domains), only one of the options is
set. This is a bug and an oversight when adding the original code for the dnsmasq options. Fixed by using
a list instead of a map.

### DIFF
--- a/pkg/tfvars/libvirt/libvirt.go
+++ b/pkg/tfvars/libvirt/libvirt.go
@@ -18,16 +18,16 @@ import (
 )
 
 type config struct {
-	URI             string            `json:"libvirt_uri,omitempty"`
-	Image           string            `json:"os_image,omitempty"`
-	IfName          string            `json:"libvirt_network_if"`
-	MasterIPs       []string          `json:"libvirt_master_ips,omitempty"`
-	BootstrapIP     string            `json:"libvirt_bootstrap_ip,omitempty"`
-	MasterMemory    string            `json:"libvirt_master_memory,omitempty"`
-	MasterVcpu      string            `json:"libvirt_master_vcpu,omitempty"`
-	BootstrapMemory int               `json:"libvirt_bootstrap_memory,omitempty"`
-	MasterDiskSize  string            `json:"libvirt_master_size,omitempty"`
-	DnsmasqOptions  map[string]string `json:"libvirt_dnsmasq_options,omitempty"`
+	URI             string              `json:"libvirt_uri,omitempty"`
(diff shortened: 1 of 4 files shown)


pkg/tfvars/libvirt/libvirt.go
```diff
@@ -18,16 +18,16 @@ import (
...
+	URI             string     `json:"libvirt_uri,omitempty"`
+	Image           string     `json:"os_image,omitempty"`
+	IfName          string     `json:"libvirt_network_if"`
+	MasterIPs       []string   `json:"libvirt_master_ips,omitempty"`
+	BootstrapIP     string     `json:"libvirt_bootstrap_ip,omitempty"`
+	MasterMemory    string     `json:"libvirt_master_memory,omitempty"`
+	MasterVcpu      string     `json:"libvirt_master_vcpu,omitempty"`
+	BootstrapMemory int        `json:"libvirt_bootstrap_memory,omitempty"`
+	MasterDiskSize  string     `json:"libvirt_master_size,omitempty"`
+	DnsmasqOptions  [][]string `json:"libvirt_dnsmasq_options,omitempty"`
```
staebler, line 30: The way this is done elsewhere is using a `[]map[string]string`.
https://github.com/openshift/installer/blob/d0462d8b5074448e1917da7f0a5d7a904bd60359/pkg/tfvars/baremetal/baremetal.go#L24

Prashanth684: ah..interesting..but do you have any objections with the way it is now or do you want to change it to be consistent ?

Prashanth684: @staebler not urgent by any means - but are you ok with this approach or do you want this to be changed?

staebler: This is code that you own, so I won't block it the way that you have it. With that being said, here are my arguments for changing it.
1) Consistency in the code so that it is easier to move from one platform to another as a developer.
2) Easier to read the terraform code since the values are read by key rather than by index. With reading by index, a developer reading the code would need to switch over to a far-away file in the repo to see what is stored in each index.

Prashanth684: makes sense. changed the code to make it a list of mappings. thanks @staebler .

PR #4859, opened 2021-04-19
Bug 1948359: aws: add permissions for untagging shared byo instance roles

staebler:
Add checks to the Platform Permissions Check asset to verify that the user has the necessary permissions to remove the shared tag from the IAM roles that the user is supplying.

### DIFF
--- a/pkg/asset/installconfig/aws/permissions.go
+++ b/pkg/asset/installconfig/aws/permissions.go
@@ -28,6 +28,10 @@ const (
 
 	// PermissionDeleteSharedNetworking is a set of permissions required when the installer destroys resources from a shared-network cluster.
 	PermissionDeleteSharedNetworking PermissionGroup = "delete-shared-networking"
+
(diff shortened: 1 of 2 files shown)


mtnbikenc (approved):
/lgtm

pkg/asset/installconfig/aws/permissions.go
```diff
@@ -223,6 +227,11 @@ var permissions = map[PermissionGroup][]string{
 	PermissionDeleteSharedNetworking: {
 		"tag:UnTagResources",
 	},
+	// Permissions required for deleting a cluster with shared instance roles
+	PermissionDeleteSharedRoles: {
```
mtnbikenc, line 231: `PermissionDeleteSharedInstanceRoles`?

mtnbikenc (changes requested):
/lgtm cancel

pkg/asset/installconfig/aws/permissions.go
staebler: Fixed.

mtnbikenc (approved):

PR #4860, opened 2021-04-20
Bug 1949923: apply defaults for root volume AZs if they aren't set

Fedosin:
This commit applies default values for root volume AZs before validating control plane and compute machine pools.

### DIFF
--- a/pkg/asset/installconfig/openstack/validate.go
+++ b/pkg/asset/installconfig/openstack/validate.go
@@ -8,6 +8,7 @@ import (
 	"github.com/openshift/installer/pkg/asset/installconfig/openstack/validation"
 	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/openstack"
+	openstackdefaults "github.com/openshift/installer/pkg/types/openstack/defaults"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -46,7 +46,7 @@ func ValidateMachinePool(p *openstack.MachinePool, ci *CloudInfo, controlPlane b
 		}
 
 		allErrs = append(allErrs, validateZones(p.RootVolume.Zones, ci.VolumeZones, fldPath.Child("rootVolume").Child("zones"))...)
-		if len(p.RootVolume.Zones) != 1 && len(p.RootVolume.Zones) != len(p.Zones) {
+		if len(p.RootVolume.Zones) > 1 && len(p.RootVolume.Zones) != len(p.Zones) {
```
mdbooth, line 49: Just trying to work out the expected semantics of p.RootVolume.Zones here. Below in validateZones we have:

``` go
func validateZones(input []string, available []string, fldPath *field.Path) field.ErrorList {
        // check if machinepool default
        if len(input) == 1 && input[0] == "" {
                return field.ErrorList{}
        }
...
```

which suggests that 'empty' RootVolume.Zones actually contains a single value of ""? I'll be honest: I don't fully understand the context here, but that seems like a potential source of error.

mandre (commented):
/approve
/lgtm

PR #4864, opened 2021-04-20
Bug 1943219: azure: remove bootstrap ssh rule on private clusters

e-tienne:
Removes ssh from source ANY to destination ANY on azure
private clusters. The private cluster design should not
include SSH access to the bootstrap unless you're on the
same local network.

https://bugzilla.redhat.com/show_bug.cgi?id=1943219

### DIFF
--- a/data/data/azure/bootstrap/main.tf
+++ b/data/data/azure/bootstrap/main.tf
@@ -216,6 +216,8 @@ resource "azurerm_linux_virtual_machine" "bootstrap" {
 }
 
 resource "azurerm_network_security_rule" "bootstrap_ssh_in" {
+  count = var.private ? 0 : 1
+
   name                        = "bootstrap_ssh_in"
   priority                    = 103
   direction                   = "Inbound"


e-tienne:
/retitle Bug 1943219: azure: remove bootstrap ssh rule on private clusters

data/data/azure/bootstrap/main.tf
```diff
@@ -216,6 +216,8 @@ resource "azurerm_linux_virtual_machine" "bootstrap" {
 }
 
 resource "azurerm_network_security_rule" "bootstrap_ssh_in" {
+  count = var.private ? 0 : 1
```
staebler, line 219: I thought that we had talked about keeping this security rule for private clusters but changing the allowed source addresses to the vnet cidrs, but I may be forgetting the details. Did we decide that was not needed because, with Azure, the intra-vnet traffic is all allowed?

e-tienne: That is my understanding. When you create a Network Security Group in azure, it will allow intra VNET traffic ([default NSG security rule](https://docs.microsoft.com/en-us/azure/virtual-network/network-security-groups-overview#default-security-rules))

PR #4865, opened 2021-04-21
azure: allow VM image to be overriden

m1kola:
* Allows VM image to be overriden
* Adds some unit tests for `provider` and `getNetworkInfo` from `pkg/asset/machines/azure/machines.go`

---

This PR doesn't change default behaviour of the installer: `openshift-install create manifests` will generate same manifests compared to manifests generated by the installer compiled from the commit before this PR. In general, all `openshift-install create` subcommands should not be affected.

This PR however adds an ability to specifiy a custom VM image for Azure. To do so:

1. Run
    ```yaml
    openshift-install create install-config
    ```
1. Under `platform.azure` add `image` object and specify an image by either refering to it by `resourceID` or `publisher`/`offer`/`sku`/`version`. For example:
    ```yaml
    platform:
      azure:
        # ...
        image:
          offer: aro4
          publisher: azureopenshift
          sku: aro_43
          version: 43.81.20200311
    ```
    or 
    ```yaml
    platform:
      azure:
        # ...
        image:
          resourceID: /resourceGroups/test-rg/providers/Microsoft.Compute/images/test-image
    ```
1. Run `openshift-install create manifests` to generate manifests
2. Check master `Machine` and worker `MachineSet` manifests: the change should be reflected in the `spec.providerSpec.value.image`.

Also see:
```
openshift-install explain installconfig.platform.azure.image
```

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -1176,6 +1176,28 @@ spec:
                           type: string
                         type: array
                     type: object
+                  image:
+                    description: Image specifies the image parameters with which a
+                      cluster should be built. Either ResourceID or Publisher/Offer/SKU/Version
+                      should be set.
+                    properties:
+                      offer:
+                        description: Offer is the image offer
+                        type: string
+                      publisher:
+                        description: Publisher is the image publisher
+                        type: string
+                      resourceId:
+                        description: ResourceID is the resource ID of an existing
+                          Image resource
+                        type: string
+                      sku:
+                        description: SKU is the image SKU
+                        type: string
+                      version:
+                        description: Version is the image version
+                        type: string
+                    type: object
                   networkResourceGroupName:
                     description: NetworkResourceGroupName specifies the network resource
                       group that contains an existing VNet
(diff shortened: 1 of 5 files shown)


tsze-redhat:
For the record, from previous discussion: this is for ARO internal use and not exposed to customer.

m1kola:
> For the record, from previous discussion: this is for ARO internal use and not exposed to customer.

@tsze-redhat this is correct, but want to clarify: despite the fact that we are adressing ARO internal use case in this PR it also enables end users of the installer to be able to override an image in the install config in a similar way to [other customisation](https://github.com/openshift/installer/blob/26d24bd526c43e83a6a4127205a8172d31cfd44e/docs/user/azure/customization.md) (see PR descritpion on how to do this). But because I wasn't sure whether this feature is benefinital to anyone apart from ARO, I did not udpate the documentation.

If you think that there is a value for the wider community of users to be able to override an image - I'm happy to update the [customisation doc](https://github.com/openshift/installer/blob/26d24bd526c43e83a6a4127205a8172d31cfd44e/docs/user/azure/customization.md). Note that `openshift-install explain installconfig.platform.azure.image` already exposes some help info on these new fields.

m1kola:
Closing in favour of https://issues.redhat.com/browse/CORS-1757

PR #4861, opened 2021-04-20
Bug 1955701: Bump boot images for RHEL 8.4 Beta

travier:
```
x86_64  48.84.202104271417-0
s390x   48.84.202104271619-0
ppc64le 48.84.202104261624-0
```

PR #4870, opened 2021-04-23
Add arm64 support

yselkowitz:
- Add ARM architecture types
- Fix libvirt cluster destroy on ARM
- Add ARM support to AWS installation
- cmd/openshift-install: Add architecture as an option for install-config
- Make order of install-config questions more consistent
- aws: treat a1 as backup to m6g
- Add Makefile
- Add rhcos-stream data for aarch64


staebler:
This PR is covering a lot of changes, some of which are independent from each other. Please separate into discrete PRs that can be discussed on their own. Also, please add references to enhancements or Jira cards that provide more context for these changes.

yselkowitz:
I moved the Makefile to #4891 where it belongs, and asked if the rhcos stream data could be added to #4861.  I think that's about as much as we could sensibly split out.

data/data/rhcos-stream.json
```diff
@@ -1,9 +1,154 @@
...
+          "regions": {
+            "ap-east-1": {
+              "release": "48.84.202105102319-0",
+              "image": "ami-0b28d982f7ff321f0"
+            },
+            "ap-northeast-1": {
+              "release": "48.84.202105102319-0",
+              "image": "ami-0ffc57b891e093438"
+            },
+            "ap-northeast-2": {
```
staebler, line 92: Do you know what would need to happen to get an ARM AMI for ap-northeast-3?

pkg/asset/installconfig/platform.go
```diff
@@ -131,6 +149,29 @@ func (a *platform) queryUserForPlatform() (platform string, err error) {
 	return
 }
 
+func (a *platform) queryUserForArchitecture(archlist []string) (architecture types.Architecture, err error) {
```
staebler, line 152: I don't think that this meets the bar for inclusion in the survey. The user can manually set the architecture in the install-config.yaml.

pkg/asset/machines/aws/instance_types.go
```diff
@@ -47,8 +47,14 @@ func PreferredInstanceType(ctx context.Context, meta *awsconfig.Metadata, types
...
-			return t, nil
+			return t, zones, nil
 		}
 	}
-	return types[0], errors.New("no instance type found for the zone constraint")
+	// HACK: in the ARM case, the m6g and m6gd instances are not available in all the zones. This might be OK as is, but the
+	// manifests generated list all the zones as available which is incorrect. Instead return only the available zones
+	// Delete the zones in which the instance is not available and return the rest
+	for k := range reqZones.Difference(sets.NewString(found[types[0]]...)) {
+		reqZones.Delete(k)
```
staebler, line 57: This should only be done if the user did not specify which zones to use. If the user did specify the zones to use, either explicitly or implicitly via provided subnets, then the installation should fail if any of those zones cannot be used.

pkg/types/machinepools.go
```diff
@@ -34,8 +34,16 @@ const (
...
 	ArchitecturePPC64LE = "ppc64le"
+	// ArchitectureARM64 indicates arm (aarch64) systems
+	ArchitectureARM64 = "arm64"
 )
 
+// ArchitectureList contains all possible architectures
+var ArchitectureList = []Architecture{ArchitectureAMD64, ArchitectureARM64, ArchitecturePPC64LE, ArchitectureS390X}
+
+// ArchitectureListAWS contains the architectures available on AWS
+var ArchitectureListAWS = []Architecture{ArchitectureAMD64, ArchitectureARM64}
```
staebler, line 45: It looks like we need validation to ensure that the user does not select an architecture other than amd or arm when installing to AWS.

pkg/types/aws/defaults/platform.go
```diff
@@ -17,18 +18,26 @@ func SetPlatformDefaults(p *aws.Platform) {
...
 // InstanceClass returns the instance "class" we should use for a given
 // region. Default is m5 unless a region override is defined in defaultMachineClass.
-func InstanceClass(region string) string {
+func InstanceClass(region string, arch types.Architecture) string {
 	if classes, ok := defaultMachineClass[region]; ok {
 		return classes[0]
 	}
+
+	if arch == types.ArchitectureARM64 {
+		return "m6g"
```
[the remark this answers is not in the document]
Prashanth684: i believe this PR: https://github.com/openshift/machine-api-operator/pull/850 takes care of this issue ?

pkg/destroy/libvirt/libvirt.go
```diff
@@ -123,8 +123,15 @@ func deleteDomainsSinglePass(conn *libvirt.Connect, filter filterFunc, logger lo
...
-		if err := domain.Undefine(); err != nil {
-			return false, errors.Wrapf(err, "undefine domain %q", dName)
+		if err := domain.UndefineFlags(libvirt.DOMAIN_UNDEFINE_NVRAM); err != nil {
+			if e := err.(libvirt.Error); e.Code == libvirt.ERR_NO_SUPPORT || e.Code == libvirt.ERR_INVALID_ARG {
+				logger.WithField("domain", dName).Info("libvirt does not support undefine flags: will try again without flags")
+				if err := domain.Undefine(); err != nil {
+					return false, errors.Wrapf(err, "couldn't undefine libvirt domain: %q", dName)
+				}
+			} else {
+				return false, errors.Wrapf(err, "couldn't undefine libvirt domain %q with flags", dName)
```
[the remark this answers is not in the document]
Prashanth684: done.

pkg/types/aws/defaults/platform.go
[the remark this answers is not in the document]
staebler: I don't think so, unless I am missing something subtle. The installer does not omit the instance type from the machien provider spec, so defaults in machine-api-operator do not apply here.

pkg/types/machinepools.go
Prashanth684: is this validation when the user edits install-config manually? because if created through the `create install-config` command the only options available for aws are `amd64` and `arm64`.

pkg/asset/machines/aws/instance_types.go
staebler: https://github.com/openshift/installer/blob/296b0b10da2a5a9a5afca6c10b03ee73ce98578c/pkg/asset/installconfig/aws/validation.go#L178-L200

pkg/types/machinepools.go
Prashanth684: added arch validation for control plane, computes and AMIs..

pkg/asset/installconfig/platform.go
Prashanth684: so ARM will be a supported architecture from hereon for AWS, so don't we need that option for the user to select the architecture? I'm perfectly fine manually changing the architecture, but trying to understand the reason why we would not want this.

pkg/asset/installconfig/installconfig.go
```diff
@@ -42,32 +42,32 @@ var _ asset.WritableAsset = (*InstallConfig)(nil)
 func (a *InstallConfig) Dependencies() []asset.Asset {
 	return []asset.Asset{
 		&sshPublicKey{},
+		&platform{},
```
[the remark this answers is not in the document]
Prashanth684: i believe the reasoning was that instead of having architecture as the last query, it should up front as it is a basic platform specific setting. And since we support different arches for different platforms, the platform query needed to be moved up too.

pkg/asset/installconfig/platform.go
yselkowitz: Manual change should not be required.  If we could set the AWS architecture based on the payload architecture, that might suffice in place of a question, but not sure how to go about that.

staebler: > so ARM will be a supported architecture from hereon for AWS, so don't we need that option for the user to select the architecture? I'm perfectly fine manually changing the architecture, but trying to understand the reason why we would not want this.

The set of survey questions is not meant to be exhaustive of all of the configuration that a user may do. The set is meant to be small and curated to only configuration that the installer cannot provide opinionated defaults for. In this case, the installer can choose a default architecture.

yselkowitz: Not sure that choosing an architecture is a configuration option, it's a fundamental decision regarding where and on what to host a cluster, since x86 and ARM clusters aren't binary compatible.  It also affects the decision about regions, which is already a question.
From the user experience perspective, the results of `create install-config` "just work" without modification for deploying a default cluster.  ARM users should have the same experience.

data/data/rhcos-stream.json
Prashanth684: the issue is that the m6g instances are not available in certain regions (https://aws.amazon.com/about-aws/whats-new/2021/02/aws-graviton2-m6g-c6g-r6g-instances-available-additional-regions/), so there is no point pushing to them

staebler: That makes sense. Are there no available ARM instances in certain regions? Or just no m6g instances?

pkg/asset/installconfig/platform.go
staebler: The question to ask is "Can I install a cluster without the user providing this configuration?"
* SSH key - yes, but in older versions the answer was no, so this has been grandfathered in
* Platform - no
* Region - arguably yes, but the user may not have access to whatever the default region would be in their account
* Base domain - no
* Cluster name - no
* Pull secret - no

The answer for architecture is yes, so does not meet the requirements for inclusion. The criteria is not whether configuring through the survey makes the experience better for the user that wants that configuration.

What percentage of users do you expect would choose ARM over x86?

data/data/rhcos-stream.json
Prashanth684: i think none of the m6g, m6gd or a1 instances...so basically i think none of the ARM instances. @yselkowitz  am i correct that the excluded regions had no ARM instances? not just m6g,m6gd ? .

yselkowitz: m6g is the most prevalent and widely available ARM instance category at this point, so it's basically one and the same.

pkg/asset/installconfig/platform.go
yselkowitz: Payloads are architecture specific.  Therefore, if the installer just defaults to amd64, then the answer to your question will actually be "no", because the installation will fail with a non-x86 payload without manual modification of the configuration.  Furthermore, even if the user were to adjust the architecture, if they selected a region in which ARM instances aren't available, the installation will fail in that case too.

staebler: > Payloads are architecture specific. Therefore, if the installer just defaults to amd64, then the answer to your question will actually be "no", because the installation will fail with a non-x86 payload without manual modification of the configuration.
Then the user does not have any choice in this, and asking the user to make a selection is wrong. The installer needs to make this decision for the user.

> Furthermore, even if the user were to adjust the architecture, if they selected a region in which ARM instances aren't available, the installation will fail in that case too.

I don't follow how that is relevant here. If the user updates the install-config.yaml to select the ARM architecture, then they are obligated to select an appropriate region as well. If they do not, then the installer should rightly fail validation.

yselkowitz: Is there a way for the installer to know the payload architecture during `create install-config`?

staebler: It would need to be built into the installer binary.

Prashanth684: should we pursue this discussion in a separate issue and remove the addition of the install-config option here so we can proceed with this PR? this discussion encompasses more than just ARM and includes all other architectures in general.

staebler: https://github.com/openshift/installer/pull/4870#issuecomment-828909878

cgwalters:
Seems sane to me to split out the rhcos stream data commit FWIW.

yselkowitz:
> Seems sane to me to split out the rhcos stream data commit FWIW.

Only if that would be expedited, because the rest of this won't work without it.

staebler:
> > Seems sane to me to split out the rhcos stream data commit FWIW.
> 
> Only if that would be expedited, because the rest of this won't work without it.

I don't care about separating that out. It is non-controversial. I just want to separate out the stuff that is controversial so that we can have discussions about those separately.

And note that nothing can be merged for another 4 weeks anyway.

yselkowitz:
Added an initial attempt to drop the survey question in favour of deriving based on payload architecture (assumed from the build host).

pkg/types/aws/defaults/platform.go
```diff
@@ -17,18 +18,26 @@ func SetPlatformDefaults(p *aws.Platform) {
...
 // InstanceClasses returns a list of instance "class", in decreasing priority order, which we should use for a given
 // region. Default is m5 then m4 unless a region override is defined in defaultMachineClass.
-func InstanceClasses(region string) []string {
+func InstanceClasses(region string, arch types.Architecture) []string {
 	if classes, ok := defaultMachineClass[region]; ok {
 		return classes
 	}
+
+	if arch == types.ArchitectureARM64 {
+		return []string{"m6g", "a1"}
```
yselkowitz, line 40: Would just `[]string{"m6g"}` without a fallback (until the next generation of ARM instance classes) break any code assumptions elsewhere?  AFAIK there are no regions with a1 that don't also have m6g, and given the complications of auto-handling a1 due to its different sizing, it would probably be cleaner if we just drop it.

pkg/types/machinepools.go
yselkowitz: Should the validation be a separate commit?

hack/build.sh
```diff
@@ -16,8 +16,9 @@ fi
 MODE="${MODE:-release}"
 GIT_COMMIT="${SOURCE_GIT_COMMIT:-$(git rev-parse --verify 'HEAD^{commit}')}"
 GIT_TAG="${BUILD_VERSION:-$(git describe --always --abbrev=40 --dirty)}"
+HOSTARCH="$(go env GOHOSTARCH)"
```
Prashanth684, line 19: ok so iiuc - the architecture is now set to the host architecture if built manually and set to the architecture of the payload if it is extracted from the payload? so if i actually custom build the installer on an ARM laptop it would set the architecture in the install-config to arm64 - but which might not match the payload architecture necessarily.

Also note that someone could still override the payload using OPENSHIFT_RELEASE_IMAGE_OVERRIDE. so if these are called out it should make things clear for upstream as well as downstream.

pkg/asset/installconfig/aws/platform.go
```diff
@@ -9,12 +9,13 @@ import (
 	"github.com/sirupsen/logrus"
 	survey "gopkg.in/AlecAivazis/survey.v1"
 
-	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/aws"
+	"github.com/openshift/installer/pkg/version"
```
Prashanth684, line 13: probably want to make it clear in the commit message and description that it is set to the payload architecture only if the binary is extracted from the payload and not custom built. this might be misleading if someone looks at this commit and assumes that it always matches the payload in general

pkg/asset/machines/aws/instance_types.go
Prashanth684: ok..changed the logic to return a filtered list of zones only if the option has not been specified through subnets/zones in the install config.

pkg/asset/installconfig/aws/platform.go
yselkowitz: when you build the installer yourself, then you're probably already overriding the payload at create cluster time anyway.  at that point you're already on your own, so having to possibly edit the install-config shouldn't be onerous.

yselkowitz:
#4942 also added aarch64 RHCOS data, so I dropped that from this PR.

yselkowitz:
The CI failures do not seem related.
What else is needed here to get this merged?
/assign @staebler

pkg/asset/machines/aws/instance_types.go
```diff
@@ -23,7 +24,50 @@ func PreferredInstanceType(ctx context.Context, meta *awsconfig.Metadata, types
...
+		return zones, nil
+	}
+
+	for k := range reqZones.Difference(sets.NewString(found[types[0]]...)) {
+		reqZones.Delete(k)
+	}
+	return reqZones.List(), nil
+}
+
+func getInstanceTypeZoneInfo(ctx context.Context, session *session.Session, region string, types []string, zones []string) (map[string][]string, error) {
```
staebler, line 67:
```suggestion
func getInstanceTypeZoneInfo(ctx context.Context, session *session.Session, region string, types []string, zones []string) (map[string]sets.String, error) {
```

pkg/asset/machines/aws/instance_types.go
```diff
@@ -23,7 +24,50 @@ func PreferredInstanceType(ctx context.Context, meta *awsconfig.Metadata, types
...
 
-	client := ec2.New(sess, aws.NewConfig().WithRegion(meta.Region))
+	found, err := getInstanceTypeZoneInfo(ctx, sess, meta.Region, types, zones)
+	if err != nil {
+		return types[0], err
+	}
+
+	reqZones := sets.NewString(zones...)
+	for _, t := range types {
+		if reqZones.Difference(sets.NewString(found[t]...)).Len() == 0 {
```
staebler, line 34: I understand that this is existing code, but let's fix this.
```suggestion
		if sets.NewString(found[t]...).HasAll(zones...) {
```

pkg/asset/machines/aws/instance_types.go
```diff
@@ -23,7 +24,50 @@ func PreferredInstanceType(ctx context.Context, meta *awsconfig.Metadata, types
...
+
+	reqZones := sets.NewString(zones...)
+	if reqZones.Difference(sets.NewString(found[instanceType]...)).Len() == 0 {
+		return zones, nil
+	}
+
+	for k := range reqZones.Difference(sets.NewString(found[types[0]]...)) {
+		reqZones.Delete(k)
+	}
+	return reqZones.List(), nil
```
staebler, line 64:
```suggestion
	return sets.NewString(found[instanceType]...).Intersection(sets.NewString(zones...)).List(), nil
```

pkg/types/aws/defaults/platform.go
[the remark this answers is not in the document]
staebler: I still have concerns that this is ignored if there are default machine classes. The `defaultMachineClass` should be based on architecture first and region second.

pkg/version/version.go
```diff
@@ -26,6 +28,12 @@ var (
 	// Set in hack/build.sh.
 	Commit = ""
 
+	// HostArch is the payload architecture for which the installer was built,
```
staebler, line 31: Why do we care about the host architecture? I would expect instead to have something like `defaultReleaseArchitecture`, which is would be populated when extracting the installer from the release image.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -47,18 +48,32 @@ func Validate(ctx context.Context, meta *Metadata, config *types.InstallConfig)
 	allErrs = append(allErrs, validateAMI(ctx, config)...)
 	allErrs = append(allErrs, validatePlatform(ctx, meta, field.NewPath("platform", "aws"), config.Platform.AWS, config.Networking, config.Publish)...)
 
-	if config.ControlPlane != nil && config.ControlPlane.Platform.AWS != nil {
-		allErrs = append(allErrs, validateMachinePool(ctx, meta, field.NewPath("controlPlane", "platform", "aws"), config.Platform.AWS, config.ControlPlane.Platform.AWS, controlPlaneReq)...)
+	if config.ControlPlane != nil {
+		allErrs = append(allErrs, validateArchitecture(ctx, field.NewPath("controlPlane", "architecture"), config.ControlPlane.Architecture)...)
```
staebler, line 52: This is not the correct place for this validation. Static validation should be in `types/aws/validation`.

pkg/types/machinepools.go
```diff
@@ -34,8 +34,16 @@ const (
 	ArchitectureS390X = "s390x"
 	// ArchitecturePPC64LE indicates ppc64 little endian (Power PC)
 	ArchitecturePPC64LE = "ppc64le"
+	// ArchitectureARM64 indicates arm (aarch64) systems
+	ArchitectureARM64 = "arm64"
 )
 
+// ArchitectureList contains all possible architectures
+var ArchitectureList = []Architecture{ArchitectureAMD64, ArchitectureARM64, ArchitecturePPC64LE, ArchitectureS390X}
```
staebler, line 42: This is unused.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -47,18 +48,32 @@ func Validate(ctx context.Context, meta *Metadata, config *types.InstallConfig)
...
 		}
 	}
 	return allErrs.ToAggregate()
 }
 
+func validateArchitecture(ctx context.Context, fldPath *field.Path, configArch types.Architecture) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	index := sort.SearchStrings(*(*[]string)(unsafe.Pointer(&types.ArchitectureListAWS)), string(configArch))
+	if index == len(types.ArchitectureListAWS) || types.ArchitectureListAWS[index] != configArch {
```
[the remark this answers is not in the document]
Prashanth684: the default value of the architecture is set to `amd64` when setting machine pool defaults

pkg/asset/installconfig/aws/validation.go
```diff
@@ -47,18 +48,32 @@ func Validate(ctx context.Context, meta *Metadata, config *types.InstallConfig)
...
 			allErrs = append(allErrs, validateMachinePool(ctx, meta, fldPath.Child("platform", "aws"), config.Platform.AWS, compute.Platform.AWS, computeReq)...)
 		}
 	}
 	return allErrs.ToAggregate()
 }
 
+func validateArchitecture(ctx context.Context, fldPath *field.Path, configArch types.Architecture) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	index := sort.SearchStrings(*(*[]string)(unsafe.Pointer(&types.ArchitectureListAWS)), string(configArch))
```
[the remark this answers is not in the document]
Prashanth684: i agree that using `unsafe` is dangerous - but in this case the type `Architecture` is nothing but the string type, so shouldn't that be ok?

pkg/asset/installconfig/aws/validation.go
```diff
@@ -80,8 +95,18 @@ func validatePlatform(ctx context.Context, meta *Metadata, fldPath *field.Path,
...
+		if sets.NewString(rhcos.AMIRegionsX86_64...).Has(config.Platform.AWS.Region) {
+			return nil
+		}
+	case types.ArchitectureARM64:
+		if sets.NewString(rhcos.AMIRegionsAARCH64...).Has(config.Platform.AWS.Region) {
+			return nil
+		}
+	default:
+		errMsg := fmt.Sprintf("No AMIs available for architecture: %s", config.ControlPlane.Architecture)
+		return field.ErrorList{field.Required(field.NewPath("platform", "aws", "amiID"), errMsg)}
```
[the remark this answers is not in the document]
Prashanth684: yes.corrected the error message

pkg/types/machinepools.go
Prashanth684: removed.

pkg/asset/machines/aws/instance_types.go
Prashanth684: i had to revert this back to `map[string][]string` because a list of strings is needed to create a new `sets.String`

pkg/asset/installconfig/aws/validation.go
```diff
@@ -47,18 +48,32 @@ func Validate(ctx context.Context, meta *Metadata, config *types.InstallConfig)
...
 	}
 	return allErrs.ToAggregate()
 }
 
+func validateArchitecture(ctx context.Context, fldPath *field.Path, configArch types.Architecture) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	index := sort.SearchStrings(*(*[]string)(unsafe.Pointer(&types.ArchitectureListAWS)), string(configArch))
+	if index == len(types.ArchitectureListAWS) || types.ArchitectureListAWS[index] != configArch {
+		allErrs = append(allErrs, field.Invalid(fldPath, configArch, "Architecture not supported"))
```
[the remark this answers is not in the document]
yselkowitz: `field.NotSupported`?

pkg/version/version.go
yselkowitz: Don't recall if we discussed such an option, but that would require support in (and therefore a matching version of) `oc`.  I thought we did settle on relying on the build architecture, in that an installer binary built in the containers is going to be part of a payload of the container build architecture, or `GOHOSTARCH`, which is why I named it `HostArch`.

staebler: I must be misunderstanding something about the build pipeline. Let's take for example that we want an installer that runs on amd64 and installs a cluster running arm64. In that case, will the container from which the build will be performed use the arm64 architecture, since the release image will be targeting arm64? And, then, will the build pipeline be doing a cross-architecture build for producing installers targeting all of the platforms that we support for the installer?

Let's assume that my understanding above is accurate. In that case, you are correct that we would not need the release architecture set during the extraction process. So we are good there. I still don't think that `HostArch` is a great name for this variable, however. That name refers to how the value is determined but not what we want the variable to represent.

I still have concerns about breaking our backwards compatibility by defaulting the target architecture to the host architecture. This may not be very problematic for OCP, but I could see it being problematic for OKD, where users do their own builds more frequently. Can we pass the target architecture as an optional argument to `./hack/build.sh` instead so that the target architecture can be set in the dockerfile?

pkg/asset/installconfig/aws/validation.go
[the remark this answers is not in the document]
staebler: Doh. Yep, that's the one.

pkg/asset/installconfig/aws/validation.go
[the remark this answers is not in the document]
staebler: It should be OK, but being able to reason about it being OK is not a compelling argument for using it. Using `unsafe` means that everyone coming to look at this later has to reason about it again, especially when making changes in this area.

pkg/types/aws/defaults/platform.go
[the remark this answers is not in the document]
Prashanth684: ah..ok...yes..i added machine classes based on arch.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -80,8 +95,18 @@ func validatePlatform(ctx context.Context, meta *Metadata, fldPath *field.Path,
...
+		if sets.NewString(rhcos.AMIRegionsX86_64...).Has(config.Platform.AWS.Region) {
+			return nil
+		}
+	case types.ArchitectureARM64:
+		if sets.NewString(rhcos.AMIRegionsAARCH64...).Has(config.Platform.AWS.Region) {
+			return nil
+		}
+	default:
+		errMsg := fmt.Sprintf("Architecture: %s is not supported", config.ControlPlane.Architecture)
+		return field.ErrorList{field.Required(field.NewPath("platform", "aws", "amiID"), errMsg)}
```
staebler, line 109: The error message talks about architecture, but the field that the error points to is the AMI ID.
```suggestion
		return field.ErrorList{field.NotSupported(field.NewPath("controlPlane", "architecture"), config.ControlPlane.Architecture, ...)}
```

pkg/asset/installconfig/aws/validation.go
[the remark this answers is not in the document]
staebler: Ah, right. And that default needs to be changed to the default architecture from `pkg/version/version.go`.

pkg/asset/machines/aws/instance_types.go
staebler: I don't understand your comment.

pkg/asset/installconfig/aws/validation.go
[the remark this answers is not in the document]
Prashanth684: done.

pkg/asset/installconfig/aws/validation.go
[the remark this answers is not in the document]
Prashanth684: made the ArchitectureListAWS a list of strings as it is only used for validation.

pkg/asset/machines/aws/instance_types.go
Prashanth684: i had committed this suggestion and then realized that `getInstanceTypeZoneInfo` needs to return a a list of strings because `FilterZonesBasedOnInstanceType` and `PreferredInstanceType` use the list to create a new `sets.String` type to act on it. i could make it return `sets.String` so we don't have to call `sets.NewString()`.

staebler: Yes, that is my suggestion. Return a map of `sets.String` here so that callers don't have to convert the slice into a `sets.String`.

pkg/asset/installconfig/aws/validation.go
Prashanth684: fixed and added as a separate commit: https://github.com/openshift/installer/pull/4870/commits/03d6b77734f452aa9a8ab76533a3c207c5715913

pkg/asset/machines/aws/instance_types.go
Prashanth684: done.

pkg/version/version.go
yselkowitz: You understand the build pipeline correctly.
I have renamed the variable to `defaultArch`.
Given that your average user does not have access to P&Z, and the installer was only recently made buildable on arm64, and any other architecture would still default to amd64, what is the case that you worry about this breaking?

staebler: My concern is that an install-config.yaml that produced a cluster in OpenShift 4.8 should continue to produce the same cluster in OpenShift 4.9. An install-config.yaml that produced a cluster on my machine should produce the same cluster on your machine, regardless of the architecture of our machines. I thought that was the whole point of this effort, namely to isolate the architecture of the machine running the installer from the architecture of the cluster created.

yselkowitz: Right now, if you `create install-config` with 4.8 for aws, you get `compute[0].architecture: amd64` and `controlPlane.architecture: amd64`.  There should be no reason that this file would not work after this PR.  The point of this PR is to change what `create install-config` writes to match the payload regardless of where it is *run*, so that even a linux/amd64 or darwin/amd64 installer from an ARM payload (and hence *built* on ARM) will write `arm64` there, and the user doesn't have to modify the install-config even in that use case.

ausil: It is not clear to me that we have articulated clearly the problem that we are trying to solve here. Perhaps our best path forward here is to exclude this change from the patchset so that we can get arm64 enabled, then followup with an issue and/or call to articulate and demonstrate the issues that we see and are trying to solve with this particular piece of code

yselkowitz: Right now, when someone downloads an installer from an x86 payload (which could run either on linux or mac) and chooses to install to AWS, the install-config.yaml generated by `create install-config` "just works", iow no modification is *necessary* to the yaml to get a working, default cluster.  The purpose this serves is to provide the very same experience for ARM payloads: that an installer from an ARM payload -- regardless of where it *runs* -- will write an install-config that "just works", without modification, to install that OpenShift payload to ARM on AWS.

staebler: The `controlPlane.architecture` field is optional. If a user is supplying their own install-config.yaml file, rather than using `create install-config`, the `controlPlane.architecture` field can be omitted. In 4.8, the installer would default the architecture to amd64. If the user then takes the same install-config.yaml, with the field omitted, and does an installation on 4.9, the installer must continue to default to amd64. It cannot default to whatever the host architecture is where the installer is run.

yselkowitz: So you're asking us to separate the `create install-config` default from the `create cluster` if unspecified default.  OK, @Prashanth684 any ideas on how to do that?

yselkowitz: As for setting the arch in the Dockerfile, I just pushed a new commit to that end.

Prashanth684: @yselkowitz i think that investigation warrants a separate discussion as a separate issue/PR. i do not want to pollute this PR with more changes back and forth on this. this is not that trivial a problem to solve and also an important one to get it done right, so i suggest breaking this out.

staebler:
> Any idea why the openstack tests are failing so quickly?

They were broken by #4999.

pkg/types/aws/defaults/platform.go
```diff
@@ -1,14 +1,19 @@
...
-	defaultMachineClass = map[string][]string{
-		// Example region default machine class override:
+	defaultMachineClassAMD64 = map[string][]string{
+		// Example region default machine class override for AMD64:
 		// "ap-east-1":      {"m5", "m4"},
 	}
+	defaultMachineClassARM64 = map[string][]string{
+		// Example region default machine class override for ARM64:
+		// "us-east-1":      {"m6g", "m6gd"},
+	}
```
staebler, line 16:
```suggestion
	defaultMachineClass = map[types.Architecture]map[string][]string{
		types.ArchitectureAMD64: {
			// Example region default machine class override for AMD64:
			// "ap-east-1":      {"m5", "m4"},
		},
		types.ArchitectureARM64: {
			// Example region default machine class override for ARM64:
			// "us-east-1":      {"m6g", "m6gd"},
		},
	}
```

pkg/types/aws/defaults/platform.go
```diff
@@ -17,17 +22,33 @@ func SetPlatformDefaults(p *aws.Platform) {
 
 // InstanceClass returns the instance "class" we should use for a given
 // region. Default is m5 unless a region override is defined in defaultMachineClass.
-func InstanceClass(region string) string {
-	if classes, ok := defaultMachineClass[region]; ok {
+func InstanceClass(region string, arch types.Architecture) string {
+	if arch == types.ArchitectureARM64 {
```
staebler, line 26:
```
if classesForArch, ok := defaultMachineClass[arch]; ok {
  if classes, ok := classesForArch[region]; ok {
    return classes[0]
  }
}

switch arch {
case types.ArchitectureARM64:
  return "m6g"
default:
  return "m5"
}
```

yselkowitz:
Any remaining issues to address here?

pkg/version/version.go
```diff
@@ -62,3 +70,17 @@ func Version() (string, error) {
...
+	switch defaultArch {
+	case "arm64":
+		return types.ArchitectureARM64
+	case "ppc64le":
+		return types.ArchitecturePPC64LE
+	case "s390x":
+		return types.ArchitectureS390X
+	default:
+		return types.ArchitectureAMD64
+	}
```
staebler, line 85: Are the values of defaultArch different than the values of types.Architecture? We want to end up with a validation error when an unknown `defaultArch` is used, so we don't want to default in this function to amd64.
```suggestion
	return types.Architecture(defaultArch)
```

pkg/types/aws/validation/machinepool.go
```diff
@@ -43,3 +45,14 @@ func ValidateAMIID(platform *aws.Platform, p *aws.MachinePool, fldPath *field.Pa
 	}
 	return allErrs
 }
+
+// ValidateMachinePoolArchitecture checks that a valid architecture is set for a machine pool.
+func ValidateMachinePoolArchitecture(pool *types.MachinePool, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	index := sort.SearchStrings(types.ArchitectureListAWS, string(pool.Architecture))
```
staebler, line 53: This is fragile because it requires `types.ArchitectureListAWS` to be sorted.
I recommend you set up a map of acceptable values like [here](https://github.com/openshift/installer/blob/4e60a8e317a164d3645bf370f303094fe2e18f8c/pkg/types/validation/machinepools.go#L43-L47).

pkg/asset/machines/aws/instance_types.go
```diff
@@ -23,7 +24,42 @@ func PreferredInstanceType(ctx context.Context, meta *awsconfig.Metadata, types
...
+	found, err := getInstanceTypeZoneInfo(ctx, sess, meta.Region, types, zones)
+	if err != nil {
+		return zones, err
+	}
+
+	return found[instanceType].Intersection(sets.NewString(zones...)).List(), nil
+}
+
+func getInstanceTypeZoneInfo(ctx context.Context, session *session.Session, region string, types []string, zones []string) (map[string]sets.String, error) {
+	found := map[string][]string{}
```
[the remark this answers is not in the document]
Prashanth684: ah..ok..there is an insert function..great..changed.

pkg/types/aws/defaults/platform.go
Prashanth684: commited this and the below change into a single commit

pkg/types/aws/defaults/platform.go
Prashanth684: done.

pkg/types/aws/validation/machinepool.go
Prashanth684: yup.done

pkg/asset/machines/aws/instance_types.go
```diff
@@ -38,17 +73,17 @@ func PreferredInstanceType(ctx context.Context, meta *awsconfig.Metadata, types
 		LocationType: aws.String("availability-zone"),
 	})
 	if err != nil {
-		return types[0], err
+		return found, err
 	}
-	reqZones := sets.NewString(zones...)
-	found := map[string][]string{}
+
+	// iterate through the offerings and create a new map of sets.String
```
staebler, line 79: This comment is not quite accurate any more.

pkg/types/aws/validation/machinepool.go
```diff
@@ -7,9 +7,25 @@ import (
...
 	"github.com/openshift/installer/pkg/types/aws"
 )
 
+var (
+	validArchitecturesAWS = map[types.Architecture]bool{
+		types.ArchitectureAMD64: true,
+		types.ArchitectureARM64: true,
+	}
+
+	validArchitectureValuesAWS = func() []string {
```
staebler, line 20: This is the same as `types.ArchitectureListAWS`, right? Can we keep this in a single place, probably best in this package?

staebler (commented):
Looks good save for a couple nits. Let's squash this down into one or a few commits.

pkg/asset/machines/aws/instance_types.go
Prashanth684: changed.

pkg/types/aws/validation/machinepool.go
Prashanth684: done.removed the `types.ArchitectureListAWS`

Prashanth684:
> Looks good save for a couple nits. Let's squash this down into one or a few commits.

squashed the later commits relating to fixes in the comments.

pkg/types/aws/validation/machinepool.go
```diff
@@ -7,9 +7,26 @@ import (
...
 )
 
+var (
+	validArchitecturesAWS = map[types.Architecture]bool{
+		types.ArchitectureAMD64: true,
+		types.ArchitectureARM64: true,
+	}
+
+	// ValidArchitectureValuesAWS lists the supported arches for AWS
+	ValidArchitectureValuesAWS = func() []string {
```
staebler, line 21: The AWS in these fields is redundant with the package name.
```suggestion
	ValidArchitectureValues = func() []string {
```

staebler (commented):
Please put the changes in the last commit into their appropriate earlier commits. I do not want a commit in the history that is "miscellaneous fixes".

pkg/types/aws/validation/machinepool.go
yselkowitz: Renamed and squashed.

staebler (approved):
/lgtm

PR #4872, opened 2021-04-25
Bug 1951670: Error gathering bootstrap log after pivot: The bootstrap…

eranco74:
… machine did not execute the release-image.service systemd unit

In case of installing single node OpenShift with bootstrap in place, the bootstrap gather artifacts prior to the node reboot are added to the log-bundle under "log-bundle-bootstrap"
Since the analyze code didn't expect that extra folder it failed to find them and logged the error.

Added optional "/log-bundle-bootstrap" to the serviceEntriesFilePathRegex.
This allows the rest of the analyze code to work with the bootstrap-in-place gather artifacts collected prior to the pivot


staebler (commented):
/lgtm

PR #4876, opened 2021-04-27
Terraform: Split AWS into separate stages

jhixson74:
This allows for more control over the terraform stages. It's also a fix to
newer versions of terraform that destroy cluster resources when destroying
bootstrap resources.

https://issues.redhat.com/browse/CORS-1693


cmd/openshift-install/gather.go
```diff
@@ -136,17 +136,17 @@ func runGatherBootstrapCmd(directory string) (string, error) {
 		return "", errors.Wrapf(err, "failed to fetch %s", config.Name())
 	}
 
-	tfstate, err := terraform.ReadState(tfStateFilePath)
+	tfstate, err := terraform.ReadState(tfBootstrapStateFilePath)
 	if err != nil {
-		return "", errors.Wrapf(err, "failed to read state from %q", tfStateFilePath)
+		return "", errors.Wrapf(err, "failed to read state from %q", tfBootstrapStateFilePath)
 	}
 	bootstrap, port, masters, err := extractHostAddresses(config.Config, tfstate)
```
patrickdillon, line 143: Does this now need to be modified to get the IP addresses from both tf state files? The information is now split between two files.

patrickdillon (commented):
I have not been knee deep in terraform so I may only have a high-level understanding of how these internals are working, but I enjoyed poking around and left some comments/questions that I hope will be helpful.

I think this could benefit from some high-level design doc, either in the pr or jira or wherever. Nothing too in-depth, but what is popping up to me at first glance is how are we going to handle the resource dependencies, e.g. VPC, between the two directories? 

It would seem to me that we would need to split further than bootstrap/cluster so that resources like the VPC could be effectively shared. Perhaps by breaking into more dirs and using [dependency inversion](https://www.terraform.io/docs/language/modules/develop/composition.html#dependency-inversion). There are definitely some drawbacks to this suggestion, such as losing terraform's dependency management.

data/data/aws/bootstrap/main.tf
```diff
@@ -1,196 +1,86 @@
...
-    # avoids accidental deletion of nodes whenever a new OS release comes out.
-    ignore_changes = [ami]
-  }
-
-  tags = merge(
-    {
-      "Name" = "${var.cluster_id}-bootstrap"
-    },
-    var.tags,
+module "vpc" {
```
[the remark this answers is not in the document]
patrickdillon: I see how this works now: separate VPC modules for bootstrap & cluster. Cluster VPC module uses a data source, so that pattern looks good.

pkg/terraform/terraform.go
```diff
@@ -23,27 +24,49 @@ const (
...
 )
 
+// GetStateFileName returns the name of the terraform state file. If no target
+// is specified or compat is specified, default to the old state file name.
+func GetStateFileName(target string) string {
+	if target != "" && target != TargetCompat && !LegacyMode() {
+		return fmt.Sprintf("terraform.%s.tfstate", target)
+	}
+
+	return StateFileName
```
staebler, line 42: nit: The new format is the happy scenario. The old format is the exceptional case. This will also make it cleaner when it comes time to remove the old format.
```suggestion
    	if target == "" || target == "compat" {
    	   	return StateFileName
    	}
    	return fmt.Sprintf("terraform.%s.tfstate", target)
```
Although, I suggest that we can also just get rid of the old format entirely and use `terraform.compat.tfstate`.

pkg/terraform/terraform.go
```diff
@@ -23,27 +24,49 @@ const (
...
+	TargetBootstrap string = "bootstrap"
+
+	// TargetCompat is the name for the legacy terraform directory
+	TargetCompat string = "compat"
 )
 
+// GetStateFileName returns the name of the terraform state file. If no target
+// is specified or compat is specified, default to the old state file name.
+func GetStateFileName(target string) string {
+	if target != "" && target != TargetCompat && !LegacyMode() {
```
staebler, line 38: Why do we need to check for legacy mode?

pkg/terraform/exec/plugins/aws.go
```diff
@@ -3,13 +3,23 @@ package plugins
...
-	KnownPlugins["terraform-provider-aws"] = exec
+	KnownPlugins["terraform-provider-aws"] = &TFPlugin{
+		Name:      awstypes.Name,
+		Exec:      awsProvider,
+		Resources: Resources(awsResources),
+	}
+}
+
+func awsResources() []string {
+	return []string{"vpc", "bootstrap", "cluster"}
```
staebler, line 24: Is it possible to re-order this as `"cluster", "bootstrap"`, where the vpc part is done as part of the cluster target? I would expect that most platforms would do the bootstrap target after the cluster target.

staebler (commented):
Here are some initial comments on the go portion. Overall, I think this looks really good. I like that you were able to get all of the terraform done as part of the cluster asset instead of breaking it into multiple assets.

pkg/asset/cluster/cluster.go
```diff
@@ -129,13 +139,19 @@ func (c *Cluster) Files() []*asset.File {
 // Load returns error if the tfstate file is already on-disk, because we want to
 // prevent user from accidentally re-launching the cluster.
 func (c *Cluster) Load(f asset.FileFetcher) (found bool, err error) {
-	_, err = f.FetchByName(terraform.StateFileName)
-	if err != nil {
-		if os.IsNotExist(err) {
-			return false, nil
+	return false, nil
+
+	// XXX: We need to check for all the terraform state files here.
```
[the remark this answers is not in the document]
jhixson74: I have updated this to check for all terraform state files.

pkg/asset/cluster/cluster.go
```diff
@@ -92,32 +83,51 @@ func (c *Cluster) Generate(parents asset.Parents) (err error) {
...
-		err = errors.Wrap(err, "failed to create cluster")
-		if stateFile == "" {
-			return err
+		// Copy the terraform.tfvars to a temp directory where the terraform
+		// will be invoked within.
+		tmpDir, err := ioutil.TempDir("", fmt.Sprintf("openshift-install-%s-", r))
+		if err != nil {
+			return errors.Wrap(err, "failed to create temp dir for terraform execution")
+		}
+		//defer os.RemoveAll(tmpDir)
```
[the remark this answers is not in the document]
jhixson74: Yes, this was left in for CI debugging. It is removed now.

pkg/asset/cluster/cluster.go
```diff
@@ -92,32 +83,51 @@ func (c *Cluster) Generate(parents asset.Parents) (err error) {
...
+
+		extraArgs := []string{}
```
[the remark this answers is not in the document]
jhixson74: This seems more of a style recommendation than actual fix. I simply moved this block of code. I'm happy to change it, however, this is handled the same way in other files (with the same name!) the same. I can change all of them, or, even better, can we come up with a style guide (or use an existing one) so that style issues aren't an issue in PR's?

pkg/terraform/exec/plugins/azure.go
```diff
@@ -11,5 +13,9 @@ func init() {
 			ProviderFunc: azurerm.Provider,
 		})
 	}
-	KnownPlugins["terraform-provider-azurerm"] = azurermProvider
+	KnownPlugins["terraform-provider-azurerm"] = &TFPlugin{
+		Name:      azuretypes.Name,
+		Exec:      azurermProvider,
+		Resources: Resources(nil),
```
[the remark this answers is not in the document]
jhixson74: The original purpose for this is no longer necessary, so this has been removed and is now just a slice of strings.

pkg/terraform/exec/plugins/plugins.go
```diff
@@ -1,5 +1,66 @@
...
 
+import (
+	"os"
+	"strconv"
+
+	"github.com/pkg/errors"
+)
+
+// TFPluginFunc is a callback function that runs the terraform plugin
+type TFpluginFunc func()
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/terraform/terraform.go
```diff
@@ -23,27 +24,46 @@ const (
 
 	// VarFileName is the default name for Terraform var file.
 	VarFileName string = "terraform.tfvars"
+
+	// TargetBootstrap is the default name for the bootstrap directory
+	TargetBootstrap string = "bootstrap"
```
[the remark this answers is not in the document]
jhixson74: The PR was out of date when you reviewed this, and indeed was using this constant ;-) It's now updated.

pkg/destroy/bootstrap/bootstrap.go
```diff
@@ -32,14 +32,15 @@ func Destroy(dir string) (err error) {
...
-	tempDir, err := ioutil.TempDir("", "openshift-install-")
+	tempDir, err := ioutil.TempDir("", "openshift-install-bootstrap-")
 	if err != nil {
 		return errors.Wrap(err, "failed to create temporary directory for Terraform execution")
 	}
 	defer os.RemoveAll(tempDir)
 
 	extraArgs := []string{}
-	for _, filename := range []string{terraform.StateFileName, cluster.TfVarsFileName, tfPlatformVarsFileName} {
+	stateFileName := terraform.GetStateFileName("bootstrap")
```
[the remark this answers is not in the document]
jhixson74: Updated.

pkg/terraform/terraform.go
jhixson74: I've gone with your second recommendation here. The original goal was to keep the code compatible with the way the original terraform worked. This makes things easier.

pkg/terraform/terraform.go
jhixson74: Legacy mode was supposed to determine if running in legacy mode ;-) It's not so important anymore so I've removed it.

pkg/terraform/exec/plugins/aws.go
jhixson74: We had previously discussed this and agreed to create the vpc separately, so that is what was done here. If we were to change it back to the way it was and move it into "cluster", how would this even work? cluster requires the bootstrap resources to have been created, unless I am misunderstanding something here. I can say that this currently works "as-is" without any issues.

pkg/asset/cluster/cluster.go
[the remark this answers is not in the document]
staebler: It is a bit more than just style, but I am fine leaving it as is.

The reason why `var extraArgs []string` is generally better is because it does not allocate the array if the slice ends up not being written to. In this case, it would be even better to allocate the array to the size needed since we know the size at this point (ie, `len(terraformVariables.Files())`).

pkg/terraform/exec/plugins/aws.go
staebler: That is not my recollection of where we left the previous discussion, but we can leave it as is for now.

Which bootstrap resources are required for the cluster resources? The bootstrap resources require cluster resources, specifically the vpc resources that you had to break out separately. I am not aware of any cluster resources that require bootstrap resources, however. But I admittedly have not dug into the tf file changes in depth yet.

data/data/aws/bootstrap/vpc/master-elb.tf
```diff
@@ -0,0 +1,74 @@
+data "aws_lb" "api_internal" {
```
staebler, line 1: Do we really need all of these `data` objects? All I see being used from the vpc module are the target group ARNs, the VPC ID, and the master security group ID.

data/data/aws/cluster/main.tf
```diff
@@ -0,0 +1,115 @@
...
+    concat(
+      var.aws_master_availability_zones,
+      var.aws_worker_availability_zones,
+    ),
+  )
+
+  tags = local.tags
+}
+
+resource "aws_ami_copy" "imported" {
```
staebler: We should not be doing this AMI copy in both the cluster and bootstrap tf file.

pkg/asset/cluster/cluster.go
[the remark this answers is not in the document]
jhixson74: I've updated the syntax to your preferred usage. The same usage is scattered in various other files I've seen and we should probably change it in those as well.

data/data/aws/bootstrap/vpc/master-elb.tf
jhixson74: I've removed  many of the data sources. There still may be some that are unused.

data/data/aws/cluster/main.tf
jhixson74: I've moved this to the vpc module and referenced it in the bootstrap and cluster modules.

pkg/terraform/exec/plugins/aws.go
jhixson74: Maybe you can explain this to me on slack. If running the bootstrap sequence after cluster is possible, why do we even have a bootstrap sequence? My understanding of how nearly all openshift platforms are setup is this (simplified):

1. Create network resources
2. Setup bootstrap host
3. Setup cluster (control plane which I thought required the bootstrap host be up and configured). 

What am I missing here?

staebler: The terraform is only creating the cloud resources, not directly running any OpenShift processes. In most cases, the control plane machines can be created before, after, or concurrently with the bootstrap machine. After the machines are created, the bootstrapping processes run on the bootstrap machine, and those processes allow the control plane machines to grab the ignition configs to ignite.

The notable exception to this is for Azure. Azure does not report its machines as created until after the machines ignite. So far Azure, the "cluster" setup would need to go after the "bootstrap" setup.

Let's leave the order as you have it for now. We can shuffle things around later (or not).

data/data/aws/bootstrap/vpc/master-elb.tf
jhixson74: It seems that anything more removed causes the install to fail. I think everything here is required.

pkg/terraform/exec/plugins/aws.go
jhixson74: Thank you for the explanation. I can certainly test out using a different order, however, I suspect it won't go as planned ;-) I think the original Terraform and code was written with assumptions on the order and probably requires more reworking.

jhixson74: It looks like you were right. I changed the ordering to have bootstrap last and the cluster installed successfully!

pkg/terraform/exec/plugins/ignition.go
```diff
@@ -6,10 +6,14 @@ import (
 )
 
 func init() {
-	exec := func() {
+	ignitionProvider := func() {
```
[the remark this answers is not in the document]
jhixson74: Yup

staebler (commented):
This is looking very close. My biggest remaining concern is the conflation of plugin and platform.

staebler:
We are getting the following error pretty frequently.
```
level=error msg=Error: Error launching source instance: InvalidParameterValue: Value (ci-op-fmxfm9j5-1d3f3-2x68t-bootstrap-profile) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name
level=error msg=	status code: 400, request id: d1177d97-64e4-4a55-84ff-b55e4e54a82c
level=error
level=error msg=  on ../tmp/openshift-install-bootstrap-234577263/bootstrap/main.tf line 111, in resource "aws_instance" "bootstrap":
level=error msg= 111: resource "aws_instance" "bootstrap" {
```
Then the installer incorrectly proceeds to the next stop of waiting for the cluster to bootstrap rather than failing immediately.

pkg/asset/cluster/cluster.go
```diff
@@ -92,32 +84,51 @@ func (c *Cluster) Generate(parents asset.Parents) (err error) {
...
+				Data:     data,
+			})
+		} else if err == nil {
+			err = err2
+		} else {
+			logrus.Errorf("Failed to read tfstate: %v", err2)
+		}
+
+		timer.StopTimer(r)
 	}
```
staebler, line 130: There should be the following here at the end of each loop.
```
if err != nil {
  return err
}
```

staebler:
> We are getting the following error pretty frequently.
> 
> ```
> level=error msg=Error: Error launching source instance: InvalidParameterValue: Value (ci-op-fmxfm9j5-1d3f3-2x68t-bootstrap-profile) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name
> level=error msg=	status code: 400, request id: d1177d97-64e4-4a55-84ff-b55e4e54a82c
> level=error
> level=error msg=  on ../tmp/openshift-install-bootstrap-234577263/bootstrap/main.tf line 111, in resource "aws_instance" "bootstrap":
> level=error msg= 111: resource "aws_instance" "bootstrap" {
> ```
> 
> Then the installer incorrectly proceeds to the next stop of waiting for the cluster to bootstrap rather than failing immediately.

The terraform errors are a race where the instance is created before the instance profile is actually ready. According to [this](https://github.com/hashicorp/terraform/issues/15341#issuecomment-360965525), the issue was fixed in v1.7.1 of the aws provider. But we are using v2.67.0, which should already have the fix.

This race exists in the current code. We just don't see it because the instance profile is created pretty early whereas the instance has to wait for a bunch of other resources to be created first (namely, the resources that are now being created in the other stages).

staebler:
The other platforms are failing with errors such as the following.
```
 INFO[2021-06-11T03:16:24Z] level=fatal msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to unpack Terraform modules: open /gcp/compat: file does not exist 
```

jhixson74:
> The other platforms are failing with errors such as the following.
> 
> ```
>  INFO[2021-06-11T03:16:24Z] level=fatal msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to unpack Terraform modules: open /gcp/compat: file does not exist 
> ```

The other platforms are not dealt with this in this PR. See PR #4909 for those.

staebler:
> > The other platforms are failing with errors such as the following.
> > ```
> >  INFO[2021-06-11T03:16:24Z] level=fatal msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to unpack Terraform modules: open /gcp/compat: file does not exist 
> > ```
> 
> The other platforms are not dealt with in this PR. See PR #4909 for those.

I don't understand this comment. I understood #4909 to be just a PR that you were using for testing. My expectation is that the other platforms should continue to work if we were to merge this PR.

jhixson74:
> > > The other platforms are failing with errors such as the following.
> > > ```
> > >  INFO[2021-06-11T03:16:24Z] level=fatal msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to unpack Terraform modules: open /gcp/compat: file does not exist 
> > > ```
> > 
> > 
> > The other platforms are not dealt with in this PR. See PR #4909 for those.
> 
> I don't understand this comment. I understood #4909 to be just a PR that you were using for testing. My expectation is that the other platforms should continue to work if we were to merge this PR.

I've updated this PR to have the changes from #4909. Now the the other platforms should work fine.

staebler:
> > > > The other platforms are failing with errors such as the following.
> > > > ```
> > > >  INFO[2021-06-11T03:16:24Z] level=fatal msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to unpack Terraform modules: open /gcp/compat: file does not exist 
> > > > ```
> > > 
> > > 
> > > The other platforms are not dealt with in this PR. See PR #4909 for those.
> > 
> > 
> > I don't understand this comment. I understood #4909 to be just a PR that you were using for testing. My expectation is that the other platforms should continue to work if we were to merge this PR.
> 
> I've updated this PR to have the changes from #4909. Now the the other platforms should work fine.

Ah. I get it now. The other PR was moving non-split platforms down into the compat folder.
For what it's worth, my PR got around this a different way. But the compat stuff is all temporary, so whatever works is fine.

staebler:
This has been combined into #5010.
/close

PR #4879, opened 2021-04-27
azure: don't require BaseDomainResourceGroupName on ARO

m1kola:
This prevents a validating failure when omitting `BaseDomainResourceGroupName` configurable which is never used in ARO.


m1kola:
Rebased on top of the `master`.

/test e2e-azure
/test e2e-azure-upi
/test e2e-azure-shared-vpc

jhixson74 (approved):

PR #4883, opened 2021-04-27
Bug 1943175: azure: set storage account tls default to 1.2

e-tienne:
Set azure storage account TLS version default to 1.2
    
    Set the default TLS version to 1.2 for azure storage
    accounts. We only set this attribute in public and
    usgovernment clouds as it's not supported in other
    ones. The azurerm upgrade also sets  to  by default.


data/data/azure/main.tf
```diff
@@ -146,6 +146,7 @@ resource "azurerm_storage_account" "cluster" {
   location                 = var.azure_region
   account_tier             = "Standard"
   account_replication_type = "LRS"
+  min_tls_version          = (var.azure_environment == "public" || var.azure_environment == "usgovernment") ? "TLS1_2" : null
```
staebler, line 149:
```suggestion
  min_tls_version          = contains(["public, "usgovernment"], var.azure_environment) ? "TLS1_2" : null
```
Or even move the `["public", "usgovernment"]` list to a local variable named something like `envirornments_with_tls1_1` to provide some hint from the name of the variable about why we are only choosing to use TLS 1.2 in those environments.

staebler (commented):
Nice work!

staebler (commented):
Looks good. Can you squash the last commit into the previous commit?

PR #4875, opened 2021-04-26
Bug 1953551: Link ppc64le binary dynamically

yselkowitz:
Recent versions of Go have had issues statically linking some ppc64le
binaries:

https://github.com/golang/go/issues/45564

/cc @joepvd

PR #4890, opened 2021-04-29
Bug 1953551: Revert "Link ppc64le binary dynamically"

yselkowitz:
This unblocked the build on other architectures, but the resulting
ppc64le binary had issues.  The golang bug which affected static linking
has been fixed with a one-line patch, which we are carrying downstream
until the upstream backports land.

This reverts commit 24ac0a15e785b2cabcdd4af2c1281efe8d29eed6.

/cc @joepvd @wking @sosiouxme

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -42,13 +42,11 @@ if (echo "${TAGS}" | grep -q 'libvirt')
 then
 	export CGO_ENABLED=1
 fi
-case "$(go env GOARCH)" in
-arm64|ppc64le)
-	# arm64: https://github.com/golang/go/issues/40492
-	# ppc64le: https://github.com/golang/go/issues/45564
+if test "$(go env GOARCH)" = "arm64"
+then
+	# https://github.com/golang/go/issues/40492
 	LDFLAGS="${LDFLAGS} -linkmode external"
 	export CGO_ENABLED=1
-	;;
-esac
+fi
 
 go build "${GOFLAGS}" -ldflags "${LDFLAGS}" -tags "${TAGS}" -o "${OUTPUT}" ./cmd/openshift-install


yselkowitz:
Holding this as a draft until the patched golang lands in a builder.

sosiouxme:
@yselkowitz patched golang builder is built and hopefully syncing out to CI shortly

yselkowitz:
/retest
Still need a LGTM and probably some overrides to get this in ASAP.

staebler:
> /retest
> Still need a LGTM and probably some overrides to get this in ASAP.

Can someone post the output from running an installer binary built with this revert? I am not aware of any tests jobs that we have in the installer repo that even build for the ppc64le architecture let alone run the installer.

dbenoit17:
Without this revert:
```[installer]$ ./bin/openshift-install version
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x0]goroutine 1 [running]:
github.com/aws/aws-sdk-go/service/efs.init()
	/root/installer/vendor/github.com/aws/aws-sdk-go/service/efs/errors.go:194 +0x24
```

After applying the PR:
```
[installer]$ ./bin/openshift-install version
./bin/openshift-install unreleased-master-4536-gb330ac1b5e2d62b89300e770c0f624d37b96d9fc-dirty
built from commit b330ac1b5e2d62b89300e770c0f624d37b96d9fc
release image registry.ci.openshift.org/origin/release:4.8
```

One additional thing to note is that [this patch](https://github.com/golang/go/commit/983dea90c169930e35721232afe39fd4e3fbe4a6) needs to be applied to the Go compiler to fix the original issue https://github.com/openshift/installer/pull/4875.  The CI builders should already have a compatible version of Go.

staebler:
> One additional thing to note is that [this patch](https://github.com/golang/go/commit/983dea90c169930e35721232afe39fd4e3fbe4a6) needs to be applied to the Go compiler to fix the original issue #4875. The CI builders should already have a compatible version of Go.

Do we need to change https://github.com/openshift/installer/blob/master/docs/dev/dependencies.md#go to reflect that we need to use a go version with this change? Or is this a fix specific to go 1.16, and go 1.14 should continue to work?

dbenoit17:
It looks like upstream backports are under consideration for Go 1.16.  It could be worth updating the minimum required Go version if/when backports land, but currently the compiler patches to fix the issue have not landed in a stable upstream release.

yselkowitz:
Still need a LGTM, and possibly an override
/retest

yselkowitz:
Test error seems unrelated, please fix or override:
```

Error downloading packages:
  azure-cli-2.23.0-1.el7.x86_64: [Errno 256] No more mirrors to try.
```

staebler (approved):
/lgtm

PR #4891, opened 2021-04-30
Add linux-amd64 binary to installer-artifacts

yselkowitz:


yselkowitz:
Per our discussions on adding this to oc
/cc @soltysh

images/installer-artifacts/Dockerfile.rhel
```diff
@@ -4,8 +4,8 @@
 FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.8 AS builder
 WORKDIR /go/src/github.com/openshift/installer
 COPY . .
-RUN go generate ./data && \
-    SKIP_GENERATION=y GOOS=darwin GOARCH=amd64 hack/build.sh
+RUN make cross-build-darwin-amd64 cross-build-linux-amd64
```
wking, line 7: Why do you need to build for Linux here?  Don't we get that via the `FROM ...:installer` line below?

staebler:
@abhinavdahiya From comments in past PRs that have tried to add a makefile, you seem to be the biggest advocate against adding a makefile. I would love your input here on whether adding a makefile to ease building for cross-platform architectures would alleviate your concerns that using make is no easier that using the build script.

images/installer-artifacts/Dockerfile.rhel
yselkowitz: We want to have a linux/amd64 installer even in the other architecture's payloads.  For remote deployments, just like we enable installing from mac regardless of cluster architecture, we also want to enable doing such from Linux on commodity hardware, which is still most commonly x86.

images/installer-artifacts/Dockerfile.rhel
```diff
@@ -1,11 +1,16 @@
 # This Dockerfile builds an image containing the Mac version of the installer layered
 # on top of the Linux installer image.
```
staebler, line 2: Is this comment still accurate?

images/installer-artifacts/Dockerfile.rhel
```diff
@@ -1,11 +1,16 @@
...
+
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.8 AS linuxbuilder
+WORKDIR /go/src/github.com/openshift/installer
+COPY . .
+RUN make cross-build-linux-amd64
 
 FROM registry.ci.openshift.org/ocp/4.8:installer
-COPY --from=builder /go/src/github.com/openshift/installer/bin/openshift-install /usr/share/openshift/mac/openshift-install
+COPY --from=macbuilder /go/src/github.com/openshift/installer/_output/bin/darwin_amd64/openshift-install /usr/share/openshift/mac/openshift-install
+COPY --from=linuxbuilder /go/src/github.com/openshift/installer/_output/bin/linux_amd64/openshift-install /usr/share/openshift/linux_amd64/openshift-install
```
staebler, line 16: Help me better understand this. Is ART replacing `registry.ci.openshift.org/ocp/4.8:installer` with architecture-specific images based on the target architecture of the OpenShift release image?
Depending on the architecture of the release image, this binary may be a duplicate of the one from `registry.ci.openshift.org/ocp/4.8:installer`, right? But we want to include the duplicate
1. so that `oc` has a consistent way of extracting the installer binary for the user's desired architecture and
2. because there is no good way to use a link instead since the Dockerfile does not know which architecture ART is going to use when replacing `registry.ci.openshift.org/ocp/4.8:installer`.

staebler (commented):
Please keep the makefile changes as a separate PR so that we can discuss the merits of those changes independently.

images/installer-artifacts/Dockerfile.rhel
yselkowitz: That pretty well describes it.

yselkowitz:
don't see how this affecths the upgrade test
/retest
/assign @staebler

staebler (approved):
/lgtm

PR #4894, opened 2021-04-30
Bug 1955697: tfvars/vsphere: use explicit path for datacenter.

jstuever:
This change specifies the explicit path to the datacenter when passing
it to Terraform. Prior to this change, only the datacenter name was
passed, which caused a collision when the same name was used by
datacenters in different paths.


staebler (commented):
/lgtm

staebler:
> /lgtm
> e2e-vsphere passed

/hold cancel

PR #4895, opened 2021-04-30
Bug 1947067: vendor/AlecAivazis/survey: Bump API to v2

mtnbikenc:
Updated survey.AskOne to pass survey.WithValidator option instead of function

https://github.com/AlecAivazis/survey/releases/tag/v2.0.0


pkg/asset/installconfig/aws/basedomain.go
```diff
@@ -61,14 +61,16 @@ func GetBaseDomain() (string, error) {
...
+	},
+		&domain,
+		survey.WithValidator(func(ans interface{}) error {
+			choice := ans.(string)
+			i := sort.SearchStrings(publicZones, choice)
+			if i == len(publicZones) || publicZones[i] != choice {
+				return errors.Errorf("invalid base domain %q", choice)
+			}
+			return nil
+		})); err != nil {
```
staebler, line 73: This indention does not look quite right. The parameters are indented at different levels.
```suggestion
		},
		&domain,
		survey.WithValidator(func(ans interface{}) error {
			choice := ans.(string)
			i := sort.SearchStrings(publicZones, choice)
			if i == len(publicZones) || publicZones[i] != choice {
				return errors.Errorf("invalid base domain %q", choice)
			}
			return nil
		}),
	); err != nil {
```

staebler (commented):
This looks good save for a couple nits.
Could you separate the vendoring changes (go.mod, go.sum, and vendor directory) into their own commit?

mtnbikenc:
> This looks good save for a couple nits.
> Could you separate the vendoring changes (go.mod, go.sum, and vendor directory) into their own commit?

I will update the indentation and split the vendoring into it's own commit.  I had considered doing that at first, but I was also trying to maintain each commit being a buildable commit.  Since vendoring would break code I decided to keep it together.

mtnbikenc:
@staebler I've split the changes into two commits and updated the indentation.  Please look it over and let me know if there is anything else.

mtnbikenc:
Updated `survey.Select` questions to use `core.OptionAnswer` answer types.
/hold cancel

staebler (approved):
/lgtm

PR #4904, opened 2021-05-05
Bug 1956208: openstack: add validation for volume types

EmilienM:
When an user wants to deploy with rootVolume, they need to specify
a volume type.
We already checked that the volume type was set but we didn't check that
the type was actually existing.

With this patch, we now check that the type exists.

If the Volume Type doesn't exist, it'll show this error message:
```
FATAL failed to fetch Metadata: failed to load asset "Install Config": compute[0].platform.openstack.rootVolume.type.type: Invalid value: "wrong": Volume Type either does not exist in this cloud, or is not available
```

Note: this patch also adds test coverage when the type is an empty string, which wasn't done before.

### DIFF
--- a/pkg/asset/installconfig/openstack/validation/cloudinfo.go
+++ b/pkg/asset/installconfig/openstack/validation/cloudinfo.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/availabilityzones"
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumetypes"
 	computequotasets "github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/quotasets"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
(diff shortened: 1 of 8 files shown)


pierreprinetti:
The failing CI tests are unrelated to this change. However, it seems that there are unit tests to fix

pkg/asset/installconfig/openstack/validation/cloudinfo.go
```diff
@@ -195,6 +197,11 @@ func (ci *CloudInfo) collectInfo(ic *types.InstallConfig, opts *clientconfig.Cli
 		return err
 	}
 
+	ci.VolumeTypes, err = ci.getVolumeTypes()
```
mandre, line 200: Possibly, we only gather the volume types when using root volumes.

mandre (commented):
/lgtm

mandre:
/assign @staebler 
For approving the vendoring change.

pkg/asset/installconfig/openstack/validation/cloudinfo.go
EmilienM: ack I'll do it later

EmilienM:
my patch has nothing to do with e2e-libvirt, can we override it please?

PR #4905, opened 2021-05-06
Bug 1953035: Disallow publish internal for non-cloud installations

rna-afk:
The publish: internal strategy is only available in cloud platforms,
GCP, Azure, AWS, Alibaba and IBM cloud, and should not be allowed as a
strategy for the other platforms.

Added a validation check to check if the strategy is internal and
the platform used is one of the valid cloud platforms. Check for IBM
and Alibaba is not yet added as they are not supported yet.

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -115,6 +115,14 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
 	}
 	allErrs = append(allErrs, validateCloudCredentialsMode(c.CredentialsMode, field.NewPath("credentialsMode"), c.Platform.Name())...)
 
+	if c.Publish == types.InternalPublishingStrategy {
+		switch platformName := c.Platform.Name(); platformName {
+		case aws.Name, azure.Name, gcp.Name:
+		default:
+			allErrs = append(allErrs, field.Invalid(field.NewPath("publish"), c.Publish, fmt.Sprintf("Internal publish strategy is not supported on %q platform", platformName)))
+		}
+	}
+
 	return allErrs
 }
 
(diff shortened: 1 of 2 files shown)


tsze-redhat:
What about Alibaba cloud?
Checking with team members who know about Alibaba cloud.

tsze-redhat:
Confirmed that Alibaba cloud probably will allow external cluster too.

pkg/types/validation/installconfig_test.go
```diff
@@ -1208,6 +1208,34 @@ func TestValidateInstallConfig(t *testing.T) {
...
+			expectedError: `publish: Invalid value: "Internal": Internal publish strategy does not work with non-cloud installations`,
+		}, {
+			name: "publish external for non-cloud platform",
+			installConfig: func() *types.InstallConfig {
+				c := validInstallConfig()
+				c.Platform = types.Platform{VSphere: validVSpherePlatform()}
+				c.Publish = types.InternalPublishingStrategy
+				return c
+			}(),
+			expectedError: `publish: Invalid value: "Internal": Internal publish strategy does not work with non-cloud installations`,
```
e-tienne, line 1228: Is the goal testing `ExternalPublishingStrategy` for vSphere?

pkg/types/validation/installconfig.go
```diff
@@ -115,6 +115,10 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
 	}
 	allErrs = append(allErrs, validateCloudCredentialsMode(c.CredentialsMode, field.NewPath("credentialsMode"), c.Platform.Name())...)
 
+	if c.Publish == types.InternalPublishingStrategy && c.Platform.AWS == nil && c.Platform.Azure == nil && c.Platform.GCP == nil {
```
e-tienne, line 118: Since the list of public clouds will grow over time, I wonder if there's a better way to group them instead have to add `&& c.Platform.Cloud == nil` for each platform.

pkg/types/validation/installconfig.go
```diff
@@ -115,6 +115,10 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
...
+		allErrs = append(allErrs, field.Invalid(field.NewPath("publish"), c.Publish, "Internal publish strategy does not work with non-cloud installations"))
```
e-tienne, line 119: Nit: I would like to align the error message with our docs if possible. I was thinking of something like "Internal publish strategy is only supported on public clouds."

staebler (approved):
/lgtm

PR #4906, opened 2021-05-06
Bug 1956776: Check if network is valid in install-config

rna-afk:
Added a validation that checks the install-config specifically
the vsphere platform network field to see if the network is
valid and present in the datacenter provided by the user and
returns error if the network is not found in the vCenter.


pkg/types/vsphere/validation/platform.go
```diff
@@ -104,3 +110,29 @@ func validateFolder(p *vsphere.Platform, fldPath *field.Path) field.ErrorList {
 
 	return allErrs
 }
+
+func validateNetwork(p *vsphere.Platform, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	vim25Client, _, err := vspheretypes.CreateVSphereClients(context.TODO(),
```
staebler, line 116: The validation in `pkg/types` should only do static checking: It should not connect to the vCenter. For that type of checking, add validation to https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/vsphere/validation.go.

pkg/types/vsphere/validation/platform.go
```diff
@@ -104,3 +110,29 @@ func validateFolder(p *vsphere.Platform, fldPath *field.Path) field.ErrorList {
...
+		p.Password)
+
+	if err != nil {
+		return append(allErrs, field.InternalError(fldPath, errors.Wrapf(err, "unable to connect to vCenter %s.", p.VCenter)))
+	}
+
+	finder := find.NewFinder(vim25Client)
+	ctx := context.TODO()
+
+	dataCenter, err := finder.Datacenter(ctx, fmt.Sprintf("./%s", p.Datacenter))
```
staebler, line 128: Check out the logic being add in https://github.com/openshift/installer/pull/4894/files#diff-cdefcb6fa878811aa273941a4ac54baf6ce971ab287ef81f2736a9d9904cbd61R55-R61 for name of the datacenter.

pkg/types/vsphere/validation/platform.go
```diff
@@ -104,3 +110,29 @@ func validateFolder(p *vsphere.Platform, fldPath *field.Path) field.ErrorList {
 
 	return allErrs
 }
+
+func validateNetwork(p *vsphere.Platform, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
```
staebler, line 115: The only places that use this variable add a single error and return immediately. We may as well do the following instead, for example.
```
return field.ErrorList{field.InternalError(...)}
```

patrickdillon (commented):
This looks good. I left a small suggestion and a nit comment.

pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -50,19 +83,31 @@ func folderExists(ic *types.InstallConfig, fldPath *field.Path) field.ErrorList
...
+	dcName := p.Datacenter
+	if !strings.HasPrefix(dcName, "/") && !strings.HasPrefix(dcName, "./") {
+		dcName = "./" + dcName
+	}
 
-	if _, err = finder.Folder(ctx, cfg.Folder); err != nil {
-		return append(allErrs, field.Invalid(fldPath, cfg.Folder, err.Error()))
+	dataCenter, err := finder.Datacenter(ctx, dcName)
+	if err != nil {
+		return field.ErrorList{field.Invalid(fldPath, p.Datacenter, "unable to find datacenter provided")}
```
staebler, line 105: Should we include `err` in the field error?
```suggestion
		return field.ErrorList{field.Invalid(fldPath, p.Datacenter, fmt.Sprintf("unable to find datacenter provided: %v", err))}
```
Or, should we keep consistent with the field error when we cannot find the folder?
```suggestion
		return field.ErrorList{field.Invalid(fldPath, p.Datacenter, err.Error())}
```

pkg/asset/installconfig/vsphere/client.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"context"
+
+	"github.com/vmware/govmomi/find"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25"
+)
+
+type VSphereClient interface {
```
staebler, line 11: We are using the term "finder" for this everywhere else. Let's stay consistent.
```suggestion
type Finder interface {
```

pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -2,46 +2,79 @@ package vsphere
...
+
+	if err != nil {
+		return errors.New(field.InternalError(field.NewPath("platform", "vsphere"), errors.Wrapf(err, "unable to connect to vCenter %s.", p.VCenter)).Error())
+	}
+
+	finder := NewFinder(vim25Client)
+	return validateVsphere(finder, ic)
+}
 
+func validateVsphere(finder VSphereClient, ic *types.InstallConfig) error {
```
staebler, line 38: The "Vsphere" part is redundant with the package name.
```suggestion
func validate(finder VSphereClient, ic *types.InstallConfig) error {
```

pkg/asset/installconfig/vsphere/validation_test.go
```diff
@@ -88,13 +91,18 @@ func TestValidate(t *testing.T) {
...
-		validationMethod: ValidateForProvisioning,
+		validationMethod: validateProvisioning,
 		expectErr:        `^platform\.vsphere\.cluster: Required value: must specify the cluster$`,
 	}}
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
 
+	vsphereClient := mock.NewMockVSphereClient(mockCtrl)
+	vsphereClient.EXPECT().Datacenter(gomock.Any(), gomock.Any()).Return(&object.Datacenter{Common: object.Common{InventoryPath: "valid_dc"}}, nil).AnyTimes()
+	vsphereClient.EXPECT().Network(gomock.Any(), gomock.Any()).Return(nil, nil).AnyTimes()
```
staebler, line 102: Let's also create tests that cover failures to find the datacenter and network.

staebler (commented):
This looks promising with the new mocks. Can you separate the new interface and mocks into its own commit separate from the commit that adds the new validation?

pkg/asset/installconfig/vsphere/validation.go
rna-afk: I'll change it to the latter for consistency. Let me know if the former is better.

pkg/asset/installconfig/vsphere/validation.go
rna-afk: just validate is not allowed.
```
validate already declared through import of package validate ("github.com/openshift/installer/pkg/validate")
```

jstuever (changes requested):
I think this is good overall. One minor change in timing of a particular validation. We may need to work on how we validate the individual parameters after we sort out the pathing issue(s) being addressed by other bugs. However, that's not a task that has to be addressed in this PR.

jstuever:
/hold
This is more an RFE than a bug, and one that only validates user input. Probably best to wait until master opens for 4.10 work before we merge this.

PR #4908, opened 2021-05-06
Bug 1957869: vendor: update baremetal-operator to align irmc PowerInterface

hardys:
This is to include https://github.com/metal3-io/baremetal-operator/pull/848

This is needed because we reference PowerInterface() in
pkg/tfvars/baremetal/baremetal.go, so Ironic power_interface
will be wrong for master deployment via terraform when irmc is
used.

See https://bugzilla.redhat.com/show_bug.cgi?id=1957869


hardys:
/retitle Bug 1957869: vendor: update baremetal-operator to align irmc PowerInterface

PR #4914, opened 2021-05-07
Bug 1958428: aws: support more auth options in manual mode

gregsheremeta:
aws: support more auth options in manual mode

This is a manual cherry-pick of #4596 and its followup fix #4655

/assign @staebler


staebler (approved):
/lgtm

wking:
@gregsheremeta confirmed this worked in 4.8, so twiddling GitHub labels directly to avoid blocking on QE verification (which we will still want for both 4.8 and 4.7 before we release this code).

Issue #4523, opened 2021-01-06
Secret mask confuses subsequent output


kgibm:
When the ~2KB secret is pasted in the macOS Terminal, the `*` characters from [AlecAivazis/survey](https://github.com/AlecAivazis/survey) that the openshift installer uses confuse subsequent output (see below). [This issue](https://github.com/AlecAivazis/survey/issues/306) has now been resolved upstream so please ensure you bump your version of `survey` at the next opportunity.

```
? Pull Secret [? for help] ***************************************************************************************************************************************************************************************************************************************************INFO Creating
infrastructure resources...         ****************************************************************************************************************************************************************************************************************************INFO Waiting 
up to 20m0s for the Kubernetes API at https://api.cluster-f1bd.sandbox1365.opentlc.com:6443... ***************************************************************************************************************************************************INFO API 
v1.17.1 up                               ***********************************************************************************************************************************************************************************************************************INFO Waiting 
up to 40m0s for bootstrapping to complete... ***************************************************************************************************************************************************************************************************************INFO Destroying 
the bootstrap resources...        ***********************************************************************************************************************************************************************************************************************INFO Waiting 
up to 30m0s for the cluster at https://api.cluster-f1bd.sandbox1365.opentlc.com:6443 to initialize... ******************************************************************************************************************************************************INFO Waiting up 
to 10m0s for the openshift-console route to be created... ***********************************************************************************************************************************************************************************************INFO Install 
complete!                            ***********************************************************************************************************************************************************************************************************************INFO To 
access the cluster as the system:admin user when using 'oc', run 'export KUBECONFIG=/home/[...]/cluster-f1bd/auth/kubeconfig' *************************************************************************************************************INFO Access 
the OpenShift web-console here: https://console-openshift-console.apps.cluster-f1bd.sandbox1365.opentlc.com ***************
```

tsze-redhat:
Created Bugzilla
https://bugzilla.redhat.com/show_bug.cgi?id=1947067

pruan-rht:
@kgibm I've tried it on iTerm2 and the output is much cleaner.  
### iTerm2
```
$ openshift-install create install-config  --dir=pruan0407b
? SSH Public Key /Users/pruan/.ssh/openshift-qe.pub
? Platform azure
? azure subscription id xxx
? azure tenant id xxx
? azure service principal client id xxx
? azure service principal client secret [? for help] **********************************
[...]

PR #4915, opened 2021-05-08
Bump survey to v2

ecordell:
I bumped survey to v2 in anticipation of pulling in a release that includes https://github.com/AlecAivazis/survey/pull/353

If/when that merges, I can bump survey again to fix the input overwrite issue. This PR takes care of all of the other v1-> v2 changes ahead of time.

Example output with the patch:
Screen Shot 2021-05-07 at 9 24 52 PM

Related to #4523


staebler:
There is already a PR for this. https://github.com/openshift/installer/pull/4895

PR #4916, opened 2021-05-08
Bug 1958420: aws: restore setting aws platform spec in infra resource

staebler:
The code that set an empty `AWSPlatformSpec` in the infrastructure resource was accidentally removed when adding support for [propagating user tags](https://github.com/openshift/installer/commit/2033421c8dcba057a643cdd35e040d7ddcd4c355). This causes the installer to panic when the install config included service endpoints.


wking (approved):
/bugzilla refresh
/lgtm

PR #4917, opened 2021-05-08
Bug 1957809: Validation of platform.openstack.machineSubnet

adduarte:
If the user provides an UUID under platform.openstack.machinesSubnet
of a subnet which does not exist, the installer should not fail with a
backtrace.
Instead, the installer should exit gracefully and provide the correct error
to the user.

### DIFF
--- a/pkg/asset/installconfig/openstack/validation/platform.go
+++ b/pkg/asset/installconfig/openstack/validation/platform.go
@@ -38,7 +38,7 @@ func validateMachinesSubnet(p *openstack.Platform, n *types.Networking, ci *Clou
 		if len(p.ExternalDNS) > 0 {
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("externalDNS"), p.ExternalDNS, "externalDNS is set, externalDNS is not supported when machinesSubnet is set"))
 		}
-		if !validUUIDv4(p.MachinesSubnet) {
+		if !validUUIDv4(p.MachinesSubnet) || ci.MachinesSubnet == nil {
 			allErrs = append(allErrs, field.InternalError(fldPath.Child("machinesSubnet"), errors.New("invalid subnet ID")))
 		} else {
 			if n.MachineNetwork[0].CIDR.String() != ci.MachinesSubnet.CIDR {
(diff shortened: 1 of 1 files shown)


pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -38,7 +38,7 @@ func validateMachinesSubnet(p *openstack.Platform, n *types.Networking, ci *Clou
 		if len(p.ExternalDNS) > 0 {
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("externalDNS"), p.ExternalDNS, "externalDNS is set, externalDNS is not supported when machinesSubnet is set"))
 		}
-		if !validUUIDv4(p.MachinesSubnet) {
+		if !validUUIDv4(p.MachinesSubnet) || ci.MachinesSubnet == nil {
```
Fedosin, line 41: I think we need to differentiate this errors. If there is no machine subnet with provided id, then we should generate NotFound error.
```go
if !validUUIDv4(p.MachinesSubnet) {
	allErrs = append(allErrs, field.InternalError(fldPath.Child("machinesSubnet"), errors.New("invalid subnet ID")))
} else if ci.MachinesSubnet == nil {
	allErrs = append(allErrs, field.NotFound(fldPath.Child("machinesSubnet"), p.MachinesSubnet))
} else if n.MachineNetwork[0].CIDR.String() != ci.MachinesSubnet.CIDR {
	allErrs = append(allErrs, field.InternalError(fldPath.Child("machinesSubnet"), fmt.Errorf("the first CIDR in machineNetwork, %s, doesn't match the CIDR of the machineSubnet, %s", n.MachineNetwork[0].CIDR.String(), ci.MachinesSubnet.CIDR)))
}
```

Fedosin (changes requested):
Just one suggestion. And also it would be nice to have a unit test for this case.

pkg/asset/installconfig/openstack/validation/platform.go
mandre, line 41: We should rather use the `field.NotFound` error for consistency:
```suggestion
		if ci.MachinesSubnet == nil {
			allErrs = append(allErrs, field.NotFound(fldPath.Child("machinesSubnet"), p.MachinesSubnet))
		}
		if !validUUIDv4(p.MachinesSubnet) {
```

mandre (commented):
/lgtm
/approve

PR #4924, opened 2021-05-11
Bug 1940337: openstack: Report missing flavor

pierreprinetti:
Before this patch, when the user provided a non-existant flavor name in
install-config, the error message was about missing GB of RAM or disk.

This happened because when a flavor was not found in the OpenStack
cloud, the "cloud info collector" code created an empty Flavor object
that responded to the name of the missing flavor.

In this patch, the absence of a flavor is signaled in the pre-flight
checks by avoiding to populate the "cloudinfo" object, so that an
explicit error message can correctly be output to the user.


EmilienM:
/approve
/lgtm

Thanks :-)

PR #4925, opened 2021-05-11
Bug 1955336: Check cluster name for similarities with word 'google'

rna-afk:
GCP does not allow certain resources to have names with words
similar to 'google'. Adding a regex check for known invalid
cluster names in GCP to inform the user before trying to create
the cluster.

Tested values:
Passed: test-googl, testgoogl-test, testgoogl, goo-test
Failed: goog-test, test-google, testgoogletest

### DIFF
--- a/pkg/asset/installconfig/clustername.go
+++ b/pkg/asset/installconfig/clustername.go
@@ -35,7 +35,13 @@ func (a *clusterName) Generate(parents asset.Parents) error {
 		validator = survey.ComposeValidators(validator, func(ans interface{}) error {
 			return validate.ClusterName1035(ans.(string))
 		})
+		if platform.GCP != nil {
+			validator = survey.ComposeValidators(validator, func(ans interface{}) error {
+				return validate.GCPClusterName(ans.(string))
+			})
+		}
 	}
+
 	if platform.Ovirt != nil {
 		// FIX-ME: As soon bz#1915122 get resolved remove the limitation of 14 chars for the clustername
 		validator = survey.ComposeValidators(validator, func(ans interface{}) error {
--- a/pkg/asset/installconfig/gcp/validation.go
+++ b/pkg/asset/installconfig/gcp/validation.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"github.com/openshift/installer/pkg/types"
+	"github.com/openshift/installer/pkg/validate"
 )
 
 type resourceRequirements struct {
@@ -35,6 +36,10 @@ var computeReq = resourceRequirements{
 // Validate executes platform-specific validation.
 func Validate(client API, ic *types.InstallConfig) error {
 	allErrs := field.ErrorList{}
+
+	if err := validate.GCPClusterName(ic.ObjectMeta.Name); err != nil {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("clusterName"), ic.ObjectMeta.Name, err.Error()))
+	}
 
 	allErrs = append(allErrs, validateProject(client, ic, field.NewPath("platform").Child("gcp"))...)
 	allErrs = append(allErrs, validateNetworks(client, ic, field.NewPath("platform").Child("gcp"))...)
--- a/pkg/validate/validate.go
+++ b/pkg/validate/validate.go
@@ -115,6 +115,17 @@ func ClusterName1035(v string) error {
 	return ClusterName(v)
 }
 
+// GCPClusterName checks if the provided cluster name has words similar to the word 'google'
+// since resources with that name are not allowed in GCP.
+func GCPClusterName(v string) error {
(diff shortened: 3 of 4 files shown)


jstuever (changes requested):
We probably also want to validate this if it was manually input into the install-config.yaml. (pkg/asset/installconfig/gcp/validation.go)

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -36,6 +37,7 @@ var computeReq = resourceRequirements{
 func Validate(client API, ic *types.InstallConfig) error {
 	allErrs := field.ErrorList{}
 
+	allErrs = append(allErrs, validateGCPClusterName(ic, field.NewPath("clusterName"))...)
```
staebler, line 40: Static validation should be in package under `pkg/types`.

pkg/validate/validate.go
```diff
@@ -115,6 +115,17 @@ func ClusterName1035(v string) error {
 	return ClusterName(v)
 }
 
+// GCPClusterName checks if the provided cluster name has words similar to the word 'google'
+// since resources with that name are not allowed in GCP.
+func GCPClusterName(v string) error {
+	reStartsWith := regexp.MustCompile("^goog")
+	reContains := regexp.MustCompile(".*google.*")
```
staebler, line 122: You don't need the `.*` parts. Also, we should check for "0" in place of "o", too.
```suggestion
	reContains := regexp.MustCompile("g[o0]{2}gle")
```

pkg/validate/validate.go
```diff
@@ -115,6 +115,17 @@ func ClusterName1035(v string) error {
 	return ClusterName(v)
 }
 
+// GCPClusterName checks if the provided cluster name has words similar to the word 'google'
+// since resources with that name are not allowed in GCP.
+func GCPClusterName(v string) error {
+	reStartsWith := regexp.MustCompile("^goog")
+	reContains := regexp.MustCompile(".*g[o0]{2}gle.*")
+	if reStartsWith.MatchString(v) || reContains.MatchString(v) {
+		return errors.New("names that start with 'goog' or contain google are reserved in GCP and are not allowed")
```
staebler, line 124: There is inconsistent quoting. The `goog` is quoted with single quotes and the `google` has no quoting. I recommend quoting both with double quotes. Let's also remove the text around why we are rejecting the name.
```suggestion
		return errors.New("cluster name must not start with \"goog\" or contain variations of \"google\"")
```

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -36,6 +37,10 @@ var computeReq = resourceRequirements{
 func Validate(client API, ic *types.InstallConfig) error {
 	allErrs := field.ErrorList{}
 
+	if err := validate.GCPClusterName(ic.ObjectMeta.Name); err != nil {
```
staebler, line 40: If a cluster name violates the RFC-1035 requirements, the installer will report that twice.
https://github.com/openshift/installer/blob/296b0b10da2a5a9a5afca6c10b03ee73ce98578c/pkg/types/validation/installconfig.go#L73

rna-afk: Removed the RFC check in the GCPClusterName function.

pkg/asset/installconfig/clustername.go
```diff
@@ -35,7 +35,17 @@ func (a *clusterName) Generate(parents asset.Parents) error {
 		validator = survey.ComposeValidators(validator, func(ans interface{}) error {
 			return validate.ClusterName1035(ans.(string))
 		})
+		if platform.GCP != nil {
+			validator = survey.ComposeValidators(validator, func(ans interface{}) error {
```
staebler, line 39: This is going to replace the RFC-1035 validator above rather than adding to it.

pkg/asset/installconfig/clustername.go
```diff
@@ -35,7 +35,17 @@ func (a *clusterName) Generate(parents asset.Parents) error {
...
+				return validate.GCPClusterName(ans.(string))
+			})
+		}
 	}
+
+	if platform.GCP != nil {
+
+	}
+
```
staebler, line 48: Remove this.

pkg/asset/installconfig/clustername.go
rna-afk: I changed the logic for the GCPClusterName to just validate against the goog and google parts and return nil if success. Do I set it to the old logic?

staebler: We need the new check for goog and google as well as the old RFC-1035 check.

staebler: Oh, sorry. I misread that the code was doing composition. This looks good as is.

rna-afk: Yes but doesn't this chain the two validators together into one?

staebler (approved):
/lgtm

PR #4929, opened 2021-05-12
libvirt: README updated for Fedora 33+ systemd-resolved

timhansendev:
Fedora 34 uses systemd-resolved, which adds an additional step for properly resolving to the dnsmasq process for the openshift installer.

Signed-off-by: Tim Hansen <tihansen@redhat.com>

### DIFF
--- a/docs/dev/libvirt/README.md
+++ b/docs/dev/libvirt/README.md
@@ -247,6 +247,34 @@ This step allows installer and users to resolve cluster-internal hostnames from 
 
 3. Reload NetworkManager to pick up the `dns` configuration change: `sudo systemctl reload NetworkManager`
 
+### Setting the resolver on Fedora 33+ with systemd-resolved
+
+Fedora 33 switched to utilizing systemd-resolved and changes how DNS resolution works on a per-interface basis. 
+
+If you are running Fedora 33, you must indicate that you want your resolver to point to the local dnsmasq resolver on the `tt0` interface.
+
+The below example uses the default `tt0` interface, change the interface name if you are utilizing a custom interface/bridge instead.
+
+```sh
+resolvectl dns tt0 127.0.0.1
+```
+
+You may see which interfaces point to which DNS resolvers by typing `resolvectl`. `tt0` should be now pointing to 127.0.0.1, where the dnsmasq process is listening.
+
+```sh
+resolvectl
+...
+Link 100 (tt0)
+    Current Scopes: none
+         Protocols: -DefaultRoute +LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
+Current DNS Server: 127.0.0.1
+       DNS Servers: 127.0.0.1
+...
+```
+
+Please note that this is not persistant across libvirt network creation/deletion of the tt0 interface. Once the tt0 interface is deleted and recreated, it will need to updated again with the proper resolver.
+
+
 ## Build the installer
 
 Set `TAGS=libvirt` to add support for libvirt; this is not enabled by default because libvirt is [development only](../../../README.md#supported-platforms).


docs/dev/libvirt/README.md
```diff
@@ -247,6 +247,31 @@ This step allows installer and users to resolve cluster-internal hostnames from
...
+
+Fedora 34 switched to utilizing systemd-resolved and changes how DNS resolution works on a per-interface basis. 
+
+If you are running Fedora 34, you must indicate that you want your resolver to point to the local dnsmasq resolver on the `tt0` interface.
+
+The below example uses the default `tt0` interface, change the interface name if you are utilizing a custom interface/bridge instead.
+
+```sh
+resolvectl dns tt0 127.0.0.1
+```
```
cfergeau, line 260: I wonder if on f33+, it would be workable to not setup dnsmasq at all on the host, and make use of https://github.com/openshift/installer/commit/a6ef6c7d0418ba81ecf6e88d9bcc6e65e5ac6ca7 instead (+ maybe some `resolvectl domain tt0 ~tt.testing` configuration?)

docs/dev/libvirt/README.md
```diff
@@ -247,6 +247,31 @@ This step allows installer and users to resolve cluster-internal hostnames from
 
 3. Reload NetworkManager to pick up the `dns` configuration change: `sudo systemctl reload NetworkManager`
 
+### Setting the resolver on Fedora 34+ with systemd-resolved
+
+Fedora 34 switched to utilizing systemd-resolved and changes how DNS resolution works on a per-interface basis. 
+
+If you are running Fedora 34, you must indicate that you want your resolver to point to the local dnsmasq resolver on the `tt0` interface.
```
cfergeau, line 254: I find this sentence a bit confusing, we want to configure systemd-resolved to use the dnsmasq resolver we configured before for the `tt0` interface rather than the one libvirt set up automatically for this interface.

docs/dev/libvirt/README.md
```diff
@@ -247,6 +247,31 @@ This step allows installer and users to resolve cluster-internal hostnames from
...
+
+The below example uses the default `tt0` interface, change the interface name if you are utilizing a custom interface/bridge instead.
+
+```sh
+resolvectl dns tt0 127.0.0.1
+```
+
+You may see which interfaces point to which DNS resolvers by typing `resolvectl`. `tt0` should be now pointing to 127.0.0.1, where the dnsmasq process is listening.
+
```
cfergeau, line 263: Does this configuration survives a destroy/create sequence (when the libvirt network is destroyed/recreated)?

docs/dev/libvirt/README.md
```diff
@@ -247,6 +247,31 @@ This step allows installer and users to resolve cluster-internal hostnames from
 
 3. Reload NetworkManager to pick up the `dns` configuration change: `sudo systemctl reload NetworkManager`
 
+### Setting the resolver on Fedora 34+ with systemd-resolved
+
```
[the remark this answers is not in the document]
timhansendev: Noted, will change.

docs/dev/libvirt/README.md
timhansendev: I did not notice this section before. Thanks for pointing it out- I will test with this solution as this is cleaner than setting it up the way I've proposed.

docs/dev/libvirt/README.md
timhansendev: It does not, this needs to be re-created every time.  I will most likely change this section favoring the method outlined in your comment above https://github.com/openshift/installer/pull/4929#discussion_r633562305

docs/dev/libvirt/README.md
timhansendev: Noted, I will rephrase this to make it clearer as to what is actually happening.

docs/dev/libvirt/README.md
```diff
@@ -247,11 +247,11 @@ This step allows installer and users to resolve cluster-internal hostnames from
 
 3. Reload NetworkManager to pick up the `dns` configuration change: `sudo systemctl reload NetworkManager`
 
-### Setting the resolver on Fedora 34+ with systemd-resolved
+### Setting the resolver on Fedora 33+ with systemd-resolved
```
cfergeau, line 250: :+1:

docs/dev/libvirt/README.md
cfergeau: Did you have a chance to try https://github.com/openshift/installer/pull/4929#discussion_r633562305 ?
If not/if it did not work, we can go with this PR, but mentioning that this change is not persistent would be useful.

cfergeau (commented):
The f33 commit could be squashed into the main one.

docs/dev/libvirt/README.md
timhansendev: I would like to address this in a separate PR to simplify how networking is done for libvirt development setups.  I was not able to get a working setup without dnsmasq running on the host. I think there could be some discussion on what the best architecture would be to make sure it is more automated in terms of deployment.

timhansendev:
> The f33 commit could be squashed into the main one.

Done. Please let me know if you want to see any other fixes. Thank you

docs/dev/libvirt/README.md
```diff
@@ -247,6 +247,34 @@ This step allows installer and users to resolve cluster-internal hostnames from
...
+...
+Link 100 (tt0)
+    Current Scopes: none
+         Protocols: -DefaultRoute +LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
+Current DNS Server: 127.0.0.1
+       DNS Servers: 127.0.0.1
+...
+```
+
+Please note that this is not persistant across libvirt network creation/deletion of the tt0 interface. Once the tt0 interface is deleted and recreated, it will need to updated again with the proper resolver.
```
timhansendev, line 275: Added the note here that this does not persist across network creation/deletion.

cfergeau: I've attempted to consolidate this PR and https://github.com/openshift/installer/pull/4564 in https://github.com/openshift/installer/pull/5132 so that we can finally merge this.

timhansendev: Thanks, do you want me to close this PR in favor of #5132 ?

timhansendev:
Work has moved towards #4564

PR #4931, opened 2021-05-12
Add Makefile

yselkowitz:
Separated from https://github.com/openshift/installer/pull/4891

/cc @staebler


yselkowitz:
This is less necessary than it was when first proposed.

PR #4932, opened 2021-05-13
Bug 1960323: Address issues raised by coverity security scan

rna-afk:
This fixes a few small defects in the installer code as
mentioned here: http://download.eng.bos.redhat.com/rcm-guest/puddles/RHAOS/coverity/results/ose-installer/fe9e87fb04ae777d913e5787792756208392cf3a/stage_1/all_results.html

### DIFF
--- a/pkg/terraform/exec/plugins/azureprivatedns/resource_private_dns_srv_record.go
+++ b/pkg/terraform/exec/plugins/azureprivatedns/resource_private_dns_srv_record.go
@@ -172,17 +172,18 @@ func resourceArmPrivateDNSArvRecordDelete(d *schema.ResourceData, meta interface
 }
 
 func flattenAzureRmPrivateDNSSrvRecords(records *[]privatedns.SrvRecord) []map[string]interface{} {
+	if records == nil {
+		return nil
+	}
 	results := make([]map[string]interface{}, 0, len(*records))
 
-	if records != nil {
-		for _, record := range *records {
-			results = append(results, map[string]interface{}{
-				"priority": *record.Priority,
-				"weight":   *record.Weight,
-				"port":     *record.Port,
-				"target":   *record.Target,
-			})
-		}
+	for _, record := range *records {
+		results = append(results, map[string]interface{}{
+			"priority": *record.Priority,
+			"weight":   *record.Weight,
+			"port":     *record.Port,
+			"target":   *record.Target,
+		})
 	}
 
 	return results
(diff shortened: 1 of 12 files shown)


pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go
```diff
@@ -190,9 +190,6 @@ func findImportOvaParams(client *vim25.Client, datacenter, cluster, datastore, n
 		foundDatastore := false
 		foundNetwork := false
 		hostObj.Properties(ctx, hostObj.Reference(), []string{"network", "datastore", "runtime"}, &hostSystemManagedObject)
-		if err != nil {
```
staebler, line 193: Rather than removing the check for an error, we should be looking at the error returned from the call to `hostObj.Properties` instead of ignoring it.

pkg/terraform/exec/plugins/azureprivatedns/resource_private_dns_srv_record.go
```diff
@@ -174,15 +174,13 @@ func resourceArmPrivateDNSArvRecordDelete(d *schema.ResourceData, meta interface
 func flattenAzureRmPrivateDNSSrvRecords(records *[]privatedns.SrvRecord) []map[string]interface{} {
 	results := make([]map[string]interface{}, 0, len(*records))
 
-	if records != nil {
```
staebler, line 177: Rather than getting rid of the check for nil, the check for nil should be done earlier in the function. The scan is telling you that, if `records` were nil, then the earlier dereference in the function would have thrown a panic before getting to this nil check.

Add the following at the top of the function.
```
if records == nil {
  return nil
}
```

pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go
```diff
@@ -189,11 +189,10 @@ func findImportOvaParams(client *vim25.Client, datacenter, cluster, datastore, n
 	for _, hostObj := range hosts {
 		foundDatastore := false
 		foundNetwork := false
-		hostObj.Properties(ctx, hostObj.Reference(), []string{"network", "datastore", "runtime"}, &hostSystemManagedObject)
+		err = hostObj.Properties(ctx, hostObj.Reference(), []string{"network", "datastore", "runtime"}, &hostSystemManagedObject)
```
staebler, line 192:
```suggestion
		err := hostObj.Properties(ctx, hostObj.Reference(), []string{"network", "datastore", "runtime"}, &hostSystemManagedObject)
```

pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go
```diff
@@ -189,11 +189,10 @@ func findImportOvaParams(client *vim25.Client, datacenter, cluster, datastore, n
...
 		if err != nil {
 			return nil, err
 		}
-
```
staebler, line 196: Leave in this empty line.

staebler (commented):
Looks good. Just two small nits.

staebler (approved):
/lgtm

PR #4933, opened 2021-05-14
Bug 1957597: aws: using dotted domain when looking for public hosted zone

staebler:
When destroying a cluster that has a bring-your-own private hosted zone, the search for the public hosted zone should be using the dotted domain (i.e., the domain that ends in a dot). Since the search was incorrectly using the un-dotted domain, the search was never able to find a match.


patrickdillon:
CI shows record sets being deleted.
/lgtm
/approve

staebler:
> CI shows record sets being deleted.
> /lgtm
> /approve

CI does not test this code path. We need https://github.com/openshift/release/pull/17062 for that.

PR #4935, opened 2021-05-14
Bug 1959981: oVirt: fix empty AffinityGroupsNames handling

Gal-Zaidman:
- Handle empty affinity group names field

### DIFF
--- a/pkg/types/ovirt/defaults/machinepools.go
+++ b/pkg/types/ovirt/defaults/machinepools.go
@@ -12,7 +12,7 @@ func setMachinePool(p *types.MachinePool) {
 }
 
 func setDefaultAffinityGroups(p *ovirt.Platform, mp *types.MachinePool, agName string) {
-	if len(mp.Platform.Ovirt.AffinityGroupsNames) == 0 {
+	if mp.Platform.Ovirt.AffinityGroupsNames == nil {
 		for _, ag := range p.AffinityGroups {
 			if ag.Name == agName {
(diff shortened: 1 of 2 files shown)


pkg/types/ovirt/defaults/machinepools.go
```diff
@@ -12,7 +12,7 @@ func setMachinePool(p *types.MachinePool) {
 }
 
 func setDefaultAffinityGroups(p *ovirt.Platform, mp *types.MachinePool, agName string) {
-	if len(mp.Platform.Ovirt.AffinityGroupsNames) == 0 {
+	if mp.Platform.Ovirt.AffinityGroupsNames == nil {
```
janoszen, line 15: Do we want this code to run if the AGN is an empty slice? Len seems to be the better option here.

Gal-Zaidman: And yes I want to handle an empty slice, meaning:
- affinity_groups_names: [] - disables it
- not writing anything will be the default behavior

janoszen: This should be documented since [] is usually equivalent to nil.

staebler: See my concerns [here](https://github.com/openshift/installer/pull/4674#discussion_r580295829).

Gal-Zaidman: handled by removing the "omitempty" from the field.

Regarding the docs this is the same logic as the platform fields, we will document both in the Docs of course :)

PR #4938, opened 2021-05-17
bug 1925343: upi/metal: add more configuration options for bare metal

staebler:
Allow facilities, plan and hardware_reservation_id to be configured.


staebler:
@jhixson74 This is working now. I've run 3 installs without issue. I did not change anything, so I don't have any clue why it was always failing before. I confirm that the machines are being created in ny5 using reserved c3.small.x86 instances.

mtnbikenc (approved):
/lgtm

PR #4950, opened 2021-05-20
Bug: 1947293 Baremetal: Validate provisioning network size

squidboylan:
Due to a limitation in dnsmasq an IPv6 provisioning network cannot be
larger than a /64 network.

bug: https://bugzilla.redhat.com/show_bug.cgi?id=1947293

### DIFF
--- a/pkg/types/baremetal/validation/platform.go
+++ b/pkg/types/baremetal/validation/platform.go
@@ -42,6 +42,17 @@ func validateIPinMachineCIDR(vip string, n *types.Networking) error {
 	return fmt.Errorf("IP expected to be in one of the machine networks: %s", strings.Join(networks, ","))
 }
 
+func validateCIDRSize(p *baremetal.Platform) error {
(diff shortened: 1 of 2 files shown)


hardys:
Looks good, but could you add a test like for other validations in `platform_test.go` please?

pkg/types/baremetal/validation/platform_test.go.orig
```diff
@@ -0,0 +1,721 @@
+package validation
```
hardys, line 1: I think this file was added by mistake? :)

hardys:
lgtm just a stray file that needs to be removed - thanks!

pkg/types/baremetal/validation/platform_test.go.orig
squidboylan: Yep, I had some trouble with `patch` which led to adding some extra files, fixed now, thanks!

PR #4952, opened 2021-05-21
Bug 1820238: delete manila shares and snapshots along with the cluster

Fedosin:
This commit adds a function that deletes manila shares and their snapshots when the cluster is being destroyed.

### DIFF
--- a/pkg/destroy/openstack/openstack.go
+++ b/pkg/destroy/openstack/openstack.go
@@ -27,6 +27,8 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
 	"github.com/gophercloud/gophercloud/openstack/objectstorage/v1/containers"
 	"github.com/gophercloud/gophercloud/openstack/objectstorage/v1/objects"
+	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/v2/shares"
+	sharesnapshots "github.com/gophercloud/gophercloud/openstack/sharedfilesystems/v2/snapshots"
 	"github.com/gophercloud/gophercloud/pagination"
 	"github.com/gophercloud/utils/openstack/clientconfig"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


pkg/destroy/openstack/openstack.go
```diff
@@ -1230,6 +1233,57 @@ func deleteVolumes(opts *clientconfig.ClientOpts, filter Filter, logger logrus.F
...
+
+	allShares, err := shares.ExtractShares(allPages)
+	if err != nil {
+		logger.Error(err)
+		return false, nil
+	}
+
+	for _, share := range allShares {
+		logger.Debugf("Deleting share %q", share.ID)
+		err = shares.Delete(conn, share.ID).ExtractErr()
```
gouthampacha, line 1272: If a share has one or more snapshots, this operation will fail; but since the cluster's being deleted anyway, should we delete snapshots first?

Fedosin: you are right! I updated my patch to delete snapshots as well

mandre (commented):
/approve
/lgtm
/hold for the green CI

mandre (commented):
/lgtm
/retest

mandre:
/lgtm cancel

The failure is legit:

```
ERRO[2021-06-02T05:09:47Z] level=debug msg=Deleting OpenStack shares    
ERRO[2021-06-02T05:09:47Z] level=error msg=No suitable endpoint could be found in the service catalog. 
```

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/4952/pull-ci-openshift-installer-master-e2e-openstack/1399892154360270848

Manila is not a core openstack service and we need to accommodate for when it's not available.

pkg/destroy/openstack/openstack.go
```diff
@@ -1274,6 +1278,102 @@ func deleteSnapshots(conn *gophercloud.ServiceClient, volumeID string, logger lo
...
+		if strings.ToLower(k) == "openshiftclusterid" {
+			clusterID = v
+			break
+		}
+	}
+
+	conn, err := clientconfig.NewServiceClient("sharev2", opts)
+	if err != nil {
+		logger.Error(err)
+		return false, nil
```
mandre, line 1296: This causes the destroy code to loop over `deleteShares()` when Manila is not available in the target cloud. We can't assume it's available since Manila is not a core OpenStack service.

mandre (commented):
/lgtm
Feel free to remove the hold once the tests are passing

PR #4957, opened 2021-05-25
Add SR-IOV notes specific to Kuryr

dulek:
There are some special considerations regarding SR-IOV and Kuryr in 4.8
and this commit makes sure to list them.

### DIFF
--- a/docs/user/openstack/deploy_sriov_workers.md
+++ b/docs/user/openstack/deploy_sriov_workers.md
@@ -203,6 +203,25 @@ spec:
           configDrive: true
 ```
 
+If you are using Kuryr and your SR-IOV network has `port_security_enabled=False`, then you must create the worker node, setting its primary port with `portSecurity: false` and empty `spec.template.spec.providerSpec.value.securityGroups`. E.g.:
+
+```
+          networks:
+            - subnets:
+              - uuid: <machines_subnet_uuid>
+                portSecurityEnabled: false
+              portSecurityEnabled: false
+          securityGroups: []
+```
+
+In that case you can still apply the workers SG to the primary VM interface *after* the VM is created:
+
+```
+openstack port set --enable-port-security --security-group <infrastructure_ID>-<node_role> <main_port_id>
+```
+
+Those Kuryr specifics will be fixed in the future OpenShift versions.
+
 After you finish editing your machineSet, upload it to your OpenShift cluster:
 
 ```sh


mandre:
Can we perhaps combine with https://github.com/openshift/installer/pull/4958/? Both seem to address the same issue of documenting the deployment of SR-IOV capable nodes on networks with port security disabled.

maxwelldb:
@dulek @iamemilio Would you be more comfortable with including this in the 4.8 downstream docs, or should it be kept to the known issues in the 4.8 RNs?

Impacts: https://github.com/openshift/openshift-docs/pull/32876

dulek:
@maxwelldb: IMO it can go into downstream 4.8 docs, sure.

docs/user/openstack/deploy_sriov_workers.md
```diff
@@ -203,6 +203,27 @@ spec:
           configDrive: true
 ```
 
+If you are using Kuryr and your SR-IOV network has `port_security_enabled=False`, then
+you'll be forced to create the worker node setting its primary port with `portSecurity: false`
```
maxwelldb, line 207: "You'll be forced" or "you must"?

maxwelldb: Docs typically prefers "must" in contexts like that.

dulek: Alright!

docs/user/openstack/deploy_sriov_workers.md
```diff
@@ -203,6 +203,25 @@ spec:
...
+          securityGroups: []
+```
+
+In that case you can still apply the workers SG to the primary VM interface *after* the VM is created:
+
+```
+openstack port set --enable-port-security --security-group <infrastructure_ID>-<node_role> <main_port_id>
+```
+
+Those Kuryr specifics will be fixed in the future OpenShift versions.
```
mandre, line 223: Is there a BZ to track the limitation? If not, could you please create one (and explicitly mentions reverting this PR) and reference it here?
We will also need a different BZ to include this documentation patch in the release-4.8 branch.

dulek: There's https://bugzilla.redhat.com/show_bug.cgi?id=1964540, I added a note linking this patch there.

mandre:
There is a chance [the fix](https://github.com/openshift/cluster-api-provider-openstack/pull/185) is going to make it to 4.8 before the limitation note, which may not be necessary in this case.

@dulek please check if you still want this patch in.

dulek:
True, this might not be necessary anymore.

PR #4958, opened 2021-05-25
Bug 1970415: Document SR-IOV use case: using network with port security disabled

iamemilio:
This use case requires a unique usage pattern in comparison to what the
customer may be use to, so we chose to document it in its entirety to
avoid any confusion down the line.

### DIFF
--- a/docs/user/openstack/deploy_sriov_workers.md
+++ b/docs/user/openstack/deploy_sriov_workers.md
@@ -209,6 +209,76 @@ After you finish editing your machineSet, upload it to your OpenShift cluster:
 oc create -f sriov_machineset.yaml
 ```
 
+To create SR-IOV ports on a network with the port security disabled, you need to make additional changes to your machineSet due to security groups being set on the instance by default, and allowed address pairs automatically getting added to ports created through the `networks` and `subnets` interfaces. The solution is to define all of your ports with the `ports` interface in your machineSet. Remember that the port for the machines subnet needs:
+- allowed address pairs for your API and ingress vip ports
+- the worker security group
+- to be attached to the machines network and subnet
+
+```yaml
+apiVersion: machine.openshift.io/v1beta1
+kind: MachineSet
+metadata:
+  labels:
+    machine.openshift.io/cluster-api-cluster: <infrastructure_ID>
+    machine.openshift.io/cluster-api-machine-role: <node_role>
+    machine.openshift.io/cluster-api-machine-type: <node_role>
+  name: <infrastructure_ID>-<node_role>
+  namespace: openshift-machine-api
+spec:
+  replicas: <number_of_replicas>
+  selector:
+    matchLabels:
+      machine.openshift.io/cluster-api-cluster: <infrastructure_ID>
+      machine.openshift.io/cluster-api-machineset: <infrastructure_ID>-<node_role>
+  template:
+    metadata:
+      labels:
+        machine.openshift.io/cluster-api-cluster: <infrastructure_ID>
+        machine.openshift.io/cluster-api-machine-role: <node_role>
+        machine.openshift.io/cluster-api-machine-type: <node_role>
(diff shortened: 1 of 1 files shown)


docs/user/openstack/deploy_sriov_workers.md
```diff
@@ -209,6 +209,73 @@ After you finish editing your machineSet, upload it to your OpenShift cluster:
...
...ork with the port security disabled, you will have to make additional changes to your machine definiton. OpenStack forbids setting allowed address pairs or security groups on ports that have their port security disabled. The machines generated by the installer set security groups on the instance, which then applies those security groups to all ports attached as an interface to that instance. We also automatically apply allowed address pairs for the VIP ports to all ports created with the `networks` and `subnets` interfaces in the machine api. This means that in order to use a network with the port security disabled, you will have to define all of your ports using the `ports` interface in the machine api and specify the allowed address pairs and security groups on a port by port basis. To replace the port that attaches your node to the machines subnet, and by extension the kubernetes api, you will need to create a port that has the allowed address pairs for your api and ingress ports, and the worker security group. An example of how to do this can be seen in the first port defined in the verified example below. Then, the additional ports for your SR-IOV NICs can be defined as usual.
```
maxwelldb, line 212:
```suggestion
When creating SR-IOV worker ports, we recommend disabling the port security at a port level, as shown in the example above. This method is easier for the machine API to manage, and works nicely with the machine definitions generated by the installer automatically. However, if your architecture calls for you to create SR-IOV ports on a network with the port security disabled, you will have to make additional changes to your machine definiton. OpenStack forbids setting allowed address pairs or security groups on ports that have their port security disabled. The machines generated by the installer set security groups on the instance, which then applies those security groups to all ports attached as an interface to that instance. We also automatically apply allowed address pairs for the VIP ports to all ports created with the `networks` and `subnets` interfaces in the machine API. This means that in order to use a network with the port security disabled, you will have to define all of your ports using the `ports` interface in the machine API and specify the allowed address pairs and security groups on a port by port basis. To replace the port that attaches your node to the machines subnet, and by extension the kubernetes API, you will need to create a port that has the allowed address pairs for your API and ingress ports, and the worker security group. An example of how to do this can be seen in the first port defined in the verified example below. Then, the additional ports for your SR-IOV NICs can be defined as usual.
```
Noticed this. :grin:

docs/user/openstack/deploy_sriov_workers.md
```diff
@@ -209,6 +209,73 @@ After you finish editing your machineSet, upload it to your OpenShift cluster:
...
...ork with the port security disabled, you will have to make additional changes to your machine definiton. OpenStack forbids setting allowed address pairs or security groups on ports that have their port security disabled. The machines generated by the installer set security groups on the instance, which then applies those security groups to all ports attached as an interface to that instance. We also automatically apply allowed address pairs for the VIP ports to all ports created with the `networks` and `subnets` interfaces in the machine API. This means that in order to use a network with the port security disabled, you will have to define all of your ports using the `ports` interface in the machine API and specify the allowed address pairs and security groups on a port by port basis. To replace the port that attaches your node to the machines subnet, and by extension the kubernetes API, you will need to create a port that has the allowed address pairs for your API and ingress ports, and the worker security group. An example of how to do this can be seen in the first port defined in the verified example below. Then, the additional ports for your SR-IOV NICs can be defined as usual.
```
Fedosin, line 212: s/definiton/definition/g

adduarte: rephrase and drop references to "we" and "you" 
for example: 
"we recommend disabling the port " => " disable port security" 
or
"However, if your architecture" => " If the architecture" 
In other words be more direct with statements, it will allow the paragraph to be smaller and clearer. 
also who is "we"? 

the `networks` and `subnets` interfaces in the machine API => the `networks` and `subnets` sections of the machineset.

mandre (commented):
/lgtm
Please reference a valid BZ if you want this to merge before the 4.8 code freeze.

PR #4962, opened 2021-05-26
Bug 1909136: OpenStack: delete volume snapshots

Fedosin:
It turned out that Cinder CSI driver doesn't attach cluster id to volume snapshot metadata. To workaround this we delete snapshots based on their volume IDs.

This is a workaround PR to support volume snapshot deletion in 4.8. It should be reverted when the upstream fix is merged: https://github.com/kubernetes/cloud-provider-openstack/pull/1544


mandre (approved):
/approve
Let's wait for a successful CI run before merging.

Just in case vexxhost fixed their nova-metadata issue:
/retest

mandre (commented):
/hold cancel
/lgtm

PR #4965, opened 2021-05-28
 Bug 1965182: update baremetal-operator to fix miss IPMI credentials

zhouhao3:
This is to include https://github.com/metal3-io/baremetal-operator/pull/880

Fix error: missing the following IPMI credentials in node's driver_info: ['ipmi_address'].

see https://bugzilla.redhat.com/show_bug.cgi?id=1965182
Signed-off-by: Zhou Hao <zhouhao@fujitsu.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -49,7 +49,7 @@ require (
 	github.com/libvirt/libvirt-go v5.10.0+incompatible
 	github.com/masterzen/simplexml v0.0.0-20190410153822-31eea3082786 // indirect
 	github.com/masterzen/winrm v0.0.0-20190308153735-1d17eaf15943 // indirect
-	github.com/metal3-io/baremetal-operator v0.0.0-20210422153428-d22c5f710cdc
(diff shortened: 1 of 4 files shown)


zhouhao3:
@dtantsur @staebler @mandre test passed, PTAL.

zaneb:
I don't believe this will do anything. The BMO code is only vendored in the installer to get access to the BareMetalHost API. You need to merge the relevant commits into the openshift/baremetal-operator repo.

zhouhao3:
@zaneb  It has been merged in openshift/baremetal-operator, https://github.com/openshift/baremetal-operator/pull/149, and the corresponding vendor file also needs to be updated

kirankt (commented):
/lgtm
@staebler Can you PTAL at this PR. Thanks.

zaneb:
Ah, I didn't realise that the bmc module is also used by the terraform code. So this is indeed required.

PR #4971, opened 2021-06-01
bring example in line with the default value

duritong:
It is confusing that the example is different than the value,
which is the case since the default value changed: 120a9a35af715385c75ad9e246a4df9b103e12e6

### DIFF
--- a/data/data/azure/variables-azure.tf
+++ b/data/data/azure/variables-azure.tf
@@ -25,7 +25,7 @@ variable "azure_bootstrap_vm_type" {
 
 variable "azure_master_vm_type" {
   type = string
-  description = "Instance type for the master node(s). Example: `Standard_DS4_v3`."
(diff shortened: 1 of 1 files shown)


data/data/azure/variables-azure.tf
```diff
@@ -25,7 +25,7 @@ variable "azure_bootstrap_vm_type" {
 
 variable "azure_master_vm_type" {
   type = string
-  description = "Instance type for the master node(s). Example: `Standard_DS4_v3`."
+  description = "Instance type for the master node(s). Example: `Standard_D8s_v3`."
```
mandre, line 28: In that case, could you also update the description for the bootstrap node on [line 23](https://github.com/openshift/installer/pull/4971/files#diff-383bd473843d54f24fdea79ed74893eccb34e213fda1595142bc3ac1d206313fR23)?

mandre (commented):
/ok-to-test

PR #4972, opened 2021-06-01
Bug 1944196: vSphere - fixes bootstrap dhcp networkmanager race

jcpowermac:
Taking partial working examples from MCO common
applying to bootstrap node.

When testing in VMC we were hitting a consistent race
between adding the additional loopback entry and networkmanager
updating `/etc/resolv.conf`.

### DIFF
--- a/data/data/bootstrap/vsphere/files/etc/NetworkManager/dispatcher.d/30-local-dns-prepender.template
+++ b/data/data/bootstrap/vsphere/files/etc/NetworkManager/dispatcher.d/30-local-dns-prepender.template
@@ -17,6 +17,11 @@ case "$STATUS" in
             done
         fi
 EOF
+    
+    if [[ ! -e /etc/resolv.conf ]] || ! grep -q nameserver /etc/resolv.conf; then
+        cp /var/run/NetworkManager/resolv.conf /etc/resolv.conf
+    fi
+
     set +e
     logger -s "NM local-dns-prepender: Checking if local DNS IP is the first entry in resolv.conf"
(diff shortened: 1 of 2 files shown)


patrickdillon:
/lgtm
cc @jstuever

patrickdillon:
Test failures look related to this: https://bugzilla.redhat.com/show_bug.cgi?id=1963912

jcpowermac:
> Test failures look related to this: https://bugzilla.redhat.com/show_bug.cgi?id=1963912

I watched a CI job and it looks like the changes we implemented seems to have fixed it.

data/data/bootstrap/vsphere/files/etc/NetworkManager/dispatcher.d/30-local-dns-prepender.template
```diff
@@ -17,6 +17,11 @@ case "$STATUS" in
             done
         fi
 EOF
+    
+    if [[ ! -e /etc/resolv.conf ]] || ! grep -q nameserver /etc/resolv.conf; then
+        cp /var/run/NetworkManager/resolv.conf /etc/resolv.conf
```
e-tienne, line 22: I reviewed that the default `cp` command would create or overwrite `/etc/resolv.conf` if it does not exist or does not contain "nameserver".

Nit: can we have this template pass shellcheck?

jcpowermac:
/close

This change is not needed think #4973 and openshift/machine-config-operator#2586 resolved the issue.

PR #4973, opened 2021-06-02
Bug 1966862: pattern removed from sed to prevent expansion

aliasbo:
The pattern used in sed within the script /etc/NetworkManager/dispatcher.d/30-local-dns-prepender is causing the bootkube.sh not to complete sucessfully as it is not adding the 'nameserver 127.0.0.1' to /etc/resolv.conf. This is affecting vSphere IPI.

I have removed the problematic pattern

from:
  sed -i "/^# Generated by.*$/a nameserver $DNS_IP" /etc/resolv.conf

to:
  sed -i "/^# Generated by/a nameserver $DNS_IP" /etc/resolv.conf


aliasbo:
The Bug 1966862 was created to support this PR.

staebler (approved):
/lgtm

PR #4974, opened 2021-06-02
WIP: bug 1955697: vsphere: use fully qualified path for datacenter

jstuever:
This change ensures we are using the fully qualified path of the
vsphere datacenter everywhere where we interact with the vsphere API.
By doing so, we remove the errors when duplicate datacenters with the
same name exist in the environment. The change also modifies the
questionaire to set the fully qualified path for the datacenter in the
install-config. It is backwards compatible with prior install-configs in
that it will prepend a leading slash on datacenter names that do not
already have it.

### DIFF
--- a/pkg/asset/manifests/cloudproviderconfig.go
+++ b/pkg/asset/manifests/cloudproviderconfig.go
@@ -152,8 +152,7 @@ func (cpc *CloudProviderConfig) Generate(dependencies asset.Parents) error {
 	case vspheretypes.Name:
 		folderPath := installConfig.Config.Platform.VSphere.Folder
 		if len(folderPath) == 0 {
-			dataCenter := installConfig.Config.Platform.VSphere.Datacenter
-			folderPath = fmt.Sprintf("/%s/vm/%s", dataCenter, clusterID.InfraID)
+			folderPath = fmt.Sprintf("%s/vm/%s", vspheretypes.FullyQualifiedPath(installConfig.Config.Platform.VSphere.Datacenter), clusterID.InfraID)
 		}
 		vsphereConfig, err := vspheremanifests.CloudProviderConfig(
 			folderPath,
--- a/pkg/tfvars/vsphere/vsphere.go
+++ b/pkg/tfvars/vsphere/vsphere.go
@@ -8,6 +8,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/openshift/installer/pkg/tfvars/internal/cache"
+	"github.com/openshift/installer/pkg/types/vsphere"
 )
 
 type config struct {
@@ -52,14 +53,6 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
 	// /<datacenter>/vm/<folder_path> so we can split on "vm/".
 	folderRelPath := strings.SplitAfterN(controlPlaneConfig.Workspace.Folder, "vm/", 2)[1]
 
-	// The vSphere provider needs explicit path of the datacenter to avoid
-	// conflicts when multiple datacenters have the same name even though they
-	// are in different folders.
-	datacenter := controlPlaneConfig.Workspace.Datacenter
-	if !strings.HasPrefix(datacenter, "/") && !strings.HasPrefix(datacenter, "./") {
-		datacenter = "./" + datacenter
-	}
-
 	cfg := &config{
 		VSphereURL:        controlPlaneConfig.Workspace.Server,
 		VSphereUsername:   sources.Username,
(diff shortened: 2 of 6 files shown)


pkg/tfvars/vsphere/vsphere.go
```diff
@@ -52,14 +53,6 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
 	// /<datacenter>/vm/<folder_path> so we can split on "vm/".
 	folderRelPath := strings.SplitAfterN(controlPlaneConfig.Workspace.Folder, "vm/", 2)[1]
 
-	// The vSphere provider needs explicit path of the datacenter to avoid
-	// conflicts when multiple datacenters have the same name even though they
-	// are in different folders.
-	datacenter := controlPlaneConfig.Workspace.Datacenter
-	if !strings.HasPrefix(datacenter, "/") && !strings.HasPrefix(datacenter, "./") {
-		datacenter = "./" + datacenter
```
staebler, line 60: The old code was prepending "./". The new code is prepending "/". Also, the old code was accepting "./" as a fully-qualifying prefix. Is there a difference between "/" and "./"?

staebler (commented):
Looks good. I just have a question about "./" versus "/".

pkg/tfvars/vsphere/vsphere.go
jstuever: The `./` was work that I did for the routers that came over (copy and paste) to datacenters. The datacenters should not be using `./` in their path as there is no reference for "current working directory" for this type of object. As a result, the it would have become the root path anyhow. I used `./` in the prior PR[1] to force a qualified path not realizing this. However, moving to fully qualified paths achieves the same goal with additional benefits. Keeping in mind, both of these PRs should land in the same release, this is operationally the same at this place in the code.

[1] https://github.com/openshift/installer/pull/4894

staebler: Sounds good.

For my clarity, is it a bad datacenter name if the user specifies a datacenter name that starts with "./"?

pkg/asset/manifests/cloudproviderconfig.go
```diff
@@ -152,8 +152,7 @@ func (cpc *CloudProviderConfig) Generate(dependencies asset.Parents) error {
 	case vspheretypes.Name:
 		folderPath := installConfig.Config.Platform.VSphere.Folder
 		if len(folderPath) == 0 {
-			dataCenter := installConfig.Config.Platform.VSphere.Datacenter
-			folderPath = fmt.Sprintf("/%s/vm/%s", dataCenter, clusterID.InfraID)
+			folderPath = fmt.Sprintf("%s/vm/%s", vspheretypes.FullyQualifiedPath(installConfig.Config.Platform.VSphere.Datacenter), clusterID.InfraID)
```
patrickdillon, line 155: nit: I think this line is too long. For readability I would avoid calling a function as an arg

```suggestion
                        dataCenter := vspheretypes.FullyQualifiedPath(installConfig.Config.Platform.VSphere.Datacenter)
			folderPath = fmt.Sprintf("%s/vm/%s", dataCenter, clusterID.InfraID)
```

patrickdillon (commented):
I wonder if this regresses installations into nested datacenters. Have you looked into that?

That is, do we currently support installing into a `datacenter foo/SDDC-Datacenter` by specifying `dataCenter: SDDC-Datacenter` in the install config? If so, these changes might specify the wrong path to the datacenter.

I took a quick look in the MAO and Terraform but couldn't turn up anything conclusive.

pkg/tfvars/vsphere/vsphere.go
jstuever: If they specify a name starting with "./", it will end up being "/./", which technically works and will operationally do the same thing as before this change. This is what was happening with the changes in #4894.

staebler: Thanks for the explanation.

jinyunma:
use cluster-bot to build the image with this PR, and deploy cluster on VMC, installation was still failed, get error from MAC log:

```
$ oc get machine -n openshift-machine-api
NAME                      PHASE         TYPE   REGION   ZONE   AGE
jima0603-p54w4-master-0   Provisioned                          52m
jima0603-p54w4-master-1                                        52m
jima0603-p54w4-master-2                                        52m

$ oc logs machine-api-controllers-cc9f86989-hvvjb -n openshift-machine-api -c machine-controller
...
E0607 03:46:48.918272       1 controller.go:299] jima0603-p54w4-master-1: error updating machine: admission webhook "validation.machine.machine.openshift.io" denied the request: providerSpec.workspace.folder: Invalid value: "/SDDC-Datacenter/vm/jima0603-p54w4": folder must be absolute path: expected prefix "//SDDC-Datacenter/vm/", retrying in 30s seconds
```
Error is reported from https://github.com/openshift/machine-api-operator/blob/master/pkg/apis/machine/v1beta1/machine_webhook.go#L1135-L1140,  seems this part also need to be updated if  datacenter use absolute path.

PR #4976, opened 2021-06-06
Bug 1967623: oVirt: Fix password input

Gal-Zaidman:
The installer project updated the version of the survey package which cause the hack of editing the PasswordQuestionTemplate to break due to misssing funtions such as: HelpIcon, QuestionIcon and HelpInputRune.
There is no reason to keep this hack just to edit the inital help message, instead we will move the ctrl+c note to the help message.


Gal-Zaidman:
Verified:
image

PR #4978, opened 2021-06-07
Bug 1955697: Revert "tfvars/vsphere: use explicit path for datacenter."

jstuever:
This reverts commit cbc778f7608f383174605f6299b3b0cd3060f7fa.

This change was insufficient to resolve the issue. Another solution will be needed, and this bit of code will not be used.


patrickdillon:
LGTM to me too. Created #4980 for missing vsphere approvers.

PR #4979, opened 2021-06-08
Bug 1969374: document how to update domain for image registry in versions <4.8

Fedosin:
/label platform/openstack

### DIFF
--- a/docs/user/openstack/known-issues.md
+++ b/docs/user/openstack/known-issues.md
@@ -92,6 +92,29 @@ To recover from a scenario where one or multiple master nodes are in failed stat
 Even if master node was removed in Nova, it can still appear when you run `oc get machines -n openshift-machine-api`.
 To delete it, you can run: `oc delete machine --force=true -n openshift-machine-api <$NODE-NAME>`
 
+## Changing user domain for image registry
+
+When users change their user domain name or ID globally they hit an issue with the image registry, when the operator doesn't update its configuration accordingly. It leads to the fact that the operator can't authenticate in OpenStack Swift and becomes degraded.
+
+```sh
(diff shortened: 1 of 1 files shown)


docs/user/openstack/known-issues.md
```diff
@@ -92,6 +92,29 @@ To recover from a scenario where one or multiple master nodes are in failed stat
...
+I0303 17:32:15.603731      14 controller.go:291] object changed: *v1.Config, Name=cluster (status=true): 
+E0303 17:32:15.610609      14 controller.go:330] unable to sync: unable to sync storage configuration: Failed to authenticate provider client: Authentication failed, requeuing
+I0303 17:32:15.644501      14 controller.go:291] object changed: *v1.Config, Name=cluster (status=true): 
+E0303 17:32:15.650130      14 controller.go:330] unable to sync: unable to sync storage configuration: Failed to authenticate provider client: Authentication failed, requeuing
+```
+
+This issue was fixed in 4.8, which means that all new deployments work correctly and update image registry configuration in the case of credentials change. For older versions, and for clusters that have been upgraded to 4.8, you need to manually clean the config and delete the incorrect values. 
+
+To do so, execute `oc edit configs.imageregistry.operator.openshift.io/cluster` and manually remove `domain` or `domainID` in the `.spec.storage.swift` section. Alternatively you can use the `oc patch` command:
```
mandre, line 111: Should it be 'remove' or 'update'?

Fedosin: It should be removed, because the operator will insert the new value automatically.

mandre (commented):
/lgtm
/approve

PR #4756, opened 2021-03-16
Bug 1931974: Prefer IPv6 hostIP on bootstrap IPv6 deployments

celebdor:
Kubelet can end up choosing IPv4 addresses as its hostIP, which after:

https://github.com/openshift/cluster-kube-apiserver-operator/pull/1042/files

mean that the kubernetes API service ends up with an endpoint being an
IPv4 address.  This breaks the kubernetes API loadbalancer in the
service network and prevent the deployment from succeeding.

PR #4987, opened 2021-06-10
 Bug 1957700: None Platform - set Kubelet's node-ip param based on service net CIDR

yboaron:
This PR addressed issue similar to [1] for the None platform,
since API VIP isn't used for None platform, the UseIPv6ForNodeIP
parameter should be calculated based on the content of service CIDR.

[1] https://github.com/openshift/installer/pull/4756

### DIFF
--- a/pkg/asset/ignition/bootstrap/common.go
+++ b/pkg/asset/ignition/bootstrap/common.go
@@ -264,12 +264,24 @@ func (a *Common) getTemplateData(dependencies asset.Parents, bootstrapInPlace bo
 		platformData.VSphere = vsphere.GetTemplateData(installConfig.Config.Platform.VSphere)
 	}
 
-	var APIIntVIPonIPv6 bool
+	var useIPv6ForNodeIP bool
 	platformAPIVIP := apiVIP(&installConfig.Config.Platform)
 	if platformAPIVIP == "" {
-		APIIntVIPonIPv6 = false
+		var ipv4, ipv6 bool
+
+		useIPv6ForNodeIP = false
+		for _, network := range installConfig.Config.Networking.ServiceNetwork {
+			if network.IP.To4() == nil {
+				ipv6 = true
+			} else {
+				ipv4 = true
+			}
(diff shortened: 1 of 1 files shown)


yboaron:
cc @celebdor

cybertron (changes requested):
This doesn't match the behavior of the masters in UPI: https://github.com/openshift/machine-config-operator/blob/847e2216cc324cfc17904e6afa6a18a47c565674/pkg/operator/render.go#L156 and https://github.com/openshift/machine-config-operator/blob/ecb3c0e03652049b5c37d9f5d5678e9d6dc24301/templates/common/_base/units/nodeip-configuration.service.yaml#L25

The difference is that on the nodes it only forces ipv6 for single-stack deployments. For dual-stack it will still use ipv4 (which should be fine since both are available). I'm not sure whether it matters since in dual-stack either should work, but I think it would be safer to be consistent.

Also, need to fix the fmt failure.

staebler:
> [](https://github.com/openshift/machine-config-operator/blob/ecb3c0e03652049b5c37d9f5d5678e9d6dc24301/templates/common/_base/units/nodeip-configuration.service.yaml#L25)
> 
> The difference is that on the nodes it only forces ipv6 for single-stack deployments. For dual-stack it will still use ipv4 (which should be fine since both are available). I'm not sure whether it matters since in dual-stack either should work, but I think it would be safer to be consistent.

Let's stay consistent with what is done in nodeip-configuration and only force ipv6 in single-stack.

staebler:
Can you help me understand why we care about the service network for this? The errors in the bug are with connecting to the API server. The service network is not relevant to that. Why wouldn't we be looking at the machine network instead?

cybertron:
/test e2e-vsphere-upi

I'm not entirely sure why we're using the service network for this, but it's what was used in MCO for determining whether kubelet should use ipv6.

sadasu:
/close
Feel free to re-open it when needed.

PR #4989, opened 2021-06-11
Bug 1946079: baremetal: pass IP_OPTIONS to os downloader

honza:
The downloader needs to modify some kernel arguments of the OS image based on the network stack.

cc @asalkeld @sadasu

### DIFF
--- a/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
+++ b/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
@@ -100,6 +100,11 @@ IPTABLES=ip6tables
 IPTABLES=iptables
 {{ end }}
 
+{{ if .UseIPv6ForNodeIP }}
+IP_OPTIONS="ip=dhcp6"
+{{ else }}
+IP_OPTIONS="ip=dhcp"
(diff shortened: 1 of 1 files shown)


data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -96,8 +96,10 @@ podman run -d --net host --privileged --name httpd \
 
 {{ if .PlatformData.BareMetal.ProvisioningIPv6 }}
 IPTABLES=ip6tables
+IP_OPTIONS="ip=dhcp6"
```
hardys, line 99: We need to set this if the API VIP (where the MCS download happens) is v6, not if the provisioning network is v6 - there's a template data field we can probably use:

https://github.com/openshift/installer/blob/master/pkg/asset/ignition/bootstrap/common.go#L297

https://github.com/openshift/installer/blob/master/pkg/asset/ignition/machine/node.go#L36

hardys:
This version lgtm, but we need https://github.com/openshift/ironic-rhcos-downloader/pull/43 to merge before this will have any effect

PR #4994, opened 2021-06-13
gcp: split tf files into stages

staebler:
This builds on the work in #4876.
/hold

* Split the GCP tf files into three stages: cluster, bootstrap, and post-bootstrap.
* Allow for more variability in the terraform stages by creating a `Stage` interface. This also separates the terraform stages from the terraform plugins and instead aligns them with the platforms.
* Collect the outputs from each stage and use them as input variables in subsequent stages.
* Find the bootstrap and control plane IPs via outputs files using consistent names, as opposed to extracting from the terraform state files.

https://issues.redhat.com/browse/CORS-1694


staebler:
/cc jhixson
This fixes a few of the issues that I called out in https://github.com/openshift/installer/pull/4876.

staebler:
/cc @jhixson74 
This fixes a few of the issues that I called out in https://github.com/openshift/installer/pull/4876.

staebler:
I re-worked the commits so that the splitting up of the GCP tf files into stages is entirely separate from the other work. That makes it easier if we want to take the commits into #4876.

jhixson74 (approved):
/lgtm

staebler:
I am working to combine the commits and do some more testing. I will post a new PR that combines everything together coherently when done.
/hold

staebler:
This has been combined into #5010.
/close

PR #5000, opened 2021-06-15
Bug 1971518: Try deleting associated trunk after port delete failure

mandre:
There could be cases where the trunk is not properly tagged, for
example the UPI scripts do not set tags to trunk since the openstack
client doesn't support it.

Failure to delete trunks could result in the destroy command stuck in
a loop until it hits the timeout.

In these cases, the cluster destroy should be smart enough to try
deleting trunks for which the tagged port is a parent.


mandre:
e2e-openstack should be better with https://github.com/openshift/installer/pull/4999
/retest

PR #5002, opened 2021-06-16
Bug 1972582: Installing with an oVirt network with 2 vnics on the same network causes the installer to not create tfvars and fail with terraform error

Gal-Zaidman:
We need to return a custom error.
We wrapped a nil error which causes errors.Wrapf to return nil and then the installer would move to the next phase and fail due to missing tfvars.json file


janoszen (approved):

janoszen:
@jhixson74  can we please get an approval for this change?

PR #5003, opened 2021-06-16
openstack: Expose master server group policy

pierreprinetti:
Prior to this change, the Control plane nodes are created in a server
group with the hardcoded policy "soft-anti-affinity".

This change adds a new configuration knob in the OpenStack-specific
MachinePool configuration of `install-config.yaml`.

The new property `controlPlane.platform.openstack.serverGroupPolicy`
defines the policy that will be applied to the Control Plane server
group.

Acceptable values are:

* `affinity`
* `anti-affinity`
* `soft-affinity`
* `soft-anti-affinity`

Defaults to `soft-anti-affinity`, to match the previous hardcoded
behaviour.

Validation will fail if the property is set on the Compute MachinePool.

Implements [OSASINFRA-2507](https://issues.redhat.com/browse/OSASINFRA-2507)

/hold for gathering team consensus on the API.

### DIFF
--- a/pkg/tfvars/openstack/openstack.go
+++ b/pkg/tfvars/openstack/openstack.go
@@ -23,27 +23,28 @@ import (
 )
 
 type config struct {
-	BaseImageName                    string   `json:"openstack_base_image_name,omitempty"`
-	ExternalNetwork                  string   `json:"openstack_external_network,omitempty"`
-	Cloud                            string   `json:"openstack_credentials_cloud,omitempty"`
-	FlavorName                       string   `json:"openstack_master_flavor_name,omitempty"`
-	APIFloatingIP                    string   `json:"openstack_api_floating_ip,omitempty"`
-	IngressFloatingIP                string   `json:"openstack_ingress_floating_ip,omitempty"`
-	APIVIP                           string   `json:"openstack_api_int_ip,omitempty"`
-	IngressVIP                       string   `json:"openstack_ingress_ip,omitempty"`
-	TrunkSupport                     bool     `json:"openstack_trunk_support,omitempty"`
-	OctaviaSupport                   bool     `json:"openstack_octavia_support,omitempty"`
-	RootVolumeSize                   int      `json:"openstack_master_root_volume_size,omitempty"`
-	RootVolumeType                   string   `json:"openstack_master_root_volume_type,omitempty"`
-	BootstrapShim                    string   `json:"openstack_bootstrap_shim_ignition,omitempty"`
-	ExternalDNS                      []string `json:"openstack_external_dns,omitempty"`
-	MasterServerGroupName            string   `json:"openstack_master_server_group_name,omitempty"`
-	AdditionalNetworkIDs             []string `json:"openstack_additional_network_ids,omitempty"`
-	AdditionalSecurityGroupIDs       []string `json:"openstack_master_extra_sg_ids,omitempty"`
-	MachinesSubnet                   string   `json:"openstack_machines_subnet_id,omitempty"`
-	MachinesNetwork                  string   `json:"openstack_machines_network_id,omitempty"`
-	MasterAvailabilityZones          []string `json:"openstack_master_availability_zones,omitempty"`
-	MasterRootVolumeAvalabilityZones []string `json:"openstack_master_root_volume_availability_zones,omitempty"`
+	BaseImageName                    string                            `json:"openstack_base_image_name,omitempty"`
+	ExternalNetwork                  string                            `json:"openstack_external_network,omitempty"`
+	Cloud                            string                            `json:"openstack_credentials_cloud,omitempty"`
+	FlavorName                       string                            `json:"openstack_master_flavor_name,omitempty"`
+	APIFloatingIP                    string                            `json:"openstack_api_floating_ip,omitempty"`
+	IngressFloatingIP                string                            `json:"openstack_ingress_floating_ip,omitempty"`
+	APIVIP                           string                            `json:"openstack_api_int_ip,omitempty"`
+	IngressVIP                       string                            `json:"openstack_ingress_ip,omitempty"`
+	TrunkSupport                     bool                              `json:"openstack_trunk_support,omitempty"`
+	OctaviaSupport                   bool                              `json:"openstack_octavia_support,omitempty"`
+	RootVolumeSize                   int                               `json:"openstack_master_root_volume_size,omitempty"`
+	RootVolumeType                   string                            `json:"openstack_master_root_volume_type,omitempty"`
+	BootstrapShim                    string                            `json:"openstack_bootstrap_shim_ignition,omitempty"`
+	ExternalDNS                      []string                          `json:"openstack_external_dns,omitempty"`
+	MasterServerGroupName            string                            `json:"openstack_master_server_group_name,omitempty"`
+	MasterServerGroupPolicy          types_openstack.ServerGroupPolicy `json:"openstack_master_server_group_policy"`
+	AdditionalNetworkIDs             []string                          `json:"openstack_additional_network_ids,omitempty"`
+	AdditionalSecurityGroupIDs       []string                          `json:"openstack_master_extra_sg_ids,omitempty"`
+	MachinesSubnet                   string                            `json:"openstack_machines_subnet_id,omitempty"`
+	MachinesNetwork                  string                            `json:"openstack_machines_network_id,omitempty"`
+	MasterAvailabilityZones          []string                          `json:"openstack_master_availability_zones,omitempty"`
+	MasterRootVolumeAvalabilityZones []string                          `json:"openstack_master_root_volume_availability_zones,omitempty"`
 }
 
 // TFVars generates OpenStack-specific Terraform variables.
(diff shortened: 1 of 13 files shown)


mandre (commented):
/lgtm
/assign @staebler 

@staebler, could you take a look and make sure you're OK with the additional `controlPlane.platform.openstack.serverGroupPolicy` property added to install-config? Sadly the soft-anti-affinity policy doesn't suit everyone and a few customers asked us to provide a way to change the server group policy.

pierreprinetti:
Tests OK, canceling the hold. Awaiting @staebler’s approval.

/hold cancel

pierreprinetti:
AWS tests are failing because the Azure image is failing to build :open_mouth:

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -167,3 +168,34 @@ func validateFlavor(flavorName string, ci *CloudInfo, req flavorRequirements, fl
...
+	for _, knownPolicy := range [...]string{
+		"affinity",
+		"anti-affinity",
+		"soft-affinity",
+		"soft-anti-affinity",
+	} {
+		if serverGroupPolicy == knownPolicy {
+			return nil
+		}
+	}
```
staebler, line 189: Using a set give somewhat better readability of intention.
```suggestion
	knownPolicies := sets.NewString("affinity", "anti-affinity", "soft-affinity", "soft-anti-affinity")
	if knownPolicies.Has(serverGroupPolicy) {
		return nil
	}
```

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -167,3 +168,34 @@ func validateFlavor(flavorName string, ci *CloudInfo, req flavorRequirements, fl
...
+
+	return field.ErrorList{field.Invalid(fldPath, serverGroupPolicy, "unknown server group policy")}
+}
+
+// validateWorkerServerGroupPolicy checks that the setting is empy, as the
+// option is only valid for Control plane nodes.
+func validateWorkerServerGroupPolicy(serverGroupPolicy string, fldPath *field.Path) field.ErrorList {
+	if serverGroupPolicy != "" {
+		return field.ErrorList{field.Invalid(fldPath, serverGroupPolicy, "setting not supported for the Compute MachinePool")}
```
staebler, line 198:
```suggestion
		return field.ErrorList{field.Invalid(fldPath, serverGroupPolicy, "server group policy cannot be set for compute machines")}
```

pkg/tfvars/openstack/openstack.go
```diff
@@ -38,6 +38,7 @@ type config struct {
 	BootstrapShim                    string   `json:"openstack_bootstrap_shim_ignition,omitempty"`
 	ExternalDNS                      []string `json:"openstack_external_dns,omitempty"`
 	MasterServerGroupName            string   `json:"openstack_master_server_group_name,omitempty"`
+	MasterServerGroupPolicy          string   `json:"openstack_master_server_group_policy,omitempty"`
```
staebler, line 41: This should not be omitempty, as it is a required variable that must be supplied to terraform.

pkg/tfvars/openstack/openstack.go
```diff
@@ -161,6 +162,12 @@ func TFVars(masterConfigs []*v1alpha1.OpenstackProviderSpec, cloud string, exter
 
 	cfg.MasterServerGroupName = masterConfig.ServerGroupName
 
+	if mpool != nil && mpool.ServerGroupPolicy != "" {
```
staebler, line 165: The value from `defaultmpool` must be taken into consideration as well.

pkg/types/openstack/machinepool.go
```diff
@@ -23,6 +23,10 @@ type MachinePool struct {
 	// +optional
 	AdditionalSecurityGroupIDs []string `json:"additionalSecurityGroupIDs,omitempty"`
 
+	// ServerGroupPolicy will be used to create the Server Group that will contain all the machines of this MachinePool.
+	// Defaults to "soft-anti-affinity".
+	ServerGroupPolicy string `json:"serverGroupPolicy,omitempty"`
```
staebler, line 28: Use a specific type rather than `string`. And set `kubebuilder` validations on the type so that it shows up in the explain for the field.

pkg/asset/installconfig/openstack/validate.go
```diff
@@ -65,7 +65,14 @@ func ValidateForProvisioning(ic *types.InstallConfig) error {
 	return nil
 }
 
-func defaultOpenStackMachinePoolPlatform() openstack.MachinePool {
+func defaultMasterOpenStackMachinePoolPlatform() openstack.MachinePool {
+	return openstack.MachinePool{
+		ServerGroupPolicy: "soft-anti-affinity",
```
staebler, line 70: Adding this here is not going to have any effect. This is only used for configuring the `Machine` and `MachineSets`. And `Machine` does not support the field.

pkg/asset/installconfig/openstack/validation/machinepool.go
pierreprinetti: Following your advice I have switched an array for a map, which is the most similar to a set that does not require yet another external dependency.

pierreprinetti:
@mandre @staebler 
I believe I have addressed all of your comments. Can you please take another look?

pierreprinetti:
`e2e-openstack` passed and the failed tests don't seem to have failed because of this change.

I think this is ready for review.

mandre (commented):
I feel a bit hesitant to put an lgtm now, after @staebler spotted so many issues that I missed in my first review :sweat_smile: (seriously, you did an _amazing_ review, thanks).

@pierreprinetti, do I remember correctly your previous iteration included unit tests?

pierreprinetti:
Now, with tests :)

pkg/types/openstack/validation/machinepool_test.go
```diff
@@ -0,0 +1,210 @@
...
+			check(noError),
+		},
+		{
+			"with valid server group policy",
+			testMachinePool(withServerGroupPolicy("anti-affinity")),
+			check(someErrorType(field.ErrorTypeInvalid)),
+		},
+		{
+			"with invalid server group policy",
+			testMachinePool(withServerGroupPolicy("anti-gravity")),
```
mandre, line 68: Nice one :smile: 
:man_astronaut:

mandre (commented):
/lgtm

pkg/asset/installconfig/openstack/validation/machinepool.go
staebler: We already have the external dependency.

pkg/types/openstack/machinepool.go
```diff
@@ -23,6 +33,10 @@ type MachinePool struct {
 	// +optional
 	AdditionalSecurityGroupIDs []string `json:"additionalSecurityGroupIDs,omitempty"`
 
+	// ServerGroupPolicy will be used to create the Server Group that will contain all the machines of this MachinePool.
+	// Defaults to "soft-anti-affinity".
+	ServerGroupPolicy `json:"serverGroupPolicy,omitempty"`
```
staebler, line 38: Use an explicit name for this field. I don't think you intend to inline this.

pkg/types/openstack/machinepool.go
```diff
@@ -77,3 +95,9 @@ type RootVolume struct {
 	// +optional
 	Zones []string `json:"zones,omitempty"`
 }
+
+// ServerGroupPolicy is the policy to be applied to an OpenStack Server Group.
+//
+// +kubebuilder:validation:Enum=affinity;soft-affinity;anti-affinity;soft-anti-affinity
```
staebler, line 101: Need to accept an empty string, too.
```suggestion
// +kubebuilder:validation:Enum="";affinity;soft-affinity;anti-affinity;soft-anti-affinity
```

pkg/types/openstack/machinepool.go
```diff
@@ -1,5 +1,15 @@
 package openstack
 
+// AllowedServerGroupPolicies is the set of allowed values for the
+// "ServerGroupPolicy" property of the MachinePool. The value will be set on
+// the Nova server group that will contain the machines.
+var AllowedServerGroupPolicies = map[ServerGroupPolicy]struct{}{
+	"affinity":           {},
```
staebler, line 7: Can we make these constants of type `ServerGroupPolicy`? That way the constant can be used in pkg/tfvars/openstack/openstack.go, too.

pkg/types/openstack/machinepool.go
```diff
@@ -1,5 +1,15 @@
 package openstack
 
+// AllowedServerGroupPolicies is the set of allowed values for the
+// "ServerGroupPolicy" property of the MachinePool. The value will be set on
+// the Nova server group that will contain the machines.
+var AllowedServerGroupPolicies = map[ServerGroupPolicy]struct{}{
```
staebler, line 6: Move this to a private variable in the ` pkg/types/openstack/validation` package.

pierreprinetti:
@staebler ptal.

PR #5004, opened 2021-06-16
Bug 1974598: OpenStack: Optimize cluster deletion

mandre:
Previously each destroy function would exit on the first conflict and
it was expected to retry on a next iteration, hoping that in the mean
time the conflict that prevented removal of the resource was fixed.
While this strategy works, it is also slower then necessary. A better
strategy is to try deleting all resources and ignore the ones that have
conflicts. On the next iteration there will be less conflicts.

This patch was tested on a openstack cluster with openshiftSDN, and
we observed a 37% faster cluster deletion. I expect the performance
boost to be significantly higher for clusters using kuryr.

### DIFF
--- a/pkg/destroy/openstack/openstack.go
+++ b/pkg/destroy/openstack/openstack.go
@@ -229,6 +229,8 @@ func deleteServers(opts *clientconfig.ClientOpts, filter Filter, logger logrus.F
 	}
 
 	filteredServers := filterObjects(serverObjects, filter)
+	numberToDelete := len(filteredServers)
+	numberDeleted := 0
 	for _, server := range filteredServers {
 		logger.Debugf("Deleting Server %q", server.ID)
 		err = servers.Delete(conn, server.ID).ExtractErr()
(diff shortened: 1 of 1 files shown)


staebler:
Nice. This matches more closely the desired destroy behavior that we am for with cloud platforms.

mandre:
/hold
I'd like to get numbers for kuryr cluster deletion. If the performance gain is really amazing we might consider backporting the change, in which case I'd like a BZ for it.

mandre:
My test on a kuryr cluster shows a 24% improvement on the time it takes to delete a cluster, 14m37s vs 19m14s. Not as impressive as I hoped it would be but still nice to take.

EmilienM:
```
INFO Time elapsed: 8m22s
```

Wow !!

/lgtm
/approve

mandre:
/retest
e2e-openstack should be working now, thanks to https://github.com/openshift/release/pull/19272

PR #5010, opened 2021-06-17
split infrastructure creation into multiple terraform stages

staebler:
Add a Stage interface which defines a given terraform stage for a given platform. For each stage, there is an option to include the
stage in the destroy when doing a bootstrap destroy. There is also an option to provide a custom destroy function. For example, when destroying the GCP bootstrap, there is a stage where the destroy behavior is modifying the load balancers to remove the bootstrap from the backend.

Use the Stage interface to run the terraform applies and destroys. The outputs from each stage are used as input into subsequent stages. This allows stages to get information that would either not be available at all or only available via a data source. For example, with GCP there is a stage where the health check IDs are needed. There is no data source available in the gcp terraform provider for accessing these IDs. However, using the output from the stage where the health checks are created, the stage is able to get the health check IDs as input instead.

Include the IPs of the bootstrap instance and the control plane instances in the output of the terraform stage where the instances
are created. This is used by the bootstrap gather as a consistent way to grab the IPs as opposed to extracting the IPs from the terraform state files.

When looking for the IPs of the bootstrap and control plane machines, consider all of the terraform stages. For compat stages, there is no change in behavior. For newer split stages, the gather looks through the output files for the "bootstrap_ip" and "control_plane_ips" outputs.

Split AWS and GCP into stages.

https://issues.redhat.com/browse/CORS-1693


staebler:
This combines #4876 and #4994.
/cc @jhixson74

staebler:
Still getting the instance profile errors, despite the change in the AWS terraform provider to increase the timeout to 2 minutes.
```
time="2021-06-17T03:31:13Z" level=error msg="Error: Error launching source instance: InvalidParameterValue: Value (ci-op-2rrgqzdi-1d3f3-bkqvr-bootstrap-profile) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name"
time="2021-06-17T03:31:13Z" level=error msg="\tstatus code: 400, request id: 44a7c552-1ded-4ce2-8bb7-bf08fccb35ce"
time="2021-06-17T03:31:13Z" level=error
time="2021-06-17T03:31:13Z" level=error msg="  on ../tmp/openshift-install-bootstrap-995973755/main.tf line 136, in resource \"aws_instance\" \"bootstrap\":"
time="2021-06-17T03:31:13Z" level=error msg=" 136: resource \"aws_instance\" \"bootstrap\" {"
```

Also, the fix for stopping the install on terraform errors has been lost.

staebler:
adceab20b...24a1010f1
* Fixed logic in `applyTerraform` where an error in the `terraform apply` was not correctly causing the installation to fail in the cases where a state file was created.

staebler:
24a1010f1...05e59bb80
* Updated to include an increase in the timeout waiting for the AWS instance to be created. https://github.com/openshift/terraform-provider-aws/pull/7

jhixson74 (approved):

PR #5026, opened 2021-06-23
BZ 1975475: aws: move creation of bootstrap instance profile to cluster stage

staebler:
When the bootstrap instance profile is created just before the bootstrap instance is created, then the RunInstance call will return an invalid value error for a little while. The terraform provider waits up to 2 minutes for this to resolve, which is does in that time. However, the terraform provider can, and frequently does, lose a response to a subsequent RunInstance. When this happens the terraform provider does not re-attempt the RunInstance request and instead sits idly waiting for the 2-minute timeout to elapse.

To combat this, we move the creation of the bootstrap instance profile to the cluster stage. This will create a delay between when the instance profile is created and when the instance is created. This is due to the instance profile being created early in the stage and there being a long delay while waiting for other resources in the stage to be created.

### DIFF
--- a/data/data/aws/bootstrap/main.tf
+++ b/data/data/aws/bootstrap/main.tf
@@ -68,75 +68,10 @@ resource "aws_s3_bucket_object" "ignition" {
   }
 }
 
-resource "aws_iam_instance_profile" "bootstrap" {
-  name = "${var.cluster_id}-bootstrap-profile"
-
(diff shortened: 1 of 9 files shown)


staebler:
Closing as the problem appears to be with the s3 bucket instead, which is addressed in https://github.com/openshift/installer/pull/5028.

/close

PR #5028, opened 2021-06-24
Bug 1975475: aws: block creation of bootstrap instance until ignition config is uploaded

staebler:
Give the bootstrap instance a dependency on the creation of the s3 bucket object holding the bootstrap's ignition config. The bootstrap instance requires the ignition config in order to complete its creation. It takes around 2 minutes to create the bucket object, and the instance creation has a 2-minute timeout. Consequently, there are times when the bootstrap fails to complete in time because the creation of the bucket object took too long.

### DIFF
--- a/data/data/aws/bootstrap/main.tf
+++ b/data/data/aws/bootstrap/main.tf
@@ -170,6 +170,8 @@ resource "aws_instance" "bootstrap" {
     },
     local.tags,
   )
+
+  depends_on = [aws_s3_bucket_object.ignition]
 }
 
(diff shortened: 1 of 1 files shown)


staebler:
This PR is 4 for 4 on successful AWS installs, so I think it is good.
/cc @jhixson74

staebler:
These changes do not affect upgrades.
/override e2e-aws-upgrade

e-tienne (commented):
/lgtm

staebler:
> These changes do not affect upgrades.
> /override e2e-aws-upgrade

Let me try again with the correct syntax.

PR #5029, opened 2021-06-24
Bug 1962414: FIPS: validate ssh public key type compatibility

e-tienne:
If a public key is configured in the install-config.yaml and fips is set to true,
validate that the ssh public key type is compatible with FIPS mode.
Only rsa and ecdsa types are valid at the moment.

https://bugzilla.redhat.com/show_bug.cgi?id=1962414

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -5,6 +5,7 @@ import (
 	"net"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


e-tienne:
/retitle Bug 1962414: FIPS: validate ssh public key type compatibility

jhixson74 (approved):
/lgtm

PR #5031, opened 2021-06-24
WIP: ovirt tf stages using terraform v0.14.6

staebler:
This is a WIP to see what needs to be split into separate stages for the ovirt platform.

PR #5034, opened 2021-06-25
ovirt: split terraform into stages

staebler:
Split the terraform for the ovirt platform into stages.

**image**: upload the rhcos image to a temporary vm
**cluster**: create the permanent cluster resources
**bootstrap**: create the bootstrap resources

This is needed for the upcoming upgrade of terraform to v0.14.6. With the more recent terraform version, the delete of the temporary rhcos image cascade to deletions of all VMs, since they have dependencies on the temporary image. This change moves the creation (and subsequent deletion) of the temporary rhcos image to its own stage to remove that terraform dependency.


staebler:
#5031 verifies that this solution works using terraform v0.14.6.

pkg/terraform/stages/ovirt/stages.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"github.com/openshift/installer/pkg/terraform"
+	"github.com/openshift/installer/pkg/terraform/stages"
+	ovirttypes "github.com/openshift/installer/pkg/types/ovirt"
+)
+
+// PlatformStages are the stages to run to provision the infrastructure in oVirt.
+var PlatformStages = []terraform.Stage{
+	stages.NewStage(ovirttypes.Name, "image", stages.WithNormalDestroy()),
```
Gal-Zaidman, line 11: I know it is not for this PR but as a general note, I think that WithNormalDestroy and WithCustomDestroy are confusing.
From what I understand WithNormalDestroy means it will be destroyed during bootstrap, and not passing anything means that stage.destroyWithBootstrap is false and then it is destroyed on regular destroy.
I think you should change it to WithBootstrapDestroy and WithCustomBootstrapDestroy to make that clearer, its also not documented on the function

pkg/terraform/stages/ovirt/stages.go
```diff
@@ -0,0 +1,16 @@
...
+	stages.NewStage(ovirttypes.Name, "cluster"),
+	stages.NewStage(ovirttypes.Name, "bootstrap", stages.WithNormalDestroy()),
+}
+
+// Note that the bootstrap IP and control-plane IP extraction is not supported on oVirt
```
Gal-Zaidman, line 16: We support extracting the bootstrap IP with: https://github.com/openshift/installer/blob/master/pkg/terraform/gather/ovirt/ip.go#L82-L114

We didn't implement control-plane IPs extraction and I think we just forgot about it since we can, @eslutsky is there a reason for not adding control plane IPs?

pkg/terraform/stages/compat/stage.go
```diff
@@ -121,16 +116,6 @@ func extractHostAddresses(config *types.InstallConfig, tfstate *terraform.State)
 		if err != nil {
 			logrus.Error(err)
 		}
-	case ovirttypes.Name:
-		bootstrap, err = gatherovirt.BootstrapIP(tfstate)
```
Gal-Zaidman, line 125: I think that both calls should be moved to "ExtractHostAddresses" or we need to add a mechanized for providers to implement this function, similar to "WithCustomDestroy"

pkg/terraform/stages/ovirt/stages.go
eslutsky: the original PR was for extracting IP information for bootstrap VM only, so i guess we missed it,
It should be an easy fix, I'll open PR for this.

data/data/ovirt/cluster/template/variables.tf
```diff
@@ -0,0 +1,18 @@
+variable "cluster_id" {
```
Gal-Zaidman, line 1: This is mainly for my knowledge,
On this PR you removed most of the variables files, for example the "image" and" "bootstrap" modules don't have any.
I guess you get the definition of the variable from "variables-ovirt".
Why do you need to include them in the template/master/affinity group?

Gal-Zaidman:
Changes looks good and thanks for cleaning up the "osImage" logic, I think you fixed a bug we have been postponing for a long time https://bugzilla.redhat.com/show_bug.cgi?id=1858418

pkg/terraform/stages/ovirt/stages.go
Gal-Zaidman: OK please link it here, I think we can still call the function on this PR just to not be blocked on this

eslutsky: https://github.com/openshift/installer/pull/5039

jhixson74:
> /hold

Is this on hold because it needs a rebase? I can certainly do that. Otherwise, feel free to provide any more feedback and I will get it taken care of.

Gal-Zaidman:
> > /hold
> 
> Is this on hold because it needs a rebase? I can certainly do that. Otherwise, feel free to provide any more feedback and I will get it taken care of.
> 
> I have rebased this PR [here](https://github.com/openshift/installer/pull/5079).

It will need a couple of fixes due to changes in the ovirt code parts that have been entered, and I want https://github.com/openshift/installer/pull/5070 to be merged before it:
- so it will be possible to backport if the customers will ask for it.
- I prefer to rebase this PR on both changes at once and not twice.

Plus we fixed the masters gathering(https://github.com/openshift/installer/pull/5039) so this PR will actually need a couple of fixes

And I would really want to fix this note as well https://github.com/openshift/installer/pull/5034/files?file-filters%5B%5D=.go&file-filters%5B%5D=.tf&hide-deleted-files=true#diff-988c9790bf4b4c5ffb0853edb9f389e8648b051970802d5c9af1b237d7adeb26R11

data/data/ovirt/cluster/template/variables.tf
staebler: The "image" and "bootstrap" modules are top-level modules, as is the "cluster" module. As you deduced, they get variables from `data/data/config.tf` and `data/data/variables-ovirt.tf`. They can have additional files with more variables corresponding to outputs from previous stages. For example, `data/data/ovirt/bootstrap/variables.tf` defines the `release_image_template_id` variable, which corresponds to the output from the "image" stage defined in `data/data/ovirt/image/outputs.tf`.

The "afffinity_group" module is a sub-module of the top-level "cluster" module. Sub-modules only get the variables that are defined directly in them. The values of these variables must be specified explicitly in the calling "module" block.

staebler:
This cannot wait for 
/hold cancel


> > > /hold
> > 
> > 
> > Is this on hold because it needs a rebase? I can certainly do that. Otherwise, feel free to provide any more feedback and I will get it taken care of.
> > I have rebased this PR [here](https://github.com/openshift/installer/pull/5079).
> 
> It will need a couple of fixes due to changes in the ovirt code parts that have been entered, and I want #5070 to be merged before it:
> 
>     * so it will be possible to backport if the customers will ask for it.
> 
>     * I prefer to rebase this PR on both changes at once and not twice.
> 

We cannot wait for #5070 any longer. These changes in this PR need to be included in 4.10.

> 
> Plus we fixed the masters gathering(#5039) so this PR will actually need a couple of fixes

Done.

> 
> And I would really want to fix this note as well https://github.com/openshift/installer/pull/5034/files?file-filters%5B%5D=.go&file-filters%5B%5D=.tf&hide-deleted-files=true#diff-988c9790bf4b4c5ffb0853edb9f389e8648b051970802d5c9af1b237d7adeb26R11
> 
> I will probably get to it after feature freeze

/hold cancel

staebler:
@Gal-Zaidman @janosdebugs Could you review these terraform changes again, please? This PR needs to merge soon so that the terraform upgrade which is blocked on this PR can be merged in 4.10.

pkg/terraform/stages/ovirt/stages.go
staebler: https://github.com/openshift/installer/pull/5476

janosdebugs (approved):
/lgtm

staebler:
5a42bb8c0...b632e343b
* Fix go fmt issues.

jhixson74 (approved):
/lgtm

staebler:
No changes in this PR have an effect on AWS installs.

PR #5039, opened 2021-06-29
Bug 1979038: oVirt: include master IPs when running must gather

eslutsky:
`gather bootstrap` command will collect logs from the VMs only if the boostrap VM still exists.
this change adds the control planes VMs IP information, resulting the master logs
to be collected as well.

Signed-off-by: Evgeny Slutsky <eslutsky@redhat.com>

### DIFF
--- a/pkg/terraform/gather/ovirt/ip.go
+++ b/pkg/terraform/gather/ovirt/ip.go
@@ -55,11 +55,10 @@ func getReportedDevices(c *ovirtsdk4.Connection, vmID string) (*ovirtsdk4.Report
 	return reportedDeviceSlice, nil
 }
 
-func findVirtualMachineIP(c *ovirtsdk4.Connection, moRefValue string) (string, error) {
-
-	reportedDeviceSlice, err := getReportedDevices(c, moRefValue)
-	if err != nil {
-		return "", err
+func findVirtualMachineIP(moRefValue string, client *ovirtsdk4.Connection) (string, error) {
+	reportedDeviceSlice, err := getReportedDevices(client, moRefValue)
+	if err == nil {
+		return "", errors.Wrapf(err, "couldnt Find IP Address for vm id: %s", moRefValue)
 	}
 
 	for _, reportedDevice := range reportedDeviceSlice.Slice() {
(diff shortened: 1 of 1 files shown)


pkg/terraform/gather/ovirt/ip.go
```diff
@@ -55,9 +55,16 @@ func getReportedDevices(c *ovirtsdk4.Connection, vmID string) (*ovirtsdk4.Report
 	return reportedDeviceSlice, nil
 }
 
-func findVirtualMachineIP(c *ovirtsdk4.Connection, moRefValue string) (string, error) {
+func findVirtualMachineIP(moRefValue string) (string, error) {
 
-	reportedDeviceSlice, err := getReportedDevices(c, moRefValue)
+	client, err := ovirt.NewConnection()
```
janoszen, line 60: Why are we creating a new oVirt connection here?

pkg/terraform/gather/ovirt/ip.go
```diff
@@ -69,52 +76,54 @@ func findVirtualMachineIP(c *ovirtsdk4.Connection, moRefValue string) (string, e
 				ipres, hasAddress := ip.Address()
 				if hasAddress {
 					if checkPortIsOpen(ipres, bootstrapSSHPort) {
-						logrus.Debugf("ovirt vm id: %s , found usable bootstrap IP %s", moRefValue, ipres)
+						logrus.Debugf("ovirt vm id: %s , found usable  IP Address: %s", moRefValue, ipres)
```
janoszen, line 79: Is this log useful to the user anyhow or is this just debug code?

pkg/terraform/gather/ovirt/ip.go
```diff
@@ -55,9 +55,16 @@ func getReportedDevices(c *ovirtsdk4.Connection, vmID string) (*ovirtsdk4.Report
...
+	client, err := ovirt.NewConnection()
+	if err != nil {
+		return "", fmt.Errorf("failed to initialize connection to ovirt-engine's %s", err)
+	}
+	defer client.Close()
+
+
+	reportedDeviceSlice, err := getReportedDevices(client, moRefValue)
 	if err != nil {
 		return "", err
```
janoszen, line 69: This error should possibly be wrapped so we have context.

pkg/terraform/gather/ovirt/ip.go
```diff
@@ -69,52 +76,54 @@ func findVirtualMachineIP(c *ovirtsdk4.Connection, moRefValue string) (string, e
...
-
-	br, err := terraform.LookupResource(tfs, "module.bootstrap", "ovirt_vm", "bootstrap")
-	if err != nil {
-		return "", errors.Wrap(err, "failed to lookup bootstrap")
+		return []string{""}, errors.Wrapf(err, "failed to lookup %s VM",name)
 	}
 
 	if len(br.Instances) == 0 {
-		return "", errors.New("no bootstrap instance found")
+		return []string{""}, errors.New(fmt.Sprintf("no %s instance found",name))
```
janoszen: Returning an empty IP seems like a workaround for something. It seems that this should not be handled here. If yes, we should add comments explaining what this does and why it is ok to return an empty string IP.

pkg/terraform/gather/ovirt/ip.go
```diff
@@ -69,52 +76,54 @@ func findVirtualMachineIP(c *ovirtsdk4.Connection, moRefValue string) (string, e
...
 	}
-	return ip, nil
+
+	return ips, nil
+}
+
+// BootstrapIP returns the ip address for bootstrap host.
+func BootstrapIP(tfs *terraform.State) (string, error) {
+	ips, err := lookupVMResources(tfs, "module.bootstrap", "bootstrap")
+	return ips[0], err
```
janoszen, line 123: Add error handling here if ips has no entries.

pkg/terraform/gather/ovirt/ip.go
```diff
@@ -69,52 +76,54 @@ func findVirtualMachineIP(c *ovirtsdk4.Connection, moRefValue string) (string, e
...
+		vmid, found, err := unstructured.NestedString(instance.Attributes, "id")
+		if err != nil {
+			return []string{""}, errors.Wrapf(err, "failed to lookup %s managed object reference",name)
+		}
+		if !found {
+			return []string{""}, errors.Errorf("failed to lookup %s by vmID: %s", name, vmid)
+		}
+		ip, err := findVirtualMachineIP(vmid)
+		if err != nil {
+			return []string{""}, errors.Wrapf(err, "failed to lookup %s ipv4 address",name)
```
janoszen: Same as above.

janoszen (changes requested):

pkg/terraform/gather/ovirt/ip.go
eslutsky: its for debug

pkg/terraform/gather/ovirt/ip.go
eslutsky: moved to the connection one level up (outside the for loop)

pkg/terraform/gather/ovirt/ip.go
eslutsky: okay , in case of error we can return nil instead

pkg/terraform/gather/ovirt/ip.go
```diff
@@ -55,9 +55,8 @@ func getReportedDevices(c *ovirtsdk4.Connection, vmID string) (*ovirtsdk4.Report
 	return reportedDeviceSlice, nil
 }
 
-func findVirtualMachineIP(c *ovirtsdk4.Connection, moRefValue string) (string, error) {
-
-	reportedDeviceSlice, err := getReportedDevices(c, moRefValue)
+func findVirtualMachineIP(moRefValue string,client *ovirtsdk4.Connection) (string, error) {
```
janoszen, line 58: This change seems unnecessary. Not a big deal, but still.

pkg/terraform/gather/ovirt/ip.go
janoszen: After the last update this will cause a panic.

janoszen (changes requested):

pkg/terraform/gather/ovirt/ip.go
eslutsky: done

pkg/terraform/gather/ovirt/ip.go
eslutsky: done

pkg/terraform/gather/ovirt/ip.go
eslutsky: done

pkg/terraform/gather/ovirt/ip.go
eslutsky: done

janoszen (approved):

Gal-Zaidman:
/assign @staebler 

Required for https://github.com/openshift/installer/pull/5034

PR #5042, opened 2021-06-29
Azure Stack cloud provider config

patrickdillon:
These commits add specific cloud provider configuration for Azure Stack Hub. 

be4635c - Adds generic support, separating configuration that differs from public Azure. 

d210770 - Is a temporary commit to add credentials to the cloud provider config. Once bootstrap support for the cloud controller manager is added we should revert this commit and keep credentials in a secret, which is not presently supported as explained in the commit message.

Merging d210770 would allow us to make progress with fully automated IPI install and reverting later would be simple enough. I can create a revert PR now with a basic PoC of how to use the merge config based on ARO.

### DIFF
--- a/pkg/asset/machines/azure/machines.go
+++ b/pkg/asset/machines/azure/machines.go
@@ -102,7 +102,7 @@ func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string
 	}
 
 	managedIdentity := fmt.Sprintf("%s-identity", clusterID)
-	if platform.IsARO() {
+	if platform.IsARO() || platform.CloudName == azure.StackCloud {
 		managedIdentity = ""
 	}
 
--- a/pkg/asset/manifests/azure/cloudproviderconfig.go
+++ b/pkg/asset/manifests/azure/cloudproviderconfig.go
@@ -25,17 +25,13 @@ type CloudProviderConfig struct {
 // JSON generates the cloud provider json config for the azure platform.
 // managed resource names are matching the convention defined by capz
 func (params CloudProviderConfig) JSON() (string, error) {
-	useManagedIdentityExtension := true
-	if params.ARO {
-		useManagedIdentityExtension = false
-	}
 
 	config := config{
 		authConfig: authConfig{
 			Cloud:                       params.CloudName.Name(),
 			TenantID:                    params.TenantID,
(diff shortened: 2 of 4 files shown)


lobziik:
/lgtm
Used this in combination with https://github.com/openshift/installer/pull/5084. Worked well for me.

lobziik:
@patrickdillon, maybe we can add ARM endpoint to cloud config?
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/auth/azure_auth.go#L68

Actually, i think we can skip part with dumping endpoints config onto config-map and just put arm endpoint into cloud config.
Now it somewhat buggy within authorization procedure:
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/auth/azure_auth.go#L228
https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/auth/azure_auth.go#L268 - this override we miss if passing endpoints config as file without extra mutations.

patrickdillon:
> @patrickdillon, maybe we can add ARM endpoint to cloud config?
> https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/auth/azure_auth.go#L68
> 
> Actually, i think we can skip part with dumping endpoints config onto config-map and just put arm endpoint into cloud config.
> Now it somewhat buggy within authorization procedure:
> https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/auth/azure_auth.go#L228
> https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/pkg/auth/azure_auth.go#L268 - this override we miss if passing endpoints config as file without extra mutations.

ARM endpoint is in the infrastructure object: 
https://github.com/openshift/installer/blob/master/pkg/asset/manifests/infrastructure.go#L149
& 
https://github.com/openshift/api/blob/master/config/v1/types_infrastructure.go#L381

Can you pull it from there?

lobziik:
>Can you pull it from there?

Will have to mutate cloud-config in operator. No other way to pass this thing to ccm. Would be waaay more easy and convenient to put it there in installer.

patrickdillon:
Added in 12251e73ca48a0fbebad41ca342df01049d81cbc @lobziik

lobziik:
/lgtm

Thanks a lot!

PR #5044, opened 2021-06-30
azure: Updates cluster-api-provider-azure package

m1kola:
New package version brings support for VM SecurityProfile property which is required to unblock work related to encryption at host.


m1kola:
/retest

I beleive e2e failures are not related

PR #5047, opened 2021-07-01
OpenStack: explicitly disable octavia when using kuryr

mandre:
Kuryr already handles the service type Load Balancer and we need to
explicitly disable the use of octavia in the cloud provider
configuration otherwise we'll run into issues once the default for
`use-octavia` changes from false to true with the out-of-tree openstack
cloud provider.

Implements [OSASINFRA-2510](https://issues.redhat.com/browse/OSASINFRA-2510)

### DIFF
--- a/pkg/asset/manifests/openstack/cloudproviderconfig.go
+++ b/pkg/asset/manifests/openstack/cloudproviderconfig.go
@@ -91,8 +91,11 @@ secret-namespace = kube-system
 		cloudProviderConfigCABundleData = string(caFile)
 	}
 
+	cloudProviderConfigData += "[LoadBalancer]\n"
 	if installConfig.NetworkType == string(operv1.NetworkTypeKuryr) {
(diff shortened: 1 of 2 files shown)


mandre:
/hold
I need to fix the tests.

MaysaMacedo:
Looks good to me. Just a side note, this config won't be enforced to False in case the user attempts to set it to True, right?
Do we plan to include a note on the docs?

mandre:
> Looks good to me. Just a side note, this config won't be enforced to False in case the user attempts to set it to True, right?

Just like today, the user can change the cloud provider configuration, either:
- by generating the manifests and editing them prior to the deployment
- as a day 2 operation, with `oc edit cm cloud-provider-config -n openshift-config`

Nothing was changed in that regard, we don't perform validation for the above. The patch only changes the default configuration to match more closely the environment on which the user deploys. 

> Do we plan to include a note on the docs?

I've just added a note in the 4.9 release note tracker, thanks for the reminder.

PR #5048, opened 2021-07-01
Bug 1978213: Run kuryr-specific quota checks for kuryr envs

mandre:
Due to a missing capitalization we didn't run the kuryr-specific network
quota checks for OpenStack validation which could result in the
installation starting even though the user has not nearly enough quota
for a successful Kuryr deployment.


dulek:
/lgtm

Whoops, nice finding.

pierreprinetti:
/approve
/hold

Please do unhold when the CI is green.

mandre:
The jobs we care about are green.
/hold cancel

PR #5050, opened 2021-07-01
OpenStack: Remove FIPs of LBs created by cloud-provider

dulek:
To implement `LoadBalancer` `Services` OpenStack Cloud Provider creates an
LB in Octavia for each of them. Those LBs have floating IPs associated
to the LB VIPs. As this is handled by the Cloud Provider itself,
Octavia's cascade delete will not remove those FIPs and we need to take
care of that ourselves.

This commit adds deleting of the FIPs associated to the LBs on
destroying cluster. In order to do that correctly it was required to
split deleting routers into two separate functions and make sure that
FIP detach and actual removal of the routers is happening after all the
delete functions finished (meaning that LBs are gone too). This is
because if FIP detach happens before we lose information about which FIP
was attached to which LB effectively preventing us from handling
deletion of them.

Only the function detaching the subnets from the routers is running as
part of main deletion step now.

### DIFF
--- a/pkg/destroy/openstack/openstack.go
+++ b/pkg/destroy/openstack/openstack.go
@@ -89,27 +89,27 @@ func (o *ClusterUninstaller) Run() error {
 	// deleteFuncs contains the functions that will be launched as
 	// goroutines.
 	deleteFuncs := map[string]deleteFunc{
-		"deleteServers":        deleteServers,
-		"deleteServerGroups":   deleteServerGroups,
-		"deleteTrunks":         deleteTrunks,
-		"deleteLoadBalancers":  deleteLoadBalancers,
-		"deletePorts":          deletePorts,
-		"deleteSecurityGroups": deleteSecurityGroups,
-		"deleteRouters":        deleteRouters,
-		"deleteSubnets":        deleteSubnets,
-		"deleteSubnetPools":    deleteSubnetPools,
-		"deleteNetworks":       deleteNetworks,
-		"deleteContainers":     deleteContainers,
-		"deleteVolumes":        deleteVolumes,
-		"deleteShares":         deleteShares,
-		"deleteFloatingIPs":    deleteFloatingIPs,
-		"deleteImages":         deleteImages,
+		"deleteServers":          deleteServers,
+		"deleteServerGroups":     deleteServerGroups,
+		"deleteTrunks":           deleteTrunks,
+		"deleteLoadBalancers":    deleteLoadBalancers,
+		"deletePorts":            deletePorts,
+		"deleteSecurityGroups":   deleteSecurityGroups,
+		"clearRoutersInterfaces": clearRoutersInferfaces,
+		"deleteSubnets":          deleteSubnets,
+		"deleteSubnetPools":      deleteSubnetPools,
+		"deleteNetworks":         deleteNetworks,
+		"deleteContainers":       deleteContainers,
+		"deleteVolumes":          deleteVolumes,
+		"deleteShares":           deleteShares,
+		"deleteFloatingIPs":      deleteFloatingIPs,
+		"deleteImages":           deleteImages,
 	}
 	returnChannel := make(chan string)
 
 	opts := openstackdefaults.DefaultClientOpts(o.Cloud)
 
-	err := cleanRouterRunner(opts, o.Filter, o.Logger)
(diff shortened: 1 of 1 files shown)


pkg/destroy/openstack/openstack.go
```diff
@@ -449,31 +457,108 @@ func updateFips(allFIPs []floatingips.FloatingIP, opts *clientconfig.ClientOpts,
...
 	}
 
-	allRouters, err := routers.ExtractRouters(allPages)
+	numberToClear := len(allRouters)
+	numberCleaned := 0
+	for _, router := range allRouters {
+		_, err = removeRouterInterfaces(conn, filter, router, logger)
+		if err != nil {
+			continue
+		}
```
pierreprinetti, line 539: Do you want to `logger.Debug` the error instead of silently ignoring?

dulek: It was like this before [1], but yeah.

[1] https://github.com/openshift/installer/pull/5050/files/d87b0c6138b17303dd4e3e195cf15d7e735e8629#diff-e480f6c0393b6788fe8352f4f8bdc63f021f9e84de632a21607aab8b414c6e3fL512-L515

pkg/destroy/openstack/openstack.go
```diff
@@ -125,6 +125,14 @@ func (o *ClusterUninstaller) Run() error {
 		o.Logger.Debugf("goroutine %v complete", res)
 	}
 
+	// we want to remove routers as the last thing as it requires detaching the
+	// FIPs and that will cause it impossible to track which FIPs are tied to
+	// LBs being deleted.
+	err = deleteRouterRunner(opts, o.Filter, o.Logger)
```
mandre, line 131: Is there a risk the router is going to block the completion of any other resource deletion? If so you'll never reach this code. I think not but just want to make sure...

mandre: Hmm, after looking at CI this is probably what's happening:
```
 INFO[2021-07-02T11:49:09Z] Deprovisioning cluster ...
{"component":"entrypoint","file":"prow/entrypoint/run.go:165","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 45m0s timeout","severity":"error","time":"2021-07-02T11:39:06Z"} 
```

mandre: If that's really what's happening you need to extract the deletion of the blocking resources to your second runner.

dulek: Yep, I'll need to analyze that. :/

dulek: Oh, seems like if it fails to remove an interface from the router it won't keep retrying so interface is left and blocks subnet removal. Easy fix but I need to check if it works.

pkg/destroy/openstack/openstack.go
```diff
@@ -449,31 +457,109 @@ func updateFips(allFIPs []floatingips.FloatingIP, opts *clientconfig.ClientOpts,
...
+
+	allRouters, err := getRouters(opts, filter, logger)
 	if err != nil {
 		logger.Error(err)
 		return false, nil
 	}
 
-	allRouters, err := routers.ExtractRouters(allPages)
+	numberToClear := len(allRouters)
+	numberCleaned := 0
```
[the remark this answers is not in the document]
dulek: I wrote that initially and changed it. My brain is playing with me.

pkg/destroy/openstack/openstack.go
```diff
@@ -449,31 +457,112 @@ func updateFips(allFIPs []floatingips.FloatingIP, opts *clientconfig.ClientOpts,
...
+	numberCleared := 0
+	for _, router := range allRouters {
+		removed, err := removeRouterInterfaces(conn, filter, router, logger)
+		if err != nil {
+			logger.Debug(err)
+			continue
+		}
+		if !removed {
+			continue
+		}
```
dulek, line 543: @mandre: This was the problem, function's result was ignored. This piece seems ugly, I'm happy to hear suggestions how to code it better.

dulek:
Looks like it worked!

mandre (commented):
/lgtm

mandre:
Nicely done @dulek 
/approve

PR #5052, opened 2021-07-01
OpenStack: Open 0.0.0.0/0 on NodePorts

dulek:
For `LoadBalancer` `Services` OpenStack cloud provider creates an
Octavia loadbalancer and adds all the nodes as members pointing to the
NodePort on the hosts. Previously the SGs opening traffic on those ports
were constrained to CIDR of the nodes subnet. This works fine in case of
Octavia with Amphora driver, as HAProxy will change the source-IP of the
traffic to the IP of the LB:
```
  traffic source-IP: (x.x.x.x)               (y.y.y.y)
  caller (x.x.x.x)      ->     LB (y.y.y.y)     ->     member
```
The LBs are created in nodes subnet, so that worked. Things are
different in case of OVN Octavia driver:
```
  traffic source-IP: (x.x.x.x)               (x.x.x.x)
  caller (x.x.x.x)      ->     LB (y.y.y.y)     ->     member
```
OVN preserves the source-IP of the original caller, meaning that SG
opening just nodes subnet CIDR is not enough as it would only allow
traffic from inside the cluster.

This commit switches those rules opening NodePorts on nodes to allow
0.0.0.0/0.

### DIFF
--- a/data/data/openstack/topology/sg-master.tf
+++ b/data/data/openstack/topology/sg-master.tf
@@ -205,23 +205,25 @@ resource "openstack_networking_secgroup_rule_v2" "master_ingress_etcd" {
 }
 
 resource "openstack_networking_secgroup_rule_v2" "master_ingress_services_tcp" {
-  direction         = "ingress"
-  ethertype         = "IPv4"
-  protocol          = "tcp"
-  port_range_min    = 30000
-  port_range_max    = 32767
-  remote_ip_prefix  = var.cidr_block
+  direction      = "ingress"
+  ethertype      = "IPv4"
+  protocol       = "tcp"
+  port_range_min = 30000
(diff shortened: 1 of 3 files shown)


mandre (changes requested):
Looks good. Could you also update the UPI scripts in a separate commit?

mandre:
Looks like openstacksdk 0.53 broke UPI:
> fatal: [localhost]: FAILED! => {"changed": false, "msg": "BadRequestException: 400: Client Error for url: https://kaizen.massopen.cloud:13696/v2.0/security-group-rules, Unrecognized attribute(s) 'remote_address_group_id'"}

Upstream issue: https://storyboard.openstack.org/#!/story/2008577

PR #5054, opened 2021-07-02
baremetal: Update IPI docs for latest version

hardys:
BM IPI is no longer experimental, and some of the content is no
longer accurate, this PR does a first-pass update of the content.


hardys:
I noticed a markdown issue with the table formatting for `bootMode`, now fixed

PR #5061, opened 2021-07-07
vendor: update baremetal-operator v0.0.0-20210706141527-5240e42f012a

derekhiggins:
Specifically looking to include https://github.com/metal3-io/baremetal-operator/pull/853
which will allow baremetal nodes to be provisioned with ipmi when
full Administrator BMC credentials arn't available.

Also had to add a new entry in the replace section for "baremetal-operator/apis"
to allow for a change in the way BMO vendors this in.


imain:
lgtm

PR #5062, opened 2021-07-08
Fix build with multiple GOFLAGS

yselkowitz:
GOFLAGS is a space-separated value; quoting it here results in errors.

/cc @sosiouxme

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -50,4 +50,5 @@ then
 	export CGO_ENABLED=1
 fi
 
-go build "${GOFLAGS}" -ldflags "${LDFLAGS}" -tags "${TAGS}" -o "${OUTPUT}" ./cmd/openshift-install
+# shellcheck disable=SC2086
+go build ${GOFLAGS} -ldflags "${LDFLAGS}" -tags "${TAGS}" -o "${OUTPUT}" ./cmd/openshift-install


yselkowitz:
This is to prepare for a hopefully more reliable workaround for build failures on ARM.
/assign @crawford 
/retest

patrickdillon:
Can you provide more context about the error you see and the reason removing the quotes fixes the problem? It seems like the opposite would be true: that because you have whitespace separated values, you need the quotes; so I could use some help to understand the issue.

yselkowitz:
```
GOFLAGS="-mod=vendor -p=8" ./hack/build.sh
[snip]
+ go build '-mod=vendor -p=8' -ldflags ' -X github.com/openshift/installer/pkg/version.Raw=unreleased-master-4791-g93204844e228d9f8cf13e454219246c6d79c3bc3 -X github.com/openshift/installer/pkg/version.Commit=93204844e228d9f8cf13e454219246c6d79c3bc3 -X github.com/openshift/installer/pkg/version.defaultArch=amd64 -s -w' -tags ' release' -o bin/openshift-install ./cmd/openshift-install
-mod=vendor -p=8 not supported (can be '', 'mod', 'readonly', or 'vendor')
```

patrickdillon:
Thanks for the clarification.

/approve

patrickdillon:
@yselkowitz do you want to wait for input from @sosiouxme ?

sosiouxme:
/retest

this expands our options (literally) and seems unlikely to break anything (unless someone somewhere is setting a single option that has spaces in it). we have a use case where this may solve persistent compile failures. so let's get it in...

yselkowitz:
I think we're going to need some overrides here in order to unblock downstream builds..
/assign @staebler

sosiouxme:
/retest
to infinity and beyond

sosiouxme:
ffs
/retest

PR #4923, opened 2021-05-10
IBM Cloud Provider Scaffolding

BobbyRadford:
## Initial scaffolding for the IBM Cloud provider

With this PR, a user should be able to:

1. `openshift-install create install-config` and follow CLI prompts to generate a bare bones install-config.yaml
2. `openshift-install create install-config --dir=./existing-config` where `./existing-config` has a customized install-config.yaml with IBM Cloud platform and machine pool customizations.

This connects with the following enhancement doc PR: https://github.com/openshift/enhancements/pull/773 Details about the supported install config customizations can be found there. 

CI has not yet been implemented for the IBM Cloud platform, so this PR should not be merged until that is complete. The scaffolding code proposed here is fairly comprehensive, already containing a suite of unit tests. However, it is not fully done. Some validation on customizations is not yet implemented, specifically around byo VPC and subnets. This byo VPC and subnets may not end up in an MVP for the IBM Cloud platform anyway.

With this PR, I hope that we can get some clarity on direction and verification that we are doing this the "right way." Hopefully that can happen prior to CI being fully implemented.

PR #5063, opened 2021-07-08
ibmcloud: Add Platform Provisioning Check

hasueki:
This adds platform provisioning check support for IBM Cloud. The check will validate if there's already a conflicting Kube API DNS record present in the user-provided CIS instance.

Changes:
- New validation function to check DNS records, `ValidatePreExitingPublicDNS()`
- New client method for fetching DNS records, `GetDNSRecordsByName()`
- Rename client method,`GetZoneIDByName() --> GetDNSZoneIDByName()`
- Unit test for new validation function
- ~~Move validation test suite to `ibmcloud` package~~
- ~~Move custom type definitions, `DNSZoneResponse{}` and `EncryptionKeyResponse{}` to `pkg/types/ibmcloud`~~


pkg/asset/installconfig/ibmcloud/client.go
```diff
@@ -52,29 +55,6 @@ func (e *VPCResourceNotFoundError) Error() string {
 	return "Not Found"
 }
 
-// DNSZoneResponse represents a DNS zone response.
-type DNSZoneResponse struct {
```
BobbyRadford, line 56: I actually originally had this in `pkg/types/ibmcloud` and was asked to keep it here. Is there a reason for moving it?

pkg/asset/installconfig/ibmcloud/validation_test.go
```diff
@@ -36,34 +42,35 @@ var (
...
 	subnetInvalidZone      = func(ic *types.InstallConfig) { ic.IBMCloud.Subnets = []string{"subnet-invalid-zone"} }
 	machinePoolInvalidType = func(ic *types.InstallConfig) {
-		ic.ControlPlane.Platform.IBMCloud = &ibmcloudtypes.MachinePool{
+		ic.ControlPlane.Platform.IBMCloud = &ibmcloud.MachinePool{
 			InstanceType: "invalid-type",
 		}
 	}
 
-	dnsZoneResponses = []ibmcloud.DNSZoneResponse{
+	existingDnsRecordsResponse = []dnsrecordsv1.DnsrecordDetails{
```
BobbyRadford, line 58: casing inconsistency: `existingDNSRecordsResponse`

pkg/asset/installconfig/ibmcloud/validation_test.go
```diff
@@ -36,34 +42,35 @@ var (
...
-			ID:   "valid-zone-id-1",
+			ID: core.StringPtr("valid-dns-record-1"),
 		},
 		{
-			Name: "another.domain",
-			ID:   "valid-zone-id-2",
+			ID: core.StringPtr("valid-dns-record-2"),
 		},
 	}
+	noDnsRecordsResponse = []dnsrecordsv1.DnsrecordDetails{}
```
BobbyRadford, line 66:
```suggestion
	noDNSRecordsResponse = []dnsrecordsv1.DnsrecordDetails{}
```

BobbyRadford (changes requested):

pkg/asset/installconfig/ibmcloud/client.go
BobbyRadford: Is it because of this?
> Move validation test suite to ibmcloud package

pkg/asset/installconfig/ibmcloud/client.go
```diff
@@ -140,8 +123,51 @@ func (c *Client) GetCISInstance(ctx context.Context, crnstr string) (*resourceco
...
 }
 
+// GetDNSRecordsByName gets DNS records in specific Cloud Internet Services instance
+// by its CRN, zone ID, and DNS record name.
+func (c *Client) GetDNSRecordsByName(ctx context.Context, crnstr string, zoneID string, recordName string) ([]dnsrecordsv1.DnsrecordDetails, error) {
+	// Set CIS DNS record service
+	dnsService, err := dnsrecordsv1.NewDnsRecordsV1(&dnsrecordsv1.DnsRecordsV1Options{
+		Authenticator:  c.Authenticator,
+		Crn:            &crnstr,
+		ZoneIdentifier: &zoneID,
```
BobbyRadford, line 133:
```suggestion
		Crn:            core.StringPtr(crnstr),
		ZoneIdentifier: core.StringPtr(zoneID),
```

pkg/asset/installconfig/ibmcloud/client.go
```diff
@@ -140,8 +123,51 @@ func (c *Client) GetCISInstance(ctx context.Context, crnstr string) (*resourceco
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Get CIS DNS records by name
+	records, _, err := dnsService.ListAllDnsRecordsWithContext(ctx, &dnsrecordsv1.ListAllDnsRecordsOptions{
+		Name: &recordName,
```
BobbyRadford, line 141:
```suggestion
		Name: core.StringPtr(recordName),
```

pkg/asset/installconfig/ibmcloud/client.go
```diff
@@ -140,8 +123,51 @@ func (c *Client) GetCISInstance(ctx context.Context, crnstr string) (*resourceco
...
+		return nil, errors.Wrap(err, "could not retrieve DNS records")
+	}
+
+	return records.Result, nil
+}
+
+// GetDNSZoneIDByName gets the CIS zone ID from its domain name.
+func (c *Client) GetDNSZoneIDByName(ctx context.Context, name string) (string, error) {
+	_, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
```
BobbyRadford, line 153:
```suggestion
```

Don't need this. I know this is coming from other functions in here, but a recent comment from a different PR review pointed this out.

pkg/asset/installconfig/ibmcloud/client.go
```diff
@@ -52,29 +55,6 @@ func (e *VPCResourceNotFoundError) Error() string {
...
-	// CISInstanceName is the display name of the CIS instance where the DNS zone
-	// is managed.
-	CISInstanceName string
-
-	// ResourceGroupID is the resource group ID of the CIS instance.
-	ResourceGroupID string
-}
-
-// EncryptionKeyResponse represents an encryption key response.
-type EncryptionKeyResponse struct{}
```
BobbyRadford, line 76: Same question for this type

pkg/asset/installconfig/ibmcloud/validation_test.go
```diff
@@ -1,24 +1,30 @@
-package ibmcloud_test
```
BobbyRadford, line 1: This was an easy way to avoid moving types

BobbyRadford: I believe there may be a way to avoid this and keep types within the ibmcloud package using some genmock magic

BobbyRadford (changes requested):

pkg/asset/installconfig/ibmcloud/client.go
hasueki: I noticed the validation test suite for ibmcloud install config was in it's own package `ibmcloud_test`, which other providers do not do. I figured it will be good to have consistency in how the test files are structured. I attempted to simply update the package name to `ibmcloud`, but ran into `import cycle not allowed in test` error. So these type definitions needed to be moved elsewhere to resolve the issue.

BobbyRadford: Yeah. That's why I ended up putting it in the `ibmcloud_test` package. Otherwise, there would need to be some non-trivial refactoring to avoid the import cycle and keep the types contained within the `ibmcloud` package. As I mentioned in the other comment, there may be a way to do this with mockgen arguments, but I haven't gone too deep into that.

hasueki: I can't seem to find a good way to resolve this when these types are defined in the same package. I've looked into mockgen options, and also simply trying to refactor the code, but it's making it especially difficult because they are used as return types. Was there a reason that these types should not be in `pkg/types/ibmcloud`? Seems like an appropriate place to be.

BobbyRadford: https://github.com/openshift/installer/pull/4923#discussion_r632740396

hasueki: Yes, I came across this comment as well. For time sake, I can just push a commit to revert the refactoring bit.

jstuever (approved):

PR #5067, opened 2021-07-09
Bug 1980866: [4.7] Prefer IPv6 hostIP on bootstrap IPv6 deployments

sadasu:
Prefer IPv6 hostIP on bootstrap IPv6 deployments 

Kubelet can end up choosing IPv4 addresses as its hostIP, which after:

https://github.com/openshift/cluster-kube-apiserver-operator/pull/1042/files

mean that the kubernetes API service ends up with an endpoint being an
IPv4 address.  This breaks the kubernetes API loadbalancer in the
service network and prevent the deployment from succeeding.

Backport of https://github.com/openshift/installer/pull/4756

### DIFF
--- a/data/data/bootstrap/systemd/units/kubelet.service.template
+++ b/data/data/bootstrap/systemd/units/kubelet.service.template
@@ -18,6 +18,9 @@ ExecStart=/usr/bin/hyperkube \
     --container-runtime=remote \
     --container-runtime-endpoint=/var/run/crio/crio.sock \
     --runtime-request-timeout=${KUBELET_RUNTIME_REQUEST_TIMEOUT} \
+{{- if .UseIPv6ForNodeIP }}
+    --node-ip=:: \
+{{- end}}
     --pod-manifest-path=/etc/kubernetes/manifests \
     --minimum-container-ttl-duration=6m0s \
     --cluster-domain=cluster.local \
--- a/pkg/asset/ignition/bootstrap/bootstrap.go
+++ b/pkg/asset/ignition/bootstrap/bootstrap.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
 	"os"
 	"path"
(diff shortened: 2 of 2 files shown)


jhixson74 (approved):

data/data/bootstrap/files/usr/local/bin/kubelet.sh.template
```diff
@@ -0,0 +1,22 @@
...
+    --node-ip=:: \
+{{- end}}
+    --pod-manifest-path=/etc/kubernetes/manifests \
+    --minimum-container-ttl-duration=6m0s \
+    --cluster-domain=cluster.local \
+    --cgroup-driver=systemd \
+    --serialize-image-pulls=false \
+    --v=2 \
+    --volume-plugin-dir=/etc/kubernetes/kubelet-plugins/volume/exec \
+    --pod-infra-container-image="${MACHINE_CONFIG_INFRA_IMAGE}"
```
hardys, line 22: I don't think adding this template is sufficient, since we're missing the systemd unit change from https://github.com/openshift/installer/pull/4742 to run this script

I guess that means we either need to backport that as well, or else add the `--node-ip` to the templated systemd unit file e.g https://github.com/openshift/installer/blob/release-4.7/data/data/bootstrap/systemd/units/kubelet.service.template

hardys:
I think I spotted an issue - also if you're reworking this @sadasu could you perhaps also ensure the commit message matches the original commit, then it's easier to compare branches e.g when using `git log` ?

sadasu:
Errors in e2e-metal-ipi-ovn-ipv6 don't seem to be due to this change. 

/test e2e-metal-ipi-ovn-ipv6

dhellmann:
@sadasu I'm looking at this one for today's patch queue. Are those CI failures relevant or are they flakes?

dhellmann:
[patch-manager] :rocket: Approved for z-stream by score: 1.50

picked

PR #5068, opened 2021-07-09
WIP: Kubevirt: Split terraform into stages

jhixson74:
datavolume: create data volume resources
masters: create permanent cluster resources
bootstrap: create bootstrap resources

This is needed for Terraform v0.14.6 and later since Terraform deletions now
destroy all resource dependencies.

https://issues.redhat.com/browse/CORS-1698


pkg/terraform/stages/kubevirt/stages.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"github.com/openshift/installer/pkg/terraform"
+	"github.com/openshift/installer/pkg/terraform/stages"
+)
+
+// PlatformStages are the stages to run to provision the infrastructure in
+// Kubevirt.
+var PlatformStages = []terraform.Stage{
+	stages.NewStage("kubevirt", "datavolume"),
+	stages.NewStage("kubevirt", "masters"),
```
jstuever, line 12: What is the reason for separating datavolume and masters into different stages? Can they not be combined into one "cluster" stage?

jhixson74: No reason other than  ignorance on my part. It was previously configured this way as a module so I separated it out this way. I don't see any reason why it can't be combined.

jhixson74: I have updated this.

staebler:
The kubevirt platform is being removed. See https://github.com/openshift/installer/pull/5301.

PR #5069, opened 2021-07-11
Bug 1992463: libvirt: bump default memory and cpus

karmab:
This commit bumps memory to 8gb for bootstrap vm and 16gb for masters, which seem minimal defaults for a successful install.
User can still override those values when building but this provide an alternative workflow when using public binaries (the openshift-baremetal one containing libvirt bindings)

### DIFF
--- a/data/data/libvirt/variables-libvirt.tf
+++ b/data/data/libvirt/variables-libvirt.tf
@@ -32,7 +32,7 @@ variable "libvirt_master_ips" {
 variable "libvirt_master_memory" {
   type        = string
   description = "RAM in MiB allocated to masters"
-  default     = "6144"
+  default     = "16384"
 }
 
 # At some point this one is likely to default to the number
@@ -41,13 +41,13 @@ variable "libvirt_master_memory" {
 variable "libvirt_master_vcpu" {
   type        = string
   description = "CPUs allocated to masters"
-  default     = "4"
+  default     = "2"
 }
 
 variable "libvirt_bootstrap_memory" {
   type        = number
   description = "RAM in MiB allocated to the bootstrap node"
-  default     = 4096
+  default     = 8192
 }
 
 # Currently RHCOS maintain its default 16G size if that


data/data/libvirt/bootstrap/main.tf
```diff
@@ -17,7 +17,7 @@ resource "libvirt_domain" "bootstrap" {
 
   memory = var.bootstrap_memory
 
-  vcpu = "2"
+  vcpu = "8"
```
praveenkumar, line 20: @karmab Do we need `8` vcpu for bootstrap ?

cfergeau: Same question for the vcpu changes :) 
I'd go with this commit:
```
commit ece7bf0a6b2281adf60e47345b1ccd119807fa1d (HEAD -> libvirt_bump)
Author: karmab <karimboumedhel@gmail.com>
Date:   Sun Jul 11 16:15:06 2021 +0200

    libvirt: bump default memory
    
    This commit bumps memory to 8gb for bootstrap vm and 16gb for masters,
    which seem minimal defaults for a successful install.
    User can still override those values when building but this provide an
    alternative workflow when using public binaries (the openshift-baremetal
    one containing libvirt bindings)

diff --git a/data/data/libvirt/variables-libvirt.tf b/data/data/libvirt/variables-libvirt.tf
index b555210f79..0b3961d2a6 100644
--- a/data/data/libvirt/variables-libvirt.tf
+++ b/data/data/libvirt/variables-libvirt.tf
@@ -32,7 +32,7 @@ variable "libvirt_master_ips" {
 variable "libvirt_master_memory" {
   type        = string
   description = "RAM in MiB allocated to masters"
-  default     = "6144"
+  default     = "16384"
 }
 
 # At some point this one is likely to default to the number
@@ -47,7 +47,7 @@ variable "libvirt_master_vcpu" {
 variable "libvirt_bootstrap_memory" {
   type        = number
   description = "RAM in MiB allocated to the bootstrap node"
-  default     = 4096
+  default     = 8192
 }
 
 # Currently RHCOS maintain its default 16G size if that
```

jstuever:
/title Bug 1992463: libvirt: bump default memory and cpus

PR #5071, opened 2021-07-12
openstack: Expose worker server group policy

pierreprinetti:
With this change, Compute nodes within each MachineSet are automatically
created in a Server group, with a default policy of
"soft-anti-affinity".

With this change, a "serverGroupPolicy" can be set in install-config, on
the worker MachinePool and/or in the platform default.

Implements [OSASINFRA-2570](https://issues.redhat.com/browse/OSASINFRA-2570)

Co-Authored-By: Matthew Booth <mbooth@redhat.com>

### DIFF
--- a/data/data/openstack/workers/variables.tf
+++ b/data/data/openstack/workers/variables.tf
@@ -1,0 +1,9 @@
+variable "server_group_names" {
+  type        = set(string)
+  description = "Names of the server groups for the worker nodes."
+}
+
+variable "server_group_policy" {
+  type        = string
+  description = "Policy of the server groups for the worker nodes."
+}
--- a/docs/user/openstack/README.md
+++ b/docs/user/openstack/README.md
@@ -69,7 +69,7 @@ For a successful installation it is required:
 - Security Group Rules: 60
 - Routers: 1
 - Subnets: 1
-- Server Groups: 1
+- Server Groups: 2, plus one per additional Availability zone in each machine-pool
 - RAM: 112 GB
 - vCPUs: 28
 - Volume Storage: 175 GB
@@ -91,11 +91,13 @@ Once you configure the quota for your project, please ensure that the user for t
 
 The default deployment stands up 3 master nodes, which is the minimum amount required for a cluster. For each master node you stand up, you will need 1 instance, and 1 port available in your quota. They should be assigned a flavor with at least 16 GB RAM, 4 vCPUs, and 25 GB Disk (or Root Volume). It is theoretically possible to run with a smaller flavor, but be aware that if it takes too long to stand up services, or certain essential services crash, the installer could time out, leading to a failed install.
 
-The Master Nodes are placed in a single Server Group with "soft anti-affinity" policy; the machines will therefore be creted on separate hosts when possible.
+The master nodes are placed in a single Server group with "soft anti-affinity" policy by default; the machines will therefore be created on separate hosts when possible.
 
 ### Worker Nodes
 
 The default deployment stands up 3 worker nodes. Worker nodes host the applications you run on OpenShift. The flavor assigned to the worker nodes should have at least 2 vCPUs, 8 GB RAM and 25 GB Disk (or Root Volume). However, if you are experiencing `Out Of Memory` issues, or your installs are timing out, try increasing the size of your flavor to match the master nodes: 4 vCPUs and 16 GB RAM.
+
+The worker nodes are placed in a single Server group with "soft anti-affinity" policy by default; the machines will therefore be created on separate hosts when possible.
 
 See the [OpenShift documentation](https://docs.openshift.com/container-platform/4.4/architecture/control-plane.html#defining-workers_control-plane) for more information on the worker nodes.
 
--- a/docs/user/openstack/customization.md
+++ b/docs/user/openstack/customization.md
@@ -40,7 +40,10 @@ Beyond the [platform-agnostic `install-config.yaml` properties](../customization
 
 * `additionalNetworkIDs` (optional list of strings): IDs of additional networks for machines.
 * `additionalSecurityGroupIDs` (optional list of strings): IDs of additional security groups for machines.
-* `serverGroupPolicy` (optional string): Server group policy to apply to the group that will contain the machines in the pool. Defaults to "soft-anti-affinity". Only applicable to the Control plane MachinePool.
+* `serverGroupPolicy` (optional string): Server group policy to apply to the group that will contain the machines in the pool. Defaults to "soft-anti-affinity". Allowed values are "affinity", "soft-affinity", "anti-affinity", "soft-anti-affinity".
+  * It is not possible to change a server group policy or a server's affiliation to a group after creation
+  * A strict "affinity" policy prevents migrations, and therefore affects OpenStack upgrades
+  * An additional OpenStack host is needed when migrating instances with a strict "anti-affinity" policy
 * `type` (optional string): The OpenStack flavor name for machines in the pool.
 * `rootVolume` (optional object): Defines the root volume for instances in the machine pool. The instances use ephemeral disks if not set.
   * `size` (required integer): Size of the root volume in GB. Must be set to at least 25.
--- a/pkg/asset/machines/openstack/machines.go
+++ b/pkg/asset/machines/openstack/machines.go
@@ -80,9 +80,6 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 		}
 
 		provider = providerConfigs[zone]
-		if role == "master" {
-			provider.ServerGroupName = clusterID + "-master"
-		}
 
 		machine := machineapi.Machine{
 			TypeMeta: metav1.TypeMeta{
@@ -147,6 +144,10 @@ func generateProvider(clusterID string, platform *openstack.Platform, mpool *ope
 		})
 	}
 
+	serverGroupName := clusterID + "-" + role
+	if az != "" {
+		serverGroupName += "-" + az
+	}
 	spec := openstackprovider.OpenstackProviderSpec{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: openstackprovider.SchemeGroupVersion.String(),
@@ -160,6 +161,7 @@ func generateProvider(clusterID string, platform *openstack.Platform, mpool *ope
 		PrimarySubnet:    platform.MachinesSubnet,
 		AvailabilityZone: az,
 		SecurityGroups:   securityGroups,
+		ServerGroupName:  serverGroupName,
 		Trunk:            trunkSupport,
 		Tags: []string{
 			fmt.Sprintf("openshiftClusterID=%s", clusterID),
--- a/pkg/tfvars/openstack/openstack.go
+++ b/pkg/tfvars/openstack/openstack.go
@@ -44,6 +44,8 @@ type config struct {
 	ExternalDNS                      []string                          `json:"openstack_external_dns,omitempty"`
 	MasterServerGroupName            string                            `json:"openstack_master_server_group_name,omitempty"`
 	MasterServerGroupPolicy          types_openstack.ServerGroupPolicy `json:"openstack_master_server_group_policy"`
+	WorkerServerGroupNames           []string                          `json:"openstack_worker_server_group_names,omitempty"`
+	WorkerServerGroupPolicy          types_openstack.ServerGroupPolicy `json:"openstack_worker_server_group_policy"`
 	AdditionalNetworkIDs             []string                          `json:"openstack_additional_network_ids,omitempty"`
 	AdditionalSecurityGroupIDs       []string                          `json:"openstack_master_extra_sg_ids,omitempty"`
 	MachinesSubnet                   string                            `json:"openstack_machines_subnet_id,omitempty"`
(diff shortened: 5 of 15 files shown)


docs/user/openstack/customization.md
```diff
@@ -39,7 +39,7 @@ Beyond the [platform-agnostic `install-config.yaml` properties](../customization
 
 * `additionalNetworkIDs` (optional list of strings): IDs of additional networks for machines.
 * `additionalSecurityGroupIDs` (optional list of strings): IDs of additional security groups for machines.
-* `serverGroupPolicy` (optional string): Server group policy to apply to the group that will contain the machines in the pool. Defaults to "soft-anti-affinity". Only applicable to the Control plane MachinePool.
+* `serverGroupPolicy` (optional string): Server group policy to apply to the group that will contain the machines in the pool. Defaults to "soft-anti-affinity". Note that a strict "affinity" policy prevents migrations, and therefore affects OpenStack upgrades. Allowed values are "affinity", "soft-affinity", "anti-affinity", "soft-anti-affinity".
```
SeanMooney, line 42: thanks for adding "Note that a strict "affinity" policy prevents migrations, and therefore affects OpenStack upgrades."

the other caveat that you may want to consider noting is that this cannot be changed after an instance is created so  if you want to change the member ship of server groups or the policy you will have to destroy the exstign workers and recretate the new ones with the new policy.

you can do that in a rolling fashion if you need too but the point being that the server group membership and policy for any worker instnace is fixed when its created for the lifetime of that worker.

pkg/tfvars/openstack/openstack.go
```diff
@@ -174,14 +176,29 @@ func TFVars(masterConfigs []*v1alpha1.OpenstackProviderSpec, cloud string, exter
 		return nil, errors.Errorf("ServerGroupID is not implemented in the Installer. Please use ServerGroupName for automatic creation of the Control Plane server group.")
 	}
 
+	for _, workerConfig := range workerConfigs {
+		cfg.WorkerServerGroupNames = append(cfg.WorkerServerGroupNames, workerConfig.ServerGroupName)
+		if workerConfig.ServerGroupID != "" {
+			return nil, errors.Errorf("ServerGroupID is not implemented in the Installer. Please use ServerGroupName for automatic creation of the Compute server group.")
```
SeanMooney, line 182: we generally recommend the use of the resouce UUIDs in external tools by the way.
names are intended to be human readable but uuids are preferred for external orchestrateors to uses.

you are allowed to have 2 server groups in the same tenant/project with the same name which is why the uuid is preferred.

pierreprinetti: Very true. However I could not find a way to use UUID here.

The problem is that the Installer creates the Machine and MachineSet manifests (the declarative definition of the servers to be created) _before_ it hits the OpenStack API: the UUIDs are not available at that stage because the resources do not exist yet.

pkg/asset/cluster/tfvars.go
```diff
@@ -510,17 +510,38 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
-			return err
+		var masterSpecs []*openstackprovider.OpenstackProviderSpec
+		{
+			masters, err := mastersAsset.Machines()
+			if err != nil {
+				return err
+			}
+
+			for _, master := range masters {
+				masterSpecs = append(masterSpecs, master.Spec.ProviderSpec.Value.Object.(*openstackprovider.OpenstackProviderSpec))
```
mdbooth, line 521: In general, are we ok with a crash in the event that ProviderSpec is unset, or of an unexpected type?

pkg/asset/cluster/tfvars.go
```diff
@@ -510,17 +510,38 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
+
+			for _, worker := range workers {
+				workerSpecs = append(workerSpecs, worker.Spec.Template.Spec.ProviderSpec.Value.Object.(*openstackprovider.OpenstackProviderSpec))
+			}
 		}
+
+		var workermpool *openstack.MachinePool
+		if len(installConfig.Config.Compute) > 0 {
+			workermpool = installConfig.Config.Compute[0].Platform.OpenStack
+		}
```
mdbooth, line 540: This doesn't look right: we're only going to consider a server group defined in the first compute config.

mdbooth (changes requested):

pkg/asset/cluster/tfvars.go
pierreprinetti: install-config's "Compute" slice only accepts one machinepool. I should add a comment.

patrickdillon:
Have you considered how changing the default behavior will affect existing workflows? Ideally, we don't want to break any workflows that previously worked. I know the RHV team ran into this for clusters running in the lab environment. Should also consider single-node clusters.

Just food for thought.

SeanMooney:
> Have you considered how changing the default behavior will affect existing workflows? Ideally, we don't want to break any workflows that previously worked. I know the RHV team ran into this for clusters running in the lab environment. Should also consider single-node clusters.
> 
> Just food for thought.

soft anti affinity is implemented as a weigher not a filter.
as such it never elimiates a host it just chagne the order in the weight list.

for single node deployment a list of one only has one order so the filter will have not impact other than doing a tiny be of extra work.

if you use the hard affinity/anti affinity polices (which we do not recommend) as a  default they are implemented as filters and it would be a problem but soft-anti-affinity should have no change in behavior for singel node.

this change should never result in a vm that could previously have booted form failing but it wil change the distibution of host to prefer spreading vms by default. the nova schedulers default behavior behaiovr is to spead anyway so you are just enhancing that perfernce with context of meber of the group.

so form an upgrade impact point of view this shoudl be minimal

if you uses any other policy as a default it would have a negitivy impact on exisitng deployments and workflows.
that is any policy beyond no policy.

pkg/tfvars/openstack/openstack.go
SeanMooney: ah yes the server group create api does not allow you to choose the uuid
https://docs.openstack.org/api-ref/compute/?expanded=create-server-group-detail#create-server-group

we often allow the uuid to be specified to avoid this catch 22 situation but i guess this is not one of those cases.

pierreprinetti:
> Have you considered how changing the default behavior will affect existing workflows

Thank you for raising this. I've been discussing offline with @mdbooth and @SeanMooney about precisely this: if changing the default from "no-server-group" to "soft-anti-affinity" could have any negative impact.

One minor impact is the doubling of the Server group quota requirement; we are tracking related work in [OS.AS.INFRA-2571](https://issues.redhat.com/browse/OSASINFRA-2571).

That being said, in the 3-compute-node setup, the answer seems to be reassuring. "soft-anti-affinity" is a best-effort policy designed to be compatible with the most critical workflows (e.g. OSP upgrades), at the risk of being ineffective in some (not-so-corner) cases.

That is the rationale with which [we have already defaulted to soft-anti-affinity for masters in v4.5](https://issues.redhat.com/browse/OSASINFRA-1204).

Affinity policies should also have no impact on single-node deployments, as to the best of my knowledge, they never apply to the _first_ server in a group.

PS: I see that @SeanMooney has addressed the point much better than I ever could; thank you again for that.

pkg/asset/cluster/tfvars.go
pierreprinetti: I believe that at this point, that property should be exactly of type `*openstackprovider.OpenstackProviderSpec`.

The alternative would be a controlled type assertion, but then what do we do in case it's not the type we expect? The answer heavily depends on how the user got to that state, which is information I don't know how to get.

docs/user/openstack/customization.md
pierreprinetti: Added a note mentioning that a server group is for life. Thanks!

pierreprinetti:
/hold
until we have QE ack for the feature

pierreprinetti:
@mdbooth @mandre 
I think I have addressed your comments. Can you please take a look and confirm that my responses and corrections are valid?

@patrickdillon 
Do you think that this change is acceptable in principle?

Even though we can't merge until QE acks for capacity, it would be optimal to have the approve/lgtm labels all lined up for merge, when the occasion presents itself.

pierreprinetti:
One more comment (courtesy @itzikb-redhat ):

Documentation should also mention that for OpenStack migrations, you will need an extra free host for migrating from a host when you have `anti-affinity` on servers

pierreprinetti:
Added a note about migrating `anti-affinity` instances (thanks @itzikb-redhat)
/hold cancel

pierreprinetti:
/hold
While this patch is ready for review, we are still waiting for a QE ack.

pkg/asset/cluster/tfvars.go
pierreprinetti: (added a comment)

pierreprinetti:
/assign @staebler

This is ready to merge once it gets approval and QE ack. The /hold is only in place to prevent accidental merging before QE confirms they have capacity to verify.

docs/user/openstack/README.md
```diff
@@ -89,12 +89,14 @@ Once you configure the quota for your project, please ensure that the user for t
 
 The default deployment stands up 3 master nodes, which is the minimum amount required for a cluster. For each master node you stand up, you will need 1 instance, and 1 port available in your quota. They should be assigned a flavor with at least 16 GB RAM, 4 vCPUs, and 25 GB Disk (or Root Volume). It is theoretically possible to run with a smaller flavor, but be aware that if it takes too long to stand up services, or certain essential services crash, the installer could time out, leading to a failed install.
 
-The Master Nodes are placed in a single Server Group with "soft anti-affinity" policy; the machines will therefore be creted on separate hosts when possible.
+The Master Nodes are placed in a single Server Group with "soft anti-affinity" policy by default; the machines will therefore be creted on separate hosts when possible.
```
EmilienM, line 92:
```suggestion
The Master Nodes are placed in a single Server Group with "soft anti-affinity" policy by default; the machines will therefore be created on separate hosts when possible.
```

docs/user/openstack/README.md
```diff
@@ -89,12 +89,14 @@ Once you configure the quota for your project, please ensure that the user for t
...
 
 ### Worker Nodes
 
 The default deployment stands up 3 worker nodes. Worker nodes host the applications you run on OpenShift. The flavor assigned to the worker nodes should have at least 2 vCPUs, 8 GB RAM and 25 GB Disk (or Root Volume). However, if you are experiencing `Out Of Memory` issues, or your installs are timing out, try increasing the size of your flavor to match the master nodes: 4 vCPUs and 16 GB RAM.
 
+The Master Nodes are placed in a single Server Group with "soft anti-affinity" policy by default; the machines will therefore be creted on separate hosts when possible.
```
EmilienM, line 98:
```suggestion
The Master Nodes are placed in a single Server Group with "soft anti-affinity" policy by default; the machines will therefore be created on separate hosts when possible.
```

pkg/asset/machines/openstack/machines.go
```diff
@@ -160,6 +157,7 @@ func generateProvider(clusterID string, platform *openstack.Platform, mpool *ope
 		PrimarySubnet:    platform.MachinesSubnet,
 		AvailabilityZone: az,
 		SecurityGroups:   securityGroups,
+		ServerGroupName:  clusterID + "-" + role + "-" + az,
```
EmilienM, line 160: are we sure we want `az` here? what if it's empty? do we end up with `cluster123-worker-` ?

pierreprinetti: Leveraged the power of neural networks, PTAL

pierreprinetti: it was ugly indeed.

pierreprinetti:
The QE ack has landed.
/hold cancel

CC @staebler

pkg/asset/cluster/tfvars.go
```diff
@@ -510,17 +510,47 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
+			if err != nil {
+				return err
+			}
+
+			for _, worker := range workers {
+				workerSpecs = append(workerSpecs, worker.Spec.Template.Spec.ProviderSpec.Value.Object.(*openstackprovider.OpenstackProviderSpec))
+			}
 		}
+
+		// Only considering the first Compute machinepool here, because
```
mandre, line 537: Let's try to not make assumptions that there will always only be one compute machinepool if possible, because that's going to bite us if the installer lifts the 1 worker machinepool restriction.

pierreprinetti: Dropping this assumption will require us to accept a whole set of new assumptions.

At this stage, we have one MachineSet per Availability zone, and one machine-pool. As a consequence, we have multiple Server group names and one single Server group policy.

If we have multiple machine-pools, there is no straightforward way to match names with policies. The code changes required to support multiple machine-pools are non-negligeable, and I can't justify that effort since there is no proposal to support that case.

I have added an assertion that there is no more than one worker machine-pool. The code will crash loudly at the first test run, and we will clearly see the required change in that case.

WDYT?

mandre: Fair enough. If it's too complex, it's not worth plan too much ahead and instead adding an obvious error message like you did is perfectly fine.

data/data/openstack/workers/variables.tf
```diff
@@ -0,0 +1,9 @@
+variable "worker_server_group_names" {
```
mandre, line 1: We're already in the worker module, we don't need to prefix the variables with `worker`

mandre (commented):
Just a nit, then I think we're good to go.

mandre (commented):
/lgtm

mandre (commented):
Restoring my vote after the rebase.
/lgtm
/approve

PR #5083, opened 2021-07-15
[WIP] Baremetal: CoreOS Live image based installs

kirankt:
This PR adds support for Baremetal IPI CoreOS Live image-based installation.


data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -4,19 +4,19 @@ set -ex
...
 RHCOS_BOOT_IMAGE_URL="{{.BootImage}}"
+RHCOS_LIVE_IMAGE_URL="{{.PlatformData.BareMetal.ClusterOSImage}}"
 
 # This DHCP range is used by dnsmasq to serve DHCP to the cluster. If empty
 # dnsmasq will only serve TFTP, and DHCP will be disabled.
 DHCP_RANGE="{{.PlatformData.BareMetal.ProvisioningDHCPRange}}"
 
 # Used by ironic to allow ssh to running IPA instances
-IRONIC_RAMDISK_SSH_KEY="{{.SSHKey}}"
+IRONIC_RAMDISK_SSH_KEY="$(echo -n "{{.SSHKey}}")"
```
hardys, line 19: Why was the echo needed here?

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -88,12 +88,6 @@ podman run -d --net host --privileged --name mariadb \
      -v $IRONIC_SHARED_VOLUME:/shared:z --entrypoint /bin/runmariadb \
      --env MARIADB_PASSWORD=$mariadb_password ${IRONIC_IMAGE}
 
-podman run -d --net host --privileged --name httpd \
-     --restart on-failure \
-     --env IRONIC_RAMDISK_SSH_KEY="$IRONIC_RAMDISK_SSH_KEY" \
-     --env PROVISIONING_INTERFACE=$PROVISIONING_NIC \
-     -v $IRONIC_SHARED_VOLUME:/shared:z --entrypoint /bin/runhttpd ${IRONIC_IMAGE}
-
 {{ if .PlatformData.BareMetal.ProvisioningIPv6 }}
```
hardys, line 91: I guess this is to align with the changes upstream and in https://github.com/openshift/ironic-image/pull/190 and remove httpd - it'd be nice if we could break that part out into another PR? /cc @elfosardo

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -115,12 +109,12 @@ podman run -d --net host --name coreos-downloader \
...
+     -v $IRONIC_SHARED_VOLUME:/shared:z ${COREOS_DOWNLOADER_IMAGE} /usr/local/bin/get-live-images.sh $RHCOS_BOOT_IMAGE_URL
 
-podman run -d --net host --name ipa-downloader \
-     --restart on-failure \
-     --env CACHEURL=${CACHEURL} \
-     -v $IRONIC_SHARED_VOLUME:/shared:z ${IPA_DOWNLOADER_IMAGE} /usr/local/bin/get-resource.sh
+#podman run -d --net host --name ipa-downloader \
+     #--restart on-failure \
+     #--env CACHEURL=${CACHEURL} \
+     #-v $IRONIC_SHARED_VOLUME:/shared:z ${IPA_DOWNLOADER_IMAGE} /usr/local/bin/get-resource.sh
```
hardys, line 117: I think we're agreed that all the ipa-downloader references can be removed here?

go.mod
```diff
@@ -38,7 +38,7 @@ require (
 	github.com/google/martian v2.1.1-0.20190517191504-25dcb96d9e51+incompatible // indirect
 	github.com/google/uuid v1.2.0
 	github.com/gophercloud/gophercloud v0.17.0
-	github.com/gophercloud/utils v0.0.0-20210323225332-7b186010c04f
+	github.com/gophercloud/utils v0.0.0-20210530213738-7c693d7efe47
```
hardys, line 41: I think the go.mod changes should be in the update-vendor commit?

pkg/tfvars/baremetal/baremetal.go
```diff
@@ -66,6 +67,7 @@ func TFVars(libvirtURI, apiVIP, imageCacheIP, bootstrapOSImage, externalBridge,
 		driverInfo := accessDetails.DriverInfo(credentials)
 		driverInfo["deploy_kernel"] = fmt.Sprintf("http://%s/images/ironic-python-agent.kernel", net.JoinHostPort(imageCacheIP, "80"))
 		driverInfo["deploy_ramdisk"] = fmt.Sprintf("http://%s/images/ironic-python-agent.initramfs", net.JoinHostPort(imageCacheIP, "80"))
```
hardys, line 69: Do we still need the IPA references here?

pkg/tfvars/baremetal/baremetal.go
```diff
@@ -115,41 +118,28 @@ func TFVars(libvirtURI, apiVIP, imageCacheIP, bootstrapOSImage, externalBridge,
...
 			instanceInfo["capabilities"] = "secure_boot:true"
 		}
 
+		// deploy_steps is set when a custom deployment is desired. We will use ironic's custom deployment
+		// interface to use live ISO based installer. Currently this value is static but may be configurable
+		// in the future.
+		var hostDeploySteps string
+		if strings.Contains(accessDetails.BootInterface(), "redfish-virtual-media") {
+			hostDeploySteps = `[{"interface": "deploy", "step": "install_coreos", "priority": 80, "args": {}}]`
+		}
```
hardys, line 135: We need this to work for PXE too, I guess this should be removed/modified?

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
dtantsur: Yeah, we don't need these any more.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -153,11 +147,34 @@ $IPTABLES -t raw -A DHCP_IRONIC -j DROP
...
-while ! curl --fail --head http://localhost/images/ironic-python-agent.initramfs ; do sleep 1; done
-while ! curl --fail --head http://localhost/images/ironic-python-agent.kernel ; do sleep 1; done
+
+# Embed agent ignition into the rhcos live iso
+sudo podman run -d --net host --privileged --name configure-coreos \
+     --restart on-failure \
+     --env IRONIC_RAMDISK_SSH_KEY="$IRONIC_RAMDISK_SSH_KEY" \
+     --env PROVISIONING_INTERFACE=$PROVISIONING_NIC \
+     --entrypoint /bin/configure-coreos-ipa \
+     -v $AUTH_DIR:/auth:ro \
```
dtantsur, line 159: Do we have TLS here? I think no, but if yes, we need to mount its volume here.

hardys:
Overall looks good, I can see a few parts are still WIP, and it'd be nice to split the vendor updates from the functional changes.

pkg/asset/rhcos/image.go
```diff
@@ -141,13 +141,32 @@ func osImage(config *types.InstallConfig) (string, error) {
...
-		// ignition support, which isn't enabled in the UPI BM images
-		if a, ok := streamArch.Artifacts["openstack"]; ok {
-			return rhcos.FindArtifactURL(a)
+		// Note that baremetal IPI currently uses the RHCOS live-iso image.
+		// We will need to pass 4 URLs one for virtual media and 3 for PXE
+		// booting. We will concatenate the URLs with a separator for easier extraction.
+		// Note: We cannot use rhcos.FindArtifactURL for the "metal" artifact.
+		iso := streamArch.Artifacts["metal"].Formats["iso"].Disk.Location
+		if iso == "" {
+			return "", fmt.Errorf("%s: No live-iso build found", st.FormatPrefix(archName))
```
dtantsur, line 150: Based on the today's conversation, let's delay making the ISO mandatory. It's actually not, we can leave happily with only the PXE artefacts.

pkg/tfvars/baremetal/baremetal.go
dtantsur: Yes. ISO is only supported for virtual media.

pkg/tfvars/baremetal/baremetal.go
```diff
@@ -66,6 +67,7 @@ func TFVars(libvirtURI, apiVIP, imageCacheIP, bootstrapOSImage, externalBridge,
 		driverInfo := accessDetails.DriverInfo(credentials)
 		driverInfo["deploy_kernel"] = fmt.Sprintf("http://%s/images/ironic-python-agent.kernel", net.JoinHostPort(imageCacheIP, "80"))
 		driverInfo["deploy_ramdisk"] = fmt.Sprintf("http://%s/images/ironic-python-agent.initramfs", net.JoinHostPort(imageCacheIP, "80"))
+		driverInfo["deploy_iso"] = fmt.Sprintf("http://%s/images/ironic-python-agent.iso", net.JoinHostPort(imageCacheIP, "80"))
```
dtantsur, line 70: We can do without ISO until network customization is needed. Won't harm either way.

pkg/tfvars/baremetal/baremetal.go
```diff
@@ -115,41 +118,28 @@ func TFVars(libvirtURI, apiVIP, imageCacheIP, bootstrapOSImage, externalBridge,
...
+		instanceInfo := make(map[string]interface{})
 		if host.BootMode == baremetal.UEFISecureBoot {
 			instanceInfo["capabilities"] = "secure_boot:true"
 		}
 
+		// deploy_steps is set when a custom deployment is desired. We will use ironic's custom deployment
+		// interface to use live ISO based installer. Currently this value is static but may be configurable
+		// in the future.
+		var hostDeploySteps string
+		if strings.Contains(accessDetails.BootInterface(), "redfish-virtual-media") {
```
dtantsur, line 133: Should be removed, has nothing to do with virtual media.

pkg/tfvars/baremetal/baremetal.go
kirankt: These are rhcos live images, so yes, we need them.

pkg/tfvars/baremetal/baremetal.go
kirankt: Right. You did clarify this for me a while ago. This is an old vestigial condition. I'll remove it.

pkg/tfvars/baremetal/baremetal.go
```diff
@@ -77,6 +79,7 @@ func TFVars(libvirtURI, apiVIP, imageCacheIP, bootstrapOSImage, externalBridge,
 			"power_interface":      accessDetails.PowerInterface(),
 			"raid_interface":       accessDetails.RAIDInterface(),
 			"vendor_interface":     accessDetails.VendorInterface(),
+			"deploy_interface":     "custom-agent",
```
dtantsur, line 82: This should only be set if CustomDeploy is set.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
kirankt: I saw instances where there was an added newline at the end of the key, so I had to add the echo.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
kirankt: Correct. httpd is moved later in the script because within it entrypoint script contains code to configure the Ironic kernel params. Please see:
https://github.com/openshift/ironic-image/blob/4300542b663ec3dfcce8d4c452f07da00f19c7af/scripts/runhttpd#L28

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
kirankt: Sure. I'll remove it. I was testing it with it commented out.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
kirankt: I don't think we do TLS in the bootstrap node's ironic.

go.mod
kirankt: Yep. Sorry got my commits mixed up. Will fix it in the next push

pkg/asset/rhcos/image.go
kirankt: +1. Fixed in the next commit.

pkg/tfvars/baremetal/baremetal.go
kirankt: So, unless there is an installer interface to toggle the CustomDeploy value, we wont' know at this point whether we're using a custom-agent or something else. Also, won't this clash with the deploy_steps setting in the file?

pkg/tfvars/baremetal/baremetal.go
kirankt: Agreed, Steve. Going to make this unconditional

kirankt:
Lets continue to discuss here: https://github.com/openshift/installer/pull/5087

hardys:
> Lets continue to discuss here: #5087

Sorry closing this was a mistake, now reopened!

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
kirankt: I'm going to move this back to the original location. The kernel params don't depend on this container. i was mistaken.

pkg/tfvars/baremetal/baremetal.go
dtantsur: Okay, so I'm not sure if we need to support the old approach for new installations (openshift-installer is not used during upgrades, is it?). But it could be useful to have a switch, so that we can switch it off in case of regressions. This can be done in a follow-up.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
dtantsur: I wonder if Go templates have a function similar to Jinja's `| trim` filter.

pkg/tfvars/baremetal/baremetal.go
kirankt: I think we should add the switch as a follow up as a BZ, if we see regressions. The changes might cascade down the stack to CAPBM, CBO, etc.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
hardys: Ok, it seems unrelated to the rest of this change so I wonder if we should handle this in a separate bugfix PR (ideally with more details about exactly when/why this can be a problem)

pkg/tfvars/baremetal/baremetal.go
hardys: Yes the installer is only used for initial deployment, not during upgrades, so IMO we should just switch to the new approach (provided we're confident with that)

Separately there is the issue of what happens on upgrade (where existing hosts and provisioning CR will exist referencing the qcow image etc), but that's outside the scope of any installer changes.

hardys: Ok I thought of a problem after discussing this with @sadasu - some customers are using a customized qcow2 image as a workaround for missing day-1 networking support (e.g bond+vlans for the controlplane) - that essentially means we can't remove the qcow2 support until we have an alternative API for that configuration (or perhaps a way to achieve the same customization in this new flow, but I'd prefer to avoid another interim solution)

That probably means we should defer this until day-1 networking (which depends on this) is done, thoughts @kirankt @dtantsur @cybertron ?

kirankt: @hardys I agree. I think making changes this late and testing would take more time than 3 days. I am OK with closing this in favor of a more comprehensive work that supports qcow2 and live images.

kirankt:
/hold

Need to revisit and figure out upgrades and keeping the existing qcow2 support in addition to the live-iso based installs.

sadasu:
Can this PR be reworked to just add support for CoreOS live images and not remove support for qcow2 images? We will need to support qcow2 images for atleast one more release (maybe more).

kirankt:
> Can this PR be reworked to just add support for CoreOS live images and not remove support for qcow2 images? We will need to support qcow2 images for atleast one more release (maybe more).

Yes. We can do it. But it will take time to test both scenarios after adding the feature.

PR #5088, opened 2021-07-18
Add SRIOV Support for Kubevirt Provider

orenc1:
Currently, the kubevirt VirtualMachines that are created during the cluster provisioning, are hard-coded with Bridge interface binding, despite the actual type of the network that was configured in `install-config.yaml`.
This PR adds an optional new field to install-config.yaml, specifying the interface binding method to be used when creating the kubevirt VirtualMachines.
Example:
```
platform:
  kubevirt:
    apiVIP: <api_vip>
    ingressVIP: <ingress_vip>
    namespace: kubevirt-nested-cluster
    networkName: <NAD name>
    InterfaceBindingMethod: SRIOV          # <------- New Field
    storageClass: <storage class used>
```

This PR is strictly related to PR in openshift/cluster-api-provider-kubevirt - https://github.com/openshift/cluster-api-provider-kubevirt/pull/38, as it's changing the API for `KubevirtMachineProviderSpec` which is consumed in this PR.

Signed-off-by: orenc1 <ocohen@redhat.com>

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -1684,6 +1684,10 @@ spec:
                       ingress controller.
                     format: ip
                     type: string
+                  interfaceBindingMethod:
+                    description: InterfaceBindingMethod is the the interface binding
+                      method of the nodes of the tenantcluster (Bridge | SRIOV).
+                    type: string
                   namespace:
                     description: Namespace is the namespace in the infra cluster,
                       which the control plane (master vms) and the compute (worker
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -674,17 +674,12 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		}
 
 		labels := kubevirtutils.BuildLabels(clusterID.InfraID)
-		interfaceBindingMethod := "InterfaceBridge"
-		if installConfig.Config.Kubevirt.InterfaceBindingMethod != "" {
-			interfaceBindingMethod = "Interface" + installConfig.Config.Kubevirt.InterfaceBindingMethod
-		}
 		data, err := kubevirttfvars.TFVars(
 			kubevirttfvars.TFVarsSources{
-				MasterSpecs:            masterSpecs,
-				ImageURL:               string(*rhcosImage),
-				Namespace:              installConfig.Config.Kubevirt.Namespace,
-				InterfaceBindingMethod: interfaceBindingMethod,
-				ResourcesLabels:        labels,
+				MasterSpecs:     masterSpecs,
+				ImageURL:        string(*rhcosImage),
+				Namespace:       installConfig.Config.Kubevirt.Namespace,
+				ResourcesLabels: labels,
 			},
 		)
 		if err != nil {
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -674,7 +674,10 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		}
 
 		labels := kubevirtutils.BuildLabels(clusterID.InfraID)
-		interfaceBindingMethod := "Interface" + installConfig.Config.Kubevirt.InterfaceBindingMethod
+		interfaceBindingMethod := "InterfaceBridge"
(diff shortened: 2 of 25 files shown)


data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -1706,6 +1710,7 @@ spec:
                 required:
                 - apiVIP
                 - ingressVIP
+                - interfaceBindingMethod
```
ravidbro, line 1713: I'm not sure we want it to be required, such change will break existing scripts and tests

orenc1: Right.
The intention was to make it optional, with default setting for BridgeInterface, for backward compatibility.
I've updated the code.

ravidbro:
I don't see where you set the default value, 
also, did you check that this option can be applied by the terraform kubevirt provider?

orenc1:
> I don't see where you set the default value,

I added a default value for the `kubevirt_interface_binding_method` variable in pkg/asset/store/data/kubevirt/variables-kubevirt.tf`
> also, did you check that this option can be applied by the terraform kubevirt provider?

Yes, I checked with @nirarg and the InterfaceSRIOV is recognized in the terraform kubevirt provider:
https://github.com/kubevirt/terraform-provider-kubevirt/blob/5c4934a8bda862edfbcc99945a2ed4ea3cc33359/kubevirt/schema/virtualmachineinstance/domain_spec.go#L115

pkg/tfvars/kubevirt/kubevirt.go
```diff
@@ -16,23 +15,25 @@ type config struct {
...
 // TFVarsSources contains the parameters to be converted into Terraform variables
 type TFVarsSources struct {
-	MasterSpecs     []*v1.KubevirtMachineProviderSpec
-	ImageURL        string
-	Namespace       string
-	ResourcesLabels map[string]string
+	MasterSpecs            []*v1.KubevirtMachineProviderSpec
+	ImageURL               string
+	Namespace              string
+	InterfaceBindingMethod string
```
patrickdillon, line 28: Can we pull the value from `MasterSpecs.InterfaceBindingMethod` instead of adding a new field?

pkg/asset/cluster/tfvars.go
```diff
@@ -673,12 +673,17 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		}
 
 		labels := kubevirtutils.BuildLabels(clusterID.InfraID)
+		interfaceBindingMethod := "InterfaceBridge"
+		if installConfig.Config.Kubevirt.InterfaceBindingMethod != "" {
+			interfaceBindingMethod = "Interface" + installConfig.Config.Kubevirt.InterfaceBindingMethod
+		}
```
patrickdillon, line 679: Should this value be set here, or in the machineproviderspec (machines.go) and then read here? There is a framework for setting default values in the machine providers. machines.go. It seems like in this current code there would be a mismatch between your masters created by terraform and the machine provider spec which doesn't seem correct.

patrickdillon (commented):
Generally this looks good, except you are setting the variables directly in the TF vars rather than in the machine provider spec and then using the machine provider spec to set the tf vars, which seems like it would result in differences between the created machines and those specified in the machine provider spec.

pkg/tfvars/kubevirt/kubevirt.go
orenc1: fixed, removed the new field from `TFVarsSources`

pkg/asset/cluster/tfvars.go
orenc1: Fixed, move that logic to `pkg/asset/machines/kubevirt/machines.go`.
Had to change the signature of `provider` function to pass the installConfig.

orenc1:
kubevirt lane seems to become flaky. retrying.
/test e2e-kubevirt

pkg/tfvars/kubevirt/kubevirt.go
```diff
@@ -3,7 +3,6 @@ package kubevirt
 
 import (
 	"encoding/json"
-
 	v1 "github.com/openshift/cluster-api-provider-kubevirt/pkg/apis/kubevirtprovider/v1alpha1"
```
patrickdillon, line 6: nit: this line should remain.

orenc1: Re-added the space between the built-in and kubevirt imported packages.

PR #5090, opened 2021-07-19
Add documentation on setting cloud provider opts

dulek:
In 4.9 we introduce support for LoadBalancer services. This means that
user might need to tweak the cloud provider options to match their
OpenStack cluster configuration. This commit adds documentation on how
to do it before and after the installation.

### DIFF
--- a/docs/user/openstack/README.md
+++ b/docs/user/openstack/README.md
@@ -39,7 +39,9 @@ In addition, it covers the installation with the default CNI (OpenShiftSDN), as 
       - [Using a Server Group](#using-a-server-group)
       - [Setting Nova Availability Zones](#setting-nova-availability-zones)
     - [Using a Custom External Load Balancer](#using-a-custom-external-load-balancer)
-    - [Refreshing a CA Certificate](#refreshing-a-ca-certificate)
+    - [Reconfiguring cloud provider](#reconfiguring-cloud-provider)
+      - [Modifying cloud provider options](#modifying-cloud-provider-options)
+      - [Refreshing a CA Certificate](#refreshing-a-ca-certificate)
   - [Reporting Issues](#reporting-issues)
 
 ## Reference Documents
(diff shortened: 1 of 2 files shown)


docs/user/openstack/README.md
```diff
@@ -646,14 +648,26 @@ set-cookie: 1e2670d92730b515ce3a1bb65da45062=9b714eb87e93cf34853e87a92d6894be; p
...
 oc edit configmap -n openshift-config cloud-provider-config
 ```
 
+There are several things you can change:
+
+#### Modifying cloud provider options
+
+If you need to modify the direct cloud provider options, then edit the `config` key in the ConfigMap. A brief list of possible options is shown in [Cloud Provider configuration](./customization.md#cloud-provider-configuration) section.
+
+Please note that it can take a while to reconfigure the cluster depending on the size of it. The reconfiguration is completed once no node is getting `SchedulingDisabled` taint anymore.
```
mandre, line 665: Should we move the note outside of the subsection,  on line 658, since it applies to bot modifying cloud provider options and refreshing the CA certificate?

docs/user/openstack/customization.md
```diff
@@ -234,6 +235,32 @@ controlPlane:
...
 
+## Cloud Provider configuration
+
+You may want to modify cloud provider configuration in order to make it work with your OpenStack cloud. This is possible if you'll let the installer generate the manifests before running the installation:
+
+```sh
+openshift-install --dir <directory> create manifests
+```
+
+Then modify the manifest containing the cloud provider configuration:
```
mandre, line 246: Should we perhaps point to https://v1-18.docs.kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#cloud-conf for all available cloud provider options?

docs/user/openstack/customization.md
```diff
@@ -234,6 +235,32 @@ controlPlane:
...
+
+```sh
+vi <directory>/manifests/cloud-provider-config.yaml
+```
+
+In order to tweak support for LoadBalancer Services you can modify options regarding Octavia configuration in the `[LoadBalancer]` section of `config` key. In particular:
```
mandre, line 252: It's not clear if this is an example or rather the only config options we allow setting.

docs/user/openstack/customization.md
dulek: Depends on if we want to advertise options that we don't support or would simply not work because legacy cloud provider is somehow broken with newer Octavia?

docs/user/openstack/customization.md
dulek: Those are the only options I personally tested and I'm sure they work.

docs/user/openstack/README.md
dulek: Your call, I only noticed that it involves the same file, but I'm okay with reverting to the previous version.

mandre: What was the previous version? I'm asking to move line 665 a few lines above so that the note applies to both `modifying cloud provider options` and `refreshing the CA certificate` subsections.

docs/user/openstack/customization.md
mandre: I still believe we should point to the cloud provider cloud.conf page. We may chose to be more prescriptive in the product doc, downstream.

docs/user/openstack/README.md
dulek: Ah, okay, I misunderstood!

docs/user/openstack/customization.md
dulek: Ok.

docs/user/openstack/customization.md
dulek: Ok.

mandre (commented):
/lgtm :+1:

PR #5092, opened 2021-07-20
Destroy AWS EFS volumes

jsafrane:
Destroy AWS EFS volumes that have owner tag when destroying a cluster. Recursively destroy all access points and mount targets used by the volume, even when they are not tagged, otherwise EFS cannot be deleted. Not sure what's common practice for recursive deletion on `cluster destroy` - access points could be tagged (and thus we may require users to tag them), but mount targets do not have any tags.

AWS EFS CSI driver is an optional OCP component (installed via OLM). Right now, an EFS must be created manually by user. Since VPC and subnets cannot be deleted if they are used by an EFS volume, allow user to tag the EFS volume so cluster destruction deletes EFS, VPC + subnets (with two iterations to get the ordering right...)

KEP for whole AWS EFS support: https://github.com/openshift/enhancements/pull/687

### DIFF
--- a/pkg/destroy/aws/aws.go
+++ b/pkg/destroy/aws/aws.go
@@ -2073,6 +2073,9 @@ func deleteFileSystem(ctx context.Context, client *efs.EFS, fsid string, logger 
 
 	_, err = client.DeleteFileSystemWithContext(ctx, &efs.DeleteFileSystemInput{FileSystemId: aws.String(fsid)})
 	if err != nil {
+		if err.(awserr.Error).Code() == efs.ErrCodeFileSystemNotFound {
+			return nil
+		}
 		return err
 	}
 
@@ -2130,6 +2133,10 @@ func deleteAccessPoint(ctx context.Context, client *efs.EFS, id string, logger l
 	logger = logger.WithField("AccessPoint ID", id)
 	_, err := client.DeleteAccessPointWithContext(ctx, &efs.DeleteAccessPointInput{AccessPointId: aws.String(id)})
 	if err != nil {
+		if err.(awserr.Error).Code() == efs.ErrCodeAccessPointNotFound {
+			return nil
+		}
+
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/destroy/aws/aws.go
```diff
@@ -2026,3 +2029,121 @@ func isBucketNotFound(err interface{}) bool {
...
+	for _, ap := range accessPointIDs {
+		err := deleteAccessPoint(ctx, client, ap, logger)
+		if err != nil {
+			return err
+		}
+	}
+
+	_, err = client.DeleteFileSystemWithContext(ctx, &efs.DeleteFileSystemInput{FileSystemId: aws.String(fsid)})
+	if err != nil {
+		return err
```
jstuever, line 2076: Is there a case here where the error is that the file system doesn't exist (as it may have been previously deleted)? If you look elsewhere in the code, we handle this differently from other errors.

pkg/destroy/aws/aws.go
```diff
@@ -2026,3 +2029,121 @@ func isBucketNotFound(err interface{}) bool {
...
+	}
+
+	return mountTargetIDs, nil
+}
+
+func deleteAccessPoint(ctx context.Context, client *efs.EFS, id string, logger logrus.FieldLogger) error {
+	logger = logger.WithField("AccessPoint ID", id)
+	_, err := client.DeleteAccessPointWithContext(ctx, &efs.DeleteAccessPointInput{AccessPointId: aws.String(id)})
+	if err != nil {
+		return err
```
jstuever, line 2133: Same here, case where the error is that it does not exist (because it was already deleted)?

pkg/destroy/aws/aws.go
```diff
@@ -2026,3 +2029,121 @@ func isBucketNotFound(err interface{}) bool {
...
+
+	logger.Info("Deleted")
+	return nil
+}
+
+func deleteMountTarget(ctx context.Context, client *efs.EFS, id string, logger logrus.FieldLogger) error {
+	logger = logger.WithField("Mount Target ID", id)
+	_, err := client.DeleteMountTargetWithContext(ctx, &efs.DeleteMountTargetInput{MountTargetId: aws.String(id)})
+	if err != nil {
+		return err
```
jstuever, line 2144: Same here, case where the error is that it does not exist (because it was already deleted)?

jstuever (commented):
A few questions regarding how errors are handled when resources were previously deleted. It's important that we handle these properly as the destroy is ran in a loop and some resources are deleted between loops, etc.

pkg/destroy/aws/aws.go
jsafrane: Fixed everywhere

jsafrane:
I added check for various NotFound errors and tested in a real cluster (with some hacks around to use invalid IDs)

PR #5093, opened 2021-07-20
Bug 2000352: vSphere: Add validation for vCenter and ESXi versions

jcpowermac:
The out of tree csi driver requires HW15
and vCenter 6.7U3 or greater. This PR
adds warnings if that isn't met.

vCenter and ESXi version validation
- [x] Add check for vCenter build number that equates to 6.7u3
- [x] Add check for ESXi version and build number

Unit tests
- [x] Validation now requires access to vCenter. During unit tests use the govmomi simulator.
- [x] Hostname validation failure, determine workaround, is #4254 required? Case should not matter.
- [x] Better documentation


jcpowermac:
CI is unhappy
/retest

jsafrane:
Since we don't install the CSI driver by default in 4.9, it does not make much sense to warn about the HW version during 4.9 install time. Only few clusters will have TechPreviewNoUpgrade enabled.

It will become more useful in 4.10, where the driver will be installed by default. I already changed corresponding story to 4.10 epic. https://issues.redhat.com/browse/SPLAT-238

jcpowermac:
> Since we don't install the CSI driver by default in 4.9, it does not make much sense to warn about the HW version during 4.9 install time. Only few clusters will have TechPreviewNoUpgrade enabled.
> 
> It will become more useful in 4.10, where the driver will be installed by default. I already changed corresponding story to 4.10 epic. https://issues.redhat.com/browse/SPLAT-238

Thanks @jsafrane still trying to get unit tests sorted - at least that gives me some extra time.

pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -66,3 +81,78 @@ func folderExists(ic *types.InstallConfig, fldPath *field.Path) field.ErrorList
...
+	if err != nil {
+		return err
+	}
+
+	vCenterVersion, err := version.NewVersion(vim25.ServiceContent.About.Version)
+	if err != nil {
+		return err
+	}
+
+	if constraints.Check(vCenterVersion) {
```
jcpowermac, line 98: TODO: The version is checked but the exact build of U3 is not. Add that check as well so that we don't notify when a vSphere environment is up-to-date.

jcpowermac:
I will rebase after reviews

pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -66,3 +82,99 @@ func folderExists(ic *types.InstallConfig, fldPath *field.Path) field.ErrorList
...
+	hosts, err := clusters[0].Hosts(context.TODO())
+	if err != nil {
+		return errors.Wrapf(err, "unable to find hosts from cluster on path: %s", clusterPath)
+	}
+
+	for _, h := range hosts {
+		var mh mo.HostSystem
+		h.Properties(context.TODO(), h.Reference(), []string{"config.product"}, &mh)
+
+		v1, err := version.NewVersion(mh.Config.Product.Version)
```
jcpowermac, line 150: Why did I call this `v1` that makes no sense, change

pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -14,12 +21,21 @@ import (
...
 	}
 
+	client, _, err := vspheretypes.CreateVSphereClients(context.TODO(), ic.VSphere.VCenter, ic.VSphere.Username, ic.VSphere.Password, certificates...)
+	if err != nil {
+		return errors.Wrap(err, "unable to connect to vCenter API")
+	}
+
+	validatevCenterVersion(client)
+
+	validateESXiVersion(client, ic)
```
patrickdillon, line 37: it looks like the errors returned by these functions are being ignored

pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -28,7 +44,7 @@ func Validate(ic *types.InstallConfig) error {
 // ValidateForProvisioning performs platform validation specifically for installer-
 // provisioned infrastructure. In this case, self-hosted networking is a requirement
 // when the installer creates infrastructure for vSphere clusters.
-func ValidateForProvisioning(ic *types.InstallConfig) error {
+func ValidateForProvisioning(ic *types.InstallConfig, certificates ...x509.Certificate) error {
```
patrickdillon, line 47: it looks like this new `certificates` parameter is not being used, or maybe I'm just missing it

pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -66,3 +82,99 @@ func folderExists(ic *types.InstallConfig, fldPath *field.Path) field.ErrorList
...
+		logrus.Warnf("The out-of-tree storage driver requires virtual hardware 15 and vSphere 6.7 U3. Current vCenter version: %s, build: %s",
+			vim25.ServiceContent.About.Version, vim25.ServiceContent.About.Build)
+
+	} else {
+		// This is the vCenter 6.7 U3 build number
+		// Anything less than this version is unsupported with the
+		// out-of-tree CSI.
+		// https://kb.vmware.com/s/article/2143838
+		// https://vsphere-csi-driver.sigs.k8s.io/supported_features_matrix.html
+		if build < 14368073 {
```
patrickdillon, line 112: similar to the comment from the other PR, would be good to have this variable as a constant.

patrickdillon (commented):
I did an initial first pass, by and large it looks good.

pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -66,3 +82,99 @@ func folderExists(ic *types.InstallConfig, fldPath *field.Path) field.ErrorList
...
+				build, err := strconv.Atoi(mh.Config.Product.Build)
+				if err != nil {
+					return err
+				}
+				// This is the ESXi 6.7 U3 build number
+				// Anything less than this version is unsupported with the
+				// out-of-tree CSI.
+				// https://kb.vmware.com/s/article/2143838
+				// https://vsphere-csi-driver.sigs.k8s.io/supported_features_matrix.html
+				if build < 14320388 {
```
gnufied, line 168: I was thinking some more about it and can we use apiVersion rather than build number? It looks like 6.7u3 sets API version to be `6.7.3` and I expect all earlier versions to have lower numbers than that. 

Screenshot from 2021-09-01 13-02-18

jcpowermac: I would rather use what I have I the PR vs the apiversion in which I have no idea what VMware could put there.

pkg/asset/installconfig/vsphere/validation.go
gnufied: if possible I would like us to be consistent with how upstream reports similar warnings for using older versions of vcenter - https://github.com/kubernetes/kubernetes/pull/98546/files#diff-494c25f28f8cd90f6fe6fa4a32a6f59e457c34bc6bbdb992d4c75d7f6138e683R205

We have to carry similar code in vsphere-problem-detector and we should be consistent everywhere in checks we perform.

pkg/asset/installconfig/vsphere/validation.go
rvanderp3: That would be my concern as well @jcpowermac .  It appears according to the schema that it __should__ be in format x.y.z https://code.vmware.com/apis/968/vsphere .  I think the build number provides a similar level of accuracy though IMO.

gnufied: I was just talking with @jcpowermac offline and we will need similar logic in vsphere-problem-detector. I just opened - https://github.com/openshift/vsphere-problem-detector/pull/48 for reporting apiVersion string.

gnufied: It should be noted that 6.5 on the other hand is setting different apiVersion. It does not seem proper but for our purposes it works. 
Screenshot from 2021-09-01 15-13-49

jcpowermac: @gnufied discussed back and forth on slack in regards to this. If VMware actually followed semver for apiVersion it would have been way easier to use that field for this check. Unfortunately they are unreliable (6.5 is "6.5", 7U2 is "7.0.2.0"). Just moving forward with version and build.

pkg/asset/installconfig/vsphere/validation.go
jcpowermac: @patrickdillon for the new validation methods to pass unit testing we need a vSphere API - this is provided by the simulator. The next problem is the simulator when started with SSL is a invalid self-signed certificate. The installer requires this certificate to be valid. The way I worked around this was to create a CA and certificate, pass the CA to `CreateVSphereClients`. (maybe there is a better way, all ears ;-) )

jcpowermac:
/hold
troubleshooting unit test failures
```
--- FAIL: TestValidate (5.64s)
    --- FAIL: TestValidate/valid_UPI_install_config (0.17s)
        validation_test.go:174: 
            	Error Trace:	validation_test.go:174
            	Error:      	Received unexpected error:
            	            	platform.vsphere: Internal error: unable to find cluster on path: /DC0/host/: NotAuthenticated
            	Test:       	TestValidate/valid_UPI_install_config
FAIL
```

jcpowermac:
e2e-vsphere job failure do to individual test failure unrelated to this PR.

PR #5094, opened 2021-07-20
vSphere: Split terraform into stages

e-tienne:
pre-bootstrap: prerequisites for next stages - template
bootstrap: create bootstrap resources
masters: create permanent cluster resources

https://issues.redhat.com/browse/CORS-1696


e-tienne:
> /lgtm This looks good to me. Let's have Patrick and Jeremiah have the final verdict.

Thanks for reviewing John!

I'm also looking for feedback regarding the possible issue with `bootstrap_ip` and `control_plane_ips` output values timeout (assigned from the vm's `default_ip_address`) even if network waiters are configured. 

P.S.: I'm aware of root module variable declarations warnings that we should we get rid of.

data/data/vsphere/master/main.tf
```diff
@@ -1,31 +1,47 @@
...
-resource "vsphere_virtual_machine" "vm" {
-  count = var.instance_count
+provider "vsphere" {
+  user                 = var.vsphere_username
+  password             = var.vsphere_password
+  vsphere_server       = var.vsphere_url
+  allow_unverified_ssl = false
+}
+
+provider "vsphereprivate" {
```
jcpowermac, line 14: If the `vsphereprivate_import_ova` resource is not being used then this isn't required
https://github.com/openshift/installer/blob/master/data/data/vsphere/main.tf#L44-L53

data/data/vsphere/bootstrap/main.tf
```diff
@@ -2,7 +2,21 @@ locals {
...
 
-resource "vsphere_virtual_machine" "vm" {
+provider "vsphere" {
+  user                 = var.vsphere_username
+  password             = var.vsphere_password
+  vsphere_server       = var.vsphere_url
+  allow_unverified_ssl = false
+}
+
+provider "vsphereprivate" {
```
jcpowermac, line 12: Same as comment below

jcpowermac:
In [.openshift_install.log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_installer/5094/pull-ci-openshift-installer-master-e2e-vsphere/1417530852111290368/artifacts/e2e-vsphere/ipi-install-install/artifacts/.openshift_install.log)

```
time="2021-07-20T17:49:04Z" level=debug msg="Warning: Value for undeclared variable"
time="2021-07-20T17:49:04Z" level=debug
time="2021-07-20T17:49:04Z" level=debug msg="The root module does not declare a variable named \"master_cores_per_socket\""
time="2021-07-20T17:49:04Z" level=debug msg="but a value was found in file"
time="2021-07-20T17:49:04Z" level=debug msg="\"/tmp/openshift-install-bootstrap-803409321/pre-bootstrap.tfvars.json\". To use"
time="2021-07-20T17:49:04Z" level=debug msg="this value, add a \"variable\" block to the configuration."
time="2021-07-20T17:49:04Z" level=debug
time="2021-07-20T17:49:04Z" level=debug msg="Using a variables file to set an undeclared variable is deprecated and will"
time="2021-07-20T17:49:04Z" level=debug msg="become an error in a future release. If you wish to provide certain \"global\""
time="2021-07-20T17:49:04Z" level=debug msg="settings to all configurations in your organization, use TF_VAR_..."
time="2021-07-20T17:49:04Z" level=debug msg="environment variables to set these instead."
time="2021-07-20T17:49:04Z" level=debug
time="2021-07-20T17:49:04Z" level=debug
time="2021-07-20T17:49:04Z" level=debug msg="Warning: Value for undeclared variable"
time="2021-07-20T17:49:04Z" level=debug
time="2021-07-20T17:49:04Z" level=debug msg="The root module does not declare a variable named \"master_disk_size\" but a"
time="2021-07-20T17:49:04Z" level=debug msg="value was found in file"
time="2021-07-20T17:49:04Z" level=debug msg="\"/tmp/openshift-install-bootstrap-803409321/pre-bootstrap.tfvars.json\". To use"
time="2021-07-20T17:49:04Z" level=debug msg="this value, add a \"variable\" block to the configuration."
time="2021-07-20T17:49:04Z" level=debug
time="2021-07-20T17:49:04Z" level=debug msg="Using a variables file to set an undeclared variable is deprecated and will"
time="2021-07-20T17:49:04Z" level=debug msg="become an error in a future release. If you wish to provide certain \"global\""
time="2021-07-20T17:49:04Z" level=debug msg="settings to all configurations in your organization, use TF_VAR_..."
time="2021-07-20T17:49:04Z" level=debug msg="environment variables to set these instead."
time="2021-07-20T17:49:04Z" level=debug
time="2021-07-20T17:49:04Z" level=debug
time="2021-07-20T17:49:04Z" level=debug msg="Warning: Value for undeclared variable"
time="2021-07-20T17:49:04Z" level=debug
time="2021-07-20T17:49:04Z" level=debug msg="The root module does not declare a variable named \"master_memory\" but a value"
time="2021-07-20T17:49:04Z" level=debug msg="was found in file"
time="2021-07-20T17:49:04Z" level=debug msg="\"/tmp/openshift-install-bootstrap-803409321/pre-bootstrap.tfvars.json\". To use"
time="2021-07-20T17:49:04Z" level=debug msg="this value, add a \"variable\" block to the configuration."
time="2021-07-20T17:49:04Z" level=debug
time="2021-07-20T17:49:04Z" level=debug msg="Using a variables file to set an undeclared variable is deprecated and will"
time="2021-07-20T17:49:04Z" level=debug msg="become an error in a future release. If you wish to provide certain \"global\""
time="2021-07-20T17:49:04Z" level=debug msg="settings to all configurations in your organization, use TF_VAR_..."
time="2021-07-20T17:49:04Z" level=debug msg="environment variables to set these instead."
time="2021-07-20T17:49:04Z" level=debug
time="2021-07-20T17:49:04Z" level=debug
time="2021-07-20T17:49:04Z" level=debug msg="Warning: Values for undeclared variables"
```

data/data/vsphere/master/main.tf
```diff
@@ -1,31 +1,47 @@
 locals {
+  // limitation of baremetal-runtimecfg.  The hostname must be master
```
jstuever, line 2: Is this correct? `baremetal-runtimecfg`

pkg/terraform/stages/vsphere/stages.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"github.com/openshift/installer/pkg/terraform"
+	"github.com/openshift/installer/pkg/terraform/stages"
+)
+
+// PlatformStages are the stages to run to provision the infrastructure in vsphere.
+var PlatformStages = []terraform.Stage{
+	stages.NewStage("vsphere", "pre-bootstrap"),
+	stages.NewStage("vsphere", "bootstrap", stages.WithNormalDestroy()),
+	stages.NewStage("vsphere", "master"),
```
jstuever, line 12: Is there a reason to use three stages here? Can we not use "Cluster" and then "Bootstrap" where cluster is the combination of pre-bootstrap and master? It's entirely possible to deploy the master instances before we deploy the bootstrap host as they will simply be waiting on ignition until bootstrap is up. I'd prefer consistency... and one of these is not like the others.

- https://github.com/openshift/installer/pull/5010/commits/5c9e55ca35de20ee5e71264fc72e367036f36ee6#diff-0bbc45f383d06f3383bae5fb8e48e47f2ffa3182726b922e6e66321e93c55a5dR10-R11
- https://github.com/openshift/installer/pull/5010/commits/05e59bb8082491eff09ba594c73f42dc7bf27be5#diff-160e82277dbca268921ba3315fc07e8abac5e7bc6c5a94527ab1b92fa2d11599R13-R15
- https://github.com/openshift/installer/pull/5119/files#diff-d7310ec3465d4577b8f70742e7ecd5a22f463698d1aa807da17041613bcd486eR9-R11

jstuever (changes requested):
I think this can be simplified by combining pre-bootstrap with master, and be more like the other platforms as a result.

data/data/vsphere/master/main.tf
e-tienne: I tried to find more information on this, it was there since the creation of the file.
@jcpowermac do you know more?

pkg/terraform/stages/vsphere/stages.go
e-tienne: The idea was to have the first stage which was a prerequisite (template) for both the bootstrap and master stages.
It sounds like a good idea to standardize. Thanks for the input!

patrickdillon: I like @jstuever's suggestion. Azure has a limitation that master VMs will not create successfully until they have pulled ignition (this is different than other platforms) from bootstrap. So in the case of Azure I think it makes sense to create masters in a separate stage after bootstrap, but we should only do that as needed.

patrickdillon:
> In [.openshift_install.log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_installer/5094/pull-ci-openshift-installer-master-e2e-vsphere/1417530852111290368/artifacts/e2e-vsphere/ipi-install-install/artifacts/.openshift_install.log)
> 
> ```
> time="2021-07-20T17:49:04Z" level=debug msg="Warning: Value for undeclared variable"
> time="2021-07-20T17:49:04Z" level=debug
> time="2021-07-20T17:49:04Z" level=debug msg="The root module does not declare a variable named \"master_cores_per_socket\""
> time="2021-07-20T17:49:04Z" level=debug msg="but a value was found in file"
> time="2021-07-20T17:49:04Z" level=debug msg="\"/tmp/openshift-install-bootstrap-803409321/pre-bootstrap.tfvars.json\". To use"
> time="2021-07-20T17:49:04Z" level=debug msg="this value, add a \"variable\" block to the configuration."
> time="2021-07-20T17:49:04Z" level=debug
> time="2021-07-20T17:49:04Z" level=debug msg="Using a variables file to set an undeclared variable is deprecated and will"
> time="2021-07-20T17:49:04Z" level=debug msg="become an error in a future release. If you wish to provide certain \"global\""
> time="2021-07-20T17:49:04Z" level=debug msg="settings to all configurations in your organization, use TF_VAR_..."
> time="2021-07-20T17:49:04Z" level=debug msg="environment variables to set these instead."
> time="2021-07-20T17:49:04Z" level=debug
> time="2021-07-20T17:49:04Z" level=debug
> time="2021-07-20T17:49:04Z" level=debug msg="Warning: Value for undeclared variable"
> time="2021-07-20T17:49:04Z" level=debug
> time="2021-07-20T17:49:04Z" level=debug msg="The root module does not declare a variable named \"master_disk_size\" but a"
> time="2021-07-20T17:49:04Z" level=debug msg="value was found in file"
> time="2021-07-20T17:49:04Z" level=debug msg="\"/tmp/openshift-install-bootstrap-803409321/pre-bootstrap.tfvars.json\". To use"
> time="2021-07-20T17:49:04Z" level=debug msg="this value, add a \"variable\" block to the configuration."
> time="2021-07-20T17:49:04Z" level=debug
> time="2021-07-20T17:49:04Z" level=debug msg="Using a variables file to set an undeclared variable is deprecated and will"
> time="2021-07-20T17:49:04Z" level=debug msg="become an error in a future release. If you wish to provide certain \"global\""
> time="2021-07-20T17:49:04Z" level=debug msg="settings to all configurations in your organization, use TF_VAR_..."
> time="2021-07-20T17:49:04Z" level=debug msg="environment variables to set these instead."
> time="2021-07-20T17:49:04Z" level=debug
> time="2021-07-20T17:49:04Z" level=debug
> time="2021-07-20T17:49:04Z" level=debug msg="Warning: Value for undeclared variable"
> time="2021-07-20T17:49:04Z" level=debug
> time="2021-07-20T17:49:04Z" level=debug msg="The root module does not declare a variable named \"master_memory\" but a value"
> time="2021-07-20T17:49:04Z" level=debug msg="was found in file"
> time="2021-07-20T17:49:04Z" level=debug msg="\"/tmp/openshift-install-bootstrap-803409321/pre-bootstrap.tfvars.json\". To use"
> time="2021-07-20T17:49:04Z" level=debug msg="this value, add a \"variable\" block to the configuration."
> time="2021-07-20T17:49:04Z" level=debug
> time="2021-07-20T17:49:04Z" level=debug msg="Using a variables file to set an undeclared variable is deprecated and will"
> time="2021-07-20T17:49:04Z" level=debug msg="become an error in a future release. If you wish to provide certain \"global\""
> time="2021-07-20T17:49:04Z" level=debug msg="settings to all configurations in your organization, use TF_VAR_..."
> time="2021-07-20T17:49:04Z" level=debug msg="environment variables to set these instead."
> time="2021-07-20T17:49:04Z" level=debug
> time="2021-07-20T17:49:04Z" level=debug
> time="2021-07-20T17:49:04Z" level=debug msg="Warning: Values for undeclared variables"
> ```

This is common across all providers we have switched to stages and we need to fix it by making sure all output variables are declared.

data/data/vsphere/master/main.tf
jcpowermac: @e-tienne I can't find a ref to "master" anymore within:
https://github.com/openshift/baremetal-runtimecfg
Let's remove it and see if CI passes.

pkg/terraform/stages/vsphere/stages.go
e-tienne: @jstuever @patrickdillon I get a similar result to Azure by combining the 2 stages as a first "Cluster" stage, when using vSphere + output variables since we require the vmware tools on the vm to be first installed, before we can gather data in the output variables (bootstrap has to spin up and mcs, then the masters have to ignite and then reboot). 

I confirmed that the actual result of doing that in the suggested order is that the control nodes will wait for ignition until they time out with `"Error: timeout waiting for an available IP address"`

e-tienne: @jcpowermac I think your input could be useful here as well!

jcpowermac: - open-vm-tools is already installed in the image, just isn't started until after ignition
- The length of time for masters to be built, ignited might extend beyond the allocated time to grab the IP from tools - which is my concern with this.

e-tienne: Thanks! What I've seen is an infinite loop on the masters trying to fetch the ignition: `ignition: GET error: Get "https://$apiVIP/22623/config/master" dial tcp $apiVIP:22623: connect: no route to host` from the control nodes console, and the output variable for the control node IP address timing out (even after extending the timeouts to 45 minutes), the control node's IP address could not be found through this process.

e-tienne: I rolled back to 3 stages, and I was not able to reproduce the timeouts for the IP addresses with vmware tools so unless we hit this issue (I believe it was of my own) again, we should be able to roll use the stages design for vsphere.

jcpowermac:
@e-tienne it looks like you have openstack commits in here ;-)

data/data/vsphere/master/main.tf
```diff
@@ -2,30 +2,37 @@ locals {
...
+  memory               = var.vsphere_control_plane_memory_mib
   guest_id             = var.guest_id
   folder               = var.folder
   enable_disk_uuid     = "true"
   annotation           = local.description
 
-  wait_for_guest_net_timeout  = "0"
-  wait_for_guest_net_routable = "false"
+  wait_for_guest_net_timeout = 0
+  wait_for_guest_ip_timeout  = 15
```
jcpowermac, line 27: @e-tienne our dev/ci environments are very high performance. I am concerned with this number. Probably should be at least 30

e-tienne: Thanks, I will set it to 30 for now. Do we have other environments to test with?

jcpowermac: > Do we have other environments to test with?

We have two vSphere CI envs boskos determines where it lands.

e-tienne: ok, let's see how it goes. If you think that 30 is not enough, please let me know. 

I'll mark it as resolved for now.

jcpowermac:
/test e2e-vsphere

We aren't 100% where the EOF came from, let's run this again

staebler (commented):
This looks good to me, save for the remaining work to more reliably fetch the IP addresses for the machines.

data/data/vsphere/master/main.tf
```diff
@@ -2,30 +2,37 @@ locals {
...
+  memory               = var.vsphere_control_plane_memory_mib
   guest_id             = var.guest_id
   folder               = var.folder
   enable_disk_uuid     = "true"
   annotation           = local.description
 
-  wait_for_guest_net_timeout  = "0"
-  wait_for_guest_net_routable = "false"
+  wait_for_guest_net_timeout  = 0
+  wait_for_guest_net_routable = false
```
staebler, line 27: Although superficial, let's leave out these changes.

data/data/vsphere/bootstrap/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "bootstrap_ip" {
+  value = vsphere_virtual_machine.vm_bootstrap.default_ip_address
```
jcpowermac, line 2: Is this going to be a problem if this value is an empty string?

pkg/terraform/stages/compat/stage.go
```diff
@@ -104,15 +102,6 @@ func extractHostAddresses(config *types.InstallConfig, tfstate *terraform.State)
...
 
-	case vspheretypes.Name:
-		bootstrap, err = gathervsphere.BootstrapIP(config, tfstate)
-		if err != nil {
-			return
-		}
-		masters, err = gathervsphere.ControlPlaneIPs(config, tfstate)
-		if err != nil {
-			logrus.Error(err)
-		}
```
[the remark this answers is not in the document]
jcpowermac: > The plan is to handle the gather with the output variables like the other platform stages as we could not reproduce issue with it at the moment. This should give us the opportunity to have other people than me test these changes as well.
> 
> However, if there is any issue, we would want to implement a custom host address extraction that should talk directly to vcenter instead 

This _already_ exists why not just use it?

> of looking up the tfstate files, because my understanding is that the tfstate file content should reflect the output variables for the bootstrap ip and control node ips

@e-tienne if you are not waiting for the ip address from vmware tools in terraform we cannot guarantee that the ip addresses will be available.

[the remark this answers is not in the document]
e-tienne: > > The plan is to handle the gather with the output variables like the other platform stages as we could not reproduce issue with it at the moment. This should give us the opportunity to have other people than me test these changes as well.
> > However, if there is any issue, we would want to implement a custom host address extraction that should talk directly to vcenter instead
> 
> This _already_ exists why not just use it?
> 
> > of looking up the tfstate files, because my understanding is that the tfstate file content should reflect the output variables for the bootstrap ip and control node ips
> 
> @e-tienne if you are not waiting for the ip address from vmware tools in terraform we cannot guarantee that the ip addresses will be available.

Clarification: 
- we will use the IP gather based on the moid for vsphere that already exists.
- the configuration for waiting for the IP address from vmware tools was removed after my last comment, because we plan to reliably get the machine IP addresses from vsphere.

data/data/vsphere/bootstrap/outputs.tf
staebler: No, it's not a problem. It is only used for the bootstrap gather. There will be code in a follow-up PR to handle querying the IP address from the vCenter (as we currently do) when this variable is empty.

PR #5095, opened 2021-07-20
azurestack: Modify destroy code to handle public dns record deletion

rna-afk:
Modifying the Azure deletion code to check if the environment is
Azure Stack Hub and if so, check and delete the A records in the
public DNS that is relevant to the current cluster during cluster
destroy.


pkg/destroy/azure/azure.go
```diff
@@ -196,6 +205,73 @@ func (o *ClusterUninstaller) Run() error {
...
+
+	privateDNSRecordName := fmt.Sprintf("api-int.%s", clusterName)
+	publicDNSRecordName := fmt.Sprintf("api.%s", clusterName)
+	for _, zone := range allZones.List() {
+		for recordPages, err := recordsClient.ListByDNSZone(ctx, rgName, zone, to.Int32Ptr(100), ""); recordPages.NotDone(); err = recordPages.NextWithContext(ctx) {
+			if err != nil {
+				return err
+			}
+			for _, record := range recordPages.Values() {
+				if t := to.String(record.Name); t == privateDNSRecordName || t == publicDNSRecordName {
```
patrickdillon, line 257: I wonder if in addition we can check based on a tag/metadata. I see that there is support for creating with a tag/metadata. 

So would the client support a check to see if the record has the tag with key/value:
```
"kubernetes.io_cluster.${var.cluster_id}" = "owned"
```
I don't believe we can locate the DNS record by this tag, which would be the ideal scenario (then we would not need to add new values to the metadata). 

This can also be followed up in a subsequent PR to add the tag in the terraform code & check the tag here.

patrickdillon (commented):
This is in good shape. I don't love deleting based on the name but I don't think we have many other options. Adding a check against tag/metadata will provide some more piece of mind, but as I said we can follow that up later if time does not allow doing it now.

pkg/destroy/azure/azure.go
```diff
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	azurestackdns "github.com/Azure/azure-sdk-for-go/services/dns/mgmt/2016-04-01/dns"
```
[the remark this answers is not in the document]
rna-afk: Testing it again right now. Should make this simple if it works.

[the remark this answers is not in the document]
rna-afk: Nope not working. Error message
```
DEBUG dns.ZonesClient#ListByResourceGroup: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="InvalidResourceType" Message="The resource type 'dnsZones' could not be found in the namespace 'Microsoft.Network' for api version '2018-03-01-preview'. The supported api-versions are '2015-05-04-preview,2016-04-01'." 

```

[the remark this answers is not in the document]
rna-afk: Nope does not work. `2018-03-01` does not have a dns service and I tested it with the `2018-05-01` and it gave me this error.
```
DEBUG dns.ZonesClient#ListByResourceGroup: Failure responding to request: StatusCode=404 
-- Original Error: autorest/azure: Service returned an error. Status=404 Code="InvalidResourceType" 
Message="The resource type 'dnsZones' could not be found in the namespace 
'Microsoft.Network' for api version '2018-05-01'. 
The supported api-versions are '2015-05-04-preview,2016-04-01'."       
```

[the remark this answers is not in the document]
patrickdillon:
```suggestion
		azurestackdns "github.com/Azure/azure-sdk-for-go/profiles/2018-03-01/dns/mgmt/dns"
```
Then `go mod tidy` `go mod vendor` should work. But this might be a bit pedantic because it is ultimately just[ using `2016-04-01` under the hood](https://github.com/Azure/azure-sdk-for-go/blob/main/profiles/2018-03-01/dns/mgmt/dns/models.go#L14), which is what you're already using.

pkg/destroy/azure/azure.go
```diff
@@ -196,6 +205,72 @@ func (o *ClusterUninstaller) Run() error {
...
+			errs = append(errs, errors.Wrap(err, "failed to advance to next dns zone"))
+			continue
+		}
+		for _, zone := range zonesPage.Values() {
+			allZones.Insert(to.String(zone.Name))
+		}
+	}
+
+	privateDNSRecordName := fmt.Sprintf("api-int.%s", clusterName)
+	publicDNSRecordName := fmt.Sprintf("api.%s", clusterName)
```
patrickdillon, line 249: We need to delete the ingress controller record as well:
`
"*.apps.%s"
`

pkg/destroy/azure/azure.go
patrickdillon: Actually if we can check on tags, then we wouldn't need/want to check the names. That is, we could just loop through all the records and see if they match the tag (not the name). But yeah that can be a subsequent PR once we confirm installer, ingress and any other dns records can be tagged.

rna-afk: I'll check it out. Will update this PR.

rna-afk: Are we adding the tags right now? I don't see them in the DNS records on the portal.

pkg/destroy/azure/azure.go
```diff
@@ -116,7 +121,11 @@ func (o *ClusterUninstaller) Run() error {
 		waitCtx,
 		func(ctx context.Context) {
 			o.Logger.Debugf("deleting public records")
-			err = deletePublicRecords(ctx, o.zonesClient, o.recordsClient, o.privateZonesClient, o.privateRecordSetsClient, o.Logger, o.ResourceGroupName)
+			if o.Environment.Name == "AzureStackCloud" {
```
patrickdillon, line 124: Sorry I missed this earlier but instead of `"AzureStackClod"` you should use the constant, I believe it would be `azure.StackCloud`

pkg/destroy/azure/azure.go
patrickdillon: No. We can add them in a subsequent pr.

PR #5099, opened 2021-07-21
ibmcloud: Destroy cluster

BobbyRadford:
Remove all IBM Cloud resources that are created with an OpenShift cluster.

### DIFF
--- a/pkg/destroy/ibmcloud/cloudobjectstorage.go
+++ b/pkg/destroy/ibmcloud/cloudobjectstorage.go
@@ -8,6 +8,7 @@ import (
 )
 
 const cosTypeName = "cos instance"
+const cosResourceID = "dff97f5c-bc5e-4455-b470-411c3edbe49c"
 
 // listCOSInstances lists COS service instances
 func (o *ClusterUninstaller) listCOSInstances() (cloudResources, error) {
@@ -22,6 +23,7 @@ func (o *ClusterUninstaller) listCOSInstances() (cloudResources, error) {
 
 	options := o.controllerSvc.NewListResourceInstancesOptions()
 	options.SetResourceGroupID(resourceGroupID)
+	options.SetResourceID(cosResourceID)
 	resources, _, err := o.controllerSvc.ListResourceInstancesWithContext(ctx, options)
 	if err != nil {
 		return nil, errors.Wrapf(err, "Failed to list COS instances")
@@ -29,8 +31,10 @@ func (o *ClusterUninstaller) listCOSInstances() (cloudResources, error) {
 
 	result := []cloudResource{}
 	for _, instance := range resources.Resources {
-		cosInstanceName := fmt.Sprintf("%s-cos", o.InfraID)
-		if cosInstanceName == *instance.Name {
+		// Match the COS instances created by both the installer and the
+		// cluster-image-registry-operator.
+		if fmt.Sprintf("%s-cos", o.InfraID) == *instance.Name ||
+			fmt.Sprintf("%s-image-registry", o.InfraID) == *instance.Name {
 			result = append(result, cloudResource{
 				key:      *instance.ID,
 				name:     *instance.Name,
@@ -105,13 +109,16 @@ func (o *ClusterUninstaller) COSInstanceID() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	if len(cosInstances.list()) == 0 {
-		return "", errors.Errorf("No COS instances found")
+	instanceList := cosInstances.list()
+	if len(instanceList) == 0 {
+		return "", errors.Errorf("COS instance not found")
 	}
-	if len(cosInstances.list()) > 1 {
-		return "", errors.Errorf("Too many COS instances found")
+
+	// Locate the installer's COS instance by name.
+	for _, instance := range instanceList {
+		if instance.name == fmt.Sprintf("%s-cos", o.InfraID) {
+			return instance.id, nil
+		}
 	}
-	cosInstanceID := cosInstances.list()[0].id
-	o.cosInstanceID = cosInstanceID
-	return cosInstanceID, nil
+	return "", errors.Errorf("COS instance not found")
 }
--- a/pkg/destroy/ibmcloud/cloudobjectstorage.go
+++ b/pkg/destroy/ibmcloud/cloudobjectstorage.go
@@ -24,6 +24,7 @@ func (o *ClusterUninstaller) listCOSInstances() (cloudResources, error) {
 	options := o.controllerSvc.NewListResourceInstancesOptions()
 	options.SetResourceGroupID(resourceGroupID)
 	options.SetResourceID(cosResourceID)
+	options.SetType("service_instance")
 	resources, _, err := o.controllerSvc.ListResourceInstancesWithContext(ctx, options)
 	if err != nil {
 		return nil, errors.Wrapf(err, "Failed to list COS instances")
@@ -117,6 +118,7 @@ func (o *ClusterUninstaller) COSInstanceID() (string, error) {
 	// Locate the installer's COS instance by name.
 	for _, instance := range instanceList {
 		if instance.name == fmt.Sprintf("%s-cos", o.InfraID) {
+			o.cosInstanceID = instance.id
 			return instance.id, nil
 		}
 	}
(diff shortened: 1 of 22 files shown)


pkg/destroy/ibmcloud/ibmcloud.go
```diff
@@ -0,0 +1,382 @@
...
+	err = o.destroyCluster()
+	if err != nil {
+		return errors.Wrap(err, "failed to destroy cluster")
+	}
+
+	return nil
+}
+
+func (o *ClusterUninstaller) destroyCluster() error {
+	stagedFuncs := [][]struct {
```
hasueki, line 97: Should this also delete the resource group, if created by the installer?

pkg/destroy/ibmcloud/instance.go
```diff
@@ -0,0 +1,155 @@
+package ibmcloud
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/IBM/go-sdk-core/core"
```
hasueki, line 8:
```suggestion
	"github.com/IBM/go-sdk-core/v5/core"
```

pkg/destroy/ibmcloud/cloudobjectstorage.go
```diff
@@ -0,0 +1,117 @@
...
+	defer cancel()
+
+	resourceGroupID, err := o.ResourceGroupID()
+	if err != nil {
+		return nil, err
+	}
+
+	options := o.controllerSvc.NewListResourceInstancesOptions()
+	options.SetResourceGroupID(resourceGroupID)
+	resources, _, err := o.controllerSvc.ListResourceInstancesWithContext(ctx, options)
```
gsrajpal, line 25: We may also want to set a `Limit` option here (Number of items returned)
...not sure what the default is, but might be a good idea.


Also, if there are more than `100` Instances in the ResourceGroup there is a possibility we may not find the `cosInstanceName`. Might be a good idea to use `next_url` from Response and request the next page of Instances.

gsrajpal: OR you can use `ResourceID` **"is.cos"** or `Type` option to get a filtered response. Will save time while iterating over the list.

pkg/destroy/ibmcloud/vpc.go
```diff
@@ -0,0 +1,99 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/IBM/vpc-go-sdk/vpcv1"
+	"github.com/pkg/errors"
+)
+
+const vpcTypeName = "vpc"
+
+// listVPCs lists subnets in the vpc
```
hasueki, line 13: Update comment

hasueki (changes requested):

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/ChrisTrenkamp/goxpath v0.0.0-20190607011252-c5096ec8773d // indirect
 	github.com/IBM-Cloud/bluemix-go v0.0.0-20210611051827-cdc80c935c05
 	github.com/IBM-Cloud/terraform-provider-ibm v1.26.2
+	github.com/IBM/go-sdk-core v1.1.0
```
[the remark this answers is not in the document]
BobbyRadford: This was added when I did go mod tidy ?

pkg/destroy/ibmcloud/cloudobjectstorage.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to list COS instances")
+	}
+
+	result := []cloudResource{}
+	for _, instance := range resources.Resources {
+		cosInstanceName := fmt.Sprintf("%s-cos", o.InfraID)
```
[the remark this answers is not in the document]
BobbyRadford: ok 👍

pkg/destroy/ibmcloud/instance.go
BobbyRadford: Ah i see. Nice catch

pkg/destroy/ibmcloud/dns.go
```diff
@@ -0,0 +1,97 @@
...
+		options := o.dnsRecordsSvc.NewListAllDnsRecordsOptions()
+		resources, _, err := o.dnsRecordsSvc.ListAllDnsRecordsWithContext(ctx, options)
+
+		if err != nil {
+			return nil, errors.Wrapf(err, "Failed to list DNS records")
+		}
+
+		for _, record := range resources.Result {
+			if strings.Contains(*record.Name, o.ClusterName) ||
+				strings.Contains(*record.Content, o.ClusterName) {
```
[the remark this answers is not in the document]
BobbyRadford: The cluster ID is a GUID (not in the record name or content. What do you mean by cluster ID?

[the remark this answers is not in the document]
BobbyRadford: This is the metadata we have access to for deleting resources:
```json
{
  "clusterName": "ocp4-ipi-072103",
  "clusterID": "5ee7b41c-1699-4b79-8530-06775548e232",
  "infraID": "ocp4-ipi-072103-b6576",
  "ibmcloud": {
    "accountID": "1e1f75646aef447814a6d907cc83fb3c",
    "baseDomain": "ibm.bobbyradford.com",
    "cisInstanceCRN": "crn:v1:bluemix:public:internet-svcs:global:a/1e1f75646aef447814a6d907cc83fb3c:5a8989d1-54c2-4634-9a33-63c57c5f129a::",
    "region": "us-east",
    "resourceGroupName": "ocp4-ipi-072103-b6576"
  }
}
```

If you mean `infraID`, that won't work because the DNS record uses the `cluster_name` TF var, which doesn't include the random suffix that the infraID has:

```json
{
"cluster_domain": "ocp4-ipi-072103.ibm.bobbyradford.com"
}
```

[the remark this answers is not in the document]
hasueki: > The cluster ID is a GUID (not in the record name or content. What do you mean by cluster ID?

Ah I see, yes I was confused with the naming there. I was referring to the `infraID`.

pkg/destroy/ibmcloud/cloudobjectstorage.go
BobbyRadford: "is.cos"? Wouldn't this have to be the COS service GUID?

pkg/destroy/ibmcloud/ibmcloud.go
BobbyRadford: Yeah. It should. Meant to hit that one as well.

pkg/destroy/ibmcloud/cloudobjectstorage.go
BobbyRadford: Ideally, we'd handle the pagination, but the vast majority of the time, there would never be >100 COS instances within the cluster resource group. Something very weird would have happened in that case. So, I'll make sure we only pull COS instances and not all types of RC Resources.

pkg/destroy/ibmcloud/dns.go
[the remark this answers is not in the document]
hasueki: > If you mean infraID, that won't work because the DNS record uses the cluster_name TF var, which doesn't include the random suffix that the infraID has...

Sorry, which part will not work? I think it should look for DNS records with..
- {infraID}-master-{index}.{clusterName}
- api.{clusterName}
- api-int.{clusterName}

..or we can simply delete all DNS records under the cluster subdomain. For this, we need to make sure the `clusterName` string value is located in the proper segment in the subdomain.

e.g. (if clusterName = "cluster1")
- api.cluster1 (Match)
- api.cluster1.test (No match)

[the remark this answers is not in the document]
BobbyRadford: OK. I see what you're saying. I'm onboard.

pkg/destroy/ibmcloud/cloudobjectstorage.go
gsrajpal: I meant `>100` all instances under a ResourceGroup.

Yup, only pulling COS would be perfect.

You can pass in `ResourceID`: e.g.`is.volume`, `is.image`,` is.vpc` - Not sure, but there might be `is.cos`

pkg/destroy/ibmcloud/dns.go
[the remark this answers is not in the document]
BobbyRadford: Added logic to support your second suggestion here.

PR #5100, opened 2021-07-22
Bug 1984576: baremetal: reinstate provisioningInterface for provisioning CR

hardys:
This was removed in #5015 but we should still include it if a user
provides a value - since it's now optional we'll pass an empty
string if no value is provided, which will trigger the new logic
that detects the interface name by MAC.

### DIFF
--- a/data/data/manifests/openshift/baremetal-provisioning-config.yaml.template
+++ b/data/data/manifests/openshift/baremetal-provisioning-config.yaml.template
@@ -3,6 +3,7 @@ kind: Provisioning
 metadata:
   name: provisioning-configuration
 spec:
+  provisioningInterface: "{{.Baremetal.ProvisioningNetworkInterface}}"
   provisioningIP: "{{.Baremetal.ClusterProvisioningIP}}"
   provisioningNetworkCIDR: "{{.Baremetal.ProvisioningNetworkCIDR}}"
   provisioningNetwork: "{{.Baremetal.ProvisioningNetwork}}"


derekhiggins:
lgtm

hardys:
I tested with and without the optional provisioningNetworkInterface and can see the templating results in an empty string when it's not provided which I think should work fine with the related https://github.com/metal3-io/ironic-image/pull/272/files changes:

```
$ cat ocp/ostest/openshift/99_baremetal-provisioning-config.yaml 
apiVersion: metal3.io/v1alpha1
kind: Provisioning
metadata:
  name: provisioning-configuration
spec:
  provisioningInterface: ""
```

hardys:
/retitle Bug 1984576: baremetal: reinstate provisioningInterface for provisioning CR

derekhiggins:
e2e-metal-ipi-ovn-ipv6 failed for a different reason (BMO failed to start)

> {"level":"error","ts":1626958012.7065668,"logger":"setup","msg":"unable to start manager","error":"error listening on :9440: listen tcp :9440: bind: address already in use","stacktrace":"main.main\n\t/go/src/github.com/metal3-io/baremetal-operator/main.go:134\nruntime.main\n\t/usr/lib/golang/src/runtime/proc.go:225"}

Looks like the bug is fixed, static-ip-manager did its job
> + /usr/sbin/ip addr add fd00:1101::3/64 dev enp1s0 valid_lft 10 preferred_lft 10

hardys:
> e2e-metal-ipi-ovn-ipv6 failed for a different reason (BMO failed to start)
> 
> > {"level":"error","ts":1626958012.7065668,"logger":"setup","msg":"unable to start manager","error":"error listening on :9440: listen tcp :9440: bind: address already in use","stacktrace":"main.main\n\t/go/src/github.com/metal3-io/baremetal-operator/main.go:134\nruntime.main\n\t/usr/lib/golang/src/runtime/proc.go:225"}

Ack, that is https://bugzilla.redhat.com/show_bug.cgi?id=1983975 which I believe @andfasano is looking into
 
> Looks like the bug is fixed, static-ip-manager did its job
> 
> > * /usr/sbin/ip addr add fd00:1101::3/64 dev enp1s0 valid_lft 10 preferred_lft 10

\o/

PR #783, opened 2018-12-04
Remove clusterID from installconfig and move it to cluster

rajatchopra:
ClusterID is now removed from installconfig. The reason is that it should not be possible for a user to override this value. The clusterID is still needed for destroy - and hence it is now a separate asset which gets stored in ClusterMetadata. Other assets needing the clusterID (e.g. legacy manifests) issue a separate dependency on this asset.
For convenience of package depenencies, the asset still lives in the installconfig package.
A follow up PR can move it to cluster package if needed (currently blocked by legacy CVO overrides).

Fixes issue #761

Issue #761, opened 2018-11-29
Drop ClusterID from InstallConfig


dgoodwin:
As discussed in arch call today and at Clayton's request, we decided we should remove the ClusterID from the InstallConfig. It should not be possible for a user to override this value and it should be generated fresh on every installer invocation.

crawford:
Duplicate of #538.

PR #5101, opened 2021-07-22
Allow to set the 'InfraID' with an environment variable.

mmartinv:
This change allows the user to set the 'INFRA_ID' environment
variable to modify the 'InfraID'.

The contents of the 'INFRA_ID' must be valid according to the current
generation, this is: it must start with the cluster name
followed by ('-') and 5 random alphanumeric chars with no vowels,
otherwise the 'INFRA_ID' env variable will be ignored and 'InfraID'
will be generated as usual.

If the cluster name contains special characters or if it has more than
27 chars long, the following rules will be applied in order to generate
the final cluster name:

- replace all characters that are not `alphanum` or `-` with `-`.
- replace all multiple dashes in a sequence with single one.
- truncate to 27 chars long
- if the resulting cluster name ends with '-', remove this last char.

An example on how to use it:

  openshift-install create install-config --dir=.
  grep -A 2 metadata install-config.yaml
  metadata:
    creationTimestamp: null
    name: cloud

  INFRA_ID="cloud-xxxxx" openshift-install create manifests --dir=.

This change is needed to implement external provider support in the
assisted-installer project.


patrickdillon:
> This change is needed to implement external provider support in the
assisted-installer project.

Can you elaborate? It seems like this a dependency on something that is designed to be non-deterministic.

pkg/asset/installconfig/clusterid.go
```diff
@@ -74,6 +75,15 @@ func generateInfraID(base string, maxLen int) string {
 	}
 	base = strings.TrimRight(base, "-")
 
+	// Return the InfraID provided by the "INFRA_ID" environment variable if valid
+	if envInfraID, ok := os.LookupEnv("INFRA_ID"); ok && envInfraID != "" {
```
Gal-Zaidman, line 79: - Can you change to "OPENSHIFT_INSTALL_INFRA_ID_OVERRIDE" to comply with the rest of the env vars
- Please change to:
if envInfraID, ok := os.LookupEnv

and add a warn log message like in:
https://github.com/openshift/installer/blob/6d778f911e79afad8ba2ff4301eda5b5cf4d8e9e/pkg/asset/rhcos/image.go#L51-L55

pkg/asset/installconfig/clusterid.go
```diff
@@ -74,6 +75,15 @@ func generateInfraID(base string, maxLen int) string {
...
+		validInfraIDString := fmt.Sprintf("^%s-([b-df-hj-np-tv-z0-9]){%d}$", base, randomLen)
```
Gal-Zaidman, line 80: Maybe move this pattern to a constant and explain what are you looking for

pkg/asset/installconfig/clusterid.go
```diff
@@ -74,6 +75,15 @@ func generateInfraID(base string, maxLen int) string {
...
+		validInfraIDRegex := regexp.MustCompile(validInfraIDString)
+		if validInfraIDRegex.MatchString(envInfraID) {
+			return envInfraID
```
Gal-Zaidman, line 83: Incase there is no match you need to return with an error.
Think that the user set the "OPENSHIFT_INSTALL_INFRA_ID_OVERRIDE" env var, he doesn't need to wait till the installation starts to see that the env var is not good

Gal-Zaidman:
> > This change is needed to implement external provider support in the
> > assisted-installer project.
> 
> Can you elaborate? It seems like this a dependency on something that is designed to be non-deterministic.

Sure in the installer the InfraID is used to set the names of the VMs and for oVirt it is also used to set a cluster tag (in oVirt).
We are working on an effort to introduce support for external providers (oVirt, vsphere, and anyone else) to the assisted installer.
In the assisted installer the user provisions the OCP nodes, that are discovered by the assisted installer service, and then the assisted installer service uses the openshift-installer to create the manifests and install config.
One of the earliest problems we encountered is that due to the random InfraID the names of the Nodes(in the manifest) are different from the nodes the user provisions, and since it is random we can't tell the user to create the Nodes with specific names.
Also, the InfraID is later used by the platforms in various ways to identify the cluster, for example in oVirt we use it to set an oVirt tag which helps us group the cluster resources in oVirt, and we rely on it in various places such as cluster destroy or CSI/Machine provider logic to filter cluster resources quickly and prevent unnecessary API calls.

Having a hook in the installer to set the InfraID prior to the installation will help us overcome does issues.

pkg/asset/installconfig/clusterid.go
```diff
@@ -74,6 +75,15 @@ func generateInfraID(base string, maxLen int) string {
...
+		}
+	}
+
 	// add random chars to the end to randomize
 	return fmt.Sprintf("%s-%s", base, utilrand.String(randomLen))
```
Gal-Zaidman, line 88: Not sure you need all the complicated regix logic, if you verify that the infraID is the cluster name followed by ('-') and 5 random alphanumeric chars with no vowels, then you can just have the env var replace the `utilrand.String(randomLen)` And have some validation to make sure it is 5 chars and if not return an error

mmartinv: > Not sure you need all the complicated regix logic, if you verify that the infraID is the cluster name followed by ('-') and 5 random alphanumeric chars with no vowels, 

Not sure if I understand what you mean. This is precisely what the regex does. We must verify the 'cluster name' contained in the 'install-config.yaml' with the modifications done by the original generation, matches the 'cluster name' part in the envInfraID or we can end up using an incorrect 'cluster name' (with invalid characters or wrong length) or even a totally different 'cluster name'.  The only part allowed to be modified by the user is the 'random' part of the InfraID and must be provided with the same original rules (5 alphanumeric lowercase chars with no vowels and ).  Besides that I don't think it's a "complicated regex logic" but the opposite, it's the most simple way to do it following the current rules.

>then you can just have the env var replace the `utilrand.String(randomLen)` And have some validation to make sure it is 5 chars and if not return an error

The env var cannot replace the `utilrand.String(randomLen)` only because it contains the cluster name also. Still, verifying the random part is 5 lowercase chars long with no vowels would require a regex anyway.

pkg/asset/installconfig/clusterid.go
mmartinv: Sure

pkg/asset/installconfig/clusterid.go
mmartinv: Ok

pkg/asset/installconfig/clusterid.go
mmartinv: I tried to explain more in the comment, moving the pattern to a constant make the code less readable IMO.

patrickdillon:
/hold
See #761
This should be discussed at greater length (something like enhancement/arch call). If these discussions have taken place apologies for not being aware of them already.

patrickdillon:
/hold
I would like to hold this for some more discussion. 

Has this been discussed in an enhancement/arch call? I am concerned about the dependencies we are introducing here. These topics have been discussed with regard to the cluster ID eg. #761 & #783 but I am not certain how this relates to the infra id.

Gal-Zaidman:
> /hold
> I would like to hold this for some more discussion.
> 
> Has this been discussed in an enhancement/arch call?
Added it to my enhancement on assisted installer: https://github.com/openshift/assisted-service/pull/2272/files#diff-988c4e5d655472b2174837171f5d89978ca0f86ec8554329b386dccb9990ea83R134-R147

>I am concerned about the dependencies we are introducing here. These topics have been discussed with regard to the cluster ID eg. #761 & #783 but I am not certain how this relates to the infra id.

Those two seem unrelated to me (unless I missed something) anyway I don't think we are adding dependencies, we keep the situation as is, but add a hook to preset it (while making sure that the structure is identical to the current state)

patrickdillon:
> > /hold
> > I would like to hold this for some more discussion.
> > Has this been discussed in an enhancement/arch call?
> 
> Added it to my enhancement on assisted installer: https://github.com/openshift/assisted-service/pull/2272/files#diff-988c4e5d655472b2174837171f5d89978ca0f86ec8554329b386dccb9990ea83R134-R147
> 
> > I am concerned about the dependencies we are introducing here. These topics have been discussed with regard to the cluster ID eg. #761 & #783 but I am not certain how this relates to the infra id.
> 
> Those two seem unrelated to me (unless I missed something) anyway I don't think we are adding dependencies, we keep the situation as is, but add a hook to preset it (while making sure that the structure is identical to the current state)

Thanks I left a comment in the enhancement. The dependency is that we are creating reliable behavior through deterministic naming conventions so consumers may start to depend on resources with certain names. Although this is an environment variable it is essentially creating an API.

pkg/asset/installconfig/clusterid.go
Gal-Zaidman: Yes but the user knows the clustername in advance, so I don't see why we should override it (or override it and then validate it is the same).
The problem that this change is supposed to fix is that we will don't know the random part prior to the installation so we can't instruct the users regarding the naming scheme of the VMs they will provision, we can verify anything else.

Plus why no vowels?

Gal-Zaidman:
> > > /hold
> > > I would like to hold this for some more discussion.
> > > Has this been discussed in an enhancement/arch call?
> > 
> > 
> > Added it to my enhancement on assisted installer: https://github.com/openshift/assisted-service/pull/2272/files#diff-988c4e5d655472b2174837171f5d89978ca0f86ec8554329b386dccb9990ea83R134-R147
> > > I am concerned about the dependencies we are introducing here. These topics have been discussed with regard to the cluster ID eg. #761 & #783 but I am not certain how this relates to the infra id.
> > 
> > 
> > Those two seem unrelated to me (unless I missed something) anyway I don't think we are adding dependencies, we keep the situation as is, but add a hook to preset it (while making sure that the structure is identical to the current state)
> 
> Thanks I left a comment in the enhancement. The dependency is that we are creating reliable behavior through deterministic naming conventions so consumers may start to depend on resources with certain names. Although this is an environment variable it is essentially creating an API.

Thanks now I understand your concerns a bit better, but I think that as long as the env var(or API) follows the same deterministic naming conventions then it shouldn't be a problem.
Yes then can be a situation in which a user will deploy 2 clusters with the same env var resulting in 2 clusters with the same infraID, but since it will not be documented for customers, includes a warning in the installation log, and is for our internal usage I think it is OK.
Worse and much harder to detect problems can occure with using the other 2 env vars ;)

pkg/asset/installconfig/clusterid.go
mmartinv: > Yes but the user knows the clustername in advance, so I don't see why we should override it (or override it and then validate it is the same).
> The problem that this change is supposed to fix is that we will don't know the random part prior to the installation so we can't instruct the users regarding the naming scheme of the VMs they will provision, we can verify anything else.
>
 
Ok, I get it now, so the `OPENSHIFT_INSTALL_INFRA_ID_OVERRIDE` will contain the random part only.

> Plus why no vowels?

Don't know the actual reason this is what `utilrand.String(randomLen)` generates. https://pkg.go.dev/k8s.io/apimachinery/pkg/util/rand#String

staebler:
In general, the use of environment variables to drive configuration of the cluster is discouraged in the installer. The install-config.yaml is the place where configuration should live. We have allowed the use of environment variables in a few rare cases, where the alternatives have been onerous or where the environment variable is only for testing and not supported for use by the end user.

I am not yet convinced that this is one of the cases where an exception is justified. If we want to provide the option for the user to be able to use a pre-determine Infra ID, then we should add that to the install-config.yaml. However, It seems reasonable enough to me to have a user of the installer that needs the infra ID get it from the infrastructure manifest rather than taking the approach of the user supplying the infra ID.

PR #5084, opened 2021-07-15
Azure Stack IPI Support

patrickdillon:
Adds Terraform configuration to build Azure Stack clusters. Successfully creates all infrastructure, but cluster bootstrapping depends on further upstream changes.

Currently marked as WIP because:

- [x] Need to revendor for OpenShift fork of Terraform Provider. That can happen now that this has been tested succesfully
- [x] Align Terraform configuration with new staged approach. We can begin this work now, but we may want to push that into a separate PR to allow parallel work.

This should unblock CI work in openshift/release#18796

PR #5102, opened 2021-07-22
Azure Stack: Worker availability set

patrickdillon:
The Azure provider is throwing an error when trying to create worker machines outside of an availability set. The machine api operator does not have the ability to create availability sets, so the installer must create one for worker nodes.

0406764a5154feb830899739e6057a188b48dd44 is the significant commit, while the rest of the PR is built on top of #5084 

/hold
This is meant to enable the cloud team to test as a PoC while we determine the best way forward. 

If we are to move forward with the installer creating this resource for the MAO, we will need to establish some API interface such as a field in the the machine(set) spec.


patrickdillon:
cc @JoelSpeed @alexander-demichev @lobziik

tsze-redhat:
@patrickdillon 
What is the best way to test this PR?
Create a cluster with only 2 zones and then try to add a machine in 3rd zone using a ymal?

patrickdillon:
> @patrickdillon
> What is the best way to test this PR?
> Create a cluster with only 2 zones and then try to add a machine in 3rd zone using a ymal?

As I mentioned in the description, we have not decided if we are going forward with this approach, so it is probably best to wait. 

This PR would enable the creation of workers, so ensuring workers are created would indicate success. 

I think you may be confusing availability zones (do not exist in ASH) with availability sets (the topic of this PR).

PR #5105, opened 2021-07-23
open cluster-policy-controller ports for metrics

atiratree:
This opens ports for gathering cluster-policy-controller metrics. The container itself is running on the same pod as kube-controller-manager.

I updated the configs in similar manner as kube-controller-manager. Please let me know if there are other steps needed.

### DIFF
--- a/data/data/aws/cluster/vpc/sg-master.tf
+++ b/data/data/aws/cluster/vpc/sg-master.tf
@@ -289,6 +289,28 @@ resource "aws_security_group_rule" "master_ingress_kube_controller_manager_from_
   to_port   = 10257
 }
 
+resource "aws_security_group_rule" "master_ingress_cluster_policy_controller" {
+  type              = "ingress"
+  security_group_id = aws_security_group.master.id
+  description       = local.description
+
+  protocol  = "tcp"
+  from_port = 10357
+  to_port   = 10357
+  self      = true
+}
+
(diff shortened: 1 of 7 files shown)


jstuever:
Is there an enhancement doc for this we can reference?

atiratree:
@jstuever no there is not. Should I create one?

atiratree:
@jstuever fyi it is also a part of a Jira issue: WRKLDS-308

atiratree:
rebased

atiratree:
> Is there an enhancement doc for this we can reference?

I just found out we have a doc describing this port: https://github.com/openshift/enhancements/blob/master/dev-guide/host-port-registry.md

we can add metrics to the notes there once this PR merges

data/data/aws/cluster/vpc/sg-master.tf
```diff
@@ -289,6 +289,28 @@ resource "aws_security_group_rule" "master_ingress_kube_controller_manager_from_
   to_port   = 10257
 }
 
+resource "aws_security_group_rule" "master_ingress_cluster_policy_controller" {
```
staebler, line 292: The AWS UPI should be updated with this as well. `upi/aws/cloudformation/03_cluster_security.yaml`

atiratree: Thanks,  updated.

staebler (approved):
/lgtm

PR #5106, opened 2021-07-26
Bug 1977129: Remove runlevel label from openshift-kubevirt-infra

rna-afk:
Runlevel was required in installer version <4.6 due to significant
start times of components waiting for the apiserver but it is no
longer necessary. More info is in this bug:
https://bugzilla.redhat.com/show_bug.cgi?id=1977129


mtnbikenc (approved):
/lgtm

PR #5107, opened 2021-07-27
Bug 1996124: version: display release architecture

yselkowitz:
This should make it easier to identify the release/payload architecture
which a particular installer binary will install, since that is often
not clear from the release image name.


yselkowitz:
fyi @cben

jstuever (approved):
/lgtm
/approve

PR #5108, opened 2021-07-27
Bug 1986420: GCP: make cluster_ip_address optional post-bootstrap

patrickdillon:
The cluster_ip_address output variable is null in the cluster stage when the publishing strategy is internal. A null output variable results in no value actually being written to the tfvars. This results in an error when running with publish: internal: "No value for required variable."


mtnbikenc (approved):
/lgtm

PR #5116, opened 2021-07-30
Bug 1994103: ibmcloud: Support Terraform stages

hasueki:
Splits main Terraform module into stages:
- network: VPC, LBs, DNS, VM image, etc
- bootstrap: Bootstrap machine
- master: Master nodes


pkg/terraform/stages/ibmcloud/stages.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"github.com/openshift/installer/pkg/terraform"
+	"github.com/openshift/installer/pkg/terraform/stages"
+)
+
+// PlatformStages are the stages to run to provision the infrastructure in IBM Cloud.
+var PlatformStages = []terraform.Stage{
+	stages.NewStage("ibmcloud", "network"),
+	stages.NewStage("ibmcloud", "bootstrap", stages.WithNormalDestroy()),
+	stages.NewStage("ibmcloud", "master"),
```
jstuever, line 12: Is there a particular reason to split these into three stages? I was trying to determine that myself looking at these changes, but it wasn't really clear. In contrast, many of the other platforms are keeping "master" and "network" together.... and rolling them out before "bootstrap". What you have here looks like it works, and there's nothing logically wrong with it as far as I can tell. But, I did see some dependency declarations were deleted with regard to LBs, so I wanted to point this out as a possible alternative. See my PR for libvirt as reference:

https://github.com/openshift/installer/pull/5119/files#diff-d7310ec3465d4577b8f70742e7ecd5a22f463698d1aa807da17041613bcd486eR10

jstuever (commented):
Overall looks good. I did have a few questions.

jstuever:
/hold
Holding additional terraform stage changes until master opens for 4.10 development. While this work is necessary to fix other bugs, we don't want to overload QE with significant changes like this during feature freeze.

jstuever:
/hold cancel
Sometimes I forget master branch is protected by the bugzilla/valid-bug label requirement during feature freeze. No need to specify a hold on top of that.

pkg/asset/installconfig/ibmcloud/dns.go
```diff
@@ -46,6 +46,7 @@ func GetDNSZone() (*Zone, error) {
 		}
 		options = append(options, option)
 	}
+	sort.Strings(options)
```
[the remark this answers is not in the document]
hasueki: I discovered this bug while testing this PR, where the survey selection fails if the options are not sorted. I can gladly open a separate PR for this fix, if that is more favorable.

pkg/terraform/stages/ibmcloud/stages.go
hasueki: Ok thanks for the pointer - no reason particular really, just made logical sense for me to split them this way. Let me refactor this to just 2 stages and see if that plays nicely on IBM Cloud VPC.

hasueki: I have tried to refactor this to 2 stages (cluster, then bootstrap), and attempted a cluster creation, but is not behaving as expected.

From what I am seeing, while the master nodes are polling the MCS LB for its ignition, if there is an update to this LB (adding bootstrap to pool), the master nodes gets into a state where it cannot reach the network. And it seems the only way to recover from this, is rebooting the master nodes. This is an issue with IBM Cloud VPC.

I will try to get some attention on this from the appropriate folks, but in the interim, I propose we stick to the 3 stages (network, bootstrap, master) since we know this has been working pretty well thus far.

pkg/asset/installconfig/ibmcloud/dns.go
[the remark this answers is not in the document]
jstuever: This is fine here.

pkg/terraform/stages/ibmcloud/stages.go
jstuever: +1

PR #5119, opened 2021-08-02
libvirt: split up tf files into stages

jstuever:
Split the tf files for libvirt into two stages.
* cluster: Create the permanent infrastructure for the cluster.
* bootstrap: Create the bootstrap infrastructure.


data/data/libvirt/cluster/output.tf
```diff
@@ -0,0 +1,19 @@
...
+
+output "base_volume_id" {
+  value = libvirt_volume.coreos_base.id
+}
+
+output "network_id" {
+  value = libvirt_network.net.id
+}
+
+output "bootstrap_ip" {
```
jhixson74, line 13: This should probably be output in the bootstrap stage.

jhixson74:
This looks good to me from my limited perspective ;-)

data/data/libvirt/cluster/output.tf
jstuever: Good idea, and done. Ironically... these IPs are set in the installer, passed into Terraform, and then output back to the installer.

jstuever:
/hold cancel
I sometimes forget master is protected by bugzilla/valid-bug label.

PR #5121, opened 2021-08-02
Bug 1978213: openstack - relax value for minNetworkConstraint

EmilienM:
We currently document that a tenant should have ports quota set to 15
(or 1500 with Kuryr), but we didn't take in account the ports that are
used for the machines:

3 masters + 3 workers = 6 ports.

With these new numbers, we can deploy clusters with the existing
documentation with enough resources used from available quotas.


mandre (commented):
/lgtm

EmilienM:
/hold cancel

The Kuryr job failed on a known issue.

PR #5122, opened 2021-08-02
Bug 1989279: openstack: backport quotas/BYON improvements

EmilienM:
## Run kuryr-specific quota checks for kuryr envs

Due to a missing capitalization we didn't run the kuryr-specific network
quota checks for OpenStack validation which could result in the
installation starting even though the user has not nearly enough quota
for a successful Kuryr deployment.

## openstack: allow to skip quota checks

Allow to skip quota checks by re-using
OPENSHIFT_INSTALL_SKIP_PREFLIGHT_VALIDATIONS variable that has been used
for validations already.
When set to 1, we'll skip quota checks. Note that this is not suggested
and please use it at your own risks.

## openstack: relax quota checks in BYON

When deploying OCP with pre-provisioned networks, we do not need to have
available quotas for routers, networks and subnets for the machines
since they are created already.

This patch does the following:

* On Kuryr based deployment:
- Do not touch the quotas on networks / subnets, since we still need to
  have a bunch of quotas to create what's needed for Kuryr.
- Relax the quota on Routers, since it wouldn't create one in BYON mode.

* For other network types:
Relax routers, subnets, networks to be 0 by default and if we don't
configure OCP in BYON mode, then we add the necessary quotas (1 router,
1 network and 1 subnet).

## openstack - relax value for minNetworkConstraint

We currently document that a tenant should have ports quota set to 15
(or 1500 with Kuryr), but we didn't take in account the ports that are
used for the machines:

3 masters + 3 workers = 6 ports.

With these new numbers, we can deploy clusters with the existing
documentation with enough resources used from available quotas.

This is a combined cherry-pick to make the review easier, and the overall solves a single problem related to quotas.

### DIFF
--- a/pkg/asset/quota/openstack/openstack.go
+++ b/pkg/asset/quota/openstack/openstack.go
@@ -14,9 +14,10 @@ import (
 // These numbers should reflect what is documented here:
 // https://github.com/openshift/installer/tree/master/docs/user/openstack
 // https://github.com/openshift/installer/blob/master/docs/user/openstack/kuryr.md
-// Number or ports here don't include the constraints needed for each machine, which are calculated later
-var networkConstraint = buildNetworkConstraint(15, 1, 1, 1, 3, 60)
-var networkConstraintWithKuryr = buildNetworkConstraint(1500, 1, 250, 250, 250, 1000)
+// Number of ports, routers, subnets and routers here don't include the constraints needed
+// for each machine, which are calculated later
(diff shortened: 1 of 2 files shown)


mandre (commented):
/lgtm
/approve

EmilienM:
note for myself, I need to add https://github.com/openshift/installer/pull/5133

deads2k:
(patch manager)

choosing not to approve the cherrypick for this z-stream in order to prioritize more important fixes and reduce risk.

PR #5130, opened 2021-08-03
Record rpm-ostree status and history in log-bundle

rvanderp3:
During log-bundle analysis it should be trivial to find out which FCOS/RHCOS os image was used during installation.  

Fixes https://github.com/openshift/okd/issues/652


rvanderp3:
Installation completed; `e2e-aws` failed due to exceeding the dockerhub rate limit.

PR #5133, opened 2021-08-04
Bug 1987845: openstack: relax quotas with Kuryr

EmilienM:
We missed Kuryr when we relaxed networking quotas and decremented the
minimal constraints.


mandre (commented):
/lgtm
/approve

mandre (commented):
/lgtm
/hold for tests

EmilienM:
the openstack jobs fail because of limited quotas
/retest

EmilienM:
/retest
FYI @MaysaMacedo we have a problem on Kuryr cloud where the *-external routers were not cleaned, so no floating IP were available. I removed them all and now it should be fine. I'll investigate.

PR #5134, opened 2021-08-05
Server-Side Encryption and Encryption At Host support

m1kola:
## What this PR does

Adds support for:

* [Server-Side Encryption](https://docs.microsoft.com/en-us/azure/virtual-machines/disk-encryption) with customer managed keys
* [Encryption At Host](https://docs.microsoft.com/en-us/azure/virtual-machines/disk-encryption#encryption-at-host---end-to-end-encryption-for-your-vm-data) support

This PR is based on the work done for Azure Red Hat OpenShift and hopefully gives the installer team a good starting point for [CORS-1494](https://issues.redhat.com/browse/CORS-1494).

In the current form the PR allows to create clusters with server side encryption (customer managed keys) and encryption at host, but it requires a user to provision key vault and disk encryption set manually, set required permissions, etc (see the steps below). Basically It works similar to bringing own vnet.

UX for IPI can be improved by giving an option to provision key vault and disk encryption set for the user.

Also cluster's default storage class is not going to use customer managed keys which might be confusing/missleading as mentioned [here](https://issues.redhat.com/browse/CORS-1494?focusedCommentId=16497490&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16497490). The installer team might need to work together with the storage team to allow the default storage class to use disk encryption set straight after the installation (**OR** at least big warning is required somewhere in the documentation).

### Prerequisite: Enable EncryptionAtHost feature

```
az feature register --namespace "Microsoft.Compute" --name "EncryptionAtHost"
```

To check the status run

```
az feature show --namespace Microsoft.Compute --name EncryptionAtHost
```

Once the feature 'EncryptionAtHost' is registered, invoking 'az provider register -n Microsoft.Compute' is required to get the change propagated.

```
az provider register -n Microsoft.Compute
```

### Set basic variables

```
OUTPUT_DIR=<some-dir>   # Output dir for the installer assets directory (install config will be here)
RESOURCEGROUP=$USER-test-encryption
LOCATION=eastus

KEYVAULT_NAME=$USER-kv  # Needs to be globally unique
KEYVAULT_KEY_NAME=$USER-disk-encryption-key
DISK_ENCRYPTION_SET_NAME=$USER-disk-encryption-set

CLUSTER_SP_ID=<appId>   # Service principal ID which is going to be used to create a cluster
```

### Create resource group for key vault and disk encryption set

Resource group creation:

```
az group create --name $RESOURCEGROUP --location $LOCATION
```

### Create Key Vault

```
az keyvault create -n $KEYVAULT_NAME -g $RESOURCEGROUP -l $LOCATION --enable-purge-protection true --enable-soft-delete true

az keyvault key create --vault-name $KEYVAULT_NAME -n $KEYVAULT_KEY_NAME --protection software
```

### Create an instance of a DiskEncryptionSet

```
KEYVAULT_ID=$(az keyvault show --name $KEYVAULT_NAME --query "[id]" -o tsv)

KEYVAULT_KEY_URL=$(az keyvault key show --vault-name $KEYVAULT_NAME --name $KEYVAULT_KEY_NAME --query "[key.kid]" -o tsv)

az disk-encryption-set create -n $DISK_ENCRYPTION_SET_NAME -l $LOCATION -g $RESOURCEGROUP --source-vault $KEYVAULT_ID --key-url $KEYVAULT_KEY_URL
```

### Grant the DiskEncryptionSet resource access to the key vault:

```
DES_IDENTITY=$(az disk-encryption-set show -n $DISK_ENCRYPTION_SET_NAME -g $RESOURCEGROUP --query "[identity.principalId]" -o tsv)

az keyvault set-policy -n $KEYVAULT_NAME -g $RESOURCEGROUP --object-id $DES_IDENTITY --key-permissions wrapkey unwrapkey get
```

### Grant service principal Reader permissions to DiskEncryptionSet

```
DES_RESOURCE_ID=$(az disk-encryption-set show -n $DISK_ENCRYPTION_SET_NAME -g $RESOURCEGROUP --query "[id]" -o tsv)

az role assignment create --assignee $CLUSTER_SP_ID --role Reader --scope $DES_RESOURCE_ID -o jsonc
```

### Get disk encryption set ID

```
echo $DES_RESOURCE_ID
```

You will need to later so you can put it into the install config yaml.

### Generate and modify an install config

```
openshift-install create install-config --dir $OUTPUT_DIR
```

In the editor of choice open `$OUTPUT_DIR/install-config.yaml` and modify both `compute` and `controlPlane` sections:
replace `platform` with the following:

```yaml
  platform:
    azure:
      encryptionAtHost: true
      osDisk:
        diskEncryptionSetId: "<output of echo $DES_RESOURCE_ID from the previous step>"
```

### Create a cluster

```
openshift-install create cluster --dir $OUTPUT_DIR
```

### Validate that all VMs have correct properties:

```
CLUSTER_RESOURCEGROUP=<cluster-resource-group-provisioned-by-installer>
az vm list -g $CLUSTER_RESOURCEGROUP --query "[].{Name:name, diskEncryptionSet: storageProfile.osDisk.managedDisk.diskEncryptionSet.id, encryptionAtHost: securityProfile.encryptionAtHost}"
```

Note: you can run this command while bootstrap nodes still exist: they should have the same properties as set in `masterProfile`.

You should expect something like:

```
[
  {
    "Name": "foo-bar-master-0",
    "diskEncryptionSet": "<disk encryption set ID provided earlier>",
    "encryptionAtHost": true
  },
  # ...
]
```

### DIFF
--- a/data/data/azure/variables-azure.tf
+++ b/data/data/azure/variables-azure.tf
@@ -28,6 +28,12 @@ variable "azure_master_vm_type" {
   description = "Instance type for the master node(s). Example: `Standard_D8s_v3`."
 }
 
+variable "azure_master_disk_encryption_set_id" {
+  default = null
+  type = string
+  description = "The ID of the Disk Encryption Set which should be used to encrypt OS disk for the master node(s)."
+}
+
 variable "azure_extra_tags" {
   type = map(string)
 
--- a/data/data/azure/variables-azure.tf
+++ b/data/data/azure/variables-azure.tf
@@ -34,6 +34,12 @@ variable "azure_master_disk_encryption_set_id" {
   description = "The ID of the Disk Encryption Set which should be used to encrypt OS disk for the master node(s)."
 }
 
+variable "azure_master_encryption_at_host_enabled" {
+  default = false
+  type = bool
+  description = "Enables encryption at the VM host for the master node(s)."
+}
+
 variable "azure_extra_tags" {
   type = map(string)
 
--- a/data/data/azure/bootstrap/main.tf
+++ b/data/data/azure/bootstrap/main.tf
@@ -195,6 +195,7 @@ resource "azurerm_linux_virtual_machine" "bootstrap" {
   # included here because it is required by the Azure ARM API.
   admin_password                  = "NotActuallyApplied!"
   disable_password_authentication = false
+  encryption_at_host_enabled      = var.azure_master_encryption_at_host_enabled
 
   identity {
     type         = "UserAssigned"
--- a/data/data/azure/bootstrap/main.tf
+++ b/data/data/azure/bootstrap/main.tf
@@ -202,10 +202,11 @@ resource "azurerm_linux_virtual_machine" "bootstrap" {
   }
 
   os_disk {
-    name                 = "${var.cluster_id}-bootstrap_OSDisk" # os disk name needs to match cluster-api convention
-    caching              = "ReadWrite"
-    storage_account_type = "Premium_LRS"
-    disk_size_gb         = 100
+    name                   = "${var.cluster_id}-bootstrap_OSDisk" # os disk name needs to match cluster-api convention
+    caching                = "ReadWrite"
+    storage_account_type   = "Premium_LRS"
+    disk_size_gb           = 100
+    disk_encryption_set_id = var.azure_master_disk_encryption_set_id
   }
 
   source_image_id = var.vm_image
--- a/data/data/azure/cluster/main.tf
+++ b/data/data/azure/cluster/main.tf
@@ -29,26 +29,27 @@ provider "azureprivatedns" {
 }
 
 module "master" {
-  source                 = "./master"
-  resource_group_name    = var.resource_group_name
-  cluster_id             = var.cluster_id
-  region                 = var.azure_region
-  availability_zones     = var.azure_master_availability_zones
-  vm_size                = var.azure_master_vm_type
-  vm_image               = var.vm_image
-  identity               = var.identity
-  ignition               = var.ignition_master
-  elb_backend_pool_v4_id = var.elb_backend_pool_v4_id
-  elb_backend_pool_v6_id = var.elb_backend_pool_v6_id
-  ilb_backend_pool_v4_id = var.ilb_backend_pool_v4_id
-  ilb_backend_pool_v6_id = var.ilb_backend_pool_v6_id
-  subnet_id              = var.master_subnet_id
-  instance_count         = var.master_count
-  storage_account        = var.storage_account
-  os_volume_type         = var.azure_master_root_volume_type
-  os_volume_size         = var.azure_master_root_volume_size
-  private                = var.azure_private
-  outbound_udr           = var.azure_outbound_user_defined_routing
+  source                     = "./master"
+  resource_group_name        = var.resource_group_name
+  cluster_id                 = var.cluster_id
+  region                     = var.azure_region
+  availability_zones         = var.azure_master_availability_zones
+  vm_size                    = var.azure_master_vm_type
+  disk_encryption_set_id     = var.azure_master_disk_encryption_set_id
+  vm_image                   = var.vm_image
+  identity                   = var.identity
+  ignition                   = var.ignition_master
+  elb_backend_pool_v4_id     = var.elb_backend_pool_v4_id
+  elb_backend_pool_v6_id     = var.elb_backend_pool_v6_id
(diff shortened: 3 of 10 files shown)


m1kola:
@jhixson74 I think CI failures in `e2e-azure` and `e2e-azure-shared-vpc` were legit: azure terraform provider wasn't happy with empty string as a default value for for `disk_encryption_set_id` param. I think I addressed it now by using `null` as a variable's default value and using `omitempty` in relevant tfvars fields (`pkg/tfvars/azure/azure.go`).

But I would like to spend a bit more time to make sure that my changes make sense. Will ping you once I'm more confident that I covered everything.

/test e2e-azure
/test e2e-azure-upi
/test e2e-azure-shared-vpc

m1kola:
@jhixson74 I think this is ready for review. I see that azure e2es are failing, but I belive these failures are not related to this PR: I checked other builds and there are similar failures there. [`4.9.0-0.nightly-2021-08-06-060446`](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.9-e2e-azure/1423526010262917120), for example, failed with the same/very similar reason.

I did some manual testing with [`OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=registry.ci.openshift.org/ocp/release:4.9.0-0.nightly-2021-08-03-135217`](https://amd64.ocp.releases.ci.openshift.org/releasestream/4.9.0-0.nightly/release/4.9.0-0.nightly-2021-08-03-135217) and was able to install a cluster successfully.

jhixson74:
> @jhixson74 @staebler could you please take a look at this? Any feedback?

Everything looks good to me, I just can't seem to get it to work for me. I will give it another run and see how it goes.

staebler:
> Also cluster's default storage class is not going to use customer managed keys which might be confusing/missleading as mentioned [here](https://issues.redhat.com/browse/CORS-1494?focusedCommentId=16497490&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16497490). The installer team might need to work together with the storage team to allow the default storage class to use disk encryption set straight after the installation (**OR** at least big warning is required somewhere in the documentation).

If I recall correctly, this is the same behavior in AWS and GCP. The disk encryption for the nodes only effects the node's disks and not cluster storage.

jhixson74:
@m1kola I've tested and confirmed this works. I have a local branch that I've updated. Can you update this or shall I create a PR with the updates?

m1kola:
Rebased on top of master to fix conflicts. @jhixson74 please take another look.


/test e2e-azure
/test e2e-azure-upi
/test e2e-azure-shared-vpc

jhixson74 (approved):
/lgtm

staebler (commented):
We need to handle Azure Stack as well. Either Azure Stack supports this, and the terraform for Azure Stack needs to be updated. Or Azure Stack does not support this, and we need validation to reject setting the field when installing on Azure Stack.

/lgtm cancel

patrickdillon:
> We need to handle Azure Stack as well. Either Azure Stack supports this, and the terraform for Azure Stack needs to be updated. Or Azure Stack does not support this, and we need validation to reject setting the field when installing on Azure Stack.

Azure Stack [supports encryption through Bitlocker](https://docs.microsoft.com/en-us/azure-stack/user/azure-stack-managed-disk-considerations?view=azs-2108&tabs=az1%2Caz2) but I do not see this encryption supported in the [Azure Stack Terraform provider](https://registry.terraform.io/providers/hashicorp/azurestack/latest/docs/resources/managed_disk). So I suggest we add validation to reject the field in Azure Stack

data/data/azure/variables-azure.tf
```diff
@@ -28,6 +28,18 @@ variable "azure_master_vm_type" {
   description = "Instance type for the master node(s). Example: `Standard_D8s_v3`."
 }
 
+variable "azure_master_disk_encryption_set_id" {
+  default = null
```
patrickdillon, line 32: We have decided to no longer provide default values in Terraform configs.

data/data/azure/variables-azure.tf
```diff
@@ -28,6 +28,18 @@ variable "azure_master_vm_type" {
...
+  type = string
+  description = "The ID of the Disk Encryption Set which should be used to encrypt OS disk for the master node(s)."
+}
+
+variable "azure_master_encryption_at_host_enabled" {
+  default = false
```
patrickdillon, line 38: We have decided to no longer provide default values in Terraform configs.

PR #5138, opened 2021-08-06
Azure Stack Hub Manual Credentials

patrickdillon:
This sets up Azure Stack Hub to use manual credentials mode. It also reverts a prior commit to include credentials in the cloud provider config. 

cc @jstuever @hasueki we probably want to do something with IBM Cloud for setting the credentials mode, but let's not copy the pattern in this PR without review first. I don't think we have a precedent for setting default Manual credentials mode.


pkg/types/defaults/installconfig.go
```diff
@@ -80,6 +81,9 @@ func SetInstallConfigDefaults(c *types.InstallConfig) {
 		awsdefaults.SetPlatformDefaults(c.Platform.AWS)
 	case c.Platform.Azure != nil:
 		azuredefaults.SetPlatformDefaults(c.Platform.Azure)
+		if c.Platform.Azure.CloudName == azure.StackCloud {
```
patrickdillon, line 84: this should also check if creds are already set

jstuever: Is there a reason not to put this inside `azuredefaults.SetPlatformDefaults`?

pkg/asset/manifests/azure/cloudproviderconfig.go
```diff
@@ -71,9 +69,6 @@ func (params CloudProviderConfig) JSON() (string, error) {
 	}
 
 	if params.CloudName == azure.StackCloud {
-		config.authConfig.AADClientID = params.AADClientID
-		config.authConfig.AADClientSecret = params.AADClientSecret
-		config.authConfig.ResourceManagerEndpoint = params.ResourceManagerEndpoint
```
lobziik, line 76: Can we leave this parameter in cloud config?

PR #5140, opened 2021-08-09
Bug 1989917: openstack: relax Security Group quotas

EmilienM:
OpenStack clouds deployed by Red Hat have a `default` security group
which contains 4 rules.

This patch will relax the number of security groups and rules so it
takes in account that a number of these resources could already be there
before installing OpenShift.

This is fine because the installer will create 2 security groups (one
for masters and one for workers, when using OpenShiftSDN, etc) and less
than 45 rules, so 56 rules is enough.
Same with Kuryr, new numbers will be enough as well.

This will allow our users following the documentation and the quotas, to
deploy on Red Hat OpenStack without modifying the quota numbers.


pierreprinetti:
Merging is blocked by https://github.com/openshift/release/pull/21001

PR #5147, opened 2021-08-12
Possible solution for BZ#1992777

pamoedom:
My 2 cents for BZ#1992777, please review when you have a moment, thanks.

### DIFF
--- a/pkg/destroy/ibmcloud/cloudobjectstorage.go
+++ b/pkg/destroy/ibmcloud/cloudobjectstorage.go
@@ -3,6 +3,7 @@ package ibmcloud
 import (
 	"fmt"
 	"net/http"
+	"strings"
 
 	"github.com/pkg/errors"
 )
@@ -118,8 +119,13 @@ func (o *ClusterUninstaller) COSInstanceID() (string, error) {
 	// Locate the installer's COS instance by name.
 	for _, instance := range instanceList {
 		if instance.name == fmt.Sprintf("%s-cos", o.InfraID) {
-			o.cosInstanceID = instance.id
-			return instance.id, nil
+                        split := strings.Split(instance.id, ":")
+                        if len(split) >= 8 { //CRN string
+                                o.cosInstanceID = split[7]
+                        } else {
+                                o.cosInstanceID = instance.id
+                        }
+                        return o.cosInstanceID, nil
 		}
 	}
 	return "", errors.Errorf("COS instance not found")
--- a/data/data/ibmcloud/image/main.tf
+++ b/data/data/ibmcloud/image/main.tf
@@ -21,7 +21,7 @@ resource "ibm_iam_authorization_policy" "policy" {
   source_service_name         = "is"
   source_resource_type        = "image"
   target_service_name         = "cloud-object-storage"
-  target_resource_instance_id = var.cos_resource_instance_id
+  target_resource_instance_id = length(split(":", var.cos_resource_instance_id)) >= 8 ? "${element(split(":", var.cos_resource_instance_id),7)}" : var.cos_resource_instance_id
   roles                       = ["Reader"]
 }
 


pamoedom:
@hasueki, please review this proposed fix for BZ#1992777 when you have a moment, thanks.

data/data/ibmcloud/network/image/main.tf
```diff
@@ -21,7 +21,7 @@ resource "ibm_iam_authorization_policy" "policy" {
   source_service_name         = "is"
   source_resource_type        = "image"
   target_service_name         = "cloud-object-storage"
-  target_resource_instance_id = var.cos_resource_instance_id
+  target_resource_instance_id = length(split(":", var.cos_resource_instance_id)) >= 8 ? "${element(split(":", var.cos_resource_instance_id),7)}" : var.cos_resource_instance_id
```
hasueki, line 24: Looks like you're parsing out the instance guid out of the CRN. Instead of doing this here, you can just update, https://github.com/pamoedom/installer/blob/1398598c915359c9b1f3f927a9397bacc57ea93a/data/data/ibmcloud/network/outputs.tf#L6, to..
```
value = ibm_resource_instance.cos.guid
```

hasueki:
@pamoedom I realized that it's not setting..

image

.. in `ibm_iam_authorization_policy.policy`. I wonder if that's causing the issue, in cases where users do not have access to all resource groups in the account.

Reference: https://registry.terraform.io/providers/IBM-Cloud/ibm/latest/docs/resources/iam_authorization_policy#source_resource_group_id

data/data/ibmcloud/network/image/main.tf
pamoedom: ACK, I'll check that ASAP

pamoedom:
> @pamoedom I realized that it's not setting..
> 
> image
> 
> .. in `ibm_iam_authorization_policy.policy`. I wonder if that's causing the issue, in cases where users do not have access to all resource groups in the account.
> 
> Reference: https://registry.terraform.io/providers/IBM-Cloud/ibm/latest/docs/resources/iam_authorization_policy#source_resource_group_id

That's optional, I recall trying also with those parameters and the whole CRN and no luck, the only thing that worked is just setting a wide policy to all resources in the same resource_group, like the following:
~~~
resource "ibm_iam_authorization_policy" "policy" {
  source_service_name         = "is"
  source_resource_type        = "image"
  target_service_name         = "cloud-object-storage"
  target_resource_group_id    = var.resource_group_id
  roles                       = ["Reader"]
}
~~~

But that policy is too wide and can conflict with previous existing one and abort the installation

data/data/ibmcloud/network/image/main.tf
pamoedom: Same error with that change:
~~~
[...]
DEBUG module.vpc.ibm_is_lb_listener.machine_config: Creation complete after 2m10s [id=r010-70459343-175b-42db-87f3-10377d459269/r010-119f5451-3a13-4b57-aa44-972073c2234e] 
ERROR                                              
ERROR Error: [DEBUG] Image creation err The request is not authorized to access the Cloud Object Storage resource. 
ERROR {                                            
ERROR     "StatusCode": 403,                       
ERROR     "Headers": {                             
ERROR         "Cache-Control": [                   
ERROR             "max-age=0, no-cache, no-store, must-revalidate" 
ERROR         ],                                   
ERROR         "Cf-Cache-Status": [                 
ERROR             "DYNAMIC"                        
ERROR         ],                                   
ERROR         "Cf-Ray": [                          
ERROR             "6888a5695f7fd665-MAD"           
ERROR         ],                                   
ERROR         "Connection": [                      
ERROR             "keep-alive"                     
ERROR         ],                                   
ERROR         "Content-Type": [                    
ERROR             "application/json; charset=utf-8" 
ERROR         ],                                   
ERROR         "Date": [                            
ERROR             "Thu, 02 Sep 2021 18:01:55 GMT"  
ERROR         ],                                   
ERROR         "Expect-Ct": [                       
ERROR             "max-age=604800, report-uri=\"https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct\"" 
ERROR         ],                                   
ERROR         "Expires": [                         
ERROR             "-1"                             
ERROR         ],                                   
ERROR         "Pragma": [                          
ERROR             "no-cache"                       
ERROR         ],                                   
ERROR         "Server": [                          
ERROR             "cloudflare"                     
ERROR         ],                                   
ERROR         "Strict-Transport-Security": [       
ERROR             "max-age=31536000; includeSubDomains" 
ERROR         ],                                   
ERROR         "Transaction-Id": [                  
ERROR             "7d4092b6d1f533314cce56897047d7e2" 
ERROR         ],                                   
ERROR         "Vary": [                            
ERROR             "Accept-Encoding"                
ERROR         ],                                   
ERROR         "X-Content-Type-Options": [          
ERROR             "nosniff"                        
ERROR         ],                                   
ERROR         "X-Request-Id": [                    
ERROR             "7d4092b6d1f533314cce56897047d7e2" 
ERROR         ],                                   
ERROR         "X-Xss-Protection": [                
ERROR             "1; mode=block"                  
ERROR         ]                                    
ERROR     },                                       
ERROR     "Result": {                              
ERROR         "errors": [                          
ERROR             {                                
ERROR                 "code": "cos_not_authorized", 
ERROR                 "message": "The request is not authorized to access the Cloud Object Storage resource.", 
ERROR                 "more_info": "http://cloud.ibm.com/docs/vpc-on-classic?topic=vpc-on-classic-rias-error-messages#cos_not_authorized", 
ERROR                 "target": {                  
ERROR                     "name": "file.href",     
ERROR                     "type": "field"          
ERROR                 }                            
ERROR             }                                
ERROR         ],                                   
ERROR         "trace": "7d4092b6d1f533314cce56897047d7e2" 
ERROR     },                                       
ERROR     "RawResult": null                        
ERROR }                                            
ERROR                                              
ERROR                                              
ERROR   on ../../../../tmp/openshift-install-network-328621160/image/main.tf line 28, in resource "ibm_is_image" "image": 
ERROR   28: resource "ibm_is_image" "image" {      
ERROR                                              
ERROR                                              
FATAL failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to apply Terraform: failed to complete the change
~~~
Are you sure that the short ID that I'm parsing from the CRN is the `cos.guid`? I think is more like the individual resourceID and I'm not sure how to obtain that from the `ibm_resource_instance.cos` object.

pamoedom: Hi again @hasueki, any other suggestion about this issue? we are still suffering the same limitation within our ibmcloud account when launching the installer and I'm afraid this problem will also appear when the end customers start using it more widely.

hasueki: Hi @pamoedom

> Are you sure that the short ID that I'm parsing from the CRN is the cos.guid? I think is more like the individual resourceID and I'm not sure how to obtain that from the ibm_resource_instance.cos object.

Yes, those values should be equivalent.

> any other suggestion about this issue?...

Let me try to recreate this behavior on my side and see if I can spot the issue. I will get back to you.

hasueki: @pamoedom Did a little investigation, and I was able to reproduce this issue. It turns out that my original [suggestion](https://github.com/openshift/installer/pull/5147/files#r701121468) will not work because `ibm_cos_bucket.images.resource_instance_id` requires a CRN value and `ibm_iam_authorization_policy.policy.target_resource_instance_id` required a GUID, but it is using the same TF variable.

With that being said, I'm ok with your proposed approach by parsing out the GUID from the CRN, so we can avoid passing in two TF variables (one for CRN, one for GUID). It may be a nitpick, but the only thing I would suggest is to change the `cos_resource_instance_id` input variable in the `images` module to `cos_resource_instance_crn` so it is clear that this variable is in fact a CRN value. Then you can just remove `length(split(":", var.cos_resource_instance_id)) >= 8 ?` that is checking if this variable is a CRN string.

pkg/destroy/ibmcloud/cloudobjectstorage.go
```diff
@@ -119,8 +120,13 @@ func (o *ClusterUninstaller) COSInstanceID() (string, error) {
...
 		if instance.name == fmt.Sprintf("%s-cos", o.InfraID) {
-			o.cosInstanceID = instance.id
-			return instance.id, nil
+                        split := strings.Split(instance.id, ":")
+                        if len(split) >= 8 { //CRN string
+                                o.cosInstanceID = split[7]
+                        } else {
+                                o.cosInstanceID = instance.id
+                        }
+                        return o.cosInstanceID, nil
```
hasueki, line 129: @pamoedom Was there a reason you had add this logic to grab the instance guid on destroy?

data/data/ibmcloud/network/image/main.tf
pamoedom: ACK @hasueki, thanks for the confirmation and the tips, I'll try your suggestion ASAP.

pkg/destroy/ibmcloud/cloudobjectstorage.go
pamoedom: If I recall correctly, I did that to homogenize the code based on the CRN parsing but not sure if this step is really necessary, let me try again removing it, thks.

data/data/ibmcloud/network/image/main.tf
pamoedom: @hasueki, the changes work as expected, however please note that I had to change more files than expected, not just the `network/image` module because the same variable is used within bootstrap and master modules depending on the value generated by `data/data/ibmcloud/network/image/outputs.tf`.

pkg/destroy/ibmcloud/cloudobjectstorage.go
pamoedom: Ok, I've tested the above changes of `cos_resource_instance_crn` but restoring `pkg/destroy/ibmcloud/cloudobjectstorage.go` to its previous content and I have something relevant to share, the authorization is not being removed on this manner because we are using now the GUID parsed from the CRN, therefore the destroy operation also needs to use the same value and not the whole CRN string.

pamoedom: @hasueki, what do you think could be a better approach to properly remove the authorization? parsing again the GUID at `pkg/destroy/ibmcloud/cloudobjectstorage.go` like I was doing or modify the `resourceMatches` parameters at `pkg/destroy/ibmcloud/iam.go`?

data/data/ibmcloud/network/main.tf
```diff
@@ -35,13 +35,13 @@ resource "ibm_resource_instance" "cos" {
...
-  resource_group_id        = local.resource_group_id
-  tags                     = local.tags
-  cos_resource_instance_id = ibm_resource_instance.cos.id
+  name                      = "${var.cluster_id}-rhcos"
+  image_filepath            = var.ibmcloud_image_filepath
+  cluster_id                = var.cluster_id
+  region                    = var.ibmcloud_region
+  resource_group_id         = local.resource_group_id
+  tags                      = local.tags
+  cos_resource_instance_crn = ibm_resource_instance.cos.id
```
hasueki, line 44:
```suggestion
  cos_resource_instance_crn = ibm_resource_instance.cos.crn
```

pkg/destroy/ibmcloud/cloudobjectstorage.go
hasueki: @pamoedom Ah gotcha. I would just parse it in `pkg/destroy/ibmcloud/iam.go`. So it's clear that IAM authorization explicitly uses a GUID.

pamoedom: ACK, let me try to modify `pkg/destroy/ibmcloud/iam.go` accordingly and test again.

data/data/ibmcloud/network/main.tf
pamoedom: ACK, let me change it.

data/data/ibmcloud/network/outputs.tf
```diff
@@ -2,7 +2,7 @@
 # Network module outputs
 #######################################
 
-output "cos_resource_instance_id" {
+output "cos_resource_instance_crn" {
   value = ibm_resource_instance.cos.id
```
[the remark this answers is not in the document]
pamoedom: ACK, let me change it.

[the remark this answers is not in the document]
pamoedom: Changed as suggested and working as expected ;)

pkg/destroy/ibmcloud/cloudobjectstorage.go
pamoedom: @hasueki, I've done the following simply change at `pkg/destroy/ibmcloud/iam.go` that does the trick:
~~~
strings.Contains(cosInstanceID, *policy.Resources[0].Attributes[2].Value) &&
~~~
Instead of string comparing the `Value` against `cosInstanceID`, I'm doing it on reverse because the value is now just the GUID.
Destroy operation works now as expected and IAM auth is also removed.

hasueki (approved):

jstuever (approved):
/lgtm
/approve

PR #5148, opened 2021-08-12
WIP - Bug 1993364: openstack/destroy: fix Kuryr/BYON

EmilienM:
* Relax the rule that lists the ports to figure out which
  network needs to be removed from the router. Now we list
  all ports that are tagged with the cluster name, and are not
  tagged with "cluster-api-provider-openstack".
  Before we were filtering the ports created by Nova for the
  machines, but we were missing the case of other ports
  (network gateways), when the machine was already destroyed.

* Cleanup networks that are tagged with the cluster name (instead of
  not tagged). I'm not sure why the logic was like this before
  but this was not clearing the networks from the router created by the
  operator (BYON).

### DIFF
--- a/pkg/destroy/openstack/openstack.go
+++ b/pkg/destroy/openstack/openstack.go
@@ -89,30 +89,25 @@ func (o *ClusterUninstaller) Run() error {
 	// deleteFuncs contains the functions that will be launched as
 	// goroutines.
 	deleteFuncs := map[string]deleteFunc{
-		"deleteServers":          deleteServers,
-		"deleteServerGroups":     deleteServerGroups,
-		"deleteTrunks":           deleteTrunks,
-		"deleteLoadBalancers":    deleteLoadBalancers,
-		"deletePorts":            deletePorts,
-		"deleteSecurityGroups":   deleteSecurityGroups,
-		"clearRoutersInterfaces": clearRoutersInferfaces,
-		"deleteSubnets":          deleteSubnets,
-		"deleteSubnetPools":      deleteSubnetPools,
-		"deleteNetworks":         deleteNetworks,
-		"deleteContainers":       deleteContainers,
-		"deleteVolumes":          deleteVolumes,
-		"deleteShares":           deleteShares,
-		"deleteFloatingIPs":      deleteFloatingIPs,
-		"deleteImages":           deleteImages,
+		"deleteServers":         deleteServers,
+		"deleteServerGroups":    deleteServerGroups,
+		"deleteTrunks":          deleteTrunks,
+		"deleteLoadBalancers":   deleteLoadBalancers,
+		"deletePorts":           deletePorts,
+		"deleteSecurityGroups":  deleteSecurityGroups,
+		"clearRouterInterfaces": clearRouterInterfaces,
+		"deleteSubnets":         deleteSubnets,
+		"deleteSubnetPools":     deleteSubnetPools,
+		"deleteNetworks":        deleteNetworks,
+		"deleteContainers":      deleteContainers,
+		"deleteVolumes":         deleteVolumes,
+		"deleteShares":          deleteShares,
+		"deleteFloatingIPs":     deleteFloatingIPs,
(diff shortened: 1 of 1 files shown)


EmilienM:
/hold

I'm not sure about this one, ping @MaysaMacedo for review & testing.

EmilienM:
Unfortunately, our BYON CI job isn't using Kuryr, so we'll have to do manual testing for this one.
Note that this scenario will be enabled via https://github.com/openshift/release/pull/21092.

pkg/destroy/openstack/openstack.go
```diff
@@ -624,15 +624,14 @@ func deleteCustomRouterInterfaces(opts *clientconfig.ClientOpts, filter Filter,
...
 
-	// Find the machines Network by the master machines Ports.
+	// Find the ports that should be removed from the router.
 	// Any worker nodes ports, including the ones used for additionalNetworks,
 	// are tagged with cluster-api-provider-openstack and should be excluded
 	// to ensure only the primary Network ports are filtered.
 	tags := filterTags(filter)
 	listOpts := ports.ListOpts{
-		Tags:        strings.Join(tags, ","),
-		DeviceOwner: "compute:nova",
```
MaysaMacedo, line 634: Probably the reason this code didn't work on the CI is that the device-owner is `compute:central` instead of `compute:nova`

pkg/destroy/openstack/openstack.go
```diff
@@ -651,10 +650,10 @@ func deleteCustomRouterInterfaces(opts *clientconfig.ClientOpts, filter Filter,
 		return true, nil
 	}
 
-	// Only proceed with clean up if is a provided Network
+	// Only proceed with clean up if the network was created by OpenShift
 	networkListOpts := networks.ListOpts{
-		NotTags: strings.Join(tags, ","),
```
MaysaMacedo, line 656: Removing it will remove the provided Network and Router

pkg/destroy/openstack/openstack.go
```diff
@@ -624,15 +624,14 @@ func deleteCustomRouterInterfaces(opts *clientconfig.ClientOpts, filter Filter,
...
 	// Any worker nodes ports, including the ones used for additionalNetworks,
 	// are tagged with cluster-api-provider-openstack and should be excluded
 	// to ensure only the primary Network ports are filtered.
 	tags := filterTags(filter)
 	listOpts := ports.ListOpts{
-		Tags:        strings.Join(tags, ","),
-		DeviceOwner: "compute:nova",
-		NotTags:     "cluster-api-provider-openstack",
+		Tags:    strings.Join(tags, ","),
+		NotTags: "cluster-api-provider-openstack",
```
MaysaMacedo, line 634: This is needed to avoid handling a possible secondary Network on the Worker Nodes, so we fetch only master Ports and grab Network from there

pkg/destroy/openstack/openstack.go
EmilienM: yeah we don't want that for sure.

dulek:
Seeing this touching router deletions - I'd suggest testing it with OpenShiftSDN and LoadBalancer services to make sure it still correctly removes FIPs attached to them.

pkg/destroy/openstack/openstack.go
```diff
@@ -669,12 +610,9 @@ func deleteCustomRouterInterfaces(opts *clientconfig.ClientOpts, filter Filter,
...
 	}
 
-	if len(allNetworks) == 0 {
-		return true, nil
-	}
-
 	portListOpts := ports.ListOpts{
-		NetworkID: allServerPorts[0].NetworkID,
+		NetworkID:   allNetworks[0].ID,
+		DeviceOwner: "network:router_interface",
```
MaysaMacedo, line 615: in osp 13 the deviceowner is `network:ha_router_replicated_interface`

MaysaMacedo:
created another PR for updating the PR title https://github.com/openshift/installer/pull/5191

PR #5149, opened 2021-08-16
images: Add az cli for CI ash testing

rna-afk:
In order to do IPI testing, login to the Azure Stack Hub is necessary
as it is private and not accessible ordinarily. Installing az cli
in the IPI image to allow for login to the Azure Stack Hub.

### DIFF
--- a/images/installer/Dockerfile.ci
+++ b/images/installer/Dockerfile.ci
@@ -8,6 +8,9 @@ COPY . .
 RUN DEFAULT_ARCH="$(go env GOHOSTARCH)" hack/build.sh
 RUN go run -mod=vendor hack/build-coreos-manifest.go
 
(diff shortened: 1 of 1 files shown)


patrickdillon:
/hold

We can't add any external dependencies to the base installer image.

> login to the Azure Stack Hub is necessary
as it is private and not accessible ordinarily. Installing az cli
in the IPI image to allow for login to the Azure Stack Hub. 

Also are you sure this is accurate? I think this is only a requirement for UPI.

rna-afk:
I'll try again but I vaguely remember something about subscription not found error being reported if I didn't login.

rna-afk:
Seems to work without the az client now. Thanks, I'll close this PR.

PR #5151, opened 2021-08-16
Metrics: Add Metrics API for installer telemetry data gathering

rna-afk:
Adds the main functions that the installer will call to create metrics,
store information and push to any endpoint required. The system has
an opt out policy where if the user sets the OPENSHIFT_INSTALL_DISABLE_METRICS
environment flag, the metrics will not be sent.

The user can also change the endpoint by setting the environment
variable OPENSHIFT_INSTALL_METRICS_ENDPOINT to a required URL
endpoint and the system will redirect the metrics.

The system creates the metrics when the values are being sent and pushes them
only if the value/labels are set to them.

The configuration right now is to switch the system on only when the URL is
set.

### DIFF
--- a/cmd/openshift-install/create.go
+++ b/cmd/openshift-install/create.go
@@ -33,6 +33,7 @@ import (
 	targetassets "github.com/openshift/installer/pkg/asset/targets"
 	destroybootstrap "github.com/openshift/installer/pkg/destroy/bootstrap"
 	"github.com/openshift/installer/pkg/gather/service"
+	"github.com/openshift/installer/pkg/metrics/gatherer"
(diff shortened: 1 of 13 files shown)


bwplotka:
I think we are missing higher level decision on what pipeline we want to use. Let's postpone this work until we do proposal. WDYT? (https://issues.redhat.com/browse/MON-1583?focusedCommentId=18926376&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-18926376)

rna-afk:
This is mostly collecting metrics from the installer. The pipeline part of the code is already in so we need to change that. This code will be disabled anyway until we figure out what pipeline we need. I guess this can be pushed too.

PR #5153, opened 2021-08-17
Bug 1978213: openstack/quota: relax min ports

EmilienM:
For Kuryr:
Reduce from 1493 to 1490:
* 6 ports taken by machines (and added later in the requirements)
* 2 for floating IPs
* 1 for router gateway
* 1 for bootstrap machine
= 10
1500 - 10 = 1490
So we have now 1490 as a new relaxed number which should satisfy the
documented quotas.

For non-Kuryr:
Reduce from 8 to 4:
* 1 taken by network gateway
* 2 taken by API & Ingress ports or FIPs
* 1 taken by bootstrap
-> this makes 4
Then the installer adds one port per machine, so 6 by default (3 masters
and 3 workers) which makes 10 total. The documented quota is 15 so have
some margin error.

### DIFF
--- a/pkg/asset/quota/openstack/openstack.go
+++ b/pkg/asset/quota/openstack/openstack.go
@@ -16,8 +16,8 @@ import (
 // https://github.com/openshift/installer/blob/master/docs/user/openstack/kuryr.md
 // Number of ports, routers, subnets and routers here don't include the constraints needed
 // for each machine, which are calculated later
-var minNetworkConstraint = buildNetworkConstraint(8, 0, 0, 0, 2, 56)
-var minNetworkConstraintWithKuryr = buildNetworkConstraint(1493, 0, 249, 249, 249, 996)
+var minNetworkConstraint = buildNetworkConstraint(4, 0, 0, 0, 2, 56)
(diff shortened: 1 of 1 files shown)


iamemilio:
I don't think we should allow customers to install with less than the quota required for 3 masters and 3 workers. I believe that is the minimum supported install that we recommend.

EmilienM:
> I don't think we should allow customers to install with less than the quota required for 3 masters and 3 workers. I believe that is the minimum supported install that we recommend.

With these numbers, we DO NOT allow less than the quota required for 3 masters and 3 workers:
e.g. with non-Kuryr, we document a minimum of ports quota set to 15.
We need to reserve 4 ports for:
* keep one port for bootstrap
* keep two ports for API & Ingress FIPs or VIPs
* keep one port for Router gateway (needed to plug to the external network)
Then later in code we bump numbers for machines 3 + 3 = 6 then 4 + 6 = 10.
Official doc recommends 15 available ports, which gives us some margin (e.g. disconnected installs, etc).

For Kuryr, we just do 1500 (documented number) - 10 = 1490.

These numbers were tested by QE and they seem good enough for now.

iamemilio:
Ahh ok, I see. LGTM then. 
/lgtm

PR #5154, opened 2021-08-17
Bug 1958154: Restrict number of AWS user tags

rna-afk:
AWS S3 bucket does not allow more than 10 user tags to be applied
to them. Since the installer adds two tags, the "Name" and the
"kubernetes.io" tags, the user specified tags should only be a maximum
of 8.

Adding this restriction to the validation of user tags to prevent
error from terraform during cluster creation.


jstuever (approved):
/lgtm
/approve

PR #5155, opened 2021-08-18
Bug 1974640: Write user credentials to specified env location

rna-afk:
If the user specified the AWS_SHARED_CREDENTIALS_FILE which is
either invalid or empty file, the credentials are collected by
the installer and are written to a default location (~/.aws/credentials)
when it can be written to the location specified in the above location.

Adding a small fix to write to the location specified in the location
specified by AWS_SHARED_CREDENTIALS_FILE if the file is either empty
or is invalid.

### DIFF
--- a/pkg/asset/installconfig/aws/session.go
+++ b/pkg/asset/installconfig/aws/session.go
@@ -1,6 +1,7 @@
 package aws
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -190,6 +191,9 @@ func getUserCredentials() error {
 	}
 
 	path := defaults.SharedCredentialsFilename()
+	if env := os.Getenv("AWS_SHARED_CREDENTIALS_FILE"); env != "" {
+		path = env
+	}
 	logrus.Infof("Writing AWS credentials to %q (https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html)", path)
 	err = os.MkdirAll(filepath.Dir(path), 0700)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/asset/installconfig/aws/session.go
```diff
@@ -198,12 +202,14 @@ func getUserCredentials() error {
 
 	creds, err := ini.Load(path)
 	if err != nil {
-		if os.IsNotExist(err) {
-			creds = ini.Empty()
-			creds.Section("").Comment = "https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html"
-		} else {
-			return err
+		if !os.IsNotExist(err) {
```
jhixson74, line 205: How about using os.IsExist(err) here instead of negating os.IsNotExist()?

jhixson74 (commented):
/lgtm

pkg/asset/installconfig/aws/session.go
rna-afk: Yeah that's better thanks!

rna-afk: Will the error message contain that information? I was under the impression that ini.Load(path) would only return `doesn't exist` and we can only check for that.

jhixson74: Looks like you are right. My mistake ;-)

pkg/asset/installconfig/aws/session.go
```diff
@@ -198,12 +202,14 @@ func getUserCredentials() error {
...
+			if err2 := os.Remove(path); err2 != nil {
```
patrickdillon, line 206: If IIUC, this will (attempt to) remove a file at `AWS_SHARED_CREDENTIALS_FILE` if it is not a valid ini file. I don't think we should be deleting user's data. It would be safer to bomb out and return a good error message so the user can correct the issue.

pkg/asset/installconfig/aws/session.go
```diff
@@ -198,12 +202,14 @@ func getUserCredentials() error {
...
+				logrus.Error(errors.Wrap(err2, fmt.Sprintf("failed to overwrite invalid file %s", path)))
```
patrickdillon, line 207: I think we should just return the error. We don't need to log it here.

pkg/asset/installconfig/aws/session.go
```diff
@@ -198,12 +202,14 @@ func getUserCredentials() error {
...
+				return err
```
patrickdillon, line 208: Should this return `err`, `err2` or both?

patrickdillon (commented):
I'm fine with this approach of writing to the value of `AWS_SHARED_CREDENTIALS_FILE` as long as we remove the part where we delete an existing file.

pkg/asset/installconfig/aws/session.go
rna-afk: Was thinking of logging err2 and returning err.

PR #5158, opened 2021-08-18
pkg/destroy/gcp: report cluster footprint in quota

stevekuznetsov:
When we destroy an entity in the public cloud relating to a cluster, we
now record the impact that the item had to the quota in the account that
the cluster was provisioned in. This will allow for downstream users of
the installer to reason about the footprint of clusters they run,
allowing for more automated reasoning about how many clusters of a type
can fit into an account.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/assign @jstuever 

I *really* don't know how to test this ... thoughts?

### DIFF
--- a/cmd/openshift-install/destroy.go
+++ b/cmd/openshift-install/destroy.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/openshift/installer/pkg/asset/cluster"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -21,7 +22,7 @@ import (
 	_ "github.com/openshift/installer/pkg/destroy/openstack"
 	_ "github.com/openshift/installer/pkg/destroy/ovirt"
 	_ "github.com/openshift/installer/pkg/destroy/vsphere"
-	timer "github.com/openshift/installer/pkg/metrics/timer"
+	"github.com/openshift/installer/pkg/metrics/timer"
 )
 
 func newDestroyCmd() *cobra.Command {
@@ -39,7 +40,8 @@ func newDestroyCmd() *cobra.Command {
 }
 
 func newDestroyClusterCmd() *cobra.Command {
-	return &cobra.Command{
+	var reportQuota bool
+	cmd := &cobra.Command{
 		Use:   "cluster",
 		Short: "Destroy an OpenShift cluster",
 		Args:  cobra.ExactArgs(0),
@@ -47,22 +49,33 @@ func newDestroyClusterCmd() *cobra.Command {
 			cleanup := setupFileHook(rootOpts.dir)
 			defer cleanup()
 
-			err := runDestroyCmd(rootOpts.dir)
+			err := runDestroyCmd(rootOpts.dir, reportQuota)
 			if err != nil {
 				logrus.Fatal(err)
 			}
 		},
 	}
+	quotaFlag := "report-quota-footprint"
+	cmd.Flags().BoolVar(&reportQuota, quotaFlag, false, "Report the quota footprint that the cluster had.")
+	_ = cmd.Flags().MarkHidden(quotaFlag) // only error possible is the flag does not exist, but we just added it
+	return cmd
 }
 
-func runDestroyCmd(directory string) error {
+func runDestroyCmd(directory string, reportQuota bool) error {
 	timer.StartTimer(timer.TotalTimeElapsed)
 	destroyer, err := destroy.New(logrus.StandardLogger(), directory)
 	if err != nil {
 		return errors.Wrap(err, "Failed while preparing to destroy cluster")
 	}
-	if err := destroyer.Run(); err != nil {
+	quota, err := destroyer.Run()
+	if err != nil {
 		return errors.Wrap(err, "Failed to destroy cluster")
+	}
+
+	if reportQuota {
+		if err := cluster.WriteQuota(directory, quota); err != nil {
+			return errors.Wrap(err, "Failed to record quota")
+		}
 	}
 
 	store, err := assetstore.NewStore(directory)
--- a/cmd/openshift-install/destroy.go
+++ b/cmd/openshift-install/destroy.go
@@ -4,7 +4,6 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/openshift/installer/pkg/asset/cluster"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -21,6 +20,7 @@ import (
 	_ "github.com/openshift/installer/pkg/destroy/libvirt"
 	_ "github.com/openshift/installer/pkg/destroy/openstack"
 	_ "github.com/openshift/installer/pkg/destroy/ovirt"
+	quotaasset "github.com/openshift/installer/pkg/destroy/quota"
 	_ "github.com/openshift/installer/pkg/destroy/vsphere"
 	"github.com/openshift/installer/pkg/metrics/timer"
 )
@@ -40,8 +40,7 @@ func newDestroyCmd() *cobra.Command {
 }
 
 func newDestroyClusterCmd() *cobra.Command {
-	var reportQuota bool
-	cmd := &cobra.Command{
+	return &cobra.Command{
 		Use:   "cluster",
 		Short: "Destroy an OpenShift cluster",
 		Args:  cobra.ExactArgs(0),
@@ -49,16 +48,12 @@ func newDestroyClusterCmd() *cobra.Command {
 			cleanup := setupFileHook(rootOpts.dir)
 			defer cleanup()
 
-			err := runDestroyCmd(rootOpts.dir, reportQuota)
+			err := runDestroyCmd(rootOpts.dir, os.Getenv("OPENSHIFT_INSTALL_REPORT_QUOTA_FOOTPRINT") == "true")
 			if err != nil {
 				logrus.Fatal(err)
 			}
 		},
 	}
-	quotaFlag := "report-quota-footprint"
-	cmd.Flags().BoolVar(&reportQuota, quotaFlag, false, "Report the quota footprint that the cluster had.")
-	_ = cmd.Flags().MarkHidden(quotaFlag) // only error possible is the flag does not exist, but we just added it
-	return cmd
 }
 
 func runDestroyCmd(directory string, reportQuota bool) error {
@@ -73,8 +68,8 @@ func runDestroyCmd(directory string, reportQuota bool) error {
 	}
 
 	if reportQuota {
-		if err := cluster.WriteQuota(directory, quota); err != nil {
-			return errors.Wrap(err, "Failed to record quota")
+		if err := quotaasset.WriteQuota(directory, quota); err != nil {
+			return errors.Wrap(err, "failed to record quota")
 		}
 	}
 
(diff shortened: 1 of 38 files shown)


pkg/destroy/gcp/gcp.go
```diff
@@ -90,6 +92,18 @@ func (o *ClusterUninstaller) Run() error {
 		return errors.Wrap(err, "failed to create compute service")
 	}
 
+	req := o.computeSvc.MachineTypes.AggregatedList(o.ProjectID).Fields("items/*/machineTypes(name,guestCpus),nextPageToken")
```
stevekuznetsov, line 95: I imagine this will be contentious - I need a place to put this cache, open to ideas.

pkg/destroy/gcp/gcp.go
```diff
@@ -118,6 +132,11 @@ func (o *ClusterUninstaller) Run() error {
 		return errors.Wrap(err, "failed to destroy cluster")
 	}
 
+	raw, err := json.Marshal(o.pendingItemTracker.removedQuota)
```
stevekuznetsov, line 135: There does not seem to be any way to pass options down from the top level to any individual destroyer, and I did not want to mess with those method signatures since they're pretty high-level. Open to ideas about how we can expose this output.

pkg/destroy/gcp/gcp.go
```diff
@@ -176,6 +195,29 @@ func (o *ClusterUninstaller) getZoneName(zoneURL string) string {
 	return ""
 }
 
+// getRegionFromZone extracts a region name from a zone name of the form: us-central1-a
+// Splitting the name with the last delimiter `-`, leaves a string like: us-central1
+func (o *ClusterUninstaller) getRegionFromZone(zoneName string) string {
```
stevekuznetsov, line 200: Surely there's a utility for this already ... ?

pkg/destroy/gcp/gcp.go
```diff
@@ -297,6 +340,9 @@ func (t pendingItemTracker) deletePendingItems(itemType string, items []cloudRes
 	if !exists {
 		lastFound = cloudResources{}
 	}
+	for _, item := range items {
```
stevekuznetsov, line 343: We only record the quota on deletion so we know we never double-record something.

pkg/destroy/gcp/httphealthcheck.go
```diff
@@ -3,7 +3,7 @@ package gcp
 import (
 	"github.com/pkg/errors"
 
-	compute "google.golang.org/api/compute/v1"
```
stevekuznetsov, line 6: `gofmt -s -w` did this for me ... is that not being run on this repo?

stevekuznetsov:
Part of DPTP-2397

pkg/destroy/gcp/gcp.go
```diff
@@ -261,7 +302,7 @@ func newPendingItemTracker() pendingItemTracker {
 }
 
 // GetAllPendintItems returns a slice of all of the pending items across all types.
-func (t pendingItemTracker) GetAllPendingItems() []cloudResource {
+func (t *pendingItemTracker) GetAllPendingItems() []cloudResource {
```
stevekuznetsov, line 305: I don't see a valid reason for not having this be a pointer receiver - right now you get away with it since the map is implicitly a pointer type, but it makes everything a massive pain (I'd need to define a pointer type for my slice ... )

pkg/destroy/gcp/gcp.go
patrickdillon: In general, this seems like a fine approach to handle the serialization/publishing here, but maybe I'm just not following: if you could pass options down what would you do differently/what would that allow?

> Open to ideas about how we can expose this output.

Other than the current approach of writing to the install log, we could, of course, write to a new file. Brainstorming here, but we could consider adding this data to `metadata.json`. 

The [expressed purpose](https://github.com/openshift/installer/blob/master/pkg/asset/cluster/metadata.go#L41) of `metadata.json` is `// Metadata contains information needed to destroy clusters.` so this *may* be a bit of a stretch., but I don't think this is too far afield for consideration  I would just want to double check that metadata.json is left over after a successful destroy.

patrickdillon (commented):
I haven't been able to go through this as carefully as I would like, but don't want to keep you waiting too long.

pkg/destroy/gcp/gcp.go
```diff
@@ -118,6 +133,11 @@ func (o *ClusterUninstaller) Run() error {
 		return errors.Wrap(err, "failed to destroy cluster")
 	}
 
+	raw, err := json.Marshal(o.pendingItemTracker.removedQuota)
+	if err != nil {
+		return errors.Wrap(err, "failed to marshal quota")
+	}
+	fmt.Println(string(raw))
```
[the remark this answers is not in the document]
jstuever: Alternatively, it might be worth placing this behind a flag and outputting to an file for consumption.

pkg/destroy/gcp/firewall.go
```diff
@@ -32,6 +32,13 @@ func (o *ClusterUninstaller) listFirewallsWithFilter(fields string, filter strin
 					key:      item.Name,
 					name:     item.Name,
 					typeName: "firewall",
+					quota: []quotaUsage{{
+						metric: &metric{
+							Service: serviceComputeEngineAPI,
+							Limit:   "firewalls",
```
jstuever, line 38: A lot of these types have a `kind` parameter, such as `compute#firewall`. this includes both the service and the item in its response. Not sure if it's worth trying to use these in most cases. I realize there are some special circumstances.

pkg/destroy/gcp/gcp.go
jstuever: The zone has a field for region, but this requires an API call to obtain. Might be better to do it this way. I don't know that we have this anywhere else in the code.

pkg/destroy/gcp/gcp.go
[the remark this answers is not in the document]
stevekuznetsov: Yep - I was hoping to flag-gate it, but since there's no way to currently pass any run-time parameters down to the Destroyers I was not sure how to best do that.

pkg/destroy/gcp/firewall.go
stevekuznetsov: I'm happy to factor this to do that, but wanted to make the least invasive change possible - what would you prefer?

jstuever: What you have is probably sufficient. I'd just make sure the limit strings match closely to some google defined string for the same resource.

stevekuznetsov: Oh - they match 100% to the quota API names. They do not match the `cloudResource.typeName` field

stevekuznetsov:
OK, updated for feedback and to use an entry in `metadata.json` to direct the output of this - now this will depend on https://github.com/openshift/release/pull/21352

pkg/destroy/gcp/gcp.go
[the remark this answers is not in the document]
patrickdillon: Can we still get rid of this line or switch to logging? Or are you just keeping it around for debugging?

cmd/openshift-install/destroy.go
```diff
@@ -39,32 +40,44 @@ func newDestroyCmd() *cobra.Command {
...
+			err := runDestroyCmd(rootOpts.dir, reportQuota)
 			if err != nil {
 				logrus.Fatal(err)
 			}
 		},
 	}
+	quotaFlag := "report-quota-footprint"
+	cmd.Flags().BoolVar(&reportQuota, quotaFlag, false, "Report the quota footprint that the cluster had.")
+	_ = cmd.Flags().MarkHidden(quotaFlag) // only error possible is the flag does not exist, but we just added it
+	return cmd
```
patrickdillon, line 61: I'm reluctant to even bring this up due to risk of bike shedding, but did we discuss hiding this behind an environment variable? Environment variables seem to be the preferred way of changing behavior for CI, and we don't frequently use flags.

That said, I'm totally fine with this and the fact that it is hidden would allow us to make subsequent changes if we need.

pkg/destroy/gcp/backendservice.go
```diff
@@ -1,9 +1,10 @@
 package gcp
 
 import (
+	"github.com/openshift/installer/pkg/types/gcp"
 	"github.com/pkg/errors"
 
-	compute "google.golang.org/api/compute/v1"
+	"google.golang.org/api/compute/v1"
 	"google.golang.org/api/googleapi"
```
patrickdillon, line 8: nit: we try to group imports like:
```
<stdlib>

<external>

<installer/openshift>
```
But that is not 100% consistent throughout the codebase.

pkg/asset/cluster/quota.go
```diff
@@ -0,0 +1,29 @@
+package cluster
```
patrickdillon, line 1: This shouldn't live here as it's not a cluster asset. A cluster asset would be an asset that is a dependency for creating a cluster.

I don't think we should think of this file as an "[asset](https://github.com/openshift/installer/blob/master/docs/design/assetgeneration.md)". Although it should be written to the asset directory, as all logs, output, metadata is.

Better to move this to the destroy package.

pkg/types/clusterquota.go
```diff
@@ -0,0 +1,24 @@
...
+// ClusterQuota contains the size, in cloud quota, of
+// the cluster that was created by installer.
+type ClusterQuota struct {
+	GCP *gcp.Quota `json:"gcp,omitempty"`
+}
+
+// Platform returns a string representation of the platform
+// (e.g. "aws" if AWS is non-nil).  It returns an empty string if no
+// platform is configured.
+func (cpm *ClusterQuota) Platform() string {
```
patrickdillon, line 16: Is this needed/used?

patrickdillon (commented):
Thanks again for doing this. Mostly nits, but we do need to take quota out of assets package.

cmd/openshift-install/destroy.go
stevekuznetsov: I have no preference, whichever your preference I will implement.

pkg/destroy/gcp/backendservice.go
stevekuznetsov: There's an automated tool to enforce this - do you want me to add a presubmit for the repo to enforce it?

cmd/openshift-install/destroy.go
```diff
@@ -39,32 +40,44 @@ func newDestroyCmd() *cobra.Command {
...
 	}
-	if err := destroyer.Run(); err != nil {
+	quota, err := destroyer.Run()
+	if err != nil {
 		return errors.Wrap(err, "Failed to destroy cluster")
 	}
 
+	if reportQuota {
+		if err := cluster.WriteQuota(directory, quota); err != nil {
+			return errors.Wrap(err, "Failed to record quota")
```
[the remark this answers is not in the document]
stevekuznetsov: FYI all the other error messages in this function (and file) seem to all have upper-case :)

pkg/asset/cluster/quota.go
stevekuznetsov: I put this here to mirror metadata. It's as much of an "asset" as metadata is, no?

stevekuznetsov:
/retitle Bug 2001009: pkg/destroy/gcp: report cluster footprint in quota

stevekuznetsov:
/retitle pkg/destroy/gcp: report cluster footprint in quota

PR #5135, opened 2021-08-05
Azure Stack Hub UPI README & Templates

patrickdillon:
This creates the initial templates and README for UPI deployments to Azure Stack Hub.

PR #5160, opened 2021-08-19
UPI: Add yq to images for ASH CI

rna-afk:
Adding yq for ASH CI to pick up information from yaml file to
login to the ASH environment which is needed for cluster
installation.


patrickdillon:
@jhixson74 [made the point](https://github.com/openshift/installer/pull/5135#discussion_r691573426) that azure upi uses a different form of `yq` (which is installed with pip).

Also this version of `yq` is deprecated. So we should either switch to the other version or use the new one. 

I think John's recommendation is right we should use the same as Azure UPI. This will require minute changes to the syntax.

PR #5162, opened 2021-08-20
refact: remove instance option m4 from IPI

mtulio:
```
                count(m5.large)  count(m4.large)  count(m5VSm4)
region                                                         
af-south-1                    3                0              3
ap-east-1                     3                0              3
ap-northeast-1                3                3              0
ap-northeast-2                4                2              2
ap-south-1                    3                2              1
ap-southeast-1                3                3              0
ap-southeast-2                3                3              0
ca-central-1                  3                2              1
eu-central-1                  3                3              0
eu-north-1                    3                0              3
eu-south-1                    3                0              3
eu-west-1                     3                3              0
eu-west-2                     3                3              0
eu-west-3                     3                0              3
me-south-1                    3                0              3
sa-east-1                     3                2              1
us-east-1                     5                6             -1  <-- here
us-east-2                     3                3              0
us-west-1                     2                2              0
us-west-2                     4                3              1
```


jstuever:
There may be an issue here if m5 does not exist in all six of us-east-1 zones. The [PreDefinedInstanceType](https://github.com/openshift/installer/blob/7b65e29f56d8fa26aa190d69ad209882da6d51ea/pkg/asset/machines/aws/instance_types.go#L33) ensures that the selected type exists in all provided zones. For the [control plane machines](https://github.com/openshift/installer/blob/7b65e29f56d8fa26aa190d69ad209882da6d51ea/pkg/asset/machines/master.go#L188), the provided zones [defaults](https://github.com/openshift/installer/blob/7b65e29f56d8fa26aa190d69ad209882da6d51ea/pkg/asset/installconfig/aws/availabilityzones.go#L13) to all zones. Therefore, with the most basic install-config.yaml, I expect this change is always breaking for us-east-1 region. The current e2e-aws test succeeded in us-west-1, which is not the targeted region.

jstuever:
Also keeping in mind our CI sets the default to [m5](https://github.com/openshift/release/blob/49ff0fa694b9be45fd49d0a7f808f322c4af7158/ci-operator/step-registry/ipi/conf/aws/ipi-conf-aws-commands.sh#L21) and [selects](https://github.com/openshift/release/blob/49ff0fa694b9be45fd49d0a7f808f322c4af7158/ci-operator/step-registry/ipi/conf/aws/ipi-conf-aws-commands.sh#L43-L49) only the zones where this type exists. As these values are patched into the install-config.yaml, we will probably not see the issue I mentioned above in CI.

mtulio:
> I appear to have been wrong. I tested this manually using 6 masters. It picked up the m5.xlarge type without issue and simply ignored the problematic zone reusing the first one a second time.

@jstuever yes, the [getInstanceTypeZoneInfo()](https://github.com/openshift/installer/blob/master/pkg/asset/machines/aws/instance_types.go#L58) will not return the "missing AZ" if the query set  (`types`) has only `m5`.

PR #5166, opened 2021-08-23
Bug 1996501: Remove worker disk types below 8GB

rna-afk:
If the workers are set with less than 8GB, the monitor operator
degrades. This has been fixed for IPI but the UPI templates still
has the instance types lower than 8GB as valid values. Removing
the types to match the minimum requirement for the clusters in
AWS.


jstuever:
I wonder if we should drop the AllowedValues section(s) completely. This is just a reference architecture and the minimum requirements are already covered in docs. Doing so here feels like an unnecessary maintenance burden.

rna-afk:
I'm worried that the user might have a really bad experience if the wrong values are entered. Failing after 2 hours for a small error would be annoying. But on the other hand, since the minimum requirement is mentioned on the docs, we shouldn't have this problem.

PR #5169, opened 2021-08-25
machines: Extract root disk sizes to constants

cgwalters:
I was looking at this issue:
https://github.com/openshift/openshift-docs/issues/35793
Which then led me to wonder exactly what the defaults are per
platform for disk sizes.

And then I stumbled into this 120 vs 128 thing that as best
I can tell from the git history is mostly an accident.  If someone
happens to know differently (e.g. can/should we unify on 120 or 128?)
then that'd be good.

In the meantime, an additional benefit here is that if one does
e.g. `git grep -i volume.*size` these constants will turn up.
It took me a little while to figure out how to backtrack from
the Terraform code to this.  (The lines here only contained
`Size` which is too generic to grep)

### DIFF
--- a/pkg/asset/machines/worker.go
+++ b/pkg/asset/machines/worker.go
@@ -71,6 +71,16 @@ const (
 
 	// workerUserDataFileName is the filename used for the worker user-data secret.
 	workerUserDataFileName = "99_openshift-cluster-api_worker-user-data-secret.yaml"
+
+	// decimalRootVolumeSize is the size in GB we use for some platforms.
+	// See below.
+	decimalRootVolumeSize = 120
+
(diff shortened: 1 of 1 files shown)


e-tienne (commented):
/lgtm

staebler:
@cgwalters Sorry that this got lost. Would you mind rebasing this when you get a chance?
/approve

cgwalters:
On Fri, Dec 3, 2021, at 3:36 PM, Matthew Staebler wrote:
> @cgwalters <https://github.com/cgwalters> Sorry that this got lost. 
> Would you mind rebasing this when you get a chance?

Sure, done 🏄  Was just a minor conflict with
https://github.com/openshift/installer/pull/5301

staebler (approved):
/lgtm
/approve

PR #5168, opened 2021-08-24
Bug 1981999: bump RHCOS boot images for 4.9

miabbott:
This updates the RHCOS boot image metadata in the installer with the
most recent version of RHCOS 4.9 artifacts.

This includes fixes for the following BZs:

1967483 - coreos-installer fails to download Ignition (DNS error, failed to lookup address)
1974411 - Installation with multipath parameters in parmfile fails (DNS resolution missing)
1980679 - On a Azure IPI installation MCO fails to create new nodes

Changes generated with the following:

```
$ plume cosa2stream --target data/data/rhcos-stream.json --distro rhcos aarch64=49.84.202108221747-0 ppc64le=49.84.202108241610-0 s390x=49.84.202108221604-0 x86_64=49.84.202108221651-0 --url https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases
$ ./hack/update-rhcos-bootimage.py https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos-4.9-aarch64/49.84.202108221747-0/aarch64/meta.json aarch64
$ ./hack/update-rhcos-bootimage.py https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos-4.9-s390x/49.84.202108221604-0/s390x/meta.json s390x
$ ./hack/update-rhcos-bootimage.py https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos-4.9/49.84.202108221651-0/x86_64/meta.json x86_64
$ ./hack/update-rhcos-bootimage.py https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos-4.9/49.84.202108221651-0/x86_64/meta.json amd64
```

PR #5180, opened 2021-08-31
Bug 1998643: Revert "bump RHCOS boot images for 4.9"

ardaguclu:
`e2e-metal-ipi-ovn-ipv6` and `e2e-metal-ipi-ovn-dualstack` jobs are failing in bootstrap stage. Service network endpoint can not be up and pods are crashlooping.

This PR reverts https://github.com/openshift/installer/pull/5168

### DIFF
--- a/data/data/rhcos-aarch64.json
+++ b/data/data/rhcos-aarch64.json
@@ -1,68 +1,68 @@
 {
-    "baseURI": "https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos-4.9-aarch64/49.84.202108221747-0/aarch64/",
-    "buildid": "49.84.202108221747-0",
+    "baseURI": "https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos-4.9-aarch64/49.84.202106302247-0/aarch64/",
+    "buildid": "49.84.202106302247-0",
     "images": {
         "aws": {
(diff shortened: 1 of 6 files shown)


ardaguclu:
@ashcrow @miabbott investigations continue but so far there is no clue yet what is causing this network problem. It can not complete cluster setup within installer timeout, but on local after waiting more(of course getting the same error as CI jobs), it becomes ready.

miabbott:
I really don't like the idea of reverting the boot image without knowing the underlying root cause, but if we are blocked on shipping in this state, it seems like this is the best path forward.

/approve

patrickdillon:
/approve 

@miabbott I am giving high-level installer approval. I will leave it to you to lgtm

miabbott:
One thing to note is that we have identified another blocker in RHCOS that we need a new boot image update to fully solve - https://bugzilla.redhat.com/show_bug.cgi?id=1999577

Which means we need to sort out the underlying problem that required us to pursue this revert before we can fully fix BZ#1999577

I'm going to give the `/lgtm` in hopes that we can unblock other folks from more reliable CI jobs, but we really need to understand the root cause that is requiring us to revert.

cc: @ardaguclu @hardys 

/lgtm

ashcrow:
@ardaguclu @hardys please keep us updated with your debugging. Note that https://bugzilla.redhat.com/show_bug.cgi?id=1999577 is also related to bare metal installs.

PR #5182, opened 2021-08-31
Bug 1999734: fix(ibmcloud): Add CIS CRN to infrastructure manifest

cjschaef:
Bug fix for 1999734 to inject the IBM Cloud CIS Instance CRN into the infrastructure/cluster resource,
along with pulling in the API support for IBMCloudPlatformStatus to include the CIS CRN.

- Pull in latest IBMCloudPlatformStatus
- Add CIS CRN to IBM Infrastructure resource


cjschaef:
I moved this over to `4.10`, although I still cannot set `Target Release`

/bugzilla refresh

pamoedom:
Hi @cjschaef, one question, this patch is designed to do an early check within the the install-config in order to verify that contains the corresponding `cisInstanceCRN` or just to properly abort if there is no CIS reachable within the account?
I'm asking because during my tests, when I don't include the `cisInstanceCRN` within the install-config, the installer proceeds anyway and gathers the CRN from the API, is this the expected behavior?
Regards.

cjschaef:
@pamoedom 
So, this adds the prompt (during `openshift-install create manifests`) to allow one to be selected (allowing multiple to be available within the designated account), which is then automatically injected in the necessary manifests (`infrastructure` primarily).

As far as performing installations using a predefined `install-config`, without the CIS CRN, that part was likely overlooked. As the CRN needs to be injected into the `infrastructure/cluster` resource in order to result in a successful deployment as it is required via the [`ingress-operator`](https://github.com/openshift/cluster-ingress-operator/blob/2f8718289d769b260c17a84cdf07b3ca6eae1c24/pkg/operator/controller/dns/controller.go#L583).

I will see if validation can make sure that value is in the installconfig, since it is a hard requirement for `ingress-operator`.

cjschaef:
@pamoedom 
Sorry, I think I misunderstood your statement above. I did some testing to create manifests using an `install-config` (versus the standard prompts) to see what you mean about the CIS CRN being collected.

As I stated, the CIS CRN is a requirement for `ingress-operator`, so without it, the IPI installation would eventually fail.
However, as long as a `BaseDomain` is provided, the installer should be able to look up the CRN
https://github.com/openshift/installer/blob/422e17192462f2b5d5572d8600096949d3c14634/pkg/asset/installconfig/ibmcloud/metadata.go#L51-L75

So, I think the CRN value should be available, based off the selected (prompted) or pre-configured `BaseDomain`, which would allow it to be injected into the `infrastructure/cluster` resource, for `ingress-operator`. Which is what this PR is specifically attempting to fix. If the CRN cannot be found for some reason, that would be an issue and should result in an error.

cjschaef:
Per testing with/without `install-config`, the CIS CRN is successfully looked up, provided a valid `baseDomain` is selected/exists (errors are returned otherwise).
That CRN is properly injected into the `infrastructure/cluster` resource as required.

/hold cancel

pamoedom:
Thanks @cjschaef, more clear now, my confusion was the use of `cisInstanceCRN` within the `install-config.yaml` but I can see now that is not really needed, the CIS lookup will only depend on the `baseDomain`.
Let me do some tests from my side to corroborate the behaviour in all scenarios, I'll keep you posted.

pamoedom:
Here you have my test results:

[Without `baseDomain` but using `cisInstanceCRN`]
~~~
$ ./openshift-install-local create manifests --dir bz1999734_nodomain_cis/ --log-level debug
DEBUG OpenShift Installer unreleased-master-4958-g3ecc5db80828cb596b63f02582533fc8c5ec9c10 
DEBUG Built from commit 3ecc5db80828cb596b63f02582533fc8c5ec9c10 
DEBUG Fetching Master Machines...                  
DEBUG Loading Master Machines...                   
DEBUG   Loading Cluster ID...                      
DEBUG     Loading Install Config...                
DEBUG       Loading SSH Key...                     
DEBUG       Loading Base Domain...                 
DEBUG         Loading Platform...                  
DEBUG       Loading Cluster Name...                
DEBUG         Loading Base Domain...               
DEBUG         Loading Platform...                  
DEBUG       Loading Networking...                  
DEBUG         Loading Platform...                  
DEBUG       Loading Pull Secret...                 
DEBUG       Loading Platform...                    
FATAL failed to fetch Master Machines: failed to load asset "Install Config": invalid "install-config.yaml" file: baseDomain: Invalid value: "": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
~~~

[With correct `baseDomain` but using a wrong `cisInstanceCRN`]
~~~
$ ./openshift-install-local create manifests --dir bz1999734_domain_badcis/ --log-level debug
DEBUG OpenShift Installer unreleased-master-4958-g3ecc5db80828cb596b63f02582533fc8c5ec9c10 
DEBUG Built from commit 3ecc5db80828cb596b63f02582533fc8c5ec9c10 
DEBUG Fetching Master Machines...                  
DEBUG Loading Master Machines...                   
DEBUG   Loading Cluster ID...                      
DEBUG     Loading Install Config...                
DEBUG       Loading SSH Key...                     
DEBUG       Loading Base Domain...                 
DEBUG         Loading Platform...                  
DEBUG       Loading Cluster Name...                
DEBUG         Loading Base Domain...               
DEBUG         Loading Platform...                  
DEBUG       Loading Networking...                  
DEBUG         Loading Platform...                  
DEBUG       Loading Pull Secret...                 
DEBUG       Loading Platform...                    
[...]   
DEBUG Generating Openshift Manifests...            
INFO Manifests created in: bz1999734_domain_badcis/manifests and bz1999734_domain_badcis/openshift 

$ grep -nr "cis" bz1999734_domain_badcis/*
bz1999734_domain_badcis/manifests/cluster-infrastructure-02-config.yml:22:      cisInstanceCRN: 'crn:v1:bluemix:public:internet-svcs:global:xxx::'
~~~

[With wrong `baseDomain` but correct `cisInstanceCRN`]
~~~
$ ./openshift-install-local create manifests --dir bz1999734_baddomain_cis/ --log-level debug
DEBUG OpenShift Installer unreleased-master-4958-g3ecc5db80828cb596b63f02582533fc8c5ec9c10 
DEBUG Built from commit 3ecc5db80828cb596b63f02582533fc8c5ec9c10 
DEBUG Fetching Master Machines...                  
DEBUG Loading Master Machines...                   
DEBUG   Loading Cluster ID...                      
DEBUG     Loading Install Config...                
DEBUG       Loading SSH Key...                     
DEBUG       Loading Base Domain...                 
DEBUG         Loading Platform...                  
DEBUG       Loading Cluster Name...                
DEBUG         Loading Base Domain...               
DEBUG         Loading Platform...                  
DEBUG       Loading Networking...                  
DEBUG         Loading Platform...
[...]
DEBUG   Fetching DNS Config...                     
DEBUG     Fetching Install Config...               
DEBUG     Reusing previously-fetched Install Config 
DEBUG     Fetching Cluster ID...                   
DEBUG     Reusing previously-fetched Cluster ID    
DEBUG     Fetching Platform Credentials Check...   
DEBUG     Reusing previously-fetched Platform Credentials Check 
DEBUG   Generating DNS Config...                   
FATAL failed to fetch Common Manifests: failed to fetch dependency of "Common Manifests": failed to generate asset "DNS Config": failed ot get DNS zone ID: DNS zone "imbcloud.qe.devcluster.openshift.com" not found
~~~

**Summary**: no matter what you try to hardcode using `cisInstanceCRN` within the `install-config.yaml`, that variable will depend from the proper lookup of the `baseDomain` within the account, which means the error message provided by this PR will not be easy to trigger unless some unexpected error occurs between the installer and the ibmcloud API.

@cjschaef, do my conclusions makes sense? any other scenario we can test to force that error to appear? AFAIK, no more that one CIS can be available per account and no external domain can be active without the proper CIS instance, right?

cjschaef:
@pamoedom That all makes sense to me. At the moment, I would think the only way to get to that error is if the CIS instance and/or domain were removed at exactly the right time. I do believe you can have multiple Cloud Internet Services for an IBM Cloud account, but you are correct that a domain cannot be active without the backing CIS.

However, the main target of this PR was to add this line, injecting the CIS CRN into the `infrastructure/cluster` resource.
https://github.com/openshift/installer/pull/5182/files#diff-306ba1253c4b1cd58fb44dbca55765f1aa8a2bea356a3a095070c327a9a2880aR182

So, while retrieving that value, I assumed it would be best to handle the error from the function used, supplying this error message in such a case.
https://github.com/openshift/installer/pull/5182/files#diff-306ba1253c4b1cd58fb44dbca55765f1aa8a2bea356a3a095070c327a9a2880aR175

But if there is a better function or way to get the CIS CRN, I'd be open to using that so as not to have to check the potential error returned by the function I propose to use here.

pamoedom:
The part of injecting the `cisInstanceCRN` within the manifests is working as expected:

~~~
$ cat bz1999734_domain_badcis/manifests/cluster-infrastructure-02-config.yml 
kind: Infrastructure
metadata:
  creationTimestamp: null
  name: cluster
spec:
  cloudConfig:
    key: config
    name: cloud-provider-config
  platformSpec:
    type: IBMCloud
status:
  apiServerInternalURI: https://api-int.pamoedo-test.ibmcloud.qe.devcluster.openshift.com:6443
  apiServerURL: https://api.pamoedo-test.ibmcloud.qe.devcluster.openshift.com:6443
  controlPlaneTopology: HighlyAvailable
  etcdDiscoveryDomain: ""
  infrastructureName: pamoedo-test-fjmgd
  infrastructureTopology: HighlyAvailable
  platform: IBMCloud
  platformStatus:
    ibmcloud:
      cisInstanceCRN: 'crn:v1:bluemix:public:internet-svcs:global:<obfuscated>::'
      location: eu-de
      resourceGroupName: pamoedom-rg
    type: IBMCloud
~~~

Regarding alternatives to get the CIS I'm afraid I cannot help much with that, but IMHO, having an extra error check in place for unexpected scenarios is always a good idea, looks good to me.

pamoedom (commented):
/lgtm

cjschaef:
I'll move out of draft, once I confirm the change of permission scope (or update accordingly).

patrickdillon:
@cjschaef the vendor unit test is failing. I haven't looked at it closely but the test runs `go mod tidy` && `go mod vendor` did you perhaps not run `tidy`?

cjschaef:
Dropped the API update, as it has been updated to a more recent version already, which should have the support needed for the changes in this PR.

pamoedom (commented):
/lgtm

PR #5186, opened 2021-09-02
baremetal: fail if apiVIP and ingressVIP are set to the same value

karmab:
baremetal: fail if apiVIP and ingressVIP are set to the same value


karmab:
> Looks good, thanks! Would you mind adding a test please? https://github.com/openshift/installer/blob/master/pkg/types/baremetal/validation/platform_test.go#L45

done!

hardys:
Thanks @karmab - lgtm but there seems to be a mismatch in the expected error in the new test: 

```
             	Error Trace:	platform_test.go:172
            	Error:      	Expect "baremetal.apiVIP: Invalid value: "192.168.111.100": apiVIP and ingressVIP must not be set to the same value" to match "Incorrect value: "192.168.111.100": APIVIP is expected to be different than IngressVIP"
            	Test:       	TestValidatePlatform/identical_apivip_ingressvip 
```

karmab:
> Thanks @karmab - lgtm but there seems to be a mismatch in the expected error in the new test:
> 
> ```
>              	Error Trace:	platform_test.go:172
>             	Error:      	Expect "baremetal.apiVIP: Invalid value: "192.168.111.100": apiVIP and ingressVIP must not be set to the same value" to match "Incorrect value: "192.168.111.100": APIVIP is expected to be different than IngressVIP"
>             	Test:       	TestValidatePlatform/identical_apivip_ingressvip 
> ```

my bad, i misinterpreted the test logic.
Fixed!

PR #5187, opened 2021-09-02
Terraform Stages: ensure inputs match outputs

patrickdillon:
If a Terraform stage receives input (from tf vars)
which is not declared, it results in a warning:

level=debug msg="Warning: Value for undeclared variable"

This suppresses those warnings for GA platforms by
declaring input for all declared outputs from previous
stages.


patrickdillon:
Must have missed some in Azure. AWS & GCP look good.

e-tienne:
> Must have missed some in Azure. AWS & GCP look good.

I looked at the last GCP and AWS logs, and I still see the errors at the bottom of the log file. Are we looking at the same output?

PR #5188, opened 2021-09-02
fix(ibmcloud): Allow NodePort access from all subnets

hasueki:
Allow access to Kubernetes NodePorts from within both the control plane and compute subnets. This will allow for cloud provider (CCM) managed LBs to reach to the cluster nodes.


rtheis (approved):
/lgtm

PR #5189, opened 2021-09-02
pkg/asset/installconfig/aws: public DNS validation

jhixson74:
Verify no DNS records exist in the public DNS zone. This helps to prevent
installing into an existing cluster. The goal here is to not leak public DNS
records. This should help make the race condition less likely.

https://issues.redhat.com/browse/CORS-1195

### DIFF
--- a/data/data/aws/cluster/route53/base.tf
+++ b/data/data/aws/cluster/route53/base.tf
@@ -13,6 +13,8 @@ data "aws_route53_zone" "public" {
   count = local.public_endpoints ? 1 : 0
 
   name = var.base_domain
+
+  depends_on = [aws_route53_record.api_external_internal_zone_alias, aws_route53_record.api_external_internal_zone_cname]
 }
 
 data "aws_route53_zone" "int" {
@@ -35,8 +37,6 @@ resource "aws_route53_zone" "new_int" {
     },
     var.tags,
   )
-
-  depends_on = [aws_route53_record.api_external_alias, aws_route53_record.api_external_cname]
 }
 
 resource "aws_route53_record" "api_external_alias" {
--- a/pkg/asset/installconfig/aws/validation.go
+++ b/pkg/asset/installconfig/aws/validation.go
@@ -319,50 +319,116 @@ var requiredServices = []string{
 
 // ValidateForProvisioning validates if the install config is valid for provisioning the cluster.
 func ValidateForProvisioning(session *session.Session, ic *types.InstallConfig, metadata *Metadata) error {
-	allErrs := field.ErrorList{}
-	allErrs = append(allErrs, validateExistingHostedZone(session, ic, metadata)...)
-	return allErrs.ToAggregate()
-}
-
-func validateExistingHostedZone(session *session.Session, ic *types.InstallConfig, metadata *Metadata) field.ErrorList {
-	if ic.AWS.HostedZone == "" {
+	if ic.Publish == types.InternalPublishingStrategy {
 		return nil
 	}
 
-	// validate that the hosted zone exists
-	hostedZonePath := field.NewPath("aws", "hostedZone")
+	var zoneName string
+	var zonePath *field.Path
+	var zone *route53.HostedZone
+
+	errors := field.ErrorList{}
+	allErrs := field.ErrorList{}
 	client := route53.New(session)
-	zone, err := client.GetHostedZone(&route53.GetHostedZoneInput{Id: aws.String(ic.AWS.HostedZone)})
-	if err != nil {
-		return field.ErrorList{
-			field.Invalid(hostedZonePath, ic.AWS.HostedZone, "cannot find hosted zone"),
+
+	if ic.AWS.HostedZone != "" {
+		zoneName := ic.AWS.HostedZone
+		zonePath := field.NewPath("aws", "hostedZone")
+		zoneOutput, errors := getHostedZone(client, zonePath, zoneName)
+		if len(errors) > 0 {
+			allErrs = append(allErrs, errors...)
+		}
+
+		if errors = validateHostedZone(zoneOutput, zonePath, zoneName, metadata); len(errors) > 0 {
+			allErrs = append(allErrs, errors...)
+		}
+
+		zone = zoneOutput.HostedZone
+	} else {
+		zoneName := ic.BaseDomain
+		zonePath := field.NewPath("baseDomain")
+		zone, errors = getBaseDomain(session, zonePath, zoneName)
+		if len(errors) > 0 {
+			allErrs = append(allErrs, errors...)
 		}
 	}
 
+	if errors = validateZoneRecords(client, zone, zoneName, zonePath, ic); len(errors) > 0 {
+		allErrs = append(allErrs, errors...)
+	}
+
+	return allErrs.ToAggregate()
+}
+
+func validateZoneRecords(client *route53.Route53, zone *route53.HostedZone, zoneName string, zonePath *field.Path, ic *types.InstallConfig) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	problematicRecords, err := getSubDomainDNSRecords(client, zone, ic)
+	if err != nil {
+		allErrs = append(allErrs, field.InternalError(zonePath,
+			errors.Wrapf(err, "could not list record sets for domain %q", zoneName)))
+	}
+
+	if len(problematicRecords) > 0 {
+		detail := fmt.Sprintf(
+			"the zone already has record sets for the domain of the cluster: [%s]",
+			strings.Join(problematicRecords, ", "),
+		)
+		allErrs = append(allErrs, field.Invalid(zonePath, zoneName, detail))
+	}
+
+	return allErrs
+}
+
+func getBaseDomain(session *session.Session, baseDomainPath *field.Path, baseDomainName string) (*route53.HostedZone, field.ErrorList) {
+	baseDomainZone, err := GetPublicZone(session, baseDomainName)
+	if err != nil {
+		return nil, field.ErrorList{
+			field.Invalid(baseDomainPath, baseDomainName, "cannot find base domain"),
+		}
+	}
+	return baseDomainZone, nil
+}
+
+func getHostedZone(client *route53.Route53, hostedZonePath *field.Path, hostedZoneName string) (*route53.GetHostedZoneOutput, field.ErrorList) {
+	// validate that the hosted zone exists
+	hostedZoneOutput, err := client.GetHostedZone(&route53.GetHostedZoneInput{Id: aws.String(hostedZoneName)})
+	if err != nil {
+		return nil, field.ErrorList{
+			field.Invalid(hostedZonePath, hostedZoneName, "cannot find hosted zone"),
+		}
+	}
+	return hostedZoneOutput, nil
+}
+
+func validateHostedZone(hostedZoneOutput *route53.GetHostedZoneOutput, hostedZonePath *field.Path, hostedZoneName string, metadata *Metadata) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	// validate that the hosted zone is associated with the VPC containing the existing subnets for the cluster
 	vpcID, err := metadata.VPC(context.TODO())
 	if err == nil {
-		if !isHostedZoneAssociatedWithVPC(zone, vpcID) {
-			allErrs = append(allErrs, field.Invalid(hostedZonePath, ic.AWS.HostedZone, "hosted zone is not associated with the VPC"))
+		if !isHostedZoneAssociatedWithVPC(hostedZoneOutput, vpcID) {
+			allErrs = append(allErrs, field.Invalid(hostedZonePath, hostedZoneName, "hosted zone is not associated with the VPC"))
 		}
 	} else {
-		allErrs = append(allErrs, field.Invalid(hostedZonePath, ic.AWS.HostedZone, "no VPC found"))
+		allErrs = append(allErrs, field.Invalid(hostedZonePath, hostedZoneName, "no VPC found"))
 	}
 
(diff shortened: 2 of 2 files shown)


wking:
Previously, an existing public A record would cause an install failure when Terraform tried to create the public A record and failed on the collision.  With this PR's precheck, you'll fail a bit faster.  But the overall result will be similar.  If you want to avoid the leak entirely, you'll need to reverse ae9cbafafaaefd93 (#1508), and start claiming the private record before you claim the public record.  That way, the deletion logic will find the private record and know it can safely remove the public record.  It will be a bit racy, although your new precheck will limit the exposure.  You'll be vulnerable to:

1. Install A's precheck looks for the public record, confirms it doesn't exist.
2. Install B creates the public record.
3. Install A creates the private record.
4. Install A tries to create the public record, but fails because B won the race.
5. Attempting to remove cluster A breaks cluster B by removing its public record (which A thought it owned, because of the private record from step 3).

It's possible you could move the precheck into a Terraform data request or some such, to reduce the racy gap between step 1 and step 4.  But even with your Go-side precheck, the gap may be small enough that we can say "probably won't happen too often, and we can manually restore B's public A record if it does" and decide that that risk is less annoying than the current A record leak.  I personally don't have much opinion on the annoyance level of leaking public records on deletion vs. removing public records that really belong to one cluster when deleting a separate cluster.  Parallel installs using the same name seems like something that will probably never happen for production clusters.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -330,15 +330,50 @@ var requiredServices = []string{
...
+	} else {
+		allErrs = append(allErrs, validateExistingHostedZone(session, ic, metadata)...)
+	}
 	return allErrs.ToAggregate()
 }
 
-func validateExistingHostedZone(session *session.Session, ic *types.InstallConfig, metadata *Metadata) field.ErrorList {
-	if ic.AWS.HostedZone == "" {
+// validate no pre-existing DNS records exist in the public DNS zone
+func validatePreExistingPublicDNS(session *session.Session, ic *types.InstallConfig, metadata *Metadata) field.ErrorList {
```
staebler, line 342: Can we change the name of this function to indicate that it is validating that there is **not** a pre-existing public DNS record?
```suggestion
func validateNoExistingPublicDNS(session *session.Session, ic *types.InstallConfig, metadata *Metadata) field.ErrorList {
```

pkg/asset/installconfig/aws/validation.go
```diff
@@ -330,15 +330,50 @@ var requiredServices = []string{
...
+	if ic.Publish == types.InternalPublishingStrategy {
 		return nil
 	}
 
+	baseDomainPath := field.NewPath("aws", "baseDomain")
```
staebler, line 347: This is not the correct field path.
```suggestion
	baseDomainPath := field.NewPath("baseDomain")
```

pkg/asset/installconfig/aws/validation.go
```diff
@@ -330,15 +330,50 @@ var requiredServices = []string{
...
+	baseDomainZone, err := GetPublicZone(session, ic.BaseDomain)
+	if err != nil {
+		return field.ErrorList{
+			field.Invalid(baseDomainPath, ic.BaseDomain, "cannot find base domain"),
+		}
+	}
+
+	allErrs := field.ErrorList{}
```
staebler, line 355:
```suggestion
	var allErrs field.ErrorList
```

pkg/asset/installconfig/aws/validation.go
```diff
@@ -369,10 +404,28 @@ func validateExistingHostedZone(session *session.Session, ic *types.InstallConfi
 			fmt.Sprintf("hosted zone domain %q is not a parent of the cluster domain %q", *zone.HostedZone.Name, dottedClusterDomain)))
 	}
 
+	problematicRecords, err := getProblematicRecords(client, zone.HostedZone, dottedClusterDomain)
+	if err != nil {
+		allErrs = append(allErrs, field.InternalError(hostedZonePath,
+			errors.Wrapf(err, "could not list record sets for hosted zone %q", ic.AWS.HostedZone)))
+
```
staebler, line 411:
```suggestion
```

pkg/asset/installconfig/aws/validation.go
```diff
@@ -369,10 +404,28 @@ func validateExistingHostedZone(session *session.Session, ic *types.InstallConfi
...
+			"hosted zone already has record sets for the domain of the cluster: [%s]",
+			strings.Join(problematicRecords, ", "),
+		)
+		allErrs = append(allErrs, field.Invalid(hostedZonePath, ic.AWS.HostedZone, detail))
+	}
+
+	return allErrs
+}
+
+func getProblematicRecords(client *route53.Route53, hostedZone *route53.HostedZone, dottedClusterDomain string) (problematicRecords []string, err error) {
```
staebler, line 425: This function name does not give much information about what it does.
Also, I don't see a compelling reason to use named returns here.
```suggestion
func getSubDomainDNSRecords(client *route53.Route53, hostedZone *route53.HostedZone, dottedClusterDomain string) ([]string, error) {
```

pkg/asset/installconfig/aws/validation.go
```diff
@@ -330,15 +330,50 @@ var requiredServices = []string{
...
+	baseDomainPath := field.NewPath("aws", "baseDomain")
+	baseDomainZone, err := GetPublicZone(session, ic.BaseDomain)
+	if err != nil {
+		return field.ErrorList{
+			field.Invalid(baseDomainPath, ic.BaseDomain, "cannot find base domain"),
+		}
+	}
+
+	allErrs := field.ErrorList{}
+	dottedClusterDomain := ic.ClusterDomain() + "."
```
staebler, line 356: We should leave the details of whether the cluster domain should be dotted to the inners of the `getProblematicRecords` and `isHostedZoneDomainParentOfClusterDomain` functions. In other words, change those two functions to take in the canonical cluster domain rather than the dotted cluster domain, and have the functions add the dot as appropriate.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,50 +329,120 @@ var requiredServices = []string{
...
+		zoneName := ic.AWS.HostedZone
+		zonePath := field.NewPath("aws", "hostedZone")
+		zoneOutput, err := getHostedZone(client, zonePath, zoneName)
+		if err != nil {
+			allErrs = append(allErrs, err...)
+		}
+
+		err = validateHostedZone(zoneOutput, zonePath, zoneName, metadata)
+		if err != nil {
+			allErrs = append(allErrs, err...)
```
staebler, line 354:
```suggestion
		if errors := validateHostedZone(zoneOutput, zonePath, zoneName, metadata); len(errors) > 0 {
			allErrs = append(allErrs, errors...)
```

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,50 +329,120 @@ var requiredServices = []string{
...
+		}
+
+		zone = zoneOutput.HostedZone
+	} else {
+		zoneName := ic.BaseDomain
+		zonePath := field.NewPath("baseDomain")
+		zone, err = getBaseDomain(session, zonePath, zoneName)
+		if err != nil {
+			allErrs = append(allErrs, err...)
```
staebler, line 363:
```suggestion
		zone, errors = getBaseDomain(session, zonePath, zoneName)
		if len(errors) > 0 {
			allErrs = append(allErrs, errors...)
```

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,50 +329,120 @@ var requiredServices = []string{
...
+		}
+	}
+
+	err = validateZoneRecords(client, zone, zoneName, zonePath, ic)
+	if err != nil {
+		allErrs = append(allErrs, err...)
```
staebler, line 369:
```suggestion
	if errors := validateZoneRecords(client, zone, zoneName, zonePath, ic); len(errors) > 0 {
		allErrs = append(allErrs, errors...)
```

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,50 +329,120 @@ var requiredServices = []string{
...
 	} else {
-		allErrs = append(allErrs, field.Invalid(hostedZonePath, ic.AWS.HostedZone, "no VPC found"))
+		allErrs = append(allErrs, field.Invalid(hostedZonePath, hostedZoneName, "no VPC found"))
 	}
 
+	return allErrs
+}
+
+func getSubDomainDNSRecords(client *route53.Route53, hostedZone *route53.HostedZone, ic *types.InstallConfig) ([]string, error) {
+	var err error
```
staebler, line 433: Don't declare this variable here. Declare it where it is used via `:=`.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,50 +329,120 @@ var requiredServices = []string{
...
+	if !isHostedZoneDomainParentOfClusterDomain(hostedZone, dottedClusterDomain) {
+		return nil, errors.New(fmt.Sprintf("hosted zone domain %q is not a parent of the cluster domain %q", hostedZone.Name, dottedClusterDomain))
 	}
 
-	// validate that the hosted zone does not already have any record sets for the cluster domain
 	var problematicRecords []string
-	if err := client.ListResourceRecordSetsPages(
-		&route53.ListResourceRecordSetsInput{HostedZoneId: zone.HostedZone.Id},
+	// validate that the hosted zone does not already have any record sets for the cluster domain
+	if err = client.ListResourceRecordSetsPages(
```
staebler, line 444:
```suggestion
	if err := client.ListResourceRecordSetsPages(
```

data/data/aws/cluster/route53/base.tf
```diff
@@ -13,6 +13,8 @@ data "aws_route53_zone" "public" {
   count = local.public_endpoints ? 1 : 0
 
   name = var.base_domain
+
+  depends_on = [aws_route53_record.api_internal_alias, aws_route53_record.api_internal_cname]
```
staebler, line 17: I think that we should put the `depends_on` on the resource for the external records themselves. And it should be the records for the api name not the api-int name.
`aws_route53_record.api_external_alias` -> `aws_route53_record.api_external_internal_zone_alias`
`aws_route53_record.api_external_cname` -> `aws_route53_record.api_external_internal_zone_cname`

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,50 +329,120 @@ var requiredServices = []string{
...
+	var err error
+
 	dottedClusterDomain := ic.ClusterDomain() + "."
 
 	// validate that the domain of the hosted zone is the cluster domain or a parent of the cluster domain
-	if !isHostedZoneDomainParentOfClusterDomain(zone.HostedZone, dottedClusterDomain) {
-		allErrs = append(allErrs, field.Invalid(hostedZonePath, ic.AWS.HostedZone,
-			fmt.Sprintf("hosted zone domain %q is not a parent of the cluster domain %q", *zone.HostedZone.Name, dottedClusterDomain)))
+	if !isHostedZoneDomainParentOfClusterDomain(hostedZone, dottedClusterDomain) {
+		return nil, errors.New(fmt.Sprintf("hosted zone domain %q is not a parent of the cluster domain %q", hostedZone.Name, dottedClusterDomain))
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,50 +329,120 @@ var requiredServices = []string{
...
+	var zoneName string
+	var zonePath *field.Path
+	var zone *route53.HostedZone
+
+	if ic.AWS.HostedZone != "" {
+		zoneName := ic.AWS.HostedZone
+		zonePath := field.NewPath("aws", "hostedZone")
+		zoneOutput, err := getHostedZone(client, zonePath, zoneName)
+		if err != nil {
+			allErrs = append(allErrs, err...)
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,50 +329,120 @@ var requiredServices = []string{
 
 // ValidateForProvisioning validates if the install config is valid for provisioning the cluster.
 func ValidateForProvisioning(session *session.Session, ic *types.InstallConfig, metadata *Metadata) error {
+	if ic.Publish == types.InternalPublishingStrategy {
+		return nil
+	}
+
+	err := field.ErrorList{}
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/asset/installconfig/aws/validation.go
jhixson74: Fixed.

pkg/asset/installconfig/aws/validation.go
jhixson74: Fixed.

pkg/asset/installconfig/aws/validation.go
jhixson74: Fixed.

pkg/asset/installconfig/aws/validation.go
jhixson74: Fixed.

pkg/asset/installconfig/aws/validation.go
jhixson74: Fixed.

data/data/aws/cluster/route53/base.tf
jhixson74: Fixed.

staebler (approved):
/lgtm
/approve

PR #5193, opened 2021-09-06
Bug 1993364: Consider all Networks on Router clean up

MaysaMacedo:
When performing a cluster destroy we should
look for all cluster tagged Networks that might be
connected to the router, regardless if it was created
by CAPO. This commit updates the filtering for the
look up of Networks and ensure to skip the Subnet
when no Gateway is set.


mandre (commented):
/lgtm
/hold for CI

EmilienM:
/hold cancel
CI looks fine, the failing tests are unrelated.

PR #5205, opened 2021-09-10
Bug 2003113: Improve host role management during assets creation

andfasano:
The current patch provides a fix for the issue BZ 2003113.
The logic for generating baremetal manifests and Terraform variables now takes into account the host `role` tag properly, if defined.
To remain backward compatible, if an host doesn not have any role defined (or supported value), it's considered eligible for the ControlPlane or Compute respectively, based on the configured requirements.
For that, the validation rule has been enhanced to verify that the correct number of hosts have been configured.

### DIFF
--- a/pkg/asset/machines/baremetal/hosts.go
+++ b/pkg/asset/machines/baremetal/hosts.go
@@ -12,6 +12,7 @@ import (
 	baremetalhost "github.com/metal3-io/baremetal-operator/apis/metal3.io/v1alpha1"
 
 	"github.com/openshift/installer/pkg/types"
+	"github.com/openshift/installer/pkg/types/baremetal"
 )
 
 // HostSettings hold the information needed to build the manifests to
@@ -25,6 +26,64 @@ type HostSettings struct {
 	Secrets []corev1.Secret
 }
 
+func createSecret(host *baremetal.Host) (*corev1.Secret, baremetalhost.BMCDetails) {
+	bmc := baremetalhost.BMCDetails{}
+	if host.BMC.Username != "" && host.BMC.Password != "" {
+		// Each host needs a secret to hold the credentials for
+		// communicating with the management controller that drives
+		// that host.
+		secret := &corev1.Secret{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "v1",
+				Kind:       "Secret",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      fmt.Sprintf("%s-bmc-secret", host.Name),
+				Namespace: "openshift-machine-api",
+			},
+			Data: map[string][]byte{
+				"username": []byte(host.BMC.Username),
+				"password": []byte(host.BMC.Password),
+			},
+		}
+		bmc.Address = host.BMC.Address
+		bmc.CredentialsName = secret.Name
+		bmc.DisableCertificateVerification = host.BMC.DisableCertificateVerification
+
+		return secret, bmc
+	}
+	return nil, bmc
+}
+
+func createBaremetalHost(host *baremetal.Host, bmc baremetalhost.BMCDetails) baremetalhost.BareMetalHost {
+
+	// Map string 'default' to hardware.DefaultProfileName
+	if host.HardwareProfile == "default" {
+		host.HardwareProfile = hardware.DefaultProfileName
+	}
+
+	newHost := baremetalhost.BareMetalHost{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: baremetalhost.GroupVersion.String(),
+			Kind:       "BareMetalHost",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      host.Name,
+			Namespace: "openshift-machine-api",
+		},
+		Spec: baremetalhost.BareMetalHostSpec{
+			Online:          true,
+			BMC:             bmc,
+			BootMACAddress:  host.BootMACAddress,
+			HardwareProfile: host.HardwareProfile,
+			BootMode:        baremetalhost.BootMode(host.BootMode),
+			RootDeviceHints: host.RootDeviceHints.MakeCRDHints(),
+		},
+	}
+
+	return newHost
+}
+
 // Hosts returns the HostSettings with details of the hardware being
 // used to construct the cluster.
 func Hosts(config *types.InstallConfig, machines []machineapi.Machine) (*HostSettings, error) {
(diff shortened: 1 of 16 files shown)


pkg/tfvars/baremetal/baremetal_test.go
```diff
@@ -0,0 +1,105 @@
...
+			var cfg config
+			err = json.Unmarshal(data, &cfg)
+			assert.Nil(t, err)
+
+			for i, hostName := range tc.expectedHosts {
+				assert.Equal(t, hostName, cfg.Masters[i]["name"])
+			}
+		})
+	}
+}
```
sadasu, line 105: Do we need a test case here where the masters and workers are out of order?

pkg/types/baremetal/validation/platform.go
```diff
@@ -270,21 +270,44 @@ func validateOSImages(p *baremetal.Platform, fldPath *field.Path) field.ErrorLis
 	return platformErrs
 }
 
-func validateHostsCount(hosts []*baremetal.Host, installConfig *types.InstallConfig) error {
+func validateMastersCount(hosts []*baremetal.Host, installConfig *types.InstallConfig) error {
 
-	hostsNum := int64(len(hosts))
 	counter := int64(0)
+	if installConfig.ControlPlane != nil && installConfig.ControlPlane.Replicas != nil {
```
sadasu, line 276: If ControlPlane is nil or ControlPlane.Replicas is nil, is this still a valid config?

pkg/types/validation/installconfig_test.go
```diff
@@ -109,6 +109,7 @@ func validBareMetalPlatform() *baremetal.Platform {
 		Hosts: []*baremetal.Host{
 			{
 				Name:           "host1",
+				Role:           "master",
```
sadasu, line 112: Is it possible that the config is provided without a defined "Role"/ Do we have defaults when that happens?

pkg/tfvars/baremetal/baremetal_test.go
andfasano: Good idea

pkg/types/validation/installconfig_test.go
andfasano: We don't have defaults (if not in the wizard). In any case the new validation rules for the hosts count will detect now any host without the `role` tag (now filtered per `master` and `worker`).

pkg/types/baremetal/validation/platform.go
andfasano: For such cases there's a reasonable default [here](https://github.com/openshift/installer/blob/40373d2fbebcb79117b7fcf1bd92d6252630fdbe/pkg/types/defaults/installconfig.go#L68-L72)

pkg/asset/machines/baremetal/hosts.go
```diff
@@ -34,76 +93,37 @@ func Hosts(config *types.InstallConfig, machines []machineapi.Machine) (*HostSet
...
+	for _, host := range config.Platform.BareMetal.Hosts {
 
-		// Map string 'default' to hardware.DefaultProfileName
-		if host.HardwareProfile == "default" {
-			host.HardwareProfile = hardware.DefaultProfileName
+		secret, bmc := createSecret(host)
+		if secret != nil {
+			settings.Secrets = append(settings.Secrets, *secret)
 		}
+		newHost := createBaremetalHost(host, bmc)
```
sadasu, line 104: Should newHost be created inside the check for if we have insufficient masters later in https://github.com/openshift/installer/pull/5205/files#diff-ae5c907c1b30aadff3caf9124cfc777ab5a253bf12b56b63e4b2f428132c9022R106?

andfasano: For that I kept the the same previous logic: here we're preparing the data for generating all the baremetal related manifests (secrets, masters and moreover workers). The assumption to have enough hosts is guaranteed by the previous validation rules applied by the `InstallConfig` asset

pkg/types/baremetal/validation/platform.go
staebler: Note that the default is applied prior to validation.

pkg/tfvars/baremetal/baremetal.go
```diff
@@ -45,17 +45,19 @@ func init() {
...
 	bootstrapOSImage, err := imageDownloader(bootstrapOSImage)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to use cached bootstrap libvirt image")
 	}
 
 	var masters, rootDevices, properties, driverInfos, instanceInfos []map[string]interface{}
 
-	for _, host := range platformHosts {
+	// Select the first N hosts as masters, excluding the workers
+	for i := 0; len(masters) < int(numControlPlaneReplicas); i++ {
```
staebler, line 57: Can we add a check to avoid overrunning the slice, too, rather than relying on validation to ensure that the number of control-plane replicas is not too large?
```suggestion
	for i := 0; i < len(platformHosts) && len(masters) < int(numControlPlaneReplicas); i++ {
```

My personal preference would be the following. However, I am not opposed to what you have.
```suggestion
	for _, host := range platformHosts {
		if len(masters) >= int(numControlPlaneReplicas) {
			break
		}
```

staebler (commented):
Please improve on the messages for the last 3 commits prior to merging.

pkg/tfvars/baremetal/baremetal_test.go
```diff
@@ -0,0 +1,164 @@
...
+			platformHosts: platformHosts(
+				host("worker-0", "worker"),
+				host("master-0", ""),
+				host("worker-1", "worker"),
+				host("master-1", ""),
+				host("worker-2", "worker"),
+				host("master-2", ""),
+			),
+			expectedHosts: []string{"master-0", "master-1"},
+		},
```
zaneb, line 107: We appear to be missing a test case like:

```
	numControlPlaneReplicas: 1,
	platformHosts: platformHosts(
		host("worker-0", ""),
		host("worker-1", ""),
		host("master-0", "master"),
	),
	expectedHosts: []string{"master-0"},
```

Also (spoiler alert!) I don't think it will pass.

pkg/tfvars/baremetal/baremetal_test.go
```diff
@@ -0,0 +1,174 @@
...
+		{
+			scenario:                "more-masters-than-required",
+			numControlPlaneReplicas: 1,
+			platformHosts: platformHosts(
+				host("server-0", ""),
+				host("server-1", ""),
+				host("server-2", "master"),
+			),
+			expectedHosts: []string{"server-0"},
```
zaneb, line 116: This doesn't seem right to me, because we are ignoring information that the user has explicitly specified and doing something else.

You could easily fix this by doing:

```
sortIndex := map[string]int{"master" : -1, "": 0, "worker": 1}
sort.SliceStable(platformHosts, func (i, j int) bool {return sortIndex[platformHosts[i].Role] < sortIndex[platformHosts[j].Role]}
```

and then just taking the first _n_ from the list as before.

andfasano: The original point was to try avoiding any behavioral change (for already working configurations), so that we could backport the fix in 4.8. In any case the lack of role specification is problematic, and IMO the best solution would be making the role tag as mandatory (which could be something to be introduced later).

@hardys wdyt?

kirankt (approved):
/approve

pkg/tfvars/baremetal/baremetal_test.go
andfasano: @hardys this will mean considering the [mixed case](https://github.com/openshift/installer/pull/5205#discussion_r733852440) as invalid.
Since I guess we all agree to make the role tag as mandatory in future, and since in the current implementation a clear warning is displayed in the output for the mixed case, maybe we could keep the current approach which anyhow address the original ordering issue.

Given an install-config.yaml with 5 hosts configured (2 with master role and 1 with worker role), that's the output:
```
$ openshift-install --dir=/test/cluster create cluster
WARNING Host ostest-master-norole-2 hasn't any role configured 
WARNING Host ostest-worker-norole-1 hasn't any role configured 
...
```

zaneb:
Heh, there that [one line changed](https://github.com/openshift/installer/pull/5205/files#diff-347c2595fa8d376188ec9b35951221328d5abc35ddd6c9e5f9e6f6cf960c1646) that means we have to pass All The Tests :facepalm:

mandre:
By the way, the second commit has the longest commit title I've seen in a while :smile:

PR #5207, opened 2021-09-13
baremetal: add NetworkConfig field

andfasano:
This PR adds a new `install-config` field to store network configuration data (with validation) as described in the enhancement https://github.com/openshift/enhancements/pull/817

### DIFF
--- a/pkg/types/baremetal/platform.go
+++ b/pkg/types/baremetal/platform.go
@@ -39,6 +39,7 @@ type Host struct {
 	HardwareProfile string           `json:"hardwareProfile"`
 	RootDeviceHints *RootDeviceHints `json:"rootDeviceHints,omitempty"`
 	BootMode        BootMode         `json:"bootMode,omitempty"`
+	NetworkConfig   string           `json:"networkConfig,omitempty"`
 }
 
 // IsMaster checks if the current host is a master
--- a/pkg/types/baremetal/validation/platform.go
+++ b/pkg/types/baremetal/validation/platform.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/apparentlymart/go-cidr/cidr"
+	"github.com/ghodss/yaml"
 	"github.com/go-playground/validator/v10"
 	"github.com/metal3-io/baremetal-operator/pkg/bmc"
 	"github.com/pkg/errors"
@@ -325,6 +326,18 @@ func validateHostsCount(hosts []*baremetal.Host, installConfig *types.InstallCon
 	return nil
 }
 
+// ensure that the NetworkConfig field contains a valid Yaml string
+func validateNetworkConfig(hosts []*baremetal.Host, fldPath *field.Path) (errors field.ErrorList) {
+	for idx, host := range hosts {
+		networkConfig := make(map[string]interface{})
+		err := yaml.Unmarshal([]byte(host.NetworkConfig), &networkConfig)
+		if err != nil {
+			errors = append(errors, field.Invalid(fldPath.Index(idx).Child("networkConfig"), host.NetworkConfig, fmt.Sprintf("Not a valid yaml: %s", err.Error())))
+		}
+	}
+	return
+}
+
 // ensure that the bootMode field contains a valid value
 func validateBootMode(hosts []*baremetal.Host, fldPath *field.Path) (errors field.ErrorList) {
 	for idx, host := range hosts {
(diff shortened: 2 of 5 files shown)


pkg/types/baremetal/validation/platform.go
```diff
@@ -290,6 +291,18 @@ func validateHostsCount(hosts []*baremetal.Host, installConfig *types.InstallCon
...
 }
 
+// ensure that the NetworkConfig field contains a valid Yaml string
+func validateNetworkConfig(hosts []*baremetal.Host, fldPath *field.Path) (errors field.ErrorList) {
+	for idx, host := range hosts {
+		networkConfig := make(map[interface{}]interface{})
+		err := yaml.Unmarshal([]byte(host.NetworkConfig), &networkConfig)
+		if err != nil {
+			errors = append(errors, field.Invalid(fldPath.Index(idx).Child("networkConfig"), host.NetworkConfig, "Not a valid yaml"))
+		}
```
kirankt, line 301: Suggestion: perhaps we can print the contents of `err` to gain more insight as to why the yaml is invalid?

kirankt (changes requested):

pkg/types/baremetal/validation/platform.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/apparentlymart/go-cidr/cidr"
 	"github.com/go-playground/validator/v10"
 	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
```
asalkeld, line 15: shouldn't this be "github.com/ghodss/yaml" ? it looks like most other yaml imports are using ghodss

asalkeld (commented):
looks mostly good to me, but not an installer guru..

pkg/types/baremetal/validation/platform.go
andfasano: Thanks @asalkeld it sounds probably a better choice, especially for validation messages as pointed out by @kirankt

pkg/types/baremetal/validation/platform.go
andfasano: Thanks, done! The message is slightly too verbose/technical, but at least could give a better idea about the location of the error

kirankt (approved):
Looks good to me, but lets wait for the enhancement to merge before committing this.
/hold

kirankt:
/approve
Please remove the hold once the enhancement merges

imain:
LGTM

romfreiman:
> Do you have a reference to a specific failed mandatory build?

Why failed? :) I dont know, that's why I suggested to verify it prior to pushing

kirankt (approved):
/approve

pkg/types/baremetal/validation/platform.go
```diff
@@ -291,6 +292,18 @@ func validateHostsCount(hosts []*baremetal.Host, installConfig *types.InstallCon
 	return nil
 }
 
+// ensure that the NetworkConfig field contains a valid Yaml string
+func validateNetworkConfig(hosts []*baremetal.Host, fldPath *field.Path) (errors field.ErrorList) {
+	for idx, host := range hosts {
+		networkConfig := make(map[string]interface{})
+		err := yaml.Unmarshal([]byte(host.NetworkConfig), &networkConfig)
+		if err != nil {
+			errors = append(errors, field.Invalid(fldPath.Index(idx).Child("networkConfig"), host.NetworkConfig, fmt.Sprintf("Not a valid yaml: %s", err.Error())))
```
staebler, line 301:
```suggestion
			errors = append(errors, field.Invalid(fldPath.Index(idx).Child("networkConfig"), host.NetworkConfig, fmt.Sprintf("Not valid yaml: %s", err.Error())))
```

pkg/types/baremetal/platform.go
```diff
@@ -34,6 +34,7 @@ type Host struct {
 	HardwareProfile string           `json:"hardwareProfile"`
 	RootDeviceHints *RootDeviceHints `json:"rootDeviceHints,omitempty"`
 	BootMode        BootMode         `json:"bootMode,omitempty"`
+	NetworkConfig   string           `json:"networkConfig,omitempty"`
```
staebler, line 37: Please add documentation here for this field so that it shows up in the `openshift-install explain` output.

staebler, line 37: Are we sure that we want this to be a `string` rather than a `map[string]interface{}`? I would expect the user to want to enter in-line yaml rather than a text string.

staebler (commented):
/approve

pkg/types/baremetal/platform.go
andfasano: The installer manages it essentially as a blob (it will be stored also in a secret, see PR [5247](https://github.com/openshift/installer/pull/5247)) and will pass it as it is to the image builder container (with a just a minimal validation).

PR #5208, opened 2021-09-13
baremetal: Specify entrypoint when starting inspector

derekhiggins:
Since https://github.com/openshift/ironic-image/pull/190 the inspector components and entrypoint moved to the ironic-image, so we need to update the image and entrypoint to match, also aligning with https://github.com/openshift/cluster-baremetal-operator/pull/132

### DIFF
--- a/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
+++ b/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
@@ -4,7 +4,6 @@ set -ex
 . /usr/local/bin/release-image.sh
 
 IRONIC_IMAGE=$(image_for ironic)
-IRONIC_INSPECTOR_IMAGE=$(image_for ironic-inspector)
 IPA_DOWNLOADER_IMAGE=$(image_for ironic-ipa-downloader)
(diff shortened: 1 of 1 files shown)


data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -177,6 +177,7 @@ podman run -d --net host --privileged --name ironic-inspector \
      --restart on-failure \
      --env PROVISIONING_INTERFACE=$PROVISIONING_NIC \
      --env HTTP_BASIC_HTPASSWD=${IRONIC_HTPASSWD} \
+     --entrypoint /bin/runironic-inspector \
```
hardys, line 180: I think we also need to replace all `IRONIC_INSPECTOR_IMAGE` references with `IRONIC_IMAGE` right?

I still see the inspector image in the release payload, but it seems likely we'll want to remove that when this is fixed to use the combined ironic-image?

derekhiggins: yup, will do

hardys: lgtm, could you just update the commit message to reflect the change, e.g move to using the ironic-image which now also contains the inspector dependencies and entrypoint?

kirankt (approved):
/lgtm

Issue #5179, opened 2021-08-31
vSphere provider doesn't validate ratio between CPUs and cpusPerSocket


Fodoj:
In the latest version it's possible to set cpus: 1 and coresPerSocket: 8 for vSphere IPI, and installer won't complain.

In the background, it will create nodes with 1 CPU, which would result in broken installation. 

Though for sure this is a user error in preparing install-config.yaml, it would be nice if the installer could validate that `coresPerSocket < cpus`

Shanduur:
I created [PoC](https://github.com/Shanduur/installer/commit/29eb69af3a25b234c881f7f1a804e468d814a9cf) for this issue, I would love if you have a look at it and validate if it now works as you expected.

PR #5217, opened 2021-09-15
feat(VSphere): validation of cpus/coresPerSocket in config

Shanduur:
As mentioned in #5179:

> it's possible to set cpus: 1 and coresPerSocket: 8 for vSphere IPI, and installer won't complain. In the background, it will create nodes with 1 CPU, which would result in broken installation.

This PR is aiming to address that issue.


pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -20,6 +20,10 @@ func Validate(ic *types.InstallConfig) error {
 		return errors.New(field.Required(field.NewPath("platform", "vsphere"), "vSphere validation requires a vSphere platform configuration").Error())
 	}
 
+	if err := validateCPUConfig(ic); err != nil {
+		return err
+	}
```
patrickdillon, line 25: All static validation--that is validation of user input where it is not necessary to call the vsphere API--should go in types, specifically: https://github.com/openshift/installer/blob/master/pkg/types/vsphere/validation/machinepool.go

pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -66,3 +74,16 @@ func folderExists(ic *types.InstallConfig, fldPath *field.Path) field.ErrorList
 	}
 	return nil
 }
+
+// validateCPUConfig returns an error if a number of cores per socket (coresPerSocket) has lower value than number of CPUs (cpus).
+func validateCPUConfig(ic *types.InstallConfig) error {
+	if ic.Platform.VSphere.DefaultMachinePlatform != nil {
```
patrickdillon, line 80: You will want to validate the default, worker, and master machine pools, but if you move the validation to pkg types as I suggested in the other comment that should be handled for you.

Shanduur:
Hi @patrickdillon . Thank you for your feedback. I applied your suggestions, this should be implemented as intended now.

PR #5218, opened 2021-09-15
Block `create cluster` if bootstrapInPlace is set in install-config

omertuc:
bootstrapInPlace configuration is only meant for generating ignition and embedding
it manually in an RHCOS ISO. It is not meant for IPI installation.

It seems that trying to do single-node IPI installation with this configuration
set in the install-config file leads to odd installation failures with confusing
error messages. This commit makes it so that the error is much earlier and
more clear.


eranco74 (commented):
/lgtm

PR #5219, opened 2021-09-15
Bug 2004600: Remove duplicate ramdisk log container

derekhiggins:
The runlogwatch.sh script is now intended to watch
both directories, so we only need one instance of
this container.

Side Note: gathering the inspection logs is also currently broken due to a bug,
this is fixed upstream metal3-io/ironic-image#289


kirankt (approved):
/approve

PR #5223, opened 2021-09-16
Add create_bump_commit

ravanelli:
- Add a helper to run the necessary steps for
 bump boot images for rhcos in different versions,
 creating the final commit with git

Signed-off-by: Renata Ravanelli <rravanel@redhat.com>

### DIFF
--- a/hack/create_bump_commit.py
+++ b/hack/create_bump_commit.py
@@ -1,0 +1,141 @@
+#!/usr/bin/env python
+
+'''
+Create a commit for boot image bump for RHCOS
+Usage: python create_bump_commit.py -v 4.6 -b 37124
+
+Requires python3-GitPython
+
+'''
+import argparse
+import git
+import os
+import subprocess
+import sys
+
+ARCH_LIST   = ['ppc64le', 's390x', 'amd64', 'aarch64','all']
+URL         = 'https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos'
+VERSIONS    = ('4.6', '4.7', '4.8', '4.9', '4.10')
+COMMANDS    = []
+
+OLD_ARCH_LIST   =  ['ppc64le', 's390x', 'amd64']
+NEW_ARCH_LIST   =  OLD_ARCH_LIST.copy()
+NEW_ARCH_LIST.append('aarch64')
+
+ARCH_ALLOWED = {}
+ARCH_ALLOWED['4.6']    = OLD_ARCH_LIST
+ARCH_ALLOWED['4.7']    = OLD_ARCH_LIST
+ARCH_ALLOWED['4.8']    = NEW_ARCH_LIST
+ARCH_ALLOWED['4.9']    = NEW_ARCH_LIST
+ARCH_ALLOWED['4.10']   = NEW_ARCH_LIST
+
+def usage():
+    '''
+    # Handle parameters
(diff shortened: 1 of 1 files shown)


ravanelli:
@miabbott @mike-nguyen @gursewak1997

hack/create_bump_commit.py
```diff
@@ -0,0 +1,137 @@
...
+    # Handle parameters
+    @param version string rhcos version
+    '''
+
+    parser = argparse.ArgumentParser(
+        description='python boot-image-bumps.py -v 4.6',
+    )
+
+    parser.add_argument("-a", "--arch", choices=ARCH_LIST, default="all")
+    parser.add_argument("-b", "--bugnumber", required=True)
```
patrickdillon, line 43: Is this requiring a bug for every bump? What about when we're bumping during a new release and we don't need a bz?

ravanelli: Thanks for checking it @patrickdillon! Done.

miabbott:
> /retest-required

Since this change isn't going to affect any of the CI jobs, I think we can skip the `/retest` commands.  Just need the :+1 from someone on the installer team.

hack/create_bump_commit.py
```diff
@@ -0,0 +1,141 @@
...
+         arch1 = "x86_64"
+     cmd = 'python update-rhcos-bootimage.py %s/%s/%s/meta.json %s' % (url, release, arch1, arch)
+     runScript(cmd)
+
+def createCommit(bugNumber=None):
+    ## Requires python3-GitPython
+    repo = git.Repo("../")
+    repo.git.add(update=True)
+    if bugNumber:
+        title = ('Bug %s: Bump boot images for RHCOS fixes' % (bugNumber))
```
miabbott, line 122: It would be nice if the title included the RHCOS version, too.

```suggestion
        title = ('Bug %s: Bump boot images for RHCOS %s fixes' % (bugNumber, version))
```

ravanelli: Done!

ravanelli:
> https://github.com/openshift/installer/blob/master/docs/dev/pinned-coreos.md

Umm, looks https://github.com/openshift/installer/commit/d773ee557311fda59565bb82e947d7d55001252b removed the hack used `update-rhcos-bootimage.py` before, where this PR depends on it.

@cgwalters would you mind to explain how the new process for bump will be without this script? So I can understand and also update our internal docs as well?

cgwalters:
> @cgwalters would you mind to explain how the new process for bump will be without this script? So I can understand and also update our internal docs as well?

Ah wow we had a lot of overlapping stuff here.   Hmm...skimming the script here, this seems to be mostly be about synchronizing the old `meta.json` which is unnecessary now, right?

I think the only thing we need now is documented in 
https://github.com/openshift/installer/blob/master/docs/dev/pinned-coreos.md#updating-pinned-stream-metadata
right?  Or am I missing something?

sadasu:
@ravanelli just checking to see if you are still interested in seeing this merged. Also, we are in 4.12 now, this script might need accommodations for 4.11 and 4.12?

sadasu:
@cgwalters @ravanelli should this PR just modify documentation mention in https://github.com/openshift/installer/pull/5223#issuecomment-956345072?

sadasu:
/close
Received confirmation from @ravanelli that it can be closed.

PR #5216, opened 2021-09-15
vendor: Bump OpenShift API to Add Alibaba Platform

patrickdillon:
```
go mod edit -require=github.com/openshift/api@master \
            -replace=k8s.io/client-go=k8s.io/client-go@v0.22.0 \
            -replace=k8s.io/kube-openapi=k8s.io/kube-openapi@v0.0.0-20210305001622-591a79e4bda7
```
Bumps the OpenShift & k8s APIs. There seems to be a breaking change
in kube-openshift which breaks our kubevirt.io dependencies. I'm pinning
this to the earlier kube-openapi until these dependencies can be
updated.

cc @kwoodson 

I believe this is a functional solution. I'm marking this as a work in progress as I am still looking into this breaking change from the kube-openapi, which were introduced, I believe in kubernetes/kube-openapi#234

PR #523, opened 2018-10-23
Update libvirt cluster-api provider config version

bison:
With the move to the kubebuilder CRD based cluster-api components, the
API version for the libvirt provider config subresources has changed
to: `libvirtproviderconfig.k8s.io/v1alpha1`

PR #5224, opened 2021-09-16
Add IPI Support for IBM Power Virtual Servers Offering

clnperez:

We have upstream support for Power VS in the API
(openshift/enhancements#736) and cluster api
provider (https://github.com/openshift/cluster-api-provider-powervs).
Other required work is actively ongoing in the community.

Also actively worked is support for creating networking and boot images
in a way that work with the OCP installer, so our present implemenation
requires that these items be pre-created and passed in via the
install-config. This is temporary and we will add support for IPI-creation
of all required components.

Since Power VS is an offering as part of IBM Cloud (but with seperate APIs for
provisioning), we are able to re-use some of the ibmcloud provider
funtionality. More info in the enhancement proposal on the need to have
a completley seperate provider.

Co-authored-by: Pradipta Banerjee <bpradipta@in.ibm.com>
Co-authored-by: Manjunath Kumatagi <mkumatag@in.ibm.com>
Co-authored-by: Mike Turek <mjturek@us.ibm.com>
Co-authored-by: wolf <cory.klokman@ibm.com>
Co-authored-by: Karthik K.N. <karthik.k.n@ibm.com>


go.mod
```diff
@@ -133,6 +136,7 @@ replace (
 	github.com/metal3-io/baremetal-operator => github.com/openshift/baremetal-operator v0.0.0-20210706141527-5240e42f012a // Use OpenShift fork
 	github.com/metal3-io/baremetal-operator/apis => github.com/openshift/baremetal-operator/apis v0.0.0-20210706141527-5240e42f012a // Use OpenShift fork
 	github.com/metal3-io/cluster-api-provider-baremetal => github.com/openshift/cluster-api-provider-baremetal v0.0.0-20190821174549-a2a477909c1d // Pin OpenShift fork
+	github.com/openshift/api => github.com/openshift-powervs/api v0.0.0-20210729234416-b883aff9d9db
```
mkumatag, line 139: we shouldn't be doing this anymore, we can use the latest code from `github.com/openshift/api` itself

mkumatag (changes requested):

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/Azure/go-ntlmssp v0.0.0-20191115210519-2b2be6cc8ed4 // indirect
 	github.com/ChrisTrenkamp/goxpath v0.0.0-20190607011252-c5096ec8773d // indirect
 	github.com/IBM-Cloud/bluemix-go v0.0.0-20210611051827-cdc80c935c05
+	github.com/IBM-Cloud/power-go-client v1.0.56
 	github.com/IBM-Cloud/terraform-provider-ibm v1.26.2
```
Prashanth684, line 17: i see packages changes in go.mod but no vendored files?

pkg/types/defaults/machinepools.go
```diff
@@ -11,7 +12,10 @@ func SetMachinePoolDefaults(p *types.MachinePool, platform string) {
 	defaultReplicaCount := int64(3)
 	if platform == libvirt.Name {
 		defaultReplicaCount = 1
+	} else if platform == powervs.Name {
+		p.Architecture = "ppc64le"
```
Prashanth684, line 16: isn't the architecture specified explicitly in the control plane and compute. PowerVS only makes sense for ppc64le so there should be a validation failure if the user chooses something other than it.

pkg/types/powervs/defaults/platform.go
```diff
@@ -0,0 +1,34 @@
...
+	return "m5"
+}
+
+// InstanceClasses returns a list of instance "class", in decreasing priority order, which we should use for a given
+// region. Default is m5 then m4 unless a region override is defined in defaultMachineClass.
+func InstanceClasses(region string) []string {
+	if classes, ok := defaultMachineClass[region]; ok {
+		return classes
+	}
+	return []string{"m5", "m4"}
```
Prashanth684, line 33: these need to be changed i assume.

pkg/types/clusterquota.go
```diff
@@ -0,0 +1,11 @@
+package types
+
+import (
+	"github.com/openshift/installer/pkg/types/gcp"
+)
+
+// ClusterQuota contains the size, in cloud quota, of
+// the cluster that was created by installer.
+type ClusterQuota struct {
+	GCP *gcp.Quota `json:"gcp,omitempty"`
```
Prashanth684, line 10: why are these new. these shouldn't have shown up if you had rebased the branch.

pkg/rhcos/powervs_regions.go
```diff
@@ -0,0 +1,70 @@
...
+	"syd":     []string{"syd04"},
+	"sao":     []string{"sao01"},
+	"tor":     []string{"tor01"},
+	"tok":     []string{"tok04"},
+	"us-east": []string{"us-east"},
+}
+
+// PowerVSRegions holds the regions for IBM Power VS, and descriptions used during the survey
+// TODO(cklokman): These are fictional regtions for testing
+var PowerVSRegions = []map[string]string{
```
Prashanth684, line 28: testing regions should probably go under the _test.go files

pkg/destroy/quota/quota.go
```diff
@@ -0,0 +1,28 @@
...
+
+	raw, err := json.Marshal(quota)
+	if err != nil {
+		return errors.Wrap(err, "failed to marshal quota")
+	}
+	if err := ioutil.WriteFile(path, raw, 0777); err != nil {
+		return errors.Wrap(err, "failed to write quota")
+	}
+	return nil
+}
```
Prashanth684, line 28: this shouldn't show up as new either.

pkg/asset/manifests/ibmcloud/cloudproviderconfig_test.go
```diff
@@ -15,11 +15,14 @@ config-file = ""
...
+g2Credentials = /etc/vpc/ibmcloud_api_key
+g2VpcName = ocp4-8pxks-vpc
 g2workerServiceAccountID = 1e1f75646aef447814a6d907cc83fb3c
 
 `
 
-	actualConfig, err := CloudProviderConfig("ocp4-8pxks", "1e1f75646aef447814a6d907cc83fb3c")
+	actualConfig, err := CloudProviderConfig("ocp4-8pxks", "1e1f75646aef447814a6d907cc83fb3c", "us-east")
 	assert.NoError(t, err, "failed to create cloud provider config")
 	assert.Equal(t, expectedConfig, actualConfig, "unexpected cloud provider config")
```
Prashanth684, line 27: these changes shouldn't be showing up as new

pkg/asset/manifests/ibmcloud/cloudproviderconfig.go
```diff
@@ -59,6 +66,9 @@ config-file = {{ if ne .Kubernetes.ConfigFile "" }}{{ .Kubernetes.ConfigFile }}{
 accountID = {{.Provider.AccountID}}
 clusterID = {{.Provider.ClusterID}}
 cluster-default-provider = {{.Provider.ClusterDefaultProvider}}
+region = {{.Provider.Region}}
+g2Credentials = {{.Provider.G2CredentialsFilePath}}
+g2VpcName = {{.Provider.G2VPCName}}
```
Prashanth684, line 71: these changes are not new and not part of this PR

pkg/asset/machines/worker.go
```diff
@@ -163,8 +168,18 @@ func defaultKubevirtMachinePoolPlatform() kubevirttypes.MachinePool {
 	}
 }
 
+func defaultPowerVSMachinePoolPlatform() powervstypes.MachinePool {
+	return powervstypes.MachinePool{
```
Prashanth684, line 172: would be good to document that unlike AWS/Azure etc..there is no "instance type" that can be specified and so we have to define these explicitly.

pkg/types/installconfig.go
```diff
@@ -37,6 +38,7 @@ var (
 		gcp.Name,
 		openstack.Name,
 		ovirt.Name,
+		powervs.Name,
```
mjturek, line 41: Should we move to hidden platforms initially?

go.mod
clnperez: ah, right. i wasn't sure what the process was for that with this repo. do you know what the process is? just run `go mod vendor` and check in changes?

pkg/types/clusterquota.go
clnperez: hm. will look into. thanks

pkg/types/installconfig.go
clnperez: yep

go.mod
Prashanth684: yes. `go mod vendor` should be fine.

pkg/types/defaults/machinepools.go
Prashanth684: yes that's true. but the idea was that the defaultarch would be detected based on payload: https://github.com/openshift/installer/commit/729882a8a2c208a815d35a1923c92d0d0225852b. so if you use an openshift-install binary corresponding to P and create an install config with that, the arch would be set to ppc64le for you irrespective of the platform you are installing on (because the payload was built on a P host). in that sense the "default" when building your own installer will be based on the host system on which you build the installer. the setting of this arch based on platform conflicts with that idea.

clnperez: so, there's going to be a specific installer binary that is an amd/x86 binary, but "corresponding to P"? (b/c clearly i'm not running a binary on my laptop to do an install that was built on a power box)

or -- there needs to be doc stating that if you're installing from a different arch than you're installing onto, you need to manually edit your install-config?

Prashanth684: exactly so for AWS on ARM : https://console.redhat.com/openshift/install/aws/arm we have an option to select an openshift-install binary which is an x86 binary but which is built for an ARM payload, hence the install-config will  have the arch set to `arm64` when created through this binary. we would need to have the same for PowerVS too. cc @yselkowitz .

Prashanth684: and if you locally build the installer and use that install binary, you would have to manually edit the architecture. So, the tl;dr is - arch is based on payload. If you use an openhsift-install binary from the payload you shouldn't worry about the arch. it will be set correctly. if you build the install binary locally, you will have to change the arch manually (i.e, if you are not building on a host corresponding to the arch of the platform)

yselkowitz: As of 4.9 all payloads regardless of architecture have a linux/amd64 installer in addition to linux/"native".  Also, when you get the installer from the payload, it will default to the payload architecture, regardless of the installer binary runtime arch.  If you build the installer yourself, you can mimic this by running e.g. `DEFAULT_ARCH=ppc64le ./hack/build.sh`

go.mod
clnperez: looks like https://github.com/openshift/installer/pull/5216 will be the one they'll pick up

pkg/rhcos/powervs_regions.go
clnperez: ah, misleading comment there. testing in this context should have been _development_. that TODO needs to removed and these zones need to be triple-checked. i missed a few stray TODOs there we should clean up -- including the one about programmatic generation (since we don't have a way to query zones)

clnperez:
looks like now we're hitting https://github.com/openshift/installer/pull/523

clnperez:
/retest
hoping those jq errors go away

pkg/types/powervs/OWNERS
```diff
@@ -0,0 +1,7 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+# This file just uses aliases defined in OWNERS_ALIASES.
+
+approvers:
+  - multiarch-approvers
```
Prashanth684, line 5: these aliases need to be mapped to people in OWNERS_ALIASES

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -434,6 +434,58 @@ spec:
...
+                          description: SysType defines the system type for instance.
+                          type: string
+                        volumeIDs:
+                          description: VolumeIDs is the list of volumes attached to
+                            the instance.
+                          items:
+                            type: string
+                          type: array
+                      required:
+                      - imageName
```
Prashanth684, line 481: imageName shouldn't be required. this should come from the rhcos stream json once we have the pipeline setup and generating images

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -434,6 +434,58 @@ spec:
...
+                      - keypairname
+                      - memory
+                      - name
+                      - networkIDs
+                      - processors
+                      - serviceinstance
```
Prashanth684, line 487: can't we have well defined defaults for most of these fields so that most of them are optional ?

data/data/install.openshift.io_installconfigs.yaml
Prashanth684: hmm..are these the boot images and not the ovas? i think long term you want the bootimage creation handled by TF or by a separate component in the installer itself by consuming the ova image i would think. in any case because ovas are what are coming from the pipeline, this needs to be optional.

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -2059,6 +2163,136 @@ spec:
...
+                  SSHKeyName:
+                    description: SSHKeyName is the name of an SSH key stored in the
+                      Service Instance.
+                    type: string
+                  bootstrapOSImage:
+                    description: BootstrapOSImage is a pre-created boot image to override
+                      the default OS image for the bootstrap node.
+                    type: string
+                  clusterOSImage:
+                    description: "ClusterOSImage is a pre-created Power VS boot image
```
Prashanth684, line 2182: once again these are referring to the bootimages - which is not what is supplied through the rhcos json. these would be internally created through the ova images?

data/data/install.openshift.io_installconfigs.yaml
clnperez: I'm not clear on how to make this file show what we'd like to be required in the install-config.yaml. These seem to be pulled from the Config (type MachinePool) in the install config. Those items _are_ required (see https://github.com/openshift/terraform-provider-ibm/blob/master/website/docs/r/pi_instance.html.markdown). But they're not required in the install-config.yaml itself. 

To answer your questions though, -- yes, they should come from the rhcos stream json by default; and yes, they are the boot images; and yes, long-term the bootimage creation will be handled by TF.

data/data/install.openshift.io_installconfigs.yaml
clnperez: same comment as above. we do have defaults though

data/data/install.openshift.io_installconfigs.yaml
clnperez: lol, once again -- bootimages. and yes :)

data/data/install.openshift.io_installconfigs.yaml
clnperez: ah. i just wasn't looking at the file as a whole. there's a compute section, a machinepools section, and a platform section all for powervs. and i didn't understand how the +optional directives would be used for each of those things w/in the code. after looking more closely, iiuc, every time i listed something as +optional, it was referring to it being defined in the install config. so i'll go band and update those. 

this leads to the question though -- for everything that shows up in this yaml (install.openshift.io_installconfigs.yaml) -- we don't have to make it configurable via the install-config.yaml do we?

Prashanth684: so this is my view - things that can be configured optionally (for example through providerspec in the manifests) should not even be included in the install config. the user can modify them after doing a `openshift-install create manifests` and modifying the provider spec which will then be pulled in by the tfvars for master and for workers, it will be handled by the cluster api provider. so IMO those options shouldn't even be exposed here. Only things which are absolutely essential to change and makes sense to do so should be mentioned here like image name, region and even for that we need to have defaults. basically we are trying to make the users life easier and not complicate it by providing too many options even though they are optional.

Prashanth684: so tl;dr is don't even define the ones that are optional in `pkg/types/powervs/platform.go` . they can be changed through the provider spec defined in cluster-api-provider-powervs. only add things to the platform.go that absolutely must be exposed to the user through install-config

clnperez:
Closing this PR and will submit smaller PRs for review. Starting with #5270 

fyi @Prashanth684 @mkumatag @mjturek @23TNC

PR #5230, opened 2021-09-20
libvirt: bump memory of machines to 16G in the provider config

Prashanth684:
While the terraform defaults for the libvirt master memory size was set to 16G through this PR:https://github.com/openshift/installer/pull/5069, the sizes were still fixed to 8G because they were being overwritten by the value set in the provider. Change the provider value to 16G to address this issue.

Note: This will also set the worker memory size to 16G which should be fine, but can also be overriden by changing the machineset manifest.

### DIFF
--- a/pkg/asset/machines/libvirt/machines.go
+++ b/pkg/asset/machines/libvirt/machines.go
@@ -63,7 +63,7 @@ func provider(clusterID string, networkInterfaceAddress string, platform *libvir
 			APIVersion: "libvirtproviderconfig.openshift.io/v1beta1",
 			Kind:       "LibvirtMachineProviderConfig",
 		},
-		DomainMemory: 8192,
+		DomainMemory: 16384,
 		DomainVcpu:   4,
 		Ignition: &libvirtprovider.Ignition{
 			UserDataSecret: userDataSecret,


Prashanth684:
cc @praveenkumar @cfergeau

Prashanth684:
@praveenkumar this will also set the worker memory to 16G , is that ok? or do we want to keep that at 8G?

pkg/asset/machines/libvirt/machines.go
```diff
@@ -63,7 +63,7 @@ func provider(clusterID string, networkInterfaceAddress string, platform *libvir
 			APIVersion: "libvirtproviderconfig.openshift.io/v1beta1",
 			Kind:       "LibvirtMachineProviderConfig",
 		},
-		DomainMemory: 8192,
+		DomainMemory: 16384,
```
praveenkumar, line 66: @Prashanth684 we are using https://cloud.google.com/compute/docs/general-purpose-machines#n1_machines (n1-standard-16) and if we are increased the resource that much it means we are not able to accommodate cluster in this machine (I think standard e2e-libvirt runs 3master-3worker mode).

Prashanth684: hmm..ok yeah n1-standard-15 has only 60G..so should we reduce worker nodes to 2 and keep it at 8192 and increase master memory to 12G then ?

praveenkumar: @Prashanth684 I will take a look tomorrow how it all link together and update you here, first we need to manual create a gcp instance (same resource) and try to start the cluster and see if it able to execute test as expected.

cfergeau: Did you get a chance to take a look at this?

Prashanth684: i did kick off a libvirt e2e and as I am monitoring, i see that Praveen is right. the standard machines have 64G memory which is not sufficient for this change. But even otherwise, i think we need to look into bumping the gcp instance type to something beefier because the libvirt master defaults need to be  16G anyway for consistent CI runs

barbacbd: @Prashanth684 this should be accomplished in [#5841](https://github.com/openshift/installer/pull/5841)

sadasu:
@Prashanth684  It appears that we are not pursuing this change anymore. If that is the case, could you please close this PR?

PR #5232, opened 2021-09-22
Bug 2021364: aws: remove invalid s3 permission

wedge-jarrad:
s3:GetBucketReplication is not a valid IAM action. The correct action is s3:GetReplicationConfiguration which is also included.

https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html

Since the installer attempts to verify that it has all of the permissions that it thinks it needs, it fails when it checks for this permission and aborts the install. This means that we must include the invalid permission in our IAM policy and ignore the warnings that this generates in order to run the cluster install.


patrickdillon:
The documentation confirming `s3:GetReplicationConfiguration` is the proper permission is here: https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketReplication.html

patrickdillon:
/retitle Bug 2021364: aws: remove invalid s3 permission

PR #5233, opened 2021-09-22
Bug 2006291: baremetal: Prevent ProvisioningNetworkCIDR with host bits set

derekhiggins:
If ProvisioningNetworkCIDR is provided to "cidr.Host"
with host bits set e.g. 172.22.0.1/24 vs 172.22.0.0/24,
then the same host bits are set in the
BootstrapProvisioningIP and ClusterProvisioningIP.

This happens because "cidr.Host" is expecting a network address
as returned by net.ParseCIDR but ipnet.MustParseCIDR returns
a host CIDR.

### DIFF
--- a/pkg/types/baremetal/validation/platform.go
+++ b/pkg/types/baremetal/validation/platform.go
@@ -413,6 +413,14 @@ func ValidateProvisioning(p *baremetal.Platform, n *types.Networking, fldPath *f
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("clusterProvisioningIP"), p.ClusterProvisioningIP, fmt.Sprintf("%q is not in the provisioning network", p.ClusterProvisioningIP)))
 		}
 
+		// Ensure provisioningNetworkCIDR does not have any host bits set
+		expectedIP := p.ProvisioningNetworkCIDR.IP.Mask(p.ProvisioningNetworkCIDR.Mask)
+		expectedLen, _ := p.ProvisioningNetworkCIDR.Mask.Size()
+		if !p.ProvisioningNetworkCIDR.IP.Equal(expectedIP) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("provisioningNetworkCIDR"), p.ProvisioningNetworkCIDR,
+				fmt.Sprintf("provisioningNetworkCIDR has host bits set, expected %s/%d", expectedIP, expectedLen)))
+		}
+
 		if err := validateIPNotinMachineCIDR(p.BootstrapProvisioningIP, n); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/types/baremetal/validation/platform.go
```diff
@@ -405,6 +405,11 @@ func ValidateProvisioning(p *baremetal.Platform, n *types.Networking, fldPath *f
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("clusterProvisioningIP"), p.ClusterProvisioningIP, fmt.Sprintf("%q is not in the provisioning network", p.ClusterProvisioningIP)))
 		}
 
+		// Ensure provisioningNetworkCIDR does not have any host bits set
+		if(string(p.ProvisioningNetworkCIDR.IP) != string(p.ProvisioningNetworkCIDR.IP.Mask(p.ProvisioningNetworkCIDR.Mask))){
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("provisioningNetworkCIDR"), p.ProvisioningNetworkCIDR, fmt.Sprintf("%q has host bits set", p.ProvisioningNetworkCIDR)))
```
hardys, line 410: Just to give users the best possible chance of understanding how to fix this, could we modify the output to say something like:

```
172.22.0.2/24 has host bits set, should be e.g 172.22.0.0/24
```

When described that way, it's clear exactly what host bits refers to, and how to fix it?

hardys:
/label platform/baremetal

lgtm - just one comment about the error message, could we calculate the expected value and show that in the message?

pkg/types/baremetal/validation/platform.go
```diff
@@ -413,6 +413,14 @@ func ValidateProvisioning(p *baremetal.Platform, n *types.Networking, fldPath *f
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("clusterProvisioningIP"), p.ClusterProvisioningIP, fmt.Sprintf("%q is not in the provisioning network", p.ClusterProvisioningIP)))
 		}
 
+		// Ensure provisioningNetworkCIDR does not have any host bits set
+		expectedIP := p.ProvisioningNetworkCIDR.IP.Mask(p.ProvisioningNetworkCIDR.Mask)
+		expectedLen, _ := p.ProvisioningNetworkCIDR.Mask.Size()
+		if string(p.ProvisioningNetworkCIDR.IP) != string(expectedIP) {
```
hardys, line 419: I think it may be cleaner to do `if ! p.ProvisioningNetworkCIDR.IP.Equal(expectedIP)` here?

FWIW the reason I initially thought that was I was worried the string comparison wouldn't handle long format ipv6 CIDRs, that's not actually the case but if you wanted to grab the tests to prove it I pushed them to https://github.com/hardys/installer/commit/b487e94a7539e418503cb165e80df32554eb4aed

PR #5234, opened 2021-09-22
Inject subnet names into IBM Cloud provider config

cjschaef:
Add the set of subnet names created for the VPC into the IBM Cloud
provider configuration.


jstuever (approved):
/lgtm

PR #5236, opened 2021-09-23
Bug 1955544: Add ingress rules to master SG for compact clusters

mandre:
In the case of compact clusters, where master nodes are schedulable, we
need to open ports for ingress, same as for workers.

### DIFF
--- a/data/data/openstack/topology/variables.tf
+++ b/data/data/openstack/topology/variables.tf
@@ -39,6 +39,10 @@ variable "masters_count" {
   type = string
 }
 
+variable "masters_schedulable" {
+  type = bool
+}
+
 variable "api_int_ip" {
   type = string
 }
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -10,6 +10,7 @@ import (
 
 	igntypes "github.com/coreos/ignition/v2/config/v3_2/types"
 	coreosarch "github.com/coreos/stream-metadata-go/arch"
+	"github.com/ghodss/yaml"
 	gcpprovider "github.com/openshift/cluster-api-provider-gcp/pkg/apis/gcpprovider/v1beta1"
 	kubevirtprovider "github.com/openshift/cluster-api-provider-kubevirt/pkg/apis/kubevirtprovider/v1alpha1"
 	kubevirtutils "github.com/openshift/cluster-api-provider-kubevirt/pkg/utils"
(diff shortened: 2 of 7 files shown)


mandre:
I'm trying to see if there isn't a way to add these rules dynamically when the master nodes are made schedulable.
/hold

mandre:
Hi @staebler, could you PTAL on this patch? Especially on the second commit - I wonder if it's fine to add a dependency on the Manifests asset like I did to generate TF variables.

/assign @staebler

pkg/asset/cluster/tfvars.go
```diff
@@ -148,6 +153,16 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 	}
 
 	masterCount := len(mastersAsset.MachineFiles)
+	schedulerConfig := configv1.Scheduler{}
+	if len(schedulerAsset.Files()) != 1 {
```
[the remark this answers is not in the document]
mandre: Oh, good catch. I'll see if I can change it.

pkg/asset/cluster/tfvars.go
```diff
@@ -148,6 +154,19 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 	}
 
 	masterCount := len(mastersAsset.MachineFiles)
+	mastersSchedulable := false
+	for _, f := range manifestsAsset.Files() {
+		if filepath.Base(f.Filename) == "cluster-scheduler-02-config.yml" {
+			schedulerConfig := configv1.Scheduler{}
+			err = yaml.Unmarshal(f.Data, &schedulerConfig)
+			if err != nil {
+				return err
```
staebler, line 163: Please wrap this error in a more descriptive message.

pkg/asset/cluster/tfvars.go
```diff
@@ -148,6 +154,19 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 	}
 
 	masterCount := len(mastersAsset.MachineFiles)
+	mastersSchedulable := false
+	for _, f := range manifestsAsset.Files() {
+		if filepath.Base(f.Filename) == "cluster-scheduler-02-config.yml" {
```
staebler, line 159: Should we make `schedulerCfgFilename` in pkg/asset/manifests/scheduler.go public so that we are not duplicating the filename here?

data/data/openstack/topology/variables.tf
```diff
@@ -39,6 +39,10 @@ variable "masters_count" {
   type = string
 }
 
+variable "masters_schedulable" {
```
staebler, line 42: I am thinking that this may be a useful variable to add to data/data/config.tf. It is obviously only be used by OpenStack now, but the concept is not specific to OpenStack.

pkg/asset/cluster/tfvars.go
mandre: I was wondering about it too, and was bother by the fact that schedulerCfgFilename is a var and not a const. That said, it's indeed cleaner to re-use the same var.

data/data/openstack/topology/variables.tf
staebler: Duh. I don't know what I was looking at at the time. Sorry.

pkg/asset/cluster/tfvars.go
staebler: Unfortunately, it cannot be a const since it is the result of a function call. So, to make it a const, you would have to not use `filepath.Join`

staebler (approved):
/lgtm

PR #5239, opened 2021-09-24
WIP | aws: changing default EBS type on masters to gp3.

mtulio:
This is WIP validation to support gp3 as default on IPI, due the benefits compared to gp2.

Overview of improvements of gp3 compared to gp2:
- The baseline IOPS was increased to 3,000 : gp2 volumes is allowed to burst up to 3,000 IOPS with a baseline of 1:3 GiB:IOPS, so volumes with low capacity will have a low baseline and can be impacted with long periods of I/O
- Now it is allowed to increase the throughput regardless of capacity with a baseline of 125 MiB/s (extra fee is charged per MiB/s provisioned above the baseline)
- Now it is allowed to increase the IOPS regardless of capacity with a baseline of 3,000 IOPS (extra fee is charged per IOPS provisioned above the baseline)
- Cheaper : The cost of gp3 is 20% lower than gp2 in a standard baseline

More details on the study case on SPLAT-253.


hexfusion:
cc @rsevilla87 @lilic @hasbro17 @jtaleric

hexfusion:
> Cheaper : The cost of gp3 is 20% lower than gp2 in a standard baseline

To clarify in this exact case if we were to use gp3 as IPI default it would be 20% lower cost?

mtulio:
> To clarify in this exact case if we were to use gp3 as IPI default it would be 20% lower cost?

@hexfusion yes, it will be 20% lower cost. Currently the default IPI is 120 GiB of gp2 (baseline of 360 IOPS [burst up to 3k] and 128MiB/s throughput*), with gp3 the baseline IOPS will be increased 10x to 3,000 and throughput to will decrease a bit to 125 MiB/s)

'* The throughput limit is between 128 MiB/s and 250 MiB/s, depending on the volume size. Volumes smaller than or equal to 170 GiB deliver a maximum throughput of 128 MiB/s. Volumes larger than 170 GiB but smaller than 334 GiB deliver a maximum throughput of 250 MiB/s if burst credits are available. [EBS doc](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html)

hexfusion:
gp3
image

gp2
image

aws-e2e seems to show very similar storage metrics. perf and scale team will try some more tests next week. But if we save 20% then why not.

hexfusion:
The above is a mixture of fsync and backend commit so I have separated those.
fsync gp3
image

fsync gp2
image

mtulio:
@hexfusion yeah, the perf should be similar, no much perceptions in a fresh cluster, but as described in the study case, in a stress scenario (gp2 with no burst credits) gp3 should be much faster and more resilient than gp2.

staebler:
My one concern here is with regards to backwards compatibility. Are there instance types that a user could select that do not support gp3? We cannot have an install-config.yaml that was valid in one OpenShift version and no longer valid in a later OpenShift version.

mtulio:
@staebler ptal? the last one for gp3! \o/

``` shell
$ oc get  infrastructures cluster -o jsonpath='{.status.infrastructureName}'
mrbgp3final-7h9rc

NAME                                       VERSION                              AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
authentication                             4.10.0-0.nightly-2021-12-10-123024   True        False         False      27s     
baremetal                                  4.10.0-0.nightly-2021-12-10-123024   True        False         False      16m     
cloud-controller-manager                   4.10.0-0.nightly-2021-12-10-123024   True        False         False      19m     
cloud-credential                           4.10.0-0.nightly-2021-12-10-123024   True        False         False      18m     
cluster-api                                4.10.0-0.nightly-2021-12-10-123024   True        False         False      18m     
cluster-autoscaler                         4.10.0-0.nightly-2021-12-10-123024   True        False         False      16m     
config-operator                            4.10.0-0.nightly-2021-12-10-123024   True        False         False      18m     
console                                    4.10.0-0.nightly-2021-12-10-123024   True        False         False      4m17s   
csi-snapshot-controller                    4.10.0-0.nightly-2021-12-10-123024   True        False         False      17m     
dns                                        4.10.0-0.nightly-2021-12-10-123024   True        False         False      16m     
etcd                                       4.10.0-0.nightly-2021-12-10-123024   True        False         False      16m     
image-registry                             4.10.0-0.nightly-2021-12-10-123024   True        False         False      7m54s   
ingress                                    4.10.0-0.nightly-2021-12-10-123024   True        False         False      7m13s   
insights                                   4.10.0-0.nightly-2021-12-10-123024   True        False         False      11m     
kube-apiserver                             4.10.0-0.nightly-2021-12-10-123024   True        False         False      8m48s   
kube-controller-manager                    4.10.0-0.nightly-2021-12-10-123024   True        False         False      15m     
kube-scheduler                             4.10.0-0.nightly-2021-12-10-123024   True        False         False      15m     
kube-storage-version-migrator              4.10.0-0.nightly-2021-12-10-123024   True        False         False      18m     
machine-api                                4.10.0-0.nightly-2021-12-10-123024   True        False         False      8m19s   
machine-approver                           4.10.0-0.nightly-2021-12-10-123024   True        False         False      17m     
machine-config                             4.10.0-0.nightly-2021-12-10-123024   True        False         False      7m49s   
marketplace                                4.10.0-0.nightly-2021-12-10-123024   True        False         False      17m     
monitoring                                 4.10.0-0.nightly-2021-12-10-123024   True        False         False      7m8s    
network                                    4.10.0-0.nightly-2021-12-10-123024   True        False         False      18m     
node-tuning                                4.10.0-0.nightly-2021-12-10-123024   True        False         False      17m     
openshift-apiserver                        4.10.0-0.nightly-2021-12-10-123024   True        False         False      9m28s   
openshift-controller-manager               4.10.0-0.nightly-2021-12-10-123024   True        False         False      7m33s   
openshift-samples                          4.10.0-0.nightly-2021-12-10-123024   True        False         False      11m     
operator-lifecycle-manager                 4.10.0-0.nightly-2021-12-10-123024   True        False         False      18m     
operator-lifecycle-manager-catalog         4.10.0-0.nightly-2021-12-10-123024   True        False         False      17m     
operator-lifecycle-manager-packageserver   4.10.0-0.nightly-2021-12-10-123024   True        False         False      11m     
service-ca                                 4.10.0-0.nightly-2021-12-10-123024   True        False         False      18m     
storage                                    4.10.0-0.nightly-2021-12-10-123024   True        False         False      17m 

$ aws ec2 describe-volumes --filters "Name=tag:kubernetes.io/cluster/mrbgp3final-7h9rc,Values=owned" |jq -r '.Volumes[].VolumeType' |sort |uniq -c
      6 gp3

$ aws ec2 describe-instances --filters "Name=tag:kubernetes.io/cluster/mrbgp3final-7h9rc,Values=owned" |jq -r '.Reservations[].Instances[].InstanceType' |sort |uniq -c
      3 m6i.large
      4 m6i.xlarge
```

mtulio:
> My one concern here is with regards to backwards compatibility. Are there instance types that a user could select that do not support gp3? We cannot have an install-config.yaml that was valid in one OpenShift version and no longer valid in a later OpenShift version.

I didn't remember to read, and I didn't find reference on the doc, about the limitations of using gp3 in old instance generations. There's a recommendation to use EBS-Optimized instances to gain better performance, but it's not a limitation. Furthermore, it's possible to migrate from gp2 to gp3 without needing to detach the volume. AWS is also retiring old instances frequently, for example, now it's available only Gen4 and later.

[Note](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSPerformance.html) about EBS-Optimized:
***Some EBS-optimized instance configurations incur an extra cost (such as C3, R3, and M3), while others are always EBS-optimized at no extra cost (such as M4, C4, C5, and D2).***

We can see that the current instance generation has already EBS-Optimized enabled, and no extra cost.

Reference:
- [EC2 page](https://aws.amazon.com/ec2/instance-types/)
- [AWS Blog/Migrate your Amazon EBS volumes from gp2 to gp3 and save up to 20% on costs](https://aws.amazon.com/blogs/storage/migrate-your-amazon-ebs-volumes-from-gp2-to-gp3-and-save-up-to-20-on-costs/)

staebler:
Looking over the instances that support gp3, it looks like it covers all of the instance type generations that we support: namely m4, c4, r4, t3, and later.
/lgtm
/approve

PR #5247, opened 2021-09-27
Generate network config secrets for baremetal platform

andfasano:
This patch adds the generation of secrets to store the network configuration defined per host in the baremetal platform.

Currently based on PR #5207


andfasano:
/hold

Requires #5207 to land first

zaneb:
I think all the prerequisites for this have landed. @andfasano could you rebase?

/hold cancel

andfasano:
requires https://github.com/openshift/installer/pull/5441

zaneb:
@andfasano @ardaguclu given that both this and #5441 require tests to pass for All of The Things, I think we should submit them both in the same PR. That should result in this merging a lot faster.

pkg/asset/machines/master.go
```diff
@@ -606,3 +609,31 @@ func IsMachineManifest(file *asset.File) bool {
 		return matched
 	}
 }
+
+func generateAsset(objects interface{}, fileName string) (assets []*asset.File, err error) {
+	switch reflect.TypeOf(objects).Kind() {
```
staebler, line 614: Do not use reflection as a means to avoid code duplication. Either create a `[]interface{}` to pass into the function or have the function operator on each element individually. I realize that there will still be quite a bit of boilerplate code with either approach, but that is go.

pkg/asset/machines/master.go
```diff
@@ -606,3 +609,31 @@ func IsMachineManifest(file *asset.File) bool {
 		return matched
 	}
 }
+
+func generateAsset(objects interface{}, fileName string) (assets []*asset.File, err error) {
```
staebler, line 613: Do not overload the term asset, which has a specific meaning in the context of the installer.
```suggestion
func generateAssetFile(objects interface{}, fileName string) (assets []*asset.File, err error) {
```

pkg/asset/machines/baremetal/hosts.go
```diff
@@ -103,6 +120,15 @@ func Hosts(config *types.InstallConfig, machines []machineapi.Machine) (*HostSet
 		}
 		newHost := createBaremetalHost(host, bmc)
 
+		if host.NetworkConfig != "" {
+			networkConfigSecret := createNetworkConfigSecret(host)
+			settings.NetworkConfigSecrets = append(settings.NetworkConfigSecrets, networkConfigSecret)
+			// newHost.Spec.PreprovisioningNetworkDataName = &corev1.SecretReference{
```
staebler: Remove this commented out code, especially as it is removed in a latter commit in this PR.

pkg/asset/machines/master.go
andfasano: Unfortunately it's not possible to use `[]interface{}` as function parameter as [disallowed](https://go.dev/doc/faq#convert_slice_of_interface) by the language. I've thus kept the conversion code within the `generateAsset` boundaries to minimize the noise outside, but that small portion cannot be simplified further, so I guess this could be a reasonable compromise between not using reflection and removing code duplication (while operating on each element individually appeared to make the code more complicated). I guess that in future Type Parameters could help for that.

pkg/asset/machines/baremetal/hosts.go
andfasano: Done

pkg/asset/machines/master.go
```diff
@@ -606,3 +610,37 @@ func IsMachineManifest(file *asset.File) bool {
 		return matched
 	}
 }
+
+func generateAsset(elements interface{}, fileName string) (assets []*asset.File, err error) {
```
ardaguclu, line 614: This change's main idea looks good to me to avoid repetitions. But I think, it would be better to have 2 functions, `generateAssetForHost`, `generateAssetForSecret` instead this generic function, thanks to that, we can pass concrete types instead interface{} and we can benefit from compile time checks.

andfasano: Thanks @ardaguclu good idea

pkg/asset/machines/master.go
staebler: That is exactly what I was trying to suggest with "create a `[]interface{}` to pass into the function".

pkg/asset/machines/master.go
```diff
@@ -606,3 +610,42 @@ func IsMachineManifest(file *asset.File) bool {
...
+
+	var objects []interface{}
+	for _, r := range resources {
+		objects = append(objects, r)
+	}
+
+	return generateAssets(objects, fileName)
+}
+
+func generateAssets(objects []interface{}, fileName string) (assets []*asset.File, err error) {
```
staebler, line 634: Still looking for this function (and now the other two generateXXXAssets) functions to be renamed since the functions do not generate assets.

pkg/asset/machines/master.go
```diff
@@ -606,3 +610,42 @@ func IsMachineManifest(file *asset.File) bool {
...
+
+	assets = make([]*asset.File, len(objects))
+	padFormat := fmt.Sprintf("%%0%dd", len(fmt.Sprintf("%d", len(objects))))
+	for i := 0; i < len(objects); i++ {
+		data, err := yaml.Marshal(objects[i])
```
staebler, line 639:
```suggestion
	for i, obj := range objects {
		data, err := yaml.Marshal(obj)
```

pkg/asset/machines/master.go
staebler, line 634: The use of named result parameters here does not add anything to document the behavior of the function. As such, please refrain from naming the result parameters.

staebler (commented):
Looks good. Just a few nits.

staebler (commented):
/lgtm
/approve

PR #2658, opened 2019-11-12
Bug 1771564: pkg/asset/manifests/proxy: Link trustedCA for transparent proxies

wking:
Folks may want to configure additional CAs even in the absence of an explicitly-configured proxy (e.g. because they are using a transparent man-in-the-middle proxy).  With this commit, we set the proxy config to point at the trusted CA bundle regardless of whether the httpProxy or httpsProxy properties were set.

This also opens us up to folks who want to set additional CAs for other purposes (e.g. connecting to a local registry) by leaning on the network-operator's injection tooling and the proxy config's cluster-scoped `trustedCA` property.  That's cheating a bit, and for things like registry access there may be more appropriate operator-level configs that we could be setting.  But the installer-level `additionalTrustBundle` isn't documented as proxy-specific, so I think we want to adjust the installer as we go to put the configured trust bundle in the appropriate place to get the cluster components to pick it up, and at the moment the proxy config is that place.

PR #5251, opened 2021-09-29
Ensure cluster proxy object sets TrustedCA if only AdditionalTrustBundle is provided

montaguethomas:
Before this change, the installer would generate an empty spec for the cluster proxy object unless `installConfig.Config.Proxy` is defined.

This change implements support for the use cases where there is no proxy configured, but there is a desire to include additional trusted CAs.

Related:
- https://access.redhat.com/solutions/6063031
- Case #03038183

### DIFF
--- a/pkg/asset/manifests/proxy.go
+++ b/pkg/asset/manifests/proxy.go
@@ -59,20 +59,19 @@ func (p *Proxy) Generate(dependencies asset.Parents) error {
 			Name: "cluster",
 			// not namespaced
 		},
+		Spec = configv1.ProxySpec{},
+	}
+
+	if installConfig.Config.AdditionalTrustBundle != "" {
+		p.Config.Spec.TrustedCA = configv1.ConfigMapNameReference{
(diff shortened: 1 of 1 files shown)


wking:
This has been rejected before: #2658, #4082.  I recommend looking at [rhbz#2009629][1].  I'm going to close this based on the previous two PRs, but I'm not really an installer maintainer anymore, so if any installer maintainers feel like I'm overreaching, feel free to re-open.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=2009629

montaguethomas:
I'll pass these resources along to our support case and our TAM. Guess for now we'll just have to keep using our hack:

```
openshift-install create manifests
[[ -f ./manifests/user-ca-bundle-config.yaml ]] \
  && yq e -P -i '.spec.trustedCA.name = "user-ca-bundle"' ./manifests/cluster-proxy-01-config.yaml
openshift-install create cluster
```

PR #5252, opened 2021-09-29
Remove old RHCOS metadata (not stream metadata)

cgwalters:
In https://github.com/openshift/enhancements/pull/679
we landed support for a stream metadata format already used
by FCOS and now by RHCOS/OCP consumers to find the bootimages.

We kept the old metadata because the UPI CI jobs used it.
In https://github.com/openshift/release/pull/17482 I tried
to port the UPI CI jobs, but ran into huge levels of technical debt.

As best I can tell, the UPI CI jobs are not running on this repo
now and are likely broken for other reasons.

Let's remove the old data and avoid the confusing duplication.
Anyone who goes to work on the UPI CI jobs and sanitizes things
should be able to pick up the work to port to stream metadata
relatively easily.


miabbott:
/test e2e-aws-upi

let's just see how a UPI job handles this

bgilbert:
Should this PR remove `hack/update-rhcos-bootimage.py` also?

cgwalters:
> Should this PR remove hack/update-rhcos-bootimage.py also?

Yep, done!

Bigger picture, I think the next step here is adding a small controller process that generates the merged stream metadata server side by default, then we just copy from that.

bgilbert (approved):
LGTM

travier (approved):

cgwalters:
All the CI failures here seem unrelated to this PR; for e.g. `e2e-aws-fips` we got into cluster provisioning but failed tests, and there's no way this PR can cause test failures.  traditional rhel worker jobs are failing to find packages.  The UPI job seems to have bitrotted enough it fails trying to run much earlier than the code would be looking for the old `rhcos.json`.

images/installer/Dockerfile.upi.ci.rhel8
```diff
@@ -15,7 +15,8 @@ FROM registry.ci.openshift.org/ocp/4.8:base
 COPY --from=cli /usr/bin/oc /bin/oc
 COPY --from=builder /go/src/github.com/openshift/installer/bin/openshift-install /bin/openshift-install
 COPY --from=builder /go/src/github.com/openshift/installer/upi /var/lib/openshift-install/upi
-COPY --from=builder /go/src/github.com/openshift/installer/data/data/rhcos.json /var/lib/openshift-install/rhcos.json
+# This was removed, see https://github.com/openshift/enhancements/pull/679
+# COPY --from=builder /go/src/github.com/openshift/installer/data/data/rhcos.json /var/lib/openshift-install/rhcos.json
```
staebler, line 19: Why do we need to keep a comment that this was removed, as opposed to just removing it?

staebler (commented):
I am on board with ripping out the old metadata. I will create a card for the installer team to update the CI jobs appropriately.

This needs a rebase.

/approve

cgwalters:
Oops, right.  

Man, I love that when Prow comments like:

- This looks broken
(second later)
- This looks quite broken
(second later)
- WHAT IS THIS GARBAGE
(second later)
- I CAN'T EVEN
(second later)
- I'M DONE WITH HUMANS, GOING TO GO TALK TO SKYNET ABOUT THIS
(ominous silence)

data/data/coreos/OWNERS
```diff
@@ -0,0 +1,20 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+# This file should mirror the list of users defined on coreos-approvers in:
+# https://github.com/openshift/release/blob/master/OWNERS_ALIASES
+
+approvers:
+  - miabbott
```
staebler, line 6: Please add this list as an alias in OWNER_ALIAS instead of here. That way we can use the same list for multiple OWNER files.
https://github.com/openshift/installer/blob/c653122585431b6beb9759efc0a67eb8de4a6d28/OWNERS_ALIASES#L3

staebler (commented):
/approve

I already approved earlier, but I wanted to make it explicitly that I also approve of the added commit the coreos team as owners.

PR #5253, opened 2021-09-29
OKD: bump FCOS 34 boot images

LorbusChris:


LorbusChris:
/hold
The issue at hand might not yet be fixed in FCOS

PR #5254, opened 2021-09-29
update-rhcos-bootimage: check that RHCOS metadata has necessary artif…

miabbott:
…acts

It is possible to feed the `update-rhcos-bootimage.py` script a
`meta.json` input that does not contain all the boot image artifacts
required when doing a boot image bump. This changes the script to bail
out if the length of the `images` dict is less than or equal to 2,
which indicates that only the `ostree` tarball and `qemu` qcow2 image
were produced as part of an RHCOS build.

### DIFF
--- a/hack/update-rhcos-bootimage.py
+++ b/hack/update-rhcos-bootimage.py
@@ -23,6 +23,16 @@ with urllib.request.urlopen(args.meta) as f:
     string_f = codecs.getreader('utf-8')(f)  # support for Python < 3.6
     meta = json.load(string_f)
 newmeta = {}
+
+# x86_64 boot image bumps should contain more than just the ostree tarball and
+# qemu qcow2 image. check the contents of the images dict for well known artifacts.
+if args.arch == 'amd64':
+    for k in ['aws', 'gcp', 'metal', 'metal4k']:
+        if k not in meta['images'].keys():
+            raise SystemExit("You appear to be missing most of the x86_64 boot "
+                             + "image artifacts. Check the meta.json for this "
+                             + "RHCOS version. Also check that boot images were "
+                             + "enabled in the RHCOS build pipeline.")
 for k in ['images', 'buildid', 'oscontainer',
           'ostree-commit', 'ostree-version',
           'azure', 'gcp']:


cgwalters (approved):

hack/update-rhcos-bootimage.py
```diff
@@ -23,6 +23,12 @@
     string_f = codecs.getreader('utf-8')(f)  # support for Python < 3.6
     meta = json.load(string_f)
 newmeta = {}
+
+# x86_64 boot image bumps should contain more than just the ostree tarball and
+# qemu qcow2 image
+if args.arch == 'amd64' and len(meta['images']) <= 2:
```
bgilbert, line 29: `2` seems pretty brittle.  Can we spot-check for specific artifacts we expect to exist?

hack/update-rhcos-bootimage.py
```diff
@@ -23,6 +23,12 @@
...
+    raise SystemExit("You appear to be missing most of the x86_64 boot image "
+                     + "artifacts. Check the meta.json for this RHCOS version.")
```
bgilbert, line 31: While we're here, we might as well say something like "Check that full artifact generation was enabled for this build", or whatever is the appropriate terminology.

hack/update-rhcos-bootimage.py
miabbott: We can...though that means maintaining an allowlist of artifacts we expect, which has it's own challenges.

I'll take another swing at it with the list of artifacts.

bgilbert: I'm not proposing listing all the affected artifacts, just checking for a couple well-known ones like `aws`.  The problem with `2` is that if light builds start having 3 artifacts, they'll pass this test.

miabbott:
This change doesn't affect any of the CI jobs, so I think any failures can be safely ignored in the context of this change.

cgwalters:
Note https://github.com/openshift/installer/pull/5252 obsoletes this.
(But, this PR is obviously safer to backport, so we could in theory merge this first and then rebase 5252 on it.  Or we could merge 5252 and just open backport PRs with this change)

miabbott:
With #5252 being merged, this change is no longer necessary. A similar check should be added to the `plume cosa2stream` function

PR #5257, opened 2021-09-30
[wip] UPI Image: update root CAs

patrickdillon:
Let's Encrypt Certificates are expiring. This adds a package to update
CAs through a RHEL repo.
https://access.redhat.com/articles/6338021

Marking this as WIP because I'm testing here. I don't have a setup to use a rhel subscription locally.


patrickdillon:
f5ca51e545924592b01abdd01be9907888a7751d has `yum install ca-certificates` and fails on `e2e-azurestack-upi` with: 

> ERROR: Unable to get endpoints from the cloud.
Please ensure you have network connection. Error detail: HTTPSConnectionPool(host='management.ppe3.stackpoc.com', port=443): Max retries exceeded with url: /metadata/endpoints?api-version=2015-01-01 (Caused by SSLError(SSLError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:877)'),)

patrickdillon:
Testing the [Let's Encrypt R3 CA](https://letsencrypt.org/certs/lets-encrypt-r3.pem) works.

Note that PPE3 is now down, but the test still shows that this cert will work:
```shell
[padillon@thinkpad test]$ curl https://management.ppe3.stackpoc.com/metadata/endpoints?api-version=1.0
curl: (60) SSL certificate problem: certificate has expired
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
[padillon@thinkpad test]$ curl --cacert lets-encrypt-r3.pem https://management.ppe3.stackpoc.com/metadata/endpoints?api-version=1.0
<!DOCTYPE html>
<html>
    <head>
        <title>Runtime Error</title>
        <meta name="viewport" content="width=device-width" />
        <style>
```
I have not been able to build this into the image and test successfully yet. 

It is also a possibility that the PPE3 crash is related to the certs and we may see them fix the problem outright.

PR #5258, opened 2021-09-30
Bug 2008119: force cvo to ignore installer-provided resources

staebler:
When the installer provides manifests for resources that are included as create-only resources in the release payload, there is a race between whether cluster-bootstrap will apply the installer-provided manifest or cluster-version-operato will apply the payload-provided manifest. Whichever one is applied first will be the only one applied. This is problematic since the expectation is that the installer-provided manifests should always be applied.

To get around this, the installer now adds to the .spec.overrides field of the ClusterVersion resource all of the resources for which the installer is providing manifests. This tells the cluster-version-operator to not apply any manifests for these resources. After the cluster-bootstrap completes, the bootkube service will remove those resources from the .spec.overrides field so that the cluster-version-operator can take over management of those resources during upgrades.

The cvoignore asset has been added during the ignition configs stage. This asset is responsible for creating two files for the bootstrap node: cvo-overrides.yaml and original_cvo_overrides.patch.

The cvo-overrides.yaml file is a replacement of the one created by the Common Manifests asset. The replacement adds to the .spec.overrides field all of the resources included in the Common Manifests and Openshift Manifests assets.

The original_cvo_overrides.patch file is a json patch that the bootkube service applies to the ClusterVersion resource after cluster-bootstrap completes. The patch restores any value that the user supplied for the .spec.overrides field.


data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -404,6 +404,11 @@ run_cluster_bootstrap() {
         --volume /etc/kubernetes:/etc/kubernetes:z \
         "${CLUSTER_BOOTSTRAP_IMAGE}" \
         start --tear-down-early=false --asset-dir=/assets --required-pods="${REQUIRED_PODS}"
+     # remove overrides for installer manifests and restore any user-supplied overrides
+    oc patch clusterversion.config.openshift.io version \
```
wking, line 408: is it worth putting this in it's own `clear-overrides.done` block, so if the patch fails (e.g. on an API-server hiccup), we don't come back in and re-launch cluster-bootstrap?

pkg/asset/ignition/bootstrap/cvoignore.go
```diff
@@ -0,0 +1,123 @@
...
+	return nil
+}
+
+// Files returns the files generated by the asset.
+func (a *CVOIgnore) Files() []*asset.File {
+	return a.FileList
+}
+
+// Load does nothing as the file should not be loaded from disk.
+func (a *CVOIgnore) Load(f asset.FileFetcher) (bool, error) {
```
wking, line 121: it seems like it might be possible to avoid the `this must come after manifests.Manifests so that it replaces cvo-overrides.yaml` tweak by having this asset directly read that particular manifest from disk.  But perhaps the order-based clobber is more straightforward after all?

staebler: If the asset reads the manifest from disk, then it will write the manifest to disk, too. Unfortunately, that is problematic because it gives the user an opportunity to make changes to the manifest. Also, I am not sure that it wouldn't cause the original manifest asset to become dirty and create a never-ending cycle.

staebler:
158fb6088...0f65703a0
* Moved the resetting of the overrides in the bootkube script to its own stage after the cluster-bootstrap stage.

pkg/asset/ignition/bootstrap/cvoignore.go
```diff
@@ -0,0 +1,123 @@
...
+	cvoOverridesFilename      = "manifests/cvo-overrides.yaml"
+	originalOverridesFilename = "original_cvo_overrides.patch"
+)
+
+// CVOIgnore adds bootstrap files needed to inform CVO to ignore resources for which the installer is providing manifests.
+type CVOIgnore struct {
+	FileList []*asset.File
+}
+
+// Name returns a human friendly name for the operator
```
patrickdillon, line 30: nit: missing period

pkg/asset/ignition/bootstrap/cvoignore.go
```diff
@@ -0,0 +1,123 @@
...
+func (a *CVOIgnore) Name() string {
+	return "CVO Ignore"
+}
+
+// Dependencies returns all of the dependencies directly needed by the CVOIgnore asset
```
patrickdillon, line 35: nit: missing period

pkg/asset/ignition/bootstrap/cvoignore.go
```diff
@@ -0,0 +1,123 @@
...
+func (a *CVOIgnore) Dependencies() []asset.Asset {
+	return []asset.Asset{
+		&manifests.Manifests{},
+		&manifests.Openshift{},
+	}
+}
+
+// Generate generates the respective operator config.yml files
```
patrickdillon, line 43: nit: missing period

patrickdillon (commented):
/approve

This looks good to me. For the sake of discussion, I will say ideally we should be creating the override during the manifests stage. What is the biggest obstacle to creating this manifest during manifest stage? We must read through all of the manifests, so it would seem like we could generate the cvo-override.yaml immediately before writing the manifests. 

I am wondering if the openshift/operators manifest split is the biggest blocker to the implementation? 

I am asking in case we should capture some of this as tech debt. I am happy to LGTM.

PR #5264, opened 2021-10-01
openstack: Move TFVars logic to own package

pierreprinetti:
This refactoring moves code from the shared `tfvars` asset directory to
the OpenStack-specific package.

The behaviour is left completely unchanged.

Implements [OSASINFRA-2695](https://issues.redhat.com/browse/OSASINFRA-2695)

### DIFF
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pierreprinetti:
/hold
We also need workersAsset

mandre (commented):
Yay! Thanks for taking care of it. Possibly other platforms would be interested in adopting the same pattern.

/cc @patrickdillon @sadasu @Gal-Zaidman @hasueki 
/assign @staebler 

/lgtm

staebler (approved):
/approve

PR #5266, opened 2021-10-05
New flag for cluster create command to avoid printing login info

andfasano:
When an installation went fine, the installer prints the login info (user/password and console url). This is not always required, and especially for CI jobs this requires to sanitize the output. This patch adds a new flag for the create cluster command, so that it will be possible to avoid printing such info in the stdout

### DIFF
--- a/cmd/openshift-install/create.go
+++ b/cmd/openshift-install/create.go
@@ -158,6 +158,10 @@ var (
 	}
 
 	targets = []target{installConfigTarget, manifestsTarget, ignitionConfigsTarget, clusterTarget, singleNodeIgnitionConfigTarget}
+
+	createClusterOpts struct {
+		hideLoginInfo bool
+	}
 )
 
 // clusterCreateError defines a custom error type that would help identify where the error occurs
@@ -217,6 +221,7 @@ func newCreateCmd() *cobra.Command {
 			return cmd.Help()
 		},
 	}
+	cmd.PersistentFlags().BoolVar(&createClusterOpts.hideLoginInfo, "hide-login-info", false, "Do not print login password after the installation is complete, but the file name where it's stored")
 
 	for _, t := range targets {
 		t.command.Args = cobra.ExactArgs(0)
@@ -552,7 +557,11 @@ func logComplete(directory, consoleURL string) error {
 	logrus.Info("Install complete!")
 	logrus.Infof("To access the cluster as the system:admin user when using 'oc', run 'export KUBECONFIG=%s'", kubeconfig)
 	logrus.Infof("Access the OpenShift web-console here: %s", consoleURL)
-	logrus.Infof("Login to the console with user: %q, and password: %q", "kubeadmin", pw)
+	if !createClusterOpts.hideLoginInfo {
+		logrus.Infof("Login to the console with user: %q, and password: %q", "kubeadmin", pw)
+	} else {
+		logrus.Infof("Login to the console with user: %q, and password from %s", "kubeadmin", pwFile)
+	}
 	return nil
 }
 


cmd/openshift-install/create.go
```diff
@@ -550,9 +555,11 @@ func logComplete(directory, consoleURL string) error {
 		return err
 	}
 	logrus.Info("Install complete!")
-	logrus.Infof("To access the cluster as the system:admin user when using 'oc', run 'export KUBECONFIG=%s'", kubeconfig)
-	logrus.Infof("Access the OpenShift web-console here: %s", consoleURL)
-	logrus.Infof("Login to the console with user: %q, and password: %q", "kubeadmin", pw)
+	if !createClusterOpts.suppressLoginInfo {
+		logrus.Infof("To access the cluster as the system:admin user when using 'oc', run 'export KUBECONFIG=%s'", kubeconfig)
+		logrus.Infof("Access the OpenShift web-console here: %s", consoleURL)
```
staebler, line 560: Do these lines need to be suppressed? Or is it only the line containing the username and password for the console?

patrickdillon:
I think it would be better to `s/suppress/hide/g`, "hide" is shorter, easier to spell, has fewer negative connotations.

staebler:
> I think it would be better to `s/suppress/hide/g`, "hide" is shorter, easier to spell, has fewer negative connotations.

I was thinking along the same lines. I was wondering whether it would be easier to explain if it were actually `--show-login-info`, defaulted to true.

derekhiggins:
> This is not always required, and especially for CI jobs this requires to sanitize the output.

I curious about why we care, this password will no longer be valid when CI is done?

cmd/openshift-install/create.go
andfasano: My preference is to hide as well such url

andfasano:
> > > This is not always required, and especially for CI jobs this requires to sanitize the output.
> > 
> > 
> > I curious about why we care, this password will no longer be valid when CI is done?
> 
> There is a period of time between when the cluster is installed and when it is destroyed during which the artifact containing the installer output is publicly available. During this time, someone could log into the cluster and obtain cloud credentials.

Yes, in particular the e2e test time, which usually is quite long. Also, in another context, the same job will be used for the ci-chat-bot.

andfasano:
> > I think it would be better to `s/suppress/hide/g`, "hide" is shorter, easier to spell, has fewer negative connotations.
> 
> I was thinking along the same lines. I was wondering whether it would be easier to explain if it were actually `--show-login-info`, defaulted to true.

I'm fine with both options, even though personally I'd prefer "hide" considering that the default is to show the info

cmd/openshift-install/create.go
staebler: Can you elaborate? My reasoning for not wanting to hide the console URL is that the URL is useful to the end user and not available elsewhere in the installation directory. On that note, I would prefer that the username/password line be replaced with a line directing the user to the location of the file containing the password.

andfasano: Sure, especially in a CI job context that piece of info can be extracted programmatically in some other way (ie through a `oc -n openshift-console get routes console -o=jsonpath='{.spec.host}'`), so no need to print out when no human interaction is required, minimizing thus the info provided.
The idea of the flag was to preserve the current existing default behavior.

cmd/openshift-install/create.go
```diff
@@ -217,6 +221,7 @@ func newCreateCmd() *cobra.Command {
 			return cmd.Help()
 		},
 	}
+	cmd.PersistentFlags().BoolVar(&createClusterOpts.hideLoginInfo, "hide-login-info", false, "Do not print login info after the installation is complete")
```
staebler, line 224: The description of this flag could use some cleanup given the changes to what is actually printed.

andfasano: @staebler could you please check if it's fine now? Thanks

ardaguclu:
For the cleaning up PRs I have subscribed, assigned, etc.

/uncc

PR #5267, opened 2021-10-05
Make bootstrap image asset depend on rhcos image. 

patrickdillon:
The bootstrap image asset calls the rhcos package to determine the rhcos image for the boostrap node. Instead it should
look for the rhcos image in the asset store.


kwoodson:
@patrickdillon Thanks, I'll try this out.

kwoodson:
@patrickdillon This worked for me. 

/lgtm

PR #5270, opened 2021-10-06
Power VS IPI Support: TF Only

clnperez:
For background see https://github.com/openshift/installer/pull/5224

Instead of one massive PR, splitting the Power VS work into multiple PRs. This one contains the Terraform changes for supporting Power VS. There are currently three resources that aren't created via TF:
- The Power VS DHCP Network (available to test 10/18)
- The IBM VPC Network (available now)
- Direct Link (available now) to connect the Power VS network to the VPC Network

The two that are available aren't added here yet as it would make dev trickier. We plan to add these as soon as the functionality is available. 

Everything is tested and is functional.


pkg/rhcos/powervs_regions.go
```diff
@@ -0,0 +1,86 @@
+package rhcos
+
+// Since there is no API to query these, we have to hard-code them here.
+
```
Prashanth684, line 4: this should come from the regions that RHCOS is pushed to which is parsed from the stream metadata.similar to how AMI regions are done through the generate script. 

But the question here is - are these the regions where the bootimage will be imported to ?

clnperez: ok, so we'll probably have to come up with a kind of hybrid approach then, or move this somewhere outside the rhcos purview, and combine what's there with the stream metadata. the regions at the top level are the Power VS regions -- not the IBM Cloud DC/COS regions the bootimage will be imported to.

clnperez:
this is very outdated. closing, and will create new ones when the installer team has bandwidth to review.

PR #5275, opened 2021-10-06
IBM Cloud: provider config zones

cjschaef:
Fix bugs due to assumptions when using install-config data when
generating the IBM cloud provider configuration. Add checks to use
a default MachinePool as well, if one is provided when collecting
ControlPlane and Compute Zones for cloud provider configuration.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2009791

### DIFF
--- a/pkg/asset/manifests/cloudproviderconfig.go
+++ b/pkg/asset/manifests/cloudproviderconfig.go
@@ -170,21 +170,28 @@ func (cpc *CloudProviderConfig) Generate(dependencies asset.Parents) error {
 		if err != nil {
 			return err
 		}
-		controlPlaneZones := installConfig.Config.ControlPlane.Platform.IBMCloud.Zones
-		if len(controlPlaneZones) < 1 {
-			controlPlaneZones, err = ibmcloudmachines.AvailabilityZones(installConfig.Config.IBMCloud.Region)
+
+		controlPlane := &ibmcloudtypes.MachinePool{}
+		controlPlane.Set(installConfig.Config.Platform.IBMCloud.DefaultMachinePlatform)
+		controlPlane.Set(installConfig.Config.ControlPlane.Platform.IBMCloud)
+		compute := &ibmcloudtypes.MachinePool{}
+		compute.Set(installConfig.Config.Platform.IBMCloud.DefaultMachinePlatform)
+		compute.Set(installConfig.Config.WorkerMachinePool().Platform.IBMCloud)
+
+		if len(controlPlane.Zones) == 0 || len(compute.Zones) == 0 {
+			zones, err := ibmcloudmachines.AvailabilityZones(installConfig.Config.IBMCloud.Region)
 			if err != nil {
-				return errors.Wrap(err, "could not create cloud provider config")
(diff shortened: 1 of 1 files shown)


staebler (commented):
I have a couple issues with this approach.
1) We should use the existing mechanisms for merging machine pools. Namely, use the `Set` [function](https://github.com/openshift/installer/blob/5a50e364ac1d61a4237b4169d0e3ab5cc2bb3f9a/pkg/types/ibmcloud/machinepool.go#L27).
2) We should only fetch the availability zones once for the region, rather than fetching them separately for the control plane nodes and the compute nodes.
3) The message wrapped around the error returned when there is a problem fetching the availability zones is not meaningful. The message should indicate the specification action that failed--namely the fetching of the availability zones--rather than generally that there was a failure creating the cloud provider config.

Here is a suggested approach. Play with it to see how it works for you. Feel free to adjust as needed.
```
controlPlane := &ibmcloudtypes.MachinePool{}
controlPlane.Set(installConfig.Config.IBMCloud.DefaultMachinePlatform)
controlPlane.Set(installConfig.Config.ControlPlane.Platform.IBMCloud)
compute := &ibmcloudtypes.MachinePool{}
compute.Set(installConfig.Config.IBMCloud.DefaultMachinePlatform)
compute.Set(installConfig.Config.WorkerMachinePool().Platform.IBMCloud)
if len(controlPlane.Zones) == 0 || len(compute.Zones) == 0 {
	zones, err := ibmcloudmachines.AvailabilityZones(installConfig.Config.IBMCloud.Region)
	if err != nil {
		return errors.Wrap(err, "could not get the availability zones for the region")
	}
	if len(controlPlane.Zones) == 0 {
		controlPlane.Zones = zones
	}
	if len(compute.Zones) == 0 {
		compute.Zones = zones
	}
}
```

pkg/asset/manifests/cloudproviderconfig.go
```diff
@@ -170,21 +170,28 @@ func (cpc *CloudProviderConfig) Generate(dependencies asset.Parents) error {
...
+		controlPlane.Set(installConfig.Config.ControlPlane.Platform.IBMCloud)
+		compute := &ibmcloudtypes.MachinePool{}
+		compute.Set(installConfig.Config.Platform.IBMCloud.DefaultMachinePlatform)
+		compute.Set(installConfig.Config.WorkerMachinePool().Platform.IBMCloud)
+
+		if len(controlPlane.Zones) == 0 || len(compute.Zones) == 0 {
+			zones, err := ibmcloudmachines.AvailabilityZones(installConfig.Config.IBMCloud.Region)
 			if err != nil {
-				return errors.Wrap(err, "could not create cloud provider config")
+				return errors.Wrap(err, fmt.Sprintf("could not get availability zones for %s", installConfig.Config.IBMCloud.Region))
```
staebler, line 184:
```suggestion
				return errors.Wrapf(err, "could not get availability zones for %s", installConfig.Config.IBMCloud.Region)
```

cjschaef: ohh, okay I can patch that quick

PR #5276, opened 2021-10-06
Fix tf format failure on ibmcloud/network/image

cjschaef:
Fix a terraform formatting failure from linting in the
ibmcloud/network/image/main.tf file.


staebler:
https://github.com/openshift/installer/pull/5273

/close

PR #5281, opened 2021-10-08
Bug 2012326: gather-bootstrap: execute dbginfo script to collect debugging info on IBM Z systems

Prashanth684:
This is to address https://bugzilla.redhat.com/show_bug.cgi?id=2012326. Today as part of the docs for IBM Z installation [1], we ask users to run a script to collect debug information[2] by doing an oc debug node. There should be no need for this as this should be part of the must-gather and the installer-gather command in case of bootstrapping failures.

[1] https://access.redhat.com/documentation/en-us/openshift_container_platform/4.8/html/installing/installing-with-rhel-kvm-on-ibm-z-and-linuxone
[2] https://github.com/ibm-s390-linux/s390-tools/blob/master/scripts/dbginfo.sh


data/data/bootstrap/files/usr/local/bin/installer-gather.sh
```diff
@@ -76,6 +79,17 @@ find "${ARTIFACTS}/rendered-assets" -name "*kubeconfig*" -print0 | xargs -0 rm
...
 
+# Collect system information specific to IBM Linux Z (s390x) systems. The dbginfo
+# script is available by default as part of the s390-utils rpm package
+if [ "$arch" == "s390x" ]
+then
+    echo "Gathering dbginfo for the 390x system"
+    mkdir -p "${ARTIFACTS}/node-dbginfo"
+    sudo /usr/sbin/dbginfo.sh -d "${ARTIFACTS}/node-dbginfo"
+    sudo chown -R "${USER}":"${USER}" "${ARTIFACTS}/node-dbginfo"
+    find "${ARTIFACTS}/node-dbginfo" -print0 | xargs -0 chmod a+r
```
staebler, line 90: Can you explain a bit why this is needed? I am confused about why, if we are changing the owner of the collecting info to the user that is collecting the info, then why do we need to allow read permissions to all users? Which other user do we expect to need to read the collected info?

data/data/bootstrap/files/usr/local/bin/installer-masters-gather.sh
```diff
@@ -54,5 +57,15 @@ mkdir -p "${ARTIFACTS}/rpm-ostree"
 sudo rpm-ostree status >& "${ARTIFACTS}/rpm-ostree/status"
 sudo rpm-ostree ex history >& "${ARTIFACTS}/rpm-ostree/history"
 
+# Collect system information specific to IBM Linux Z (s390x) systems. The dbginfo
+# script is available by default as part of the s390-utils rpm package
+if [ "$arch" == "s390x" ]
+then
+    echo "Gathering dbginfo for the s390x system"
+    mkdir -p "${ARTIFACTS}/node-dbginfo"
+    /usr/sbin/dbginfo.sh -d "${ARTIFACTS}/node-dbginfo"
```
staebler, line 66: What is the difference between gathering the dgbinfo on the bootstrap machine and the master machine that the former needs to use `sudo` while the latter does not?

Prashanth684:
@staebler is this good for an approval ?

PR #5284, opened 2021-10-11
Bug 2009253: Add validation to check APIVIP is IPv4 in dual-stack for Bare Metal

ardaguclu:
This PR adds validation to check VIP for API belongs to IPv4 family
for bare metal dual-stack. Because in dual-stack, primary network
must be in IPv4. Hence, APIVIP must also be in IPv4.

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -232,6 +232,15 @@ func validateNetworkingIPVersion(n *types.Networking, p *types.Platform) field.E
 			if apiVIPIPFamily != presence["machineNetwork"].Primary {
 				allErrs = append(allErrs, field.Invalid(field.NewPath("networking", "baremetal", "apiVIP"), p.BareMetal.APIVIP, "VIP for the API must be of the same IP family with machine network's primary IP Family for dual-stack IPv4/IPv6"))
 			}
+
+			ingressVIPIPFamily := corev1.IPv6Protocol
+			if net.ParseIP(p.BareMetal.IngressVIP).To4() != nil {
+				ingressVIPIPFamily = corev1.IPv4Protocol
+			}
+
+			if ingressVIPIPFamily != presence["machineNetwork"].Primary {
+				allErrs = append(allErrs, field.Invalid(field.NewPath("networking", "baremetal", "ingressVIP"), p.BareMetal.IngressVIP, "VIP for the Ingress must be of the same IP family with machine network's primary IP Family for dual-stack IPv4/IPv6"))
+			}
 		case p.None != nil:
 		default:
 			allErrs = append(allErrs, field.Invalid(field.NewPath("networking"), "DualStack", "dual-stack IPv4/IPv6 is not supported for this platform, specify only one type of address"))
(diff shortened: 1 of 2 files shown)


pkg/types/validation/installconfig.go
```diff
@@ -224,6 +224,12 @@ func validateNetworkingIPVersion(n *types.Networking, p *types.Platform) field.E
 		case p.Azure != nil && experimentalDualStackEnabled:
 			logrus.Warnf("Using experimental Azure dual-stack support")
 		case p.BareMetal != nil:
+			// VIP for API should belong to IPv4 family in Bare Metal dual-stack
+			// Because primary network should also be IPv4.
+			apiVIP := net.ParseIP(p.BareMetal.APIVIP)
+			if apiVIP.To4() == nil {
+				allErrs = append(allErrs, field.Invalid(field.NewPath("networking", "baremetal", "apiVIP"), p.BareMetal.APIVIP, "VIP for the API must be in IPv4 network for dual-stack IPv4/IPv6"))
```
danwinship, line 231: The fact that "primary network should be IPv4" is a *bug*. We are supposed to support either IPv4-primary or IPv6-primary, and eventually we will, and we should minimize the number of places that we need to change things in the installer when that happens.

The constraint on the API VIP should be "the API VIP must be of the cluster's primary IP family, whatever family that is",

pkg/types/validation/installconfig.go
```diff
@@ -224,6 +224,14 @@ func validateNetworkingIPVersion(n *types.Networking, p *types.Platform) field.E
 		case p.Azure != nil && experimentalDualStackEnabled:
 			logrus.Warnf("Using experimental Azure dual-stack support")
 		case p.BareMetal != nil:
+			apiVIPIPFamily := corev1.IPv6Protocol
+			if net.ParseIP(p.BareMetal.APIVIP).To4() != nil {
+				apiVIPIPFamily = corev1.IPv4Protocol
+			}
+
+			if apiVIPIPFamily != presence["machineNetwork"].Primary {
```
danwinship, line 232: This could be simplified to

    if utilnet.IsIPv6String(p.BareMetal.APIVIP) != utilnet.IsIPv6CIDR(n.MachineNetwork[0].CIDR) {

(I think that's right... not tested)

but it looks like nothing else in the installer is using the utilnet (`"k8s.io/utils/net"`) IP family functions currently, so maybe that should be part of a larger future cleanup

/lgtm

danwinship:
> Overall this looks good, but we should also validate the Ingress VIP in a similar way I think?

I wasn't sure. With APIVIP, it's definitely the case that having them mismatched will result in a failed install (https://bugzilla.redhat.com/show_bug.cgi?id=2009253#c19) but that's in part because the apiserver is still always single-stack (https://github.com/kubernetes/enhancements/issues/2438). The router has dual-stack support, so it's possible that it will happily work with an IP of either family.

yboaron:
I'm not sure that the current Keepalived code will work for such a mix.
I think that for V4 API-VIP and V6 INGRESS-VIP we'll have a wrong netmask config for keepalived ingress , check  [1] and [2] 

[1] https://github.com/openshift/machine-config-operator/blob/master/templates/master/00-master/on-prem/files/keepalived-keepalived.yaml#L109
[2] https://github.com/openshift/baremetal-runtimecfg/blob/5dab105faaa43f46871405d809c05736a5062837/pkg/config/node.go#L326-#L330

PR #5289, opened 2021-10-11
Bug 2012233: IBMCloud: Handle 5 rule SG limit

cjschaef:
IBM Cloud has a 5 (remote) rule limit on each SecurityGroup. A
recent change placed the rules over that limit on the ControlPlane
SG. Split rules on that SG into 2 SG for the ControlPlane traffic.


pamoedom (commented):
/lgtm

PR #5018, opened 2021-06-21
[WIP] Add Alibaba Cloud platform

bd233:

PR #5291, opened 2021-10-12
[wip] Alibaba recommitted

patrickdillon:
The Alibaba PR #5018 up until this point has been divided among dozens of commits; the PR has recently been squashed down into two large commits, one of all code/configuration, the other for all vendoring.

This PR, takes the most recent state of the PR with the two commits e7297fa443e64e842c7e7fa3166bd7f380ab4339 and 8962496f84393e5c6668330d5a054c622a599977, attempts to help reorganize them in a logical manner for easier review. This PR simply organizes the commits around the code structure of the Installer. There are separate commits for:
- types
- assets
- terraform code and configuration
- destroy code
- vendoring

I propose that @bd233 and his team take the commits from this PR and either update #5018 with the new organization or open a new PR to replace #5018. Again this PR simply reorganizes the current state of #5018 with the goal of making it easier to review. 

Moving forward changes to the PR would either be rebased into the appropriate commit or added using FIXUP commits. Let's make an agreement here before proceeding.

@staebler and @kwoodson thoughts on this plan?


pkg/asset/ignition/node.go
```diff
@@ -91,3 +93,59 @@ func ConvertToAppendix(file *igntypes.File) {
...
+
+		userCA = bytes.TrimSpace(userCA)
+	}
+
+	return carefs, nil
+}
+
+// GenerateIgnitionShim is used to generate an ignition file that contains a user ca bundle
+// in its Security section.
+func GenerateIgnitionShim(bootstrapConfigURL string, userCA string) ([]byte, error) {
```
patrickdillon, line 121: this moves the AWS ignition shim into a reusable function

pkg/asset/installconfig/alibabacloud/metadata.go
```diff
@@ -0,0 +1,27 @@
+package alibabacloud
+
+// Metadata holds additional metadata for InstallConfig resources that
+// does not need to be user-supplied (e.g. because it can be retrieved
+// from external APIs).
+type Metadata struct {
+	client     *Client
```
patrickdillon, line 7: This stored `*Client` is not being utilized, instead `NewClient` is called throughout the code.

pkg/types/alibabacloud/metadata.go
```diff
@@ -0,0 +1,13 @@
...
+
+// Metadata contains Alibaba Cloud metadata (e.g. for uninstalling the cluster).
+type Metadata struct {
+	Region string `json:"region"`
+	// The system checks whether the resource group contains resources after deleted,
+	// This process takes 7 days. The resource group remains in the 'Deleting' state
+	// and couldn't create resource group with the same name during this period.
+	// Before deploying the cluster, the user must manually create a resource group.
+	// The parameter ResourceGroupID is required.
+	ResourceGroupID string `json:"resourceGroupID"`
```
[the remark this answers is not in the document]
bd233: There is a 7-day buffer time after alicloud resource groups are deleted. During this period, resource groups with the same name cannot be created. Therefore, in the initial design, users need to create them manually. If this change is not necessary, we plan to support the function of creating new resource groups in later versions.

pkg/asset/installconfig/alibabacloud/client.go
```diff
@@ -0,0 +1,304 @@
...
+	return map[string]string{
+		"pvtz":            "pvtz.aliyuncs.com",
+		"resourcemanager": "resourcemanager.aliyuncs.com",
+		"ecs":             "ecs.aliyuncs.com",
+	}
+}
+
+func storeCredentials(accessKeyID string, accessKeySecret string) (err error) {
+	os.Setenv(envAccessKeyID, accessKeyID)
+	os.Setenv(envAccessKeySecret, accessKeySecret)
```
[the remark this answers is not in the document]
bd233: OK, I'll delete it

pkg/asset/machines/alibabacloud/machines.go
```diff
@@ -0,0 +1,124 @@
...
+		},
+		ImageID:            mpool.ImageID,
+		InstanceType:       mpool.InstanceType,
+		SystemDiskCategory: string(mpool.SystemDiskCategory),
+		SystemDiskSize:     mpool.SystemDiskSize,
+		RegionID:           platform.Region,
+		ResourceGroupID:    resourceGroup,
+		ZoneID:             az,
+		UserDataSecret:     &corev1.LocalObjectReference{Name: userDataSecret},
+		CredentialsSecret:  &corev1.LocalObjectReference{Name: "alibabacloud-credentials"},
```
patrickdillon, line 103: How does this work with manual credential mode?

pkg/asset/manifests/openshift.go
```diff
@@ -107,7 +109,17 @@ func (o *Openshift) Generate(dependencies asset.Parents) error {
...
+		client, err := alibabacloud.NewClient(installConfig.Config.AlibabaCloud.Region)
+		if err != nil {
+			return err
+		}
+		cloudCreds = cloudCredsSecretData{
+			AlibabaCloud: &AlibabaCloudCredsSecretData{
+				Base64encodeAccessKeyID:     base64.StdEncoding.EncodeToString([]byte(client.AccessKeyID)),
+				Base64encodeSecretAccessKey: base64.StdEncoding.EncodeToString([]byte(client.AccessKeySecret)),
+			},
+		}
```
patrickdillon, line 122: See above comment. I don't think we should be doing this for manual mode.

pkg/asset/manifests/template.go
```diff
@@ -2,6 +2,12 @@ package manifests
 
 import "github.com/openshift/installer/pkg/types/baremetal"
 
+// AlibabaCloudCredsSecretData holds encoded credentials and is used to generate cloud-creds secret
+type AlibabaCloudCredsSecretData struct {
+	Base64encodeAccessKeyID     string
+	Base64encodeSecretAccessKey string
+}
```
patrickdillon, line 9: see above comments, I don't think we should be doing this for manual credential mode

pkg/asset/ignition/node.go
patrickdillon: @bd233 yes this is good. Sorry for the confusing comment. I meant that as a note to myself.

staebler:
> Since the installation occurs before these variables are set I'm not sure how to resolve these until after the cluster installation has started. I believe this can be done but wanted to report this as an extra step that is required before installation can complete successfully. If we need to merge this PR and fix this afterwards that should be okay as then the Alibaba team can reproduce. I wanted to bring this up and begin to think about how we populate these fields during the installation?

I don't think we'll be able to use a VPC ID in the machinesets. As you point out, the actual VPC ID is not known until after the terraform runs. Other platforms use a well-known VPC name instead.

PR #5295, opened 2021-10-13
feat(ibmcloud): Support VPC dedicated hosts

hasueki:
New feature: OpenShift IPI on IBM Cloud VPC dedicated hosts (single-tenant hosts)
IBM Cloud docs: https://cloud.ibm.com/docs/vpc?topic=vpc-creating-dedicated-hosts-instances&interface=ui

This change includes the ability to:
- Dynamically create dedicated hosts (control plane and/or compute) based on install-config params
- Import and use existing dedicated hosts based on install-config params
- Provision  instances on dedicated hosts, if specified
- Static validation checks for dedicated host related install-config params (IBM Cloud Machine Pool)
- API validation checks for dedicated host related install-config params
- Machine/MachineSet manifest generation with dedicated host field set
- Destroy installer created dedicated hosts and dedicated host groups

Enabling the use of this dedicated hosts feature will be controlled through the IBM Cloud MachinePool fields in the install-config. Supported for both control plane and compute machine pools.

Examples of valid `install-config.yaml` snippet with dedicated hosts:
```
platform:
  ibmcloud:
    zones:
    - us-east-1
    dedicatedHosts:
    - profile: cx2-host-152x304  # New dedicated host will be created
      zone: us-east-1
    type: cx2-16x32
```
```
platform:
  ibmcloud:
    zones:
    - us-east-1
    dedicatedHosts:
    - id: my-dedicated-host-id  # Use existing dedicated host
      zone: us-east-1
    type: cx2-16x32
```
```
platform:
  ibmcloud:
    zones:
    - us-east-1
    - us-east-2
    - us-east-3
    dedicatedHosts:
    - profile: cx2-host-152x304  # New dedicated host will be created
      zone: us-east-1
    - id: my-dedicated-host-id   # Use existing dedicated host
      zone: us-east-2
    - profile: cx2-host-152x304  # New dedicated host will be created
      zone: us-east-3
    type: cx2-16x32
 ```
 ---
 Updated
 ```
platform:
  ibmcloud:
    zones:
    - us-east-1
    dedicatedHosts:
    - profile: cx2-host-152x304  # New dedicated host will be created
    type: cx2-16x32
```
```
platform:
  ibmcloud:
    zones:
    - us-east-1
    dedicatedHosts:
    - name: my-dedicated-host-name  # Use existing dedicated host
    type: cx2-16x32
```
```
platform:
  ibmcloud:
    zones:
    - us-east-1
    - us-east-2
    - us-east-3
    dedicatedHosts:
    - profile: cx2-host-152x304     # New dedicated host will be created
    - name: my-dedicated-host-name  # Use existing dedicated host
    - profile: cx2-host-152x304     # New dedicated host will be created
    type: cx2-16x32
 ```

### DIFF
--- a/data/data/ibmcloud/bootstrap/main.tf
+++ b/data/data/ibmcloud/bootstrap/main.tf
@@ -21,6 +21,8 @@ resource "ibm_is_instance" "bootstrap_node" {
     security_groups = concat(var.control_plane_security_group_id_list, [ibm_is_security_group.bootstrap.id])
   }
 
+  dedicated_host = length(var.control_plane_dedicated_host_id_list) > 0 ? var.control_plane_dedicated_host_id_list[0] : null
+
   vpc  = var.vpc_id
   zone = var.control_plane_subnet_zone_list[0]
   keys = []
--- a/go.mod
+++ b/go.mod
@@ -67,7 +67,7 @@ require (
 	github.com/openshift/client-go v0.0.0-20210409155308-a8e62c60e930
 	github.com/openshift/cloud-credential-operator v0.0.0-20200316201045-d10080b52c9e
 	github.com/openshift/cluster-api-provider-gcp v0.0.1-0.20201203141909-4dc702fd57a5
-	github.com/openshift/cluster-api-provider-ibmcloud v0.0.0-20210702173623-676faba9895d
+	github.com/openshift/cluster-api-provider-ibmcloud v0.0.0-20211008100740-4d7907adbd6b
 	github.com/openshift/cluster-api-provider-libvirt v0.2.1-0.20191219173431-2336783d4603
 	github.com/openshift/cluster-api-provider-ovirt v0.1.1-0.20210817084941-2262c7c6cece
 	github.com/openshift/library-go v0.0.0-20210408164723-7a65fdb398e2
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -447,6 +447,56 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			workerConfigs[i] = w.Spec.Template.Spec.ProviderSpec.Value.Object.(*ibmcloudprovider.IBMCloudMachineProviderSpec)
 		}
 
+		// Set machine pool info
+		var masterMachinePool ibmcloud.MachinePool
+		var workerMachinePool ibmcloud.MachinePool
+		if installConfig.Config.Platform.IBMCloud.DefaultMachinePlatform != nil {
+			masterMachinePool.Set(installConfig.Config.Platform.IBMCloud.DefaultMachinePlatform)
+			workerMachinePool.Set(installConfig.Config.Platform.IBMCloud.DefaultMachinePlatform)
+		}
+		if installConfig.Config.ControlPlane.Platform.IBMCloud != nil {
+			masterMachinePool.Set(installConfig.Config.ControlPlane.Platform.IBMCloud)
+		}
+		if worker := installConfig.Config.WorkerMachinePool(); worker != nil {
+			workerMachinePool.Set(worker.Platform.IBMCloud)
+		}
+
+		// Get master dedicated host info
+		var masterDedicatedHosts []ibmcloudtfvars.DedicatedHost
+		for _, dhost := range masterMachinePool.DedicatedHosts {
+			if dhost.Name != "" {
+				dh, err := client.GetDedicatedHostByName(ctx, dhost.Name, installConfig.Config.Platform.IBMCloud.Region)
+				if err != nil {
+					return err
+				}
+				masterDedicatedHosts = append(masterDedicatedHosts, ibmcloudtfvars.DedicatedHost{
+					ID: *dh.ID,
+				})
+			} else {
+				masterDedicatedHosts = append(masterDedicatedHosts, ibmcloudtfvars.DedicatedHost{
+					Profile: dhost.Profile,
+				})
+			}
+		}
+
+		// Get worker dedicated host info
+		var workerDedicatedHosts []ibmcloudtfvars.DedicatedHost
+		for _, dhost := range workerMachinePool.DedicatedHosts {
+			if dhost.Name != "" {
+				dh, err := client.GetDedicatedHostByName(ctx, dhost.Name, installConfig.Config.Platform.IBMCloud.Region)
+				if err != nil {
+					return err
+				}
+				workerDedicatedHosts = append(workerDedicatedHosts, ibmcloudtfvars.DedicatedHost{
+					ID: *dh.ID,
+				})
+			} else {
+				workerDedicatedHosts = append(workerDedicatedHosts, ibmcloudtfvars.DedicatedHost{
+					Profile: dhost.Profile,
+				})
+			}
+		}
+
 		// Get CISInstanceCRN from InstallConfig metadata
 		crn, err := installConfig.IBMCloud.CISInstanceCRN(ctx)
 		if err != nil {
@@ -455,13 +505,15 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 
 		data, err = ibmcloudtfvars.TFVars(
 			ibmcloudtfvars.TFVarsSources{
-				Auth:              auth,
-				CISInstanceCRN:    crn,
-				ImageURL:          string(*rhcosImage),
-				MasterConfigs:     masterConfigs,
-				PublishStrategy:   installConfig.Config.Publish,
-				ResourceGroupName: installConfig.Config.Platform.IBMCloud.ResourceGroupName,
-				WorkerConfigs:     workerConfigs,
+				Auth:                 auth,
+				CISInstanceCRN:       crn,
+				ImageURL:             string(*rhcosImage),
+				MasterConfigs:        masterConfigs,
+				MasterDedicatedHosts: masterDedicatedHosts,
+				PublishStrategy:      installConfig.Config.Publish,
+				ResourceGroupName:    installConfig.Config.Platform.IBMCloud.ResourceGroupName,
+				WorkerConfigs:        workerConfigs,
+				WorkerDedicatedHosts: workerDedicatedHosts,
 			},
 		)
 		if err != nil {
--- a/pkg/asset/installconfig/ibmcloud/validation.go
+++ b/pkg/asset/installconfig/ibmcloud/validation.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/IBM/vpc-go-sdk/vpcv1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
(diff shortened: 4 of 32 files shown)

### CODE REVIEW

File: pkg/destroy/ibmcloud/dedicatedhost.go [hunk 1]
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	options := o.vpcSvc.NewListDedicatedHostsOptions()
+	options.SetResourceGroupID(resourceGroupID)
+	resources, _, err := o.vpcSvc.ListDedicatedHostsWithContext(ctx, options)
+	if err != nil {
+		return nil, err
+	}
+
+	result := []cloudResource{}
```
staebler, line 36: Why collect in an intermediate slice rather than insert each `cloudResource` directly into the `cloudResources` object?

pkg/types/ibmcloud/validation/machinepool.go
```diff
@@ -25,3 +42,40 @@ func validateBootVolume(bv *ibmcloud.BootVolume, path *field.Path) field.ErrorLi
 	}
 	return allErrs
 }
+
+func validateDedicatedHosts(dhosts []ibmcloud.DedicatedHost, itype string, zones []string, path *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	// Length of dedicated hosts must match platform zones
+	if len(dhosts) != len(zones) {
```
staebler, line 50: Why do we require the user to specify the zones in two different locations in the exact same order? Can't we use the zones defined at the machine-pool level and omit the zones from the dedicated hosts?

data/data/ibmcloud/network/vpc/dedicated-hosts.tf
```diff
@@ -0,0 +1,70 @@
...
+  dhosts_master_merged = [
+    for dhost in var.dedicated_hosts_master :
+    lookup(dhost, "id", "") == ""
+    ? ibm_is_dedicated_host.control_plane[index(ibm_is_dedicated_host.control_plane.*.zone, dhost.zone)].id
+    : dhost.id
+  ]
+}
+
+############################################
+# Dedicated hosts (Control Plane)
```
staebler, line 13: Is the vpc module really the proper place for creating the dedicated hosts? At first blush, dedicated hosts are not part of a VPC.

pkg/asset/machines/ibmcloud/machines.go
```diff
@@ -120,6 +132,44 @@ func provider(clusterID string,
...
+		defaultName = fmt.Sprintf("%s-dhost-control-plane-%s", clusterID, zone)
+	case "worker":
+		defaultName = fmt.Sprintf("%s-dhost-compute-%s", clusterID, zone)
+	default:
+		return "", fmt.Errorf("invalid machine role %v", role)
+	}
+
+	for _, dhost := range dhosts {
+		if dhost.Zone == zone {
+			return defaultName, nil
```
staebler, line 148: Should there be a check here that the profile is not empty?

pkg/asset/machines/ibmcloud/machines.go
```diff
@@ -120,6 +132,44 @@ func provider(clusterID string,
...
+
+	for _, dhost := range dhosts {
+		if dhost.ID != "" {
+			dh, err := client.GetDedicatedHostByID(context.TODO(), dhost.ID, strings.Join(strings.Split(zone, "-")[:2], "-"))
+			if err != nil {
+				return "", err
+			}
+
+			if *dh.Zone.Name == zone {
+				return *dh.Name, nil
```
staebler, line 165: If the MAO is expecting the name of the dedicated host rather than the ID, why are we asking the user to provide the ID rather than the name?

data/data/ibmcloud/bootstrap/main.tf
```diff
@@ -21,6 +21,8 @@ resource "ibm_is_instance" "bootstrap_node" {
     security_groups = concat(var.control_plane_security_group_id_list, [ibm_is_security_group.bootstrap.id])
   }
 
+  dedicated_host = length(var.control_plane_dedicated_host_id_list) > 0 ? var.control_plane_dedicated_host_id_list[0] : null
```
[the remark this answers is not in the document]
hasueki: Is this something that we can just simply doc? It's kind of a similar situation of how bootstrap node gets placed in the first control plane subnet, or how bootstrap node will need to be accounted for as far as resource quotas.

pkg/asset/cluster/tfvars.go
```diff
@@ -447,6 +447,16 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			workerConfigs[i] = w.Spec.Template.Spec.ProviderSpec.Value.Object.(*ibmcloudprovider.IBMCloudMachineProviderSpec)
 		}
 
+		// Get dedicated host info
+		var masterDedicatedHosts []ibmcloud.DedicatedHost
+		if installConfig.Config.ControlPlane.Platform.IBMCloud != nil {
+			masterDedicatedHosts = installConfig.Config.ControlPlane.Platform.IBMCloud.DedicatedHosts
```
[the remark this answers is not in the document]
hasueki: We want to get it from the install-config. Ok will add validation there.

go.mod
```diff
@@ -67,7 +67,7 @@ require (
 	github.com/openshift/client-go v0.0.0-20210409155308-a8e62c60e930
 	github.com/openshift/cloud-credential-operator v0.0.0-20200316201045-d10080b52c9e
 	github.com/openshift/cluster-api-provider-gcp v0.0.1-0.20201203141909-4dc702fd57a5
-	github.com/openshift/cluster-api-provider-ibmcloud v0.0.0-20210702173623-676faba9895d
+	github.com/openshift/cluster-api-provider-ibmcloud v0.0.0-20211008100740-4d7907adbd6b
```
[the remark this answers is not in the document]
hasueki: Will do.

pkg/asset/installconfig/ibmcloud/validation.go
```diff
@@ -55,24 +58,100 @@ func validateMachinePool(client API, platform *ibmcloud.Platform, machinePool *i
...
+			}
+
+			if dh != nil {
+				// Check if instance is provisionable on host
+				if !*dh.InstancePlacementEnabled || !*dh.Provisionable {
+					allErrs = append(allErrs, field.Invalid(path.Index(i).Child("id"), dhost.ID, "dedicated host is unable to provision instances"))
+				}
+
+				// Check if host is in region
+				if !strings.HasPrefix(*dh.Zone.Name, region) {
```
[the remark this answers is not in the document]
hasueki: Ah good catch. Yes that was the intention - will adjust.

pkg/asset/cluster/tfvars.go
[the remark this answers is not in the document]
staebler: Are you for sure sure? Typically, at this stage, the machines and machinesets are the source of truth for machine-specific details rather than the install config.

pkg/destroy/ibmcloud/dedicatedhost.go
hasueki: I left it as that to keep it consistent with other resources' destroy/list logic

data/data/ibmcloud/bootstrap/main.tf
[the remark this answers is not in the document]
staebler: Sure. I don't know enough about dedicated hosts to make a judgement call on the impact to the user for a typical dedicated-host use case. If you think that it will rarely be an issue, then let's leave it to docs.

pkg/destroy/ibmcloud/dedicatedhost.go
staebler: Sounds good. I did not check to see whether this was an existing pattern.

pkg/types/ibmcloud/validation/machinepool.go
hasueki: Yes, we can. I didn't know if it was an anti-pattern to make that sort of linkage between 2 params.

data/data/ibmcloud/network/vpc/dedicated-hosts.tf
hasueki: You're right there. I just put it there for the sake of ease, but I can create a new module.

pkg/asset/machines/ibmcloud/machines.go
hasueki: Just to make it a little more clear that the dedicated host already exists (not that installer will create a new dedicated host with that name). But I can update the MachinePool to use name, if that's better.

hasueki:
Thank you for the review @staebler! I have responded to your feedback and will make those updates as discussed.

pkg/asset/machines/ibmcloud/machines.go
staebler: I'm interested in consistency between the installer config and the MAO config. The installer can (and should) check to verify that the dedicated host with the name exists, so I don't see an issue with the user incorrectly running an installation expecting the dedicated host to be created with the specified name. I also don't think that the user would even have that expectation since it would be pretty unique behavior for the installer to take a name for a resource that the installer is going to create.

pkg/asset/cluster/tfvars.go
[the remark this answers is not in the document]
hasueki: The caveat here is that once the machine/machinesets are generated, all it has in the API spec is the "name" of the dedicated host. This makes it difficult to determine if the dedicated host needed to be created or not.

pkg/asset/machines/ibmcloud/machines.go
hasueki: Ok understood - will adjust

pkg/asset/cluster/tfvars.go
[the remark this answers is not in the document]
staebler: Ah, fair point. What do you expect the behavior to be if the user manually modifies the machines/machinesets manifests?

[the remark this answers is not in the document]
hasueki: Hm I haven't accounted for that use case.. In that case, is it safe to assume if the dedicated host name in the machine/machineset is the default name including the cluster_id, it needs to be created? Not sure if there would be a case where user would have access to the cluster_id (with the hash suffix) to manually create a dedicated host with that name.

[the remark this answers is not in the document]
staebler: I'm probably overthinking this. I think we are fine with not supporting the user making manual modifications to the dedicated hosts in the machine and machinesets. The supported way to set the dedicated hosts is via the install config. So, the only additional part needed here is to account for the use case where the dedicated hosts are set in the default machine pool and not in either the control plane or compute pools.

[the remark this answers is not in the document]
hasueki: Got it - thanks for the feedback. Will modify to account for the default machine pool.

pkg/asset/cluster/tfvars.go
```diff
@@ -447,43 +447,53 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
+		if installConfig.Config.Platform.IBMCloud.DefaultMachinePlatform != nil {
+			masterMachinePool.Set(installConfig.Config.Platform.IBMCloud.DefaultMachinePlatform)
+			workerMachinePool.Set(installConfig.Config.Platform.IBMCloud.DefaultMachinePlatform)
+		}
+		if installConfig.Config.ControlPlane.Platform.IBMCloud != nil {
+			masterMachinePool.Set(installConfig.Config.ControlPlane.Platform.IBMCloud)
+		}
+		if installConfig.Config.Compute[0].Platform.IBMCloud != nil {
+			workerMachinePool.Set(installConfig.Config.Compute[0].Platform.IBMCloud)
+		}
```
staebler, line 462:
```suggestion
		if worker := installConfig.Config.WorkerMachinePool; worker != nil {
			workerMachinePool.Set(worker.Platform.IBMCloud)
		}
```

pkg/asset/installconfig/ibmcloud/validation.go
```diff
@@ -2,14 +2,14 @@ package ibmcloud
...
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
-	"reflect"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
+	"github.com/IBM/vpc-go-sdk/vpcv1"
```
staebler, line 11: Combine this with the k8s.io imports above.

pkg/asset/cluster/tfvars.go
```diff
@@ -447,6 +447,56 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
+
+		// Get master dedicated host info
+		var masterDedicatedHosts []ibmcloudtfvars.DedicatedHost
+		for _, dhost := range masterMachinePool.DedicatedHosts {
+			if dhost.Name != "" {
+				dh, err := client.GetDedicatedHostByName(ctx, dhost.Name, installConfig.Config.Platform.IBMCloud.Region)
```
staebler, line 468: Would it be better to leave it up to the terraform to do the lookup from name to ID via an `ibm_is_dedicated_host` data source?

staebler (commented):
Looks pretty solid to me. Feel free to squash down the commits (leaving the vendor commit separate).

pkg/asset/cluster/tfvars.go
hasueki: Yes it would be and that was the route I was hoping for.. But the IBM TF provider at it's current state requires the dedicated host group to be specified to do a lookup (which technically isn't necessary, but requires it for some reason). So installer lib would have to make an API call regardless, so figured might as well grab the ID and pass it to TF.

We can look into getting rid of this dedicated host group requirement in the TF provider data source. That way we can get rid of this lookup from the installer and let TF handle it like mentioned.

hasueki:
@staebler Updated to address your most recent feedback and commits squashed. Left one comment.

pkg/asset/cluster/tfvars.go
staebler: OK. If the terraform provider does not support this as is, then I have no issue with doing the query here.

staebler (approved):
/lgtm
/approve

PR #5296, opened 2021-10-14
Bug 2004313: azure: Skip making azure bootstrap ignition temp file

squidboylan:
Instead of creating a temp file and then uploading that to an azure
storage blob, we can upload the ignition file created by the go code to
azure.

fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2004313


m1kola:
@squidboylan could you please give more details on why we need this PR?

staebler:
External testing of this PR has shown that it is not sufficient to resolve the BZ. I think we still want this change, but there is some other issue as well.

PR #5297, opened 2021-10-15
Bug 1960378:  Relax the imageContentSource validation

kirankt:
Relax the imageContentSource validation to allow for mirroring root registry.

Currently passing a full hostname in the imageContentSource entries will result in an canonical error. This PR relaxes the validation check to allow passing of the registry root.

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -536,6 +536,18 @@ func validateImageContentSources(groups []types.ImageContentSource, fldPath *fie
 func validateNamedRepository(r string) error {
 	ref, err := dockerref.ParseNamed(r)
 	if err != nil {
+		// If a mirror name is provided without the named reference,
+		// then the name is not considered canonical and will cause
+		// an error. e.g. registry.lab.redhat.com:5000 will result
+		// in an error. Instead we will check whether the input is
+		// a valid hostname as a workaround.
+		if err == dockerref.ErrNameNotCanonical {
+			_, err := url.ParseRequestURI(r)
+			if err != nil {
+				return fmt.Errorf("the repository provided is invalid")
+			}
+			return nil
+		}
 		return errors.Wrap(err, "failed to parse")
 	}
 	if !dockerref.IsNameOnly(ref) {
--- a/pkg/types/validation/installconfig_test.go
+++ b/pkg/types/validation/installconfig_test.go
@@ -1038,7 +1038,7 @@ func TestValidateInstallConfig(t *testing.T) {
 			expectedError: `^credentialsMode: Unsupported value: "Mint": supported values: "Manual"$`,
 		},
 		{
-			name: "release image source is not canonical",
+			name: "release image source is not valid",
 			installConfig: func() *types.InstallConfig {
 				c := validInstallConfig()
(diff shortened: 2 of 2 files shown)


pkg/types/validation/installconfig.go
```diff
@@ -543,6 +543,18 @@ func validateImageContentSources(groups []types.ImageContentSource, fldPath *fie
...
 	if err != nil {
+		// If a mirror name is provided without the named reference,
+		// then the name is not considered canonical and will cause
+		// an error. e.g. registry.lab.redhat.com:5000 will result
+		// in an error. Instead we will check whether the input is
+		// a valid hostname as a workaround.
+		if err == dockerref.ErrNameNotCanonical {
+			_, err := url.ParseRequestURI(r)
+			if err != nil {
+				return fmt.Errorf("the repository provided is invalid")
```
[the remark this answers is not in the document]
staebler: We do not need to include the URI in the returned error. The URI will already be included in the field error presented to the user.

See the following as an example of the construction of the field error, where `group.Source` is the URI.
https://github.com/openshift/installer/blob/ae323fb7434ca859988ada53ab064a6cffac2788/pkg/types/validation/installconfig.go#L530

pkg/types/validation/installconfig.go
```diff
@@ -543,6 +543,18 @@ func validateImageContentSources(groups []types.ImageContentSource, fldPath *fie
 func validateNamedRepository(r string) error {
 	ref, err := dockerref.ParseNamed(r)
 	if err != nil {
+		// If a mirror name is provided without the named reference,
+		// then the name is not considered canonical and will cause
+		// an error. e.g. registry.lab.redhat.com:5000 will result
+		// in an error. Instead we will check whether the input is
+		// a valid hostname as a workaround.
+		if err == dockerref.ErrNameNotCanonical {
```
staebler, line 551: Can we call `dockerref.ParseNormalizedNamed` instead of `dockerref.ParseNamed` to avoid the canonical check? Or is there an example of a non-canonical URI that would be accepted that we want to reject?

staebler, line 551: Please add unit test cases covering a non-canonical reference that should be rejected and one that should be accepted.

kirankt: After calling `ParseNormalizedNamed`, `ParseNamed` also checks to see if the original string is the same as what `ParseNormalizedNamed` spits out. If they're the same then the original input was kosher and all is well. If you don't pass anything that is not canonical, then `ParseNormalizedNamed` makes changes and we will encounter the `dockerref.ErrNameNotCanonical` error.

https://github.com/containers/image/blob/b7361977328b2d8cd0a75333e4124589405cd45f/docker/reference/reference.go#L246

e.g.
`host.example.com:5000/library/package` : Input is Canonical. All is well.
`foo/bar` transforms to `docker.io/library/foo/bar` : Name transformed into a canonical string but fails string check
host.example.com:5000 transforms to docker.io/library/host.example.com:5000: This fails the string check in ParseNamed and we get the canonical error. This is a corner case, we want to succeed. So we perform an additional URI check here to see if its a valid hostname and continue if true. Or else we exit.

PR #5307, opened 2021-10-18
validation: Enforce strict unmarshalling of config

rna-afk:
In order to stop the user from adding unrecognized properties
in the install config, a validation check is added to force the
user to remove them. This is done by enforcing a strict
unmarshalling of the install config yaml file that will raise an
error when an unknown property is added.

This will force the installer to return an error for the first
occurence of an unknown property since the current library
returns only the first unknown field.

### DIFF
--- a/pkg/asset/installconfig/installconfig.go
+++ b/pkg/asset/installconfig/installconfig.go
@@ -3,6 +3,7 @@ package installconfig
 import (
 	"context"
 	"os"
+	"strings"
 
 	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
@@ -123,8 +124,12 @@ func (a *InstallConfig) Load(f asset.FileFetcher) (found bool, err error) {
 	}
 
 	config := &types.InstallConfig{}
-	if err := yaml.Unmarshal(file.Data, config); err != nil {
-		return false, errors.Wrapf(err, "failed to unmarshal %s", installConfigFilename)
+	if err := yaml.UnmarshalStrict(file.Data, config, yaml.DisallowUnknownFields); err != nil {
+		if strings.Contains(err.Error(), "unknown field") {
+			err = errors.Wrapf(err, "failed to parse first occurence of unknown field")
+		}
+		err = errors.Wrapf(err, "failed to unmarshal %s", installConfigFilename)
+		return false, err
 	}
 	a.Config = config
 
--- a/pkg/asset/installconfig/installconfig_test.go
+++ b/pkg/asset/installconfig/installconfig_test.go
@@ -178,6 +178,21 @@ metadata:
 			expectedError: true,
 		},
 		{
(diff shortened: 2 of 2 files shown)


pkg/asset/installconfig/installconfig.go
```diff
@@ -123,8 +124,12 @@ func (a *InstallConfig) Load(f asset.FileFetcher) (found bool, err error) {
 	}
 
 	config := &types.InstallConfig{}
-	if err := yaml.Unmarshal(file.Data, config); err != nil {
-		return false, errors.Wrapf(err, "failed to unmarshal %s", installConfigFilename)
+	if err := yaml.UnmarshalStrict(file.Data, config, yaml.DisallowUnknownFields); err != nil {
+		err = errors.Wrapf(err, "failed to unmarshal %s", installConfigFilename)
+		if strings.Contains(err.Error(), "unknown field") {
+			err = errors.Wrapf(err, "failed to parse first occurence of unknown field")
```
staebler, line 130: The order of this wrapping is backwards. The "failed to unmarshal" error should wrap around the "failed to parse" error.

pkg/asset/installconfig/installconfig_test.go
```diff
@@ -189,7 +189,21 @@ platform:
     region: us-east-1
 pullSecret: "{\"auths\":{\"example.com\":{\"auth\":\"authorization value\"}}}"
 network:
-  type: OpenShiftSDN
+  type: OpenshiftSDN
```
staebler, line 192: Why did you change this to an invalid network type?

pkg/asset/installconfig/installconfig_test.go
```diff
@@ -189,7 +189,21 @@ platform:
     region: us-east-1
 pullSecret: "{\"auths\":{\"example.com\":{\"auth\":\"authorization value\"}}}"
 network:
```
staebler, line 191: Use a field name that is **clearly** unknown rather than a field name that is similar to a known field. It is hard reading the test case input to spot immediately which field is wrong.

pkg/asset/installconfig/installconfig_test.go
```diff
@@ -177,6 +177,21 @@ metadata:
 			fetchError:    errors.New("fetch failed"),
 			expectedError: true,
 		},
+		{
+			name: "unknown field",
+			data: `
+apiVersion: v1beta3
```
staebler, line 183:
```suggestion
apiVersion: v1
```

patrickdillon:
I assume you're  already looking into this, but openstack-manifests is failing with what looks like a related error.

rna-afk:
Yep working on fixing the test first

rna-afk:
/hold cancel
The fix for openstack-manifests has merged
https://github.com/openshift/installer/pull/5349

PR #5310, opened 2021-10-18
Add 'auth_tls = none' to libvirt dev docs

machadovilaca:
Signed-off-by: João Vilaça <jvilaca@redhat.com>

`libvirtd.service` was not starting up correctly because of `libvirtd-tcp.socket` which was failing with the error `systemd[1]: Failed to listen on Libvirt non-TLS IP socket`

Making the change suggested in this PR fixed the issue

### DIFF
--- a/docs/dev/libvirt/README.md
+++ b/docs/dev/libvirt/README.md
@@ -160,6 +160,7 @@ set the following:
 listen_tls = 0
 listen_tcp = 1
 auth_tcp = "none"
+auth_tls = "none"
 tcp_port = "16509"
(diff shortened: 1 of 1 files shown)


praveenkumar:
@machadovilaca We already disable `listen_tls = 0` so why we have to set the `tls_auth` ?

cfergeau:
More details would indeed be needed, on my rhel8 system, starting `libvirtd-tcp.socket` works with the default `libvirtd.conf` file (the one from the libvirt-daemon RPM)

PR #5314, opened 2021-10-19
Bug 1999668: Azure: Ensure destroy doesn't panic when invalid credent…

kirankt:
Ensure destroy doesn't panic when invalid credentials are provided.

### DIFF
--- a/pkg/destroy/azure/azure.go
+++ b/pkg/destroy/azure/azure.go
@@ -468,17 +468,11 @@ func deleteResourceGroup(ctx context.Context, client resources.GroupsClient, log
 	defer cancel()
 
 	delFuture, err := client.Delete(ctx, name)
-	if err != nil {
-		if wasNotFound(delFuture.Response()) {
-			logger.Debug("already deleted")
-			return nil
-		}
-		return err
+	if err == nil {
+		err = delFuture.WaitForCompletionRef(ctx, client.Client)
 	}
-
-	err = delFuture.WaitForCompletionRef(ctx, client.Client)
 	if err != nil {
-		if wasNotFound(delFuture.Response()) {
+		if isNotFoundError(err) {
 			logger.Debug("already deleted")
 			return nil
 		}
@@ -490,6 +484,23 @@ func deleteResourceGroup(ctx context.Context, client resources.GroupsClient, log
 
 func wasNotFound(resp *http.Response) bool {
 	return resp != nil && resp.StatusCode == http.StatusNotFound
+}
+
+func isNotFoundError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	var dErr autorest.DetailedError
+	if errors.As(err, &dErr) {
+		switch statusCode := dErr.StatusCode.(type) {
+		case int:
+			if statusCode == http.StatusNotFound {
+				return true
+			}
+		}
+	}
+	return false
 }
 
 func isAuthError(err error) bool {


pkg/destroy/azure/azure.go
```diff
@@ -467,16 +467,16 @@ func deleteResourceGroup(ctx context.Context, client resources.GroupsClient, log
...
 			return nil
 		}
-		return err
+		return errors.Wrapf(err, "failed to delete %s", name)
 	}
 
 	err = delFuture.WaitForCompletionRef(ctx, client.Client)
 	if err != nil {
-		if wasNotFound(delFuture.Response()) {
+		if delFuture.FutureAPI != nil && wasNotFound(delFuture.Response()) {
```
staebler, line 479: The added check for a non-nil `FutureAPI` is not needed here. The contract with the `client.Delete` function is that `delFuture` will have a non-nil `FutureAPI` unless an error was returned. Regardless, if the code got to this point, then `FutureAPI` cannot be nil, otherwise the call to `delFuture.WaitForCompletionRef` would have raised a panic.

pkg/destroy/azure/azure.go
```diff
@@ -469,11 +469,11 @@ func deleteResourceGroup(ctx context.Context, client resources.GroupsClient, log
 
 	delFuture, err := client.Delete(ctx, name)
 	if err != nil {
-		if wasNotFound(delFuture.Response()) {
+		if delFuture.FutureAPI != nil {
```
staebler, line 472: I feel like we should be checking something in `err` here rather than assuming that a nil `delFuture.FutureAPI` means that the resource group does not exist. I expect that, if there was an error, then `delFuture.FutureAPI` will be nil, regardless of whether the error was due to the resource group not existing or not.

Maybe it would be helpful to run the destroyer using a non-existent resource group to see what the value of `err` is in that case.

m1kola: @staebler you are right: `err` should have information on 404 objects. This is the pattern we use to check for not founds in [ARO](https://github.com/azure/aro-rp):

```go
if detailedErr, ok := err.(autorest.DetailedError); ok &&
	detailedErr.StatusCode == http.StatusNotFound {
	err = nil
}
if err != nil {
	return err
}
```

m1kola: Note: the above is an example from a get request. It might be enough for deleting resource groups.

We also have [this](https://github.com/Azure/ARO-RP/blob/37b20109e06813e38cbfb6f0edf433119ba8edba/pkg/util/azureerrors/error.go#L93-L108) function to check for `ResourceGroupNotFound` specificly and [this](https://github.com/Azure/ARO-RP/blob/37b20109e06813e38cbfb6f0edf433119ba8edba/pkg/cluster/delete.go#L332-L349) is how we delete cluster's resource group.

Checking for errors in Azure is hard :)

kirankt: @m1kola and @staebler Thanks for the examples and suggestions. PTAL at the new changes and let me know what you guys think.

pkg/destroy/azure/azure.go
```diff
@@ -467,25 +467,24 @@ func deleteResourceGroup(ctx context.Context, client resources.GroupsClient, log
...
 
-	delFuture, err := client.Delete(ctx, name)
+	err := deleteAndWait(ctx, client, name)
 	if err != nil {
-		if wasNotFound(delFuture.Response()) {
-			logger.Debug("already deleted")
-			return nil
+		if detailedErr, ok := err.(autorest.DetailedError); ok &&
+			(detailedErr.StatusCode == http.StatusForbidden || detailedErr.StatusCode == http.StatusNotFound) {
+			return errors.Wrapf(err, "failed to delete %s", name)
```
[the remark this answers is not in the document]
2uasimojo: If we go with this approach, can I suggest for consistency:
a) adding a helper function similar to [isAuthError](https://github.com/openshift/installer/blob/99b6117014de7f27fce54ff8ac2d9f13df7bbd3c/pkg/destroy/azure/azure.go#L494-L510) and [isResourceGroupBlockedError](https://github.com/openshift/installer/blob/99b6117014de7f27fce54ff8ac2d9f13df7bbd3c/pkg/destroy/azure/azure.go#L512-L528)
b) replacing _all_ uses of `wasNotFound` with the above.

pkg/destroy/azure/azure.go
```diff
@@ -467,25 +467,24 @@ func deleteResourceGroup(ctx context.Context, client resources.GroupsClient, log
...
 		}
-		return err
 	}
+	logger.Info("deleted")
+	return nil
+}
 
-	err = delFuture.WaitForCompletionRef(ctx, client.Client)
+func deleteAndWait(ctx context.Context, client resources.GroupsClient, resourceGroupName string) error {
```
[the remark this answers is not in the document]
2uasimojo: > forego the function entirely

+1.

Another option, which might be more go-ish (if more verbose):

```go
delFuture, err := client.Delete(ctx, name)
if err != nil {
    return err
}
err = delFuture.WaitForCompletionRef(ctx, client.Client)
if err != nil {
    if wasNotFound(delFuture.Response()) {   // or per above comment: if isNotFoundError(err) {
        logger.Debug("already deleted")
        return nil
    }
    return err
}
```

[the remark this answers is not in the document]
m1kola: @2uasimojo  I think with the above option we will still have nil dereference mentioned in #5428. No?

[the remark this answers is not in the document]
2uasimojo: I don't think so.
When calling `WaitForCompletionRef` (which is also part of FutureAPI) the assumption is that `delFuture.FutureAPI` can't be `nil` if `err == nil`, which (IIUC) is a typical pattern in go.
And since that's the case, subsequently calling `Response` is safe.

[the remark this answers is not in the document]
staebler: My objection to the option proposed by @2uasimojo is that it effectively reverts https://github.com/openshift/installer/pull/3416, which was at least a partial fix for a real problem. There are some scenarios where the call to `client.Delete` will return an error when the resource group does not exist.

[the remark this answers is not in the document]
2uasimojo: Oh, so we *do* have to check `err` for 404 immediately after the `Delete` call? Then yeah, what you said. Which, for completeness, is approximately the following if you accept my above comment about the helper function:
```go
delFuture, err := client.Delete(ctx, name)
if err == nil {
    err = delFuture.WaitForCompletionRef(ctx, client.Client)
}
if err != nil {
    if isNotFoundError(err) {
        logger.Debug("already deleted")
        return nil
    }
    return errors.Wrapf(err, "failed to delete %s", name)
}
logger.Info("deleted")
return nil
```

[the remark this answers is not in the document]
kirankt: Thanks you @2uasimojo and @staebler! I've incorporated your suggestions into the PR. PTAL. Thanks.

pkg/destroy/azure/azure.go
```diff
@@ -488,6 +482,13 @@ func deleteResourceGroup(ctx context.Context, client resources.GroupsClient, log
 	return nil
 }
 
+func isNotFoundError(err error) bool {
+	if detailedErr, ok := err.(autorest.DetailedError); ok && detailedErr.StatusCode == http.StatusNotFound {
```
2uasimojo, line 486: I noticed the other methods use `errors.As()` rather than casting like this. Then again, they also use hardcoded response codes rather than consts. :man_shrugging:

2uasimojo (commented):
/lgtm

pkg/destroy/azure/azure.go
staebler: Let's stick with `errors.As`. That will accept a DetailedError which is wrapped in another error.

pkg/destroy/azure/azure.go
```diff
@@ -488,6 +482,13 @@ func deleteResourceGroup(ctx context.Context, client resources.GroupsClient, log
 	return nil
 }
 
+func isNotFoundError(err error) bool {
```
staebler, line 485: nit: move this function down with the other `is***Error` functions.

pkg/destroy/azure/azure.go
```diff
@@ -492,6 +486,23 @@ func wasNotFound(resp *http.Response) bool {
...
+	}
+
+	var dErr autorest.DetailedError
+	if errors.As(err, &dErr) {
+		switch statusCode := dErr.StatusCode.(type) {
+		case int:
+			if statusCode == http.StatusNotFound {
+				return true
+			}
+		}
```
staebler, line 501: This is overkill for an equality check. I don't know why the `isResourceGroupBlockedError` function does this. But let's keep it as is for consistency. We can change them both as a separate PR if we like.
```suggestion
		return dErr.StatusCode == http.StatusNotFound
```

staebler (commented):
/lgtm
/approve

staebler:
@kirankt Can you verify for me that this change does not regress the destroyer when attempting to destroy a cluster using a resource group that has already been deleted?

pkg/destroy/azure/azure.go
kirankt: Yep. I thought about it but kept it to make it look consistent with the other fns. In any case, I'll update. Please lgtm again. Thanks

kirankt: Doh. Got merged before I can make this change. Will create a separate PR to simplify these.

PR #5317, opened 2021-10-20
Bug 1956776: Fix network validation

rna-afk:
The validateNetwork function is in place but is never
being called due to a slight if condition where the error
returned is always not nil. Fixing the code in order for
subsequent validations to run.

This is to fix #4906

### DIFF
--- a/pkg/asset/installconfig/vsphere/validation.go
+++ b/pkg/asset/installconfig/vsphere/validation.go
@@ -22,8 +22,8 @@ func Validate(ic *types.InstallConfig) error {
 	}
 
 	p := ic.Platform.VSphere
-	if err2 := validation.ValidatePlatform(p, field.NewPath("platform").Child("vsphere")); err2 != nil {
-		return err2.ToAggregate()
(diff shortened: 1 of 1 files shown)


pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -22,8 +22,8 @@ func Validate(ic *types.InstallConfig) error {
 	}
 
 	p := ic.Platform.VSphere
-	if err2 := validation.ValidatePlatform(p, field.NewPath("platform").Child("vsphere")); err2 != nil {
-		return err2.ToAggregate()
+	if err2 := validation.ValidatePlatform(p, field.NewPath("platform").Child("vsphere")).ToAggregate(); err2 != nil {
+		return err2
```
staebler, line 26: I find it easier to look at this if we treat the result of `ValidatePlatform` as it is meant to be treated--as a collection of errors.
```suggestion
	if errs := validation.ValidatePlatform(p, field.NewPath("platform").Child("vsphere")); len(errs) > 0 {
		return errs.ToAggregate()
```

PR #5325, opened 2021-10-25
Set CredentialsMode on Manual-only Platforms

patrickdillon:
CredentialsMode should be set for platforms that only allow manual mode so that no cloud credentials are uploaded to the cluster by the installer. #4416 created logic in the installer to not create cluster credentials when running in manual mode, but unless `CredentialsMode: Manual` is explicitly set in the install config, that logic would be ignored. 

With Azure Stack Hub, IBM, and Alibaba being manual-only platforms, we should use a default credential mode for these platforms to ensure that manual-mode logic is being observed. We should also consider future work of disabling the CCO when manual mode is set.


SHAKY404 (approved):

sadasu:
@patrickdillon is this still an issue we want to solve?

PR #5328, opened 2021-10-26
LAB/cluster-bot: change default AWS IPI to instance=m6i and EBS=gp3

mtulio:
**LAB Only**: to be used in cluster-bot as it seems to not support specifying specific GH user's PR (eg mtulio/installer#6)

Changes to increase control plane performance just changing to the most recent AWS instance (m6i) and volume type (gp3). 

Joint PRs for https://github.com/openshift/installer/pull/5327 and https://github.com/openshift/installer/pull/5239


mtulio:
We don't need more this PR to be used in cluster-bot, the https://github.com/openshift/installer/pull/5373 was merged

PR #5330, opened 2021-10-26
azure/validation: Skip "resource group empty" check for ARO

mbarnes:
As of release 4.9, the `Cluster` target now inherits the `PlatformProvisionCheck` dependency by way of [this commit for baremetal](https://github.com/openshift/installer/commit/e6ea0c6364837204428a08abfc9a2cdae2076960).

In the context of [Azure Red Hat OpenShift (ARO)](https://azure.microsoft.com/en-us/services/openshift/), all provisioning checks are passing except for one, which we need to disable.  Since the [ARO Resource Provider](https://github.com/Azure/ARO-RP) provisions Azure resources _before_ resolving the installer asset graph, it's normal for the Azure resource group to be populated.

cc: @m1kola

### DIFF
--- a/pkg/asset/installconfig/azure/validation_test.go
+++ b/pkg/asset/installconfig/azure/validation_test.go
@@ -340,6 +340,7 @@ var validGroupWithConflictinsTagsResult = &azres.Group{
 func Test_validateResourceGroup(t *testing.T) {
 	cases := []struct {
 		groupName string
+		wantSkip  bool
 		err       string
 	}{{
 		groupName: "non-existent-group",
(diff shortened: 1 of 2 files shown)


pkg/asset/installconfig/azure/validation_test.go
```diff
@@ -349,7 +351,10 @@ func Test_validateResourceGroup(t *testing.T) {
 		err:       `^\Qplatform.azure.resourceGroupName: Invalid value: "valid-resource-group-conf-tags": resource group has conflicting tags kubernetes.io_cluster.test-cluster-12345\E$`,
 	}, {
 		groupName: "valid-resource-group-with-resources",
-		err:       `^\Qplatform.azure.resourceGroupName: Invalid value: "valid-resource-group-with-resources": resource group must be empty but it has 3 resources like id1, id2 ...\E$`,
+		// ARO provisions Azure resources before resolving the asset graph,
+		// so there will always be resources in its resource group.
+		wantSkip: p.IsARO(),
```
staebler, line 356: The `p` variable is too far away to use the name `p` for the variable. So either rename the variable so that it is clear what the variable is, or don't use a variable.
```suggestion
		wantSkip: (&azure.Platform{}).IsARO(),
```

staebler (commented):
Looks fine to me. Please squash down the commits.

mbarnes:
Hey @staebler, sorry I missed your comment here way back in November.

ARO is re-evaluating our installer patches so I came across this again and applied your requests.  This should be good to go now.

If you're not already aware, ARO is pursing integrating Hive and dropping our installer fork, and that will mean letting the installer use Terraform for ARO resource provisioning.  So it's not clear if we'll need this workaround for the long term, but I'd still like to have it merged until we're sure we _don't_ need it.

PR #5331, opened 2021-10-26
Use Openshift Cluster API instead of MachineAPIOperator

kirankt:
Reduce the references to MachineAPIOperator for Cluster API and instead we will utilize OpenShift API. The MachineAPIOperator must still be vendored for the time being because of the vsphere CAPI provider resides within its repo unlike other providers who have their own separate repo.


patrickdillon:
/hold cancel

This looks good, but I suspect we might be able to do much more in this area. I will follow up on it and report back here or in slack.

patrickdillon:
It looks like we should also be able to remove the cluster-api provider dependencies for aws, gcp, and azure as well. Incremental progress is good, so let's merge this and close [CORS-1788](https://issues.redhat.com/browse/CORS-1788) but I think we should open a new card as well. I will leave a comment in CORS-1788

PR #5333, opened 2021-10-27
Alibabacloud recommitted and fixed

bd233:
This PR is based on [#5018](https://github.com/openshift/installer/pull/5018), and applies the modification of [#5291](https://github.com/openshift/installer/pull/5291),  and fixes the issues raised in #5291.


bd233:
@patrickdillon @kwoodson I have removed the pkg/destroy commit and updated the vendor. Please continue to review the code.

patrickdillon:
@bd233 This looks to be in very good shape. Can you please squash as many of the fix commits as possible? I would prefer that you squash them all and it looks like most of them should squash easily. Please make sure to squash 28d5d9cbad52f84a114e5b8200443a402c5d5758 Alibaba: fix: reconstruction GenerateIgnitionShim function

I think it is ok to not squash the last commit, Alibaba: fix: add VSwitch and security group info for machines but it would be better if you could update the commit message to remove `fix` from the title and provide a little more detail.

@kwoodson does the importing in c8c38d3277b84f273b4b2e12988c2babf2e05dc3 look correct to you? I believe you were talking about this in the call today? 

I think if we get those commits squished and Kenny confirms about the importing/testing we should be in good shape to merge. Thank you!

kwoodson:
@bd233 @dongchen126

One simple update remaining to this file:
https://github.com/openshift/installer/pull/5333/files#diff-d8695174d83d5a37969524935693f06fda5f8f2c947f3d979a7162dca4721f64R8

Should be:
```
```

bd233:
> @bd233 @dongchen126
> 
> One simple update remaining to this file: https://github.com/openshift/installer/pull/5333/files#diff-d8695174d83d5a37969524935693f06fda5f8f2c947f3d979a7162dca4721f64R8
> 
> Should be:
> 
> ```
> 	machineapi "github.com/openshift/api/machine/v1beta1"
> ```

Have updated

kwoodson:
@bd233 @dongchen126 Congrats!

Please update #5348 so that we can finish that work.

PR #1242, opened 2019-02-13
Append router CA to cluster CA in kubeconfig

sallyom:
With users created with an identity-provider, OAuth flow goes through router and router-ca is not trusted.  This prohibits the user from using `oc login` from command line, without manually appending the router ca to the cluster ca.  The  openshift-ingress-operator creates the router-CA and this is not available until the very end of an install.

This PR adds the router CA from `configmap router-ca -n openshift-config-managed` to the kubeconfig certificate-authority-data. This allows an identity-provided user to `oc login` from the terminal.

PR #1132, opened 2019-01-26
cmd: wait for the cluster to be initialized

staebler:
After creating the cluster, wait until the ClusterVersion object indicates that the cluster has been initialized prior to exiting the installer.

PR #806, opened 2018-12-06
add logs at end of install for kubeadmin, consoleURL

sallyom:
@wking @eparis 
This PR adds logging at end of install, for kubeadmin login and console url.
This is what I have come up with so far:
```
DEBUG module.bootstrap.aws_iam_role.bootstrap: Destruction complete after 1s 
DEBUG                                              
DEBUG Destroy complete! Resources: 9 destroyed.    
INFO Install complete! To manage the cluster with 'oc', the OpenShift CLI, first run 'export KUBECONFIG=/home/somalley/installer/somalley-test/auth/kubeconfig' 
INFO The cluster is ready when this command succeeds (wait a few minutes): 'oc login -u kubeadmin -p azsvQ-7chno-A6KSa-6cta2' 
INFO Access the OpenShift web-console here with user: kubeadmin and pw: azsvQ-7chno-A6KSa-6cta2, `https://console-openshift-console.apps.somalley-test.devcluster.openshift.com` 
```
I'm wondering if it makes sense to add a wait/retry with that oc login command after the bootstrap destroy, or if it's enough to tell a user to wait until it works.

PR #5336, opened 2021-10-27
cmd/openshift-install/create: One shot console access

wking:
The console may become optional (openshift/enhancements#922), so teach the installer to handle its absence gracefully.

We've waited on the console since way back in ff53523b11 (#806).  Back then, install-complete timing was much less organized, and since e17ba3c571 (#1132) we've blocked on ClusterVersion going `Available=True`.  So the current dependency chain is:

1. Console route admission blocks console operator from going `Available=True` in its ClusterOperator.
2. Console ClusterOperator blocks cluster-version operator from going `Available=True` in ClusterVersion.
3. ClusterVersion blocks installer's `waitForInitializedCluster`.

So we no longer need to wait for the route to show up, and can fail fast if we get a clear `IsNotFound`.  I'm keeping a bit of polling so we don't fail an install on a temporary network hiccup.

We don't want to drop the console check entirely, because when it is found, we want:

* To continue to log that access pathway on install-complete.
* To continue to append the router CA to the kubeconfig.

That latter point has been done since 40335776f6 (#1242).  The motication in that commit message is not explicit, but the idea is to support folks who [naively run `oc login` with the kubeadmin kubeconfig][2] (despite that kubeconfig already having cluster-root access) when the console route's cert's CA happens to be something that the user's local trust store doesn't include by default.

[1]: https://github.com/openshift/enhancements/pull/922

### DIFF
--- a/cmd/openshift-install/create.go
+++ b/cmd/openshift-install/create.go
@@ -13,6 +13,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


wking:
I can revive this if/when folks want it, just let me know.  Until then, I'm not going to try to keep on top of rebases, so closing.

wking:
Re-opened and rebased onto master with dca82dfd03 -> b05f71aff3 now that openshift/console-operator#665 is failing on the lack of this change or something similar.

PR #5338, opened 2021-10-27
Bug 2017986: Set AWS Bootstrap Type == Master

patrickdillon:
This commit sets the boostrap node to use the same instance type as the control-plane nodes in order to avoid failures that we have seen when the hard-coded instance type of m5 is unavailable in the zone.


sdodson:
Overriding bugzilla/valid-bug so that we ensure this makes it back to 4.8 for the next z-stream.

PR #5342, opened 2021-10-28
[splat-330] vSphere: Modify the platform spec

jcpowermac:
Modification of the platform spec to include
`VCenters`, `Regions` and `Zones` that will allow
deployment of a OpenShift cluster in vSphere across
multiple vCenter datacenters and clusters. This will also
enable Regions and Zones as implemented by the cloud-provider.


jcpowermac:
don't break the current install...
/test e2e-vsphere

PR #5346, opened 2021-11-01
bug 1960378: Relax the imageContentSource validation

kirankt:
Relax the imageContentSource validation to allow for mirroring root registry.

This is a follow up to a previous attempt: #5297 which didn't quite fix the issue. The use of url.Parse does not work as expected. This PR utilized validate.Host function to check for a valid host name/IP when the canonical check fails.

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -566,9 +566,14 @@ func validateNamedRepository(r string) error {
 		// in an error. Instead we will check whether the input is
 		// a valid hostname as a workaround.
 		if err == dockerref.ErrNameNotCanonical {
-			_, err := url.ParseRequestURI(r)
+			// If the hostname string contains a port, lets attempt
+			// to split them
+			host, _, err := net.SplitHostPort(r)
 			if err != nil {
-				return fmt.Errorf("the repository provided is invalid")
+				host = r
+			}
+			if err = validate.Host(host); err != nil {
+				return errors.Wrap(err, "the repository provided is invalid")
 			}
 			return nil
 		}
--- a/pkg/types/validation/installconfig_test.go
+++ b/pkg/types/validation/installconfig_test.go
@@ -1092,7 +1092,7 @@ func TestValidateInstallConfig(t *testing.T) {
 				}}
 				return c
 			}(),
-			expectedError: `^imageContentSources\[0\]\.source: Invalid value: "ocp/release-x\.y": the repository provided is invalid$`,
+			expectedError: `^imageContentSources\[0\]\.source: Invalid value: "ocp/release-x\.y": the repository provided is invalid: a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, \'\-\' or \'\.\', and must start and end with an alphanumeric character \(e.g. \'example\.com\', regex used for validation is \'\[a\-z0\-9\]\(\[\-a\-z0\-9\]\*\[a\-z0\-9\]\)\?\(\\\.\[a\-z0\-9\]\(\[\-a\-z0\-9\]\*\[a\-z0\-9\]\)\?\)\*\'\)`,
 		},
 		{
 			name: "release image source's mirror is not valid",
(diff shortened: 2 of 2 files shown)


pkg/types/validation/installconfig.go
```diff
@@ -542,9 +542,22 @@ func validateNamedRepository(r string) error {
...
-			if err != nil {
-				return fmt.Errorf("the repository provided is invalid")
+			// First check whether the repo name provided contains
+			// host:port. If so, split the host part and lookup
+			// to see if it is a valid host. If the port is
+			// not provided, then just check the name to see if
+			// the repo name resolves correctly.
+			host, _, err := net.SplitHostPort(r)
+			if err == nil {
+				_, err = net.LookupHost(host)
```
patrickdillon, line 552: A valid registry might only be accessible to the bootstrap node but not the installer host. In which case this would incorrectly fail validation.

patrickdillon (commented):
Can you provide some more context as how the QE test case was not covered by the previous solution? I think that might help determine the best solution.

pkg/types/validation/installconfig.go
kirankt: You're right. If we can't resolve the name then I'll keep it simple and look for a valid dns name string characters.

pkg/types/validation/installconfig.go
```diff
@@ -542,8 +543,13 @@ func validateNamedRepository(r string) error {
...
 		// a valid hostname as a workaround.
 		if err == dockerref.ErrNameNotCanonical {
-			_, err := url.ParseRequestURI(r)
+			// If the hostname string contains a port, lets attempt
+			// to split them
+			host, _, err := net.SplitHostPort(r)
 			if err != nil {
+				host = r
+			}
+			if !govalidator.IsHost(host) {
```
staebler, line 552: Can we use https://github.com/openshift/installer/blob/c88793ce714db6856284faa91542a2cfb7b51437/pkg/validate/validate.go#L275 (once https://github.com/openshift/installer/pull/5367 merges)? I would rather that than bring in another dependency.

kirankt: +1. Will change. Thanks.

pkg/types/validation/installconfig.go
```diff
@@ -566,8 +566,13 @@ func validateNamedRepository(r string) error {
...
 		if err == dockerref.ErrNameNotCanonical {
-			_, err := url.ParseRequestURI(r)
+			// If the hostname string contains a port, lets attempt
+			// to split them
+			host, _, err := net.SplitHostPort(r)
 			if err != nil {
+				host = r
+			}
+			if err = validate.Host(host); err != nil {
 				return fmt.Errorf("the repository provided is invalid")
```
staebler, line 576: Let's wrap around the error returned from `validate.Host`.
```suggestion
				return errors.Wrap(err, "the repository provided is invalid")
```

pkg/types/validation/installconfig.go
```diff
@@ -566,8 +566,13 @@ func validateNamedRepository(r string) error {
 		// in an error. Instead we will check whether the input is
 		// a valid hostname as a workaround.
 		if err == dockerref.ErrNameNotCanonical {
-			_, err := url.ParseRequestURI(r)
+			// If the hostname string contains a port, lets attempt
+			// to split them
+			host, _, err := net.SplitHostPort(r)
 			if err != nil {
+				host = r
```
staebler, line 573: Can you add a short comment describing that this case is when the repository does not contain a port? It requires a bit of thinking when reading this to understand why we can simply ignore the error.

pkg/types/validation/installconfig.go
kirankt: Usually I like wrapping things but this one is unnecessarily verbose. This is what gets spit out:
```Invalid value: "ocp/release-x.y": the repository provided is invalid: a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

pkg/types/validation/installconfig.go
kirankt: Done. Thanks.

pkg/types/validation/installconfig.go
staebler: I personally prefer the more verbose error rather than a statement that the repository is invalid without any details about how it is invalid.

kirankt: Done.
Just so that you know, I was cursing at you while escaping the regex in the validation. :smile:

staebler: Fair enough. For what it's worth, I would have happily accepted `.*` for the majority of the error message.

pkg/types/validation/installconfig.go
```diff
@@ -566,9 +566,14 @@ func validateNamedRepository(r string) error {
 		// in an error. Instead we will check whether the input is
 		// a valid hostname as a workaround.
 		if err == dockerref.ErrNameNotCanonical {
-			_, err := url.ParseRequestURI(r)
+			// If the hostname string contains a port, lets attempt
```
staebler, line 569: nit
```suggestion
			// If the hostname string contains a port, let's attempt
```

staebler (approved):
/lgtm

PR #5348, opened 2021-11-02
alibaba: implement cluster destroy

bd233:
Add code for the `openshift-install destroy cluster` command for cluster using the alibaba platform.


kwoodson:
@bd233 Please rebase this now that #5333 has merged.

I rely on the destroy code to clean up my clusters. Thanks!

kwoodson:
@patrickdillon I have been testing this destroy code and it does clean up the cluster. I'll go through and verify again.

bd233:
@kwoodson I have fixed it

bd233:
@patrickdillon Please review whether this is the desired validation [Alibaba: fix: add metadata server IP validation](https://github.com/openshift/installer/pull/5348/commits/a0f6d805dfe88cd5436e34a2c31635a6d72cbb83)

bd233:
@kwoodson I have a question, what should the number of worker nodes in the default configuration (if I do not modify the install-config file) depend on? Is the number of all available zones available?

kwoodson:
@bd233 
It appears the replicas are set by default by the number of zones:
https://github.com/openshift/installer/blob/master/pkg/asset/machines/alibabacloud/machinesets.go#L25-L38

The zones are determined here https://github.com/openshift/installer/blob/master/pkg/asset/machines/worker.go#L239-L249

As you know, this can be determined in any method you desire. I think the current behavior is 1 worker per zone but I think the standard node count is generally 3 worker nodes. That provides sufficient space to run the ingress with HA as well as the other cluster resources (registry, console, etc).

It appears that other providers do something similar. AWS looks at subnets or number of AZs. I think you will always get number of availabilityzones + 1. Due to this code:
```
		if int64(idx) < total%numOfAZs {
			replicas++
		}
```

patrickdillon:
Default cluster should always have 3 compute nodes.

kwoodson:
@bd233 I've noticed that the SLB that is created by the ingress operator has DeletionProtection enabled. Therefore the ingress controller's SLB does not get removed.

Also appears that the OSS bucket that is created for the bootstrap node also does not get deleted.

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -172,6 +173,7 @@ func validatePrivateZoneID(client *Client, ic *types.InstallConfig, path *field.
 // ValidateForProvisioning validates if the install config is valid for provisioning the cluster.
 func ValidateForProvisioning(client *Client, ic *types.InstallConfig, metadata *Metadata) error {
 	allErrs := field.ErrorList{}
+	allErrs = append(allErrs, validateMetadataServerIPNotInMachineCIDR(ic.Networking)...)
```
patrickdillon, line 176: Because these functions are static validation--that is, they do not require connecting to the alibaba api--they should be moved to `pkg/types`. Specifically here: https://github.com/openshift/installer/blob/master/pkg/types/validation/installconfig.go#L465

the `validatePlatform` function will need to pass in the networking struct, similar to how OpenStack does.

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -190,3 +192,23 @@ func validateClusterName(client *Client, ic *types.InstallConfig) field.ErrorLis
 	}
 	return allErrs
 }
+
+func validateMetadataServerIPNotInMachineCIDR(n *types.Networking) field.ErrorList {
+	allErrs := field.ErrorList{}
+	fldPath := field.NewPath("networking").Child("machineNetwork")
+	matedataServerIP := "100.100.100.200"
```
patrickdillon, line 199: `matedataServerIP` -> `metadataServerIP`

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -190,3 +192,23 @@ func validateClusterName(client *Client, ic *types.InstallConfig) field.ErrorLis
...
+	if err := validateIPNotinMachineCIDR(matedataServerIP, n); err != nil {
+		allErrs = append(allErrs, field.Invalid(fldPath, matedataServerIP, err.Error()))
+	}
+	return allErrs
+
+}
+
+func validateIPNotinMachineCIDR(ip string, n *types.Networking) error {
+	for _, network := range n.MachineNetwork {
```
patrickdillon, line 208: we should probably check the cluster network and service network as well

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -190,3 +192,23 @@ func validateClusterName(client *Client, ic *types.InstallConfig) field.ErrorLis
...
+		if network.CIDR.Contains(net.ParseIP(ip)) {
+			return fmt.Errorf("the IP must not be in one of the machine networks")
```
patrickdillon, line 210: "the IP must not be in one of the machine networks" -> "the <network>  contains 100.100.100.200 which is reserved for the metadata service"

patrickdillon (commented):
I just reviewed the metadata server validation. This stands alone as it's own PR so I would suggest moving it to a new one.

bd233:
> @bd233 It appears the replicas are set by default by the number of zones+1: https://github.com/openshift/installer/blob/master/pkg/asset/machines/alibabacloud/machinesets.go#L25-L38
> 

thx, I think I understand the logic here, `total` is the total number of nodes (from the pool.Replicas, the default is 3), `replicas` is the number of nodes in each availability zone, if total is less than the number of availability zones, `replicas` is 0 in some availability zones.

bd233:
> @bd233 I've noticed that the SLB that is created by the ingress operator has DeletionProtection enabled. Therefore the ingress controller's SLB does not get removed.
> 
@kwoodson I think this has nothing to do with deleting protection. If you run the `installer destroy cluster` command directly, the ECS instance will be released immediately, and CCM has no chance to delete the SLB, so the installer should be responsible for deleting this SLB.

In addition, as you know, the Installer will create two SLB instances. The installer searches for the two SLB instances through tags `kubernetes.io/cluster/<cluster_id>: owned`, turns off the delete protection status and deletes them.
I will use a tag `ack.aliyun.com:<cluster_id>` to find the SLB created by ingress operator and delete it. (I need you to help me verify that this is feasible, I can’t create it to this point)

> Also appears that the OSS bucket that is created for the bootstrap node also does not get deleted.

I created the cluster and deleted it, but it did not reproduce the problem. Is there any more detailed information to help me reproduce it?

bd233:
@kwoodson [Alibaba: fix: destroy SLB created by ingress operator](https://github.com/openshift/installer/pull/5348/commits/67d08922d94d91678ef2e382113af018267eee05)

kwoodson:
@bd233 Let me test the latest code and verify if the SLB gets removed. I'll also look into the OSS buckets.

pkg/destroy/alibabacloud/alibabacloud.go
```diff
@@ -0,0 +1,1380 @@
...
+	if !isSuccessful {
+		notDeletedResources := []string{}
+		tagResources, err := o.findResourcesByTag()
+		if err != nil {
+			return err
+		}
+		for _, arn := range tagResources {
+			notDeletedResources = append(notDeletedResources, arn.ResourceARN)
+		}
+		return errors.New(fmt.Sprintf("There are undeleted cloud resources %q", notDeletedResources))
```
patrickdillon, line 352: As staebler pointed out in the previous PR, the destroy code should never stop running if it knows that there are outstanding resources to be deleted. The destroyer should loop, attempting to destroy any resources as long as they exist.

patrickdillon (commented):
I think it would be easier to review if you broke this in to multiple PRs. One for destroy, one for existing VPC, etc. 

Also please squash where possible. I know there is at least one destroy commit that could be squashed.

pkg/destroy/alibabacloud/alibabacloud.go
bd233: Yes, it is inappropriate to raise an error here.
But I think `waitComplete` here is redundant. The delete function of each cloud resource already contains the logic to wait for the deletion to be completed, so I should remove `waitComplete`.

pkg/destroy/alibabacloud/alibabacloud.go
```diff
@@ -0,0 +1,1343 @@
...
+			close(wgDone)
+		}()
+
+		select {
+		case <-wgDone:
+			// On to the next stage
+			continue
+		case err := <-errCh:
+			return err
+		}
```
patrickdillon, line 232: If I'm reading this code correctly, an error returned from any of the destroy functions will cause the destroyer to stop. So if we get an error when trying to destroy a DNS record in the first stage, none of the other stages will be executed.

The destroyer cannot stop on an error. It must continuously loop trying to destroy all known resources. It stops when it cannot find more resources to delete.

It looks like this destroy code may be modeled after IBM, which seems to have this same issue. I will bring this up with them to change. 

If you want to follow the current async model you have you can look at the openstack code: https://github.com/openshift/installer/blob/master/pkg/destroy/openstack/openstack.go

Destroy does not have to be async, in which case you can follow the model from AWS.

patrickdillon: Oops. I see now I was misreading this code. errCh is not passed to the stage functions

pkg/destroy/alibabacloud/alibabacloud.go
```diff
@@ -0,0 +1,1343 @@
...
+		resources, err := o.ListTagResources(tags)
+		if err != nil {
+			return nil, err
+		}
+		tagResources = append(tagResources, resources...)
+	}
+	return tagResources, nil
+}
+
+func (o *ClusterUninstaller) ListTagResources(tags map[string]string) (tagResources []tag.TagResource, err error) {
```
patrickdillon, line 987: This is causing the go-lint test to fail: 

```
 /go/src/github.com/openshift/installer/pkg/destroy/alibabacloud/alibabacloud.go:987:1: exported method ClusterUninstaller.ListTagResources should have comment or be unexported
```

patrickdillon (commented):
Can you fix this one small error to pass the test, and we should be ready to merge?

pkg/destroy/alibabacloud/alibabacloud.go
bd233: done, please continue

kwoodson:
@bd233 I attempted to run this destroy code today but it continued to run. I ran this with `--log-level debug` and was able to see that it was attempting to delete the DNS records.
```
DEBUG Start to search DNS records                  
DEBUG DNS records: SDK.ServerError                 
DEBUG ErrorCode: InvalidDomainName.NoExist         
DEBUG Recommend: https://error-center.aliyun.com/status/search?Keyword=InvalidDomainName.NoExist&source=PopGw 
DEBUG RequestId: 470B1762-645B-53AA-A907-A5796994BD4D 
DEBUG Message: The specified domain name does not exist. Refresh the page and try again. 
...
```

If the record `InvalidDomainName.NoExist` is returned, I would assume that the deletion occurred already. Is it possible to include the `NoExist` as a successful delete?

cc @patrickdillon

pkg/destroy/alibabacloud/alibabacloud.go
```diff
@@ -0,0 +1,1343 @@
...
+func (o *ClusterUninstaller) executeStageFunction(f struct {
+	name    string
+	execute func() error
+}, errCh chan error, wg *sync.WaitGroup) error {
+	defer wg.Done()
+
+	err := wait.PollImmediateInfinite(
+		time.Second*10,
+		func() (bool, error) {
+			ferr := f.execute()
```
kwoodson, line 247: I do not love this solution and I am not proposing it but this addition was able to skip the resources that are missing:
```
			if ferr != nil {
				if strings.Contains(ferr.Error(), "not exist") {
					return true, nil
				}
				o.Logger.Debugf("%s: %v", f.name, ferr)
				return false, nil
			}
```

I recommend using something similar to this to skip waiting if an object returns `NoExist`.
@bd233 @patrickdillon WDYT?

patrickdillon: Is the `deleteDNSRecords` function incorrectly returning an error when it can't find the domain?

kwoodson: The `deleteDNSRecords` is returning an error that the DNS record does not exist. I think this behavior is expected from the Alibaba API. We need to gracefully handle `NoExist` errors so that deletion can complete successfully.

@bd233 Can you confirm this behavior?

patrickdillon: Another option is to check to see if it exists before trying to delete. That is what I thought this code was doing: https://github.com/openshift/installer/pull/5348/files#diff-5d31e7ab73dc252cc09331a0790fcc8b6cd3add977402effbe19a1b1b24fba39R1249-R1255

bd233: I'm sorry to reply so late

bd233: @kwoodson @patrickdillon  Sorry to reply so late.
I think this is an abnormal scenario. I read the `Start to search DNS records` log, but did not see `Start to delete DNS records`, so the error should come from `records, err: = o.listrecord (basedomain)`, But this should not be, if the domain name does not exist, it should be returned here:

> 	 if len(domains) == 0 {
> 		 return
> 	}

I am trying to reproduce and fix this problem as soon as possible.

>  I recommend checking the ferr message for objects that do not exist and return true to continue the deletion.
>  @bd233 @patrickdillon WDYT?

I don't think 'ferr' should be handled. This error should be resolved in 'deleteDnsRecords'. This solution may cause some resources not to be deleted

@kwoodson Are there any manual deletion operations in the process of executing the installer？

kwoodson: @bd233 Generally speaking there shouldn't be any but that doesn't stop customers from updating or modifying the cluster objects. The reason this DNS item was modified is that we set up valid DNS and removed the original before removing the cluster. That caused the missing DNS record. I think that it is not unreasonable to have components get modified or deleted outside of the cluster. I think if objects return `NoExist` they should be considered to be deleted and the destroy code can continue.

bd233: @kwoodson @patrickdillon 
I think I understand this problem. Querying domainName is a fuzzy query. If baseDomain is `openshift.com`, then` bd.openshift.com` may appear in the result, and `len (domains) == 0 ` is false, so...

I'll fix it as soon as possible.

bd233: I think `ListPrivateZones` and `ListDNSDomain` in `pkg/asset/installconfig/alibabacloud/client.go` also has this problem. I will create a PR to fix it.

pkg/destroy/alibabacloud/alibabacloud.go
```diff
@@ -0,0 +1,1345 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return response.Zones.Zone, nil
+}
+
+func (o *ClusterUninstaller) deleteDNSRecords() (err error) {
+	o.Logger.Debug("Start to search DNS records")
+
+	baseDomain := strings.Join(strings.Split(o.ClusterDomain, ".")[1:], ".")
```
patrickdillon, line 1249: Is this the fix for the fuzzy domain search error? If so can you explain how this solves the problem? Also it would be nice to have a comment in the code  for why we need the string manipulation

patrickdillon (commented):
Can you explain (briefly) the fix for the dns issue? I am not sure I understand how it is being fixed.

Also it looks like an azure stack commit snuck in during a rebase.

pkg/destroy/alibabacloud/alibabacloud.go
bd233: Aha~, about azure stack commit, it is a rebase mistake, I will updata it later.
The `LIKE` is default search mode on  DescribeDomains api. In this mode, if you search `openshift.com`,  can match `xx.openshift.com` and `openshift.com.xx`, but using `EXACT` mode, only return `openshift.com`.

> Also it would be nice to have a comment in the code for why we need the string manipulation

I need base domain to search domain and delete it records, but I can not get base domain, so I use cluster domin to splite by `.` (the format of cluster domain is `<cluster name>.<base domain>`).

I will add comment here.

kwoodson:
I tested this yesterday:
```
~/go/src/github.com/openshift/installer/bin/openshift-install destroy cluster --dir cluster --log-level debug
DEBUG OpenShift Installer unreleased-master-5214-gc80668d4b12c8a2d89f2914b24877d6d7a43c8f3 
DEBUG Built from commit c80668d4b12c8a2d89f2914b24877d6d7a43c8f3 
DEBUG Retrieving cloud resources by tag {"kubernetes.io/cluster/test-p79w9":"owned"} 
DEBUG Retrieving cloud resources by tag {"ack.aliyun.com":"test-p79w9"} 
DEBUG Start to search DNS records                  
DEBUG Start to search RAM policy "test-p79w9-policy-bootstrap" attachments 
DEBUG Start to check and turn off ECS instances ["i-0xi32ol31nc3j9j5akfz" "i-0xicg6qlyjeopjflu4y1" "i-0xig89d4vg0r8mwf6ov9" "i-0xibcqegh9hceeerbb8h" "i-0xie1u8s8lhavztcj64a" "i-0xifkuk3z82eyof7b6kp" "i-0xi5hd6brb1uccksrzc5"] deletion protection 
DEBUG Start to delete ECS instances ["i-0xi32ol31nc3j9j5akfz" "i-0xicg6qlyjeopjflu4y1" "i-0xig89d4vg0r8mwf6ov9" "i-0xibcqegh9hceeerbb8h" "i-0xie1u8s8lhavztcj64a" "i-0xifkuk3z82eyof7b6kp" "i-0xi5hd6brb1uccksrzc5"] 
DEBUG Start to delete buckets "test-p79w9-bootstrap" 
DEBUG Untag cloud resources &["arn:acs:oss:us-east-1:*:bucket/test-p79w9-bootstrap"] with tags &["kubernetes.io/cluster/test-p79w9"] 
DEBUG Start to detach RAM policy "test-p79w9-policy-bootstrap" 
DEBUG Start to detach RAM policy "test-p79w9-policy-bootstrap" with "test-p79w9-role-bootstrap@role.5807403157081522.onaliyunservice.com" 
DEBUG Start to delete DNS records                  
DEBUG Start to delete DNS record "731226929327065088" 
DEBUG Start to delete objects of buckets test-p79w9-bootstrap 
DEBUG Start to search and delete RAM policy "test-p79w9-policy-bootstrap" 
DEBUG Start to search and delete RAM role "test-p79w9-role-bootstrap" 
DEBUG Start to search RAM policy "test-p79w9-policy-master" attachments 
DEBUG Start to detach RAM policy "test-p79w9-policy-master" 
DEBUG Start to detach RAM policy "test-p79w9-policy-master" with "test-p79w9-role-master@role.5807403157081522.onaliyunservice.com" 
DEBUG Start to search and delete RAM policy "test-p79w9-policy-master" 
DEBUG Start to search and delete RAM role "test-p79w9-role-master" 
DEBUG Start to search RAM policy "test-p79w9-policy-worker" attachments 
DEBUG Start to detach RAM policy "test-p79w9-policy-worker" 
DEBUG Start to detach RAM policy "test-p79w9-policy-worker" with "test-p79w9-role-worker@role.5807403157081522.onaliyunservice.com" 
DEBUG Start to search and delete RAM policy "test-p79w9-policy-worker" 
DEBUG Start to search and delete RAM role "test-p79w9-role-worker" 
DEBUG Start to search private zones                
DEBUG Start to delete security groups ["sg-0xig7un0qyij3hguy35x" "sg-0xibkdteueprzcms7h4a" "sg-0xi3kcj0cd4may8umsb5"] 
DEBUG Start to delete SLBs ["lb-7goby9h2trqyoobuw6a8f" "lb-7go0jssnzhyycbh5qv7tl" "lb-7gop47wfz838slf1ak7s5" "lb-7gokgvsp68kukoroxmb4a"] 
DEBUG Start to delete NAT gateways ["ngw-0xihgz0tfk9thxvwzc1nv"] 
DEBUG Start to delete security group "sg-0xig7un0qyij3hguy35x" rules  
DEBUG Start to delete NAT gateway "ngw-0xihgz0tfk9thxvwzc1nv" 
DEBUG Start to unbind/bind private zone "dfefb8bf19aa992a7d68432f1216a33a" with vpc 
DEBUG Start to delete SLB "lb-7goby9h2trqyoobuw6a8f" 
DEBUG Start to delete private zones                
DEBUG Start to delete private zone "dfefb8bf19aa992a7d68432f1216a33a" 
DEBUG Start to delete SLB "lb-7go0jssnzhyycbh5qv7tl" 
DEBUG Start to delete SLB "lb-7gop47wfz838slf1ak7s5" 
DEBUG Start to delete SLB "lb-7gokgvsp68kukoroxmb4a" 
DEBUG Start to delete security group "sg-0xibkdteueprzcms7h4a" rules  
DEBUG Start to delete security group "sg-0xi3kcj0cd4may8umsb5" rules  
DEBUG Start to delete security group "sg-0xig7un0qyij3hguy35x" 
DEBUG Start to delete security group "sg-0xibkdteueprzcms7h4a" 
DEBUG Start to delete security group "sg-0xi3kcj0cd4may8umsb5" 
DEBUG Start to delete EIPs ["eip-0xi8mxqtfpfzwqfpm93yf"] 
DEBUG Start to delete EIP "eip-0xi8mxqtfpfzwqfpm93yf" 
DEBUG Start to delete VSwitchs ["vsw-0xihwtxkkqqkbj8nflbpy" "vsw-0xi87gk3s7vhkyh6xguyb" "vsw-0xivf3rei4llqzvz48y6j" "vsw-0xiomujawrayubh04nz04"] 
DEBUG Start to delete VSwitch "vsw-0xihwtxkkqqkbj8nflbpy" 
DEBUG Start to delete VSwitch "vsw-0xi87gk3s7vhkyh6xguyb" 
DEBUG Start to delete VSwitch "vsw-0xivf3rei4llqzvz48y6j" 
DEBUG Start to delete VSwitch "vsw-0xiomujawrayubh04nz04" 
DEBUG Start to delete VPCs ["vpc-0xive4iz8few58zdak5oq"] 
DEBUG Start to delete VPC "vpc-0xive4iz8few58zdak5oq" 
DEBUG Purging asset "Metadata" from disk           
DEBUG Purging asset "Master Ignition Customization Check" from disk 
DEBUG Purging asset "Worker Ignition Customization Check" from disk 
DEBUG Purging asset "Terraform Variables" from disk 
DEBUG Purging asset "Kubeconfig Admin Client" from disk 
DEBUG Purging asset "Kubeadmin Password" from disk 
DEBUG Purging asset "Certificate (journal-gatewayd)" from disk 
DEBUG Purging asset "Cluster" from disk            
INFO Time elapsed: 1m48s                         
```

PR #5349, opened 2021-11-03
openstack: Fix install config for openstack manifests

rna-afk:
The openstack manifests tests use an install config with the
clusterID field that is no longer supported by the installer.
Changes to the installer to enforce strict unmarshalling of the
install config is in place but is being hindered by the openstack
manifests tests in this PR.

https://github.com/openshift/installer/pull/5307


patrickdillon:
/approve

Will wait for an @openshift/team-openstack reviewer to do a final review.

EmilienM:
/lgtm

Thanks!

cc @pierreprinetti

PR #5350, opened 2021-11-03
Use openshift/api instead of machine-api-operator

kirankt:
This PR is a follow up to #5331.

In the previous PR, references to machine-api-operator was left behind. This PR removes the last references to machine-api-operator in the installer code. Please note that the MAO is still vendored indirectly because other packages haven't moved to openshift/api yet.

e.g. https://github.com/openshift/cluster-api-provider-ovirt/blob/master/pkg/apis/ovirtprovider/v1beta1/register.go


pkg/asset/machines/master.go
```diff
@@ -514,7 +512,7 @@ func (m *Master) Machines() ([]machineapi.Machine, error) {
 	libvirtapi.AddToScheme(scheme)
 	openstackapi.AddToScheme(scheme)
 	ovirtproviderapi.AddToScheme(scheme)
-	vsphereapi.AddToScheme(scheme)
+	machineapi.AddToScheme(scheme)
```
staebler, line 515: This is not an adequate replacement. The scheme from github.com/openshift/api/machine/v1beta1 does not include the provider specs [1]. If we want to use the types defined in openshift/api, then we will need to manual add the vSphere provider spec to the scheme.

[1] https://github.com/openshift/api/blob/master/machine/v1beta1/register.go#L34-L41

kirankt: I agree. I was aware of the CI failure and was looking for a solution for this.

kirankt: Added the vspherespec into scheme.

staebler:
From the latest [e2e-vsphere job](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_installer/5350/pull-ci-openshift-installer-master-e2e-vsphere/1461391361948258304/artifacts/e2e-vsphere/ipi-install-install/build-log.txt),
```
level=info msg=Install complete!
```

/lgtm
/approve
/retitle Use openshift/api instead of machine-api-operator

PR #5360, opened 2021-11-09
Alibaba:  add metadata server IP validation

bd233:
Validate if  machine, service and cluster networking CIDR  contains metadata server IP

### DIFF
--- a/pkg/types/alibabacloud/validation/platform.go
+++ b/pkg/types/alibabacloud/validation/platform.go
@@ -1,20 +1,49 @@
 package validation
 
 import (
+	"fmt"
+	"net"
+
+	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/alibabacloud"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 )
 
 // ValidatePlatform checks that the specified platform is valid.
-func ValidatePlatform(p *alibabacloud.Platform, fldPath *field.Path) field.ErrorList {
+func ValidatePlatform(p *alibabacloud.Platform, n *types.Networking, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	if p.Region == "" {
 		allErrs = append(allErrs, field.Required(fldPath.Child("region"), "region must be specified"))
 	}
 
-	if p.ResourceGroupID == "" {
-		allErrs = append(allErrs, field.Required(fldPath.Child("resourceGroupID"), "resource group ID must be specified"))
+	allErrs = append(allErrs, validateMetadataServerIP(n)...)
+
+	return allErrs
+}
+
+func validateMetadataServerIP(n *types.Networking) field.ErrorList {
+	allErrs := field.ErrorList{}
+	metadataServerIP := "100.100.100.200"
+	networkPath := field.NewPath("networking")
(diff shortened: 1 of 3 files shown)


bd233:
@patrickdillon I have updated here

pkg/types/alibabacloud/validation/platform.go
```diff
@@ -1,20 +1,49 @@
...
+	allErrs := field.ErrorList{}
+	metadataServerIP := "100.100.100.200"
+	networkPath := field.NewPath("networking")
+	machineNetworkPath := networkPath.Child("machineNetwork")
+	clusterNetworkPath := networkPath.Child("clusterNetwork")
+	serviceNetworkPath := networkPath.Child("serviceNetwork")
+
+	for _, network := range n.MachineNetwork {
+		if network.CIDR.Contains(net.ParseIP(metadataServerIP)) {
+			allErrs = append(allErrs, field.Invalid(machineNetworkPath, metadataServerIP, fmt.Errorf("the contains %s which is reserved for the metadata service", metadataServerIP).Error()))
```
patrickdillon, line 35:
```suggestion
			allErrs = append(allErrs, field.Invalid(machineNetworkPath, metadataServerIP, fmt.Errorf("contains %s which is reserved for the metadata service", metadataServerIP).Error()))
```

pkg/types/alibabacloud/validation/platform.go
```diff
@@ -1,20 +1,49 @@
...
+		}
+	}
+	for _, network := range n.ClusterNetwork {
+		if network.CIDR.Contains(net.ParseIP(metadataServerIP)) {
+			allErrs = append(allErrs, field.Invalid(clusterNetworkPath, metadataServerIP, fmt.Errorf("the contains %s which is reserved for the metadata service", metadataServerIP).Error()))
```
patrickdillon, line 40:
```suggestion
			allErrs = append(allErrs, field.Invalid(clusterNetworkPath, metadataServerIP, fmt.Errorf("contains %s which is reserved for the metadata service", metadataServerIP).Error()))
```

pkg/types/alibabacloud/validation/platform.go
```diff
@@ -1,20 +1,49 @@
...
+		}
+	}
+	for _, network := range n.ServiceNetwork {
+		if network.Contains(net.ParseIP(metadataServerIP)) {
+			allErrs = append(allErrs, field.Invalid(serviceNetworkPath, metadataServerIP, fmt.Errorf("the contains %s which is reserved for the metadata service", metadataServerIP).Error()))
```
patrickdillon, line 45:
```suggestion
			allErrs = append(allErrs, field.Invalid(serviceNetworkPath, metadataServerIP, fmt.Errorf("contains %s which is reserved for the metadata service", metadataServerIP).Error()))
```

patrickdillon (commented):
Typically we have tests for the install config and validation like this would have a test. You can use AWS as an example: https://github.com/openshift/installer/blob/master/pkg/types/aws/validation/platform_test.go

bd233:
> Typically we have tests for the install config and validation like this would have a test. You can use AWS as an example: https://github.com/openshift/installer/blob/master/pkg/types/aws/validation/platform_test.go

I have updated

kwoodson:
@patrickdillon Please TAL

PR #5362, opened 2021-11-09
azurestack: Fix cluster tag check for A record destroy

rna-afk:
The azurestack A record destroy was checking for cluster tag
with the cluster name and not the Infra ID which is the actual
name used for cluster tag population. Changing the name from
clusterName to InfraID to pick the right A records to destroy.

### DIFF
--- a/pkg/destroy/azure/azure.go
+++ b/pkg/destroy/azure/azure.go
@@ -39,7 +39,6 @@ type ClusterUninstaller struct {
 
 	InfraID                     string
 	ResourceGroupName           string
-	ClusterName                 string
 	BaseDomainResourceGroupName string
 
 	Logger logrus.FieldLogger
@@ -102,7 +101,6 @@ func New(logger logrus.FieldLogger, metadata *types.ClusterMetadata) (providers.
 		ResourceGroupName:           group,
 		Logger:                      logger,
 		BaseDomainResourceGroupName: metadata.Azure.BaseDomainResourceGroupName,
-		ClusterName:                 metadata.ClusterName,
 		CloudName:                   cloudName,
 	}, nil
 }
@@ -222,7 +220,6 @@ func deleteAzureStackPublicRecords(ctx context.Context, o *ClusterUninstaller) e
 
 	recordsClient := azurestackdns.NewRecordSetsClientWithBaseURI(o.Environment.ResourceManagerEndpoint, o.SubscriptionID)
 	recordsClient.Authorizer = o.Authorizer
-	clusterName := o.ClusterName
 
 	var errs []error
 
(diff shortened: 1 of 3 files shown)


pkg/destroy/azure/azure.go
```diff
@@ -222,7 +220,7 @@ func deleteAzureStackPublicRecords(ctx context.Context, o *ClusterUninstaller) e
 
 	recordsClient := azurestackdns.NewRecordSetsClientWithBaseURI(o.Environment.ResourceManagerEndpoint, o.SubscriptionID)
 	recordsClient.Authorizer = o.Authorizer
-	clusterName := o.ClusterName
+	clusterName := o.InfraID
```
patrickdillon, line 224: Should change the variable name now that this is no longer the cluster name. Not sure if we even need a variable? Could just use `o.InfraID` driectly.

rna-afk: Yeah would be better actually

staebler:
These changes seem reasonable to me. However, the e2e-azurestack job is broken. The name is being set in the install-config.yaml twice.
First,
https://github.com/openshift/release/blob/4be8a966864e97782fc766adf4a3fff70dfc576f/ci-operator/step-registry/ipi/conf/ipi-conf-commands.sh#L23-L24
Then,
https://github.com/openshift/release/blob/4be8a966864e97782fc766adf4a3fff70dfc576f/ci-operator/step-registry/ipi/conf/azurestack/ipi-conf-azurestack-commands.sh#L33-L34

We should run the e2e-azure job, too, for sanity since public azure uses the same destroyer.
/test e2e-azure

patrickdillon:
> These changes seem reasonable to me. However, the e2e-azurestack job is broken. The name is being set in the install-config.yaml twice. First, https://github.com/openshift/release/blob/4be8a966864e97782fc766adf4a3fff70dfc576f/ci-operator/step-registry/ipi/conf/ipi-conf-commands.sh#L23-L24 Then, https://github.com/openshift/release/blob/4be8a966864e97782fc766adf4a3fff70dfc576f/ci-operator/step-registry/ipi/conf/azurestack/ipi-conf-azurestack-commands.sh#L33-L34

@rna-afk has a fix in flight here: openshift/release#23453

staebler:
> > These changes seem reasonable to me. However, the e2e-azurestack job is broken. The name is being set in the install-config.yaml twice. First, https://github.com/openshift/release/blob/4be8a966864e97782fc766adf4a3fff70dfc576f/ci-operator/step-registry/ipi/conf/ipi-conf-commands.sh#L23-L24 Then, https://github.com/openshift/release/blob/4be8a966864e97782fc766adf4a3fff70dfc576f/ci-operator/step-registry/ipi/conf/azurestack/ipi-conf-azurestack-commands.sh#L33-L34
> 
> @rna-afk has a fix in flight here: [openshift/release#23453](https://github.com/openshift/release/pull/23453)

Of course he does! :)

staebler (approved):
/lgtm
/approve

PR #5375, opened 2021-11-11
Bug 1964380: azure upi: do not create etcd DNS records

MayXuQQ:
Applies to 4.6+ 
https://bugzilla.redhat.com/show_bug.cgi?id=1964380#c2


jianlinliu:
There is a dependency in https://github.com/openshift/release/blob/master/ci-operator/templates/openshift/installer/cluster-launch-installer-upi-e2e.yaml#L2101, so I think need a followup PR to amended it.

MayXuQQ:
/assign @fabianofranz  
/assign  @jhixson74 
https://github.com/openshift/release/pull/23567 depend on this PR

staebler (commented):
Please squash the commits into a single commit as they are small and highly dependent upon each other.
Please improve the commit message. A link to a BZ is not sufficient. Remove references to which versions the change applies to as the change only applies to the current version.

staebler:
/retitle BZ1964380: azure upi: do not create etcd DNS records

upi/azure/05_masters.json
```diff
@@ -32,6 +32,7 @@
     },
     "privateDNSZoneName" : {
       "type" : "string",
+      "defaultValue" : "privateDNSZone",
```
jianlinliu, line 35: I think empty string - "" should be better here.

PR #5376, opened 2021-11-11
vsphere: Use Managed Object ID for networks instead of potentially duplicate name.

bostrt:
vSphere Network names are not necessarily unique, especially in NSX-T environments so we must use the unique Managed Object ID for a network to fix installations in these type of environments. 

Related to [SPLAT-340](https://issues.redhat.com/browse/SPLAT-340)


jcpowermac:
This PR is also required for the region/zone, multiple datacenters/clusters support

pkg/tfvars/vsphere/vsphere.go
```diff
@@ -76,3 +94,39 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
...
 }
+
+func getNetworkMoID(vcenter, username, password, datacenter, cluster, network string) (string, error) {
+	client, _, err := vsphere.CreateVSphereClients(context.TODO(), vcenter, username, password)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to setup vSphere client")
+	}
+
+	finder := find.NewFinder(client)
+	path := fmt.Sprintf("/%s/host/*", datacenter)
```
jcpowermac, line 105: Wondering since we know the cluster if we just use it in the path?
And then remove the branch starting on line 113

bostrt: I think you are right. I believe we can get rid of the loop as well if we use `finder.ClusterComputeResource` instead of `List`.

jcpowermac:
@staebler @jstuever can you ptal, thanks!

data/data/vsphere/pre-bootstrap/outputs.tf
```diff
@@ -11,7 +11,7 @@ output "folder" {
 }
 
 output "network" {
```
staebler, line 13: This output should be removed. Later stages that were using this output should be changed to use `vsphere_network` instead.

Also, the description of the `vsphere_network` variable should be updated to reflect that it is the ID now rather than the name.

pkg/tfvars/vsphere/vsphere.go
```diff
@@ -55,6 +60,19 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
...
 
+	// Must use the Managed Object ID for a port group (e.g. dvportgroup-5258)
+	// instead of the name since port group names aren't always unique in vSphere.
+	// https://bugzilla.redhat.com/show_bug.cgi?id=1918005
+	networkID, err := getNetworkMoID(controlPlaneConfig.Workspace.Server,
+		sources.Username,
+		sources.Password,
+		controlPlaneConfig.Workspace.Datacenter,
+		sources.Cluster,
+		controlPlaneConfig.Network.Devices[0].NetworkName)
```
staebler, line 71: Are there any corresponding changes to the machine-api-operator's vSphere provider to ensure that workers use the correct network as well?

pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -109,7 +109,7 @@ func validateNetwork(finder Finder, p *vsphere.Platform, fldPath *field.Path) fi
 		return field.ErrorList{field.Invalid(fldPath, p.Datacenter, err.Error())}
 	}
 	networkPath := fmt.Sprintf("%s/network/%s", dataCenter.InventoryPath, p.Network)
-	_, err = finder.Network(ctx, networkPath)
+	_, err = finder.NetworkList(ctx, networkPath)
```
staebler, line 112: Should we do the same method of finding the network that we are doing in `getNetworkMoID` here to validate that the network exists in the specified cluster? Otherwise, this validation may give a false positive.

pkg/tfvars/vsphere/vsphere.go
```diff
@@ -1,7 +1,12 @@
 package vsphere
 
 import (
+	"context"
 	"encoding/json"
+	"fmt"
+	"github.com/vmware/govmomi/find"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/mo"
```
staebler, line 9: These should be in the next section of imports.

pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go
```diff
@@ -157,16 +157,12 @@ func findImportOvaParams(client *vim25.Client, datacenter, cluster, datastore, n
 
 	// Find the network object using the provided network name
 	for _, networkMoRef := range ccrMo.Network {
```
staebler, line 159: Do we have to do any of the lookup for the network any more? Don't we know that `networkMoRef` already?
```
networkMoRef := &ManagedObjectReference{Type: "Network", Value: network}
```

pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go
```diff
@@ -157,16 +157,12 @@ func findImportOvaParams(client *vim25.Client, datacenter, cluster, datastore, n
 
 	// Find the network object using the provided network name
```
staebler, line 158: A couple comments are left over referring to finding by network name, which is not being done any more.

staebler (commented):
@jstuever I would love your opinion on these changes, since you've spent a lot of time looking at the BZ.

pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go
jcpowermac: Trying this out since DistributedVirtualPortgroup and OpaqueNetwork extend Network. Otherwise the problem is we don't know the type

pkg/tfvars/vsphere/vsphere.go
bostrt: Yes, SPLAT-345 is where we will track those changes. I'll cross link in this PR too.

jstuever:
> @jstuever I would love your opinion on these changes, since you've spent a lot of time looking at the BZ.

I'm not convinced this resolves the underlying issue for the associated BZ. Maybe I'm missing something here. The root cause was that the survey in `pkg/asset/installconfig/vsphere/vsphere.go` returned the names of the network(s) without regard to their path. As a result, if there were two networks with the same name (one nested in a subfolder) the installer would only capture the un-pathed name as part of the survey. Then when it came time to consume that name it would select the FIRST network that matched.

Getting the path into the results of the survey was the easy part... we did that with other components in the same file. Using that path was more difficult because 1) remaining backward compatible, while 2) ensuring it is properly consumed by all of the dependencies (such as operators). Passing the ID for the matched network is probably a good idea, but it doesn't matter if we are still choosing the wrong network to begin with. Passing a fully pathed value is probably better.

I believe fixing the survey to output the path, and then ensuring that works with all of the dependencies is required to fully satisfy this BZ.

pkg/tfvars/vsphere/vsphere.go
```diff
@@ -94,4 +94,4 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
 	}
 
 	return json.MarshalIndent(cfg, "", "  ")
-}
+}
```
bostrt, line 97: I'll squash before merge to get rid of this.

data/data/vsphere/pre-bootstrap/outputs.tf
bostrt: Thanks, both points addressed in f10b387455edeb5314646abfe5f8231b8fa34e74

pkg/asset/installconfig/vsphere/validation.go
bostrt: This has been addressed in f10b387455edeb5314646abfe5f8231b8fa34e74 and required some rework to get unit tests (w/ mock) to run properly.

bostrt:
> I'm not convinced this resolves the underlying issue for the associated BZ. Maybe I'm missing something here. The root cause was that the survey in `pkg/asset/installconfig/vsphere/vsphere.go` returned the names of the network(s) without regard to their path. As a result, if there were two networks with the same name (one nested in a subfolder) the installer would only capture the un-pathed name as part of the survey. Then when it came time to consume that name it would select the FIRST network that matched.
> 
> Getting the path into the results of the survey was the easy part... we did that with other components in the same file. Using that path was more difficult because 1) remaining backward compatible, while 2) ensuring it is properly consumed by all of the dependencies (such as operators). Passing the ID for the matched network is probably a good idea, but it doesn't matter if we are still choosing the wrong network to begin with. Passing a fully pathed value is probably better.
> 
> I believe fixing the survey to output the path, and then ensuring that works with all of the dependencies is required to fully satisfy this BZ.

@jstuever Unfortunately, there's no unique network path that could be used; leaving the Manged Object ID as the only unique identifier for vSphere networks. I've updated this PR so the installer correctly prompt users with a list of networks that are filtered on the Datacenter and Cluster the user specifies _before_ the network prompt.

bostrt:
Sorting out lots of rebase issues that just came up.

jcpowermac:
e2e-vsphere test failures unrelated to this PR (vcenter sessions overrun issue - 503s)

jcpowermac:
Failed on destroy with a 503 error caused by: Bug 2022824 
Unrelated to this PR

bostrt:
Found a problem in similar MAPI changes and need to verify this PR again.

/hold

bostrt:
/hold

Hopefully we can move forward with https://github.com/openshift/installer/pull/5673 instead of this PR.

PR #5280, opened 2021-10-08
Bug 2004596: bump RHCOS 4.10 boot image metadata

miabbott:
This updates the RHCOS 4.10 boot image metadata in the installer. This
change includes fixes in the boot media for the following BZs:
    
2002215 - Multipath day1 not working on s390x
2004391 - RHCOS-4.9 failed to boot in FIPS mode on s390x
2004449 - Boot option recovery menu prevents image boot
2006690 - OS boot failure "x64 Exception Type 06 - Invalid Opcode Exception"
2007085 - Intermittent failure mounting /run/media/iso when booting live ISO from USB stick
2011956 - [tracker] Kubelet rejects pods that use resources that should be freed by completed pods
2011960 - [tracker] Storage operator is not available after reboot cluster instances
    
Changes generated with:
    
```
./hack/update-rhcos-bootimage.py https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos-4.10-aarch64/410.84.202110141117-0/aarch64/meta.json aarch64
./hack/update-rhcos-bootimage.py https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos-4.10-ppc64le/410.84.202110141003-0/ppc64le/meta.json ppc64le
./hack/update-rhcos-bootimage.py https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos-4.10-s390x/410.84.202110141003-0/s390x/meta.json s390x
./hack/update-rhcos-bootimage.py https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos-4.10/410.84.202110140201-0/x86_64/meta.json amd64
plume cosa2stream --target data/data/rhcos-stream.json --distro rhcos --url https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases aarch64=410.84.202110141117-0 ppc64le=410.84.202110141003-0 s390x=410.84.202110141003-0 x86_64=410.84.202110140201-0
```
    
**Verification Steps:**
    
1. Install a new 4.10 cluster
2. `oc debug node/<node name> -- chroot /host rpm-ostree status`
3. Verify that the deployment version matches the version from this PR
that matches the architecture you are testing on. (i.e. `aarch64`
should have version `410.84.202110141117-0`)

PR #5377, opened 2021-11-11
Bug 2016008: aws: add newly supported regions for arm64 

AnnaZivkovic:
This PR enables 3 regions for arm64:
eu-north-1
eu-west-3
me-south-1

m6g instances are now available in these regions: 
https://aws.amazon.com/about-aws/whats-new/2021/02/aws-graviton2-based-m6g-c6g-r6g-instances-available-additional-regions/

The boot images for all architectures have been bumped to accommodate this change


miabbott:
It looks like this is also including PowerVS regions; is that intended?

Could you amend the commit message to include the bootimage bump BZ for 4.10 - https://bugzilla.redhat.com/show_bug.cgi?id=2016008

Also, please note in the commit message the dependent BZs on the bootimage bump which are fixed/delivered via this change.

For example - https://github.com/openshift/installer/pull/5280

The changes to the metadata look sane, but I would like to have a good paper trail via the commit messages.

Prashanth684:
> It looks like this is also including PowerVS regions; is that intended?
> 
> Could you amend the commit message to include the bootimage bump BZ for 4.10 - https://bugzilla.redhat.com/show_bug.cgi?id=2016008
> 
> Also, please note in the commit message the dependent BZs on the bootimage bump which are fixed/delivered via this change.
> 
> For example - #5280
> 
> The changes to the metadata look sane, but I would like to have a good paper trail via the commit messages.

tagged Bug 2016008 for the bootimage bump and listed the associated dependent BZs. The coreos-assembler now has support for parsing IBMCloud and PowerVS metadata (https://github.com/coreos/coreos-assembler/pull/2535) and since we have PowerVS ova images being uploaded to IBMCloud, it gets included when we run the cosa2stream command, so it is intended. In any case nothing is being done with it for now until the support for the platform itself gets in.

miabbott (commented):
/approve

PR #5313, opened 2021-10-19
Bug 1917893: bump oVirt terraform provider version which fix "Disk is locked" bug

Gal-Zaidman:
On RHV 4.4.7 the image upload process changed which lead to a bug that affected many customers.
This change bumps the oVirt terraform version to a version that contains the fix for the Bug.

PR #5379, opened 2021-11-12
Alibaba: support existing VPC, VSwitchs and PrivateZone

bd233:
This PR is based on #5378
If the user wants the Installer to install the cluster in an existing network, it is now supported.

### DIFF
--- a/data/data/alibabacloud/cluster/dns/privatezone.tf
+++ b/data/data/alibabacloud/cluster/dns/privatezone.tf
@@ -1,7 +1,8 @@
 locals {
-  description  = "Created By OpenShift Installer"
-  prefix       = var.cluster_id
-  cluster_name = split(".", var.cluster_domain)[0]
+  description     = "Created By OpenShift Installer"
+  prefix          = var.cluster_id
+  cluster_name    = split(".", var.cluster_domain)[0]
+  private_zone_id = var.private_zone_id == "" ? alicloud_pvtz_zone.pvtz.0.id : var.private_zone_id
 }
 
 // Using this data source can open Private Zone service automatically.
@@ -24,17 +25,20 @@ resource "alicloud_alidns_record" "dns_public_record" {
 }
 
 resource "alicloud_pvtz_zone" "pvtz" {
+  count = var.private_zone_id == "" ? 1 : 0
+
   resource_group_id = var.resource_group_id
   zone_name         = var.cluster_domain
 }
 
 resource "alicloud_pvtz_zone_attachment" "pvtz_attachment" {
-  zone_id = alicloud_pvtz_zone.pvtz.id
+  count   = var.private_zone_id == "" ? 1 : 0
+  zone_id = local.private_zone_id
   vpc_ids = [var.vpc_id]
 }
 
 resource "alicloud_pvtz_zone_record" "pvtz_record_api_int" {
-  zone_id = alicloud_pvtz_zone.pvtz.id
+  zone_id = local.private_zone_id
   type    = "A"
   rr      = "api-int"
   value   = var.slb_internal_ip
@@ -42,7 +46,7 @@ resource "alicloud_pvtz_zone_record" "pvtz_record_api_int" {
 }
 
 resource "alicloud_pvtz_zone_record" "pvtz_record_api" {
-  zone_id = alicloud_pvtz_zone.pvtz.id
+  zone_id = local.private_zone_id
   type    = "A"
   rr      = "api"
   value   = var.slb_internal_ip
--- a/data/data/alibabacloud/cluster/master/variables.tf
+++ b/data/data/alibabacloud/cluster/master/variables.tf
@@ -11,9 +11,14 @@ variable "vpc_id" {
   description = "The VPC ID of the master ECS."
 }
 
-variable "vswitch_ids" {
+variable "zone_ids" {
   type        = list(string)
-  description = "The VSwitch IDs of the master ECS. Example: [vsw-xxx1, vsw-xxx2, vsw-xxx3]"
+  description = "The availability zones in which to create the masters. The length of this list must match instance_count."
+}
+
+variable "az_to_vswitch_id" {
+  type        = map(string)
+  description = "Map from availability zone ID to the ID of the VSwitch in that availability zone"
 }
 
 variable "sg_id" {
@@ -23,6 +28,10 @@ variable "sg_id" {
 
 variable "slb_ids" {
   type = list(string)
+}
+
+variable "instance_count" {
+  type = string
 }
 
 variable "instance_type" {
--- a/data/data/alibabacloud/cluster/vpc/vpc.tf
+++ b/data/data/alibabacloud/cluster/vpc/vpc.tf
@@ -3,9 +3,17 @@ locals {
   description = "Created By OpenShift Installer"
   prefix      = var.cluster_id
   newbits     = max(4, 20 - tonumber(split("/", var.vpc_cidr_block)[1]))
+  vpc_id      = var.vpc_id == "" ? alicloud_vpc.vpc.0.id : var.vpc_id
+  vswitch_ids = length(var.vswitch_ids) == 0 ? alicloud_vswitch.vswitches.*.id : var.vswitch_ids
+}
+
+data "alicloud_vswitches" "vswitches" {
+  ids = local.vswitch_ids
 }
 
 resource "alicloud_vpc" "vpc" {
+  count = var.vpc_id == "" ? 1 : 0
+
   resource_group_id = var.resource_group_id
   vpc_name          = "${local.prefix}-vpc"
   cidr_block        = var.vpc_cidr_block
@@ -18,12 +26,12 @@ resource "alicloud_vpc" "vpc" {
   )
 }
 
-resource "alicloud_vswitch" "vswitchs" {
-  count = length(var.zone_ids)
+resource "alicloud_vswitch" "vswitches" {
+  count = length(var.vswitch_ids) == 0 ? length(var.zone_ids) : 0
 
   vswitch_name = "${local.prefix}-vswitch-${var.zone_ids[count.index]}"
   description  = local.description
-  vpc_id       = alicloud_vpc.vpc.id
+  vpc_id       = local.vpc_id
   cidr_block   = cidrsubnet(var.vpc_cidr_block, local.newbits, count.index)
   zone_id      = var.zone_ids[count.index]
   tags = merge(
@@ -37,7 +45,7 @@ resource "alicloud_vswitch" "vswitchs" {
 resource "alicloud_vswitch" "vswitch_nat_gateway" {
   vswitch_name = "${local.prefix}-vswitch-nat-gateway"
   description  = local.description
-  vpc_id       = alicloud_vpc.vpc.id
+  vpc_id       = local.vpc_id
   cidr_block   = cidrsubnet(var.vpc_cidr_block, local.newbits, local.newbits)
   zone_id      = var.nat_gateway_zone_id
   tags = merge(
--- a/pkg/asset/installconfig/alibabacloud/client.go
+++ b/pkg/asset/installconfig/alibabacloud/client.go
@@ -246,13 +246,51 @@ func (client *Client) ListDNSDomain(baseDomain string) (response *alidns.Describ
 	return
 }
 
+// ListPrivateZonesByName gets the list of privatezones with the specified name.
+func (client *Client) ListPrivateZonesByName(zoneName string) (response *pvtz.DescribeZonesResponse, err error) {
+	return client.ListPrivateZones(zoneName, "")
+}
+
+// ListPrivateZonesByVPC gets the list of privatezones attached to the specified VPC.
+func (client *Client) ListPrivateZonesByVPC(queryVpcID string) (response *pvtz.DescribeZonesResponse, err error) {
+	return client.ListPrivateZones("", queryVpcID)
+}
+
 // ListPrivateZones gets the list of privatzones.
-func (client *Client) ListPrivateZones(zoneName string) (response *pvtz.DescribeZonesResponse, err error) {
+func (client *Client) ListPrivateZones(zoneName string, queryVpcID string) (response *pvtz.DescribeZonesResponse, err error) {
 	request := pvtz.CreateDescribeZonesRequest()
 	request.Lang = "en"
-	request.Keyword = zoneName
 	request.SearchMode = "EXACT"
+	request.Keyword = zoneName
+	request.QueryVpcId = queryVpcID
+
 	response = &pvtz.DescribeZonesResponse{
+		BaseResponse: &responses.BaseResponse{},
+	}
+	err = client.doActionWithSetDomain(request, response)
+	return
+}
+
+// ListVpcs gets the list of VPCs.
+func (client *Client) ListVpcs(vpcID string) (response *vpc.DescribeVpcsResponse, err error) {
+	request := vpc.CreateDescribeVpcsRequest()
+	request.RegionId = client.RegionID
+	request.VpcId = vpcID
+
+	response = &vpc.DescribeVpcsResponse{
+		BaseResponse: &responses.BaseResponse{},
+	}
+	err = client.doActionWithSetDomain(request, response)
+	return
+}
+
+// ListVSwitches gets the list of VSwitches.
+func (client *Client) ListVSwitches(vswitchID string) (response *vpc.DescribeVSwitchesResponse, err error) {
+	request := vpc.CreateDescribeVSwitchesRequest()
+	request.RegionId = client.RegionID
+	request.VSwitchId = vswitchID
+
+	response = &vpc.DescribeVSwitchesResponse{
 		BaseResponse: &responses.BaseResponse{},
 	}
 	err = client.doActionWithSetDomain(request, response)
--- a/pkg/asset/installconfig/alibabacloud/validation.go
+++ b/pkg/asset/installconfig/alibabacloud/validation.go
@@ -109,7 +109,18 @@ func validateInstanceType(client *Client, zones []string, instanceType string, f
 
 func validatePlatform(client *Client, ic *types.InstallConfig, path *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
-	allErrs = append(allErrs, validateResourceGroup(client, ic, path)...)
+	if ic.AlibabaCloud.ResourceGroupID != "" {
+		allErrs = append(allErrs, validateResourceGroup(client, ic, path)...)
+	}
+	if ic.AlibabaCloud.VpcID != "" {
+		allErrs = append(allErrs, validateVpc(client, ic, path)...)
+	}
+	if len(ic.AlibabaCloud.VSwitchIDs) != 0 {
+		allErrs = append(allErrs, validateVSwitches(client, ic, path)...)
+	}
+	if ic.AlibabaCloud.PrivateZoneID != "" {
+		allErrs = append(allErrs, validatePrivateZoneID(client, ic, path)...)
+	}
 	return allErrs
 }
 
@@ -127,6 +138,73 @@ func validateResourceGroup(client *Client, ic *types.InstallConfig, path *field.
 	return append(allErrs, field.NotFound(path.Child("resourceGroupID"), ic.AlibabaCloud.ResourceGroupID))
 }
 
+func validateVpc(client *Client, ic *types.InstallConfig, path *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	vpcs, err := client.ListVpcs(ic.AlibabaCloud.VpcID)
+	if err != nil {
+		return append(allErrs, field.InternalError(path.Child("vpcID"), err))
+	}
+	if vpcs.TotalCount == 0 {
+		allErrs = append(allErrs, field.NotFound(path.Child("vpcID"), ic.AlibabaCloud.VpcID))
+	}
+	return allErrs
+}
+
+func validateVSwitches(client *Client, ic *types.InstallConfig, path *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	zoneIDs := map[string]bool{}
+	for idx, id := range ic.AlibabaCloud.VSwitchIDs {
+		vswitches, err := client.ListVSwitches(id)
+		if err != nil {
+			return append(allErrs, field.InternalError(path.Child("vswitchIDs"), err))
+		}
+		if vswitches.TotalCount != 1 {
+			allErrs = append(allErrs, field.NotFound(path.Child("vswitchIDs").Index(idx), id))
+			continue
+		}
+
+		vswitch := vswitches.VSwitches.VSwitch[0]
+		if vswitch.VpcId != ic.AlibabaCloud.VpcID {
+			allErrs = append(allErrs, field.Invalid(path.Child("vswitchIDs").Index(idx), id, fmt.Sprintf("the VSwitch does not belong to vpc %s", ic.AlibabaCloud.VpcID)))
+		}
+
+		if zoneIDs[vswitch.ZoneId] {
+			allErrs = append(allErrs, field.Invalid(path.Child("vswitchIDs").Index(idx), id, "the availability zone of the VSwitch overlapped with other VSwitches"))
+		} else {
+			zoneIDs[vswitch.ZoneId] = true
+		}
+	}
+
+	return allErrs
+}
+
+func validatePrivateZoneID(client *Client, ic *types.InstallConfig, path *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	fldpath := path.Child("privateZoneID")
+	zoneName := ic.ClusterDomain()
+	isExist := false
+
+	zones, err := client.ListPrivateZonesByVPC(ic.AlibabaCloud.VpcID)
+	if err != nil {
+		return append(allErrs, field.InternalError(fldpath, err))
+	}
+
+	for _, zone := range zones.Zones.Zone {
+		if zone.ZoneId == ic.AlibabaCloud.PrivateZoneID {
+			isExist = true
+			if zone.ZoneName != zoneName {
+				allErrs = append(allErrs, field.Invalid(fldpath, ic.AlibabaCloud.PrivateZoneID, fmt.Sprintf("the name %s of the existing private zone does not match the expected zone name %s", zone.ZoneName, zoneName)))
+			}
+			break
+		}
+	}
+	if !isExist {
+		return append(allErrs, field.Invalid(fldpath, ic.AlibabaCloud.PrivateZoneID, fmt.Sprintf("the private zone is not found, or not associated with the VPC %s", ic.AlibabaCloud.VpcID)))
+	}
+	return allErrs
+}
+
 // ValidateForProvisioning validates if the install config is valid for provisioning the cluster.
 func ValidateForProvisioning(client *Client, ic *types.InstallConfig, metadata *Metadata) error {
 	allErrs := field.ErrorList{}
@@ -136,15 +214,21 @@ func ValidateForProvisioning(client *Client, ic *types.InstallConfig, metadata *
 
 func validateClusterName(client *Client, ic *types.InstallConfig) field.ErrorList {
 	allErrs := field.ErrorList{}
-	namePath := field.NewPath("metadata").Child("name")
+	if ic.AlibabaCloud.PrivateZoneID != "" {
+		return allErrs
+	}
 
+	namePath := field.NewPath("metadata").Child("name")
 	zoneName := ic.ClusterDomain()
-	response, err := client.ListPrivateZones(zoneName)
+	response, err := client.ListPrivateZonesByName(zoneName)
 	if err != nil {
 		allErrs = append(allErrs, field.InternalError(namePath, err))
 	}
-	if response.TotalItems > 0 {
-		allErrs = append(allErrs, field.Invalid(namePath, ic.ObjectMeta.Name, fmt.Sprintf("cluster name is unavailable, private zone name %s already exists", zoneName)))
+	for _, zone := range response.Zones.Zone {
+		if zone.ZoneName == zoneName {
+			allErrs = append(allErrs, field.Invalid(namePath, ic.ObjectMeta.Name, fmt.Sprintf("cluster name is unavailable, private zone name %s already exists", zoneName)))
+			break
+		}
 	}
 	return allErrs
 }
--- a/pkg/asset/machines/alibabacloud/machines.go
+++ b/pkg/asset/machines/alibabacloud/machines.go
@@ -17,7 +17,7 @@ import (
 )
 
 // Machines returns a list of machines for a machinepool.
-func Machines(clusterID string, config *types.InstallConfig, pool *types.MachinePool, role, userDataSecret string, resourceTags map[string]string) ([]machineapi.Machine, error) {
+func Machines(clusterID string, config *types.InstallConfig, pool *types.MachinePool, role, userDataSecret string, resourceTags map[string]string, vswitchMaps map[string]string) ([]machineapi.Machine, error) {
 	if configPlatform := config.Platform.Name(); configPlatform != alibabacloud.Name {
 		return nil, fmt.Errorf("non-AlibabaCloud configuration: %q", configPlatform)
 	}
@@ -35,8 +35,9 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 
 	var machines []machineapi.Machine
 	for idx := int64(0); idx < total; idx++ {
-		azIndex := int(idx) % len(azs)
-		provider, err := provider(clusterID, platform, mpool, azIndex, role, userDataSecret, resourceTags)
+		zoneID := azs[int(idx)%len(azs)]
+		vswitchID := vswitchMaps[zoneID]
+		provider, err := provider(clusterID, platform, mpool, zoneID, role, userDataSecret, resourceTags, vswitchID)
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to create provider")
 		}
@@ -69,12 +70,12 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 func provider(clusterID string,
 	platform *alibabacloud.Platform,
 	mpool *alibabacloud.MachinePool,
-	azIdx int,
+	zoneID string,
 	role string,
 	userDataSecret string,
 	resourceTags map[string]string,
+	vswitchID string,
 ) (*alibabacloudprovider.AlibabaCloudMachineProviderConfig, error) {
-	az := mpool.Zones[azIdx]
 	tags, err := tagsFromResourceTags(clusterID, resourceTags)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to create alibabacloudprovider.Tag from Tags")
@@ -87,13 +88,7 @@ func provider(clusterID string,
 			}),
 		},
 	}
-	vSwitchResourceRef := alibabacloudprovider.AlibabaResourceReference{
-		Tags: append(tags, alibabacloudprovider.Tag{
-			Key:   "Name",
-			Value: fmt.Sprintf("%s-vswitch-%s", clusterID, az),
-		}),
-	}
-	return &alibabacloudprovider.AlibabaCloudMachineProviderConfig{
+	config := &alibabacloudprovider.AlibabaCloudMachineProviderConfig{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "alibabacloudmachineproviderconfig.openshift.io/v1beta1",
 			Kind:       "AlibabaCloudMachineProviderConfig",
@@ -106,13 +101,21 @@ func provider(clusterID string,
 		},
 		RegionID:          platform.Region,
 		ResourceGroupID:   platform.ResourceGroupID,
-		ZoneID:            az,
+		ZoneID:            zoneID,
 		UserDataSecret:    &corev1.LocalObjectReference{Name: userDataSecret},
 		CredentialsSecret: &corev1.LocalObjectReference{Name: "alibabacloud-credentials"},
 		Tags:              tags,
+		RAMRoleName:       fmt.Sprintf("%s-role-%s", clusterID, role),
 		SecurityGroups:    sgResourceRef,
-		VSwitch:           vSwitchResourceRef,
-	}, nil
+	}
+	config.VSwitch = alibabacloudprovider.AlibabaResourceReference{
+		ID: vswitchID,
+		Tags: append(tags, alibabacloudprovider.Tag{
+			Key:   "Name",
+			Value: fmt.Sprintf("%s-vswitch-%s", clusterID, zoneID),
+		}),
+	}
+	return config, nil
 }
 
 func tagsFromResourceTags(clusterID string, resourceTags map[string]string) ([]alibabacloudprovider.Tag, error) {
--- a/pkg/asset/machines/alibabacloud/machinesets.go
+++ b/pkg/asset/machines/alibabacloud/machinesets.go
@@ -14,7 +14,7 @@ import (
 )
 
 // MachineSets returns a list of machinesets for a machinepool.
-func MachineSets(clusterID string, config *types.InstallConfig, pool *types.MachinePool, role, userDataSecret string, resourceTags map[string]string) ([]*machineapi.MachineSet, error) {
+func MachineSets(clusterID string, config *types.InstallConfig, pool *types.MachinePool, role, userDataSecret string, resourceTags map[string]string, vswitchMaps map[string]string) ([]*machineapi.MachineSet, error) {
 	if configPlatform := config.Platform.Name(); configPlatform != alibabacloud.Name {
 		return nil, fmt.Errorf("non-AlibabaCloud configuration: %q", configPlatform)
 	}
@@ -37,7 +37,11 @@ func MachineSets(clusterID string, config *types.InstallConfig, pool *types.Mach
 			replicas++
 		}
 
-		provider, err := provider(clusterID, platform, mpool, idx, role, userDataSecret, resourceTags)
+		vswitchID, ok := vswitchMaps[az]
+		if len(vswitchMaps) > 0 && !ok {
+			return nil, errors.Errorf("no VSwitch for zone %s", az)
+		}
+		provider, err := provider(clusterID, platform, mpool, az, role, userDataSecret, resourceTags, vswitchID)
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to create provider")
 		}
--- a/pkg/asset/machines/master.go
+++ b/pkg/asset/machines/master.go
@@ -168,19 +168,30 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 		if err != nil {
 			return err
 		}
+		vswitchMaps, err := installConfig.AlibabaCloud.VSwitchMaps()
+		if err != nil {
+			return errors.Wrap(err, "failed to get VSwitchs map")
+		}
 		mpool := alibabacloudtypes.DefaultMasterMachinePoolPlatform()
 		mpool.ImageID = string(*rhcosImage)
 		mpool.Set(ic.Platform.AlibabaCloud.DefaultMachinePlatform)
 		mpool.Set(pool.Platform.AlibabaCloud)
 		if len(mpool.Zones) == 0 {
-			azs, err := client.GetAvailableZonesByInstanceType(mpool.InstanceType)
-			if err != nil || len(azs) == 0 {
-				return errors.Wrap(err, "failed to fetch availability zones")
+			if len(vswitchMaps) > 0 {
+				for zone := range vswitchMaps {
+					mpool.Zones = append(mpool.Zones, zone)
+				}
+			} else {
+				azs, err := client.GetAvailableZonesByInstanceType(mpool.InstanceType)
+				if err != nil || len(azs) == 0 {
+					return errors.Wrap(err, "failed to fetch availability zones")
+				}
+				mpool.Zones = azs
 			}
-			mpool.Zones = azs
 		}
+
 		pool.Platform.AlibabaCloud = &mpool
-		machines, err = alibabacloud.Machines(clusterID.InfraID, ic, &pool, "master", "master-user-data", installConfig.Config.Platform.AlibabaCloud.Tags)
+		machines, err = alibabacloud.Machines(clusterID.InfraID, ic, &pool, "master", "master-user-data", installConfig.Config.Platform.AlibabaCloud.Tags, vswitchMaps)
 		if err != nil {
 			return errors.Wrap(err, "failed to create master machine objects")
 		}
--- a/pkg/types/alibabacloud/platform.go
+++ b/pkg/types/alibabacloud/platform.go
@@ -10,6 +10,26 @@ type Platform struct {
 	// +optional
 	ResourceGroupID string `json:"resourceGroupID,omitempty"`
 
+	// VpcID is the ID of an already existing VPC where the cluster should be installed.
+	// If empty, the installer will create a new VPC for the cluster.
+	// +optional
+	VpcID string `json:"vpcID,omitempty"`
+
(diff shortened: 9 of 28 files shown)


pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -100,6 +119,56 @@ func validateResourceGroup(client *Client, ic *types.InstallConfig, path *field.
...
+func validateVpc(client *Client, ic *types.InstallConfig, path *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	vpcs, err := client.ListVpcs(ic.AlibabaCloud.VpcID)
+	if err != nil {
+		return append(allErrs, field.InternalError(path.Child("vpcID"), err))
+	}
+	if vpcs.TotalCount == 1 {
+		return allErrs
+	}
+	return append(allErrs, field.NotFound(path.Child("vpcID"), ic.AlibabaCloud.VpcID))
```
staebler, line 131:
```suggestion
	if vpcs.TotalCount == 0 {
		allErrs = append(allErrs, field.NotFound(path.Child("vpcID"), ic.AlibabaCloud.VpcID))
	}
	return allErrs
```

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -100,6 +119,56 @@ func validateResourceGroup(client *Client, ic *types.InstallConfig, path *field.
...
+}
+
+func validateVSwitchs(client *Client, ic *types.InstallConfig, path *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if ic.AlibabaCloud.VpcID == "" {
+		return append(allErrs, field.Required(path.Child("vswitchIDs"), "Use the existing VSwitchs, vpcID must be specified"))
```
staebler, line 138:
```suggestion
		return append(allErrs, field.Required(path.Child("vswitchIDs"), "when using existing VSwitches, an existing VPC must be used"))
```

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -100,6 +119,56 @@ func validateResourceGroup(client *Client, ic *types.InstallConfig, path *field.
...
+	}
+
+	for _, id := range ic.AlibabaCloud.VSwitchIDs {
```
staebler, line 141: Use an index variable so that the paths used in the field errors can reference the index.
Example,
```
allErrs = append(allErrs, field.NotFound(path.Child("vswitchIDs").Index(i), id))
```

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -100,6 +119,56 @@ func validateResourceGroup(client *Client, ic *types.InstallConfig, path *field.
...
+		vswitchs, err := client.ListVSwitchs(id)
+		if err != nil {
+			return append(allErrs, field.InternalError(path.Child("vswitchIDs"), err))
+		}
+		if vswitchs.TotalCount != 1 {
+			allErrs = append(allErrs, field.NotFound(path.Child("vswitchIDs"), id))
+			continue
+		}
+		if vswitchs.VSwitches.VSwitch[0].VpcId != ic.AlibabaCloud.VpcID {
+			allErrs = append(allErrs, field.Invalid(path.Child("vswitchIDs"), id, fmt.Sprintf("The VSwitch %s does not belong to vpc %s", id, ic.AlibabaCloud.VpcID)))
```
staebler, line 151: Do not include the ID of the vswitch in the details. That is already included in the field error.

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -116,7 +185,7 @@ func validateClusterName(client *Client, ic *types.InstallConfig) field.ErrorLis
 	if err != nil {
 		allErrs = append(allErrs, field.InternalError(namePath, err))
 	}
-	if response.TotalItems > 0 {
+	if response.TotalItems > 0 && ic.AlibabaCloud.PrivateZoneID == "" {
```
staebler, line 188: Should we validate that the existing private zone has the expected name? If not, then `validateClusterName` should return immediately when using an existing private zone, since there is nothing to validate. There is no sense in listing the private zones if they are not going to be used.

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -73,12 +81,23 @@ func validateMachinePool(client *Client, ic *types.InstallConfig, fldPath *field
 			zones = zones[:3]
 		}
 
+		response, err := client.DescribeAvailableZoneByInstanceType(pool.InstanceType)
```
staebler, line 84: This change does not seem related to adding support for existing VPC, vSwitches, or private zones.

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -100,6 +119,56 @@ func validateResourceGroup(client *Client, ic *types.InstallConfig, path *field.
...
+	if err != nil {
+		return append(allErrs, field.InternalError(path.Child("vpcID"), err))
+	}
+	if vpcs.TotalCount == 1 {
+		return allErrs
+	}
+	return append(allErrs, field.NotFound(path.Child("vpcID"), ic.AlibabaCloud.VpcID))
+}
+
+func validateVSwitchs(client *Client, ic *types.InstallConfig, path *field.Path) field.ErrorList {
```
staebler, line 134: I think we also need validation that there is only a single vSwitch in each availability zone.

pkg/asset/machines/alibabacloud/machines.go
```diff
@@ -35,8 +35,12 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 
 	var machines []machineapi.Machine
 	for idx := int64(0); idx < total; idx++ {
-		azIndex := int(idx) % len(azs)
-		provider, err := provider(clusterID, platform, mpool, azIndex, role, userDataSecret, resourceTags)
+		zoneID := azs[int(idx)%len(azs)]
+		vswitchID, ok := vswitchMaps[zoneID]
+		if len(vswitchMaps) > 0 && !ok {
+			return nil, errors.Errorf("no VSwitch for zone %s", zoneID)
```
staebler, line 41: This check is fine to leave here, but the user should never encounter this error. This must be caught by earlier validation.

pkg/types/alibabacloud/platform.go
```diff
@@ -12,6 +12,27 @@ type Platform struct {
 	// +optional
 	ResourceGroupID string `json:"resourceGroupID"`
 
+	// VpcID is the ID of an already existing VPC where the cluster should be installed.
+	// If empty, a new VPC will created for the cluster.
+	// Destroying the cluster using installer will delete this VPC.
```
staebler, line 17: Why would the cluster destroy delete the VPC? Is that because the VPC must be in the existing resource group? If so, then we are missing some validation. If not, then I don't think that the cluster destroy should delete the existing VPC.

pkg/types/alibabacloud/platform.go
```diff
@@ -12,6 +12,27 @@ type Platform struct {
...
+	// VSwitchIDs is the ID list of already existing VSwitchs where the master should be created.
+	// If empty, the new VSwitchs will created for the cluster.
+	// Destroying the cluster using installer will delete these VSwitchs.
+	// +optional
+	VSwitchIDs []string `json:"vswitchIDs,omitempty"`
+
+	// PrivateZoneID is the ID of an existing private zone into which to add DNS
+	// records for the cluster's internal API. An existing private zone can
+	// only be used when also using existing vpc. The private zone must be
+	// associated with the VPC containing the subnets.
```
staebler, line 30: Validation for this is missing.

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -100,6 +119,56 @@ func validateResourceGroup(client *Client, ic *types.InstallConfig, path *field.
...
+	allErrs := field.ErrorList{}
+
+	if ic.AlibabaCloud.VpcID == "" {
```
staebler, line 137: This should be checked via static validation in pkg/types/alibabacloud/validation/platform.go.

pkg/types/alibabacloud/platform.go
```diff
@@ -12,6 +12,27 @@ type Platform struct {
 	// +optional
 	ResourceGroupID string `json:"resourceGroupID"`
 
+	// VpcID is the ID of an already existing VPC where the cluster should be installed.
+	// If empty, a new VPC will created for the cluster.
+	// Destroying the cluster using installer will delete this VPC.
+	// +optional
+	VpcID string `json:"vpcID,omitempty"`
+
+	// VSwitchIDs is the ID list of already existing VSwitchs where the master should be created.
```
staebler, line 21: Globally, replace VSwitchs with VSwitches.
```suggestion
	// VSwitchIDs is the ID list of already existing VSwitches where the masters should be created.
```

staebler, line 21: The vswitches are not just for the masters.

staebler (commented):
In the future, please refrain from adding so many separate parts in one commit.
For example, it would have been better to have one commit each for the existing VPC, the existing VSwitches, and the existing private zones. They could also be done as separate PRs, which would make it much faster to get approval.

pkg/asset/installconfig/alibabacloud/alibabacloud.go
```diff
@@ -126,6 +129,7 @@ func selectResourceGroup(client *Client) (string, error) {
 				Message: "Resource Group",
 				Help:    "The resource group where the cluster will be provisioned.",
 				Options: options,
+				Default: createNew,
```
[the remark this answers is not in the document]
staebler: Moving comment to #5378.

pkg/types/alibabacloud/platform.go
```diff
@@ -5,8 +5,11 @@ type Platform struct {
...
 	Region string `json:"region"`
 
-	// ResourceGroupID is the ID of an already existing resource group where the
-	// cluster should be installed.
+	// ResourceGroupID is the ID of an already existing resource group where the cluster should be installed.
+	// This resource group must be empty with no other resources when trying to use it for creating a cluster.
+	// If empty, a new resource group will created for the cluster.
+	// Destroying the cluster using installer will delete this resource group.
+	// +optional
 	ResourceGroupID string `json:"resourceGroupID"`
```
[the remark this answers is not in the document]
staebler: Moving comment to #5378.

data/data/alibabacloud/cluster/outputs.tf
```diff
@@ -1,3 +1,6 @@
+output "resource_group_id" {
```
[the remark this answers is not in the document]
staebler: Moving comment to #5378.

pkg/asset/installconfig/alibabacloud/validation.go
bd233: Sorry, I don’t understand why there can only be one switch in each zone. Can you explain more details?

pkg/asset/machines/alibabacloud/machines.go
bd233: You are right, I removed it

pkg/types/alibabacloud/platform.go
bd233: I think there is no clear description here. If VpcID is empty, a new VPC will be created, and released when the cluster is destroyed.

bd233:
@staebler I have updated, please continue to review

pkg/asset/installconfig/alibabacloud/validation.go
staebler: You are creating snat entries based on the number of zones. If there are more vswitches than zones, then some vswitches will not have snat entries created for them.
https://github.com/openshift/installer/blob/8be8890de7f0d634c1defa19920f66eab5904821/data/data/alibabacloud/cluster/vpc/nat_gateway.tf#L17-L24

When creating the machine and machineset manifests, a map is built mapping a zone to a vswitch. If there are multiple vswitches for the same zone, then the map will only contain a mapping to the last vswitch for the zone.
https://github.com/openshift/installer/blob/8be8890de7f0d634c1defa19920f66eab5904821/pkg/asset/machines/master.go#L173-L179

But, then you are creating masters based on the number of vswitches. That implies that the number of vswitches must match the number of control-plane replicas.
https://github.com/openshift/installer/blob/8be8890de7f0d634c1defa19920f66eab5904821/data/data/alibabacloud/cluster/master/main.tf#L11

pkg/types/alibabacloud/platform.go
staebler: My suggestion is to remove the language around deleting the VPC.
```
	// VpcID is the ID of the VPC into which the cluster's machines will be created.
	// Leave the VPC ID unset to have the installer create a VPC on your behalf.
```

pkg/asset/installconfig/alibabacloud/validation.go
bd233: Thank you very much. I realized the problem. I have done the following:
1.  Use master_count as the number of master instances (I think this is the correct way)
2.  Add to verify whether there are VSwitches in the same availability zone and whether there are duplicate VSwitches

data/data/alibabacloud/cluster/vpc/vpc.tf
```diff
@@ -19,12 +27,12 @@ resource "alicloud_vpc" "vpc" {
   )
 }
 
-resource "alicloud_vswitch" "vswitchs" {
-  count = length(var.zone_ids)
+resource "alicloud_vswitch" "vswitches" {
+  count = length(var.vswitch_ids) == 0 ? length(var.zone_ids) : 0
```
staebler, line 31: This count is not correct when the installer is creating the vSwitches. Using `length(var.zone_ids)` assumes that the zones in `var.zone_ids` are unique. If the master replica count is greater than the number of availability zones used, then this will create multiple vSwitches in some of the zones, and those extras will be omitted from the mapping from zone ID to vSwitch ID.

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -36,6 +36,15 @@ func validatePlatform(client *Client, ic *types.InstallConfig, path *field.Path)
 	if ic.AlibabaCloud.ResourceGroupID != "" {
 		allErrs = append(allErrs, validateResourceGroup(client, ic, path)...)
 	}
+	if ic.AlibabaCloud.VpcID != "" {
+		allErrs = append(allErrs, validateVpc(client, ic, path)...)
+	}
+	if ic.AlibabaCloud.VSwitchIDs != nil {
```
staebler, line 42:
```suggestion
	if len(ic.AlibabaCloud.VSwitchIDs) != 0 {
```

pkg/asset/installconfig/alibabacloud/validation.go
staebler: Is this change related to this PR? If not, please move it to a new PR.

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -100,6 +121,92 @@ func validateResourceGroup(client *Client, ic *types.InstallConfig, path *field.
...
+	if vpcs.TotalCount == 0 {
+		allErrs = append(allErrs, field.NotFound(path.Child("vpcID"), ic.AlibabaCloud.VpcID))
+	}
+	return allErrs
+}
+
+func validateVSwitches(client *Client, ic *types.InstallConfig, path *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	zoneIDs := []string{}
```
staebler, line 139: Use a map instead of a list.
```suggestion
	zoneIDs := map[string]bool{}
```
Then you can do a lookup into the map rather than having to iterate over the list.
```
if zoneIDs[vswitches.VSwitches.VSwitch[0].ZoneId {
```

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -100,6 +121,92 @@ func validateResourceGroup(client *Client, ic *types.InstallConfig, path *field.
...
+		vswitches, err := client.ListVSwitches(id)
+		if err != nil {
+			return append(allErrs, field.InternalError(path.Child("vswitchIDs"), err))
+		}
+		if vswitches.TotalCount != 1 {
+			allErrs = append(allErrs, field.NotFound(path.Child("vswitchIDs").Index(idx), id))
+			continue
+		}
+		if vswitches.VSwitches.VSwitch[0].VpcId != ic.AlibabaCloud.VpcID {
+			allErrs = append(allErrs, field.Invalid(path.Child("vswitchIDs").Index(idx), id, fmt.Sprintf("The VSwitch does not belong to vpc %s", ic.AlibabaCloud.VpcID)))
```
staebler, line 150: All of the field error messages should being with lowercase letters.
```suggestion
 			allErrs = append(allErrs, field.Invalid(path.Child("vswitchIDs").Index(idx), id, fmt.Sprintf("the VSwitch does not belong to vpc %s", ic.AlibabaCloud.VpcID)))
```

pkg/asset/installconfig/alibabacloud/validation.go
staebler, line 139: At the end, there should be validation that there is a vSwitch for every zone specified in a machine pool. For example, if the control-plane machine pool specifies that zone Z should be used for control plane nodes, then there *must* be a vSwitch specified in zone Z.

pkg/asset/machines/alibabacloud/machinesets.go
```diff
@@ -37,7 +37,11 @@ func MachineSets(clusterID string, config *types.InstallConfig, pool *types.Mach
 			replicas++
 		}
 
-		provider, err := provider(clusterID, platform, mpool, idx, role, userDataSecret, resourceTags)
+		vswitchID, ok := vswitchMaps[az]
+		if len(vswitchMaps) > 0 && !ok {
+			return nil, errors.Errorf("no subnet for zone %s", az)
```
staebler, line 42: subnet?

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -73,13 +82,25 @@ func validateMachinePool(client *Client, ic *types.InstallConfig, fldPath *field
 			zones = zones[:3]
```
staebler, line 82: Cannot assume 3 here.
1. This should use the number of control plane replicas.
2. There may be less zones than there are control plane replicas.

pkg/types/alibabacloud/platform.go
```diff
@@ -7,11 +7,32 @@ type Platform struct {
 
 	// ResourceGroupID is the ID of an already existing resource group where the cluster should be installed.
 	// This resource group must be empty with no other resources when trying to use it for creating a cluster.
-	// If empty, a new resource group will created for the cluster.
-	// Destroying the cluster using installer will delete this resource group.
+	// If empty, a new resource group will created for the cluster,
+	// and destroying the cluster using installer will delete this resource group.
```
staebler, line 11: This change states that the resource group will only be when the user let the installer create the resource group. Instead, the resource group is *always* deleted, regardless of whether it was user-supplied or installer-created.

pkg/types/alibabacloud/platform.go
```diff
@@ -7,11 +7,32 @@ type Platform struct {
...
 	// +optional
 	ResourceGroupID string `json:"resourceGroupID,omitempty"`
 
+	// VpcID is the ID of an already existing VPC where the cluster should be installed.
+	// If empty, a new VPC will created for the cluster, and destroying
```
staebler, line 16: I do not think that the text about destroying the VPC should be added. If the installer creates the VPC, it is contracted that the installer will delete the VPC. If the installer does not delete the VPC, then that needs to be called out explicitly. If the installer is deleting the user-supplied VPC, then that needs to be called out explicitly (as was done for the resource group). Otherwise, nothing needs to be stated.

pkg/types/alibabacloud/platform.go
```diff
@@ -7,11 +7,32 @@ type Platform struct {
...
+	// the cluster using installer will delete this VPC.
+	// +optional
+	VpcID string `json:"vpcID,omitempty"`
+
+	// VSwitchIDs is the ID list of already existing VSwitches where cluster resources will be created.
+	// If empty, the new VSwitches will created for the cluster,
+	// and destroying the cluster using installer will delete these VSwitches.
```
staebler, line 23: See above for the comment regarding deleting the VPC.

pkg/types/alibabacloud/platform.go
```diff
@@ -7,11 +7,32 @@ type Platform struct {
...
 	// +optional
 	ResourceGroupID string `json:"resourceGroupID,omitempty"`
 
+	// VpcID is the ID of an already existing VPC where the cluster should be installed.
+	// If empty, a new VPC will created for the cluster, and destroying
+	// the cluster using installer will delete this VPC.
+	// +optional
+	VpcID string `json:"vpcID,omitempty"`
+
+	// VSwitchIDs is the ID list of already existing VSwitches where cluster resources will be created.
```
staebler, line 21: I assume that since the vSwitch is part of the VPC, that the vSwitches can only be user-provided if the VPC is also user-provided. If so, then there should (1) be text to that affect in this description, (2) be validation that the VPC ID is specified when the vSwitch IDs are specified, and (3) be validation that the vSwitches are part of the VPC.

data/data/alibabacloud/cluster/vpc/vpc.tf
```diff
@@ -4,9 +4,17 @@ locals {
   prefix            = var.cluster_id
   newbits           = tonumber(split("/", var.vpc_cidr_block)[1]) < 16 ? 20 - tonumber(split("/", var.vpc_cidr_block)[1]) : 4
```
staebler, line 5: nit not related to this PR: This would be a lot easier to read if it used the `max` function.
```suggestion
  newbits           = max(4, 20 - tonumber(split("/", var.vpc_cidr_block)[1]))
```

pkg/types/alibabacloud/validation/platform.go
```diff
@@ -17,11 +17,46 @@ func ValidatePlatform(p *alibabacloud.Platform, n *types.Networking, fldPath *fi
...
 }
 
+func validateVSwitches(p *alibabacloud.Platform, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if p.VpcID == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("vpcID"), "when using existing VSwitches, an existing VPC must be used"))
+	}
+
+	vswitchIDs := []string{}
```
staebler, line 42: Use a map instead of a list.

pkg/types/alibabacloud/validation/platform.go
```diff
@@ -17,11 +17,46 @@ func ValidatePlatform(p *alibabacloud.Platform, n *types.Networking, fldPath *fi
...
+	for _, vswitchID := range p.VSwitchIDs {
+		isRepeated := false
+		for _, id := range vswitchIDs {
+			if vswitchID == id {
+				isRepeated = true
+				break
+			}
+		}
+		if isRepeated {
+			allErrs = append(allErrs, field.Duplicate(fldPath.Child("vswitchIDs"), vswitchID))
```
staebler, line 52: Include the index to the duplicate in the field path.

pkg/types/alibabacloud/platform.go
staebler: Moved this to #5378.

pkg/types/alibabacloud/platform.go
```diff
@@ -7,11 +7,32 @@ type Platform struct {
 
 	// ResourceGroupID is the ID of an already existing resource group where the cluster should be installed.
 	// This resource group must be empty with no other resources when trying to use it for creating a cluster.
```
staebler, line 9: Is this strictly true? Or could the user-supplied VPC be in the resource group?

pkg/types/alibabacloud/metadata.go
```diff
@@ -2,7 +2,10 @@ package alibabacloud
...
 // Metadata contains Alibaba Cloud metadata (e.g. for uninstalling the cluster).
 type Metadata struct {
-	Region          string `json:"region"`
-	ResourceGroupID string `json:"resourceGroupID"`
-	ClusterDomain   string `json:"clusterDomain"`
+	Region          string   `json:"region"`
+	ResourceGroupID string   `json:"resourceGroupID"`
+	VpcID           string   `json:"vpcID"`
+	VSwitchIDs      []string `json:"vswitchIDs"`
+	PrivateZoneID   string   `json:"privateZoneID"`
```
staebler, line 9: In what way are these values going to be used while destroying the cluster?

data/data/alibabacloud/cluster/dns/privatezone.tf
```diff
@@ -52,7 +56,7 @@ resource "alicloud_pvtz_zone_record" "pvtz_record_api" {
 resource "alicloud_pvtz_zone_record" "pvtz_record_masters" {
```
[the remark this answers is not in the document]
bd233: I just referred to [Table 5. Required DNS records
Component](https://docs.openshift.com/container-platform/4.9/installing/installing_bare_metal/installing-bare-metal.html#installation-dns-user-infra_installing-bare-metal), if this is not needed in the installer, I can remove them.
image

data/data/alibabacloud/cluster/master/variables.tf
```diff
@@ -11,9 +11,14 @@ variable "vpc_id" {
   description = "The VPC ID of the master ECS."
 }
 
-variable "vswitch_ids" {
+variable "zone_ids" {
   type        = list(string)
-  description = "The VSwitch IDs of the master ECS. Example: [vsw-xxx1, vsw-xxx2, vsw-xxx3]"
+  description = "The availability zones in which to create the masters. The length of this list must match master_count."
```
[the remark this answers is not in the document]
kwoodson: @bd233 Please update, thanks.

data/data/alibabacloud/cluster/dns/privatezone.tf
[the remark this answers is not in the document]
staebler: Those requirements are for on-prem installations. My assumption is that the alibaba cloud already provides DNS resolution from one VM instance to another via the hostname of the VM. I also assume that the hostname of the VMs will not contain the cluster domain but instead contain a domain name specific to the alibaba cloud, in which case the A records created here would not even be relevant. Are either or both of those assumptions invalid?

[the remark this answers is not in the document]
bd233: > My assumption is that the alibaba cloud already provides DNS resolution from one VM instance to another via the hostname of the VM. 

@mtulio @kwoodson As you know, there is no open API for the function of Automatically synchronize hostnames. After the Installer is executed, the user needs to manually set it to the open state (I guess). I am not sure if this requires adding the records of `<master><n>.<cluster name>.<base domain>` in terrform, and whether this will affect the installation process of the installer?
image

data/data/alibabacloud/cluster/vpc/vpc.tf
bd233: I can solve this by `length(toset(var.zone_ids))`, but I think this may not be the most suitable solution. I should ensure that the zones in var_zone_ids is unique,  at the same time, the length of ali_zone_ids need not be equal master_count. WDYT?

pkg/asset/installconfig/alibabacloud/validation.go
bd233: Yes, I think it is more appropriate to add  the validation of the machine pool, I will create a new PR.

pkg/asset/installconfig/alibabacloud/validation.go
bd233: Ok, get it

pkg/types/alibabacloud/platform.go
bd233: Regarding (2) and (3), relevant  validation  have already been made, and I will add  text  in this description

> that the vSwitches can only be user-provided if the VPC is also user-provided.

I disagree with this point. I think it’s okay for the user to only provide the VPC ID, and the installer will create the VSwitches

data/data/alibabacloud/cluster/vpc/vpc.tf
bd233: Okey, thanks

pkg/types/alibabacloud/platform.go
bd233: In fact, it has no effect if it is not empty. we will not query and destroy resources through resource group. So this description is incorrect, I will delete it.

pkg/types/alibabacloud/metadata.go
bd233: I have to admit that these are not used, I will  delete it.

data/data/alibabacloud/cluster/vpc/vpc.tf
bd233: # 5413

bd233: PR #5414

data/data/alibabacloud/cluster/dns/privatezone.tf
```diff
@@ -52,7 +56,7 @@ resource "alicloud_pvtz_zone_record" "pvtz_record_api" {
...
   count = var.master_count
 
-  zone_id = alicloud_pvtz_zone.pvtz.id
+  zone_id = local.private_zone_id
   type    = "A"
   rr      = "master${reverse(split("-", keys(var.master_ips)[count.index]))[0]}"
```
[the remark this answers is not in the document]
bd233: PR #5313

data/data/alibabacloud/cluster/vpc/vpc.tf
staebler: Yes, `var.zone_ids` should be a distinct list of availability zones. It should also include zones for the workers, which appears to be missing currently.

pkg/types/alibabacloud/platform.go
staebler: Sorry, I thought that I had deleted parts (2) and (3) after re-checking that the validation was indeed already there.

> > that the vSwitches can only be user-provided if the VPC is also user-provided.
> 
> I disagree with this point. I think it’s okay for the user to only provide the VPC ID, and the installer will create the VSwitches

Yes, I agree that it is OK to use a user-provided VPC and installer-generated VSwitches. My point is the converse. If the user provides the VSwitches, then they *must* all provide the VPC. I would like to call that out specifically in the description for the field so that the user does not assume that the installer will deduce the VPC to use from the VSwitches (which the installer could presumably do).

pkg/types/alibabacloud/platform.go
staebler: Will the destroyer delete the resource group if it is user-supplied?

data/data/alibabacloud/cluster/dns/privatezone.tf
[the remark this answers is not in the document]
kwoodson: We are able to proceed without this functionality.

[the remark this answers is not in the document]
bd233: @kwoodson Okey, so I remove these records about master. @staebler  And the PR #5413 may be meaningless, close it? or update removed code on #5413?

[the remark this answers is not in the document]
staebler: If we don't need these A records, let's close #5413 and open a new PR to remove the A records.

data/data/alibabacloud/cluster/master/variables.tf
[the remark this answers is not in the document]
mtulio: @staebler @kwoodson in us-east-1 region I can see only two zones, not sure if it is a general information to all regions.
Screenshot from 2021-11-24 14-24-12

data/data/alibabacloud/cluster/dns/privatezone.tf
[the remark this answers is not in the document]
bd233: New PR #5422 : remove the A records

data/data/alibabacloud/cluster/vpc/vpc.tf
bd233: Yes, thank you for your reminder, the worker zones is missing currently, I will create a new PR to update.

bd233:
@staebler I should have updated everything, please continue to review

pkg/asset/machines/alibabacloud/machines.go
```diff
@@ -104,16 +99,29 @@ func provider(clusterID string,
...
 		IoOptimized:        "optimized",
 		RAMRoleName:        fmt.Sprintf("%s-role-%s", clusterID, role),
 		SecurityGroups:     sgTags,
-		VSwitch:            &vSwitchTags,
-	}, nil
+	}
+
+	if vswitchID == "" {
+		vSwitchTags := alibabacloudprovider.ResourceTagReference{
+			Tags: append(tags, alibabacloudprovider.Tag{
```
staebler, line 114: How does the alibaba machine-api provider work when trying to find the VSwitch? Do the tags on the VSwitch have to match exactly the tags specified in the machine spec? Or do the tags on the VSwitch just need to be a superset of the tags specified in the machine spec? I am a bit surprised that we are adding *all* of the tags for the machine to the tags that we are looking for on the VSwitch, as opposed to just including the "Name" tag.

pkg/asset/machines/master.go
```diff
@@ -162,19 +162,41 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 	machines := []machineapi.Machine{}
 	switch ic.Platform.Name() {
 	case alibabacloudtypes.Name:
+		client, err := installConfig.AlibabaCloud.Client()
```
staebler, line 165: For future work, we should cache the VSwitches and availability zones so that we don't have to look them up multiple times.

pkg/types/alibabacloud/platform.go
```diff
@@ -12,6 +12,26 @@ type Platform struct {
 	// +optional
 	ResourceGroupID string `json:"resourceGroupID,omitempty"`
 
+	// VpcID is the ID of an already existing VPC where the cluster should be installed.
+	// If empty, a new VPC will created for the cluster.
```
staebler, line 16:
```suggestion
	// If empty, the installer will create a new VPC for the cluster.
```

pkg/types/alibabacloud/platform.go
```diff
@@ -12,6 +12,26 @@ type Platform struct {
...
+	// +optional
+	VpcID string `json:"vpcID,omitempty"`
+
+	// VSwitchIDs is the ID list of already existing VSwitches where cluster resources will be created.
+	// The existing VSwitches can only be used when also using existing VPC.
+	// If empty, the new VSwitches will created for the cluster.
```
staebler, line 22:
```suggestion
	// If empty, the installer will create new VSwitches for the cluster.
```

pkg/types/alibabacloud/validation/platform.go
staebler: Still need to include the index in the field path.
```suggestion
			allErrs = append(allErrs, field.Duplicate(fldPath.Child("vswitchIDs").Index(i), vswitchID))
```

pkg/asset/installconfig/alibabacloud/client.go
```diff
@@ -236,14 +236,58 @@ func (client *Client) ListDNSDomain(baseDomain string) (response *alidns.Describ
 func (client *Client) ListPrivateZones(zoneName string) (response *pvtz.DescribeZonesResponse, err error) {
 	request := pvtz.CreateDescribeZonesRequest()
 	request.Lang = "en"
-	request.Keyword = zoneName
+	if zoneName != "" {
```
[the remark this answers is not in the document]
bd233: > I don't understand why this change is necessary. There is no change to request.Keyword when zoneName is an empty string, since request.Keyword is already an empty string.
> Moreover, it seems like a violation of the contract of the ListPrivateZones function to pass in an empty value for zoneName anyway.

I understand your doubts. In the initial assumption, the zoneName of ListPrivateZones is an optional parameter. At present, if users do not want to use zoneName, they must specify an empty string, which is obviously unreasonable.

pkg/asset/machines/alibabacloud/machines.go
bd233: > Or do the tags on the VSwitch just need to be a superset of the tags specified in the machine spec?

Provider will match a superset of the tags specified in the machine spec.

>  I am a bit surprised that we are adding all of the tags for the machine to the tags that we are looking for on the VSwitch, as opposed to just including the "Name" tag.

In fact, the Name tag can be matched to the only VSwitch.

pkg/asset/cluster/alibabacloud/alibabacloud.go
```diff
@@ -8,8 +8,7 @@ import (
 // Metadata converts an install configuration to Alibaba Cloud metadata.
 func Metadata(config *types.InstallConfig) *alibabacloud.Metadata {
 	return &alibabacloud.Metadata{
-		Region:          config.Platform.AlibabaCloud.Region,
-		ResourceGroupID: config.Platform.AlibabaCloud.ResourceGroupID,
```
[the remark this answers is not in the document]
bd233: #5431

pkg/types/alibabacloud/metadata.go
```diff
@@ -2,7 +2,6 @@ package alibabacloud
 
 // Metadata contains Alibaba Cloud metadata (e.g. for uninstalling the cluster).
 type Metadata struct {
-	Region          string `json:"region"`
-	ResourceGroupID string `json:"resourceGroupID"`
```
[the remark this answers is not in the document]
bd233: #5431

pkg/asset/installconfig/alibabacloud/client.go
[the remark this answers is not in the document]
bd233: @staebler In the latest update, I used Params type to store parameters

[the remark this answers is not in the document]
bd233: @staebler In the latest update, I used Params type to store parameters

data/data/alibabacloud/cluster/vpc/vpc.tf
bd233: > Yes, var.zone_ids should be a distinct list of availability zones. It should also include zones for the workers, which appears to be missing currently.

@staebler #5438

kwoodson:
@bd233 @staebler Where are we with this one? Would you mind updating the status here?

bd233:
> @bd233 @staebler Where are we with this one? Would you mind updating the status here?

@staebler I have updated, please continue to review, thank you.

pkg/asset/installconfig/alibabacloud/client.go
```diff
@@ -233,19 +233,75 @@ func (client *Client) ListDNSDomain(baseDomain string) (response *alidns.Describ
...
 }
 
+// ListPrivateZonesParams holds the params of ListPrivateZones function
+type ListPrivateZonesParams struct {
+	zoneName string
+}
+
 // ListPrivateZones gets the list of privatzones.
-func (client *Client) ListPrivateZones(zoneName string) (response *pvtz.DescribeZonesResponse, err error) {
+func (client *Client) ListPrivateZones(params *ListPrivateZonesParams) (response *pvtz.DescribeZonesResponse, err error) {
```
staebler, line 242: I don't understand the purpose of this change?

pkg/asset/installconfig/alibabacloud/client.go
```diff
@@ -233,19 +233,75 @@ func (client *Client) ListDNSDomain(baseDomain string) (response *alidns.Describ
...
+// ListVpcsParams holds the params of ListVpcs function
+type ListVpcsParams struct {
+	vpcID string
+}
+
+// ListVpcs gets the list of VPCs.
+func (client *Client) ListVpcs(params *ListVpcsParams) (response *vpc.DescribeVpcsResponse, err error) {
+	request := vpc.CreateDescribeVpcsRequest()
+	request.RegionId = client.RegionID
+	if params.vpcID != "" {
```
staebler, line 275: Remove this check. It is not necessary.

pkg/asset/installconfig/alibabacloud/client.go
```diff
@@ -233,19 +233,75 @@ func (client *Client) ListDNSDomain(baseDomain string) (response *alidns.Describ
...
+// ListVSwitchesParams holds the params of ListVSwitches function
+type ListVSwitchesParams struct {
+	vswitchID string
+}
+
+// ListVSwitches gets the list of VSwitches.
+func (client *Client) ListVSwitches(params *ListVSwitchesParams) (response *vpc.DescribeVSwitchesResponse, err error) {
+	request := vpc.CreateDescribeVSwitchesRequest()
+	request.RegionId = client.RegionID
+	if params.vswitchID != "" {
```
staebler, line 295: Remove this check. It is not necessary.

kwoodson:
@bd233 Please review the comments and address. Thanks!

pkg/asset/installconfig/alibabacloud/client.go
bd233: I agree that this is not a good solution. I hope that `ListPrivateZones` can accept an optional parameter more flexibly, but compared to specifying at least an empty string before, now,  must specify a `ListPrivateZonesParams`, which becomes more complicated.

staebler: Why do you need an optional parameter? You are not passing nil in either of the two calls to the function.

The one call where it seems like you don't want to search for a particular zone by name is still passing in an empty struct instead of nil. In that case, however, why wouldn't we want to search for the private zone using the zone name from the cluster domain and validate that the zone has the specified zone ID? That seems much more efficient than trying to pull down all of the private zones looking for the one with the specified zone ID.

If you don't like that change, then at least go back to the `zoneName` directly to `ListPrivateZones`. All I wanted you to do originally was to remove the unneeded check for whether `zoneName` was empty.

staebler: Out of curiosity, are you not able to use `QueryVpcId` in the `DescribeZonesRequest` struct to search for the zone by ID?

bd233: > why wouldn't we want to search for the private zone using the zone name from the cluster domain and validate that the zone has the specified zone ID? That seems much more efficient than trying to pull down all of the private zones looking for the one with the specified zone ID.

Yes, this is indeed a better solution, but in order to distinguish whether it is an unbound VPC error or a different zone name error when using `QueryVpcId` (I even ignored this parameter), I specified an empty string to `zoneName`

pkg/asset/installconfig/alibabacloud/client.go
```diff
@@ -234,18 +234,46 @@ func (client *Client) ListDNSDomain(baseDomain string) (response *alidns.Describ
 }
 
 // ListPrivateZones gets the list of privatzones.
-func (client *Client) ListPrivateZones(zoneName string) (response *pvtz.DescribeZonesResponse, err error) {
+func (client *Client) ListPrivateZones(zoneName string, queryVpcId string) (response *pvtz.DescribeZonesResponse, err error) {
```
staebler, line 237: I would rather these be two separate functions.
```
func (client *Client) ListPrivateZonesByName(zoneName string) (response *pvtz.DescribeZonesResponse, err error) {
```
```
func (client *Client) ListPrivateZonesByVPC(vpcId string) (response *pvtz.DescribeZonesResponse, err error) {
```

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -100,6 +109,73 @@ func validateResourceGroup(client *Client, ic *types.InstallConfig, path *field.
...
+	return allErrs
+}
+
+func validatePrivateZoneID(client *Client, ic *types.InstallConfig, path *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	fldpath := path.Child("privateZoneID")
+	zoneName := ic.ClusterDomain()
+	isExist := false
+
+	zones, err := client.ListPrivateZones("", ic.AlibabaCloud.VpcID)
```
staebler, line 159: Doh. I missed that `QueryVpcId` was VPC ID instead of Zone ID. Nevertheless, I think this is fine.

staebler (commented):
Just the one nit.

pkg/asset/installconfig/alibabacloud/client.go
bd233: Can't agree more. But I added two functions on the basis of ListPrivateZones

pkg/asset/installconfig/alibabacloud/client.go
```diff
@@ -233,19 +233,55 @@ func (client *Client) ListDNSDomain(baseDomain string) (response *alidns.Describ
 	return
 }
 
+func (client *Client) ListPrivateZonesByName(zoneName string) (response *pvtz.DescribeZonesResponse, err error) {
```
staebler, line 236:
```suggestion
// ListPrivateZonesByName gets the list of privatezones with the specified name.
func (client *Client) ListPrivateZonesByName(zoneName string) (response *pvtz.DescribeZonesResponse, err error) {
```

pkg/asset/installconfig/alibabacloud/client.go
```diff
@@ -233,19 +233,55 @@ func (client *Client) ListDNSDomain(baseDomain string) (response *alidns.Describ
...
+	return client.ListPrivateZones(zoneName, "")
+}
+
+func (client *Client) ListPrivateZonesByVPC(queryVpcID string) (response *pvtz.DescribeZonesResponse, err error) {
```
staebler, line 240:
```suggestion
// ListPrivateZonesByVPC gets the list of privatezones attached to the specified VPC.
func (client *Client) ListPrivateZonesByVPC(queryVpcID string) (response *pvtz.DescribeZonesResponse, err error) {
```

bd233: Aha...I have add the comments.

staebler (approved):
/lgtm
/approve

mtulio:
/hold
Yesterday I ran some tests in this PR and I found some errors. Let me wait to build a new cluster using existing VPC and vSwitchs, I will give feedback soon.

mtulio:
@bd233 as we've commented on Slack, could you please rebase this PR to get the fix from PR #5426  that was recently merged? Thanks!

mtulio:
I ran the following tests and it worked as expected:
1. Install a cluster using existing VPC
2. Install a cluster using existing vSwitches
3. Install a cluster using the existing private DNS zone

The errors I've mentioned were fixed in the latest change. The behavior that I saw of more than one vSwitch created by Zone, was the NatGW which is expected to be created on Zone A.

I would like to take some notes, that's not directly related or blocking this PR, but I collect when running those tests:
- There is a vSwitch for NatGW  allocating /20 subnet (4092 IP Address) to be used with only one asset (NatGW)
- Regarding the NatGW placed on Zone A: need to review what happens if Zone A is unavailable.
- The vSwitch for the secondary zone of an SLB is not explicitly added when it is provisioned, I am not sure how it can handle when the Primary zone is unavailable. I tested removing the vSW-AZ_A (secondary) and it allowed, even if I said that Zone A should be the fallback from the SLB, when I tried to remove vSW-AZ_B (Primary zone), it does not allow due to association with vSwitch. It seems that SLB does not have an explicit association with the Secondary zone (vSwitch), it also makes the HA question open about the behavior when one zone is unavailable, in this case, the Primary zone.

LGTM

cc @staebler @kwoodson @bd233

kwoodson:
@staebler  Looks like an error in the AWS terraform generation. Any ideas on this one?
```
level=error
level=error msg=Error: Provider produced inconsistent result after apply
level=error
level=error msg=When applying changes to module.vpc.aws_vpc_dhcp_options_association.main[0],
level=error msg=provider "registry.terraform.io/-/aws" produced an unexpected new value for
level=error msg=was present, but now absent.
level=error
level=error msg=This is a bug in the provider, which should be reported in the provider's own
level=error msg=issue tracker.
level=error
level=fatal msg=failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to apply Terraform: failed to complete the change
```

mtulio:
@kwoodson this looks like a possible flake similar to what was reported in https://github.com/hashicorp/terraform-provider-aws/issues/16142
/retest-required

PR #5382, opened 2021-11-12
azurestack: Fix gather bootstrap for azurestack

rna-afk:
The terraform stages needs just a list of control plane ips to
gather bootstrap logs but the azurestack terraform output is
a list of lists causing the gather command to fail. Fixing the
output of the terraform stage to flatten the ips to a single
list. This will enable azurestack gather command.

### DIFF
--- a/data/data/azurestack/cluster/master/outputs.tf
+++ b/data/data/azurestack/cluster/master/outputs.tf
@@ -1,3 +1,3 @@
 output "ip_addresses" {
-  value = azurestack_network_interface.master.*.private_ip_addresses
+  value = azurestack_network_interface.master.*.private_ip_address
 }


staebler (commented):
1. The output from the master module should be flattened rather than the output from the cluster root module.
2. We don't want to naively flatten the list of lists. Instead, we want to select the first IP address from each master to include in the list. Conveniently, you can use `private_ip_address` instead of `private_ip_addresses` to get just the first IP address.
3. It looks like the tf for public azure has the same issue.

patrickdillon:
In my testing this is failing with:
```
FATAL could not read control plane IPs from outputs file "/home/padillon/ct/test-clusters/202112021228-ash/cluster.tfvars.json" 
```

As pointed out in https://github.com/openshift/installer/pull/5443#issuecomment-984923878, this gather code is actually executing the gather code for public Azure stages.

patrickdillon:
Doh. I'm deleting my previous comment about this not working. My test was wrong. Hopefully you missed my incorrect comment. otherwise, ignore it.

patrickdillon:
I do think this is invoking the code for the Azure stages, but they are compatible so it works.

staebler:
These changes look good to me.
/approve

@rna-afk Since the e2e-azurestack job is broken, could you do a manual azurestack install and provide the contents of the cluster.tfvars.json file as proof that this is working correctly now?

rna-afk:
I'm always getting a timeout error so the cluster.tfvars.json file isn't getting generated. but I can see the right format in the terraform.cluster.tfstate

```
"control_plane_ips": {
      "value": [
        "10.0.0.8",
        "10.0.0.6",
        "10.0.0.7"
      ],

```

staebler:
/lgtm

@rna-afk Would you mind making this same PR for public Azure as well?

rna-afk:
Sure. This should do it.
https://github.com/openshift/installer/pull/5448

PR #5383, opened 2021-11-14
remove MCO manifests

yuvalk:
this is no longer needed


sinnykumari:
LGTM
/retest

yuvalk:
/assign @staebler 
will appreciate your help with this one

staebler:
@yuvalk Can you add some color for why this is no longer needed?

yuvalk:
@staebler see bz https://bugzilla.redhat.com/show_bug.cgi?id=1978581
and https://bugzilla.redhat.com/show_bug.cgi?id=1805488

tl;dr:
These appear to be either copy and paste based on bad documentation, or previously required before OCP 4.7 due to significant boot delays.
With some basic testing, it appears that these can be removed.

let me know if you want me to include this info in the commit msg itself

staebler:
I would like to take a step back from the changes made in this PR and evaluate whether the installer should even be creating these manifests. I feel like this is leftover from the tectonic days. The machine-config-operator should be rendering any manifests that it needs during bootstrapping.

yuvalk:
> I would like to take a step back from the changes made in this PR and evaluate whether the installer should even be creating these manifests. I feel like this is leftover from the tectonic days. The machine-config-operator should be rendering any manifests that it needs during bootstrapping or including them as part of the release payload.
> 
> The machine-config-operator is already including the manifest for the openshift-machine-config-operator namespace as part of the release payload. Is there some reason why the namespace needs to exist earlier? Even so, rendering it during bootstrap would accomplish the same thing and push the ownership of the manifest to the MCO, where it belongs.
> 
> As an aside, the MCO uses a runlevel of 1 for the openshift-machine-config-operator namespace manifest that is included in the release payload. https://github.com/openshift/machine-config-operator/blob/6d2b493e0f4ed4dcbe32a4666bd0fa8facc6e5b1/install/0000_80_machine-config-operator_00_namespace.yaml#L13

We are addressing that in https://github.com/openshift/machine-config-operator/pull/2655
and the whole reason for this PR, is to be on par with that from installer side too.

Let's try and remove the file altogether (from installer)

yuvalk:
/test e2e-aws-upgrade
build cluster had issues

staebler:
Looks like removing the manifest entirely works just fine. The changes look good to me. Clean up the commit message and the PR title, and I'll give it a lgtm.

staebler (approved):
/lgtm
/approve

PR #5384, opened 2021-11-15
Use neutral pronoun in doc.

flopotok:


staebler (commented):
Thank you for this change. Could you please improve on the commit message?
See https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#commit-message-format.

PR #5386, opened 2021-11-15
[release-4.8] Bug 2023363: Add ingress rules to master SG for compact clusters

mandre:
In the case of compact clusters, where master nodes are schedulable, we
need to open ports for ingress, same as for workers.

Manual cherry-pick of https://github.com/openshift/installer/pull/5320 onto release-4.8.

### DIFF
--- a/data/data/config.tf
+++ b/data/data/config.tf
@@ -32,6 +32,17 @@ EOF
 
 }
 
+variable "masters_schedulable" {
+  type = bool
+
+  default = false
+
+  description = <<EOF
+Whether master nodes are schedulables.
+EOF
+
+}
+
 variable "base_domain" {
   type = string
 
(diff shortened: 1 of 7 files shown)


data/data/config.tf
```diff
@@ -32,6 +32,17 @@ EOF
 
 }
 
+variable "masters_schedulable" {
+  type = bool
+
+  default = false
+
+  description = <<EOF
+Whether master nodes are schedulables.
```
staebler, line 41:
```suggestion
Whether master nodes are schedulable.
```

staebler (approved):

data/data/config.tf
mandre: Oops. Since it's a backport, I'd prefer we keep the patch unchanged and leave the typo there.

PR #5392, opened 2021-11-16
Bug 1997059: aws: Filter out local zones when generating a default list of zones

squidboylan:
Certain resources cannot be created in local zones such as load
balancers, this causes an openshift installation to fail, because of
this we need to filter out local zones when generating a default list of
zones to use

https://bugzilla.redhat.com/show_bug.cgi?id=1997059

### DIFF
--- a/pkg/asset/installconfig/aws/availabilityzones.go
+++ b/pkg/asset/installconfig/aws/availabilityzones.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"context"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
@@ -28,9 +29,21 @@ func availabilityZones(ctx context.Context, session *session.Session, region str
 		return nil, errors.Wrap(err, "fetching availability zones")
 	}
 
+	// This finds local zones and allows us to filter them out. It does so by
+	// using the fact that local zones have a predictable format based on the
+	// region, for example a local zone in us-west-2 will look like
+	// "us-west-2-las-1a". This is calculated relative to the region name as not
+	// all regions follow the same format with 2 hyphens.
+	// This is a temporary fix until we update our aws-sdk-go package to a
+	// version that adds the ZoneType field to the AvailabilityZone struct
+	numberOfHyphensInRegion := strings.Count(region, "-")
+
 	zones := []string{}
 	for _, zone := range resp.AvailabilityZones {
-		zones = append(zones, *zone.ZoneName)
+		numberOfHyphensInZone := strings.Count(*zone.ZoneName, "-")
+		if numberOfHyphensInZone != numberOfHyphensInRegion+2 {
+			zones = append(zones, *zone.ZoneName)
+		}
 	}
 
 	if len(zones) == 0 {


jstuever (commented):
Looks good.

pkg/asset/installconfig/aws/availabilityzones.go
```diff
@@ -28,9 +29,18 @@ func availabilityZones(ctx context.Context, session *session.Session, region str
 		return nil, errors.Wrap(err, "fetching availability zones")
 	}
 
+	// This regex finds local zones and allows us to filter them out.
+	// This is a temporary fix until we update our aws-sdk-go package to a
+	// version that adds the ZoneType field to the AvailabilityZone struct
+	regex, err := regexp.Compile("^\\w+-\\w+-\\w+-\\w+-\\w+$")
```
patrickdillon, line 35: nit: because I'm not great at parsing regex, I find it helpful to have examples. 

```
us-east-1-bos-1a
--

```
us-east-1a
```

**Comment by patrickdillon** [comment]
/retitle Bug 1997059: aws: Filter out local zones when generating a default list of zones

File: pkg/asset/installconfig/aws/availabilityzones.go [hunk 2]
```diff
@@ -28,9 +29,18 @@ func availabilityZones(ctx context.Context, session *session.Session, region str
 		return nil, errors.Wrap(err, "fetching availability zones")
 	}
 
+	// This regex finds local zones and allows us to filter them out.
+	// This is a temporary fix until we update our aws-sdk-go package to a
```
patrickdillon, line 33: can you create a jira card for doing this once we update the sdk?

pkg/asset/installconfig/aws/availabilityzones.go
staebler, line 35: I find it a bit easier to read without needed to have the escapes for the backslash.
```suggestion
	regex, err := regexp.Compile(`^\w+-\w+-\w+-\w+-\w+$`)
```
I would also make this a package-level variable and use `MustCompile` instead. The regex is static so we know it must compile (or it is a coding error).

However, a regex seems like it may be overkill here. All we need to do is search for hyphens in the string. I'm not sure which people think is easier to understand. They both need examples, either way.
```
len(strings.Split(zoneName, "-")) == 5
```

staebler, line 35: Will this test hold up for other AWS partitions? For example, in the us gov partition, the regions are `us-gov-east-1` and `us-gov-west-1`. Presumably, if c2s did support local zones, then the zone name would have 5 hyphens instead of 4. Should we instead check the number of parts after the region name in the zone name?

jstuever: The local zones are currently only in us-east-1 and us-west-2. IMHO this is sufficient to hold us over until we are able to update the packages and provide a more permanent solution.

squidboylan: Yeah this makes sense to me

staebler (commented):
/approve

staebler (approved):
/lgtm

staebler:
since these changes only affect whether the install was allowed to proceed or not

PR #5393, opened 2021-11-16
Vendor: Bump API for Azure Marketplace Images

patrickdillon:
Bump API to get Machine Provider Spec for updated Azure Marketplace images.

Depends on openshift/api#1057


patrickdillon:
I started writing the code that would directly add this [Image API ](https://github.com/openshift/api/blob/master/machine/v1beta1/types_azureprovider.go#L160-L185) to the Installer types machinepools to allow users to enter these values in the install config, but I decided to hold off. The current changes will allow users to edit the manifests, but for the install config, I am uncertain about whether we want expose the entire API or keep something simpler (e.g. "Use marketplace image" rather than specify all the values of an image."

cc @joelddiaz @2uasimojo for hive use case.

patrickdillon:
/retitle Bug 2023434: Vendor: Bump API for Azure Marketplace Images

This will need to be backported.

patrickdillon:
> @patrickdillon: This pull request references [Bugzilla bug 2023434](https://bugzilla.redhat.com/show_bug.cgi?id=2023434), which is invalid:
> 
> * expected the bug to be in one of the following states: NEW, ASSIGNED, ON_DEV, POST, POST, but it is ON_QA instead
> 
> Comment `/bugzilla refresh` to re-evaluate validity if changes to the Bugzilla bug are made, or edit the title of this pull request to link to a different bug.

Not sure if it is possible to have this one BZ be a catchall BZ for all azure marketplace PRs, or if we need individual ones.

patrickdillon:
/retitle Vendor: Bump API for Azure Marketplace Images

patrickdillon:
> @patrickdillon: No Bugzilla bug is referenced in the title of this pull request. To reference a bug, add 'Bug XXX:' to the title of this pull request and request another bug refresh with `/bugzilla refresh`.

We only need a bz for the 4.9 backport, which can be based on the general bz.

staebler (approved):
/approve

2uasimojo:
> I started writing the code that would directly add this [Image API ](https://github.com/openshift/api/blob/master/machine/v1beta1/types_azureprovider.go#L160-L185) to the Installer types machinepools to allow users to enter these values in the install config, but I decided to hold off. The current changes will allow users to edit the manifests, but for the install config, I am uncertain about whether we want expose the entire API or keep something simpler (e.g. "Use marketplace image" rather than specify all the values of an image."

Wouldn't that be the behavior if the `image` field was nil/unspecified?

PR #4664, opened 2021-02-17
vSphere IPI: Enable thin provisioning via the OVA import

jcpowermac:
This PR enables thin provisioning via the importing process
of the RHCOS ova. Terraform will then clone the virtual machine
based on the existing template disk type.

This adds an additional option in the vsphere platform struct `Thin`.

- [x] Performance testing needs to be completed before offering this as an option
- [ ] Virtual machine provisioning with storage options can be problematic, additional testing in a nested environment is required.

PR #5395, opened 2021-11-17
vSphere: Set default disktype to thick not eager

jcpowermac:
In #4664 I incorrectly commented to use `eagerZeroedThick`.
The correct disktype to keep the vSphere install the same
as previous versions is `thick`. eagerZeroed as named
will write zeros when creating the vmdk which significantly
extends the time to import and clone virtual machines.

### DIFF
--- a/data/data/vsphere/variables-vsphere.tf
+++ b/data/data/vsphere/variables-vsphere.tf
@@ -83,5 +83,5 @@ variable "vsphere_control_plane_cores_per_socket" {
 }
 variable "vsphere_disk_type" {
   type    = string
-  default = "eagerZeroedThick"
+  default = "thick"
 }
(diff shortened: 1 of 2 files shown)


jcpowermac:
> @jcpowermac As a follow-up, can you do the following, which should have been done as part of #4664 as well?
> 
> 1. Add a `kubebuilder:validation:Enum` annotation to the `DiskType` type.
> 2. Add documentation to the `DiskType` field making it explicit what the default value is.
> 3. Add code to the `SetPlatformDefaults` to set the default value when the user omits it.
> 
> /approve

Thanks @staebler 

@ayesha54 is this something you would want to take as an action?

ayesha54:
> > @jcpowermac As a follow-up, can you do the following, which should have been done as part of #4664 as well?
> > 
> > 1. Add a `kubebuilder:validation:Enum` annotation to the `DiskType` type.
> > 2. Add documentation to the `DiskType` field making it explicit what the default value is.
> > 3. Add code to the `SetPlatformDefaults` to set the default value when the user omits it.
> > 
> > /approve
> 
> Thanks @staebler
> 
> @ayesha54 is this something you would want to take as an action?

yes sure

PR #5396, opened 2021-11-18
Alibaba Cloud Provider: remove the ingress record.

kwoodson:
The ingress record is created by the ingress-operator is no longer required to be created by the installer.


kwoodson:
cc @bd233

staebler (approved):
/lgtm

PR #5397, opened 2021-11-18
Update code and Dockerfiles for golang 1.17

yselkowitz:
- Fix gofmt for 1.17
- Update CI to golang-1.17

Closes #5364
Closes #5365
Closes #5368

### DIFF
--- a/data/assets.go
+++ b/data/assets.go
@@ -1,4 +1,6 @@
+//go:build !release
 // +build !release
+
 //go:generate go run assets_generate.go
 
 package data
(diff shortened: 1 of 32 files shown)


data/assets.go
```diff
@@ -1,4 +1,6 @@
+//go:build !release
 // +build !release
```
staebler, line 2: Should we remove the `+build` directives, too, and force the move to go1.17 by changing https://github.com/openshift/installer/blob/cd170249fb5d7ce041ee35aae1486872cb076ef6/hack/build.sh#L8?

yselkowitz: I think that's a bit premature.  For now I think it's sufficient to make sure installer *can* build with 1.17.

staebler (approved):
/lgtm

PR #5398, opened 2021-11-18
Bug 2015837: openstack: unset OS_CLOUD for running Terraform

pierreprinetti:
In terraform-provider-openstack, the OS_CLOUD environment variable takes
precedence over what's passed directly.

### DIFF
--- a/pkg/terraform/exec/plugins/openstack.go
+++ b/pkg/terraform/exec/plugins/openstack.go
@@ -1,11 +1,19 @@
 package plugins
 
 import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


pierreprinetti:
After some investigation, my conclusion is that bumping terraform-provider-openstack would require an effort that this bug does not justify. Going forward with `os.Unset("OS_CLOUD")` seems like the rational choice here.

pierreprinetti:
/retitle Bug 2015837: openstack: unset OS_CLOUD for running Terraform

PR #5400, opened 2021-11-18
upi/aws/cloudformation: make bootstrap instance type configurable

Prashanth684:
this is needed to spin up ARM UPI clusters. given the fact that the release payloads are not manifest listed yet, the bootstrap instance type has to correspond to the specific architecture of the release payload.


yselkowitz:
This is the one that really matters here
/test e2e-aws-upi

Prashanth684:
looks like we need this: https://github.com/openshift/release/pull/22317 for  e2e-aws-upi to run.

PR #5402, opened 2021-11-18
Revert "Workaround golang linux/arm64 link error"

yselkowitz:
golang 1.17 fixed the underlying bug, so we can now make the linux/arm64 default/cloud installer statically linked just like the others.  This is also a prerequisite for a darwin/arm64 installer binary.

This reverts commit 801ece65139d7314fd34144e6b1cefc281f833e9.


staebler (commented):
If this fix requires go1.17, then once again I think we should be updated the minimum go version required to 1.17. Other than that, this looks good to me.

yselkowitz:
For our purposes #5397 is sufficient for this, and only compiling with `GOARCH=arm64` is affected.

staebler:
> For our purposes #5397 is sufficient for this, and only compiling with `GOARCH=arm64` is affected.

I am not only concerned about your purposes, though. Maybe you are indeed the only ones building for arm64. Nevertheless, we are adding in an implicit requirement for go 1.17, we should adjust the explicit requirement accordingly. If you would rather adjust the explicit requirement so that it only comes into effect when the target arch is arm64, then I can accept that, but I expect that it is much easier to just raise the requirement across the board.

yselkowitz:
Unfortunately that's not entirely possible at the moment though, as there is no `docker.io/openshift/origin-release:golang-1.17` to be used by the hack/ scripts.  There is a CI image though; would that work, or does it need to be something publicly available?

staebler:
All I am asking is that the minimum required version set in ./hack/build.sh be changed to go1.17. Is there some reason why docker.io/openshift/origin-release:golang-1.17 is needed for that? We can change the other hack scripts to update the image that they use when such an image is available.

https://github.com/openshift/installer/blob/5473c9334b491bcdb7a1bd10a7e7521f08c04543/hack/build.sh#L8

staebler (approved):
/lgtm
/approve

yselkowitz:
just to quiet the bot
/retest

PR #5404, opened 2021-11-19
hack: include goimports script

bostrt:
This script runs goimports to make sure import sections are consistent.


bostrt:
Should we consider dropping `gofmt` in favor of `goimports`? 

https://pkg.go.dev/golang.org/x/tools/cmd/goimports
> goimports also formats your code in the same style as gofmt

staebler:
What is the motivation for this? Can you provide an example of something that would get caught by this that is not getting caught by hack/go-fmt.sh?

bostrt:
https://github.com/openshift/installer/pull/5376#discussion_r749906017 specifically. 

Gofmt won't reorder imports but goimports will (in addition to formatting other code). Here's an example showing that goimports catches the bad imports ordering and gofmt doesn't note anything wrong:

```
$ head -n 15 main.go
package main

import (
	"context"

	"fmt"
	"github.com/vmware/govmomi"
	"github.com/vmware/govmomi/find"
	"net/url"
	"time"

	"github.com/vmware/govmomi/object"
	"github.com/vmware/govmomi/vim25/soap"
)

$ goimports -d main.go
diff -u main.go.orig main.go
--- main.go.orig	2021-11-24 14:12:13.010454773 -0700
+++ main.go	2021-11-24 14:12:13.010454773 -0700
@@ -4,11 +4,12 @@
 	"context"
 
 	"fmt"
-	"github.com/vmware/govmomi"
-	"github.com/vmware/govmomi/find"
 	"net/url"
 	"time"
 
+	"github.com/vmware/govmomi"
+
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/soap"
 )

$ gofmt -d main.go
<no output>
```

staebler:
Oh yeah, I was getting confused in the samples that I was trying because the script captures all unstaged changes.

staebler:
> Should we consider dropping `gofmt` in favor of `goimports`?
> 
> https://pkg.go.dev/golang.org/x/tools/cmd/goimports
> 
> > goimports also formats your code in the same style as gofmt

Because we use the code-simplification feature of gofmts, goimports is not a drop-in replacement. I am not keen to add another script and test to run in addition to the gofmt ones. I would be more inclined to accept adding a call to goimports in addition to the call to gofmt in the hack/go-fmt.sh script.

PR #5405, opened 2021-11-20
Bug 2025093: Remove The Default Value For The Disk Provisioning Type of Vsphere

ayesha54:
Resolved the following bug by removing the default values for the disk provisioning type of Vsphere.
https://bugzilla.redhat.com/show_bug.cgi?id=2025093

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -2141,8 +2141,14 @@ spec:
                         type: object
                     type: object
                   diskType:
-                    description: Disk Type Thin specifies if thin disks should be
-                      use instead of thick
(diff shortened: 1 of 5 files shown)


jcpowermac:
Based on the test failure @ayesha54 I think you need to change https://github.com/openshift/installer/blob/master/pkg/terraform/exec/plugins/vsphereprivate/resource_vsphereprivate_import_ova.go#L103-L108
to `Required: false`

staebler:
/retitle Bug 2025093: Remove The Default Value For The Disk Provisioning Type of Vsphere

staebler (commented):
Please add details in your commit message regarding why we are removing the default value.

PR #5406, opened 2021-11-22
Alibaba: fix kubebuilder validation

patrickdillon:
DiskCategory validation had string type instead of enum.

On master without this change:
```
$ go generate ./pkg/types/installconfig.go
github.com/openshift/installer/pkg/types:-: conflicting types in allOf branches in schema: string vs DiskCategory
github.com/openshift/installer/pkg/types:-: conflicting types in allOf branches in schema: string vs DiskCategory
github.com/openshift/installer/pkg/types:-: conflicting types in allOf branches in schema: string vs DiskCategory
Error: not all generators ran successfully
run `controller-gen crd:crdVersions=v1 paths=. output:dir=../../data/data/ -w` to see all available markers, or `controller-gen crd:crdVersions=v1 paths=. output:dir=../../data/data/ -h` for usage
exit status 1
pkg/types/installconfig.go:67: running "go": exit status 1

```

### DIFF
--- a/pkg/types/alibabacloud/machinepool.go
+++ b/pkg/types/alibabacloud/machinepool.go
@@ -31,7 +31,6 @@ type MachinePool struct {
 
 	// SystemDiskCategory defines the category of the system disk.
 	//
-	// +kubebuilder:validation:Type=DiskCategory
 	// +optional
 	SystemDiskCategory DiskCategory `json:"systemDiskCategory,omitempty"`
 


kwoodson:
@patrickdillon Thanks for this. I haven't seen this. Am I missing a step when I build my clusters?

pkg/types/alibabacloud/machinepool.go
```diff
@@ -31,7 +31,7 @@ type MachinePool struct {
 
 	// SystemDiskCategory defines the category of the system disk.
 	//
-	// +kubebuilder:validation:Type=DiskCategory
+	// +kubebuilder:validation:Enum="";cloud_efficiency;cloud_essd
```
staebler, line 34: This line should be removed entirely. The validation should be inherited from the validation on the `DiskCategory` type.

patrickdillon:
> @patrickdillon Thanks for this. I haven't seen this. Am I missing a step when I build my clusters?

@kwoodson no this is only an error when generating new code for `openshift-install explain`

pkg/types/alibabacloud/machinepool.go
patrickdillon: Ok I have fixed this. Thanks!

patrickdillon:
We should figure out how this managed to get through CI

clnperez:
i think just 

```
diff --git a/hack/verify-codegen.sh b/hack/verify-codegen.sh
index 58336a517..6db1faf2f 100755
--- a/hack/verify-codegen.sh
+++ b/hack/verify-codegen.sh
@@ -3,7 +3,13 @@
   set -x
+  if [[ $? != 0 ]]; then
+      exit 1
+  fi
   go generate ./pkg/rhcos/ami.go
+  if [[ $? != 0 ]]; then
+      exit 1
+  fi
   set +x
   git diff --exit-code
 else
``` 

would fix up the CI loophole. am happy to submit a PR if no one is already on it

patrickdillon:
> would fix up the CI loophole. am happy to submit a PR if no one is already on it

That would be great. Thank you

clnperez:
submitted that pr, so as soon as this one merges we can merge it

staebler:
> i think just
> 
> ```
> diff --git a/hack/verify-codegen.sh b/hack/verify-codegen.sh
> index 58336a517..6db1faf2f 100755
> --- a/hack/verify-codegen.sh
> +++ b/hack/verify-codegen.sh
> @@ -3,7 +3,13 @@
>  if [ "$IS_CONTAINER" != "" ]; then
>    set -x
>    go generate ./pkg/types/installconfig.go
> +  if [[ $? != 0 ]]; then
> +      exit 1
> +  fi
>    go generate ./pkg/rhcos/ami.go
> +  if [[ $? != 0 ]]; then
> +      exit 1
> +  fi
>    set +x
>    git diff --exit-code
>  else
> ```
> 
> would fix up the CI loophole. am happy to submit a PR if no one is already on it

Or just remove the `set -x` and `set +x`. But this would also require fixing the fact that the `./pkg/rhcos/ami.go` no longer exists. See https://issues.redhat.com/browse/CORS-1838.

PR #5407, opened 2021-11-22
Bug 1968364: azure: fix upi bug preventing using ed25519 keys

squidboylan:
This updates the azure UPI arm templates to not add the ssh keys.
The azure API does not support using ed25519 ssh keys which
caused the deployment to fail when using ed25519 keys, we are already
providing ssh keys in the ignition so we can bypass this limitation.

fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1968364

### DIFF
--- a/upi/azure/04_bootstrap.json
+++ b/upi/azure/04_bootstrap.json
@@ -18,8 +18,9 @@
     },
     "sshKeyData" : {
       "type" : "securestring",
+      "defaultValue" : "Unused",
       "metadata" : {
-        "description" : "SSH RSA public key file as a string."
+        "description" : "Unused"
       }
     },
     "bootstrapVMSize" : {
@@ -184,17 +185,10 @@
         "osProfile" : {
           "computerName" : "[variables('vmName')]",
           "adminUsername" : "core",
+          "adminPassword" : "NotActuallyApplied!",
           "customData" : "[parameters('bootstrapIgnition')]",
           "linuxConfiguration" : {
-            "disablePasswordAuthentication" : true,
-            "ssh" : {
-              "publicKeys" : [
-                {
-                  "path" : "[variables('sshKeyPath')]",
-                  "keyData" : "[parameters('sshKeyData')]"
-                }
-              ]
-            }
+            "disablePasswordAuthentication" : false
           }
         },
         "storageProfile" : {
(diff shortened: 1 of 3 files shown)


MayXuQQ:
Removing  "sshKeyData" will cause the job to fail for earlier OpenShift versions that still require the parameter.
That change will only affect jobs running against the master branch of the installer. This template is run against other branches of the installer as well. -- @staebler
ref: https://github.com/openshift/release/blob/311617f07117e909073c6ce7f01c8e7b188fcc2d/ci-operator/templates/openshift/installer/cluster-launch-installer-upi-e2e.yaml#L2089
suggest  :
"sshKeyData" : {
      "type" : "securestring",
      "default": "",
      "metadata" : {
        "description" : "unused"
      }
    },

/assign @staebler

upi/azure/04_bootstrap.json
```diff
@@ -16,12 +16,6 @@
         "description" : "Bootstrap ignition content for the bootstrap cluster"
       }
     },
-    "sshKeyData" : {
```
staebler, line 19: Unfortunately, I think you are going to need to keep these parameters. The CI jobs will need to keep passing in the parameter so that we can support earlier OpenShift versions that require the parameter. You can provide a default value so that users do not have to provide the parameter. And you should change the description to "unused".

staebler (commented):
Looks good other than needing to keep the `sshKeyData` parameters.

upi/azure/04_bootstrap.json
staebler: Which I know see that @MayXuQQ pointed out the same up above.

squidboylan: That makes sense, thanks, i've updated the PR

upi/azure/04_bootstrap.json
```diff
@@ -19,7 +19,7 @@
     "sshKeyData" : {
       "type" : "securestring",
       "metadata" : {
-        "description" : "SSH RSA public key file as a string."
+        "description" : "Unused"
```
staebler, line 22: These should have a default value specified as well so that users do not have to supply a dummy value as so that we can eventually remove the parameter once support for older versions make their way out of CI.

staebler (approved):
/lgtm
/approve

PR #5410, opened 2021-11-23
Alibaba: fix: remove the default value

bd233:
Remove the default value in terraform variables, because the root module
and tfvars file should always be  explicitly supplying a value.

### DIFF
--- a/data/data/alibabacloud/cluster/master/variables.tf
+++ b/data/data/alibabacloud/cluster/master/variables.tf
@@ -42,8 +42,7 @@ variable "system_disk_size" {
 
 variable "system_disk_category" {
   type        = string
-  description = "The system disk category of the master ECS.Valid values are cloud_efficiency, cloud_ssd, cloud_essd. Default value is cloud_essd."
-  default     = "cloud_essd"
+  description = "The system disk category of the master ECS. Valid values are cloud_efficiency, cloud_ssd, cloud_essd."
 }
 
 variable "role_name" {
@@ -57,6 +56,5 @@ variable "user_data_ign" {
 
 variable "tags" {
   type        = map(string)
-  default     = {}
   description = "Tags to be applied to created resources."
 }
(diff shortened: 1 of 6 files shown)


data/data/alibabacloud/cluster/master/variables.tf
```diff
@@ -43,7 +43,6 @@ variable "system_disk_size" {
 variable "system_disk_category" {
   type        = string
   description = "The system disk category of the master ECS.Valid values are cloud_efficiency, cloud_ssd, cloud_essd. Default value is cloud_essd."
```
staebler, line 45:
```suggestion
  description = "The system disk category of the master ECS. Valid values are cloud_efficiency, cloud_ssd, cloud_essd."
```

data/data/alibabacloud/variables-alibabacloud.tf
```diff
@@ -48,7 +48,6 @@ variable "ali_system_disk_size" {
 variable "ali_system_disk_category" {
   type        = string
   description = "The system disk category of the master ECS.Valid values are cloud_efficiency, cloud_ssd, cloud_essd. Default value is cloud_essd."
```
staebler, line 50:
```suggestion
  description = "The system disk category of the master ECS. Valid values are cloud_efficiency, cloud_ssd, cloud_essd."
```

data/data/alibabacloud/variables-alibabacloud.tf
```diff
@@ -59,7 +58,6 @@ variable "ali_extra_tags" {
 
 Example: `{ "key" = "value", "foo" = "bar" }`
 EOF
-  default = {}
```
staebler, line 62: Removing this default also requires removing the `omitempty` from the field in the tfvars struct.
https://github.com/openshift/installer/blob/c1c9b61f0c9b568af9d9ecd825a221e9f5b041f7/pkg/tfvars/alibabacloud/alibabacloud.go#L30

bd233:
@staebler Have updated.

staebler (commented):
Looks good to me.
/approve

kwoodson:
This is pretty straightforward.
/ok-to-test
/lgtm

PR #5411, opened 2021-11-23
Alibaba: fix: use EXACT search mode

bd233:
The `LIKE` is default search mode on DescribeDomains api, In this mode,
if you search `openshift.com`, can match `xx.openshift.com` and
`openshift.com.xx`, but using `EXACT` mode, only return `openshift.com`.


bd233:
@kwoodson @patrickdillon  This solves the same problem as [this](https://github.com/openshift/installer/pull/5348#issuecomment-973095117)

staebler (commented):
Looks good to me, but I cannot test this.

/approve

kwoodson:
@bd233 I tested this along with #5421. This did work as intended. Please address feedback in that PR. Thanks!

/lgtm

PR #5413, opened 2021-11-24
Alibaba: fix: Use master ip list to replace map

bd233:
Using `{<master name>: <ip>}` map is a redundant operation, using IP list is simpler


staebler (commented):
This looks good to me. But I have no way to verify the changes. @kwoodson can you verify this change?

/approve

bd233:
@staebler Unfortunately, my test found some problems, `length(var.master_ips)` does not work properly, I updated it again. @kwoodson  Sorry, please use the latest submission test.

PR #5415, opened 2021-11-24
Alibaba: fix: update comment on ResourceGroupID

bd233:
In fact, it has no effect if it is not empty. we do not query and destroy resources through resource group. So this description is incorrect

### DIFF
--- a/pkg/types/alibabacloud/platform.go
+++ b/pkg/types/alibabacloud/platform.go
@@ -6,9 +6,7 @@ type Platform struct {
 	Region string `json:"region"`
 
 	// ResourceGroupID is the ID of an already existing resource group where the cluster should be installed.
-	// This resource group must be empty with no other resources when trying to use it for creating a cluster.
-	// If empty, a new resource group will created for the cluster.
-	// Destroying the cluster using installer will delete this resource group.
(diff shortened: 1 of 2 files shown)


staebler (commented):
Need to run `go generate ./pkg/types`. But that depends on #5406.

pkg/types/alibabacloud/platform.go
```diff
@@ -6,9 +6,7 @@ type Platform struct {
 	Region string `json:"region"`
 
 	// ResourceGroupID is the ID of an already existing resource group where the cluster should be installed.
-	// This resource group must be empty with no other resources when trying to use it for creating a cluster.
 	// If empty, a new resource group will created for the cluster.
```
staebler, line 9: Let's fix this up this grammar while we are here.
```suggestion
	// If empty, the installer will create a new resource group for the cluster.
```

bd233: Okey, updated

staebler (approved):
/lgtm
/approve

staebler:
Need to run `go generate ./pkg/types` again.
/lgtm cancel

PR #5418, opened 2021-11-24
installer-artifacts: add darwin/arm64 binary

yselkowitz:
This will require a simultaneous commit to ocp-build-data, so this should not be merged until both are ready.


yselkowitz:
Pending ocp-build-data PR
/hold

staebler:
This PR needs context for the motivation behind this addition.

yselkowitz:
Apple's latest hardware is moving to their M1 aarch64 processors.  Currently we provide a darwin/amd64 binary, but openshift-install is too large for Rosetta 2 to translate effectively or at all.  Now that installer is using go1.17 which fixed the arm64 static linking bug (which affected both linux/arm64 and darwin/arm64), we can provide a darwin/arm64 binary which will allow MacOS users with the newest hardware to run installer too.  This is the first step in that process.

staebler (commented):
/lgtm
/approve

PR #5421, opened 2021-11-26
Alibaba: fix: destroy the records of the current cluster

bd233:
Filter the resolution records of the base domain through `api.<cluster name>`,
otherwise it will cause all records to be released.

### DIFF
--- a/pkg/destroy/alibabacloud/alibabacloud.go
+++ b/pkg/destroy/alibabacloud/alibabacloud.go
@@ -1146,19 +1146,14 @@ func (o *ClusterUninstaller) detachPolicy(policyName string, policyType string, 
 
 func (o *ClusterUninstaller) deletePrivateZones() (err error) {
 	clusterDomain := o.ClusterDomain
-	o.Logger.Debug("Start to search private zones")
-	zones, err := o.listPrivateZone(clusterDomain)
+	zoneID, err := o.getPrivateZoneID()
 	if err != nil {
 		return err
 	}
-	if len(zones) == 0 {
+	if zoneID == "" {
 		return nil
 	}
-	if len(zones) > 1 {
-		return errors.Wrap(err, fmt.Sprintf("matched to multiple private zones by clustedomain %q", clusterDomain))
-	}
 
-	zoneID := zones[0].ZoneId
 	err = o.bindZoneVpc(zoneID)
 	if err != nil {
 		return err
@@ -1214,6 +1209,23 @@ func (o *ClusterUninstaller) deletePrivateZones() (err error) {
 	return nil
 }
 
+func (o *ClusterUninstaller) getPrivateZoneID() (zoneID string, err error) {
+	clusterDomain := o.ClusterDomain
+	o.Logger.Debug("Start to search private zones")
+	zones, err := o.listPrivateZone(clusterDomain)
+	if err != nil {
+		return "", err
+	}
+	if len(zones) == 0 {
+		return "", nil
+	}
+	if len(zones) > 1 {
+		return "", errors.Wrap(err, fmt.Sprintf("matched to multiple private zones by clusterdomain %q", clusterDomain))
+	}
+
+	return zones[0].ZoneId, nil
+}
+
 func (o *ClusterUninstaller) deletePrivateZone(zoneID string) (err error) {
 	o.Logger.Debugf("Start to delete private zone %q", zoneID)
 	request := pvtz.CreateDeleteZoneRequest()
@@ -1243,11 +1255,29 @@ func (o *ClusterUninstaller) listPrivateZone(clusterDomain string) ([]pvtz.Zone,
 	return response.Zones.Zone, nil
 }
 
+func (o *ClusterUninstaller) listPrivateZoneRecords(zoneID string) ([]pvtz.Record, error) {
+	request := pvtz.CreateDescribeZoneRecordsRequest()
+	request.Lang = "en"
+	request.ZoneId = zoneID
+
+	response, err := o.pvtzClient.DescribeZoneRecords(request)
+	if err != nil {
+		return nil, err
+	}
+	return response.Records.Record, nil
+}
+
 func (o *ClusterUninstaller) deleteDNSRecords() (err error) {
 	o.Logger.Debug("Start to search DNS records")
 
 	// Get the base domain from the cluster domain. the format of cluster domain is '<cluster name>.<base domain>'.
-	baseDomain := strings.Join(strings.Split(o.ClusterDomain, ".")[1:], ".")
+	domainParts := strings.Split(o.ClusterDomain, ".")
+	if len(domainParts) < 2 {
+		return errors.New("could not determine cluster name from cluster domain")
+	}
+	clusterName := domainParts[0]
+	baseDomain := strings.Join(domainParts[1:], ".")
+
 	domains, err := o.listDomain(baseDomain)
 	if err != nil {
 		return
@@ -1256,21 +1286,50 @@ func (o *ClusterUninstaller) deleteDNSRecords() (err error) {
 		return
 	}
 
-	records, err := o.listRecord(baseDomain)
+	recordSetKey := func(recordType string, rr string) string {
+		return fmt.Sprintf("%s %s", recordType, rr)
+	}
+
+	// Get the parsing record of privatezone and delete the record in publiczone
+	// When the user manually deletes the private zone and records, it may cause the public records to be leaked.
+	privateZoneID, err := o.getPrivateZoneID()
 	if err != nil {
 		return
 	}
-	if len(records) == 0 {
+	if privateZoneID == "" {
+		o.Logger.Info("The private zone ID is not obtained, and the DNS public records cannot be deleted")
+		return nil
+	}
+
+	privateRecords := map[string]bool{}
+	records, err := o.listPrivateZoneRecords(privateZoneID)
+	if err != nil {
+		return
+	}
+	for _, record := range records {
+		key := recordSetKey(record.Type, fmt.Sprintf("%s.%s", record.Rr, clusterName))
+		privateRecords[key] = true
+	}
+
+	publicRecords, err := o.listRecord(baseDomain)
+	if err != nil {
+		return
+	}
+	if len(publicRecords) == 0 {
 		return
 	}
 
+	var lastErr error
 	o.Logger.Debug("Start to delete DNS records")
-	for _, record := range records {
-		err = o.deleteRecord(record.RecordId)
-		if err != nil {
-			err = errors.Wrap(err, fmt.Sprintf("DNS record %q", record.RecordId))
-			o.Logger.Info(err)
-			return
+	for _, record := range publicRecords {
+		key := recordSetKey(record.Type, record.RR)
+		if privateRecords[key] {
+			err = o.deleteRecord(record.RecordId)
+			if err != nil {
+				privateRecords[key] = false
+				lastErr = errors.Wrap(err, fmt.Sprintf("DNS record %q", record.RecordId))
+				o.Logger.Info(lastErr)
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/destroy/alibabacloud/alibabacloud.go
```diff
@@ -1248,6 +1248,7 @@ func (o *ClusterUninstaller) deleteDNSRecords() (err error) {
 
 	// Get the base domain from the cluster domain. the format of cluster domain is '<cluster name>.<base domain>'.
 	baseDomain := strings.Join(strings.Split(o.ClusterDomain, ".")[1:], ".")
+	rr := fmt.Sprintf("api.%s", strings.Split(o.ClusterDomain, ".")[0])
```
mtulio, line 1251: You could consider using a structured logger here. See the example on the destroy flow for Route53 in AWS:
https://github.com/openshift/installer/blob/06e3fe64f3616f6d6baff95dde5cd3389cf785aa/pkg/destroy/aws/aws.go#L1824-L1832

The `logger` for this function is being used in every function that will call, tracking the id of hosted zone.

mtulio:
``` go
rr := fmt.Sprintf("api.%s", strings.Split(o.ClusterDomain, ".")[0])
```
@kwoodson @staebler I am also not sure if we should delete all RRs with a suffix of the `o.ClusterDomain`, rather than only the `api.*`, I guess we will have a leak of RR `*.apps.(o.ClusterDomain)` using this approach. wdyt?

kwoodson: @bd233 I tested this code and the code from the #5411.

The domain I was using was `*.apps.test` (created by the ingress controller) as well as `api.test` (created by the installer). I created another domain to verify that it was not removed (`*.apps.test1` and `api.test1`). When calling the `destroy` cluster I verified that `api.test` was removed. This is an improvement but I would also like to see the record from the ingress removed `*.apps.test`. We do not want to leak these resource records.

I *think* we will want to remove all resource records with the `o.ClusterDomain`. That should be relatively safe since the cluster domain should be unique.

mtulio:
@bd233 ptal on those comments above?:
1. We should be using structured logs (as [commented on slack](https://coreos.slack.com/archives/C01PES793LL/p1637815446441600?thread_ts=1637793763.435500&cid=C01PES793LL))
2. We need to make sure that all RRs for the cluster must be removed when the cluster is destroyed. In this approach, the RR `*.apps` is leaking.

Thanks!

pkg/destroy/alibabacloud/alibabacloud.go
staebler: Not exactly. We cannot guarantee that the cluster domain is unique. What we do for AWS, which I recommend following here as well, is that we only delete the public DNS record if there is also a private DNS record for the same domain.
See https://github.com/openshift/installer/blob/d3736c2e9e10b0b8e5a0ecbf666ff2f739a02683/pkg/destroy/aws/aws.go#L1832-L1865

bd233:
> @bd233 ptal on those comments above?:
> 
> 1. We should be using structured logs (as [commented on slack](https://coreos.slack.com/archives/C01PES793LL/p1637815446441600?thread_ts=1637793763.435500&cid=C01PES793LL))
> 2. We need to make sure that all RRs for the cluster must be removed when the cluster is destroyed. In this approach, the RR `*.apps` is leaking.
> 
> Thanks!

Okey, thanks, I will update it as soon as possible.

pkg/destroy/alibabacloud/alibabacloud.go
bd233: Sorry, I don't understand why the cluster domain is not unique. Can you provide more details. Can users create two clusters with the same cluster name and base domain (I think this is not feasible)? Or consider that the user may manually add the records containing the cluster domain？

staebler: The installer does not know about any other clusters that exist. So the installer cannot directly verify that someone has not attempted to install a cluster with the exact same cluster domain. Moreover, the destroyer must be resilient to any changes that have been made in the cloud account in between when the installation attempt occurred and when the destroyer is called. The destroyer should always err on the side of caution because the worst thing that it can do is to delete a resource that did not actually belong to the cluster.

Let's look at an example. The user attempts to create a cluster with the cluster domain test.example.com. The installation failed at some point before creating the public DNS records. The next day, the user fixes whatever was causing the installation to fail the first time and tries another installation with the same cluster domain. This installation was successful. The next day, the user realizes that there are some leftover resources from the first installation attempt because the user forgot to destroy the cluster. The user runs the destroyer on the partial cluster from the first attempt. The destroyer sees that there is a *.apps.test.example.com public DNS record, so the destroyer deletes it. Now the user cannot access the apps on their working cluster because the public DNS record has been deleted.

bd233:
@staebler @mtulio I submitted the changes to formatting the log on #5435 PR. If it meets expectations, I will continue to modify all logs.

kwoodson:
I'll test this PR this morning when removing my cluster.

mtulio:
@bd233 : @kwoodson and I tested this today and it's deleting all the records from the public zone with suffix of `clusterName`. It is not the expected behavior because we could have RRs created by any other component outside of OpenShift (example: manual) that is being deleted.

In my tests I create on the console one RR `mrb.test` on the public zone `alicloud-dev.devcluster.openshift.com` and it has been deleted when we triggered the destroy flow.

Please review the suggestion above of @staebler what is done in AWS in [this comment](https://github.com/openshift/installer/pull/5421#discussion_r759489684).

Could you review it? Thanks!

bd233:
> Please review the suggestion above of @staebler what is done in AWS in this comment.

@mtulio @staebler @kwoodson  I have updated the deletion strategy according to the suggestion, please review it.

mtulio:
@bd233 , I just test this PR and the DNS records created by the installer were successfully removed from the public DNS zone, alongside the private zone.

```
DEBUG Start to search DNS records    
DEBUG Start to search private zones                
DEBUG Start to delete DNS records                  
DEBUG Start to delete DNS record "734306694439286784" 
DEBUG Start to search private zones                
DEBUG Start to delete private zones                
DEBUG Start to delete private zone "550502d3a7eda57b049634b699a553f6"
```

In my tests, my cluster was not successfully installed, for that reason only one record was removed (the `api.${clusterName}` one). FYI @kwoodson 

I also would like to mention that the output will be improved when the PR https://github.com/openshift/installer/pull/5435 will be merged.
/lgtm

mtulio:
@kwoodson @staebler @patrickdillon ptal?

staebler (commented):
Shoot. Forgot to press the "Submit review" button. Sorry for the delay.

mtulio:
/lgtm cancel
@bd233 could you please take a look at the @staebler comments above? I can run new tests when it's ready. Thanks!

kwoodson:
@bd233 Please update this PR. I believe this is a small fix. Thanks!

pkg/destroy/alibabacloud/alibabacloud.go
```diff
@@ -1256,21 +1278,43 @@ func (o *ClusterUninstaller) deleteDNSRecords() (err error) {
...
+	records, err := o.listPrivateZoneRecords(privateZoneID)
+	if err != nil {
+		return
+	}
+	for _, record := range records {
+		key := recordSetKey(record.Type, fmt.Sprintf("%s.%s", record.Rr, clusterName))
+		privateRecords[key] = true
+	}
+
+	publicRecords, err := o.listRecord(baseDomain)
```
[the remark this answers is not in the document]
bd233: The default page size is 20, and the SDK will not return all results. If this has caused a problem, I can temporarily increase the page size, such as 50. This is not a good solution, I will create a new PR to add the function of automatically obtaining all paging to `alibabacloud.client`.

pkg/destroy/alibabacloud/alibabacloud.go
```diff
@@ -1214,6 +1206,23 @@ func (o *ClusterUninstaller) deletePrivateZones() (err error) {
...
 
+func (o *ClusterUninstaller) getPrivateZoneID() (zoneID string, err error) {
+	clusterDomain := o.ClusterDomain
+	o.Logger.Debug("Start to search private zones")
+	zones, err := o.listPrivateZone(clusterDomain)
+	if err != nil {
+		return "", err
+	}
+	if len(zones) == 0 {
+		return "", errors.Wrap(err, fmt.Sprintf("cannot found privatezone using cluster domain %q", clusterDomain))
```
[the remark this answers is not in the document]
bd233: Thank you. At the same time, I realized a problem. Determine whether to delete the public analysis record through the analysis record of privateZone, but the privateZone or its analysis record may not exist (for example, the user manually deleted it). I am referring to AWS to try to get inspiration.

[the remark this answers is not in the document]
staebler: If the user manually deletes the private zone or the analysis record, then the public record will be leaked. There is no way around that. The same thing would happen with AWS.

pkg/destroy/alibabacloud/alibabacloud.go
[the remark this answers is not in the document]
staebler: I do not think that the paging issue is limited to just this call. We need to handle the possibility of paging in *all* of the calls that we make to the SDK.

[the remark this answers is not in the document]
staebler: But, yes, let's move that to a separate PR.

pkg/destroy/alibabacloud/alibabacloud.go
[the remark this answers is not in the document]
bd233: OK...I have updated.

pkg/destroy/alibabacloud/alibabacloud.go
```diff
@@ -1214,6 +1209,23 @@ func (o *ClusterUninstaller) deletePrivateZones() (err error) {
...
+	o.Logger.Debug("Start to search private zones")
+	zones, err := o.listPrivateZone(clusterDomain)
+	if err != nil {
+		return "", err
+	}
+	if len(zones) == 0 {
+		return "", nil
+	}
+	if len(zones) > 1 {
+		return "", errors.Wrap(err, fmt.Sprintf("matched to multiple private zones by clustedomain %q", clusterDomain))
```
staebler, line 1223:
```suggestion
		return "", errors.Wrap(err, fmt.Sprintf("matched to multiple private zones by clusterdomain %q", clusterDomain))
```

pkg/destroy/alibabacloud/alibabacloud.go
```diff
@@ -1243,11 +1255,24 @@ func (o *ClusterUninstaller) listPrivateZone(clusterDomain string) ([]pvtz.Zone,
...
+	}
+	return response.Records.Record, nil
+}
+
 func (o *ClusterUninstaller) deleteDNSRecords() (err error) {
 	o.Logger.Debug("Start to search DNS records")
 
 	// Get the base domain from the cluster domain. the format of cluster domain is '<cluster name>.<base domain>'.
 	baseDomain := strings.Join(strings.Split(o.ClusterDomain, ".")[1:], ".")
+	clusterName := strings.Split(o.ClusterDomain, ".")[0]
```
staebler, line 1275:
```suggestion
	domainParts := strings.Split(o.ClusterDomain, ".", 1)
	if len(domainParts) < 2 {
		return errors.New("could not determine cluster name from cluster domain")
	}
	clusterName := domainParts[0]
	baseDomain := domainParts[1]
```

pkg/destroy/alibabacloud/alibabacloud.go
```diff
@@ -1256,21 +1281,48 @@ func (o *ClusterUninstaller) deleteDNSRecords() (err error) {
...
-		if err != nil {
-			err = errors.Wrap(err, fmt.Sprintf("DNS record %q", record.RecordId))
-			o.Logger.Info(err)
-			return
+	for _, record := range publicRecords {
+		key := recordSetKey(record.Type, record.RR)
+		if privateRecords[key] {
+			err = o.deleteRecord(record.RecordId)
+			if err != nil {
+				err = errors.Wrap(err, fmt.Sprintf("DNS record %q", record.RecordId))
```
staebler, line 1323: When there is an error deleting the record, remove the key from `privateRecords` so that we don't wait the full minute for the record to be deleted. But also capture that there was an error so that `deleteDNSRecords` returns an error later after the wait.

bd233: Have updated^^

staebler (commented):
/lgtm
/approve

PR #5422, opened 2021-11-26
Alibaba: fix: remove the master node records

bd233:
The `<master><n>.<cluster name>.<base domain>` records is not needed on
 base domain


staebler (commented):
/approve
/cc @kwoodson

kwoodson:
@staebler Thanks for the mention. I tested this yesterday and verified that the cluster installed successfully without using the DNS.  
image

/lgtm

kwoodson:
*sigh* 
We need to skip unrelated tests when testing specific Alibaba terraform code. This compounds across multiple repositories.

PR #5424, opened 2021-11-26
Bug 2027329: Handle unset default machine pool

stephenfin:
If the user has explicitly provided `controlPlane` and `compute` configuration, then there won't be a default to consider.

This was hit in CI once `platform.openstack.computeFlavor` was no longer defined in a generated `install-config.yaml` file. This had been masking the bug since setting that attribute resulted in us initializing `config.Platform.OpenStack.DefaultMachinePlatform` with an `openstack.MachinePool` instance and populating the `FlavorName` attribute of this struct (see `pkg/types/conversion/installconfig.go` for more info).

/label platform/openstack


pierreprinetti:
/approve
/lgtm
:+1:

mandre (commented):
/lgtm
/label backport-risk-assessed

PR #5425, opened 2021-11-27
[Baremetal] Support For Deploying with ISOs for Baremetal IPI

imain:
Switch to using an ISO for deployments. Use image-customization-server
to serve up custom ignition settings.  Set up custom deploy steps for
RHCOS imaging.

Marked as work in progress because I'm sure there's still more clean up
to do.

Co-Authored-By: Kiran Thyagaraja <kthyagar@redhat.com>
Co-Authored-By: Caleb Boylan <cboylan@redhat.com>


imain:
EOD Friday here.  Putting this up so people can see where we're at.  I'll fix it up Monday :)

dtantsur:
Looks okay at first glance, will review in depth once ready.

imain:
OK, so there are go mod issues here.  Zane and/or I are going to post a patch with just the switch to a newer openshift/cluster-api-provider-baremetal change.

zaneb (commented):
There's a few different logical changes here and splitting them into separate commits might help with reviewing, but also with potentially reverting later:

* Run image customisation server and pass URLs to Terraform
* Use deploy_steps deployment method with Terraform
* Add ISO/kernel/initramfs/rootfs preProvisioningOSDownloadURLs to Provisioning CR (we _may_ or may not use this for testing, but if we do we will probably revert later)
* Add ISO/kernel/initramfs/rootfs URLs to installer config - this plan has been superseded and we don't want to do this at all

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -7,23 +7,26 @@ IRONIC_IMAGE=$(image_for ironic)
...
 DHCP_RANGE="{{.PlatformData.BareMetal.ProvisioningDHCPRange}}"
 
 # Used by ironic to allow ssh to running IPA instances
-IRONIC_RAMDISK_SSH_KEY="{{.SSHKey}}"
+IRONIC_RAMDISK_SSH_KEY="$(echo -n "{{.SSHKey}}")"
 
 # First we stop any previously started containers, because ExecStop only runs when the ExecStart process
 # e.g this script is still running, but we exit if *any* of the containers exits unexpectedly
-for name in ironic-api ironic-conductor ironic-inspector ironic-ramdisk-logs dnsmasq httpd mariadb ipa-downloader coreos-downloader; do
+for name in ironic-api ironic-conductor ironic-inspector ironic-ramdisk-logs dnsmasq httpd mariadb ipa-downloader coreos-downloader image-customization; do
```
hardys, line 23: Can we remove ipa-downloader and the related `IPA_DOWNLOADER_IMAGE` above?

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -152,11 +150,31 @@ $IPTABLES -t raw -A DHCP_IRONIC -j DROP
...
+
+# Write master network config to /tmp/nmstate for image_customization_controller to serve our merged images
+mkdir -p /tmp/nmstate
+{{range .PlatformData.BareMetal.Hosts}}
+    cat >/tmp/nmstate/{{.Name}}.yaml <<EOF
+{{.NetworkConfig}}
+EOF
+{{end}}
+
+CUSTOMIZATION_IMAGE="virthost.ostest.test.metalkube.org:5000/localimages/image-customization-controller"
```
hardys, line 164: This needs to be referenced via image_for like IPA_DOWNLOADER_IMAGE etc at the start of the script

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -215,8 +234,8 @@ while true; do
         sleep 10
       done
 
-      echo "Stopping provisioning services..."
-      podman stop ironic-api ironic-conductor ironic-inspector ironic-ramdisk-logs $dnsmasq_container_name httpd mariadb
+      echo "PRETENDING to Stopping provisioning services..."
+      #podman stop ironic-api ironic-conductor ironic-inspector ironic-ramdisk-logs $dnsmasq_container_name httpd mariadb
```
hardys, line 238: Presumably this is disabled for debugging, but when re-enabled we can probably stop the image-customization container here too?

pkg/asset/ignition/bootstrap/baremetal/template.go
```diff
@@ -46,12 +46,23 @@ type TemplateData struct {
 
 	// BaremetalIntrospectionEndpointOverride contains the url for the baremetal introspection endpoint
 	BaremetalIntrospectionEndpointOverride string
+
+	// ClusterOSImage contains 4 URLs to download RHCOS live iso, kernel, rootfs and initramfs
+	ClusterOSImage string
+
+        // API VIP for use by ironic during bootstrap.
+        APIVIP string
```
hardys, line 54: Whitespace issue here

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -152,11 +150,31 @@ $IPTABLES -t raw -A DHCP_IRONIC -j DROP
...
 podman wait -i 1000ms coreos-downloader
-while ! curl --fail http://localhost/images/rhcos-ootpa-latest.qcow2.md5sum ; do sleep 1; done
-while ! curl --fail --head http://localhost/images/ironic-python-agent.initramfs ; do sleep 1; done
-while ! curl --fail --head http://localhost/images/ironic-python-agent.kernel ; do sleep 1; done
+
+# Write master network config to /tmp/nmstate for image_customization_controller to serve our merged images
+mkdir -p /tmp/nmstate
+{{range .PlatformData.BareMetal.Hosts}}
+    cat >/tmp/nmstate/{{.Name}}.yaml <<EOF
+{{.NetworkConfig}}
```
hardys, line 160: I think we'll need to create secrets from the NetworkConfig, similar to how BMC credentials are handled:

https://github.com/openshift/installer/blob/master/pkg/asset/machines/baremetal/hosts.go#L29

That also means we'll have to consumer the manifest (not the install-config) data, since in theory a user can modify the manifests after running `create manifests` prior to running `create cluster`

So I'm wondering if we can just consume the secrets here (the manifests will get copied onto the bootstrap VM via ignition, like the BMH and BMC-secret manifests)

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -152,11 +150,31 @@ $IPTABLES -t raw -A DHCP_IRONIC -j DROP
...
+EOF
+{{end}}
+
+CUSTOMIZATION_IMAGE="virthost.ostest.test.metalkube.org:5000/localimages/image-customization-controller"
+
+# Embed agent ignition into the rhcos live iso
+sudo podman run -d --net host --privileged --name image-customization \
+    --env DEPLOY_ISO="/shared/html/images/ironic-python-agent.iso" \
```
hardys, line 168: Do we need net=host and privileged here?

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
imain: Is that not what is happening here? https://github.com/openshift/installer/pull/5247/files

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -152,11 +150,31 @@ $IPTABLES -t raw -A DHCP_IRONIC -j DROP
...
+    --env IRONIC_BASE_URL="http://{{.PlatformData.BareMetal.APIVIP}}" \
+    --env IRONIC_AGENT_IMAGE="" \
+    --env IRONIC_AGENT_PULL_SECRET="" \
```
zaneb, line 171: We need to set these to stop it from pulling from @dtantsur's quay.io account, which is temporarily the default.

I had a go at this [here](https://github.com/zaneb/openshift-installer/commit/24cf61e62c5371cd9b5817f2d98b36be5a3f33a1) but it turns out that, at least in dev-scripts, the pull secret we have is for the mirror (virthost.ostest.test.metalkube.org:5000). So I think we need to add the file https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/etc/containers/registries.conf.template to the ignition in image-customization-controller.
I'm not clear on how this will work once we are running in the cluster. Presumably for disconnected installs at least there will be something similar. Where do we get the mirror list from in that case? @hardys, do you know?

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
hardys: Ah thanks I missed that, but I think the point remains that in startironic.sh we should consume those manifests, not the install-config data directly (I guess that could be addressed as a follow-up, but if we want to fix that now, this PR needs to be rebased onto https://github.com/openshift/installer/pull/5247 and the /tmp/nmstate stuff needs to consume the generated secrets)

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
zaneb: https://github.com/openshift/image-customization-controller/pull/20 will enable us to embed the registries.conf file.

hardys: @zaneb when the image exists in the release payload we should just be able to use `image_for` like the other images at the top of this script - we shouldn't need to care about disconnected or not.

We pass a reference to the virthost registry via [imageContentSources in the install-config](https://github.com/openshift-metal3/dev-scripts/blob/master/utils.sh#L382), and that ends up templated into registries.conf already.

hardys: More info after re-reading the question :)  The same registries.conf content ends up on the running cluster:

```
[root@master-0 ~]# cat /etc/containers/registries.conf
unqualified-search-registries = ["registry.access.redhat.com", "docker.io"]
short-name-mode = ""

[[registry]]
  prefix = ""
  location = "quay.io/openshift-release-dev/ocp-v4.0-art-dev"
  mirror-by-digest-only = true

  [[registry.mirror]]
    location = "virthost.ostest.test.metalkube.org:5000/localimages/local-release-image"

[[registry]]
  prefix = ""
  location = "registry.ci.openshift.org/ocp/release"
  mirror-by-digest-only = true

  [[registry.mirror]]
    location = "virthost.ostest.test.metalkube.org:5000/localimages/local-release-image"
```

That gets created via ignition, from the MCS rendered content:

```
    "filesystem": "root",
    "overwrite": true,
    "path": "/etc/containers/registries.conf",
    "contents": {
...
```
Looks like that comes from the `01-master-container-runtime` MachineConfig object (not entirely sure how that's populated though, I'm guessing it copies the bootstrap file since I don't see it populated via installer manifests).

zaneb: Ah, if you override images then they point to virthost.ostest.test.metalkube.org (which works), so that is possibly how we tested.
If you don't then they point to teh Internets, so we'll need to pass the registries.conf file in the IPA Ignition in the cluster as well as on the bootstrap.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
imain: yup

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
imain: hrrm, is this part of the release?  I've been building it by hand and sticking it in the local repo to pull here.. :/

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
imain: Gotcha.  I'll do that. I see it merged now.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
imain: hah! probably not.. I'm so used to it now :).

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
zaneb: It's not part of the release yet, but it should be in the next couple of days I hope.

data/data/manifests/openshift/baremetal-provisioning-config.yaml.template
```diff
@@ -10,3 +10,8 @@ spec:
   provisioningDHCPRange: "{{.Baremetal.ProvisioningDHCPRange}}"
   provisioningOSDownloadURL: "{{.ProvisioningOSDownloadURL}}"
   watchAllNamespaces: false
+  preProvisioningOSDownloadURLs:
+    initramfsURL: "{{.InitramfsURL}}"
+    isoURL: "{{.IsoURL}}"
+    kernelURL: "{{.KernelURL}}"
+    rootfsURL: "{{.RootfsURL}}"
```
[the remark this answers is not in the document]
sadasu: I think it is safe to remove this as part of the cleanup activity for this PR.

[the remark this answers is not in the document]
zaneb: @sadasu on reflection, I think this is needed for us to test your CBO PR. We'll likely want to do a separate PR to switch to machine-os-images. So I'd vote to keep this in for now.

imain:
Didn't get as far as I had hoped.  This is most of the cleanups.  Will split the patch and use secrets for network data tomorrow.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -152,11 +150,31 @@ $IPTABLES -t raw -A DHCP_IRONIC -j DROP
 {{end}}
 
 # Wait for images to be downloaded/ready
-podman wait -i 1000ms ipa-downloader
+#podman wait -i 1000ms ipa-downloader
```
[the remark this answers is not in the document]
imain: I think I got all these

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
imain: Added a placeholder for now.  I'm still testing the patch after changes.

data/data/manifests/openshift/baremetal-provisioning-config.yaml.template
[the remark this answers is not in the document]
imain: I'll make it a separate patch.

zaneb (commented):
Thanks, splitting it up like this helps a lot! So, to summarise what this patch is doing:

* We're not adding any fields to the installer config (yay!)
* We're changing the default value of the OS URL from the qcow to the comma-separated iso,kernel,ramdisk,rootfs
* With the default values, we're using the CoreOS IPA for virtualmedia hosts. (Not yet for PXE, see comment inline.)
* If the user overrides the default with a QCOW (as they may have been doing up until now for disconnected installs) the installer would generate the right Provisioning CR manifests for CBO to handle it, but installation will fail (on virtualmedia) because we won't have the URLs for the masters.
* This will be resolved when we switch to getting images from machine-os-images - at that point everything will go through the CoreOS IPA.

I don't know that there are any test jobs (i.e. disconnected configuration) that this would break (@hardys @andfasano can you confirm?), so I tend to think this is OK and a lot simpler than supporting both the CoreOS IPA and the RHEL IPA depending on the image input, as we do in CBO.

pkg/tfvars/baremetal/baremetal.go
```diff
@@ -86,6 +86,7 @@ func TFVars(numControlPlaneReplicas int64, libvirtURI, apiVIP, imageCacheIP, boo
 		driverInfo := accessDetails.DriverInfo(credentials)
 		driverInfo["deploy_kernel"] = fmt.Sprintf("http://%s/images/ironic-python-agent.kernel", net.JoinHostPort(imageCacheIP, "80"))
 		driverInfo["deploy_ramdisk"] = fmt.Sprintf("http://%s/images/ironic-python-agent.initramfs", net.JoinHostPort(imageCacheIP, "80"))
```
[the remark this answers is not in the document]
zaneb: Actually, it looks like I was wrong about this.

The [`get_live_images.sh` script](https://github.com/openshift/ironic-rhcos-downloader/blob/master/get-live-images.sh) in ironic-rhcos-downloader puts the kernel/ramdisk (alongside the rootfs) in the same place, so these URLs will still work.

The ironic configuration script [automatically adds a kernel arg](https://github.com/metal3-io/ironic-image/blob/master/scripts/coreos-ipa-common.sh#L7-L19) (for PXE only) to tell CoreOS where to find the rootfs - we decided to do this instead of concatenating.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -152,11 +147,41 @@ $IPTABLES -t raw -A DHCP_IRONIC -j DROP
...
+    oc get secrets {{.Name}}-network-config-secret -n openshift-machine-api -o json | jq -r '.data.nmstate' | base64 -d > /tmp/nmstate/{{.Name}}.yaml
+{{end}}
+
+# When this makes the release we can use this instead.
+#CUSTOMIZATION_IMAGE=$(image_for image-customization-controller)
+# For now, this is being built locally:
+CUSTOMIZATION_IMAGE="virthost.ostest.test.metalkube.org:5000/localimages/image-customization-controller"
+
+# Embed agent ignition into the rhcos live iso
+sudo podman run -d -p 8084:8084 --privileged --name image-customization \
```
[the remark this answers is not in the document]
zaneb: It should just be `"/shared/html/images/ironic-python-agent.initramfs"` afaict.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
zaneb: That patch looks like it got messed up. https://github.com/zaneb/openshift-installer/commit/73d656816d1491f12e07f3387357fda464df5d80 is a rebased one.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -152,11 +147,41 @@ $IPTABLES -t raw -A DHCP_IRONIC -j DROP
...
-podman wait -i 1000ms ipa-downloader
 podman wait -i 1000ms coreos-downloader
-while ! curl --fail http://localhost/images/rhcos-ootpa-latest.qcow2.md5sum ; do sleep 1; done
-while ! curl --fail --head http://localhost/images/ironic-python-agent.initramfs ; do sleep 1; done
-while ! curl --fail --head http://localhost/images/ironic-python-agent.kernel ; do sleep 1; done
+
+export KUBECONFIG=/opt/openshift/auth/kubeconfig-loopback
+
+# Wait till the network secrets are populated
+until oc get secrets -n openshift-machine-api | grep network-config-secret; do
```
zaneb, line 155: I tested this and it just waited forever - I assume it doesn't create a Secret if there's no network config for a particular host.

Maybe we could wait for the BareMetalHosts to show up +10s?

zaneb: https://github.com/zaneb/openshift-installer/commit/3262834896d795a4aba45386ab7ba2c1a0bdcae8 should be safe (testing now).

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
imain: Done now.

imain:
> 
Pretty sure I got all these.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
imain: Changed to be like yours

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -152,11 +149,47 @@ $IPTABLES -t raw -A DHCP_IRONIC -j DROP
...
+        until oc get -n openshift-machine-api secret "${secret_name}"; do
+            echo Waiting for Secret "${secret_name}" to appear...
+            sleep 10
+        done
+        oc get -n openshift-machine-api secret ${secret_name} -o jsonpath="{.data.nmstate}" | base64 -d > /tmp/nmstate/{{.Name}}.yaml
+    else
+        touch /tmp/nmstate/{{.Name}}.yaml
+    fi
+{{end}}
+
```
andfasano, line 174: @imain I think that this whole block could be simplified, given that the `NetworkConfig` field is already available as a template parameter. Ie: 

```
{{range .PlatformData.BareMetal.Hosts}}
  echo "{{.NetworkConfig}}" > /tmp/nmstate/{{.Name}}.yaml
{{end}}
```

imain: That's actually what I had before.  Steve wanted to use the network secrets to get the data.  Apparently they could be modified by the user after configuration and before/during deployment.

zaneb:
Moved to #5473
/close

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
andfasano: @hardys just to clarify, do you have more details about this point? In understood that the main purpose of the secret was for day-2 operation.
During the installation the user is expected to provide the configuration in the install-config.yaml, from where the `NetworkConfig` field is extracted

PR #5426, opened 2021-11-29
Alibaba: refactor validation of machine pool

bd233:
1.Add validation for instanceType to check whether the user-specified
    or default instance type in the availability zones are `Available` 
    and `WithStock` .
    2.Add verification of the merged machine pool.
    3.Add DefaultSystemDiskSize type, this is helpful for validate the default machine pool


pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -15,168 +14,140 @@ func Validate(client *Client, ic *types.InstallConfig) error {
...
 	for _, zone := range response.AvailableZones.AvailableZone {
-		if zone.Status == "Available" {
-			zones = append(zones, zone.ZoneId)
+		if zone.Status == "Available" && zone.StatusCategory == "WithStock" {
+			zoneWithStock[zone.ZoneId] = true
 		}
 	}
+	if len(zoneWithStock) == 0 {
+		return append(allErrs, field.Invalid(fldpath, instanceType, "no availability zone with stock found"))
+	}
```
staebler, line 136: This should only be checked if `zones` is empty. Otherwise, we want to create an error for each zone explicitly specified by the user.

staebler (commented):
There are some comments in the first commit that the third commits makes obsolete. But we should not be merging in commits with issues. My preference would be to re-order the commits so that the third commit comes before the first commit. But I can also accept simply squashing everything into a single commit.

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -15,168 +14,140 @@ func Validate(client *Client, ic *types.InstallConfig) error {
...
+	allErrs = append(allErrs, validateMachinePools(client, ic)...)
 
-	allErrs = append(allErrs, validateDefaultInstanceType(client, ic)...)
+	return allErrs.ToAggregate()
+}
+
+func validateMachinePools(client *Client, ic *types.InstallConfig) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if ic.Platform.AlibabaCloud.DefaultMachinePlatform != nil {
+		allErrs = append(allErrs, validateMachinePool(client, ic, field.NewPath("platform", "alibabacloud", "defaultMachinePlatform"), ic.Platform.AlibabaCloud.DefaultMachinePlatform)...)
```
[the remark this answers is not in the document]
bd233: Yes, this is meaningless, I removed the validation on the unmerged machine pools

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -15,168 +14,140 @@ func Validate(client *Client, ic *types.InstallConfig) error {
...
-			continue
-		}
-
-		for _, zoneId := range compute.Platform.AlibabaCloud.Zones {
-			for _, id := range zones {
-				if zoneId != id {
-					zones = append(zones, id)
-				}
-			}
+	fldPath := field.NewPath("machinePool", "compute")
```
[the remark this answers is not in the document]
bd233: There is still a problem. After using `MachinePool.Set` to merge polls, the source path of the value cannot be obtained. I don't have a good idea yet. Do you have any better suggestions?

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -15,168 +14,140 @@ func Validate(client *Client, ic *types.InstallConfig) error {
...
+				allErrs = append(allErrs, field.Invalid(fldPath, zone, fmt.Sprintf("zone ID is unavailable in region %q", ic.Platform.AlibabaCloud.Region)))
 			}
 		}
-		if !isWithStock {
-			allErrs = append(allErrs, field.Invalid(path, instanceType, fmt.Sprintf("instance type is out of stock in zone %q", zoneID)))
+	}
+	// InstanceType and zones are related.
+	// If the availability zone is not available, the instanceType will not be validated.
+	if len(allErrs) == 0 {
+		if pool.InstanceType != "" {
```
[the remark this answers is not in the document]
bd233: Now that the validation on the unmerged pools has been removed, this check `pool.InstanceType != ""`  also needs to be removed.

pkg/asset/installconfig/alibabacloud/validation.go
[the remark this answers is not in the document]
staebler: Although not ideal, I suggest that we just use the path to the field in the compute or control-plane machine pool.

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -16,76 +15,108 @@ func Validate(client *Client, ic *types.InstallConfig) error {
...
+	allErrs = append(allErrs, validateComputeMachinePool(client, ic)...)
+
+	return allErrs.ToAggregate()
+}
+
+func validateControlPlaneMachinePool(client *Client, ic *types.InstallConfig) field.ErrorList {
+	allErrs := field.ErrorList{}
+	var mpool mergedMachinePool
+	defaultPool := alibabacloudtypes.DefaultMasterMachinePoolPlatform()
+	mpool.SetWithFieldPath(&defaultPool, field.NewPath("default"))
```
bd233, line 28: @staebler I added a `mergedMachinePool` type to store the field path, but there is still one unsatisfactory place, which is the use of `field.NewPath("default")`

kwoodson:
@bd233 Where are we at with this one? What's the status?

bd233:
> @bd233 Where are we at with this one? What's the status?

I have updated, please @staebler  help review whether `field.NewPath("default")` is allowed, and whether there are other problems.

pkg/asset/installconfig/alibabacloud/validation.go
staebler: No, the "default" field path is not acceptable.

Ideally, the default value should never be the cause of the validation failure. If we use the default value and it fails validation, then there should be some other field that caused that validation instead. For example, if the user did not specify the instance type, but the default instance type is not available in the region that the user specified, then the validation error should ideally be on the region rather than the instance type.

However, I understand that the gets pretty tricky, so I can accept the causes of the validation failures as we have them currently. If we use the default value and it fails validation, then let's attribute that to the specific machine pool. For example, if the default instance type fails validation for the control plane, then the field path reported in the error should be "controlPlane.platform.alibabacloud.instanceType".

bd233: > For example, if the default instance type fails validation for the control plane, then the field path reported in the error should be "controlPlane.platform.alibabacloud.instanceType".

Okay, thank you, I used the machine pool field path for the default value.

bd233: @staebler Can you review again？

go.mod
```diff
@@ -83,7 +83,6 @@ require (
 	github.com/ulikunitz/xz v0.5.8
 	github.com/vincent-petithory/dataurl v0.0.0-20191104211930-d1553a71de50
 	github.com/vmware/govmomi v0.24.0
-	github.com/wxnacy/wgo v1.0.4
```
staebler, line 86: Please separate vendor changes into their own commit.

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -16,76 +15,109 @@ func Validate(client *Client, ic *types.InstallConfig) error {
...
+	allErrs = append(allErrs, validateMachinePool(client, ic, &mpool)...)
+	return allErrs
+}
+
+func validateComputeMachinePool(client *Client, ic *types.InstallConfig) field.ErrorList {
+	allErrs := field.ErrorList{}
+	var mpool mergedMachinePool
 
 	for idx, compute := range ic.Compute {
 		fldPath := field.NewPath("compute").Index(idx)
```
staebler, line 43: This field path is only used to later get the platform.alibabacloud child. The variable name "fldPath" is not descriptive enough.
```suggestion
		computePoolFieldPath := field.NewPath("compute").Index(idx).Child("platform", "alibabacloud")
```

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -16,76 +15,109 @@ func Validate(client *Client, ic *types.InstallConfig) error {
...
+
+	response, err := client.DescribeAvailableZoneByInstanceType(instanceType)
 	if err != nil {
 		return append(allErrs, field.InternalError(fldPath, err))
 	}
-	for _, zone := range response.AvailableZones.AvailableZone {
-		if zone.Status == "Available" {
-			zones = append(zones, zone.ZoneId)
-		}
+	if len(response.AvailableZones.AvailableZone) == 0 {
```
staebler, line 97: This check should only be done if the user did not specify specific zones to use.
```suggestion
	if len(zones) == 0 && len(response.AvailableZones.AvailableZone) == 0 {
```

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -121,3 +153,24 @@ func validateClusterName(client *Client, ic *types.InstallConfig) field.ErrorLis
...
 	return allErrs
 }
+
+type mergedMachinePool struct {
+	alibabacloudtypes.MachinePool
+	zonesFieldPath        *field.Path
+	instanceTypeFieldPath *field.Path
+}
+
+func (a *mergedMachinePool) SetWithFieldPath(required *alibabacloudtypes.MachinePool, fldPath *field.Path) {
```
staebler, line 163: This function can be private.
```suggestion
func (a *mergedMachinePool) setWithFieldPath(required *alibabacloudtypes.MachinePool, fldPath *field.Path) {
```

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -16,76 +15,109 @@ func Validate(client *Client, ic *types.InstallConfig) error {
...
 	}
 
-	if len(pool.Zones) > 0 {
-		for _, zone := range pool.Zones {
-			if index := arrays.ContainsString(zones, zone); index == -1 {
-				allErrs = append(allErrs, field.Invalid(fldPath, zone, fmt.Sprintf("zone ID is unavailable in region %q", ic.Platform.AlibabaCloud.Region)))
-			}
-
+	for _, zone := range response.AvailableZones.AvailableZone {
+		if zone.Status == "Available" && zone.StatusCategory == "WithStock" {
```
staebler, line 102: Can we put `GetAvailabilityZonesByInstanceType` somewhere so that it can be called here rather than duplicating the logic of searching for zones with stock?

bd233: @staebler  Have updated

pkg/asset/installconfig/alibabacloud/client.go
```diff
@@ -262,6 +275,23 @@ func (client *Client) GetOSSObjectSignURL(bucketName string, objectName string)
 	return
 }
 
+// GetAvailabilityZonesByInstanceType returns a list of supported zones for the specified instance type.
```
staebler, line 278: Please improve on this function description to make it clear how this function differs from `DescribeAvailableZoneByInstanceType`.

pkg/asset/installconfig/alibabacloud/client.go
```diff
@@ -262,6 +275,23 @@ func (client *Client) GetOSSObjectSignURL(bucketName string, objectName string)
...
+func (client *Client) GetAvailabilityZonesByInstanceType(instanceType string) ([]string, error) {
```
staebler, line 279: Should the name of this function use "AvailableZone" rather than "AvailabilityZone"? The former seems to be the term use by the sdk.
```suggestion
func (client *Client) GetAvailableZonesByInstanceType(instanceType string) ([]string, error) {
```

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -16,76 +15,107 @@ func Validate(client *Client, ic *types.InstallConfig) error {
...
-		allErrs = append(allErrs, validateMachinePool(client, ic, field.NewPath("controlPlane", "platform", "alibabacloud"), ic.ControlPlane.Platform.AlibabaCloud, ic.ControlPlane.Replicas)...)
+	allErrs = append(allErrs, validateControlPlaneMachinePool(client, ic)...)
+	allErrs = append(allErrs, validateComputeMachinePool(client, ic)...)
+
+	return allErrs.ToAggregate()
+}
+
+func validateControlPlaneMachinePool(client *Client, ic *types.InstallConfig) field.ErrorList {
+	allErrs := field.ErrorList{}
+	var mpool mergedMachinePool
```
staebler, line 26: Although the behavior is the same, the typical way to get an zero-value struct is the following.
```suggestion
	mpool := mergedMachinePool{}
```

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -16,76 +15,107 @@ func Validate(client *Client, ic *types.InstallConfig) error {
...
 	}
+	mpool.setWithFieldPath(ic.ControlPlane.Platform.AlibabaCloud, field.NewPath("controlPlane", "platform", "alibabacloud"))
+
+	allErrs = append(allErrs, validateMachinePool(client, ic, &mpool)...)
+	return allErrs
+}
+
+func validateComputeMachinePool(client *Client, ic *types.InstallConfig) field.ErrorList {
+	allErrs := field.ErrorList{}
+	var mpool mergedMachinePool
```
staebler, line 40: Same comment here about typically code style.
```suggestion
	mpool := mergedMachinePool{}
```

staebler, line 40: This variable needs to be declared inside the context of the for loop so that it is zero-valued for each compute pool.

pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -16,76 +15,107 @@ func Validate(client *Client, ic *types.InstallConfig) error {
...
 
 	return allErrs
 }
 
-func validateMachinePool(client *Client, ic *types.InstallConfig, fldPath *field.Path, pool *alibabacloudtypes.MachinePool, replicas *int64) field.ErrorList {
+func validateInstanceType(client *Client, zones []string, instanceType string, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
-	var zones []string
-	response, err := client.DescribeAvailableResource("Zone")
+	zoneWithStock := map[string]bool{}
```
staebler, line 91: I think I steered you to `map[string]bool` here, but it would be a bit cleaner if you used `sets.NewString` instead.
https://github.com/kubernetes/apimachinery/blob/5f072755815add51938b52f806c555ed8c86a1ce/pkg/util/sets/string.go#L30
```
zonesWithStock := sets.NewString(availableZones)
```
```
for_, zoneID := range zones {
  if zonesWithStock.Has(zoneID) {
    ...
  }
}
```

pkg/types/alibabacloud/machinepool.go
```diff
@@ -76,3 +76,30 @@ func (a *MachinePool) Set(required *MachinePool) {
 
 // DefaultDiskCategory holds the default Alibaba Cloud disk type used by the ECS.
 const DefaultDiskCategory DiskCategory = CloudESSDDiskCategory
+
+// DefaultDiskSize holds the default Alibaba Cloud system disk size used by the ECS.
```
staebler, line 80:
```suggestion
/ DefaultSystemDiskSize holds the default Alibaba Cloud system disk size used by the ECS.
```

staebler (commented):
Functionally, looks good. Just a handful of nits and lint problems.

pkg/types/alibabacloud/machinepool.go
bd233: Thank you for your suggestion, I have updated

staebler (commented):
Looks good to me. I will give it a lgtm once I do some testing (unless someone beats me to it).
/approve

mtulio:
> ```
> 1.Add validation for instanceType to check whether the user-specified
> or default instance type in the availability zones are `Available` 
> and `WithStock` .
> 2.Add verification of the merged machine pool.
> 3.Add DefaultSystemDiskSize type, this is helpful for validate the default machine pool
> ```

I ran some tests changing the instanceType, checking unavailable. Unfortunately, the test "WithNoInStock" is hard due we need to test-and-tried with some luck to get one instance that is missing in one zone. So I skip this one.

I can't see restrictions to move it forward.

/lgtm

cc @staebler 
__
## Tests

Test => When setting the Instance that did not exists on the region
Result => PASS
Log:
```
$ yq .controlPlane.platform.alibabacloud mrb5426/install-config.yaml
{
  "instanceType": "ecs.g7.large"
}


$ aliyun ecs DescribeAvailableResource --RegionId us-east-1 --InstanceType ecs.g7.large --DestinationResource InstanceType 
{
	"RequestId": "2D2B115A-C3C3-3860-9136-299E0C43315E"
}

$ aliyun ecs DescribeAvailableResource --RegionId us-east-1 --InstanceType ecs.g6.large --DestinationResource InstanceType |jq -r '.AvailableZones.AvailableZone[] | ("=>", .ZoneId, .StatusCategory, .Status, .StatusCategory)'
=>
us-east-1b
WithStock
Available
WithStock
=>
us-east-1a
WithStock
Available
WithStock

# trying to use g7
$ ${BIN_INSTALLER_CREATE} create manifests --dir "${NEW_INSTALL_DIR}"
FATAL failed to fetch Master Machines: failed to load asset "Install Config": controlPlane.platform.alibabacloud.instanceType: Invalid value: "ecs.g7.large": no available availability zones found
```

Test => When setting the InstanceType that is available in only one AZ
Result => PASS

```
$ yq .controlPlane.platform.alibabacloud mrb5426_install-config.yaml-instance-g5.16x 
{
  "instanceType": "ecs.g5.16xlarge"
}

$ aliyun ecs DescribeAvailableResource --RegionId us-east-1 --InstanceType ecs.g5.16xlarge --DestinationResource InstanceType |jq -r '.AvailableZones.AvailableZone[] | ("=>", .ZoneId, .StatusCategory, .Status, .StatusCategory)'
=>
us-east-1a
WithStock
Available
WithStock

$ ${BIN_INSTALLER_CREATE} create manifests --dir "${NEW_INSTALL_DIR}"
WARNING Found override for OS Image. Please be warned, this is not advised 
INFO Manifests created in: mrb5426/manifests and mrb5426/openshift 

$ yq -r '.spec.providerSpec.value|(.instanceType, .zoneId, .systemDiskSize)' mrb5426/openshift/99_openshift-cluster-api_master-machines-*.yaml
ecs.g5.16xlarge
us-east-1a
120
ecs.g5.16xlarge
us-east-1aTes
120
ecs.g5.16xlarge
us-east-1a
120

$ yq -r '.spec.template |(.metadata.labels["machine.openshift.io/cluster-api-machineset"], .spec.providerSpec.value.zoneId, .spec.providerSpec.value.instanceType, .spec.providerSpec.value.systemDiskSize)' mrb5426/openshift/99_openshift-cluster-api_worker-machineset-*.yaml
mrb5426-4j47v-worker-us-east-1b
us-east-1b
ecs.g6.large
120
mrb5426-4j47v-worker-us-east-1a
us-east-1a
ecs.g6.large
120

```

Test =>  Creating a cluster with different type of default
Result => Pass

```
]$ yq .controlPlane.platform.alibabacloud mrb5426_install-config.yaml-instance-g5.2x 
{
  "instanceType": "ecs.g5.2xlarge"
}

```
Screenshot from 2021-12-10 18-59-19

pkg/asset/installconfig/alibabacloud/validation.go
bd233: I am sorry to realize this problem after the merge, I think it has nothing to do with whether the user specifies zones, this is just to judge whether the available zone of `instanceType` (default or user specified) exists
I have create a PR #5479 to fix it.

PR #5429, opened 2021-11-29
OpenStack: Parallelize port deletion

dulek:
OpenStack Neutron is slow but it's sometimes able to handle quite a few
requests in parallel. This commit splits the operation of Neutron port
deletion into 10 concurrent goroutines to speed it up. This mostly aids
Kuryr case when we can have hundreds of ports created that block
deletion of other resources.

### DIFF
--- a/pkg/destroy/openstack/openstack.go
+++ b/pkg/destroy/openstack/openstack.go
@@ -347,49 +347,80 @@ func deletePorts(opts *clientconfig.ClientOpts, filter Filter, logger logrus.Fie
 	}
 	numberToDelete := len(allPorts)
 	numberDeleted := 0
-	for _, port := range allPorts {
-		listOpts := floatingips.ListOpts{
-			PortID: port.ID,
-		}
-		allPages, err := floatingips.List(conn, listOpts).AllPages()
-		if err != nil {
-			logger.Error(err)
-			return false, nil
-		}
-		allFIPs, err := floatingips.ExtractFloatingIPs(allPages)
-		if err != nil {
-			logger.Error(err)
-			return false, nil
-		}
-		// If a user provisioned floating ip was used, it needs to be dissociated.
-		// Any floating Ip's associated with ports that are going to be deleted will be dissociated.
-		for _, fip := range allFIPs {
-			logger.Debugf("Dissociating Floating IP %q", fip.ID)
-			_, err := floatingips.Update(conn, fip.ID, floatingips.UpdateOpts{}).Extract()
-			if err != nil {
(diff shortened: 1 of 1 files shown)


dulek:
@MaysaMacedo, @mandre, @pierreprinetti:  Can you please take a look? This seems to speed up `destroy cluster` with Kuryr quite a few times.

dulek:
Without the patch:
> Step e2e-openstack-kuryr-ipi-deprovision-deprovision succeeded after 6m30s. 
With the patch:
> Step e2e-openstack-kuryr-ipi-deprovision-deprovision succeeded after 2m40s.

pkg/destroy/openstack/openstack.go
```diff
@@ -347,49 +347,78 @@ func deletePorts(opts *clientconfig.ClientOpts, filter Filter, logger logrus.Fie
...
+		deletedChannel <- localDeleted
+	}
+
+	deletedChannel := make(chan int)
+	const chunksNumber = 10
+
+	// run processing in chunks
+	for i := 0; i < chunksNumber; i++ {
+		// This is at trick, but it'll produce chunksNumber chunks of roughly the same size.
+		// It may produce empty chunks, but why would we care?
```
mandre, line 410: This doesn't seem right. Take for example `len(allPorts) = 1`, you'd have 10 workers trying to delete the same port, and in the end `numberDeleted` is equal to 10.

You should instead use a buffered channel and run each port delete in its own goroutine.
See https://gobyexample.com/worker-pools for example.

dulek: This is actually fun, in such case the trick in 411-412 would produce 10 chunks, 9 of which would be empty. Check out and play with [1]. It's suboptimal but I opted to do it that way to make code cleaner.

Now the way you're proposing seems to be even simpler, let me try it out.

[1] https://go.dev/play/p/ogvz3ARDGc9

dulek:
/hold

@mandre proposed alternative approach that seems to be more natural.

pkg/destroy/openstack/openstack.go
pierreprinetti: Beautiful!
I used this model to test your code: https://go.dev/play/p/6h2w0YgXts3

dulek: Right, you've tested the way @mandre proposed, which can also be called "the correct way of doing this". ;)

mandre (commented):
Nice, thanks @dulek 
/lgtm

dulek:
> Step e2e-openstack-kuryr-ipi-deprovision-deprovision succeeded after 2m50s. 

Still looks good after latest changes.

PR #5430, opened 2021-11-30
Azure Stack: add CA support in bootstrap ign stub

patrickdillon:
On Azure Stack, this adds CAs from the additionalTrustBundle to the ignition stub. Azure Stack uses shared-access signatures (SAS) to manage access to the ignition storage account. [Creating a SAS requires an existing storage account](https://pkg.go.dev/github.com/Azure/azure-storage-blob-go/azblob?utm_source=godoc#BlobSASSignatureValues), so we cannot use a similar approach to AWS, which creates a pre-signed URL and passes it through the tfvars.

Instead, this approach manipulates the ignition stub in the Terraform config to append the SAS to the storage account URL.

### DIFF
--- a/data/data/azurestack/bootstrap/main.tf
+++ b/data/data/azurestack/bootstrap/main.tf
@@ -70,12 +70,6 @@ resource "azurestack_storage_blob" "ignition" {
   type                   = "block"
 }
 
-data "ignition_config" "redirect" {
-  replace {
-    source = "${azurestack_storage_blob.ignition.url}${data.azurestack_storage_account_sas.ignition.sas}"
-  }
-}
-
 resource "azurestack_public_ip" "bootstrap_public_ip_v4" {
   count = var.azure_private ? 0 : 1
 
@@ -125,7 +119,10 @@ resource "azurestack_virtual_machine" "bootstrap" {
     # isn't installed in RHCOS. As a result, this password is never set. It is
     # included here because it is required by the Azure ARM API.
     admin_password = "NotActuallyApplied!"
-    custom_data    = base64encode(data.ignition_config.redirect.rendered)
+
+    custom_data = base64encode(replace(var.azure_bootstrap_ignition_stub,
+      var.azure_bootstrap_ignition_url_placeholder,
+    "${azurestack_storage_blob.ignition.url}${data.azurestack_storage_account_sas.ignition.sas}"))
   }
 
   os_profile_linux_config {
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -324,19 +324,34 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		}
 
 		preexistingnetwork := installConfig.Config.Azure.VirtualNetwork != ""
+
+		var bootstrapIgnStub, bootstrapIgnURLPlaceholder string
+		if installConfig.Azure.CloudName == azure.StackCloud {
+			// Due to the SAS created in Terraform to limit access to bootstrap ignition, we cannot know the URL in advance.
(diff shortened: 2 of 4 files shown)


data/data/azurestack/bootstrap/main.tf
```diff
@@ -125,7 +119,10 @@ resource "azurestack_virtual_machine" "bootstrap" {
     # isn't installed in RHCOS. As a result, this password is never set. It is
     # included here because it is required by the Azure ARM API.
     admin_password = "NotActuallyApplied!"
-    custom_data    = base64encode(data.ignition_config.redirect.rendered)
+
+    custom_data = base64encode(replace(var.azure_bootstrap_ignition_stub,
+      azurestack_storage_blob.ignition.url,
+    "${azurestack_storage_blob.ignition.url}${data.azurestack_storage_account_sas.ignition.sas}"))
```
patrickdillon, line 125: This is a complicated line so I tried to make it more readable with formatting. I also considered handling this in a local. Not sure if there is any way to make this easier to read.

pkg/asset/cluster/tfvars.go
```diff
@@ -325,6 +325,24 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
+
+		var bootstrapIgnStub, storageAccountName string
+		if installConfig.Azure.CloudName == azure.StackCloud {
+			// Our Azure Terraform config uses a random string to name storage accounts as the names must be globally unique & only alphanumeric.
+			// In order to create the bootstrap stub in advance, that name needs to be deterministic, so we can use the
+			// infrastructure ID suffix, which is also a random string. Keep the "cluster" prefix to be consistent with public Azure.
+			infraIDSuffix := strings.Split(clusterID.InfraID, "-")[1]
+			storageAccountName = fmt.Sprintf("cluster%s", infraIDSuffix)
+
+			storageSuffix := strings.TrimPrefix(installConfig.Azure.ARMEndpoint, "https://management.")
```
patrickdillon, line 337: Should probably change this based on this comment: https://github.com/openshift/installer/pull/5103/files#r725223700

I might need to update the metadata.

pkg/asset/cluster/tfvars.go
```diff
@@ -325,6 +325,24 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
 
 		preexistingnetwork := installConfig.Config.Azure.VirtualNetwork != ""
+
+		var bootstrapIgnStub, storageAccountName string
+		if installConfig.Azure.CloudName == azure.StackCloud {
+			// Our Azure Terraform config uses a random string to name storage accounts as the names must be globally unique & only alphanumeric.
+			// In order to create the bootstrap stub in advance, that name needs to be deterministic, so we can use the
+			// infrastructure ID suffix, which is also a random string. Keep the "cluster" prefix to be consistent with public Azure.
+			infraIDSuffix := strings.Split(clusterID.InfraID, "-")[1]
+			storageAccountName = fmt.Sprintf("cluster%s", infraIDSuffix)
```
staebler, line 335: I am not sure that I am understanding the need for this. As far as I can tell, we don't need the storage account name to be deterministic. The only thing that we need to be deterministic is the string that we are searching for in the ignition stub to replace with the actual URL that we want to use for the bootstrap ignition config. The string that we search for does not have to be a URL: It can be a constant value unrelated to the specific cluster being installed.

patrickdillon: I think you are correct re constant string and that will be a great improvement. I will put it in.

staebler (commented):
Just a nit. Otherwise, it looks good.
/approve

patrickdillon:
Ok tf-fmt is now passing and I have rebased and removed optional tf vars. I think this is ready for review

/test e2e-azure

staebler (approved):
/lgtm

PR #5432, opened 2021-11-30
Bug 2019977: Validate region provided in install config

rna-afk:
When an invalid region is passed to the installer, the binary
spends 60 minutes trying to find the region and fails. Adding
a check to see if the region provided is present in the list of
supported regions in the AWS SDK.

### DIFF
--- a/pkg/asset/installconfig/aws/validation.go
+++ b/pkg/asset/installconfig/aws/validation.go
@@ -11,6 +11,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/route53"
 	"github.com/pkg/errors"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -276,29 +277,25 @@ func validateDuplicateSubnetZones(fldPath *field.Path, subnets map[string]Subnet
 
 func validateServiceEndpoints(fldPath *field.Path, region string, services []awstypes.ServiceEndpoint) field.ErrorList {
 	allErrs := field.ErrorList{}
-	// For each provided service endpoint, verify we can resolve and connect with net.Dial.
+	ec2Endpoint := ""
 	for id, service := range services {
-		// Ignore e2e.local from unit tests.
-		if service.URL == "e2e.local" {
-			continue
-		}
-		URL, err := url.Parse(service.URL)
+		err := validateEndpointAccessibility(service.URL)
 		if err != nil {
 			allErrs = append(allErrs, field.Invalid(fldPath.Index(id).Child("url"), service.URL, err.Error()))
 			continue
 		}
-		port := URL.Port()
-		if port == "" {
-			port = "https"
+		if service.Name == ec2.ServiceName {
+			ec2Endpoint = service.URL
 		}
-		conn, err := net.Dial("tcp", net.JoinHostPort(URL.Hostname(), port))
-		if err != nil {
-			allErrs = append(allErrs, field.Invalid(fldPath.Index(id).Child("url"), service.URL, err.Error()))
-			continue
-		}
-		conn.Close()
 	}
-	if _, partitionFound := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), region); partitionFound {
+
+	if partition, partitionFound := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), region); partitionFound {
+		if _, ok := partition.Regions()[region]; !ok && ec2Endpoint == "" {
+			err := validateRegion(region)
+			if err != nil {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("region"), region, err.Error()))
+			}
+		}
 		return allErrs
 	}
 
@@ -313,8 +310,40 @@ func validateServiceEndpoints(fldPath *field.Path, region string, services []aws
 	if err := utilerrors.NewAggregate(errs); err != nil {
 		allErrs = append(allErrs, field.Invalid(fldPath, services, err.Error()))
 	}
-
 	return allErrs
+}
+
+func validateRegion(region string) error {
+	ses, err := GetSessionWithOptions(func(sess *session.Options) {
+		sess.Config.Region = aws.String(region)
+	})
+	if err != nil {
+		return err
+	}
+	ec2Session := ec2.New(ses)
+	return validateEndpointAccessibility(ec2Session.Endpoint)
+}
+
+func validateEndpointAccessibility(endpointURL string) error {
+	// For each provided service endpoint, verify we can resolve and connect with net.Dial.
+	// Ignore e2e.local from unit tests.
+	if endpointURL == "e2e.local" {
+		return nil
+	}
+	URL, err := url.Parse(endpointURL)
+	if err != nil {
+		return err
+	}
+	port := URL.Port()
+	if port == "" {
+		port = "https"
+	}
+	conn, err := net.Dial("tcp", net.JoinHostPort(URL.Hostname(), port))
+	if err != nil {
+		return err
+	}
+	conn.Close()
+	return nil
 }
 
 var requiredServices = []string{
--- a/pkg/asset/installconfig/aws/validation_test.go
+++ b/pkg/asset/installconfig/aws/validation_test.go
@@ -458,6 +458,7 @@ func TestValidate(t *testing.T) {
 		installConfig: func() *types.InstallConfig {
 			c := validInstallConfig()
 			c.Platform.AWS.Region = "us-newregion-1"
+			c.Platform.AWS.ServiceEndpoints = validServiceEndpoints()
 			return c
 		}(),
 		availZones:     validAvailZones(),


pkg/asset/installconfig/aws/validation.go
```diff
@@ -59,7 +60,13 @@ func Validate(ctx context.Context, meta *Metadata, config *types.InstallConfig)
 	}
 	return allErrs.ToAggregate()
 }
-
+func validateRegion(region string) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if _, ok := endpoints.AwsPartition().Regions()[region]; !ok {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("platform", "aws", "region"), region, "invalid region"))
```
jstuever, line 66: It is unnecessary (and likely less optimal execution) to call an append when there are not multiple errors to be combined and returned. Instead, return `field.ErrorList{field.Invalid(field.NewPath("platform", "aws", "region"), region, "invalid region")}` here and return `nil` below. You can refer to `func validateAMI()` below for similar usage.

jstuever (changes requested):
This looks generally good. I've suggested an minor optimization.

pkg/asset/installconfig/aws/validation.go
rna-afk: Is this approach correct? I was under the (vague) impression that we allowed custom regions along with custom service endpoints? There was a validation test like that here.
https://github.com/openshift/installer/blob/46acb726bbebb7d2511dc7f83ed92b6385eb286b/pkg/asset/installconfig/aws/validation_test.go#L445

staebler: No it is not correct. I just realized that I never submitted my comment from yesterday. See below.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -59,7 +60,13 @@ func Validate(ctx context.Context, meta *Metadata, config *types.InstallConfig)
 	}
 	return allErrs.ToAggregate()
 }
-
+func validateRegion(region string) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if _, ok := endpoints.AwsPartition().Regions()[region]; !ok {
```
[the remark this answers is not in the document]
staebler: I suggest that we query AWS for the valid regions and check that the selected region is in that list.
The sdk equivalent of the following.
```
$ aws ec2 describe-regions
REGIONS ec2.af-south-1.amazonaws.com    opted-in        af-south-1
REGIONS ec2.eu-north-1.amazonaws.com    opt-in-not-required     eu-north-1
REGIONS ec2.ap-south-1.amazonaws.com    opt-in-not-required     ap-south-1
REGIONS ec2.eu-west-3.amazonaws.com     opt-in-not-required     eu-west-3
REGIONS ec2.eu-west-2.amazonaws.com     opt-in-not-required     eu-west-2
REGIONS ec2.eu-south-1.amazonaws.com    opted-in        eu-south-1
REGIONS ec2.eu-west-1.amazonaws.com     opt-in-not-required     eu-west-1
REGIONS ec2.ap-northeast-3.amazonaws.com        opt-in-not-required     ap-northeast-3
REGIONS ec2.ap-northeast-2.amazonaws.com        opt-in-not-required     ap-northeast-2
REGIONS ec2.me-south-1.amazonaws.com    opted-in        me-south-1
REGIONS ec2.ap-northeast-1.amazonaws.com        opt-in-not-required     ap-northeast-1
REGIONS ec2.sa-east-1.amazonaws.com     opt-in-not-required     sa-east-1
REGIONS ec2.ca-central-1.amazonaws.com  opt-in-not-required     ca-central-1
REGIONS ec2.ap-east-1.amazonaws.com     opted-in        ap-east-1
REGIONS ec2.ap-southeast-1.amazonaws.com        opt-in-not-required     ap-southeast-1
REGIONS ec2.ap-southeast-2.amazonaws.com        opt-in-not-required     ap-southeast-2
REGIONS ec2.eu-central-1.amazonaws.com  opt-in-not-required     eu-central-1
REGIONS ec2.us-east-1.amazonaws.com     opt-in-not-required     us-east-1
REGIONS ec2.us-east-2.amazonaws.com     opt-in-not-required     us-east-2
REGIONS ec2.us-west-1.amazonaws.com     opt-in-not-required     us-west-1
REGIONS ec2.us-west-2.amazonaws.com     opt-in-not-required     us-west-2
```

pkg/asset/installconfig/aws/regions.go
```diff
@@ -37,3 +41,20 @@ func IsKnownRegion(region string, architecture types.Architecture) bool {
 	}
 	return false
 }
+
+// IsKnownAWSRegion gets all the supported AWS regions currently using the SDK session call and
+// checks if the given region is present in the list of regions returned.
+func IsKnownAWSRegion(region string) (bool, error) {
+	session := ec2.New(session.New(), &aws.Config{Region: aws.String("us-east-1")})
```
staebler, line 48: Hmm. I didn't consider that we need to hit a known region in order to list the available regions. We cannot use "us-east-1" when the partition requested is not the standard partition. We would need to find the partition for the selected region first, and then use a region in that partition to validate that the region exists. There would also need to be consideration for whether the user provided an override for the ec2 service endpoint.

I am coming back around to the idea of running a test query against the selected region with a short timeout rather than querying for whether the region exists directly.

rna-afk: Yeah that works locally but the tests are failing with almost all regions. Not really sure but is it an awscred file error?

rna-afk:
This is a better solution than the one I am doing right now. I took this approach hoping that I could try and tackle another bug where ap-northeast-3 was not showing up in the installer. Fetching all the regions will work for both but I feel I have to do way more work for just two small requirements. I will try this.

pkg/asset/machines/master.go
```diff
@@ -220,7 +221,9 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 					mpool.Zones = append(mpool.Zones, zone)
 				}
 			} else {
-				mpool.Zones, err = installConfig.AWS.AvailabilityZones(ctx)
+				ctx2, cancel := context.WithTimeout(ctx, 1*time.Minute)
```
staebler, line 224: I would rather not put the timeout in this call. Putting it here is somewhat fragile because it makes an assumption that this is going to be the very first call to the service endpoint. I would rather make a separate call where we are explicitly testing the accessibility of the service endpoints. Conveniently, we already have a function where we validate the service endpoints, namely `validateServiceEndpoints`.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -298,8 +299,17 @@ func validateServiceEndpoints(fldPath *field.Path, region string, services []aws
...
 	}
-	if _, partitionFound := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), region); partitionFound {
-		return allErrs
+
+	if partition, partitionFound := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), region); partitionFound {
+		if _, ok := partition.Regions()[region]; ok {
+			return allErrs
+		} else if len(services) == 0 {
+			err2 := validateRegion(region)
+			if err2 != nil {
```
staebler, line 308: No need to use `err2` here instead of `err`.
```suggestion
			if err := validateRegion(region); err != nil {
```

pkg/asset/installconfig/aws/validation.go
```diff
@@ -313,10 +323,24 @@ func validateServiceEndpoints(fldPath *field.Path, region string, services []aws
...
 
+// validateRegion checks if the given region is supported by AWS.
+func validateRegion(region string) error {
+	ses, err := GetSession()
+	if err != nil {
+		return err
+	}
+	ec2Session := ec2.New(ses)
+	input := &ec2.DescribeRegionsInput{RegionNames: aws.StringSlice([]string{region})}
+	_, err = ec2Session.DescribeRegions(input)
```
staebler, line 337: We still need something here to limit the amount of time that it takes for the request to time out, right? Otherwise we will just wait here for 60 minutes.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -313,10 +323,24 @@ func validateServiceEndpoints(fldPath *field.Path, region string, services []aws
...
+	if err != nil {
+		return err
+	}
+	return nil
```
staebler, line 341:
```suggestion
	return err
```

pkg/asset/installconfig/aws/validation.go
```diff
@@ -313,10 +323,24 @@ func validateServiceEndpoints(fldPath *field.Path, region string, services []aws
 	if err := utilerrors.NewAggregate(errs); err != nil {
 		allErrs = append(allErrs, field.Invalid(fldPath, services, err.Error()))
 	}
-
 	return allErrs
 }
 
+// validateRegion checks if the given region is supported by AWS.
+func validateRegion(region string) error {
+	ses, err := GetSession()
```
staebler, line 331: We need a way to sub out the as calls for the unit tests, right?

rna-afk: Not sure. The aws calls are working in CI.

pkg/asset/installconfig/aws/validation.go
rna-afk: Not sure. If I pass an invalid region I get error message real quick but there might be a possibility that the API is unreachable.

pkg/asset/installconfig/aws/validation.go
staebler: I think that what you are seeing is the call to `GetSession` throwing an error that the calling code is interpreting as an unknown region.

pkg/asset/installconfig/aws/validation.go
staebler: Ah, I see. The `GetSession` function is using the default region of us-east-1. So this won't work for other partitions.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -313,10 +324,24 @@ func validateServiceEndpoints(fldPath *field.Path, region string, services []aws
...
+	if err != nil {
+		return err
+	}
+	ec2Session := ec2.New(ses)
+	input := &ec2.DescribeRegionsInput{RegionNames: aws.StringSlice([]string{region})}
+	ctx, cancel := context.WithTimeout(context.TODO(), 25*time.Second)
+	defer cancel()
+	_, err = ec2Session.DescribeRegionsWithContext(ctx, input)
+	return err
+}
```
staebler, line 343: I know that using the context has been discussed previously in this PR. But I cannot get over the seemingly arbitrary 25-second timeout. I still think it is better to pull-up the URL validation code from `validateServiceEndpoints` and re-use it here to validate accessibility of the endpoint URL.
```suggestion
	return validateEndpointAccessibility(ec2Session.Endpoint)
}

func validateEndpointAccessibility(endpointURL string) error {
	// Ignore e2e.local from unit tests.
	if endpointURL == "e2e.local" {
		return nil
	}
	URL, err := url.Parse(endpointURL)
	if err != nil {
		return err
	}
	port := URL.Port()
	if port == "" {
		port = "https"
	}
	conn, err := net.Dial("tcp", net.JoinHostPort(URL.Hostname(), port))
	if err != nil {
		return err
	}
	conn.Close()
	return nil
}
```

pkg/asset/installconfig/aws/validation.go
```diff
@@ -298,8 +300,17 @@ func validateServiceEndpoints(fldPath *field.Path, region string, services []aws
...
+	if partition, partitionFound := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), region); partitionFound {
+		if _, ok := partition.Regions()[region]; ok {
+			return allErrs
+		} else if len(services) == 0 {
+			err := validateRegion(region)
+			if err != nil {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("region"), region, "region not found"))
+				return allErrs
+			}
+		}
```
staebler, line 313: When the partition is found, this needs to *always* return so that the resolver code is not called. Otherwise, valid regions that are not known by the AWS SDK will fail.

This also means that we should not be checking whether the user provided services. Instead, the `validateRegion` function needs to take as a parameter the services so that the user-provided endpoint for ec2 can be used if provided. This leaves a gap where the user could provide an invalid region and a valid ec2 endpoint, but I think I can live with that.
```suggestion
		if _, ok := partition.Regions()[region]; !ok {
			err := validateRegion(region, serviceEndpoints)
			if err != nil {
				allErrs = append(allErrs, field.Invalid(fldPath.Child("region"), region, "region not found"))
			}
		}
		return allErrs
```

pkg/asset/installconfig/aws/validation.go
```diff
@@ -298,8 +300,17 @@ func validateServiceEndpoints(fldPath *field.Path, region string, services []aws
...
-	if _, partitionFound := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), region); partitionFound {
-		return allErrs
+
+	if partition, partitionFound := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), region); partitionFound {
+		if _, ok := partition.Regions()[region]; ok {
+			return allErrs
+		} else if len(services) == 0 {
+			err := validateRegion(region)
+			if err != nil {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("region"), region, "region not found"))
```
staebler, line 310: Let's include the underlying error in the field error.
```suggestion
				allErrs = append(allErrs, field.Invalid(fldPath.Child("region"), region, err.Error()))
```

staebler (commented):
Looks good. I just have 2 small nits that shouldn't have any effect on any behavior.

staebler (approved):
/lgtm

PR #5436, opened 2021-12-01
Bug 2027387: fix(ibmcloud): Do not compute image file checksum

hasueki:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2027387

Remove the usage of the `etag` TF variable with `filemd()`. This variable is intended for TF to compute the diff of a file to perform an update of an existing object. The installer is not designed to run update operations, so we can just remove.

The `filemd()` TF function is causing high memory consumption according to https://github.com/hashicorp/terraform/issues/23890.


staebler (approved):
/lgtm
/approve

pamoedom (commented):
/lgtm

With this change, the installer (along with tf-ibm-provider) is not exceeding 300MB during all installations phases, initiation, COS bucket image creation, bootstrap, etc.

PR #5437, opened 2021-12-01
Remove GCP cluster api provider

patrickdillon:
I ended up doing this work while trying to eliminate our need to `replace` the cluster-api-provider-azure. More work is needed beyond this to achieve that, but we might as well commit it. I am doing this work for Azure, but we should be able to do something similar for all other providers that have migrated to openshift/api.


patrickdillon:
/test e2e-gcp

I thought these tests should be run automatically with recent changes, but it looks like they are not getting queued.

pkg/asset/machines/master.go
```diff
@@ -537,7 +535,6 @@ func (m *Master) Machines() ([]machineapi.Machine, error) {
 	awsapi.AddToScheme(scheme)
 	azureapi.AddToScheme(scheme)
 	baremetalapi.AddToScheme(scheme)
-	gcpapi.AddToScheme(scheme)
```
kirankt, line 540: I think we will need to add the GCP types into the schema here.
e.g. https://github.com/openshift/installer/blob/c3a4dbc69860b9b66b019461ce3fc3c546e1914f/pkg/asset/machines/master.go#L545-L548

pkg/asset/machines/worker.go
```diff
@@ -555,7 +552,6 @@ func (w *Worker) MachineSets() ([]machineapi.MachineSet, error) {
 		awsprovider.SchemeGroupVersion,
 		azureprovider.SchemeGroupVersion,
 		baremetalprovider.SchemeGroupVersion,
-		gcpprovider.SchemeGroupVersion,
```
kirankt, line 558: Ditto.

kirankt (commented):
Thanks @patrickdillon I actaully meant to work on this today as well. Left a few comments. PTAL.

pkg/tfvars/gcp/gcp.go
```diff
@@ -4,8 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 
-	gcpprovider "github.com/openshift/cluster-api-provider-gcp/pkg/apis/gcpprovider/v1beta1"
-
+	machineapi "github.com/openshift/api/machine/v1beta1"
```
staebler, line 7: Leave the machineapi import in a section separate from the installer import.

staebler (commented):
Please separate the vendor changes in their own commit.

patrickdillon:
/test e2e-gcp

Ok, fixed it up based on the recommendations. Let's see if we get further in the tests now.

patrickdillon:
install suceeds, 2 failed tests that look unrelated

/test e2e-gcp

kirankt (commented):
It would be nice to see 2 separate commits. One for the code changes and the other for vendoring changes. Other than that it looks like the gcp test is passing, so lgtm.

patrickdillon:
Oops, I lost track of this but have now updated/rebased.

kirankt (commented):
GCP test passed.
/lgtm

kirankt:
Should we skip the other tests? They're not relevant to this PR...

patrickdillon:
> Should we skip the other tests? They're not relevant to this PR...

Sure we can 
/skip

staebler:
98e2d2a30...3e4b438b1
* Rebased and ran `go mod tidy` to fix up the go.sum file.

/lgtm

staebler:
3e4b438b1...830d0d73f
* Actually include the go.sum changes this time.

/lgtm

PR #5438, opened 2021-12-02
Alibaba: fix: add the availability zone of worker

bd233:
The `ali_zone_ids` needs to contain the availability zone of the workers,
and create the VSwitches

### DIFF
--- a/data/data/alibabacloud/variables-alibabacloud.tf
+++ b/data/data/alibabacloud/variables-alibabacloud.tf
@@ -23,9 +23,14 @@ variable "ali_private_zone_id" {
   type = string
 }
 
-variable "ali_zone_ids" {
+variable "ali_master_availability_zone_ids" {
   type        = list(string)
   description = "The availability zones in which to create the masters. The length of this list must match master_count."
+}
+
+variable "ali_worker_availability_zone_ids" {
+  type        = list(string)
+  description = "The availability zones to provision for workers. Worker instances are created by the machine-API operator, but this variable controls their supporting VSwitches."
 }
 
 variable "ali_nat_gateway_zone_id" {
--- a/data/data/alibabacloud/cluster/main.tf
+++ b/data/data/alibabacloud/cluster/main.tf
@@ -23,12 +23,17 @@ module "resource_group" {
 }
 
 module "vpc" {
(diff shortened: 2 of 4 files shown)


patrickdillon:
LGTM. I will let @kwoodson take a look
/approve

kwoodson:
I will test this one now. brb

staebler (commented):
There is other work that needs to happen first before this can go in as is. Currently, the number of master instances created is equal to the number of vswitches. There is one vswitch for every zone. If we make this change to add workers to the zones, then we will end up creating masters in every zone where we have workers, which is not what the user may want. I think this PR is dependent upon https://github.com/openshift/installer/pull/5379.

The comments for the following terraform variables need to be changed.
https://github.com/openshift/installer/blob/cd69a4bda52fe5e4cf4d250717529ba9e8811ba4/data/data/alibabacloud/variables-alibabacloud.tf#L16
https://github.com/openshift/installer/blob/cd69a4bda52fe5e4cf4d250717529ba9e8811ba4/data/data/alibabacloud/cluster/vpc/variables.tf#L11

bd233:
@staebler  I have update description.

> I think this PR is dependent upon #5379.

That's it.

data/data/alibabacloud/variables-alibabacloud.tf
```diff
@@ -25,7 +25,7 @@ variable "ali_private_zone_id" {
 
 variable "ali_zone_ids" {
   type        = list(string)
-  description = "The availability zones in which to create the masters. The length of this list must match master_count."
+  description = "The availability zones in which to create the masters, workers and bootstrap node."
```
staebler, line 28: This change will break the masters module. The zone ids passed into the master module must be an ordered list corresponding to the masters that need to be created. There needs to be one variable for the master zone ids and a separate variable for the worker zone ids.

staebler (changes requested):

data/data/alibabacloud/variables-alibabacloud.tf
bd233: Yes, your are right. I have updated.

staebler (approved):
/lgtm
/approve

PR #5439, opened 2021-12-02
[MCO-8] machineconfig: Use KernelArguments field instead of kernel-args file

LorbusChris:
The MC KernelArguments field is the recommended way to set kernel args,
and should be used in new installs as opposed to writing kernel args to
the `/etc/pivot/kernel-args` file, which should be considered
deprecated.

This should be considered a stop-gap solution until in-Ignition kargs support has landed in the MCO.

/cc @sinnykumari @yuqi-zhang @craychee @cgwalters 

/hold
for testing

### DIFF
--- a/pkg/asset/machines/machineconfig/hyperthreading.go
+++ b/pkg/asset/machines/machineconfig/hyperthreading.go
@@ -11,16 +11,11 @@ import (
 )
 
 // ForHyperthreadingDisabled creates the MachineConfig to disable hyperthreading.
-// RHCOS ships with pivot.service that uses the `/etc/pivot/kernel-args` to override the kernel arguments for hosts.
+// RHCOS ships with pivot.service to override the kernel arguments for hosts.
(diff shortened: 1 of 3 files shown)


LorbusChris:
I'll have to update the tests, too.
/label wip

sinnykumari:
Thanks for the PR, can you also add steps on how to test this?

LorbusChris:
I've updated the unit tests, but I don't think the kargs that are actually in use are gathered anywhere in CI, so that might have to be checked manually.

staebler (commented):
/approve

I will leave it up to the CoreOS team to give it a lgtm once they are satisfied with the test results.

cgwalters:
I think using the old interface was semi-intentional as far as ensuring it was tested.  But...OTOH, yes everyone should be using the MCO path now.  So...
/approve

LorbusChris:
@sinnykumari I have added instructions for testing this on the commit message
/hold cancel

sinnykumari:
Missed this PR earlier.
/lgtm

PR #5443, opened 2021-12-02
Azure Stack Bootstrap Destroy Bug

patrickdillon:
There was a bug where the wrong platform name was being set for Azure Stack Hub bootstrap destroy.

### DIFF
--- a/pkg/asset/cluster/cluster.go
+++ b/pkg/asset/cluster/cluster.go
@@ -72,6 +72,11 @@ func (c *Cluster) Generate(parents asset.Parents) (err error) {
 	}
 
 	platform := installConfig.Config.Platform.Name()
+
+	if azure := installConfig.Config.Platform.Azure; azure != nil && azure.CloudName == typesazure.StackCloud {
+		platform = typesazure.StackTerraformName
+	}
+
 	stages := platformstages.StagesForPlatform(platform)
 
 	logrus.Infof("Creating infrastructure resources...")
@@ -80,12 +85,9 @@ func (c *Cluster) Generate(parents asset.Parents) (err error) {
 		if err := aws.PreTerraform(context.TODO(), clusterID.InfraID, installConfig); err != nil {
 			return err
 		}
-	case typesazure.Name:
+	case typesazure.Name, typesazure.StackTerraformName:
 		if err := azure.PreTerraform(context.TODO(), clusterID.InfraID, installConfig); err != nil {
 			return err
-		}
-		if installConfig.Config.Platform.Azure.CloudName == typesazure.StackCloud {
-			platform = "azurestack"
 		}
 	}
 
--- a/pkg/destroy/bootstrap/bootstrap.go
+++ b/pkg/destroy/bootstrap/bootstrap.go
@@ -41,7 +41,7 @@ func Destroy(dir string) (err error) {
 	// Set platform to azurestack after setting tfPlatformVarsFileName, because the
 	// tfvars file is still terraform.azure.auto.tfvars.json in the case of Azure Stack.
 	if platform == typesazure.Name && metadata.Azure.CloudName == typesazure.StackCloud {
-		platform = "azurestack"
+		platform = typesazure.StackTerraformName
 	}
 
 	varFiles := []string{cluster.TfVarsFileName, tfPlatformVarsFileName}
--- a/pkg/terraform/stages/azure/stages.go
+++ b/pkg/terraform/stages/azure/stages.go
@@ -3,21 +3,19 @@ package azure
 import (
 	"github.com/openshift/installer/pkg/terraform"
 	"github.com/openshift/installer/pkg/terraform/stages"
+	typesazure "github.com/openshift/installer/pkg/types/azure"
 )
-
-const azure string = "azure"
-const azurestack string = "azurestack"
 
 // PlatformStages are the stages to run to provision the infrastructure in Azure.
 var PlatformStages = []terraform.Stage{
-	stages.NewStage(azure, "vnet"),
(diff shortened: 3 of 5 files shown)


rna-afk:
/lgtm 
I'm now wondering how my bootstrap gather code fix was working :(

pkg/destroy/bootstrap/bootstrap.go
```diff
@@ -41,7 +41,7 @@ func Destroy(dir string) (err error) {
 	// Set platform to azurestack after setting tfPlatformVarsFileName, because the
```
squidboylan, line 41: Should we update this comment? I assume `platform = typesazure.StackCloud.Name()` returns the name of the specific azurestack in use.

patrickdillon: Fixed, but the new value is still a constant `AzureStackCloud` instead of `azurestack`

patrickdillon:
> /lgtm I'm now wondering how my bootstrap destroy code fix was working :(

I don't think it is working now, perhaps because of the recent change, but originally it probably worked because the Azure & Azure Stack code must have been compatible. That is, the gather code is invoking the Azure code, but it works for Azure Stack Hub. Probably not what we want to do. 

I will update that PR.

patrickdillon:
Doh. I'm deleting my previous comment about this not working. My test was wrong. Hopefully you missed my incorrect comment.

pkg/destroy/bootstrap/bootstrap.go
```diff
@@ -38,10 +38,10 @@ func Destroy(dir string) (err error) {
 	tfPlatformVarsFileName := fmt.Sprintf(cluster.TfPlatformVarsFileName, platform)
 
 	// Azure Stack uses the Azure platform but has its own Terraform configuration.
-	// Set platform to azurestack after setting tfPlatformVarsFileName, because the
+	// Set platform to AzureStackCloud after setting tfPlatformVarsFileName, because the
 	// tfvars file is still terraform.azure.auto.tfvars.json in the case of Azure Stack.
 	if platform == typesazure.Name && metadata.Azure.CloudName == typesazure.StackCloud {
-		platform = "azurestack"
+		platform = typesazure.StackCloud.Name()
```
staebler, line 44: I don't think that this is the best solution. We put the tf files under data/data/azurestack as so we call the platform "azurestack" in the stages.
https://github.com/openshift/installer/blob/ec1c83027c3963abb273f412c60472db58bf1860/pkg/terraform/stages/azure/stages.go#L20-L22
So I think we should call the platform "azurestack" here as well, for consistency. We should instead the following line to "azurestack".
https://github.com/openshift/installer/blob/ec1c83027c3963abb273f412c60472db58bf1860/pkg/terraform/stages/platform/stages.go#L31

patrickdillon: @staebler, your point makes sense. I ended up going a bit further and eliminated the use of the "azurestack" magic string. IMHO, it is better to nip this in the bud and clean it up now (mess I introduced); but if you think this is overkill let me know.

pkg/asset/cluster/cluster.go
```diff
@@ -80,13 +85,12 @@ func (c *Cluster) Generate(parents asset.Parents) (err error) {
 		if err := aws.PreTerraform(context.TODO(), clusterID.InfraID, installConfig); err != nil {
 			return err
 		}
+	case typesazure.StackName:
+		fallthrough
 	case typesazure.Name:
```
staebler, line 90:
```suggestion
	case typesazure.Name, typesazure.StackName:
```

pkg/terraform/stages/azure/stages.go
```diff
@@ -3,10 +3,11 @@ package azure
 import (
 	"github.com/openshift/installer/pkg/terraform"
 	"github.com/openshift/installer/pkg/terraform/stages"
+	typesazure "github.com/openshift/installer/pkg/types/azure"
 )
 
-const azure string = "azure"
-const azurestack string = "azurestack"
+const azure string = typesazure.Name
+const azurestack string = typesazure.StackName
```
staebler, line 10: Let's just get rid of these constants.

pkg/types/azure/doc.go
```diff
@@ -2,5 +2,8 @@
 // configuration and management.
 package azure
 
-// Name is name for the Azure platform.
+// Name is the name for the Azure platform.
 const Name string = "azure"
+
+// StackName is the name for the Azure Stack platform.
+const StackName string = "azurestack"
```
staebler, line 9: I am concerned that this constant here is going to add to confusion making it seem like azurestack is a separate platform.
```suggestion
// StackTerraformName is the name used for the terraform code when installing to the Azure Stack cloud environment.
const StackTerraformName string = "azurestack"
```

pkg/destroy/bootstrap/bootstrap.go
```diff
@@ -41,7 +41,7 @@ func Destroy(dir string) (err error) {
 	// Set platform to azurestack after setting tfPlatformVarsFileName, because the
 	// tfvars file is still terraform.azure.auto.tfvars.json in the case of Azure Stack.
```
staebler, line 42: Not for this PR, but we should consider whether there is any benefit in using a different name for the tfvars file for each platform. We never have tfvars for multiple platforms simultaneously, so we could just always use the name `terraform.platform.auto.tfvars.json` for the file.

patrickdillon:
Updated. Ready for review.

staebler (approved):
/lgtm
/approve

pkg/destroy/bootstrap/bootstrap.go
staebler: https://github.com/openshift/installer/pull/5477

PR #5444, opened 2021-12-02
pkg/types: Correct docs for deprecated options

stephenfin:
The `machineCIDR` config option wasn't replaced by `machineCIDRs` - it was replaced by `machineNetwork[*].cidr`. Similarly, `serviceNetwork` isn't a like-for-like replacement for `serviceCIDR`: the latter took a single value while the former takes a list. Update docs to reflect this, fixing an unrelated bug in the process.

### DIFF
--- a/pkg/types/installconfig.go
+++ b/pkg/types/installconfig.go
@@ -281,8 +281,8 @@ type Networking struct {
 
 	// Deprecated types, scheduled to be removed
 
-	// Deprecated name for MachineCIDRs. If set, MachineCIDRs must
-	// be empty or the first index must match.
+	// Deprecated way to configure an IP address pool for machines.
+	// Replaced by MachineNetwork which allows for multiple pools.
 	// +optional
 	DeprecatedMachineCIDR *ipnet.IPNet `json:"machineCIDR,omitempty"`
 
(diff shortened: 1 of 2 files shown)


stephenfin:
Looks like the unrelated bug that I spotted here is https://github.com/openshift/installer/pull/5406. I'm trusting that git will auto-resolve the conflict if that merges first.

pkg/types/installconfig.go
```diff
@@ -281,16 +281,17 @@ type Networking struct {
 
 	// Deprecated types, scheduled to be removed
 
-	// Deprecated name for MachineCIDRs. If set, MachineCIDRs must
-	// be empty or the first index must match.
+	// Deprecated way to configure an IP address pool for machines.
+	// Replaced by MachineNetwork which allows for multiple pools
```
staebler, line 285:
```suggestion
	// Replaced by MachineNetwork which allows for multiple pools.
```

pkg/types/installconfig.go
```diff
@@ -281,16 +281,17 @@ type Networking struct {
...
 	// +optional
 	DeprecatedMachineCIDR *ipnet.IPNet `json:"machineCIDR,omitempty"`
 
 	// Deprecated name for NetworkType
 	// +optional
 	DeprecatedType string `json:"type,omitempty"`
 
-	// Deprecated name for ServiceNetwork
+	// Deprecated way to configure an IP address pool for services.
+	// Replaced by ServiceNetwork which allows for multiple pools
```
staebler, line 294:
```suggestion
	// Replaced by ServiceNetwork which allows for multiple pools.
```

staebler (commented):
Looks good to me except for the nit of missing periods.
/approve

stephenfin:
Thanks for the review :+1: I've added the missing periods

staebler:
> Thanks for the review +1 I've added the missing periods

Need to re-run `go generate ./pkg/types/installconfig.go`.

staebler (approved):
/lgtm

PR #5446, opened 2021-12-02
Enforce IBM Cloud cluster name limitations

cjschaef:
Validate the cluster name provided meets the restrictions on IBM
Cloud. Compress the SecurityGroup names for IBM Cloud to make sure
those names can meet the IBM Cloud restricitions (character limits).

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2027498

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -87,6 +87,13 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
 		// FIX-ME: As soon bz#1915122 get resolved remove the limitation of 14 chars for the clustername
 		nameErr = validate.ClusterNameMaxLength(c.ObjectMeta.Name, 14)
 	}
+	if c.Platform.IBMCloud != nil {
+		ibmNameErr := validate.ClusterNameIBMMaxLength(c.ObjectMeta.Name)
+		// Override cluster name error if name does not meet IBM Cloud restrictions
+		if ibmNameErr != nil {
+			nameErr = ibmNameErr
+		}
+	}
 	if nameErr != nil {
 		allErrs = append(allErrs, field.Invalid(field.NewPath("metadata", "name"), c.ObjectMeta.Name, nameErr.Error()))
 	}
--- a/pkg/validate/validate.go
+++ b/pkg/validate/validate.go
@@ -18,6 +18,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation"
 )
 
+// IBMCloudClusterNameMaxLength is the maximum allowed length for a cluster name on IBM Cloud
+const IBMCloudClusterNameMaxLength = 32
+
 var (
 	// DockerBridgeCIDR is the network range that is used by default network for docker.
 	DockerBridgeCIDR = func() *net.IPNet {
@@ -145,6 +148,11 @@ func ClusterName(v string) error {
 		return err
 	}
 	return validateSubdomain(v)
+}
+
+// ClusterNameIBMMaxLength checks if the given string meets the length restrictions for IBM Cloud
+func ClusterNameIBMMaxLength(v string) error {
+	return ClusterNameMaxLength(v, IBMCloudClusterNameMaxLength)
 }
 
 // SubnetCIDR checks if the given IP net is a valid CIDR.
--- a/pkg/types/validation/installconfig_test.go
+++ b/pkg/types/validation/installconfig_test.go
@@ -3,6 +3,7 @@ package validation
 import (
 	"fmt"
 	"net"
+	"strings"
 	"testing"
 
 	"github.com/pborman/uuid"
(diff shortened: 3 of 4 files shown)


data/data/ibmcloud/network/vpc/security-groups.tf
```diff
@@ -17,7 +17,7 @@ locals {
 ############################################
 
 resource "ibm_is_security_group" "cluster_wide" {
-  name           = "${local.prefix}-security-group-cluster-wide"
+  name           = "${local.prefix}-sg-cluster-wide"
```
kirankt, line 20: The var names are clear and descriptive. IMO, we should leave them be and not shorten them.

kirankt, line 20: The var names are clear and descriptive. IMO, we should leave them be and not shorten them.

kirankt (commented):
Looks like gofmt is unhappy. Please either run "hack/gofmt.sh ." or "gofmt -w" directly on the changed files.

pkg/asset/installconfig/ibmcloud/validation.go
```diff
@@ -13,10 +14,15 @@ import (
...
 
+var (
+	ibmCloudClusterNameRegex = regexp.MustCompile(`^([0-9A-Za-z_.-]){1,32}$`)
+)
+
 // Validate executes platform-specific validation.
 func Validate(client API, ic *types.InstallConfig) error {
 	allErrs := field.ErrorList{}
 	platformPath := field.NewPath("platform").Child("ibmcloud")
+	allErrs = append(allErrs, validateClusterName(ic)...)
```
staebler, line 25: This should be done in static validation similar to https://github.com/openshift/installer/blob/794d9514f71f587360b7f755fb1cb6f65f1510c0/pkg/types/validation/installconfig.go#L86-L89.

data/data/ibmcloud/network/vpc/security-groups.tf
staebler: Unfortunately, the names need to be shortened because of limitations that IBM Cloud imposes on the length of the name.

pkg/types/validation/installconfig.go
```diff
@@ -87,6 +87,9 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
 		// FIX-ME: As soon bz#1915122 get resolved remove the limitation of 14 chars for the clustername
 		nameErr = validate.ClusterNameMaxLength(c.ObjectMeta.Name, 14)
 	}
+	if c.Platform.IBMCloud != nil {
+		nameErr = validate.ClusterNameIBMRestrictions(c.ObjectMeta.Name)
```
staebler, line 91: This validation accepts cluster names that would be rejected by `validate.ClusterName`. It looks like there is actually a bug in the validation for ovirt where the validation is only checking the length and ignoring any errors from `validate.ClusterName`. Let's not make the same mistake here. My suggestion is to either (1) create a new function in the `validate` package that will accept a max length and do the same thing that `validate.ClusterName` does or (2) validate the length first and then call `validate.ClusterName` if and only if the length is valid.

staebler: Your changes now change the behavior for other platforms. Please keep your changes specific to the ibmcloud platform.

cjschaef: Ah, yes, I think I follow what you mean, given platforms can override the `error` with `nil` potentially.

data/data/ibmcloud/network/vpc/security-groups.tf
cjschaef: I moved these to a different PR, since for now, they are currently blocking CI setup
https://github.com/openshift/installer/pull/5455

pkg/types/validation/installconfig.go
cjschaef: Not sure if the new approach is acceptable (since `validate.ClusterName` is still run first), as I think it reduces code complexity.

pkg/validate/validate.go
```diff
@@ -147,6 +148,14 @@ func ClusterName(v string) error {
 	return validateSubdomain(v)
 }
 
+// ClusterNameIBMRestrictions checks if the given string is a valid name for a cluster on IBM Cloud and returns an error if not
+func ClusterNameIBMRestrictions(v string) error {
+	if !ibmCloudClusterNameRegex.MatchString(v) {
```
staebler, line 153: Do we need a regex for this? Or can we just validate the name length? I think the character restrictions are already handled by `validate.ClusterName`.

cjschaef: Sure, I will simplify the check to be length only, and leave the character restrictions to the `validate.ClusterName` instead.

pkg/validate/validate.go
```diff
@@ -18,6 +18,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation"
 )
 
+// IBMCloudClusterNameMaxLength is the maximum allowed length for a cluster name on IBM Cloud
+const IBMCloudClusterNameMaxLength = 32
```
staebler, line 22: Please add to this comment for how this number is determined.

cjschaef: Oh, this was based on an IBM Kubernetes restriction, which isn't applicable here.
I shall add the formula and new value, once I work that out calculation.

cjschaef:
Closing this PR, as the length restrictions for IBM Cloud cluster names is not necessary, the existing `ClusterID` logic is sufficient, now that we have trimmed down the `SecurityGroup` names in a separate PR, which was causing the major issue related to name character length.

PR #5447, opened 2021-12-02
close a verify-codegen ci failure loophole

clnperez:
if the `go codegen` commands failed, the the `hack/verify-codegen.sh` script continued and the CI check passed. see https://github.com/openshift/installer/pull/5406 for an example. the script was not set to fail on a non-zero return code. this pr adds the bash `e` option to the `set` builtin to catch errors and exit on failure.

### DIFF
--- a/hack/verify-codegen.sh
+++ b/hack/verify-codegen.sh
@@ -1,10 +1,11 @@
 #!/bin/sh
 
 if [ "$IS_CONTAINER" != "" ]; then
-  set -x
+  set -xe
   go generate ./pkg/types/installconfig.go
-  go generate ./pkg/rhcos/ami.go
-  set +x
+  # See https://github.com/openshift/installer/pull/5447#discussion_r762340594
+  # go generate ./pkg/rhcos/ami.go
+  set +ex
   git diff --exit-code
 else
   podman run --rm \


clnperez:
The codegen tests fail now as they should:

```
Container test exited with code 1, reason Error
---
+ go generate ./pkg/types/installconfig.go
github.com/openshift/installer/pkg/types:-: conflicting types in allOf branches in schema: string vs DiskCategory
github.com/openshift/installer/pkg/types:-: conflicting types in allOf branches in schema: string vs DiskCategory
github.com/openshift/installer/pkg/types:-: conflicting types in allOf branches in schema: string vs DiskCategory
Error: not all generators ran successfully 
```

We should merge https://github.com/openshift/installer/pull/5406 first

clnperez:
fyi @patrickdillon

squidboylan:
looks good to me

hack/verify-codegen.sh
```diff
@@ -3,7 +3,15 @@
 if [ "$IS_CONTAINER" != "" ]; then
   set -x
```
staebler, line 4: Just remove the `set -x` and `set +x` instead of going through the machinations of checking the return codes.

hack/verify-codegen.sh
```diff
@@ -3,7 +3,15 @@
...
   go generate ./pkg/types/installconfig.go
+  rc=$?
+  if [ $rc != 0 ]; then
+      exit 1
+  fi
   go generate ./pkg/rhcos/ami.go
```
staebler, line 10: This is going to fail because the file no longer exists. We need a replacement for this. But we can comment it out for now.

hack/verify-codegen.sh
clnperez: i added `e` instead

clnperez:
@staebler @jhixson74 PTAL at the updates

staebler (commented):
The changes look good.
Please improve upon the PR and commit title so that they are not quite so generic.

clnperez:
i updated the summary comment and pr title. hopefully gh picks those up.

staebler:
> i updated the summary comment and pr title. hopefully gh picks those up.

You need to push up the commit with the updated message.

clnperez:
@staebler done

hack/verify-codegen.sh
```diff
@@ -1,11 +1,12 @@
...
 if [ "$IS_CONTAINER" != "" ]; then
-  set -x
+  set -xe
   go generate ./pkg/types/installconfig.go
-  go generate ./pkg/rhcos/ami.go
-  set +x
+  # See https://github.com/openshift/installer/pull/5447#discussion_r762340594
+  # go generate ./pkg/rhcos/ami.go
   git diff --exit-code
+  set +ex
```
staebler, line 9: Let's leave this up above the `git diff` as it was.

clnperez: oops. yup.

staebler (approved):
/lgtm
/approve

hack/verify-codegen.sh
```diff
@@ -1,10 +1,11 @@
 #!/bin/sh
 
 if [ "$IS_CONTAINER" != "" ]; then
-  set -x
+  set -xe
   go generate ./pkg/types/installconfig.go
-  go generate ./pkg/rhcos/ami.go
-  set +x
+  # See https://github.com/openshift/installer/pull/5447#discussion_r762340594
+  # go generate ./pkg/rhcos/ami.go
```
staebler, line 7: Fixed with https://github.com/openshift/installer/pull/5466.

clnperez: sounds like i should just delete those two commented-out lines now then @staebler?

staebler: No, leave them commented out. This PR will likely merge before mine.

PR #5448, opened 2021-12-03
azure: Fix terraform cluster IP output

rna-afk:
Fixing an output produced by the cluster stage of create command
in Azure where the stage provided a list of lists as output for
the IP addresses. This could be made as simple as a list hence
making the change to pick only the first IP address of each
control plane node.


staebler (commented):
/lgtm
/approve
/hold for azure tests
/test e2e-azure

rna-afk:
e2e-azure test passed
/hold cancel

PR #5451, opened 2021-12-06
Revert "open cluster-policy-controller ports for metrics"

atiratree:
This reverts commit 2636aef6cdf0f897f98446e29c969d61b6b009a7 (https://github.com/openshift/installer/pull/5105).

Opening port 10357 has to be reverted due to the problems with opening ports during cluster upgrades and moving to a different solution.


staebler (commented):
/lgtm
/approve

PR #5453, opened 2021-12-06
docs/dev/pinned-coreos.md: update rhcos-stream.json example

saqibali-2k:
The file data/data/rhcos-stream.json was deleted in https://github.com/openshift/installer/commit/d773ee557311fda59565bb82e947d7d55001252b
and the corresponding data now lives in data/data/coreos/rhcos.json. 
Let's update the documentation to reflect this change.


travier (approved):
/lgtm

PR #5454, opened 2021-12-06
Migrate `ibmcloud` from hidden to an installer recognized/supported platform

jeffnowicki:
Migrate `ibmcloud` platform from hidden so that installer builds recognizes `ibmcloud` as a supported platform.


staebler (commented):
/lgtm
/approve

kirankt (commented):
/lgtm

PR #5457, opened 2021-12-07
Bug 2020216: replace public-access with sas token in bootstrap deployment

MayXuQQ:
remove --public-access blob,use sas token in BOOTSTRAP_URL
https://bugzilla.redhat.com/show_bug.cgi?id=2020216#c7

### DIFF
--- a/docs/user/azure/install_upi.md
+++ b/docs/user/azure/install_upi.md
@@ -229,7 +229,7 @@ done
 Create a blob storage container and upload the generated `bootstrap.ign` file:
 
 ```sh
-az storage container create --name files --account-name ${CLUSTER_NAME}sa --public-access blob
+az storage container create --name files --account-name ${CLUSTER_NAME}sa
 az storage blob upload --account-name ${CLUSTER_NAME}sa --account-key $ACCOUNT_KEY -c "files" -f "bootstrap.ign" -n "bootstrap.ign"
 ```
 
@@ -342,7 +342,8 @@ Copy the [`04_bootstrap.json`](../../../upi/azure/04_bootstrap.json) ARM templat
 Create the deployment using the `az` client:
 
 ```sh
-export BOOTSTRAP_URL=`az storage blob url --account-name ${CLUSTER_NAME}sa --account-key $ACCOUNT_KEY -c "files" -n "bootstrap.ign" -o tsv`
+bootstrap_url_expiry=`date -u -d "10 hours" '+%Y-%m-%dT%H:%MZ'`
+export BOOTSTRAP_URL=`az storage blob generate-sas -c 'files' -n 'bootstrap.ign' --https-only --full-uri --permissions r --expiry $bootstrap_url_expiry --account-name ${CLUSTER_NAME}sa --account-key $ACCOUNT_KEY -o tsv`
 export BOOTSTRAP_IGNITION=`jq -rcnM --arg v "3.1.0" --arg url $BOOTSTRAP_URL '{ignition:{version:$v,config:{replace:{source:$url}}}}' | base64 | tr -d '\n'`
 
 az deployment group create -g $RESOURCE_GROUP \


docs/user/azure/install_upi.md
```diff
@@ -209,7 +209,7 @@ export VHD_URL=`curl -s https://raw.githubusercontent.com/openshift/installer/ma
 Copy the chosen VHD to a blob:
 
 ```sh
-az storage container create --name vhd --account-name ${CLUSTER_NAME}sa
+az storage container create --name vhd --account-name ${CLUSTER_NAME}sa --account-key ${ACCOUNT_KEY}
```
jianlinliu, line 212: Per our QE testing result, I think we do not need "--account-key" option on this step, right?

docs/user/azure/install_upi.md
```diff
@@ -229,7 +229,7 @@ done
 Create a blob storage container and upload the generated `bootstrap.ign` file:
 
 ```sh
-az storage container create --name files --account-name ${CLUSTER_NAME}sa --public-access blob
+az storage container create --name files --account-name ${CLUSTER_NAME}sa --account-key $ACCOUNT_KEY
```
jianlinliu, line 232: The same comments as the above.

jianlinliu:
suggest open separate PR for different issues.

docs/user/azure/install_upi.md
MayXuQQ: az storage container create --name vhd --account-name ${CLUSTER_NAME}sa
There are no credentials provided in your command and environment, we will query for account key for your storage account.
It is recommended to provide --connection-string, --account-key or --sas-token in your command as credentials.

jianlinliu: It is just a warning, when no `--account-key` is provided, it will default to use the az login's service principal to access. And QE's original script was working well, also prove this command does not need `--account-key` as a MUST. If we have no specific complaint about that, I think it is better to not touch with it, only focus on `--public-access blob` issue.

docs/user/azure/install_upi.md
```diff
@@ -194,10 +194,10 @@ export ACCOUNT_KEY=`az storage account keys list -g $RESOURCE_GROUP --account-na
 Given the size of the RHCOS VHD, it's not possible to run the deployments with this file stored locally on your machine.
 We must copy and store it in a storage container instead. To do so, first create a blob storage container and then copy the VHD.
 
-Choose the RHCOS version you'd like to use and export the URL of its VHD to an environment variable. For example, to use the latest release available for the 4.3 version, use:
+Choose the RHCOS version you'd like to use and export the URL of its VHD to an environment variable. For example, to use the available of the openshift-installer version, use:
```
staebler, line 197: The grammar in the changed sentence does not make sense.

docs/user/azure/install_upi.md
```diff
@@ -194,10 +194,10 @@ export ACCOUNT_KEY=`az storage account keys list -g $RESOURCE_GROUP --account-na
...
 
 ```sh
-export VHD_URL=`curl -s https://raw.githubusercontent.com/openshift/installer/release-4.3/data/data/rhcos.json | jq -r .azure.url`
+export VHD_URL=$(./openshift-install coreos print-stream-json | jq -r '.architectures.x86_64."rhel-coreos-extensions"."azure-disk".url')
```
staebler, line 200:
```suggestion
export VHD_URL=$(openshift-install coreos print-stream-json | jq -r '.architectures.x86_64."rhel-coreos-extensions"."azure-disk".url')
```

docs/user/azure/install_upi.md
```diff
@@ -342,7 +342,8 @@ Copy the [`04_bootstrap.json`](../../../upi/azure/04_bootstrap.json) ARM templat
 Create the deployment using the `az` client:
 
 ```sh
-export BOOTSTRAP_URL=`az storage blob url --account-name ${CLUSTER_NAME}sa --account-key $ACCOUNT_KEY -c "files" -n "bootstrap.ign" -o tsv`
+export end=`date -u -d "10 hours" '+%Y-%m-%dT%H:%MZ'`
```
staebler, line 345: This does not need to b exported. I would also recommend a more specific name than just "end".
```suggestion
bootstrap_url_expiry=`date -u -d "10 hours" '+%Y-%m-%dT%H:%MZ'`
```

staebler (commented):
I agree with the sentiment from @jianlinliu that there should be separate PRs for each issue.

docs/user/azure/install_upi.md
patrickdillon: As stated here, https://bugzilla.redhat.com/show_bug.cgi?id=2020216#c8 we cannot remove `--public-access blob` (without more work).

staebler (commented):
/lgtm
/approve

PR #5458, opened 2021-12-07
Validate BMC driver supported and requires provisioning network

ardaguclu:
This PR adds returning error if bmc driver requires provisioning network
and does not disabled support provisioning network cases.

### DIFF
--- a/pkg/types/baremetal/validation/platform.go
+++ b/pkg/types/baremetal/validation/platform.go
@@ -360,6 +360,21 @@ func validateBootMode(hosts []*baremetal.Host, fldPath *field.Path) (errors fiel
 	return
 }
 
+// validateProvisioningNetworkDisabledSupported validates hosts bmc address support provisioning network is disabled
+func validateProvisioningNetworkDisabledSupported(hosts []*baremetal.Host, fldPath *field.Path) (errors field.ErrorList) {
+	for idx, host := range hosts {
+		accessDetails, err := bmc.NewAccessDetails(host.BMC.Address, host.BMC.DisableCertificateVerification)
+		if err != nil {
+			errors = append(errors, field.Invalid(fldPath.Index(idx).Child("BMC"), host.BMC.Address, err.Error()))
+		} else if accessDetails.RequiresProvisioningNetwork() {
+			msg := fmt.Sprintf("driver %s requires provisioning network", accessDetails.Driver())
+			errors = append(errors, field.Invalid(fldPath.Index(idx).Child("BMC"), host.BMC.Address, msg))
+		}
+	}
+
+	return
+}
+
 // ValidatePlatform checks that the specified platform is valid.
 func ValidatePlatform(p *baremetal.Platform, n *types.Networking, fldPath *field.Path, c *types.InstallConfig) field.ErrorList {
 	allErrs := field.ErrorList{}
@@ -433,6 +448,8 @@ func ValidateProvisioning(p *baremetal.Platform, n *types.Networking, fldPath *f
 	// will be run on the external network. Users must provide IP's on the
 	// machine networks to host those services.
 	case baremetal.DisabledProvisioningNetwork:
+		allErrs = validateProvisioningNetworkDisabledSupported(p.Hosts, fldPath.Child("Hosts"))
+
 		// If set, ensure bootstrapProvisioningIP is in one of the machine networks
 		if p.BootstrapProvisioningIP != "" {
 			if err := validateIPinMachineCIDR(p.BootstrapProvisioningIP, n); err != nil {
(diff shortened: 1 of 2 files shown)


ardaguclu:
@dtantsur you commented in here https://github.com/openshift/installer/blob/f0b973614bca9296b2384127372ca206a02048d9/pkg/types/baremetal/validation/platform.go#L352 for the cases what we should do when bmc driver is not supported or get error.

Perhaps, validating all hosts' bmc addresses in any case and catching unsupported ones is a good way to detect misconfiguration. I changed the code that way and wonder your opinions about this. Thanks.

dtantsur:
/lgtm

Not sure I'm fond of iterating of BMCs and creating AccessDetails twice.. but it shouldn't be a big deal.

ardaguclu:
> /lgtm
> 
> Not sure I'm fond of iterating of BMCs and creating AccessDetails twice.. but it shouldn't be a big deal.

Thanks @dtantsur. Yes, I'm curious about the exact same thing. But one of them is ValidatePlatform and the other is ValidateProvision, I thought that they are different phases and will not hurt do that twice

ardaguclu:
@hardys @andfasano could you please take a look at this validation?, thanks.

pkg/types/baremetal/validation/platform.go
```diff
@@ -338,24 +338,52 @@ func validateNetworkConfig(hosts []*baremetal.Host, fldPath *field.Path) (errors
...
 
-// ensure that the bootMode field contains a valid value
-func validateBootMode(hosts []*baremetal.Host, fldPath *field.Path) (errors field.ErrorList) {
+// validateBMC validates all host BMC addresses and calls custom validate function
+func validateBMC(hosts []*baremetal.Host, fldPath *field.Path, validateFn func(host *baremetal.Host, b bmc.AccessDetails, idx int, fldPath *field.Path) field.ErrorList) field.ErrorList {
+	allErrs := field.ErrorList{}
+
 	for idx, host := range hosts {
+		accessDetails, err := bmc.NewAccessDetails(host.BMC.Address, host.BMC.DisableCertificateVerification)
+		if err != nil || accessDetails == nil {
```
andfasano, line 347: Wouldn't be better to have a separate check and a specific error message for `accessDetails == nil`?

pkg/types/baremetal/validation/platform.go
```diff
@@ -338,24 +338,52 @@ func validateNetworkConfig(hosts []*baremetal.Host, fldPath *field.Path) (errors
 	return
 }
 
-// ensure that the bootMode field contains a valid value
-func validateBootMode(hosts []*baremetal.Host, fldPath *field.Path) (errors field.ErrorList) {
+// validateBMC validates all host BMC addresses and calls custom validate function
+func validateBMC(hosts []*baremetal.Host, fldPath *field.Path, validateFn func(host *baremetal.Host, b bmc.AccessDetails, idx int, fldPath *field.Path) field.ErrorList) field.ErrorList {
```
andfasano, line 342: Don't want to be a blocking comment, but maybe the setup could be slightly simplified/more readable by adding a specific validation rule for the access details (to ensure that everything it's fine) to be invoked before `validateBootMode` and `validateProvisioningNetworkDisabledSupported` (respectively, in `ValidateProvisioning` and `ValidatePlatform`).
At this point, each validate function could simply iterate over all the hosts (as it happens in other validation rules), fetch the access details (now assuming that no error will happen, so just an additional line of code), and apply its own validation logic

ardaguclu: This PR is working, however I was thinking the same thing that hurts readability and agree with you. I'll update to simpler version like you suggested.

pkg/types/baremetal/validation/platform.go
```diff
@@ -360,6 +360,21 @@ func validateBootMode(hosts []*baremetal.Host, fldPath *field.Path) (errors fiel
 	return
 }
 
+// validateProvisioningNetworkDisabledSupported validates hosts bmc address support provisioning network is disabled
+func validateProvisioningNetworkDisabledSupported(hosts []*baremetal.Host, fldPath *field.Path) (errors field.ErrorList) {
+	for idx, host := range hosts {
+		accessDetails, err := bmc.NewAccessDetails(host.BMC.Address, host.BMC.DisableCertificateVerification)
+		if err != nil {
+			errors = append(errors, field.Invalid(fldPath.Index(idx).Child("BMC"), host.BMC.Address, err.Error()))
+		} else if accessDetails.RequiresProvisioningNetwork() {
```
andfasano, line 369: minor: for the happy path I guess there could be an immediate return in case of error (and else removed)?

andfasano: Yep, good point

ardaguclu:
@sadasu could you PTAL?. Thanks

pkg/types/baremetal/validation/platform.go
```diff
@@ -360,6 +360,21 @@ func validateBootMode(hosts []*baremetal.Host, fldPath *field.Path) (errors fiel
...
+			msg := fmt.Sprintf("driver %s requires provisioning network", accessDetails.Driver())
```
sadasu, line 370: Nit: Is it possible to add the host's name in addition to the host.BMC.Address for easier identification of the host that doesn't support this configuration?

ardaguclu:
@sadasu thanks for lgtm. Would you mind approve it?, because currently it is the label this PR needs. Thanks.

PR #5461, opened 2021-12-08
Alibaba: fix: replace the ic.ClusterName in the validateClusterName

bd233:
The `ic.ClusterName` is empty in the `validateClusterName`, and the
`ic.ObjectMeta.Name` is correct, not `ic.ClusterName`.

### DIFF
--- a/pkg/asset/installconfig/alibabacloud/validation.go
+++ b/pkg/asset/installconfig/alibabacloud/validation.go
@@ -109,7 +109,7 @@ func ValidateForProvisioning(client *Client, ic *types.InstallConfig, metadata *
 
 func validateClusterName(client *Client, ic *types.InstallConfig) field.ErrorList {
 	allErrs := field.ErrorList{}
-	namePath := field.NewPath("matedata").Child("name")
+	namePath := field.NewPath("metadata").Child("name")
 
 	zoneName := ic.ClusterDomain()
 	response, err := client.ListPrivateZones(zoneName)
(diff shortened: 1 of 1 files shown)


pkg/asset/installconfig/alibabacloud/validation.go
```diff
@@ -117,7 +117,7 @@ func validateClusterName(client *Client, ic *types.InstallConfig) field.ErrorLis
 		allErrs = append(allErrs, field.InternalError(namePath, err))
 	}
 	if response.TotalItems > 0 {
-		allErrs = append(allErrs, field.Invalid(namePath, ic.ClusterName, fmt.Sprintf("cluster name is unavailable, private zone name %s already exists", zoneName)))
+		allErrs = append(allErrs, field.Invalid(namePath, ic.ObjectMeta.Name, fmt.Sprintf("cluster name is unavailable, private zone name %s already exists", zoneName)))
```
staebler, line 120:
```suggestion
		allErrs = append(allErrs, field.Invalid(namePath, ic.Name, fmt.Sprintf("cluster name is unavailable, private zone name %s already exists", zoneName)))
```

staebler (commented):
While we are here fixing field errors, can we fix the typo in `namePath` where it uses "matedata" instead of "metadata"?

pkg/asset/installconfig/alibabacloud/validation.go
bd233: > While we are here fixing field errors, can we fix the typo in namePath where it uses "matedata" instead of "metadata"?

Thanks for reminding！

Sorry, I did not understand this suggestion, the type `types.InstallConfig` does not exist `Name`, or modify `validateClusterName(client *Client, ic *types.InstallConfig)` to `validateClusterName(client *Client, ic *ObjectMeta) `?

bd233: I remember, because `ic.ObjectMeta.Name` and `ic.Platform.Name` exist at the same time, so `ic.Name` will cause "ambiguous selector" error.

staebler (approved):
/lgtm
/approve

PR #5462, opened 2021-12-08
ibmcloud: remove resource group question from survey

staebler:
The installer supports creating a resource group for the user wheninstalling to ibmcloud. This should be the default, opinionated way that the installer in which the installer is configured. As such,the installer should not be asking in the survey for the user to provide an existing resource group to use.


kirankt (commented):
/lgtm

PR #5463, opened 2021-12-08
WIP Use MCD once-from instead of coreos-installer in BIP mode

vrutkovs:
This reprovisions bootstrap node into a master in BIP mode

### DIFF
--- a/data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/install-to-disk.sh.template
+++ b/data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/install-to-disk.sh.template
@@ -14,9 +14,22 @@ record_service_stage_success
 
 if [ ! -f coreos-installer.done ]; then
   record_service_stage_start "coreos-installer"
+  {{ if .IsOKD -}}
+
+  # Delete bootstrap manifests
+  rm -rf /etc/kubernetes/manifests/*
+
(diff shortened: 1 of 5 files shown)


data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/install-to-disk.sh.template
```diff
@@ -14,9 +14,22 @@ record_service_stage_success
 
 if [ ! -f coreos-installer.done ]; then
   record_service_stage_start "coreos-installer"
+  {{ if .IsOKD -}}
+
+  # Delete bootstrap manifests
+  rm -rf /etc/kubernetes/manifests/*
+
+  # Apply new ignition using start --once-from. This doesn't require a full disk, but may
+  # leave some bootstrap files around
```
staebler, line 23: For OCP, leaving bootstrap files around is unacceptable. I would assume that OKD has the same requirements.

vrutkovs: Yes, leaving bootstrap files around is not a great idea - but this PR is the best we've got so far, I'll leave a FIXME note

vrutkovs:
No longer required for OKD AI flow, we found a less invasive way

PR #5464, opened 2021-12-08
bootstrap: remove support for certs lacking SAN

staebler:
This removes the deprecated support for certs which lack a SAN.

See https://github.com/openshift/installer/pull/4248


kirankt (approved):
/lgtm

PR #5465, opened 2021-12-08
Bug 2025464: aws: get ips for all control plane instances for bootstrap gather

staebler:
The tf code for aws outputs only the IP address of the first control plane instance rather than the IPs of all of the control plane instances. The issue is that all of the IPs for all of the instances are collected into a single list and only then the first IP is taken from the list. Instead, we want to take the first IP from each instance and then collect those IPs into a list.

### DIFF
--- a/data/data/aws/cluster/master/outputs.tf
+++ b/data/data/aws/cluster/master/outputs.tf
@@ -1,3 +1,3 @@
 output "ip_addresses" {
-  value = aws_network_interface.master.*.private_ips[0]
+  value = [for m in aws_network_interface.master : tolist(m.private_ips)[0]]
 }


jstuever:
@staebler Does it make sense to do this the same way as it was handled in GCP?
https://github.com/openshift/installer/blob/fca2ee6a557dc194129ead6a7d06fcf99d84532c/data/data/gcp/cluster/master/outputs.tf#L10

staebler:
> @staebler Does it make sense to do this the same way as it was handled in GCP? I expect these might be different data structures, but it's not clear to me without looking at the specifics.
> 
> https://github.com/openshift/installer/blob/fca2ee6a557dc194129ead6a7d06fcf99d84532c/data/data/gcp/cluster/master/outputs.tf#L10

That does not work because the `private_ip` field is a set instead of a list. You can get an element by index from a set. That is the purpose of the `tolist` call, so that we can get what should be the only element in the set. In more recent version of terraform, we would have been able to use the `one` function instead.

jstuever (approved):
/lgtm
/approve

jstuever:
> That does not work because the `private_ip` field is a set instead of a list. You cannot get an element by index from a set. That is the purpose of the `tolist` call, so that we can get what should be the only element in the set. In more recent version of terraform, there is the `one` function that could be used instead.

I was struggling with the placement of the `[0]` without seeing it in action. This response totally cleared that up.

PR #5468, opened 2021-12-08
azurestack: Add a custom rhcos image field for ASH

squidboylan:
Add a new field to the install config so users can specify an RHCOS image
to use in their ASH environment. This field is required for ASH, and not
allowed to be used for azure.

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -1342,6 +1342,11 @@ spec:
                     - AzureGermanCloud
                     - AzureStackCloud
                     type: string
+                  clusterOSImage:
+                    description: ClusterOSImage is the url of a storage blob in the
+                      Azure Stack environment containing an RHCOS VHD. This field
+                      is required for Azure Stack and not applicable to Azure.
+                    type: string
                   computeSubnet:
                     description: ComputeSubnet specifies an existing subnet for use
                       by compute nodes
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"net/url"
 	"sort"
 	"strconv"
 	"strings"
@@ -41,6 +42,13 @@ func Validate(client API, ic *types.InstallConfig) error {
 	allErrs = append(allErrs, validateNetworks(client, ic.Azure, ic.Networking.MachineNetwork, field.NewPath("platform").Child("azure"))...)
 	allErrs = append(allErrs, validateRegion(client, field.NewPath("platform").Child("azure").Child("region"), ic.Azure)...)
 	allErrs = append(allErrs, validateInstanceTypes(client, ic)...)
+	if ic.Azure.CloudName == aztypes.StackCloud && ic.Azure.ClusterOSImage != "" {
+		StorageEndpointSuffix, err := client.GetStorageEndpointSuffix(context.TODO())
+		if err != nil {
+			return err
+		}
+		allErrs = append(allErrs, validateAzureStackClusterOSImage(StorageEndpointSuffix, ic.Azure.ClusterOSImage, field.NewPath("platform").Child("azure"))...)
+	}
 	return allErrs.ToAggregate()
 }
 
@@ -274,6 +282,9 @@ func ValidatePublicDNS(ic *types.InstallConfig, azureDNS *DNSConfig) error {
 func ValidateForProvisioning(client API, ic *types.InstallConfig) error {
 	allErrs := field.ErrorList{}
 	allErrs = append(allErrs, validateResourceGroup(client, field.NewPath("platform").Child("azure"), ic.Azure)...)
+	if ic.Azure.CloudName == aztypes.StackCloud {
+		allErrs = append(allErrs, checkAzureStackClusterOSImageSet(ic.Azure.ClusterOSImage, field.NewPath("platform").Child("azure"))...)
+	}
 	return allErrs.ToAggregate()
 }
 
@@ -316,6 +327,27 @@ func validateResourceGroup(client API, fieldPath *field.Path, platform *aztypes.
 			ids = ids[:2]
 		}
 		allErrs = append(allErrs, field.Invalid(fieldPath.Child("resourceGroupName"), platform.ResourceGroupName, fmt.Sprintf("resource group must be empty but it has %d resources like %s ...", l, strings.Join(ids, ", "))))
+	}
+	return allErrs
+}
+
+func checkAzureStackClusterOSImageSet(ClusterOSImage string, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	if ClusterOSImage == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("clusterOSImage"), "clusterOSImage must be set when installing on Azure Stack"))
+	}
+	return allErrs
+}
+
+func validateAzureStackClusterOSImage(StorageEndpointSuffix string, ClusterOSImage string, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	imageParsedURL, err := url.Parse(ClusterOSImage)
+	if err != nil {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("clusterOSImage"), ClusterOSImage, fmt.Errorf("clusterOSImage URL is invalid: %w", err).Error()))
+	}
+	// If the URL for the image isn't in the Azure Stack environment we can't use it.
+	if !strings.HasSuffix(imageParsedURL.Host, StorageEndpointSuffix) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("clusterOSImage"), ClusterOSImage, "clusterOSImage must be in the Azure Stack environment"))
 	}
 	return allErrs
 }
--- a/pkg/types/azure/platform.go
+++ b/pkg/types/azure/platform.go
@@ -31,6 +31,9 @@ type Platform struct {
 	// ARMEndpoint is the endpoint for the Azure API when installing on Azure Stack.
 	ARMEndpoint string `json:"armEndpoint,omitempty"`
 
+	// ClusterOSImage is the url of a storage blob in the Azure Stack environment containing an RHCOS VHD. This field is required for Azure Stack and not applicable to Azure.
+	ClusterOSImage string `json:"clusterOSImage,omitempty"`
+
 	// BaseDomainResourceGroupName specifies the resource group where the Azure DNS zone for the base domain is found.
 	BaseDomainResourceGroupName string `json:"baseDomainResourceGroupName,omitempty"`
 
--- a/pkg/types/azure/validation/platform.go
+++ b/pkg/types/azure/validation/platform.go
@@ -80,6 +80,9 @@ func ValidatePlatform(p *azure.Platform, publish types.PublishingStrategy, fldPa
 		if p.ARMEndpoint != "" {
 			allErrs = append(allErrs, field.Required(fldPath.Child("armEndpoint"), fmt.Sprintf("ARM endpoint must not be set when the cloud name is %s", cloud)))
 		}
+		if p.ClusterOSImage != "" {
+			allErrs = append(allErrs, field.Required(fldPath.Child("clusterOSImage"), fmt.Sprintf("clusterOSImage must not be set when the cloud name is %s", cloud)))
+		}
 	}
 
 	return allErrs
--- a/pkg/asset/installconfig/azure/client.go
+++ b/pkg/asset/installconfig/azure/client.go
@@ -26,6 +26,7 @@ type API interface {
 	GetDiskSkus(ctx context.Context, region string) ([]azsku.ResourceSku, error)
 	GetGroup(ctx context.Context, groupName string) (*azres.Group, error)
 	ListResourceIDsByGroup(ctx context.Context, groupName string) ([]string, error)
+	GetStorageEndpointSuffix(ctx context.Context) (string, error)
 }
 
 // Client makes calls to the Azure API.
@@ -92,6 +93,12 @@ func (c *Client) getVirtualNetworksClient(ctx context.Context) (*aznetwork.Virtu
 	vnetsClient := aznetwork.NewVirtualNetworksClientWithBaseURI(c.ssn.Environment.ResourceManagerEndpoint, c.ssn.Credentials.SubscriptionID)
 	vnetsClient.Authorizer = c.ssn.Authorizer
 	return &vnetsClient, nil
+}
+
+// GetStorageEndpointSuffix retrieves the StorageEndpointSuffix from the
+// session environment
+func (c *Client) GetStorageEndpointSuffix(ctx context.Context) (string, error) {
+	return c.ssn.Environment.StorageEndpointSuffix, nil
 }
 
 // getSubnetsClient sets up a new client to retrieve a subnet
--- a/pkg/asset/installconfig/azure/mock/azureclient_generated.go
+++ b/pkg/asset/installconfig/azure/mock/azureclient_generated.go
@@ -6,53 +6,39 @@ package mock
 
 import (
 	context "context"
+	reflect "reflect"
+
 	compute "github.com/Azure/azure-sdk-for-go/profiles/2018-03-01/compute/mgmt/compute"
 	network "github.com/Azure/azure-sdk-for-go/profiles/2018-03-01/network/mgmt/network"
 	resources "github.com/Azure/azure-sdk-for-go/profiles/2018-03-01/resources/mgmt/resources"
 	subscriptions "github.com/Azure/azure-sdk-for-go/profiles/2018-03-01/resources/mgmt/subscriptions"
 	gomock "github.com/golang/mock/gomock"
-	reflect "reflect"
 )
 
-// MockAPI is a mock of API interface
+// MockAPI is a mock of API interface.
 type MockAPI struct {
 	ctrl     *gomock.Controller
 	recorder *MockAPIMockRecorder
 }
 
-// MockAPIMockRecorder is the mock recorder for MockAPI
+// MockAPIMockRecorder is the mock recorder for MockAPI.
 type MockAPIMockRecorder struct {
 	mock *MockAPI
 }
 
-// NewMockAPI creates a new mock instance
+// NewMockAPI creates a new mock instance.
 func NewMockAPI(ctrl *gomock.Controller) *MockAPI {
 	mock := &MockAPI{ctrl: ctrl}
 	mock.recorder = &MockAPIMockRecorder{mock}
 	return mock
 }
 
-// EXPECT returns an object that allows the caller to indicate expected use
+// EXPECT returns an object that allows the caller to indicate expected use.
 func (m *MockAPI) EXPECT() *MockAPIMockRecorder {
 	return m.recorder
 }
 
-// GetVirtualNetwork mocks base method
-func (m *MockAPI) GetVirtualNetwork(ctx context.Context, resourceGroupName, virtualNetwork string) (*network.VirtualNetwork, error) {
-	m.ctrl.T.Helper()
-	ret := m.ctrl.Call(m, "GetVirtualNetwork", ctx, resourceGroupName, virtualNetwork)
-	ret0, _ := ret[0].(*network.VirtualNetwork)
-	ret1, _ := ret[1].(error)
-	return ret0, ret1
-}
-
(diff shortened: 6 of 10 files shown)


pkg/types/azure/platform.go
```diff
@@ -31,6 +31,9 @@ type Platform struct {
 	// ARMEndpoint is the endpoint for the Azure API when installing on Azure Stack.
 	ARMEndpoint string `json:"armEndpoint,omitempty"`
 
+	// RHCOSImage is the url for the RHCOS image to use, required for Azure Stack and cannot be used with Azure.
+	RHCOSImage string `json:"RHCOSImage,omitempty"`
```
rna-afk, line 35: I forgot about the explain command. You can add it to the description of explain command by running
```
```

pkg/tfvars/azure/azure.go
```diff
@@ -76,6 +78,16 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
 		return nil, errors.Wrap(err, "could not determine Azure environment to use for Terraform")
 	}
 
+	imageURL := sources.ImageURL
+	_, ok := os.LookupEnv("OPENSHIFT_INSTALL_OS_IMAGE_OVERRIDE")
```
rna-afk, line 82: os seems to be not imported here.

pkg/types/azure/validation/platform.go
```diff
@@ -80,6 +80,9 @@ func ValidatePlatform(p *azure.Platform, publish types.PublishingStrategy, fldPa
 		if p.ARMEndpoint != "" {
 			allErrs = append(allErrs, field.Required(fldPath.Child("armEndpoint"), fmt.Sprintf("ARM endpoint must not be set when the cloud name is %s", cloud)))
 		}
+		if p.RHCOSImage != "" {
+			allErrs = append(allErrs, field.Required(fldPath.Child("RHCOSImage"), fmt.Sprintf("RHCOS Image must not be set when the cloud name is %s", cloud)))
```
rna-afk, line 84: I forgot to mention you also need to check the platform_test.go to see if the existing tests are impacted by this change and also might need to add tests.

rna-afk (commented):
Looks good. Just a few changes.

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -1300,6 +1300,10 @@ spec:
               azure:
                 description: Azure is the configuration used when installing on Azure.
                 properties:
+                  RHCOSImage:
```
rna-afk, line 1303: There's also a test in the explain pkg that needs to be changed. 
https://github.com/openshift/installer/blob/master/pkg/explain/printer_test.go

pkg/types/azure/validation/platform.go
```diff
@@ -106,6 +110,15 @@ func validateAzureStack(p *azure.Platform, fldPath *field.Path) field.ErrorList
 	if p.ARMEndpoint == "" {
 		allErrs = append(allErrs, field.Required(fldPath.Child("armEndpoint"), "ARM endpoint must be set when installing on Azure Stack"))
 	}
+	if p.RHCOSImage == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("RHCOSImage"), "RHCOS Image must be set when installing on Azure Stack"))
+	} else {
+		storageBaseURL := strings.TrimPrefix(p.ARMEndpoint, "https://management.")
+		// If the URL for the image isn't in the ASH environment we can't use it.
+		if !strings.Contains(p.RHCOSImage, storageBaseURL) {
```
rna-afk, line 118: I think we also need to address Matthew's original comment here 
https://github.com/openshift/installer/pull/5103/commits/f4de0c078b5fb4d8f01e76b94befcf34695a40a9#r725196893

pkg/tfvars/azure/azure.go
```diff
@@ -76,6 +78,12 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
...
+	if sources.CloudName == azure.StackCloud && ok {
+		imageURL = sources.RHCOSImage
+	}
```
patrickdillon, line 85: Instead of doing the switch here, consider returning the value of `RHCOSImage` here https://github.com/openshift/installer/blob/master/pkg/asset/rhcos/image.go#L128-L137  (when it's azurestack). To me this is cleaner, and the OVERRIDE environment variable is already handled.

patrickdillon:
This documentation should be updated: https://github.com/openshift/installer/blob/master/docs/user/azure/customization.md

pkg/types/azure/validation/platform.go
```diff
@@ -106,6 +110,15 @@ func validateAzureStack(p *azure.Platform, fldPath *field.Path) field.ErrorList
 	if p.ARMEndpoint == "" {
 		allErrs = append(allErrs, field.Required(fldPath.Child("armEndpoint"), "ARM endpoint must be set when installing on Azure Stack"))
 	}
+	if p.RHCOSImage == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("RHCOSImage"), "RHCOS Image must be set when installing on Azure Stack"))
```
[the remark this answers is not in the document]
patrickdillon: When I say "we have already released UPI" I mean that we cannot break existing workflows, so once an API is released we cannot require new fields without an API version bump; instead when we add new fields we must make them optional with sane defaults.

pkg/types/azure/platform.go
```diff
@@ -31,6 +31,9 @@ type Platform struct {
 	// ARMEndpoint is the endpoint for the Azure API when installing on Azure Stack.
 	ARMEndpoint string `json:"armEndpoint,omitempty"`
 
+	// RHCOSImage is the url for the RHCOS image to use, required for Azure Stack and cannot be used with Azure.
```
patrickdillon, line 34:
```suggestion
	// RHCOSImage is the url of a storage blob in the Azure Stack environment containing an RHCOS VHD. This field is required for Azure Stack and not applicable to Azure.
```

pkg/types/validation/installconfig_test.go
```diff
@@ -1095,6 +1096,18 @@ func TestValidateInstallConfig(t *testing.T) {
...
+			name: "invalid azurestack platform RHCOSImage field",
+			installConfig: func() *types.InstallConfig {
+				c := validInstallConfig()
+				c.Platform = types.Platform{
+					Azure: validAzureStackPlatform(),
+				}
+				c.Azure.RHCOSImage = "invalid-image"
+				return c
+			}(),
+			expectedError: `RHCOSImage must be in the ASH environment`,
```
patrickdillon, line 1109: use the regex pattern of [^] [message] [$] to ensure this is the complete error message.

pkg/types/azure/validation/platform.go
```diff
@@ -106,6 +110,15 @@ func validateAzureStack(p *azure.Platform, fldPath *field.Path) field.ErrorList
...
+	} else {
+		storageBaseURL := strings.TrimPrefix(p.ARMEndpoint, "https://management.")
+		// If the URL for the image isn't in the ASH environment we can't use it.
+		if !strings.Contains(p.RHCOSImage, storageBaseURL) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("outboundType"), p.RHCOSImage, "RHCOSImage must be in the ASH environment"))
```
patrickdillon, line 119: `outboundType`? Copypasta?

"ASH" -> "Azure Stack". Don't use "ASH" in public-facing messages.

pkg/tfvars/azure/azure.go
```diff
@@ -49,6 +50,7 @@ type TFVarsSources struct {
 	Auth                        Auth
 	CloudName                   azure.CloudEnvironment
 	ARMEndpoint                 string
+	RHCOSImage                  string
```
patrickdillon, line 53: I'm happy with the field name `RHCOSImage` but let's consider others for the sake of brainstorming. 

This field refers to the URL of a storage blob which contains a VHD with RHCOS. Personally I'm leaning to the name `VHDURL`

Other possibilities: `OSImageURL`, `VHDBlob`. 

To me `VHD

squidboylan: Yeah I was also unsure about the field name, I'm happy for whatever and dont really have strong feelings one way or another

pkg/tfvars/azure/azure.go
squidboylan: This makes sense, I wasnt aware of that code path, i think i've added what you suggested

pkg/asset/installconfig/azure/validation.go
```diff
@@ -319,3 +321,25 @@ func validateResourceGroup(client API, fieldPath *field.Path, platform *aztypes.
 	}
 	return allErrs
 }
+
+func validateAzureStackRHCOSImage(ARMEndpoint string, RHCOSImage string, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	if RHCOSImage == "" {
```
patrickdillon, line 327: Won't this fail on public Azure or other non-ASH environments?

pkg/tfvars/azure/azure.go
patrickdillon: Let's go with `VHDURL`

pkg/asset/installconfig/azure/validation.go
```diff
@@ -274,6 +275,7 @@ func ValidatePublicDNS(ic *types.InstallConfig, azureDNS *DNSConfig) error {
 func ValidateForProvisioning(client API, ic *types.InstallConfig) error {
 	allErrs := field.ErrorList{}
 	allErrs = append(allErrs, validateResourceGroup(client, field.NewPath("platform").Child("azure"), ic.Azure)...)
+	allErrs = append(allErrs, validateAzureStackRHCOSImage(ic.Azure.ARMEndpoint, ic.Azure.RHCOSImage, field.NewPath("platform").Child("azure"))...)
```
patrickdillon, line 278: Same as my other comment , we should either only run this conditionally for Azure Stack or return nil for non-azurestack.

patrickdillon (commented):
Big idea looks good, just some minor changes. 

/test e2e-azure

I would expect this to fail based on my comments about how the field seems to be incorrectly required for azure.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -319,3 +323,25 @@ func validateResourceGroup(client API, fieldPath *field.Path, platform *aztypes.
...
+		storageParsedURL, err := url.Parse(ARMEndpoint)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("armEndpoint"), ARMEndpoint, "Failed to parse ARMEndpoint URL"))
+		}
+		storageBaseHost := strings.TrimPrefix(storageParsedURL.Host, "management.")
+		imageParsedURL, err := url.Parse(VHDURL)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("VHDURL"), VHDURL, "Failed to parse VHDURL URL"))
+		}
+		// If the URL for the image isn't in the Azure Stack environment we can't use it.
```
staebler, line 341: If the URL is required to be part of the stack environment, is there benefit in querying azure to validate that the blob actually exists?

pkg/types/azure/platform.go
```diff
@@ -31,6 +31,9 @@ type Platform struct {
 	// ARMEndpoint is the endpoint for the Azure API when installing on Azure Stack.
 	ARMEndpoint string `json:"armEndpoint,omitempty"`
 
+	// VHDURL is the url of a storage blob in the Azure Stack environment containing an RHCOS VHD. This field is required for Azure Stack and not applicable to Azure.
+	VHDURL string `json:"VHDURL,omitempty"`
```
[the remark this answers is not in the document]
patrickdillon: I didn't realize there was a convention for other platforms. So I agree `ClusterOSImage` is good. Sorry for the mixed signals. Hopefully renaming a symbol is easy in your IDE.

pkg/asset/installconfig/azure/validation.go
patrickdillon: Yes in the sense that we would fail earlier and more gracefully than a Terraform failure.

staebler: I guess my question is do we have a sense of (1) how easy it is to go from the URL to querying the azure api for the blob and (2) how fragile that translation may be. The check for the existence would need to wait until the pre-terraform step, too.

patrickdillon: The blob URL should be predictable and easily translatable. But thinking about this more I wonder about permissions. I think it would be a valid use case to have a VHD in a blob in a separate subscription that the installer would not necessarily have permission to access via the api.

staebler: What determines whether the VM is allowed to use the image if it is not the permissions used to create the VM? In other words, I would expect that if I have permission to create a VM using an image that I should also have permission to read that image. But, then, that is just my expectation and not a reflection on the reality of things.

staebler: Nevertheless, we can certainly skip doing the extra validation that the blob exists. The error will come soon enough when the terraform runs.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -319,3 +323,25 @@ func validateResourceGroup(client API, fieldPath *field.Path, platform *aztypes.
...
+		if !strings.HasSuffix(imageParsedURL.Host, storageBaseHost) {
```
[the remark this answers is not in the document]
squidboylan: From what I can tell it doesn't look like the installer knows that value at this point.

pkg/asset/installconfig/azure/validation.go
patrickdillon: It is not permission to read the image, but rather the storage blob containing the VHD that will be used to create the image. In public Azure, Red Hat has a public storage blob `https://rhcos.blob.core.windows.net/imagebucket/rhcos-410.84.202112040202-0-azure.x86_64.vhd` that we copy into the installation target resource group, but I am doubting whether a user's credentials could use the Azure storage blob API to verify the existence of `https://rhcos.blob.core.windows.net/imagebucket/rhcos-410.84.202112040202-0-azure.x86_64.vhd`. I *think* there could be a similar valid use case in Azure Stack.

Also, this makes me wonder if the step of copying the VHD into the storage account is redundant for Azure Stack: https://github.com/openshift/installer/blob/master/data/data/azurestack/vnet/main.tf#L54-L67

I think we have the capability of testing both of these questions, but perhaps it would be better to scope those into a new PR.

pkg/asset/installconfig/azure/validation.go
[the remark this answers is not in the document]
patrickdillon: I think it should be the `StorageEndpointSuffix` (although not 100% certain). The installer should be able to retrieve that value from the session environment. You can create a session like this: https://github.com/openshift/installer/blob/master/pkg/asset/cluster/azure/azure.go#L38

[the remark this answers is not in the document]
squidboylan: I dont think I have access to that `installconfig.InstallConfig` object, the InstallConfig provided seems to be https://github.com/openshift/installer/blob/master/pkg/types/installconfig.go#L70 and I don't think I can get to the Session through that.

pkg/types/azure/platform.go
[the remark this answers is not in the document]
squidboylan: No worries, xargs and sed make things easy

pkg/asset/installconfig/azure/validation.go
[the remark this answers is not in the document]
patrickdillon: Ah, I see. You can change the parameters to `ValidateForProvisioning`:  https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/platformprovisioncheck.go#L70. For instance you could pass in the `ic.Azure.Environment`

patrickdillon:
In general this looks good to me. I'm going to do some testing.

Note there are some linting/formatting failures in the tests.

squidboylan:
I updated the `ClusterOSImage` to be represented in json as `clusterOSImage` as well for consistency

pkg/asset/installconfig/azure/validation.go
```diff
@@ -319,3 +327,20 @@ func validateResourceGroup(client API, fieldPath *field.Path, platform *aztypes.
 	}
 	return allErrs
 }
+
+func validateAzureStackClusterOSImage(StorageEndpointSuffix string, ClusterOSImage string, fldPath *field.Path) field.ErrorList {
+	var allErrs field.ErrorList
+	if ClusterOSImage == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("clusterOSImage"), "clusterOSImage must be set when installing on Azure Stack"))
+	} else {
```
patrickdillon, line 335: This should be the only check that is run in validateForProvisioning.

The other checks should be performed in the main Validate function of install config asset: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/azure/validation.go#L38

That validate function already has a client where you can grab the environment

patrickdillon (commented):
I steered you astray in our earlier conversations. Instead of the current solution, we should:
platformProvisionCheck: only check that ClusterOSImage is set
normal install config validation: if ClusterOSImage is set, do the rest of the validation

This has two beneifts:
1. Code is a little cleaner because you don't have to jump through hoops to get the environment (this is where I misled earlier)
2. invalid install config will fail earlier in the `create manifest` phase

patrickdillon:
LGTM. Can you squash the commits?

pkg/asset/installconfig/azure/validation.go
```diff
@@ -271,9 +279,12 @@ func ValidatePublicDNS(ic *types.InstallConfig, azureDNS *DNSConfig) error {
 }
 
 // ValidateForProvisioning validates if the isntall config if valid for provisioning the cluster.
-func ValidateForProvisioning(client API, ic *types.InstallConfig) error {
+func ValidateForProvisioning(client API, ic *types.InstallConfig, metadata *Metadata) error {
```
patrickdillon, line 282: We no longer need the `metadata` passed in here.

pkg/tfvars/azure/azure.go
```diff
@@ -50,6 +50,7 @@ type TFVarsSources struct {
 	Auth                            Auth
 	CloudName                       azure.CloudEnvironment
 	ARMEndpoint                     string
+	ClusterOSImage                  string
```
staebler, line 53: This field is unused.

staebler (commented):
/test e2e-azure
/test e2e-azurestack
/hold

pkg/asset/cluster/tfvars.go
```diff
@@ -342,6 +342,7 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 				Auth:                            auth,
 				CloudName:                       installConfig.Config.Azure.CloudName,
 				ARMEndpoint:                     installConfig.Config.Azure.ARMEndpoint,
+				ClusterOSImage:                  installConfig.Config.Azure.ClusterOSImage,
```
patrickdillon, line 345: Ah, similar to as staebler pointed out in the other comment, I don't think this is used either.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -41,6 +42,13 @@ func Validate(client API, ic *types.InstallConfig) error {
 	allErrs = append(allErrs, validateNetworks(client, ic.Azure, ic.Networking.MachineNetwork, field.NewPath("platform").Child("azure"))...)
 	allErrs = append(allErrs, validateRegion(client, field.NewPath("platform").Child("azure").Child("region"), ic.Azure)...)
 	allErrs = append(allErrs, validateInstanceTypes(client, ic)...)
+	if ic.Azure.CloudName == aztypes.StackCloud {
```
patrickdillon, line 45: either
```suggestion
	if ic.Azure.CloudName == aztypes.StackCloud && ic.Azure.ClusterOSImage != "" {
```

or just

```suggestion
	if ic.Azure.ClusterOSImage != "" {
```

The current scenario will cause failures in UPI or not the best error message on IPI:

```shell
# yq e .platform.azure c/install-config.yaml 
armEndpoint: https://management.ppe3.stackpoc.com
baseDomainResourceGroupName: openshiftInstallerRG
cloudName: AzureStackCloud
region: ppe3

# ./openshift-install create manifests --dir c
INFO Credentials loaded from file "/root/.azure/osServicePrincipal.json" 
FATAL failed to fetch Master Machines: failed to load asset "Install Config": platform.azure.clusterOSImage: Invalid value: "": clusterOSImage must be in the Azure Stack environment 
```

patrickdillon:
/hold cancel

Locally tested azure.

staebler:
No changes that affect baremetal.
/override ci/prow/e2e-metal-ipi-ovn-ipv6-required

PR #5469, opened 2021-12-09
azure upi: get VHD_URL with openshift-install coreos print-stream-json

MayXuQQ:
To avoid changing the VHD_URL with the release version, get VHD_URL with "openshift-install coreos print-stream-json"

### DIFF
--- a/docs/user/azure/install_upi.md
+++ b/docs/user/azure/install_upi.md
@@ -194,22 +194,9 @@ export ACCOUNT_KEY=`az storage account keys list -g $RESOURCE_GROUP --account-na
 Given the size of the RHCOS VHD, it's not possible to run the deployments with this file stored locally on your machine.
 We must copy and store it in a storage container instead. To do so, first create a blob storage container and then copy the VHD.
 
-Choose the RHCOS version you'd like to use and export the URL of its VHD to an environment variable. For example, to use the latest release available for the 4.3 version, use:
-
-```sh
-export VHD_URL=`curl -s https://raw.githubusercontent.com/openshift/installer/release-4.3/data/data/rhcos.json | jq -r .azure.url`
-```
-
-If you'd just like to use the latest _development_ version available (master branch), use:
-
-```sh
-export VHD_URL=`curl -s https://raw.githubusercontent.com/openshift/installer/master/data/data/rhcos.json | jq -r .azure.url`
-```
-
-Copy the chosen VHD to a blob:
-
 ```sh
 az storage container create --name vhd --account-name ${CLUSTER_NAME}sa
+export VHD_URL=$(openshift-install coreos print-stream-json | jq -r '.architectures.x86_64."rhel-coreos-extensions"."azure-disk".url')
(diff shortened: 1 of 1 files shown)


patrickdillon:
This LGTM 

Would you mind making a similar change for Azure Stack? 
The variable is called COMPRESSED_VHD_URL https://github.com/openshift/installer/blob/master/docs/user/azure/install_upi_azurestack.md#copy-the-cluster-image

It could be in a separate PR

docs/user/azure/install_upi.md
```diff
@@ -194,10 +194,10 @@ export ACCOUNT_KEY=`az storage account keys list -g $RESOURCE_GROUP --account-na
 Given the size of the RHCOS VHD, it's not possible to run the deployments with this file stored locally on your machine.
 We must copy and store it in a storage container instead. To do so, first create a blob storage container and then copy the VHD.
 
-Choose the RHCOS version you'd like to use and export the URL of its VHD to an environment variable. For example, to use the latest release available for the 4.3 version, use:
+Choose the RHCOS version you'd like to use and export the URL of its VHD to an environment variable. For example, to use the version of the openshift-installer bootimages, use:
 
 ```sh
-export VHD_URL=`curl -s https://raw.githubusercontent.com/openshift/installer/release-4.3/data/data/rhcos.json | jq -r .azure.url`
+export VHD_URL=$(openshift-install coreos print-stream-json | jq -r '.architectures.x86_64."rhel-coreos-extensions"."azure-disk".url')
```
staebler, line 200: The [next part right after this](https://github.com/openshift/installer/blob/20ee3cbe302fb196ee55953ac37320e0e0639dab/docs/user/azure/install_upi.md?plain=1#L206) needs to be changed, too.

I recommend that we just get rid of the section. Even if the user wants to use the master branch, the recommended approach is still to run `openshift-install coreos print-stream-json`.

MayXuQQ: How  about update as above?

docs/user/azure/install_upi.md
```diff
@@ -194,16 +194,10 @@ export ACCOUNT_KEY=`az storage account keys list -g $RESOURCE_GROUP --account-na
 Given the size of the RHCOS VHD, it's not possible to run the deployments with this file stored locally on your machine.
 We must copy and store it in a storage container instead. To do so, first create a blob storage container and then copy the VHD.
 
-Choose the RHCOS version you'd like to use and export the URL of its VHD to an environment variable. For example, to use the latest release available for the 4.3 version, use:
+For example, to use the version of the openshift-installer bootimages, use:
```
staebler, line 197: I recommend we omit this entirely. We are not providing an *example* way of obtaining the VHD URL. We are providing the one recommended way of obtaining the VHD URL.

Let's combine the export of the VHD_URL environment variable in with the shell code for copying to the blob.

staebler (approved):
/lgtm
/approve

staebler:
@MayXuQQ What was the acccc67 to 40d9f5d change?

MayXuQQ:
> @MayXuQQ What was the [acccc67](https://github.com/openshift/installer/commit/acccc6769402ce5a19d91c9629f368fa73d7c33a) to [40d9f5d](https://github.com/openshift/installer/commit/40d9f5de8b9438a3f7931e4c8d22bca758dbe417) change?

I just squash the two commits.

PR #5470, opened 2021-12-10
azurestack upi : get VHD_URL with openshift-install coreos print-stream-json

MayXuQQ:
To avoid changing the COMPRESSED_VHD_URL with the release version, get COMPRESSED_VHD_URL with "openshift-install coreos print-stream-json"

### DIFF
--- a/docs/user/azure/install_upi_azurestack.md
+++ b/docs/user/azure/install_upi_azurestack.md
@@ -276,7 +276,7 @@ storage blob.
 First, download and decompress the VHD, note that the decompressed file is 16GB:
 
 ```sh
-$ export COMPRESSED_VHD_URL=$(curl -s https://raw.githubusercontent.com/openshift/installer/release-4.9/data/data/rhcos-amd64.json | jq -r '(.baseURI + .images.azurestack.path)')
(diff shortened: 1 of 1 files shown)


docs/user/azure/install_upi_azurestack.md
```diff
@@ -276,7 +276,7 @@ storage blob.
 First, download and decompress the VHD, note that the decompressed file is 16GB:
 
 ```sh
-$ export COMPRESSED_VHD_URL=$(curl -s https://raw.githubusercontent.com/openshift/installer/release-4.9/data/data/rhcos-amd64.json | jq -r '(.baseURI + .images.azurestack.path)')
+$ export COMPRESSED_VHD_URL=$(openshift-install coreos print-stream-json | jq -r '.architectures.x86_64."rhel-coreos-extensions"."azure-disk".url')
```
patrickdillon, line 279: Azure Stack uses a different image than Azure (and this is using the Azure disk). I think the jq command should be:

```
openshift-install coreos print-stream-json | jq -r '.architectures.x86_64.artifacts.azurestack.formats."vhd.gz".disk.location'
```

PR #5473, opened 2021-12-11
Baremetal: Deploy using CoreOS ISO from the release payload

zaneb:
Complete the implementation of [Include the CoreOS image in the release for baremetal](https://github.com/openshift/enhancements/blob/master/enhancements/baremetal/coreos-image-in-release.md#include-the-coreos-image-in-the-release-for-baremetal) and [Baremetal IPI Network Configuration](https://github.com/openshift/enhancements/blob/master/enhancements/network/baremetal-ipi-network-configuration.md#baremetal-ipi-network-configuration) by switching installation to use the CoreOS images from the machine-os-images container in the release payload, for both masters and workers. Also, customise the master network config using the image-customization-controller. (Worker network config was already set up by #5247.)

This PR depends on openshift/cluster-baremetal-operator#208.


data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -152,11 +140,45 @@ $IPTABLES -t raw -A DHCP_IRONIC -j DROP
...
+sudo podman run -d -p 8084:8084 --privileged --name image-customization \
+    --env DEPLOY_ISO="/shared/html/images/ironic-python-agent.iso" \
+    --env DEPLOY_INITRD="/shared/html/images/ironic-python-agent.initramfs" \
+    --env IRONIC_BASE_URL="http://{{.PlatformData.BareMetal.APIVIP}}" \
+    --env IRONIC_RAMDISK_SSH_KEY="$IRONIC_RAMDISK_SSH_KEY" \
+    --env IRONIC_AGENT_IMAGE="$IRONIC_AGENT_IMAGE" \
+    --env IRONIC_AGENT_PULL_SECRET="{{.PlatformData.BareMetal.PullSecretBase64}}" \
+    --env PROVISIONING_INTERFACE=$PROVISIONING_NIC \
+    --env REGISTRIES_CONF_PATH=/tmp/containers/registries.conf \
+    --entrypoint '["/image-customization-server", "--nmstate-dir=/tmp/nmstate/", "--images-publish-addr=http://0.0.0.0:8084"]' \
```
dtantsur, line 177: Hardcoding 0.0.0.0 may not play well with IPv6-only environments.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -152,11 +140,45 @@ $IPTABLES -t raw -A DHCP_IRONIC -j DROP
...
+        done
+        oc get -n openshift-machine-api secret ${secret_name} -o jsonpath="{.data.nmstate}" | base64 -d > /tmp/nmstate/{{.Name}}.yaml
+    else
+        touch /tmp/nmstate/{{.Name}}.yaml
+    fi
+{{end}}
+
+
+# Embed agent ignition into the rhcos live iso
+sudo podman run -d -p 8084:8084 --privileged --name image-customization \
```
dtantsur, line 168: Does it really have to be privileged?

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
zaneb: This is not actually used anywhere - the server doesn't report back the URL to us (it's not like 0.0.0.0 would work either). We just infer what the URL will be. So I think it's fine.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
zaneb: Based on experience with CBO, probably yes because all of the image files in the cache are owned by root. This may be fixable with careful examination of what we are running as privileged.

zaneb:
Baremetal tests are not expected to pass until after openshift/cluster-baremetal-operator#208 merges.

pkg/asset/machines/master.go
```diff
@@ -155,6 +155,8 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 	mign := &machine.Master{}
 	dependencies.Get(clusterID, installConfig, rhcosImage, mign)
 
+	masterUserDataSecretName := "master-user-data"
```
staebler, line 158: I am happy to take these changes, but please include them in a separate PR.

Also, use a constant at the package scope rather than a variable at the function scope.

pkg/asset/machines/master.go
```diff
@@ -359,6 +359,9 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 		mpool.Set(pool.Platform.BareMetal)
 		pool.Platform.BareMetal = &mpool
 
+		// Use managed user data secret, since we always have up to date images
+		// available in the cluster
+		masterUserDataSecretName = "master-user-data-managed"
```
staebler, line 364: Is this change a requirement for deploying the CoreOS ISO from the release payload? Or was it just convenient to include this change in this PR?

staebler (commented):
Mostly looks fine to me. My biggest concern is over continued support for the `ClusterOSImage` field in the install config.

pkg/asset/rhcos/image.go
```diff
@@ -140,14 +140,8 @@ func osImage(config *types.InstallConfig) (string, error) {
...
-
-		// Note that baremetal IPI currently uses the OpenStack image
-		// because this contains the necessary ironic config drive
-		// ignition support, which isn't enabled in the UPI BM images
-		if a, ok := streamArch.Artifacts["openstack"]; ok {
-			return rhcos.FindArtifactURL(a)
-		}
-		return "", fmt.Errorf("%s: No openstack build found", st.FormatPrefix(archName))
+		// Use image from release payload
+		return "", nil
```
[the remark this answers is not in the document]
zaneb: Yes, I failed to notice that it could be collapsed, although the comment is perhaps helpful to make explicit what is going on.

If the user provides a value here we'll still pass it to the CBO in the Provisioning resource, and CBO will still ensure it is downloaded and cached. Default MachineSets won't refer to it any more but the user could still create a MachineSet that used the image they specified here and it would work.

We just don't default it to the OpenStack QCOW any more because we don't need that image by default; the new default is not to download any image.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -152,11 +149,53 @@ $IPTABLES -t raw -A DHCP_IRONIC -j DROP
...
+        oc get -n openshift-machine-api secret ${secret_name} -o jsonpath="{.data.nmstate}" | base64 -d > /tmp/nmstate/{{.Name}}.yaml
+    else
+        touch /tmp/nmstate/{{.Name}}.yaml
+    fi
+{{end}}
+
+IRONIC_IP="{{.PlatformData.BareMetal.APIVIP}}"
+# If the IP contains a colon, then it's an IPv6 address, and the HTTP
+# host needs surrounding with brackets
+if [[ "$IRONIC_IP" =~ .*:.* ]]; then
```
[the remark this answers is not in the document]
zaneb: I copied this from a [working example](https://github.com/openshift/ironic-image/blob/master/scripts/ironic-common.sh#L41).

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
zaneb: Nope, we just need something url-like. We should probably get rid of this parameter and generate the default internally - there are two binaries in the package (this static server and a k8s controller) and a sane URL is needed only for the latter, but the args to both are the same.

pkg/asset/machines/master.go
staebler: Ah, sorry. I missed the part where this was used in the secret name at the end. Sounds good.

pkg/asset/machines/master.go
staebler: Fair enough. I was just checking because the intricacies of these changes are a bit too much for me on a quick work through the PR.

pkg/asset/rhcos/image.go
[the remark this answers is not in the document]
staebler: I agree that the comments are helpful and should be retained. If it is easier to describe the behavior by keeping the two separate returns, then that is perfectly acceptable.

zaneb:
e2e-metal-ipi-ovn-ipv6-required job timed out, so kicking it again
/test e2e-metal-ipi-ovn-ipv6-required

staebler:
Since e2e-metal-ipi-ovn-ipv6 passed,
/override ci/prow/e2e-metal-ipi-ovn-ipv6-required

PR #5475, opened 2021-12-13
openstack: Document Control plane node migration

pierreprinetti:
Provide an example script.

The script is a combination of:
* [OpenStack node migration][1]
* [OpenShift graceful node reboot][2]

Implements [OSASINFRA-2508](https://issues.redhat.com/browse/OSASINFRA-2508)

[1]: https://docs.openstack.org/nova/latest/admin/migration.html
[2]: https://docs.openshift.com/container-platform/4.9/nodes/nodes/nodes-nodes-rebooting.html#nodes-nodes-rebooting-gracefully_nodes-nodes-rebooting


mandre (commented):
Tested the script a few times to move master nodes from one host to another, it works really well, and takes between 8 and 9 minutes per node. 
/lgtm

PR #5477, opened 2021-12-13
asset/tfvars: use generic name for platform tfvars file

staebler:
Since we are not installing multiple platforms at once, there is no need to include the platform name in the tfvars file containing the platform-specific variables. Removing the platform name from the file makes the code cleaner for Azure Stack.

See https://github.com/openshift/installer/pull/5443#discussion_r766337074.


patrickdillon:
/test e2e-gcp
/test e2e-azure
/approve
/lgtm
/hold 
for tests, feel free to cancel

PR #5479, opened 2021-12-13
Alibaba: fix instance type validation

bd233:
Should not add `len(zones) == 0` to judge. When the user does not specify `zones`, `availableZones` is an
empty list (this means that the available zones are not queried according to `instanceType`) errors will not be return.


staebler:
From the description of the PR, it is not clear to me what the issue is. Can you provide example input, expected output, and actual output for the issue?

PR #5481, opened 2021-12-13
[WIP] Decouple Terraform from installer

jhixson74:
Split out Terraform and providers into embedded binaries. There is much to do here, but would like to start getting eyes on it and iron everything out. I will update this later with more details.


fabianofranz:
This looks very promising, thank you folks! @jhixson74 @staebler

clnperez:
Is the plan to eventually not need the openshift terraform forks for some of the providers?

staebler:
> Is the plan to eventually not need the openshift terraform forks for some of the providers?

That is not an explicit goal of this work. We will be moving to the latest upstream releases, and if we don't need to carry any patches, then we would not use a fork.

clnperez:
thanks @staebler. I think I may have misunderstood some other conversations I've seen spread around and thought that the v2 sdk would be supported in 4.11, so we (ibmcloud/power vs) wouldn't need to rely on the openshift fork for the v1 sdk.

staebler:
@clnperez You are not incorrect. The v2 sdk will be supported in 4.11. Each terraform provider will be run from its own binary, as intended by terraform. Consequently, each terraform provider can use whichever version of the SDK they want.

PR #5483, opened 2021-12-13
aws: Update our local zone filtering code

squidboylan:
We have updated our aws-sdk-go dependency to a version that exposes the
ZoneType field on AvailabilityZone objects, we can use this field to
filter out local zones instead of our temporary fix that used the format
of the zone name.

### DIFF
--- a/pkg/asset/installconfig/aws/availabilityzones.go
+++ b/pkg/asset/installconfig/aws/availabilityzones.go
@@ -2,7 +2,6 @@ package aws
 
 import (
 	"context"
-	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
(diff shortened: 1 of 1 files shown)


pkg/asset/installconfig/aws/availabilityzones.go
```diff
@@ -29,19 +28,9 @@ func availabilityZones(ctx context.Context, session *session.Session, region str
...
-	// all regions follow the same format with 2 hyphens.
-	// This is a temporary fix until we update our aws-sdk-go package to a
-	// version that adds the ZoneType field to the AvailabilityZone struct
-	numberOfHyphensInRegion := strings.Count(region, "-")
-
 	zones := []string{}
 	for _, zone := range resp.AvailabilityZones {
-		numberOfHyphensInZone := strings.Count(*zone.ZoneName, "-")
-		if numberOfHyphensInZone != numberOfHyphensInRegion+2 {
+		if *zone.ZoneType != "local-zone" {
```
staebler, line 33: I suggest that we enforce that the zone *must be* an availability zone rather than simply not a local zone.
```suggestion
		if *zone.ZoneType == "availability-zone" {
```

staebler (approved):
/lgtm
/approve

PR #5486, opened 2021-12-14
images/libvirt: provide specific version of google cloud sdk

cooktheryan:
Proposed fix based on the changes to CRC https://github.com/code-ready/crc/commit/b85050b66be6b0923e73db5d7de1c3fabe1f4633

Currently, google-cloud-sdk is broken and issuing the following error when launching and running jobs.  

err
```gcloud crashed (SystemError): ffi_prep_closure(): bad user_data (it seems that the version of the libffi library seen at runtime is different from the 'ffi.h' file seen at compile-time```

Providing a specific version should allow jobs to no longer fail

### DIFF
--- a/images/libvirt/Dockerfile.ci
+++ b/images/libvirt/Dockerfile.ci
@@ -18,7 +18,7 @@ RUN yum update -y && \
     yum install --setopt=tsflags=nodocs -y \
     genisoimage \
     gettext \
-    google-cloud-sdk \
+    google-cloud-sdk-365.0.1 \
     libvirt-client \
     libvirt-libs \
     nss_wrapper \


cooktheryan:
based on @sallyom sugggestion

miabbott (approved):
This looks like it will fix the problem I was trying to solve in #5467, so LGTM

staebler:
Please improve on the PR title and the commit message.

Also, please open a separate PR to add an OWNERS file under images/libvirt.

staebler:
> @staebler patched. I don't actually know who should be the owners of the file. I know that a few of the jobs that run out of openshift/releases use this image

https://github.com/openshift/installer/blob/master/data/data/libvirt/OWNERS

staebler:
> @staebler patched. I don't actually know who should be the owners of the file. I know that a few of the jobs that run out of openshift/releases use this image

You still need to push up the amended commit. The commit title should indicate that the change is to the libvirt image.

PR #5490, opened 2021-12-16
Bug 2033239: Alibaba: fix the wrong default region

bd233:
Because the `defaultRegion` is `cn-hangzhou`, so when the step
 of selecting the region is executed, if you do nothing and just
 press enter, the default value `cn-hangzhou` will be displayed,
 although it seems to be the first choice options.

Signed-off-by: sunhui <wb-sh373163@alibaba-inc.com>

### DIFF
--- a/pkg/asset/installconfig/alibabacloud/alibabacloud.go
+++ b/pkg/asset/installconfig/alibabacloud/alibabacloud.go
@@ -41,12 +41,21 @@ func selectRegion(client *Client) (string, error) {
 	}
 	regions := regionsResponse.Regions.Region
 
+	var defaultLongRegion string
 	longRegions := make([]string, 0, len(regions))
 	shortRegions := make([]string, 0, len(regions))
 	for _, location := range regions {
-		longRegions = append(longRegions, fmt.Sprintf("%s (%s)", location.RegionId, location.LocalName))
+		longRegion := fmt.Sprintf("%s (%s)", location.RegionId, location.LocalName)
+		longRegions = append(longRegions, longRegion)
 		shortRegions = append(shortRegions, location.RegionId)
+		if location.RegionId == defaultRegion {
+			defaultLongRegion = longRegion
+		}
 	}
+	if defaultLongRegion == "" {
+		return "", errors.Errorf("installer bug: invalid default alibabacloud region %q", defaultRegion)
+	}
+
 	var regionTransform survey.Transformer = func(ans interface{}) interface{} {
 		switch v := ans.(type) {
 		case core.OptionAnswer:
@@ -66,7 +75,7 @@ func selectRegion(client *Client) (string, error) {
 			Prompt: &survey.Select{
 				Message: "Region",
 				Help:    "The Alibaba Cloud region to be used for installation.",
-				Default: defaultRegion,
+				Default: defaultLongRegion,
 				Options: longRegions,
 			},
 			Validate: survey.ComposeValidators(survey.Required, func(ans interface{}) error {


pkg/asset/installconfig/alibabacloud/alibabacloud.go
```diff
@@ -66,7 +66,7 @@ func selectRegion(client *Client) (string, error) {
 			Prompt: &survey.Select{
 				Message: "Region",
 				Help:    "The Alibaba Cloud region to be used for installation.",
-				Default: defaultRegion,
+				Default: longRegions[0],
```
staebler, line 69: I don't think that this is the correct solution. We want the region to default to cn-hangzhou if the user does not make an explicit selection, rather than defaulting to whatever happens to be the first region in the list. The issue in the BZ was that the region shown to the user as the default was the first region rather than cn-hangzhou.

The value for "Default" should be the entry in `longRegions` that corresponds to the cn-hangzhou region.

jianli-wei:
@bd233 @staebler I tried with the Slack app "cluster-bot", by "build openshift/installer#5490". Then with the generated build, there's no the original issue of https://bugzilla.redhat.com/show_bug.cgi?id=2033239. But according to the above comment from Matthew, the expected behavior should be, the region "cn-hangzhou (China (Hangzhou))" should be shown as the default selection, rather than the alpha-numerically first one "ap-northeast-1 (Japan (Tokyo))". Could you please confirm? Thanks!

bd233:
@jianli-wei  @staebler I have updated the behavior of specifying default region, please review

pkg/asset/installconfig/alibabacloud/alibabacloud.go
```diff
@@ -41,12 +41,19 @@ func selectRegion(client *Client) (string, error) {
...
 	shortRegions := make([]string, 0, len(regions))
 	for _, location := range regions {
 		longRegions = append(longRegions, fmt.Sprintf("%s (%s)", location.RegionId, location.LocalName))
 		shortRegions = append(shortRegions, location.RegionId)
+		regionMap[location.RegionId] = location.LocalName
 	}
+	_, ok := regionMap[defaultRegion]
+	if !ok {
+		return "", errors.Errorf("installer bug: invalid default alibabacloud region %q", defaultRegion)
+	}
```
staebler, line 55: I don't think there is much value in using a map here.
```suggestion
	var defaultLongRegion string
	longRegions := make([]string, 0, len(regions))
	shortRegions := make([]string, 0, len(regions))
	for _, location := range regions {
		longRegion := fmt.Sprintf("%s (%s)", location.RegionId, location.LocalName)
		longRegions = append(longRegions, longRegion)
		shortRegions = append(shortRegions, location.RegionId)
		if location.RegionId == defaultRegion {
			defaultLongRegion = longRegion
		}
	}
```

bd233: Have  changed^^

pkg/asset/installconfig/alibabacloud/alibabacloud.go
```diff
@@ -66,7 +75,7 @@ func selectRegion(client *Client) (string, error) {
 			Prompt: &survey.Select{
 				Message: "Region",
 				Help:    "The Alibaba Cloud region to be used for installation.",
-				Default: defaultRegion,
+				Default: fmt.Sprintf("%s (%s)", defaultRegion, defaultLongRegion),
```
staebler, line 78:
```suggestion
				Default: defaultLongRegion,
```

bd233: AHa...have updated

staebler (commented):
/lgtm
/approve

PR #5494, opened 2021-12-16
Bug 1965969: aws: Fix dns destroy logs

squidboylan:


pkg/destroy/aws/shared.go
```diff
@@ -181,7 +181,6 @@ func (o *ClusterUninstaller) cleanSharedRoute53(ctx context.Context, session *se
 	if err != nil {
 		return err
 	}
-	logger = logger.WithField("id", id)
```
staebler: I think that we should keep this so that the log events such as "Cleaned record sets from hosted zone" include the ID of the hosted zone. It is not a huge deal, because the log events will already have the ARN of the hosted zone, but I think it is a bit nicer to be consistent by having the "id" field.

In the `if publicZoneID != ""` section of `cleanSharedHostedZone`, create a new logger with the "id" field set to the public hosted zone for use in deleting the matching public record.
```
publicZoneLogger := logger.WithField("id", id)
```

pkg/destroy/aws/shared.go
```diff
@@ -218,21 +217,22 @@ func (o *ClusterUninstaller) cleanSharedHostedZone(ctx context.Context, client *
...
 					continue
 				}
-				recordSetLogger := logger.WithField(
-					"recordset",
-					fmt.Sprintf("%s (%s)", aws.StringValue(recordSet.Name), aws.StringValue(recordSet.Type)),
-				)
 				// delete any matching record sets in the public hosted zone
 				if publicZoneID != "" {
-					if err := deleteMatchingRecordSetInPublicZone(ctx, client, publicZoneID, recordSet, logger); err != nil {
+					publicZoneLogger := logger.WithField("id", id)
```
staebler, line 222:
```suggestion
					publicZoneLogger := logger.WithField("id", publicZoneID)
```

pkg/destroy/aws/shared.go
```diff
@@ -218,21 +217,22 @@ func (o *ClusterUninstaller) cleanSharedHostedZone(ctx context.Context, client *
...
 						lastError = errors.Wrapf(err, "deleting record set matching %#v from public zone %s", recordSet, publicZoneID)
 						// do not delete the record set in the private zone if the delete failed in the public zone;
 						// otherwise the record set in the public zone will get leaked
 						continue
 					}
+					recordSetLogger := logger.WithField("id", publicZoneID).WithField(
+						"recordset",
+						fmt.Sprintf("%s (%s)", aws.StringValue(recordSet.Name), aws.StringValue(recordSet.Type)),
+					)
 					recordSetLogger.Debug("Deleted from public zone")
```
staebler, line 236:
```suggestion
					publicZoneLogger.WithField(
						"recordset",
						fmt.Sprintf("%s (%s)", aws.StringValue(recordSet.Name), aws.StringValue(recordSet.Type)),
					).Debug("Deleted from public zone")
```

pkg/destroy/aws/shared.go
staebler: Keep this line.

pkg/destroy/aws/shared.go
```diff
@@ -242,6 +242,10 @@ func (o *ClusterUninstaller) cleanSharedHostedZone(ctx context.Context, client *
 					}
 					lastError = errors.Wrapf(err, "deleting record set %#v from zone %s", recordSet, id)
 				}
+				recordSetLogger := logger.WithField("id", id).WithField(
+					"recordset",
+					fmt.Sprintf("%s (%s)", aws.StringValue(recordSet.Name), aws.StringValue(recordSet.Type)),
+				)
 				recordSetLogger.Debug("Deleted")
```
staebler, line 249: Assuming that we've already set the "id" logger field (which we want to do in `cleanSharedRoute53`,
```suggestion
				logger.WithField(
					"recordset",
					fmt.Sprintf("%s (%s)", aws.StringValue(recordSet.Name), aws.StringValue(recordSet.Type)),
				).Debug("Deleted")
```

pkg/destroy/aws/shared.go
```diff
@@ -218,21 +217,22 @@ func (o *ClusterUninstaller) cleanSharedHostedZone(ctx context.Context, client *
...
+							publicZoneLogger.Debug(lastError)
 						}
 						lastError = errors.Wrapf(err, "deleting record set matching %#v from public zone %s", recordSet, publicZoneID)
 						// do not delete the record set in the private zone if the delete failed in the public zone;
 						// otherwise the record set in the public zone will get leaked
 						continue
 					}
+					recordSetLogger := logger.WithField("id", publicZoneID).WithField(
+						"recordset",
+						fmt.Sprintf("%s (%s)", aws.StringValue(recordSet.Name), aws.StringValue(recordSet.Type)),
```
staebler, line 234: Extract this as a variable that can be shared by the two places were the "recordset" logger field is set, rather than duplicating this code. This will make the code a bit easier to read by being able to put the entire log event on a single line.
```
publicZoneLogger.WithField("recordset", recordsetValue).Debug("Deleted from public zone")
```

Or even use a `Fields` variable.
```
recordsetFields := logrus.Fields{"recordset": fmt.Sprintf("%s (%s)", aws.StringValue(recordSet.Name), aws.StringValue(recordSet.Type))}

publicZoneLogger.WithFields(recordsetFields).Debug("Deleted from public zone")
```

staebler (commented):
/lgtm
/approve

PR #5495, opened 2021-12-16
Bug 2020480:  [vsphere] installation failure sometimes with error: the object 'vim.Folder:group-v******' has already been deleted or has not been completely created

ayesha54:
Updating the version of govmomi to fix the bugs we are experiencing in terms of failing tests for the CI jobs.

Attaching some useful links below:

https://github.com/openshift/release/blob/master/ci-operator/step-registry/upi/conf/vsphere/ova/upi-conf-vsphere-ova-commands.sh#L38-L44

https://github.com/openshift/installer/blob/master/images/installer/Dockerfile.upi.ci#L59-L61

https://github.com/vmware/govmomi/pull/1916/files

### DIFF
--- a/images/installer/Dockerfile.upi.ci
+++ b/images/installer/Dockerfile.upi.ci
@@ -56,9 +56,7 @@ ENV IGNITION_PROVIDER_VERSION=v2.1.0
 RUN curl -L -O https://github.com/community-terraform-providers/terraform-provider-ignition/releases/download/${IGNITION_PROVIDER_VERSION}/terraform-provider-ignition-${IGNITION_PROVIDER_VERSION}-linux-amd64.tar.gz && \
     tar xzf terraform-provider-ignition-${IGNITION_PROVIDER_VERSION}-linux-amd64.tar.gz && \
     mv terraform-provider-ignition-${IGNITION_PROVIDER_VERSION}-linux-amd64/terraform-provider-ignition /bin/terraform-provider-ignition
-RUN curl -L -O https://github.com/vmware/govmomi/releases/download/v0.20.0/govc_linux_amd64.gz && \
-    gzip -d govc_linux_amd64.gz && \
-    chmod +x govc_linux_amd64 && mv govc_linux_amd64 /bin/govc
+RUN curl -L -o - "https://github.com/vmware/govmomi/releases/latest/download/govc_$(uname -s)_$(uname -m).tar.gz" | tar -C /bin -xvzf - govc
 RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
     unzip awscliv2.zip && \
     ./aws/install -b /bin && \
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -106,276 +106,16 @@ require (
 )
 
 require (
-	cloud.google.com/go/bigtable v1.5.0 // indirect
```
kirankt, line 109: Removing these modules does not look right. Please re-run verify vendor.

kirankt (changes requested):

ayesha54:
@jcpowermac

staebler (commented):
Changes seem fine to me.
Need to re-run `go mod vendor`.
/approve

jcpowermac (commented):
I don't see a change to the dockerfile. The govc binary in the container is very out of date and was our guess to the cause of the problem.

ayesha54:
> Changes seem fine to me. Need to re-run `go mod vendor`. /approve

I have ran the go mod vendor but didn't pushed the untracked files.

staebler:
> I don't see a change to the dockerfile. The govc binary in the container is very out of date and was our guess to the cause of the problem.

Can you elaborate on this? Do you think that it is *both* the govmomi module and the govc binary that need to be updated? Or do you think that it is *only* the govc binary that needs to be updated? Can you provide more details about the root cause?

jcpowermac:
This is the change I was expecting.
/lgtm

ayesha54:
> > I don't see a change to the dockerfile. The govc binary in the container is very out of date and was our guess to the cause of the problem.
> 
> Can you elaborate on this? Do you think that it is _both_ the govmomi module and the govc binary that need to be updated? Or do you think that it is _only_ the govc binary that needs to be updated? Can you provide more details about the root cause?

@staebler your comment is definitely helpful, so for now I am upgrading the govc at the container level, as we are guessing that, since its a very old version of govc that can be the cause of the failing jobs with that error but we really don't know, once the PR is merge, I can see the CI jobs results.

staebler:
I am fine with putting this in as an initial step. But the BZ report was that this occurred for manual installations as well, which presumably were not using the UPI CI image.

jcpowermac:
> I am fine with putting this in as an initial step. But the BZ report was that this occurred for manual installations as well, which presumably were not using the UPI CI image.

Right, this change is more to validate a guess that its related to a bug with govmomi (and govc). If this resolve the issue we were seeing with UPI and the OVA import we can move on to the terraform provider.

https://github.com/vmware/govmomi/issues/1908

jcpowermac:
@ayesha54 govc is now in a tar.gz

https://github.com/vmware/govmomi/releases

jcpowermac:
@ayesha54 something isn't quite right
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/5495/pull-ci-openshift-installer-master-e2e-vsphere-upi/1479096065838813184#1:build-log.txt%3A55

jcpowermac:
Reviewed vsphere-upi installation was fine, test failure unrelated

/lgtm

PR #5496, opened 2021-12-17
stop considering Mint mode as supported on Azure

joelddiaz:
xref: https://issues.redhat.com/browse/CORS-1866

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -607,7 +607,7 @@ func validateCloudCredentialsMode(mode types.CredentialsMode, fldPath *field.Pat
 	}
 	allErrs := field.ErrorList{}
 
-	allowedAzureModes := []types.CredentialsMode{types.MintCredentialsMode, types.PassthroughCredentialsMode, types.ManualCredentialsMode}
+	allowedAzureModes := []types.CredentialsMode{types.PassthroughCredentialsMode, types.ManualCredentialsMode}
 	if platform.Azure != nil && platform.Azure.CloudName == azure.StackCloud {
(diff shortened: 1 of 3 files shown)


pkg/types/validation/installconfig.go
```diff
@@ -607,7 +607,7 @@ func validateCloudCredentialsMode(mode types.CredentialsMode, fldPath *field.Pat
 	}
 	allErrs := field.ErrorList{}
 
-	allowedAzureModes := []types.CredentialsMode{types.MintCredentialsMode, types.PassthroughCredentialsMode, types.ManualCredentialsMode}
+	allowedAzureModes := []types.CredentialsMode{types.PassthroughCredentialsMode, types.ManualCredentialsMode}
```
patrickdillon, line 610: This seems reasonable to me. #5325 is a start to clean up some loose ends in how the installer handles credential modes (I think we want to move toward choosing defaults rather than allowing credentialmode to be `""`). #5325 would need to handle this case as well, where we have two modes.

staebler: The issue with choosing a default instead of an empty string is that they mean different things. When an explicit selection is made, it (1) tells the cloud-credential-operator to use that mode and (2) skips permissions validation. When an empty string is used, it tells the cloud-credential-operator to inspect the credentials provided and make an opinionated decision about which mode to use based upon what is allowed with the credentials.

staebler:
Changes need to be made to https://github.com/openshift/installer/blob/f6ea846f7a8a2357191dd2e2c4cec5b73023d0f0/pkg/types/installconfig.go#L146 as well.

/lgtm cancel

patrickdillon:
/close
in favor of #5699

PR #5498, opened 2021-12-17
vendor: update openshift/api to include some alibaba infra changes

staebler:
* Update openshift/api to 6e0b1eb97188.
* Update kube modules to v0.23.0.
* Update controller-runtime to v0.11.0.
* Remove unused terraform-provider-kubernetes.
* Updated the hack/verify-codegen.sh script to use an image containing go 1.17, rather than go 1.16.


kwoodson:
@staebler I was able to install a cluster without this PR. I'm not saying we shouldn't take it but I'm not sure it is required at this time.

staebler:
> @staebler I was able to install a cluster without this PR. I'm not saying we shouldn't take it but I'm not sure it is required at this time.

Since this is not required to fix the BZ, I am removing the link to the BZ and closing the PR.
/retitle vendor: update openshift/api to include some alibaba infra changes
/close

PR #5501, opened 2021-12-20
Bug 2034245: terraform: use the correct os for terraform providers

staebler:
The symlinks created for the terraform providers are hard-coded to
use "linux" as the OS for the provider. Instead, it should be using
the OS from the runtime.


p0lyn0mial:
the fix seems to be real. I have just tested it on my local machine. Thanks!

PR #5505, opened 2021-12-22
Bug 2025788: Check vCPUsAvailable for given instance type

rna-afk:
The installer must also check the vCPUsAvailable for the given
instance type to see if the value is greater than the minimum
required to create a cluster. Adding an extra check for this
requirement.

### DIFF
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -68,13 +68,13 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 
 	for _, capability := range *typeMeta.Capabilities {
 
-		if strings.EqualFold(*capability.Name, "vCPUs") {
+		if strings.EqualFold(*capability.Name, "vCPUsAvailable") {
 			cpus, err := strconv.ParseFloat(*capability.Value, 0)
 			if err != nil {
 				return append(allErrs, field.InternalError(fieldPath, err))
 			}
 			if cpus < float64(req.minimumVCpus) {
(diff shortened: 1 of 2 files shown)


pkg/asset/installconfig/azure/validation.go
```diff
@@ -75,7 +75,16 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 			}
 			if cpus < float64(req.minimumVCpus) {
 				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d vCPUs", req.minimumVCpus)
-				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type", "vCPUs"), instanceType, errMsg))
```
staebler, line 78: The field path should not be changed. This should point to a field in the install-config.yaml.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -75,7 +75,16 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
...
+			}
+		} else if strings.EqualFold(*capability.Name, "vCPUsAvailable") {
```
staebler, line 80: We should be checking `vCPUsAvailable` **instead** of `vCPUs` rather than in addition to.

staebler: See https://docs.microsoft.com/en-us/azure/virtual-machines/constrained-vcpu.
Essentially, the number of available vCPUs will always be less than or equal to the number of vCPUs for an instance type.

staebler:
The e2e-azure job passed.
/hold cancel

PR #5506, opened 2021-12-22
Bug 2028695: Delete bootstrap profiles during cluster destroy

rna-afk:
The installer has been ignoring the bootstrap profiles during
cluster destroy and needs to be cleaned up along with the
cluster. Added bootstrap-profiles to the list of profiles to
check during destroy.


staebler (approved):
/lgtm
/approve

PR #5509, opened 2022-01-04
Bug 2025868: Check HyperVGenerations for instance type

rna-afk:
If an instance type that does not support HyperVGeneration version 1
then terraform returns an error mentioning there's support only for
V1. Adding a check during install config to check for the versions
supported by the instance type provided.

### DIFF
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -13,6 +13,7 @@ import (
 	aznetwork "github.com/Azure/azure-sdk-for-go/profiles/2018-03-01/network/mgmt/network"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"github.com/openshift/installer/pkg/types"
@@ -76,6 +77,16 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 			if cpus < float64(req.minimumVCpus) {
 				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d vCPUsAvailable", req.minimumVCpus)
 				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+			}
+		} else if strings.EqualFold(*capability.Name, "HyperVGenerations") {
+			generations := sets.NewString()
+			for _, g := range strings.Split(to.String(capability.Value), ",") {
+				g = strings.TrimSpace(g)
+				g = strings.ToUpper(g)
+				generations.Insert(g)
+			}
+			if !generations.Has("V1") {
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "only disks with HyperVGeneration V1 are supported"))
 			}
 		} else if strings.EqualFold(*capability.Name, "MemoryGB") {
 			memory, err := strconv.ParseFloat(*capability.Value, 0)
--- a/pkg/asset/installconfig/azure/validation_test.go
+++ b/pkg/asset/installconfig/azure/validation_test.go
@@ -37,13 +37,14 @@ var (
 	validResourceSkuRegions        = "southeastasia"
 
 	instanceTypeSku = []*azsku.ResourceSku{
-		{Name: to.StringPtr("Standard_D4s_v3"), Capabilities: &[]azsku.ResourceSkuCapabilities{{Name: to.StringPtr("vCPUsAvailable"), Value: to.StringPtr("4")}, {Name: to.StringPtr("MemoryGB"), Value: to.StringPtr("16")}, {Name: to.StringPtr("PremiumIO"), Value: to.StringPtr("True")}}},
-		{Name: to.StringPtr("Standard_A1_v2"), Capabilities: &[]azsku.ResourceSkuCapabilities{{Name: to.StringPtr("vCPUsAvailable"), Value: to.StringPtr("1")}, {Name: to.StringPtr("MemoryGB"), Value: to.StringPtr("2")}, {Name: to.StringPtr("PremiumIO"), Value: to.StringPtr("True")}}},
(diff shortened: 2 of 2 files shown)


pkg/asset/installconfig/azure/validation.go
```diff
@@ -77,6 +77,11 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d vCPUsAvailable", req.minimumVCpus)
 				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
 			}
+		} else if strings.EqualFold(*capability.Name, "HyperVGenerations") {
+			generation := to.String(capability.Value)
+			if !strings.Contains(generation, "V1") {
```
staebler, line 82: This probably won't ever bite us, but this is fragile to generations greater than 9.
```suggestion
			if !sets.NewString(strings.Split(generation, ",")...).Has("V1")
```
https://go.dev/play/p/xkPDVfR9kK0

For completeness, it is probably worthwhile to convert `generation` to upper-case and trim whitespace around the generations, too.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -77,6 +78,11 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d vCPUsAvailable", req.minimumVCpus)
 				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
 			}
+		} else if strings.EqualFold(*capability.Name, "HyperVGenerations") {
+			generation := strings.ReplaceAll(strings.ToUpper(to.String(capability.Value)), " ", "")
+			if !sets.NewString(strings.Split(generation, ",")...).Has("V1") {
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "only disks with HyperVGeneration V1 is supported"))
```
patrickdillon, line 84:
```suggestion
				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "only disks with HyperVGeneration V1 are supported"))
```

patrickdillon:
I believe (am trying to get images in order to test) that Gen2 VMs may work with compute nodes when we use images from the Azure Marketplace. 

Therefore, I think we should only run this validation against control plane nodes. Or perhaps also against compute nodes when a marketplace image is not being used.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -77,6 +78,11 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d vCPUsAvailable", req.minimumVCpus)
 				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
 			}
+		} else if strings.EqualFold(*capability.Name, "HyperVGenerations") {
+			generation := strings.ReplaceAll(strings.ToUpper(to.String(capability.Value)), " ", "")
+			if !sets.NewString(strings.Split(generation, ",")...).Has("V1") {
```
staebler, line 83: Let's use `strings.TrimSpace` instead of `strings.ReplaceAll` so that it is clearer on read what the intention is.
```suggestion
			generations := sets.NewString()
			for _, g := range strings.Split(to.String(capability.Value), ",") {
				g = strings.TrimSpace(g)
				g = strings.ToUpper(g)
				generations.Insert(g)
			}
			if !generations.Has("V1") {
```

staebler:
> I believe (am trying to get images in order to test) that Gen2 VMs may work with compute nodes when we use images from the Azure Marketplace.
> 
> Therefore, I think we should only run this validation against control plane nodes. Or perhaps also against compute nodes when a marketplace image is not being used.

The install config does not support using marketplace images for the compute nodes yet, though, right? It still requires the user manually adjusting the machineset manifests.

patrickdillon:
> > I believe (am trying to get images in order to test) that Gen2 VMs may work with compute nodes when we use images from the Azure Marketplace.
> > Therefore, I think we should only run this validation against control plane nodes. Or perhaps also against compute nodes when a marketplace image is not being used.
> 
> The install config does not support using marketplace images for the compute nodes yet, though, right? It still requires the user manually adjusting the machineset manifests.

Ah, yes, that is correct. So you can ignore my comment and once we add support in the install config we can skip this check (assuming the images do support Gen2).

patrickdillon:
On closer inspection, I believe what is happening is that we are [creating the cluster image](https://github.com/openshift/installer/blob/master/data/data/azure/vnet/main.tf#L94) with the default Gen1 value (see https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/image#hyper_v_generation). This is going to create all VMs created from the image as Gen1.

I believe this image is used throughout the lifespace of the cluster. I think it would be worth investigating whether there are any gen1-only instance types we want to support. Otherwise, I think we could just set the image to have generation = 2. I will try a test install with that configuration.

patrickdillon:
By changing the `hyper_v_generation` on the image I was able to create a cluster of gen2 VMs. Seems like there are a few different ways we could take this.

rna-afk:
Interesting so instead of this, we just pass the generation value to terraform?

patrickdillon:
> Interesting so instead of this, can we just pass the generation value to terraform?

I have been looking into this more and I think the best course of action is to add support for both Gen1 & Gen2 VMs in our next version. 

I was hoping we could either just choose all Gen1 or Gen2, but I think the Instance Type landscape is too diverse to do that. 

So I think what we should do as a new feature, is have the installer create two images, a gen1 and gen2 image. We would use code similar to what you have in this PR, but when an instance type has only gen1, we would set the gen1 image in the machineset and pass the values through to terraform for the masters. Otherwise we default to the gen2 image.

I'm not sure if code changes would be needed for the machine-api operator. I don't think they would be needed, but it would be necessary if a new machineset were created that a compatible image and instance type were specified. This would probably just be a matter of documentation.

So until we add this feature I am happy with this approach.

PR #5510, opened 2022-01-05
Bug 2017874: Check quota for G and VT instances

rna-afk:
The AWS ec2 instances with instance types of G and VT were being
treated as type X instance and are not treated as their own
individual instances. Adding G and VT as valid instance types
in the quota check for AWS.

### DIFF
--- a/pkg/asset/quota/aws/aws.go
+++ b/pkg/asset/quota/aws/aws.go
@@ -1,9 +1,13 @@
 package aws
 
 import (
+	"fmt"
+	"regexp"
 	"sort"
+	"strings"
 
 	machineapi "github.com/openshift/api/machine/v1beta1"
+	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/sets"
 	awsprovider "sigs.k8s.io/cluster-api-provider-aws/pkg/apis/awsprovider/v1beta1"
 
@@ -130,14 +134,26 @@ func others() []quota.Constraint {
 
 func machineTypeToQuota(t string, instanceTypes map[string]InstanceTypeInfo) quota.Constraint {
 	info, ok := instanceTypes[t]
+	warnMessage := fmt.Sprintf("The instance class is unknown for the instance type %q. The vCPU quota check will be skipped.", t)
 	if !ok {
+		logrus.Warnf(warnMessage)
 		return quota.Constraint{Name: "ec2/L-7295265B", Count: 0}
 	}
-	class := string(t[0])
-	switch class {
-	case "a", "c", "d", "h", "i", "m", "r", "t", "z":
+	r := regexp.MustCompile(`^([A-Za-z]+)[0-9]`)
+	match := r.FindStringSubmatch(strings.ToLower(t))
+	if match == nil {
+		logrus.Warnf(warnMessage)
+		return quota.Constraint{Name: "ec2/L-7295265B", Count: 0}
+	}
+	switch match[1] {
+	case "a", "c", "d", "h", "i", "is", "im", "m", "r", "t", "z":
 		return quota.Constraint{Name: "ec2/L-1216C47A", Count: info.vCPU}
-	default:
+	case "g", "vt":
+		return quota.Constraint{Name: "ec2/L-DB2E81BA", Count: info.vCPU}
+	case "x":
 		return quota.Constraint{Name: "ec2/L-7295265B", Count: info.vCPU}
+	default:
+		logrus.Warnf(warnMessage)
+		return quota.Constraint{Name: "ec2/L-7295265B", Count: 0}
 	}
 }


rna-afk:
Hope that quota code for G and VT are right. Based it off of the output of this command.
```
aws service-quotas list-service-quotas --service-code ec2
...
"ServiceCode": "ec2",
"ServiceName": "Amazon Elastic Compute Cloud (Amazon EC2)",
"QuotaArn": "arn:aws:servicequotas:us-east-1:269733383066:ec2/L-3819A6DF",
"QuotaCode": "L-3819A6DF",
"QuotaName": "All G and VT Spot Instance Requests",
...
```

pkg/asset/quota/aws/aws.go
```diff
@@ -137,7 +138,12 @@ func machineTypeToQuota(t string, instanceTypes map[string]InstanceTypeInfo) quo
 	switch class {
 	case "a", "c", "d", "h", "i", "m", "r", "t", "z":
 		return quota.Constraint{Name: "ec2/L-1216C47A", Count: info.vCPU}
+	case "g":
+		return quota.Constraint{Name: "ec2/L-3819A6DF", Count: info.vCPU}
 	default:
+		if strings.HasPrefix(strings.ToLower(t), "vt") {
+			return quota.Constraint{Name: "ec2/L-3819A6DF", Count: info.vCPU}
+		}
```
staebler, line 146: What is unique about the VT instance class that we are looking for prefixes rather than the entire string?
I would have expected the following.
```suggestion
	case "vt":
		return quota.Constraint{Name: "ec2/L-3819A6DF", Count: info.vCPU}
	default:
```

rna-afk: Class was set to t[0] so I thought just the first character is being set to it

staebler: Ah, right. That makes the code awkward. I think I can live with what you have, but give me a bit to ruminate on it.

pkg/asset/quota/aws/aws.go
```diff
@@ -137,7 +138,12 @@ func machineTypeToQuota(t string, instanceTypes map[string]InstanceTypeInfo) quo
 	switch class {
 	case "a", "c", "d", "h", "i", "m", "r", "t", "z":
 		return quota.Constraint{Name: "ec2/L-1216C47A", Count: info.vCPU}
+	case "g":
+		return quota.Constraint{Name: "ec2/L-3819A6DF", Count: info.vCPU}
 	default:
+		if strings.HasPrefix(strings.ToLower(t), "vt") {
+			return quota.Constraint{Name: "ec2/L-3819A6DF", Count: info.vCPU}
```
staebler, line 145: This is the wrong code. We want on-demand instances not spot instances.
```
$ aws service-quotas list-service-quotas --service-code ec2 --output json | jq '.Quotas[] | select(.QuotaName | contains("On-Demand G and VT instances")) | .QuotaCode' -r
L-DB2E81BA
```

```suggestion
		return quota.Constraint{Name: "ec2/L-DB2E81BA", Count: info.vCPU}
	default:
		if strings.HasPrefix(strings.ToLower(t), "vt") {
			return quota.Constraint{Name: "ec2/L-DB2E81BA", Count: info.vCPU}
```

pkg/asset/quota/aws/aws.go
```diff
@@ -137,7 +138,12 @@ func machineTypeToQuota(t string, instanceTypes map[string]InstanceTypeInfo) quo
...
+		}
 		return quota.Constraint{Name: "ec2/L-7295265B", Count: info.vCPU}
```
staebler, line 147: Let's add a case for "X" class explicitly rather than defaulting to "X". If the class type is one that the installer does not know, then we should warn the user that we are not checking that quota and return a constraint with a 0 count. That way we are protected from new classes being added in the future that we would otherwise assume are "X".

pkg/asset/quota/aws/aws.go
staebler: The logic of just looking at the first character is flawed to begin with, unfortunately. For example, the DL instance types are incorrectly interpreted as part of the D class instead of the DL class. I was hoping that we could use a regex with the assumption that the instance class is always followed by a number (eg, M4), but there are the Im4gn and Is4gen types that are part of the I instance class. So then I thought maybe a number or a lower-case letter, but there are the Inf1 types that are part of the Inf instance class.

Nevertheless, I propose that we are better off with something akin to the following.
https://go.dev/play/p/OBx-EcfMjVS

pkg/asset/quota/aws/aws.go
```diff
@@ -131,13 +133,21 @@ func others() []quota.Constraint {
 func machineTypeToQuota(t string, instanceTypes map[string]InstanceTypeInfo) quota.Constraint {
 	info, ok := instanceTypes[t]
 	if !ok {
-		return quota.Constraint{Name: "ec2/L-7295265B", Count: 0}
+		return quota.Constraint{Name: "unknown", Count: 0}
```
staebler, line 136: I think that this needs to stay as a valid quota label. The installer is going to check this quota but check for enough quota for 0 cpus. That is unless there is conditional somewhere to skip the quota check when the count is 0.

pkg/asset/quota/aws/aws.go
```diff
@@ -131,13 +133,21 @@ func others() []quota.Constraint {
...
-	class := string(t[0])
-	switch class {
-	case "a", "c", "d", "h", "i", "m", "r", "t", "z":
+	switch match[1] {
+	case "a", "c", "d", "h", "i", "is", "im", "m", "r", "t", "z":
 		return quota.Constraint{Name: "ec2/L-1216C47A", Count: info.vCPU}
+	case "g", "vt":
+		return quota.Constraint{Name: "ec2/L-DB2E81BA", Count: info.vCPU}
+	case "x":
+		return quota.Constraint{Name: "ec2/L-7295265B", Count: 0}
```
staebler, line 149:
```suggestion
		return quota.Constraint{Name: "ec2/L-7295265B", Count: info.vCPU}
```

pkg/asset/quota/aws/aws.go
```diff
@@ -133,11 +135,19 @@ func machineTypeToQuota(t string, instanceTypes map[string]InstanceTypeInfo) quo
 	if !ok {
 		return quota.Constraint{Name: "ec2/L-7295265B", Count: 0}
```
staebler, line 136: Can we add a warning to the places where we are returning a zero-count quota so that the user is informed that we are not actually checking the quota? That will help us know in the future when users start trying to use new instance classes.

pkg/asset/quota/aws/aws.go
```diff
@@ -130,14 +133,26 @@ func others() []quota.Constraint {
 
 func machineTypeToQuota(t string, instanceTypes map[string]InstanceTypeInfo) quota.Constraint {
 	info, ok := instanceTypes[t]
+	warnMessage := "unrecognized instance type specified, using a quota with 0 count"
```
staebler, line 136:
```suggestion
	warnMessage := "unrecognized instance type specified, skipping vCPU quota check"
```

pkg/asset/quota/aws/aws.go
```diff
@@ -130,14 +133,26 @@ func others() []quota.Constraint {
 
 func machineTypeToQuota(t string, instanceTypes map[string]InstanceTypeInfo) quota.Constraint {
 	info, ok := instanceTypes[t]
+	warnMessage := "unrecognized instance type specified, skipping vCPU quota check"
```
staebler, line 136: Sorry that I missed this in my last comment. The warning message should use sentences and start with an upper-case letter. Let's also take the opportunity to make it a bit more user-friendly.
```suggestion
	warnMessage := fmt.Sprintf("The instance class is unknown for the instance type %q. The vCPU quota check will be skipped.", t)
```

staebler (commented):
/lgtm
/approve

staebler:
> /hold for e2e-azure tests /test e2e-azure

Ugh, I'm getting things mixed up. These changes do not affect Azure.
/hold cancel

Issue #5504, opened 2021-12-21
IPI fails on iRMC servers due to inspect timeout


hs0210:
# Version

```console
$ openshift-baremetal-install version
openshift-baremetal-install 4.10.0-0.nightly-2021-12-20-231053
built from commit 9f37ece3620d14e48507f2afc5cf6a667ca2cef0
release image quay.io/openshift-release-dev/ocp-release-nightly@sha256:26f811bd37c593564093aa8e323cf81637c49a9527dbe6772c885fa9f55ab684
release architecture amd64
```

# Platform:

baremetal

IPI

# What happened?

### Initial error phenomenon

Bootstrap VM does not start the ironic related containers.

### What we tried

We manually pull `image-customization-controller` container, and after that the ironic related containers can successfully start.

### Next error phenomenon

The master nodes got stuck in a restart loop and eventually the IPI failed due to inspect timeout.

The following is the error log from .openshift_install.log
```
time="2021-12-21T07:30:59Z" level=error msg="Error: could not inspect: could not inspect node, node is currently 'inspect failed' , last error was 'timeout reached while inspecting the node'"
time="2021-12-21T07:30:59Z" level=error
time="2021-12-21T07:30:59Z" level=error msg="  on ../../tmp/openshift-install-masters-4047236787/main.tf line 13, in resource \"ironic_node_v1\" \"openshift-master-host\":"
time="2021-12-21T07:30:59Z" level=error msg="  13: resource \"ironic_node_v1\" \"openshift-master-host\" {"
```

The following is the error log from ServerView
![errorFormSeverView](https://user-images.githubusercontent.com/53652350/146900533-9779ade6-5f76-4f00-94db-fe99cdd4ca77.png)

# What you expected to happen?

IPI successfully run on iRMC.

# How to reproduce it (as minimally and precisely as possible)?

1. Download the latest OCP nightly
2. Fill in install-config
3. Execute `create cluster`
4. Pull `image-customization-controller` after bootstrap started.

```console
$ mkdir ~/clusterconfigs
$ cp ~/install-config.yaml ~/clusterconfigs/
$ openshift-baremetal-install --dir ~/clusterconfigs create manifests
$ openshift-baremetal-install --dir ~/clusterconfigs --log-level debug create cluster
# wait for bootstrap starts, then execute the following command on Bootstrap VM
[...]

PR #5513, opened 2022-01-06
Fix image-customization image pull failure

hs0210:
*startironic.sh* itself is executed under the root user, so there is no need to use `sudo`. And using `sudo` leads to the loss of proxy env, which causes the failure of image-customization pull.

Before #5473 `sudo` has been used to run Ironic related containers, but no error was reported because the ironic image had been [downloaded to the local](https://github.com/openshift/installer/blob/master/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template#L53) beofre. But the image-customization container newly added by #5473 did not pull the image locally before, so it triggered this long-standing error.

This PR fix the first error in #5504 

Signed-off-by: Hu Shuai <hus.fnst@fujitsu.com>

### DIFF
--- a/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
+++ b/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
@@ -173,7 +173,7 @@ else
 fi
 
 # Embed agent ignition into the rhcos live iso
-sudo podman run -d --net host --privileged --name image-customization \
+podman run -d --net host --privileged --name image-customization \
     --env DEPLOY_ISO="/shared/html/images/ironic-python-agent.iso" \
     --env DEPLOY_INITRD="/shared/html/images/ironic-python-agent.initramfs" \
     --env IRONIC_BASE_URL="http://${IRONIC_HOST}" \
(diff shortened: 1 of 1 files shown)


zaneb:
We'll need to open a bugzilla for this.

/label platform/baremetal
/approve

zaneb:
We're also using sudo for some iptables commands, it wouldn't hurt to get rid of those as well so that we're less likely to repeat the error.

hs0210:
> We're also using sudo for some iptables commands, it wouldn't hurt to get rid of those as well so that we're less likely to repeat the error.

Done. I removed sudo from iptables commands.

hs0210:
> We'll need to open a bugzilla for this.

Understood, we will do.

hs0210:
> We'll need to open a bugzilla for this.

@zaneb We have sent the [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2039109), PTAL.

dtantsur:
/retitle Bug 2039109: Fix image-customization-controller image pull failure

dtantsur:
@hs0210 as you can see from the bot, private bugzillas cannot be attached to pull requests. Could you please open it or file a public one with limited information? In the latter case, please update the title again.

PR #29, opened 2018-07-06
*: Remove remaining tectonic-torcx references

wking:
These should have been removed by 45fa0e49 (coreos/tectonic-installer#3275).

PR #5514, opened 2022-01-06
Bug 2034527: Base IPA kernel params on provisioning network IP version

derekhiggins:
We need the kernel param for dhcp version to be based on the
provisioning network version for IPA and the external network
version for the machine image. These can differ.

### DIFF
--- a/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
+++ b/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
@@ -98,14 +98,22 @@ IPTABLES=iptables
 {{ end }}
 
 {{ if .UseIPv6ForNodeIP }}
-IP_OPTIONS="ip=dhcp6"
+EXTERNAL_IP_OPTIONS="ip=dhcp6"
 {{ else }}
-IP_OPTIONS="ip=dhcp"
+EXTERNAL_IP_OPTIONS="ip=dhcp"
 {{ end }}
+
+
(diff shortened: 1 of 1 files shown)


data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -188,6 +188,10 @@ sudo podman run -d --net host --privileged --name image-customization \
     -v /etc/containers:/tmp/containers:z \
     ${CUSTOMIZATION_IMAGE}
 
+{{ if .PlatformData.BareMetal.ProvisioningIPv6 }}
+IP_OPTIONS="ip=dhcp6"
```
zaneb, line 192: Does this do the right thing if the Provisioning network is IPv4 and the external network is IPv6? Or can this not happen?

If they're separate I think we should rename the two variables to PROVISIONING_IP_OPTIONS and EXTERNAL_IP_OPTIONS to avoid confusion.

zaneb:
Actually, I'm no longer convinced by this. See https://github.com/openshift/ironic-agent-image/pull/29#issuecomment-1006978285

We're not actually setting different options for IPA vs the final image here, we're just setting different options for PXE vs virtualmedia. That's unlikely to be the right thing.

derekhiggins:
Closing in favour of [#29 (comment)
](https://github.com/openshift/ironic-agent-image/pull/29#issuecomment-1007882330)

PR #5515, opened 2022-01-06
Bug 2026356: Fix bootstrap disk instance type

rna-afk:
The bootstrap VM type is set to the default value for the given
region but must be instead set to the value given for the master
VM type. Making the changes to use the master VM type.

### DIFF
--- a/data/data/azure/bootstrap/main.tf
+++ b/data/data/azure/bootstrap/main.tf
@@ -188,7 +188,7 @@ resource "azurerm_linux_virtual_machine" "bootstrap" {
   location              = var.azure_region
   resource_group_name   = var.resource_group_name
   network_interface_ids = [azurerm_network_interface.bootstrap.id]
-  size                  = var.azure_bootstrap_vm_type
+  size                  = var.azure_master_vm_type
   admin_username        = "core"
   # The password is normally applied by WALA (the Azure agent), but this
   # isn't installed in RHCOS. As a result, this password is never set. It is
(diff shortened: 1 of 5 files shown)


staebler (commented):
The focus of the BZ is that the **VM type** of the bootstrap should match the VM type of the masters.

data/data/azure/bootstrap/main.tf
```diff
@@ -188,7 +188,7 @@ resource "azurerm_linux_virtual_machine" "bootstrap" {
   location              = var.azure_region
   resource_group_name   = var.resource_group_name
   network_interface_ids = [azurerm_network_interface.bootstrap.id]
-  size                  = var.azure_bootstrap_vm_type
+  size                  = var.azure_master_vm_type
```
staebler, line 191: We should do one of the following.
1) Change the value set for the `azure_bootstrap_vm_type` variable and use that variable.
1) Or, remove the `azure_bootstrap_vm_type` variable.

If we do (2), then we need to make changes to the azurestack tf files, too.

staebler, line 191: We also need to do something about the type for the os_disk. If the user selects a VM type that does not support premium IO, then the install will fail to create the bootstrap VM. Consequently, we need to use the same disk type for the bootstrap VM as we do for the master VMs.

pkg/tfvars/azure/azure.go
```diff
@@ -84,7 +82,6 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
 		Environment:                     environment,
 		ARMEndpoint:                     sources.ARMEndpoint,
 		Region:                          region,
-		BootstrapInstanceType:           defaults.BootstrapInstanceType(sources.CloudName, region),
```
staebler, line 87: Can the `BootstrapInstanceType` function be removed?

rna-afk: It's not being used anywhere now so removing it :+1:

staebler (commented):
This looks good. But let's wait on this on for master to open back up.
/approve

staebler (commented):
/lgtm

PR #5516, opened 2022-01-06
Bug 2035705: Azure: Only attempt to destroy resourcegroups if present

kirankt:
This is a follow up PR to https://github.com/openshift/installer/pull/5314 which didn't correctly check whether the resourcegroups were already deleted before attempting deletion.

This PR fixes the bug and has been locally tested by attempting to delete a missing resourcegroup.

### DIFF
--- a/pkg/destroy/azure/azure.go
+++ b/pkg/destroy/azure/azure.go
@@ -489,14 +489,21 @@ func isNotFoundError(err error) bool {
 	}
 
 	var dErr autorest.DetailedError
-	if errors.As(err, &dErr) {
-		switch statusCode := dErr.StatusCode.(type) {
-		case int:
-			if statusCode == http.StatusNotFound {
+	errors.As(err, &dErr)
+
+	if dErr.StatusCode == http.StatusNotFound {
+		return true
+	}
+
+	if dErr.StatusCode == 0 {
+		serviceErr, ok := dErr.Original.(*azureenv.ServiceError)
+		if ok {
+			if strings.HasSuffix(serviceErr.Code, "NotFound") {
 				return true
 			}
 		}
 	}
+
 	return false
 }
 


pkg/destroy/azure/azure.go
```diff
@@ -469,7 +469,10 @@ func deleteResourceGroup(ctx context.Context, client resources.GroupsClient, log
 		err = delFuture.WaitForCompletionRef(ctx, client.Client)
 	}
 	if err != nil {
-		if isNotFoundError(err) {
+		// FIXME. This is a workaround while using Azure SDK
+		// versions v5x and older which don't seem to pass
+		// the correct error code. They send a 0 instead of a 404.
+		if strings.Contains(err.Error(), "ResourceGroupNotFound") {
```
staebler, line 475: I think we should go ahead and check `isNotFoundError`, too. That will cover us when we inevitably update the azure sdk and forget to change the logic here.

This is the code for `isNotFoundError` that worked for me to cover both SDK versions.
```
func isNotFoundError(err error) bool {
	if err == nil {
		return false
	}

	var dErr autorest.DetailedError
	errors.As(err, &dErr)

	if dErr.StatusCode == 404 {
		return true
	}

	if dErr.StatusCode == 0 {
		serviceErr, ok := dErr.Original.(*azure2.ServiceError)
		if ok {
			if strings.HasSuffix(serviceErr.Code, "NotFound") {
				return true
			}
		}
	}

	return false
}
```

pkg/destroy/azure/azure.go
```diff
@@ -489,14 +489,21 @@ func isNotFoundError(err error) bool {
 	}
 
 	var dErr autorest.DetailedError
-	if errors.As(err, &dErr) {
-		switch statusCode := dErr.StatusCode.(type) {
-		case int:
-			if statusCode == http.StatusNotFound {
+	errors.As(err, &dErr)
+
+	if dErr.StatusCode == 404 {
```
staebler, line 494:
```suggestion
	if dErr.StatusCode == http.StatusNotFound {
```

staebler (commented):
Looks good. Can you squash the commits?

staebler (commented):
/lgtm
/approve
/hold for e2e-azure tests
/test e2e-azure

staebler:
e2e-azure job passed.
/hold cancel

PR #5521, opened 2022-01-08
Bug 2034527: Pass different IP options to installed CoreOS image and IPA

zaneb:
The installed CoreOS that is written to disk needs IP options set such that it can contact the MCO over the external network to obtain the Ignition file for the Machine.
These options may be different from the ones needed for pre-provisioning, when IPA has to contact ironic, usually over the provisioning network.

Pass the option to the image-customization-controller, so it can build it into the IPA ignition (in openshift/image-customization-controller#31)


zaneb:
The dualstack job won't pass until all 4 patches for this bug are merged (and the last one needs to be the ironic-agent one, otherwise we'd likely break the IPv6 jobs too), so I think this is ready to go.
/assign @andfasano

PR #5529, opened 2022-01-11
Bug 2039689: Alibaba: fix payment method for NAT gateway

bd233:
Pay-by-specification NAT is no longer supported. Newly purchased pay-as-you-go NAT gateways only support the pay-by-CU metering method.

The wrong information is as follows:

image


bd233:
@elmiko I submitted a PR here, maybe you need to create a bugzilla and change its title

elmiko:
@bd233 thanks for the ping, we will need to make a bugzilla for this PR but we should get @staebler and @patrickdillon involved too as they are the installer contacts.

PR #5532, opened 2022-01-12
Bug 2026321: Remove allowed values for nodeVMsize

rna-afk:
Providing allowedValues for the node VMsize for an Azure UPI installation
seems to be an overhead for the installer where we have to
constantly update the list of values that can be used every
time azure changes or creates a new type.

Removing the allowed value list to prevent this overhead.


staebler (commented):
/lgtm
/approve

PR #5533, opened 2022-01-13
Bug 2033524: [Alibaba] support for list existing domains

bd233:
Existing base domains should be automatically listed.


staebler (approved):
/lgtm
/approve

PR #5540, opened 2022-01-17
Bug 2041361: [Alibaba] fix load balance listener strategy

bd233:
Disable session persistence and removebBandwidth peak of listener

### DIFF
--- a/data/data/alibabacloud/cluster/vpc/slb.tf
+++ b/data/data/alibabacloud/cluster/vpc/slb.tf
@@ -18,9 +18,8 @@ resource "alicloud_slb_listener" "listener_external_6443" {
   backend_port              = 6443
   frontend_port             = 6443
   protocol                  = "tcp"
-  bandwidth                 = 10
-  sticky_session            = "on"
-  sticky_session_type       = "insert"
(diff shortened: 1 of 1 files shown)


data/data/alibabacloud/cluster/vpc/slb.tf
```diff
@@ -18,9 +18,6 @@ resource "alicloud_slb_listener" "listener_external_6443" {
   backend_port              = 6443
   frontend_port             = 6443
   protocol                  = "tcp"
-  bandwidth                 = 10
-  sticky_session            = "on"
```
mtulio, line 22: @bd233 if we don't want session enabled and it can cause issues when enabled, I believe we should explicitly disable it, even if the default is off.

staebler (commented):
/approve

data/data/alibabacloud/cluster/vpc/slb.tf
bd233: Thanks. Have  updated.

staebler (commented):
Are there similar changes that should be made to the ingress load balancer creating in-cluster?

/lgtm

PR #5541, opened 2022-01-17
Bug 2041765: Adjust the startup order of httpd container

zhouhao3:
Run the httpd container after the coreos-downloader completes to ensure that the kernel parameters can be added correctly.

Fixes: #5504 

Signed-off-by: Zhou Hao <zhouhao@fujitsu.com>

### DIFF
--- a/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
+++ b/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
@@ -85,12 +85,6 @@ podman run -d --net host --privileged --name mariadb \
      -v $IRONIC_SHARED_VOLUME:/shared:z --entrypoint /bin/runmariadb \
      --env MARIADB_PASSWORD=$mariadb_password ${IRONIC_IMAGE}
 
-podman run -d --net host --privileged --name httpd \
-     --restart on-failure \
-     --env IRONIC_RAMDISK_SSH_KEY="$IRONIC_RAMDISK_SSH_KEY" \
-     --env PROVISIONING_INTERFACE=$PROVISIONING_NIC \
-     -v $IRONIC_SHARED_VOLUME:/shared:z --entrypoint /bin/runhttpd ${IRONIC_IMAGE}
-
 {{ if .PlatformData.BareMetal.ProvisioningIPv6 }}
 IPTABLES=ip6tables
 {{ else }}
@@ -116,6 +110,15 @@ podman run -d --name coreos-downloader \
      --env IP_OPTIONS=${PROVISIONING_IP_OPTIONS} \
      -v $IRONIC_SHARED_VOLUME:/shared:z \
      ${MACHINE_OS_IMAGES_IMAGE} /bin/copy-metal --all /shared/html/images/
+
+# Wait for images to be downloaded/ready
+podman wait -i 1000ms coreos-downloader
+
+podman run -d --net host --privileged --name httpd \
+     --restart on-failure \
+     --env IRONIC_RAMDISK_SSH_KEY="$IRONIC_RAMDISK_SSH_KEY" \
+     --env PROVISIONING_INTERFACE=$PROVISIONING_NIC \
+     -v $IRONIC_SHARED_VOLUME:/shared:z --entrypoint /bin/runhttpd ${IRONIC_IMAGE}
 
 # Add firewall rules to ensure the IPA ramdisk can reach httpd, Ironic and the Inspector API on the host
 for port in 80 5050 6385 ; do
@@ -146,9 +149,6 @@ done
 
 $IPTABLES -t raw -A DHCP_IRONIC -j DROP
 {{end}}
-
-# Wait for images to be downloaded/ready
-podman wait -i 1000ms coreos-downloader
 
 export KUBECONFIG=/opt/openshift/auth/kubeconfig-loopback
 


zaneb:
So the runhttpd script is [rendering](https://github.com/openshift/ironic-image/blob/master/scripts/runhttpd#L33) `inspector.ipxe`, which [contains](https://github.com/openshift/ironic-image/blob/master/ironic-config/inspector.ipxe.j2#L8) a reference to `IRONIC_KERNEL_PARAMS`, which is only set correctly [if the rootfs file exists](https://github.com/openshift/ironic-image/blob/master/scripts/coreos-ipa-common.sh#L15-L21).

What's most confusing about this is how it could have been working anywhere?

We'll need to link this to a (public) bugzilla in order to merge it.

/approve
/label platform/baremetal
/test e2e-metal-ipi

zhouhao3:
> What's most confusing about this is how it could have been working anywhere?

We are also very confused, what environment do you test in so that this issue will not occur? Maybe it's just an iPXE issue.

zhouhao3:
/retitle Bug 2041765: Adjust the startup order of httpd container

zhouhao3:
> We'll need to link this to a (public) bugzilla in order to merge it.

Sent bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=2041765

derekhiggins:
> What's most confusing about this is how it could have been working anywhere?

AIUI there are two paths to inspecting a host, requesting ironic-conducor to inspect a specific host will cause ironic to create a pxe config specific for the host, this uses dualboot.ipxe this doesn't not use inspector.ipxe as a config is present for the MAC of the host

The alternative also uses dualboot.ipxe but fails to get a host specific config and then fallsback to inspector.ipxe , this could happen if
 inspection is not being managed by ironic-conductor, perhaps by talking directly to inspector?
or the MAC address for the host is incorrect, causing the config created for the host to fail to be found

zhouhao3:
>The alternative also uses dualboot.ipxe but fails to get a host specific config and then fallsback to inspector.ipxe , this could happen if
inspection is not being managed by ironic-conductor, perhaps by talking directly to inspector?
or the MAC address for the host is incorrect, causing the config created for the host to fail to be found

As we can see from the PXE startup interface, inspector.ipxe is used because there is no corresponding file in the pxelinux.cfg folder.

MicrosoftTeams-image

We can see that the pxelinux.cfg folder is empty in bootstrap.

```
[root@bootstrap html]# pwd
/var/lib/containers/storage/volumes/ironic/_data/html
[root@bootstrap html]# ls
boot.ipxe  dualboot.ipxe  images  inspector.ipxe  pxelinux.cfg  uefi_esp.img
[root@bootstrap html]# ls pxelinux.cfg/
```

There are relevant files in the tftpboot folder, which are produced by the dnsmasq container.

```
[root@bootstrap _data]# pwd
/var/lib/containers/storage/volumes/ironic/_data
[root@bootstrap _data]# ls
html  ironic_prometheus_exporter  log  tftpboot
[root@bootstrap _data]# ls tftpboot/pxelinux.cfg/
01-90-1b-0e-e6-37-dd  01-90-1b-0e-e6-7b-70  01-90-1b-0e-e6-7b-88
[root@bootstrap _data]# pwd
/var/lib/containers/storage/volumes/ironic/_data
[root@bootstrap _data]# ls
html  ironic_prometheus_exporter  log  tftpboot
[root@bootstrap _data]# ls tftpboot/
01e9193d-b574-4daa-922b-96204e9b0ca6            90:1b:0e:e6:37:dd.conf                          grub.cfg-01-90-1b-0e-e6-37-dd  snponly.efi
19c783cc-80f1-56d9-b495-de809b584454.converted  90:1b:0e:e6:7b:70.conf                          grub.cfg-01-90-1b-0e-e6-7b-70  undionly.kpxe
207cca95-4faf-5c21-b7ac-05c88802adec.converted  90:1b:0e:e6:7b:88.conf                          grub.cfg-01-90-1b-0e-e6-7b-88
5dcaec5b-7e86-5ba6-849b-5d554febc5fc.converted  bb03b882-cb00-41ad-bf24-1dd6dd6fc8a9            ipxe.efi
80c6b42d-fc31-42c9-b712-0d968f261488            efab1372-ace5-5041-922e-851381838617.converted  pxelinux.cfg
[root@bootstrap _data]# ls -l tftpboot/pxelinux.cfg/
total 0
lrwxrwxrwx. 1 root root 46 Jan 19 05:47 01-90-1b-0e-e6-37-dd -> ../80c6b42d-fc31-42c9-b712-0d968f261488/config
lrwxrwxrwx. 1 root root 46 Jan 19 05:47 01-90-1b-0e-e6-7b-70 -> ../01e9193d-b574-4daa-922b-96204e9b0ca6/config
lrwxrwxrwx. 1 root root 46 Jan 19 05:47 01-90-1b-0e-e6-7b-88 -> ../bb03b882-cb00-41ad-bf24-1dd6dd6fc8a9/config

```
It looks like the tftp service configured in the dnsmasq container is correct, but it is not being used.

zaneb:
The files in the tftpboot dir have `01-` prepended to the MAC, which happens when `ipxe_enabled` is false, but the dualboot.ipxe script is just looking for the MAC. So it sounds to me like we are somehow using ipxe but with `ipxe_enabled = False` in the irmc driver/config somehow.

We are setting the boot interface to `pxe`, so that is probably wrong: https://github.com/metal3-io/baremetal-operator/blob/main/pkg/hardwareutils/bmc/irmc.go#L77-L79

According to https://docs.openstack.org/ironic/latest/admin/drivers/irmc.html#hardware-interfaces the `pxe` interface is deprecated and we should use `irmc-pxe`.
https://docs.openstack.org/ironic/latest/admin/drivers/irmc.html#boot-from-remote-volume says that both of those use iPXE, but that does not appear to be true - `pxe` and `ipxe` are different boot interfaces, and `IRMCPXEBoot` inherits from `pxe.PXEBoot` (with `ipxe_enabled = False), not `ipxe.iPXEBoot`.

I suspect if we change the boot interface to `ipxe`, this will work.

zhouhao3:
>I suspect if we change the boot interface to ipxe, this will work.

After changing the boot interface to ipxe, there will be no previous issue, thank you very much, I will close this PR, and I will create a fixed PR later.

zhouhao3:
After solving this problem by changing the boot interface to ipxe, there is a issue that the Master cannot download the Ironic Agent image. See #5552 for details.

zaneb:
I think we should still merge this, since it's clearly correct even if it isn't strictly needed in this case.
/reopen

patrickdillon:
alibaba should no longer be required.

/skip

PR #5543, opened 2022-01-17
Bug 2041598: Azure Stack remove CA bundle from CPC

patrickdillon:
Remove the CA bundle from the clooud provider config, because the MCO does not expect it. This leads to a discrepency between the bootstrap generated and cluster generated master configs.

Considering there is a discrepency, we could add support in the MCO for the trust bundle in the Azure cloud provider config. But we have decided that the long-term plan should be a unified approach to internal CAs. Until then, operators on Azure Stack should use the CA from the proxy object.

So there is no need for the CA bundle in the Azure configmap so let's just remove it.


staebler (commented):
/lgtm
/approve

PR #5545, opened 2022-01-18
Bug 2037813: [baremetal] networkConfig field now accepts yaml instead of string value

andfasano:
The current patch allows the user to specify the content of the install-config networkConfig field directly as a yaml object. Content validation (for a generic yaml) is now carried on by the install config asset

/label platform/baremetal

### DIFF
--- a/pkg/asset/machines/baremetal/hosts.go
+++ b/pkg/asset/machines/baremetal/hosts.go
@@ -3,7 +3,9 @@ package baremetal
 import (
 	"fmt"
 
+	"github.com/ghodss/yaml"
 	"github.com/metal3-io/baremetal-operator/pkg/hardware"
+	"github.com/pkg/errors"
 
 	machineapi "github.com/openshift/api/machine/v1beta1"
 	corev1 "k8s.io/api/core/v1"
@@ -29,8 +31,14 @@ type HostSettings struct {
 	NetworkConfigSecrets []corev1.Secret
 }
 
-func createNetworkConfigSecret(host *baremetal.Host) corev1.Secret {
-	return corev1.Secret{
+func createNetworkConfigSecret(host *baremetal.Host) (*corev1.Secret, error) {
+
+	yamlNetworkConfig, err := yaml.Marshal(host.NetworkConfig)
+	if err != nil {
+		return nil, errors.Wrap(err, "Error while creating network config secret")
+	}
+
+	return &corev1.Secret{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "v1",
 			Kind:       "Secret",
@@ -39,8 +47,8 @@ func createNetworkConfigSecret(host *baremetal.Host) corev1.Secret {
 			Name:      fmt.Sprintf("%s-network-config-secret", host.Name),
 			Namespace: "openshift-machine-api",
 		},
-		Data: map[string][]byte{"nmstate": []byte(host.NetworkConfig)},
-	}
+		Data: map[string][]byte{"nmstate": yamlNetworkConfig},
+	}, nil
 }
 
 func createSecret(host *baremetal.Host) (*corev1.Secret, baremetalhost.BMCDetails) {
@@ -120,9 +128,12 @@ func Hosts(config *types.InstallConfig, machines []machineapi.Machine) (*HostSet
 		}
 		newHost := createBaremetalHost(host, bmc)
 
-		if host.NetworkConfig != "" {
-			networkConfigSecret := createNetworkConfigSecret(host)
-			settings.NetworkConfigSecrets = append(settings.NetworkConfigSecrets, networkConfigSecret)
+		if host.NetworkConfig != nil {
+			networkConfigSecret, err := createNetworkConfigSecret(host)
+			if err != nil {
+				return nil, err
+			}
+			settings.NetworkConfigSecrets = append(settings.NetworkConfigSecrets, *networkConfigSecret)
 			newHost.Spec.PreprovisioningNetworkDataName = networkConfigSecret.Name
 		}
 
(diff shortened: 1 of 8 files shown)


pkg/asset/machines/baremetal/hosts.go
```diff
@@ -29,8 +32,27 @@ type HostSettings struct {
...
 
-func createNetworkConfigSecret(host *baremetal.Host) corev1.Secret {
-	return corev1.Secret{
+func convertNetworkConfigToYaml(host *baremetal.Host) ([]byte, error) {
+	var yamlNetworkConfig []byte
+
+	jsonNetworkConfig, err := json.Marshal(host.NetworkConfig)
+	if err == nil {
+		yamlNetworkConfig, err = yaml.JSONToYAML([]byte(jsonNetworkConfig))
+	}
```
staebler, line 41: Why do we need to marshal to JSON and then convert the JSON to YAML as opposed to directly marshaling to YAML?

staebler (commented):
> The current patch allows the user to specify the content of the install-config networkConfig field directly as a yaml object.

Hooray for this!

pkg/asset/machines/baremetal/hosts.go
andfasano: Thanks for the suggestion @staebler even though the implementation was the same I could save few lines of code

zaneb:
Unfortunate that we can't allow a string as well, but it appears to me that would be banned by rule 3 of https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema

/approve

staebler:
> Unfortunate that we can't allow a string as well, but it appears to me that would be banned by rule 3 of https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema
> 
> /approve

You could have two separate fields--one that takes yaml and one that takes a string. The installer would fill out the one that is easier to use internally with the value from the other, if set. The installer would need to validate that if both are set that they are equivalent.

sadasu:
/lgtm

Nice improvement.

andfasano:
/hold 

let's wait for the results of the pending jobs

zaneb:
> You could have two separate fields--one that takes yaml and one that takes a string. The installer would fill out the one that is easier to use internally with the value from the other, if set. The installer would need to validate that if both are set that they are equivalent.

The main reason to support both would be so that we don't have to worry about users doing it wrong, so I think having two fields is just giving people twice as many ways to mess up :)

andfasano:
/hold cancel
metal-ipi jobs went fine

Agree also on just keeping one field for the sake of simplicity

andfasano:
There was a problem with the `controller-gen`, since `interface{}` type wasn't allowed, so I had to switch type.

pkg/asset/machines/baremetal/hosts.go
```diff
@@ -29,8 +31,14 @@ type HostSettings struct {
 	NetworkConfigSecrets []corev1.Secret
 }
 
-func createNetworkConfigSecret(host *baremetal.Host) corev1.Secret {
-	return corev1.Secret{
+func createNetworkConfigSecret(host *baremetal.Host) (*corev1.Secret, error) {
+
+	yamlNetworkConfig, err := yaml.Marshal(host.NetworkConfig)
+	if err != nil {
+		return nil, errors.Wrap(err, "Error while creating network config secret")
```
ardaguclu, line 38: Maybe we can note that this field only can be in yaml format?

andfasano: Doc updated

pkg/types/baremetal/platform.go
```diff
@@ -39,7 +40,7 @@ type Host struct {
 	HardwareProfile string           `json:"hardwareProfile"`
 	RootDeviceHints *RootDeviceHints `json:"rootDeviceHints,omitempty"`
 	BootMode        BootMode         `json:"bootMode,omitempty"`
-	NetworkConfig   string           `json:"networkConfig,omitempty"`
+	NetworkConfig   *apiextv1.JSON   `json:"networkConfig,omitempty"`
```
staebler, line 43: I guess the plus side of this is that you **could** accept a string here as well. I don't know if we want to, but you could.

pkg/types/baremetal/validation/platform.go
```diff
@@ -326,18 +325,6 @@ func validateHostsCount(hosts []*baremetal.Host, installConfig *types.InstallCon
 	return nil
 }
 
-// ensure that the NetworkConfig field contains a valid Yaml string
-func validateNetworkConfig(hosts []*baremetal.Host, fldPath *field.Path) (errors field.ErrorList) {
```
staebler, line 330: I think we still want this validation that the network config can be unmarshaled into a `map[string]interface{}`. The apiextensions JSON can accept any valid JSON value.
```
// JSON represents any valid JSON value.
// These types are supported: bool, int64, float64, string, []interface{}, map[string]interface{} and nil.
```

pkg/asset/machines/baremetal/hosts_test.go
```diff
@@ -4,8 +4,10 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/ghodss/yaml"
 	machineapi "github.com/openshift/api/machine/v1beta1"
 	"github.com/stretchr/testify/assert"
+
```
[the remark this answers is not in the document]
andfasano: Not sure if it was my ide, rearranged

pkg/types/baremetal/validation/platform.go
andfasano: Yep, good idea

andfasano:
/hold 
Checking ipv6 job failure

PR #5547, opened 2022-01-19
Bug 2040160: [Alibaba] fix EIP metering method

bd233:
In some regions, 'PayByBandwidth' is not supported, and it is changed to 'PayByTraffic'.


mtulio:
@bd233 the BZ target on the title was marked as duplicated to [2040160](https://bugzilla.redhat.com/show_bug.cgi?id=2040160), could you please rename it?

PR #5548, opened 2022-01-19
Bug 2042315: Fallback to External IP_OPTIONS when ProvisioningNetwork is Disabled

derekhiggins:
If ProvisioningNetwork is Disabled the PROVISIONING_IP_OPTIONS should be
based on the IP version on the external network.

### DIFF
--- a/pkg/asset/ignition/bootstrap/baremetal/template.go
+++ b/pkg/asset/ignition/bootstrap/baremetal/template.go
@@ -55,6 +55,9 @@ type TemplateData struct {
 
 	// Hosts is the information needed to create the objects in Ironic.
 	Hosts []*baremetal.Host
+
+	// ProvisioningNetwork displays the type of provisioning network being used
+	ProvisioningNetwork string
 }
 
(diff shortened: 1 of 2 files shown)


pkg/asset/ignition/bootstrap/baremetal/template.go
```diff
@@ -55,6 +55,9 @@ type TemplateData struct {
 
 	// Hosts is the information needed to create the objects in Ironic.
 	Hosts []*baremetal.Host
+
+	// ProvisioningNetwork displays the type of provisioning network being used
+	ProvisioningNetwork string
```
dtantsur, line 60: We could probably just have a boolean flag here.

dtantsur (commented):
/lgtm

Thanks!

pkg/asset/ignition/bootstrap/baremetal/template.go
derekhiggins: There are 3 potential values, I thought we might need more detail in future

PR #5550, opened 2022-01-20
Bug 2042655: revert delay of bootstrap control plane teardown for alibaba

staebler:
This reverts commit 6e2d76b61c98ce6b3289a2bbee2c25778a1a1d4a.

With https://github.com/openshift/machine-config-operator/pull/2919, it is no longer necessary to delay the teardown
of the bootstrap control plane. The cluster will no longer get into an unusable state when there is only a single
kube-apiserver pod running.

/hold for https://github.com/openshift/machine-config-operator/pull/2919


patrickdillon:
/approve

openshift/machine-config-operator#2919 has merged

/hold cancel

PR #5553, opened 2022-01-20
[WIP] baremetal: use combined Ironic, drop MariaDB

dtantsur:
This replaces 3 containers (ironic-api, ironic-conductor and mariadb)
with 1 (using sqlite as a backend), significantly reducing the resource
footprint.


data/data/bootstrap/baremetal/files/usr/local/bin/stopironic.sh
```diff
@@ -2,7 +2,7 @@
 
 set -x 
 
-for name in ironic-api ironic-conductor ironic-inspector ironic-deploy-ramdisk-logs ironic-inspector-ramdisk-logs dnsmasq httpd mariadb ipa-downloader coreos-downloader; do
+for name in ironic ironic-api ironic-conductor ironic-inspector ironic-deploy-ramdisk-logs ironic-inspector-ramdisk-logs dnsmasq httpd mariadb ipa-downloader coreos-downloader; do
```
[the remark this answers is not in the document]
dtantsur: I wonder if we should be able to stop old containers? Or do we assume that bootstrap is too temporary?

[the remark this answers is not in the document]
derekhiggins: Ya bootstrap is ephemeral, these files are put into the bootstrap VM when the installer creates it, they should match as any changes would involve rerunning the installer and starting a new VM.

ardaguclu:
Probably it is naive question but in which container SQLite is installed?, because there is no clue about SQLite.

dtantsur:
> Probably it is naive question but in which container SQLite is installed?, because there is no clue about SQLite.

@ardaguclu SQLite is a library, not a service, so you won't see it here when starting containers. Inspector and Ironic both have their own local databases.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -222,45 +214,35 @@ podman run -d --net host --privileged --name image-customization \
...
-     -v $AUTH_DIR:/auth:ro \
-     -v $IRONIC_SHARED_VOLUME:/shared:z ${IRONIC_IMAGE}
-
 podman run -d --name ironic-ramdisk-logs \
      --restart on-failure \
      --entrypoint /bin/runlogwatch.sh \
      -v $IRONIC_SHARED_VOLUME:/shared:z ${IRONIC_IMAGE}
 
 # Failure to start image-customization results in a very confusing error
+sleep 10
```
hardys, line 245: Can we do something better than sleep here, e.g wait for the desired status instead?

elfosardo, line 245: wow

andfasano: I agree that even though this part could be improved, the sleep was present also before

PR #5554, opened 2022-01-20
Add a way to override the interface used by bootstrap node

bn222:
Currently, the provision interface name is hardcoded to ens4. On my setup, this interface has a different name but there is no way to override this.

This patch adds a way to specify a different name.

### DIFF
--- a/pkg/asset/ignition/bootstrap/baremetal/template.go
+++ b/pkg/asset/ignition/bootstrap/baremetal/template.go
@@ -75,7 +75,11 @@ func GetTemplateData(config *baremetal.Platform, networks []types.MachineNetwork
 		cidr, _ := config.ProvisioningNetworkCIDR.Mask.Size()
 		templateData.ProvisioningCIDR = cidr
 		templateData.ProvisioningIPv6 = config.ProvisioningNetworkCIDR.IP.To4() == nil
-		templateData.ProvisioningInterface = "ens4"
+		if config.BootstrapProvisioningInterfaceOverride == "" {
+			templateData.ProvisioningInterface = "ens4"
+		} else {
+			templateData.ProvisioningInterface = config.BootstrapProvisioningInterfaceOverride
+		}
 		templateData.ProvisioningDNSMasq = true
 	}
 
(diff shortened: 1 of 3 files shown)


zshi-redhat:
@bn222 Some CI jobs are failing (e.g. `verify-codegen`), you may need to run the `hack/verify-codegen.sh` to generate the code and commit it to the same PR.

zshi-redhat:
For other reviewers, this is an issue Balazs found on the ARM platform where it has a different interface name to ens4 (on x86 VMs). I think we can either provides an override interface as what this PR does or detect the platform to use the right interface name (assuming the interface name would persist across all ARM VMs).

bn222:
@zshi-redhat Thanks for adding the comment. I've run `hack/verify-codegen.sh` and added it to the PR.

pkg/asset/ignition/bootstrap/baremetal/template.go
```diff
@@ -95,7 +99,11 @@ func GetTemplateData(config *baremetal.Platform, networks []types.MachineNetwork
 		}
 		templateData.ProvisioningDHCPAllowList = strings.Join(dhcpAllowList, " ")
 	case baremetal.DisabledProvisioningNetwork:
-		templateData.ProvisioningInterface = "ens3"
+		if config.BootstrapProvisioningInterfaceOverride == "" {
+			templateData.ProvisioningInterface = "ens3"
+		} else {
+			templateData.ProvisioningInterface = config.BootstrapProvisioningInterfaceOverride
```
hardys, line 105: This nic is defined in the (installer created) bootstrap VM - so I'm not sure it's reasonable to expect the user to somehow know what the appropriate nic name is here?

I'm wondering why the VM nic name is different in your environment (which is ARM I believe) - is there some way we can make the nic name consistent between architectures instead, or else detect the required value based on the architecture?

hardys:
> For other reviewers, this is an issue Balazs found on the ARM platform where it has a different interface name to ens4 (on x86 VMs)

Do we know why the name is different?  It seems weird that the nic number would increment if the VM definition is the same e.g we didn't add any nics?

I'm wondering if there's some way to make the name consistent, and if that's not possible then IMO we should set the value inside the installer and not expose this implementation detail to the user.

jeffdyoung:
@hardys For me the nics are named different between arm/x86 because the bootstrap vm created by libvirt uses pci vs pcie controllers:
x86: ```<controller type="pci" index="0" model="pci-root"/>``` results in `ensX`
arm64: ```<controller type="pci" index="0" model="pcie-root"/>``` results in `enpXs0`

More on nic naming in the [rhel 8 network doc](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/consistent-network-interface-device-naming_configuring-and-managing-networking#predictable-network-interface-device-names-on-the-x86_64-platform-explained_consistent-network-interface-device-naming)

I have time to look into what it would take to discover the right nic name in the bootstrap container based on mac address (or something we don't have to hardcode), if you all agree and think it's feasible.

bn222:
Superseded by https://github.com/openshift/installer/pull/5698

PR #5556, opened 2022-01-20
vsphere: make cluster destroy more resilient to api failures

r4f4:
Trying to find/delete a resource may fail, so keep trying until it
succeeds or the user stops the application.

### DIFF
--- a/pkg/destroy/vsphere/virtualmachines.go
+++ b/pkg/destroy/vsphere/virtualmachines.go
@@ -1,0 +1,114 @@
+package vsphere
+
+import (
+	"context"
+	"time"
+
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/property"
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/types"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+func (o *ClusterUninstaller) getVirtualMachineManagedObjects(ctx context.Context, moRef []types.ManagedObjectReference) ([]mo.VirtualMachine, error) {
+	var virtualMachineMoList []mo.VirtualMachine
+	if len(moRef) > 0 {
+		pc := property.DefaultCollector(o.Client)
+		err := pc.Retrieve(ctx, moRef, nil, &virtualMachineMoList)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return virtualMachineMoList, nil
+}
+
+func (o *ClusterUninstaller) listVirtualMachines(ctx context.Context) ([]mo.VirtualMachine, error) {
+	virtualMachineList, err := o.getAttachedObjectsOnTag("VirtualMachine")
+	if err != nil {
+		return nil, err
+	}
+
+	return o.getVirtualMachineManagedObjects(ctx, virtualMachineList)
+}
+
+func isPoweredOff(vmMO mo.VirtualMachine) bool {
+	return vmMO.Summary.Runtime.PowerState == "poweredOff"
+}
+
+func (o *ClusterUninstaller) stopVirtualMachine(ctx context.Context, vmMO mo.VirtualMachine) error {
+	virtualMachineLogger := o.Logger.WithField("VirtualMachine", vmMO.Name)
+	if !isPoweredOff(vmMO) {
+		vm := object.NewVirtualMachine(o.Client, vmMO.Reference())
+		task, err := vm.PowerOff(ctx)
+		if err == nil {
+			err = task.Wait(ctx)
+		}
+		if err != nil {
+			virtualMachineLogger.Debug(err)
+			return err
+		}
+	}
+	virtualMachineLogger.Debug("Powered off")
+	return nil
+}
+
+func (o *ClusterUninstaller) stopVirtualMachines() error {
+	ctx, cancel := context.WithTimeout(o.context, time.Minute*30)
+	defer cancel()
+
+	o.Logger.Debug("Power Off Virtual Machines")
+	found, err := o.listVirtualMachines(ctx)
+	if err != nil {
+		o.Logger.Debug(err)
+		return err
+	}
+
+	var errs []error
+	for _, vmMO := range found {
+		if !isPoweredOff(vmMO) {
+			if err := o.stopVirtualMachine(ctx, vmMO); err != nil {
+				errs = append(errs, err)
+			}
+		}
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+func (o *ClusterUninstaller) deleteVirtualMachine(ctx context.Context, vmMO mo.VirtualMachine) error {
+	virtualMachineLogger := o.Logger.WithField("VirtualMachine", vmMO.Name)
+	vm := object.NewVirtualMachine(o.Client, vmMO.Reference())
+	task, err := vm.Destroy(ctx)
+	if err == nil {
+		err = task.Wait(ctx)
+	}
+	if err != nil {
+		virtualMachineLogger.Debug(err)
+		return err
+	}
+	virtualMachineLogger.Info("Destroyed")
+	return nil
+}
+
+func (o *ClusterUninstaller) deleteVirtualMachines() error {
+	ctx, cancel := context.WithTimeout(o.context, time.Minute*30)
+	defer cancel()
+
+	o.Logger.Debug("Delete Virtual Machines")
+	found, err := o.listVirtualMachines(ctx)
+	if err != nil {
+		o.Logger.Debug(err)
+		return err
+	}
+
+	var errs []error
+	for _, vmMO := range found {
+		if err := o.deleteVirtualMachine(ctx, vmMO); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
--- a/pkg/destroy/vsphere/vsphere.go
+++ b/pkg/destroy/vsphere/vsphere.go
@@ -19,11 +19,14 @@ import (
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/openshift/installer/pkg/destroy/providers"
 	installertypes "github.com/openshift/installer/pkg/types"
 	vspheretypes "github.com/openshift/installer/pkg/types/vsphere"
 )
+
+var defaultTimeout = time.Minute * 5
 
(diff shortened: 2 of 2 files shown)


jcpowermac:
Two concerns only related to CI:
- for loops without sleep will really hammer the vCenter APIs - that are already taxed
- the destroy operation never exiting - causing eventual CI pod timeouts

pkg/destroy/vsphere/vsphere.go
```diff
@@ -227,59 +244,92 @@ func (o *ClusterUninstaller) Run() (*installertypes.ClusterQuota, error) {
...
-			return nil, err
-		}
+	// First, make sure all the VMs are terminated so the cluster is not creating any new resources
+	var virtualMachineMoList []mo.VirtualMachine
+	err = wait.PollInfinite(
+		time.Second*10,
+		func() (done bool, err error) {
+			if len(virtualMachineList) > 0 && len(virtualMachineMoList) == 0 {
+				o.Logger.Debug("Find VirtualMachine objects")
+				virtualMachineMoList, err = getVirtualMachineManagedObjects(context.TODO(), o.Client, virtualMachineList)
```
staebler, line 258: Let's get the list anew each time through. We want to loop until the machine is reporting powered off. This requires re-querying for the machine each time. We also want a separate list for machines that we have already sent the power-off request for, so that we are not repeating that request.

Moreover, I think we need to actually re-query for object by tag each time through this loop. Until all of the machines are powered-off, the cluster may attempt to create new machines. We should be resilient to that.

staebler (commented):
We should query for the objects by tag after all of the machines have been terminated. If we are only querying before, then there is no benefit to waiting for the machines to terminate first.

r4f4:
Tried to follow the GCP model of stages with retry. I'm just waiting on #5558 to be reviewed so I can integrate the same solution here before removing the `hold` label.

pkg/destroy/vsphere/virtualmachines.go
```diff
@@ -0,0 +1,128 @@
...
+	for _, attachedObject := range tagAttachedObjects {
+		for _, ref := range attachedObject.ObjectIDs {
+			if ref.Reference().Type == "VirtualMachine" {
+				virtualMachineList = append(virtualMachineList, ref.Reference())
+			}
+		}
+	}
+
+	var virtualMachineMoList []mo.VirtualMachine
+	if len(virtualMachineList) > 0 {
```
staebler, line 43: Move the check for an empty list into `getVirtualMachineManagedObjects`.

pkg/destroy/vsphere/virtualmachines.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (o *ClusterUninstaller) deleteVirtualMachine(vmMO mo.VirtualMachine) error {
+	ctx, cancel := o.contextWithTimeout()
+	defer cancel()
+
+	virtualMachineLogger := o.Logger.WithField("VirtualMachine", vmMO.Name)
+	vm := object.NewVirtualMachine(o.Client, vmMO.Reference())
+	task, err := vm.Destroy(ctx)
+	if err == nil {
+		err = task.Wait(ctx)
```
staebler, line 102: Do you know if the `Destroy` waits for the VM to actually be destroyed or does it just wait for the VM to be put into a destroying state? If it is the former, then should we try to parallelize the destroys?

pkg/destroy/vsphere/vsphere.go
```diff
@@ -21,6 +23,13 @@ import (
 	"github.com/openshift/installer/pkg/destroy/providers"
 	installertypes "github.com/openshift/installer/pkg/types"
 	vspheretypes "github.com/openshift/installer/pkg/types/vsphere"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
```
staebler, line 28: Move these to the group above the installer imports.

pkg/destroy/vsphere/virtualmachines.go
```diff
@@ -0,0 +1,128 @@
...
+				errs = append(errs, err)
+			}
+		}
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+func (o *ClusterUninstaller) deleteVirtualMachine(vmMO mo.VirtualMachine) error {
+	ctx, cancel := o.contextWithTimeout()
```
staebler, line 95: Is the 2-minute timeout enough? We were using a 30-minute timeout for the entire operation of powering off and destroying all of the VMs.

pkg/destroy/vsphere/vsphere.go
```diff
@@ -52,98 +63,127 @@ func New(logger logrus.FieldLogger, metadata *installertypes.ClusterMetadata) (p
...
-	return virtualMachineMoList, nil
+
+	var folderList []types.ManagedObjectReference
+	for _, attachedObject := range tagAttachedObjects {
+		for _, ref := range attachedObject.ObjectIDs {
+			if ref.Reference().Type == "Folder" {
+				folderList = append(folderList, ref.Reference())
+			}
+		}
+	}
```
staebler, line 118: Can you add the filtering on type to the `getAttachedObjectsOnTag` function so that we are not duplicating this logic?

pkg/destroy/vsphere/vsphere.go
```diff
@@ -52,98 +63,127 @@ func New(logger logrus.FieldLogger, metadata *installertypes.ClusterMetadata) (p
...
+
+	var folderMoList []mo.Folder
+	if len(folderList) > 0 {
```
staebler, line 121: As with `listVirtualMachines`, can you move the check for an empty list into the `getFolderManagedObjects` function?

pkg/destroy/vsphere/virtualmachines.go
```diff
@@ -0,0 +1,128 @@
...
+	return virtualMachineMoList, nil
+}
+
+func (o *ClusterUninstaller) listVirtualMachines() ([]mo.VirtualMachine, error) {
+	tagAttachedObjects, err := o.getAttachedObjectsOnTag()
+	if err != nil && !isNotFound(err) {
+		return nil, err
+	}
+
+	// Seperate the objects attached to the tag based on type
```
staebler, line 31:
```suggestion
	// Separate the objects attached to the tag based on type
```

pkg/destroy/vsphere/vsphere.go
```diff
@@ -52,98 +63,127 @@ func New(logger logrus.FieldLogger, metadata *installertypes.ClusterMetadata) (p
...
+	if len(folderMoList) > 1 {
+		return errors.Errorf("Expected 1 Folder per tag but got %d", len(folderMoList))
+	}
+
+	if len(folderMoList) == 0 {
+		o.Logger.Debug("All folders deleted")
+		return nil
+	}
+
+	// If there are no children in the folder go ahead an remove it
```
staebler, line 154:
```suggestion
	// If there are no children in the folder, go ahead and remove it
```

pkg/destroy/vsphere/virtualmachines.go
```diff
@@ -0,0 +1,128 @@
...
+func (o *ClusterUninstaller) stopVirtualMachine(vmMO mo.VirtualMachine) error {
+	ctx, cancel := o.contextWithTimeout()
+	defer cancel()
+
+	virtualMachineLogger := o.Logger.WithField("VirtualMachine", vmMO.Name)
+	if !isPoweredOff(vmMO) {
+		vm := object.NewVirtualMachine(o.Client, vmMO.Reference())
+		task, err := vm.PowerOff(ctx)
+		if err == nil {
+			err = task.Wait(ctx)
```
staebler, line 63: Similar question here to the question about calling `Destroy`. Does calling `PowerOff` wait for the VM to actually power off before returning? If not, then we need some subsequent wait for the VM to be reporting that it is actually powered off before proceeding to the stage where we are destroying the VM.

staebler (commented):
The break down into stages looks great. It is a nice improvement. Thank you.

pkg/destroy/vsphere/virtualmachines.go
jcpowermac: It won't block if you don't wait for the task and I am unsure how otherwise you would get the status - there is probably a way just not finding. it.

https://vdc-repo.vmware.com/vmwb-repository/dcr-public/6b586ed2-655c-49d9-9029-bc416323cb22/fa0b429a-a695-4c11-b7d2-2cbc284049dc/doc/vim.ManagedEntity.html#destroy

pkg/destroy/vsphere/virtualmachines.go
r4f4: `PowerOff` [1] returns a `Task` [2] from a soap roundtrip request. There is `WaitForPowerState` [3] in the API but I haven't seen that one used anywhere. I'm new to the `vsphere` API, if you know of a better way, let me know and I'll change the code.

[1]  https://github.com/vmware/govmomi/blob/master/object/virtual_machine.go#L113-L124
[2] https://github.com/vmware/govmomi/blob/master/vim25/methods/methods.go#L9114-L9124
[3] https://github.com/vmware/govmomi/blob/master/object/virtual_machine.go#L819-L840

pkg/destroy/vsphere/virtualmachines.go
r4f4: I can increase the `defaultTimeout` from 2min to something else (I was just using the same value as in GCP) or I can use a non-default here. Just let me know which timeout value you'd prefer.

staebler: I think GCP is different because the requests made in GCP do not wait for long-running operations to complete before returning. The requests return when the operation has started and the state of the resource has been changed to indicate that. It appears that the vSphere API requests wait until the long-running operation completes before returning.

r4f4:
Fixed the timeouts and rebased on top of current master.

pkg/destroy/vsphere/vsphere.go
```diff
@@ -54,98 +57,122 @@ func New(logger logrus.FieldLogger, metadata *installertypes.ClusterMetadata) (p
...
-			vm := object.NewVirtualMachine(client, vmMO.Reference())
-			if vmMO.Summary.Runtime.PowerState == "poweredOn" {
-				task, err := vm.PowerOff(ctx)
-				if err != nil {
-					return err
-				}
-				task.Wait(ctx)
-				virtualMachineLogger.Debug("Powered off")
-			}
+	o.Logger.Debug("Find attached %s on tag", objType)
```
r4f4, line 76: Ops, this should be `o.Logger.Debugf`

staebler (commented):
Looks good to me, save for the nit from https://github.com/openshift/installer/pull/5556#pullrequestreview-914669628.
/approve

staebler (commented):
/lgtm

PR #5557, opened 2022-01-21
Bug 2042826: Set 1 ingresscontroller replica for private SNO

Miciah:
When generating a default ingresscontroller manifest, check the infrastructure topology, and set the ingresscontroller's replicas to 1 if the topology is "SingleReplica" (i.e., Single-Node OpenShift).

When the publishing strategy is "Internal", the installer generates a default ingresscontroller manifest.  Before this PR, the installer did not check the cluster topology and always left replicas unspecified in the generated ingresscontroller.  As a result, the ingress operator used the default value of 2 replicas, causing the ingress operator to try to schedule 2 pod replicas for the ingresscontroller, which would fail on single-node clusters because only 1 pod replica can be scheduled per ingresscontroller per node.

This issue does not arise when the publishing strategy is "External" because in this case, the installer does not generate a default ingresscontroller manifest; instead, the ingress operator creates the default ingresscontroller, and the ingress operator does check the topology and set replicas appropriately in the ingresscontroller that it creates.

* `pkg/asset/manifests/infrastructure.go` (`Generate`): Factor out logic for computing the control-plane topology and infrastructure topology.
* `pkg/asset/manifests/ingress.go` (`generateDefaultIngressController`): Check the infrastructure topology and set replicas to 1 if the topology is "SingleReplica".
* `pkg/asset/manifests/ingress_test.go`: New file.
(`TestGenerateDefaultIngressController`): New test.  Verify that `generateDefaultIngressController` behaves correctly.
(`withPublish`): New helper for `icBuildNamespace`.
(`ingresscontrollerOption`, `ingresscontrollerBuildNamespace`): New types for building test inputs.
(`ingresscontrollerBuild`): New variable for building test inputs.
(`build`, `withLoadBalancer`, `withReplicas`, `withScope`): New helpers for `ingresscontrollerBuildNamespace`.
* `pkg/asset/manifests/utils.go` (`getControlPlaneTopology`)
(`getInfrastructureTopology`): New functions.
* `pkg/asset/manifests/utils_test.go`: New file.
(`TestGetInfrastructureTopology`): New test.  Verify that `getInfrastructureTopology` behaves correctly.
(`withControlPlaneReplicas`, `withWorkerReplicas`): New helpers for `TestGetInfrastructureTopology`.

### DIFF
--- a/pkg/asset/manifests/ingress.go
+++ b/pkg/asset/manifests/ingress.go
@@ -97,6 +97,11 @@ func (ing *Ingress) generateClusterConfig(config *types.InstallConfig) ([]byte, 
 func (ing *Ingress) generateDefaultIngressController(config *types.InstallConfig) ([]byte, error) {
 	switch config.Publish {
 	case types.InternalPublishingStrategy:
+		var replicas *int32
+		if getInfrastructureTopology(config) == configv1.SingleReplicaTopologyMode {
+			one := int32(1)
+			replicas = &one
+		}
 		obj := &operatorv1.IngressController{
 			TypeMeta: metav1.TypeMeta{
 				APIVersion: operatorv1.GroupVersion.String(),
@@ -113,6 +118,7 @@ func (ing *Ingress) generateDefaultIngressController(config *types.InstallConfig
 						Scope: operatorv1.InternalLoadBalancer,
 					},
 				},
+				Replicas: replicas,
 			},
 		}
 		return yaml.Marshal(obj)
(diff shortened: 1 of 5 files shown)


pkg/asset/manifests/ingress.go
```diff
@@ -113,6 +118,7 @@ func (ing *Ingress) generateDefaultIngressController(config *types.InstallConfig
 						Scope: operatorv1.InternalLoadBalancer,
 					},
 				},
+				Replicas: replicas,
```
staebler, line 121: I would have expected that the ingress operator would be deducing the correct number of replicas to use by inspecting the Infrastructure resource.

staebler (commented):
The code changes look good. I just have the one open question about whether this is the appropriate way to handle the topology.

Miciah:
Agreed, we can resolve the issue in the ingress operator.  
/close

PR #5558, opened 2022-01-21
Bug 2021041: vsphere: Not found TagCategory when destroying ipi cluster

r4f4:
While https://github.com/vmware/govmomi/issues/2710 is not fixed
upstream, let's carry our own fixed version: try out all the category
ids before erroring out.

Signed-off-by: Rafael Fonseca <r4f4rfs@gmail.com>

### DIFF
--- a/pkg/destroy/vsphere/vsphere.go
+++ b/pkg/destroy/vsphere/vsphere.go
@@ -3,11 +3,12 @@ package vsphere
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/pbm"
 	pbmtypes "github.com/vmware/govmomi/pbm/types"
@@ -17,6 +18,7 @@ import (
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/openshift/installer/pkg/destroy/providers"
 	installertypes "github.com/openshift/installer/pkg/types"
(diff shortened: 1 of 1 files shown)


pkg/destroy/vsphere/vsphere.go
```diff
@@ -189,11 +189,19 @@ func deleteTag(ctx context.Context, client *rest.Client, tagID string) error {
...
-	return err
+
+	for _, id := range ids {
+		category, err := tagManager.GetCategory(ctx, id)
+		if err == nil && category.Name == categoryID {
+			return tagManager.DeleteCategory(ctx, category)
+		}
+	}
+
+	return errors.Errorf("category %s not found", categoryID)
```
staebler, line 204: I think this should not return an error. We can print a debug statement. But the destroyer should accept a missing category as acceptable.

pkg/destroy/vsphere/vsphere.go
```diff
@@ -189,11 +189,19 @@ func deleteTag(ctx context.Context, client *rest.Client, tagID string) error {
...
-		err = tagManager.DeleteCategory(ctx, category)
+	ids, err := tagManager.ListCategories(ctx)
+	if err != nil {
+		return err
 	}
-	return err
+
+	for _, id := range ids {
+		category, err := tagManager.GetCategory(ctx, id)
+		if err == nil && category.Name == categoryID {
```
staebler, line 199: We should not be ignoring all errors here. We can ignore not found errors, so that we don't get into the same problem that the `GetCategories` function has in govmomi. But if an error is something else, such as failure to connect, then that error should be handled.

staebler (commented):
This is looking promising.

staebler (commented):
Looks good to me, save for a small nit.

pkg/destroy/vsphere/vsphere.go
```diff
@@ -187,13 +189,39 @@ func deleteTag(ctx context.Context, client *rest.Client, tagID string) error {
...
+		return err
 	}
-	return err
+
+	var loopError error
+	for _, id := range ids {
+		category, err := tagManager.GetCategory(ctx, id)
+		if err != nil {
+			if !strings.HasSuffix(err.Error(), http.StatusText(http.StatusNotFound)) {
+				loopError = err
```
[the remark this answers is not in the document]
staebler: It is probably also helpful to include in the collected error the ID that was queried that resulted in the error.
```
errs = append(errs, errors.Wrapf(err, "could not get category %q", id))
```

pkg/destroy/vsphere/vsphere.go
```diff
@@ -21,6 +23,8 @@ import (
 	"github.com/openshift/installer/pkg/destroy/providers"
 	installertypes "github.com/openshift/installer/pkg/types"
 	vspheretypes "github.com/openshift/installer/pkg/types/vsphere"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
```
staebler, line 27: Move this up into the section above the installer imports, with the govmomi imports (which should really be in the same section as the logrus imports).

staebler (commented):
Looks good. Just needs the import grouping tidied up.

jcpowermac:
Since its fixed now upstream should we just update govmomi?
https://github.com/vmware/govmomi/releases/tag/v0.27.4

staebler:
> Since its fixed now upstream should we just update govmomi? https://github.com/vmware/govmomi/releases/tag/v0.27.4

It may be good to get this change in for the purposes of backporting it. We need to evaluate whether we can update govmomi in previous versions.

staebler (commented):
/lgtm
/approve

PR #5564, opened 2022-01-22
Bug 2041694: [Alibaba] fix system disk category of bootstrap

bd233:
Remove hard coding, support users can specify cloud_efficiency in regions that do not support cloud_essd disk category


kirankt (approved):
/lgtm

PR #5565, opened 2022-01-22
Bug 2041926: [Alibaba] fix creating public record being skipped

bd233:
If the user chooses a base domain for which there is no zone, creating the A record in the zone is simply skipped rather than raising an error.


patrickdillon:
In addition to the changes in this PR, it should be validated that the provided base domain is valid. It is better to fail early than wait for terraform to fail. That validation can be added in a followup PR.

/approve
/lgtm

PR #5570, opened 2022-01-24
openstack: Migration script should --force drain

pierreprinetti:
Draining a node for ops purposes needs a `--force` flag to not get the
following error:

```
cannot delete Pods not managed by ReplicationController, ReplicaSet,
Job, DaemonSet or StatefulSet (use --force to override)
```

### DIFF
--- a/docs/user/openstack/README.md
+++ b/docs/user/openstack/README.md
@@ -725,7 +725,7 @@ readonly server_id
 
 # Drain the node
 oc adm cordon "$node_name"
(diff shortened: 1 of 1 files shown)


pierreprinetti:
/skip
This is a docs fix.

mandre:
The `--force` shouldn't be needed, and it's likely caused by https://bugzilla.redhat.com/show_bug.cgi?id=2038481
/hold

pierreprinetti:
/close
Thank you @mandre !

PR #5571, opened 2022-01-24
Bug 2041750: [Alibaba] fix endpoint error in some regions

bd233:
Update sdk and terraform provier version, and add some endpoints of ECS
service to fix endpoint error.


patrickdillon:
Can you separate the vendor code into a separate commit so this is easier to review?

bd233:
> Can you separate the vendor code into a separate commit so this is easier to review?

Have done.

kirankt:
The golint job failure is real. Please fix. Always run the checks listed in CONTRIBUTING.md file before your commits to catch these types of failures.

pkg/asset/installconfig/alibabacloud/client.go
```diff
@@ -331,6 +323,23 @@ func (client *Client) GetAvailableZonesByInstanceType(instanceType string) ([]st
 	return zones, nil
 }
 
+func getEndpoint(productName string, regionId string) string {
```
patrickdillon, line 326: golint is failing because `regionId` should be `regionID`

patrickdillon (commented):
/approve

This looks good but you need to fix the issue I pointed out with the golint to pass the required tests.

pkg/asset/installconfig/alibabacloud/client.go
bd233: Have done

kirankt (approved):
/lgtm

PR #5572, opened 2022-01-24
Bug 2040143: [Alibaba] remove unsupport regions

bd233:
Remove unsupport region Nanjing and Dubai. Nanjing is a local cloud and Dubai does not support private zone service.


pkg/asset/installconfig/alibabacloud/alibabacloud.go
```diff
@@ -17,6 +17,13 @@ const (
 	defaultAcceptLanguage = "en-US"
 )
 
+var unsupportedRegions = map[string]bool{
```
patrickdillon, line 20: Can you use `sets` from here: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#String

Here is an example using sets in this package: https://github.com/openshift/installer/blob/master/pkg/types/azure/validation/machinepool.go#L21-L24

patrickdillon (commented):
Generally looks good, but using `sets` as I suggested in the comment would be more in line with the rest of the codebase.

pkg/asset/installconfig/alibabacloud/alibabacloud.go
bd233: Thanks for your suggestion, I have updated.

PR #5575, opened 2022-01-25
Bug 2042370: [Alibaba] fix installer index panic

bd233:
And add NAT gateway validation to check the region whether support NAT gateway

### DIFF
--- a/pkg/asset/installconfig/alibabacloud/validation.go
+++ b/pkg/asset/installconfig/alibabacloud/validation.go
@@ -209,6 +209,7 @@ func validatePrivateZoneID(client *Client, ic *types.InstallConfig, path *field.
 func ValidateForProvisioning(client *Client, ic *types.InstallConfig, metadata *Metadata) error {
 	allErrs := field.ErrorList{}
 	allErrs = append(allErrs, validateClusterName(client, ic)...)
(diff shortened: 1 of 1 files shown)


pkg/asset/cluster/tfvars.go
```diff
@@ -741,6 +741,9 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		if err != nil {
 			return errors.Wrapf(err, "failed to list avaliable zones for NAT gateway")
 		}
+		if len(natGatewayZones.Zones) == 0 {
+			return errors.Wrapf(err, "enhanced NAT gateway is not supported in the region %s", installConfig.AlibabaCloud.Region)
+		}
```
patrickdillon, line 746: Why is this needed here as well as in the machine configs? Isn't it sufficient to check once?

bd233: After I added this check, I realized it would be more appropriate to add a validation, but I'm not sure if it's superfluous, so I kept it

patrickdillon: Yeah you can remove this entire check from here. The install config validation will cover it.

bd233: Have removed

PR #5579, opened 2022-01-26
Bug 2046181: baremetal: wait for image-customization to come up

dtantsur:
It crashes if the provided network configuration is invalid. In this
case inspection currently fails with a generic message.

### DIFF
--- a/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
+++ b/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
@@ -3,6 +3,9 @@ set -ex
 
 . /usr/local/bin/release-image.sh
 
+# Must match ironic.service
+EXIT_CODE_NO_RESTART=42
+
 IRONIC_IMAGE=$(image_for ironic)
 IRONIC_AGENT_IMAGE=$(image_for ironic-agent)
 CUSTOMIZATION_IMAGE=$(image_for image-customization-controller)
@@ -241,6 +244,13 @@ podman run -d --name ironic-ramdisk-logs \
      --restart on-failure \
      --entrypoint /bin/runlogwatch.sh \
      -v $IRONIC_SHARED_VOLUME:/shared:z ${IRONIC_IMAGE}
+
+# Failure to start image-customization results in a very confusing error
+icc_id=$(podman ps --filter name=image-customization --filter status=running --format '{{`{{.ID}}`}}')
+if [ -z "$icc_id" ]; then
+    echo The image-customization service crashed after start, check its logs
+    exit $EXIT_CODE_NO_RESTART
+fi
 
 set +x
 AUTH_DIR=/opt/metal3/auth
--- a/data/data/bootstrap/baremetal/systemd/units/ironic.service
+++ b/data/data/bootstrap/baremetal/systemd/units/ironic.service
@@ -11,6 +11,8 @@ RemainAfterExit=yes
 
 Restart=on-failure
 RestartSec=10
+# Used in startironic.sh to indicate a fatal failure
+RestartPreventExitStatus=42
 TimeoutStartSec=600
 
 [Install]


dtantsur:
I guess we have a whole different story that the installer doesn't care if bootstrap fails. I don't think we can have a better fix though, not unless we rework how ironic is started.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -204,6 +204,19 @@ podman run -d --net host --privileged --name image-customization \
...
+attempt=0
+until [ $attempt -eq 15 ] || curl -I "http://localhost:8084" > /dev/null; do
+    echo Waiting for image-customization controller to come up...
+    sleep 2
+    attempt=$(( $attempt + 1 ))
+done
+if [ $attempt -eq 15 ]; then
+    echo The image-customization controller did not come up in 30 seconds
+    podman logs image-customization
+    exit 1
```
zaneb, line 217: This will just restart the script I think? https://github.com/openshift/installer/blob/master/data/data/bootstrap/baremetal/systemd/units/ironic.service#L12

We really need a way to stop installation.

Using https://github.com/openshift/installer/blob/master/docs/dev/bootstrap_services.md#bootstrap-service-records looks like it would really help for debugging, but I'm not clear on whether anything in the installer will actually bail out once a failure is recorded. @staebler what is the recommended way of bailing out from a non-recoverable error?

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -204,6 +204,19 @@ podman run -d --net host --privileged --name image-customization \
     --secret pull-secret,mode=400 \
     ${CUSTOMIZATION_IMAGE}
 
+# We're not interesting for the exit code, just that the server is available
```
zaneb, line 207: It would actually be sufficient to just check if the container has exited after a short interval (perhaps after the sleep 10 on line 233).

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
dtantsur: We can tell systemd to treat some exit codes as non-restartable `RestartPreventExitStatus=42` and use `exit 42` here. Will it work as a quick workaround?

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
dtantsur: Are we sure 10 seconds are always enough? I guess it's not going to take too long...

elfosardo:
e2e-alibaba is permanently failing, I think we should override it for the time being

ardaguclu:
Yes, e2e-alibaba is permanently failing.

/approve

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
staebler: There is no current way for the bootstrap machine to inform the installer that it should bail out of waiting for the installation to complete.

dtantsur: Thank you! Then, I think, aborting this script is the most obvious thing we can do.

elfosardo (approved):

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -242,6 +245,14 @@ podman run -d --name ironic-ramdisk-logs \
      --entrypoint /bin/runlogwatch.sh \
      -v $IRONIC_SHARED_VOLUME:/shared:z ${IRONIC_IMAGE}
 
+# Failure to start image-customization results in a very confusing error
+icc_id=$(podman ps --filter name=image-customization --filter status=running --format '{{`{{.ID}}`}}')
```
andfasano, line 249: Shouldn't be `... --format '{{.ID}}')` ?

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -242,6 +245,14 @@ podman run -d --name ironic-ramdisk-logs \
...
+if [ -z "$icc_id" ]; then
+    echo The image-customization controller crashed after start
```
andfasano, line 251: typo: it's the server

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -242,6 +245,14 @@ podman run -d --name ironic-ramdisk-logs \
...
+    podman logs image-customization || true
```
andfasano, line 252: I'm not sure about this one (as potentially could expose unwanted information): in case of bootstrap failure the user couldn't check the logs of the container?

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
dtantsur: Nope. This file is already a Go template, we need to escape {{ in it. That's the way I found on the internet.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
dtantsur: I assume that people who have access to journald logs also have access to podman? I also don't think ICC actually logs anything sensitive.

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
andfasano: Oh yes that's right, I've missed that, thanks

data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
andfasano: Here we don't have control on what the ICC could log, in future (and it could be captured by the journalctl reported in CI jobs in case of bootstrap failure). The echo msg seems clear enough to point whoever is troubleshooting the problem in the icc direction.

dtantsur: What threat model do you have in mind? People who can read this log, can also read podman logs. It's pure convenience here.

PR #5583, opened 2022-01-27
Bug 2046025: [Alibaba] fix destory exist private zone

bd233:
Should not destroy pre-configured alicloud DNS private zone

### DIFF
--- a/pkg/asset/cluster/alibabacloud/alibabacloud.go
+++ b/pkg/asset/cluster/alibabacloud/alibabacloud.go
@@ -10,5 +10,6 @@ func Metadata(config *types.InstallConfig) *alibabacloud.Metadata {
 	return &alibabacloud.Metadata{
 		Region:        config.Platform.AlibabaCloud.Region,
 		ClusterDomain: config.ClusterDomain(),
+		PrivateZoneID: config.AlibabaCloud.PrivateZoneID,
 	}
 }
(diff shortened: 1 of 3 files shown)


pkg/asset/cluster/alibabacloud/alibabacloud.go
```diff
@@ -10,5 +10,6 @@ func Metadata(config *types.InstallConfig) *alibabacloud.Metadata {
 	return &alibabacloud.Metadata{
 		Region:        config.Platform.AlibabaCloud.Region,
 		ClusterDomain: config.ClusterDomain(),
+		PrivateZoneID: config.AlibabaCloud.PrivateZoneID,
```
patrickdillon, line 13: Generally we try to avoid adding data to the Metadata when it is not necessary, but I do not see another way of determining for certain that the user provided the private dns zone.

pkg/destroy/alibabacloud/alibabacloud.go
```diff
@@ -1205,6 +1207,10 @@ func (o *ClusterUninstaller) detachPolicy(policyName string, policyType string,
 }
 
 func (o *ClusterUninstaller) deletePrivateZones(logger logrus.FieldLogger) (err error) {
+	if o.PrivateZoneID != "" {
+		return nil
```
patrickdillon, line 1211: I believe we were relying on this to delete the installer created records in the private dns zone, so I think new logic would need to be added (I could be wrong). I think we can merge this anyway and create a new bug. I would rather leak records than destroy records unintentionally.

patrickdillon (commented):
This looks sufficient but we need to follow up on whether private records are being leaked.

/lgtm
/approve

PR #5585, opened 2022-01-27
Bug 2047258: Read GovCloud from RHCOS stream

patrickdillon:
AMIs for GovCloud regions have been added to the RHCOS stream. Remove
validation requiring users to provide an AMI.


jstuever:
Looks like some unit tests need modification to no longer use these regions....
https://github.com/openshift/installer/blob/4208655768d0f08d5913d786f7bab0b08d38ed70/pkg/types/aws/validation/machinepool_test.go#L83-L104

patrickdillon:
> Looks like some unit tests need modification to no longer use these regions....
> 
> https://github.com/openshift/installer/blob/4208655768d0f08d5913d786f7bab0b08d38ed70/pkg/types/aws/validation/machinepool_test.go#L83-L104

Thanks for catching that. Let's see if this newest push fixes it.

patrickdillon:
ok it's fixed!

PR #5586, opened 2022-01-27
oVirt Gather Debug statements

patrickdillon:
/hold

This is for debugging why oVirt is failing to get IP addresses for gather bootstrap. We can merge if we want but not necessary

### DIFF
--- a/cmd/openshift-install/gather.go
+++ b/cmd/openshift-install/gather.go
@@ -106,7 +106,16 @@ func runGatherBootstrapCmd(directory string) (string, error) {
 	bootstrap := gatherBootstrapOpts.bootstrap
 	port := 22
 	masters := gatherBootstrapOpts.masters
-	if bootstrap == "" && len(masters) == 0 {
+
+	if len(masters) > 0 && bootstrap == "" {
+		return "", fmt.Errorf("when control plane IP addresses are provided, the bootstrap IP address is required")
+	}
+
+	if bootstrap != "" && len(masters) == 0 {
+		logrus.Warn("Only bootstrap IP address was supplied. Supplying control-plane IP addresses as well allows more complete log gathering.")
+	}
+
+	if bootstrap == "" {
 		config := &installconfig.InstallConfig{}
 		if err := assetStore.Fetch(config); err != nil {
 			return "", errors.Wrapf(err, "failed to fetch %s", config.Name())
@@ -115,24 +124,29 @@ func runGatherBootstrapCmd(directory string) (string, error) {
 		for _, stage := range platformstages.StagesForPlatform(config.Config.Platform.Name()) {
 			stageBootstrap, stagePort, stageMasters, err := stage.ExtractHostAddresses(directory, config.Config)
 			if err != nil {
-				return "", err
+				logrus.Debugf("Error when obtaining host IP addresses, continuing with best effort. Error message: %s", err)
 			}
+
 			if stageBootstrap != "" {
 				bootstrap = stageBootstrap
 			}
+
 			if stagePort != 0 {
 				port = stagePort
 			}
+
 			if len(stageMasters) > 0 {
 				masters = stageMasters
 			}
 		}
+	}
 
-		if bootstrap == "" || len(masters) == 0 {
-			return "", errors.New("bootstrap host address and at least one control plane host address must be provided")
-		}
-	} else if bootstrap == "" || len(masters) == 0 {
-		return "", errors.New("must provide both bootstrap host address and at least one control plane host address when providing one")
+	if bootstrap == "" {
+		return "", fmt.Errorf("failed to obtain IP address for bootstrap host")
+	}
+
+	if len(masters) == 0 {
+		logrus.Warn("No control-plane IP addresses found. Attempting to gather logs from bootstrap host only...")
 	}
 
 	return logGatherBootstrap(bootstrap, port, masters, directory)
--- a/pkg/terraform/stages/ovirt/stages.go
+++ b/pkg/terraform/stages/ovirt/stages.go
@@ -73,7 +73,7 @@ func extractOutputHostAddresses(s stages.SplitStage, directory string, ic *types
 			}
 			ip, err := findVirtualMachineIP(vmID, client)
 			if err != nil {
-				returnErr = errors.Wrapf(err, "could not find IP address for bootstrap instance %q", vmID)
+				returnErr = errors.Wrapf(err, "could not find IP address for control-plane instance %s", vmID)
 				return
 			}
 			controlPlaneIPs[i] = ip
@@ -87,7 +87,7 @@ func checkPortIsOpen(host string, port string) bool {
 	timeout := time.Second
 	conn, err := net.DialTimeout("tcp", net.JoinHostPort(host, port), timeout)
 	if err != nil {
-		logrus.Debugf("connection error: %v", err)
+		logrus.Debugf("Connection error: %v", err)
 		return false
 	}
 	if conn != nil {
@@ -123,11 +123,16 @@ func getReportedDevices(c *ovirtsdk4.Connection, vmID string) (*ovirtsdk4.Report
 
 func findVirtualMachineIP(instanceID string, client *ovirtsdk4.Connection) (string, error) {
 	reportedDeviceSlice, err := getReportedDevices(client, instanceID)
-	if err == nil {
-		return "", errors.Wrapf(err, "could not Find IP Address for vm id: %s", instanceID)
+	if err != nil {
+		return "", errors.Wrapf(err, "error retrieving reported devices when determining IP address")
 	}
 
-	for _, reportedDevice := range reportedDeviceSlice.Slice() {
+	reportedDevices := reportedDeviceSlice.Slice()
+	if len(reportedDevices) == 0 {
+		return "", fmt.Errorf("no reported devices found when determining IP address")
+	}
+
+	for _, reportedDevice := range reportedDevices {
 		ips, hasIps := reportedDevice.Ips()
 		if hasIps {
 			for _, ip := range ips.Slice() {
@@ -141,5 +146,5 @@ func findVirtualMachineIP(instanceID string, client *ovirtsdk4.Connection) (stri
 			}
 		}
 	}
-	return "", fmt.Errorf("could not find usable IP address for vm id: %s", instanceID)
+	return "", fmt.Errorf("no usable IP addresses found")
 }
--- a/pkg/asset/manifests/infrastructure.go
+++ b/pkg/asset/manifests/infrastructure.go
@@ -206,11 +206,6 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
 			}
 		}
 	case ovirt.Name:
-		config.Spec.PlatformSpec.Type = configv1.OvirtPlatformType
-		config.Status.PlatformStatus.Ovirt = &configv1.OvirtPlatformStatus{
-			APIServerInternalIP: installConfig.Config.Ovirt.APIVIP,
-			IngressIP:           installConfig.Config.Ovirt.IngressVIP,
-		}
 	default:
 		config.Spec.PlatformSpec.Type = configv1.NonePlatformType
 	}


pkg/terraform/stages/ovirt/stages.go
```diff
@@ -56,6 +58,10 @@ func extractOutputHostAddresses(s stages.SplitStage, directory string, ic *types
 			return
 		}
 		bootstrapIP = ip
+		logrus.Debug("Found bootstrapIP: %s", bootstrapIP)
+	} else {
+		returnErr = errors.New("bootstrap vm id not found in Terraform outputs")
+		return
```
staebler, line 64: You don't want the return here. Only the output file from the bootstrap stage will have `bootstrap_vm_id`. The output file from the cluster stage will have `control_plane_vm_ids`.

pkg/terraform/stages/ovirt/stages.go
```diff
@@ -65,21 +71,28 @@ func extractOutputHostAddresses(s stages.SplitStage, directory string, ic *types
...
 			if err != nil {
 				returnErr = errors.Wrapf(err, "could not find IP address for bootstrap instance %q", vmID)
 				return
 			}
 			controlPlaneIPs[i] = ip
+			logrus.Debug("Found %s", ip)
 		}
+	} else {
+		returnErr = errors.New("control plane VM IDs not found in terraform outputs")
+		return
```
staebler, line 92: Same here. There should not be a return.

pkg/terraform/stages/ovirt/stages.go
```diff
@@ -56,6 +58,10 @@ func extractOutputHostAddresses(s stages.SplitStage, directory string, ic *types
 			return
 		}
 		bootstrapIP = ip
+		logrus.Debug("Found bootstrapIP: %s", bootstrapIP)
```
staebler, line 61:
```suggestion
		logrus.Debugf("Found bootstrapIP: %s", bootstrapIP)
```

pkg/terraform/stages/ovirt/stages.go
```diff
@@ -65,21 +71,28 @@ func extractOutputHostAddresses(s stages.SplitStage, directory string, ic *types
 			return
 		}
 		controlPlaneIPs = make([]string, len(vmIDs))
+		logrus.Debug("Retrieving control plane IPs from VMs: %v", vmIDs)
```
staebler, line 74: This will only print out the address to the `vmIDs` slice and not the contents of the slice.

pkg/terraform/stages/ovirt/stages.go
```diff
@@ -65,21 +71,28 @@ func extractOutputHostAddresses(s stages.SplitStage, directory string, ic *types
...
 		for i, vmIDRaw := range vmIDs {
 			vmID, ok := vmIDRaw.(string)
 			if !ok {
 				returnErr = errors.New("could not read control plane VM ID from terraform outputs")
 				return
 			}
+			logrus.Debug("Checking vmID: ", vmID)
```
staebler, line 81:
```suggestion
			logrus.Debugf("Checking vmID: %s", vmID)
```

pkg/terraform/stages/ovirt/stages.go
```diff
@@ -65,21 +71,28 @@ func extractOutputHostAddresses(s stages.SplitStage, directory string, ic *types
...
 			ip, err := findVirtualMachineIP(vmID, client)
 			if err != nil {
 				returnErr = errors.Wrapf(err, "could not find IP address for bootstrap instance %q", vmID)
 				return
 			}
 			controlPlaneIPs[i] = ip
+			logrus.Debug("Found %s", ip)
```
staebler, line 88:
```suggestion
			logrus.Debugf("Found %s", ip)
```

pkg/terraform/stages/ovirt/stages.go
```diff
@@ -65,21 +71,28 @@ func extractOutputHostAddresses(s stages.SplitStage, directory string, ic *types
...
 		}
+	} else {
+		returnErr = errors.New("control plane VM IDs not found in terraform outputs")
+		return
 	}
 
+	logrus.Debug("Returning controlPlaneIPs %v", controlPlaneIPs)
```
staebler, line 95:
```suggestion
	logrus.Debugf("Returning controlPlaneIPs %v", controlPlaneIPs)
```

pkg/terraform/stages/ovirt/stages.go
patrickdillon: I think `vmIDs` -> `&vmIDs` should resolve this but I'm testing in CI, not locally so it might still be wrong.

I wasn't necessarily planning on merging this so I probably should have just used `Spew`

patrickdillon:
Ok these debug statements did show up in the last run but my premature return statement meant no meaningful data was collected. Let's see if the new run turns something up.

pkg/terraform/stages/ovirt/stages.go
staebler: It seems that I am out of date on what the formatter does anyway.
https://go.dev/play/p/MgH1wx_VfZ4

patrickdillon:
Not sure if those newest changes will force bootstrap failures

staebler:
> Not sure if those newest changes will force bootstrap failures

No. That is just going to keep the cluster from fully initializing. You want a manifest that cannot be applied by cluster-bootstrap.

But I suppose you many as well just change the following line so that the installer is never informed of the bootstrap completion.
https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/report-progress.sh#L29

patrickdillon:
```
time="2022-02-02T20:04:50Z" level=debug msg="Extracting oVirt Host Addresses to Gather Logs..."
time="2022-02-02T20:04:50Z" level=debug msg="Reading VM IDs from Terraform Outputs..."
time="2022-02-02T20:04:50Z" level=debug msg="Retrieving control plane IPs from VMs: &[df379c21-5638-4657-8eaa-0d4db0768fd9 69d13ae4-f935-4967-9e40-a9655427cdce 0bfd0a80-2c8c-400a-b798-2c320533fc9d]"
time="2022-02-02T20:04:50Z" level=debug msg="Checking vmID: df379c21-5638-4657-8eaa-0d4db0768fd9"
time="2022-02-02T20:04:51Z" level=debug msg="Found "
time="2022-02-02T20:04:51Z" level=debug msg="Checking vmID: 69d13ae4-f935-4967-9e40-a9655427cdce"
time="2022-02-02T20:04:51Z" level=debug msg="Found "
time="2022-02-02T20:04:51Z" level=debug msg="Checking vmID: 0bfd0a80-2c8c-400a-b798-2c320533fc9d"
time="2022-02-02T20:04:51Z" level=debug msg="Found "
time="2022-02-02T20:04:51Z" level=debug msg="Returning controlPlaneIPs &[  ]"
time="2022-02-02T20:04:52Z" level=debug msg="Extracting oVirt Host Addresses to Gather Logs..."
time="2022-02-02T20:04:52Z" level=debug msg="Reading VM IDs from Terraform Outputs..."
time="2022-02-02T20:04:53Z" level=debug msg="Found bootstrapIP: "
time="2022-02-02T20:04:53Z" level=debug msg="Returning controlPlaneIPs &[]"
time="2022-02-02T20:04:53Z" level=error msg="Attempted to gather debug logs after installation failure: bootstrap host address and at least one control plane host address must be provided"
```

patrickdillon:
Looks like this might be the problem: https://github.com/openshift/installer/blob/master/pkg/terraform/stages/ovirt/stages.go#L126

staebler:
> Looks like this might be the problem: https://github.com/openshift/installer/blob/master/pkg/terraform/stages/ovirt/stages.go#L126

LOL. Yeah, that looks pretty suspect.

patrickdillon:
> LOL. Yeah, that looks pretty suspect.

Haha. I'll try fixing i there. I am guessing wrapping a nil error returns nil, so this makes it kind of hard to diagnose.

staebler:
```
time="2022-02-02T22:49:51Z" level=debug msg="Extracting oVirt Host Addresses to Gather Logs..."
time="2022-02-02T22:49:51Z" level=debug msg="Reading VM IDs from Terraform Outputs..."
time="2022-02-02T22:49:51Z" level=debug msg="Retrieving control plane IPs from VMs: &[98a7dd84-281b-4b61-a88d-8cf7d5f77ad2 f59629a5-f69a-4258-a3f0-37747b4f577f 7e6a5ea0-76ff-4a30-a34d-20cefb18fe63]"
time="2022-02-02T22:49:51Z" level=debug msg="Checking vmID: 98a7dd84-281b-4b61-a88d-8cf7d5f77ad2"
time="2022-02-02T22:49:52Z" level=error msg="Attempted to gather debug logs after installation failure: could not find IP address for control-plane instance \"98a7dd84-281b-4b61-a88d-8cf7d5f77ad2\": could not find usable IP address for vm id: 98a7dd84-281b-4b61-a88d-8cf7d5f77ad2"
```

staebler:
I wonder if there is only an IP registered for the VM after the OS ignites.

staebler:
This is the spew of reported devices.
```
(*ovirtsdk.ReportedDeviceSlice)(0xc002dcafc0)({
 href: (*string)(<nil>),
 slice: ([]*ovirtsdk.ReportedDevice) <nil>
})
([]*ovirtsdk.ReportedDevice) <nil>
```

patrickdillon:
> I wonder if there is only an IP registered for the VM after the OS ignites.

This makes sense; and if so, I think the best I can hope for here is to add a more coherent error message. The code/current error messages make it look like there *should be* an error if there are no IP addresses.

staebler:
> > I wonder if there is only an IP registered for the VM after the OS ignites.
> 
> This makes sense; and if so, I think the best I can hope for here is to add a more coherent error message. The code/current error messages make it look like there _should be_ an error if there are no IP addresses.

I think that the installer should allow the bootstrap gather to run when there are no control plane IPs. This would be useful for azure if we wanted to do a bootstrap gather when the masters do not ignite and the terraform fails.

janosdebugs:
@patrickdillon @staebler the IP address is reported by the QEMU guest agent, which is run when the OS boots. I believe (and I may be incorrect) that it runs in a container, so it may very well be that it depends on ignition.

patrickdillon:
> I think that the installer should allow the bootstrap gather to run when there are no control plane IPs. This would be useful for azure if we wanted to do a bootstrap gather when the masters do not ignite and the terraform fails.

I have added 61d37c64e9bb1ca1683a82f1fb94ffdf392bb914 as a quick stab at this.

cmd/openshift-install/gather.go
```diff
@@ -114,25 +114,26 @@ func runGatherBootstrapCmd(directory string) (string, error) {
...
-			if err != nil {
-				return "", err
-			}
-			if stageBootstrap != "" {
-				bootstrap = stageBootstrap
+			if err != nil || stageBootstrap == "" {
+				return "", errors.Wrapf(err, "failed to obtain IP address for bootstrap host")
+			} else if err != nil {
+				logrus.Debugf("Error when obtaining host IP addresses, continuing with best effort. Error message: %s", err)
 			}
```
staebler, line 121: This logic is not quite right. (1) It should be done *after* going through all of the stages. (2) An error in extracting the host addresses does not imply that there was an error obtaining the IP address specifically for the bootstrap host. Keep in mind that each stage may have all, some, or none of the host addresses, depending upon the platform.

cmd/openshift-install/gather.go
```diff
@@ -115,24 +115,29 @@ func runGatherBootstrapCmd(directory string) (string, error) {
...
 				masters = stageMasters
 			}
 		}
+	}
 
-		if bootstrap == "" || len(masters) == 0 {
-			return "", errors.New("bootstrap host address and at least one control plane host address must be provided")
-		}
-	} else if bootstrap == "" || len(masters) == 0 {
-		return "", errors.New("must provide both bootstrap host address and at least one control plane host address when providing one")
```
patrickdillon, line 135: I'm going to add some more logic around the checking here.

cmd/openshift-install/gather.go
```diff
@@ -106,7 +106,11 @@ func runGatherBootstrapCmd(directory string) (string, error) {
 	bootstrap := gatherBootstrapOpts.bootstrap
 	port := 22
 	masters := gatherBootstrapOpts.masters
-	if bootstrap == "" && len(masters) == 0 {
+	if len(masters) > 0 && bootstrap == "" {
+		return "", fmt.Errorf("when control plane IP addresses are provided, the bootstrap IP address is required")
+	}
```
patrickdillon, line 111: I've introduced this somewhat arbitrary requirement. Alternatively we could try to determine the bootstrap IP if only control plane ips are specified. I put this in to maintain some of the prior behavior, but I think this is a super rare case.

I am wondering a bit about the usability here, though. If we start allowing users to input only the bootstrap ip will they get lazy and not enter the control plane ips, not realizing how important having the control plane IPs is?

I guess the solution is to enter a strong warning message

staebler: I wonder how many non-machine users supply IPs rather than relying on the gathering of the IPs from the terraform results. I suggest that we allow them to omit control plane IPs and warn if they are omitted. We should always require them to specify the bootstrap IP when specifying the control plane IP, though, like you have.

staebler: So, yeah, exactly what you wrote.

patrickdillon: ok so I will relax the current code to allow bootstrap ip without control plane

patrickdillon:
```
time="2022-02-03T20:20:16Z" level=debug msg="Checking vmID: 4cf09038-a60d-47f2-8418-c508886a0f50"
time="2022-02-03T20:20:17Z" level=debug msg="Error when obtaining host IP addresses, continuing with best effort. Error message: could not find IP address for control-plane instance 4cf09038-a60d-47f2-8418-c508886a0f50: no reported devices found when determining IP address"
time="2022-02-03T20:20:17Z" level=debug msg="Extracting oVirt Host Addresses to Gather Logs..."
time="2022-02-03T20:20:17Z" level=debug msg="Reading VM IDs from Terraform Outputs..."
time="2022-02-03T20:20:18Z" level=debug msg="connection error: dial tcp 192.168.220.134:22: i/o timeout"
time="2022-02-03T20:20:18Z" level=debug msg="connection error: dial tcp [fe80::ccd6:5013:afb9:9c7b]:22: connect: invalid argument"
time="2022-02-03T20:20:18Z" level=debug msg="Error when obtaining host IP addresses, continuing with best effort. Error message: could not find IP address for bootstrap instance \"c18303b0-0a15-4baf-9e29-c324fd12ae48\": no usable IP addresses found"
time="2022-02-03T20:20:18Z" level=error msg="Attempted to gather debug logs after installation failure: failed to obtain IP address for bootstrap host"
```

Not sure what to make of the connection errors.

patrickdillon:
oh nevermind. I see them right in the code :/

janosdebugs:
@patrickdillon FYI this will not fix the gathering because the CI system is running behind a jump host which doesn't proxy through SSH.

PR #5594, opened 2022-01-31
Bug 2056460: Preallocated disks for OCP nodes on oVirt

janosdebugs:
This change implements [RFE 2035334](https://bugzilla.redhat.com/show_bug.cgi?id=2035334), which enables provisioning OCP nodes with preallocated disks.

---

- Depends on openshift/cluster-api-provider-ovirt#129
- Depends on ovirt/terraform-provider-ovirt#313


data/data/ovirt/cluster/masters/variables.tf
```diff
@@ -57,6 +57,7 @@ variable "ovirt_master_instance_type_id" {
   description = "master VM instance type ID"
 }
 
+<<<<<<< HEAD:data/data/ovirt/cluster/masters/variables.tf
```
mykaul, line 60: ?

pkg/types/ovirt/machinepool.go
```diff
@@ -39,6 +39,10 @@ type MachinePool struct {
 	// Hugepages is the size of a VM's hugepages to use in KiBs.
 	// +optional
 	Hugepages Hugepages `json:"hugepages,omitempty"`
+
+	// PreallocatedDisks sets the disks to be prepovisioned.
```
mykaul, line 43:
```suggestion
	// PreallocatedDisks sets the disks to be preprovisioned.
```

But I think it sets them to be pre-allocated, not pre-provisioned.

janosdebugs:
@michalskrivanek @mykaul @ahadas please provide input for the default values for the following flags:

- **clone** clones the VM instead of linking it to the template. (was: false)
- **sparse** creates a thin-provisioned disks (was: true)
- **format** influences the disk format. Raw doesn't support backups. (was: cow)

We need these defaults for both control plane nodes, as well as worker nodes. These defaults will be applied only to newly created install-configs, old configs will be imported keeping their old values.

ahadas:
> @michalskrivanek @mykaul @ahadas please provide input for the default values for the following flags:
> 
> * **clone** clones the VM instead of linking it to the template. (was: `false`; my suggestion: `true`)

+1
This is the default setting for 'server'/'high-performance' VMs in RHV

> * **sparse** creates a thin-provisioned disks (was: `true`; my suggestion: `false`)

+1
Because of the issue that was reported recently with the disk extension

> * **format** influences the disk format. Raw doesn't support backups. (was: `cow`; my suggestion: `cow` to make backups work)

Makes sense but I don't know if it's the right thing to say when it comes to worker nodes - don't we expect users to back up their workloads with Openshift tools?

janosdebugs:
@ahadas for the worker nodes you don't need backups, unless you are using the local storage operator (which I would personally not recommend under any circumstances). For control plane nodes you should back up your etcd, but since we don't have a good description on how to recover a RHV IPI control plane node without full backups, I would currently still recommend cow for general purpose and customers can use raw as a customization.

michalskrivanek:
clone - true by default (implemented via honoring the VM type, not just "true/false")
format - raw - to quote https://github.com/oVirt/terraform-provider-ovirt/pull/313 - "For disk format it's a bit ore questionable. I think it still makes sense to overcommit in many cases, and it's a question whether that is going to cause any problems once we improve the thin provisioning thresholds in bug https://bugzilla.redhat.com/show_bug.cgi?id=2051997. Still or performance reasons it may make sense to use "raw". I'd lean towards making it a default, but it needs to be configurable."
sparse - not sure yet:) Sparse with RAW still does preallocation AFAIK (but please check it's not returned as error)

ahadas:
> Sparse with RAW still does preallocation AFAIK (but please check it's not returned as error)

on file storage? I don't think so

michalskrivanek:
> > Sparse with RAW still does preallocation AFAIK (but please check it's not returned as error)
> 
> On file storage? I don't think so

well, on "file" the point is it doesn't really matter. overall "sparseness" doesn't really matter, what people are after is thin/preallocation in the logical sense. Whether you have sparse support on NFS 4.2 or not on <4.2....sure, actual size differs but the performance impact is negligent as the physical sparseness is "cheap".

ahadas:
> > > Sparse with RAW still does preallocation AFAIK (but please check it's not returned as error)
> > 
> > 
> > On file storage? I don't think so
> 
> well, on "file" the point is it doesn't really matter. overall "sparseness" doesn't really matter, what people are after is thin/preallocation in the logical sense. Whether you have sparse support on NFS 4.2 or not on <4.2....sure, actual size differs but the performance impact is negligent as the physical sparseness is "cheap".

So I don't understand you then..
RAW+SPARSE is [not supported on block storage](https://github.com/oVirt/ovirt-engine/blob/ovirt-engine-4.4.10.6/backend/manager/modules/bll/src/main/java/org/ovirt/engine/core/bll/storage/disk/image/ImagesHandler.java#L533). If you don't care about sparseness on file storage, then just set RAW+PREALLOCATED

eslutsky (commented):
does it make sense to add additional field validation for the MasterFormat   here :
https://github.com/openshift/installer/blob/3c4a5fb3d859cc150a2dd936517501ba07d5e825/pkg/types/ovirt/validation/platform.go#L13

janosdebugs:
> does it make sense to add additional field validation for the MasterFormat here :
> 
> https://github.com/openshift/installer/blob/3c4a5fb3d859cc150a2dd936517501ba07d5e825/pkg/types/ovirt/validation/platform.go#L13

Addressed in `ValidateMachinePool` now.

eslutsky (approved):

janosdebugs:
There is still a dependency bug I need to resolve. :(

patrickdillon:
The code looks fine, but please fix the commit structure:

1. Squash commits to where you're not introducing a change in one commit then changing in another within the same pr
2.  Commit messages according to https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#commit-message-format

janosdebugs:
Thanks @patrickdillon I've squashed and retitled the changes.

patrickdillon:
/approve

Can you get an ovirt reviewer to lgtm?

PR #5595, opened 2022-02-01
Bug 2048222: Remove non-public AWS regions from list of regions

rna-afk:
When creating the install-config, the installer displays regions
of all partitions of AWS. Certain regions also need extra information
for the validation to work and should not be taken as input since we
only ask for the bare minimum amount of information to create the
install config.

The best approach here would be to only display all the public regions
of AWS and allow for other regions after the install-config is created
to allow for the user to add the extra information.

### DIFF
--- a/pkg/asset/installconfig/aws/regions.go
+++ b/pkg/asset/installconfig/aws/regions.go
@@ -14,12 +14,9 @@ func knownRegions(architecture types.Architecture) map[string]string {
 	required := rhcos.AMIRegions(architecture)
 
 	regions := make(map[string]string)
-	for _, partition := range endpoints.DefaultPartitions() {
-		for _, partitionRegion := range partition.Regions() {
-			partitionRegion := partitionRegion
-			if required.Has(partitionRegion.ID()) {
-				regions[partitionRegion.ID()] = partitionRegion.Description()
-			}
+	for _, region := range endpoints.AwsPartition().Regions() {
+		if required.Has(region.ID()) {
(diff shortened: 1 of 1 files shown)


pkg/asset/installconfig/aws/regions.go
```diff
@@ -14,12 +14,10 @@ func knownRegions(architecture types.Architecture) map[string]string {
...
-	for _, partition := range endpoints.DefaultPartitions() {
-		for _, partitionRegion := range partition.Regions() {
-			partitionRegion := partitionRegion
-			if required.Has(partitionRegion.ID()) {
-				regions[partitionRegion.ID()] = partitionRegion.Description()
-			}
+	for _, partitionRegion := range endpoints.AwsPartition().Regions() {
+		partitionRegion := partitionRegion
+		if required.Has(partitionRegion.ID()) {
+			regions[partitionRegion.ID()] = partitionRegion.Description()
```
staebler, line 20: nit: Change `partitionRegion` to `region`. The partition part of the name is not informative.
Also, I have no idea what the intention of the following line was.
```
		partitionRegion := partitionRegion
```
```suggestion
	for _, region := range endpoints.AwsPartition().Regions() {
		if required.Has(region.ID()) {
			regions[region.ID()] = region.Description()
```

staebler (commented):
Looks good save for a naming nit.

staebler (commented):
/lgtm
/approve

PR #5596, opened 2022-02-02
Bug 1926975: Add cloudProviderConfigDataKey to cloud-provider-config map in AWS C2S

dmc5179:
resolve BZ 1926975

I have tested this fix in C2S and it resolves the issue. The newline looks strange but is needed to generate properly formatted yaml. It is similar to the newlines here: pkg/asset/manifests/openstack/cloudproviderconfig.go

### DIFF
--- a/pkg/asset/manifests/cloudproviderconfig.go
+++ b/pkg/asset/manifests/cloudproviderconfig.go
@@ -99,7 +99,10 @@ func (cpc *CloudProviderConfig) Generate(dependencies asset.Parents) error {
 			return nil
 		}
 		cm.Data[cloudProviderConfigCABundleDataKey] = trustBundle
-                // Add DataKey to address BZ 1926975. The newline is required or the yaml is invalid
+                // Include a non-empty kube config to appease components--such as the kube-apiserver--that
+                // expect there to be a kube config if the cloud-provider-config ConfigMap exists. See
+                // https://bugzilla.redhat.com/show_bug.cgi?id=1926975.
+                // Note that the newline is required in order to be valid yaml.
                 cm.Data[cloudProviderConfigDataKey] = `[Global]
 `
 	case alibabacloudtypes.Name:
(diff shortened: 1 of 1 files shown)


pkg/asset/manifests/cloudproviderconfig.go
```diff
@@ -99,6 +99,9 @@ func (cpc *CloudProviderConfig) Generate(dependencies asset.Parents) error {
 			return nil
 		}
 		cm.Data[cloudProviderConfigCABundleDataKey] = trustBundle
+                // Add DataKey to address BZ 1926975. The newline is required or the yaml is invalid
```
staebler, line 102: Please give the details of why this is needed in the comment as opposed to referencing the BZ and requiring the reader to have to context switch to looking up the BZ.
```suggestion
                // Include a non-empty kube config to appease components--such as the kube-apiserver--that
                // expect there to be a kube config if the cloud-provider-config ConfigMap exists. See
                // https://bugzilla.redhat.com/show_bug.cgi?id=1926975.
                // Note that the newline is required in order to be valid yaml.
```

staebler (commented):
I would prefer that the kube-apiserver were resilient to an absent kube config. But the solution here is demonstrably better than requiring the user to manually update the manifest.
/approve

staebler (commented):
/lgtm

pkg/asset/manifests/cloudproviderconfig.go
```diff
@@ -99,6 +99,12 @@ func (cpc *CloudProviderConfig) Generate(dependencies asset.Parents) error {
 			return nil
 		}
 		cm.Data[cloudProviderConfigCABundleDataKey] = trustBundle
+                // Include a non-empty kube config to appease components--such as the kube-apiserver--that
+                // expect there to be a kube config if the cloud-provider-config ConfigMap exists. See
+                // https://bugzilla.redhat.com/show_bug.cgi?id=1926975.
+                // Note that the newline is required in order to be valid yaml.
+                cm.Data[cloudProviderConfigDataKey] = `[Global]
```
staebler, line 106: This needs to use tabs rather than spaces.
```suggestion
		// Include a non-empty kube config to appease components--such as the kube-apiserver--that
		// expect there to be a kube config if the cloud-provider-config ConfigMap exists. See
		// https://bugzilla.redhat.com/show_bug.cgi?id=1926975.
		// Note that the newline is required in order to be valid yaml.
		cm.Data[cloudProviderConfigDataKey] = `[Global]
```

staebler (commented):
/lgtm cancel

PR #5597, opened 2022-02-02
aws: support the entirety of the secret partitions

staebler:
When C2S support was added, there was only the one region, us-iso-east-1, in the C2S partition. The us-iso-west-1 region is being added. The installer should be able to support new regions in the partition without requiring code changes going forward, just like it does for the public partition.

Additionally, the installer should be able to support the SC2S partition, as it has all of the same conditional behavior as the C2S partition does.


staebler:
https://issues.redhat.com/browse/CORS-1900

kirankt (commented):
/lgtm

patrickdillon:
/retitle Bug 2050940: aws: support the entirety of the secret partition

PR #5569, opened 2022-01-24
Bug 2045927: Add proxy for ironic-agent.service

zhouhao3:
Avoid the issue that ironic agent image cannot be downloaded due to network proxy.

See #5552 for details.

Signed-off-by: Zhou Hao <zhouhao@fujitsu.com>

PR #5598, opened 2022-02-02
baremetal: remove redundant proxy setting

andfasano:
During the bootstrap, proxy related variables should be already injected in the service by https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/etc/systemd/system.conf.d/10-default-env.conf.template.
In addition, proxy environment variable should be already passed to the container by podman

### DIFF
--- a/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
+++ b/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
@@ -184,18 +184,6 @@ else
     IRONIC_HOST="${IRONIC_IP}"
 fi
 
-{{if .Proxy -}}
-{{if .Proxy.HTTPProxy -}}
-HTTP_PROXY="{{.Proxy.HTTPProxy}}"
-{{end -}}
-{{if .Proxy.HTTPSProxy -}}
-HTTPS_PROXY="{{.Proxy.HTTPSProxy}}"
-{{end -}}
-{{if .Proxy.NoProxy -}}
-NO_PROXY="{{.Proxy.NoProxy}}"
-{{end -}}
-{{end -}}
-
 # Create a podman secret for the image-customization-server 
 podman secret rm pull-secret || true
 base64 -w 0 /root/.docker/config.json | podman secret create pull-secret -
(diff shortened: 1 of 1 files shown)


andfasano:
cc @zhouhao3 @rhjanders

ardaguclu:
Good catch!

/approve

elfosardo (approved):

zhouhao3:
lgtm, I'll test it later.

zhouhao3 (approved):

hardys:
I did wonder why we needed the explicit proxy settings in #5569 - thanks @andfasano for pointing out how the global variables are defined.

This lgtm, but we need test confirmation from @zhouhao3 given this effectively reverts #5569

I'm unclear atm why that was needed, https://bugzilla.redhat.com/show_bug.cgi?id=2045927 says this change in behavior was related to the day-1 networking changes, but I'm unclear why the icc started via startironic.sh won't just use the global environment variables mentioned?

@andfasano if we merge https://github.com/openshift-metal3/dev-scripts/pull/1341 ASAP can we get a new CI job to prove this definitely works (and stays working)?

andfasano:
> I'm unclear atm why that was needed, https://bugzilla.redhat.com/show_bug.cgi?id=2045927 says this change in behavior was related to the day-1 networking changes, but I'm unclear why the icc started via startironic.sh won't just use the global environment variables mentioned?

I think the main problem was captured in https://github.com/openshift/image-customization-controller/pull/33, since ICC didn't use the proxy vars for the `ironic-agent.service`
(note that there also an improvement by @zaneb  on that, see https://github.com/openshift/image-customization-controller/pull/39)

> 
> @andfasano if we merge [openshift-metal3/dev-scripts#1341](https://github.com/openshift-metal3/dev-scripts/pull/1341) ASAP can we get a new CI job to prove this definitely works (and stays working)?

Yes, I've got a pending PR ready for that (https://github.com/openshift/release/pull/25912) - and planned also to open a new one for the installer repo as well

hardys:
> I think the main problem was captured in https://github.com/openshift/image-customization-controller/pull/33, since ICC didn't use the proxy vars for the ironic-agent.service

Ah Ok so we only actually needed the icc fix - @zhouhao3 please can you confirm if you're happy with this change based on your testing, thanks!

zaneb:
This looks correct, podman passes these vars automagically: https://docs.podman.io/en/latest/markdown/podman-run.1.html#http-proxy
/lgtm
/hold pending testing by @zhouhao3

zhouhao3:
> @zhouhao3 please can you confirm if you're happy with this change based on your testing, thanks!
> pending testing by @zhouhao3

Sorry for the late reply. I've tested this PR and the change is correct and podman can correctly pass the proxy into the container.

PR #5600, opened 2022-02-02
Bug 2048451: Use proxy dial to validate endpoints

rna-afk:
Using proxy dial instead of the current net dial for validating
service endpoints in AWS since there could be a proxy setup
when an IPI restricted installation occurs that might make the
the endpoints unreachable without using them.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -84,6 +84,7 @@ require (
 	github.com/zclconf/go-cty v1.8.1
 	golang.org/x/crypto v0.0.0-20210817164053-32db794688a5
 	golang.org/x/lint v0.0.0-20210508222113-6edffad5e616
+	golang.org/x/net v0.0.0-20210825183410-e898025ed96a
 	golang.org/x/oauth2 v0.0.0-20210819190943-2bc19b11175f
 	golang.org/x/sys v0.0.0-20211029165221-6e7872819dc8
 	google.golang.org/api v0.44.0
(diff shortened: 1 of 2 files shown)


pkg/asset/installconfig/aws/validation.go
```diff
@@ -317,6 +319,8 @@ func validateRegion(region string) error {
 func validateEndpointAccessibility(endpointURL string) error {
 	// For each provided service endpoint, verify we can resolve and connect with net.Dial.
 	// Ignore e2e.local from unit tests.
+	ctx, cancel := context.WithDeadline(context.TODO(), time.Now().Add(1*time.Minute))
```
staebler, line 322: Can you explain the 1-minute timeout chosen? By default, the OS will impose a timeout on the dial (typically 3 minutes).

rna-afk: I didn't know the OS did that. I'll change it to 3 minutes.

staebler: My suggestion is that you omit the timeout entirely and leave it up to the OS defaults.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -328,7 +330,7 @@ func validateEndpointAccessibility(endpointURL string) error {
 	if port == "" {
 		port = "https"
 	}
-	conn, err := net.Dial("tcp", net.JoinHostPort(URL.Hostname(), port))
+	conn, err := proxy.Dial(ctx, "tcp", net.JoinHostPort(URL.Hostname(), port))
```
staebler, line 333: I don't see much reason to use a `ctx` variable.
```suggestion
	conn, err := proxy.Dial(context.Background(), "tcp", net.JoinHostPort(URL.Hostname(), port))
```

staebler:
You need to update the vendor, too, to get the golang.org/x/net/proxy module.

go.mod
```diff
@@ -328,7 +328,7 @@ require (
 	go.mongodb.org/mongo-driver v1.5.1 // indirect
 	go.opencensus.io v0.22.5 // indirect
 	golang.org/x/mod v0.4.2 // indirect
-	golang.org/x/net v0.0.0-20210825183410-e898025ed96a // indirect
+	golang.org/x/net v0.0.0-20210825183410-e898025ed96a
```
staebler, line 331: Move this up to the first `require` block, since it is a direct requirement.

PR #5601, opened 2022-02-03
Bug 2050146: Don't shortcut OpenStack scraping if quota is unavailable

mdbooth:
This results in an incorrect failure to validate network capabilities
because network extensions weren't loaded.


mandre (commented):
/lgtm
/approve
/hold for CI

PR #4112, opened 2020-08-28
data/manifests/bootkube/cvo-overrides: Drop the explicit upstream

wking:
Not bug-worthy, but floating now for 4.7 to centralize discussion.

The cluster-version operator has been falling back to a default URI when the ClusterVersion upstream is empty since [way][1] [back][2], and this behavior is [enshrined in the API][3].  Drop the installer-side
opinion, so:

* There is a single location where we version-control the default upstream (the CVO).
* Folks consuming in-cluster ClusterVersion objects have one less property to distract them (unless they want to override the default, in which case it's not distracting).

Related to openshift/openshift-docs#22252.

CC @vrutkovs , since I'm not sure how this will square with FCOS and a71f4245a.  FCOS options:

1. Continue to override the CVO default with explicit ClusterVersion.  No change from today, but breaks [nominal API][3] of:

    > By default it will use the appropriate update server for the cluster"

2. Teach the CVO to recognize FCOS vs. OCP and use the right default for each.
3. Fork the CVO and add an fcos branch, as you currently do for the installer.

Thoughts?

[1]: https://github.com/openshift/cluster-version-operator/blame/2c4931dc283c551938be1a00fee290de0b79d99c/pkg/cvo/availableupdates.go#L27-L31
[2]: https://github.com/openshift/cluster-version-operator/commit/ab4d84a021c38fea5e5973287e1f5580928907c1#diff-78f2af341fa49292dd6930f378018867R24
[3]: https://github.com/openshift/api/blame/0422dc17083e9e8df18d029f3f34322e96e9c326/config/v1/types_cluster_version.go#L56-L57

PR #5605, opened 2022-02-03
Add user selectable capabilities in installconfig

kirankt:
This is a partial implementation of [user selectable install solutions](https://github.com/openshift/enhancements/blob/master/enhancements/installer/component-selection.md)

This PR only adds install-config configuration to manage capabilities. 

Edit: New and additional details improving on the enhancement can be found [here](https://github.com/openshift/api/pull/1106) and is incorporated in this PR


pkg/types/installconfig.go
```diff
@@ -383,3 +386,33 @@ func (c *InstallConfig) WorkerMachinePool() *MachinePool {
...
+
+// Capabilities defines the configuration for user to install/exclude optional OpenShift components
+type Capabilities struct {
+	// InclusionDefault declares the default policy for optional, core cluster components
+	// +optional
+	InclusionDefault ClusterVersionCapabilitiesDefault `json:"inclusionDefault,omitempty"`
+
+	// Include declares capabilities that should be included, regardless of inclusionDefault
+	// +optional
+	Include []string `json:"include,omitempty"`
```
staebler, line 413: Do we want to use a type for the capability names?
```suggestion
	Include []Capability `json:"include,omitempty"`
```

pkg/types/installconfig.go
```diff
@@ -154,6 +154,9 @@ type InstallConfig struct {
 	// BootstrapInPlace is the configuration for installing a single node
 	// with bootstrap in place installation.
 	BootstrapInPlace *BootstrapInPlace `json:"bootstrapInPlace,omitempty"`
+
+	// Capabilities configures the installation of optional, core cluster components
+	Capabilities *Capabilities `json:"capabilities,omitempty"`
```
staebler, line 159: Should this be optional? Are we going to omit the capabilities configuration from the ClusterVersion resource when this is omitted? I would expect that we would still set a default value in the ClusterVersion resource.

kirankt: Yes. This should be optional. I've added the macro comment.

staebler: OK. Optional does make sense. But I do think we want to populate with the default, even when the entire `Capabilities` has been omitted. I am also leaning towards **if** the user provided any values for `Include` or `Exclude`, then they are also required to make an explicit decision about `InclusionDefault` (despite it being redundant with whether they chose `Include` or `Exclude`).

kirankt: The design has changed since the enhancement and I've updated the PR with the new ideas. PTAL

pkg/types/defaults/installconfig.go
```diff
@@ -82,6 +84,10 @@ func SetInstallConfigDefaults(c *types.InstallConfig) {
 		}
 	}
 
+	if c.Capabilities != nil && c.Capabilities.BaselineCapabilitySet == "" {
+		c.Capabilities.BaselineCapabilitySet = configv1.ClusterVersionCapabilitySetCurrent
```
wking, line 88: This puts the installer on the hook for updating as the API-declared default changes.  Doesn't seem like a high risk, but I'm also not clear on the benefit vs. just leaving the value unset so the CVO can default it.  For precedent, see #4112 where I removed the need for the installer to care about a property that the CVO can default on its own.

pkg/types/installconfig.go
```diff
@@ -383,3 +389,20 @@ func (c *InstallConfig) WorkerMachinePool() *MachinePool {
 
 	return nil
 }
+
+// Capabilities defines the configuration for user to install/exclude optional OpenShift components
+type Capabilities struct {
```
wking, line 394: Is it useful to declare your own type?  I'd have expected you to use `*configv1.ClusterVersionCapabilitiesSpec` as the type for the new `InstallConfig` property.

pkg/types/defaults/installconfig.go
kirankt: +1. That makes sense. I'll remove this here.

pkg/types/installconfig.go
kirankt: +1. Will remove. I was working myself backwards into reusing all the defs from configv1 but missed this one.

wking:
Looks good to me :).  I'll let the installer maintainers take a look (and I've filed #5644 to drop myself from the approver set).

pkg/types/installconfig.go
staebler: The reason that we want to use our own type is because it allows us to control changes to meet our forward-compatibility guarantees.

wking: Won't openshift/api have those same forward-compat commitments?

kirankt:
Closing this PR in favor of #5645

PR #5292, opened 2021-10-12
Power VS IPI Support: types only

clnperez:
Depends on https://github.com/openshift/installer/pull/5270

Signed-off-by: Christy Norman <christy@linux.vnet.ibm.com>

PR #5609, opened 2022-02-04
Add IBM Power VS: types

clnperez:

Older discussions on some of the code here can be found in
https://github.com/openshift/installer/pull/5224.
An older unreviewed version of this PR can be found here:
https://github.com/openshift/installer/pull/5292

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -469,6 +469,36 @@ spec:
                           - high_performance
                           type: string
                       type: object
+                    powervs:
+                      description: PowerVS is the configuration used when installing
+                        on IBM Power VS.
+                      properties:
+                        memory:
+                          description: Memory defines the memory in GB for the instance.
+                          type: string
+                        procType:
+                          description: ProcType defines the processor sharing model
+                            for the instance. Must be one of {capped, dedicated, shared}.
+                          enum:
+                          - ""
+                          - capped
+                          - dedicated
+                          - shared
+                          type: string
+                        processors:
+                          description: Processors defines the processing units for
+                            the instance.
+                          type: string
+                        sysType:
+                          description: SysType defines the system type for instance.
+                          type: string
+                        volumeIDs:
+                          description: VolumeIDs is the list of volumes attached to
+                            the instance.
+                          items:
+                            type: string
+                          type: array
+                      type: object
                     vsphere:
                       description: VSphere is the configuration used when installing
                         on vSphere.
@@ -929,6 +959,36 @@ spec:
                         - high_performance
                         type: string
                     type: object
+                  powervs:
+                    description: PowerVS is the configuration used when installing
+                      on IBM Power VS.
+                    properties:
+                      memory:
+                        description: Memory defines the memory in GB for the instance.
+                        type: string
+                      procType:
+                        description: ProcType defines the processor sharing model
+                          for the instance. Must be one of {capped, dedicated, shared}.
+                        enum:
+                        - ""
+                        - capped
+                        - dedicated
+                        - shared
+                        type: string
+                      processors:
+                        description: Processors defines the processing units for the
+                          instance.
+                        type: string
+                      sysType:
+                        description: SysType defines the system type for instance.
+                        type: string
+                      volumeIDs:
+                        description: VolumeIDs is the list of volumes attached to
+                          the instance.
+                        items:
+                          type: string
+                        type: array
+                    type: object
                   vsphere:
                     description: VSphere is the configuration used when installing
                       on vSphere.
@@ -981,7 +1041,7 @@ spec:
               values. For all other platforms, the field must not be set. AWS: \"Mint\",
               \"Passthrough\", \"Manual\" Azure: \"Mint\", \"Passthrough\", \"Manual\"
               AzureStack: \"Manual\" GCP: \"Mint\", \"Passthrough\", \"Manual\" IBMCloud:
-              \"Manual\" AlibabaCloud: \"Manual\""
+              \"Manual\" AlibabaCloud: \"Manual\" PowerVS: \"Manual\""
             enum:
             - ""
             - Mint
@@ -2158,6 +2218,92 @@ spec:
                 - ingress_vip
                 - ovirt_cluster_id
                 - ovirt_storage_domain_id
+                type: object
+              powervs:
+                description: PowerVS is the configuration used when installing on
+                  Power VS.
+                properties:
+                  clusterOSImage:
+                    description: ClusterOSImage is a pre-created Power VS boot image
+                      that overrides the default image for cluster nodes.
+                    type: string
+                  defaultMachinePlatform:
+                    description: DefaultMachinePlatform is the default configuration
+                      used when installing on Power VS for machine pools which do
+                      not define their own platform configuration.
+                    properties:
+                      memory:
+                        description: Memory defines the memory in GB for the instance.
+                        type: string
+                      procType:
+                        description: ProcType defines the processor sharing model
+                          for the instance. Must be one of {capped, dedicated, shared}.
+                        enum:
+                        - ""
+                        - capped
+                        - dedicated
+                        - shared
+                        type: string
+                      processors:
+                        description: Processors defines the processing units for the
+                          instance.
+                        type: string
+                      sysType:
+                        description: SysType defines the system type for instance.
+                        type: string
+                      volumeIDs:
+                        description: VolumeIDs is the list of volumes attached to
+                          the instance.
+                        items:
+                          type: string
+                        type: array
+                    type: object
+                  powervsResourceGroup:
+                    description: PowerVSResourceGroup is the resource group in which
+                      Power VS resources will be created.
+                    type: string
+                  pvsNetworkName:
+                    description: PVSNetworkName specifies an existing network within
+                      the Power VS Service Instance.
+                    type: string
+                  region:
+                    description: Region specifies the IBM Cloud colo region where
+                      the cluster will be created.
+                    type: string
+                  serviceInstanceID:
+                    description: ServiceInstanceID is the ID of the Power IAAS instance
+                      created from the IBM Cloud Catalog
+                    type: string
+                  subnets:
+                    description: Subnets specifies existing subnets (by ID) where
+                      cluster resources will be created.  Leave unset to have the
+                      installer create subnets in a new VPC on your behalf.
+                    items:
+                      type: string
+                    type: array
+                  userID:
+                    description: UserID is the login for the user's IBM Cloud account.
+                    type: string
+                  vpc:
+                    description: VPC is a VPC inside IBM Cloud. Needed in order to
+                      create VPC Load Balancers.
+                    type: string
+                  vpcRegion:
+                    description: VPCRegion specifies the IBM Cloud region in which
+                      to create VPC resources. Leave unset to allow installer to select
+                      the closest VPC region.
+                    type: string
+                  zone:
+                    description: Zone specifies the IBM Cloud colo region where the
+                      cluster will be created. At this time, only single-zone clusters
+                      are supported.
+                    type: string
+                required:
+                - powervsResourceGroup
+                - region
+                - serviceInstanceID
+                - userID
+                - zone
                 type: object
               vsphere:
                 description: VSphere is the configuration used when installing on
--- a/pkg/types/powervs/machinepools.go
+++ b/pkg/types/powervs/machinepools.go
@@ -1,0 +1,65 @@
+package powervs
+
+// ProcType defines valid types for a ppc64le processor in Power VS
+// +kubebuilder:validation:Enum="";capped;dedicated;shared
+type ProcType string
+
+// Capped type for capped processor consumption
+const Capped ProcType = "capped"
+
+// Dedicated type for dedicated processor(s)
+const Dedicated ProcType = "dedicated"
+
+// Shared type shared type for shared processor(s)
+const Shared ProcType = "shared"
+
+// MachinePool stores the configuration for a machine pool installed on IBM Power VS.
+type MachinePool struct {
+	// VolumeIDs is the list of volumes attached to the instance.
+	//
+	// +optional
+	VolumeIDs []string `json:"volumeIDs,omitempty"`
+
+	// Memory defines the memory in GB for the instance.
+	//
+	// +optional
+	Memory string `json:"memory,omitempty"`
+
+	// Processors defines the processing units for the instance.
+	//
+	// +optional
+	Processors string `json:"processors,omitempty"`
+
+	// ProcType defines the processor sharing model for the instance.
+	// Must be one of {capped, dedicated, shared}.
+	//
+	// +optional
+	ProcType ProcType `json:"procType,omitempty"`
+
+	// SysType defines the system type for instance.
+	//
+	// +optional
+	SysType string `json:"sysType,omitempty"`
+}
+
+// Set stores values from required into a
+func (a *MachinePool) Set(required *MachinePool) {
+	if required == nil || a == nil {
+		return
+	}
+	if len(required.VolumeIDs) != 0 {
+		a.VolumeIDs = required.VolumeIDs
+	}
+	if required.Memory != "" {
+		a.Memory = required.Memory
+	}
+	if required.Processors != "" {
+		a.Processors = required.Processors
+	}
+	if required.ProcType != "" {
+		a.ProcType = required.ProcType
+	}
+	if required.SysType != "" {
+		a.SysType = required.SysType
+	}
+}
--- a/pkg/types/powervs/platform.go
+++ b/pkg/types/powervs/platform.go
@@ -1,0 +1,57 @@
+package powervs
+
+// Platform stores all the global configuration that all machinesets
+// use.
+type Platform struct {
+
+	// ServiceInstanceID is the ID of the Power IAAS instance created from the IBM Cloud Catalog
+	ServiceInstanceID string `json:"serviceInstanceID"`
+
+	// PowerVSResourceGroup is the resource group in which Power VS resources will be created.
+	PowerVSResourceGroup string `json:"powervsResourceGroup"`
+
+	// Region specifies the IBM Cloud colo region where the cluster will be created.
+	Region string `json:"region"`
+
+	// Zone specifies the IBM Cloud colo region where the cluster will be created.
+	// At this time, only single-zone clusters are supported.
+	Zone string `json:"zone"`
+
+	// VPCRegion specifies the IBM Cloud region in which to create VPC resources.
+	// Leave unset to allow installer to select the closest VPC region.
+	//
+	// +optional
+	VPCRegion string `json:"vpcRegion,omitempty"`
+
+	// UserID is the login for the user's IBM Cloud account.
+	UserID string `json:"userID"`
+
+	// VPC is a VPC inside IBM Cloud. Needed in order to create VPC Load Balancers.
+	//
+	// +optional
+	VPC string `json:"vpc,omitempty"`
+
+	// Subnets specifies existing subnets (by ID) where cluster
+	// resources will be created.  Leave unset to have the installer
+	// create subnets in a new VPC on your behalf.
+	//
+	// +optional
+	Subnets []string `json:"subnets,omitempty"`
+
+	// PVSNetworkName specifies an existing network within the Power VS Service Instance.
+	//
+	// +optional
+	PVSNetworkName string `json:"pvsNetworkName,omitempty"`
+
+	// ClusterOSImage is a pre-created Power VS boot image that overrides the
+	// default image for cluster nodes.
+	//
+	// +optional
+	ClusterOSImage string `json:"clusterOSImage,omitempty"`
+
+	// DefaultMachinePlatform is the default configuration used when
+	// installing on Power VS for machine pools which do not define their own
+	// platform configuration.
+	// +optional
+	DefaultMachinePlatform *MachinePool `json:"defaultMachinePlatform,omitempty"`
+}
--- a/pkg/types/powervs/powervs_regions.go
+++ b/pkg/types/powervs/powervs_regions.go
@@ -1,0 +1,67 @@
+package powervs
+
+// Since there is no API to query these, we have to hard-code them here.
+
+// Region describes resources associated with a region in Power VS.
+// We're using a few items from the IBM Cloud VPC offering. The region names
+// for VPC are different so another function of this is to correlate those.
+type Region struct {
+	Description string
+	VPCRegion   string
+	Zones       []string
+}
+
+// Regions holds the regions for IBM Power VS, and descriptions used during the survey
+var Regions = map[string]Region{
+	"dal": {
+		Description: "Dallas, USA",
+		VPCRegion:   "us-south",
+		Zones:       []string{"dal12"},
+	},
+	"eu-de": {
+		Description: "Frankfurt, Germany",
+		VPCRegion:   "eu-de",
+		Zones: []string{
+			"eu-de-1",
+			"eu-de-2",
+		},
+	},
+	"lon": {
+		Description: "London, UK.",
+		VPCRegion:   "eu-gb",
+		Zones: []string{
+			"lon04",
+			"lon06",
+		},
+	},
+	"osa": {
+		Description: "Osaka, Japan",
+		VPCRegion:   "jp-osa",
+		Zones:       []string{"osa21"},
+	},
+	"syd": {
+		Description: "Sydney, Australia",
+		VPCRegion:   "au-syd",
+		Zones:       []string{"syd04"},
+	},
+	"sao": {
+		Description: "São Paulo, Brazil",
+		VPCRegion:   "br-sao",
+		Zones:       []string{"sao01"},
+	},
+	"tor": {
+		Description: "Toronto, Canada",
+		VPCRegion:   "ca-tor",
+		Zones:       []string{"tor01"},
+	},
+	"tok": {
+		Description: "Tokyo, Japan",
+		VPCRegion:   "jp-tok",
+		Zones:       []string{"tok04"},
+	},
+	"us-east": {
+		Description: "Washington DC, USA",
+		VPCRegion:   "us-east",
+		Zones:       []string{"us-east"},
+	},
+}
--- a/pkg/types/powervs/validation/machinepool.go
+++ b/pkg/types/powervs/validation/machinepool.go
@@ -1,0 +1,69 @@
+package validation
+
+import (
+	"math"
+	"regexp"
+	"strconv"
+
+	"github.com/google/uuid"
+	"github.com/openshift/installer/pkg/types/powervs"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// ValidateMachinePool checks that the specified machine pool is valid.
+func ValidateMachinePool(p *powervs.MachinePool, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	// Validate VolumeIDs
+	for i, volumeID := range p.VolumeIDs {
+		_, err := uuid.Parse(volumeID)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("volumeIDs").Index(i), volumeID, "volume ID must be a valid UUID"))
+		}
+	}
+
+	// Validate Memory
+	if p.Memory != "" {
+		memory, err := strconv.ParseInt(p.Memory, 10, 64)
+		if err == nil {
+			if memory < 2 || memory > 64 {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("memory"), p.Memory, "memory must be an integer number of GB that is at least 2 and no more than 64"))
+			}
+		} else {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("memory"), p.Memory, "memory must be an integer number of GB that is at least 2 and no more than 64"))
+		}
+	}
+
+	// Validate Processors
+	if p.Processors != "" {
+		processors, err := strconv.ParseFloat(p.Processors, 64)
+		if err == nil {
+			if processors < 0.25 || processors > 32 {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("processors"), p.Processors, "number of processors must be from .25 to 32 cores"))
+			}
+			if math.Mod(processors*1000, 2) != 0 || math.Mod(processors*100, 25) != 0 {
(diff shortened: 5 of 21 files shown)


clnperez:
@rna-afk can you advise on how we should add the right reviewers and owners to some of these files? Back in October the answer was to add what the checks are complaining about, so sounds like we need to be educated on the new best practices there.

pkg/types/powervs/machinepools.go
```diff
@@ -0,0 +1,51 @@
+package powervs
+
+// MachinePool stores the configuration for a machine pool installed on IBM Power VS.
+type MachinePool struct {
+	// VolumeIDs is the list of volumes attached to the instance.
+	//
+	// +optional
+	VolumeIDs []string `json:"volumeIDs"`
```
staebler, line 8: Use `omitempty` as well for optional fields.
```suggestion
	VolumeIDs []string `json:"volumeIDs,omitempty"`
```

pkg/types/powervs/platform.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Platform stores all the global configuration that all machinesets
+// use.
+// Note: The subsequent mentions of future-TF support refer to work that is
+// undergoing and should be available to test well in time for 4.10 feature-
+// freeze. We do not plan to GA with these as required inputs.
+type Platform struct {
+
+	// ServiceInstanceID is the ID of the Power IAAS instance created from the IBM Cloud Catalog
+	ServiceInstanceID string `json:"serviceInstance"`
```
staebler, line 11:
```suggestion
	ServiceInstanceID string `json:"serviceInstanceID"`
```

pkg/types/powervs/platform.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// PowerVSResourceGroup is the resource group for creating Power VS resources.
+	PowerVSResourceGroup string `json:"powervsResourceGroup"`
+
+	// Region specifies the IBM Cloud region where the cluster will be created.
+	Region string `json:"region"`
+
+	// Zone specifies the IBM Cloud colo region where the cluster will be created.
```
staebler, line 19: Do we not want to spread the cluster across multiple zones in the region, where possible?

pkg/types/powervs/platform.go
```diff
@@ -0,0 +1,63 @@
...
+	// to allow installer to randomly select a zone.
+	//
+	// +optional
+	VPCZone string `json:"vpcRegion,omitempty"`
+
+	// UserID is the login for the user's IBM Cloud account.
+	UserID string `json:"userID"`
+
+	// APIKey is the API key for the user's IBM Cloud account.
+	APIKey string `json:"APIKey,omitempty"`
```
staebler, line 33: These should not be in the install-config.yaml. Can these be read from another file instead? Or from the IC_API_KEY environment variable as with the ibmcloud platform?

pkg/types/powervs/platform.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// VPC is a VPC inside IBM Cloud. Needed in order to create VPC Load Balancers.
```
staebler, line 35: The text here implies to me that you can have a successful install without specifying the VPC, in which case your cluster will not use VPC Load Balancers. Is that accurate? Or is the optional aspect of this like it is for other platforms where omitting the VPC tells the installer to create a VPC for you?

pkg/types/powervs/platform.go
```diff
@@ -0,0 +1,63 @@
...
+	VPC string `json:"vpc,omitempty"`
+
+	// Subnets specifies existing subnets (by ID) where cluster
+	// resources will be created.  Leave unset to have the installer
+	// create subnets in a new VPC on your behalf.
+	//
+	// +optional
+	Subnets []string `json:"subnets,omitempty"`
+
+	// PVSNetworkName specifies an existing network within the Power VS Service Instance.
```
staebler, line 47: What is the behavior when omitted?

pkg/types/powervs/powervs_regions.go
```diff
@@ -0,0 +1,86 @@
...
+	"us-east": {
+		Name:        "us-east",
+		Description: "Washington DC, USA",
+		VPCRegion:   "us-east",
+		Zones:       []string{"us-east"},
+	},
+}
+
+// Zones retrieves a slice of all zones in Power VS
+func Zones() []string {
```
staebler, line 80: How do you envision this function being used? What is the use case for wanting to know *all* of the zones across all regions?

pkg/types/powervs/validation/machinepool.go
```diff
@@ -0,0 +1,65 @@
...
+
+// ValidateMachinePool checks that the specified machine pool is valid.
+func ValidateMachinePool(p *powervs.MachinePool, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	// Validate VolumeIDs
+	for i, volumeID := range p.VolumeIDs {
+		_, err := uuid.Parse(volumeID)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("volumeIDs").Index(i), volumeID, "Volume ID provided is not a UUID"))
```
staebler, line 20: Field error messages should start with a lower case letter. Also, the message should tell the user what the field *must* be not what the invalid field is.
```suggestion
			allErrs = append(allErrs, field.Invalid(fldPath.Child("volumeIDs").Index(i), volumeID, "volume ID must be a UUID"))
```

pkg/types/powervs/machinepools.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// Memory defines the memory in GB for the instance.
+	//
+	// +optional
+	Memory string `json:"memory"`
```
staebler, line 13: Why is this a string instead of an int64?

pkg/types/powervs/validation/machinepool.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	// Validqte Processors
+	processors, err := strconv.ParseFloat(p.Processors, 64)
+	if p.Processors != "" {
+		if err == nil {
+			if processors < 0.25 || processors > 32 {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("processors"), p.Processors, "Number of processors must be from .25 to 32 cores"))
+			}
+			if math.Mod(processors, 0.25) != 0 {
```
staebler, line 43: Floating point arithmetic does not allow for this type of precision.

pkg/types/powervs/validation/machinepool.go
```diff
@@ -0,0 +1,65 @@
...
+
+	// Validate Memory
+	if p.Memory != "" {
+		memory, err := strconv.ParseInt(p.Memory, 10, 64)
+		if err == nil {
+			if memory < 2 || memory > 64 {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("memory"), p.Memory, "Memory must be from 2 to 64 GB"))
+			}
+		} else {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("memory"), p.Memory, "Memory must be a valid integer"))
```
staebler, line 32: Be more precise about the message. Do not tell the user that it must be an integer, allow them to put in an integer outside of the prescribed bounds, and then later tell them the bounds.
```suggestion
			allErrs = append(allErrs, field.Invalid(fldPath.Child("memory"), p.Memory, "memory must be an integer number of GB that is at least 2 and no more than 64"))
```

pkg/types/powervs/machinepools.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// Processors defines the processing units for the instance.
+	//
+	// +optional
+	Processors string `json:"processors"`
+
+	// ProcType defines the processor sharing model for the instance.
+	//
+	// +optional
+	ProcType string `json:"procType"`
```
staebler, line 23: Use a type with known values. Detail the accepted values in the description of the field. Add kubebuilder validation on the type.

pkg/types/powervs/validation/machinepool.go
```diff
@@ -0,0 +1,65 @@
...
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("processors"), p.Processors, "Processors must be in increments of .25"))
+			}
+		} else {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("processors"), p.Processors, "Processors must be a valid floating point number"))
+		}
+	}
+	// Validate ProcType
+	if p.ProcType != "" {
+		if p.ProcType != "shared" && p.ProcType != "dedicated" {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("procType"), p.ProcType, "ProcType must be either 'shared' or 'dedicated'"))
```
staebler, line 53: Use `field.NotSupported` instead of `field.Invalid`.

pkg/types/powervs/validation/machinepool.go
```diff
@@ -0,0 +1,65 @@
...
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("memory"), p.Memory, "Memory must be from 2 to 64 GB"))
+			}
+		} else {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("memory"), p.Memory, "Memory must be a valid integer"))
+		}
+	}
+
+	// Validqte Processors
+	processors, err := strconv.ParseFloat(p.Processors, 64)
+	if p.Processors != "" {
```
staebler, line 38:
```suggestion
	if p.Processors != "" {
		processors, err := strconv.ParseFloat(p.Processors, 64)
```

pkg/types/powervs/validation/machinepool.go
```diff
@@ -0,0 +1,65 @@
...
+	// Validate ProcType
+	if p.ProcType != "" {
+		if p.ProcType != "shared" && p.ProcType != "dedicated" {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("procType"), p.ProcType, "ProcType must be either 'shared' or 'dedicated'"))
+		}
+	}
+
+	// Validate SysType
+	if p.SysType != "" {
+		sysTypeRegex := `^(?:e980|s922(-.*|))$`
```
staebler, line 59:
```suggestion
		const sysTypeRegex = `^(?:e980|s922(-.*|))$`
```

pkg/types/powervs/machinepools.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// SysType defines the system type for instance.
+	//
+	// +optional
+	SysType string `json:"sysType"`
```
staebler, line 28: Add in the description what the valid values are.

pkg/types/powervs/validation/platform.go
```diff
@@ -0,0 +1,23 @@
...
+
+// ValidatePlatform checks that the specified platform is valid.
+func ValidatePlatform(p *powervs.Platform, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if p.Region == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("region"), "region must be specified"))
+	}
+
+	/*if p.DefaultMachinePlatform != nil {
```
staebler, line 17: Do not include commented-out code.

pkg/types/powervs/platform.go
```diff
@@ -0,0 +1,63 @@
+package powervs
+
+// Platform stores all the global configuration that all machinesets
+// use.
+// Note: The subsequent mentions of future-TF support refer to work that is
```
staebler, line 5: There are no mentions of future-TF support.

staebler:
> @rna-afk can you advise on how we should add the right reviewers and owners to some of these files? Back in October the answer was to add what the checks are complaining about, so sounds like we need to be educated on the new best practices there.

Add the following to all your installer PRs. Aditya will be the point for doing reviews.
/cc @rna-afk

rna-afk:
For the owners file, [this](https://github.com/openshift/installer/blob/69521ee60a4902743cc7269a32cdaecd813e9642/pkg/types/powervs/OWNERS#L4) looks right but you should add the alias (multiarch-reviewers) in this [file](https://github.com/openshift/installer/blob/master/OWNERS_ALIASES). Also you need to add the former in every powervs folder you create in the installer.

pkg/types/powervs/platform.go
clnperez: it's not possible at this time without setting up additional tunnelling. should i note that here or somewhere else?

pkg/types/powervs/platform.go
clnperez: yah, we can remove that. @cklokman would you mind working on this?

pkg/types/powervs/platform.go
clnperez: ah -- i blanked on explaining that a few things here are temporary. we've had to pause on including the vpc creation, so for now there are manual steps for dev that we need to take to create a vpc and the networks in both ibmcloud and power vs. after that, vpc will be created by the installer. it's not going to be optional when we've got networking a fix into prod.

pkg/types/powervs/platform.go
clnperez: same as above. this is temporarily a dev pre-req but will be auto-created in the future every time

pkg/types/powervs/machinepools.go
clnperez: i can't remember specifically what, but something somewhere complained and i changed it months ago.

pkg/types/powervs/validation/machinepool.go
clnperez: @mjturek can you fix?

pkg/types/powervs/validation/machinepool.go
clnperez: @mjturek fyi

pkg/types/powervs/validation/platform.go
clnperez: oops. thought i'd removed all that.

pkg/types/powervs/powervs_regions.go
clnperez: this is no longer used. removing.

pkg/types/powervs/platform.go
clnperez: n/m @cklokman

pkg/types/powervs/validation/machinepool.go
clnperez: @mjturek just noticing that `capped` isn't in the list of accepted types. was this intentional?

pkg/types/powervs/validation/machinepool.go
clnperez: n/m @mjturek

kirankt:
> For the owners file, [this](https://github.com/openshift/installer/blob/69521ee60a4902743cc7269a32cdaecd813e9642/pkg/types/powervs/OWNERS#L4) looks right but you should add the alias (multiarch-reviewers/approvers) in this [file](https://github.com/openshift/installer/blob/master/OWNERS_ALIASES). Also you need to add the former in almost every powervs folder you create in the installer. I can guide you on which folders.

Question. Why multiarch-{reviewers,approvers} as opposed to powervs-{reviewers,approvers}? Looks like the file follows platform-{reviewers,approvers} format.

rna-afk:
> Question. Why multiarch-{reviewers,approvers} as opposed to powervs-{reviewers,approvers}? Looks like the file follows platform-{reviewers,approvers} format.

Not sure if there's a preference on the naming and well now I think powervs would be better. Thanks!

clnperez:
`multiarch-reviewers` comes from the name of the team I work closely with inside Red Hat. @Prashanth684 is the dev I work the most with. Thoughts, Prashanth?

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -469,6 +469,35 @@ spec:
                           - high_performance
                           type: string
                       type: object
+                    powervs:
+                      description: PowerVS is the configuration used when installing
```
rna-afk, line 473: Changes to this file will cause this test to fail. 
https://github.com/openshift/installer/blob/02ccd27538b080bcc09983f27dcb4aa42d26c5dc/pkg/explain/printer_test.go#L11

rna-afk (commented):
Looks good. Just a few nits/questions

pkg/types/powervs/validation/machinepool.go
```diff
@@ -0,0 +1,67 @@
...
+	// Validate ProcType
+	if p.ProcType != "" {
+		if p.ProcType != "shared" && p.ProcType != "dedicated" && p.ProcType != "capped" {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("procType"), p.ProcType, "procType must be one of: 'capped', 'shared' or 'dedicated'"))
+		}
+	}
+
+	// Validate SysType
+	if p.SysType != "" {
+		const sysTypeRegex = `^(?:e980|s922(-.*|))$`
```
Prashanth684, line 60: are there concrete values here we could list rather than a regex? i mean i would assume there would be invalid values even in this case ?

pkg/types/powervs/validation/machinepool.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	// Validqte Processors
+	if p.Processors != "" {
+		processors, err := strconv.ParseFloat(p.Processors, 64)
+		if err == nil {
+			if processors < 0.25 || processors > 32 {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("processors"), p.Processors, "number of processors must be from .25 to 32 cores"))
+			}
+			if math.Mod(processors*1000, 2) == 0 && math.Mod(processors*100, 25) == 0 {
```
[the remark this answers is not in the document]
clnperez: .5 is valid. maybe that statement needs some tweaking to be more clear? since those both result in zeros and the lhs and rhs are both true: true && true == true 
to be sure, i did make a little test program:
```
func main() {
        processors := []float64{0.0, 0.5, 0.25, 0.6, 1.4, 1.5, 2.0, 4, 5, 32, 16.111, 100.999}
        for _, n := range processors {
                valid := math.Mod(n*1000, 2) == 0 && math.Mod(n*100, 25) == 0
                fmt.Println(fmt.Sprintf("%f: %v", n, valid))
        }
}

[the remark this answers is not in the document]
Comment (rna-afk) [reply pkg/types/powervs/validation/machinepool.go:44]: But then it shouldn't report an error if it's valid?

File: pkg/types/powervs/validation/machinepool.go (hunk 20 above)
Comment (clnperez) [reply pkg/types/powervs/validation/machinepool.go:60]: the dev team creates us special system pools sometimes in staging that will follow the pattern s922-* (so like, s922-ocp). if we don't account for that here we'd have to recompile with a specific system pool for testing in staging. (and it's never with the e980s since those are so large, expensive, and therefore few)

Comment (clnperez) [reply pkg/types/powervs/validation/machinepool.go:60]: oops. also, system type is really just another word for system pool in practicality but we shouldn't really use the two terms interchangeably so apologies if that was confusing

File: pkg/types/powervs/validation/machinepool.go (hunk 21 above)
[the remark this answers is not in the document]
Comment (clnperez) [reply pkg/types/powervs/validation/machinepool.go:44]: ah! now i see your point. will fix. thanks

File: data/data/install.openshift.io_installconfigs.yaml (hunk 19 above)
Comment (clnperez) [reply data/data/install.openshift.io_installconfigs.yaml:473]: not including this file caused the codegen checks to fail. i'll see if i can just include that one tweaked test and this file.

Comment (rna-afk) [reply data/data/install.openshift.io_installconfigs.yaml:473]: Ah sorry I meant you should also update the test with these fields

File: pkg/types/powervs/validation/machinepool.go (hunk 20 above)
Comment (Prashanth684) [reply pkg/types/powervs/validation/machinepool.go:62]: ah i see..so these have to be pre-created before we start the install.

**Comment by clnperez** [comment]
after a quick chat with @Prashanth684 we decided to switch to `powervs-[reviewers|approvers]` vs `multiarch`, and not add the other two IDs he listed. 

hopefully good to go now. i've looked through failing tests on each run (with a special focus on the ibmcloud ones, since we share some things -- though nothing in this PR so just being extra careful) and nothing looks related.

File: pkg/types/powervs/validation/machinepool_test.go [hunk 22]
```diff
@@ -0,0 +1,132 @@
...
+			pool: &powervs.MachinePool{
+				ProcType: "shared",
+			},
+		},
+		{
+			name: "invalid procType",
+			pool: &powervs.MachinePool{
+				ProcType: "none",
+			},
+			expected: `^test-path\.procType: Invalid value: "none": ProcType must be either 'shared' or 'dedicated'$`,
```
mjturek, line 106: At some point `capped` was introduced as a valid processor type but haven't tracked down the change where it happened. Either way capped should be part of the list of valid ProcTypes.

pkg/types/powervs/powervs_regions.go
```diff
@@ -0,0 +1,77 @@
...
+		},
+	},
+	"osa": {
+		Name:        "osa",
+		Description: "Osaka, Japan",
+		VPCRegion:   "jp-osa",
+		Zones:       []string{"osa21"},
+	},
+	"syd": {
+		Name:        "syd",
```
Prashanth684, line 48: the key itself is the name. do you want to include a duplicate name field as well?

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -2159,6 +2217,91 @@ spec:
...
+                    type: array
+                  userID:
+                    description: UserID is the login for the user's IBM Cloud account.
+                    type: string
+                  vpc:
+                    description: VPC is a VPC inside IBM Cloud. Needed in order to
+                      create VPC Load Balancers.
+                    type: string
+                  vpcRegion:
+                    description: Zone in the region used to create VPC resources.
```
Prashanth684, line 2289: is this vpcRegion or vpcZone...there seems to be some inconsistency here with the naming

pkg/types/powervs/powervs_regions.go
clnperez: nope, we can remove that `Name` field from `Region`.

data/data/install.openshift.io_installconfigs.yaml
clnperez: region. that looks like a copy-paste problem in the description

clnperez:
@staebler ok to merge?

pkg/types/powervs/platform.go
staebler: Let's still clean up the text in the description to represent the current behavior.

pkg/types/powervs/machinepools.go
staebler: Please either detail why this needs to be a string or make it an int64.

pkg/types/powervs/machinepools.go
```diff
@@ -0,0 +1,55 @@
+package powervs
+
+// +kubebuilder:validation:Enum=capped;dedicated;shared
+type procType string
```
staebler, line 4: This must be a public type so that it can be used in other packages.
```suggestion
type ProcType string
```

pkg/types/powervs/machinepools.go
staebler: Still need valid values added to the description of this field.

pkg/types/powervs/machinepools.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// Memory defines the memory in GB for the instance.
+	//
+	// +optional
+	Memory string `json:"memory"`
+
+	// Processors defines the processing units for the instance.
+	//
+	// +optional
+	Processors string `json:"processors"`
```
staebler, line 21: All optional values should be `omitempty`.
```suggestion
	Processors string `json:"processors,omitempty"`
```

pkg/types/powervs/machinepools.go
```diff
@@ -0,0 +1,55 @@
+package powervs
+
+// +kubebuilder:validation:Enum=capped;dedicated;shared
```
staebler, line 3: An empty string is a valid value, too.
```suggestion
// +kubebuilder:validation:Enum="";capped;dedicated;shared
```

pkg/types/powervs/machinepools.go
staebler, line 4: Provide constants for the values of the enum.

pkg/types/powervs/platform.go
```diff
@@ -0,0 +1,57 @@
+package powervs
+
+// Platform stores all the global configuration that all machinesets
+// use.
+type Platform struct {
+
+	// ServiceInstanceID is the ID of the Power IAAS instance created from the IBM Cloud Catalog
+	ServiceInstanceID string `json:"serviceInstanceID"`
+
+	// PowerVSResourceGroup is the resource group for creating Power VS resources.
```
staebler, line 10:
```suggestion
	// PowerVSResourceGroup is the resource group in which the Power VS resources will be created.
```

pkg/types/powervs/platform.go
```diff
@@ -0,0 +1,57 @@
...
+	PowerVSResourceGroup string `json:"powervsResourceGroup"`
+
+	// Region specifies the IBM Cloud colo region where the cluster will be created.
+	Region string `json:"region"`
+
+	// Zone specifies the IBM Cloud colo region where the cluster will be created.
+	// At this time, only single-zone clusters are supported.
```
staebler, line 17: "At this time" implies that at some point in the future we may support mutli-zone clusters. At that point, the `Zone` field will need to be deprecated in favor of a `Zones` field. Do you have a feel for (1) the likelihood of supporting multi-zone clusters and (2) the timeframe for said support?

pkg/types/powervs/validation/machinepool.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/openshift/installer/pkg/types/powervs"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+// ValidateMachinePool checks that the specified machine pool is valid.
+func ValidateMachinePool(p *powervs.MachinePool, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	// Validate VolumeIDs
```
staebler, line 18: Do you want to validate that the volume IDs are unique?

pkg/types/powervs/validation/machinepool.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	// Validate Processors
+	if p.Processors != "" {
+		processors, err := strconv.ParseFloat(p.Processors, 64)
+		if err == nil {
+			if processors < 0.25 || processors > 32 {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("processors"), p.Processors, "number of processors must be from .25 to 32 cores"))
+			}
+			if math.Mod(processors*1000, 2) != 0 || math.Mod(processors*100, 25) != 0 {
```
staebler, line 45: This is still problematic with floating-point arithmetic. Multiplying does not change anything with regards to precision.

pkg/types/powervs/validation/platform.go
```diff
@@ -0,0 +1,18 @@
+package validation
+
+import (
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	"github.com/openshift/installer/pkg/types/powervs"
+)
+
+// ValidatePlatform checks that the specified platform is valid.
+func ValidatePlatform(p *powervs.Platform, fldPath *field.Path) field.ErrorList {
```
staebler, line 10: This is missing validation for most of the fields.

pkg/types/powervs/platform.go
```diff
@@ -0,0 +1,57 @@
...
+	Zone string `json:"zone"`
+
+	// VPCRegion specifies the IBM Cloud region in which to create VPC resources.
```
staebler, line 20: What is the use case for having the VPC resources in a different region than the other cluster resources?

pkg/types/powervs/validation/machinepool.go
clnperez: can you give me a sample input for which this would be problematic? all the test inputs i've given have worked so i'm missing something

pkg/types/powervs/platform.go
clnperez: Good point, but no I don't. AFAIK it's not in the roadmap yet.

pkg/types/powervs/platform.go
clnperez: are you asking why we have a Region and a VPCRegion? VPC Region is for the IBM Cloud region, e.g. `eu-gb`. Region is for the Power VS (different DCs and DC names) that is colocated with that IBM Cloud region, e.g. `lon`.

pkg/types/powervs/validation/platform.go
clnperez: yah. this is a wip in our fork: https://github.com/openshift-powervs/installer/pull/102/files
ok to pull that in in a followon PR? for now would like to get these in so we can stop using that fork

pkg/types/powervs/validation/machinepool.go
clnperez: yes, can we do that in a followon PR as well?

clnperez:
@staebler i believe all your most recent comments have been addressed or followed up on. ptal at the latest

pkg/types/powervs/validation/machinepool.go
staebler: There are no examples of it failing for the valid values since floating points have binary precision and we are dealing with quarters. However, this is fragile and easily avoided by using regex matching of the string instead of floating point arithmetic.

pkg/types/powervs/machinepools.go
```diff
@@ -0,0 +1,65 @@
...
+type ProcType string
+
+// Capped type for capped processor consumption
+const Capped ProcType = "capped"
+
+// Dedicated type for dedicated processor(s)
+const Dedicated ProcType = "dedicated"
+
+// Shared type shared type for shared processor(s)
+const Shared ProcType = "shared"
```
staebler, line 14: Tie these together in a `const` block so that it is easier to see that they are all of the values of `ProcType`.
```suggestion
const (
	// Capped type for capped processor consumption
	Capped ProcType = "capped"

	// Dedicated type for dedicated processor(s)
	Dedicated ProcType = "dedicated"

	// Shared type shared type for shared processor(s)
	Shared ProcType = "shared"
)
```

staebler (commented):
Just a nit. We can merge as is.
/lgtm
/approve

staebler:
/lgtm cancel
The `unit` and `verify-codegen` pre-submit checks are failing.

clnperez:
sorry missed a typo. fixed and those are passing now. PTAL @staebler

clnperez:
/test e2e-aws
this one's being weird. timeouts and image pull issues. @rna-afk any idea if something's up with ci?

rna-afk:
Not aware of any problems in CI right now. The e2e-aws test is failing since this morning. Probably some kind of performance issue with AWS.

PR #5610, opened 2022-02-04
Add IBM Power VS: manifest assets

clnperez:

Older discussions on some of the code here can be found in
https://github.com/openshift/installer/pull/5224

### DIFF
--- a/pkg/asset/manifests/dns.go
+++ b/pkg/asset/manifests/dns.go
@@ -28,6 +28,7 @@ import (
 	nonetypes "github.com/openshift/installer/pkg/types/none"
 	openstacktypes "github.com/openshift/installer/pkg/types/openstack"
 	ovirttypes "github.com/openshift/installer/pkg/types/ovirt"
+	powervstypes "github.com/openshift/installer/pkg/types/powervs"
 	vspheretypes "github.com/openshift/installer/pkg/types/vsphere"
 )
 
@@ -139,7 +140,7 @@ func (d *DNS) Generate(dependencies asset.Parents) error {
 			config.Spec.PublicZone = &configv1.DNSZone{ID: zone.Name}
 		}
 		config.Spec.PrivateZone = &configv1.DNSZone{ID: fmt.Sprintf("%s-private-zone", clusterID.InfraID)}
-	case ibmcloudtypes.Name:
+	case ibmcloudtypes.Name, powervstypes.Name:
 		client, err := icibmcloud.NewClient()
 		if err != nil {
 			return errors.Wrap(err, "failed to get IBM Cloud client")
--- a/pkg/asset/manifests/infrastructure.go
+++ b/pkg/asset/manifests/infrastructure.go
@@ -23,6 +23,7 @@ import (
 	"github.com/openshift/installer/pkg/types/none"
 	"github.com/openshift/installer/pkg/types/openstack"
 	"github.com/openshift/installer/pkg/types/ovirt"
+	"github.com/openshift/installer/pkg/types/powervs"
 	"github.com/openshift/installer/pkg/types/vsphere"
 )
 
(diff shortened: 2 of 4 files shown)


pkg/asset/manifests/infrastructure.go
```diff
@@ -211,6 +212,17 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
 			APIServerInternalIP: installConfig.Config.Ovirt.APIVIP,
 			IngressIP:           installConfig.Config.Ovirt.IngressVIP,
 		}
+	case powervs.Name:
+		config.Spec.PlatformSpec.Type = configv1.PowerVSPlatformType
+		cisInstanceCRN, err := installConfig.PowerVS.CISInstanceCRN(context.TODO())
+		if err != nil {
+			return err
```
rna-afk, line 219:
```suggestion
			return errors.Wrapf(err, "failed to get instance CRN")
```

kirankt (commented):
Before you commit changes to the PR, please make sure that you've run the linters and tests. This will help cut down failed CI tests.
https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#contribution-flow

clnperez:
@kirankt yep, thanks.i thought we had all those covered in our fork where we've been doing dev for the past several months but it looks like we missed some. will address those for sure

clnperez:
I think we'll need to get at least https://github.com/openshift/installer/pull/5609 merged before this will build and really run any tests

pkg/asset/manifests/dns.go
```diff
@@ -158,7 +159,7 @@ func (d *DNS) Generate(dependencies asset.Parents) error {
 		config.Spec.PrivateZone = &configv1.DNSZone{
 			ID: zoneID,
 		}
-	case libvirttypes.Name, openstacktypes.Name, baremetaltypes.Name, nonetypes.Name, vspheretypes.Name, ovirttypes.Name:
+	case libvirttypes.Name, openstacktypes.Name, baremetaltypes.Name, nonetypes.Name, vspheretypes.Name, ovirttypes.Name, powervstypes.Name:
```
Prashanth684, line 162: wouldn't this be the same as ibmcloud above?

clnperez: @Karthik-K-N -- can you answer this? i can't tell from looking at your Power VS support pr here https://github.com/openshift/cluster-ingress-operator/pull/658/files but am thinking we may be able to just add the powervs case to the ibmcloud case here (which is what i'm pretty sure Prashanth is suggesting)?

Karthik-K-N: Yes, We need to add PowerVS to IBMCloud case block and its tracked here https://issues.redhat.com/browse/MULTIARCH-1603 and also I think as per the slack discussion I see, @miyamotoh already done this.

clnperez:
looks like this also depends on https://github.com/openshift/installer/pull/5612

clnperez:
added cloud-config while we wait for the others, but this does now require https://github.com/openshift/installer/pull/5700

clnperez:
i think this one is good to go now @patrickdillon @Prashanth684 @rna-afk

pkg/asset/manifests/powervs/cloudproviderconfig.go
```diff
@@ -0,0 +1,79 @@
...
+	Region                   string `gcfg:"region"`
+	G2CredentialsFilePath    string `gcfg:"g2Credentials"`
+	G2ResourceGroupName      string `gcfg:"g2ResourceGroupName"`
+	G2VPCName                string `gcfg:"g2VpcName"`
+	G2WorkerServiceAccountID string `gcfg:"g2workerServiceAccountID"`
+	G2VPCSubnetNames         string `gcfg:"g2VpcSubnetNames"`
+}
+
+// CloudProviderConfig generates the cloud provider config for the IBMCloud platform.
+func CloudProviderConfig(infraID string, accountID string, vpcName string, region string, resourceGroupName string, subnets []string) (string, error) {
```
Prashanth684, line 37: can this not be common with ibmcloud? i don't see any difference other than the subnets?

clnperez: i believe @Karthik-K-N said they'll be adding some additional items when they get the ccm fork available, so i'd defer to him on that. but if we do agree to that let's make a tracker so we don't misplace that thought. in the meantime, ok to merge this as-is?

Prashanth684: if there's going to be some deviation then yeah, it's ok

Prashanth684: also a nit to change the above comment from IBMCloud to PowerVS

PR #5507, opened 2021-12-23
isolate terraform

staebler:
More progress on #5481 for removing terraform and terraform providers from the installer's module dependencies.

- [x] Terraform and all providers removed from go.mod.
- [x] All providers that are published in a public registry are downloaded from that registry rather than built locally. The exceptions are azurestack and vsphere for which we are carrying patches. All of the downloaded providers have been updated to the most recent version available.
- [x] Support for other architectures and OSes.

Follow-up work:
- [x] Add verification scripts for downloaded and vendored providers.
- [x] Add developer documentation for adding, updating, and maintaining providers.

PR #5613, opened 2022-02-04
Add IBM Power VS: tf plugin for ibm presign

clnperez:
This plugin is needed in order to securely host the ignition config
files for bootstrap machines. It uses the COS offering from IBM Cloud,
which is s3-compatible.

Older discussions on some of the code here can be found in #5224

Authored-by: Manjunath Kumatagi <mailto:mkumatag@in.ibm.com>

### DIFF
--- a/pkg/terraform/exec/plugins/ibms3presign.go
+++ b/pkg/terraform/exec/plugins/ibms3presign.go
@@ -1,0 +1,15 @@
+package plugins
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/plugin"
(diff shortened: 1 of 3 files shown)


clnperez:
Since this is now in: https://github.com/openshift/installer/pull/5507 -- we'll do what `ibmcloud` is doing for the bootstrap ignition config and won't need this. 

see https://github.com/openshift/installer/blob/86003e9797ec1c9a1160c156944556f557f9d3a7/data/data/ibmcloud/bootstrap/main.tf#L34:L38

cc @mkumatag @mjturek

PR #5614, opened 2022-02-04
Add IBM Power VS: tf data

clnperez:

Older discussions on some of the code here can be found in #5224

### DIFF
--- a/data/data/powervs/cluster/bootstrap/main.tf
+++ b/data/data/powervs/cluster/bootstrap/main.tf
@@ -1,0 +1,77 @@
+# TODO(mjturek): network and image data blocks can be in main module
+#                as master and bootstrap will be using the same
+#                network and image. Once we add in master module, make
+#                the move.
+data "ibm_pi_network" "network" {
+  pi_network_name      = var.network_name
+  pi_cloud_instance_id = var.cloud_instance_id
+}
+
+data "ibm_resource_group" "cos_group" {
+  name = var.resource_group
+}
+
+resource "ibm_resource_instance" "cos_instance" {
+  name              = "${var.cluster_id}-cos"
+  resource_group_id = data.ibm_resource_group.cos_group.id
+  service           = "cloud-object-storage"
+  plan              = "standard"
+  location          = var.cos_instance_location
+  tags              = [var.cluster_id]
+}
+
+# Create an IBM COS Bucket to store ignition
+resource "ibm_cos_bucket" "ignition" {
+  bucket_name          = "${var.cluster_id}-bootstrap-ign"
+  resource_instance_id = ibm_resource_instance.cos_instance.id
+  region_location      = var.cos_bucket_location
+  storage_class        = var.cos_storage_class
+}
+
+resource "ibm_resource_key" "cos_service_cred" {
+  name                 = "${var.cluster_id}-cred"
+  role                 = "Reader"
+  resource_instance_id = ibm_resource_instance.cos_instance.id
+  parameters           = { HMAC = true }
+}
+
+# Place the bootstrap ignition file in the ignition COS bucket
+resource "ibm_cos_bucket_object" "ignition" {
+  bucket_crn      = ibm_cos_bucket.ignition.crn
+  bucket_location = ibm_cos_bucket.ignition.region_location
+  content         = var.ignition
+  key             = "bootstrap.ign"
+  etag            = md5(var.ignition)
+}
+
+data "ibm_iam_auth_token" "iam_token" {}
+
+# Create the bootstrap instance
+resource "ibm_pi_instance" "bootstrap" {
+  pi_memory            = var.memory
+  pi_processors        = var.processors
+  pi_instance_name     = "${var.cluster_id}-bootstrap"
+  pi_proc_type         = var.proc_type
+  pi_image_id          = var.image_id
+  pi_sys_type          = var.sys_type
+  pi_cloud_instance_id = var.cloud_instance_id
+  pi_network {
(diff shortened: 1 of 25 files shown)


clnperez:
this needs some rework post https://github.com/openshift/installer/pull/5507 and to stop relying on the custom s3 plugin for the bootstrap ignition config

clnperez:
i'm going to add a few more bits to this so we have fewer PRs to review, b/c of the timing with the tests mentioned in other Power VS PRs.

clnperez:
@patrickdillon @rna-afk  -- I'm taking this back out of WIP and will push new changes to a new PR. The "small thing" I wanted to add here turned out to be a little more like a medium thing.

clnperez:
@rna-afk @patrickdillon @Prashanth684 can you take a look at this as well? Depends on https://github.com/openshift/installer/pull/5615

clnperez:
ping @rna-afk do you know if the list of required/optional tests only shows up after an lgtm?

rna-afk:
@patrickdillon has been working on clearing up the CI tests that are blocking PRs from merging. I'm hoping that's the reason we aren't seeing any required tests right now.

data/data/powervs/topology/pi_network.tf
```diff
@@ -0,0 +1,40 @@
+## Network
+## These are be optional arguments in the install-config (e.g. Platform)
+## so that users may specify them. Have them be "hidden" in that the survey doesn't ask for them
+## unless the OCP leads disagree.
+## And since they're optional, use the count = construct to conditionally create them if the tf
+## vars are unset.
+
+## Note, the following are incomplete placeholders to be tested and reviewed later when the TF
```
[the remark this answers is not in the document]
clnperez: no those i probably should have deleted but they're removed in #5780

clnperez:
@mjturek -- can you double-check the descriptions I just added (esp the COS ones for bootstrap)?

data/data/powervs/cluster/bootstrap/main.tf
```diff
@@ -0,0 +1,77 @@
+# TODO(mjturek): network and image data blocks can be in main module
+#                as master and bootstrap will be using the same
+#                network and image. Once we add in master module, make
+#                the move.
```
mjturek, line 4: This comment is no longer relevant, we can drop it.

data/data/powervs/cluster/bootstrap/variables.tf
```diff
@@ -0,0 +1,69 @@
...
+}
+
+variable "processors" {
+  type = string
+  description = "The processor count for each node."
+}
+
+variable "ignition" {
+  type = string
+  description = "The ignition file."
```
mjturek, line 13: I think this is okay, but "The contents of the ignition file" may be more accurate.

data/data/powervs/cluster/bootstrap/variables.tf
```diff
@@ -0,0 +1,69 @@
...
+}
+
+variable "cos_instance_location" {
+  type = string
+  description = "The region in which the Cloud Object Store resides. Used for the ignition file."
+}
+
+variable "cos_bucket_location" {
+  type = string
+  description = "The URL of the Cloud Object Store. Used for the igntion file."
```
mjturek, line 63: This is actually used for `region_location` [here](https://registry.terraform.io/providers/IBM-Cloud/ibm/latest/docs/resources/cos_bucket#region_location).

So it's not a URL, but a region like `us-south` or `eu-gb`.

mjturek (changes requested):
Comments inline

data/data/powervs/cluster/bootstrap/main.tf
clnperez: that's also dropped in #5780

data/data/powervs/cluster/bootstrap/variables.tf
clnperez: how would i differentiate that from `cos_instance_location` above?

mjturek: There isn't much difference, but `cos_instance_location` can be set to `global` unlike `region_location`. We actually use the `global` setting for `cos_instance_location` so as far as I can tell, we should keep both.

clnperez:
thanks @mjturek!

should be all good now @rna-afk

clnperez:
/retest-required

@patrickdillon these two keep failing for this and #5615. do i need to keep retrying? if a PR gets an approve does mergebot automatically retry for me?

data/data/powervs/topology/pi_network.tf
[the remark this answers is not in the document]
clnperez: @rna-afk i'm going to resolve this an assume it's okay to leave as-is b/c it's taken care of in the other pr. lmk if that's not okay.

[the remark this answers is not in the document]
patrickdillon: I would prefer not to merge code to remove it in a later pr/commit & #5780 still needs to be broken into separate commits. Can we get rid of this and the other unnecessary comments?

patrickdillon (commented):
This looks very close. But I would prefer to remove the unneeded comments, especially considering one is an entire file.

data/data/powervs/topology/pi_network.tf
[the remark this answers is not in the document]
clnperez: sure. deleted this whole file and re-pushed

clnperez:
@patrickdillon can you skip those two for this pr as well? nothing seems to be re-running anyway

PR #5621, opened 2022-02-07
data/manifests/bootkube/cvo-overrides: Default to stable-4.11

wking:
Like d1bf677d62 (#5312) and earlier, now that 4.10 has forked off from the development branch.

### DIFF
--- a/data/data/manifests/bootkube/cvo-overrides.yaml.template
+++ b/data/data/manifests/bootkube/cvo-overrides.yaml.template
@@ -8,6 +8,6 @@ spec:
   upstream: https://amd64.origin.releases.ci.openshift.org/graph
   channel: stable-4
 {{- else }}
-  channel: stable-4.10
+  channel: stable-4.11
 {{- end }}
   clusterID: {{.CVOClusterID}}


data/data/manifests/bootkube/cvo-overrides.yaml.template
```diff
@@ -8,6 +8,6 @@ spec:
   upstream: https://amd64.origin.releases.ci.openshift.org/graph
   channel: stable-4
 {{- else }}
-  channel: stable-4.10
+  channel: stable-4.11
```
LalatenduMohanty, line 11: Can a test be added so that every time the test runs in a new branch it will fail unless the version is added in this file? It is bad situation is someone has to remember every time to send a PR manually

LalatenduMohanty (changes requested):

data/data/manifests/bootkube/cvo-overrides.yaml.template
LalatenduMohanty: I meant unit test.

PR #5623, opened 2022-02-08
Bug 2035642: Report bootstrap is complete only when the bootstrap static pod is no longer used

eranco74:
From time to time Single node installation with bootstrap in place fails.
When the installation fails after the bootstrap phase is complete and prior of having a running kube-apiserver we fail to get the logs.

This change updates the progress-report script to not report that the bootstrap is complete, instead it will set the status to `bootstrap-in-place`.
The status will get updated to complete in a later phase once the OCP kube-apiserver is running.
This will prolong the time the installer waits for bootstrap to complete, and in case the installation fails while the bootstrap before we have a running OCP kube-apiserver the installer will fall back to execute installer-gather and get the relevant logs.


eranco74:
The log collection issue happens on AWS IPI and isn't related to bootstrap in place.
Closing.

PR #5629, opened 2022-02-09
Validate user selected capabilities

kirankt:
This is a follow up to #5605 and has a hard dependency on it.

This PR adds validation to the user-selectable capabilities api in the install-config.


pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,31 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+		// Populate the available capability set map for the chosen baselineCapablitySet
+		for _, capability := range availableAdditionalCapabilties {
+			capabilitySet[capability] = true
+		}
+		// Check for additionalEnabledCapabilities input by user
+		for _, inputCapability := range c.AdditionalEnabledCapabilities {
+			if capabilitySet[inputCapability] != true {
+				allErrs = append(allErrs, field.Invalid(fldPath, c.AdditionalEnabledCapabilities, fmt.Sprintf("unknown additional enabled capability specified")))
+			}
+		}
```
rna-afk, line 728: nit: Just a thought but I think using sets would be easier here. I'm fine either way
```
// Check to see if the additionalEnabledCapabilities is defined in openshift/api
if availableCapabilities, found := configv1.ClusterVersionCapabilitySets[c.BaselineCapabilitySet]; found {
	// Populate the available capability set map for the chosen baselineCapablitySet
	capabilitySet := sets.NewString(availableCapabilities...)
	// Check for additionalEnabledCapabilities input by user
	missingCapability := sets.NewString(c.AdditionalEnabledCapabilities...).Difference(capabilitySet)
	for _, missing := range missingCapability.List() {
		allErrs = append(allErrs, field.Invalid(fldPath, missing, fmt.Sprintf("unknown additional enabled capability specified")))
	}
}
```

kirankt: Looks a lot more elegant than using a map. Will change. Thanks for the tip.

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,35 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
 	}
 	return allErrs
 }
+
+// validateCapabilities checks if additional, optional openshift components are specified in the
+// install-config to be included in the installation
+func validateCapabilities(c *types.Capabilities, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if c.BaselineCapabilitySet == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("baselineCapabilitySet"), "baseline capability set required"))
```
wking, line 712: This is [optional in the ClusterVersion API][1], so I dunno if we need this guard.

[1]: https://github.com/openshift/api/blob/b632c5fc10c0ea86cb18577db9388109a43d465f/config/v1/types_cluster_version.go#L281-L282

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,35 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+		for _, capability := range c.AdditionalEnabledCapabilities {
+			additionalCapabilitySet.Insert(string(capability))
+		}
+		// Check for unknown capabilities between available and additional sets
+		missingCapability := additionalCapabilitySet.Difference(availableCapabilitySet)
+		for _, missing := range missingCapability.List() {
+			allErrs = append(allErrs, field.Invalid(fldPath, missing, fmt.Sprintf("unknown additional enabled capability specified")))
+		}
+	} else {
+		allErrs = append(allErrs, field.Invalid(fldPath, c.BaselineCapabilitySet, fmt.Sprintf("unknown baseline capability set specified")))
```
wking, line 734: possibly helpful to include a list of `configv1.ClusterVersionCapabilitySets` keys here, so folks see the valid choices and can copy/paste typo corrections?

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,35 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+	allErrs := field.ErrorList{}
+	if c.BaselineCapabilitySet == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("baselineCapabilitySet"), "baseline capability set required"))
+
+		return allErrs
+	}
+	// Check to see if the additionalEnabledCapabilities is defined in openshift/api
+	if availableCapabilities, found := configv1.ClusterVersionCapabilitySets[c.BaselineCapabilitySet]; found {
+		// create a set of the available capability set
+		availableCapabilitySet := sets.NewString()
```
patrickdillon, line 719: This looks like Hungarian Notation, where you are including `Set` in the variable name because that is type, but I think this is confusing because two lines earlier `CapabilitySet` means something else. I would instead call this something like `validCapabilities`

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,35 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+		for _, capability := range availableCapabilities {
+			availableCapabilitySet.Insert(string(capability))
+		}
+		// create a set of the user specified additional capability set
+		additionalCapabilitySet := sets.NewString()
```
patrickdillon, line 724: Same comment as above. Perhaps something like `userCapabilities` or whatever is consistent with similar checks in the code.

patrickdillon (commented):
Looks pretty good, but left some comments, mostly nits.

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,35 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+	if c.BaselineCapabilitySet == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("baselineCapabilitySet"), "baseline capability set required"))
+
+		return allErrs
+	}
+	// Check to see if the additionalEnabledCapabilities is defined in openshift/api
+	if availableCapabilities, found := configv1.ClusterVersionCapabilitySets[c.BaselineCapabilitySet]; found {
+		// create a set of the available capability set
+		availableCapabilitySet := sets.NewString()
+		for _, capability := range availableCapabilities {
```
wking, line 720: `additionalEnabledCapabilities` is expected to contain caps that are not in the baseline set.  Some cases:

* An `additionalEnabledCapabilities` cap is in the baseline set.  This is redundant stutter, because the set already enables the capability.  The user could drop the cap from `additionalEnabledCapabilities` and get exactly the same set of enabled caps.  But still, redundant isn't broken, so probably not worth pestering users about these.
* An `additionalEnabledCapabilities` cap is not in the baseline set, but it is in some other set.  This is great, the user is extending their requested cap set beyond their selected base set.  Lovely.
* An `additionalEnabledCapabilities` cap is not in any set.  Probably a typo, and we should reject the install-config.

So instead of building `availableCapabilitySet` from the selected baseline set, we should build it to include all caps from every set while iterating over every set in `configv1.ClusterVersionCapabilitySets`.  And then we can complain about unrecognized caps here.

pkg/types/validation/installconfig.go
wking: Folding into [this thread][1], which is making the same point.

[1]: https://github.com/openshift/installer/pull/5629#discussion_r804093636

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,35 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
 	}
 	return allErrs
 }
+
+// validateCapabilities checks if additional, optional openshift components are specified in the
+// install-config to be included in the installation
```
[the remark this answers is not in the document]
kirankt: Done. Thanks.

pkg/types/validation/installconfig.go
kirankt: I agree with you that this logic is broken. I'll work on a fix. Looking at the ClusterVersionCapabilitySets, it would be nice to maintain a *everything included* set which we could use to look up and figure out whether the user input is kosher or not. WDYT?

kirankt: Also, wouldn't you think that a union of all the sets available in `configv1.ClusterVersionCapabilitySets` could potentially leave out an list of actual optional caps? Hence my mention of the "all possible capabilities" set.

pkg/types/validation/installconfig.go
kirankt: Done.

pkg/types/installconfig.go
```diff
@@ -383,3 +389,20 @@ func (c *InstallConfig) WorkerMachinePool() *MachinePool {
 
 	return nil
 }
+
+// Capabilities defines the configuration for user to install/exclude optional OpenShift components
+type Capabilities struct {
+	// baselineCapabilitySet selects an initial set of
```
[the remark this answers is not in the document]
kirankt: Removing this because we'll end up using openshift/api's types.

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,35 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+func validateCapabilities(c *types.Capabilities, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if c.BaselineCapabilitySet == "" {
+		allErrs = append(allErrs, field.Required(fldPath.Child("baselineCapabilitySet"), "baseline capability set required"))
+
+		return allErrs
+	}
```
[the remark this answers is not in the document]
kirankt: Thanks. Will remove.

pkg/types/validation/installconfig.go
kirankt: Reworked the logic a bit here and completely removed the use of sets. Thanks.

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,35 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+		}
+		// create a set of the user specified additional capability set
+		additionalCapabilitySet := sets.NewString()
+		for _, capability := range c.AdditionalEnabledCapabilities {
+			additionalCapabilitySet.Insert(string(capability))
+		}
+		// Check for unknown capabilities between available and additional sets
+		missingCapability := additionalCapabilitySet.Difference(availableCapabilitySet)
+		for _, missing := range missingCapability.List() {
+			allErrs = append(allErrs, field.Invalid(fldPath, missing, fmt.Sprintf("unknown additional enabled capability specified")))
```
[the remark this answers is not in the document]
kirankt: I'm keeping the sprintf here because I'm going to list all possible caps values if the user ends up making a mistake.

pkg/types/installconfig.go
```diff
@@ -383,3 +389,20 @@ func (c *InstallConfig) WorkerMachinePool() *MachinePool {
 
 	return nil
 }
+
+// Capabilities defines the configuration for user to install/exclude optional OpenShift components
+type Capabilities struct {
```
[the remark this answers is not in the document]
kirankt: I agree. I'll create a new card and will add it.

pkg/types/validation/installconfig.go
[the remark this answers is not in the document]
patrickdillon: If it helps there is an example of doing that here: https://github.com/openshift/installer/blob/master/pkg/types/azure/validation/platform.go#L65-L67

Not sure if this translates exactly

pkg/types/validation/installconfig.go
wking: I agree that it's theoretically possible to have a capability that belongs to no sets.  But I would be very surprised if we ever had a capability like that, and even if we did, the folks testing that capability would notice when the installer complained about the unrecognized value, and we could patch things up at that point.

But today we do not have any capabilities like that.  In fact, `vCurrent` will contain the complete set of caps for 4.10, so the super simple approach is just to use `vCurrent` to get the valid caps.  Iterating over all sets in the map gives you some protection against some hypothetical future where `vCurrent` leaves out some new capability and a new `allTheCaps` set or whatever takes over that role.

pkg/types/installconfig.go
```diff
@@ -389,20 +389,3 @@ func (c *InstallConfig) WorkerMachinePool() *MachinePool {
 
 	return nil
 }
-
-// Capabilities defines the configuration for user to install/exclude optional OpenShift components
-type Capabilities struct {
```
wking: Needs a rebase to set this commit back on top of the current #5605.

```console
$ git --no-pager log --oneline --graph -6 origin/pr/5605 origin/pr/5629
* c3b26f3f1 (origin/pr/5629) Validate additional user-specified openshift component capabilities
* 3c7d85c76 Vendor the latest version of openshift/api which contains support for user-selectable capabilities
* 77dfc61e8 Add user-selectable capabilities api in the installconfig
| * 65d684eef (origin/pr/5605) Vendor the latest version of openshift/api which contains support for user-selectable capabilities
| * 0542412a1 Add user-selectable capabilities api in the installconfig
|/  
*   d436da243 Merge pull request #5640 from nutanix-cloud-native/update-owner-aliases
|\ 
```

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,33 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+
+	if _, found := configv1.ClusterVersionCapabilitySets[c.BaselineCapabilitySet]; !found {
+		allErrs = append(allErrs, field.Invalid(fldPath, c.BaselineCapabilitySet, "unknown baseline capability set specified"))
+	}
+
+	allAvailableCapabilities := map[string]bool{}
+	allCaps := []string{}
+	// Create a map and slice of *all* available capabilities across all capability sets
+	for k, _ := range configv1.ClusterVersionCapabilitySets {
+		for _, capability := range configv1.ClusterVersionCapabilitySets[k] {
```
wking, line 720: nit: you should be able to use:

```go
for _, capabilities := range configv1.ClusterVersionCapabilitySets {
    for _, capability := range capabilities {
```

to avoid having to touch the keys or lookup from the map twice.

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,33 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+			if _, found := allAvailableCapabilities[string(capability)]; !found {
+				allAvailableCapabilities[string(capability)] = true
+				allCaps = append(allCaps, string(capability))
+			}
+		}
+	}
+	sort.Strings(allCaps)
```
wking, line 727: micro-nit: you can save fractions of CPU in the happy, error-free case by moving this sort down into the `!found` case below.  If we end up confirming the sort a few times when folks typo multiple capabilities, that seems like it would still be less energy than sorting all the runs where the user didn't typo anything.

pkg/types/validation/installconfig_test.go
```diff
@@ -1399,6 +1400,42 @@ func TestValidateInstallConfig(t *testing.T) {
...
+		},
+		{
+			name: "invalid additional enabled capability specified",
+			installConfig: func() *types.InstallConfig {
+				c := validInstallConfig()
+				c.Capabilities = &configv1.ClusterVersionCapabilitiesSpec{BaselineCapabilitySet: "v4.11",
+					AdditionalEnabledCapabilities: []configv1.ClusterVersionCapability{"not-valid"}}
+				return c
+			}(),
+			expectedError: `"not-valid": Unknown capability. Must specify one of the following \(openshift-samples\)`,
```
wking, line 1437: We will add more caps over time, so probably worth future-proofing this regexp to `\(.*openshift-samples.*\)` or some such.

pkg/types/validation/installconfig_test.go
```diff
@@ -1399,6 +1400,42 @@ func TestValidateInstallConfig(t *testing.T) {
...
+			}(),
+		},
+		{
+			name: "invalid baseline capability set specified",
+			installConfig: func() *types.InstallConfig {
+				c := validInstallConfig()
+				c.Capabilities = &configv1.ClusterVersionCapabilitiesSpec{BaselineCapabilitySet: "notValid"}
+				return c
+			}(),
+			expectedError: `"notValid": unknown baseline capability set specified`,
```
wking, line 1418: nit: here's another place where we could help the user out with a list of options.  But probably a rare enough problem that I don't think we'd be fine as you have it too.

pkg/types/validation/installconfig.go
kirankt: That makes sense. Thanks.

pkg/types/validation/installconfig.go
wking: Fixed in a10be333d

pkg/types/validation/installconfig_test.go
wking: Fixed in a10be333d

pkg/types/validation/installconfig_test.go
wking: Fixed in a10be333d

pkg/types/validation/installconfig.go
wking: Fixed in a10be333d

kirankt:
Closing this PR in favor of #5645

PR #5631, opened 2022-02-10
Add Nutanix-specific customizations to manifests

thunderboltsid:
- Update templates in data/data/manifests
- Add nutanix platform customizations in pkg/asset/manifests

### DIFF
--- a/data/data/manifests/openshift/cloud-creds-secret.yaml.template
+++ b/data/data/manifests/openshift/cloud-creds-secret.yaml.template
@@ -16,6 +16,8 @@ metadata:
   name: vsphere-creds
 {{- else if .CloudCreds.Ovirt}}
   name: ovirt-credentials
+{{- else if .CloudCreds.Nutanix}}
+  name: nutanix-credentials
 {{- end}}
 data:
 {{- if .CloudCreds.AWS}}
@@ -45,4 +47,9 @@ data:
   ovirt_password: {{.CloudCreds.Ovirt.Base64encodePassword}}
   ovirt_insecure: {{.CloudCreds.Ovirt.Base64encodeInsecure}}
   ovirt_ca_bundle: {{.CloudCreds.Ovirt.Base64encodeCABundle}}
+{{- else if .CloudCreds.Nutanix}}
+  NUTANIX_ENDPOINT: {{.CloudCreds.Nutanix.Base64encodeEndpoint}}
+  NUTANIX_PORT: {{.CloudCreds.Nutanix.Base64encodePort}}
+  NUTANIX_USER: {{.CloudCreds.Nutanix.Base64encodeUsername}}
+  NUTANIX_PASSWORD: {{.CloudCreds.Nutanix.Base64encodePassword}}
 {{- end}}
--- a/pkg/asset/manifests/cloudproviderconfig.go
+++ b/pkg/asset/manifests/cloudproviderconfig.go
@@ -30,6 +30,7 @@ import (
 	ibmcloudtypes "github.com/openshift/installer/pkg/types/ibmcloud"
 	libvirttypes "github.com/openshift/installer/pkg/types/libvirt"
 	nonetypes "github.com/openshift/installer/pkg/types/none"
+	nutanixtypes "github.com/openshift/installer/pkg/types/nutanix"
 	openstacktypes "github.com/openshift/installer/pkg/types/openstack"
 	ovirttypes "github.com/openshift/installer/pkg/types/ovirt"
 	powervstypes "github.com/openshift/installer/pkg/types/powervs"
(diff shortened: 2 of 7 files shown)


data/data/manifests/openshift/cloud-creds-secret.yaml.template
```diff
@@ -45,4 +47,9 @@ data:
   ovirt_password: {{.CloudCreds.Ovirt.Base64encodePassword}}
   ovirt_insecure: {{.CloudCreds.Ovirt.Base64encodeInsecure}}
   ovirt_ca_bundle: {{.CloudCreds.Ovirt.Base64encodeCABundle}}
+{{- else if .CloudCreds.Nutanix}}
```
rvanderp3, line 50: What is, or is there, a relationship between this secret and the CSI driver operator secret `ntnx-secret`?  The [enhancement](https://github.com/openshift/enhancements/blob/8e71d3860e433fa5d118f31c69825df5913da7c1/enhancements/installer/nutanix-ahv-ipi.md?plain=1#L124) references the Nutanix CSI driver operator which depends on `ntnx-secret` in the `ntnx-system` namespace.

thunderboltsid: From my understanding, these secrets have the same contents albeit in the case of CSI driver they tend to follow a string based encoding of the values (prism-element-ip:prism-port:username:password).

rvanderp3: @thunderboltsid I think this the [discussion](https://github.com/openshift/cloud-credential-operator/pull/450#discussion_r800788502) you are referring to.  Do you happen to have a timeline on when the transition to moving the configuration of the ip/port to the infrastructure CR might take place and would moving the configuration to the infrastructure CR at a later time cause issues?

thunderboltsid: We can move the IP/Port over to infra CR at a later date without any issues.

pkg/asset/manifests/cloudproviderconfig.go
```diff
@@ -92,7 +93,7 @@ func (cpc *CloudProviderConfig) Generate(dependencies asset.Parents) error {
 	}
 
 	switch installConfig.Config.Platform.Name() {
-	case libvirttypes.Name, nonetypes.Name, baremetaltypes.Name, ovirttypes.Name:
+	case libvirttypes.Name, nonetypes.Name, baremetaltypes.Name, ovirttypes.Name, nutanixtypes.Name:
```
staebler, line 96: I am a bit surprised that there is not a cloud provider config. Is the name of the secret with the creds hard-coded somewhere?

pkg/asset/manifests/template.go
```diff
@@ -54,6 +54,14 @@ type OvirtCredsSecretData struct {
 	Base64encodeCABundle string
 }
 
+// NutanixCredsSecretData holds encoded credentials and is used to generate cloud-creds secret
+type NutanixCredsSecretData struct {
+	Base64encodeNutanixEndpoint string
+	Base64encodeNutanixPort     string
+	Base64encodeNutanixUsername string
+	Base64encodeNutanixPassword string
```
staebler, line 62: "Nutanix" in these names is redundant.
```suggestion
	Base64encodeEndpoint string
	Base64encodePort     string
	Base64encodeUsername string
	Base64encodePassword string
```

pkg/asset/manifests/cloudproviderconfig.go
thunderboltsid: Currently, yes. `nutanix-credentials` is the credential name.

data/data/manifests/openshift/cloud-creds-secret.yaml.template
```diff
@@ -45,4 +47,9 @@ data:
...
+  NUTANIX_ENDPOINT: {{.CloudCreds.Nutanix.Base64encodeNutanixEndpoint}}
+  NUTANIX_PORT: {{.CloudCreds.Nutanix.Base64encodeNutanixPort}}
+  NUTANIX_USER: {{.CloudCreds.Nutanix.Base64encodeNutanixUsername}}
+  NUTANIX_PASSWORD: {{.CloudCreds.Nutanix.Base64encodeNutanixPassword}}
```
rvanderp3, line 54: @thunderboltsid does this need to refactored in light of https://github.com/openshift/installer/pull/5631/files#diff-89d415ff2b74dae9eb1035ae2b655b8f099235198e424c4108e000601cc1dec6R218 ?

thunderboltsid: Indeed

PR #5635, opened 2022-02-10
Add the terraform plugin for Nutanix

thunderboltsid:
Add Nutanix terraform plugin in pkg/terraform/exec/plugins.


rvanderp3:
Can you separate vendor changes into their own commits, to make reviewing and perusing history easier.

rvanderp3:
@thunderboltsid there have been some updates regarding how additional terraform providers are added to the installer.  Please see this document for details.

rvanderp3:
/lgtm 

My only comment here is that it would have been helpful to have tools.go in its own commit.

thunderboltsid:
> /lgtm
> 
> My only comment here is that it would have been helpful to have tools.go in its own commit.

Given the entire file is a single import statement, does that really help?

staebler (commented):
1. Please separate the terraform/providers/nutanix/vendor files into their own commit.
2. Fix the commit message, as it still refers to an old directory.
3. Add the provider in pkg/terraform/providers/providers.go.

staebler:
> > /lgtm
> > My only comment here is that it would have been helpful to have tools.go in its own commit.
> 
> Given the entire file is a single import statement, does that really help?

Yes, it is helpful to have the tools.go, go.mod, and go.sum in their own commit. With it all in one commit, it is difficult to review the commit in github. I am not looking over each file in the vendor directory, so it is convenient to have them separated from the files that I do want to look at.

staebler:
See https://github.com/openshift/installer/commit/7c216a7c1104a610004bd0a23cecbc38bf6e688d and https://github.com/openshift/installer/commit/d7fe250d6c51c6a770f880166a184b9b7abe36a5 for precedent.

thunderboltsid:
> * Please separate the terraform/providers/nutanix/vendor files into their own commit.
> * Fix the commit message, as it still refers to an old directory.
> * Add the provider in pkg/terraform/providers/providers.go.

Reg. 3: No such file exists https://github.com/openshift/installer/tree/master/terraform/providers

thunderboltsid:
> * Please separate the terraform/providers/nutanix/vendor files into their own commit.
> * Fix the commit message, as it still refers to an old directory.
> * Add the provider in pkg/terraform/providers/providers.go.

Done.

staebler (commented):
/lgtm
/approve

PR #5636, opened 2022-02-10
Update declarative assets for Nutanix

thunderboltsid:
- Update bootkube template to copy static resources for nutanix under data/data/bootstrap/files
- Add terraform files for Nutanix platform under data/data/nutanix
- Add terraform stages for Nutanix platform under pkg/terraform/stages/nutanix

### DIFF
--- a/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
+++ b/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
@@ -337,6 +337,7 @@ then
 	copy_static_resources_for openstack
 	copy_static_resources_for ovirt
 	copy_static_resources_for vsphere
+	copy_static_resources_for nutanix
 
 	cp mco-bootstrap/manifests/* manifests/
 
--- a/data/data/nutanix/bootstrap/main.tf
+++ b/data/data/nutanix/bootstrap/main.tf
@@ -1,0 +1,74 @@
+locals {
+  description = "Created By OpenShift Installer"
+}
+
+provider "nutanix" {
+  wait_timeout = 60
+  username     = var.nutanix_username
+  password     = var.nutanix_password
+  endpoint     = var.nutanix_prism_central_address
+  port         = var.nutanix_prism_central_port
+}
+
+resource "nutanix_image" "bootstrap_ignition" {
+  name        = var.nutanix_bootstrap_ignition_image
+  source_path = var.nutanix_bootstrap_ignition_image_filepath
+  description = local.description
+
+  categories {
+    name  = var.ocp_category_key_id
+    value = var.ocp_category_value_owned_id
+  }
+}
+
+resource "nutanix_virtual_machine" "vm_bootstrap" {
+  name                 = "${var.cluster_id}-bootstrap"
+  description          = local.description
+  cluster_uuid         = var.nutanix_prism_element_uuid
+  num_vcpus_per_socket = 4
+  num_sockets          = 1
+  memory_size_mib      = 16384
+  boot_device_order_list = [
+    "DISK",
+    "CDROM",
+    "NETWORK"
+  ]
+
+  disk_list {
+    device_properties {
+      device_type = "DISK"
+      disk_address = {
+        device_index = 0
+        adapter_type = "SCSI"
+      }
+    }
+    data_source_reference = {
+      kind = "image"
+      uuid = var.image_id
+    }
+    disk_size_mib = var.nutanix_control_plane_disk_mib
+  }
+
+  disk_list {
+    device_properties {
+      device_type = "CDROM"
+      disk_address = {
+        adapter_type = "IDE"
+        device_index = 0
+      }
+    }
+    data_source_reference = {
+      kind = "image"
+      uuid = nutanix_image.bootstrap_ignition.id
+    }
+  }
+
+  categories {
+    name  = var.ocp_category_key_id
+    value = var.ocp_category_value_owned_id
+  }
+
+  nic_list {
+    subnet_uuid = var.nutanix_subnet_uuid
+  }
+}
--- a/data/data/nutanix/variables-nutanix.tf
+++ b/data/data/nutanix/variables-nutanix.tf
@@ -1,0 +1,73 @@
+//////
+// Nutanix variables
+//////
+
+variable "nutanix_prism_central_address" {
+  type        = string
+  description = "Address to connect to Prism Central."
+}
+
+variable "nutanix_prism_central_port" {
+  type        = string
+  description = "Port to connect to Prism Central."
+}
+
+variable "nutanix_username" {
+  type        = string
+  description = "Prism Central user for the environment."
+}
+
+variable "nutanix_password" {
+  type        = string
+  description = "Prism Central user password"
+}
+
+variable "nutanix_prism_element_uuid" {
+  type        = string
+  description = "This is the uuid of the Prism Element cluster."
+}
+
+variable "nutanix_image_filepath" {
+  type        = string
+  description = "This is the filepath to the image file that will be imported into Prism Central."
+}
+
+variable "nutanix_image" {
+  type        = string
+  description = "This is the name to the image that will be imported into Prism Central."
+}
+
+variable "nutanix_subnet_uuid" {
+  type        = string
+  description = "This is the uuid of the publicly accessible subnet for cluster ingress and access."
+}
+
+variable "nutanix_bootstrap_ignition_image" {
+  type        = string
+  description = "Name of the image containing the bootstrap ignition files"
+}
+
+variable "nutanix_bootstrap_ignition_image_filepath" {
+  type        = string
+  description = "Path to the image containing the bootstrap ignition files"
+}
+
+///////////
+// Control Plane machine variables
+///////////
+
+variable "nutanix_control_plane_memory_mib" {
+  type = number
+}
+
+variable "nutanix_control_plane_disk_mib" {
+  type = number
+}
+
+variable "nutanix_control_plane_num_cpus" {
+  type = number
+}
+
+variable "nutanix_control_plane_cores_per_socket" {
+  type = number
+}
--- a/pkg/terraform/stages/nutanix/stages.go
+++ b/pkg/terraform/stages/nutanix/stages.go
@@ -1,0 +1,13 @@
+package nutanix
+
+import (
+	"github.com/openshift/installer/pkg/terraform"
+	"github.com/openshift/installer/pkg/terraform/providers"
+	"github.com/openshift/installer/pkg/terraform/stages"
+)
+
+// PlatformStages are the stages to run to provision the infrastructure in Nutanix.
+var PlatformStages = []terraform.Stage{
+	stages.NewStage("nutanix", "cluster", []providers.Provider{providers.Nutanix}),
+	stages.NewStage("nutanix", "bootstrap", []providers.Provider{providers.Nutanix}, stages.WithNormalBootstrapDestroy()),
+}
--- a/data/data/nutanix/OWNERS
+++ b/data/data/nutanix/OWNERS
@@ -1,0 +1,7 @@
+# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+# This file just uses aliases defined in OWNERS_ALIASES.
+
+approvers:
+  - nutanix-approvers
+reviewers:
+  - nutanix-reviewers
--- a/data/data/nutanix/bootstrap/outputs.tf
+++ b/data/data/nutanix/bootstrap/outputs.tf
@@ -1,0 +1,3 @@
+output "bootstrap_ip" {
+  value = nutanix_virtual_machine.vm_bootstrap.nic_list_status.0.ip_endpoint_list.0.ip
+}
--- a/data/data/nutanix/bootstrap/variables.tf
+++ b/data/data/nutanix/bootstrap/variables.tf
@@ -1,0 +1,11 @@
+variable "image_id" {
+  type = string
+}
+
+variable "ocp_category_key_id" {
+  type = string
+}
+
+variable "ocp_category_value_owned_id" {
+  type = string
+}
--- a/data/data/nutanix/cluster/main.tf
+++ b/data/data/nutanix/cluster/main.tf
@@ -1,0 +1,78 @@
+locals {
+  description = "Created By OpenShift Installer"
+}
+
+provider "nutanix" {
+  wait_timeout = 60
+  username     = var.nutanix_username
+  password     = var.nutanix_password
+  endpoint     = var.nutanix_prism_central_address
+  port         = var.nutanix_prism_central_port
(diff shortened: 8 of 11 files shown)


data/data/nutanix/pre-bootstrap/main.tf
```diff
@@ -0,0 +1,43 @@
...
+  password     = var.password
+  endpoint     = var.prism_central
+  insecure     = var.insecure
+  port         = var.port
+}
+
+#TODO: Upload instead of query
+resource "nutanix_image" "rhcos" {
+  name        = var.nutanix_image
+  source_path  = var.nutanix_image_filepath
```
rvanderp3, line 17: tf-fmt is failing
```suggestion
  source_path = var.nutanix_image_filepath
```

data/data/nutanix/master/main.tf
```diff
@@ -0,0 +1,51 @@
...
+  wait_timeout = 60
+  username     = var.username
+  password     = var.password
+  endpoint     = var.prism_central
+  insecure     = var.insecure
+  port         = var.port
+}
+
+resource "nutanix_virtual_machine" "vm_master" {
+  count = var.master_count
```
rvanderp3, line 15: tf-fmt is failing due '=' not lining up with the rest of the block
```suggestion
  count                = var.master_count
```

data/data/nutanix/variables-nutanix.tf
```diff
@@ -0,0 +1,69 @@
+//////
+// Nutanix variables
+//////
+
+variable "prism_central" {
+  type        = string
+  description = "This is the Prism Central for the environment."
```
rvanderp3, line 7: Is this the hostname/IP for prism central?

rvanderp3: the description seems ambiguous

rvanderp3:
> /lgtm
> 
> My only comment here is that it would have been helpful to have `tools.go` in its own commit.

apologies, this comment is not relevant to this PR

rvanderp3:
@thunderboltsid would you mind taking a look at the merge conflict?

data/data/nutanix/bootstrap/main.tf
```diff
@@ -0,0 +1,63 @@
+locals {
+  description = "Created By OpenShift Installer"
+}
+
+provider "nutanix" {
+  wait_timeout = 60
```
rvanderp3, line 6: Is a 60 minute creation timeout reasonable?  I would expect the bootstrap to be provisioned within a few minutes at most if all is well.  There may be a rationale behind this timeout that I'm not aware of.

data/data/nutanix/master/main.tf
```diff
@@ -0,0 +1,50 @@
+locals {
+  description = "Created By OpenShift Installer"
+}
+
+provider "nutanix" {
+  wait_timeout = 60
```
rvanderp3, line 6: Is a 60 minute creation timeout reasonable?  I would expect the control plane nodes to be provisioned within a few minutes at most if all is well.  There may be a rationale behind this timeout that I'm not aware of.

data/data/nutanix/pre-bootstrap/main.tf
```diff
@@ -0,0 +1,36 @@
+locals {
+  description = "Created By OpenShift Installer"
+}
+
+provider "nutanix" {
+  wait_timeout = 60
```
rvanderp3, line 6: Is a 60 minute creation timeout reasonable?

data/data/nutanix/variables-nutanix.tf
```diff
@@ -0,0 +1,69 @@
...
+}
+
+variable "username" {
+  type        = string
+  description = "Prism Central user for the environment."
+}
+
+variable "password" {
+  type        = string
+  description = "Prism Central server password"
```
rvanderp3, line 23: Should this be `Prism Central user password`?

rvanderp3 (commented):
Found a nit and had a few questions on the terraform provider timeout value.

data/data/nutanix/master/main.tf
thunderboltsid: Good point. We haven't instrumented these stage times and as a result this is just a safe timeout. We can always reduce in future.

data/data/nutanix/bootstrap/main.tf
thunderboltsid: Good point. We haven't instrumented these stage times and as a result this is just a safe timeout. We can always reduce in future.

data/data/nutanix/pre-bootstrap/outputs.tf
```diff
@@ -0,0 +1,23 @@
...
+  value = nutanix_image.bootstrap_ignition.id
+}
+
+output "cluster_domain" {
+  value = var.cluster_domain
+}
+
+output "cluster_id" {
+  value = var.cluster_id
+}
```
staebler, line 15: Why are these outputs needed? They are not dependent on anything created by the stage.

data/data/nutanix/pre-bootstrap/main.tf
```diff
@@ -0,0 +1,36 @@
...
+  name        = var.nutanix_image
+  source_path = var.nutanix_image_filepath
+  description = local.description
+}
+
+resource "nutanix_category_key" "ocp_category_key" {
+  name = "openshift-${var.cluster_id}"
+}
+
+resource "nutanix_category_value" "ocp_category_value" {
```
staebler, line 23: Should we add the `description` field to these?

data/data/nutanix/master/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "control_plane_ips" {
+  value = nutanix_virtual_machine.vm_master.*.nic_list_status.0.ip_endpoint_list.0.ip
```
staebler, line 2: Let's use the new style for splat expressions.
```suggestion
  value = nutanix_virtual_machine.vm_master[*].nic_list_status[0].ip_endpoint_list[0].ip
```

pkg/terraform/stages/nutanix/stages.go
```diff
@@ -0,0 +1,14 @@
...
+	"github.com/openshift/installer/pkg/terraform"
+	"github.com/openshift/installer/pkg/terraform/providers"
+	"github.com/openshift/installer/pkg/terraform/stages"
+)
+
+// PlatformStages are the stages to run to provision the infrastructure in Nutanix.
+var PlatformStages = []terraform.Stage{
+	stages.NewStage("nutanix", "pre-bootstrap", []providers.Provider{providers.Nutanix}),
+	stages.NewStage("nutanix", "bootstrap", []providers.Provider{providers.Nutanix}, stages.WithNormalBootstrapDestroy()),
+	stages.NewStage("nutanix", "master", []providers.Provider{providers.Nutanix}),
```
staebler, line 13: Is there any technical reason why the "pre-bootstrap" and the "master" stages cannot be combined into a single stage which is run prior to the "bootstrap" stage? For example, will the nutanix terraform provider wait for the masters to fully ignite before the provider deems that master VMs to have been successfully created?

data/data/nutanix/variables-nutanix.tf
```diff
@@ -0,0 +1,69 @@
+//////
+// Nutanix variables
+//////
+
+variable "prism_central_address" {
```
staebler, line 5: All of the variables in this file should have the "nutanix_" prefix.

staebler (commented):
Please add OWNERS files at data/data/nutanix and pkg/terraform/stages/nutanix.

data/data/nutanix/pre-bootstrap/main.tf
```diff
@@ -0,0 +1,36 @@
...
+  name  = nutanix_category_key.ocp_category_key.id
+  value = "openshift-ipi-installations"
+}
+
+resource "nutanix_image" "bootstrap_ignition" {
```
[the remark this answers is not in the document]
thunderboltsid: I don't think that's a good idea. Image upload to PC can be a rather slow process and sometimes fail due to network disruption. Case in which I'd rather have it fail early on.

data/data/nutanix/pre-bootstrap/main.tf
```diff
@@ -0,0 +1,36 @@
...
+  name        = var.nutanix_bootstrap_ignition_image
+  source_path = var.nutanix_bootstrap_ignition_image
```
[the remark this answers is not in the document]
thunderboltsid: No. Changing this to name and absolute path to make it clearer.

pkg/terraform/stages/nutanix/stages.go
thunderboltsid: The masters nodes get created after the bootstrap node gets created. The three stages are reflective of phases of the install pre-bootstrap(create images on Prism Central), bootstrap (provision bootstrap node), master (provision master nodes).

data/data/nutanix/pre-bootstrap/main.tf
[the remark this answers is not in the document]
thunderboltsid: Would it make sense to delete the image also as part of a successful cluster install workflow from the installer?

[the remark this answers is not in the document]
staebler: Yes, the image should be deleted as part of a successful install as well as deleted as part of deleting a failed install.

pkg/terraform/stages/nutanix/stages.go
staebler: My question was whether there is any technical reason why the masters could not be created **before** the bootstrap node is created.

thunderboltsid: None really, added a change to merge those two. See if you like it more this way.

data/data/nutanix/pre-bootstrap/outputs.tf
thunderboltsid: Cleaned up some outputs (were useful for debug output)

data/data/nutanix/pre-bootstrap/main.tf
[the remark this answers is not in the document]
thunderboltsid: Moved the ignition image to the bootstrap phase so now it gets deleted when bootstrap gets destroyed. Ignition image takes less than a minute to upload so shouldn't be a problem.

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -337,6 +337,7 @@ then
 	copy_static_resources_for openstack
 	copy_static_resources_for ovirt
 	copy_static_resources_for vsphere
+	copy_static_resources_for nutanix
```
rvanderp3, line 340: can we pull this change in to it's own commit since its not related to the core terraform file addition?

staebler, line 340: This PR does not add any static Nutanix resources to copy. Let's hold off on adding the copy until there are files to copy.

data/data/nutanix/pre-bootstrap/main.tf
```diff
@@ -0,0 +1,74 @@
...
+    name  = nutanix_category_key.ocp_category_key.name
+    value = nutanix_category_value.ocp_category_value.value
+  }
+
+  guest_customization_cloud_init_user_data = base64encode(var.ignition_master)
+  nic_list {
+    subnet_uuid = var.nutanix_subnet_uuid
+  }
+
+  depends_on = [nutanix_image.rhcos]
```
staebler, line 73: Is this depends_on needed? Do we not get this for free since the image is used in the disk list?

pkg/terraform/stages/nutanix/stages.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"github.com/openshift/installer/pkg/terraform"
+	"github.com/openshift/installer/pkg/terraform/providers"
+	"github.com/openshift/installer/pkg/terraform/stages"
+)
+
+// PlatformStages are the stages to run to provision the infrastructure in Nutanix.
+var PlatformStages = []terraform.Stage{
+	stages.NewStage("nutanix", "pre-bootstrap", []providers.Provider{providers.Nutanix}),
```
staebler, line 11: Can we call this stage "cluster" instead of "pre-bootstrap"? The resources for the cluster are created by the stage.

data/data/nutanix/pre-bootstrap/main.tf
```diff
@@ -0,0 +1,74 @@
...
+provider "nutanix" {
+  wait_timeout = 60
+  username     = var.nutanix_username
+  password     = var.nutanix_password
+  endpoint     = var.nutanix_prism_central_address
+  port         = var.nutanix_prism_central_port
+}
+
+resource "nutanix_category_key" "ocp_category_key" {
+  name        = "openshift-${var.cluster_id}"
```
staebler, line 14: Is this similar to tags in AWS? If so, then we should stick with the common pattern of using "kubernetes.io/${var.cluster_id}/owned" as the key.

data/data/nutanix/pre-bootstrap/main.tf
thunderboltsid: Good catch; I don't think it is. TF should be able to infer this.

pkg/terraform/stages/nutanix/stages.go
thunderboltsid: Sure.

data/data/nutanix/pre-bootstrap/main.tf
thunderboltsid: That would not be a valid key format for Nutanix Prism Categories :)

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
thunderboltsid: Yeah, seems like this is required to copy openshift resources. Files to copy are already in the codebase: i.e. coredns and keepalived templates.

data/data/nutanix/pre-bootstrap/main.tf
staebler: How about `kubernetes-io-cluster-${var.cluster_id}` for a key, like is used for gcp?

data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
staebler: Ah, I am mistaken about what this is copying. Sorry for the noise.

thunderboltsid: No worries. As discussed offline, we can keep the commit here.

data/data/nutanix/pre-bootstrap/main.tf
thunderboltsid: That works.

staebler (commented):
/lgtm
/approve

PR #5633, opened 2022-02-10
Add nutanix-specific customizations to installconfig

thunderboltsid:
Add nutanix platform customizations in pkg/asset/installconfig.

PR #5638, opened 2022-02-10
Add cluster destroy logic for Nutanix clusters

thunderboltsid:
Add cluster destroy code to pkg/destroy

### DIFF
--- a/pkg/destroy/nutanix/nutanix.go
+++ b/pkg/destroy/nutanix/nutanix.go
@@ -1,0 +1,224 @@
+package nutanix
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	nutanixclientv3 "github.com/terraform-providers/terraform-provider-nutanix/client/v3"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/openshift/installer/pkg/destroy/providers"
+	installertypes "github.com/openshift/installer/pkg/types"
+	nutanixtypes "github.com/openshift/installer/pkg/types/nutanix"
+)
+
+const (
+	emptyFilter                = ""
+	expectedCategoryKeyFormat  = "kubernetes-io-cluster-%s"
+	expectedCategoryValueOwned = "owned"
+)
+
+// clusterUninstaller holds the various options for the cluster we want to delete.
+type clusterUninstaller struct {
+	clusterID string
+	infraID   string
+	v3Client  *nutanixclientv3.Client
+	logger    logrus.FieldLogger
+}
+
+// New returns an Nutanix destroyer from ClusterMetadata.
+func New(logger logrus.FieldLogger, metadata *installertypes.ClusterMetadata) (providers.Destroyer, error) {
+	v3Client, err := nutanixtypes.CreateNutanixClient(context.TODO(),
+		metadata.ClusterPlatformMetadata.Nutanix.PrismCentral,
+		metadata.ClusterPlatformMetadata.Nutanix.Port,
+		metadata.ClusterPlatformMetadata.Nutanix.Username,
+		metadata.ClusterPlatformMetadata.Nutanix.Password,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return &clusterUninstaller{
+		clusterID: metadata.ClusterID,
+		infraID:   metadata.InfraID,
+		v3Client:  v3Client,
+		logger:    logger,
+	}, nil
+}
+
+// Run is the entrypoint to start the uninstall process.
+func (o *clusterUninstaller) Run() (*installertypes.ClusterQuota, error) {
+	o.logger.Infof("Starting deletion of Nutanix infrastructure for Openshift cluster %q", o.infraID)
+	err := wait.PollImmediateInfinite(time.Second*30, o.destroyCluster)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to destroy cluster")
+	}
+
+	return nil, nil
+}
+
+func (o *clusterUninstaller) destroyCluster() (bool, error) {
+	cleanupFuncs := []struct {
+		name    string
+		execute func(*clusterUninstaller) error
+	}{
+		{name: "VMs", execute: cleanupVMs},
+		{name: "Images", execute: cleanupImages},
+		{name: "Categories", execute: cleanupCategories},
+	}
+
+	done := true
+	for _, cleanupFunc := range cleanupFuncs {
+		if err := cleanupFunc.execute(o); err != nil {
+			o.logger.Debugf("%s: %v", cleanupFunc.name, err)
+			done = false
+		}
+	}
+	return done, nil
+}
+
+func cleanupVMs(o *clusterUninstaller) error {
+	matchedVirtualMachineList := make([]*nutanixclientv3.VMIntentResource, 0)
+	allVMs, err := o.v3Client.V3.ListAllVM(emptyFilter)
+	if err != nil {
+		return err
+	}
+
+	for _, v := range allVMs.Entities {
+		if hasCategoryOwned(v.Metadata, expectedCategoryKey(o.infraID)) {
+			matchedVirtualMachineList = append(matchedVirtualMachineList, v)
+		}
+	}
+
+	if len(matchedVirtualMachineList) == 0 {
+		o.logger.Infof("No VMs found that require deletion for cluster %q", o.clusterID)
+	} else {
+		logToBeDeletedVMs(matchedVirtualMachineList, o.logger)
+		err := deleteVMs(o.v3Client.V3, matchedVirtualMachineList, o.logger)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func cleanupImages(o *clusterUninstaller) error {
+	allImages, err := o.v3Client.V3.ListAllImage(emptyFilter)
+	if err != nil {
+		return err
+	}
+
+	imageDeletionFailed := false
+	for _, image := range allImages.Entities {
+		if hasCategoryOwned(image.Metadata, expectedCategoryKey(o.infraID)) {
+			imageName := *image.Spec.Name
+			imageUUID := *image.Metadata.UUID
+			o.logger.Infof("Deleting image %q with UUID %q", imageName, imageUUID)
+			response, err := o.v3Client.V3.DeleteImage(imageUUID)
+			if err != nil {
+				o.logger.Errorf("Failed to delete image %q: %v", imageUUID, err)
+				imageDeletionFailed = true
+				continue
+			}
+
+			if err := nutanixtypes.WaitForTask(o.v3Client.V3, response.Status.ExecutionContext.TaskUUID.(string)); err != nil {
+				o.logger.Errorf("Failed to confirm image deletion %q: %v", imageUUID, err)
+				imageDeletionFailed = true
+			}
+		}
+	}
+
+	if imageDeletionFailed {
+		return fmt.Errorf("failed to cleanup images")
+	}
+
+	return nil
+}
+
+func cleanupCategories(o *clusterUninstaller) error {
+	expCatKey := expectedCategoryKey(o.infraID)
+	key, err := o.v3Client.V3.GetCategoryKey(expCatKey)
+	if err != nil {
+		if strings.Contains(err.Error(), "does not exist") {
+			//Already deleted
+			return nil
+		}
+		return err
+	}
+
+	values, err := o.v3Client.V3.ListCategoryValues(*key.Name, &nutanixclientv3.CategoryListMetadata{})
+	if err != nil {
+		return err
+	}
+
+	categoryDeletionFailed := false
+	for _, value := range values.Entities {
+		o.logger.Infof("Deleting category value : %s", *value.Value)
+		err := o.v3Client.V3.DeleteCategoryValue(expCatKey, *value.Value)
+		if err != nil {
+			o.logger.Errorf("Failed to delete category value %q: %v", *value.Value, err)
+			categoryDeletionFailed = true
+		}
+	}
+
+	o.logger.Infof("Deleting category key : %s", expCatKey)
+	err = o.v3Client.V3.DeleteCategoryKey(expCatKey)
+	if err != nil {
+		o.logger.Errorf("Failed to delete category key %q: %v", expCatKey, err)
+		categoryDeletionFailed = true
+	}
+
+	if categoryDeletionFailed {
+		return fmt.Errorf("failed to delete category")
+	}
+
+	return nil
+}
+
+func deleteVMs(clientV3 nutanixclientv3.Service, vms []*nutanixclientv3.VMIntentResource, l logrus.FieldLogger) error {
+	taskUUIDs := make([]string, 0)
+	vmDeletionFailed := false
+	for _, vm := range vms {
+		l.Infof("Deleting VM %s with ID %s", *vm.Spec.Name, *vm.Metadata.UUID)
+		response, err := clientV3.DeleteVM(*vm.Metadata.UUID)
+		if err != nil {
+			l.Errorf("Failed to delete VM %q: %v", *vm.Metadata.UUID, err)
+			vmDeletionFailed = true
+			continue
+		}
+
+		taskUUIDs = append(taskUUIDs, response.Status.ExecutionContext.TaskUUID.(string))
+	}
+
+	err := nutanixtypes.WaitForTasks(clientV3, taskUUIDs)
+	if err != nil {
+		l.Errorf("Failed to confirm deletion of VMs: %v", err)
+		vmDeletionFailed = true
+	}
+
+	if vmDeletionFailed {
+		return fmt.Errorf("failed to delete VMs")
+	}
+
+	return nil
+}
+
+func logToBeDeletedVMs(vms []*nutanixclientv3.VMIntentResource, l logrus.FieldLogger) {
+	l.Info("Virtual machines scheduled to be deleted: ")
+	for _, vm := range vms {
+		l.Infof("- %s", *vm.Spec.Name)
+	}
+}
+
+func expectedCategoryKey(infraID string) string {
+	return fmt.Sprintf(expectedCategoryKeyFormat, infraID)
+}
+
+func hasCategoryOwned(metadata *nutanixclientv3.Metadata, expectedCategoryKey string) bool {
+	value, keyExists := metadata.Categories[expectedCategoryKey]
+	return keyExists && value == expectedCategoryValueOwned
+}
--- a/pkg/types/nutanix/helpers.go
+++ b/pkg/types/nutanix/helpers.go
@@ -11,7 +11,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/kdomanski/iso9660"
 	"github.com/pkg/errors"
-	nutanixClientV3 "github.com/terraform-providers/terraform-provider-nutanix/client/v3"
+	nutanixclientv3 "github.com/terraform-providers/terraform-provider-nutanix/client/v3"
 	"github.com/terraform-providers/terraform-provider-nutanix/utils"
 )
 
@@ -48,7 +48,6 @@ func CreateBootstrapISO(infraID, userData string) (string, error) {
 	metaObj := &metadataCloudInit{
 		UUID: id.String(),
 	}
-
 	metadata, err := json.Marshal(metaObj)
 	if err != nil {
 		return "", errors.Wrap(err, fmt.Sprintf("failed to marshal metadata struct to json"))
@@ -97,9 +96,10 @@ func CreateBootstrapISO(infraID, userData string) (string, error) {
 	return fullISOFile, nil
 }
 
-func waitForTasks(clientV3 nutanixClientV3.Service, taskUUIDs []string) error {
+// WaitForTasks is a wrapper for WaitForTask
+func WaitForTasks(clientV3 nutanixclientv3.Service, taskUUIDs []string) error {
 	for _, t := range taskUUIDs {
-		err := waitForTask(clientV3, t)
+		err := WaitForTask(clientV3, t)
 		if err != nil {
 			return err
 		}
@@ -107,7 +107,8 @@ func waitForTasks(clientV3 nutanixClientV3.Service, taskUUIDs []string) error {
 	return nil
 }
 
-func waitForTask(clientV3 nutanixClientV3.Service, taskUUID string) error {
+// WaitForTask waits until a queued task has been finished or timeout has been reached
+func WaitForTask(clientV3 nutanixclientv3.Service, taskUUID string) error {
 	finished := false
 	var err error
 	for start := time.Now(); time.Since(start) < timeout; {
@@ -127,7 +128,7 @@ func waitForTask(clientV3 nutanixClientV3.Service, taskUUID string) error {
 	return nil
 }
 
-func isTaskFinished(clientV3 nutanixClientV3.Service, taskUUID string) (bool, error) {
+func isTaskFinished(clientV3 nutanixclientv3.Service, taskUUID string) (bool, error) {
 	isFinished := map[string]bool{
 		"QUEUED":    false,
 		"RUNNING":   false,
@@ -140,11 +141,10 @@ func isTaskFinished(clientV3 nutanixClientV3.Service, taskUUID string) (bool, er
 	if val, ok := isFinished[status]; ok {
 		return val, nil
 	}
-	return false, errors.Errorf("Retrieved unexpected task status: %s", status)
-
+	return false, errors.Errorf("retrieved unexpected task status: %s", status)
 }
 
-func getTaskStatus(clientV3 nutanixClientV3.Service, taskUUID string) (string, error) {
+func getTaskStatus(clientV3 nutanixclientv3.Service, taskUUID string) (string, error) {
 	v, err := clientV3.GetTask(taskUUID)
 
 	if err != nil {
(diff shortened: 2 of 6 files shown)


pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,224 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+
+	"github.com/openshift/installer/pkg/destroy/providers"
+	installertypes "github.com/openshift/installer/pkg/types"
+	nutanixtypes "github.com/openshift/installer/pkg/types/nutanix"
+	nutanixClientV3 "github.com/terraform-providers/terraform-provider-nutanix/client/v3"
```
rvanderp3, line 14: package names should be lower case

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,224 @@
...
+				if hasCategoryAssigned(i.Metadata, ec, o.Logger) {
+					if i.Spec.Description != nil {
+						if *i.Spec.Description == createdByOpenshiftInstallerDesc {
+							found = true
+							o.Logger.Infof("Deleting image %s with UUID %s", imageName, imageUUID)
+							response, err := clientV3.DeleteImage(imageUUID)
+							if err != nil {
+								return err
+							}
+							nutanixtypes.WaitForTask(clientV3, response.Status.ExecutionContext.TaskUUID.(string))
```
rvanderp3, line 140: the error from the call to `WaitForTask` is not being handled

pkg/types/nutanix/helpers.go
```diff
@@ -0,0 +1,146 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/kdomanski/iso9660"
+	"github.com/pkg/errors"
+	nutanixClientV3 "github.com/terraform-providers/terraform-provider-nutanix/client/v3"
```
rvanderp3, line 13: package name should be lower case

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,224 @@
...
+		ClusterID: metadata.ClusterID,
+		InfraID:   metadata.InfraID,
+		V3Client:  v3Client,
+		Logger:    logger,
+	}, nil
+}
+
+// Run is the entrypoint to start the uninstall process.
+func (o *ClusterUninstaller) Run() (*installertypes.ClusterQuota, error) {
+	ec := createExpectedCategory(o.InfraID)
```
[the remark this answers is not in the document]
thunderboltsid: yes & yes.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,227 @@
...
+						if *i.Spec.Description == createdByOpenshiftInstallerDesc {
+							found = true
+							o.Logger.Infof("Deleting image %s with UUID %s", imageName, imageUUID)
+							response, err := clientV3.DeleteImage(imageUUID)
+							if err != nil {
+								return err
+							}
+							err = nutanixtypes.WaitForTask(clientV3, response.Status.ExecutionContext.TaskUUID.(string))
+							if err != nil {
+								o.Logger.Warnf("error while deleting images: %s", err)
```
rvanderp3, line 142: Should this error propagate back out instead of being logged?  It seems like this could lead to images being leaked if the image deletion never completes.

pkg/types/nutanix/client.go
```diff
@@ -0,0 +1,34 @@
...
+	ctx, cancel := context.WithTimeout(ctx, 60*time.Second)
+	defer cancel()
+
+	cred := nutanixClient.Credentials{
+		URL:      fmt.Sprintf("%s:%s", prismCentral, port),
+		Username: username,
+		Password: password,
+		Port:     port,
+		Endpoint: prismCentral,
+		Insecure: false,
```
rvanderp3, line 24: this file appears to be in conflict with what is in https://github.com/openshift/installer/pull/5633/commits/8150c70a911381dfefdf621769b14a8afb1c3c22#diff-aa325f2e2fa65a7eb01b9fb635844901f1d27126471742a88661c666feb83f8aR22 .  Is the intention to use client.go from 5633?

thunderboltsid: Yes, when 5633 is merged, we can resolve this.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,227 @@
...
+
+func hasCategoryAssigned(metadata *nutanixclientv3.Metadata, ec *expectedCategory, l logrus.FieldLogger) bool {
+	if metadata.Categories != nil {
+		if val, ok := metadata.Categories[ec.Key]; ok {
+			if val == ec.Value {
+				return true
+			}
+		}
+	}
+	return false
```
staebler, line 226:
```suggestion
	value, keyExists := metadata.Categories[ec.Key]
	return keyExists && value == ec.Value
```

staebler (commented):
Two main concerns:
1. We should have an authoritative way of finding resources owned by the cluster. It seems like the category is that way.
2. The destroyer should continue deleting other resources when there are errors encountered. And the destroyer should retry deleting resource indefinitely. For example, let's say the destroyer needs to delete resources A, B, C, D, and E. If there are errors deleting B and C, the destroyer should still delete D and E before going back to retry the deletes of B and C. The retries should happen in a loop with a delay between attempts until the user stop execution of the destroyer.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,227 @@
...
+	if err != nil {
+		return err
+	}
+	allVMs := allVMsRaw.Entities
+	for _, v := range allVMs {
+		vmName := *v.Spec.Name
+		if strings.HasPrefix(vmName, o.InfraID) {
+			if hasCategoryAssigned(v.Metadata, ec, o.Logger) {
+				if v.Spec.Description != nil {
+					if *v.Spec.Description == createdByOpenshiftInstallerDesc {
```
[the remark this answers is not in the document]
thunderboltsid: You are right about the description, I think we can get rid of it. The prefix+category should be sufficient.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,227 @@
...
+	allImages := allImagesRaw.Entities
+	for _, expectedImageName := range expectedImageNames {
+		found = false
+		for _, i := range allImages {
+			imageName := *i.Spec.Name
+			imageUUID := *i.Metadata.UUID
+			if imageName == expectedImageName {
+				if hasCategoryAssigned(i.Metadata, ec, o.Logger) {
+					if i.Spec.Description != nil {
+						if *i.Spec.Description == createdByOpenshiftInstallerDesc {
```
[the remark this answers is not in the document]
thunderboltsid: You are right about the description, I think we can get rid of it. The prefix+category should be sufficient.

pkg/destroy/nutanix/nutanix.go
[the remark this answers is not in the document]
staebler: Why do we need the prefix? If the VM has the category, it is owned by the cluster and should be deleted when the cluster is destroyed.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,227 @@
...
+	ClusterID string
+	InfraID   string
+	V3Client  *nutanixclientv3.Client
+	Logger    logrus.FieldLogger
+}
+
+type expectedCategory struct {
+	Key   string
+	Value string
+}
```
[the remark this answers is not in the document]
thunderboltsid: nope; not really. made them unexported.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,227 @@
...
+func cleanupVMs(clientV3 nutanixclientv3.Service, ec *expectedCategory, o *ClusterUninstaller) error {
+	matchedVirtualMachineList := make([]*nutanixclientv3.VMIntentResource, 0)
+	allVMsRaw, err := clientV3.ListAllVM(emptyFilter)
+	if err != nil {
+		return err
+	}
+	allVMs := allVMsRaw.Entities
+	for _, v := range allVMs {
+		vmName := *v.Spec.Name
+		if strings.HasPrefix(vmName, o.InfraID) {
```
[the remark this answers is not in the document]
thunderboltsid: It's a good first check. I'm inclined to keep it here for now.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,227 @@
...
+				if v.Spec.Description != nil {
+					if *v.Spec.Description == createdByOpenshiftInstallerDesc {
+						matchedVirtualMachineList = append(matchedVirtualMachineList, v)
+					}
+				}
+			}
+		}
+	}
+
+	amountOfTBD := len(matchedVirtualMachineList)
```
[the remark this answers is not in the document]
thunderboltsid: to be deleted.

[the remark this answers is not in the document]
thunderboltsid: Removing the variable as it's not necessary.

thunderboltsid:
> Two main concerns:
> 
> 1. We should have an authoritative way of finding resources owned by the cluster. It seems like the category is that way.
Agreed. Removed the description and prefix based checks.
> 2. The destroyer should continue deleting other resources when there are errors encountered. And the destroyer should retry deleting resource indefinitely. For example, let's say the destroyer needs to delete resources A, B, C, D, and E. If there are errors deleting B and C, the destroyer should still delete D and E before going back to retry the deletes of B and C. The retries should happen in a loop with a delay between attempts until the user stop execution of the destroyer.
Did some refactoring. Take a look and see if this looks better.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+}
+
+func deleteVMs(clientV3 nutanixclientv3.Service, tbd []*nutanixclientv3.VMIntentResource, l logrus.FieldLogger) error {
+	taskUUIDs := make([]string, 0)
+	for _, v := range tbd {
+		l.Infof("Deleting vm %s with ID %s", *v.Spec.Name, *v.Metadata.UUID)
+		response, err := clientV3.DeleteVM(*v.Metadata.UUID)
+		if err != nil {
+			return err
```
rvanderp3, line 202: I think staebler was looking to have the deletion of VMs to continue even if a failure was encountered.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,223 @@
...
+func logToBeDeletedVMs(tbd []*nutanixclientv3.VMIntentResource, l logrus.FieldLogger) {
+	l.Info("VMs scheduled to be deleted: ")
+	for _, v := range tbd {
+		l.Infof("- %s", *v.Spec.Name)
+	}
+}
+
+func deleteVMs(clientV3 nutanixclientv3.Service, tbd []*nutanixclientv3.VMIntentResource, l logrus.FieldLogger) error {
+	taskUUIDs := make([]string, 0)
+	for _, v := range tbd {
```
rvanderp3, line 198: `tbd` is ambiguous, can this be changed to be more clear?

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+func (o *clusterUninstaller) destroyCluster() (bool, error) {
+	cleanupFuncs := []struct {
+		name    string
+		execute func(*clusterUninstaller) error
+	}{
+		{name: "VMs", execute: cleanupVMs},
+		{name: "Images", execute: cleanupImages},
+		{name: "Categories", execute: cleanupCategories},
```
staebler, line 77: What is the behavior if the destroyer attempts to delete the categories while there are still VMs and/or images using the categories? Will the delete fail? If the delete succeeds, will the categories be removed from the remaining VMs and images? In other words, do we need to wait until the VMs and images are deleted before attempting to delete the categories?

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,226 @@
...
+	if err != nil {
+		return err
+	}
+
+	for _, expectedImageName := range expectedImageNames {
+		found = false
+		for _, i := range allImages.Entities {
+			imageName := *i.Spec.Name
+			imageUUID := *i.Metadata.UUID
+			if imageName == expectedImageName && hasCategoryAssigned(i.Metadata, ec) {
```
staebler, line 136: For the same reason that we do not want the destroyer to consider the name of the VM when determining whether to delete a VM, the destroyer should not consider the name of the image. If the image has the expected category, the destroyer should delete it.

pkg/destroy/nutanix/nutanix.go
thunderboltsid: No, categories can't be deleted as long as resources with the categories exist.

staebler: Perfect. That works then. The delete of the categories will fail, which is fine.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+func cleanupImages(o *clusterUninstaller) error {
+	ec := createExpectedCategory(o.infraID)
+	allImages, err := o.v3Client.V3.ListAllImage(emptyFilter)
+	if err != nil {
+		return err
+	}
+
+	for _, i := range allImages.Entities {
```
staebler, line 123: The variable name `i` is confusing to use in a `range`.
```suggestion
	for _, image := range allImages.Entities {
```

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,211 @@
...
+		imageName := *i.Spec.Name
+		imageUUID := *i.Metadata.UUID
+		if hasCategoryAssigned(i.Metadata, ec) {
```
staebler, line 126:
```suggestion
		if hasCategoryAssigned(i.Metadata, ec) {
			imageName := *i.Spec.Name
			imageUUID := *i.Metadata.UUID
```

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,211 @@
...
+			o.logger.Infof("deleting image %s with UUID %s", imageName, imageUUID)
+			response, err := o.v3Client.V3.DeleteImage(imageUUID)
+			if err != nil {
+				o.logger.Errorf("failed to delete image %s: %q", imageUUID, err)
+				break
```
staebler, line 131: Don't break on an error. You can continue. But you need to store the error to return.
```suggestion
				continue
```

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,211 @@
...
+			}
+
+			if err = nutanixtypes.WaitForTask(o.v3Client.V3, response.Status.ExecutionContext.TaskUUID.(string)); err != nil {
+				o.logger.Errorf("failed to confirm image deletion %s: %q", imageUUID, err)
+			}
+
+			break
```
staebler, line 138: This seems wrong. We don't want to break after the first image.
```suggestion
```

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+func createExpectedCategory(infraID string) *expectedCategory {
+	return &expectedCategory{
+		key:    fmt.Sprintf(expectedCategoryKeyFormat, infraID),
+		values: []string{expectedCategoryValueOwned, expectedCategoryValueShared},
+	}
+}
+
+func logToBeDeletedVMs(tbd []*nutanixclientv3.VMIntentResource, l logrus.FieldLogger) {
```
staebler, line 175: Please use a better variable name than `tbd`.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,211 @@
...
+	for _, v := range tbd {
+		l.Infof("- %s", *v.Spec.Name)
+	}
+}
+
+func deleteVMs(clientV3 nutanixclientv3.Service, vms []*nutanixclientv3.VMIntentResource, l logrus.FieldLogger) error {
+	taskUUIDs := make([]string, 0)
+	for _, vm := range vms {
+		l.Infof("deleting VM %s with ID %s", *vm.Spec.Name, *vm.Metadata.UUID)
+		response, err := clientV3.DeleteVM(*vm.Metadata.UUID)
```
staebler, line 186: We need to save this error to return it after attempting to delete all of the VMs. As is, if there is a failure in a call to `DeleteVM`, the `deleteVMs` function still returns nil.

pkg/destroy/nutanix/nutanix.go
thunderboltsid: indeed, this does not make sense.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+func cleanupImages(o *clusterUninstaller) error {
+	ec := createExpectedCategory(o.infraID)
+	allImages, err := o.v3Client.V3.ListAllImage(emptyFilter)
+	if err != nil {
+		return err
+	}
+
+	allErrs := field.ErrorList{}
```
staebler, line 124: Don't use field errors. That is for validation. The errors are logged when encountered already, so the function does not need to return all of the errors. The function just needs to return **an** error.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	for _, v := range allVMs.Entities {
+		if hasCategoryAssigned(v.Metadata, ec) {
+			matchedVirtualMachineList = append(matchedVirtualMachineList, v)
+		}
+	}
+
+	if len(matchedVirtualMachineList) == 0 {
+		o.logger.Infof("no VMs found that require deletion for cluster %s", o.clusterID)
```
staebler, line 106: Log messages should start with a capital letter.
```suggestion
		o.logger.Infof("No VMs found that require deletion for cluster %s", o.clusterID)
```

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,229 @@
...
+
+	allErrs := field.ErrorList{}
+	for _, image := range allImages.Entities {
+		if hasCategoryAssigned(image.Metadata, ec) {
+			imageName := *image.Spec.Name
+			imageUUID := *image.Metadata.UUID
+			o.logger.Infof("deleting image %s with UUID %s", imageName, imageUUID)
+			response, err := o.v3Client.V3.DeleteImage(imageUUID)
+			if err != nil {
+				o.logger.Errorf("failed to delete image %s: %q", imageUUID, err)
```
staebler, line 132: Use `%v` for formatting the error. You don't want to wrap the entire error in quotes. The `%q` would be good for the image UUID, however.
```suggestion
				o.logger.Errorf("failed to delete image %q: %v", imageUUID, err)
```

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,229 @@
...
+		allErrs = append(allErrs, &field.Error{Type: field.ErrorTypeInternal, Detail: err.Error()})
+	}
+
+	return allErrs.ToAggregate()
+}
+
+func createExpectedCategory(infraID string) *expectedCategory {
+	return &expectedCategory{
+		key:    fmt.Sprintf(expectedCategoryKeyFormat, infraID),
+		values: []string{expectedCategoryValueOwned, expectedCategoryValueShared},
```
staebler, line 181: The "shared" value should not be included. We do not want to delete resources with the "shared" value.
```suggestion
		values: []string{expectedCategoryValueOwned},
```

thunderboltsid: We still want to delete category value as that is a prerequisite for deleting the category key.

thunderboltsid: We will look for all values of the category and delete them all instead of deleting one or two.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	if len(matchedVirtualMachineList) == 0 {
+		o.logger.Infof("No VMs found that require deletion for cluster %q", o.clusterID)
+	} else {
+		logToBeDeletedVMs(matchedVirtualMachineList, o.logger)
+		err := deleteVMs(o.v3Client.V3, matchedVirtualMachineList, o.logger)
+		if err != nil {
+			return err
+		}
```
staebler, line 104: The destroyer should continue to delete the other VMs despite receiving an error deleting one of the VMs.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,208 @@
...
+	for _, image := range allImages.Entities {
+		if hasCategoryOwned(image.Metadata, expectedCategoryKey(o.infraID)) {
+			imageName := *image.Spec.Name
+			imageUUID := *image.Metadata.UUID
+			o.logger.Infof("Deleting image %q with UUID %q", imageName, imageUUID)
+			response, err := o.v3Client.V3.DeleteImage(imageUUID)
+			if err != nil {
+				o.logger.Errorf("Failed to delete image %q: %v", imageUUID, err)
+				continue
+			}
```
staebler, line 124: An error deleting an image **must** cause an error to be returned from `cleanupImages`.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,208 @@
...
+
+			if err = nutanixtypes.WaitForTask(o.v3Client.V3, response.Status.ExecutionContext.TaskUUID.(string)); err != nil {
```
staebler, line 126: An error deleting image 1 will be overwritten by a successful delete of image 2, causing `cleanupImages` to return successfully.

pkg/destroy/nutanix/nutanix.go
thunderboltsid: This method call takes a list of VMs to delete. if one of them fails, it returns an error and `cleanupVMs` gets called again.

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,224 @@
...
+			imageUUID := *image.Metadata.UUID
+			o.logger.Infof("Deleting image %q with UUID %q", imageName, imageUUID)
+			response, err := o.v3Client.V3.DeleteImage(imageUUID)
+			if err != nil {
+				o.logger.Errorf("Failed to delete image %q: %v", imageUUID, err)
+				imageDeletionFailed = true
+				continue
+			}
+
+			if err = nutanixtypes.WaitForTask(o.v3Client.V3, response.Status.ExecutionContext.TaskUUID.(string)); err != nil {
```
staebler, line 128:
```suggestion
			if err := nutanixtypes.WaitForTask(o.v3Client.V3, response.Status.ExecutionContext.TaskUUID.(string)); err != nil {
```

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,224 @@
...
+			vmDeletionFailed = true
+			continue
+		}
+
+		taskUUIDs = append(taskUUIDs, response.Status.ExecutionContext.TaskUUID.(string))
+	}
+
+	err := nutanixtypes.WaitForTasks(clientV3, taskUUIDs)
+	if err != nil {
+		l.Errorf("failed to confirm deletion of VMs: %v", err)
```
staebler, line 199:
```suggestion
		l.Errorf("Failed to confirm deletion of VMs: %v", err)
```

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,224 @@
...
+		if err != nil {
+			o.logger.Errorf("failed to delete category value %q: %v", *value.Value, err)
+			categoryDeletionFailed = true
+		}
+	}
+
+	o.logger.Infof("Deleting category key : %s", expCatKey)
+	err = o.v3Client.V3.DeleteCategoryKey(expCatKey)
+	if err != nil {
+		o.logger.Errorf("failed to delete category key %q: %v", expCatKey, err)
```
staebler, line 171:
```suggestion
		o.logger.Errorf("Failed to delete category key %q: %v", expCatKey, err)
```

pkg/destroy/nutanix/nutanix.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return err
+	}
+
+	categoryDeletionFailed := false
+	for _, value := range values.Entities {
+		o.logger.Infof("Deleting category value : %s", *value.Value)
+		err := o.v3Client.V3.DeleteCategoryValue(expCatKey, *value.Value)
+		if err != nil {
+			o.logger.Errorf("failed to delete category value %q: %v", *value.Value, err)
```
staebler, line 163:
```suggestion
			o.logger.Errorf("Failed to delete category value %q: %v", *value.Value, err)
```

staebler (commented):
Just a few very minor nits.

pkg/destroy/nutanix/nutanix.go
thunderboltsid: Given err has already been declared before in the scope, what's the rationale for using := instead of = ?

staebler: It reduces the scope of `err` to only the block where it is used. We don't have to worry about whether the `err` in the outer scope is meaningful in some way where overwriting it has side effects.

staebler (commented):
/lgtm
/approve

PR #5639, opened 2022-02-10
Bug 2046181: baremetal: reduce API timeout to 15 minutes

dtantsur:
There is no point in waiting 1 hours for Ironic API to start: even given
the bootstrap VM start-up, it takes much less. Currently, users have to
wait 1 hour to learn that API failed to start for some reason (e.g.
because the network config has an error).


kirankt (approved):
/lgtm

PR #5641, opened 2022-02-10
Azure disk encryption support

jhixson74:
Add support for encryption at host and disk encryption with a disk encryption set ID. 

https://issues.redhat.com/browse/CORS-1889
https://issues.redhat.com/browse/CORS-1890
https://issues.redhat.com/browse/CORS-1891

### DIFF
--- a/data/data/azure/bootstrap/main.tf
+++ b/data/data/azure/bootstrap/main.tf
@@ -200,6 +200,7 @@ resource "azurerm_linux_virtual_machine" "bootstrap" {
   # included here because it is required by the Azure ARM API.
   admin_password                  = "NotActuallyApplied!"
   disable_password_authentication = false
+  encryption_at_host_enabled      = var.azure_master_encryption_at_host_enabled
 
   identity {
     type         = "UserAssigned"
@@ -207,10 +208,11 @@ resource "azurerm_linux_virtual_machine" "bootstrap" {
   }
 
   os_disk {
-    name                 = "${var.cluster_id}-bootstrap_OSDisk" # os disk name needs to match cluster-api convention
-    caching              = "ReadWrite"
-    storage_account_type = var.azure_master_root_volume_type
-    disk_size_gb         = 100
+    name                   = "${var.cluster_id}-bootstrap_OSDisk" # os disk name needs to match cluster-api convention
+    caching                = "ReadWrite"
+    storage_account_type   = var.azure_master_root_volume_type
+    disk_size_gb           = 100
+    disk_encryption_set_id = var.azure_master_disk_encryption_set_id
   }
 
   source_image_id = var.vm_image
--- a/data/data/azure/cluster/main.tf
+++ b/data/data/azure/cluster/main.tf
@@ -29,26 +29,28 @@ provider "azureprivatedns" {
 }
 
 module "master" {
-  source                 = "./master"
-  resource_group_name    = var.resource_group_name
-  cluster_id             = var.cluster_id
-  region                 = var.azure_region
-  availability_zones     = var.azure_master_availability_zones
-  vm_size                = var.azure_master_vm_type
-  vm_image               = var.vm_image
-  identity               = var.identity
-  ignition               = var.ignition_master
-  elb_backend_pool_v4_id = var.elb_backend_pool_v4_id
-  elb_backend_pool_v6_id = var.elb_backend_pool_v6_id
-  ilb_backend_pool_v4_id = var.ilb_backend_pool_v4_id
-  ilb_backend_pool_v6_id = var.ilb_backend_pool_v6_id
-  subnet_id              = var.master_subnet_id
-  instance_count         = var.master_count
-  storage_account_name   = var.storage_account_name
-  os_volume_type         = var.azure_master_root_volume_type
-  os_volume_size         = var.azure_master_root_volume_size
-  private                = var.azure_private
-  outbound_udr           = var.azure_outbound_user_defined_routing
+  source                     = "./master"
+  resource_group_name        = var.resource_group_name
+  cluster_id                 = var.cluster_id
+  region                     = var.azure_region
+  availability_zones         = var.azure_master_availability_zones
+  vm_size                    = var.azure_master_vm_type
+  disk_encryption_set_id     = var.azure_master_disk_encryption_set_id
+  encryption_at_host_enabled = var.azure_master_encryption_at_host_enabled
+  vm_image                   = var.vm_image
+  identity                   = var.identity
+  ignition                   = var.ignition_master
+  elb_backend_pool_v4_id     = var.elb_backend_pool_v4_id
+  elb_backend_pool_v6_id     = var.elb_backend_pool_v6_id
+  ilb_backend_pool_v4_id     = var.ilb_backend_pool_v4_id
+  ilb_backend_pool_v6_id     = var.ilb_backend_pool_v6_id
+  subnet_id                  = var.master_subnet_id
+  instance_count             = var.master_count
+  storage_account_name       = var.storage_account_name
+  os_volume_type             = var.azure_master_root_volume_type
+  os_volume_size             = var.azure_master_root_volume_size
+  private                    = var.azure_private
+  outbound_udr               = var.azure_outbound_user_defined_routing
 
   use_ipv4 = var.use_ipv4
   use_ipv6 = var.use_ipv6
--- a/data/data/azure/cluster/master/variables.tf
+++ b/data/data/azure/cluster/master/variables.tf
@@ -16,6 +16,17 @@ variable "vm_size" {
   type = string
 }
 
+variable "disk_encryption_set_id" {
+  type        = string
+  default     = null
+  description = "The ID of the Disk Encryption Set which should be used to encrypt OS disk."
+}
+
+variable "encryption_at_host_enabled" {
+  type        = bool
+  description = "Enables encryption at the VM host."
+}
+
 variable "vm_image" {
   type        = string
   description = "The resource id of the vm image used for masters."
--- a/data/data/azure/variables-azure.tf
+++ b/data/data/azure/variables-azure.tf
@@ -23,6 +23,17 @@ variable "azure_master_vm_type" {
   description = "Instance type for the master node(s). Example: `Standard_D8s_v3`."
 }
 
+variable "azure_master_disk_encryption_set_id" {
+  type = string
+  default = null
+  description = "The ID of the Disk Encryption Set which should be used to encrypt OS disk for the master node(s)."
+}
+
+variable "azure_master_encryption_at_host_enabled" {
+  type = bool
+  description = "Enables encryption at the VM host for the master node(s)."
+}
+
 variable "azure_extra_tags" {
   type = map(string)
 
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -190,9 +190,33 @@ spec:
                       description: Azure is the configuration used when installing
                         on Azure.
                       properties:
+                        encryptionAtHost:
+                          description: EncryptionAtHost enables encryption at the
+                            VM host.
+                          type: boolean
                         osDisk:
                           description: OSDisk defines the storage for instance.
                           properties:
+                            diskEncryptionSet:
+                              description: DiskEncryptionSet defines a disk encryption
+                                set.
+                              properties:
+                                name:
+                                  description: Name is the name of the disk encryption
+                                    set.
+                                  type: string
+                                resourceGroup:
+                                  description: ResourceGroup defines the Azure resource
+                                    group used by the disk encryption set.
+                                  type: string
+                                subscriptionId:
+                                  description: SubscriptionID defines the Azure subscription
+                                    the disk encryption set is in.
+                                  type: string
+                              required:
+                              - name
+                              - resourceGroup
+                              type: object
                             diskSizeGB:
                               description: DiskSizeGB defines the size of disk in
                                 GB.
@@ -201,9 +225,9 @@ spec:
                               type: integer
                             diskType:
                               description: DiskType defines the type of disk. The
-                                valid values are Standard_LRS, Premium_LRS, StandardSSD_LRS
+                                valid values are Standard_LRS, Premium_LRS, StandardSSD_LRS.
                                 For control plane nodes, the valid values are Premium_LRS
-                                and StandardSSD_LRS. Default is Premium_LRS
+                                and StandardSSD_LRS. Default is Premium_LRS.
                               enum:
                               - Standard_LRS
                               - Premium_LRS
@@ -682,9 +706,33 @@ spec:
                     description: Azure is the configuration used when installing on
                       Azure.
                     properties:
+                      encryptionAtHost:
+                        description: EncryptionAtHost enables encryption at the VM
+                          host.
+                        type: boolean
                       osDisk:
                         description: OSDisk defines the storage for instance.
                         properties:
+                          diskEncryptionSet:
+                            description: DiskEncryptionSet defines a disk encryption
+                              set.
+                            properties:
+                              name:
+                                description: Name is the name of the disk encryption
+                                  set.
+                                type: string
+                              resourceGroup:
+                                description: ResourceGroup defines the Azure resource
+                                  group used by the disk encryption set.
+                                type: string
+                              subscriptionId:
+                                description: SubscriptionID defines the Azure subscription
+                                  the disk encryption set is in.
+                                type: string
+                            required:
+                            - name
+                            - resourceGroup
+                            type: object
                           diskSizeGB:
                             description: DiskSizeGB defines the size of disk in GB.
                             format: int32
@@ -692,9 +740,9 @@ spec:
                             type: integer
                           diskType:
                             description: DiskType defines the type of disk. The valid
-                              values are Standard_LRS, Premium_LRS, StandardSSD_LRS
+                              values are Standard_LRS, Premium_LRS, StandardSSD_LRS.
                               For control plane nodes, the valid values are Premium_LRS
-                              and StandardSSD_LRS. Default is Premium_LRS
+                              and StandardSSD_LRS. Default is Premium_LRS.
                             enum:
                             - Standard_LRS
                             - Premium_LRS
@@ -1445,9 +1493,33 @@ spec:
                       used when installing on Azure for machine pools which do not
                       define their own platform configuration.
                     properties:
+                      encryptionAtHost:
+                        description: EncryptionAtHost enables encryption at the VM
+                          host.
+                        type: boolean
                       osDisk:
                         description: OSDisk defines the storage for instance.
                         properties:
+                          diskEncryptionSet:
+                            description: DiskEncryptionSet defines a disk encryption
+                              set.
+                            properties:
+                              name:
+                                description: Name is the name of the disk encryption
+                                  set.
+                                type: string
+                              resourceGroup:
+                                description: ResourceGroup defines the Azure resource
+                                  group used by the disk encryption set.
+                                type: string
+                              subscriptionId:
+                                description: SubscriptionID defines the Azure subscription
+                                  the disk encryption set is in.
+                                type: string
+                            required:
+                            - name
+                            - resourceGroup
+                            type: object
                           diskSizeGB:
                             description: DiskSizeGB defines the size of disk in GB.
                             format: int32
@@ -1455,9 +1527,9 @@ spec:
                             type: integer
                           diskType:
                             description: DiskType defines the type of disk. The valid
-                              values are Standard_LRS, Premium_LRS, StandardSSD_LRS
+                              values are Standard_LRS, Premium_LRS, StandardSSD_LRS.
                               For control plane nodes, the valid values are Premium_LRS
-                              and StandardSSD_LRS. Default is Premium_LRS
+                              and StandardSSD_LRS. Default is Premium_LRS.
                             enum:
                             - Standard_LRS
                             - Premium_LRS
--- a/pkg/asset/installconfig/azure/validation_test.go
+++ b/pkg/asset/installconfig/azure/validation_test.go
@@ -294,6 +294,7 @@ func TestAzureInstallConfigValidation(t *testing.T) {
 	//Resource SKUs
 	azureClient.EXPECT().GetDiskSkus(gomock.Any(), validResourceSkuRegions).Return(nil, fmt.Errorf("invalid disk type")).AnyTimes()
 	azureClient.EXPECT().GetDiskSkus(gomock.Any(), invalidResourceSkuRegion).Return(nil, fmt.Errorf("invalid region")).AnyTimes()
+
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
 			editedInstallConfig := validInstallConfig()
--- a/pkg/asset/machines/azure/machines.go
+++ b/pkg/asset/machines/azure/machines.go
@@ -105,6 +105,20 @@ func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string
 		managedIdentity = ""
 	}
 
+	var diskEncryptionSet *machineapi.DiskEncryptionSetParameters
+	if mpool.OSDisk.DiskEncryptionSet != nil {
+		diskEncryptionSet = &machineapi.DiskEncryptionSetParameters{
+			ID: mpool.OSDisk.DiskEncryptionSet.ToID(),
+		}
+	}
+
+	var securityProfile *machineapi.SecurityProfile
+	if mpool.EncryptionAtHost {
+		securityProfile = &machineapi.SecurityProfile{
+			EncryptionAtHost: &mpool.EncryptionAtHost,
+		}
+	}
+
 	spec := &machineapi.AzureMachineProviderSpec{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "machine.openshift.io/v1beta1",
@@ -122,8 +136,10 @@ func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string
 			DiskSizeGB: mpool.OSDisk.DiskSizeGB,
 			ManagedDisk: machineapi.ManagedDiskParameters{
 				StorageAccountType: mpool.OSDisk.DiskType,
+				DiskEncryptionSet:  diskEncryptionSet,
 			},
 		},
+		SecurityProfile:      securityProfile,
 		Zone:                 az,
 		Subnet:               subnet,
 		ManagedIdentity:      managedIdentity,
--- a/pkg/tfvars/azure/azure.go
+++ b/pkg/tfvars/azure/azure.go
@@ -26,6 +26,8 @@ type config struct {
 	ExtraTags                       map[string]string `json:"azure_extra_tags,omitempty"`
 	MasterInstanceType              string            `json:"azure_master_vm_type,omitempty"`
 	MasterAvailabilityZones         []string          `json:"azure_master_availability_zones"`
+	MasterEncryptionAtHostEnabled   bool              `json:"azure_master_encryption_at_host_enabled"`
+	MasterDiskEncryptionSetID       string            `json:"azure_master_disk_encryption_set_id,omitempty"`
 	VolumeType                      string            `json:"azure_master_root_volume_type"`
 	VolumeSize                      int32             `json:"azure_master_root_volume_size"`
 	ImageURL                        string            `json:"azure_image_url,omitempty"`
@@ -77,6 +79,15 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
 		return nil, errors.Wrap(err, "could not determine Azure environment to use for Terraform")
 	}
 
+	masterEncryptionAtHostEnabled := masterConfig.SecurityProfile != nil &&
+		(*masterConfig.SecurityProfile).EncryptionAtHost != nil &&
+		*masterConfig.SecurityProfile.EncryptionAtHost
+
+	var masterDiskEncryptionSetID string
+	if masterConfig.OSDisk.ManagedDisk.DiskEncryptionSet != nil {
+		masterDiskEncryptionSetID = masterConfig.OSDisk.ManagedDisk.DiskEncryptionSet.ID
+	}
+
 	cfg := &config{
 		Auth:                            sources.Auth,
 		Environment:                     environment,
@@ -84,6 +95,8 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
 		Region:                          region,
 		MasterInstanceType:              masterConfig.VMSize,
 		MasterAvailabilityZones:         masterAvailabilityZones,
+		MasterEncryptionAtHostEnabled:   masterEncryptionAtHostEnabled,
+		MasterDiskEncryptionSetID:       masterDiskEncryptionSetID,
 		VolumeType:                      masterConfig.OSDisk.ManagedDisk.StorageAccountType,
 		VolumeSize:                      masterConfig.OSDisk.DiskSizeGB,
 		ImageURL:                        sources.ImageURL,
--- a/pkg/types/azure/disk.go
+++ b/pkg/types/azure/disk.go
@@ -1,0 +1,49 @@
+package azure
+
+import "fmt"
+
+// ToID creates an Azure resource ID for the disk encryption set.
+// It is possible to return a non-valid ID when SubscriptionID is empty. This
+// should never happen since if SubscriptionID is empty, it is set to the
+// current subscription. Also, should it somehow be empty and this returns an
+// invalid ID, the validation code will produce an error when checked  against
+// the validation.RxDiskEncryptionSetID regular expression.
+func (d *DiskEncryptionSet) ToID() string {
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/diskEncryptionSets/%s",
+		d.SubscriptionID, d.ResourceGroup, d.Name)
+}
+
+// OSDisk defines the disk for machines on Azure.
+type OSDisk struct {
+	// DiskSizeGB defines the size of disk in GB.
+	//
+	// +kubebuilder:validation:Minimum=0
+	DiskSizeGB int32 `json:"diskSizeGB"`
+	// DiskType defines the type of disk.
+	// The valid values are Standard_LRS, Premium_LRS, StandardSSD_LRS.
+	// For control plane nodes, the valid values are Premium_LRS and StandardSSD_LRS.
+	// Default is Premium_LRS.
+	// +optional
+	// +kubebuilder:validation:Enum=Standard_LRS;Premium_LRS;StandardSSD_LRS
+	DiskType string `json:"diskType"`
+
+	// DiskEncryptionSet defines a disk encryption set.
+	//
+	// +optional
+	*DiskEncryptionSet `json:"diskEncryptionSet,omitempty"`
+}
+
+// DiskEncryptionSet defines the configuration for a disk encryption set.
+type DiskEncryptionSet struct {
+	// SubscriptionID defines the Azure subscription the disk encryption
+	// set is in.
+	SubscriptionID string `json:"subscriptionId,omitempty"`
+	// ResourceGroup defines the Azure resource group used by the disk
+	// encryption set.
+	ResourceGroup string `json:"resourceGroup"`
+	// Name is the name of the disk encryption set.
+	Name string `json:"name"`
+}
+
+// DefaultDiskType holds the default Azure disk type used by the VMs.
+const DefaultDiskType string = "Premium_LRS"
--- a/pkg/types/azure/machinepool.go
+++ b/pkg/types/azure/machinepool.go
@@ -15,29 +15,16 @@ type MachinePool struct {
 	// +optional
 	InstanceType string `json:"type"`
 
+	// EncryptionAtHost enables encryption at the VM host.
+	//
+	// +optional
+	EncryptionAtHost bool `json:"encryptionAtHost,omitempty"`
+
 	// OSDisk defines the storage for instance.
 	//
 	// +optional
 	OSDisk `json:"osDisk"`
 }
-
-// OSDisk defines the disk for machines on Azure.
-type OSDisk struct {
-	// DiskSizeGB defines the size of disk in GB.
-	//
-	// +kubebuilder:validation:Minimum=0
-	DiskSizeGB int32 `json:"diskSizeGB"`
-	// DiskType defines the type of disk.
-	// The valid values are Standard_LRS, Premium_LRS, StandardSSD_LRS
-	// For control plane nodes, the valid values are Premium_LRS and StandardSSD_LRS.
-	// Default is Premium_LRS
-	// +optional
-	// +kubebuilder:validation:Enum=Standard_LRS;Premium_LRS;StandardSSD_LRS
-	DiskType string `json:"diskType"`
-}
-
-// DefaultDiskType holds the default Azure disk type used by the VMs.
-const DefaultDiskType string = "Premium_LRS"
 
 // Set sets the values from `required` to `a`.
 func (a *MachinePool) Set(required *MachinePool) {
@@ -53,11 +40,19 @@ func (a *MachinePool) Set(required *MachinePool) {
 		a.InstanceType = required.InstanceType
 	}
 
+	if required.EncryptionAtHost {
+		a.EncryptionAtHost = required.EncryptionAtHost
+	}
+
 	if required.OSDisk.DiskSizeGB != 0 {
 		a.OSDisk.DiskSizeGB = required.OSDisk.DiskSizeGB
 	}
 
 	if required.OSDisk.DiskType != "" {
 		a.OSDisk.DiskType = required.OSDisk.DiskType
+	}
+
+	if required.DiskEncryptionSet != nil {
+		a.DiskEncryptionSet = required.DiskEncryptionSet
 	}
 }
--- a/pkg/types/azure/validation/disk.go
+++ b/pkg/types/azure/validation/disk.go
@@ -1,0 +1,57 @@
+package validation
+
+import (
+	"regexp"
+
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	"github.com/openshift/installer/pkg/types/azure"
+)
+
+var (
+	// RxDiskEncryptionSetID is a regular expression that validates a disk encryption set ID.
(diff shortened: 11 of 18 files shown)


pkg/asset/installconfig/azure/validation.go
```diff
@@ -161,6 +163,30 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+	}
+
+	encryptionType := ic.Platform.Azure.DefaultMachinePlatform.OSDisk.EncryptionType
+	if ic.Azure.CloudName == aztypes.StackCloud && encryptionType != "" {
+		return append(allErrs, field.Invalid(field.NewPath("encryptionType"), encryptionType, "customer managed keys encryption is not supported on this platform"))
+	}
+
+	if encryptionAtHost == true && encryptionType != string(azcompute.EncryptionAtRestWithCustomerKey) {
+		return append(allErrs, field.Invalid(field.NewPath("encryptionType"), encryptionType, "encryption at host is only supported with customer managed keys"))
+	}
```
m1kola, line 185: This is not correct. Encryption at host can be enabled independently of server side encrpytion (encryption at rest) and doesn't require customer managed keys ([example in docs](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/disks-enable-host-based-encryption-cli#create-a-vm-with-encryption-at-host-enabled-with-platform-managed-keys)). It is a separate feature. Just a bool on a VM.

However it can be used together with server side encrpytion ([example in docs](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/disks-enable-host-based-encryption-cli#create-a-vm-with-encryption-at-host-enabled-with-customer-managed-keys)). In that case temp disks are still encrypted with platform-managed keys.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -161,6 +163,30 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+
+	encryptionAtHost := ic.Platform.Azure.DefaultMachinePlatform.EncryptionAtHost
+	if ic.Azure.CloudName == aztypes.StackCloud && encryptionAtHost == true {
+		return append(allErrs, field.Invalid(field.NewPath("encryptionAtHost"), encryptionAtHost, "encryption at host is not supported not this platform"))
+	}
+
+	encryptionType := ic.Platform.Azure.DefaultMachinePlatform.OSDisk.EncryptionType
+	if ic.Azure.CloudName == aztypes.StackCloud && encryptionType != "" {
+		return append(allErrs, field.Invalid(field.NewPath("encryptionType"), encryptionType, "customer managed keys encryption is not supported on this platform"))
+	}
```
m1kola, line 181: I know very little about azure stack, but I was under the impression that it has the same set of features as the public cloud. Am I wrong?

Or do we just want to limit the scope of this feature to non-Azure Stack clouds for now?

pkg/asset/installconfig/azure/validation.go
m1kola: [This document](https://docs.microsoft.com/en-us/azure/virtual-machines/disk-encryption-overview) will either help you to get better understanding or confuse you more :) My journey with these docs was like this: I have no idea what I'm doing -> Ok, I think I understnad it now -> Nope, I'm lost again -> Ok, I think I finally understand it.

Hopefully this sumarry will help. There are three things related to disk encryption:

* Azure Disk Encryption (out of scope)
* Server-Side Encryption or encryption at rest
  * Enabled by default with platform managed keys
  * Customer can provide their own keys (this is what we want to achive here for our OCP/ARO clusters)
* Encryption at host
  * DIsabled by default
  * Can be enabled separately (in scope)
  * Can be enabled together with Server-Side Encryption with patform managed keys (in scope)
  * Can be enabled together with Server-Side Encryption with customer managed keys (in scope)

When I say in scope/not in scope - I refer to what I originally tried to do in #5134

pkg/asset/machines/azure/machines.go
```diff
@@ -105,6 +107,22 @@ func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string
...
+
+	var securityProfile *machineapi.SecurityProfile
+	if mpool.EncryptionAtHost {
+		securityProfile = &machineapi.SecurityProfile{
+			EncryptionAtHost: &mpool.EncryptionAtHost,
+		}
+	}
+
+	if mpool.OSDisk.EncryptionType != "" {
+		managedDisk.EncryptionType = mpool.OSDisk.EncryptionType
```
m1kola, line 123: There is no such thing as [`encryptionType` in Azure API](https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update#manageddiskparameters) in the VM create or update endpoint.

We need `diskEncryptionSet` as it was in #5134. And I think we don't need https://github.com/openshift/api/pull/1118.

Or am I missing something?

m1kola (changes requested):
@jhixson74 I don't have enough bandwith right now to continue [my original PR](https://github.com/openshift/installer/pull/5134). Thanks a lot for picking this up.

I found few issues (I think) in this PR. Please see my comments below.

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -165,9 +165,18 @@ spec:
...
+                        encryptionAtHost:
+                          description: EncryptionAtHost enables encryption at the
+                            VM host.
+                          type: boolean
                         osDisk:
                           description: OSDisk defines the storage for instance.
                           properties:
+                            encryptionType:
+                              description: encryptionType is the type of encryption used for disks.
+                                Possible values include: 'EncryptionAtRestWithPlatformKey',
```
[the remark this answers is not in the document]
jhixson74: This type, as with the other types for disk encryption sets, is here: 

https://github.com/Azure/azure-sdk-for-go/blob/165ee2301805cbda1463088ae0dd403a14395a1f/services/compute/mgmt/2020-06-01/compute/enums.go#L376 . 

The encryption type is a property of DiskProperties: 

https://github.com/Azure/azure-sdk-for-go/blob/165ee2301805cbda1463088ae0dd403a14395a1f/services/compute/mgmt/2020-06-01/compute/models.go#L3390 

DiskProperties -> Encryption -> EncryptionType. 

I am not familiar with openshift/api, however it does not appear to map 1 to 1 to the Azure API, so I placed the encryptionType under osDisk since it was the closest mapping. 

After reading what you wrote above, I am in agreement that we should not provision the keyvault and keys.

pkg/asset/installconfig/azure/validation.go
jhixson74: Thanks for the clarification. My understanding was SSE is always on, no customer managed keys required. Encryption at host required customer managed keys (I was not able to find clarification on this in the docs). If encryption at host does not require customer managed keys, I will update things. Thanks for pointing these things out.

pkg/asset/installconfig/azure/validation.go
jhixson74: It's been decided (for now), to just throw an error if these fields are set for Azure stack.

pkg/asset/machines/azure/machines.go
jhixson74: I will be reworking this PR. Thanks for all your feedback.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -161,6 +162,26 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
 	return allErrs
 }
 
+// validateDiskEncryption checks that azure disk encryption is not set for Azure stack cloud
+func validateDiskEncryption(client API, ic *types.InstallConfig) field.ErrorList {
```
m1kola, line 166: I'm not familiar with what kinds of validation installer performs, but here is what we validate on ARO:

* Disk Encryption Set ID is a [valid resoruce id](https://github.com/Azure/ARO-RP/blob/339345cbb4ea1521545560f9c6e6f919ff46185f/pkg/api/v20210901preview/openshiftcluster_validatestatic.go#L246-L252)
* Disk Encryption Set is [in the same subscription as cluster](https://github.com/Azure/ARO-RP/blob/339345cbb4ea1521545560f9c6e6f919ff46185f/pkg/api/v20210901preview/openshiftcluster_validatestatic.go#L253-L255)
* Cluster service principal (note: we don't use managed identities) [has permissions to read disk encryption set](https://github.com/Azure/ARO-RP/blob/339345cbb4ea1521545560f9c6e6f919ff46185f/pkg/api/validate/dynamic/diskencryptionset.go#L68-L89).
* Disk encryption set [is in the same location as cluster/VMs](https://github.com/Azure/ARO-RP/blob/339345cbb4ea1521545560f9c6e6f919ff46185f/pkg/api/validate/dynamic/diskencryptionset.go#L91-L108) (IIRC, it is an Azure limitation/requirement, but I might be wrong)

data/data/azure/cluster/master/variables.tf
```diff
@@ -16,6 +16,18 @@ variable "vm_size" {
   type = string
 }
 
+variable "disk_encryption_set_id" {
+  default     = null
```
m1kola, line 20: Note: Patrick had some feedback about default terraform values in the original PR https://github.com/openshift/installer/pull/5134#pullrequestreview-862702928

I'm not sure that I understand what needs to be done. Do you @jhixson74? If not - let's ask Patrick to clarify.

m1kola (commented):
I added a note on what we validate in ARO, but in general looks good to me.

pkg/asset/installconfig/azure/validation_test.go
```diff
@@ -255,6 +273,26 @@ func TestAzureInstallConfigValidation(t *testing.T) {
...
+		},
+		{
+			name:     "Valid disk encryption set",
+			edits:    editFunctions{validDiskEncryptionSet},
+			errorMsg: "",
+		},
+		{
+			name:     "Invalid encryption at host",
+			edits:    editFunctions{invalidEncryptionAtHost},
+			errorMsg: `^\S+`,
```
patrickdillon, line 289: We should probably check that the correct error message is being used

patrickdillon (commented):
Although the machine api requires the ID for the disk encryption set, it seems like having users enter an ID into the install config would be clunky. I think we should rethink the type. 

The GCP encryption work could be a potential guide, although they do not need to reassemble into an ID: https://github.com/openshift/installer/pull/4318

Also we should see what sort of validation we can do to retrieve the encryption set and what information that provides. It seems like there are different requirements for user- vs platform-managed keys

pkg/asset/installconfig/azure/validation.go
```diff
@@ -161,6 +164,42 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+		if !RxDiskEncryptionSetID.MatchString(diskEncryptionSetID) {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("diskEncryptionSetID"), diskEncryptionSetID, fmt.Sprintf("The provided disk encryption set '%s' is invalid", diskEncryptionSetID)))
+		}
+		desr, err := azautorest.ParseResourceID(diskEncryptionSetID)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("diskEncryptionSetID"), diskEncryptionSetID, err.Error()))
+		}
+		if desr.SubscriptionID != metadata.Credentials.SubscriptionID {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("diskEncryptionSetID"), diskEncryptionSetID, fmt.Sprintf("The provided disk encryption set '%s' is invalid: must be in same subscription as cluster", diskEncryptionSetID)))
+		}
```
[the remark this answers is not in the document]
m1kola: I believe cross subscription of encryption keys is not possible. This is from [Azure docs](https://docs.microsoft.com/en-us/azure/virtual-machines/disk-encryption#restrictions):

> All resources related to your customer-managed keys (Azure Key Vaults, disk encryption sets, VMs, disks, and snapshots) must be in the same subscription and region.

So it not only must be in the same sub, but also in the same region.

This is [one of the validations we do](https://github.com/openshift/installer/pull/5641/files#r805670299).

pkg/types/azure/machinepool.go
```diff
@@ -34,6 +39,10 @@ type OSDisk struct {
 	// +optional
 	// +kubebuilder:validation:Enum=Standard_LRS;Premium_LRS;StandardSSD_LRS
 	DiskType string `json:"diskType"`
+
+	// DiskEncryptionSetID is a resource ID of disk encryption set
+	// +optional
+	DiskEncryptionSetID string `json:"diskEncryptionSetId,omitempty"`
```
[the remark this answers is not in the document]
m1kola: On ARO we already ask customers for a bunch of IDs: for example, we ask for vnet subnet IDs. That's why we went for resource ID for disk encryption set. Just for consistency.

But if you decide to change this - that is fine from ARO perspective. We can provide values for 3 fields separately into the installer, no problem.

pkg/asset/installconfig/azure/validation.go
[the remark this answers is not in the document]
patrickdillon: > I believe cross subscription of encryption keys is not possible. This is from [Azure docs](https://docs.microsoft.com/en-us/azure/virtual-machines/disk-encryption#restrictions):

I agree. I was confused about customer-managed vs platform-managed keys. Now I understand this code only applies to customer-managed keys, this validation makes sense and should stay (although it should be rewritten to use the API interface).

pkg/types/azure/machinepool.go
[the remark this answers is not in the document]
patrickdillon: > On ARO we already ask customers for a bunch of IDs: for example, we ask for vnet subnet IDs. That's why we went for resource ID for disk encryption set. Just for consistency.

Ah, I see. It is the opposite on the installer, for subnets we ask for the name. The fact that the openshift API uses the ID I think is an argument for keeping the ID, but I think usability and consistency are better arguments:

```yaml
osDisk:
    diskEncryptionSetID: /subscriptions/12345678-1234-1234-1234-1234567890ab/resourceGroups/es-resource-group/providers/Microsoft.Compute/diskEncryptionSets/my-encryption-set
```

vs

```yaml
osDisk:
        diskEncryptionSet:
            name: my-encryption-set
            resourceGroup: es-resource-group
            subscription: 12345678-1234-1234-1234-1234567890ab
```

> But if you decide to change this - that is fine from ARO perspective. We can provide values for 3 fields separately into the installer, no problem.

So if this works for ARO, I think the latter is better for users. We should be able to use sane defaults:

So grabbing the service principal subscription, users supply two values:

```yaml
osDisk:
        diskEncryptionSet:
            name: my-encryption-set
            resourceGroup: es-resource-group
```

Perhaps it is reasonable to assume that if an existing resource group is supplied in the platform, the encryption set can default to that, in which case we can reduce to a single value:

```yaml
osDisk:
        diskEncryptionSet:
            name: my-encryption-set
```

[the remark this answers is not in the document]
m1kola: @patrickdillon sounds good to me. Apart from bit:

> Perhaps it is reasonable to assume that if an existing resource group is supplied in the platform, the encryption set can default to that, in which case we can reduce to a single value:

Not sure that I understand "supplied in the platform" point. One of the previous implementations of this PR was with automatic provisioning of key vault, key and disk encryption set by installer inside the cluster resource group. Is this what you are suggesting?

I think this stuff needs to be provisioned and provided by a customer. Especially on ARO. Some of this was discussed here: https://github.com/openshift/installer/pull/5641#discussion_r804537388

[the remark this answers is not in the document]
patrickdillon: >  Is this what you are suggesting?
> 
> I think this stuff needs to be provisioned and provided by a customer. Especially on ARO. 

No, that is not what I intended to suggest. I was/am trying to determine if there is a sane default we could use for the machineset `azure.osDisk.diskEncryptionSet.resourceGroup`. What I was considering--and reading your comment from the other PR, it sounds like this is not the case--is whether when `installConfig.platform.azure.resourceGroupName != ""` if that would be a suitable default value for `diskEncryptionSet.resourceGroup`. 

So it sounds like `diskEncryptionSet.resourceGroup` should always be supplied and a default value cannot be inferred.

data/data/azure/cluster/master/variables.tf
```diff
@@ -22,6 +22,12 @@ variable "disk_encryption_set_id" {
   description = "The ID of the Disk Encryption Set which should be used to encrypt OS disk."
 }
 
+variable "encryption_at_host_enabled" {
+  default     = false
```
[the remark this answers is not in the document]
jhixson74: Fixed.

data/data/azure/variables-azure.tf
```diff
@@ -34,6 +34,12 @@ variable "azure_master_disk_encryption_set_id" {
   description = "The ID of the Disk Encryption Set which should be used to encrypt OS disk for the master node(s)."
 }
 
+variable "azure_master_encryption_at_host_enabled" {
+  default = false
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/types/azure/machinepool.go
```diff
@@ -34,6 +34,10 @@ type OSDisk struct {
 	// +optional
 	// +kubebuilder:validation:Enum=Standard_LRS;Premium_LRS;StandardSSD_LRS
 	DiskType string `json:"diskType"`
+
+	// DiskEncryptionSetID is a resource ID of disk encryption set
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -161,6 +164,42 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
 	return allErrs
 }
 
+// validateDiskEncryption checks that azure disk encryption is not set for Azure stack cloud
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -161,6 +164,42 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+	RxDiskEncryptionSetID := regexp.MustCompile(`(?i)^/subscriptions/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/resourceGroups/[-a-z0-9_().]{0,89}[-a-z0-9_()]/providers/Microsoft\.Compute/diskEncryptionSets/[-a-z0-9_]{1,80}$`)
+
+	allErrs := field.ErrorList{}
+	if ic.Platform.Azure == nil || ic.Platform.Azure.DefaultMachinePlatform == nil {
+		return allErrs
+	}
+
+	encryptionAtHost := ic.Platform.Azure.DefaultMachinePlatform.EncryptionAtHost
+	if encryptionAtHost == true && ic.Azure.CloudName == aztypes.StackCloud {
+		return append(allErrs, field.Invalid(field.NewPath("encryptionAtHost"), encryptionAtHost, "encryption at host is not supported not this platform"))
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -161,6 +164,42 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+	}
+
+	diskEncryptionSetID := ic.Platform.Azure.DefaultMachinePlatform.OSDisk.DiskEncryptionSetID
+	if diskEncryptionSetID != "" && ic.Azure.CloudName == aztypes.StackCloud {
+		return append(allErrs, field.Invalid(field.NewPath("diskEncryptionSetID"), diskEncryptionSetID, "disk encryption set id's are not supported not this platform"))
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -161,6 +164,42 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+	}
+
+	metadata := NewMetadata(ic.Azure.CloudName, ic.Platform.Azure.ARMEndpoint)
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/asset/installconfig/azure/validation_test.go
jhixson74: Fixed.

pkg/asset/installconfig/azure/validation_test.go
```diff
@@ -69,6 +69,24 @@ var (
 		ic.Platform.Azure.DefaultMachinePlatform.InstanceType = "Dne_D2_v4"
 	}
 
+	validEncryptionAtHost = func(ic *types.InstallConfig) {
+		ic.Platform.Azure.DefaultMachinePlatform.EncryptionAtHost = true
+	}
+
+	validDiskEncryptionSet = func(ic *types.InstallConfig) {
+		ic.Platform.Azure.DefaultMachinePlatform.OSDisk.DiskEncryptionSetID = "azure-encryption-set-id"
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/tfvars/azure/azure.go
```diff
@@ -28,6 +28,7 @@ type config struct {
 	BootstrapInstanceType           string            `json:"azure_bootstrap_vm_type,omitempty"`
 	MasterInstanceType              string            `json:"azure_master_vm_type,omitempty"`
 	MasterAvailabilityZones         []string          `json:"azure_master_availability_zones"`
+	MasterDiskEncryptionSetID       string            `json:"azure_master_disk_encryption_set_id,omitempty"`
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -161,6 +164,42 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+	if diskEncryptionSetID != "" && (Metadata{}) != *metadata && metadata.Credentials != nil {
+		if !RxDiskEncryptionSetID.MatchString(diskEncryptionSetID) {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("diskEncryptionSetID"), diskEncryptionSetID, fmt.Sprintf("The provided disk encryption set '%s' is invalid", diskEncryptionSetID)))
```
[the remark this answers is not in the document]
jhixson74: Fixed.

data/data/azure/variables-azure.tf
```diff
@@ -28,6 +28,12 @@ variable "azure_master_vm_type" {
   description = "Instance type for the master node(s). Example: `Standard_D8s_v3`."
 }
 
+variable "azure_master_disk_encryption_set_id" {
+  default = null
```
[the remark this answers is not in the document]
jhixson74: Fixed.

data/data/azure/cluster/master/variables.tf
jhixson74: Fixed.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -161,6 +164,42 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+		return allErrs
+	}
+
+	encryptionAtHost := ic.Platform.Azure.DefaultMachinePlatform.EncryptionAtHost
+	if encryptionAtHost == true && ic.Azure.CloudName == aztypes.StackCloud {
+		return append(allErrs, field.Invalid(field.NewPath("encryptionAtHost"), encryptionAtHost, "encryption at host is not supported not this platform"))
+	}
+
+	diskEncryptionSetID := ic.Platform.Azure.DefaultMachinePlatform.OSDisk.DiskEncryptionSetID
+	if diskEncryptionSetID != "" && ic.Azure.CloudName == aztypes.StackCloud {
```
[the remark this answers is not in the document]
jhixson74: It appears so: https://docs.microsoft.com/en-us/azure/azure-government/documentation-government-plan-security

pkg/asset/installconfig/azure/validation.go
```diff
@@ -161,6 +164,42 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+		return append(allErrs, field.Invalid(field.NewPath("diskEncryptionSetID"), diskEncryptionSetID, "disk encryption set id's are not supported not this platform"))
+	}
+
+	metadata := NewMetadata(ic.Azure.CloudName, ic.Platform.Azure.ARMEndpoint)
+	if diskEncryptionSetID != "" && (Metadata{}) != *metadata && metadata.Credentials != nil {
+		if !RxDiskEncryptionSetID.MatchString(diskEncryptionSetID) {
```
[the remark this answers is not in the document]
jhixson74: I'm not sure I follow, can you explain? Currently, it's not being used anywhere else. If the need arises, I'm fine making it more modular and moving it.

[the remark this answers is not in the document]
patrickdillon: Static validation (validation not requiring a connection to the API) should go here: https://github.com/openshift/installer/tree/master/pkg/types/azure/validation

`pkg/types` is the public API for the Installer. That is other repos can import the types, and use this validation. 

All validation that connects to the APIs goes here in pkg/asset.

data/data/azure/bootstrap/main.tf
```diff
@@ -200,17 +200,19 @@ resource "azurerm_linux_virtual_machine" "bootstrap" {
...
   os_disk {
-    name                 = "${var.cluster_id}-bootstrap_OSDisk" # os disk name needs to match cluster-api convention
-    caching              = "ReadWrite"
-    storage_account_type = var.azure_master_root_volume_type
-    disk_size_gb         = 100
+    name                   = "${var.cluster_id}-bootstrap_OSDisk" # os disk name needs to match cluster-api convention
+    caching                = "ReadWrite"
+    storage_account_type   = var.azure_master_root_volume_type
+    disk_size_gb           = 100
+    disk_encryption_set_id = var.azure_master_disk_encryption_set_id == "" ? null : var.azure_master_disk_encryption_set_id
```
patrickdillon, line 215: I steered you wrong here: the way you had it before, with a default of `null` is much better. I was under the impression that an empty string would be the equivalent of null, but I was mistaken. Sorry about that.

data/data/azure/cluster/main.tf
```diff
@@ -29,26 +29,28 @@ provider "azureprivatedns" {
...
-  os_volume_size         = var.azure_master_root_volume_size
-  private                = var.azure_private
-  outbound_udr           = var.azure_outbound_user_defined_routing
+  source                     = "./master"
+  resource_group_name        = var.resource_group_name
+  cluster_id                 = var.cluster_id
+  region                     = var.azure_region
+  availability_zones         = var.azure_master_availability_zones
+  vm_size                    = var.azure_master_vm_type
+  disk_encryption_set_id     = var.azure_master_disk_encryption_set_id == "" ? null : var.azure_master_disk_encryption_set_id
```
patrickdillon, line 38: Same as other comment.

pkg/asset/installconfig/azure/validation.go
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/types/azure/machinepool.go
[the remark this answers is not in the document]
jhixson74: I've split the ID up into the individual fields.

pkg/types/azure/disk.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// ToID creates an Azure resource ID for the disk encryption set.
+func (d *DiskEncryptionSet) ToID() string {
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/diskEncryptionSets/%s",
+		d.SubscriptionID, d.ResourceGroup, d.Name)
+}
+
+// Set fills a DiskEncryptionSet struct from a disk encryption set ID string
+func (d *DiskEncryptionSet) Set(ID string) error {
```
patrickdillon, line 31: I'm not following this: why are we setting from the ID rather than `*DiskEncryptionSet`?  If this is only used in tests it should be moved to tests, but I think it makes more sense to have this Set from `*DiskEncryptionSet` and used by https://github.com/openshift/installer/blob/master/pkg/types/azure/machinepool.go#L43 with a pattern similar to the GCP encryption keys: https://github.com/openshift/installer/blob/master/pkg/types/gcp/machinepools.go#L65-L70

pkg/types/azure/disk.go
```diff
@@ -0,0 +1,65 @@
...
+
+var (
+	RxDiskEncryptionSetID   = regexp.MustCompile(`(?i)^/subscriptions/([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})/resourceGroups/([-a-z0-9_().]{0,89}[-a-z0-9_()])/providers/Microsoft\.Compute/diskEncryptionSets/([-a-z0-9_]{1,80})$`)
+	RxSubscriptionID        = regexp.MustCompile(`^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`)
+	RxResourceGroup         = regexp.MustCompile(`^[-a-z0-9_().]{0,89}[-a-z0-9_()]$`)
+	RxDiskEncryptionSetName = regexp.MustCompile(`^[-a-z0-9_]{1,80}$`)
+)
+
+// DiskEncryptionSet defines the configuration for a disk encryption set.
+type DiskEncryptionSet struct {
```
patrickdillon, line 18: As `DiskEncryptionSet` is used as a field in `OSDisk` I think it would be more readable to have `OSDisk` appear first  and `DiskEncryptionSet` would be defined below.

pkg/types/azure/disk.go
```diff
@@ -0,0 +1,65 @@
...
+	SubscriptionID string `json:"subscriptionId"`
+	ResourceGroup  string `json:"resourceGroup"`
+	Name           string `json:"name"`
```
patrickdillon, line 21: These need comments for Go Docs and explain. Writing the comments, adding kubebuilder tags, and generating the explain docs should help clear up errors in the govet and explain tests.

patrickdillon (commented):
Moving the EncryptionSet to a struct looks good, but still needs further refinement. 

I suggested validating that the referenced encryption set actually exists. I think that work should get a new card and be merged separately from this.

data/data/azure/cluster/master/variables.tf
```diff
@@ -16,6 +16,16 @@ variable "vm_size" {
   type = string
 }
 
+variable "disk_encryption_set_id" {
+  type        = string
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/types/azure/disk.go
```diff
@@ -0,0 +1,65 @@
...
+	// For control plane nodes, the valid values are Premium_LRS and StandardSSD_LRS.
+	// Default is Premium_LRS.
+	// +optional
+	// +kubebuilder:validation:Enum=Standard_LRS;Premium_LRS;StandardSSD_LRS
+	DiskType string `json:"diskType"`
+
+	// DiskEncryptionSet defines a disk encryption set.
+	//
+	// +optional
+	DiskEncryptionSet `json:"diskEncryptionSet,omitempty"`
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/types/azure/disk.go
```diff
@@ -0,0 +1,65 @@
...
+var (
+	RxDiskEncryptionSetID   = regexp.MustCompile(`(?i)^/subscriptions/([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})/resourceGroups/([-a-z0-9_().]{0,89}[-a-z0-9_()])/providers/Microsoft\.Compute/diskEncryptionSets/([-a-z0-9_]{1,80})$`)
+	RxSubscriptionID        = regexp.MustCompile(`^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`)
+	RxResourceGroup         = regexp.MustCompile(`^[-a-z0-9_().]{0,89}[-a-z0-9_()]$`)
+	RxDiskEncryptionSetName = regexp.MustCompile(`^[-a-z0-9_]{1,80}$`)
+)
+
+// DiskEncryptionSet defines the configuration for a disk encryption set.
+type DiskEncryptionSet struct {
+	SubscriptionID string `json:"subscriptionId"`
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/types/azure/disk.go
jhixson74: Fixed.

pkg/types/azure/disk.go
jhixson74: Fixed.

pkg/types/azure/disk.go
jhixson74: Nuked.

jhixson74:
[](https://github.com/openshift/installer/pull/5641#issuecomment-1061362776)/test e2e-azure

pkg/asset/machines/azure/machines.go
```diff
@@ -105,6 +105,23 @@ func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string
 		managedIdentity = ""
 	}
 
+	var diskEncryptionSet *machineapi.DiskEncryptionSetParameters
+	if mpool.OSDisk.DiskEncryptionSet != nil &&
+		mpool.OSDisk.DiskEncryptionSet.SubscriptionID != "" &&
+		mpool.OSDisk.DiskEncryptionSet.ResourceGroup != "" &&
+		mpool.OSDisk.DiskEncryptionSet.Name != "" {
```
patrickdillon, line 112: These empty checks shouldn't be required because the fields have already been validated or have a default

pkg/asset/machines/azure/machines.go
```diff
@@ -105,6 +105,23 @@ func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string
...
+		diskEncryptionSet = &machineapi.DiskEncryptionSetParameters{
+			ID: mpool.OSDisk.DiskEncryptionSet.ToID(),
```
patrickdillon, line 114: :heart_eyes:

pkg/types/azure/disk.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"fmt"
+	"regexp"
+)
+
+var (
+	RxDiskEncryptionSetID   = regexp.MustCompile(`(?i)^/subscriptions/([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})/resourceGroups/([-a-z0-9_().]{0,89}[-a-z0-9_()])/providers/Microsoft\.Compute/diskEncryptionSets/([-a-z0-9_]{1,80})$`)
+	RxSubscriptionID        = regexp.MustCompile(`^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`)
+	RxResourceGroup         = regexp.MustCompile(`^[-a-z0-9_().]{0,89}[-a-z0-9_()]$`)
+	RxDiskEncryptionSetName = regexp.MustCompile(`^[-a-z0-9_]{1,80}$`)
```
patrickdillon, line 12: I think it makes more sense to move this to the validation package not export it.

pkg/types/azure/disk.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+// ToID creates an Azure resource ID for the disk encryption set.
+func (d *DiskEncryptionSet) ToID() string {
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/Microsoft.Compute/diskEncryptionSets/%s",
+		d.SubscriptionID, d.ResourceGroup, d.Name)
+}
```
patrickdillon, line 19: nit: This function can return an invalid id when subscription is empty. (but it won't in the current code flow so I don't see it as a blocker, just mention it in case you see a room for improvement).

pkg/types/azure/validation/disk.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/openshift/installer/pkg/types/azure"
+)
+
+// ValidateDiskEncryption checks that the specified disk encryption configuration is valid.
+func ValidateDiskEncryption(p *azure.MachinePool, cloudName azure.CloudEnvironment, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	childFldPath := fldPath.Child("osDisk", "diskEncryptionSet")
+
+	diskEncryptionSet := p.OSDisk.DiskEncryptionSet
+	diskEncryptionSetID := diskEncryptionSet.ToID()
```
patrickdillon, line 15: At this stage, we can't assume subscription is set and therefore have a valid id.

pkg/types/azure/validation/disk.go
```diff
@@ -0,0 +1,45 @@
...
+	if diskEncryptionSetID != "" && cloudName == azure.StackCloud {
```
patrickdillon, line 16:
```suggestion
	if diskEncryptionSet != nil && cloudName == azure.StackCloud {
```

pkg/types/azure/validation/disk.go
```diff
@@ -0,0 +1,45 @@
...
+		return append(allErrs, field.Invalid(childFldPath.Child("diskEncryptionSetID"), diskEncryptionSetID, "disk encryption sets are not supported not this platform"))
```
patrickdillon, line 17:
```suggestion
		return append(allErrs, field.Invalid(childFldPath.Child("diskEncryptionSet"), diskEncryptionSet, "disk encryption sets are not supported not this platform"))
```

pkg/types/azure/validation/disk.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	if !azure.RxSubscriptionID.MatchString(diskEncryptionSet.SubscriptionID) {
```
patrickdillon, line 19:
```suggestion
	if diskEncryptionSet.SubscriptionID != "" && !azure.RxSubscriptionID.MatchString(diskEncryptionSet.SubscriptionID) {
```

pkg/types/azure/validation/disk.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	if !azure.RxResourceGroup.MatchString(diskEncryptionSet.ResourceGroup) {
+		return append(allErrs, field.Invalid(childFldPath.Child("resourceGroup"), diskEncryptionSet.ResourceGroup, "invalid resource group format"))
+	}
+	if !azure.RxDiskEncryptionSetName.MatchString(diskEncryptionSet.Name) {
+		return append(allErrs, field.Invalid(childFldPath.Child("diskEncryptionSetName"), diskEncryptionSet.Name, "invalid name format"))
+	}
+	if !azure.RxDiskEncryptionSetID.MatchString(diskEncryptionSetID) {
+		return append(allErrs, field.Invalid(childFldPath.Child("diskEncryptionSetID"), diskEncryptionSetID, "invalid encryption set ID format"))
+	}
```
patrickdillon, line 30:
```suggestion
```
We can't check this yet because we may not have the subscription, but since we will construct it later and have validated all sub parts we know it is valid.

pkg/types/azure/validation/disk_test.go
```diff
@@ -0,0 +1,135 @@
...
+		{
+			name: "invalid disk encryption set (invalid subscription ID)",
+			pool: func() *azure.MachinePool {
+				p := validDiskEncryptionMachinePool()
+				p.OSDisk.DiskEncryptionSet.SubscriptionID = ""
+				return p
+			}(),
+			cloudName: azure.PublicCloud,
+			expected:  `subscriptionID: Invalid value: "": invalid subscription ID format`,
+		},
```
patrickdillon, line 61:
```suggestion
```
empty subscription is ok, we default to account subscription

pkg/types/azure/validation/disk.go
```diff
@@ -0,0 +1,45 @@
...
+	return allErrs
+}
+
+// ValidateEncryptionAtHost checks that the encryption at host configuration is valid.
+func ValidateEncryptionAtHost(p *azure.MachinePool, cloudName azure.CloudEnvironment, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	encryptionAtHost := p.EncryptionAtHost
+	if encryptionAtHost == true && cloudName == azure.StackCloud {
+		return append(allErrs, field.Invalid(fldPath.Child("encryptionAtHost"), encryptionAtHost, "encryption at host is not supported not this platform"))
```
patrickdillon, line 41:
```suggestion
		return append(allErrs, field.Invalid(fldPath.Child("encryptionAtHost"), encryptionAtHost, "encryption at host is not supported on this platform"))
```
this pops up a few places if you can do find and replace

patrickdillon (commented):
Mostly fine tuning validation.

It looks like the tests may be failing on the explain command, but I did not look into it too closely.

pkg/asset/machines/azure/machines.go
jhixson74: Fixed.

pkg/types/azure/disk.go
jhixson74: Fixed.

pkg/types/azure/validation/disk.go
jhixson74: Fixed.

pkg/types/azure/validation/disk.go
jhixson74: Fixed.

pkg/types/azure/validation/disk.go
jhixson74: Fixed.

pkg/types/azure/validation/disk.go
jhixson74: Fixed.

patrickdillon:
/approve

This LGTM once the tests are fixed. Also, can you squash any changes that are spread over commits (e.g. moving validation from assets to types)? I'm fine if you want to squash it all into a single commit--or not!

patrickdillon:
Interesting. Install succeeded but the step failed: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/5641/pull-ci-openshift-installer-master-e2e-azure/1502011338124169216

PR #5643, opened 2022-02-11
Set container engine in scripts based on OS

### LINKED ISSUE #5642: hack/go-fmt, hack/go-vet, and other sanity checking scripts fail to run locally on Mac
When running the scripts locally on a Mac, I see the following errors. Note: the directory indeed exists (see `pwd` output)
```
➜  openshift-installer git:(ntnx-2) pwd
/Users/sid.shukla/go/src/github.com/nutanix-cloud-native/openshift-installer
➜  openshift-installer git:(ntnx-2) ./hack/go-fmt.sh
Error: statfs /Users/sid.shukla/go/src/github.com/nutanix-cloud-native/openshift-installer: no such file or directory
➜  openshift-installer git:(ntnx-2) ./hack/go-vet.sh
Error: statfs /Users/sid.shukla/go/src/github.com/nutanix-cloud-native/openshift-installer: no such file or directory
➜  openshift-installer git:(ntnx-2) ./hack/go-lint.sh
Error: statfs /Users/sid.shukla/go/src/github.com/nutanix-cloud-native/openshift-installer: no such file or directory
```
This seems to be coming from podman. Switching out the podman commands to use docker does not produce these errors and scripts succeed.

thunderboltsid:
Use docker for Mac and podman for everything else.

Fixes #5642

### DIFF
--- a/hack/go-fmt.sh
+++ b/hack/go-fmt.sh
@@ -5,7 +5,11 @@ if [ "$IS_CONTAINER" != "" ]; then
   done
   git diff --exit-code
 else
-  podman run --rm \
+  ENGINE="podman"
+  if [ "$(uname)" = "Darwin" ]; then
+    ENGINE="docker"
+  fi
+  "$ENGINE" run --rm \
     --env IS_CONTAINER=TRUE \
     --volume "${PWD}:/go/src/github.com/openshift/installer:z" \
     --workdir /go/src/github.com/openshift/installer \
--- a/hack/go-genmock.sh
+++ b/hack/go-genmock.sh
@@ -5,7 +5,11 @@ if [ "$IS_CONTAINER" != "" ]; then
   go install github.com/golang/mock/mockgen
   go generate ./pkg/asset/installconfig/... "${@}"
 else
-  podman run --rm \
+  ENGINE="podman"
+  if [ "$(uname)" = "Darwin" ]; then
+    ENGINE="docker"
+  fi
+  "$ENGINE" run --rm \
     --env IS_CONTAINER=TRUE \
     --volume "${PWD}:/go/src/github.com/openshift/installer:z" \
     --workdir /go/src/github.com/openshift/installer \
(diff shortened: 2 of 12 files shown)


hack/go-fmt.sh
```diff
@@ -5,7 +5,11 @@ if [ "$IS_CONTAINER" != "" ]; then
   done
   git diff --exit-code
 else
-  podman run --rm \
+  ENGINE="podman"
+  if [ "$(uname)" = "Darwin" ]; then
```
barbacbd, line 9: Should we add this in if there is a work around? Could the user 1) alias podman=docker or 2) if we are going to follow through with these changes, warn someone as to the changes. 

```if ! command -v podman >/dev/null 2>&1 ; then
    if command -v docker >/dev/null 2>&1 ; then
        echo "[WARNING]: Aliasing docker to podman. Install podman if your system meets requirements.";
        alias podman=docker;
    else
        echo "[ERROR]: Failed to find services in the following order [podman, docker]." && exit
    fi
fi
```
It doesn't have to be this solution, but it would at least warn the user.

jhixson74:
This looks straight forward, however it's just not passing CI. Any ideas on why not?

sadasu:
/test e2e-nutanix

@thunderboltsid @jhixson74 e2e-nutanix CI tests seem to have a very low pass rate. But seeing some successful runs in the recent past. So, kicking off a test. Ideally, we would like the PR submitter to take the lead on figuring out what is going on with the failed CIs.

thunderboltsid:
> /test e2e-nutanix
> 
> @thunderboltsid @jhixson74 e2e-nutanix CI tests seem to have a very low pass rate. But seeing some successful runs in the recent past. So, kicking off a test. Ideally, we would like the PR submitter to take the lead on figuring out what is going on with the failed CIs.

@sadasu Looking at the e2e-nutanix suite runs, every time the tests have been run, they have passed. Most of the times, the suite doesn't get to run and the build fails early on due to a missing dependency on GOVC or some linux deps (misses on the NYU and RH mirror).
cc @rvanderp3

thunderboltsid:
Closing the PR as it doesn't add much value and is just generating github noise as some required tests never seem to pass.

PR #5645, opened 2022-02-11
Render the CVO manifest with user-specified capabilities from the install-config

kirankt:
This PR is depends on: #5605 and #5629.

The CVO manifest is rendered using capabilities specified in the install-config.


wking:
These are stacking up a bit deep, but looks like this change doesn't actually need to depend on #5629?  Hopefully #5605 will land soon, and we can save some time on rebasing.

pkg/types/installconfig.go
```diff
@@ -154,6 +156,10 @@ type InstallConfig struct {
 	// BootstrapInPlace is the configuration for installing a single node
 	// with bootstrap in place installation.
 	BootstrapInPlace *BootstrapInPlace `json:"bootstrapInPlace,omitempty"`
+
+	// Capabilities configures the installation of optional core cluster components.
+	// +optional
+	Capabilities *configv1.ClusterVersionCapabilitiesSpec `json:"capabilities,omitempty"`
```
staebler, line 162: https://github.com/openshift/installer/pull/5605#discussion_r812493276

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,36 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
 	}
 	return allErrs
 }
+
+// validateCapabilities checks if additional, optional OpenShift components are specified in the
+// install-config to be included in the installation.
+func validateCapabilities(c *configv1.ClusterVersionCapabilitiesSpec, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	allAvailableCapabilities := map[string]bool{}
```
staebler, line 712: Use a [set](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/sets/string.go) instead of a map. This will obviate the need for the `allCaps` variable, too.
```suggestion
	allAvailableCapabilities := sets.NewString()
```

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,36 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+			if _, found := allAvailableCapabilities[string(capability)]; !found {
+				allAvailableCapabilities[string(capability)] = true
+				allCaps = append(allCaps, string(capability))
+			}
+		}
+	}
+
+	if _, found := configv1.ClusterVersionCapabilitySets[c.BaselineCapabilitySet]; !found {
+		sort.Strings(allSets)
+		allErrs = append(allErrs, field.Invalid(fldPath, c.BaselineCapabilitySet, fmt.Sprintf("Unknown baseline set. Must specify one of the following (%s)", strings.Join(allSets, ", "))))
```
staebler, line 727: Use `field.NotSupported` instead of `field.Invalid`.

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,36 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+	}
+
+	// Check to see the validity of additionalEnabledCapabilities specified by the user
+	for _, capability := range c.AdditionalEnabledCapabilities {
+		if _, found := allAvailableCapabilities[string(capability)]; !found {
+			sort.Strings(allCaps)
+			allErrs = append(allErrs, field.Invalid(fldPath, string(capability), fmt.Sprintf("Unknown capability. Must specify one of the following (%s)", strings.Join(allCaps, ", "))))
```
staebler, line 734: Similarly, here, use `field.NotSupported`.

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,36 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+	}
+
+	if _, found := configv1.ClusterVersionCapabilitySets[c.BaselineCapabilitySet]; !found {
+		sort.Strings(allSets)
+		allErrs = append(allErrs, field.Invalid(fldPath, c.BaselineCapabilitySet, fmt.Sprintf("Unknown baseline set. Must specify one of the following (%s)", strings.Join(allSets, ", "))))
+	}
+
+	// Check to see the validity of additionalEnabledCapabilities specified by the user
+	for _, capability := range c.AdditionalEnabledCapabilities {
+		if _, found := allAvailableCapabilities[string(capability)]; !found {
```
staebler, line 732: If you use a set, then you can use `Difference` here. This will make this a bit more concise and easier to reason about.

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,36 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+		allSets = append(allSets, string(baselineSet))
+		for _, capability := range capabilities {
+			if _, found := allAvailableCapabilities[string(capability)]; !found {
+				allAvailableCapabilities[string(capability)] = true
+				allCaps = append(allCaps, string(capability))
+			}
+		}
+	}
+
+	if _, found := configv1.ClusterVersionCapabilitySets[c.BaselineCapabilitySet]; !found {
```
staebler, line 725: This will raise an error if `BaselineCapabilitySet` is omitted. We want the field to be optional, right? We should add a test case for this.

pkg/asset/manifests/template.go
```diff
@@ -1,6 +1,9 @@
 package manifests
 
-import "github.com/openshift/installer/pkg/types/baremetal"
+import (
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/installer/pkg/types/baremetal"
```
staebler, line 5: Separate imports of installer packages from other packages.
```suggestion
	configv1 "github.com/openshift/api/config/v1"

	"github.com/openshift/installer/pkg/types/baremetal"
```

staebler (commented):
Added reviews for all three PRs into this PR.

pkg/types/validation/installconfig.go
kirankt: True. I've added a check to catch empty values.

pkg/types/validation/installconfig.go
kirankt: Done.

pkg/types/validation/installconfig.go
kirankt: Done. Thanks.

pkg/types/validation/installconfig.go
kirankt: +1.

pkg/types/installconfig.go
wking: Won't openshift/api have those same forward-compat commitments?

pkg/types/installconfig.go
```diff
@@ -373,6 +379,24 @@ type BootstrapInPlace struct {
 	InstallationDisk string `json:"installationDisk"`
 }
 
+// Capabilities selects the managed set of optional, core cluster components.
+type Capabilities struct {
+	// baselineCapabilitySet selects an initial set of
+	// optional capabilities to enable, which can be extended via
+	// additionalEnabledCapabilities.  If unset, the cluster will
+	// choose a default, and the default may change over time.
```
staebler, line 387: Since the installer is installing a particular static version, should we removed the language about the default changing?

pkg/types/installconfig.go
```diff
@@ -373,6 +379,24 @@ type BootstrapInPlace struct {
...
+	// The current default is vCurrent.
+	// +optional
+	BaselineCapabilitySet configv1.ClusterVersionCapabilitySet `json:"baselineCapabilitySet,omitempty"`
+
+	// additionalEnabledCapabilities extends the set of managed
+	// capabilities beyond the baseline defined in
+	// baselineCapabilitySet.  The default is an empty set.
+	// +listType=atomic
```
staebler, line 395: Is this meaningful in the context of the install config?

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,35 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
 }
+
+// validateCapabilities checks if additional, optional OpenShift components are specified in the
+// install-config to be included in the installation.
+func validateCapabilities(c *types.Capabilities, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	allCapabilitySets := sets.NewString()
+	allAvailableCapabilities := sets.NewString()
+	// Create a map and slice of *all* available sets and capabilities across all capability sets
```
staebler, line 714:
```suggestion
	// Create sets of all capability sets and *all* available capabilities across those capability sets
```

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,35 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+	for baselineSet, capabilities := range configv1.ClusterVersionCapabilitySets {
+		if !allCapabilitySets.Has(string(baselineSet)) {
+			allCapabilitySets.Insert(string(baselineSet))
+		}
```
staebler, line 718: It is a set, so the insert is guaranteed not to create a duplicate.
```suggestion
		allCapabilitySets.Insert(string(baselineSet))
```

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,35 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+		for _, capability := range capabilities {
+			if !allAvailableCapabilities.Has(string(capability)) {
+				allAvailableCapabilities.Insert(string(capability))
+			}
```
staebler, line 722:
```suggestion
			allAvailableCapabilities.Insert(string(capability))
```

pkg/types/validation/installconfig.go
```diff
@@ -699,3 +703,35 @@ func validateFIPSconfig(c *types.InstallConfig) field.ErrorList {
...
+	}
+
+	if len(c.BaselineCapabilitySet) > 0 && !allCapabilitySets.Has(string(c.BaselineCapabilitySet)) {
+		allErrs = append(allErrs, field.NotSupported(fldPath.Child("baselineCapabilitySet"), c.BaselineCapabilitySet, allCapabilitySets.List()))
+	}
+
+	// Check to see the validity of additionalEnabledCapabilities specified by the user
+	for _, capability := range c.AdditionalEnabledCapabilities {
+		if len(capability) > 0 && !allAvailableCapabilities.Has(string(capability)) {
+			allErrs = append(allErrs, field.NotSupported(fldPath.Child("additionalEnabledCapabilities"), capability, allAvailableCapabilities.List()))
```
staebler, line 733: 1. An empty capability string should be rejected not accepted.
2. Include the index of the offending capability in the field error.
```suggestion
	for i, capability := range c.AdditionalEnabledCapabilities {
		if !allAvailableCapabilities.Has(string(capability)) {
			allErrs = append(allErrs, field.NotSupported(fldPath.Child("additionalEnabledCapabilities").Index(i), capability, allAvailableCapabilities.List()))
```

pkg/types/installconfig.go
kirankt: Removing it. Thanks.

pkg/types/installconfig.go
kirankt: +1

pkg/types/validation/installconfig.go
kirankt: I was just going through this file and had changed it before I saw this. I'll use your suggestion

staebler (commented):
/lgtm
/approve

wking:
No CI coverage today, so avoid all the non-required flavors:

/skip

wking:
[`registry.redhat.io` outage took out e2e-aws][1].  But looks like GitHub is complaining about vendor merge conflicts due to some parallel change.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/5645/pull-ci-openshift-installer-master-e2e-aws/1496942777617879040

staebler (commented):
/lgtm

PR #5652, opened 2022-02-17
Bug 2048686: Add mac address validation for hosts bootMACAddress

ardaguclu:
This PR adds mac address validation for all hosts bootMACAddress
field. If bootMACAddress is not empty, it should be in standard
mac address format.

### DIFF
--- a/pkg/types/baremetal/validation/platform.go
+++ b/pkg/types/baremetal/validation/platform.go
@@ -218,6 +218,8 @@ func validateHostsBase(hosts []*baremetal.Host, fldPath *field.Path, filter vali
 					hostErrs = append(hostErrs, field.Required(childName, "missing "+err.Field()))
 				case "uniqueField":
 					hostErrs = append(hostErrs, field.Duplicate(childName, err.Value()))
+				case "mac":
+					hostErrs = append(hostErrs, field.Invalid(childName, err.Value(), "invalid "+err.Field()))
 				}
 			}
 		}
--- a/pkg/types/baremetal/platform.go
+++ b/pkg/types/baremetal/platform.go
@@ -36,7 +36,7 @@ type Host struct {
 	Name            string           `json:"name,omitempty" validate:"required,uniqueField"`
 	BMC             BMC              `json:"bmc"`
 	Role            string           `json:"role"`
-	BootMACAddress  string           `json:"bootMACAddress" validate:"required,uniqueField"`
+	BootMACAddress  string           `json:"bootMACAddress" validate:"required,uniqueField,mac"`
 	HardwareProfile string           `json:"hardwareProfile"`
 	RootDeviceHints *RootDeviceHints `json:"rootDeviceHints,omitempty"`
 	BootMode        BootMode         `json:"bootMode,omitempty"`
--- a/pkg/types/baremetal/validation/platform_test.go
+++ b/pkg/types/baremetal/validation/platform_test.go
@@ -206,6 +206,12 @@ func TestValidatePlatform(t *testing.T) {
 			expected: "baremetal.hosts\\[0\\].BootMACAddress: Required value: missing BootMACAddress",
 		},
 		{
+			name: "invalid_mac",
+			platform: platform().
+				Hosts(host1().BootMACAddress("00:00:00")).build(),
+			expected: "baremetal.hosts\\[0\\].BootMACAddress: Invalid value: \"00:00:00\": invalid BootMACAddress",
+		},
+		{
 			name: "duplicate_host_name",
 			platform: platform().
 				Hosts(


ardaguclu:
@andfasano @dtantsur could you PTAL?, thanks.

pkg/types/baremetal/validation/platform.go
```diff
@@ -243,6 +243,23 @@ func validateHostsBMCOnly(hosts []*baremetal.Host, fldPath *field.Path) field.Er
 	})
 }
 
+// validateHostsBootMACAddress validates all hosts BootMACAddress are valid or not
+func validateHostsBootMACAddress(hosts []*baremetal.Host, fldPath *field.Path) field.ErrorList {
+	hostErrs := field.ErrorList{}
+
+	for idx, host := range hosts {
+		if host.BootMACAddress == "" {
```
andfasano, line 251: There is already a couple of validation rule on BootMacAddress field, in particular the `required` one should ensure that  the field is not empty: https://github.com/openshift/installer/blob/969378dc498336d1ac29baa89f75324a44467d5d/pkg/types/baremetal/platform.go#L39

pkg/types/baremetal/validation/platform.go
```diff
@@ -243,6 +243,23 @@ func validateHostsBMCOnly(hosts []*baremetal.Host, fldPath *field.Path) field.Er
...
+			continue
+		}
+
+		if err := validate.MAC(host.BootMACAddress); err != nil {
```
andfasano, line 255: Could be worth trying the already baked-in validator for mac? https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Media_Access_Control_Address_MAC

pkg/types/baremetal/validation/platform.go
ardaguclu: Thanks for review @andfasano.

That is interesting, because it is optional in upstream https://github.com/metal3-io/baremetal-operator/blob/3612e86973f1ce0fef606557c7e20dd482d8fddc/config/crd/bases/metal3.io_baremetalhosts.yaml#L108.

I guess, this is intentionally marked as required in OpenShift?

pkg/types/baremetal/validation/platform.go
ardaguclu: Yes, `validate.MAC` function just wraps up ^ function https://github.com/openshift/installer/blob/3d19350885d593ee2b1d9ecd7612c2d697dab2a3/pkg/validate/validate.go#L231.

pkg/types/baremetal/validation/platform.go
andfasano: Yes, originally it was set in https://github.com/openshift/installer/pull/3208

pkg/types/baremetal/validation/platform.go
andfasano: I think it could be enough to just add `mac` here: https://github.com/openshift/installer/blob/969378dc498336d1ac29baa89f75324a44467d5d/pkg/types/baremetal/platform.go#L39?

ardaguclu: If it works, that would be much more simpler. I'll try.

ardaguclu:
@andfasano thanks for guiding me throughout this PR. I have implemented `mac` validator and unit tests to cover it. Could you PTAL?, thanks.

pkg/types/baremetal/validation/platform.go
```diff
@@ -204,6 +204,14 @@ func validateHostsBase(hosts []*baremetal.Host, fldPath *field.Path, filter vali
 		return !valueFound
 	})
 
+	validate.RegisterValidation("mac", func(fl validator.FieldLevel) bool {
```
andfasano, line 207: Isn't `mac` an already baked-in validation? `RegisterValidation` should be used just for custom validations AFAIK

ardaguclu: I'd expect the same thing but after adding https://github.com/openshift/installer/pull/5652/files#diff-534defd3e5e7453c872c0992bd665e8c6b241d7f3baab87d8893dc0995829cbfR39 and running this test https://github.com/openshift/installer/pull/5652/files#diff-534defd3e5e7453c872c0992bd665e8c6b241d7f3baab87d8893dc0995829cbfR39. It failed. I think it is not baked-in validation or there is something I could not see.

andfasano: Probably it was a typo or cached test result? AFAIK it's a baked-in validator, so I'd expect to be invoked automatically in the `validateHostsBase`. I'd suggest giving it a try by removing the registration

ardaguclu: As you said it is baked-in validation but it just needs this;

https://github.com/openshift/installer/pull/5652/files#diff-3294dd7b4b68c688149586052c12a8cf764e28ab6ada68f60138f10fc331c4e4R221-R222

Now I think it is ready for one more review.

ardaguclu:
@sadasu could you PTAL?. Thanks.

ardaguclu:
It is clear that ipv6 is perm failing. Until it is passing;

/hold

ardaguclu:
4 hour timeout exceed problem still exists, thus;
/hold

ardaguclu:
Unfortunately after a lot of re-tests ipv6 could not have passed and this PR could not be merged. I'm closing it and I'll reopen it until ipv6 is green.

/close

PR #5657, opened 2022-02-24
Azure Stack UPI Docs: Remove Feature Gate CRs

patrickdillon:
Updates the instructions to have users remove any credential requests for feature-gated resources. Creating such credentials will cause installs to fail as the bootstrap node continuously fails to create a resource in a namespace that does not exist.

cc @mjpytlak

### DIFF
--- a/docs/user/azure/install_upi_azurestack.md
+++ b/docs/user/azure/install_upi_azurestack.md
@@ -229,6 +229,16 @@ stringData:
   azure_region: <$REGION>
 ```
 
+NOTE: Any credentials for a credential request in Tech Preview must be excluded, or they will cause the
+installation to fail. As of 4.10, there is one credential request, for the `capi-opearator`, which is in
+tech preview. Any credential requests from a feature gate can simply be removed before you create the credentials:
+
+```shell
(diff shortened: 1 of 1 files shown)


docs/user/azure/install_upi_azurestack.md
```diff
@@ -229,6 +229,16 @@ stringData:
   azure_region: <$REGION>
 ```
 
+NOTE: Any credentials for a credential request in Tech Preview must be excluded, or they will cause the
```
barbacbd, line 232: Should we consider moving this Note/Comment up or highlighting it differently as it is inserted after a large file block which means it can be skipped quite easily?

patrickdillon: I agree this could be an improvement but I don’t think we need to fine tune the Azure Stack docs too much. I think we should just merge this as is.

sadasu: I am OK to merge it as is since it is in the Cluster Credentials section where an user might look in case of failures due to cluster credentials.

sadasu:
/skip
This is a change to docs and should have no effect on the above failing CI jobs.

PR #5664, opened 2022-02-28
Bug 2048067: [Alibaba] fix location service endpoint

bd233:
Add LocationProduct, LocationEndpointType and CommonApi paramaters to use the LocationResolver get endpoint.


kirankt (approved):
/lgtm

PR #5665, opened 2022-02-28
Azure azidentity adapter

r4f4:
Microsoft is in the process of upgrading the Azure SDK for Go from V1 to V2. Our clients and authentication are on V1. The V1 authentication utilzes ADAL which will be deprecated June 30, 2022. All V2 clients, except the V2 auth client azidentity, are in beta (azidentity is scheduled to be stable in Q2 2022).[0]

These changes remove the dependency on the ADAL API, replace the authentication with azidentity, and use an adapter so the auth will work with V1 clients.

[0] https://azure.github.io/azure-sdk/releases/latest/index.html#go


pkg/asset/installconfig/azure/session.go
```diff
@@ -224,11 +226,14 @@ func newSessionFromCredentials(cloudEnv azureenv.Environment, credentials *Crede
...
 	c.Resource = cloudEnv.TokenAudience
-	authorizer, err := c.Authorizer()
+
+	cred, err := azidentity.NewClientSecretCredential(credentials.TenantID, credentials.ClientID, credentials.ClientSecret, nil)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to get client credentials authorizer")
+		return nil, errors.Wrap(err, "failed to get client credentials secret")
 	}
 
+	authorizer := azidext.NewTokenCredentialAdapter(cred, []string{azidext.DefaultManagementScope})
```
patrickdillon, line 235: I ran the azure stack CI test, but I'm pretty sure this is not going to work because the `DefaultManagementScope` is pointing to public Azure. Instead, we need to populate [DefaultAzureCredentialOptions](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#DefaultAzureCredentialOptions) with the appropriate values for the environment.

patrickdillon (commented):
In general this looks good but as I pointed out we will need to make some adjustments for azure stack and gov cloud. 

Your commit messages already follow [our contributing guidelines](https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#commit-message-format) but I would add some more context to the message for the third commit as well as the PR description (they can be the same).

Something to the effect of:
```
Microsoft is in the process of upgrading the Azure SDK for Go from V1 to V2. Our clients and authentication are on V1. The V1 authentication utilzes ADAL which will be deprecated June 30, 2022. All V2 clients, except the V2 auth client azidentity, are in beta (azidentity is scheduled to be stable in Q2 2022).[0]

These changes remove the dependency on the ADAL API, replace the authentication with azidentity, and use an adapter so the auth will work with V1 clients.

[0] https://azure.github.io/azure-sdk/releases/latest/index.html#go
```

pkg/asset/installconfig/azure/session.go
patrickdillon, line 235: I think we will also want a comment here about how/why we are using `azidext` because we will eventually plan on removing it once all the clients are from the `/sdk` modules.

patrickdillon:
Ok, so the azurestack tests are failing as we expected:

> Service returned an error. Status=401 Code="InvalidAuthenticationTokenAudience" Message="The access token has been obtained for wrong audience or resource 'https://management.azure.com/'. It should exactly match with one of the allowed audiences 'https://management.stackpoc.com/81c9b804-ec9e-4b5a-8845-1d197268b1f5','https://graph.windows.net/'."

I think we have a good idea of how to fix this.

pkg/asset/installconfig/azure/session.go
```diff
@@ -224,11 +235,17 @@ func newSessionFromCredentials(cloudEnv azureenv.Environment, credentials *Crede
...
+	cred, err := azidentity.NewClientSecretCredential(credentials.TenantID, credentials.ClientID, credentials.ClientSecret, nil)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to get client credentials authorizer")
+		return nil, errors.Wrap(err, "failed to get client credentials secret")
 	}
 
+	scopes := []string{endpointToScope(cloudEnv.ResourceManagerEndpoint)}
+	// Use an adapter so azidentity in the Azure Management libraries can be
+	// used in place of Authorizer when calling Azure Management APIs.
+	authorizer := azidext.NewTokenCredentialAdapter(cred, scopes)
```
patrickdillon, line 247: The error message you're seeing now in CI is because no tenant is being provided. Instead of using `azidext.NewTokenCredentialAdapter` try using `azidext.NewDefaultAzureCredentialAdapter`. The `*azidentity.DefaultAzureCredentialOptions` will take the tenantID and I _think_ `AuthorityHost` should be set to `cloudEnv.ActiveDirectoryEndpoint`.

r4f4: The problem with the `azidext.NewDefaultAzureCredentialAdapter` is that it uses the `azidentity.NewDefaultAzureCredential` which, according to the docs,
 ```
It attempts to authenticate with each of these credential types, in the following order, stopping when one provides a token: - EnvironmentCredential - ManagedIdentityCredential - AzureCLICredential
```
So I don't see a way for us to pass in the client credentials through code.

Comment (patrickdillon) [reply pkg/asset/installconfig/azure/session.go:247]: You are correct, my mistake! You are indeed already passing the tenant to `NewClientSecretCredentials`, but   `AuthorityHost` is nil. So I think we need to figure out what `AuthorityHost` should be, perhaps `cloudEnv.ActiveDirectoryEndpoint`.

**Comment by r4f4** [comment]
/hold
Needs https://github.com/openshift/release/pull/26725 so we can test against Azure Gov cloud as well.

File: pkg/asset/installconfig/azure/session.go [hunk 3]
```diff
@@ -224,9 +235,31 @@ func newSessionFromCredentials(cloudEnv azureenv.Environment, credentials *Crede
...
+	case azure.StackCloud.Name():
+		logrus.Debug("Falling back to deprecated ADAL authentication")
+		// The new authorization method is not yet supported for private clouds
+		// See https://github.com/Azure/azure-sdk-for-go/pull/16942
+		var err error
+		authorizer, err = c.Authorizer()
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to get client credentials authorizer")
+		}
+	default:
```
patrickdillon, line 250: It may take some time to get CI setup on govcloud so I'm testing this manually:

```
FATAL failed to fetch Master Machines: failed to load asset "Install Config": platform.azure.region: Invalid value: "usdodcentral": region "usdodcentral" does not support resource creation 
```

I think we need another switch statement here for govcloud and to pass in govcloud as the authority host as described here (this example uses China, but I believe govcloud should behave similarly): https://github.com/Azure/azure-sdk-for-go/issues/16638

r4f4:
/unhold
We are not blocking this one on the CI job anymore

pkg/asset/installconfig/azure/session.go
```diff
@@ -216,6 +218,15 @@ func saveCredentials(credentials Credentials, filePath string) error {
...
 }
 
+const defaultScope = "/.default"
+
+func endpointToScope(endpoint string) string {
+	if endpoint[len(endpoint)-1] != '/' {
+		endpoint += "/"
+	}
+	return endpoint + defaultScope
+}
```
patrickdillon, line 228: nit: can you move this helper function down below?

patrickdillon (commented):
This looks good. Can you squash the commits? Also some minor suggestions.

patrickdillon:
looks like there are some `gofmt` linting errors

r4f4:
I had to invert the logic in the `switch` statement because the `cloudEnv.Name` for the AzureStack case was being set to `HybridEnvironment` and it seems it can also be set to other arbitrary values.

pkg/asset/installconfig/azure/session.go
```diff
@@ -224,9 +226,38 @@ func newSessionFromCredentials(cloudEnv azureenv.Environment, credentials *Crede
...
+
+		scopes := []string{endpointToScope(cloudEnv.ResourceManagerEndpoint)}
+		// Use an adapter so azidentity in the Azure SDK can be used as
+		// Authorizer when calling the Azure Management Packages, which we
+		// currently use. Once the Azure SDK clients (found in /sdk) move to
+		// stable, we can update our clients and they will be able to use the
+		// creds directly without the authorizer. The schedule is here:
+		// https://azure.github.io/azure-sdk/releases/latest/index.html#go
+		authorizer = azidext.NewTokenCredentialAdapter(cred, scopes)
+	default:
```
patrickdillon, line 252: This works, but having Azure Stack as a default is counter intuitive because it is a much less common use case. I would like to make this more explicit. Let's go back to the earlier option we discussed of passing `cloudName azure.CloudEnvironment` as a parameter to this function and keep the switch statement logic similar to what it was before.

patrickdillon:
This LGTM. Just would like to see the tests run

PR #5666, opened 2022-02-28
Bug 2059213: build all terraform providers and terraform binary locally

staebler:
The installer targets architectures (s390x, for example) that are not supported officially by terraform. This means that there are no official terraform binaries to download and no public registries for the terraform providers. Consequently, we must abandon the idea of downloading publicly-available, pre-built binaries.


staebler:
b7e448747...07c03fe8a
* Change the versions used for the terraform providers and the terraform binary to match the versions tested in https://github.com/openshift/installer/pull/5507.
* Add descriptions to commit messages.

staebler:
07c03fe8a...dca52957d
* Use the right version of the aws terraform provider--namely v3.70.0 rather than v3.7.0.

PR #5670, opened 2022-03-02
doc: terraform maintenance document

staebler:
Add a dev document for describing how to maintain the terraform binaries embedded in the installer.

### DIFF
--- a/docs/dev/terraform.md
+++ b/docs/dev/terraform.md
@@ -1,0 +1,109 @@
+# Embedded Terraform in the Installer
+
+The installer runs terraform in order to provision the infrastructure for a cluster. However, there is a design goal
+for the installer that the user performing the installation should not be required to have any software pre-installed
+in their environment. This design goal precludes the user from, for example, being required to have any containerization
+software available. To get around this for terraform, the installer embeds within its binary the needed terraform
+binaries. At installation time, the terraform binaries are extracted from the installer into the local file system.
+
+## Design
+
+### Building terraform binaries
+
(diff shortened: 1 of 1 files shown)


docs/dev/terraform.md
```diff
@@ -0,0 +1,108 @@
...
+terraform provider.
+
+## Update the version of terraform or a provider
+
+To update the version of terraform or a provider, modify the version in the go.mod file for the relevant sub-package.
+To update the version of terraform, change the version of the required module in ./terraform/terraform/go.mod. To update
+the version of the aws terraform provider, change the version of the required module in ./terraform/aws/go.mod.
+
+After updating the require statement in the go.mod file, perform the following actions in the sub-package.
+1. Remove all of the replace statements in the go.mod file.
```
patrickdillon, line 106: I don't understand the context/significance of this step. Can you spell it out a bit? My confusion is: if there were replaces I would expect them to be there for a reason...

Also I spot checked the go.mod in a handful of providers and the only replace statement I saw was in vsphere for our fork. I'm not sure how that replace fits in with this step.

staebler: Ah, this is a mistake on my part. I meant to say remove all of the indirect requires. I will fix it up.

As for replaces, if there are any, they should match the replaces from the terraform provider. For example, the ibm provider has replaces for github.com/softlayer/softlayer-go and github.com/dgrijalva/jwt-go.

staebler: Fix in 28ff94e5f...bfa3cd984.

PR #5671, opened 2022-03-02
OCPBUGS-2384: [Alibaba] fix the creation of public record

bd233:
When the publish is 'Internal', need not to create public record


patrickdillon:
/retitle OCPBUGS-2384: [Alibaba] fix the creation of public record

PR #5672, opened 2022-03-02
terraform: build binaries on go.mod changes

staebler:
Only re-build the terraform binary or a terraform provider binary when there has been a change to the go.mod file in the relevant sub-module.


patrickdillon:
/approve

This LGTM but still wrapping my head around (learning to read) the Makefile

PR #5673, opened 2022-03-02
Bug 1918005: vsphere: Use Managed Object ID for networks instead of potentially duplicate name.

bostrt:
vSphere Network names are not necessarily unique, especially in NSX-T environments so we must use the unique Managed Object ID for a network to fix installations in these type of environments.

Related to [SPLAT-340](https://issues.redhat.com/browse/SPLAT-340) and [SPLAT-379](https://issues.redhat.com/browse/SPLAT-379)

Originally worked in https://github.com/openshift/installer/pull/5376 but submitting new PR to review usage of vSphere simulator for unit tests.


pkg/asset/installconfig/vsphere/validation_test.go
```diff
@@ -91,16 +93,28 @@ func TestValidate(t *testing.T) {
...
+	}
+
+	client, _, err := mock.GetClient(server)
+	if err != nil {
+		t.Error(err)
+		return
+	}
 
-	vsphereClient := mock.NewMockFinder(mockCtrl)
+	/*vsphereClient := mock.NewMockFinder(mockCtrl)
```
kirankt, line 108: Let's remove this comment block.

kirankt (commented):
It would be nice to see changes squashed into 2 commits: one for code changes and another for vendoring.

pkg/asset/installconfig/vsphere/validation_test.go
bostrt: Thanks for catching this!

bostrt:
> It would be nice to see changes squashed into 2 commits: one for code changes and another for vendoring.

Cleaned up in latest rebase

bostrt:
If these changes looks acceptable, let me know and I will move the bugzilla link from https://github.com/openshift/installer/pull/5376 to _this_ PR

patrickdillon (commented):
Thanks for tackling this.

There are quite a few changes in this PR and some of them are a bit subtle, at least for us who may not work on vSphere regularly. 

One question I have after reading [some history](https://github.com/openshift/installer/pull/5376#issuecomment-970741931) is about the nature of these duplicate names. Is it possible for the port groups to have duplicate names and also the same datacenter and cluster, or is that prevented by vSphere? Or can only port groups in different datacenters and/or clusters have the same name?

bostrt:
>  Is it possible for the port groups to have duplicate names and also the same datacenter and cluster, or is that prevented by vSphere? Or can only port groups in different datacenters and/or clusters have the same name?

Unfortunately, yes and no. vSphere alone will _not_ allow for duplicated port groups within a datacenter (or in a cluster). 

However, we recently learned that NSX-T _will_ [allow for duplicated port group names at basically any level](https://kb.vmware.com/s/article/79872#Duplicate_names); previously we only thought it would allow duplication at a datacenter level. 

We're reevaluating right now to see what the options are. 

/hold

bostrt:
> we recently learned that NSX-T will [allow for duplicated port group names at basically any level](https://kb.vmware.com/s/article/79872#Duplicate_names); previously we only thought it would allow duplication at a datacenter level.
> We're reevaluating right now to see what the options are.

tl;dr We can fix the biggest duplicate network name situation and keep the door open for other less likely situations. 
.

@patrickdillon So there are 3 known situations in which duplicate network names could appear:

A. NSX-T Segment in a single transport zone could be duplicated as defined in bugzilla.
B. NSX-T Segments with same name under same transport zone. However, we believe this situation would be unlikely.
C. A standard port group and distributed port group with same name. We've seen this reported from the field before but it was found to be a mis-configuration. 

This PR (as it stands right now) would resolve issue A.

I think this PR is still a step in the right direction to handle network naming in vSphere/NSX-T but as mentioned, it won't address each situation. Unfortunately, if we were to handle situations B and C, we would have to start using Managed Object IDs which I think would start diminishing user experience.

pkg/asset/installconfig/vsphere/vsphere.go
```diff
@@ -303,9 +305,12 @@ func getNetwork(ctx context.Context, path string, finder Finder, client *vim25.C
 		return "", errors.New("did not find any networks")
 	}
 	if len(networks) == 1 {
-		n := networks[0].(networkNamer)
-		logrus.Infof("Defaulting to only available network: %s", n.Name())
-		return n.Name(), nil
+		n, err := vspheretypes.GetNetworkName(ctx, client, networks[0])
```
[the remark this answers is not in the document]
bostrt: Good question :) This change was needed since the networks slice now contains generic `ManagedObjectReference` and no longer the `NetworkReference` that contains `Name()`. AFAIK, we can't easily get a `NetworkReference` out of a `ManagedObjectReference` without going through the network factory used in `GetNetworkName`.

pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -77,7 +78,12 @@ func folderExists(finder Finder, ic *types.InstallConfig, fldPath *field.Path) f
 	return nil
 }
 
-func validateNetwork(finder Finder, p *vsphere.Platform, fldPath *field.Path) field.ErrorList {
+func validateNetwork(client *vim25.Client, finder *find.Finder, p *vsphere.Platform, fldPath *field.Path) field.ErrorList {
+	// It's not possible to validate a network if datacenter or cluster are empty strings
+	if p.Datacenter == "" || p.Cluster == "" {
```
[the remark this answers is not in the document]
bostrt: I'll take a look a this. This was added due to a failing test, but the test can be adjusted instead of adding this check.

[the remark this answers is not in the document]
bostrt: Looking at this again, without this check in place, it would result in unfriendly errors like this when the Cluster isn't specified in install config:
```
FATAL failed to fetch Terraform Variables: failed to fetch dependency of "Terraform Variables": failed to generate asset "Platform Provisioning Check": [platform.vsphere.cluster: Required value: must specify the cluster, platform.vsphere.network: Invalid value: "TestDC": datacenter './TestDC' not found] 
```

_With_ these changes, the error shown in case of missing cluster will remain the same and much more user friendly:
```
FATAL failed to fetch Terraform Variables: failed to fetch dependency of "Terraform Variables": failed to generate asset "Platform Provisioning Check": platform.vsphere.cluster: Required value: must specify the cluster 
```

jcpowermac:
vsphere e2e failures are unrelated to PR. Looks like vCenter restapi failing

pkg/asset/installconfig/vsphere/vsphere.go
```diff
@@ -137,7 +128,7 @@ func getClients() (*vCenterClient, error) {
 
 	// There is a noticeable delay when creating the client, so let the user know what's going on.
 	logrus.Infof("Connecting to vCenter %s", vcenter)
-	vim25Client, restClient, err := vspheretypes.CreateVSphereClients(context.TODO(),
+	vim25Client, restClient, err := CreateVSphereClients(context.TODO(),
```
patrickdillon, line 131: Other cloud platforms store the client in a session object in the install config (check out aws or azure). Would that approach work for vsphere?

pkg/tfvars/vsphere/vsphere.go
```diff
@@ -56,6 +60,22 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
...
+		sources.Username,
+		sources.Password)
+	networkID, err := vsphere.GetNetworkMoID(context.TODO(),
+		vim25Client,
+		controlPlaneConfig.Workspace.Datacenter,
+		sources.Cluster,
+		controlPlaneConfig.Network.Devices[0].NetworkName)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get vSphere network ID")
+	}
```
patrickdillon, line 77: Let's find the networkID in `tfvars` in the `asset` pkg and pass it in here in `TFVarsSources`.

patrickdillon (commented):
This PR has quite a few changes, and some substantial (but welcome) ones to the vsphere code. It would be easier to review if these were arranged around commits (e.g. a commit for moving the client to the asset package). That said, the overall intent of the PR looks good. I left comments about several details we need to iron out. We can continue to collaborate on these changes.

The simulator is not related to solving this bug and should probably be in a separate PR. If we want to cherry pick this BZ, it definitely needs to move out.

I assume the simulator allows for mocking? We are looking at expanding our unit test coverage for aws with mocking, and it would be good to do the same with vsphere. Even if the simulator does not include mocking, we can use both gomock and the simulator in the test bed.

pkg/asset/installconfig/vsphere/vsphere.go
bostrt: So technically, the installer is leaking session here (they never get logged out). We did some research on this, across OpenShift, in https://issues.redhat.com/browse/SPLAT-319 and actually meeting with the Govmomi community this week to see if improving session management _inside_ govmomi would be suitable. 

I'm a little hesitant to add yet another session management layer to an OpenShift component with the other work about to start that I mentioned above. 

If the govmomi community doesn't want to move forward with session management improvements in govmomi, I am certainly happy to revisit what you point out.

pkg/asset/installconfig/vsphere/client.go
```diff
@@ -2,30 +2,118 @@ package vsphere
...
+}
+
+// GetClusterNetworks returns a slice of Managed Object references for vSphere networks in the given Datacenter
+// and Cluster. The given NetworkIdentifier and Finder are used to query vSphere API.
+func GetClusterNetworks(ctx context.Context, finder *find.Finder, datacenter, cluster string) ([]types.ManagedObjectReference, error) {
+	ctx, cancel := context.WithTimeout(ctx, 60*time.Second)
+	defer cancel()
+
+	// Get vSphere Cluster resource in the given Datacenter.
+	path := fmt.Sprintf("/%s/host/%s", datacenter, cluster)
```
bostrt, line 68: @jcpowermac I'm afraid if a customer uses a folder for their clusters this will break. Do you know if this is common and we should account for it?

pkg/asset/installconfig/vsphere/validation_test.go
```diff
@@ -43,10 +44,11 @@ func validIPIInstallConfig() *types.InstallConfig {
 }
 
 func TestValidate(t *testing.T) {
+	t.SkipNow()
```
bostrt, line 47: I'm marking this PR WIP just to be safe while this line is present. I'm skipping so we don't fail on unit test and I can still see e2e run.

I have to skip due to the `nil` being passed in `validationMethod` further below (it would result in test fail). I'm doing this because this is where get into a tough spot with mocking Govmomi. In order for the validation methods to proceed I would need to mocking a number of Govmomi functions for validating the network.

I'll keep looking at a way we can unit test, without having to mock a whole subsystem of Govmomi.

pkg/asset/installconfig/vsphere/mock/vsphere_sim.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	context2 "context"
+	"crypto/tls"
+	"encoding/pem"
+	"errors"
+	"github.com/vmware/govmomi/find"
+	"github.com/vmware/govmomi/session"
+	"github.com/vmware/govmomi/simulator"
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/soap"
```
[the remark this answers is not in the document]
bostrt: I will address this in a separate PR for simulator addition.

pkg/asset/installconfig/vsphere/client.go
```diff
@@ -2,30 +2,118 @@ package vsphere
...
+	netObj := object.NewNetwork(client, ref)
+	name, err := netObj.ObjectName(ctx)
+	if err != nil {
+		return "", errors.Wrap(err, "could not get network name")
+	}
+	return name, nil
+}
+
+// GetNetworkMoID returns the unique Managed Object ID for given network name inside of the given Datacenter
+// and Cluster. The given NetworkIdentifier and Finder are used to query vSphere API.
```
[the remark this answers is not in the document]
bostrt: Good catch! I've addressed this.

pkg/asset/installconfig/vsphere/vsphere.go
bostrt: > I'm a little hesitant to add yet another session management layer to an OpenShift component with the other work about to start that I mentioned above.

After meeting with the Govmomi community today, there won't be any work on session management improvements in the near future, due to lots of design work needed. I will plan on a separate PR to address session issues here if that's okay. I've created [SPLAT-471](https://issues.redhat.com/browse/SPLAT-471) for this task.

pkg/tfvars/vsphere/vsphere.go
bostrt: Done.

pkg/asset/installconfig/vsphere/validation_test.go
bostrt: Since we're only doing negative tests in `vsphere/validation_test.go` right now, the `nil` client _shouldn't_ matter. Running unit tests locally seems to work out.

bostrt: Tests failed, I'll keep checking it out. My local test actually failed too, I just haven't had enough coffee yet.

pkg/asset/installconfig/vsphere/client.go
jcpowermac: Yep

jcpowermac:
With https://bugzilla.redhat.com/show_bug.cgi?id=2063829
have we tested this if port groups are contained within a folder? Doesn't need to be in the real dev enviornments, maybe just the simulator.

pkg/asset/cluster/tfvars.go
```diff
@@ -20,6 +20,7 @@ import (
 	configv1 "github.com/openshift/api/config/v1"
 	machinev1 "github.com/openshift/api/machine/v1"
 	machinev1beta1 "github.com/openshift/api/machine/v1beta1"
+
```
patrickdillon, line 23: remove the extra line

patrickdillon (commented):
This looks good to me except one nit for an extra line.

But I believe the issue with a cluster in folders is still unresolved, correct? https://github.com/openshift/installer/pull/5673/commits/928f5f5e9786e5621f58e32a16b64b47a0fdca6e#r833729403

pkg/asset/installconfig/vsphere/client.go
jcpowermac: I don't think this is a concern. We would have hit this already since the terraform provider does this:
https://github.com/openshift/installer/blob/master/terraform/providers/vsphereprivate/resource_vsphereprivate_import_ova.go#L148

bostrt:
> But I believe the issue with a cluster in folders is still unresolved, correct? https://github.com/openshift/installer/pull/5673#discussion_r833729403

There haven't been any reports (customer cases or bz) of this being an issue so I'm not sure we need to address in this PR.

patrickdillon:
/approve

This LGTM. @jcpowermac or @rvanderp3, will one of you `/lgtm`?

PR #5674, opened 2022-03-02
terraform: upgrade alicloud provider to v1.158.0

staebler:
When https://github.com/openshift/installer/pull/5507 was merged, the version of the alicloud terraform provider was set to v1.148.0, which is the version in use when the PR was created. However, in the time between creation and merge, https://github.com/openshift/installer/pull/5571 was merged, which upgrade the version to v1.153.0. This change was lost during conflict resolution.

Let's upgrade to the latest version (v1.158.0) of the alicloud terraform provider while we are at it.


staebler:
/cc @kwoodson 
Would you mind giving this a review, please?

kwoodson:
Testing cluster install.

staebler:
> Testing cluster install.

Thanks. Note that this is restoring the version that is used in 4.10. I am open to upgrading to an even more recent version, too. As of now, that would be v1.160.0.

staebler:
> Unfortunately I'm still seeing this behavior.

Sorry, for the confusion. I did not mean to imply that the upgrade of the terraform provider would fix that problem. I am just trying to rectify the mistaken change in versions from 4.10 to 4.11.

kwoodson:
@staebler I'm supportive of this change. Moving back to the 1.158 sounds good to me. If you want to try moving forward I can test that as well.

sadasu:
/close

Latest version is now 1.182.0. Will upgrade to the latest when we see a need.

PR #5676, opened 2022-03-03
creation of a node CR file

sairameshv:
These code changes help in the creation of a Node CR during the bootstrap process.
Reference: https://github.com/openshift/api/blob/master/config/v1/0000_10_config-operator_01_node.crd.yaml
EP: https://github.com/openshift/enhancements/blob/master/enhancements/worker-latency-profile/worker-latency-profile.md


sairameshv:
/assign @jhixson74

cc: @harche @rphillips

go.mod
```diff
@@ -37,8 +37,8 @@ require (
 	github.com/metal3-io/baremetal-operator v0.0.0-20220128094204-28771f489634
 	github.com/metal3-io/baremetal-operator/apis v0.0.0
 	github.com/metal3-io/baremetal-operator/pkg/hardwareutils v0.0.0
-	github.com/openshift/api v0.0.0-20220222102030-354aa98a475c
-	github.com/openshift/client-go v0.0.0-20211025111749-96ca2abfc56c
+	github.com/openshift/api v0.0.0-20220302114424-24043075985b
```
patrickdillon, line 40: Please move vendoring into its own commit to ease reviewing.

pkg/asset/manifests/node.go
```diff
@@ -0,0 +1,82 @@
+package manifests
+
+import (
+	"github.com/ghodss/yaml"
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"path/filepath"
```
patrickdillon, line 7: nit: please move this into its own import block and then delete the following blank line.

import blocks should follow:

```
<stdlib>

<external>

<openshift>
```

pkg/asset/manifests/node.go
```diff
@@ -0,0 +1,82 @@
...
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/installer/pkg/asset"
+	"github.com/openshift/installer/pkg/asset/installconfig"
+)
+
+var (
+	nodeCfgFilename = filepath.Join(manifestDir, "cluster-node-02-config.yml")
+)
+
+// Node generates the cluster-node-*.yml files.
```
patrickdillon, line 18: nit: 

```
// Node generates the cluster-node-02-config.yml file.
```

pkg/asset/manifests/node.go
```diff
@@ -0,0 +1,82 @@
...
+		&installconfig.ClusterID{},
+		&installconfig.InstallConfig{},
+	}
+}
+
+// Generate generates the Node config and its CRD.
+func (n *Node) Generate(dependencies asset.Parents) error {
+	clusterID := &installconfig.ClusterID{}
+	installConfig := &installconfig.InstallConfig{}
+	dependencies.Get(clusterID, installConfig)
```
patrickdillon, line 43: It does not look you are using these dependencies.

patrickdillon (commented):
The installer does not seem like the right place for this... do we need this now that https://github.com/openshift/cluster-config-operator/pull/233 has merged?

go.mod
sairameshv: Sure

sairameshv:
> The installer does not seem like the right place for this... do we need this now that [openshift/cluster-config-operator#233](https://github.com/openshift/cluster-config-operator/pull/233) has merged?

Yes, I think these changes are required in order to render the MCO manifests apart from the cluster-config-operator changes

pkg/asset/manifests/node.go
sairameshv: Addressed

pkg/asset/manifests/node.go
sairameshv: Addressed

pkg/asset/manifests/node.go
sairameshv: If we have to support the WorkerLatencyProfiles via the install-config (in future), the dependency may be required. Anyway, as of today, this is not really required.

staebler:
> > The installer does not seem like the right place for this... do we need this now that [openshift/cluster-config-operator#233](https://github.com/openshift/cluster-config-operator/pull/233) has merged?
> 
> Yes, I think these changes are required in order to render the MCO manifests apart from the cluster-config-operator changes

@sairameshv  Can you elaborate on this? I would expect the MCO to be resilient to the absent of the Node CR. This does not appear to be for the rendering done by the bootstrap's MCO, which is the only reason why I would expect the installer to need to define this CR. Having the manifest come from the installer does not provide any guarantees that the CR will be applied to the cluster any faster than having the manifest come from the release payload.

sairameshv:
Thanks for your time on the reviews. Based on the [above](https://github.com/openshift/installer/pull/5676#issuecomment-1062876167) conversation, closing this PR.
/close

PR #5678, opened 2022-03-04
Bug 2060508: azurestack: fix backend address pools for internal publishing

staebler:
When using internal publishing, the backend address pools of the load balancers incorrectly used a null list for the external load balancer's ID. Terraform requires that the list supplied to the concat function be an empty list rather than a null list.

Also, the elb_backend_pool_v4_id variable used as input to the bootstrap stage did not have a default. For internal publishing, the elb_backend_pool_v4_id output from the vnet stage is null, which is omitted from the outputs rather than explicitly set to null. Consequently, terraform would complain that the required elb_backend_pool_v4_id value was not set.


jstuever (commented):
/lgtm

patrickdillon:
skip

PR #5681, opened 2022-03-04
pkg: Remove redundant code

rna-afk:
Removing a few redundant nil checks and files that might not
be needed for the installer. Might help with readability. Feel
free to suggest more changes or close this if it's insignificant.

### DIFF
--- a/pkg/asset/installconfig/platform.go
+++ b/pkg/asset/installconfig/platform.go
@@ -58,66 +58,33 @@ func (a *platform) Generate(asset.Parents) error {
 	switch platform {
 	case alibabacloud.Name:
 		a.AlibabaCloud, err = alibabacloudconfig.Platform()
-		if err != nil {
-			return err
-		}
 	case aws.Name:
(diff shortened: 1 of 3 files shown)


staebler (commented):
Generally, I am not in favor of these time of changes to error handling. For the first two files, I could accept the changes, although I would prefer to deal with the error as close to where it happens as possible. That keeps me from having to wonder about why the error is not being handled and having to scroll down from where I am reading to see what the behavior is.

I am more strongly opposed to the changes in the third file. If the first function call, such as `gcpclient.NewClient`, returns an error, then we are returning immediately. If the second function call, such as `gcpconfig.ValidatePreExistingPublicDNS`, returns an error, then we are not returning that error until later. That has a few problems. (1) There is a second error variable that I have to wonder about the significance of. (2) There are different patterns for the two errors despite the observable behavior of the two being the same. (3) If a new function is added for which we need to handle the error, it would become even more difficult to reason about why only the last function does not have its error handled immediately.

PR #5688, opened 2022-03-08
Bug 2061544: azurestack: stop pinning to Standard_LRS for disk type

staebler:
Azure Stack supports disks types other than Standard_LRS. The installer should not overriding the disk type choice made by the user and silently restricting the disk type for the control plane to Standard_LRS.


patrickdillon:
Looks good to me. Will tag it after CI run.

PR #5691, opened 2022-03-08
Bug 2061549: azurestack: create the api DNS record when publishing internally

staebler:
When using internal publishing on Azure Stack Hub, the DNS record for the api should be created and point to the internal load balancer.


staebler:
This builds on https://github.com/openshift/installer/pull/5678. That PR should merge first.
/hold

staebler:
8419e9c4a...30e83662e
* Rebased after #5678 merged.

/hold cancel

PR #5692, opened 2022-03-09
Validate disk encryption sets

jhixson74:
Add validation for disk encryption sets

https://issues.redhat.com/browse/CORS-1938

This currently uses #5641 until merged.


pkg/asset/installconfig/azure/validation.go
```diff
@@ -53,6 +53,32 @@ func Validate(client API, ic *types.InstallConfig) error {
 	return allErrs.ToAggregate()
 }
 
+// ValidateDiskEncryptionSet ensures the disk encryption set exists and is valid.
+func validateDiskEncryptionSet(client API, ic *types.InstallConfig) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if ic.ControlPlane != nil && ic.ControlPlane.Platform.Azure != nil && ic.ControlPlane.Platform.Azure.OSDisk.DiskEncryptionSet != nil {
```
patrickdillon, line 60: We should also check the default pool.

pkg/asset/installconfig/azure/validation_test.go
```diff
@@ -127,6 +128,59 @@ var (
...
+		ic.Compute[0].Platform.Azure.OSDisk.DiskEncryptionSet.SubscriptionID = ""
+	}
+	invalidateDiskEncryptionSetComputeResourceGroup = func(ic *types.InstallConfig) {
+		ic.Compute[0].Platform.Azure.OSDisk.DiskEncryptionSet = validDiskEncryptionSetConfig()
+		ic.Compute[0].Platform.Azure.OSDisk.DiskEncryptionSet.ResourceGroup = ""
+	}
+	invalidateDiskEncryptionSetComputeName = func(ic *types.InstallConfig) {
+		ic.Compute[0].Platform.Azure.OSDisk.DiskEncryptionSet = validDiskEncryptionSetConfig()
+		ic.Compute[0].Platform.Azure.OSDisk.DiskEncryptionSet.Name = ""
+	}
```
patrickdillon, line 183: Having these test variations on the different fields is confusing because it does not test code we wrote: these field variations are different ways for the API to fail in the same way. 

From what I see the test state matrix is 6 different states: 3 machine pools (master, worker, default) x 2 DiskEncryptionSetStates (found, not found). I could be convinced otherwise. For example, there might be a way to factor out the machine pools, but it is not obvious to me.

patrickdillon (commented):
Looks good. Left some minor suggestions.

pkg/asset/installconfig/azure/validation.go
jhixson74: Fixed.

pkg/asset/installconfig/azure/validation_test.go
jhixson74: Fixed.

patrickdillon:
I was expecting 76ad24584320cbb36751276b26de1ae1380a3794 to disappear after #5641 merged, but it is still showing up as changes in this PR. Do you know why those changes are still here? Perhaps try rebasing?

PR #5695, opened 2022-03-09
Bug 2047670: aws: remove validation check for internal publish strategy

jhixson74:
This was keeping ValidateForProvisioning from checking if the zone was
associated with a VPC or not.

https://issues.redhat.com/browse/OCPBUGSM-40036

### DIFF
--- a/pkg/asset/installconfig/aws/validation.go
+++ b/pkg/asset/installconfig/aws/validation.go
@@ -349,7 +349,7 @@ var requiredServices = []string{
 
 // ValidateForProvisioning validates if the install config is valid for provisioning the cluster.
 func ValidateForProvisioning(session *session.Session, ic *types.InstallConfig, metadata *Metadata) error {
-	if ic.Publish == types.InternalPublishingStrategy {
+	if ic.Publish == types.InternalPublishingStrategy && ic.AWS.HostedZone == "" {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/asset/installconfig/aws/validation.go
```diff
@@ -349,10 +349,6 @@ var requiredServices = []string{
 
 // ValidateForProvisioning validates if the install config is valid for provisioning the cluster.
 func ValidateForProvisioning(session *session.Session, ic *types.InstallConfig, metadata *Metadata) error {
-	if ic.Publish == types.InternalPublishingStrategy {
```
staebler, line 352: This will cause installs with internal publishing to fail when the user is not supplying a private hosted zone. Specifically, the later check for the public hosted zone should not be performed.

jhixson74: The prior code just returned if not hosted zone was supplied. Is the correct logic here to return if no hosted zone was provided and using internal publishing?

patrickdillon:
/retest

This seems reasonable to me:
/approve

But I would like appreciate it if @staebler could take a second look. On a related note:

I think this identifies a lack in our unit test coverage. I do not have a lot of experience with the AWS assets and did not feel confident reviewing this minor change. I think there is a significant degree of state that we have to keep in mind when reviewing this kind of code ("Oh what about internal publishing?" "Oh what if they bring their own networks?") and I think we can capture that in code. I created https://issues.redhat.com/browse/CORS-1947 for this idea, and while it needs further grooming I think the idea is there and would appreciate any feedback.

mhanss:
@jhixson74 I have verified the defect using this PR but one of the test cases is still failing. Please check the defect comment ([Bugzilla bug 2047670](https://bugzilla.redhat.com/show_bug.cgi?id=2047670)) for more details.

jhixson74:
> @jhixson74 I have verified the defect using this PR but one of the test cases is still failing. Please check the defect comment ([Bugzilla bug 2047670](https://bugzilla.redhat.com/show_bug.cgi?id=2047670)) for more details.

I looked at the bug and that particular issue has been addressed. Can you verify it's been fixed?

staebler (commented):
/lgtm
/approve

PR #5697, opened 2022-03-11
[wip] gather bootstrap console logs

jhixson74:
Gather serial logs through the Azure and GCP SDK. There is some platform abstraction here, but it could be improved. Some terraform state could also be passed along. Feedback welcome. 

https://issues.redhat.com/browse/CORS-1942


cmd/openshift-install/gather.go
```diff
@@ -132,38 +137,53 @@ func runGatherBootstrapCmd(directory string) (string, error) {
...
+func gatherBootstrap(bootstrap string, port int, masters []string, directory string) (string, string, error) {
+	gatherID := time.Now().Format("20060102150405")
+
+	serialLogBundle := filepath.Join(directory, fmt.Sprintf("serial-log-bundle-%s.tar.gz", gatherID))
+	serialLogBundlePath, err := filepath.Abs(serialLogBundle)
+
+	logrus.Info("Pulling platform specific console logs from the bootstrap machine")
+	gather, err := gather.New(logrus.StandardLogger(), serialLogBundle, bootstrap, masters, directory)
+	if err != nil {
+		return "", "", errors.Wrap(err, "Failed to create gather object")
```
patrickdillon, line 158: nit: `Failed` -> `failed`
Perhaps: `failed to gather console logs`

cmd/openshift-install/gather.go
```diff
@@ -132,38 +137,53 @@ func runGatherBootstrapCmd(directory string) (string, error) {
...
+	}
+	err = gather.Run()
+	if err != nil {
+		logrus.Infof("%s: Failed to gather platform specific logs", err.Error())
```
patrickdillon, line 162: nit: instead of "platform specific"... "failed to gather VM console logs"

cmd/openshift-install/gather.go
```diff
@@ -132,38 +137,53 @@ func runGatherBootstrapCmd(directory string) (string, error) {
...
 	}
-	path, err := filepath.Abs(file)
+	logBundlePath, err := filepath.Abs(file)
 	if err != nil {
-		return "", errors.Wrap(err, "failed to stat log file")
+		return "", "", errors.Wrap(err, "failed to stat log file")
 	}
-	return path, nil
+
+	return serialLogBundlePath, logBundlePath, nil
```
patrickdillon, line 186: We should try to have the serial console logs added to the bootstrap log bundle (rather than two separate bundles).

cmd/openshift-install/gather.go
patrickdillon, line 158: Should be best effort: don't stop here if we fail, keep trying to get the rest of the logs.

patrickdillon (commented):
I haven't had a chance to review the platform-specific AWS code in-depth, but in terms of evaluating this as a general approach for how to gather serial logs across platforms: 

I think this approach seems very promising. We discussed an alternative approach of using the Terraform outputs, but this approach has at least one advantage that we don't create a dependency (a dependency that could be non-obvious) on those TF outputs.

@staebler I would be interested to get your input on the higher-level design here.

pkg/gather/azure/azure.go
```diff
@@ -0,0 +1,293 @@
...
+	defer cancel()
+
+	var bootDiagnostics []*compute.BootDiagnosticsInstanceView
+	for _, vm := range virtualMachines {
+		if vm.VirtualMachineProperties.DiagnosticsProfile != nil &&
+			vm.VirtualMachineProperties.DiagnosticsProfile.BootDiagnostics != nil &&
+			to.Bool(vm.VirtualMachineProperties.DiagnosticsProfile.BootDiagnostics.Enabled) == true {
+			instanceView, err := g.virtualMachinesClient.InstanceView(ctx, g.resourceGroupName, to.String(vm.Name))
+			if err != nil {
+				g.logger.Debugf("failed to get instance view: %s", err.Error())
```
staebler, line 189:
```suggestion
				g.logger.Debugf("failed to get instance view: %v", err)
```

pkg/gather/azure/azure.go
```diff
@@ -0,0 +1,293 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	return credentials, nil
+}
+
+func getVirtualMachines(ctx context.Context, g *Gather) ([]compute.VirtualMachine, error) {
```
staebler, line 155: The approach of fetching all machines works for Azure, where everything is contained in a resource group specific to the cluster. However, for other platforms that do not have the concept of a resource group, we will need filtering on the virtual machines. For Azure, we would grab boot diagnostics for workers. For other platforms, we would presumably not grab boot diagnostics for workers. I would rather see a consistent approach across platforms.

staebler (commented):
This general approach looks very good. My biggest concern is the filtering of which machines to consider for fetching boot diagnostics.

cmd/openshift-install/gather.go
```diff
@@ -132,38 +137,53 @@ func runGatherBootstrapCmd(directory string) (string, error) {
...
 }
 
-func gatherBootstrap(bootstrap string, port int, masters []string, directory string) (string, error) {
+func gatherBootstrap(bootstrap string, port int, masters []string, directory string) (string, string, error) {
+	gatherID := time.Now().Format("20060102150405")
+
+	serialLogBundle := filepath.Join(directory, fmt.Sprintf("serial-log-bundle-%s.tar.gz", gatherID))
+	serialLogBundlePath, err := filepath.Abs(serialLogBundle)
+
+	logrus.Info("Pulling platform specific console logs from the bootstrap machine")
```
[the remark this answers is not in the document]
jhixson74: Fixed.

cmd/openshift-install/gather.go
jhixson74: Fixed.

cmd/openshift-install/gather.go
jhixson74: Fixed.

cmd/openshift-install/gather.go
jhixson74: Fixed.

cmd/openshift-install/gather.go
jhixson74: Fixed.

cmd/openshift-install/gather.go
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/gather/azure/azure.go
```diff
@@ -0,0 +1,293 @@
...
+
+func getBootDiagnostics(ctx context.Context, virtualMachines []compute.VirtualMachine, g *Gather) []*compute.BootDiagnosticsInstanceView {
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+
+	var bootDiagnostics []*compute.BootDiagnosticsInstanceView
+	for _, vm := range virtualMachines {
+		if vm.VirtualMachineProperties.DiagnosticsProfile != nil &&
+			vm.VirtualMachineProperties.DiagnosticsProfile.BootDiagnostics != nil &&
+			to.Bool(vm.VirtualMachineProperties.DiagnosticsProfile.BootDiagnostics.Enabled) == true {
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/gather/azure/azure.go
jhixson74: Fixed.

pkg/gather/azure/azure.go
```diff
@@ -0,0 +1,293 @@
...
+
+	// clean up the mess we've made.
+	serialLogBundleDir := filepath.Join(g.directory, strings.TrimSuffix(filepath.Base(g.serialLogBundle), ".tar.gz"))
+	_, err := os.Stat(serialLogBundleDir)
+	if err == nil && !strings.HasPrefix(serialLogBundleDir, ".") {
+		err := os.RemoveAll(serialLogBundleDir)
+		if err != nil {
+			g.logger.Debugf("failed to remove archive directory: %s", err.Error())
+		}
+	}
```
r4f4, line 230: We are repeating this block of code in all of the providers. Maybe this could be part of the top-level gather like `CreateArchive`?

jhixson74: Agreed.

jhixson74: Fixed.

patrickdillon (commented):
I didn't have a chance to review all of this code thoroughly but hopefully this feedback should be helpful.

It would be easier to review this PR if it were broken up into commits. I would suggest there should at least be commits for the different providers.

In addition to the points about the structure of the log bundles, I think it would be good to remove the old log bundles if we are able to successfully combine them.

cmd/openshift-install/gather.go
```diff
@@ -151,19 +172,27 @@ func gatherBootstrap(bootstrap string, port int, masters []string, directory str
...
-	path, err := filepath.Abs(file)
+
+	clusterLogBundlePath, err := filepath.Abs(file)
 	if err != nil {
 		return "", errors.Wrap(err, "failed to stat log file")
 	}
-	return path, nil
+
+	logBundlePath := filepath.Join(filepath.Dir(clusterLogBundlePath), fmt.Sprintf("log-bundle-%s.tar.gz", gatherID))
+	err = gather.CombineArchives(logBundlePath, map[string]string{serialLogBundlePath: "serial"})
```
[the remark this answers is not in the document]
jhixson74: I'm curious to see what is happening here. The CombineArchive() function just adds a subdirectory named "serial" and places the serial logs under there. The directory structure remains the same.

pkg/gather/azure/azure.go
```diff
@@ -0,0 +1,289 @@
...
+
+	gather.accountsClient = storage.NewAccountsClientWithBaseURI(session.Environment.ResourceManagerEndpoint, gather.subscriptionID)
+	gather.accountsClient.Authorizer = gather.authorizer
+
+	return gather, nil
+}
+
+// Run is the entrypoint to start the gather process.
+func (g *Gather) Run() error {
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
```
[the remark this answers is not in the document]
jhixson74: We do this all over the installer. I've never bothered to look it up, but if it is indeed an issue, it's something that needs to be done in a lot more places. I follow this pattern because it's what has been in the installer since I began working on it.

pkg/gather/gather.go
```diff
@@ -0,0 +1,179 @@
...
+
+	combinedDirectory := strings.TrimSuffix(archiveName, suffix)
+	if archiveName[0] == '.' || archiveName[0] == '/' {
+		combinedDirectory = strings.TrimSuffix(filepath.Base(archiveName), suffix)
+	}
+
+	for archive, subDirectory := range archives {
+		_, err := os.Stat(archive)
+		if err != nil {
+			logrus.Warnf("unable to stat %s, skipping", archive)
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/gather/azure/azure.go
```diff
@@ -0,0 +1,289 @@
...
+	"bytes"
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob"
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
+	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-06-01/storage"
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/gather/azure/azure.go
```diff
@@ -0,0 +1,289 @@
...
+	environment           azureenv.Environment
+	cloudName             azure.CloudEnvironment
+	resourceGroupName     string
+	logger                logrus.FieldLogger
+	serialLogBundle       string
+	bootstrap             string
+	masters               []string
+	directory             string
+	virtualMachinesClient compute.VirtualMachinesClient
+	accountsClient        storage.AccountsClient
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/gather/azure/azure.go
```diff
@@ -0,0 +1,289 @@
...
+
+	return gather, nil
+}
+
+// Run is the entrypoint to start the gather process.
+func (g *Gather) Run() error {
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+	defer cancel()
+
+	credentials, err := getSharedKeyCredentials(ctx, g)
```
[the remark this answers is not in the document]
jhixson74: Fixed.

pkg/gather/azure/azure.go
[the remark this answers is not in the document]
patrickdillon: > We do this all over the installer. I've never bothered to look it up, but if it is indeed an issue, it's something that needs to be done in a lot more places. I follow this pattern because it's what has been in the installer since I began working on it.

Let's make sure we're talking about the same issue. I should have been more clear to say "this context timeout is being shared." Typically in the installer, we pass a background or TODO context until we're actually making the API call. That is fine.

What I'm pointing out is that in this code I think you're setting a 1 minute deadline for 3 subsequent API calls. I don't think this pattern is common in the installer, but if it is, we can examine it.

I have put together a go playground to illustrate the problem here: https://go.dev/play/p/974k2h_rCEl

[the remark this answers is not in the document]
jhixson74: Fixed.

cmd/openshift-install/gather.go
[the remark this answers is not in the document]
patrickdillon: Are you getting different results in your testing?

patrickdillon:
Why pull in 7c27e3c55f06b64b44075a5095da914fb773b6bb by @r4f4? That PR depends on this one, but the more complex we make this one the harder it becomes to review & merge. 

If it's coming down to a deadline we might want to group them together so we make sure they get in, but at this stage it would probably be most efficient to make sure #5807 is reviewed and that the pieces that #5807 depends on can be merged.

cmd/openshift-install/gather.go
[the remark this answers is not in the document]
patrickdillon: @jhixson74 it occurred to me that we should actually be able to test this in CI. If you add a DO NOT MERGE commit which would cause bootstrap to fail, say: break a critical manifest, or cause an error in bootkube.sh, then CI should run gather bootstrap and collect the bundle in CI, so reviewers can examine it there.

wking:
#5807 merged.  Sounds like there's nothing left to do here?  Feel free to re-open if I'm misunderstanding.

/close

PR #5700, opened 2022-03-11
Add VPCRegionForPowerVSRegion function

hamzy:
This function will be used to create a CloudProvider config for
Power VS among other potential uses.

### DIFF
--- a/pkg/types/powervs/powervs_regions.go
+++ b/pkg/types/powervs/powervs_regions.go
@@ -1,5 +1,9 @@
 package powervs
 
+import (
+	"fmt"
+)
+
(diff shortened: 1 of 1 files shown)


clnperez:
@patrickdillon another Power VS PR. Can we get this one merged before the manifest assets one (https://github.com/openshift/installer/pull/5610)? I'm going to make a small change to that one but need this first. Trying to cut down on the number of PRs since CI is such a time-consuming gate for these.

patrickdillon:
> since CI is such a time-consuming gate for these.

Agree. I'm really hoping we can devote some time to improving this experience ASAP.

PR #5701, opened 2022-03-11
Azure: Add ultraSSDCapability to control plane instances.

jstuever:
This change adds the ulstraSSDCapability option to the control plane
instances. Users will now be able to configure the control plane
instances to use ultra ssd disks via the install-config.yaml.

https://issues.redhat.com/browse/CORS-1893


pkg/types/azure/machinepool.go
```diff
@@ -24,6 +24,12 @@ type MachinePool struct {
 	//
 	// +optional
 	OSDisk `json:"osDisk"`
+
+	// ultraSSDCapability defines if the instance should use Ultra SSD disks.
+	// The valid values are Enabled, Disabled.
+	//
+	// +optional
+	UltraSSDCapability string `json:"ultraSSDCapability,omitempty"`
```
patrickdillon, line 32: As this is essentially an enum, creating  a type for the valid values would make reading this a little easier. See Azure's `CloudEnvironment` for example: https://github.com/openshift/installer/blob/master/pkg/types/azure/platform.go#L89-L114

patrickdillon (commented):
Looks good. One nit. The UI is pretty straight forward, worth adding tests?

pkg/types/azure/machinepool.go
jstuever: I don't believe the readability benefit outweighs the cost in terms of binary size as well as additional CPU usage requirements to wrap a string in this way.

patrickdillon:
/approve

needs rebase

data/data/azure/variables-azure.tf
```diff
@@ -57,6 +57,11 @@ variable "azure_master_root_volume_size" {
   description = "The size of the volume in gigabytes for the root block device of master nodes."
 }
 
+variable "azure_control_ultra_ssd_enabled" {
```
patrickdillon, line 60: Just noticed this:

`control` is inconsistent with `master` in the rest of the variables. I support moving from "master" to "control plane" even if it is inconsistent, which is how I'm reading the intent here. If that's not the intent, I'm fine with "master."

My issue, here, though is I think the variable name may be a little unclear. To me it sounds like this is controlling the ultra ssd enabled, rather than a reference to the control plane. 

Have you considered `azure_control_plane_ultra_ssd_enabled`, `azure_cp_ultra_ssd_enabled` or any other alternatives? Same comment would apply to the gocode variable in tfvars.

jstuever: This is indeed an effort to move toward the compute/controlPlane nomenclature. I recognize the ambiguity of the word "control" but leaned into it because `wc control` == `wc compute`. I am not against using the more verbose `control plane` naming scheme. I'll make those changes now.

PR #5702, opened 2022-03-14
Add exit code for infra, bootstrip and install failures

xueqzhan:
Attempt to add some additional exit codes to differentiate errors from infrastructure, boothstrap and cluster phases. 

- Exit code for infrastructure phase only applies to "create cluster" command. 
- Exit code for bootstrap catches the result from bootstrap watch phase. 
- Exit code for install catches the final result from cluster installation. Right now the code keeps checking until context is done, or clusterversion is Available. It does not exit because any other reasons. So exitCodeIntallFailed pretty much means the cluster installation did not complete within the specified timeframe.

### DIFF
--- a/cmd/openshift-install/create.go
+++ b/cmd/openshift-install/create.go
@@ -45,6 +45,13 @@ type target struct {
 	assets  []asset.WritableAsset
 }
 
+const (
+	exitCodeInstallConfigError = iota + 3
+	exitCodeInfrastructureFailed
+	exitCodeBootstrapFailed
+	exitCodeInstallFailed
+)
+
 // each target is a variable to preserve the order when creating subcommands and still
 // allow other functions to directly access each target individually.
 var (
@@ -128,7 +135,7 @@ var (
 						}
 						logrus.Infof("Bootstrap gather logs captured here %q", bundlePath)
 					}
-					logrus.Fatal("Bootstrap failed to complete")
+					logrus.Exit(exitCodeBootstrapFailed)
 				}
 				timer.StopTimer("Bootstrap Complete")
 				timer.StartTimer("Bootstrap Destroy")
@@ -151,7 +158,8 @@ var (
 						logrus.Error("Attempted to gather ClusterOperator status after installation failure: ", err2)
 					}
 					logTroubleshootingLink()
-					logrus.Fatal(err)
+					logrus.Error(err)
+					logrus.Exit(exitCodeInstallFailed)
 				}
 				timer.StopTimer(timer.TotalTimeElapsed)
 				timer.LogSummary()
@@ -267,6 +275,14 @@ func runTargetCmd(targets ...asset.WritableAsset) func(cmd *cobra.Command, args 
 
 		err := runner(rootOpts.dir)
 		if err != nil {
+			if strings.Contains(err.Error(), asset.InstallConfigError) {
+				logrus.Error(err)
+				logrus.Exit(exitCodeInstallConfigError)
+			}
+			if strings.Contains(err.Error(), asset.ClusterCreationError) {
+				logrus.Error(err)
+				logrus.Exit(exitCodeInfrastructureFailed)
+			}
 			logrus.Fatal(err)
 		}
 		if cmd.Name() != "cluster" {
--- a/cmd/openshift-install/waitfor.go
+++ b/cmd/openshift-install/waitfor.go
@@ -55,7 +55,7 @@ func newWaitForBootstrapCompleteCmd() *cobra.Command {
 				logrus.Info("openshift-install gather bootstrap --help")
 				logrus.Error("Bootstrap failed to complete: ", err.Unwrap())
 				logrus.Error(err.Error())
-				logrus.Fatal("Bootstrap failed to complete")
+				logrus.Exit(exitCodeBootstrapFailed)
 			}
 
 			logrus.Info("It is now safe to remove the bootstrap resources")
@@ -89,7 +89,8 @@ func newWaitForInstallCompleteCmd() *cobra.Command {
 					logrus.Error("Attempted to gather ClusterOperator status after wait failure: ", err2)
 				}
 				logTroubleshootingLink()
-				logrus.Fatal(err)
+				logrus.Error(err)
+				logrus.Exit(exitCodeInstallFailed)
 			}
 			timer.StopTimer(timer.TotalTimeElapsed)
 			timer.LogSummary()
--- a/pkg/asset/asset.go
+++ b/pkg/asset/asset.go
@@ -11,6 +11,13 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	// ClusterCreationError is the error when terraform fails, implying infrastructure failures
+	ClusterCreationError = "failed to create cluster"
+	// InstallConfigError wraps all configuration errors in one single error
+	InstallConfigError = "failed to create install config"
+)
+
 // Asset used to install OpenShift.
 type Asset interface {
 	// Dependencies returns the assets upon which this asset directly depends.
--- a/pkg/asset/cluster/cluster.go
+++ b/pkg/asset/cluster/cluster.go
@@ -163,7 +163,7 @@ func (c *Cluster) applyTerraform(tmpDir string, platform string, stage terraform
 	}
 
 	if applyErr != nil {
-		return nil, errors.Wrap(applyErr, "failed to create cluster")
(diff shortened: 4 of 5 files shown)


patrickdillon:
@xueqzhan Thanks for submitting this pull request. I think a PR such as this, which is effectively establishing an API that would be used by a range of components, should go through the enhancement process. 

We do have an ongoing project to create structured output for failures. There are various ways of conveying this information. Do you have a particular use case where codifying the error in the exit code is useful?

cmd/openshift-install/create.go
```diff
@@ -267,6 +274,10 @@ func runTargetCmd(targets ...asset.WritableAsset) func(cmd *cobra.Command, args
 
 		err := runner(rootOpts.dir)
 		if err != nil {
+			if cmd.Name() == "cluster" {
+				logrus.Error(err)
+				logrus.Exit(exitCodeInfrastructureFailed)
```
staebler, line 279: This will categorize errors that are not infrastructure errors as such. One example is if the install-config.yaml is invalid.

deads2k:
> We do have an ongoing project to create structured output for failures. There are various ways of conveying this information. Do you have a particular use case where codifying the error in the exit code is useful?

Good question.  Yes, there is a consumer ready for a coarse grained exit code for a few different states.  Our CI step registry now watches the exit code coming from the installer binary: https://github.com/openshift/release/pull/26694/files.  The goal is not fine-grained diagnostics, but coarse-grained categorization.

Starting with
1. infrastructure
2. bootstrap
3. wait-for-cluster
4. everything else

is a good starting point and does not preclude future refinement.  These different failure modes are chosen because the focus of investigation, which tools can be brought to bear, how we react in CI steps, and which teams need to investigate are different for each one.  The expectation is still that a human looks at the failures to do the investigation beyond these categories.

Adding other coarse categories based on the stage of the installer, not the failure itself, is ok for future refinement of exit codes.  However, this will never be a replacement for fine grained, machine-parseable feedback.

xueqzhan:
@patrickdillon David explained our usage for ci jobs. Do you think we still should go through the enhancement process? Our goal is for better categorization of CI failures.

cmd/openshift-install/create.go
staebler: You could have the Cluster.Generate return a specific error type when the failure is from applying the terraform.

staebler:
> @patrickdillon David explained our usage for ci jobs. Do you think we still should go through the enhancement process? Our goal is for better categorization of CI failures.

I recommend still going through the enhancement process. Your use case is CI, but using specific error codes applies to all users. That assumes that we are going to accept that the error codes are an API output of the installer and offer some compatibility guarantees as such.

xueqzhan:
@patrickdillon @staebler I can't get all the jobs passing. But I believe all required ones passed.

cmd/openshift-install/create.go
```diff
@@ -267,6 +285,10 @@ func runTargetCmd(targets ...asset.WritableAsset) func(cmd *cobra.Command, args
 
 		err := runner(rootOpts.dir)
 		if err != nil {
+			if cmd.Name() == "cluster" && isClusterCreationError(err) {
```
patrickdillon, line 288: Can you look into adding something similar for install config validation errors?

xueqzhan: @patrickdillon I looked at the code a little more. And I am a little concerned about this. There are some errors that are obvious config related errors. But there are others I am not sure. For example, can I treat all errors returned from PreTerraform config errors? Or do I have to go into the function and differentiate different errors? Take an example of aws.PreTerraform, I can categorize PrivateSubnets and PublicSubnets as config errors. But how about errors from client interaction with AWS (such as add tags to subnet, or add tags to hosted zone)? Having to differentiate errors within those calls will be a nightmare. This also applies to other platforms like azure. 

The above is only for cluster asset. Many other assets are generated for “create cluster”: cluster.Metadata, machine.MasterIgnitionCustomizations, machine.WorkerIgnitionCustomizations, cluster.TerraformVariables, kubeconfig.AdminClient, password.KubeadminPassword, tls.JournalCertKey. Each of those have their own set of config errors. Can you clarify the ask is to cover all config errors?

xueqzhan: I just pushed a commit with some sample changes for config errors. This is not a complete list in any way. 

- I ignored marshing/unmarshing errors. 
- This change only applies to aws
- I have not considered some of the dependency assets

There should be a lot more places if this is the way to go. @patrickdillon I just want to confirm this is what you have in mind.

xueqzhan: @patrickdillon Thanks for pointing out the proper place for catching the validation errors. Now it looks much cleaner.

cmd/openshift-install/create.go
```diff
@@ -230,6 +238,26 @@ func newCreateCmd() *cobra.Command {
...
+	for err != nil {
+		if err.Error() == asset.ClusterCreationError {
+			return true
+		}
+		err = errors.Unwrap(err)
+	}
+	return false
+}
+
+func isConfigurationError(err error) bool {
```
patrickdillon, line 251: It would be easy to create a common function that handles both of these and you pass in either `asset.InstallConfigError` or `assetClusterCreationError`

cmd/openshift-install/create.go
```diff
@@ -230,6 +238,26 @@ func newCreateCmd() *cobra.Command {
...
 }
 
+func isClusterCreationError(err error) bool {
+	for err != nil {
+		if err.Error() == asset.ClusterCreationError {
+			return true
+		}
+		err = errors.Unwrap(err)
+	}
+	return false
```
patrickdillon, line 248: This logic doesn't work, because the unwrapping unwraps the outermost errors first:
```
(string) (len=243) "failed to fetch Metadata: failed to load asset \"Install Config\": failed to create install config: platform.aws.serviceEndpoints.region: Invalid value: \"us-east-22\": dial tcp: lookup ec2.us-east-22.amazonaws.com on 192.168.86.1:53: no such host"
(string) (len=243) "failed to fetch Metadata: failed to load asset \"Install Config\": failed to create install config: platform.aws.serviceEndpoints.region: Invalid value: \"us-east-22\": dial tcp: lookup ec2.us-east-22.amazonaws.com on 192.168.86.1:53: no such host"
(string) (len=217) "failed to load asset \"Install Config\": failed to create install config: platform.aws.serviceEndpoints.region: Invalid value: \"us-east-22\": dial tcp: lookup ec2.us-east-22.amazonaws.com on 192.168.86.1:53: no such host"
(string) (len=217) "failed to load asset \"Install Config\": failed to create install config: platform.aws.serviceEndpoints.region: Invalid value: \"us-east-22\": dial tcp: lookup ec2.us-east-22.amazonaws.com on 192.168.86.1:53: no such host"
(string) (len=178) "failed to create install config: platform.aws.serviceEndpoints.region: Invalid value: \"us-east-22\": dial tcp: lookup ec2.us-east-22.amazonaws.com on 192.168.86.1:53: no such host"
(string) (len=178) "failed to create install config: platform.aws.serviceEndpoints.region: Invalid value: \"us-east-22\": dial tcp: lookup ec2.us-east-22.amazonaws.com on 192.168.86.1:53: no such host"
(string) (len=145) "platform.aws.serviceEndpoints.region: Invalid value: \"us-east-22\": dial tcp: lookup ec2.us-east-22.amazonaws.com on 192.168.86.1:53: no such host"
```

Perhaps we want to use [`Is`](https://pkg.go.dev/errors#Is)

cmd/openshift-install/create.go
```diff
@@ -267,6 +295,16 @@ func runTargetCmd(targets ...asset.WritableAsset) func(cmd *cobra.Command, args
 
 		err := runner(rootOpts.dir)
 		if err != nil {
+			if cmd.Name() == "cluster" {
```
patrickdillon, line 298: This either needs to be:
1. `if cmd.Name() == "cluster" || if cmd.Name() == "install-config" {` 
or
2. removed

I think removing it is fine. I don't think we need the check.

patrickdillon (changes requested):

cmd/openshift-install/create.go
xueqzhan: Good catch. Unfortunately Is does not work for us here. I debugged it a little and I think the lower layer (e.g withMessage) does not support Is(). So my option will be to use strings.Contain. As long as the final error msg contains the string we are looking for, that is a match.

cmd/openshift-install/create.go
xueqzhan: Removed

PR #5703, opened 2022-03-14
bootkube: Use `--image-references` for MCO bootstrap

cgwalters:
Depends: https://github.com/openshift/machine-config-operator/pull/3017

I want to add a new image (e.g. `rhel-coreos` as part of layering)
and this will avoid a tedious "ratching" process to plumb that
through the installer.  Now the MCO can just directly read the image
references.


cgwalters:
(Draft since this is not tested yet, also this is a good time to object or propose a different plan)

staebler (commented):
This seems like a reasonable approach to me. I like the idea of pushing the domain knowledge of which images MCO is using to the MCO repo, instead of having it in the installer repo.

PR #5712, opened 2022-03-16
Update SR-IOV & DPDK doc for OpenStack

EmilienM:
In 4.11, we did the following:

* Remove the dependency on config-drive and support OpenStack Metadata
  URL fetching.
* Automate the configuration of VFIO NO IOMMU
* Add support for vhostuser interfaces for DPDK

This patch adds more content for DPDK, with an example of a pod using
hugepages, dedicated CPUs and the DPDK port.

It also add more content for SR-IOV, like how to create a policy & a
pod.

Later, we'll add content for SR-IOV with HW offload.

### DIFF
--- a/docs/user/openstack/deploy_sriov_workers.md
+++ b/docs/user/openstack/deploy_sriov_workers.md
@@ -22,80 +22,7 @@ in OpenShift, and that your tenant has access to them. Your OpenStack cluster mu
 For all clusters that use single-root input/output virtualization (SR-IOV), RHOSP compute nodes require a flavor that supports [huge pages][huge-pages].
 Deploying worker nodes with SR-IOV networks is supported as a post-install operation for both IPI and UPI workflows. After you verify that your OpenStack cluster can support SR-IOV in OpenShift and you install an OpenShift cluster that meets the [minimum requirements](README.md#openstack-requirements), use the following steps and examples to create worker nodes with SR-IOV NICs.
 
-## Preparing your OpenShift Cluster to use SR-IOV Networks
-
-Before you use single root I/O virtualization (SR-IOV) on a cluster that runs on Red Hat OpenStack Platform (RHOSP), make the RHOSP metadata service mountable as a drive and enable the No-IOMMU Operator for the virtual function I/O (VFIO) driver.
-
-### Enabling the RHOSP metadata service as a mountable drive
-
-Create a machine config in your worker machine pool that makes the OpenStack metadata service available as a mountable drive. This machine config enables the SR-IOV operator to get the UUID of networks from OpenStack.
-
-```yaml
-kind: MachineConfig
-apiVersion: machineconfiguration.openshift.io/v1
-metadata:
-  name: 20-mount-config
-  labels:
-    machineconfiguration.openshift.io/role: worker
-spec:
-  osImageURL: ''
-  config:
-    ignition:
-      version: 2.2.0
-    systemd:
-      units:
-        - name: create-mountpoint-var-config.service
-          enabled: true
-          contents: |
-            [Unit]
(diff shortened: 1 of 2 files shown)


docs/user/openstack/ovs-dpdk.md
```diff
@@ -156,93 +156,91 @@ spec:
...
-ansible-playbook play.yaml -e network_ids="ID of VFIO network" --extra-vars "mc_config_file=/tmp/vhostuser.yaml"
-oc create -f /tmp/vhostuser.yaml
+```
+apiVersion: sriovnetwork.openshift.io/v1
+kind: SriovNetworkNodePolicy
+metadata:
+  name: dpdk1
+  namespace: openshift-sriov-network-operator
+spec:
+  deviceType: vfio-pci
```
MaysaMacedo, line 184: should we mention the `isRdma` here?

docs/user/openstack/deploy_sriov_workers.md
```diff
@@ -279,6 +204,12 @@ spec:
             name: worker-user-data
 ```
```
MaysaMacedo, line 206: Shoud we mention the the creation of the NodePolicy here from? maybe just link the example from dpdk?

MaysaMacedo (commented):
/lgtm
/hold I have a minor remarks, feel free to remove the hold if there is no need to include them

docs/user/openstack/ovs-dpdk.md
EmilienM: it's mentioned L193

docs/user/openstack/deploy_sriov_workers.md
EmilienM: Good idea.

docs/user/openstack/ovs-dpdk.md
EmilienM: i've added it, it was a good idea. Thanks

docs/user/openstack/deploy_sriov_workers.md
EmilienM: done

EmilienM:
@mandre please review

docs/user/openstack/deploy_sriov_workers.md
```diff
@@ -279,6 +204,85 @@ spec:
...
 
+Once the workers are deployed, you must label them SR-IOV capable:
```
mandre, line 207: Not sure if this is more correct but it sounds better to my ears as a non-native English speaker: 
```suggestion
Once the workers are deployed, you must label them as SR-IOV capable:
```

docs/user/openstack/deploy_sriov_workers.md
```diff
@@ -279,6 +204,85 @@ spec:
...
+  resourceName: sriov1
+```
+
+Note: If the network device plugged to the network is not from Intel and is from Mellanox, then `deviceType` must be set to `netdevice` and `isRdma` set to `true`. 
+
+The SR-IOV network operator will automatically discover the devices connected on that network for each worker, and make them available for use by the CNF pods later.
+
+## Deploy a testpmd pod
+
+This pod is an example how we can create a container that uses the hugepages, the reserved CPUs and the SR-IOV port:
```
mandre, line 245:
```suggestion
This pod is an example of how we can create a container that uses the hugepages, the reserved CPUs and the SR-IOV port:
```

mandre (commented):
This makes the workflow so much better - thanks @EmilienM for all your work on SR-IOV and DPDK.
Happy to put an LGTM once this doc has been reviewed by the SDN team.

mandre (commented):
/lgtm
/approve

PR #5713, opened 2022-03-16
baremetal: refresh owners list

andfasano:
Refresh the baremetal owners list to reflect latest organizational changes


ardaguclu:
@andfasano since I have transferred to another team and probably will not be able to review PRs anymore, would you mind also removing me from approver list. Thanks.

PR #5715, opened 2022-03-16
cluster: unpack terraform binaries into install directory

staebler:
Currently, the installer unpacks the terraform and terraform providers binaries to the same temporary directory under /tmp that is used for the plan files for a stage. Some users, however, are not able to execute files under /tmp.

A workaround for this would be for the user to set the TMPDIR environment variable to a directory where they do have the ability to execute files. That is not a great experience for a user when there is already a command-line argument that the user can use to specify the install directory.

So, let's use the install directory as the location under which to unpack the terraform binaries instead. At the same time, let's unpack once for all of the stages rather than unpacking once per stage.

One downside of this approach is that either (1) it makes the Cluster asset dependent upon the install directory or (2) it requires that the binaries be added as assets. The second option is not appropriate as it would significantly increase the size of .openshift_install_state.json. The first option is accomplished, for the time being, by having a package-scoped variable set by the create command that the Cluster asset can reference. This ensures that the asset graph itself is not polluted with the install directory and the install directory can be moved without affecting the state. However, it is an ugly encapsulation smell.

https://issues.redhat.com/browse/CORS-1952

### DIFF
--- a/cmd/openshift-install/create.go
+++ b/cmd/openshift-install/create.go
@@ -27,6 +27,7 @@ import (
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
 	routeclient "github.com/openshift/client-go/route/clientset/versioned"
 	"github.com/openshift/installer/pkg/asset"
+	"github.com/openshift/installer/pkg/asset/cluster"
 	"github.com/openshift/installer/pkg/asset/installconfig"
 	"github.com/openshift/installer/pkg/asset/logging"
 	assetstore "github.com/openshift/installer/pkg/asset/store"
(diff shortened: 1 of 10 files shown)


staebler:
2a963a79a...fc75b79aa
* Reconcile the changes made to alibaba stages in https://github.com/openshift/installer/pull/5682

hardys:
I think this may have broken baremetal-ipi like:

```
"bootstrap" stage: failed to create cluster: failed doing terraform init: fork/exec ocp/ostest/terraform/bin/terraform: no such file or directory
```

I'll propose a revert to test that in CI

PR #5357, opened 2021-11-08
vSphere: Add the ability to attach a tag to a mob

jcpowermac:
The upstream terraform provider does not provide
a resource to tag an already existing vCenter object.
To resolve this pr adds an additional resource
to the vsphereprivate terraform provider called
`vsphereprivate_tag_attach`. It takes
the tag id, the managed object id and the managed object
type.

This will be used in the region and zone modifications.

PR #5716, opened 2022-03-16
vsphereprivate: update to v2 terraform sdk

jcpowermac:


patrickdillon:
/approve

This LGTM. I'll give it some time in case any one else wants to take a look.

patrickdillon:
/hold
Let's sort out the relatoinship between this and #5357 because it looks like these changes are being pulled in there, which may make more sense as it has the relevant context.

patrickdillon:
Rereading the comments from #5357 I see that there is no connection which is the reason this was split out. My mistake.

/hold cancle

staebler (commented):
/lgtm

PR #5718, opened 2022-03-17
Azure: Fine-tune wording around ADAL auth

patrickdillon:
ADAL authentication will be deprecated soon, but is not quite deprecated yet. Make the wording a little more politic.


r4f4:
I'm proposing we revert the whole change in https://github.com/openshift/installer/pull/5722

patrickdillon:
/close
not needed with #5722

PR #5720, opened 2022-03-17
Set ignition string to sensitive

jcpowermac:
We are leaking the ignition into the .openshift_install.log


staebler (commented):
Good catch. Thanks.
/lgtm
/approve

PR #5721, opened 2022-03-17
azure: Enable HyperVGeneration setting for Azure disks

rna-afk:
Currently, azure disks do not support HyperVGeneration version 2
so adding a field in the install config and set it to either V1
or the preferred V2 if the instance type supports it.

This includes adding another OS image for version 2 and picking
the right image depending on the HyperVGeneration version set and
passing the version to terraform to output the right image name
used.

### DIFF
--- a/data/data/azure/variables-azure.tf
+++ b/data/data/azure/variables-azure.tf
@@ -163,3 +163,10 @@ This determined whether User defined routing will be used for egress to Internet
 When false, Standard LB will be used for egress to the Internet.
 EOF
 }
+
+variable "azure_hypervgeneration_version" {
+  type = string
+  description = <<EOF
+This determines the HyperVGeneration disk type to use for the control plane VMs.
+EOF
+}
--- a/data/data/azure/vnet/main.tf
+++ b/data/data/azure/vnet/main.tf
@@ -102,3 +102,16 @@ resource "azurerm_image" "cluster" {
     blob_uri = azurerm_storage_blob.rhcos_image.url
   }
 }
+
+resource "azurerm_image" "clustergen2" {
+  name                = "${var.cluster_id}-gen2"
+  resource_group_name = data.azurerm_resource_group.main.name
+  location            = var.azure_region
+  hyper_v_generation  = "V2"
+
+  os_disk {
+    os_type  = "Linux"
+    os_state = "Generalized"
+    blob_uri = azurerm_storage_blob.rhcos_image.url
+  }
+}
--- a/data/data/azure/vnet/outputs.tf
+++ b/data/data/azure/vnet/outputs.tf
@@ -51,7 +51,7 @@ output "resource_group_name" {
 }
 
 output "vm_image" {
-  value = azurerm_image.cluster.id
+  value = var.azure_hypervgeneration_version == "V2" ? azurerm_image.clustergen2.id : azurerm_image.cluster.id
 }
 
 output "identity" {
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -27,6 +27,7 @@ import (
 	"github.com/openshift/installer/pkg/asset/ignition/machine"
 	"github.com/openshift/installer/pkg/asset/installconfig"
 	awsconfig "github.com/openshift/installer/pkg/asset/installconfig/aws"
+	aztypes "github.com/openshift/installer/pkg/asset/installconfig/azure"
 	gcpconfig "github.com/openshift/installer/pkg/asset/installconfig/gcp"
 	ovirtconfig "github.com/openshift/installer/pkg/asset/installconfig/ovirt"
 	"github.com/openshift/installer/pkg/asset/machines"
@@ -298,7 +299,6 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		if err != nil {
 			return err
 		}
-
 		auth := azuretfvars.Auth{
 			SubscriptionID: session.Credentials.SubscriptionID,
 			ClientID:       session.Credentials.ClientID,
@@ -320,6 +320,12 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		workerConfigs := make([]*machinev1beta1.AzureMachineProviderSpec, len(workers))
 		for i, w := range workers {
 			workerConfigs[i] = w.Spec.Template.Spec.ProviderSpec.Value.Object.(*machinev1beta1.AzureMachineProviderSpec)
+		}
+		client := aztypes.NewClient(session)
+		hyperVGeneration, err := client.GetHyperVGenerationVersion(context.TODO(), masterConfigs[0].VMSize,
+			masterConfigs[0].OSDisk.ManagedDisk.StorageAccountType, masterConfigs[0].Location)
+		if err != nil {
+			return err
 		}
 
 		preexistingnetwork := installConfig.Config.Azure.VirtualNetwork != ""
@@ -351,6 +357,7 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 				OutboundType:                    installConfig.Config.Azure.OutboundType,
 				BootstrapIgnStub:                bootstrapIgnStub,
 				BootstrapIgnitionURLPlaceholder: bootstrapIgnURLPlaceholder,
+				HyperVGeneration:                hyperVGeneration,
 			},
 		)
 		if err != nil {
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -13,7 +13,6 @@ import (
 	aznetwork "github.com/Azure/azure-sdk-for-go/profiles/2018-03-01/network/mgmt/network"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/pkg/errors"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"github.com/openshift/installer/pkg/types"
@@ -112,16 +111,6 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 			if cpus < float64(req.minimumVCpus) {
 				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d vCPUsAvailable", req.minimumVCpus)
 				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
-			}
-		} else if strings.EqualFold(*capability.Name, "HyperVGenerations") {
-			generations := sets.NewString()
-			for _, g := range strings.Split(to.String(capability.Value), ",") {
-				g = strings.TrimSpace(g)
-				g = strings.ToUpper(g)
-				generations.Insert(g)
-			}
-			if !generations.Has("V1") {
-				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "only disks with HyperVGeneration V1 are supported"))
 			}
 		} else if strings.EqualFold(*capability.Name, "MemoryGB") {
 			memory, err := strconv.ParseFloat(*capability.Value, 0)
--- a/pkg/asset/machines/azure/machines.go
+++ b/pkg/asset/machines/azure/machines.go
@@ -20,7 +20,7 @@ const (
 )
 
 // Machines returns a list of machines for a machinepool.
-func Machines(clusterID string, config *types.InstallConfig, pool *types.MachinePool, osImage, role, userDataSecret string) ([]machineapi.Machine, error) {
+func Machines(clusterID string, config *types.InstallConfig, pool *types.MachinePool, osImage, role, userDataSecret string, hyperVGen string) ([]machineapi.Machine, error) {
 	if configPlatform := config.Platform.Name(); configPlatform != azure.Name {
 		return nil, fmt.Errorf("non-Azure configuration: %q", configPlatform)
 	}
@@ -46,7 +46,7 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 		if len(azs) > 0 {
 			azIndex = int(idx) % len(azs)
 		}
-		provider, err := provider(platform, mpool, osImage, userDataSecret, clusterID, role, &azIndex)
+		provider, err := provider(platform, mpool, osImage, userDataSecret, clusterID, role, &azIndex, hyperVGen)
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to create provider")
 		}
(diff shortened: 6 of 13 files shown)


pkg/types/azure/machinepool.go
```diff
@@ -15,6 +15,12 @@ type MachinePool struct {
 	// +optional
 	InstanceType string `json:"type"`
 
+	// HyperVGeneration is the field that determines the version of HyperVGeneration of the disk.
+	// eg. V1, V2
+	//
+	// +optional
+	HyperVGeneration string `json:"hypervgeneration,omitempty"`
```
patrickdillon, line 22: Do we need to let users specify this? if we default to gen2 when gen2 is available and gen1 for gen1 only instances, the only case this is needed is if users want to use gen1 with a gen2 capable instance. From my understanding of the generations, wanting to downgrade is unlikely. There could be an escape hatch through editing the machine manifests in the unlikely event a downgrade is desired.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -85,13 +86,19 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
...
 			}
-			if !generations.Has("V1") {
-				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "only disks with HyperVGeneration V1 are supported"))
+			if hyperVVersion == "" {
+				if generations.Has("V2") {
+					hyperVGeneration = "V2"
+				} else {
+					hyperVGeneration = "V1"
+				}
+			} else if hyperVVersion != "" && !generations.Has(hyperVVersion) {
```
staebler, line 95:
```suggestion
			} else !generations.Has(hyperVVersion) {
```

pkg/asset/installconfig/azure/validation.go
```diff
@@ -85,13 +86,19 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
...
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, fmt.Sprintf("instance type does not support provided HyperVGeneration %s", hyperVVersion)))
```
staebler, line 96:
```suggestion
				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, fmt.Sprintf("instance type does not support provided HyperVGeneration %q", hyperVVersion)))
```

pkg/asset/installconfig/azure/validation.go
```diff
@@ -54,25 +54,26 @@ func Validate(client API, ic *types.InstallConfig) error {
 }
 
 // ValidateInstanceType ensures the instance type has sufficient Vcpu and Memory.
-func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceType string, diskType string, req resourceRequirements) field.ErrorList {
+func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceType string, diskType string, hyperVVersion string, req resourceRequirements) (string, field.ErrorList) {
```
staebler, line 57: It is a smell to have a validation function returning anything other than whether the inputs passed the validation. Perhaps we should be separate the functionality of gathering from the cloud provider the information for the instance type from the functionality of validating that information against the requirements.

data/data/azure/vnet/main.tf
```diff
@@ -102,3 +102,16 @@ resource "azurerm_image" "cluster" {
     blob_uri = azurerm_storage_blob.rhcos_image.url
   }
 }
+
+resource "azurerm_image" "clusterv2" {
```
staebler, line 106: I am hesitant to add another image to the resource group. What do we think the use cases are for the customer mixing V1 and V2 images? The one that I can think of would be a customer starting with V1 images and wanting to move to an instance type post-installation that only supports V2. That is a scenario that would also come up for a customer upgrading from earlier OpenShift versions where we only create the V1 image. I would expect that we would need documentation on how the user could add the V2 image manually.

data/data/azure/variables-azure.tf
```diff
@@ -158,3 +158,11 @@ This determined whether User defined routing will be used for egress to Internet
 When false, Standard LB will be used for egress to the Internet.
 EOF
 }
+
+variable "azure_hypervgeneration_version" {
+  type = string
+  description = <<EOF
+This determines the HyperVGeneration disk type to use. This will be used to also
```
staebler, line 165:
```suggestion
This determines the HyperVGeneration disk type to use for the control plane VMs. This will be used to also
```

staebler, line 165: "also"? Isn't that the only use. I don't know that we need to include details on how the variable is being used anyway.

data/data/azure/variables-azure.tf
```diff
@@ -158,3 +158,11 @@ This determined whether User defined routing will be used for egress to Internet
...
+determine the image to be used.
+EOF
+}
```
staebler, line 168: Add a newline.

data/data/azure/vnet/main.tf
```diff
@@ -102,3 +102,16 @@ resource "azurerm_image" "cluster" {
...
+  resource_group_name = data.azurerm_resource_group.main.name
+  location            = var.azure_region
+
+  os_disk {
+    os_type  = "Linux"
+    os_state = "Generalized"
+    hyper_v_generation = "V2"
+    blob_uri = azurerm_storage_blob.rhcos_image.url
+  }
+}
```
staebler, line 117: Add a newline.

data/data/azure/vnet/outputs.tf
```diff
@@ -51,7 +51,7 @@ output "resource_group_name" {
 }
 
 output "vm_image" {
-  value = azurerm_image.cluster.id
+  value = var.azure_hypervgeneration_version == "V2" ? "${azurerm_image.cluster.id}_v2" : azurerm_image.cluster.id
```
staebler, line 54:
```suggestion
  value = var.azure_hypervgeneration_version == "V2" ? azurerm_image.clusterv2.id : azurerm_image.cluster.id
```

pkg/asset/installconfig/azure/validation.go
```diff
@@ -135,14 +147,24 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
-		allErrs = append(allErrs, ValidateInstanceType(client, field.NewPath("controlPlane", "platform", "azure"), ic.Azure.Region, instanceType, diskType, controlPlaneReq)...)
+		if hyperVGeneration == "" {
+			hyperVGeneration = defaultHyperVGeneration
+		}
+		version, errs := ValidateInstanceType(client, field.NewPath("controlPlane", "platform", "azure"),
+			ic.Azure.Region, instanceType, diskType, hyperVGeneration, controlPlaneReq)
+		if errs != nil {
+			allErrs = append(allErrs, errs...)
+		} else if version != "" {
+			ic.ControlPlane.Platform.Azure.HyperVGeneration = version
```
staebler, line 158: We should not be setting fields in a validation function.

pkg/tfvars/azure/azure.go
```diff
@@ -88,6 +90,11 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
 		masterDiskEncryptionSetID = masterConfig.OSDisk.ManagedDisk.DiskEncryptionSet.ID
 	}
 
+	var masterConfigHyperVGeneration string
```
staebler, line 93:
```suggestion
	masterConfigHyperVGeneration := "V1"
```

pkg/tfvars/azure/azure.go
```diff
@@ -88,6 +90,11 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
...
+	if strings.HasSuffix(masterConfig.Image.ResourceID, "_v2") {
```
staebler, line 94: I am not keen on parsing the image ID to determine the hyperV generation. It is fragile, prone to breaking if a change is made to how we name the image IDs. This will also break if the user uses marketplace images instead, where the image is not specified via an image ID.

data/data/azure/vnet/main.tf
patrickdillon: > What do we think the use cases are for the customer mixing V1 and V2 images? 

In general the use case is any time users want to create machines with an instance type that only supports one generation that is not the default image we would be creating. For example, currently we create a V1 image, so to create a machineset of V2-only instance types, we would need a second image.

> That is a scenario that would also come up for a customer upgrading from earlier OpenShift versions where we only create the V1 image. I would expect that we would need documentation on how the user could add the V2 image manually.

This is a great point, yes we would need this. Not to put too fine a point on in it, but it's not really tied to upgrades. RHCOS in 4.9 supports this now, so the image could be created manually in a 4.9 cluster without upgrading.

pkg/types/azure/machinepool.go
rna-afk: I'll try to bypass the install config.

pkg/tfvars/azure/azure.go
rna-afk: I tried to bypass the image check. Not sure if the current solution is ideal.

pkg/asset/cluster/tfvars.go
```diff
@@ -321,6 +321,12 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		for i, w := range workers {
 			workerConfigs[i] = w.Spec.Template.Spec.ProviderSpec.Value.Object.(*machinev1beta1.AzureMachineProviderSpec)
 		}
+		client := aztypes.NewClient(session)
+		hyperVGeneration, err := client.GetHyperVGenerationVersion(context.TODO(), masterConfigs[0].VMSize,
+			masterConfigs[0].OSDisk.ManagedDisk.StorageAccountType, masterConfigs[0].Location)
+		if err != nil {
+			return err
+		}
```
patrickdillon, line 329: This is clean, but I don't love making a second API call to retrieve the same information as was retrieved earlier.

I suppose simply checking whether the image in the master configs ends in `_v2` is too hacky.

AWS creates an install config metadata to hold this kind of information. Perhaps we could create a similar metadata for azure.

data/data/azure/vnet/main.tf
patrickdillon: @staebler did this explanation help at all? My only concern here is about usability; specifically will it be clear when users are crafting machinesets day-2. Did you have any further concerns beyond that? I have not checked pricing but I imagine this would be inconsequential.

pkg/asset/machines/azure/machines.go
```diff
@@ -118,6 +120,20 @@ func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string
...
+	if err != nil {
+		return nil, err
+	}
+	client := aztypes.NewClient(ssn)
+	hyperVGeneration, err := client.GetHyperVGenerationVersion(context.TODO(), mpool.InstanceType, mpool.OSDisk.DiskType, platform.Region)
+	if err != nil {
+		return nil, err
+	}
+	if hyperVGeneration == "V2" {
+		image += "_v2"
```
patrickdillon, line 135: I think `_gen2` might be better here.

data/data/azure/vnet/main.tf
patrickdillon, line 106:
```suggestion
resource "azurerm_image" "clustergen2" {
```

data/data/azure/vnet/outputs.tf
```diff
@@ -51,7 +51,7 @@ output "resource_group_name" {
 }
 
 output "vm_image" {
-  value = azurerm_image.cluster.id
+  value = var.azure_hypervgeneration_version == "V2" ? azurerm_image.clusterv2.id : azurerm_image.cluster.id
```
patrickdillon, line 54:
```suggestion
  value = var.azure_hypervgeneration_version == "V2" ? azurerm_image.clustergen2.id : azurerm_image.cluster.id
```

patrickdillon (commented):
This looks good to me. A couple of requests:
1. Consider using install config metadata similar to aws to get around multiple api calls for same info
2. Sorry to be pedantic about naming but I think we should use "Gen2" when there is no further context about what we're talking about; and "V2" is used specifically when answering what is the HyperVGeneration: "It's a gen2 image." "The HyperVGeneration is V2."

pkg/asset/cluster/tfvars.go
staebler: > I suppose simply checking whether the image in the master configs ends in _v2 is too hacky.

If I recall, that is what the code was doing before that I commented was too fragile.

data/data/azure/vnet/main.tf
staebler: OK. The cost looks to be on the order of dollars per month, which is pretty inconsequential. We are not transferring data, so that is not a concern either. I am satisfied.

pkg/asset/cluster/tfvars.go
patrickdillon: > If I recall, that is what the code was doing before that I commented was too fragile.

Ah yes, I see that comment now and completely agree. The only suggestion I have for avoiding a second API call is to stick a variable in the Azure install config metadata:
https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/azure/metadata.go

I am happy with either solution.

data/data/azure/vnet/main.tf
patrickdillon: Thanks for checking the cost.

pkg/asset/machines/master.go
```diff
@@ -362,7 +363,17 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 
 		pool.Platform.Azure = &mpool
 
-		machines, err = azure.Machines(clusterID.InfraID, ic, &pool, string(*rhcosImage), "master", masterUserDataSecretName)
+		ssn, err := icazure.GetSession(installConfig.Azure.CloudName, installConfig.Azure.ARMEndpoint)
+		if err != nil {
+			return err
+		}
```
patrickdillon, line 369: Do we need a new session? It looks like there is already a session in this context

pkg/asset/machines/worker.go
```diff
@@ -405,7 +406,17 @@ func (w *Worker) Generate(dependencies asset.Parents) error {
 			}
 
 			pool.Platform.Azure = &mpool
-			sets, err := azure.MachineSets(clusterID.InfraID, ic, &pool, string(*rhcosImage), "worker", workerUserDataSecretName)
+			ssn, err := icazure.GetSession(installConfig.Azure.CloudName, installConfig.Azure.ARMEndpoint)
+			if err != nil {
+				return err
+			}
```
patrickdillon, line 412: Do we need a new session? It looks like there is already a session in this context

patrickdillon (commented):
I think this is looking close except for a few minor comments. 

Can you test this again after the naming changes? I would be surprised if this weren't broken after the mismatch between `_v2` and `_gen2` between terraform and the machinesets.

PR #5723, opened 2022-03-17
Use powervs NewClient in GetDNSZone

hamzy:
The powervs code uses a new version of the power-go-client which requires
not reusing the Authenticator structure as input.  We are using our own
code instead of ibmcloud's code since ibmcloud may diverge in the future.
Also, their timetables on fixes and testing are different than ours.


mjturek:
Looks good to me!

clnperez:
lgtm as well. @patrickdillon one more for Power VS. Without this one we just can't generate an install-config using the survey questions.

PR #5725, opened 2022-03-18
generate azure storage class manifests

jhixson74:
This creates storage class manifests for disk encryption sets. 

https://issues.redhat.com/browse/CORS-1892


docs/user/azure/encryption.md
```diff
@@ -0,0 +1,115 @@
+# Azure Disk Encryption with Customer Managed Keys
+
+The installer can use an existing disk encryption set. The disk encryption set should be created in a separate resource group than the installer to avoid losing access to your keys.
```
patrickdillon, line 3: "to avoid losing access to your keys after the cluster is destroyed."

docs/user/azure/encryption.md
```diff
@@ -0,0 +1,115 @@
...
+```
+export RESOURCEGROUP="<Resource Group>"
+export LOCATION="<Location>"
+export KEYVAULT_NAME="<Keyvault Name>"
+export KEYVAULT_KEY_NAME="<Keyvault Key Name>"
+export DISK_ENCRYPTION_SET_NAME="<Disk Encryption Set Name>"
+export CLUSTER_SP_ID="<Service Principal ID>"
+```
+
+### Register features
```
patrickdillon, line 15: If I understand this correctly,

"Encryption at host is in preview, and must be enabled to be used."

docs/user/azure/encryption.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Run the installer
+```
+openshift-install create cluster --log-level debug
+```
+
+
+### Update the annotations on the default storage class
+```
+oc patch storageclass managed-premium -p '{"metadata": # {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'
```
patrickdillon, line 104: 1. Is this syntax correct or did that `#` sneak in? 
2. Have we gotten any review/feedback from storage folks about this? I am ignorant of this topic.

docs/user/azure/encryption.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+### Grab the identity of resource group created by the installer
+```
+az identity list --resource-group "<Cluster Resource Group>"
+```
+
+### Grant service principal write permissions to the DiskEncryptionSet
```
patrickdillon, line 112: Maybe it's just me, but I read this differently than I think it is intended.
IIUC, "Grant resource group permissions to the DiskEncryptionSet" may be more accurate.

I read this as "Grant DiskEncryptionSet the permission to write service principals"

pkg/asset/manifests/storageclassconfig.go
```diff
@@ -0,0 +1,135 @@
...
+				})
+				break
+			}
+		}
+	}
+
+	return nil
+}
+
+func validAzureStorageClass(pool *azuretypes.MachinePool, name string) *storagev1.StorageClass {
```
patrickdillon, line 97: Does `valid` have special meaning in this context? Can we just call it `AzureStorageClass`?

pkg/asset/manifests/storageclassconfig.go
```diff
@@ -0,0 +1,135 @@
...
+func (sc *StorageClass) Generate(dependencies asset.Parents) error {
+	installConfig := &installconfig.InstallConfig{}
+	clusterID := &installconfig.ClusterID{}
+	dependencies.Get(installConfig, clusterID)
+
+	// It's possible to have different storage class types for both the
+	// master pool and the worker pool. For the master pool it is simple to
+	// determine the storage class. For the worker pool, this is a bit of a
+	// hacky solution. As soon as a storage class is determined, we stop
+	// and assume that is the storage class. This should probably be added
```
patrickdillon, line 59: >  For the worker pool, this is a bit of a hacky solution. As soon as a storage class is determined, we stop and assume that is the storage class

If I'm understanding why you call this "hacky" it is because there could be multiple worker pools with different encryption sets. Is that correct?

If so, and if it's possible to create multiple storage classes, it doesn't seem too tough to just pass in an index to the storage class name, and create multiple worker storage classes. 

But I'm probably missing a lot of the nuance here.

pkg/asset/manifests/storageclassconfig_test.go
```diff
@@ -0,0 +1,49 @@
+package manifests
```
patrickdillon, line 1: Can we fix the commit message to indicate these are tests--not validation.

pkg/asset/manifests/storageclassconfig_test.go
```diff
@@ -0,0 +1,49 @@
...
+  name: master-managed-premium-encrypted-cmk
+parameters:
+  diskencryptionsetid: /subscriptions/08675309-1111-2222-3333-303606808909/resourceGroups/test-resource-group/providers/Microsoft.Compute/diskEncryptionSets/test-encryption-set
+  kind: Managed
+  resourcegroup: test-resource-group
+  skuname: Premium_LRS
+provisioner: kubernetes.io/azure-disk
+reclaimPolicy: Delete
+volumeBindingMode: WaitForFirstConsumer
+`
```
patrickdillon, line 45: This test concerns me a little because it seems like we would need to change the test to accommodate changes to the code. This is probably manageable if we don't expect the code to change much.

docs/user/azure/encryption.md
```diff
@@ -0,0 +1,115 @@
...
+```
+az role assignment create --role Owner --assignee "<Identity Name>"
```
patrickdillon, line 114: hm is this set of permissions very broad? That would work fine for dev purposes, but users may balk. Perhaps we could add a sentence before this command that says: "The cluster's resource group needs permissions to write to the disk encryption set. You could, for example, make the resource group identity an `Owner.`"

patrickdillon (commented):
This looks really good. I mostly just left several small suggestions.

Does this need any further work/testing?

docs/user/azure/encryption.md
jhixson74: Fixed.

docs/user/azure/encryption.md
jhixson74: That did indeed sneak in. No, I have not spoken to the storage folks about this. However, this is also how the ARO team handles disk encryption sets. I don't see how it could be a problem, but it wouldn't hurt to check.

docs/user/azure/encryption.md
jhixson74: This is why we have reviews ;-) Fixed.

pkg/asset/manifests/storageclassconfig.go
jhixson74: Fixed.

pkg/asset/manifests/storageclassconfig_test.go
jhixson74: Fixed.

docs/user/azure/encryption.md
jhixson74: I'd say it is broad for the identity name. Stricter permissions are probably possible, but after days of trying to get this to work, I stopped when it worked ;-) I'm happy to leave it to the admins to use stricter permissions. I've updated the document with your words.

docs/user/azure/encryption.md
jhixson74: Fixed.

pkg/asset/manifests/storageclassconfig_test.go
jhixson74: I don't see this being a problem.

patrickdillon:
@jsafrane would you or someone from your team mind taking a look at the general approach to make sure we're not missing anything?

pkg/asset/manifests/storageclassconfig.go
```diff
@@ -0,0 +1,135 @@
...
+				})
+				break
+			}
+		}
+	}
+
+	return nil
+}
+
+func azureStorageClass(pool *azuretypes.MachinePool, name string) *storagev1.StorageClass {
```
jsafrane, line 97: We're in process of making the CSI storage class as default in 4.11, hence this should generate a CSI storage class ([like this one](https://github.com/openshift/azure-disk-csi-driver-operator/blob/master/assets/storageclass.yaml) + encryption params).

https://github.com/openshift/azure-disk-csi-driver-operator/pull/47

pkg/asset/manifests/storageclassconfig.go
```diff
@@ -0,0 +1,135 @@
...
+		&installconfig.ClusterID{},
+		// PlatformCredsCheck just checks the creds (and asks, if needed)
+		// We do not actually use it in this asset directly, hence
+		// it is put in the dependencies but not fetched in Generate
+		&installconfig.PlatformCredsCheck{},
+	}
+}
+
+// Generate generates the storage class config file.
+func (sc *StorageClass) Generate(dependencies asset.Parents) error {
```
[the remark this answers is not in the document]
patrickdillon: This storage class requires user input of the encryption keys. @jhixson74 are these keys required to be specified at install time or can they be provided day 2? @jsafrane does the storage operator have the ability to take these keys as input? I assume the user would just generate appropriate manifests.

patrickdillon:
/approve cancel
Let's work out @jsafrane's feedback

pkg/asset/manifests/storageclassconfig.go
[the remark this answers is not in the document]
jhixson74: @patrickdillon , the keys are required prior to install, hence the need to configure the disk encryption set in the install-config.

pkg/asset/manifests/storageclassconfig.go
jhixson74: Updated.

pkg/asset/manifests/storageclassconfig.go
```diff
@@ -0,0 +1,135 @@
...
+				Filename: masterStorageClassConfig,
+				Data:     configData,
+			})
+		}
+
+		workerPool := installConfig.Config.Compute
+		for _, pool := range workerPool {
+			wpool := pool.Platform.Azure
+			if wpool.OSDisk.DiskEncryptionSet != nil {
+				storageClass := azureStorageClass(wpool, "worker")
```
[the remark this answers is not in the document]
jhixson74: Perhaps I've taken the wrong approach here. I was mimicking what is done in ARO. There is a storage class for control plane nodes, and another storage class for worker nodes. Ideally a user would use the worker nodes storage class. Let me know your thoughts on how this could be improved.

pkg/asset/manifests/storageclassconfig.go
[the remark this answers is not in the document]
jhixson74: @jsafrane If you feel this is better placed in the azure-disk-csi-driver-operator, I'm all ears. Would that responsibility fall on somebody else or me? I'm happy to do it, however I don't know where to start ;-)

pkg/asset/manifests/storageclassconfig.go
jhixson74: There can be a disk encryption set for worker nodes, and another for control plane nodes.

pkg/asset/manifests/storageclassconfig.go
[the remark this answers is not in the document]
jsafrane: It depend how the encryption works. Can be the volumes from the "worker" class attached to + decrypted by "master" VMs (and the other way around)? Then we may survive just with one SC with the worker key.

pkg/asset/manifests/storageclassconfig.go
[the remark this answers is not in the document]
jsafrane: Well, it's probably our (storage team) responsibility, but it requires an RFE or epic, which is quite late now.

So, we're adding [a new storage class controller into azure-disk CSI driver operator](https://github.com/openshift/azure-disk-csi-driver-operator/pull/47/) right now  to sort out default storage classes (in-tree -> CSI). It just blindly applies a static manifest now. The controller will be shared by many CSI drivers shortly.

We need:
* Add a new parameter to [`NewCSIStorageClassController`](https://github.com/openshift/library-go/blob/5b7d3d77b85e2aee455c9c3a4538244cd10d60e5/pkg/operator/csi/csistorageclasscontroller/csi_storageclass_controller.go#L46) with a callback function. This is a common pattern, see e.g. [here](https://github.com/openshift/library-go/blob/5b7d3d77b85e2aee455c9c3a4538244cd10d60e5/pkg/operator/csi/csidrivercontrollerservicecontroller/csi_driver_controller_service_controller.go#L70).
* The callback would be called on each [`syncStorageClass()`](https://github.com/openshift/library-go/blob/5b7d3d77b85e2aee455c9c3a4538244cd10d60e5/pkg/operator/csi/csistorageclasscontroller/csi_storageclass_controller.go#L102) to modify the `expectedSC`.
* Azure Disk operator would inject the keys to it in the callback (reading some API object(s) ... `Infrastructure`?).

Another challenge is that the controller supports only one StorageClass and you may need to have two of them. We can run two controllers in the worst case.

We've got generous 3 extra weeks before FF, I think we could squeeze these changes there. I need to talk to the rest of the team, no promises!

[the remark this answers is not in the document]
jsafrane: I talked to the team, @bertinatto will handle the changes in azure-disk-csi-driver-operator.

docs/user/azure/encryption.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Run the installer
+```
+openshift-install create cluster --log-level debug
+```
+
+
+### Update the annotations on the default storage class
+```
+oc patch storageclass managed-premium -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"false"}}}'
```
gnufied, line 109: Do we know which is the default SC now after customers disable storageclass that we create as default?

gnufied: Also in 4.11 - I do not think this is the default storageclass anymore.

docs/user/azure/encryption.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+### Grab the identity of resource group created by the installer
+```
+az identity list --resource-group "<Cluster Resource Group>"
```
gnufied, line 114: Is the `Cluster Resource Group` here different from `$RESOURCEGROUP` above? I assume yes - @patrickdillon to follow up about why we need additional permissions mentioned below.

PR #5731, opened 2022-03-21
Bug 2062998: Update region check for coreos AMIs

patrickdillon:
When determining whether an AMI is available in a region, we should use the coreos stream directly. The installer function for determining known regions has been changed to only list public regions.

### DIFF
--- a/pkg/asset/installconfig/aws/regions.go
+++ b/pkg/asset/installconfig/aws/regions.go
@@ -7,10 +7,10 @@ import (
 	"github.com/openshift/installer/pkg/types"
 )
 
-// knownRegions is a list of AWS regions that the installer recognizes.
-// This is subset of AWS regions and the regions where RHEL CoreOS images are published.
-// The result is a map of region identifier to region description
-func knownRegions(architecture types.Architecture) map[string]string {
+// knownPublicRegions is the subset of public AWS regions where RHEL CoreOS images are published.
+// This subset does not include supported regions which are found in other partitions, such as us-gov-east-1.
+// Returns: a map of region identifier to region description.
+func knownPublicRegions(architecture types.Architecture) map[string]string {
 	required := rhcos.AMIRegions(architecture)
 
 	regions := make(map[string]string)
@@ -22,10 +22,10 @@ func knownRegions(architecture types.Architecture) map[string]string {
 	return regions
 }
 
-// IsKnownRegion return true is a specified region is Known to the installer.
-// A known region is subset of AWS regions and the regions where RHEL CoreOS images are published.
-func IsKnownRegion(region string, architecture types.Architecture) bool {
-	if _, ok := knownRegions(architecture)[region]; ok {
+// IsKnownPublicRegion returns true if a specified region is Known to the installer.
+// A known region is the subset of public AWS regions where RHEL CoreOS images are published.
+func IsKnownPublicRegion(region string, architecture types.Architecture) bool {
+	if _, ok := knownPublicRegions(architecture)[region]; ok {
 		return true
 	}
 	return false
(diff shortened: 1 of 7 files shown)


pkg/asset/rhcos/image.go
```diff
@@ -86,8 +85,10 @@ func osImage(config *types.InstallConfig) (string, error) {
 			return config.Platform.AWS.AMIID, nil
 		}
 		region := config.Platform.AWS.Region
-		if !configaws.IsKnownRegion(config.Platform.AWS.Region, config.ControlPlane.Architecture) {
-			region = "us-east-1"
+		if _, ok := streamArch.Images.Aws.Regions[region]; !ok {
+			globalResourceRegion := "us-east-1"
```
patrickdillon, line 89: open to suggestions here for a better variable name

pkg/asset/installconfig/aws/regions.go
```diff
@@ -23,7 +23,7 @@ func knownRegions(architecture types.Architecture) map[string]string {
 }
 
 // IsKnownRegion return true is a specified region is Known to the installer.
-// A known region is subset of AWS regions and the regions where RHEL CoreOS images are published.
+// A known region is subset of public AWS regions and the regions where RHEL CoreOS images are published.
```
staebler, line 26:
```suggestion
// A known region is a public AWS regions where RHEL CoreOS images are published.
```

pkg/asset/rhcos/image.go
staebler: amiSourceRegion?

staebler: Also, nit that this can be a constant.
```suggestion
			const globalResourceRegion = "us-east-1"
```

pkg/asset/installconfig/aws/regions.go
```diff
@@ -23,7 +23,7 @@ func knownRegions(architecture types.Architecture) map[string]string {
...
 func IsKnownRegion(region string, architecture types.Architecture) bool {
```
staebler, line 27: The names of these functions should be updated to make it clear that it is specifically known public regions.

pkg/asset/rhcos/image.go
```diff
@@ -86,8 +85,10 @@ func osImage(config *types.InstallConfig) (string, error) {
 			return config.Platform.AWS.AMIID, nil
 		}
 		region := config.Platform.AWS.Region
-		if !configaws.IsKnownRegion(config.Platform.AWS.Region, config.ControlPlane.Architecture) {
-			region = "us-east-1"
+		if _, ok := streamArch.Images.Aws.Regions[region]; !ok {
```
staebler, line 88: Can we find a way to re-use the code in the `rhcos.AMIRegions` function here as well?

pkg/asset/installconfig/aws/regions.go
```diff
@@ -7,10 +7,10 @@ import (
 	"github.com/openshift/installer/pkg/types"
 )
 
-// knownRegions is a list of AWS regions that the installer recognizes.
-// This is subset of AWS regions and the regions where RHEL CoreOS images are published.
-// The result is a map of region identifier to region description
-func knownRegions(architecture types.Architecture) map[string]string {
+// knownPublicRegions is a subset of public AWS regions and regions where RHEL CoreOS images are published.
```
staebler, line 10: I still think that you are using the term subset incorrectly here to mean intersection.
Either
```suggestion
// knownPublicRegions is the subset of public AWS regions where RHEL CoreOS images are published.
```
or
```suggestion
// knownPublicRegions is the intersection of public AWS regions and regions where RHEL CoreOS images are published.
```

staebler (commented):
nit on comments but otherwise LGTM

patrickdillon:
I miss a reference to `aws_skip_region_validation` in bootstrap TF. Fixed now.

patrickdillon:
Skipping non-required jobs because none of the failures look related to these changes.

PR #5734, opened 2022-03-22
Added changes for validating the IBM Account Type for Provisioning Power VS resources

bkhadars:
1. Added IBM Account validation changes.
2. Added changes to use the Bluemix object for creating the IBM PISession.

### DIFF
--- a/pkg/asset/installconfig/platformcredscheck.go
+++ b/pkg/asset/installconfig/platformcredscheck.go
@@ -71,6 +71,15 @@ func (a *PlatformCredsCheck) Generate(dependencies asset.Parents) error {
 		if err != nil {
 			return errors.Wrap(err, "creating IBM Cloud session")
 		}
+	case powervs.Name:
+		bxCli, err := powervsconfig.NewBxClient()
+		if err != nil {
+			return err
+		}
+		err = bxCli.NewPISession()
+		if err != nil {
+			return errors.Wrap(err, "creating IBM Cloud session")
+		}
 	case openstack.Name:
 		_, err = openstackconfig.GetSession(ic.Config.Platform.OpenStack.Cloud)
 		if err != nil {
@@ -91,11 +100,6 @@ func (a *PlatformCredsCheck) Generate(dependencies asset.Parents) error {
 		err = con.Test()
 		if err != nil {
 			return errors.Wrap(err, "testing Engine connection")
-		}
-	case powervs.Name:
-		_, err = powervsconfig.GetSession()
-		if err != nil {
-			return errors.Wrap(err, "creating IBM Cloud session")
 		}
 	default:
 		err = fmt.Errorf("unknown platform type %q", platform)
--- a/pkg/asset/installconfig/platformpermscheck.go
+++ b/pkg/asset/installconfig/platformpermscheck.go
@@ -9,6 +9,7 @@ import (
 	"github.com/openshift/installer/pkg/asset"
 	awsconfig "github.com/openshift/installer/pkg/asset/installconfig/aws"
 	gcpconfig "github.com/openshift/installer/pkg/asset/installconfig/gcp"
+	powervsconfig "github.com/openshift/installer/pkg/asset/installconfig/powervs"
 
 	"github.com/openshift/installer/pkg/types/alibabacloud"
 	"github.com/openshift/installer/pkg/types/aws"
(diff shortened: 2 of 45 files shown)


pkg/asset/installconfig/powervs/session.go
```diff
@@ -33,65 +57,162 @@ type PISessionVars struct {
...
+	if accType == "TRIAL" {
+		return fmt.Errorf("account type must be of Pay-As-You-Go/Subscription type for provision Power VS resources")
 	}
-	s, err := ibmpisession.NewIBMPISession(&piOpts)
+	return nil
+}
+
+// GetPISession updates pisession details, return error on fail
+func (c *BxClient) GetPISession() error {
+	var pisv PISessionVars
```
Prashanth684, line 189: instead of this function name being `GetPISession` should it be something like `NewPISession` - because it is not actually getting anything

pkg/asset/installconfig/platformpermscheck.go
```diff
@@ -90,7 +91,14 @@ func (a *PlatformPermsCheck) Generate(dependencies asset.Parents) error {
 	case ibmcloud.Name:
 		// TODO: IBM[#90]: platformpermscheck
 	case powervs.Name:
-		//@TODO add check that the account plan is anything but "lite"
+		bxCli, err := powervsconfig.NewBxClient()
```
Prashanth684, line 94: we should work with the ibm x86 team to see if this can be common for both.

pkg/asset/installconfig/powervs/session.go
bkhadars: Ok will update the function

pkg/asset/installconfig/platformpermscheck.go
clnperez: @bkhadars did you already take a look at the ibmcloud code? Is it possible for us to share theirs? We may have to tweak where we store a credential, etc. But for 4.11 if that's too much maybe we don't have time to tweak that.

clnperez (approved):
If it's easy for us to use the ibmcloud code as-is, let's do that. If not, they're pretty booked so I'm hoping we can merge this as-is.

pkg/asset/installconfig/platformpermscheck.go
bkhadars: @clnperez currenlty ibmcloud is not doing the permission check https://github.com/openshift/installer/blob/950155cddc4b8018b9c6ea4de84695c558b77d24/pkg/asset/installconfig/platformpermscheck.go#L90, I have looked into the ibmcloud code as well and we can't use their code for permissioncheck validation as of now

rna-afk:
Could you move the vendor commits to a separate commit? Would make reading easier.

patrickdillon:
It seems like this PR is still needed. If so, can you please separate the vendoring into a separate commit?

pkg/asset/installconfig/platformcredscheck.go
```diff
@@ -68,6 +68,12 @@ func (a *PlatformCredsCheck) Generate(dependencies asset.Parents) error {
 		}
 	case ibmcloud.Name:
 		_, err = ibmcloudconfig.NewClient()
+	case powervs.Name:
```
jstuever, line 71: This appears to have removed the error handling from the case `ibmcloud.Name`.

bkhadars: Oh.. it's a mistake, I will add it back

bkhadars: updated

clnperez:
@mjturek did you ever figure out what the merge conflict issue was in your PR? looks like Basheer's is facing the same thing in the same test suite

pkg/asset/installconfig/powervs/session.go
```diff
@@ -33,62 +57,162 @@ type PISessionVars struct {
...
-	piOpts := ibmpisession.IBMPIOptions{
-		Authenticator: &core.IamAuthenticator{
-			ApiKey: pisv.APIKey,
-		},
-		Region:      pisv.Region,
-		UserAccount: pisv.ID,
-		Zone:        pisv.Zone,
+	c.APIKey = pisv.APIKey
+
+	os.Setenv("IC_API_KEY", pisv.APIKey)
```
rna-afk, line 134: I think we should refrain from setting environment variables since you cannot rely on them. Do we need this? I see only one place in the code where this is used and it looks like savePISessionVars function should take care of it?

bkhadars: updated the PR with changes

clnperez:
/retest-required

@rna-afk is there something going on with the `e2e-metal-ipi-ovn-ipv6` tests? It seems to fail b/c of a git conflict, which another team member saw as well. But I believe he said that test has changed from required to not, so maybe it will disappear her in a minute from the list.

rna-afk:
Need to take a look but I'd say we can ignore that test for this PR.

rna-afk:
ah I see. The list with the `e2e-metal-ipi-ovn-ipv6` is just the history of all tests that failed in this PR. That's not the list of tests that are currently running, which is the second list of tests (one just above the comments section). We can ignore that test.

clnperez:
@patrickdillon can we get an approve on this one?

PR #5739, opened 2022-03-23
terraform: compress provider binaries faster

staebler:
The amount of time that it takes to compress the terraform binaries is burdensome on development and is causing the brew builds to fail due to timeouts. Changing the compression speed from 9 to 1 increases the resultant size of the installer binary by roughly 10% but reduces the time spent compressing to roughly 20% of the original time.

On my local machine, building and compressing the aws provider is reduced from about 22 seconds to about 4 seconds. The size of the aws provider zip is increased from 99M to 108M. The size of the installer binary is increased from 413M to 450M.

### DIFF
--- a/terraform/Makefile
+++ b/terraform/Makefile
@@ -23,7 +23,7 @@ $(TERRAFORM_PROVIDER_TARGETS): bin/terraform-provider-%.zip: providers/%/go.mod
 	cd providers/$*; \
 	if [ -f main.go ]; then path="."; else path=./vendor/`grep _ tools.go|awk '{ print $$2 }'|sed 's|"||g'`; fi; \
 	go build -ldflags $(LDFLAGS) -o ../../bin/terraform-provider-$* "$$path"; \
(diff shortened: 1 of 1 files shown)


terraform/Makefile
```diff
@@ -23,7 +23,7 @@ $(TERRAFORM_PROVIDER_TARGETS): bin/terraform-provider-%.zip: providers/%/go.mod
 	cd providers/$*; \
 	if [ -f main.go ]; then path="."; else path=./vendor/`grep _ tools.go|awk '{ print $$2 }'|sed 's|"||g'`; fi; \
 	go build -ldflags $(LDFLAGS) -o ../../bin/terraform-provider-$* "$$path"; \
-	zip -9j ../../bin/terraform-provider-$*.zip ../../bin/terraform-provider-$*;
```
patrickdillon, line 26: was dropping the `-j` intentional?

staebler: No, it was not. Thanks.

PR #5743, opened 2022-03-24
Bug 2048451: Fix proxy dial to pick all proxies

rna-afk:
The current proxy dial function does not consider proxy
information provided in some environment variables like
http_proxy etc when checking if the service principal
provided is reachable. Adding a function that checks a few
env variables and dials the service to see if it is reachable.


pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,12 +330,28 @@ func validateEndpointAccessibility(endpointURL string) error {
...
+	if err == nil {
+		conn.Close()
+		return nil
+	}
+	// DefaultClient picks the http and https proxy environment variables automatically.
+	req, err := http.DefaultClient.Get(URL)
+	if err == nil {
+		resp, err := http.DefaultClient.Do(req.Request)
+		if err == nil {
+			if resp.StatusCode == 200 {
```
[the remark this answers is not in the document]
rna-afk: I was hoping to change that to CONNECT and forgot to do that. :(

patrickdillon:
/hold
for discussion

See https://go.dev/doc/go1.16#net/http: 

> The [ProxyFromEnvironment](https://go.dev/pkg/net/http/#ProxyFromEnvironment) function no longer returns the setting of the HTTP_PROXY environment variable for https:// URLs when HTTPS_PROXY is unset.

Also https://github.com/golang/go/issues/40909

It looks to me like the proxy environment could be misconfigured: `HTTPS_PROXY` needs to be set with `https://`

rna-afk:
I asked @yunjiang29 on how he tested it and he used the http_proxy and the https_proxy env variable to test the proxy connection. We still need to fix the existing code that only checks NO_PROXY and ALL_PROXY variables since proxy.Dial function does not check the http proxies.

patrickdillon:
Oh, and I now see that is essentially what you are doing.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,12 +330,28 @@ func validateEndpointAccessibility(endpointURL string) error {
...
 	}
-	conn, err := proxy.Dial(context.Background(), "tcp", net.JoinHostPort(URL.Hostname(), port))
-	if err != nil {
-		return err
+	err = proxyDialURL(context.Background(), net.JoinHostPort(URL.Hostname(), port))
+	return err
+}
+
+func proxyDialURL(ctx context.Context, URL string) error {
+	conn, err := proxy.Dial(context.Background(), "tcp", URL)
```
patrickdillon, line 338: So this satisfies the case if the proxy is setup using `ALL_PROXY`? If that is correct, can you add a comment indicating that is the reason we are using this?

Also this seems like the less common use case, so I would move this to be the second attempt

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,12 +330,28 @@ func validateEndpointAccessibility(endpointURL string) error {
...
 		port = "https"
 	}
-	conn, err := proxy.Dial(context.Background(), "tcp", net.JoinHostPort(URL.Hostname(), port))
-	if err != nil {
-		return err
+	err = proxyDialURL(context.Background(), net.JoinHostPort(URL.Hostname(), port))
+	return err
+}
+
+func proxyDialURL(ctx context.Context, URL string) error {
```
patrickdillon, line 337: Name seems inaccurate because we are trying this proxy or not.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,12 +330,28 @@ func validateEndpointAccessibility(endpointURL string) error {
...
+	conn, err := proxy.Dial(context.Background(), "tcp", URL)
+	if err == nil {
+		conn.Close()
+		return nil
+	}
+	// DefaultClient picks the http and https proxy environment variables.
+	req, err := http.NewRequestWithContext(context.Background(), "CONNECT", URL, nil)
```
patrickdillon, line 344: Why `CONNECT`? Can we replace this with a simple `http.Get`?

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,12 +330,28 @@ func validateEndpointAccessibility(endpointURL string) error {
...
+	if err == nil {
+		resp, err := http.DefaultClient.Do(req)
```
patrickdillon, line 346: Do we care what the response is? Or does any response indicate we can connect so we are good to go?

patrickdillon (commented):
It took me a moment to catch up. Perhaps I'm still catching up, but it seems like we can simplify this. If I'm missing edge cases, let's make them more explicit.

pkg/asset/installconfig/aws/validation.go
rna-afk: Don't we have to check the connection for 404 Not Found?

pkg/asset/installconfig/aws/validation.go
rna-afk: I'm trying to relearn everything about http requests but AFAIK GET will request for resources that may not be available and CONNECT just gets the connection to the server?

staebler: CONNECT is used to tunnel through a proxy. If you are going to use CONNECT, then the URL would need to be the proxy URL and you would need to build out the request to have the connection routed to the service endpoint.

pkg/asset/installconfig/aws/validation.go
patrickdillon: > Don't we have to check the connection for 404 Not Found?

Well, see: https://github.com/openshift/installer/pull/5743#discussion_r834342174

I think we're testing whether the endpoint is accessible, and a 404 response from the endpoint indicates our request got a response from the server. If there is a way to get a known response (e.g. 200) from all the endpoints, perhaps that would be a better option, but that is certainly not what was happening prior to these code changes.

pkg/asset/installconfig/aws/validation.go
rna-afk: I think the http request function expects a URL and automatically picks the proxy URL from the environment variables to connect to the service endpoint and we just need to provide the service endpoint here.

staebler: I don't think that is what is going to happen here. There is no sense in making a CONNECT without subsequent http requests through the tunnel. The http.Request does not support using the tunnel.

patrickdillon: The client--not the request--will pick up the proxy from the environment variables. To reiterate staebler's point CONNECT is used to connect to a proxy, but right now the code is trying to connect to the AWS endpoint as if the endpoint were the proxy.

I haven't heard any objections so I think we can still greatly simplify this down to just `http.Get`

rna-afk: Got it. I changed it to GET now.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,12 +330,23 @@ func validateEndpointAccessibility(endpointURL string) error {
...
+
+func dialURL(ctx context.Context, URL string) error {
+	// DefaultClient picks the http and https proxy environment variables.
+	_, err := http.Get(URL)
+	if err == nil {
+		return nil
+	}
+	// Proxy Dial picks up the ALL_PROXY environment variable and handles sock5 proxy dialing to URL.
+	conn, err := proxy.Dial(context.Background(), "tcp", URL)
+	if err == nil {
```
sadasu, line 345: Probably better to check if the conn != nil to close the connection instead of the value of err. Something like this:
    if conn != nil {
        conn.Close()
    }
    return err

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,12 +330,22 @@ func validateEndpointAccessibility(endpointURL string) error {
...
+}
+
+func dialURL(ctx context.Context, URL string) error {
+	// DefaultClient picks the http and https proxy environment variables.
+	_, err := http.Get(URL)
+	if err == nil {
+		return nil
+	}
+	// Proxy Dial picks up the ALL_PROXY environment variable and handles sock5 proxy dialing to URL.
+	conn, err := proxy.Dial(context.Background(), "tcp", URL)
```
staebler, line 344: I don't think that we need to fall back to this. If we are not handling the connection to an endpoint via http(s) proxies, then the access to the endpoint is going to fail on the local machine when attempting later.

If there was an error returned from `http.Get`, then there was a problem establishing a connection to the endpoint URL. We may get some false positives in cases where some layer-7 reverse proxy handling the endpoint URL accepts the connection and returns a 5xx error. But we should not be getting any false negatives where the installer would report that the endpoint URL is not valid when it actually is (except cases where the endpoint URL is not valid from the installer's environment but is from the cluster's environment).

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,12 +330,22 @@ func validateEndpointAccessibility(endpointURL string) error {
...
-	conn, err := proxy.Dial(context.Background(), "tcp", net.JoinHostPort(URL.Hostname(), port))
-	if err != nil {
-		return err
+	err = dialURL(context.Background(), net.JoinHostPort(URL.Hostname(), port))
+	return err
+}
+
+func dialURL(ctx context.Context, URL string) error {
+	// DefaultClient picks the http and https proxy environment variables.
+	_, err := http.Get(URL)
```
staebler, line 339: I suggest HEAD rather than GET, since we do not care about any content that may be returned.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -329,12 +329,8 @@ func validateEndpointAccessibility(endpointURL string) error {
 	if port == "" {
 		port = "https"
 	}
-	conn, err := proxy.Dial(context.Background(), "tcp", net.JoinHostPort(URL.Hostname(), port))
-	if err != nil {
-		return err
-	}
-	conn.Close()
-	return nil
+	_, err = http.Head(net.JoinHostPort(URL.Hostname(), port))
```
staebler, line 332: I don't think that we need the part where we split `endpointURL` into hostname and port and then recombine them. In fact, we actually want to retain the other parts of the URL, especially the scheme.

sadasu:
@rna-afk could we add unit tests to verify behavior when proxy env variables are set?

rna-afk:
@sadasu I tried a local http server and set the proxy env variable but the code removed it since the request URL was a loopback to the same IP address so if we need to test this, we need at least 2 systems.

pkg/asset/installconfig/aws/validation_test.go
```diff
@@ -580,6 +632,11 @@ func TestValidate(t *testing.T) {
 				publicSubnets:     test.publicSubnets,
 				instanceTypes:     test.instanceTypes,
 			}
+			if test.proxy != "" {
+				os.Setenv("HTTP_PROXY", test.proxy)
+			} else {
+				os.Unsetenv("HTTP_PROXY")
+			}
```
patrickdillon, line 639: As it stands, I don't think `test.proxy` is worth including in these tests. We're not testing whether the proxy is being used or not. So I don't think including the proxy in the test is doing much. 

If we were going to write a test for the proxy, I would imagine it would use something like `[httpTest](https://pkg.go.dev/net/http/httptest#example-Server)` and create a proxy server and make sure the proxy is being called when trying to connect to the endpoint. 

While that could be cool, it seems like a fair amount of effort to test code that is mostly in the stdlib--not code we're writing. So I think it might be best to just remove these parts from the test.

PR #5745, opened 2022-03-24
Check for CVO failing condition before exiting as success

r4f4:


sadasu:
Changes look fine to me.

sadasu:
Could you please specify the behavior we are trying to change here? 
Afaik, if a ClusterOperator reports Available and Progressing, it is still able to handle requests. So, not really a bad/failure condition.
Also, should we be checking for Degraded being False (https://github.com/openshift/library-go/blob/226b07263a02463546bfccce26c69baddf32e613/vendor/github.com/openshift/api/config/v1/types_cluster_operator.go#L177) instead of Progressing being False?

r4f4:
See the lengthy discussion in https://bugzilla.redhat.com/show_bug.cgi?id=1951835 where the conclusion is that CVO will change some of its logic around when it sets its Failing condition but that shouldn't stop us from also checking for that status now so that when CVO is updated we improve the signal back to the admin.
@staebler suggestion was to change https://github.com/openshift/installer/blob/febfc9de5cb8de8fddb0363b79e4dbfcc1b181a9/cmd/openshift-install/create.go#L461 to wait for `Available=true and Failing=false` and perhaps also for `Progressing=false`.

patrickdillon:
/approve

@r4f4 can you improve the commit message? Would be fine to keep it simple and link back to the BZ, for example.

sadasu:
@r4f4 Thanks for providing that context and updating the commit message to point to the discussion around this.
/lgtm

sadasu:

Metal team is looking into these failures. It is not expected to pass at this time.

PR #5746, opened 2022-03-25
WIP - Try defualt placement

omertuc:


eranco74 (commented):
/lgtm

staebler (commented):
> The infrastructure asset is now a dependency of the ingress asset, because the topologies determined within it affect how the new `DefaultPlacement` Ingress config status field is set.

That is undesirable. If we have dependencies within a target, then it makes for a surprising user experience when the user edits both of the manifests. In this case, if the user were to edit both the infra and ingress manifests, then the edits that the user made to the infra manifest would be overwritten by the installer.

staebler:
Other than the asset dependencies, the rest of the changes look fine. I suggest extracting into functions that can be called by both the infra and ingress assets the functionality for determining the topology based on the install config.

omertuc:
Infrastructure no longer a dependency of Ingress

pkg/asset/manifests/ingress.go
```diff
@@ -78,6 +78,24 @@ func (ing *Ingress) Generate(dependencies asset.Parents) error {
 }
 
 func (ing *Ingress) generateClusterConfig(config *types.InstallConfig) ([]byte, error) {
+	controlPlaneTopology, infrastructureTopology := determineTopologies(config)
+
+	isSingleNode := controlPlaneTopology == configv1.SingleReplicaTopologyMode &&
+		infrastructureTopology == configv1.SingleReplicaTopologyMode
```
staebler, line 84: Why is the infrastructure topology a consideration? Shouldn't the logic be that if there is a single control-plane node, then the default placement is on the control plane?

From https://github.com/openshift/enhancements/blob/master/enhancements/single-node/single-node-openshift-with-workers.md#implementation-detailsnotesconstraints,
```
The installer will detect situations in which it's unlikely the user will want to set up a load-balancer. For now, those situations only include installation of single control-plane node cluster deployments on the "none" platform.
```

pkg/asset/manifests/infrastructure_test.go
```diff
@@ -144,6 +154,20 @@ func (b infraBuildNamespace) build(opts ...infraOption) *configv1.Infrastructure
 	return infra
 }
 
+func (b infraBuildNamespace) toFile(opts ...infraOption) *asset.File {
```
staebler, line 157: This does not appear to be used.

pkg/asset/manifests/ingress.go
omertuc: (2) has the advantage of being consistent - a single control plane node cluster by default will always have ingress go through the control plane node, regardless of how many workers it has, and never requires the user to provision a load balancer because the SNO acts like one.

(1) has the advantage of leveraging the user's hardware better by default, but complicates things because the user has to provision a load balancer for ingress (but not one for the API) - which is inconsistent and would probably require good documentation to get the idea across

romfreiman: I vote for consistency here

staebler (commented):
/lgtm
/approve

PR #5747, opened 2022-03-25
Lower severity of ListAWSDefaultServiceQuotas warning msg

r4f4:
This permission is optional and only used by the installer to do
pr-flight quota checks. Logging it missing as a warning tends to
distract SREs and customers from other more serious problems in the log.

### DIFF
--- a/pkg/asset/quota/quota.go
+++ b/pkg/asset/quota/quota.go
@@ -81,7 +81,8 @@ func (a *PlatformQuotaCheck) Generate(dependencies asset.Parents) error {
 		}
 		q, err := quotaaws.Load(context.TODO(), session, ic.AWS.Region, services...)
 		if quotaaws.IsUnauthorized(err) {
-			logrus.Warnf("Missing permissions to fetch Quotas and therefore will skip checking them: %v, make sure you have `servicequotas:ListAWSDefaultServiceQuotas` permission available to the user.", err)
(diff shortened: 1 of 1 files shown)


pkg/asset/quota/quota.go
```diff
@@ -81,7 +81,8 @@ func (a *PlatformQuotaCheck) Generate(dependencies asset.Parents) error {
 		}
 		q, err := quotaaws.Load(context.TODO(), session, ic.AWS.Region, services...)
 		if quotaaws.IsUnauthorized(err) {
-			logrus.Warnf("Missing permissions to fetch Quotas and therefore will skip checking them: %v, make sure you have `servicequotas:ListAWSDefaultServiceQuotas` permission available to the user.", err)
+			logrus.Debugf("Missing permissions to fetch Quotas and therefore will skip checking them: %v, make sure you have `servicequotas:ListAWSDefaultServiceQuotas` permission available to the user.", err)
```
sadasu, line 84: Should also be a Info message? I am not sure if the Debug msgs are always visible (should a debug flag be present ?)

r4f4: I'll let @staebler weigh in on whether `Infof` is preferable over `Debugf`

staebler: I think it is appropriate as is. The user gets an INFO message that the quota checks are skipped. There is additional detail in a DEBUG message, which is always written to .openshift_install.log.

PR #5756, opened 2022-03-28
WIP: api: update to latest version to use AzureUltraSSDCapabilityEnabled

jstuever:
The name of a constant was changed in the OpenShift API related to the
UltraSSD work. This change updates the API and modifies the usage of
that constant to reflect the new name: AzureUltraSSDCapabilityEnabled

https://issues.redhat.com/browse/CORS-1894


pkg/tfvars/azure/azure.go
```diff
@@ -76,7 +76,7 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
 	}
 
 	controlPlaneUltraSSDEnabled := false
-	if masterConfig.UltraSSDCapability == machineapi.AzureUltraSSDCapabilityTrue {
+	if masterConfig.UltraSSDCapability == machineapi.AzureUltraSSDCapabilityEnabled {
 		controlPlaneUltraSSDEnabled = true
 	}
```
patrickdillon, line 81: We can refactor this code block out to where the assignment in the config becomes:

```go
		ControlPlaneUltraSSDEnabled:     masterConfig.UltraSSDCapability == machineapi.AzureUltraSSDCapabilityEnabled,

```

jstuever: That is not possible because the OpenShift API deviates from both Terraform and the Azure API precisely in this regard. Instead of "Enabled" and "Disabled", Terraform uses "True" and "False". As such, it is necessary to translate between the two. I deliberately did so here, above the assignment below, because the in-line syntax was long and not self evident.

jstuever: We could have done that, no doubt. All I'm doing in this PR is updating the API and making our code compatible.

patrickdillon: Ok. I will open a PR to refactor.

PR #5757, opened 2022-03-29
Refactor UltraSSD tfvars

patrickdillon:
Noticed a refactoring opportunity when reviewing #5756.

Depends on #5756


patrickdillon:
#5756 has merged so this is ready to go

PR #5759, opened 2022-03-30
Azure: validate UltraSSD instance types

r4f4:
Confirm an instance type is compatible with UltraSSD enabled or fail
early, before provisioning infrastructure, if type is incompatible.

### DIFF
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -54,7 +54,7 @@ func Validate(client API, ic *types.InstallConfig) error {
 }
 
 // ValidateInstanceType ensures the instance type has sufficient Vcpu and Memory.
-func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceType string, diskType string, req resourceRequirements) field.ErrorList {
+func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceType string, diskType string, req resourceRequirements, ultraSSDEnabled bool) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	typeMeta, err := client.GetVirtualMachineSku(context.TODO(), instanceType, region)
@@ -67,6 +67,7 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 		return append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
 	}
 
+	ultraSSDAvailable := false
 	for _, capability := range *typeMeta.Capabilities {
 
 		if strings.EqualFold(*capability.Name, "vCPUsAvailable") {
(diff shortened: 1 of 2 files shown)


pkg/asset/installconfig/azure/validation.go
```diff
@@ -102,9 +103,18 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
...
+		} else if strings.EqualFold(*capability.Name, "UltraSSDAvailable") {
+			ultraSSDAvailable = strings.EqualFold(*capability.Value, "True")
 		}
 	}
 
+	// The UltraSSDAvailable capability might not be present at all, in which
+	// case it must assumed to be false
+	if ultraSSDEnabled && !ultraSSDAvailable {
+		errMsg := fmt.Sprintf("UltraSSD capability not supported for instance type %s", instanceType)
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
```
rna-afk, line 115: Since we are already providing the instanceType value in the Invalid function, let's just mention it once?
```suggestion
		allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "UltraSSD capability not supported for provided instance type"))
```

patrickdillon (commented):
This looks good, but it needs to be extended to the other machine pools (compute and default)

pkg/asset/installconfig/azure/validation.go
```diff
@@ -122,9 +130,15 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
 		defaultInstanceType = ic.Platform.Azure.DefaultMachinePlatform.InstanceType
 	}
 
+	defaultUltraSSDCapability := "Disabled"
+	if ic.Platform.Azure.DefaultMachinePlatform != nil && ic.Platform.Azure.DefaultMachinePlatform.UltraSSDCapability != "" {
+		defaultUltraSSDCapability = ic.Platform.Azure.DefaultMachinePlatform.UltraSSDCapability
```
patrickdillon, line 135: The `DefaultMachinePlatform` can also set the instance type.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -102,9 +103,16 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
...
 				allErrs = append(allErrs, field.Invalid(fieldPath.Child("osDisk", "diskType"), diskType, errMsg))
 			}
+		} else if strings.EqualFold(*capability.Name, "UltraSSDAvailable") {
+			ultraSSDAvailable = strings.EqualFold(*capability.Value, "True")
 		}
 	}
 
+	// The UltraSSDAvailable capability might not be present at all, in which case it must assumed to be false
+	if ultraSSDEnabled && !ultraSSDAvailable {
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "UltraSSD capability not supported for provided instance type"))
```
patrickdillon, line 113:
```suggestion
                errMsg: fmt.Sprintf("UltraSSD capability not supported for this instance type in the %s region", region)
		allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
```

pkg/asset/installconfig/azure/validation.go
```diff
@@ -123,17 +123,18 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
 
 	defaultDiskType := aztypes.DefaultDiskType
 	defaultInstanceType := ""
```
patrickdillon, line 125: naming nit: can we change this to something like `defaultMachinePlatformInstanceType` or `defaultMPInstanceType`? I think we should distinguish that this is a default set by the user rather than the default chosen by the installer in the `defaults` package.

PR #5761, opened 2022-03-30
Bug 2069259: Set ironic httpd port to 6180

elfosardo:
We should not use default port 80 as it may cause conflicts and
it's potentially unsecure in production environments.
Also we may want to comply with the docs at
https://docs.openshift.com/container-platform/4.10/installing/installing_bare_metal_ipi/ipi-install-prerequisites.html#network-requirements-out-of-band_ipi-install-prerequisites

### DIFF
--- a/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
+++ b/data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
@@ -6,6 +6,7 @@ set -ex
 # Must match ironic.service
 EXIT_CODE_NO_RESTART=42
 
+DEFAULT_HTTP_PORT="6180"
 IRONIC_IMAGE=$(image_for ironic)
 IRONIC_AGENT_IMAGE=$(image_for ironic-agent)
 CUSTOMIZATION_IMAGE=$(image_for image-customization-controller)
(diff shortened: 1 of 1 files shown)


data/data/bootstrap/baremetal/files/usr/local/bin/startironic.sh.template
```diff
@@ -114,13 +116,14 @@ podman run -d --name coreos-downloader \
...
-     -v $IRONIC_SHARED_VOLUME:/shared:z --entrypoint /bin/runhttpd ${IRONIC_IMAGE}
+    --restart on-failure \
+    --env IRONIC_RAMDISK_SSH_KEY="$IRONIC_RAMDISK_SSH_KEY" \
+    --env PROVISIONING_INTERFACE=$PROVISIONING_NIC \
+    --env HTTP_PORT=$DEFAULT_HTTP_PORT \
+    -v $IRONIC_SHARED_VOLUME:/shared:z --entrypoint /bin/runhttpd ${IRONIC_IMAGE}
 
 # Add firewall rules to ensure the IPA ramdisk can reach httpd, Ironic and the Inspector API on the host
-for port in 80 5050 6385 ; do
+for port in 6180 5050 6385 ; do
```
zaneb, line 126: I'm no longer convinced this is the right thing to do. If it has always been on port 80, as seems to be the case, then it needs to stay there as there are likely customers with external firewalls who haven't opened port 6180 to the bootstrap, and moving the port would break them.

elfosardo: we could leave port 80 open temporarily but we should correct this anyway
we can maybe fix the docs for 4.9 and 4.10, but then leave the change for 4.11 ?
the docs say that port 6180 should be the only one needed

elfosardo:
we'll leave this as it is for the time being

PR #5765, opened 2022-03-31
Bug 2070666: Update the default channel for new 4.6 installs to eus-4.6

sdodson:
OpenShift 4.6 EUS is unique in that it doesn't pick up the lifecycle changes
which made EUS available to all subscribers, nor do we continue adding all
4.6.z to stable-4.6 channel. Therefore, so that new installs don't get abandoned
we'll set the default channel to eus-4.6.


LalatenduMohanty (approved):
/lgtm

jiajliu:
@sdodson Do we also need to update v4.6 doc for this update?

sdodson:
/override ci/prow/e2e-azure-upi
This hasn't passed since January.

PR #5769, opened 2022-04-04
Bug 2055601: Delete *.apps record by name

rna-afk:
The *.apps records in DNS do not have the cluster tag that was
used to delete all records. Adding a name check
to destroy the *.apps record along with the tag check.


patrickdillon:
Can we check whether ingress operator can add the tag when creating the record?

patrickdillon:
/hold 
in favor of https://github.com/openshift/cluster-ingress-operator/pull/737

PR #5778, opened 2022-04-06
Bug 2033482: vsphere: ensure terraform variables are defined

jstuever:
This change add missing variables to each vsphere terraform stage in order
to remove the terraform warnings regarding undeclared variables. These
warnings were the result of output variables from some stages not being
declared in the later stages.


patrickdillon:
CI logs look good. 

/approve

PR #5615, opened 2022-02-04
Add IBM Power VS: tfvars

clnperez:
For more background on IPI on Power VS, refer to the enhancement
proposal here: openshift/enhancements#736

Older discussions on some of the code here can be found in #5224

Signed-off-by: Christy Norman <christy@linux.vnet.ibm.com>

PR #5780, opened 2022-04-06
Power VS: Create Remaining TF Resources

clnperez:
Add the creation of the VPC in IBM Cloud to connect in to the load
balancers, the DHCP Network in Power VS, and the Cloud Connection (a GRE
tunnel wrapper) to connect the two.

This PR also adds `time` as a Terraform provider because of the way the
IP addresses are retrieved for the server instances. With traditional
Power VS networks, Power VS assigns a static IP. With v1 of the DHCP
service created for IPI, the DHCP server on the private network assigns
the IP. As a result, there is a small amount of time needed to wait between
the instance creation and the IP retrieval from the list of leases.

It also reorganizes some of the resource creation due to this difference in
behavior from the network we did early development with.


clnperez:
depends on #5614 and #5615

Prashanth684:
splitting the vendoring into a separate commit would help

clnperez:
i'd like to get a 2nd opinion on that from @rna-afk . it's vendoring but its also a new tf provider's vendoring -- as opposed to the normal vendoring for the installer binary

clnperez:
or maybe even from @staebler since he's done most of the tf changes wrt all that

rna-afk:
I think it's best to move them to another commit. Would make it easier to read the tf files since terraform changes are just vendor changes

staebler:
> or maybe even from @staebler since he's done most of the tf changes wrt all that

I am 100% in support of splitting this PR into multiple commits, with the vendor being one contained commit. At a glance, this is the minimum that I would do.
1. Commit for adding the new terraform provider.
2. Commit for vendoring the dependencies of the new provider.
3. Commit for the tf files.
4. Commit for the rest.

clnperez:
just what i was looking for @staebler. thanks and will do

clnperez:
got those split up and will fix the merge conflicts. heading out the door for a bit but wanted to get the first half done

clnperez:
@rna-afk @patrickdillon i don't know why verify-vendor always seems to fail for me. any suggestions? i'm still using go 1.17, and no matter whether i run go mod tidy/vendor using the container or not, nothing changes.

clnperez:
ah. rebasing and re-pushing worked even though it didn't change anything i could see 🤷🏻‍♀️

clnperez:
actually that didn't work either. i was looking at `verify-codegen`, not `verify-vendor`. there was only an issue with one repo, so i just manually moved the dependency version in go.mod: 

`github.com/vmware/govmomi v0.24.0` => `github.com/vmware/govmomi v0.27.4`

clnperez:
PR was rebased. PTAL @patrickdillon @rna-afk

data/data/powervs/cluster/bootstrap/lb/main.tf
```diff
@@ -0,0 +1,53 @@
+# Using explicit depends_on as otherwise there are issues with updating and adding of pool members
```
patrickdillon, line 1: The bootstrap node, here, is part of the cluster stage. Looking at https://github.com/openshift/installer/blob/master/pkg/terraform/stages/powervs/stages.go#L10-L16 I don't see any bootstrap destroy functionality.

Is this intentional? Typically the bootstrap node and all resources that should be destroyed along with it will go into their own stage. Nothing else should go into that stage as it would be destroyed.

clnperez: oh actually, now that i'm not in a meeting and multitasking -- yes, this specific bit was intentional, as in, i didn't add a bootstrap phase and add these resources (vpc, network, cloud connection) all in the same PR. so i did put these into their own file now to go ahead and organize them in a way that he didn't have to do a major rebase.

clnperez:
@patrickdillon given my response about the bootstrap phase and the followon PR, are we okay merging this as-is?

patrickdillon (commented):
I am fine with merging this with the understanding that a later PR will add a separate bootstrap phase. 

Can you remove the file .swp file: data/data/powervs/cluster/master/lb/.variables.tf.swp

Then I will approve, lgtm

clnperez:
@patrickdillon a nutanix commit caused a merge conflict. fixed but needs a new approve, etc.

PR #5781, opened 2022-04-06
Update installconfig validation for Nutanix

thunderboltsid:
Expect credentialsMode for Nutanix to be Manual.


staebler (commented):
Is this change necessary? Are we planning to have CredentialsRequests for Nutanix? If the answers to the previous are yes, then we also need changes to the description of the `credentialsMode` field.
https://github.com/openshift/installer/blob/73522dd14c8f0c8b905056624dc96ef3a2e632df/pkg/types/installconfig.go#L159

thunderboltsid:
> 

We do have CredentialsRequest for Nutanix. Good catch on the description. Will fix that.

thunderboltsid:
> Is this change necessary? Are we planning to have CredentialsRequests for Nutanix? If the answers to the previous are yes, then we also need changes to the description of the `credentialsMode` field.
> 
> https://github.com/openshift/installer/blob/73522dd14c8f0c8b905056624dc96ef3a2e632df/pkg/types/installconfig.go#L159

We do have CredentialsRequest for Nutanix. Good catch on the description. Will fix that.

PR #5783, opened 2022-04-07
pkg/asset/manifests/operators: Set kubernetes.io/description for cluster-config-v1

wking:
The ConfigMap is install-time archival, and may no longer express the current cluster state.  Having it in-cluster was intended to be a temporary hack until we got Infrastructure and such, and we just failed to remove the ConfigMap quickly enough, and now removing it would be a breaking change.  But with this commit, we get a warning about this into the cluster, so folks looking at the resource are more likely to understand its limitations.

The kubernetes.io/description annotation has been documented since kubernetes/website#31621.

### DIFF
--- a/pkg/asset/manifests/utils.go
+++ b/pkg/asset/manifests/utils.go
@@ -15,8 +15,9 @@ type configurationObject struct {
 }
 
 type metadata struct {
-	Name      string `json:"name,omitempty"`
-	Namespace string `json:"namespace,omitempty"`
+	Annotations map[string]string `json:"annotations,omitempty"`
+	Name        string            `json:"name,omitempty"`
+	Namespace   string            `json:"namespace,omitempty"`
 }
 
 func configMap(namespace, name string, data genericData) *configurationObject {
--- a/pkg/asset/manifests/operators.go
+++ b/pkg/asset/manifests/operators.go
@@ -92,6 +92,11 @@ func (m *Manifests) Generate(dependencies asset.Parents) error {
 	m.KubeSysConfig = configMap("kube-system", "cluster-config-v1", genericData{
 		"install-config": string(redactedConfig),
 	})
+	if m.KubeSysConfig.Metadata.Annotations == nil {
+		m.KubeSysConfig.Metadata.Annotations = make(map[string]string, 1)
+	}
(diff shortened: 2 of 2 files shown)


pkg/asset/manifests/utils.go
```diff
@@ -15,8 +15,10 @@ type configurationObject struct {
 }
 
 type metadata struct {
-	Name      string `json:"name,omitempty"`
-	Namespace string `json:"namespace,omitempty"`
+	Annotations map[string]string ``json:"annotations,omitempty"`
```
staebler, line 18:
```suggestion
	Annotations map[string]string `json:"annotations,omitempty"`
```

wking: Oops; thanks.  Fixed with c4d2ceeda3 -> 22837b812f.

wking:
Huh, checking a CI run to confirm this was working, we somehow have one of these in `openshift-etcd` too?

```console
$ curl -s https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_installer/5783/pull-ci-openshift-installer-master-e2e-gcp/1512131968819531776/artifacts/e2e-gcp/gather-extra/artifacts/configmaps.json | jq -r '.items[].metadata | select(.name == "cluster-config-v1")'
{
  "annotations": {
    "kubernetes.io/description": "The install-config content used to create the cluster.  The cluster configuration may have evolved since installation, so check cluster configuration resources directly if you are interested in the current cluster state."
  },
  "creationTimestamp": "2022-04-07T19:27:25Z",
  "name": "cluster-config-v1",
  "namespace": "kube-system",
  "resourceVersion": "608",
  "uid": "2d20ce12-d57b-465e-aafa-0d9d33c19cd1"
}
{
  "annotations": {
    "kubernetes.io/description": "The install-config content used to create the cluster.  The cluster configuration may have evolved since installation, so check cluster configuration resources directly if you are interested in the current cluster state."
  },
  "creationTimestamp": "2022-04-07T19:33:02Z",
  "name": "cluster-config-v1",
  "namespace": "openshift-etcd",
  "resourceVersion": "5272",
  "uid": "0408a811-32f5-4652-b8fc-5bcd771b00a5"
}
```

wking:
Seems like that is from openshift/cluster-etcd-operator#708.  I'll follow up with the etcd folks to see what that was about.

PR #5784, opened 2022-04-07
Remove defaultStorageContainer from nutanix platform installconfig

thunderboltsid:
Remove defaultStorageContainer from the platform, installconfig, and survey.
This was added earlier with the intention to provide a default storage class
for the cluster. However, with the Nutanix CSI driver provisioning being a day-2
operation for Nutanix, a defaultStorageContainer is no longer required.


staebler (commented):
/lgtm
/approve

PR #5787, opened 2022-04-08
baremetal: static IP for bootstrap node

honza:
Add new configuration fields to `install-config.yaml` that will enable the cluster operator to set up static networking for the bootstrap node

This is the implementation of the [Bootstrap external IP OpenShift enhancement](https://github.com/openshift/enhancements/blob/master/enhancements/baremetal/bootstrap-external-ip.md).

### DIFF
--- a/data/data/bootstrap/baremetal/files/etc/NetworkManager/system-connections/nmconnection.template
+++ b/data/data/bootstrap/baremetal/files/etc/NetworkManager/system-connections/nmconnection.template
@@ -1,0 +1,15 @@
+{{if .PlatformData.BareMetal.ExternalStaticIP}}
+[connection]
+type=ethernet
+[ethernet]
+mac-address={{.PlatformData.BareMetal.ExternalMACAddress}}
+{{ if .UseIPv6ForNodeIP }}
+[ipv6]
+{{ else }}
+[ipv4]
+{{ end }}
+method=manual
+addresses={{.PlatformData.BareMetal.ExternalStaticIP}}/{{.PlatformData.BareMetal.ExternalSubnetCIDR}}
+gateway={{.PlatformData.BareMetal.ExternalStaticGateway}}
+dns={{.PlatformData.BareMetal.ExternalStaticGateway}}
+{{end}}
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -1871,6 +1871,18 @@ spec:
                     description: APIVIP is the VIP to use for internal API communication
                     format: ip
                     type: string
+                  bootstrapExternalStaticGateway:
+                    description: BootstrapExternalStaticGateway is the static network
+                      gateway of the bootstrap node. This can be useful in environments
+                      without a DHCP server.
+                    format: ip
+                    type: string
+                  bootstrapExternalStaticIP:
+                    description: BootstrapExternalStaticIP is the static IP address
+                      of the bootstrap node. This can be useful in environments without
+                      a DHCP server.
+                    format: ip
+                    type: string
                   bootstrapOSImage:
                     description: BootstrapOSImage is a URL to override the default
                       OS image for the bootstrap node. The URL must contain a sha256
--- a/pkg/asset/ignition/bootstrap/baremetal/template.go
+++ b/pkg/asset/ignition/bootstrap/baremetal/template.go
@@ -58,6 +58,16 @@ type TemplateData struct {
 
 	// ProvisioningNetwork displays the type of provisioning network being used
 	ProvisioningNetwork string
+
+	// ExternalStaticIP is the static IP of the bootstrap node
+	ExternalStaticIP string
+
+	// ExternalStaticIP is the static gateway of the bootstrap node
+	ExternalStaticGateway string
+
+	ExternalSubnetCIDR int
+
+	ExternalMACAddress string
 }
 
 // GetTemplateData returns platform-specific data for bootstrap templates.
@@ -70,6 +80,17 @@ func GetTemplateData(config *baremetal.Platform, networks []types.MachineNetwork
 	templateData.ProvisioningNetwork = string(config.ProvisioningNetwork)
 	templateData.BaremetalEndpointOverride = fmt.Sprintf("http://%s/v1", net.JoinHostPort(config.APIVIP, "6385"))
 	templateData.BaremetalIntrospectionEndpointOverride = fmt.Sprintf("http://%s/v1", net.JoinHostPort(config.APIVIP, "5050"))
+	templateData.ExternalStaticIP = config.BootstrapExternalStaticIP
+	templateData.ExternalStaticGateway = config.BootstrapExternalStaticGateway
+	templateData.ExternalMACAddress = config.ExternalMACAddress
+
+	if config.BootstrapExternalStaticIP != "" {
+		for _, network := range networks {
+			cidr, _ := network.CIDR.Mask.Size()
+			templateData.ExternalSubnetCIDR = cidr
+			break
+		}
+	}
 
 	if config.ProvisioningNetwork != baremetal.DisabledProvisioningNetwork {
 		cidr, _ := config.ProvisioningNetworkCIDR.Mask.Size()
--- a/pkg/asset/installconfig/baremetal/validation.go
+++ b/pkg/asset/installconfig/baremetal/validation.go
@@ -20,3 +20,20 @@ func ValidateProvisioning(ic *types.InstallConfig) error {
 
 	return allErrs.ToAggregate()
 }
+
+// ValidateStaticBootstrapNetworking ensures that both or neither of  BootstrapExternalStaticIP and BootstrapExternalStaticGateway are set
+func ValidateStaticBootstrapNetworking(ic *types.InstallConfig) error {
+	if ic.Platform.BareMetal == nil {
+		return errors.New(field.Required(field.NewPath("platform", "baremetal"), "Baremetal validation requires a baremetal platform configuration").Error())
+	}
+
+	if ic.Platform.BareMetal.BootstrapExternalStaticIP != "" && ic.Platform.BareMetal.BootstrapExternalStaticGateway == "" {
+		return errors.New(field.Required(field.NewPath("platform", "baremetal"), "You must specify a value for BootstrapExternalStaticGateway when BootstrapExternalStaticIP is set.").Error())
+	}
+
+	if ic.Platform.BareMetal.BootstrapExternalStaticGateway != "" && ic.Platform.BareMetal.BootstrapExternalStaticIP == "" {
+		return errors.New(field.Required(field.NewPath("platform", "baremetal"), "You must specify a value for BootstrapExternalStaticIP when BootstrapExternalStaticGateway is set.").Error())
+	}
+
+	return nil
+}
--- a/pkg/asset/installconfig/baremetal/validation.go
+++ b/pkg/asset/installconfig/baremetal/validation.go
@@ -8,25 +8,23 @@ import (
 	"github.com/openshift/installer/pkg/types/baremetal/validation"
 )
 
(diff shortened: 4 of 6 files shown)


data/data/bootstrap/baremetal/files/etc/NetworkManager/conf.d/99-baremetal.conf
```diff
@@ -1,3 +1,14 @@
...
 ipv6.dhcp-duid=ll
 ipv6.dhcp-iaid=mac
+
+{{if .PlatformData.BareMetal.ExternalStaticIP}}
+type=ethernet
+interface-name=ens3
+[ethernet]
+[ipv4]
+method=manual
+addresses={{.PlatformData.BareMetal.ExternalStaticIP}}/24
```
hardys, line 11: The subnet here should probably be derived from the `machineNetwork` CIDR?

data/data/bootstrap/baremetal/files/etc/NetworkManager/conf.d/99-baremetal.conf
```diff
@@ -1,3 +1,14 @@
 [connection]
 ipv6.dhcp-duid=ll
 ipv6.dhcp-iaid=mac
+
+{{if .PlatformData.BareMetal.ExternalStaticIP}}
+type=ethernet
+interface-name=ens3
+[ethernet]
+[ipv4]
```
hardys, line 9: You can probably determine v4/v6 using the existing `UseIPv6ForNodeIP` variable added via #4756

data/data/bootstrap/baremetal/files/etc/NetworkManager/conf.d/99-baremetal.conf
```diff
@@ -1,3 +1,14 @@
 [connection]
 ipv6.dhcp-duid=ll
 ipv6.dhcp-iaid=mac
+
+{{if .PlatformData.BareMetal.ExternalStaticIP}}
```
hardys, line 5: I think this needs to go into a new file e.g `/etc/NetworkManager/system-connections/ens3.nmconnection`, not the global `99-baremetal-conf`?

honza: OK. Can we hardcode `ens3`?

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -1704,6 +1704,12 @@ spec:
                     description: APIVIP is the VIP to use for internal API communication
                     format: ip
                     type: string
+                  bootstrapExternalStaticGateway:
```
zhouhao3, line 1707: A description of the field should be added.

zhouhao3:
LGTM
Tested this PR locally by way of IPI and it works fine.

sadasu:
@honza would be great if you could add when the 2 new configs added to the install config should be set. Maybe here [https://github.com/openshift/installer/pull/5787#issue-1197294475] is a good place.
Also, as @zhouhao3 pointed out, the changes to install.openshift.io_installconfigs.yaml should include a description of the 2 new fields being added.

honza:
@sadasu The link in your comment is broken: where would you like me to add these comments?

sadasu:
> @sadasu The link in your comment is broken: where would you like me to add these comments?

https://github.com/openshift/installer/pull/5787#issue-1197294475

data/data/bootstrap/baremetal/files/etc/NetworkManager/conf.d/99-baremetal.conf
hardys: That's a good point - no I don't think we can, specifically the name differs for ARM, so we'll need to use the MAC address provided via https://github.com/openshift/installer/pull/5698 (if specified) then default to the name when no MAC is specified 

@jeffdyoung would you agree?

data/data/bootstrap/baremetal/files/etc/NetworkManager/system-connections/ens3.nmconnection.template
```diff
@@ -0,0 +1,15 @@
+{{if .PlatformData.BareMetal.ExternalStaticIP}}
+[connection]
+type=ethernet
+interface-name=ens3
```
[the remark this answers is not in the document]
jeffdyoung: @zaneb can confirm. ARM and x86 libvirt vms will end up with different nic names because they use different pci/pcie busses by default.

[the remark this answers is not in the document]
hardys: Since #5698 switched to referring to the nics by MAC (either user specified or generated) I think we can just modify the NM keyfile to refer to the mac instead of the device name?

pkg/asset/ignition/bootstrap/baremetal/template.go
```diff
@@ -70,6 +78,16 @@ func GetTemplateData(config *baremetal.Platform, networks []types.MachineNetwork
 	templateData.ProvisioningNetwork = string(config.ProvisioningNetwork)
 	templateData.BaremetalEndpointOverride = fmt.Sprintf("http://%s/v1", net.JoinHostPort(config.APIVIP, "6385"))
 	templateData.BaremetalIntrospectionEndpointOverride = fmt.Sprintf("http://%s/v1", net.JoinHostPort(config.APIVIP, "5050"))
+	templateData.ExternalStaticIP = config.BootstrapExternalStaticIP
+	templateData.ExternalStaticGateway = config.BootstrapExternalStaticGateway
+
+	if config.BootstrapExternalStaticIP != "" {
+		for _, network := range networks {
+			cidr, _ := network.CIDR.Mask.Size()
```
sadasu, line 86: Should we be checking that we pick the CIDR from the network that contains the BootstrapExternalStaticIP? In this implementation, we would be picking the 1st network's CIDR. Is that what we want?

pkg/asset/installconfig/baremetal/validation.go
```diff
@@ -20,3 +20,19 @@ func ValidateProvisioning(ic *types.InstallConfig) error {
 
 	return allErrs.ToAggregate()
 }
+
+func ValidateStaticBootstrapNetworking(ic *types.InstallConfig) error {
+	if ic.Platform.BareMetal == nil {
```
sadasu, line 25: With this change, a check for ic.Platform.BareMetal == nil will take place twice from  https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/platformprovisioncheck.go#L75.
We could separate the platform check into its own validation and call it from  https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/platformprovisioncheck.go#L75?

data/data/bootstrap/baremetal/files/etc/NetworkManager/system-connections/nmconnection.template
```diff
@@ -0,0 +1,15 @@
+{{if .PlatformData.BareMetal.ExternalStaticIP}}
+[connection]
+type=ethernet
+interface-name={{.PlatformData.BareMetal.ProvisioningInterfaceMAC}}
```
zaneb, line 4: I don't think this is what we're looking for. For a start, we're trying to configure the External interface, and this variable is for the Provisioning interface,
But also I don't know that you can just stick a MAC address as an interface name in NetworkManager? Nothing in https://networkmanager.dev/docs/api/latest/settings-connection.html suggests that to me.

In #5698 passing the MAC only worked because there is code that runs in the script that looks up the interface name for that MAC using NetworkManager https://github.com/openshift/installer/pull/5698/files#diff-6f24f49f3938f9f97f5429b1280c21ccb462b377705ad4f9a7f526316caa2784R31. What we have here is a much harder problem because we need networking to be set up in the initramfs, before we even get to run any code.

It looks from https://networkmanager.dev/docs/api/latest/NetworkManager.conf.html#connection-sections that there _might_ be a way to set in the NetworkManager config which device a connection gets mapped to when interface-name doesn't appear It seems like either:

```
[connection-external]
match-device=mac:<external mac>
```

or

```
[device-external]
match-device=mac:<external mac>
allowed-connections=uuid:<connection UUID>

[device-provisioning]
match-device=mac:<provisioning mac>
allowed-connections=except:uuid:<connection UUID>
```

might possibly be the magic incantation.

pkg/asset/ignition/bootstrap/baremetal/template.go
hardys: In a dual stack environment I think you'll always want the external IP to be the first (v4) network, although technically in the DHCP dual stack scenario I guess the NIC could get an IP on both networks

If we ever want IPI to support a case like https://issues.redhat.com/browse/METAL-163 where the BMCs are only accessible via v6, we'd need to revise this for the case where the provisioning network is disabled so the user can specify two IPs.

Considering that a follow-up seems reasonable to me, although we'd end up having to make the install-config interface a list like the networks if that becomes a requirement

data/data/bootstrap/baremetal/files/etc/NetworkManager/system-connections/nmconnection.template
hardys: > What we have here is a much harder problem because we need networking to be set up in the initramfs, before we even get to run any code.

@zaneb I don't think that's the case on the bootstrap VM, unlike the other hosts we're not pulling ignition in the ramdisk via the network, because it's injected via the fw-cfg qemu interface instead.

I agree we need to use the `ExternalMACAddress` though, but I think we can just use `mac-address` in the `[ethernet]` section to specify the device, e.g similar to [these docs](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/assembly_manually-creating-networkmanager-profiles-in-key-file-format_configuring-and-managing-networking) (not tested this myself though)

honza: > I agree we need to use the ExternalMACAddress though, but I think we can just use mac-address in the [ethernet] section to specify the device, e.g similar to [these docs](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/assembly_manually-creating-networkmanager-profiles-in-key-file-format_configuring-and-managing-networking) (not tested this myself though)

Just added these changes

zaneb: Ah yep, that looks correct, here is the doc: https://networkmanager.dev/docs/api/latest/settings-802-3-ethernet.html

> If specified, this connection will only apply to the Ethernet device  whose permanent MAC address matches. This property does not change the  MAC address of the device (i.e. MAC spoofing).

That's exactly what we want.

sadasu:
/approve
For reference, the ExternalMACAddress is generated and code was added via https://github.com/openshift/installer/pull/5698

elfosardo:
we need https://github.com/openshift/installer/pull/5909 for this to pass

honza:
#5909 has merged
/retest

honza:
Openstack build failures in e2e-metal jobs due to timeouts; openstack was previously built successfully

/test e2e-metal-ipi

iurygregory:
will this be converted to a BZ so we can get in 4.11 or no?

sadasu:
@honza you will need a BZ if this needs to merge after FF. 
Nit: https://github.com/openshift/installer/pull/5787/files#diff-98889c853b423cbe4c84633ca8badb2879e2288d3ba0a76ebb2ed5102c0a59baR68-R70 . These 2 new fields could use a description.

awolffredhat:
Tested successfully by deploying a cluster on VMs with no dhcp server on the baremetal network.

bparees:
applying valid-bz label so this can merge as it has gotten all necessary pre-approvals/had pre-merge testing completed.

zaneb:
The e2e-metal-ipi-ovn-ipv6 appears to be permafailing (not just here). Logs show that the cluster is being bootstrapped successfully, so the failures are not related to this patch.

PR #5788, opened 2022-04-10
Bug 2076880: for vsphere ipi add cluster domain to the uploaded vm configs so that 30-local-dns-prepender can use it

fortinj66:
for vsphere ipi add cluster domain to the uploaded vm configs so that 30-local-dns-prepender can use it

Due to changes in FCOS 35 /etc/nsswitch.conf, we need to move the name resolution of api-int.<domain> to
systemd-resolved.  The way we do this is by adding a Domain attribute to /etc/systemd/resolved.conf.d/60-kni.conf.

Since we are using a slightly different prepender than the masters and workers the domain is not being
discovered automatically.  To address this we are adding the domain in data/data/vsphere/bootstrap/main.tf
which gets uploaded to the bootstrap instance on creation.

30-local-dns-prepender then downloads this information and adds it to /etc/systemd/resolved.conf.d/60-kni.conf

Note that we check for a non-empty domain before writing /etc/systemd/resolved.conf.d/60-kni.conf.

Fixes https://github.com/openshift/okd/issues/1182

Signed-off-by: John Fortin <fortinj66@gmail.com>

### DIFF
--- a/data/data/bootstrap/vsphere/files/etc/NetworkManager/dispatcher.d/30-local-dns-prepender.template
+++ b/data/data/bootstrap/vsphere/files/etc/NetworkManager/dispatcher.d/30-local-dns-prepender.template
@@ -20,11 +20,13 @@ EOF
     set +e
     if systemctl -q is-enabled systemd-resolved; then
         >&2 echo "NM resolv-prepender: Setting up systemd-resolved for local DNS"
-        if [[ ! -f /etc/systemd/resolved.conf.d/60-kni.conf ]]; then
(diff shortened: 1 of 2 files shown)


jcpowermac:
e2e-vsphere failure unrelated to PR changes

fortinj66:
okd-e2e-vsphere failures do not seem related to the PR either

kirankt (commented):
/lgtm

vrutkovs:
/retitle Bug 2076880: for vsphere ipi add cluster domain to the uploaded vm configs so that 30-local-dns-prepender can use it

PR #5791, opened 2022-04-11
Increase Unit Testing Coverage of AWS Install Config to Include ValidateForProvisioning

barbacbd:
installconfig/aws: Created a client API and the mock tests

Utilized an interface to expose functions for both mocking and for normal
circumstances. 

Fixes: 1947


barbacbd:
Removed an old commit using git commit fixup so that we could backport a bug that was found. 

https://github.com/openshift/installer/pull/5801

barbacbd:
Removed a piece of an old commit from validation.go that included a fix for empty string in ValidateForProvisioning. This fix will need to be backported. 

https://github.com/openshift/installer/pull/5801

barbacbd:
Requires https://github.com/openshift/installer/pull/5801 for full functionality of the unit tests.

barbacbd:
=== Unit ===
Appears to be failing on tests that only fail in CI. Removed my default location creds from ~/.aws and retested. All Pass

=== gofmt ===
There is a different version of GO used in the CI than the 1.17 required. No fmt errors during local testing.

PR #5796, opened 2022-04-11
vSphere session cleanup

bostrt:
This PR adds logouts for each location a vSphere client is created (excluding inside `terraform-provider-vsphere`). 

A new `Cleanup` method was added to `Destroyer` interface (this method name is open to suggestions) with the first implementation being for vSphere where logouts occur. 

Terraform provider for vSphere is not addressed in this PR due to https://github.com/hashicorp/terraform-provider-vsphere/issues/618

Relates to [SPLAT-471](https://issues.redhat.com//browse/SPLAT-471)

### DIFF
--- a/pkg/asset/installconfig/vsphere/client.go
+++ b/pkg/asset/installconfig/vsphere/client.go
@@ -39,31 +39,38 @@ func NewFinder(client *vim25.Client, all ...bool) Finder {
 	return find.NewFinder(client, all...)
 }
 
+// ClientLogout is empty function that logs out of vSphere clients
+type ClientLogout func()
+
 // CreateVSphereClients creates the SOAP and REST client to access
 // different portions of the vSphere API
 // e.g. tags are only available in REST
-func CreateVSphereClients(ctx context.Context, vcenter, username, password string) (*vim25.Client, *rest.Client, error) {
+func CreateVSphereClients(ctx context.Context, vcenter, username, password string) (*vim25.Client, *rest.Client, ClientLogout, error) {
 	ctx, cancel := context.WithTimeout(ctx, 60*time.Second)
 	defer cancel()
 
 	u, err := soap.ParseURL(vcenter)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 	u.User = url.UserPassword(username, password)
 	c, err := govmomi.NewClient(ctx, u, false)
 
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
 	restClient := rest.NewClient(c.Client)
 	err = restClient.Login(ctx, u.User)
 	if err != nil {
-		return nil, nil, err
+		c.Logout(context.TODO())
+		return nil, nil, nil, err
 	}
 
-	return c.Client, restClient, nil
+	return c.Client, restClient, func() {
+		c.Logout(context.TODO())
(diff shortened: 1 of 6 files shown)


pkg/asset/installconfig/vsphere/client.go
```diff
@@ -39,31 +39,37 @@ func NewFinder(client *vim25.Client, all ...bool) Finder {
...
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
-	restClient := rest.NewClient(c.Client)
+	restClient := rest.NewClient(client.Client)
 	err = restClient.Login(ctx, u.User)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
```
staebler, line 66: If there is an error creating the rest client, the govmomi client logout will not happen.

pkg/destroy/vsphere/vsphere.go
```diff
@@ -33,8 +33,9 @@ type ClusterUninstaller struct {
 	ClusterID string
 	InfraID   string
 
-	Client     *vim25.Client
-	RestClient *rest.Client
+	Client       *vim25.Client
+	RestClient   *rest.Client
+	ClientLogout vsphere.ClientLogout
```
staebler, line 38: This seems like a smell that `New` is causing resources to be held rather than having all resource acquisition and cleanup contained within `Run`. It is particularly smelly that this leaks into the interface affecting the destroyers of other platforms, too.

bostrt: Yes, I see how this could be solved all inside `Run` as well, without adding the `Cleanup`. I'm happy to go that way if that looks better to you.

bostrt: > This seems like a smell that New is causing resources to be held rather than having all resource acquisition and cleanup contained within Run

It looks like there's no particular pattern with which other destroyers are managing their API client. 

For instance: AWS, alibaba, Azure, and ibmcloud all keep a session or entire client inside of their `ClusterUninstaller` struct. On the other hand, oVirt manages the API client entirely within `Run`. 

> It is particularly smelly that this leaks into the interface affecting the destroyers of other platforms, too.

If I understand what you are saying correctly, I can explain that the new `Cleanup` would promote a separation of concerns for the Destroyer. vSphere isn't the only one that would benefit from this. Alibaba, oVirt, and baremetal (not sure this is used) could all take advantage of the new `Cleanup`. 

I'm curious what you think about this or if you still think consolidating into just `Run` is still the best move.

bostrt:
Removed `Cleanup` method so touching fewer files now. I'm not completely happy with continuing to store the client in the struct based on @staebler's other comments but without a refactor of the various destroy methods, it must be done.

staebler:
> Removed `Cleanup` method so touching fewer files now. I'm not completely happy with continuing to store the client in the struct based on @staebler's other comments but without a refactor of the various destroy methods, it must be done.

It is ugly because there is a blending of two areas of concern. The `ClusterUninstaller` type is being used (1) as a means of selecting an implementation and (2) as a store to ease passing data to functions. Presumably, this was done because it was convenient.

staebler (commented):
/lgtm

bostrt:
I believe https://github.com/openshift/release/pull/27834 is going to need to merge first before e2e-vsphere test will pass.

bostrt:
e2e-vsphere passes :+1:

PR #5801, opened 2022-04-12
Bug 2074659:Fix empty string usage in ValidateForProvisioning

barbacbd:
The zoneName and zonePath were always empty variables. The variables were initialized
outside of the if/else and then reinitialized using := in the if/else causing their
values to go out of scope before the function call to validateZoneRecords.

Fixes: 43224


patrickdillon:
/override ci/prow/tf-lint
test passed but context is stuck

PR #5804, opened 2022-04-13
images: Revert terraform version

rna-afk:
Changing the version back to the older version due to failures in
CI specifically vsphere upi.


kirankt (commented):
/lgtm

jcpowermac:
/hold
I might have a solution

jcpowermac:
Two choices:

#5808 - will be alittle while before ci starts this
#5806 - looking at ci this is past install

jcpowermac:
With #5808 in flight I think this can be closed

rna-afk:
Yep I'll close this.

PR #5811, opened 2022-04-14
Bug 2074210:Add new Regions to the GCP Installer configuration list

barbacbd:
Add three new GCP regions that were not previously listed as base regions for the installer to use. 
The following regions were added:

** Asia South 2
** Australia Southeast 2
** South America West 1

The installer was able to create the cluster and tear it down when these regions were selected.


makentenza:
@patrickdillon @barbacbd note that `southamerica-west1` Region doesn't support "N1" Machine Types which are the ones the installer is defaulting. There is an [RFE](https://issues.redhat.com/browse/RFE-2747) to enable this Region and we are expecting to enable this in 4.12

pkg/types/gcp/validation/platform.go
```diff
@@ -24,19 +24,23 @@ var (
...
 		"australia-southeast1":    "Sydney, Australia",
+		"australia-southeast2":    "Melbourne, Australia",
 		"europe-central2":         "Warsaw, Poland",
 		"europe-north1":           "Hamina, Finland",
 		"europe-west1":            "St. Ghislain, Belgium",
 		"europe-west2":            "London, England, UK",
 		"europe-west3":            "Frankfurt, Germany",
 		"europe-west4":            "Eemshaven, Netherlands",
 		"europe-west6":            "Zürich, Switzerland",
+		"europe-west8":            "Milan, Italy",
```
jianli-wei, line 39: Note that "europe-west8" Region doesn't support "N1" Machine Types too.

patrickdillon: @barbacbd  Let's not add `europe-west8` then for this reason. n1 is default, at least until #5815 merges. We are currently in the process of figuring out how to best handle adding new regions.

patrickdillon:
The merge commit 925d2337f0fbdc4d9cad2c555b340e4facdd7abf should not be included in the PR.

PR #5813, opened 2022-04-14
Ensure ignition host is set to API VIP for Nutanix platform

thunderboltsid:
This is to make sure the user does not have to manually set A records
for api-int.


jcpowermac:
cc: @staebler

staebler (commented):
/approve

PR #5814, opened 2022-04-14
Terraform: use absolute path

jhixson74:
Fix for this:
```
DEBUG Generating Cluster...                        
DEBUG creating terraform/bin/terraform file        
DEBUG creating terraform/plugins/openshift/local/aws directory 
DEBUG creating terraform/plugins/openshift/local/aws/terraform-provider-aws_1.0.0_linux_amd64.zip file 
INFO Creating infrastructure resources...         
DEBUG [INFO] running Terraform command: terraform/bin/terraform version -json 
DEBUG [INFO] running Terraform command: terraform/bin/terraform version -json 
DEBUG [INFO] running Terraform command: terraform/bin/terraform init -no-color -force-copy -input=false -backend=true -get=true -upgrade=false -plugin-dir=terraform/plugins 
ERROR failed to fetch Cluster: failed to generate asset "Cluster": failure applying terraform for "cluster" stage: failed to create cluster: failed doing terraform init: fork/exec terraform/bin/terraform: no such file or directory 
```


pkg/asset/cluster/cluster.go
```diff
@@ -92,8 +93,14 @@ func (c *Cluster) Generate(parents asset.Parents) (err error) {
 	if err := os.Mkdir(terraformDir, 0777); err != nil {
 		return errors.Wrap(err, "could not create the terraform directory")
 	}
+
+	terraformDirRp, err := realpath.Realpath(terraformDir)
```
staebler, line 97: Is there something that we need from `realpath.Realpath` that we would not get with a combination of `filepath.Abs` and `filepath.EvalSymLinks`? Do we even need to evaluate symbolic links, or is the underlying issue just with relative paths?

jhixson74: This is just me not thinking like a go programmer. C, Python, Perl, etc.. all use the native system call of realpath() to resolve relative paths to absolute paths. I am really surprised that go does not have this ability without calling multiple functions. It is not just for resolving symbolic links. Yes, this can be handled as you state. The issue here is the the terraform library is failing to initialize since it can't find the terraform binary. Once it has the absolute path it works. It looks like this is moot anyhow since the commit that broke it has been reverted.

PR #5815, opened 2022-04-14
Dynamic retrieval of GCP Regions for a Project

barbacbd:
**When the user enters a project get the list of regions for the project

**Use the static list of regions and their locations to apply a location
  to the dynamically grabbed regions (this is not currently provided with
  the GO api)

Fixes 1734


pkg/asset/installconfig/gcp/client.go
```diff
@@ -222,6 +224,34 @@ func (c *Client) GetProjects(ctx context.Context) (map[string]string, error) {
...
+func (c *Client) GetProjectRegions(ctx context.Context, project string) ([]string, error) {
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+
+	gc, err := google.DefaultClient(ctx, compute.CloudPlatformScope)
+	if err != nil {
+		return nil, err
+	}
+
+	service, err := compute.New(gc)
```
patrickdillon, line 237: can we use `c.getComputeService`?

pkg/asset/installconfig/gcp/client.go
```diff
@@ -222,6 +224,34 @@ func (c *Client) GetProjects(ctx context.Context) (map[string]string, error) {
 	return projects, nil
 }
 
+// GetProjectRegions gets the regions that are valid for the project. An error is returned when unsuccessful
+func (c *Client) GetProjectRegions(ctx context.Context, project string) ([]string, error) {
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
```
patrickdillon, line 229: This context is shared between two calls. I think it should probably be reset for each.

pkg/asset/installconfig/gcp/client.go
```diff
@@ -222,6 +224,34 @@ func (c *Client) GetProjects(ctx context.Context) (map[string]string, error) {
...
+	}
+
+	service, err := compute.New(gc)
+	if err != nil {
+		return nil, err
+	}
+
+	r, err := service.Regions.List(project).Context(ctx).Do()
+	if err != nil {
+		return nil, err
```
patrickdillon, line 244: Some of these:

```go
return nil, err
```
will go away if we change how the service is instantiated, but in general when errors are returned like this we should `wrap` them so we have an idea of where failures are occurring when errors happen.

pkg/asset/installconfig/gcp/gcp.go
```diff
@@ -82,13 +82,40 @@ func selectProject(ctx context.Context) (string, error) {
...
+				shortRegions = append(shortRegions, id)
+				found = true
+				break
+			}
+		}
+		if !found {
+			longRegions = append(longRegions, fmt.Sprintf("%s", validRegions[i]))
+			shortRegions = append(shortRegions, validRegions[i])
+		}
+	}
```
patrickdillon, line 117: Would it make more sense for `GetProjectRegions` to return both the short and long names? perhaps as a map[shortregions]longRegions?

pkg/asset/installconfig/gcp/gcp.go
```diff
@@ -103,13 +130,25 @@ func selectRegion(project string) (string, error) {
...
+	for idx := range shortRegions {
+		if shortRegions[idx] == defaultRegion {
+			foundIdx = idx
+			break
+		}
+	}
+
+	if foundIdx >= 0 {
+		defaultRegion = validRegions[0]
+	}
```
patrickdillon, line 143: 1. Is this defensive programming or is it possible the list of regions would not have `us-central1`? 
2. A more idiomatic way of writing this would be something like
```suggestion
	var found int
	for i, v := range shortRegions {
		if v == defaultRegion {
			found = i
			break
		}
	}
```

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -248,3 +248,19 @@ func ValidateEnabledServices(ctx context.Context, client API, project string) er
...
+	validRegions, err := client.GetProjectRegions(ctx, project)
+	if err != nil {
+		return err
+	}
+
+	for r := range validRegions {
+		if region == validRegions[r] {
+			return nil
+		}
+	}
```
patrickdillon, line 263: Seems like another place where having `GetProjectRegions` return a map would be good.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -248,3 +248,19 @@ func ValidateEnabledServices(ctx context.Context, client API, project string) er
...
+
+	return fmt.Errorf("region: %s not found: %s", region, strings.Join(validRegions, ","))
```
patrickdillon, line 265: `strings.Join(validRegions, ","))` what does the output of this look like? I'm imagining a very long list.

patrickdillon (commented):
This is a good start. The main things I noticed were that the service invocation doesn't seem to follow the pattern of other compute services and we might be able to improve the return type of the `GetProjectRegions` function.

pkg/asset/installconfig/gcp/gcp.go
barbacbd: That was a thought, but I was allowing the caller to alter the data as they pleased. Technically the long name may not exist. Is that something that we still want returned in a map?

pkg/asset/installconfig/gcp/validation.go
barbacbd: It definitely could be a very long list. should we newline separate it? I just noticed that we have some VERY long string errors, so I wasn't entirely sure if this made sense or fit our norms.

pkg/asset/installconfig/gcp/client.go
```diff
@@ -222,6 +224,34 @@ func (c *Client) GetProjects(ctx context.Context) (map[string]string, error) {
...
+	}
+
+	regionsCP := make(map[string]string)
+	for i := range r.Items {
+		if value, ok := gcpv.Regions[r.Items[i].Name]; ok {
+			regionsCP[r.Items[i].Name] = value
+		} else {
+			regionsCP[r.Items[i].Name] = ""
+		}
+	}
```
rna-afk, line 250:
```suggestion
	regionsCP := make(map[string]string)
	for _, region := range r.Items {
		if value, ok := gcpv.Regions[region.Name]; ok {
			regionsCP[region.Name] = value
		} else {
			regionsCP[region.Name] = ""
		}
	}
```

pkg/asset/installconfig/gcp/client.go
```diff
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	gcpv "github.com/openshift/installer/pkg/types/gcp/validation"
```
rna-afk, line 9: We follow an order for imports so its best if we move this import to a new section below.

golang libraries

external libraries (non-openshift)

openshift libraries

pkg/asset/installconfig/gcp/client.go
```diff
@@ -222,6 +224,34 @@ func (c *Client) GetProjects(ctx context.Context) (map[string]string, error) {
...
+func (c *Client) GetProjectRegions(ctx context.Context, project string) (map[string]string, error) {
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+
+	svc, err := c.getComputeService(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	gctx := context.Background()
```
rna-afk, line 237: nit: Probably best to just reuse the variable. Fine either way.
```suggestion
	ctx = context.Background()
```

pkg/asset/installconfig/gcp/gcp.go
```diff
@@ -82,13 +81,33 @@ func selectProject(ctx context.Context) (string, error) {
...
+	shortRegions := make([]string, 0, len(validRegions))
+	for key, value := range validRegions {
+		shortRegions = append(shortRegions, key)
+		if value == "" {
+			longRegions = append(longRegions, key)
+		} else {
+			longRegions = append(longRegions, fmt.Sprintf("%s (%s)", key, value))
+		}
+	}
+
```
rna-afk, line 110: nit: we can avoid the found logic with this? would reduce a for loop but maybe it's not worth reducing?
```suggestion
        defaultRegionName := ""
        defaultRegion := "us-central1"
	longRegions := make([]string, 0, len(validRegions))
	shortRegions := make([]string, 0, len(validRegions))
	for key, value := range validRegions {
		shortRegions = append(shortRegions, key)
		regionDesc := key
		if value != "" {
		        regionDesc := fmt.Sprintf("%s (%s)", key, value)
		}
		longRegions = append(longRegions, regionDesc)
		if key == defaultRegion {
		       defaultRegionName = regionDesc
		}
	}

```

rna-afk (commented):
Looks good. Just a few suggestions.

pkg/asset/installconfig/gcp/gcp.go
barbacbd: @rna-afk The reason that I did not choose the above was specifically because my solution will offer a region to the user even when the region does not have an exact known location (specific city/country). This will allow us to still offer the region as acceptable, and we will be able to find missing regions easier in the future.

pkg/asset/installconfig/gcp/client.go
barbacbd: @rna-afk  I was told specifically not to reuse the context earlier. I can change this around though.

rna-afk: My suggestion is to just reuse the variable and reset the context? I think we don't really need to use another variable but yeah it's just a nitpick.

barbacbd: @rna-afk should we reset the variable since it was a parameter that was passed in ?

pkg/asset/installconfig/gcp/gcp.go
rna-afk: That sounds right but could you elaborate? I suggested this code hoping to exactly replicate your code logic but reduce the for loop count but I may have made some mistakes. IIUC your concern of adding a region to the list even if it does not have a proper description(longName) is taken care of by this code?

pkg/asset/installconfig/gcp/client.go
patrickdillon: I believe you should be deriving children contexts from the parent contexts. See: https://pkg.go.dev/context

So a parent context is passed into the function, this is typically TODO or background. And then before making an API call you can set a timeout to say: "I want this API call to take at most X amount of time." What you're doing right now is saying, "I want API Call A and API Call B to take at most X amount of time." I think what we should do is: "I want API Call A to take at most X amount of time and API Call B to take at most Y amount of time."

pkg/asset/installconfig/gcp/client.go
```diff
@@ -222,6 +225,31 @@ func (c *Client) GetProjects(ctx context.Context) (map[string]string, error) {
...
 }
 
+// GetProjectRegions gets the regions that are valid for the project. An error is returned when unsuccessful
+func (c *Client) GetProjectRegions(ctx context.Context, project string) (map[string]string, error) {
+	svc, err := c.getComputeService(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	gctx := context.Background()
```
r4f4, line 235: I see from previous reviews that you were told not to reuse contexts, but is this correct? There's no way to propagate cancellation/timeout through the argument's context here since you're creating a new one. Perhaps a similar pattern to the one used in `GetZones` below would be more appropriate?
```
	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
	defer cancel()
	svc, err := c.getComputeService(ctx)
	if err != nil {
		return nil, err
	}
        r, err := svc.Regions.List(project).Context(ctx).Do()
        ...
```

pkg/asset/installconfig/gcp/client.go
```diff
@@ -222,6 +225,31 @@ func (c *Client) GetProjects(ctx context.Context) (map[string]string, error) {
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get regions for project %s", project)
+	}
+
+	regionsCP := make(map[string]string)
+	for _, region := range r.Items {
+		if value, ok := gcpv.Regions[region.Name]; ok {
+			regionsCP[region.Name] = value
+		} else {
+			regionsCP[region.Name] = ""
```
r4f4, line 246: Later on when you use this value, you skip the regions which are not already known by the installer. Any reason for not skipping them here instead?

r4f4, line 246: Later on when you use this value, you skip the regions which are not already known by the installer. Any reason why not do this here already?

pkg/asset/installconfig/gcp/client.go
barbacbd: The concern here was that the same context that was passed in as an arg is assigned again. The other functions may do this, but is it the correct approach? Reusing contexts appears to be the accepted practice on server side software. Its easy enough to change, but should I? @r4f4  @patrickdillon

pkg/asset/installconfig/gcp/client.go
barbacbd: No thats a good point. Now that I am skipping them I can just put them in to the Interface

pkg/asset/installconfig/gcp/client.go
patrickdillon: I have been discussing this with @jhixson74 here: https://github.com/openshift/installer/pull/5697#discussion_r863925907 which includes a link to go playground that illustrates the problem I was trying to describe. Note that the playground does not indicate WHAT TO DO, but  just tries to illustrate the problem as I see it. 

Calling it a problem, might be a strong word. I just want to make sure we're aware of what we're doing. @r4f4's example above gives 1 minute total to get the compute service and list the regions. Is that what we want to do? Why not give 30 seconds each? or a minute each? 

When adding a deadline/timeout to a context a child context is created from the parent context, so it should be well supported to create multiple timeouts/deadlines from a parent context. If we do want to pursue that approach I just think we need to figure out the right style/pattern. Does it work to just reuse the `ctx variable` and shadow the existing `ctx` value? This is not what the go playground does but I think it's worth checking out.

barbacbd: It actually looks like AN accepted practice is to use the parent (context parameter) to create the one used in this function. I think that I should also use the parent to create the second context. 

```
gctx := context.WithTimeout(ctx, 1*time.Minute)
```

would become 

```
gctx := context.Background()
```

To avoid confusion, maybe we should change the name of the context that we create from the parent:
```
ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
```

to
```
child_ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
```

We can also change them to 30 seconds or something much more reasonable.

pkg/asset/installconfig/gcp/gcp.go
```diff
@@ -82,13 +81,36 @@ func selectProject(ctx context.Context) (string, error) {
...
+	longRegions := make([]string, 0, len(validRegions))
+	shortRegions := make([]string, 0, len(validRegions))
+	for key, value := range validRegions {
+		shortRegions = append(shortRegions, key)
+		regionDesc := fmt.Sprintf("%s (%s)", key, value)
+		longRegions = append(longRegions, regionDesc)
+
+		if key == defaultRegion {
+			defaultRegionName = regionDesc
+		}
```
jstuever, line 111: It might be worth using a structure here instead of using three different variables.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -248,3 +248,17 @@ func ValidateEnabledServices(ctx context.Context, client API, project string) er
...
+	validRegions, err := client.GetRegions(ctx, project)
+	if err != nil {
+		return err
+	}
+
+	if _, ok := validRegions[region]; ok {
+		return nil
+	}
+
+	return fmt.Errorf("the following region is invalid: %s", region)
```
jstuever, line 263: Reverse the logic here to enable additional validations....
```
if _, ok := validRegions[region]; not ok {
        return fmt.Errorf("the following region is invalid: %s", region)
}
return nil
```

jstuever (changes requested):
A few naming and logic concerns in addition to the context concerns pointed out by others.

pkg/asset/installconfig/gcp/gcp.go
```diff
@@ -102,14 +124,17 @@ func selectRegion(project string) (string, error) {
...
+
 	var selectedRegion string
-	err := survey.Ask([]*survey.Question{
+	err = survey.Ask([]*survey.Question{
 		{
 			Prompt: &survey.Select{
 				Message: "Region",
 				Help:    "The GCP region to be used for installation.",
-				Default: fmt.Sprintf("%s (%s)", defaultRegion, validation.Regions[defaultRegion]),
+				Default: fmt.Sprintf("%s", defaultRegionName),
```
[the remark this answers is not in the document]
barbacbd: What would happen if by some miracle it wasn't found? Is that possible or will our default always be found?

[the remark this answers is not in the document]
barbacbd: We can definitely shorten it to `Default: defaultRegionName,`

pkg/asset/installconfig/gcp/client.go
```diff
@@ -222,6 +225,34 @@ func (c *Client) GetProjects(ctx context.Context) (map[string]string, error) {
...
+	defer cancel()
+
+	svc, err := c.getComputeService(computeCtx)
+	if err != nil {
+		return nil, err
+	}
+
+	regionCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+	gcpRegionsList, err := svc.Regions.List(project).Context(regionCtx).Do()
```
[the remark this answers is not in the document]
patrickdillon: I am fine with merging this PR using the existing pattern of sharing contexts, although I disagree with it for the reasons mentioned above. Here is an example of the existing pattern: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/gcp/client.go#L72-L85

If we go that route, I would prefer this be updated to use the same variable naming conventions in case we decide to fix the problem later in bulk.

I'm also happy to have the pattern fixed here, in a separate commit.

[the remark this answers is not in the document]
barbacbd: This is simple enough to add to this PR in a separate commit. If I follow the pattern the following functions will be changed to match the pattern:

```
GetMachineType
GetNetwork
GetSubnetworks
GetZones
```

The `getComputeService` will be altered to make the child context.

pkg/asset/installconfig/gcp/client.go
```diff
@@ -222,6 +225,34 @@ func (c *Client) GetProjects(ctx context.Context) (map[string]string, error) {
...
+		return nil, errors.Wrapf(err, "failed to get regions for project %s", project)
+	}
+
+	validatedRegions := make(map[string]string)
+	for _, region := range gcpRegionsList.Items {
+		// Only add validated regions
+		if value, ok := gcpValidation.Regions[region.Name]; ok {
+			validatedRegions[region.Name] = value
+		}
+	}
```
patrickdillon, line 251: I don't *think* this is where we want to filter the list of regions from the GCP API down to the list whitelisted by QE.  I think this function should just return the complete list from the API, and that should be used to validate the region passed in the install config.

pkg/asset/installconfig/gcp/gcp.go
```diff
@@ -82,13 +81,38 @@ func selectProject(ctx context.Context) (string, error) {
...
+	client := &Client{
+		ssn: ssn,
+	}
+
+	selectRegionCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	validRegions, err := client.GetRegions(selectRegionCtx, project)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get regions")
```
patrickdillon, line 99: These are the regions that would be included in the survey, so it makes sense to filter here based on the list from pkg types.

pkg/asset/installconfig/gcp/client.go
barbacbd: Is it a waste of resources to convert it to a map? I thought that we decided to pass around the map. If/when we get the regions it will return a list. If we don't filter it here (and we want to return a map) it will be filled with empty values.

patrickdillon: My main concern is the functional correctness, so let's go into that first. The behavior we want is:

- the survey is populated by the subset of regions from the GCP API that appear in the "qe whitelist" in `gcpValidation.Regions`
- the install config will accept any region from the GCP API

The example to test this we came up with is `europe-west8`. `europe-west8` (at the moment) should not appear in the survey, but should be a valid region when passed in the install config. I believe if we filter the region list to the validated QE list here, the second scenario will not work. 

 >  If we don't filter it here (and we want to return a map) it will be filled with empty values

IMHO this should be filled with data from the API, not the static list from our code. But I am happy to go with a different preferred solution as long as the functionality works in the desired way.

patrickdillon:
`europe-west8` in the install config is still not working:
```
ERROR failed to fetch Install Config: failed to load asset "Install Config": failed to create install config: invalid "install-config.yaml" file: platform.gcp.region: Unsupported value: "europe-west8": supported values: "asia-east1", "asia-east2", "asia-northeast1", "asia-northeast2", "asia-northeast3", "asia-south1", "asia-south2", "asia-southeast1", "asia-southeast2", "australia-southeast1", "australia-southeast2", "europe-central2", "europe-north1", "europe-west1", "europe-west2", "europe-west3", "europe-west4", "europe-west6", "northamerica-northeast1", "northamerica-northeast2", "southamerica-east1", "us-central1", "us-east1", "us-east4", "us-west1", "us-west2", "us-west3", "us-west4" 
```

I think you need to get rid of this: https://github.com/openshift/installer/blob/master/pkg/types/gcp/validation/platform.go#L65-L67

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -248,3 +249,24 @@ func ValidateEnabledServices(ctx context.Context, client API, project string) er
...
+			return append(allErrs, field.InternalError(fieldPath.Child("project"), err))
+		} else if len(computeRegions) == 0 {
+			return append(allErrs, field.Invalid(fieldPath.Child("project"), ic.GCP.ProjectID, "no regions found"))
+		}
+
+		validRegions := getValidatedRegions(computeRegions)
+		if _, ok := validRegions[ic.GCP.Region]; !ok {
+			return append(allErrs, field.Invalid(fieldPath.Child("region"), ic.GCP.Region, "invalid region"))
+		}
+	}
```
patrickdillon, line 269: This doesn't look right to me... wouldn't this require that any regions in the install config be in the QE whitelist? When trying to test I hit a different issue.

patrickdillon (commented):
When trying to test whether I could create an install-config using the `europe-west8` region I was prompted incorrectly for my service account credentials. It looks like the session creation here might be incorrect:

```
# ./openshift-install create install-config --dir c
? Service Account (absolute path to file or JSON content) [Enter 2 empty lines to finish]
```

patrickdillon:
> When trying to test whether I could create an install-config using the `europe-west8` region I was prompted incorrectly for my service account credentials. It looks like the session creation here might be incorrect:
> 
> ```
> # ./openshift-install create install-config --dir c
> ? Service Account (absolute path to file or JSON content) [Enter 2 empty lines to finish]
> ```

Nevermind, it looks like the credentials issue for me are local.

pkg/asset/installconfig/gcp/validation.go
patrickdillon: So ignore my earlier comment about the credentials, but it looks like my comment here is correct. This block is preventing specifying europe-west8 in an install config:

```
ERROR failed to fetch Install Config: failed to load asset "Install Config": failed to create install config: platform.gcp.region: Invalid value: "europe-west8": invalid region 
```

patrickdillon:
/retitle Bug 2102362: Dynamic retrieval of GCP Regions for a Project

PR #5821, opened 2022-04-19
Ensure Nutanix API VIP is populated in the SAN for the MCS Cert

thunderboltsid:
Without the API VIP populated in the SAN for the MCS Cert, the
nodes created by the installer will fail to ignite due to a x509
verification error.


jcpowermac:
/lgtm

cc: @staebler

staebler (commented):
/approve

PR #5822, opened 2022-04-19
pkg/asset/installconfig: fix dropped errors

alrs:
This fixes two `err` variables that were being lost in a `select` due to use of `:=`. It also picks up an error that was being dropped outright.

### DIFF
--- a/pkg/asset/installconfig/platform.go
+++ b/pkg/asset/installconfig/platform.go
@@ -112,6 +112,9 @@ func (a *platform) Generate(asset.Parents) error {
 		}
 	case powervs.Name:
 		a.PowerVS, err = powervsconfig.Platform()
+		if err != nil {
+			return err
+		}
 	case nutanix.Name:
 		a.Nutanix, err = nutanixconfig.Platform()
 		if err != nil {
--- a/pkg/asset/installconfig/platformprovisioncheck.go
+++ b/pkg/asset/installconfig/platformprovisioncheck.go
@@ -111,13 +111,15 @@ func (a *PlatformProvisionCheck) Generate(dependencies asset.Parents) error {
 			return err
 		}
 	case alibabacloud.Name:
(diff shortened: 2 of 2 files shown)


pkg/asset/installconfig/platformprovisioncheck.go
```diff
@@ -111,13 +111,15 @@ func (a *PlatformProvisionCheck) Generate(dependencies asset.Parents) error {
 			return err
 		}
 	case alibabacloud.Name:
-		client, err := ic.AlibabaCloud.Client()
+		var client *alibabacloudconfig.Client
+		client, err = ic.AlibabaCloud.Client()
 		if err != nil {
 			return err
 		}
 		err = alibabacloudconfig.ValidateForProvisioning(client, ic.Config, ic.AlibabaCloud)
```
r4f4, line 119: We could also keep the above code as is and just return the result of the call here.

r4f4 (commented):
Most of the other `case`s just return immediately in case of error, so we might as well just get rid of the outer `var err error` declaration.

pkg/asset/installconfig/platformprovisioncheck.go
sadasu: Agree with @r4f4 that we can keep the code as is. I am not sure we are losing any error here.

pkg/asset/installconfig/platform.go
```diff
@@ -112,6 +112,9 @@ func (a *platform) Generate(asset.Parents) error {
 		}
 	case powervs.Name:
 		a.PowerVS, err = powervsconfig.Platform()
+		if err != nil {
```
sadasu, line 115: These are the only changes needed as part of this PR, IMHO.

r4f4:
/close
This PR has been made obsolete by https://github.com/openshift/installer/pull/5958 and https://github.com/openshift/installer/pull/6595

PR #5824, opened 2022-04-20
vsphere: remove unused namer interface

patrickdillon:
Removes leftover namer interface. The implementation using the interface was removed in e7dfefa7d51d5d4fbee1a37412ef6f8ab7c16d38 but the interface was left behind. We have no need for the interface so clean it
up.


kirankt (commented):
/lgtm

PR #5825, opened 2022-04-20
Unpack to install dir

jhixson74:
PR https://github.com/openshift/installer/pull/5715 with fixes for terraform path. 

https://issues.redhat.com/browse/CORS-1952

### DIFF
--- a/pkg/destroy/bootstrap/bootstrap.go
+++ b/pkg/destroy/bootstrap/bootstrap.go
@@ -47,6 +47,19 @@ func Destroy(dir string) (err error) {
 		varFiles = append(varFiles, stage.OutputsFilename())
 	}
 
+	terraformDir := filepath.Join(dir, "terraform")
+	if err := os.Mkdir(terraformDir, 0777); err != nil {
+		return errors.Wrap(err, "could not create the terraform directory")
+	}
+
+	terraformDirPath, err := filepath.Abs(terraformDir)
+	if err != nil {
+		return errors.Wrap(err, "could not get absolute path of terraform directory")
+	}
+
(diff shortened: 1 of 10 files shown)


pkg/destroy/bootstrap/bootstrap.go
```diff
@@ -64,6 +70,7 @@ func Destroy(dir string) (err error) {
 		tempDir, err := ioutil.TempDir("", fmt.Sprintf("openshift-install-%s-", stage.Name()))
 		if err != nil {
 			return errors.Wrap(err, "failed to create temporary directory for Terraform execution")
+
```
patrickdillon, line 73: extra line

patrickdillon (commented):
/approve

This looks good. Can you just remove the extra line that snuck in?

patrickdillon:
Oops. I hit CTRL + Enter instead of enter with my earlier comment.

This looks good. I tested with symlinks, creating a `c` dir with the install assets and then a symlink `s/c2` dir and installed with `openshift-install create cluster --dir s/c2` with no problems.

We should merge this as a single commit. I'm looking into now whether the tide merge label can automatically squash all commits, or whether those commits need to be labeled as fixups.

pkg/destroy/bootstrap/bootstrap.go
jhixson74: Fixed.

patrickdillon:
@jhixson74 I think `/label tide/merge-method-squash` will only squash fixup commits. Can you either squash the commit manually or make it a fixup? Thanks!

janosdebugs:
FYI @mburman5

PR #5836, opened 2022-04-22
Bug 2064170: Fixed duplicate strings in explain 

AnnaZivkovic:


patrickdillon:
/hold
Looks like a bad rebase.

MayXuQQ:
pre-merge check with registry.build01.ci.openshift.org/ci-ln-bgpsphk/release:latest
osDisk.diskType, output same for controlPlane, compute and defaultMachinePlatform, even under control plane node, valid value are different
./openshift-install explain installconfig.controlPlane.platform.azure.osDisk.diskType
KIND:     InstallConfig
VERSION:  v1

RESOURCE: <string>
  Valid Values: "Standard_LRS","Premium_LRS","StandardSSD_LRS"
  DiskType defines the type of disk. For control plane nodes, the valid values are Premium_LRS and StandardSSD_LRS. Default is Premium_LRS.
/label qe-approved
/bugzilla cc-qa

pkg/explain/printer.go
```diff
@@ -31,7 +31,7 @@ func (p printer) PrintResource(schema *apiextv1.JSONSchemaProps) {
 	}
 	io.WriteString(p.Writer, fmt.Sprintf("RESOURCE: <%s>\n", resource))
 
-	p.printLables(2, schema)
+	p.printLabels(2, schema)
```
patrickdillon, line 34: Can you squash these fixups into the first commit??

barbacbd:
I was curious about the formatting since the BZ suggests two different formats between Description and Comment 3. I ran 

`[bbarbach@bbarbach installer]$ bin/openshift-install explain installconfig.controlPlane
KIND:     InstallConfig
VERSION:  v1

RESOURCE: <object>
  ControlPlane is the configuration for the machines that comprise the control plane.

FIELDS:
    architecture <string>
      Default: "amd64"
      Valid Values: "","amd64"
      Architecture is the instruction set architecture of the machine pool. Defaults to amd64.

    hyperthreading <string>
      Default: "Enabled"
      Valid Values: "","Enabled","Disabled"
      Hyperthreading determines the mode of hyperthreading that machines in the pool will utilize. Default is for hyperthreading to be enabled.

    name <string> -required-
      Name is the name of the machine pool. For the control plane machine pool, the name will always be "master". For the compute machine pools, the only valid name is "worker".

    platform <object> -required-
      Platform is configuration for machine pool specific to the platform.

    replicas <integer>
      Format: int64
      Replicas is the machine count for the machine pool.`

the default is listed at the top of these rather than on the same line. I prefer this format, but open to change.

patrickdillon (commented):
The commit structure looks good, but can you add messages according to [the guidelines here](https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#commit-message-format).

Particularly, can you explain (for posterity( why you are making the change in the first commit. The other commits are more self explanatory so a single sentence would do.

AnnaZivkovic:
@barbacbd I think it would be harder to scan for the relevant information if the valid values and defaults are on the same line. I can see where you are coming from aesthetically tho.

barbacbd:
Like I said definitely open to change. But that begs the question, which to change? @patrickdillon This could probably be a new card to ensure that all `explain` shows the data in the same format correct?

patrickdillon:
It took some iteration with the BZ to figure out the problem. The problem was that the structured output differed between whether the explain was explaining a struct or a field. The newly updated commit message describes it:

> PrintResorces was changed so that it would have the ability to print out the valid values, defaults, and format when using explain for sub-fields in a struct. Without this change those fields would only be displayed when explaining the entire struct. Previously, using explain on a field of the struct would only print the description.

@barbacbd I'm not sure if I understand your suggestion. Are you talking about the repetition of default values between the description and the structured data? At this stage, I would say improving anything that isn't broken would be pretty low priority.  I discussed a project with @AnnaZivkovic in terms of having the same logic for explain generate the in-repo docs. We don't have a card for it yet, but any potential fine tuning of explain could be lumped in to that effort.

/lgtm
/approve

barbacbd:
@patrickdillon I am not concerned with any repetition. I am suggesting that the format of the output be the same between different explanations. We shouldn't add the default to the same line as the long explanation if it is the first line in other explanations. The problem is consistency not repetition. I don't think that there is a right or wrong way to do it, but they should be the same. 

`hyperthreading <string>
  Default: "Enabled"
  Valid Values: "","Enabled","Disabled"
  Hyperthreading determines the mode of hyperthreading that machines in the pool will utilize. Default is for hyperthreading to be enabled.`

Above does not match the output that looks like 

`hyperthreading <string>
  Valid Values: "","Enabled","Disabled"
  Hyperthreading determines the mode of hyperthreading that machines in the pool will utilize. Default is for hyperthreading to be enabled. Default: "Enabled"`

Again, not a major issue but something to think about.

PR #5838, opened 2022-04-25
Bug 2078875: Delete all the ports from tagged Neutron networks. 

gryf:
Currently, we are trying to remove ports, which are tagged, but
sometimes it happens, that port is created but tagging has failed, or we
didn't get response from Neutron API, resulting with untagged port.

Having network already tagged we can use its id to select all the port
within such network, so that all the ports, not only tagged would be
deleted.

### DIFF
--- a/pkg/destroy/openstack/openstack.go
+++ b/pkg/destroy/openstack/openstack.go
@@ -103,7 +103,7 @@ func (o *ClusterUninstaller) Run() (*types.ClusterQuota, error) {
 		"deleteServerGroups":    deleteServerGroups,
 		"deleteTrunks":          deleteTrunks,
 		"deleteLoadBalancers":   deleteLoadBalancers,
-		"deletePorts":           deletePorts,
+		"deletePorts":           deletePortsByFilter,
 		"deleteSecurityGroups":  deleteSecurityGroups,
 		"clearRouterInterfaces": clearRouterInterfaces,
 		"deleteSubnets":         deleteSubnets,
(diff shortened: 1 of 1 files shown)


mandre:
This looks like something we'll want to backport. Could you please create a BZ and reference it in the PR title?

gryf:
Yup, I'll do that.

pkg/destroy/openstack/openstack.go
```diff
@@ -103,7 +103,6 @@ func (o *ClusterUninstaller) Run() (*types.ClusterQuota, error) {
 		"deleteServerGroups":    deleteServerGroups,
 		"deleteTrunks":          deleteTrunks,
 		"deleteLoadBalancers":   deleteLoadBalancers,
-		"deletePorts":           deletePorts,
```
dulek, line 106: I don't think we're guaranteed that all ports we need to delete are on a tagged networks. If nodes have additional user-specified networks, then these might not be tagged, yet we still need to delete these ports. Will Nova take care of that? I doubt it, as MAPO/CAPO creates the ports itself.

dulek (changes requested):
@mdbooth, your opinion might help here.

pkg/destroy/openstack/openstack.go
dulek: Yes, I would think so.

MaysaMacedo: +1 and we would need to ignore not found errors, in case both goroutines try to delete same Port

pkg/destroy/openstack/openstack.go
```diff
@@ -320,20 +320,45 @@ func deleteServerGroups(opts *clientconfig.ClientOpts, filter Filter, logger log
...
 		return false, nil
 	}
+	return result, err
+}
+
+func deletePortsByFilter(opts *clientconfig.ClientOpts, filter Filter, logger logrus.FieldLogger) (bool, error) {
+
 	tags := filterTags(filter)
 	listOpts := ports.ListOpts{
 		TagsAny: strings.Join(tags, ","),
```
dulek, line 341: Wait, why is that `TagsAny`? Not a question for this patch, just curious.

pkg/destroy/openstack/openstack.go
dulek: Hm, true, seems like 404 on DELETE is not ignored now. But from the other hand it's not a fatal issue, on the next port list we'll just go through this.

pkg/destroy/openstack/openstack.go
mandre: At least one of the tags must be present. Let's just say it's future proof if we start filtering with more than just the cluster ID :man_shrugging:

mandre (commented):
/lgtm
/approve

PR #5840, opened 2022-04-25
Azure: Fix Capability Name for Accelerated Networking Validation

patrickdillon:
The validation for Azure Accelerated Networking had the incorrect name for the capability, which meant that validation was being skipped for VMs that did not have the capability.

This also adds a check in case the capability is missing from the VM
metadata.

### DIFF
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -101,6 +101,7 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 	}
 
 	ultraSSDAvailable := false
+	acceleratedNetworkingAvailable := false
 	for _, capability := range *typeMeta.Capabilities {
 
 		if strings.EqualFold(*capability.Name, "vCPUsAvailable") {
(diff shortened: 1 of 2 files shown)


pkg/asset/installconfig/azure/validation.go
```diff
@@ -128,19 +129,20 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 			}
 		} else if strings.EqualFold(*capability.Name, "UltraSSDAvailable") {
 			ultraSSDAvailable = strings.EqualFold(*capability.Value, "True")
-		} else if strings.EqualFold(*capability.Name, string(aztypes.VMnetworkingTypeAccelerated)) {
-			if vmNetworkingType == string(aztypes.VMnetworkingTypeAccelerated) && !strings.EqualFold(*capability.Value, "True") {
-				errMsg := fmt.Sprintf("vm networking type is not supported for instance type %s", instanceType)
-				allErrs = append(allErrs, field.Invalid(fieldPath.Child("vmNetworkingType"), vmNetworkingType, errMsg))
-			}
+		} else if strings.EqualFold(*capability.Name, "AcceleratedNetworkingEnabled") {
```
kwoodson, line 132: I created a type for this magic string here:
https://github.com/openshift/installer/blob/master/pkg/types/azure/machinepool.go#L50

patrickdillon: Ah! I switched to the type in the pr.

patrickdillon:
oops. maybe I broke the unit tests. looking.

patrickdillon:
> The unit tests are broken because the networkType is "accelerated" by default in the `install-config` but Disabled (i.e. not set) by default in the instanceTypes of the unit tests.

Oops. Yeah Kenny and I are discussing in slack. I will ping you.

MayXuQQ:
pre-merge check with registry.build01.ci.openshift.org/ci-ln-dqvp7ck/release:latest 
when use "Standard_B4ms" will prompt 

> ERROR failed to fetch Metadata: failed to load asset "Install Config": failed to create install config: controlPlane.platform.azure.vmNetworkingType: Invalid value: "Accelerated": vm networking type is not supported for instance type Standard_B4ms
that is expected.

patrickdillon:
/hold

Unit tests are failing because we're setting the default for enable accelerated networking to true and the error messages are not correct for the unit tests. 

In conversation with @kwoodson we decided to set defaults dynamically based on capability. This PR would depend on those fixes or get closed in favor of that new PR.

PR #5841, opened 2022-04-25
machines: Set defaults for machine instance types

rna-afk:
Setting all compute nodes to use the instance types that provide
4 VPUs and 16 GB RAM for AWS, GCP and VSphere platforms.

### DIFF
--- a/pkg/asset/machines/worker.go
+++ b/pkg/asset/machines/worker.go
@@ -115,7 +115,7 @@ func defaultAzureMachinePoolPlatform() azuretypes.MachinePool {
 
 func defaultGCPMachinePoolPlatform() gcptypes.MachinePool {
 	return gcptypes.MachinePool{
-		InstanceType: "n1-standard-4",
+		InstanceType: "n2-standard-4",
 		OSDisk: gcptypes.OSDisk{
 			DiskSizeGB: powerOfTwoRootVolumeSize,
 			DiskType:   "pd-ssd",
@@ -156,9 +156,9 @@ func defaultOvirtMachinePoolPlatform() ovirttypes.MachinePool {
 
 func defaultVSphereMachinePoolPlatform() vspheretypes.MachinePool {
 	return vspheretypes.MachinePool{
-		NumCPUs:           2,
-		NumCoresPerSocket: 2,
-		MemoryMiB:         8192,
+		NumCPUs:           4,
+		NumCoresPerSocket: 1,
+		MemoryMiB:         16384,
 		OSDisk: vspheretypes.OSDisk{
 			DiskSizeGB: decimalRootVolumeSize,
 		},
@@ -185,11 +185,11 @@ func defaultNutanixMachinePoolPlatform() nutanixtypes.MachinePool {
 	}
 }
 
-func awsDefaultWorkerMachineTypes(region string, arch types.Architecture) []string {
+func awsDefaultMachineTypes(region string, arch types.Architecture) []string {
 	classes := awsdefaults.InstanceClasses(region, arch)
 	types := make([]string, len(classes))
 	for i, c := range classes {
-		types[i] = fmt.Sprintf("%s.large", c)
+		types[i] = fmt.Sprintf("%s.xlarge", c)
 	}
 	return types
 }
@@ -341,10 +341,10 @@ func (w *Worker) Generate(dependencies asset.Parents) error {
 				}
 			}
 			if mpool.InstanceType == "" {
-				mpool.InstanceType, err = aws.PreferredInstanceType(ctx, installConfig.AWS, awsDefaultWorkerMachineTypes(installConfig.Config.Platform.AWS.Region, installConfig.Config.ControlPlane.Architecture), mpool.Zones)
+				mpool.InstanceType, err = aws.PreferredInstanceType(ctx, installConfig.AWS, awsDefaultMachineTypes(installConfig.Config.Platform.AWS.Region, installConfig.Config.ControlPlane.Architecture), mpool.Zones)
 				if err != nil {
 					logrus.Warn(errors.Wrap(err, "failed to find default instance type"))
-					mpool.InstanceType = awsDefaultWorkerMachineTypes(installConfig.Config.Platform.AWS.Region, installConfig.Config.ControlPlane.Architecture)[0]
+					mpool.InstanceType = awsDefaultMachineTypes(installConfig.Config.Platform.AWS.Region, installConfig.Config.ControlPlane.Architecture)[0]
 				}
 			}
 			// if the list of zones is the default we need to try to filter the list in case there are some zones where the instance might not be available
--- a/pkg/asset/machines/master.go
+++ b/pkg/asset/machines/master.go
@@ -54,7 +54,6 @@ import (
 	"github.com/openshift/installer/pkg/types"
 	alibabacloudtypes "github.com/openshift/installer/pkg/types/alibabacloud"
 	awstypes "github.com/openshift/installer/pkg/types/aws"
-	awsdefaults "github.com/openshift/installer/pkg/types/aws/defaults"
 	azuretypes "github.com/openshift/installer/pkg/types/azure"
 	azuredefaults "github.com/openshift/installer/pkg/types/azure/defaults"
 	baremetaltypes "github.com/openshift/installer/pkg/types/baremetal"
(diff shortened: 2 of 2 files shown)


pkg/asset/machines/worker.go
```diff
@@ -115,7 +115,7 @@ func defaultAzureMachinePoolPlatform() azuretypes.MachinePool {
 
 func defaultGCPMachinePoolPlatform() gcptypes.MachinePool {
 	return gcptypes.MachinePool{
-		InstanceType: "n1-standard-4",
+		InstanceType: "n2-standard-4",
```
patrickdillon, line 118: I don't think we need to change this one. `n1-standard-4` is close enough in terms of specs.

patrickdillon (commented):
Nice. So it looks like you identified aws, gcp, nutanix, and vsphere. gcp I think is close enough so let's leave it alone.

For vsphere and nutanix let's get owner opinions: @thunderboltsid, @rvanderp3, @jcpowermac. I suspect resources might be tighter on these platforms.

For context, it has been pointed out to us that aws has fewer resources on compute nodes that gcp/azure. While we're fixing aws we're also checking other platforms.

pkg/asset/machines/worker.go
rna-afk: Yeah I thought so too but wanted another opinion

barbacbd: I would agree with @patrickdillon the 1GB difference in memory is more than likely not noticed in a default configuration.

rna-afk: Spoke to Mak about this and he wanted this set to n2-standard-4.

barbacbd:
@rna-afk I saw that the aws default instance type was set to `gp3` but would other "generic" or general purpose instances make sense? I don't know if there are other conditions that need to be met.

rna-afk:
We need a lot of throughput for all our operations on the cluster. The general guideline is to provide a premium disk type for the control-plane nodes and any for the compute nodes but we set the default to premium for better performance. I think gp3 for the root volume type of the VMs is fine.

pkg/asset/machines/worker.go
```diff
@@ -176,20 +176,20 @@ func defaultPowerVSMachinePoolPlatform() powervstypes.MachinePool {
 
 func defaultNutanixMachinePoolPlatform() nutanixtypes.MachinePool {
 	return nutanixtypes.MachinePool{
-		NumCPUs:           2,
-		NumCoresPerSocket: 2,
-		MemoryMiB:         8192,
+		NumCPUs:           4,
+		NumCoresPerSocket: 4,
```
thunderboltsid, line 180: This implies 16 CPU units (CPUs x Cores per CPU). Is that what we want? I actually have an [open PR that was reducing these even further.](https://github.com/openshift/installer/pull/5857)

rna-afk: I just moved the code from [1] and the minimum requirement is 2 and 2 but the vsphere folks set it to 4 and 4 as part of the performance recommendation for vsphere. I'll remove the nutanix change though to let you change it to the ideal value, thanks!

thunderboltsid: @rna-afk I can also take care of bumping up the RAM to 16GB in my PR.

jcpowermac:
> For vsphere and nutanix let's get owner opinions: @thunderboltsid, @rvanderp3, @jcpowermac. I suspect resources might be tighter on these platforms.

In CI we have to up to this value anyway to pass jobs. So lgtm to me. We should update documentation right?

patrickdillon:
> We should update documentation right?

Yeah this is a good point. This *may* need more docs than the release note I originally mentioned. Although when I'm looking at the docs:
[AWS](https://docs.openshift.com/container-platform/4.9/installing/installing_aws/installing-aws-account.html#installation-aws-limits_installing-aws-account)
[GCP]https://docs.openshift.com/container-platform/4.9/installing/installing_gcp/installing-gcp-account.html#installation-gcp-limits_installing-gcp-account
Hm.. not finding the equivalent for vSphere?

It looks like AWS/GCP don't mention memory in those sections, so it doesn't look like they need updates. Not sure if anywhere else would need to be updated for AWS/GCP. @rna-afk can you  take a quick look?

patrickdillon:
/approve

Updating aws, gcp, & vsphere looks good to me.

rna-afk:
Almost all docs need changing specifically in the compute section where the minimum requirements are 2 VCPUs and 8 GB RAM. 
[AWS](https://docs.openshift.com/container-platform/4.10/installing/installing_aws/installing-aws-customizations.html#installation-minimum-resource-requirements_installing-aws-customizations)
[Azure](https://docs.openshift.com/container-platform/4.10/installing/installing_azure/installing-azure-customizations.html#installation-minimum-resource-requirements_installing-azure-customizations)

rna-afk:
> Almost all docs need changing specifically in the compute section where the minimum requirements are 2 VCPUs and 8 GB RAM.

This actually depends on if we want to increase our minimum requirements spec.

mjpytlak:
@patrickdillon @rna-afk Please let @bscott-rh and I know if doc updates beyond the release note are required and to which platforms. Based on comments, not sure which platforms this PR will ultimately land on.

sadasu:
@rna-afk it appears that we are changing the CPU and memory defaults just for AWS, GCP and vSphere. If that is the case, could you please commit message to reflect that (and not all platforms)? Also, https://github.com/openshift/installer/pull/5841#issue-1214820529 needs to updated to reflect that too. I'll defer to the other reviewers with more knowledge of the platforms to decide if the defaults are the right ones.

patrickdillon:
> > Almost all docs need changing specifically in the compute section where the minimum requirements are 2 VCPUs and 8 GB RAM.
> 
> This actually depends on if we want to increase our minimum requirements spec. Other than this I don't see any other docs change.

No, this wouldn't change minimum requirements.

patrickdillon:
/lgtm

@rna-afk would you add a note in the jira card for the docs team that says what the new default instance type is for each platform updated in this pr is?

pkg/asset/machines/worker.go
```diff
@@ -158,7 +158,7 @@ func defaultVSphereMachinePoolPlatform() vspheretypes.MachinePool {
 	return vspheretypes.MachinePool{
 		NumCPUs:           2,
 		NumCoresPerSocket: 2,
```
patrickdillon, line 160: Don't we need to bump this too?
/cc @jcpowermac

pkg/asset/machines/worker.go
```diff
@@ -156,9 +156,9 @@ func defaultOvirtMachinePoolPlatform() ovirttypes.MachinePool {
 
 func defaultVSphereMachinePoolPlatform() vspheretypes.MachinePool {
 	return vspheretypes.MachinePool{
-		NumCPUs:           2,
-		NumCoresPerSocket: 2,
-		MemoryMiB:         8192,
+		NumCPUs:           4,
+		NumCoresPerSocket: 1,
```
patrickdillon, line 160: Above it is 4 `NumCoresPerSocket` and here it is 1. Which is right?

rna-afk: 4 cpus and 1 cores per socket is right. I'll fix it

rna-afk: Changed it to set masters to 4 vcpus and 4 numcores and workers to 4 vcpus and 1 numcores

patrickdillon:
/lgtm

We could update https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/vsphere/ipi-conf-vsphere-commands.sh to no longer explicitly set the parameters that are being updated here

rvanderp3:
> /lgtm
> 
> We could update https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/vsphere/ipi-conf-vsphere-commands.sh to no longer explicitly set the parameters that are being updated here

@patrickdillon sure, good idea.  I'll follow up on that.

rvanderp3:
> > /lgtm
> > We could update https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/vsphere/ipi-conf-vsphere-commands.sh to no longer explicitly set the parameters that are being updated here
> 
> @patrickdillon sure, good idea. I'll follow up on that.

started a PR to address this: https://github.com/openshift/release/pull/29234

PR #5842, opened 2022-04-25
vsphere upi: missing etc resolv

jcpowermac:
Using rc-manager=unmanaged
for vsphere upi breaks the bootstrap
process since the RHCOS instance has
no `/etc/resolv.conf`


cybertron:
/lgtm

This logic matches what's used in the dispatcher script[0] so it should do what we want.

0: https://github.com/openshift/installer/blob/6194c1ccc459ac3de8c1cee8a4b811d6f3da81af/data/data/bootstrap/vsphere/files/etc/NetworkManager/dispatcher.d/30-local-dns-prepender.template#L6

deads2k:
UPI install worked.  per the bot it seems to only need approved labels.  Nothing has merged this was opened.  Merging

PR #5844, opened 2022-04-26
azure: move zones functionality to the client API

r4f4:
The zones.go file predates the existance of client.go. Most of the
functionality for retrieving resources through the Azure SDK is now
concentrated on the client file, except for Availability Zones. So let's
merge the zones.go code into client.go.

As an added benefit, we now have the ability to mock calls to fetch
the Availability Zones in the tests if needed.


pkg/asset/installconfig/azure/client.go
```diff
@@ -328,3 +329,28 @@ func (c *Client) AreMarketplaceImageTermsAccepted(ctx context.Context, publisher
...
+	filter := fmt.Sprintf("location eq '%s'", region)
+	for res, err := client.List(ctx, filter); res.NotDone(); err = res.NextWithContext(ctx) {
+		if err != nil {
+			return zones, err
+		}
+
+		for _, resSku := range res.Values() {
+			if strings.EqualFold(to.String(resSku.Name), instanceType) {
+				for _, locationInfo := range *resSku.LocationInfo {
+					if strings.EqualFold(to.String(locationInfo.Location), region) {
```
patrickdillon, line 347: just curious: now that we have the `filter`, is this still needed, or is this just defensive programming? I am fine with leaving it in, defensively (perhaps leave a comment), but it seems like it would not be needed?

pkg/asset/installconfig/azure/client.go
```diff
@@ -328,3 +329,28 @@ func (c *Client) AreMarketplaceImageTermsAccepted(ctx context.Context, publisher
...
+						return *locationInfo.Zones, nil
+					}
+				}
+			}
+		}
+	}
+
+	return zones, nil
```
patrickdillon, line 355: This follows the original code, so I think it is ok, but it strikes me as odd that this is not an error.

patrickdillon (commented):
Nice. This LGTM. A couple of questions but no changes needed that I see. 

/approve

pkg/asset/installconfig/azure/client.go
r4f4: I don't think it's needed any more. Maybe we can remove it and wait for a round of CI to check it doesn't impact anything?

pkg/asset/installconfig/azure/client.go
r4f4: For Azure Stack and Single VM regions that do not support availability zones, the Zones field seems to be just an empty slice. For example, `northcentralus` returns:
```
    "locationInfo": [
      {
        "location": "northcentralus",
        "zoneDetails": [],
        "zones": []
      }
    ],
```
So I agree we should probably return some kind of error at this point.

patrickdillon, line 355:
```suggestion
	return nil, errors.New(fmt.Sprintf("error retrieving availability zones for %s in %s", instanceType, region))
```

patrickdillon (commented):
@r4f4 would you mind making these changes as a new, separate commit (in case we need to revert for some reason)? One commit for both changes is fine.

r4f4:
Done

pkg/asset/installconfig/azure/client.go
```diff
@@ -344,13 +344,11 @@ func (c *Client) GetAvailabilityZones(ctx context.Context, instanceType string,
...
-						return *locationInfo.Zones, nil
-					}
+					return *locationInfo.Zones, nil
 				}
 			}
 		}
 	}
 
-	return zones, nil
+	return nil, fmt.Errorf("error retrieving availability zones for %s in %s", instanceType, region)
```
patrickdillon, line 353: this is better than my suggestion :)

patrickdillon:
LGTM. Let's just let CI do it's thing

r4f4:
Oops, somehow I had inverted `region` and `instanceType` arguments and deleted the setting of the authorizer. Those should be fixed now.

r4f4:
/hold
I forgot to regenerate the mock functions. I'll send a fix after the current test run finishes.

r4f4:
Mock regen is now done and I've squashed the fixup commits.

PR #5845, opened 2022-04-26
[wip] vSphere: Enable change block tracking

jcpowermac:
Based on upstream CSI issue:
https://github.com/kubernetes-sigs/vsphere-csi-driver/issues/1416

It looks like changed block tracking needs to be enabled.

### DIFF
--- a/data/data/vsphere/master/main.tf
+++ b/data/data/vsphere/master/main.tf
@@ -45,6 +45,7 @@ resource "vsphere_virtual_machine" "vm_master" {
     "guestinfo.ignition.config.data"          = base64encode(var.ignition_master)
     "guestinfo.ignition.config.data.encoding" = "base64"
     "guestinfo.hostname"                      = "${var.cluster_id}-master-${count.index}"
+    "ctkEnable"                               = "TRUE"
   }
 
   tags = var.tags


data/data/vsphere/master/main.tf
```diff
@@ -45,6 +45,7 @@ resource "vsphere_virtual_machine" "vm_master" {
     "guestinfo.ignition.config.data"          = base64encode(var.ignition_master)
     "guestinfo.ignition.config.data.encoding" = "base64"
     "guestinfo.hostname"                      = "${var.cluster_id}-master-${count.index}"
+    "ctkEnable"                               = "TRUE"
   }
```
gnufied, line 49: Does this needs to be continuously reconciled? It sounds like - in some cases of hard reboot/power failure, the setting is lost?

jcpowermac: @gnufied I think you are right based on the kb / issue:
- https://github.com/kubernetes-sigs/vsphere-csi-driver/issues/1416
- https://kb.vmware.com/s/article/1020128

Not sure where we could do that.

gnufied: We may have to start detecting CBT ourselves in vsphere-problem-detector and at least start creating alerts. I am not sure if we have to automatically fix that, but as long we can alert customers we should be fine.

jcpowermac:
/close

not sure we should be doing this since CBT would only be disabled if something not good happened.

PR #5846, opened 2022-04-26
Azure Provider: fix vmNetworkType defaults by checking vm capabilities

kwoodson:
Summary:
With the recent changes introduced to vmNetworkingType #5309, the default setting for was set to true (enable accelerated networking). This has broken a QE test for instance type  `Standard_B4ms`.

The idea behind this PR is to fetch the vm capabilities and then set the default to `vmNetworkingType` to whether or not the VM has the capability. This introduces a new azure client call `GetVMCapabilities`.

### DIFF
--- a/pkg/asset/installconfig/azure/client.go
+++ b/pkg/asset/installconfig/azure/client.go
@@ -35,6 +35,7 @@ type API interface {
 	GetHyperVGenerationVersion(ctx context.Context, instanceType string, diskType string, region string) (string, error)
 	GetMarketplaceImage(ctx context.Context, region, publisher, offer, sku, version string) (azsku.VirtualMachineImage, error)
 	AreMarketplaceImageTermsAccepted(ctx context.Context, publisher, offer, sku string) (bool, error)
+	GetVMCapabilities(ctx context.Context, instanceType, region string) (map[string]string, error)
 }
 
 // Client makes calls to the Azure API.
@@ -272,6 +273,24 @@ func (c *Client) GetDiskEncryptionSet(ctx context.Context, subscriptionID, group
 	}
 
 	return &diskEncryptionSet, nil
+}
+
+// GetVMCapabilities retrieves the capabilities of an instant type in a specific region. Returns these values
+// in a map with the capability name as the key and the corresponding value.
+func (c *Client) GetVMCapabilities(ctx context.Context, instanceType, region string) (map[string]string, error) {
+	typeMeta, err := c.GetVirtualMachineSku(ctx, instanceType, region)
+	if err != nil {
+		return nil, fmt.Errorf("error connecting to Azure client: %v", err)
+	}
+	if typeMeta == nil {
+		return nil, fmt.Errorf("not found in region %s", region)
+	}
+
+	capabilities := make(map[string]string)
+	for _, capability := range *typeMeta.Capabilities {
+		capabilities[to.String(capability.Name)] = to.String(capability.Value)
+	}
+	return capabilities, nil
 }
 
 // GetHyperVGenerationVersion gets the HyperVGeneration version for the given disk instance type. Defaults to V2 if either V1 or V2
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -88,59 +88,66 @@ func ValidateDiskEncryptionSet(client API, ic *types.InstallConfig) field.ErrorL
 }
 
 // ValidateInstanceType ensures the instance type has sufficient Vcpu and Memory.
-func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceType string, diskType string, req resourceRequirements, ultraSSDEnabled bool, vmNetworkingType string) field.ErrorList {
+func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceType string, diskType string, req resourceRequirements, ultraSSDEnabled bool, vmNetworkingType string, capabilities map[string]string) field.ErrorList {
 	allErrs := field.ErrorList{}
-
-	typeMeta, err := client.GetVirtualMachineSku(context.TODO(), instanceType, region)
-	if err != nil {
-		return append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, err.Error()))
+	val, ok := capabilities["vCPUsAvailable"]
+	if ok {
+		cpus, err := strconv.ParseFloat(val, 0)
+		if err != nil {
+			return append(allErrs, field.InternalError(fieldPath, err))
+		}
+		if cpus < float64(req.minimumVCpus) {
+			errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d vCPUsAvailable", req.minimumVCpus)
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+		}
+	} else {
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "capability not found: vCPUsAvailable"))
 	}
 
-	if typeMeta == nil {
-		errMsg := fmt.Sprintf("not found in region %s", region)
-		return append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+	val, ok = capabilities["MemoryGB"]
+	if ok {
+		memory, err := strconv.ParseFloat(val, 0)
+		if err != nil {
+			return append(allErrs, field.InternalError(fieldPath, err))
+		}
+		if memory < float64(req.minimumMemory) {
+			errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d GB Memory", req.minimumMemory)
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+		}
+	} else {
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "capability not found: MemoryGB"))
 	}
 
-	ultraSSDAvailable := false
-	for _, capability := range *typeMeta.Capabilities {
+	val, ok = capabilities["PremiumIO"]
+	if diskType == "Premium_LRS" && ok {
+		if strings.EqualFold(val, "False") {
+			errMsg := fmt.Sprintf("PremiumIO not supported for instance type %s", instanceType)
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("osDisk", "diskType"), diskType, errMsg))
+		}
+	} else {
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "capability not found: PremiumIO"))
+	}
 
-		if strings.EqualFold(*capability.Name, "vCPUsAvailable") {
-			cpus, err := strconv.ParseFloat(*capability.Value, 0)
-			if err != nil {
-				return append(allErrs, field.InternalError(fieldPath, err))
-			}
-			if cpus < float64(req.minimumVCpus) {
-				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d vCPUsAvailable", req.minimumVCpus)
-				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
-			}
-		} else if strings.EqualFold(*capability.Name, "MemoryGB") {
-			memory, err := strconv.ParseFloat(*capability.Value, 0)
-			if err != nil {
-				return append(allErrs, field.InternalError(fieldPath, err))
-			}
-			if memory < float64(req.minimumMemory) {
-				errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d GB Memory", req.minimumMemory)
-				allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
-			}
-		} else if diskType == "Premium_LRS" && strings.EqualFold(*capability.Name, "PremiumIO") {
-			if strings.EqualFold(*capability.Value, "False") {
-				errMsg := fmt.Sprintf("PremiumIO not supported for instance type %s", instanceType)
-				allErrs = append(allErrs, field.Invalid(fieldPath.Child("osDisk", "diskType"), diskType, errMsg))
-			}
-		} else if strings.EqualFold(*capability.Name, "UltraSSDAvailable") {
-			ultraSSDAvailable = strings.EqualFold(*capability.Value, "True")
-		} else if strings.EqualFold(*capability.Name, string(aztypes.VMnetworkingTypeAccelerated)) {
-			if vmNetworkingType == string(aztypes.VMnetworkingTypeAccelerated) && !strings.EqualFold(*capability.Value, "True") {
+	if vmNetworkingType == string(aztypes.VMnetworkingTypeAccelerated) {
+		val, ok = capabilities[string(aztypes.AcceleratedNetworkingEnabled)]
+		if ok {
+			if !strings.EqualFold(val, "True") {
 				errMsg := fmt.Sprintf("vm networking type is not supported for instance type %s", instanceType)
 				allErrs = append(allErrs, field.Invalid(fieldPath.Child("vmNetworkingType"), vmNetworkingType, errMsg))
 			}
+		} else {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "capability not found: AcceleratedNetworkingEnabled"))
 		}
 	}
 
 	// The UltraSSDAvailable capability might not be present at all, in which case it must assumed to be false
-	if ultraSSDEnabled && !ultraSSDAvailable {
-		errMsg := fmt.Sprintf("UltraSSD capability not supported for this instance type in the %s region", region)
-		allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+	val, ok = capabilities["UltraSSDAvailable"]
+	if ok {
+		ultraSSDAvailable := strings.EqualFold(val, "True")
+		if ultraSSDEnabled && !ultraSSDAvailable {
+			errMsg := fmt.Sprintf("UltraSSD capability not supported for this instance type in the %s region", region)
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+		}
 	}
 
 	return allErrs
@@ -153,7 +160,9 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList 
 	defaultDiskType := aztypes.DefaultDiskType
 	defaultInstanceType := ""
 	defaultUltraSSDCapability := "Disabled"
-	defaultVMNetworkingType := string(aztypes.VMnetworkingTypeAccelerated)
+	defaultVMNetworkingType := ""
+	var capabilities map[string]string
+	var err error
 
 	if ic.Platform.Azure.DefaultMachinePlatform != nil {
 		if ic.Platform.Azure.DefaultMachinePlatform.OSDisk.DiskType != "" {
@@ -171,6 +180,7 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList 
 	}
 
 	if ic.ControlPlane != nil && ic.ControlPlane.Platform.Azure != nil {
+		fieldPath := field.NewPath("controlPlane", "platform", "azure")
 		diskType := ic.ControlPlane.Platform.Azure.OSDisk.DiskType
 		instanceType := ic.ControlPlane.Platform.Azure.InstanceType
 		ultraSSDCapability := ic.ControlPlane.Platform.Azure.UltraSSDCapability
@@ -188,11 +198,18 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList 
 		if ultraSSDCapability == "" {
 			ultraSSDCapability = defaultUltraSSDCapability
 		}
+		capabilities, err = client.GetVMCapabilities(context.TODO(), instanceType, ic.Azure.Region)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, err.Error()))
+		}
 		if vmNetworkingType == "" {
 			vmNetworkingType = defaultVMNetworkingType
 		}
 		ultraSSDEnabled := strings.EqualFold(ultraSSDCapability, "Enabled")
-		allErrs = append(allErrs, ValidateInstanceType(client, field.NewPath("controlPlane", "platform", "azure"), ic.Azure.Region, instanceType, diskType, controlPlaneReq, ultraSSDEnabled, vmNetworkingType)...)
+		// if capabilities fails to find a valid instance type, we should skip this validation
+		if capabilities != nil {
+			allErrs = append(allErrs, ValidateInstanceType(client, fieldPath, ic.Azure.Region, instanceType, diskType, controlPlaneReq, ultraSSDEnabled, vmNetworkingType, capabilities)...)
+		}
 	}
 
 	for idx, compute := range ic.Compute {
@@ -218,9 +235,16 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList 
 			if vmNetworkingType == "" {
 				vmNetworkingType = defaultVMNetworkingType
 			}
+			capabilities, err = client.GetVMCapabilities(context.TODO(), instanceType, ic.Azure.Region)
+			if err != nil {
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("platform", "azure", "type"), instanceType, err.Error()))
+			}
 			ultraSSDEnabled := strings.EqualFold(ultraSSDCapability, "Enabled")
-			allErrs = append(allErrs, ValidateInstanceType(client, fieldPath.Child("platform", "azure"),
-				ic.Azure.Region, instanceType, diskType, computeReq, ultraSSDEnabled, vmNetworkingType)...)
+			// skip validation if instance type was invalid
+			if capabilities != nil {
+				allErrs = append(allErrs, ValidateInstanceType(client, fieldPath.Child("platform", "azure"),
+					ic.Azure.Region, instanceType, diskType, computeReq, ultraSSDEnabled, vmNetworkingType, capabilities)...)
+			}
 		}
 	}
 
--- a/pkg/asset/installconfig/azure/validation_test.go
+++ b/pkg/asset/installconfig/azure/validation_test.go
@@ -44,14 +44,23 @@ var (
 		{Name: to.StringPtr("Standard_D4_v4"), Capabilities: &[]azsku.ResourceSkuCapabilities{{Name: to.StringPtr("vCPUsAvailable"), Value: to.StringPtr("4")}, {Name: to.StringPtr("MemoryGB"), Value: to.StringPtr("16")}, {Name: to.StringPtr("PremiumIO"), Value: to.StringPtr("True")}, {Name: to.StringPtr("HyperVGenerations"), Value: to.StringPtr("V1,V2")}, {Name: to.StringPtr("AcceleratedNetworkingEnabled"), Value: to.StringPtr("True")}}},
 		{Name: to.StringPtr("Standard_D2s_v3"), Capabilities: &[]azsku.ResourceSkuCapabilities{{Name: to.StringPtr("vCPUsAvailable"), Value: to.StringPtr("4")}, {Name: to.StringPtr("MemoryGB"), Value: to.StringPtr("16")}, {Name: to.StringPtr("PremiumIO"), Value: to.StringPtr("True")}, {Name: to.StringPtr("HyperVGenerations"), Value: to.StringPtr("V1,V2")}, {Name: to.StringPtr("AcceleratedNetworkingEnabled"), Value: to.StringPtr("True")}}},
 		{Name: to.StringPtr("Standard_D8s_v3"), Capabilities: &[]azsku.ResourceSkuCapabilities{{Name: to.StringPtr("vCPUsAvailable"), Value: to.StringPtr("4")}, {Name: to.StringPtr("MemoryGB"), Value: to.StringPtr("16")}, {Name: to.StringPtr("PremiumIO"), Value: to.StringPtr("True")}, {Name: to.StringPtr("HyperVGenerations"), Value: to.StringPtr("V1,V2")}, {Name: to.StringPtr("UltraSSDAvailable"), Value: to.StringPtr("True")}, {Name: to.StringPtr("AcceleratedNetworkingEnabled"), Value: to.StringPtr("True")}}},
-		{Name: to.StringPtr("Standard_D_v4"), Capabilities: &[]azsku.ResourceSkuCapabilities{{Name: to.StringPtr("vCPUsAvailable"), Value: to.StringPtr("4")}, {Name: to.StringPtr("MemoryGB"), Value: to.StringPtr("16")}, {Name: to.StringPtr("PremiumIO"), Value: to.StringPtr("False")}, {Name: to.StringPtr("HyperVGenerations"), Value: to.StringPtr("V1,V2")}}},
 		{Name: to.StringPtr("Standard_Dc4_v4"), Capabilities: &[]azsku.ResourceSkuCapabilities{{Name: to.StringPtr("vCPUsAvailable"), Value: to.StringPtr("4")}, {Name: to.StringPtr("MemoryGB"), Value: to.StringPtr("16")}, {Name: to.StringPtr("PremiumIO"), Value: to.StringPtr("True")}, {Name: to.StringPtr("HyperVGenerations"), Value: to.StringPtr("V2")}}},
+	}
+	vmCapabilities = map[string]map[string]string{
+		"Standard_D8s_v3": {"vCPUsAvailable": "4", "MemoryGB": "16", "PremiumIO": "True", "HyperVGenerations": "V1,V2", "AcceleratedNetworkingEnabled": "True", "UltraSSDAvailable": "True"},
+		"Standard_D4s_v3": {"vCPUsAvailable": "4", "MemoryGB": "32", "PremiumIO": "True", "HyperVGenerations": "V1", "AcceleratedNetworkingEnabled": "True", "UltraSSDAvailable": "True"},
+		"Standard_A1_v2":  {"vCPUsAvailable": "1", "MemoryGB": "2", "PremiumIO": "True", "HyperVGenerations": "V1,V2", "AcceleratedNetworkingEnabled": "False"},
+		"Standard_D2_v4":  {"vCPUsAvailable": "2", "MemoryGB": "8", "PremiumIO": "True", "HyperVGenerations": "V1,V2", "AcceleratedNetworkingEnabled": "True"},
+		"Standard_D4_v4":  {"vCPUsAvailable": "4", "MemoryGB": "16", "PremiumIO": "False", "HyperVGenerations": "V1,V2", "AcceleratedNetworkingEnabled": "True"},
+		"Standard_D2s_v3": {"vCPUsAvailable": "4", "MemoryGB": "16", "PremiumIO": "True", "HyperVGenerations": "V1,V2", "AcceleratedNetworkingEnabled": "True"},
+		"Standard_Dc4_v4": {"vCPUsAvailable": "4", "MemoryGB": "16", "PremiumIO": "True", "HyperVGenerations": "V2"},
+		"Standard_B4ms":   {"vCPUsAvailable": "4", "MemoryGB": "16", "PremiumIO": "True", "HyperVGenerations": "V1,V2", "AcceleratedNetworkingEnabled": "False", "UltraSSDAvailable": "True"},
 	}
 
 	validInstanceTypes = func(ic *types.InstallConfig) {
-		ic.Platform.Azure.DefaultMachinePlatform.InstanceType = "Standard_D2_v4"
-		ic.ControlPlane.Platform.Azure.InstanceType = "Standard_D4_v4"
-		ic.Compute[0].Platform.Azure.InstanceType = "Standard_D2_v4"
+		ic.Platform.Azure.DefaultMachinePlatform.InstanceType = "Standard_D4s_v3"
+		ic.ControlPlane.Platform.Azure.InstanceType = "Standard_D8s_v3"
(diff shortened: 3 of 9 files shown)


pkg/asset/installconfig/azure/client.go
```diff
@@ -274,9 +275,21 @@ func (c *Client) GetDiskEncryptionSet(ctx context.Context, subscriptionID, group
 	return &diskEncryptionSet, nil
 }
 
+func (c *Client) GetVMCapabilities(ctx context.Context, instanceType, region string) (map[string]string, error) {
+	typeMeta, err := c.GetVirtualMachineSku(ctx, instanceType, region)
+	if err != nil {
+		return nil, fmt.Errorf("error onnecting to Azure client: %s", err.Error())
```
r4f4, line 281: small nit
```suggestion
		return nil, fmt.Errorf("error connecting to Azure client: %s", err.Error())
```

pkg/asset/installconfig/azure/validation.go
```diff
@@ -88,59 +89,63 @@ func ValidateDiskEncryptionSet(client API, ic *types.InstallConfig) field.ErrorL
...
+			errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d vCPUsAvailable", req.minimumVCpus)
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+		}
+	} else {
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "capability not found: vCPUsAvailable"))
 	}
 
-	if typeMeta == nil {
-		errMsg := fmt.Sprintf("not found in region %s", region)
-		return append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
```
r4f4, line 101: This is going to break the tests which are expecting this error message for invalid instance types. The way things are now, this should be moved to `GetVMCapabilities`

pkg/asset/installconfig/azure/client.go
```diff
@@ -274,9 +275,21 @@ func (c *Client) GetDiskEncryptionSet(ctx context.Context, subscriptionID, group
...
+	}
```
r4f4, line 282: You need to check that `typeMeta` is a valid pointer since the passed `instanceType` might not be available in the specified region.

pkg/asset/installconfig/azure/validation.go
kwoodson: Should I also move the fieldPath.Child? That seemed out-of-place.

pkg/asset/installconfig/azure/validation_test.go
```diff
@@ -430,6 +440,11 @@ func TestAzureInstallConfigValidation(t *testing.T) {
 	}
 	azureClient.EXPECT().GetVirtualMachineSku(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil).AnyTimes()
 
+	for key, value := range vmCapabilities {
+		azureClient.EXPECT().GetVMCapabilities(gomock.Any(), key, validRegion).Return(value, nil).AnyTimes()
+	}
+	azureClient.EXPECT().GetVMCapabilities(gomock.Any(), "Dne_D2_v4", validRegion).Return(nil, nil).AnyTimes()
```
r4f4, line 446: Inside `GetVMCapabilities` you return an error if the `instanceType` is `nil`. So, when mocking, you need to do the same thing here because the test is expecting to match the error message.
```suggestion
	azureClient.EXPECT().GetVMCapabilities(gomock.Any(), "Dne_D2_v4", validRegion).Return(nil, fmt.Errorf("not found in region %s", validRegion).AnyTimes()
```

pkg/asset/installconfig/azure/validation.go
```diff
@@ -216,11 +234,20 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
 				ultraSSDCapability = defaultUltraSSDCapability
 			}
 			if vmNetworkingType == "" {
-				vmNetworkingType = defaultVMNetworkingType
+				capabilities, _ = client.GetVMCapabilities(context.TODO(), instanceType, ic.Azure.Region)
```
r4f4, line 237: This is not correct. You need to fetch the capabilities whether `vmNetworkingType` was supplied or not. And you definitely need to check for the `err` value as you did in the controlPlane section. The `instanceType` for the compute nodes might be different from the one for the controlPlane nodes, so you effectively have 2 sets of capabilities.
```
capabilities, err = client.GetVMCapabilities(context.TODO(), instanceType, ic.Azure.Region)
if err != nil {
	fp := field.NewPath("compute", "platform", "azure")
	return append(allErrs, field.Invalid(fp.Child("type"), instanceType, err.Error()))
}
```

pkg/asset/installconfig/azure/validation.go
```diff
@@ -188,11 +191,26 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
 		if ultraSSDCapability == "" {
 			ultraSSDCapability = defaultUltraSSDCapability
 		}
+
+		capabilities, err = client.GetVMCapabilities(context.TODO(), instanceType, ic.Azure.Region)
+		if err != nil {
+			fp := field.NewPath("controlPlane", "platform", "azure")
+			return append(allErrs, field.Invalid(fp.Child("type"), "vmNetworkingCapability", "unable to fetch vm capabilities"))
```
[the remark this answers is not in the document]
r4f4: Also this is a change from the previous behaviour in which the compute nodes where also validated before all the errors where returned. Here, you're short-circuiting the validation on the control plane. If that's desirable, you'll have to adjust some of the error messages matching in the tests.

pkg/asset/installconfig/azure/validation_test.go
r4f4: Another point regarding this is, since your `GetVMCapabilities` is now used as `GetVirtualMachineSku`, you need to make sure that all combinations of `instanceType` and `region` expected by the tests are properly mocked.

pkg/asset/installconfig/azure/validation.go
kwoodson: @r4f4 I agree. This is why (from our slack discussion) I said that I made an attempt at this but didn't like how it was turning out. This is due to the previous comment I discussed above about setting default variables in the middle of the validation functions. The issue is when a user doesn't define the control plane or the compute or any combination of these we will get different combination of validation or none at all.

The getVmCapabilities should occur as one of the first things that happens when doing the validation if the compute || the control plane objects aren't nil.

pkg/asset/installconfig/azure/client.go
```diff
@@ -32,9 +32,10 @@ type API interface {
 	ListResourceIDsByGroup(ctx context.Context, groupName string) ([]string, error)
 	GetStorageEndpointSuffix(ctx context.Context) (string, error)
 	GetDiskEncryptionSet(ctx context.Context, subscriptionID, groupName string, diskEncryptionSetName string) (*azenc.DiskEncryptionSet, error)
-	GetHyperVGenerationVersion(ctx context.Context, instanceType string, diskType string, region string) (string, error)
+	GetHyperVGenerationVersion(ctx context.Context, instanceType string, region string) (string, error)
```
patrickdillon, line 35: We do not necessarily need to keep the `GetHyperVGenerationVersion` as `GetVMCapabilities` is a more generalized solution that could replace it.

pkg/asset/installconfig/azure/client.go
```diff
@@ -274,12 +275,28 @@ func (c *Client) GetDiskEncryptionSet(ctx context.Context, subscriptionID, group
 	return &diskEncryptionSet, nil
 }
 
+func (c *Client) GetVMCapabilities(ctx context.Context, instanceType, region string) (map[string]string, error) {
```
patrickdillon, line 278: Can all of this be replaced by the `getHyperVGenerationVersion` function you created? I don't see that function being used anywhere.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -188,11 +192,24 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
 		if vmNetworkingType == "" {
-			vmNetworkingType = defaultVMNetworkingType
+			val, ok := capabilities[string(azure.AcceleratedNetworkingEnabled)]
+			if ok {
+				if strings.EqualFold(val, "True") {
+					vmNetworkingType = string(azure.VMnetworkingTypeAccelerated)
+				} else {
+					vmNetworkingType = string(azure.VMNetworkingTypeBasic)
+				}
+			}
```
patrickdillon, line 209: I'm not aware of a need to set the default here when validating. To me the approach that makes sense is:
- if a user sets the value, validate it is possible
- if not, skip that validation, set the default value that works in the machines

pkg/asset/installconfig/azure/validation.go
```diff
@@ -215,12 +232,24 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+				val, ok := capabilities[string(azure.AcceleratedNetworkingEnabled)]
+				if ok {
+					if strings.EqualFold(val, "True") {
+						vmNetworkingType = string(azure.VMnetworkingTypeAccelerated)
+					} else {
+						vmNetworkingType = string(azure.VMNetworkingTypeBasic)
+					}
+				}
 			}
+			ic.Compute[idx].Platform.Azure.VMNetworkingType = vmNetworkingType
```
patrickdillon, line 249: Same comment as above, I don't think we need to set defaults in the validation for this.

pkg/asset/installconfig/azure/validation_test.go
```diff
@@ -430,6 +440,18 @@ func TestAzureInstallConfigValidation(t *testing.T) {
...
+	azureClient.EXPECT().GetVMCapabilities(gomock.Any(), gomock.Any(), "neverland").Return(nil, nil).AnyTimes()
+	azureClient.EXPECT().GetVMCapabilities(gomock.Any(), gomock.Any(), "australiacentral2").Return(vmCapabilities["Standard_D8s_v3"], nil).AnyTimes()
+	azureClient.EXPECT().GetVMCapabilities(gomock.Any(), gomock.Any(), "centralus").Return(vmCapabilities["Standard_D8s_v3"], nil).AnyTimes()
+	azureClient.EXPECT().GetVMCapabilities(gomock.Any(), gomock.Any(), "Central US").Return(nil, nil).AnyTimes()
```
patrickdillon, line 450: I suspect we could reduce the cardinality of these capability region combinations. Not sure if the code allows this ATM, but we really shouldn't need different regions in these calls... And it seems like we should potentially be able to get away with two different instance types: one good, one bad.

pkg/asset/machines/azure/machines.go
```diff
@@ -78,12 +79,26 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
...
+	}
+
+	if mpool.VMNetworkingType == "" {
+		acceleratedNetworking, err := getVMNetworkingCapability(capabilities)
+		if err != nil {
+			return nil, err
+		}
+		if acceleratedNetworking {
+			mpool.VMNetworkingType = string(azure.VMnetworkingTypeAccelerated)
+		}
```
patrickdillon, line 100: Let's do an `else` here with a logrus `INFO` that the selected instance type does not support accelerated networking.

patrickdillon (commented):
It looks like the validation work is a bit of a bear. I'm happy to work through it now (we definitely need to prioritize this) but if it's possible to introduce the main functionality of pulling capabailities and dynamically setting the defaults, I am happy to not have the validation refactoring be a requirement of this work.

If you want to keep on it, can we break it into a separate commit?

pkg/asset/installconfig/azure/client.go
kwoodson: I could not come up with a clean way of removing this `GetHyperVGenerationVersion` as it is called in a couple of different places. In machine, the variable is used to generate a name for the the image (https://github.com/openshift/installer/blob/master/pkg/asset/machines/azure/machines.go#L98-L99). Then later on it is used again to pass data into the terraform (https://github.com/openshift/installer/blob/master/pkg/asset/cluster/tfvars.go#L331-L332). 

Maybe if the terraform (tfvars.go) bits could look at the image value and determine what the proper value is for hypervgen?

pkg/asset/installconfig/azure/validation.go
kwoodson: I agree, I will not set the default. The only thing that is redundant is when the value is not set we are setting it to a default value based upon the capabilities and then validating the default that we just set.

patrickdillon: > The only thing that is redundant is when the value is not set we are setting it to a default value based upon the capabilities and then validating the default that we just set.

Maybe there's a disconnect or I'm missing something. In what I'm suggesting I don't think we would be: " validating the default that we just set."

My suggestion is that we:
- validate the user provider install config (if accelerated networking is set, validate that too, if not skip it)
- set a default, if needed, for the machinesets

I don't think those machinesets would be validated.

pkg/asset/installconfig/azure/client.go
patrickdillon: > I could not come up with a clean way of removing this `GetHyperVGenerationVersion` as it is called in a couple of different places. 

That's fine. It's probably best to try to keep this all as concentrated as possible.

> Maybe if the terraform (tfvars.go) bits could look at the image value and determine what the proper value is for hypervgen?

We went back and forth on this in the original PR and ultimately ended up with the current approach.

kwoodson:
Cleaned up the tests and was able to create accelerated networking hosts when not specifying the setting in the install config. 

Cleaned up the validation tests as well.
```
{
  "name": "accnet-v5l4p-master-0-nic",
  "acceleratedNetworking": true
}
{
  "name": "accnet-v5l4p-master-1-nic",
  "acceleratedNetworking": true
}
{
  "name": "accnet-v5l4p-master-2-nic",
  "acceleratedNetworking": true
}
{
  "name": "accnet-v5l4p-worker-centralus1-jfsjd-nic",
  "acceleratedNetworking": true
}
{
  "name": "accnet-v5l4p-worker-centralus2-zkpvt-nic",
  "acceleratedNetworking": true
}
{
  "name": "accnet-v5l4p-worker-centralus3-wpch8-nic",
  "acceleratedNetworking": true
}

```

kwoodson:
The e2e-azure tests are failing for different reasons.
```
error while getting build logs, could not retrieve build pod ci-op-idjz0v9n/installer-build : pod "installer-build" not foundWARN[2022-05-03T19:08:08Z] Failed to get build baremetal-installer.      error=builds.build.openshift.io "baremetal-installer" not found
43
```

kwoodson:
@patrickdillon @r4f4 I have updated. PTAL and let me know. I tested these changes and things look good.

MayXuQQ:
when use 'Standard_B4ms' in install-config.yaml 
ERROR failed to fetch Metadata: failed to load asset "Install Config": failed to create install config: controlPlane.platform.azure.vmNetworkingType: Invalid value: "Accelerated": vm networking type is not supported for instance type Standard_B4ms 
that is expected. thanks

pkg/asset/installconfig/azure/validation.go
```diff
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"github.com/openshift/installer/pkg/types"
+	"github.com/openshift/installer/pkg/types/azure"
```
r4f4, line 19: This package is already import in the following line as `aztypes`

pkg/asset/cluster/tfvars.go
```diff
@@ -327,9 +327,9 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		for i, w := range workers {
 			workerConfigs[i] = w.Spec.Template.Spec.ProviderSpec.Value.Object.(*machinev1beta1.AzureMachineProviderSpec)
 		}
+
 		client := aztypes.NewClient(session)
-		hyperVGeneration, err := client.GetHyperVGenerationVersion(context.TODO(), masterConfigs[0].VMSize,
-			masterConfigs[0].OSDisk.ManagedDisk.StorageAccountType, masterConfigs[0].Location)
+		hyperVGeneration, err := client.GetHyperVGenerationVersion(context.TODO(), masterConfigs[0].VMSize, masterConfigs[0].Location)
```
r4f4, line 332: This change now looks out of place in this PR. You're dropping one of the arguments from `GetHyperVGenerationVersion` but how is that related to the `vmNetworkType`?

pkg/asset/installconfig/azure/client.go
r4f4: Related to my previous comment, I think you can just drop this change from this PR now. Feel free to do it in a follow-up PR.

pkg/asset/installconfig/azure/client.go
```diff
@@ -274,12 +275,30 @@ func (c *Client) GetDiskEncryptionSet(ctx context.Context, subscriptionID, group
 	return &diskEncryptionSet, nil
 }
 
+// GetVMCapabilities retrieves the capabilities of an instant type in a specific region. Returns these values
+// in a map with the capability name as the key and the corresponding value.
+func (c *Client) GetVMCapabilities(ctx context.Context, instanceType, region string) (map[string]string, error) {
+	typeMeta, err := c.GetVirtualMachineSku(ctx, instanceType, region)
+	if err != nil {
+		return nil, fmt.Errorf("error connecting to Azure client: %s", err.Error())
```
r4f4, line 283: I've also seen the formatting being done as 
```suggestion
		return nil, fmt.Errorf("error connecting to Azure client: %v", err)
```
Not sure which one is more desirable

pkg/asset/installconfig/azure/validation.go
```diff
@@ -188,11 +192,27 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+				} else {
+					vmNetworkingType = string(azure.VMNetworkingTypeBasic)
+				}
+			}
 		}
 		ultraSSDEnabled := strings.EqualFold(ultraSSDCapability, "Enabled")
-		allErrs = append(allErrs, ValidateInstanceType(client, field.NewPath("controlPlane", "platform", "azure"), ic.Azure.Region, instanceType, diskType, controlPlaneReq, ultraSSDEnabled, vmNetworkingType)...)
+		// if capabilities fails to find a valid instance type, we should skip this validation
+		if capabilities != nil {
+			allErrs = append(allErrs, ValidateInstanceType(client, field.NewPath("controlPlane", "platform", "azure"), ic.Azure.Region, instanceType, diskType, controlPlaneReq, ultraSSDEnabled, vmNetworkingType, capabilities)...)
```
r4f4, line 214: Why not pass in the `fieldPath` var you create above?
```suggestion
			allErrs = append(allErrs, ValidateInstanceType(client, fieldPath, ic.Azure.Region, instanceType, diskType, controlPlaneReq, ultraSSDEnabled, vmNetworkingType, capabilities)...)
```

pkg/asset/installconfig/azure/validation.go
```diff
@@ -188,11 +192,27 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
 		}
+
+		capabilities, err = client.GetVMCapabilities(context.TODO(), instanceType, ic.Azure.Region)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, err.Error()))
+		}
+
 		if vmNetworkingType == "" {
-			vmNetworkingType = defaultVMNetworkingType
+			val, ok := capabilities[string(azure.AcceleratedNetworkingEnabled)]
```
r4f4, line 202: If there was an error, `capabilities` is going to be `nil` but you're still trying to access it here as a map. Down below you correctly skip instance validation if it's `nil`.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -215,12 +235,26 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
 			if ultraSSDCapability == "" {
 				ultraSSDCapability = defaultUltraSSDCapability
 			}
+			capabilities, err = client.GetVMCapabilities(context.TODO(), instanceType, ic.Azure.Region)
+			if err != nil {
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("platform", "azure", "type"), instanceType, err.Error()))
+			}
 			if vmNetworkingType == "" {
-				vmNetworkingType = defaultVMNetworkingType
+				val, ok := capabilities[string(azure.AcceleratedNetworkingEnabled)]
```
r4f4, line 243: Same problem as in the controlPlane: capabilities can be `nil`.

r4f4 (changes requested):
Looks good. There is only one issue with possible `nil` capabilities that I've commented inline and some nitpicking about the HyperV function arguments change.

pkg/asset/cluster/tfvars.go
kwoodson: This is something I realized as I was working through the code. There are arguments passed that are not required. I was doing some clean up as I went through. I can revert this change and leave the cruft for someone else.

pkg/asset/installconfig/azure/client.go
```diff
@@ -274,12 +275,30 @@ func (c *Client) GetDiskEncryptionSet(ctx context.Context, subscriptionID, group
...
+}
+
 // GetHyperVGenerationVersion gets the HyperVGeneration version for the given disk instance type. Defaults to V2 if either V1 or V2
 // available.
-func (c *Client) GetHyperVGenerationVersion(ctx context.Context, instanceType string, diskType string, region string) (version string, err error) {
+func (c *Client) GetHyperVGenerationVersion(ctx context.Context, instanceType string, region string) (version string, err error) {
 	typeMeta, err := c.GetVirtualMachineSku(ctx, instanceType, region)
 	if err != nil {
-		return "", fmt.Errorf("error onnecting to Azure client: %s", err.Error())
+		return "", fmt.Errorf("error connecting to Azure client: %s", err.Error())
```
kwoodson, line 301: reverting this change as well as others to the GetHyperVGenerationVersion.

pkg/asset/installconfig/azure/validation.go
kwoodson: Thanks, I'll add a check to guard against this.

pkg/asset/installconfig/azure/validation.go
kwoodson: Fixed! thanks

pkg/asset/cluster/tfvars.go
r4f4: You can at least put it in its own commit so the reason for the change is clear for someone looking at the commit history down the line.

MayXuQQ:
based on the pre-merge  (354a00) test, have not set the vmNetworkingType in install-config.yaml, 
when use the vm type is 'type: Standard_B4ms'  the enableAcceleratedNetworking is false,
if not set or set to other vm type , the enableAcceleratedNetworking is true. 
that is expected.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -88,59 +88,63 @@ func ValidateDiskEncryptionSet(client API, ic *types.InstallConfig) field.ErrorL
...
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "capability not found: PremiumIO"))
+	}
+	val, ok = capabilities[string(aztypes.AcceleratedNetworkingEnabled)]
+	if ok {
+		if vmNetworkingType == string(aztypes.VMnetworkingTypeAccelerated) && !strings.EqualFold(val, "True") {
+			errMsg := fmt.Sprintf("vm networking type is not supported for instance type %s", instanceType)
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("vmNetworkingType"), vmNetworkingType, errMsg))
 		}
+	} else {
+		allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "capability not found: AcceleratedNetworkingEnabled"))
```
r4f4, line 137: The docs say "Accelerated networking is available in all global Azure regions and Azure Government Cloud" [1]. So I'm wondering: could this capability not exist in the Azure Stack case? If so, do we want to error out here?

[1] https://docs.microsoft.com/en-us/azure/virtual-network/accelerated-networking-overview#regions

r4f4:
Don't forget to regenerate the mock functions to drop the HyperV changes introduced earlier.

r4f4:
As I suspected, Azure Stack deployment is now failing `level=fatal msg=failed to fetch Master Machines: failed to generate asset "Master Machines": failed to create master machine objects: failed to create provider: unable to determine vm network type capability`

kwoodson:
@r4f4 @patrickdillon What would the expectation be here? Should I catch the error and set the networking to false if the vm capability is missing? That seems like a logical progression for setting defaults.

r4f4:
What about just validating the value if `vmNetworkType` it's set to `Accelerated`? Otherwise you don't even need to look at the capability. Or am I missing something?

kwoodson:
@r4f4 If the value is set to `""` then we need to see if the capability exists on the vm and then default to the capability to whether it exists or not.  If we only check when it is set, then we will not enable the feature when it is not set.

I suggest not checking the error when it is returned which will default to setting the capability to false when it is not found.

r4f4:
> @r4f4 If the value is set to `""` then we need to see if the capability exists on the vm and then default to the capability to whether it exists or not. If we only check when it is set, then we will not enable the feature when it is not set.

You already do that correctly here https://github.com/openshift/installer/pull/5846/files#diff-75606c90c997f4218689ce173a97316d61be39a8df4f5bebe08a4cd0e9f24973R200-R208. In case the option was not set in the install-config and the capability does not exist, you go into `ValidateInstanceType` with `vmNetworkType == ""` and it's here that I suggested you validate only if `vmNetworkType == Accelerated`. If it's `""` it means the cap does not exist and if it's `Basic` it doesn't matter the value of the capability.

In code:
```
func validateInstanceTypes(...) {
        [...]
	if vmNetworkingType == "" && capabilities != nil {
		val, ok := capabilities[string(aztypes.AcceleratedNetworkingEnabled)]
		if ok {
			if strings.EqualFold(val, "True") {
				vmNetworkingType = string(aztypes.VMnetworkingTypeAccelerated)
			} else {
				vmNetworkingType = string(aztypes.VMNetworkingTypeBasic)
			}
		}
         }
         allErrs = ValidateInstanceType(...)
         [...]
}

func ValidateInstanceType(...) {
    [...]
    if vmNetworkingType == aztypes.VMnetworkingTypeAccelerated {
        val, ok := capabilities[string(aztypes.AcceleratedNetworkingEnabled)]; 
        if ok && !strings.EqualFold(val, "True") {
		errMsg := fmt.Sprintf("vm networking type is not supported for instance type %s", instanceType)
		allErrs = append(allErrs, field.Invalid(fieldPath.Child("vmNetworkingType"), vmNetworkingType, errMsg))
	} else {
		allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "capability not found: AcceleratedNetworkingEnabled"))
        }
    }
    [...]
}
```
> I suggest not checking the error when it is returned which will default to setting the capability to false when it is not found.

If you do that in `ValidateInstanceType`, aren't you possibly failing to validate the case when the `Accelerated` type was set in the install-config but the capability does not exist for the selected InstanceType?
Or do you mean to do that in `validateInstanceTypes`, therefore making sure `vmNetworkingType` is either `Accelerated` or `Basic` before going into `ValidateInstanceType`?

kwoodson:
@r4f4 The code referenced in both sections is different.

This is the validation and behaves differently than the machine defaults:
https://github.com/openshift/installer/pull/5846/files#diff-75606c90c997f4218689ce173a97316d61be39a8df4f5bebe08a4cd0e9f24973R200-R208

The reason for this is that the section for Azure can be nil and therefore we do not have capabilities for the VM. In this instance we skip the validation entirely.

In the machines.go code we call the capabilities check here:
https://github.com/openshift/installer/pull/5846/files#diff-54168990f7f297f463e06a145f714bd63582499b41b4be695f49f5f16dc6f76eR95-R98

In the Azurestack, the capabilities does not include the accelerated networking which is causing this to fail. We could assume that when the capability is not in the capabilities map that the instance does not support accelerated networking. We could just skip the error checking at that moment and set the value to False. 

If we only check the value when vmNetworkingType is set, then we will miss all cases where the feature is supported by the instance (has the capability) but the user did not set it. The behavior that we want is that the feature is on by default and not explicitly require the user to set it.

I think we just need to remove the error from `getVMNetworkingCapability` and it will return `False` if it does not have the capability or if the value is not supported. Only return true if the map has the capability and the value is `True`.

r4f4:
> @r4f4 Let's discuss on Monday

Sure. I see now I was looking at the wrong place :facepalm:

pkg/asset/installconfig/azure/validation.go
```diff
@@ -188,11 +191,28 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+		if vmNetworkingType == "" && capabilities != nil {
+			val, ok := capabilities[string(aztypes.AcceleratedNetworkingEnabled)]
+			if ok {
+				if strings.EqualFold(val, "True") {
+					vmNetworkingType = string(aztypes.VMnetworkingTypeAccelerated)
+				} else {
+					vmNetworkingType = string(aztypes.VMNetworkingTypeBasic)
+				}
+			}
+
```
r4f4, line 209: This part can be removed, can't it? You're assigning a value to `vmNetworkingType` and then validating the value is correct (which it should be, since you assigned the correct value). Later on, when provisioning the machines, you're going to do the same assignment according to the capabilities again.

kwoodson: Sounds good to me. Will skip validation of the vmNetworkignType when set to `""`.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -88,59 +88,66 @@ func ValidateDiskEncryptionSet(client API, ic *types.InstallConfig) field.ErrorL
...
-		} else if strings.EqualFold(*capability.Name, string(aztypes.VMnetworkingTypeAccelerated)) {
-			if vmNetworkingType == string(aztypes.VMnetworkingTypeAccelerated) && !strings.EqualFold(*capability.Value, "True") {
+	if vmNetworkingType != "" {
+		val, ok = capabilities[string(aztypes.AcceleratedNetworkingEnabled)]
+		if ok {
+			if vmNetworkingType == string(aztypes.VMnetworkingTypeAccelerated) && !strings.EqualFold(val, "True") {
 				errMsg := fmt.Sprintf("vm networking type is not supported for instance type %s", instanceType)
 				allErrs = append(allErrs, field.Invalid(fieldPath.Child("vmNetworkingType"), vmNetworkingType, errMsg))
 			}
+		} else {
```
patrickdillon, line 138: if `vmNetworkingType == Basic` and `ok == false` this will throw an error, but I believe that is a valid state.

pkg/asset/machines/azure/machines.go
```diff
@@ -78,12 +80,28 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
...
 
+	hyperVGen, err := getHyperVGenerationVersion(capabilities)
+	if err != nil {
+		return nil, err
+	}
+
+	if mpool.VMNetworkingType == "" {
+		acceleratedNetworking, err := getVMNetworkingCapability(capabilities)
+		if err != nil {
+			logrus.Info(err)
```
patrickdillon, line 97: This error should be returned, not logged.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -188,11 +194,15 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
 		}
-		if vmNetworkingType == "" {
-			vmNetworkingType = defaultVMNetworkingType
+		capabilities, err = client.GetVMCapabilities(context.TODO(), instanceType, ic.Azure.Region)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, err.Error()))
 		}
 		ultraSSDEnabled := strings.EqualFold(ultraSSDCapability, "Enabled")
-		allErrs = append(allErrs, ValidateInstanceType(client, field.NewPath("controlPlane", "platform", "azure"), ic.Azure.Region, instanceType, diskType, controlPlaneReq, ultraSSDEnabled, vmNetworkingType)...)
+		// if capabilities fails to find a valid instance type, we should skip this validation
```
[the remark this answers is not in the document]
r4f4: The way the code is currently, the validation will also check the compute nodes before returning all the errors found at the end. Maybe we can do the immediate return as a separate PR with the reasoning why?

pkg/asset/installconfig/azure/validation.go
```diff
@@ -88,59 +88,66 @@ func ValidateDiskEncryptionSet(client API, ic *types.InstallConfig) field.ErrorL
...
-		} else if diskType == "Premium_LRS" && strings.EqualFold(*capability.Name, "PremiumIO") {
-			if strings.EqualFold(*capability.Value, "False") {
-				errMsg := fmt.Sprintf("PremiumIO not supported for instance type %s", instanceType)
-				allErrs = append(allErrs, field.Invalid(fieldPath.Child("osDisk", "diskType"), diskType, errMsg))
-			}
-		} else if strings.EqualFold(*capability.Name, "UltraSSDAvailable") {
-			ultraSSDAvailable = strings.EqualFold(*capability.Value, "True")
-		} else if strings.EqualFold(*capability.Name, string(aztypes.VMnetworkingTypeAccelerated)) {
-			if vmNetworkingType == string(aztypes.VMnetworkingTypeAccelerated) && !strings.EqualFold(*capability.Value, "True") {
+	if vmNetworkingType != "" {
```
r4f4, line 131:
```suggestion
	if vmNetworkingType == string(aztypes.VMnetworkingTypeAccelerated {
```
The only thing we need to validate is that the instance type supports Accelerated network when it's set in the install-config.

pkg/asset/installconfig/azure/validation.go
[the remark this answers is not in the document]
kwoodson: We skip the validation if the instance type is not valid and we add to the list of errors. Because these sections are different in the install config means that we have to check master and compute independently. This was the previously behavior and I followed that pattern.

pkg/asset/machines/azure/machines.go
kwoodson: Maybe we should discuss the behavior here. This is getting a bit confusing. This means the capabilities did not include the key for the capability. This was happening on Azure stack hub. The issue here is if they didn't put any value in to start (vmnetworkingtype = ""), I'm not sure we want to return error here. Maybe Azure stack doesn't support it and doesn't have it in the capabilities. This logrus was added from a previous discussion.

pkg/asset/machines/azure/capabilities.go
```diff
@@ -0,0 +1,41 @@
...
+			return "V2", nil
+		}
+		return "V1", nil
+	}
+	return "", fmt.Errorf("unable to determine HyperVGeneration version")
+}
+
+func getVMNetworkingCapability(capabilities map[string]string) (bool, error) {
+	val, ok := capabilities[string(azure.AcceleratedNetworkingEnabled)]
+	if !ok {
```
patrickdillon, line 29: My understanding is that if the capability was not included, that is the equivalent of false or not supoorted--not an error. WDYT @r4f4? I think that should be the case here and in `getUltraSSDCapability`

pkg/asset/machines/azure/machines.go
patrickdillon: Hopefully this comment will clarify: https://github.com/openshift/installer/pull/5846/files#r868360255

To me it doesn't seem like getVMNetworkingCapability should return an error at all.

patrickdillon: >  I'm not sure we want to return error here.

Yeah I agree with this.

pkg/asset/installconfig/azure/validation.go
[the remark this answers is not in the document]
r4f4: > Oh, I see. If `capabilities == nil` here that means `err != nil` above? this makes sense.

Exactly.

pkg/asset/machines/azure/capabilities.go
r4f4: Yeah, that's the case for the non-zonal case of the UltraSSD. For the networking, the docs [1] say:

> "Accelerated networking is available in all global Azure regions and Azure Government Cloud"

So my understanding is that the only case when the capability will be missing is for Azure Stack Cloud. In that case, I think it's safe to say the capability is not supported.

[1] https://docs.microsoft.com/en-us/azure/virtual-network/accelerated-networking-overview#regions

r4f4: The full ultraSSD validation gets a bit more complicated in the multi-zones case because the capability is not listed with the other instance capabilites but in the `LocationInfo`. So the absence of the key from the map is not enough to determine lack of support.

patrickdillon: I see. I think the solution to this is to have `GetVMCapabilities` return the `typeMeta` (or perhaps just the `LocationInfo`) in addition to the more user-friendly capabilities map. We can handle that in #5854 and does not need to be a part of this PR.

patrickdillon (commented):
I think we are really close here but there are two changes I see that we need:

1. Handling the case where `vmNetworkingType == Basic && !ok`, comment [1](https://github.com/openshift/installer/pull/5846/files#r868263879) & [2](https://github.com/openshift/installer/pull/5846/files#r868244952)
2. [Don't return an error for getVMNetworking](https://github.com/openshift/installer/pull/5846/files#r868369777) 

It would also be great if you could add a commit message.

@r4f4 is there anything else that needs to go in here?  #5854 can update `getUltraSSDCapability`. 

Thanks @kwoodson. I know this has been a slog and appreciate your help here.

r4f4:
> @r4f4 is there anything else that needs to go in here? #5854 can update `getUltraSSDCapability`.

I can fix it in a follow-up so we can get this PR moving, but I think in `validation.go` we can move the `GetVMCapabilities` call to inside `ValidateInstanceType`. 

> Thanks @kwoodson. I know this has been a slog and appreciate your help here.

+1

pkg/asset/installconfig/azure/validation.go
kwoodson: I think we should take @r4f4's suggestion here and only validate if the vmNetworkingType is set to `Accelerated`. We don't need to validate basic as it is the default setting. I'll update.

patrickdillon:
/approve
/hold

to see azure  e2e tests, anyone should feel free to cancel the hold if the test installs the cluster (no need for test suite to pass)

patrickdillon:
/hold cancel

I am satisfied with the CI run that we're not breaking anything.

Issue #5830, opened 2022-04-21
[OKD 4.11 vSphere IPI]  Bootstrap node fails to be destroyed after bootstrap completes successfully


fortinj66:
# Version

```console
[root@os-utils-d03 <poc410> poc410]$ ./openshift-install version
./openshift-install 4.11.0-0.okd-2022-04-21-163508
built from commit a722d01342175ddc0cd95969650b79cb3dba2a63
release image registry.ci.openshift.org/origin/release@sha256:593d68a713e07eb0927f5e8987f11fd88c2d8fdd247d8bb1a58bd3d2166af980
release architecture amd64

```

# Platform:
vSphere IPI

# What happened?

When bootstrap completes and the installer attempts to destroy the bootstrap VM, the following errors are returned:
```
[root@os-utils-d03 <poc410> poc410]$ ./openshift-install destroy bootstrap --dir okd-install
ERROR
ERROR Error: disk.0: virtual disk "disk0": cannot change the value of "eagerly_scrub" - (old: false newValue: true)
ERROR
ERROR   with vsphere_virtual_machine.vm_bootstrap,
ERROR   on main.tf line 12, in resource "vsphere_virtual_machine" "vm_bootstrap":
ERROR   12: resource "vsphere_virtual_machine" "vm_bootstrap" {
ERROR
FATAL terraform destroy: failed doing terraform destroy: exit status 1
FATAL
FATAL Error: disk.0: virtual disk "disk0": cannot change the value of "eagerly_scrub" - (old: false newValue: true)
FATAL
FATAL   with vsphere_virtual_machine.vm_bootstrap,
FATAL   on main.tf line 12, in resource "vsphere_virtual_machine" "vm_bootstrap":
FATAL   12: resource "vsphere_virtual_machine" "vm_bootstrap" {
FATAL
FATAL

```
# What you expected to happen?

Bootstrap node to be removed

# How to reproduce it (as minimally and precisely as possible)?

Normal IPI install with 4.11 installer above...

jcpowermac:
@fortinj66 what kind of storage do you use?
CI uses all vSAN and we haven't ran into is why I am asking

Also see:
https://github.com/hashicorp/terraform-provider-vsphere/issues/1028

fortinj66:
HP 3PAR Flash...

```
 platform:
    vsphere:
      apiVIP: 10.102.5.11
      cluster: GSO-ESX
      datacenter: GSO-ESX
[...]

PR #5848, opened 2022-04-27
vSphere: ignore bootstrap eagerly scrub changes

jcpowermac:
The vsphere terraform provider has a long standing
bug if a disk parameter is changed underneath
terraform by vSphere the apply will fail.

This resolves reported issue:
https://github.com/openshift/installer/issues/5830


patrickdillon:
+1. LGTM. Holding off on the label because the CI job is not required, and want to wait and see if we need more.

jcpowermac:
job failure unrelated to installation and PR.

PR #5852, opened 2022-04-28
AGENT-132: Import fleeting prototype

zaneb:
Import the fleeting prototype from https://github.com/openshift-agent-team/fleeting

History is included, by using git-filter-repo to include only the files that are required, and in their correct locations.

In order to get CI testing working as soon as possible, the code is imported largely unchanged. Once this merges, our first priority will be to convert it to installer idioms.


zaneb:
I tested manually using https://github.com/openshift/installer/pull/5852 and it now successfully creates a cluster.

rwsu (commented):
Found a few more places where panic could be used. Otherwise lgtm.

bfournie:
Tested this with https://github.com/openshift-metal3/dev-scripts/pull/1386 and was able to get a full installation.

zaneb:
I think we want to eliminate both `panic()` and `os.Exit(1)` as we convert stuff to Assets. As long as lint stops yelling at us there's probably little reason to prefer one over the other in the meantime.

patrickdillon:
Can we clean up the commit structure? In general we try to be pretty strict about the commit structure in the installer repo: particularly in organizing commits in a logical fashion and having good commit messages (also separating vendoring into its own commit for reviewing). That might be difficult when bringing in a large piece of work like this so I'm happy to be flexible, but it looks like this is pulling in merge commits which probably should not be included?

zaneb:
> Can we clean up the commit structure? In general we try to be pretty strict about the commit structure in the installer repo: particularly in organizing commits in a logical fashion and having good commit messages (also separating vendoring into its own commit for reviewing). That might be difficult when bringing in a large piece of work like this so I'm happy to be flexible, but it looks like this is pulling in merge commits which probably should not be included?

It's hard to tell from GitHub, but what I did is create a branch with _all_ of the history of the fleeting prototype (including merges, using git-filter-repo) so that e.g. git blame, git revert continue to work as expected. Then there's a [single merge commit](https://github.com/openshift/installer/pull/5852/commits/89d3b0baa06872f7a737649262184c9e706fff7d) to merge that into the installer, since obviously they don't share any history before that. Vendoring is in its [own commit](https://github.com/openshift/installer/pull/5852/commits/fcba8e9247091fc92dcdc7ec9ef0ff196a9d3d6b) on top of that.

Here's a better diagram than you can get out of GitHub:

image

So I feel like this is actually pretty well-organised, and offline tools work well with this structure but GitHub kind of gives up.

zaneb:
The e2e-metal-ipi-ovn-ipv6 is borked at the moment (possible fix is https://github.com/openshift/origin/pull/27095). @patrickdillon I think it'd be reasonable to just override that one if you have a moment.

PR #5854, opened 2022-04-28
Azure: validate UltraSSD instance types in multi-zone regions

r4f4:
In order to validate UltraSSD in multi-zone regions, I have updated the Azure client to return `LocationInfo` information which we then use to check if availability zones are supported and which ones have UltraSSD available.

The error messages might still need some more polish before this is merged but I would like to get some comments, specially on 58ad37dbd400302d17c4faf121e8d13db8d99c47, before I proceed ironing out the details.

I have also added unit tests for as many combinations as I found relevant but they are by no means exhaustive.


r4f4:
/hold
Until I get feedback on 79d6587020d03c8c91a38017cb70b1fb489115cd

pkg/asset/installconfig/azure/client.go
```diff
@@ -248,12 +249,17 @@ func (c *Client) GetVirtualMachineSku(ctx context.Context, name, region string)
 			if !strings.EqualFold(name, *sku.Name) {
 				continue
 			}
+			return &sku, nil
+
+			// FIXME: I don't think the following block is needed anymore since
```
rna-afk, line 254: I don't think we need it if you are confident the filter works. Haven't tried it myself but it looks like it should work.

rna-afk (commented):
LGTM. Can't see any complications coming from switching to the service but will wait for others' opinion

pkg/asset/installconfig/azure/client.go
r4f4: Right. I'll remove it if I don't see anything suspicious in the CI runs.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -127,7 +129,7 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 				errMsg := fmt.Sprintf("PremiumIO not supported for instance type %s", instanceType)
 				allErrs = append(allErrs, field.Invalid(fieldPath.Child("osDisk", "diskType"), diskType, errMsg))
 			}
-		} else if strings.EqualFold(*capability.Name, "UltraSSDAvailable") {
+		} else if strings.EqualFold(*capability.Name, ultraSSDCapability) {
```
patrickdillon, line 132:
```suggestion
		} else if strings.EqualFold(*capability.Name, aztypes.AcceleratedNetworkingEnabled) {
```
This should be fixed by either #5840 or more probably #5846, although of course you are free to fix here.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -137,6 +139,37 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 		}
 	}
 
+	// Convenience check so the tests don't have to define LocationInfo if it's not used
+	if typeMeta.LocationInfo != nil && ultraSSDEnabled {
+		for _, locationInfo := range *typeMeta.LocationInfo {
+			// If Availability Zones are not supported (e.g StackCloud)
+			if locationInfo.ZoneDetails == nil {
+				continue
+			}
```
patrickdillon, line 148: This case is already handled by the preceding conditional two lines earlier.

patrickdillon (commented):
This looks great. A couple of suggestions.

pkg/asset/installconfig/azure/validation.go
r4f4: In my tests against the API for the Azure Stack case the `LocationInfo` struct was defined but the `ZoneDetails` field was `nil`. So I need to check that before dereferencing the pointer in the `for` loop that follows. I even added that as a test case https://github.com/openshift/installer/pull/5854/files#diff-310a64406bdb016820cd032ef2ca5cd092ba942f2b7ccbedeec0abc254fbef74R670-R676

pkg/asset/installconfig/azure/validation.go
r4f4: You mean I should add `UltraSSDAvailable` as a type in `aztypes`?

r4f4:
I've also implemented an alternative approach in https://github.com/r4f4/installer/tree/azure-ultrassd-alt

Amoghrd:
Pre-merge testing works fine in all regions with appropriate error messages.

r4f4:
/close
Closing this in favour of https://github.com/openshift/installer/pull/5937

PR #5857, opened 2022-05-01
Change the default machine pool config for Nutanix

thunderboltsid:
Adjust the default machine pool configuration for Nutanix machines
  - Reduce the number of cores per socket:
    - Control plane nodes: From 4 cores per socket to 1
    - Worker nodes: From 2 cores per socket to 1
  - Increase the number of CPUs:
    - Control plane nodes: From 4 CPUs to 8
    - Worker nodes: From 2 CPUs to 4
  - Increased the RAM to 16GB.

Note: The AHV scheduler treats socket and core allocation exactly the same.


thunderboltsid:
cc @patrickdillon for approval

PR #5864, opened 2022-05-04
Bug 2078431: Set TF_DATA_DIR to the Terraform Dir.

patrickdillon:
Sets the TF_DATA_DIR env variable so that .terraform folder and
extracted provider binaries are written to a directory that has
the same level of permissions as the Terraform binary.

### DIFF
--- a/pkg/terraform/init.go
+++ b/pkg/terraform/init.go
@@ -62,9 +62,7 @@ func unpackAndInit(dir string, platform string, target string, terraformDir stri
 	}
 
 	// Explicitly specify the CLI config file to use so that we control the providers that are used.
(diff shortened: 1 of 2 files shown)


patrickdillon:
If this approach does work, we may want to consider extracting this to a subdir within terraform dir.

janosdebugs (approved):
/lgtm

patrickdillon:
/lgtm cancel

`msg=Error: Could not load plugin`

patrickdillon:
2d7b14dab7dca69cb7ba90cb1b99719dd4ebb1ee fixes the earlier test failures. so we could remove the hold but I would like to get this verified by ovirt folks

PR #5868, opened 2022-05-05
Pull the OS Image for Nutanix via URL

thunderboltsid:
Instead of uploading the rhcos image from the installer machine
to Prism Central, ask Prism Central to download the image through
a URL. This is a lot less error prone.

### DIFF
--- a/data/data/nutanix/cluster/main.tf
+++ b/data/data/nutanix/cluster/main.tf
@@ -29,7 +29,7 @@ resource "nutanix_category_value" "ocp_category_value_shared" {
 
 resource "nutanix_image" "rhcos" {
   name        = var.nutanix_image
-  source_path = var.nutanix_image_filepath
+  source_uri  = var.nutanix_image_uri
   description = local.description
 
   categories {
--- a/data/data/nutanix/variables-nutanix.tf
+++ b/data/data/nutanix/variables-nutanix.tf
@@ -27,9 +27,9 @@ variable "nutanix_prism_element_uuid" {
   description = "This is the uuid of the Prism Element cluster."
 }
 
-variable "nutanix_image_filepath" {
+variable "nutanix_image_uri" {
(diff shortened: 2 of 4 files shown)


jcpowermac:
@thunderboltsid if `ClusterOSImage` is defined
https://github.com/openshift/installer/blob/master/pkg/types/nutanix/platform.go#L16

This URL should be replaced:
https://github.com/openshift/installer/blob/master/pkg/asset/cluster/tfvars.go#L875

Thinking the scenario of disconnected installations.

jcpowermac:
Hmm and wouldn't this be debatable if we need it anymore?
https://github.com/openshift/installer/blob/master/pkg/tfvars/nutanix/nutanix.go#L43

thunderboltsid:
> Hmm and wouldn't this be debatable if we need it anymore?
> https://github.com/openshift/installer/blob/master/pkg/tfvars/nutanix/nutanix.go#L43

You mean the `DownloadImageFile` function? There's other providers that are still using it.

jcpowermac:
> > Hmm and wouldn't this be debatable if we need it anymore?
> > https://github.com/openshift/installer/blob/master/pkg/tfvars/nutanix/nutanix.go#L43
> 
> You mean the `DownloadImageFile` function? There're other providers that are still using it.

Right but do we need to download the file locally where the installer is running if prism central is going to download it anyway?
I guess it doesn't matter but it just won't be used if I am thinking correctly.

thunderboltsid:
> @thunderboltsid if `ClusterOSImage` is defined https://github.com/openshift/installer/blob/master/pkg/types/nutanix/platform.go#L16
> 
> This URL should be replaced: https://github.com/openshift/installer/blob/master/pkg/asset/cluster/tfvars.go#L875
> 
> Thinking the scenario of disconnected installations.

Good catch!

PR #5870, opened 2022-05-06
Update to Golang v1.17

creydr:
When I run the `hack/go-test.sh` locally, I get the following errors:

```
$ ./hack/go-test.sh
# sigs.k8s.io/json/internal/golang/encoding/json
vendor/sigs.k8s.io/json/internal/golang/encoding/json/encode.go:1249:12: sf.IsExported undefined (type reflect.StructField has no field or method IsExported)
vendor/sigs.k8s.io/json/internal/golang/encoding/json/encode.go:1255:18: sf.IsExported undefined (type reflect.StructField has no field or method IsExported)
...
```
This is because of an older golang version (CI is running already on Golang 1.17: https://github.com/openshift/installer/blob/21cd5218bb58288cd7b03018b9a2513aca3a13a5/.ci-operator.yaml#L4)

This PR addresses it and updates the Golang version to 1.17.

### DIFF
--- a/docs/dev/dependencies.md
+++ b/docs/dev/dependencies.md
@@ -18,7 +18,7 @@ We follow a hard flattening approach; i.e. direct and inherited dependencies are
 
 Dependencies are managed with [Go Modules](https://github.com/golang/go/wiki/Modules) but committed directly to the repository.
 
-We require at least Go 1.16.
+We require at least Go 1.17.
 
 - Add or update a dependency with `go get <dependency>@<version>`.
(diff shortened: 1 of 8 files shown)


patrickdillon:
/approve

Thanks for taking care of this. I recently noticed this myself. Longer term we should investigate moving away from using containers in these scripts IMHO, but I think we should move forward with unbreaking things first.

creydr:
@sadasu could you check on this PR too?

PR #5871, opened 2022-05-06
AGENT-159: Assets generation for create manifests command

andfasano:
This patch introduces the assets integration of the agent installer for the create manifest command


cmd/openshift-install/agent.go
```diff
@@ -15,7 +20,67 @@ func newAgentCmd() *cobra.Command {
...
+	}
+
+	agentTargets = []target{agentManifestsTarget, agentImageTarget}
+)
+
+func newAgentCreateCmd() *cobra.Command {
+
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Commands for generating agent installer based artifacts",
```
celebdor, line 72:
```suggestion
		Short: "Commands for generating agent installation artifacts",
```
This might be more accurate.

pkg/asset/agent/manifests/infraenv.go
```diff
@@ -0,0 +1,106 @@
...
+	if i.File != nil {
+		return []*asset.File{i.File}
+	}
+	return []*asset.File{}
+}
+
+// Load returns infraenv asset from the disk.
+func (i *InfraEnv) Load(f asset.FileFetcher) (bool, error) {
+
+	infraEnvFile, err := f.FetchByName(infraEnvFilename)
```
zaneb, line 96: Arguably we should do FetchByPattern("*.yaml") and look through the contents for the one that contains an InfraEnv resource type rather than assuming it has a particular filename. That would be more compatible with ZTP manifests generated by other means than with `openshift install create cluster-manifests` (e.g. SiteConfig).

This is fine for now though.

andfasano: Ok, let's keep it by name for now, and eventually we'll revisit later.

pkg/asset/agent/manifests/agent.go
```diff
@@ -0,0 +1,77 @@
...
+		&AgentPullSecret{},
+		&InfraEnv{},
+	}
+}
+
+// Generate generates the respective manifest files.
+func (m *AgentManifests) Generate(dependencies asset.Parents) error {
+	for _, a := range []asset.WritableAsset{
+		&AgentPullSecret{},
+		&InfraEnv{},
```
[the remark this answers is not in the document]
andfasano: I'd not recommend to do that. The current contract of the `Dependencies()` method is to provide just the [types](https://github.com/openshift/installer/blob/6d778f911e79afad8ba2ff4301eda5b5cf4d8e9e/pkg/asset/store/store.go#L199) of the dependencies  - and evenutally allocate a new instance in case of load from disk. So, even though at the cost of a little bit of duplication, the current approach seems to be a reasonable compromise (considering also that you're interested in the concrete asset types, to extract the parsed/loaded data subsequently).

rwsu (approved):
/lgtm

pkg/asset/agent/manifests/agent.go
```diff
@@ -0,0 +1,77 @@
...
+	} {
+		dependencies.Get(a)
+		m.FileList = append(m.FileList, a.Files()...)
```
zaneb, line 44: It looks like there might be nothing to initialise FileList here?

PR #5872, opened 2022-05-06
AGENT-40: allow reading cert keys

zaneb:
Ensure that we can use an admin kubeconfig pre-generated by the agent
installer. This requires adding the ability to read all of the
certificates that it depends on:

overrides

Certs and keys will be read from disk only if they exist and the environment variable `OPENSHIFT_INSTALL_LOAD_CLUSTER_CERTS` is set to `true`.

### DIFF
--- a/pkg/asset/tls/certkey.go
+++ b/pkg/asset/tls/certkey.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"crypto/rsa"
 	"crypto/x509"
+	"os"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


pkg/asset/tls/certkey.go
```diff
@@ -69,6 +70,39 @@ func (c *CertKey) Load(asset.FileFetcher) (bool, error) {
 	return false, nil
 }
 
+func (c *CertKey) loadCertKey(f asset.FileFetcher, filenameBase string) (bool, error) {
+	if os.Getenv("OPENSHIFT_INSTALL_LOAD_CLUSTER_CERTS") != "true" {
```
patrickdillon, line 74: Who is setting this? Will it be set by the agent installer or users? We have avoided using any user-set environment variables.

zaneb: It'll be set by the assisted-installer. This feature is not intended for consumption by users running openshift-install directly, hence the env var check.

romfreiman: Similar to the invoker one:
https://github.com/openshift/installer/blob/6d778f911e79afad8ba2ff4301eda5b5cf4d8e9e/pkg/asset/openshiftinstall/openshiftinstall.go#L78

pkg/asset/tls/certkey.go
```diff
@@ -69,6 +70,39 @@ func (c *CertKey) Load(asset.FileFetcher) (bool, error) {
 	return false, nil
 }
 
+func (c *CertKey) loadCertKey(f asset.FileFetcher, filenameBase string) (bool, error) {
```
rwsu, line 73: Should there be a unit test for this function?

rwsu (commented):
Might be good to have a unit test for loadCertKey. Otherwise lgtm.

pkg/asset/tls/certkey.go
patrickdillon: Avoiding env vars set by users is my main concern. This looks reasonable as the env var is avoiding this functionality creeping outside of the agent installer.

/approve

PR #5809, opened 2022-04-13
Add Power VS-specific fields to cloud-provider config

clnperez:
adding new fields to align with changes in cloud-provider-powervs

Signed-off-by: Christy Norman <christy@linux.vnet.ibm.com>

PR #5875, opened 2022-05-09
Update template for CloudProviderConfig

hamzy:
The PowerVS cloud provider needs additional parameters.	 These were added previously
but not	added to the template.


clnperez:
@rna-afk @patrickdillon can we get this one in? i missed this file when i submitted https://github.com/openshift/installer/pull/5809

PR #5876, opened 2022-05-09
[release-4.10] Bug 2083335: Set rc-manager=unmanaged for on-prem bootstrap

cybertron:
This is a combined backport of https://github.com/openshift/installer/pull/5482 and https://github.com/openshift/installer/pull/5842 (which confusingly ended up with PR numbers where only the two middle digits were transposed). Both patches are needed to properly fix the bug on all platforms.


cybertron:
Looks like it needs cherry-pick-approved. Who would have permission to add that on this repo?

PR #5883, opened 2022-05-10
Power VS Install Config changes

23TNC:
This PR attempts to move Region and Zone out of .powervs/config.json as well as add ResourceGroup to the survey and default to Default.

### DIFF
--- a/pkg/asset/installconfig/azure/client.go
+++ b/pkg/asset/installconfig/azure/client.go
@@ -37,6 +37,7 @@ type API interface {
 	AreMarketplaceImageTermsAccepted(ctx context.Context, publisher, offer, sku string) (bool, error)
 	GetVMCapabilities(ctx context.Context, instanceType, region string) (map[string]string, error)
 	GetAvailabilityZones(ctx context.Context, region string, instanceType string) ([]string, error)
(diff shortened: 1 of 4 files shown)


clnperez:
remind me why we're writing the apkey to disk if we're requiring that it be set as an env var? Or, are we still prompting for that during `create install-config` so it's not mandatory that it be set as an env var?

23TNC:
API Key and ID are saved to config.json under .powervs in the users home directory. This does a few things, it prevents requiring the apikey and id every call.  It also ensures its written to disk and available to subsequent calls later in the process and therefore wont prompt for the API Key and ID 3 times.

clnperez:
i forgot about https://github.com/openshift/installer/pull/5734. it's been open for so long. it looks like these two are changing the same functions, so let's get it in first and circle back to cleaning this up

barbacbd:
@23TNC can you squash the commits? There is no need for three commits this small that are all related.

clnperez:
@23TNC did you close this intentionally?

barbacbd:
@23TNC was this reopened as a new PR [here](https://github.com/openshift/installer/pull/6015) ?

23TNC:
yes

PR #5889, opened 2022-05-11
Agent-160: Add baseIso asset

bfournie:
Download the baseIso used for the agent "create image" command.

Note that cache.go is copied from ./pkg/tfvars/internal/cache/, since its in an internal dir it can't be imported. A follow-on PR
will move this file to a common area so that it can be used by both pkgs.


pkg/asset/agent/image/baseiso.go
```diff
@@ -0,0 +1,141 @@
...
+func getIsoFromReleasePayload(config *types.InstallConfig) (string, error) {
+
+	// TODO
+	return "", nil
+}
+
+// Dependencies returns dependencies used by the asset.
+func (i *BaseIso) Dependencies() []asset.Asset {
+	return []asset.Asset{
+		&installconfig.InstallConfig{},
```
zaneb, line 92: We can't really depend on the install-config until https://issues.redhat.com/browse/AGENT-136 is implemented, because it will make it compulsory to provide an install-config.

pkg/agent/build_image.go
```diff
@@ -4,20 +4,26 @@ import (
...
-	if err != nil {
-		return err
+	baseIso := &image.BaseIso{}
+	if err := assetStore.Fetch(baseIso); err != nil {
+		return errors.Wrapf(err, "failed to fetch %s", baseIso.Name())
+	}
+	logrus.Infof("Got BaseIso From assetstore %s", baseIso.File.Filename)
+
+	if err = imagebuilder.BuildImage(baseIso.File.Filename); err != nil {
+		return errors.Wrapf(err, "failed to build IOS using %s", baseIso.Name())
```
rwsu, line 26:
```suggestion
		return errors.Wrapf(err, "failed to build ISO using %s", baseIso.Name())
```

pkg/asset/agent/image/baseiso.go
```diff
@@ -0,0 +1,139 @@
...
+	ctx, cancel := context.WithTimeout(context.TODO(), 30*time.Second)
+	defer cancel()
+
+	// Get the ISO to use from rhcos.json
+	st, err := rhcos.FetchCoreOSBuild(ctx)
+	if err != nil {
+		return "", err
+	}
+
+	// Defaults to using the X86_64 baremetal ISO for all platforms
```
rwsu, line 57:
```suggestion
	// Defaults to using the x86_64 baremetal ISO for all platforms
```

pkg/asset/agent/image/baseiso.go
```diff
@@ -0,0 +1,139 @@
...
+	// ic := &installconfig.InstallConfig{}
+	// p.Get(ic)
+	// use the GetIso function to get the BaseIso from the release payload
+	isoGetter := newGetIso(GetIsoPluggable)
+	baseIsoFileName, err := isoGetter.getter()
+	if err != nil {
+		return errors.Wrap(err, "failed to get base iso image")
+	}
+
+	i.File = &asset.File{Filename: baseIsoFileName}
```
rwsu, line 108: Shouldn't asset.Data also be set? Is it not necessary because when the ISO is generated, it reads it again from the path (cache)?

rwsu (commented):
The generated fleeting.iso booted ok. Small nits.

pkg/asset/agent/image/baseiso.go
bfournie: Fixed

pkg/agent/build_image.go
bfournie: Fixed

pkg/asset/agent/image/baseiso.go
bfournie: Good point. The file is cached and we use the cached location so don't need to copy the data here. I can change *type BaseIso struct* to just be a filename string (similar to  https://github.com/openshift/installer/blob/master/pkg/asset/rhcos/image.go#L35 which also uses the cached file) although it wasn't clear when we get the baseiso from the release payload in the disconnected case if we can just use the filename or need to copy the data.

rwsu (approved):
/lgtm

pkg/asset/agent/image/baseiso.go
bfournie: Changed to remove InstallConfig as a dependency.

pkg/asset/agent/image/baseiso.go
```diff
@@ -0,0 +1,139 @@
...
+	return &getIso{getter: getter}
+}
+
+// GetIsoPluggable defines the method to use get the baseIso file
+var GetIsoPluggable = downloadIso
+
+// Download the ISO using the URL in rhcos.json
+func downloadIso() (string, error) {
+
+	ctx, cancel := context.WithTimeout(context.TODO(), 30*time.Second)
```
andfasano, line 48: Probably this function could be refactored in future to remove some duplication (and I guess using also the `Image` asset directly once we could add a dependency to `InstallConfig`), but for now it could be reasonable for me

pkg/asset/agent/image/baseiso.go
```diff
@@ -0,0 +1,139 @@
...
+
+var (
+	baseIsoFilename = ""
+)
+
+var _ asset.WritableAsset = (*BaseIso)(nil)
+
+// Name returns the human-friendly name of the asset.
+func (i *BaseIso) Name() string {
+	return "BaseIso"
```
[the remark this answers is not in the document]
bfournie: Updated.

pkg/asset/agent/image/baseiso.go
bfournie: Yes once InstallConfig is used we can remove some of this code, will leave it here for now

rwsu (commented):
/lgtm

PR #5890, opened 2022-05-11
AGENT-169: AgentClusterInstall asset

rwsu:
Creates the asset for AgentClusterInstall.

There is a partial implementation of the translation of
install-config.yaml to AgentClusterInstall.


pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -1 +1,111 @@
...
+	"github.com/openshift/installer/pkg/asset"
+	"github.com/openshift/installer/pkg/asset/installconfig"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
+)
+
+var (
+	agentClusterInstallFilename = filepath.Join(clusterManifestDir, "agent-cluster-install.yml")
```
bfournie, line 18: Nit - using "yml" for the extension is consistent with infraenv.yml but different than pull-secret.yaml. Either is fine, just wondering if we should be consistent.

bfournie, line 18: Nit - using "yml" for the extension is consistent with infraenv.yml but different than pull-secret.yaml. Either is fine, just wondering if we should be consistent.

rwsu: I think we've been using .yaml. Thank and good catch!

pkg/asset/agent/manifests/agentclusterinstall_test.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/openshift/installer/pkg/asset/mock"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// func TestAgentClusterInstall_Generate(t *testing.T) {
+
+// 	installConfig := &installconfig.InstallConfig{
```
andfasano, line 20: nit: also t.Skip can be used

PR #5891, opened 2022-05-11
AGENT-193: Do not set api_vip for SNO cluster

pawanpinjarkar:
Fix the unwanted API call to set API_VIP in case of SNO cluster in start-cluster-installation.service. 
`{"code":"400","href":"","id":400,"kind":"Error","reason":"API VIP cannot be set with User Managed Networking"}`
Although this HTTP 400 error doesn't impact starting cluster installation in the case of SNO, there is no need to check the API_VIP in the first place.

Ref: https://issues.redhat.com/browse/AGENT-193

### DIFF
--- a/data/data/agent/files/usr/local/bin/start-cluster-installation.sh.template
+++ b/data/data/agent/files/usr/local/bin/start-cluster-installation.sh.template
@@ -39,7 +39,7 @@ done
 
 echo "All ${total_required_nodes} hosts are ready."
 
-if [[ "$total_required_nodes" > "1" ]]; then
+if [[ "${APIVIP}" != "" ]]; then
     api_vip=$(curl -s -S "${BASE_URL}/clusters" | jq -r .[].api_vip)
     if [ "${api_vip}" == null ]; then
         echo "Setting api vip"
--- a/data/data/agent/files/usr/local/bin/start-cluster-installation.sh.template
+++ b/data/data/agent/files/usr/local/bin/start-cluster-installation.sh.template
@@ -39,10 +39,12 @@ done
 
 echo "All ${total_required_nodes} hosts are ready."
 
-api_vip=$(curl -s -S "${BASE_URL}/clusters" | jq -r .[].api_vip)
-if [ "${api_vip}" == null ]; then
-    echo "Setting api vip"
-    curl -s -S -X PATCH "${BASE_URL}/clusters/${cluster_id}" -H "Content-Type: application/json" -d '{"api_vip": "{{.APIVIP}}"}'
+if [[ "$total_required_nodes" > "1" ]]; then
+    api_vip=$(curl -s -S "${BASE_URL}/clusters" | jq -r .[].api_vip)
+    if [ "${api_vip}" == null ]; then
+        echo "Setting api vip"
+        curl -s -S -X PATCH "${BASE_URL}/clusters/${cluster_id}" -H "Content-Type: application/json" -d '{"api_vip": "{{.APIVIP}}"}'
+    fi
 fi
 
 while [[ "${cluster_status}" != "ready" ]]


data/data/agent/files/usr/local/bin/start-cluster-installation.sh.template
```diff
@@ -39,10 +39,12 @@ done
 
 echo "All ${total_required_nodes} hosts are ready."
 
-api_vip=$(curl -s -S "${BASE_URL}/clusters" | jq -r .[].api_vip)
-if [ "${api_vip}" == null ]; then
-    echo "Setting api vip"
-    curl -s -S -X PATCH "${BASE_URL}/clusters/${cluster_id}" -H "Content-Type: application/json" -d '{"api_vip": "{{.APIVIP}}"}'
+if [[ "$total_required_nodes" > "1" ]]; then
```
rwsu, line 42: I don't think this script is the correct place to not set the api_vip. This should be done in the AgentClusterInstall CR. For SNO clusters, it should not contain the api_vip.

rwsu: The logic here should be if api_vip is set, then patch the cluster.

pawanpinjarkar: > I don't think this script is the correct place to not set the api_vip. This should be done in the AgentClusterInstall CR. For SNO clusters, it should not contain the api_vip.

I am doing that here https://github.com/openshift-metal3/dev-scripts/pull/1387/files#diff-f1c857ae41bcafd8ae197fe612a47b44bb97ca35b45193db1f5dc1b9e1a2496fR81-R86 in another PR. The condition in the current PR is for a different situation. Here is that ref https://coreos.slack.com/archives/CUPJTHQ5P/p1647877938543519 Basically, the next button was disabled in absence of API VIP. I will check this more once all the other bugs found recently are fixed with the cluster installation.

In non SNO cluster,

pawanpinjarkar: @rwsu 
Nit: Fix the unwanted the API call to set API_VIP in case of SNO `{"code":"400","href":"","id":400,"kind":"Error","reason":"API VIP cannot be set with User Managed Networking"}`

Current code is
https://github.com/openshift/installer/blob/13b32ccbc02fd6243982cef586afd0db3a75e868/data/data/agent/files/usr/local/bin/start-cluster-installation.sh.template#L51-L55

this results in below logs in case of SNO
```
[core@node0 ~]$ sudo systemctl status start-cluster-installation.service 
● start-cluster-installation.service - Service that starts cluster installation
   Loaded: loaded (/etc/systemd/system/start-cluster-installation.service; enabled; vendor preset: enabled)
   Active: active (exited) since Mon 2022-05-16 14:23:25 UTC; 3min 59s ago
 Main PID: 2346 (code=exited, status=0/SUCCESS)
    Tasks: 0 (limit: 205176)
   Memory: 0B
   CGroup: /system.slice/start-cluster-installation.service

May 16 14:22:49 node0 start-cluster-installation.sh[2346]: Waiting for hosts to become ready for cluster installation...
May 16 14:23:19 node0 start-cluster-installation.sh[2346]: Hosts known and ready for cluster installation (1/1)
May 16 14:23:19 node0 start-cluster-installation.sh[2346]: All 1 hosts are ready.
May 16 14:23:19 node0 start-cluster-installation.sh[2346]: Setting api vip
May 16 14:23:19 node0 start-cluster-installation.sh[2346]: {"code":"400","href":"","id":400,"kind":"Error","reason":"API VIP cannot be set with User Managed Networking"}
```

Although this HTTP 400 error doesn't impact starting cluster installation in case of SNO, I think there is no need to check the API_VIP at first place.

rwsu, line 42: Instead of conditioning on the number of nodes, we can condition on whether the api vip was set in the AgentClusterInstall.  The api vip won't be set for SNO clusters in the AgentClusterInstall. This was the change I was suggesting:
```suggestion
if [[ "${APIVIP}" != "" ]]; then
```

rwsu (changes requested):

data/data/agent/files/usr/local/bin/start-cluster-installation.sh.template
celebdor: Yes. I don't think we should limit it by nodes. Rather, by ACI VIPs. No vips, none platform

rwsu (approved):
/lgtm

PR #5865, opened 2022-05-05
Bump Azure Stack Terraform Provider for Go 1.18

patrickdillon:
Builds for Mac are failing with Go 1.18 due to an outdated dependency in the Azure Stack Terraform provider on x/sys: https://github.com/golang/go/issues/49219

I tested this using `registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.11` and it resolves the issue.

PR #5896, opened 2022-05-12
Bug 2085187: terraform: revendor golang.org/x/sys

yselkowitz:
This is necessary for these providers to build for Mac with Go 1.18.

https://github.com/golang/go/issues/51706


barbacbd:
Looks like the same versions were used from [#5865](https://github.com/openshift/installer/pull/5865)

yselkowitz:
random: our version is behind, latest upstream version builds fine with 1.18
ignition: https://github.com/community-terraform-providers/terraform-provider-ignition/pull/19

patrickdillon:
@yselkowitz your upstream ignition PR merged. Instead of revendoring x/sys locally can we bring in the new versions of these providers with the proper x/sys support?

yselkowitz:
For ignition, no new version has been spun, but the revendoring would be the only upstream code change, so that still seems appropriate.  For remote, a rebase to v3.1.3 was more than I bargained for, but if you insist...

sdodson:
@yselkowitz Do we ship builds built via this method or do we cross compile? Just trying to gauge urgency here.

yselkowitz:
yes, this affects the build of installer-artifacts, which is currently behind installer and the rest of OCP as a result.

patrickdillon:
> @yselkowitz Do we ship builds built via this method or do we cross compile? Just trying to gauge urgency here.

This is urgent. ART is currently making an exception to build installer with go 1.18 until we get this merged. Thanks @yselkowitz  for your work on this.

yselkowitz:
@patrickdillon e2e-vsphere-upi passed

PR #5898, opened 2022-05-12
GCP: Revert Instance Type from N2 to N1

patrickdillon:
GCP has separate quotas for N2 CPUs: https://cloud.google.com/compute/quotas#cpu_quota
We should revert this change until we're not breaking CI.

These changes were introduced in: https://github.com/openshift/installer/pull/5841/commits/16f2827aafa325d9ca5c7e525d5c1abb400d5dd9

### DIFF
--- a/pkg/asset/machines/worker.go
+++ b/pkg/asset/machines/worker.go
@@ -113,7 +113,7 @@ func defaultAzureMachinePoolPlatform() azuretypes.MachinePool {
 
 func defaultGCPMachinePoolPlatform() gcptypes.MachinePool {
 	return gcptypes.MachinePool{
-		InstanceType: "n2-standard-4",
+		InstanceType: "n1-standard-4",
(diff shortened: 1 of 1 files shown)


cgwalters:
Just so I understand this, the core problem is the CI account needs a quota bump, then we can re-land this change?

patrickdillon:
> Just so I understand this, the core problem is the CI account needs a quota bump, then we can re-land this change?

Precisely. (or we solve the ci problem another way.) But I don't want to break all of GCP CI for the last day of FF.

cgwalters:
Sorry I'm just going to drop in here with a hammer :hammer: 
This can't affect AWS so
/override ci/prow/e2e-aws

patrickdillon:
/override ci/prow/e2e-aws

but don't think that is required here

wking:
#6058 is taking a second run at moving to n2.

PR #5899, opened 2022-05-12
[release-4.11.0] Bug 2095226: Added changes to verify cloud connection and dhcp serviceinstance details of a  powervs instance

bkhadars:
Added changes for verifying the cloudconnection and dhcpserverinstance details of a powervs instance in the ibm cloud

### DIFF
--- a/pkg/asset/installconfig/powervs/session.go
+++ b/pkg/asset/installconfig/powervs/session.go
@@ -1,6 +1,7 @@
 package powervs
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -20,6 +21,7 @@ import (
 	"github.com/IBM-Cloud/bluemix-go/http"
 	"github.com/IBM-Cloud/bluemix-go/rest"
 	bxsession "github.com/IBM-Cloud/bluemix-go/session"
+	"github.com/IBM-Cloud/power-go-client/clients/instance"
 	"github.com/IBM-Cloud/power-go-client/ibmpisession"
 	"github.com/IBM/go-sdk-core/v5/core"
 	"github.com/form3tech-oss/jwt-go"
@@ -178,6 +180,62 @@ func (c *BxClient) ValidateAccountPermissions() error {
 	}
 	if accType == "TRIAL" {
 		return fmt.Errorf("account type must be of Pay-As-You-Go/Subscription type for provision Power VS resources")
+	}
+	return nil
+}
+
+//ValidateDhcpService checks for existing Dhcp service for the provided PowerVS cloud instance
+func (c *BxClient) ValidateDhcpService(ctx context.Context, svcInsID string) error {
+	ctx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+	defer cancel()
+
+	//create Power VS DHCP Client
+	dhcpClient := instance.NewIBMPIDhcpClient(ctx, c.PISession, svcInsID)
+	//Get all DHCP Services
+	dhcpServices, err := dhcpClient.GetAll()
+	if err != nil {
+		return errors.Wrap(err, "failed to get DHCP service details")
+	}
+	if len(dhcpServices) > 0 {
+		return fmt.Errorf("DHCP service already exists for provided cloud instance")
+	}
+	return nil
+}
+
+//ValidateCloudConnectionInPowerVSRegion counts cloud connection in PowerVS Region
+func (c *BxClient) ValidateCloudConnectionInPowerVSRegion(ctx context.Context, svcInsID string) error {
+	ctx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+	defer cancel()
+	var cloudConnectionsIDs []string
+	cloudConnectionClient := instance.NewIBMPICloudConnectionClient(ctx, c.PISession, svcInsID)
+
+	//check number of cloudconnections
+	getAllResp, err := cloudConnectionClient.GetAll()
+	if err != nil {
+		return errors.Wrap(err, "failed to get existing Cloud connection details")
+	}
+
+	if len(getAllResp.CloudConnections) >= 2 {
+		return fmt.Errorf("cannot create new Cloud connection in Power VS. Only two Cloud connections are allowed per zone")
+	}
+
+	for _, cc := range getAllResp.CloudConnections {
+		cloudConnectionsIDs = append(cloudConnectionsIDs, *cc.CloudConnectionID)
+	}
+
+	//check for Cloud connection attached to DHCP Service
+	for _, cc := range cloudConnectionsIDs {
+		cloudConn, err := cloudConnectionClient.Get(cc)
+		if err != nil {
+			return errors.Wrap(err, "failed to get Cloud connection details")
+		}
+		if cloudConn != nil {
+			for _, nw := range cloudConn.Networks {
+				if nw.DhcpManaged {
+					return fmt.Errorf("only one Cloud connection can be attached to any DHCP network per account per zone")
+				}
+			}
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/asset/installconfig/powervs/session.go
```diff
@@ -182,6 +184,37 @@ func (c *BxClient) ValidateAccountPermissions() error {
...
+//ValidateDhcpServiceInstance checks for existing dhcp instance for the provided PowerVS cloud instance
+func (c *BxClient) ValidateDhcpServiceInstance(svcInsID string) error {
+
+	//check DHCP serviceInstance
+	dhcpClient := instance.NewIBMPIDhcpClient(context.Background(), c.PISession, svcInsID)
+	dhcpServers, err := dhcpClient.GetAll()
+	if err != nil {
+		return err
+	}
+	// only one dhcp server is allowed per cloud instance
```
clnperez, line 196: same comment here. "only one DHCP service is allowed..."

clnperez (changes requested):

pkg/asset/installconfig/powervs/session.go
```diff
@@ -182,6 +185,67 @@ func (c *BxClient) ValidateAccountPermissions() error {
...
+		}
+		res <- len(getAllResp.CloudConnections)
+		close(res)
+		return nil
+	}()
+	for {
+		select {
+		case noOfCloudConnections := <-res:
+			if noOfCloudConnections >= 1 {
+				return fmt.Errorf("only one cloudconnection with dhcpservice is allowed per Zone")
```
clnperez, line 240: 1) I don't see where this checks that there is a DHCP service associated with this Cloud connection. The existence of another Cloud connection on its own isn't a problem.
2) Cloud connection is two words. All of our documentation capitalizes the first word but not the second: https://cloud.ibm.com/docs/power-iaas?topic=power-iaas-cloud-connections
3) DHCP service is two words. DHCP should always be capitalized as it is an acronym. 
4) Zone doesn't need to be capitalized. 
5) The error isn't exactly clear to an end user. Change this to something like, "only one Cloud connection can be attached to any DHCP network per account per zone."

pkg/asset/quota/quota.go
```diff
@@ -144,7 +145,26 @@ func (a *PlatformQuotaCheck) Generate(dependencies asset.Parents) error {
...
+			return errors.Wrap(err, "failed to create bluemix client")
+		}
+
+		err = bxCli.NewPISession()
+		if err != nil {
+			return errors.Wrap(err, "failed to get PowerVS connection details")
+		}
+		err = bxCli.ValidateCloudConnectionInPowerVSRegion(context.TODO(), ic.Config.Platform.PowerVS.ServiceInstanceID)
+		if err != nil {
+			return errors.Wrap(err, "failed to meet the prerequisite for cloudConnection")
```
clnperez, line 160: same issues with the grammar here. Cloud connections

pkg/asset/quota/quota.go
```diff
@@ -144,7 +145,26 @@ func (a *PlatformQuotaCheck) Generate(dependencies asset.Parents) error {
...
+		}
+
+		err = bxCli.ValidateDhcpService(context.TODO(), ic.Config.Platform.PowerVS.ServiceInstanceID)
+		if err != nil {
+			return errors.Wrap(err, "failed to meet the prerequisite of dhcpservice per Power VS instance")
```
clnperez, line 165: same here as well for DHCP

clnperez (changes requested):
the code itself is fine, but the messages exposed to the user need to be cleaned up

pkg/asset/installconfig/powervs/session.go
```diff
@@ -182,6 +185,67 @@ func (c *BxClient) ValidateAccountPermissions() error {
 	return nil
 }
 
+//ValidateDhcpService checks for existing dhcp instance for the provided PowerVS cloud instance
+func (c *BxClient) ValidateDhcpService(ctx context.Context, svcInsID string) error {
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, helpers.PIGetTimeOut)
+	defer cancel()
+	res := make(chan int)
+
+	go func() error {
```
barbacbd, line 194: With the use of context, this probably doesn't need to be an extra go routine.

pkg/asset/installconfig/powervs/session.go
```diff
@@ -182,6 +185,67 @@ func (c *BxClient) ValidateAccountPermissions() error {
...
+		//create Power VS DHCP Client
+		dhcpClient := instance.NewIBMPIDhcpClient(ctxWithTimeout, c.PISession, svcInsID)
+		//Get all DHCP Services
+		dhcpServers, err := dhcpClient.GetAll()
+		if err != nil {
+			return err
```
barbacbd, line 200: Wrap the errors that come from other APIs

pkg/asset/installconfig/powervs/session.go
```diff
@@ -182,6 +185,67 @@ func (c *BxClient) ValidateAccountPermissions() error {
...
+			return nil
+		case <-ctxWithTimeout.Done():
+			return ctxWithTimeout.Err()
+		}
+	}
+}
+
+//ValidateCloudConnectionInPowerVSRegion counts cloud connection in PowerVS Region
+func (c *BxClient) ValidateCloudConnectionInPowerVSRegion(ctx context.Context, svcInsID string) error {
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, helpers.PIGetTimeOut)
```
barbacbd, line 221: Generally the child context takes on the name ctx also. More of nit picks but it is consistent in our code. Also can look into using a specific time like 60*time.Second or whatever the equivalent is for PIGetTimeOut

pkg/asset/installconfig/powervs/session.go
bkhadars: Done

pkg/asset/installconfig/powervs/session.go
```diff
@@ -182,6 +184,59 @@ func (c *BxClient) ValidateAccountPermissions() error {
...
+	getAllResp, err := cloudConnectionClient.GetAll()
+	if err != nil {
+		return errors.Wrap(err, "failed to get existing Cloud connection details")
+	}
+	for _, cc := range getAllResp.CloudConnections {
+		cloudConnectionsIDs = append(cloudConnectionsIDs, *cc.CloudConnectionID)
+	}
+
+	//check for DHCP Service attached to Cloud connection
+	if len(cloudConnectionsIDs) >= 1 {
```
clnperez, line 222: can you omit this check? i could be wrong but i think that the range loop just won't do anything if there's nothing there, so you don't need this

bkhadars: @clnperez As discussed I have removed the above check and added the check to verify number of cloud connection count per zone..

pkg/asset/installconfig/powervs/session.go
```diff
@@ -182,6 +184,62 @@ func (c *BxClient) ValidateAccountPermissions() error {
...
+	cloudConnectionClient := instance.NewIBMPICloudConnectionClient(ctx, c.PISession, svcInsID)
+
+	//check number of cloudconnections
+	getAllResp, err := cloudConnectionClient.GetAll()
+	if err != nil {
+		return errors.Wrap(err, "failed to get existing Cloud connection details")
+	}
+
+	if len(getAllResp.CloudConnections) >= 2 {
+		return fmt.Errorf("cannot create new cloud connection in powervs zone,only 2 cloud connections allowed")
```
clnperez, line 219: i'm sorry to keep picking on grammar, but this is a run-on sentence and needs proper nouns to be capitalized. "cannot create new Cloud connection in Power VS. Only two Cloud connections are allowed per zone."

PR #5900, opened 2022-05-12
WIP: [RFE-2181] To support transparent proxy configuration 

TrilokGeer:
Added configType support during installation to signify transparent or explicit proxy configuration to enable user-ca-bundle configured with cluster proxy and determine configuration type for cluster-wide proxy.

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -2935,8 +2935,7 @@ spec:
               configType:
                 description: Config type is the proxy configuration type. The valid values
                   are transparent or explicit
-                format: int32
-                type: integer
+                type: string
                 enum:
                   - explicit
                   - transparent
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -2932,6 +2932,14 @@ spec:
             description: Proxy defines the proxy settings for the cluster. If unset,
               the cluster will not be configured to use a proxy.
             properties:
+              configType:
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -210,4 +210,4 @@ replace sigs.k8s.io/controller-tools => sigs.k8s.io/controller-tools v0.3.1-0.20
 
 replace github.com/terraform-providers/terraform-provider-nutanix => github.com/nutanix/terraform-provider-nutanix v1.2.2-0.20211029075448-e21f85ac2cf7
 
-replace github.com/openshift/api => github.com/TrilokGeer/api v0.0.2
+replace github.com/openshift/api => github.com/TrilokGeer/api v0.0.5
```
barbacbd, line 213: Could you explain the requirement to add your local api ?

pkg/asset/manifests/proxy.go
```diff
@@ -2,6 +2,7 @@ package manifests
 
 import (
 	"fmt"
+	"github.com/openshift/installer/pkg/types/openstack"
```
barbacbd, line 5: this import needs to be moved to the correct location. Looks like you need to simply squash this change.

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -2935,8 +2935,7 @@ spec:
               configType:
                 description: Config type is the proxy configuration type. The valid values
                   are transparent or explicit
-                format: int32
-                type: integer
+                type: string
```
barbacbd, line 2938: This is a simple change to a file that you already changed. Please squash this commit.

PR #5901, opened 2022-05-12
Power VS: Separating bootstrap resources to own stage for clean destroy

miyamotoh:
As eluded to in #5780, this is to move the bootstrap resource creation to its own TF stage so it can be called for quick `terraform destory` after `bootstrap-complete` for PowerVS platform. Also introduced is the `bootstrap-routing` stage which takes care of the necessary network routing components that live beyond the bootstrap phase, but need adjustments when bootstrap VM is around and gone.


patrickdillon:
@miyamotoh required tf-fmt test is failing

clnperez:
/lgtm

my only comment is that, the name `bootstrap-routing` is a bit confusing (to me at least) b/c it seems to imply that the things from this phase will be deleted, so i'd like to see it renamed to something like `loadbalancers`. but that's just a nit and we can update it later

Prashanth684:
> /lgtm
> 
> my only comment is that, the name `bootstrap-routing` is a bit confusing (to me at least) b/c it seems to imply that the things from this phase will be deleted, so i'd like to see it renamed to something like `loadbalancers`. but that's just a nit and we can update it later

maybe call it `post-bootstrap` like gcp does

PR #5903, opened 2022-05-12
Azure provider: only validate premiumio when specified in install con…

kwoodson:
…fig.

This PR is to address BZ https://bugzilla.redhat.com/show_bug.cgi?id=2084441.  In order to address API calls that might not include their vm capabilities, I suggest only attempting to validate the capability for `Premium_LRS` when specified in the install config. This gives the ability to attempt best effort validation. If the value is not found in the capability then we create a log message and proceed on with the installation.

### DIFF
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -13,6 +13,7 @@ import (
 	aznetwork "github.com/Azure/azure-sdk-for-go/profiles/2018-03-01/network/mgmt/network"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"github.com/openshift/installer/pkg/types"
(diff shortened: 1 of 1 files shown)


r4f4 (commented):
I wonder if we should add unit tests to cover this case

kwoodson:
@MayXuQQ This is my proposal for the fix for the PremiumIO issue. [BZ ](https://bugzilla.redhat.com/show_bug.cgi?id=2084441)

MayXuQQ:
[[IPI on azure]fail to check the vm capabilities in private cluster](https://bugzilla.redhat.com/show_bug.cgi?id=2084441) fixed , check with registry.build01.ci.openshift.org/ci-ln-kn91cg2/release:latest (067098)

kwoodson:
@patrickdillon Mind reviewing? Looks like May has verified this fix. I can update the title for a bz if that helps, thanks.

r4f4:
I completely forgot about this and opened https://github.com/openshift/installer/pull/5924 to solve the same problem :man_facepalming:

r4f4:
> @r4f4 That's fine, should I close this PR?

If that's ok with you. The other one ended up getting the needed labels. Feel free to leave any comments/concerns on #5924

PR #5905, opened 2022-05-12
Bug 2094902: Simplify cross-compiling

yselkowitz:
- Always run go generate natively
- Add GOOS_GOARCH to all terraform build paths

### DIFF
--- a/hack/build.sh
+++ b/hack/build.sh
@@ -56,7 +56,8 @@ release)
 	TAGS="${TAGS} release"
 	if test "${SKIP_GENERATION}" != y
 	then
-		go generate ./data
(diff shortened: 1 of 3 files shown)


yselkowitz:
@jhixson74 thanks for the review.  Since we're now past feature freeze, should I file a BZ to get this in 4.11, or does this need to wait until 4.12?

yselkowitz:
pending the previous question
/hold

patrickdillon:
> @jhixson74 thanks for the review. Since we're now past feature freeze, should I file a BZ to get this in 4.11, or does this need to wait until 4.12?

I think we can put this in 4.11 with a BZ. We might be interested in backporting too.

PR #5906, opened 2022-05-13
AGENT-131:  Ignition and iso assets for agent installer

rwsu:
The Ignition asset embeds the asset/ignition/bootstrap/Common type.
The Ignition assets fills in the SSHPublicKey and the manifests that
are currently available. Much of the ignition data will be populated
in a future PR.
    
The ISO asset writes to agent-installer.iso leaving the fleeting.iso
unchanged. The ISO asset will replace fleeting.iso when they are
equivalent.

Comment out support for AgentClusterInstall asset until PR#5890
merges.


pkg/asset/agent/iso.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	ignitionContent := &isoeditor.IgnitionContent{Config: ignitionByte}
+	custom, err := isoeditor.NewRHCOSStreamReader(baseImage, ignitionContent, nil)
+	if err != nil {
+		return err
+	}
+	defer custom.Close()
+
+	output, err := os.Create(agentISOFilename)
```
zaneb, line 56: Hmm, I hadn't thought about the fact that the ISO is absolutely massive and there is no way we can save it in memory and let the asset management code write it to disk for us.

pkg/asset/agent/iso.go
```diff
@@ -0,0 +1,92 @@
...
+	iso, err := f.FetchByName(agentISOFilename)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, errors.Wrap(err, fmt.Sprintf("failed to load %s file", agentISOFilename))
+	}
+
+	i.File = iso
+	return true, nil
```
zaneb, line 83: There's probably no point implementing `Load()` - nothing depends on this, so we are never going to read an ISO loaded from disk, and in any event `Generate()` doesn't produce a file returnable from `Files()`. We're really only implementing this because the target has to take a list of WritableAssets, not any Assets. It's safe to always just `return false, nil`.

pkg/asset/agent/iso.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Files returns the files generated by the asset.
+func (a *ISO) Files() []*asset.File {
+	if a.File != nil {
+		return []*asset.File{a.File}
```
zaneb, line 89: Likewise, I _think_ we can always return an empty list here.

pkg/asset/agent/iso.go
```diff
@@ -0,0 +1,92 @@
...
+
+// Name returns the human-friendly name of the asset.
+func (a *ISO) Name() string {
+	return "Agent Installer ISO"
+}
+
+// Load returns the ISO from disk.
+func (i *ISO) Load(f asset.FileFetcher) (bool, error) {
+
+	iso, err := f.FetchByName(agentISOFilename)
```
zaneb, line 74: `iso` here contains a `[]byte`, so it's basically 1GiB in size.

pkg/asset/agent/ignition.go
```diff
@@ -0,0 +1,67 @@
...
+		&am.AgentManifests{},
+		// &am.AgentClusterInstall{},
+	}
+}
+
+// Generate generates the agent installer ignition.
+func (a *Ignition) Generate(dependencies asset.Parents) error {
+	agentManifests := &am.AgentManifests{}
+	// aciAsset := &am.AgentClusterInstall{}
+	dependencies.Get(agentManifests)
```
zaneb, line 37: I think it's too early to try to implement this - we inadvertently started implementing assets in the wrong order, from the bottom up when they can only be integrated from the top down, and as a result we're saddling ourselves with large amounts of untested code that will have to be integrated in one big bang.

Instead I think we should just do:

```
        configBuilder, err := imagebuilder.New()
        if err != nil {
                return err
        }

        ignition, err := configBuilder.Ignition()
        if err != nil {
                return err
        }

        a.Ignition = ignition
        return nil
```

that way you could immediately remove the PostRun function from the agentImageTarget and the code in this PR would be fully tested (including by CI when we have it). Currently it's not because whatever ISO gets written, it will be overridden by the one written when we call `agentcmd.BuildImage()` in the PostRun function.

Definitely save this implementation somewhere because we will need it for https://issues.redhat.com/browse/AGENT-181 and/or https://issues.redhat.com/browse/AGENT-133.

pkg/asset/agent/ignition.go
```diff
@@ -0,0 +1,67 @@
...
+	am "github.com/openshift/installer/pkg/asset/agent/manifests"
+	"github.com/openshift/installer/pkg/asset/ignition"
+	"github.com/openshift/installer/pkg/asset/ignition/bootstrap"
+)
+
+const manifestPathInIso = "/etc/assisted/manifests"
+
+// Ignition is an asset that generates the agent installer ignition file.
+type Ignition struct {
+	bootstrap.Common
```
zaneb, line 17: If we're going to use this type we should do it via composition rather than embedding. We don't want to inherit any of the implementation from it - as it is, this makes `Ignition` a `WritableAsset` with an implementation of `Files()` that we don't want.

pkg/asset/agent/image/iso.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/openshift/installer/pkg/agent/isosource"
+	"github.com/openshift/installer/pkg/asset"
+)
+
+const (
+	agentISOFilename = "output/fleeting.iso"
+)
+
+// ISO is an asset that generates the bootable image used to install clusters.
+type ISO struct {
```
zaneb, line 18: I'd suggest AgentImage as the name, since this may not always be an ISO forever.

pkg/asset/agent/image/ignition.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	ignitionBytes, err := configBuilder.Ignition()
+	if err != nil {
+		return err
+	}
+
+	ignition := &igntypes.Config{}
+
+	err = json.Unmarshal(ignitionBytes, ignition)
```
zaneb, line 40: Why unmarshal into a struct only to marshal back to text again in the next Asset?

zaneb (commented):
I believe we can delete these files now too:

- `pkg/agent/build_image.go`
- `pkg/agent/imagebuilder/embed_ignition.go`

/approve

pkg/asset/agent/image/ignition.go
rwsu: I was split between doing that or breaking out content.Ignition() to return just the config. I'll convert it to the later.

zaneb: I meant that Ignition.Config could be a `[]bytes` and then no conversion would be required.
This won't hurt anything though.

cmd/openshift-install/agent.go
```diff
@@ -44,22 +42,10 @@ var (
 			Use:   "image",
 			Short: "Generates a bootable image containing all the information needed to deploy a cluster",
 			Args:  cobra.ExactArgs(0),
-			PostRun: func(cmd *cobra.Command, args []string) {
-				cleanup := setupFileHook(rootOpts.dir)
```
andfasano, line 48: I'm not fully sure we could easily throw away this part as part of the command execution flow

zaneb:
/lgtm

As is traditional, this change touches one file (`cmd/openshift-install/agent.go`) that needs approval from the installer-approvers.
/assign @jhixson74

PR #5904, opened 2022-05-12
Azure console logs gathering

r4f4:
https://github.com/openshift/installer/pull/5807 should merge first

PR #5807, opened 2022-04-13
Aws console logs

r4f4:
Depends on https://github.com/openshift/installer/pull/5697

https://issues.redhat.com/browse/CORS-1945

PR #5907, opened 2022-05-13
Gather Console Logs for GCP & Azure

patrickdillon:
This rebases #5902 and #5904 after #5807 merged. 

/cc @r4f4 

/approve


patrickdillon:
I added b216585ecf3f15a70a7aa3ec0e8d931d698d10b7 as a way of getting around a build error. It seems like these imports are unused so it is fine to remove them, although I'm not sure I see where exactly the gathers are running `init()` and getting registered. So I would appreciate a second opinion that this is the correct fix.

patrickdillon:
/close
Hm CI is having a build error which I thought I fixed locally and now cannot reproduce locally. Not sure why it is happening in CI and it is late. Will check it out in the morning if I have time.

r4f4:
> I added [b216585](https://github.com/openshift/installer/commit/b216585ecf3f15a70a7aa3ec0e8d931d698d10b7) as a way of getting around a build error. It seems like these imports are unused so it is fine to remove them, although I'm not sure I see where exactly the gathers are running `init()` and getting registered. So I would appreciate a second opinion that this is the correct fix.

These imports are needed so the `init()` inside `pkg/gather/$platform/register.go` can run.

PR #5911, opened 2022-05-13
[vsphere] support for multi-zone/region installation

rvanderp3:
The intention of this PR is to implement the changes to the platform specification for vsphere as defined in https://github.com/openshift/enhancements/pull/918


rvanderp3:
cc @jcpowermac

jcpowermac:
LGTM

pkg/types/vsphere/platform.go
```diff
@@ -72,4 +107,207 @@ type Platform struct {
...
+	// tagCategory is the category used for the tag
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=80
+	TagCategory string `json:"tagCategory"`
+}
+
+// FailureDomainSpec holds the region and zone failure domain and
+// the vCenter topology of that failure domain.
+type FailureDomainSpec struct {
```
patrickdillon, line 156: I left a comment on the enhancement about this, because it seems more appropriate to discuss there: https://github.com/openshift/enhancements/pull/918/files#r933507865

But generally I would lean in favor of removing the `Spec` suffix from all of these variable names.

pkg/types/vsphere/validation/machinepool.go
```diff
@@ -1,13 +1,14 @@
 package validation
 
 import (
+	"github.com/openshift/installer/pkg/validate"
```
patrickdillon, line 4: nit: this should be grouped in with the next block

pkg/asset/machines/vsphere/machines.go
```diff
@@ -3,15 +3,15 @@ package vsphere
 
 import (
 	"fmt"
+	"regexp"
 
 	machineapi "github.com/openshift/api/machine/v1beta1"
+	"github.com/openshift/installer/pkg/types"
+	"github.com/openshift/installer/pkg/types/vsphere"
```
patrickdillon, line 10: nit: these imports should go in there own block, we try to group them:

<stdlib>

<external>

<internal>

We need to document this/enforce this

pkg/asset/machines/vsphere/machines.go
```diff
@@ -22,15 +22,48 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
...
-	total := int64(1)
+	azs := mpool.Zones
+	numOfAZs := len(azs)
+	definedZones := make(map[string]*vsphere.Platform)
+
+	if numOfAZs > 0 {
+		for _, az := range azs {
+			for _, deploymentZone := range platform.DeploymentZones {
+				if az == deploymentZone.Name {
+					if deploymentZone.ControlPlane == "NotAllowed" {
```
patrickdillon, line 39: Let's use the `const` here.

pkg/asset/machines/vsphere/machinesets.go
```diff
@@ -74,7 +65,136 @@ func MachineSets(clusterID string, config *types.InstallConfig, pool *types.Mach
...
+	return mset, nil
+}
+
+func getVCenterFromServerName(server string, platformSpec *vsphere.Platform) (*vsphere.VCenterSpec, error) {
+	for _, vCenter := range platformSpec.VCenters {
+		if vCenter.Server == server {
+			return &vCenter, nil
+		}
+	}
+	return nil, errors.Errorf("%s is not a defined vCenter server", server)
```
patrickdillon, line 77: Do we ever expect this to fail? Should this be part of the validation?

pkg/asset/manifests/cloudproviderconfig.go
```diff
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 
 	"github.com/ghodss/yaml"
+	vspheremanifests "github.com/openshift/installer/pkg/asset/manifests/vsphere"
```
patrickdillon, line 10: this shouldn't be moved

data/data/manifests/openshift/cloud-creds-secret.yaml.template
```diff
@@ -37,8 +37,10 @@ data:
   clouds.yaml: {{.CloudCreds.OpenStack.Base64encodeCloudCreds}}
   clouds.conf: {{.CloudCreds.OpenStack.Base64encodeCloudCredsINI}}
 {{- else if .CloudCreds.VSphere}}
-  {{.CloudCreds.VSphere.VCenter}}.username: {{.CloudCreds.VSphere.Base64encodeUsername}}
-  {{.CloudCreds.VSphere.VCenter}}.password: {{.CloudCreds.VSphere.Base64encodePassword}}
+{{- range .CloudCreds.VSphere}}
+  {{.VCenter}}.username: {{.Base64encodeUsername}}
```
patrickdillon, line 41: The server name is not being included here... is it needed?

patrickdillon (commented):
This looks really good. I did a first pass through and left some comments. There's a lot here, so I will do another pass soon. 

In general, can you update the commit messages according to [the contributing guidelines ](https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#commit-message-format)

And also we try to organize imports as:
<stdlib>

<external>

<internal>

This is something we need to document and enforce through tests. 

Like I said, I will be doing another review asap.

pkg/types/vsphere/validation/platform.go
```diff
@@ -56,6 +53,232 @@ func ValidatePlatform(p *vsphere.Platform, fldPath *field.Path) field.ErrorList
...
+	}
+	if len(allErrs) > 0 {
+		// if failureDomains and deploymentZones don't exist, this will cascade to checks to follow
+		return allErrs
+	}
+
+	// diskType is optional, but if provided should pass validation
+	if len(p.DiskType) != 0 {
+		allErrs = append(allErrs, validateDiskType(p, fldPath)...)
+	}
```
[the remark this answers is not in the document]
rvanderp3: yep! this is duplicated after a recent refactor.  will take care of that.

pkg/types/vsphere/validation/platform.go
```diff
@@ -56,6 +53,232 @@ func ValidatePlatform(p *vsphere.Platform, fldPath *field.Path) field.ErrorList
...
+func validateFailureDomain(failureDomain *vsphere.FailureDomain, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if len(failureDomain.Name) == 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("name"), "must specify the name"))
+	}
+
+	switch failureDomain.Type {
+	case "ComputeCluster":
+	case "Datacenter":
+	case "HostGroup":
```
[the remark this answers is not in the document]
rvanderp3: yeah, that is much better, updating

pkg/asset/machines/vsphere/machinesets.go
rvanderp3: yep, it is [checked](https://github.com/openshift/installer/pull/5911/commits/e81b2abbe82a163823d7920b78bbae581f1caba8#diff-2cdb5d09122e1d26502732ec3ee5fb818c7153af73b55c95cc5f41c9ae005febR217-R220) during validation as well.  i'll strike the error from here.

data/data/manifests/openshift/cloud-creds-secret.yaml.template
rvanderp3: .VCenter resolves to the server name.  the goal here was to produce the credentials secret in such a way that would allow multiple vCenters to be defined.

rvanderp3:
/hold
placing hold until tests have a chance to run.

rvanderp3:
/hold cancel

the e2e-vsphere-zones failure is expected due to the missing terraform bits which are coming in a follow on PR.  However, confirmed that the API came up and the install-config was processed corrected.  Additionally e2e-vsphere passed confirming that non-multi-zone installs are still successful.

Thanks for reviewing this @patrickdillon ! Would you mind giving this another look when you get a chance?

cc: @jcpowermac

pkg/types/vsphere/validation/platform.go
```diff
@@ -56,6 +53,227 @@ func ValidatePlatform(p *vsphere.Platform, fldPath *field.Path) field.ErrorList
...
+
+	// If all VIPs are empty, skip IP validation.  All VIPs are required to be defined together.
+	if strings.Join([]string{p.APIVIP, p.IngressVIP}, "") != "" {
+		allErrs = append(allErrs, validateVIPs(p, fldPath)...)
+	}
+
+	return allErrs
+}
+
+func validateMultiVCenter(p *vsphere.Platform, fldPath *field.Path) field.ErrorList {
```
patrickdillon, line 68: The enhancement says:
> Note: Only a single vCenter will
be supported by this effort. While the out-of-tree CCM supports multiple vCenters
the out-of-tree CSI does not.

(sorry, can't find a good way to link the line in the pr)

IIUC it seems like we should add validation to throw an error `if len(p.VCenters) > 1`

patrickdillon (commented):
Still working through this, but found one question, and then posted some more feedback/questions on the enhancement.

pkg/types/vsphere/validation/platform.go
rvanderp3: yeah, good call out there.  i'll add that validation

rvanderp3: validation added:
https://github.com/openshift/installer/pull/5911/commits/cc00cab911ec2dd6311b87322857ce29ffae6587#diff-2cdb5d09122e1d26502732ec3ee5fb818c7153af73b55c95cc5f41c9ae005febR103-R105
https://github.com/openshift/installer/pull/5911/commits/cc00cab911ec2dd6311b87322857ce29ffae6587#diff-54ccad1462fee98b54d320708c101c5094144e30cf0d9700a44dc8d332898c22R260-R270

pkg/types/vsphere/validation/machinepool.go
```diff
@@ -35,5 +36,28 @@ func ValidateMachinePool(p *vsphere.MachinePool, fldPath *field.Path) field.Erro
...
+					zoneDefined = true
+				}
+			}
+			if zoneDefined == false {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("zones"), zone, "zone not defined in deploymentZones"))
+			}
+		}
+	} else if len(platform.DeploymentZones) > 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("zones"), "zones must be defined if deploymentZones are defined"))
+	}
```
patrickdillon, line 61: Would it be a better UX, if `deploymentZones` were defined but no zones, the machinesets were distributed over the defined `deploymentZones`? I imagine that would be the common use case, and would simplify the config. Then `zones` would only need to be defined in more specific configurations.

pkg/asset/machines/vsphere/machines.go
```diff
@@ -22,15 +23,49 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
...
-	total := int64(1)
+	azs := mpool.Zones
+	numOfAZs := len(azs)
+	definedZones := make(map[string]*vsphere.Platform)
+
+	if numOfAZs > 0 {
+		for _, az := range azs {
+			for _, deploymentZone := range platform.DeploymentZones {
+				if az == deploymentZone.Name {
+					if deploymentZone.ControlPlane == vsphere.NotAllowed {
```
patrickdillon, line 40: Hm I'm not sure about this. Two things:

1. Don't both roles (`compute` & `master`) go through this code path? So I suspect this would incorrectly fail if a zone with `ControlPlane == NotAllowed` were used for a compute machine pool.
2. I suspect this code should live in the machinepool validation code--not here.

pkg/types/vsphere/validation/machinepool.go
patrickdillon: `DeploymentSuitable` might add a wrinkle here. We would need to consider how to handle the control plane in an edge case where, say, all provided zones were `ControlPlane == NotAllowed`

pkg/asset/machines/vsphere/machines.go
rvanderp3: Machines is only used for rendering control plane nodes.  Compute nodes are rendered from [machinesets](https://github.com/openshift/installer/pull/5911/files/c41f2c10e2929213b597de909dea0fe77a8dd2b6..b14a7a5745406a924280ab9dc8cf0a4e851b5ab3).  I can look at pulling this in to the machinepool validation.

rvanderp3: its coming back to me now.  at the point of [machinepool validation](https://github.com/openshift/installer/pull/5911/commits/c41f2c10e2929213b597de909dea0fe77a8dd2b6#diff-898a198eb6ce7f9bbd3a05ff17e82ea2e5114d2b7b5694a523f64e99e17f8e5bR11), there isn't a readily available context indicating that a given pool is controlPlane or compute.

rvanderp3: oh ... wait, we know the pool name, and that is validated to ensure it matches controlPlane and compute pools.

rvanderp3: @patrickdillon changes have been committed which apply default zones based upon the machine pool role and what deploymentZones are defined.

PR #5914, opened 2022-05-16
Bug 2086936: vSphere: change socket back to cores

jcpowermac:
Virtual machines running on
vSphere should be configured for cores
over sockets for NUMA. The change
in #5841 missed that.

Changing vSphere back to using a single
socket and multiple cores.

### DIFF
--- a/pkg/asset/machines/worker.go
+++ b/pkg/asset/machines/worker.go
@@ -155,7 +155,7 @@ func defaultOvirtMachinePoolPlatform() ovirttypes.MachinePool {
 func defaultVSphereMachinePoolPlatform() vspheretypes.MachinePool {
 	return vspheretypes.MachinePool{
 		NumCPUs:           4,
-		NumCoresPerSocket: 1,
+		NumCoresPerSocket: 4,
 		MemoryMiB:         16384,
 		OSDisk: vspheretypes.OSDisk{
 			DiskSizeGB: decimalRootVolumeSize,


jcpowermac:
ptal
@patrickdillon 
@rna-afk 

https://blogs.vmware.com/performance/2017/03/virtual-machine-vcpu-and-vnuma-rightsizing-rules-of-thumb.html
and
#5001

jcpowermac:
In vSphere:

NumCPUs / NumCoresPerSocket = socket count

patrickdillon:
This is only affecting compute nodes, right? I believe this value is already `4` for control plane. The reason I felt comfortable with setting the `1` for `NumCoresPerSocket` is because that is the value used in CI: https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/vsphere/ipi-conf-vsphere-commands.sh#L57

I definitely defer to your judgment. Also, @rna-afk had this set to `4` before I questioned him about it.

patrickdillon:
/approve

@jcpowermac would you mind creating a bz for this?

jcpowermac:
> This is only affecting compute nodes, right? I believe this value is already `4` for control plane. The reason I felt comfortable with setting the `1` for `NumCoresPerSocket` is because that is the value used in CI: https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/conf/vsphere/ipi-conf-vsphere-commands.sh#L57

Yeah I should have updated it after #5001

PR #5915, opened 2022-05-16
Bug 2084280: Remove optional services from permissions check

rna-afk:
There are a few services that are not required for the installer
to run but are now causing the installer to fail if the permission
is not provided. Adding an optional set of permissions and warn
the user if these permissions are not provided.

List of required services:
compute - used for creating VMs
cloudresourcemanager - used for updating metadata for resources
iam, iamcredentials - used for creating iam accounts
dns - used for dns creation

### DIFF
--- a/pkg/asset/installconfig/gcp/validation.go
+++ b/pkg/asset/installconfig/gcp/validation.go
@@ -219,17 +219,19 @@ func validateMachineNetworksContainIP(fldPath *field.Path, networks []types.Mach
 }
 
 //ValidateEnabledServices gets all the enabled services for a project and validate if any of the required services are not enabled.
+//also warns the user if optional services are not enabled.
 func ValidateEnabledServices(ctx context.Context, client API, project string) error {
-	services := sets.NewString("compute.googleapis.com",
-		"cloudapis.googleapis.com",
+	requiredServices := sets.NewString("compute.googleapis.com",
 		"cloudresourcemanager.googleapis.com",
 		"dns.googleapis.com",
 		"iam.googleapis.com",
-		"iamcredentials.googleapis.com",
+		"iamcredentials.googleapis.com")
+	optionalServices := sets.NewString("cloudapis.googleapis.com",
 		"servicemanagement.googleapis.com",
-		"serviceusage.googleapis.com",
+		"deploymentmanager.googleapis.com",
 		"storage-api.googleapis.com",
-		"storage-component.googleapis.com")
+		"storage-component.googleapis.com",
+		"serviceusage.googleapis.com")
 	projectServices, err := client.GetEnabledServices(ctx, project)
 
 	if err != nil {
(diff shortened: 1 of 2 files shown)


patrickdillon:
Let's update this based on our slack discussion to make the list of required APIs more accurate.

jianli-wei (changes requested):

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -219,15 +219,17 @@ func validateMachineNetworksContainIP(fldPath *field.Path, networks []types.Mach
...
 func ValidateEnabledServices(ctx context.Context, client API, project string) error {
-	services := sets.NewString("compute.googleapis.com",
-		"cloudapis.googleapis.com",
+	requiredServices := sets.NewString("compute.googleapis.com",
 		"cloudresourcemanager.googleapis.com",
 		"dns.googleapis.com",
 		"iam.googleapis.com",
 		"iamcredentials.googleapis.com",
+		"deploymentmanager.googleapis.com",
+		"serviceusage.googleapis.com")
```
patrickdillon, line 230: What about `serviceusage`,  how are we determining this is required? Previously it was listed as `optional`...

patrickdillon (commented):
I have questions about the apis we're adding as required. 

For any changes we do make, can we update the commit message to have reasoning specific to each change we make.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -219,15 +219,17 @@ func validateMachineNetworksContainIP(fldPath *field.Path, networks []types.Mach
...
+//also warns the user if optional services are not enabled.
 func ValidateEnabledServices(ctx context.Context, client API, project string) error {
-	services := sets.NewString("compute.googleapis.com",
-		"cloudapis.googleapis.com",
+	requiredServices := sets.NewString("compute.googleapis.com",
 		"cloudresourcemanager.googleapis.com",
 		"dns.googleapis.com",
 		"iam.googleapis.com",
 		"iamcredentials.googleapis.com",
+		"deploymentmanager.googleapis.com",
```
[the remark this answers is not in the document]
rna-afk: All of them were added based on traffic. I think it's being used for UPI so I'll keep it as optional.

pkg/asset/installconfig/gcp/validation.go
rna-afk: I listed it as optional first but I think it's being used by the installer/google-sdk to get the list of APIs supported? If that is true then this might be the most important one.

patrickdillon: > I listed it as optional first but I think it's being used by the installer/google-sdk to get the list of APIs supported? If that is true then this might be the most important one for this logic of checking the APIs.

Ah thanks for explaining. This makes sense as to what the API is needed for but I think we should keep is as optional because the API check is optional and will be handled gracefully:

https://github.com/openshift/installer/blob/60ad1fe02453c00f5ff1a33f4e78e5ccc2dee13b/pkg/asset/installconfig/gcp/validation.go#L235-L244

So let's move it back to optional, if you agree.

rna-afk: Sounds good to me. I'll change it.

patrickdillon:
/approve
/lgtm

GCP tests are struggling but I tested this locally.

PR #5917, opened 2022-05-16
AGENT-180: Allow create image command to fetch manifests files from cluster-manifests assets folder

andfasano:
Currently the agent installer tries to fetch the required manifests from the `manifests` folder found in the present working directory. This patch introduces the capability to retrieve them from the `cluster-manifests` folder located in the assets directory configured when running the OpenShift Installer (via the `--dir` option)

### DIFF
--- a/cmd/openshift-install/agent.go
+++ b/cmd/openshift-install/agent.go
@@ -50,7 +50,7 @@ var (
 
 				timer.StartTimer("Agent Create Image Complete")
 
-				err := agentcmd.BuildImage()
+				err := agentcmd.BuildImage(rootOpts.dir)
 				if err != nil {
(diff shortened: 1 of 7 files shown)


cmd/openshift-install/agent.go
```diff
@@ -50,7 +50,7 @@ var (
 
 				timer.StartTimer("Agent Create Image Complete")
 
-				err := agentcmd.BuildImage()
+				err := agentcmd.BuildImage(rootOpts.dir)
```
zaneb, line 53: This will conflict with #5906. It would be better if we could build on top of that.

zaneb (commented):
I like this approach of doing the minimal thing to get manifests loaded from the right directory.

PR #5919, opened 2022-05-17
AGENT-194: Fix service name in dependencies

zaneb:
The create-cluster-and-infra-env service was renamed in
d87d402510aaa96553f46a415483a67d023b3dc2, but the dependencies on it
were not.


rwsu (approved):
/lgtm

PR #5922, opened 2022-05-17
Bug 1969794: Correct typo

stephenfin:
The created pvc needs to use a volumeMode of 'Filesystem', not 'Block'. Using the latter will result in the image-registry pod failing to mount the volume because it can't handle raw block devices without a filesystem [1].

[1] https://docs.openshift.com/container-platform/4.10/virt/virtual_machines/virtual_disks/virt-creating-data-volumes.html

### DIFF
--- a/docs/user/openstack/image_registry_with_custom_pvc_backends.md
+++ b/docs/user/openstack/image_registry_with_custom_pvc_backends.md
@@ -36,7 +36,7 @@ metadata:
 spec:
   accessModes:
   - ReadWriteOnce
-  volumeMode: Block
+  volumeMode: Filesystem
   resources:
     requests:
       storage: 100Gi


stephenfin:
> /lgtm
> 
> Does any external facing documentation need to change also?

What do you mean? Docs from docs.openshift.com? Those are already updated, if so.

mandre:
> > /lgtm
> > Does any external facing documentation need to change also?
> 
> What do you mean? Docs from docs.openshift.com? Those are already updated, if so.

Indeed, I just checked and upstream docs were fixed via https://github.com/openshift/openshift-docs/pull/34932.

mandre (commented):
/lgtm
/approve

sadasu:
> > > /lgtm
> > > Does any external facing documentation need to change also?
> > 
> > 
> > What do you mean? Docs from docs.openshift.com? Those are already updated, if so.
> 
> Indeed, I just checked and upstream docs were fixed via [openshift/openshift-docs#34932](https://github.com/openshift/openshift-docs/pull/34932).

Yes, that is what i was looking for. Thanks for providing the link.

PR #5925, opened 2022-05-17
BUG 2075459: IOPS was not being set even when manually configured

barbacbd:
aws: IOPS was not being set even when manually configured

** Removed the terraform parameter that was constantly set to 0 unless this
was an io1 instance.

** Left the bootstrap terraform value remain as it isn't causing any issues, but
we should be aware if we want to change the boostrap aws instance to anything
other than gp3 then we will change that too.

** Added static value tests for the volume type, volume size and iops
for AWS.

### DIFF
--- a/pkg/types/aws/machinepool.go
+++ b/pkg/types/aws/machinepool.go
@@ -78,8 +78,8 @@ func (a *MachinePool) Set(required *MachinePool) {
 
 // EC2RootVolume defines the storage for an ec2 instance.
 type EC2RootVolume struct {
-	// IOPS defines the amount of provisioned IOPS. This is only valid
-	// for type io1.
+	// IOPS defines the amount of provisioned IOPS. (KiB/s). IOPS may only be set for
+	// io1, io2, & gp3 volume types.
 	//
 	// +kubebuilder:validation:Minimum=0
 	// +optional
--- a/pkg/types/aws/validation/machinepool.go
+++ b/pkg/types/aws/validation/machinepool.go
@@ -38,15 +38,49 @@ func ValidateMachinePool(platform *aws.Platform, p *aws.MachinePool, fldPath *fi
 		}
 	}
 
-	if p.IOPS < 0 {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("iops"), p.IOPS, "Storage IOPS must be positive"))
-	}
-	if p.Size < 0 {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("size"), p.Size, "Storage size must be positive"))
+	if p.EC2RootVolume.Type != "" {
+		allErrs = append(allErrs, validateVolumeSize(p, fldPath)...)
+		allErrs = append(allErrs, validateIOPS(p, fldPath)...)
 	}
 
 	if p.EC2Metadata.Authentication != "" && !validMetadataAuthValues.Has(p.EC2Metadata.Authentication) {
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("authentication"), p.EC2Metadata.Authentication, "must be either Required or Optional"))
+	}
+
+	return allErrs
+}
+
+func validateVolumeSize(p *aws.MachinePool, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	volumeSize := p.EC2RootVolume.Size
+
+	if volumeSize <= 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("size"), volumeSize, "volume size value must be a positive number"))
+	}
+
+	return allErrs
+}
+
+func validateIOPS(p *aws.MachinePool, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	volumeType := strings.ToLower(p.EC2RootVolume.Type)
+	iops := p.EC2RootVolume.IOPS
+
+	switch volumeType {
+	case "io1", "io2":
+		if iops <= 0 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("iops"), iops, "iops must be a positive number"))
+		}
+	case "gp3":
+		if iops < 0 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("iops"), iops, "iops must be a positive number"))
+		}
+	case "gp2", "st1", "sc1", "standard":
+		if iops != 0 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("iops"), iops, fmt.Sprintf("iops not supported for type %s", volumeType)))
+		}
+	default:
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("type"), volumeType, fmt.Sprintf("failed to find volume type %s", volumeType)))
 	}
 
 	return allErrs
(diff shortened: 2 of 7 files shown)


barbacbd:
The following [link](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-volume-types.html) can be used to see where the static values came from.

pkg/types/aws/validation/machinepool.go
```diff
@@ -52,6 +113,40 @@ func ValidateMachinePool(platform *aws.Platform, p *aws.MachinePool, fldPath *fi
...
 }
 
+func validateVolumeSize(p *aws.MachinePool, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	volumeType := p.EC2RootVolume.Type
+	volumeSize := p.EC2RootVolume.Size
+
+	if volumeLimits, ok := validLimits[volumeType]; ok {
+		validSize := volumeLimits.VolumeSize
+		if volumeSize >= validSize.Max || volumeSize <= validSize.Min {
```
r4f4, line 123: So the Min, Max values themselves are not valid?

barbacbd: No but the static checks are actually being changed. The max size isn't quite that simple. There is a max size per region per instance, so there is no way to actually check that in the types package.

pkg/types/aws/validation/machinepool.go
```diff
@@ -52,6 +50,41 @@ func ValidateMachinePool(platform *aws.Platform, p *aws.MachinePool, fldPath *fi
...
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("iops"), iops, fmt.Sprintf("IOPS must be set for type %s", volumeType)))
+		}
+	case "gp3":
+		// Terraform will accept any value
+	case "gp2", "st1", "sc1", "standard":
+		if iops != 0 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("iops"), iops, fmt.Sprintf("IOPS must not be set for type %s", volumeType)))
+		}
+	default:
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("type"), volumeType, fmt.Sprintf("Failed to find volume type %s", volumeType)))
```
rna-afk, line 82: We do not use caps at the start of an error message since it's not the first word of the error message anyway due to the padding done to the message by the field.Invalid object.

pkg/types/aws/validation/machinepool.go
```diff
@@ -52,6 +50,41 @@ func ValidateMachinePool(platform *aws.Platform, p *aws.MachinePool, fldPath *fi
...
+	switch volumeType {
+	case "io1", "io2":
+		if iops <= 0 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("iops"), iops, fmt.Sprintf("IOPS must be set for type %s", volumeType)))
+		}
+	case "gp3":
+		// Terraform will accept any value
+	case "gp2", "st1", "sc1", "standard":
+		if iops != 0 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("iops"), iops, fmt.Sprintf("IOPS must not be set for type %s", volumeType)))
```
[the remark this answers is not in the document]
barbacbd: I would have chosen similar text, but decided to mirror error messages from Terraform. Do we still want to go ahead with these messages?

pkg/types/aws/validation/machinepool.go
```diff
@@ -52,6 +50,41 @@ func ValidateMachinePool(platform *aws.Platform, p *aws.MachinePool, fldPath *fi
 	return allErrs
 }
 
+func validateVolumeSize(p *aws.MachinePool, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	volumeType := strings.ToLower(p.EC2RootVolume.Type)
+	volumeSize := p.EC2RootVolume.Size
+
+	if volumeSize <= 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("size"), volumeSize, fmt.Sprintf("Storage must exist for type %s", volumeType)))
```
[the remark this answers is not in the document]
barbacbd: I would have chosen similar text, but decided to mirror error messages from Terraform. Do we still want to go ahead with these messages?

pkg/types/aws/validation/platform_test.go
```diff
@@ -128,11 +128,13 @@ func TestValidatePlatform(t *testing.T) {
...
 				DefaultMachinePlatform: &aws.MachinePool{
 					EC2RootVolume: aws.EC2RootVolume{
+						Type: "io1",
 						IOPS: -10,
+						Size: 128,
 					},
 				},
 			},
-			expected: `^test-path\.defaultMachinePlatform\.iops: Invalid value: -10: Storage IOPS must be positive$`,
+			expected: `^test-path.*iops: Invalid value: -10: iops must be set for type io1$`,
```
r4f4, line 137: Nit: this message looks a bit confusing here. The problem is not that the value must be set, because it is, but that it must be set to a positive value. Is it like this because it defaults to `0` when not set?

pkg/types/aws/validation/machinepool.go
```diff
@@ -52,6 +50,40 @@ func ValidateMachinePool(platform *aws.Platform, p *aws.MachinePool, fldPath *fi
 	return allErrs
 }
 
+func validateVolumeSize(p *aws.MachinePool, fldPath *field.Path) field.ErrorList {
```
r4f4, line 53: Nit: this function is so small/simple that it could maybe be integrated into the caller `if`.

r4f4:
/lgtm
I left some small comments but I'm nitpicking at this point.

pkg/types/aws/machinepool.go
```diff
@@ -78,8 +78,7 @@ func (a *MachinePool) Set(required *MachinePool) {
 
 // EC2RootVolume defines the storage for an ec2 instance.
 type EC2RootVolume struct {
-	// IOPS defines the amount of provisioned IOPS. This is only valid
-	// for type io1.
+	// IOPS defines the amount of provisioned IOPS. (KiB/s)
```
patrickdillon, line 81:
```suggestion
	// IOPS defines the amount of provisioned IOPS. (KiB/s). IOPS may only be set for io1, io2, & gp3 volume types.
```

patrickdillon (commented):
@barbacbd I believe you mentioned that there were some docs inconsistencies that needed to be corrected. Can you work with @mjpytlak (or ben scott, whose github username I don't know) to make these changes?

pkg/types/aws/machinepool.go
bscott-rh: @barbacbd I'm available to work on this - what docs need revising?

bscott-rh: > https://docs.openshift.com/container-platform/4.10/storage/dynamic-provisioning.html This is the first instance of the issue (not sure if there are more). In the section "AWS Elastic Block Store (EBS) object definition".
> 
> In #2 GP3 is a missing value. In #3 It says that iops is only allowed for io1. It is allowed for gp3 and io2. It is required for io1 and io2 to be a value greater than 0.

@lpettyjo Brent has identified some docs inconsistencies in the storage area. Could you take a look? unsure if we should have a separate BZ created or how it should be handled. Thank you

patrickdillon: @barbacbd I'd like to get this comment updated, because I don't think there's anywhere in the codebase that documents this behavior.

patrickdillon: (And after the comment is updated the explain code will need to be regenerated.)

PR #5927, opened 2022-05-18
Bug 2090487: SNO network type update

sadasu:
Change the default SNO networking from OpenShiftSDN to OVNKubernetes, which has the features (in particular, IPv6) required for a majority of the deployments.

In addition, also change the default networkType just for SNO deployments to OVNKubernetes.
Details in: https://issues.redhat.com/browse/CORS-2008

### DIFF
--- a/pkg/types/defaults/installconfig.go
+++ b/pkg/types/defaults/installconfig.go
@@ -25,7 +25,6 @@ var (
 	defaultClusterNetwork = ipnet.MustParseCIDR("10.128.0.0/14")
 	defaultHostPrefix     = 23
 	defaultNetworkType    = string(operv1.NetworkTypeOpenShiftSDN)
-	defaultOKDNetworkType = string(operv1.NetworkTypeOVNKubernetes)
 )
 
 // SetInstallConfigDefaults sets the defaults for the install config.
@@ -44,8 +43,8 @@ func SetInstallConfigDefaults(c *types.InstallConfig) {
 		}
 	}
 	if c.Networking.NetworkType == "" {
-		if c.IsOKD() {
-			c.Networking.NetworkType = defaultOKDNetworkType
+		if c.IsOKD() || c.IsSingleNodeOpenShift() {
+			c.Networking.NetworkType = string(operv1.NetworkTypeOVNKubernetes)
 		} else {
 			c.Networking.NetworkType = defaultNetworkType
 		}
--- a/pkg/types/installconfig.go
+++ b/pkg/types/installconfig.go
@@ -178,6 +178,12 @@ func (c *InstallConfig) IsOKD() bool {
 	return OKD
 }
 
+// IsSingleNodeOpenShift returns true if the install-config has been configured for
+// bootstrapInPlace
+func (c *InstallConfig) IsSingleNodeOpenShift() bool {
+	return c.BootstrapInPlace != nil
+}
+
 // Platform is the configuration for the specific platform upon which to perform
 // the installation. Only one of the platform configuration should be set.
 type Platform struct {
@@ -274,9 +280,9 @@ func (p *Platform) Name() string {
 
 // Networking defines the pod network provider in the cluster.
 type Networking struct {
-	// NetworkType is the type of network to install. The default is OpenShiftSDN
-	//
-	// +kubebuilder:default=OpenShiftSDN
+	// NetworkType is the type of network to install.
+	// The default value is OVNKubernetes for Single Node OpenShift
+	// and OpenShiftSDN for all other platforms.
 	// +optional
 	NetworkType string `json:"networkType,omitempty"`
 
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -106,7 +106,7 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
 		}
 	}
 	if c.Networking != nil {
-		allErrs = append(allErrs, validateNetworking(c.Networking, field.NewPath("networking"))...)
+		allErrs = append(allErrs, validateNetworking(c.Networking, c.IsSingleNodeOpenShift(), field.NewPath("networking"))...)
 		allErrs = append(allErrs, validateNetworkingIPVersion(c.Networking, &c.Platform)...)
 		allErrs = append(allErrs, validateNetworkingForPlatform(c.Networking, &c.Platform, field.NewPath("networking"))...)
 	} else {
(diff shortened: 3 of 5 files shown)


pkg/asset/ignition/bootstrap/bootstrap_in_place.go
```diff
@@ -75,5 +75,10 @@ func verifyBootstrapInPlace(installConfig *types.InstallConfig) error {
 		errorList = append(errorList, field.Required(field.NewPath("bootstrapInPlace", "installationDisk"),
 			"installationDisk must be set the target disk drive for the installation"))
 	}
+	if installConfig.BootstrapInPlace != nil && installConfig.Networking.NetworkType == "OpenShiftSDN" {
+		errorList = append(errorList, field.Invalid(field.NewPath("networking", "networkType"), installConfig.Networking.NetworkType,
+			"bootstrap in place does not support a NetworkType of OpenShiftSDN. Please use OVNKubernetes instead."))
```
romfreiman, line 80: Bootstrap in place is kinda of impdemetation details. Lets just write that sno does not support sdn. It will align with the docs

sadasu: The other error msgs in this validation method all use "bootstrap in place". If SNO is synonymous with bootstrap in place then maybe this an opportunity to reword all the other error msgs too?

pkg/asset/ignition/bootstrap/bootstrap_in_place.go
```diff
@@ -75,5 +75,10 @@ func verifyBootstrapInPlace(installConfig *types.InstallConfig) error {
 		errorList = append(errorList, field.Required(field.NewPath("bootstrapInPlace", "installationDisk"),
 			"installationDisk must be set the target disk drive for the installation"))
 	}
+	if installConfig.BootstrapInPlace != nil && installConfig.Networking.NetworkType == "OpenShiftSDN" {
+		errorList = append(errorList, field.Invalid(field.NewPath("networking", "networkType"), installConfig.Networking.NetworkType,
+			"Single Node Openshift does not support a NetworkType of OpenShiftSDN. Please use OVNKubernetes instead."))
+	}
```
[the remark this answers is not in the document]
romfreiman: yes, lets set a default :)

pkg/types/installconfig.go
```diff
@@ -178,6 +178,15 @@ func (c *InstallConfig) IsOKD() bool {
 	return OKD
 }
 
+// IsBootstrapInPlace returns true if the install-config has been configured for
+// Single Node OpenShift
+func (c *InstallConfig) IsBootstrapInPlace() bool {
+	if c.Platform.Name() == "none" && c.BootstrapInPlace != nil {
+		return true
+	}
+	return false
```
patrickdillon, line 187: nit: This could be simplified.
```suggestion
	return c.Platform.Name() == "none" && c.BootstrapInPlace != nil
```

sadasu: SNO deployments can happen on multiple platform types (None, vsphere and even Baremetal). So, the only valid check would be to check for the presence of BootstrapInPlace config.

pkg/types/defaults/installconfig.go
```diff
@@ -44,7 +44,7 @@ func SetInstallConfigDefaults(c *types.InstallConfig) {
 		}
 	}
 	if c.Networking.NetworkType == "" {
-		if c.IsOKD() {
+		if c.IsOKD() || c.IsSingleNodeOpenShift() {
 			c.Networking.NetworkType = defaultOKDNetworkType
```
patrickdillon, line 48: Sorry in advance to be nitpicky here, and I'm only leaving this comment because there was another change in this review. Also, the issue I'm complaining about here does not really come from this PR, but I think we should fix it:

I think the semantics here are bad with the variable `defaultOKDNetworkType` and it bothers me a little bit to assign the `defaultOKDNetworkType` to SNO, because I don't see a relationship between the networking type of OKD and SNO. I think we should just get rid of the `defaultOKDNetworkType` variable (I don't think it is widely used) and then:

```suggestion
		if c.IsOKD() || c.IsSingleNodeOpenShift() {
			c.Networking.NetworkType = string(operv1.NetworkTypeOVNKubernetes)
```

patrickdillon (commented):
a couple of small changes and we should be good

pkg/types/validation/installconfig.go
```diff
@@ -295,12 +295,16 @@ func validateNetworkingIPVersion(n *types.Networking, p *types.Platform) field.E
...
 
-func validateNetworking(n *types.Networking, fldPath *field.Path) field.ErrorList {
+func validateNetworking(n *types.Networking, singleNodeOpenShift bool, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 	if n.NetworkType == "" {
 		allErrs = append(allErrs, field.Required(fldPath.Child("networkType"), "network provider type required"))
 	}
 
+	if singleNodeOpenShift && n.NetworkType == string(operv1.NetworkTypeOpenShiftSDN) {
+		allErrs = append(allErrs, field.Required(fldPath.Child("networkType"), "Only NetworkType OVNKubernetes is currently supported on Single Node OpenShift"))
```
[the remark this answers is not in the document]
sadasu: Done.

pkg/types/defaults/installconfig.go
sadasu: Agree. I was reusing that variable and didn't pay too much attention to its name. It would be confusing to keep using the same variable name.

pkg/types/validation/installconfig.go
```diff
@@ -295,12 +295,16 @@ func validateNetworkingIPVersion(n *types.Networking, p *types.Platform) field.E
...
 
-func validateNetworking(n *types.Networking, fldPath *field.Path) field.ErrorList {
+func validateNetworking(n *types.Networking, singleNodeOpenShift bool, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 	if n.NetworkType == "" {
 		allErrs = append(allErrs, field.Required(fldPath.Child("networkType"), "network provider type required"))
 	}
 
+	if singleNodeOpenShift && n.NetworkType == string(operv1.NetworkTypeOpenShiftSDN) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("networkType"), n.NetworkType, "only NetworkType OVNKubernetes is currently supported on Single Node OpenShift"))
```
romfreiman, line 305: this is not accurate. CUs do inject cillium, callico .... I dont think we claim it's unsupported. 
I would rather say that SDN is not, and the default is OVN. Or somehting in between those lines.

romfreiman: This the message we present in assisted: https://github.com/openshift/assisted-service/pull/3419/files#diff-a712b20e2e375b0286d437906cfa73f4423a506449c19e71ba14633cd89dd01bR270
but i think stating explicitly about ovn as a default is better

sadasu: Ok. Thanks for pointing out that other networkTypes are supported. In this validation check, I have reworded the error to say that OpenShiftSDN is invalid but don't think it is the right place to say what the default value is. There are other sections that have already been updated to indicate the change in default.

PR #5931, opened 2022-05-19
Bug 2084580: Add check to validate cluster name for dots

rna-afk:
On prem clusters with a dot in the cluster name causes issues
in keepalived and should be validated to prevent this. Adding
an extra check to make sure there are no dots in the cluster
name.

### DIFF
--- a/pkg/validate/validate.go
+++ b/pkg/validate/validate.go
@@ -279,3 +279,11 @@ func Host(v string) error {
 	}
 	return validateSubdomain(v)
 }
+
+// OnPremClusterName verifies if the cluster name contains a '.' and returns an error if it does.
+func OnPremClusterName(v string) error {
+	if strings.Contains(v, ".") {
+		return errors.New("cluster name must not contain '.'")
+	}
+	return ClusterName(v)
+}
--- a/pkg/validate/validate_test.go
+++ b/pkg/validate/validate_test.go
@@ -49,6 +49,28 @@ func TestClusterName(t *testing.T) {
 	}
 }
 
+func TestOnPremClusterName(t *testing.T) {
+	cases := []struct {
+		name        string
+		clusterName string
+		valid       bool
+	}{
+		{"single lowercase", "a", true},
+		{"has a dot", "a.a", false},
+		{"valid name", "abcde", true},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			err := OnPremClusterName(tc.clusterName)
+			if tc.valid {
+				assert.NoError(t, err)
+			} else {
+				assert.Error(t, err)
+			}
+		})
+	}
+}
+
 func TestClusterName1035(t *testing.T) {
 	maxSizeName := "a" + strings.Repeat("123456789.", 5) + "123"
 
--- a/pkg/asset/installconfig/clustername.go
+++ b/pkg/asset/installconfig/clustername.go
@@ -48,6 +48,11 @@ func (a *clusterName) Generate(parents asset.Parents) error {
 			return validate.ClusterNameMaxLength(ans.(string), 14)
 		})
 	}
+	if platform.VSphere != nil || platform.BareMetal != nil {
+		validator = survey.ComposeValidators(validator, func(ans interface{}) error {
+			return validate.OnPremClusterName(ans.(string))
+		})
+	}
 	validator = survey.ComposeValidators(validator, func(ans interface{}) error {
 		installConfig := &types.InstallConfig{BaseDomain: bd.BaseDomain}
 		installConfig.ObjectMeta.Name = ans.(string)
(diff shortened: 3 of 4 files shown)


pkg/validate/validate.go
```diff
@@ -279,3 +279,11 @@ func Host(v string) error {
 	}
 	return validateSubdomain(v)
 }
+
+// VSphereClusterName verifies if the cluster name contains a '.' and returns an error if it does.
+func VSphereClusterName(v string) error {
```
kirankt, line 284: Can we add a test for this function?

kirankt (commented):
Looks good so far. Can we add a test to check the input for `VSphereClusterName` function?

jinyunma:
/label qe-approved
the verification is updated on https://bugzilla.redhat.com/show_bug.cgi?id=2084580#c2

pkg/validate/validate_test.go
```diff
@@ -49,6 +49,28 @@ func TestClusterName(t *testing.T) {
...
 
+func TestVSphereClusterName(t *testing.T) {
+	cases := []struct {
+		name        string
+		clusterName string
+		valid       bool
+	}{
+		{"single lowercase", "a", true},
+		{"has a dot", "a.a", false},
+		{"valid name", "abcde", true},
```
sadasu, line 60: Would be good to add a test where cluster name has other characters like "ab:cd-ef". I am assuming this is supported.

sadasu:
Can we update install-config [https://github.com/openshift/installer/blob/af6c11b063d4b6981db0c3e4f88d04ba192f080c/docs/user/customization.md#platform-customization] to document this constraint?

pkg/validate/validate_test.go
rna-afk: The main aim of the VsphereClusterName function is to just check the dot and then call ClusterName function. These cases should be tested in the TestClusterName function. I can add it there.

rna-afk:
I think [that document](https://github.com/openshift/installer/blob/af6c11b063d4b6981db0c3e4f88d04ba192f080c/docs/user/customization.md#platform-customization) is more of a common document and we should be mentioning this in a more vsphere specific docs. I checked the vsphere docs and there's no mention of cluster name. We need to think about documenting all the platforms in a separate PR then since I know a few constraints for cluster name like this in other platforms.

patrickdillon:
> We had concerns this issue could affect all baremetal & friends networking platforms. Those could be fixed in separate PRs if that's the case though.

@sadasu @rvanderp3 @jcpowermac wdyt? The bug states that the issue here is with keepalived. Should we add this check to prevent all clusters using keepalived from having a `.` in the name?

jcpowermac:
> > We had concerns this issue could affect all baremetal & friends networking platforms. Those could be fixed in separate PRs if that's the case though.
> 
> @sadasu @rvanderp3 @jcpowermac wdyt? The bug states that the issue here is with keepalived. Should we add this check to prevent all clusters using keepalived from having a `.` in the name?

Thats what I thought too when looking at this.
@cybertron

cybertron:
Yes, all of the on-prem IPI platforms are affected by this.

patrickdillon:
> Yes, all of the on-prem IPI platforms are affected by this.

@rna-afk Let's change this for all on prem. I tested on aws and it looks like the behavior is that the cluster name is accepted, but the `.` is converted to a `-` when creating infrastructure so let's not look at the other clouds.

patrickdillon:
> > but the `.` is converted to a `-` when creating infrastructure
> 
> Can we just do that for on-prem too? Replacing .'s in the cluster name would solve the underlying bug instead of working around it.

That ‘.’ -> ‘-‘ replacement is not happening in installer code (must be either terraform or aws). In general, in the installer we throw errors and let users decide how to fix their input, rather than quietly changing it.

patrickdillon:
Just fyi, I was wrong about:

> That ‘.’ -> ‘-‘ replacement is not happening in installer code (must be either terraform or aws). In general, in the installer we throw errors and let users decide how to fix their input, rather than quietly changing it.

That is indeed happening in installer code, but just for generating the infra id. That would not affect the outcome of this decision.

pkg/validate/validate.go
```diff
@@ -279,3 +279,11 @@ func Host(v string) error {
 	}
 	return validateSubdomain(v)
 }
+
+// OnPremClusterName verifies if the cluster name contains a '.' and returns an error if it does.
+func OnPremClusterName(v string) error {
+	if strings.Contains(v, ".") {
+		return errors.New("cluster name must not contain '.'")
```
patrickdillon, line 286: This error should use the kube field package we use for validating install config, e.g.:

```
allErrs = append(allErrs, field.Invalid(field.NewPath("metadata", "name"), c.ObjectMeta.Name, nameErr.Error()))
```

patrickdillon: You are correct, sir!

PR #5935, opened 2022-05-20
Bug 2088660: Fix for absence folder for bootstrap ignition iso

adiantum:
- ensure that folder exists before file creation

### DIFF
--- a/pkg/types/nutanix/helpers.go
+++ b/pkg/types/nutanix/helpers.go
@@ -78,6 +78,23 @@ func CreateBootstrapISO(infraID, userData string) (string, error) {
 	}
 
 	fullISOFile := BootISOImagePath(cacheDir, infraID)
+	fullISOFileDir, err := filepath.Abs(filepath.Dir(fullISOFile))
+	if err != nil {
+		return "", errors.Wrap(err, "unable to extract parent directory from bootstrap iso filepath")
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/types/nutanix/helpers.go
```diff
@@ -78,6 +78,23 @@ func CreateBootstrapISO(infraID, userData string) (string, error) {
 	}
 
 	fullISOFile := BootISOImagePath(cacheDir, infraID)
+	fullISOFileDir, err := filepath.Abs(filepath.Dir(fullISOFile))
```
rvanderp3, line 81: Is the call to filepath.Abs necessary here?  [Abs](https://pkg.go.dev/path/filepath#Abs) will set the provided path relative to the current working directory if a full path isn't provided and I'm not sure that was the intention here.  It looks like UserCacheDir already provides an absolute path that is consumed by BootISOImagePath.  I'm not sure the call to Abs is necessary here.  I could be wrong on that.

adiantum: Unfortunately, there is no guarantee that os.UserCacheDir() will return absolute path: https://cs.opensource.google/go/go/+/refs/tags/go1.17:src/os/file.go;l=406

I've checked it at least on Centos 7, it's possible to have conditions when the function return relative path.

PR #5937, opened 2022-05-23
Azure: validate ultrassd instance types in multi-zone regions [alternative approach]

r4f4:
This is an alternative approach to https://github.com/openshift/installer/pull/5854


pkg/asset/installconfig/azure/validation.go
```diff
@@ -146,11 +147,46 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
...
 
-	// The UltraSSDAvailable capability might not be present at all, in which case it must assumed to be false
-	val, ok = capabilities["UltraSSDAvailable"]
-	if ok {
-		ultraSSDAvailable := strings.EqualFold(val, "True")
-		if ultraSSDEnabled && !ultraSSDAvailable {
+	if ultraSSDEnabled {
+		locationInfo, err := client.GetLocationInfo(context.TODO(), region, instanceType)
+		if err != nil {
+			errMsg := fmt.Sprintf("could not determined Availability Zones support in the %s region: %v\n", region, err)
```
jhixson74, line 153: errMsg := fmt.Sprintf("could not determine Availability Zones support in the %s region: %v\n", region, err)

pkg/asset/installconfig/azure/validation_test.go
```diff
@@ -692,3 +657,285 @@ func TestAzureDiskEncryptionSet(t *testing.T) {
 		})
 	}
 }
+
+func TestAzureUltraSSDCapability(t *testing.T) {
+	// ultraSSDCapability := azenc.ResourceSkuCapabilities{
+	// 	Name:  to.StringPtr("UltraSSDAvailable"),
+	// 	Value: to.StringPtr("True"),
+	// }
+
```
jhixson74, line 666: This should probably be removed

jhixson74 (commented):
/lgtm

pkg/asset/installconfig/azure/validation.go
r4f4: Done

pkg/asset/installconfig/azure/validation.go
```diff
@@ -146,11 +147,46 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 		}
 	}
 
-	// The UltraSSDAvailable capability might not be present at all, in which case it must assumed to be false
-	val, ok = capabilities["UltraSSDAvailable"]
-	if ok {
-		ultraSSDAvailable := strings.EqualFold(val, "True")
-		if ultraSSDEnabled && !ultraSSDAvailable {
+	if ultraSSDEnabled {
```
patrickdillon, line 150: Perhaps I'm overloading this PR, but IMHO this function is doing too much, is hard to read and could be refactored to break down each check into its own function, roughly following this pattern:

https://github.com/openshift/installer/blob/0a8f60ffca5edae87a64b0f81d56e352bf6caaad/pkg/asset/installconfig/azure/validation.go#L40-L44

 So, in this case we could have something like:

```go
allErrs := field.ErrorList{} 
  
 allErrs = append(allErrs, validateCPUs(client, fieldPath, capabilities)...) 
 allErrs = append(allErrs, validateMemory(client, fieldPath, capabilities)...)
allErrs = append(allErrs, validateUltraSSD(client, fieldPath, capabilities, zones)...) 
```
This is a simplified example, but WDYT?

patrickdillon (commented):
If I'm understanding the difference between this and #5854 correctly, I prefer this approach. I think this keeps the advantages of the `GetVMCapabilities` function. It looks like testing with Azure Stack is successful and if there are any issues we could handle those with further abstraction.

pkg/asset/installconfig/azure/validation.go
r4f4: I added a new commit implementing this. LMK what you think.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -119,77 +126,99 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
...
+	if diskType == "Premium_LRS" {
+		allErrs = append(allErrs, validatePremiumDisk(fieldPath, diskType, instanceType, capabilities)...)
+	}
+
+	if vmNetworkingType == string(aztypes.VMnetworkingTypeAccelerated) {
+		allErrs = append(allErrs, validateAcceleratedNetworking(fieldPath, vmNetworkingType, instanceType, capabilities)...)
+	}
+
+	if ultraSSDEnabled {
+		allErrs = append(allErrs, validateUltraSSD(client, fieldPath.Child("type"), icZones, region, instanceType, capabilities)...)
```
patrickdillon, line 221: :heart_eyes:

r4f4:
/retitle Bug 2086791: Azure: validate ultrassd instance types in multi-zone regions

pkg/asset/installconfig/azure/validation.go
```diff
@@ -104,10 +112,10 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 		}
 		if cpus < float64(req.minimumVCpus) {
 			errMsg := fmt.Sprintf("instance type does not meet minimum resource requirements of %d vCPUsAvailable", req.minimumVCpus)
-			allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, errMsg))
+			allErrs = append(allErrs, field.Invalid(fieldPath, instanceType, errMsg))
 		}
 	} else {
-		allErrs = append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, "capability not found: vCPUsAvailable"))
+		allErrs = append(allErrs, field.Invalid(fieldPath, instanceType, "capability not found: vCPUsAvailable"))
```
patrickdillon, line 118: So the previous behavior did not fail here when checking memory or vcpus and they couldn't be found? Let's just log a WARNING or DEBUG message instead of failing if that's more consistent with the previous behavior...

patrickdillon (approved):
This looks great. One small typo. Squash?

PR #5938, opened 2022-05-23
Power VS: Readd install-config DNS validation tests

mjturek:
Some changes were made that broke install-config validation. Fix some missing mocks so that we can test again.


pkg/asset/installconfig/powervs/validation_test.go
```diff
@@ -126,3 +129,85 @@ func TestValidate(t *testing.T) {
...
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+
+	powervsClient := mock.NewMockAPI(mockCtrl)
+	metadata := mock.NewMockMetadataAPI(mockCtrl)
+
+	dnsRecordNames := [...]string{fmt.Sprintf("api.%s.%s", validClusterName, validBaseDomain), fmt.Sprintf("api-int.%s.%s", validClusterName, validBaseDomain)}
+
+	//metadata := powervs.NewMetadata(validBaseDomain)
+	//metadata.SetCISInstanceCRN(validCISInstanceCRN)
```
barbacbd, line 167: can probably get rid of these comments

pkg/asset/installconfig/powervs/validation_test.go
```diff
@@ -126,3 +129,85 @@ func TestValidate(t *testing.T) {
...
+
+	// Mocks: no pre-existing DNS records
+	metadata.EXPECT().CISInstanceCRN(gomock.Any()).Return(validCISInstanceCRN, nil)
```
barbacbd, line 170: instead of adding the same mock results you can change 
`metadata.EXPECT().CISInstanceCRN(gomock.Any()).Return(validCISInstanceCRN, nil)`
to 
`metadata.EXPECT().CISInstanceCRN(gomock.Any()).Return(validCISInstanceCRN, nil).AnyTimes()`
since it is always returning the same result.

barbacbd: same applies for your mock call to `GetDNSZoneIDByName` call

pkg/asset/installconfig/powervs/validation_test.go
```diff
@@ -126,3 +129,85 @@ func TestValidate(t *testing.T) {
...
+
+	// Mocks: cannot get zone ID
+	metadata.EXPECT().CISInstanceCRN(gomock.Any()).Return(validCISInstanceCRN, nil)
+	powervsClient.EXPECT().GetDNSZoneIDByName(gomock.Any(), validBaseDomain).Return("", fmt.Errorf(""))
+
+	// Mocks: cannot get DNS records
+	metadata.EXPECT().CISInstanceCRN(gomock.Any()).Return(validCISInstanceCRN, nil)
+	powervsClient.EXPECT().GetDNSZoneIDByName(gomock.Any(), validBaseDomain).Return(validDNSZoneID, nil)
+	for _, dnsRecordName := range dnsRecordNames {
+		powervsClient.EXPECT().GetDNSRecordsByName(gomock.Any(), validCISInstanceCRN, validDNSZoneID, dnsRecordName).Return(nil, fmt.Errorf(""))
```
barbacbd, line 191: It is a bit confusing to use the same dnsRecordNames in the tests for mocking results when you cannot find the DNS records as it is when you can find the DNS Records. Suggest making a list of records that cannot be found.

barbacbd:
I am getting some build errors on the branch.

pkg/asset/installconfig/powervs/validation_test.go
mjturek: This worked for `CISInstanceCRN` but `GetDNSZoneIDByName` expects an empty string for the `cannot get zone ID` test unlike the other tests.

pkg/asset/installconfig/powervs/metadata.go
```diff
@@ -84,7 +94,7 @@ func (m *Metadata) APIKey(ctx context.Context) (string, error) {
 }
 
 // GetCISInstanceCRN gets the CRN name for the specified base domain.
-func GetCISInstanceCRN(APIKey string, BaseDomain string) (string, error) {
+func (m *Metadata) GetCISInstanceCRN() (string, error) {
```
clnperez, line 97: While you're changing things, can you change this function name to something like CISCRNName(), or something? None of the other "getters" have `Get` in the name, and there are two the retrieve CRN-related values, so I think this would make the purpose of each more clear. Unless this ends up creating a cascade of other things that need changing and making this PR overly-large.

clnperez (commented):
very close to lgtm'ing, but would like an answer to the func question.

clnperez:
@patrickdillon PTAL at this one as well

PR #5940, opened 2022-05-24
[release-4.11.0] Bug 2092296: Changed DefaultMachineCIDR of PowerVS to 192.168.0.0/16

bkhadars:
1.Changed PowerVS MachineCIDR from `10.x.x.x/16 to192.168.0.0/16`
2.Added changes to set Default credentialMode for Powervs to `ManualCredentialsMode`

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -1440,8 +1440,8 @@ spec:
                 description: MachineNetwork is the list of IP address pools for machines.
                   This field replaces MachineCIDR, and if set MachineCIDR must be
                   empty or match the first entry in the list. Default is 10.0.0.0/16
-                  for all platforms other than libvirt. For libvirt, the default is
-                  192.168.126.0/24.
+                  for all platforms other than libvirt and Power VS. For libvirt,
+                  the default is 192.168.126.0/24. For Power VS, the default is 192.168.0.0/16.
                 items:
                   description: MachineNetworkEntry is a single IP address block for
                     node IP blocks.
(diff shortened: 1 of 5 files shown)


data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -1440,8 +1440,8 @@ spec:
                 description: MachineNetwork is the list of IP address pools for machines.
                   This field replaces MachineCIDR, and if set MachineCIDR must be
                   empty or match the first entry in the list. Default is 10.0.0.0/16
-                  for all platforms other than libvirt. For libvirt, the default is
-                  192.168.126.0/24.
+                  for all platforms other than libvirt and powervs. For libvirt, the
```
clnperez, line 1443: Unless as variable name, always use `Power VS`. Here you have a mix of `powervs` and `Powervs.`

pkg/types/installconfig.go
```diff
@@ -283,8 +283,9 @@ type Networking struct {
 	// MachineNetwork is the list of IP address pools for machines.
 	// This field replaces MachineCIDR, and if set MachineCIDR must
 	// be empty or match the first entry in the list.
-	// Default is 10.0.0.0/16 for all platforms other than libvirt.
+	// Default is 10.0.0.0/16 for all platforms other than libvirt and powervs.
```
clnperez, line 286: same comment about capitalization.

clnperez (changes requested):
only capitalization concerns. although i'd also prefer that different features be put into different commits at least, if not different PRs. not sure on how @patrickdillon feels about lumping these two nonrelated things into one PR. I'll defer to the installer team's preference

clnperez:
BTW thanks for fixing that credentials mode one. It does need to be done.

clnperez:
/lgtm as well
can we get an approve @patrickdillon ?

PR #5933, opened 2022-05-19
AGENT-165: Install release specified in ClusterImageSet

zaneb:
Instead of hard-coding the release image to be installed in the
assisted-service config, use the version passed in the ClusterImageSet
manifest. dev-scripts already provides this in test environments.

PR #5942, opened 2022-05-24
AGENT-203: Add ClusterImageSet asset

rwsu:
The ClusterImageSet is a Hive CRD used to specify which version
of OpenShift to install.


pkg/asset/agent/manifests/clusterimageset.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Files returns the files generated by the asset.
+func (a *ClusterImageSet) Files() []*asset.File {
+	if a.File != nil {
+		return []*asset.File{a.File}
+	}
+	return []*asset.File{}
+}
+
+// Load returns agentclusterinstall asset from the disk.
```
bfournie, line 50: Nit - change agentclusterinstall

pkg/asset/agent/manifests/clusterimageset.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/openshift/installer/pkg/asset"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/yaml"
+)
+
+var (
+	clusterImageSetFilename = filepath.Join(clusterManifestDir, "cluster-image-set.yaml")
+)
+
+// ClusterImageSet generates the agent-cluster-install.yaml file.
```
bfournie, line 18: Nit - "agent-cluster-install.yaml"

bfournie:
It looks like a failure in ci/prow/verify-vendor due to:
github.com/openshift/hive/apis v0.0.0-20210506000654-5c038fb05190 // indirect

I think the *indirect* needs to be removed like Zane did in https://github.com/openshift/installer/pull/5933 since its now explicitly imported.

pkg/asset/agent/manifests/clusterimageset.go
rwsu: Thanks for spotting these.

PR #5943, opened 2022-05-25
AGENT-180: use nmstateconfig asset and improved validations/tests

andfasano:
This patch introduces the usage of the StaticNetworkConfig objects fetched from the asset.
The validation rules have been split into the respective assets (there's just one cross-assets validation rule, which has been placed into the `AgentManifests` assets).
Tests have been updated accordingly, and also modified to express more clearly the reason of the failure where required.

PR #5945, opened 2022-05-26
AGENT-133: Move Ignition generation from fleeting code to asset

rwsu:
Instead of calling the fleeting code to generate the Ignition config, the Ignition asset now builds it directly.

To make it easier for core reviewers, the first commit touches core installer code and has been submitted as its own PR #5944.

~~Depends on: #5943~~
~~Depends on: #5944~~


pkg/asset/agent/image/ignition.go
```diff
@@ -31,16 +74,107 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
...
+		return fmt.Errorf("pull secret is missing")
+	}
+
+	nodeZeroIP, err := manifests.GetNodeZeroIP(agentManifests.NMStateConfigs)
+	if err != nil {
+		return err
+	}
+
+	agentTemplateData := getTemplateData(
+		pullSecret.StringData[".dockerconfigjson"],
```
andfasano, line 106: Could be retrieved from the method added in [PR 5943](https://github.com/openshift/installer/pull/5943/files#diff-ed3ddff72757f2b9f8078d820e70ac801ad04c31885e033d6722ac6ae865f10eR92)

andfasano:
In general the approach sounds good to me, I think anyhow this PR should follow https://github.com/openshift/installer/pull/5943, also to reuse some of the validations/tests introduced (not yet present here, but I guess since it's still a WIP)

bfournie:
This shouldn't hold this up but it looks like the unit test for GetNodeZeroIP() from here https://github.com/openshift/installer/blob/agent-installer/pkg/agent/manifests/nmstateconfig_test.go#L10-L89 got dropped. 
This can be added back in a follow-on.

/lgtm

bfournie:
I verified this with a successful install in dev-scripts using:
`export INSTALLER_REPO_PR=5945`

/approve

PR #5947, opened 2022-05-30
Bug 2064693: Restore ability to use local clouds.yaml

stephenfin:
We recently switched from using Terraform as a library to using it as a binary. As part of this switch, we have changed the definition of `$PWD` for the providers - it is now a temporary working directory we create rather than the user's current working directory. This breaks the ability to use a `clouds.yaml` file in the local directory when provisioning on OpenStack clouds.

Restore this by checking for the presence of such a file early in the provisioning process and setting an environment variable to point to this file if found. This isn't the nicest solution - it would probably be better to load the file as an asset and copy it to the Terraform working directory - but it should be effective.

### DIFF
--- a/pkg/asset/cluster/cluster.go
+++ b/pkg/asset/cluster/cluster.go
@@ -14,6 +14,7 @@ import (
 	"github.com/openshift/installer/pkg/asset"
 	"github.com/openshift/installer/pkg/asset/cluster/aws"
 	"github.com/openshift/installer/pkg/asset/cluster/azure"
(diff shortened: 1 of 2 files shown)


mandre (commented):
/approve
/lgtm

mandre:
@patrickdillon, this PR has some bits in `pkg/asset/cluster/cluster.go` that the OpenStack team has no ownership on. Could you take a look and approve if you're fine with the changes?
/assign @patrickdillon

PR #5949, opened 2022-05-31
agent: default openshift version when pullspec it's a digest

andfasano:
Since currently we don't have a way to extract the openshift version when the pullspec it's a digest, let's default the version to some hard-coded values. The current approach will not work properly in the general case but it will support the CI environment, where the pullspec it's referencing the current ephemeral release payload

### DIFF
--- a/pkg/agent/imagebuilder/releaseimage.go
+++ b/pkg/agent/imagebuilder/releaseimage.go
@@ -3,7 +3,10 @@ package imagebuilder
 import (
 	"encoding/json"
 	"fmt"
+	"regexp"
 	"strings"
+
+	"github.com/openshift/installer/pkg/version"
 )
 
 type releaseImage struct {
@@ -13,7 +16,28 @@ type releaseImage struct {
 	Tag            string `json:"version"`
 }
 
+func isDigest(pullspec string) bool {
+	return regexp.MustCompile(`.*sha256:[a-fA-F0-9]{64}$`).MatchString(pullspec)
+}
+
 func releaseImageFromPullSpec(pullSpec, arch string) (releaseImage, error) {
+
+	// When the pullspec it's a digest let's use the current version
+	// stored in the installer
+	if isDigest(pullSpec) {
+		versionString, err := version.Version()
+		if err != nil {
+			return releaseImage{}, err
+		}
+
+		return releaseImage{
+			ReleaseVersion: versionString,
+			Arch:           arch,
+			PullSpec:       pullSpec,
+			Tag:            versionString,
+		}, nil
+	}
+
 	components := strings.SplitN(pullSpec, ":", 2)
 	if len(components) < 2 {
 		return releaseImage{}, fmt.Errorf("invalid release image \"%s\"", pullSpec)
--- a/pkg/agent/imagebuilder/releaseimage.go
+++ b/pkg/agent/imagebuilder/releaseimage.go
@@ -35,18 +35,10 @@ func releaseImageFromPullSpec(pullSpec, arch string) (releaseImage, error) {
 }
 
 func releaseImageList(pullSpec, arch string) (string, error) {
-	// TODO: re-enable this code instead of hard-coding
-	/*
-		relImage, err := releaseImageFromPullSpec(pullSpec, arch)
-		if err != nil {
-			return "", err
-		}
-	*/
-	relImage := releaseImage{
-		ReleaseVersion: "4.10",
-		Arch:           arch,
-		PullSpec:       "quay.io/openshift-release-dev/ocp-release:4.10.0-rc.1-x86_64",
-		Tag:            "4.10.0-rc.1",
+
+	relImage, err := releaseImageFromPullSpec(pullSpec, arch)
+	if err != nil {
+		return "", err
 	}
 
 	imageList := []interface{}{relImage}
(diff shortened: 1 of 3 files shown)


pkg/agent/imagebuilder/releaseimage.go
```diff
@@ -13,7 +14,21 @@ type releaseImage struct {
...
 
+func isDigest(pullspec string) bool {
+	return regexp.MustCompile(`.*[a-zA-Z0-9-_+.]+:[a-fA-F0-9]+$`).MatchString(pullspec)
+}
+
 func releaseImageFromPullSpec(pullSpec, arch string) (releaseImage, error) {
+
+	if isDigest(pullSpec) {
+		return releaseImage{
+			ReleaseVersion: "0.99",
```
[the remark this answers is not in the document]
andfasano: Good point, the current job seems to be failing for that reason, even though I was expecting a `0.99` version in the error message of the assisted-service ```
Jun 01 08:37:05 master-0.ostest.test.metalkube.org service[3031]: time="2022-06-01T08:37:05Z" level=error 
msg="Failed to registered cluster ostest with id d729d982-5b52-4915-a6ab-579e813bf597" 
func="github.com/openshift/assisted-service/internal/bminventory.(*bareMetalInventory).RegisterClusterInternal.func1" 
file="/go/src/github.com/openshift/origin/internal/bminventory/inventory.go:390" 
cluster_id=d729d982-5b52-4915-a6ab-579e813bf597 
error="
    Openshift version 4.11.0-0.ci.test-2022-06-01-081506-ci-op-23cf6chl-latest for CPU architecture x86_64 is not supported: 
    The requested release image for version (4.11.0-0.ci.test-2022-06-01-081506-ci-op-23cf6chl-latest) and CPU architecture (x86_64) isn't specified in release images list"     
go-id=151 pkg=Inventory request_id=976f3f47-c8ab-4306-87f7-b7f1d2575fe6```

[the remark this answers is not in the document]
andfasano: I've double checked and currently it does not seem to be required, as long as OS_IMAGES it's not empty

data/data/agent/files/usr/local/share/assisted-service/images.env.template
```diff
@@ -1,3 +1,3 @@
...
+OS_IMAGES=[{"openshift_version":"4.10","cpu_architecture":"x86_64","url":"https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/pre-release/4.10.0-rc.1/rhcos-4.10.0-rc.1-x86_64-live.x86_64.iso","rootfs_url":"https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/pre-release/4.10.0-rc.1/rhcos-4.10.0-rc.1-x86_64-live-rootfs.x86_64.img","version":"410.84.202201251210-0"},{"openshift_version":"4.11","cpu_architecture":"x86_64","url":"https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos-4.11/411.85.202203181601-0/x86_64/rhcos-411.85.202203181601-0-live.x86_64.iso","rootfs_url":"https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos-4.11/411.85.202203181601-0/x86_64/rhcos-411.85.202203181601-0-live-rootfs.x86_64.img","version":"411.85.202203181601-0"}]
```
zaneb, line 3: Is this change just removing the newline at the end of the file?

pkg/agent/imagebuilder/releaseimage.go
```diff
@@ -13,7 +16,28 @@ type releaseImage struct {
 	Tag            string `json:"version"`
 }
 
+func isDigest(pullspec string) bool {
+	return regexp.MustCompile(`.*[a-zA-Z0-9-_+.]+:[a-fA-F0-9]+$`).MatchString(pullspec)
```
zaneb, line 20: The digest has a fixed length, we should probably use that instead of just `+`.

pkg/agent/imagebuilder/releaseimage_test.go
```diff
@@ -50,7 +56,6 @@ func TestReleaseImageListErrors(t *testing.T) {
 	cases := []string{
 		"",
 		"quay.io/openshift-release-dev/ocp-release-4.10",
-		"quay.io/openshift-release-dev/ocp-release:4",
```
zaneb, line 53: If we check the length of the digest we wouldn't need to remove this test.

andfasano: Thanks, and I also noticed I forgot to remove the skip in the previous test!

pkg/agent/imagebuilder/releaseimage_test.go
```diff
@@ -33,13 +31,21 @@ func TestReleaseImageList(t *testing.T) {
...
+			result:   "[{\"openshift_version\":\"was not built correctly\",\"cpu_architecture\":\"x86_64\",\"url\":\"registry.build04.ci.openshift.org/ci-op-m7rfgytz/release@sha256:ebb203f24ee060d61bdb466696a9c20b3841f9929badf9b81fc99cbedc2a679e\",\"version\":\"was not built correctly\"}]",
+		},
 	}
 
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
 			output, err := releaseImageList(tc.pullSpec, tc.arch)
 			assert.NoError(t, err)
-			assert.Equal(t, tc.result, output)
+			if err != nil {
```
zaneb, line 46: Isn't this always false, given that we did assert.NoError() above?

andfasano: I didn't see the tests running in failfast mode, so in case of wrong case (while developing) I found a little bit annoying having an unnecessary failure

zaneb: It should be `if err == nil` then.

zaneb (commented):
/lgtm

rwsu (approved):
/approve

PR #5952, opened 2022-06-01
azure: validation: avoid specifying capabilities twice

r4f4:
Instead, use the capabilities map to generate Azure types. It's much
easier to add capabilities to a simple string map and then have that
automatically converted to Azure API types. We also make sure that both
vmCapabilities and instanceTypeSku are always in sync.


barbacbd (commented):
/lgtm

barbacbd (commented):
/lgtm

r4f4:
/skip

This PR only touches on the Azure unit tests.

PR #5953, opened 2022-06-01
WIP AGENT-174: Wait for bootstrap

lranjbar:
Draft PR for the WIP on this function. Not ready for review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /bin/
+.vscode/
--- a/pkg/agent/wait_for.go
+++ b/pkg/agent/wait_for.go
@@ -2,11 +2,62 @@ package agent
 
 import (
 	"github.com/sirupsen/logrus"
+
+	ainstaller_ctrl "github.com/openshift/assisted-installer/src/assisted_installer_controller/"
+	ainstaller_common "github.com/openshift/assisted-installer/src/common"
+	ainstaller_invclient "github.com/openshift/assisted-installer/src/inventory_client"
+	ainstaller_utils "github.com/openshift/assisted-installer/src/utils"
+	aservice_models "github.com/openshift/assisted-service/models"
 )
+
+type WaitForBootstrap struct {
+	log *logrus.Logger
+	inv inventory_client.InventoryClient
+	kc  k8s_client.K8SClient
+}
 
 // WaitFor wait for the installation complete triggered by the agent installer.
 func WaitFor() error {
 	logrus.Info("WaitFor command")
+
+	return nil
+}
+
+func WaitForBootstrap(b *WaitForBootstrap) error {
+	logrus.Info("WaitForBootstrap command")
+
+	// Assisted Service Models
+	// - cluster : https://github.com/openshift/assisted-service/blob/master/models/cluster.go
+	// - host: https://github.com/openshift/assisted-service/blob/master/models/host.go
+	// - host_stage: https://github.com/openshift/assisted-service/blob/master/models/host.go
+
+	// Assisted Installer functions
+	// - invclient.GetHosts
+	// - common.GetHostsInStatus
+	// - utils.GenerateRequestContext
+
+	// Openshift Installer
+	// - waitForBootstrapComplete cmd/openshift-install/create.go#341
+
+	// psuedo --
+
+	// while bootstrap != not complete
+
+	// 	check for assisted service
+	// 	if assisted service is up log host status from assisted service ?
+
+	// after bootstrap complete return true
+
+	// Snippets from assisted installer ---
+
+	// ctxReq := ainstaller_utils.GenerateRequestContext()
+	// log := ainstaller_utils.RequestIDLogger(ctxReq, c.log)
+
+	// nodesMap, err := ainstaller_invclient.GetHosts(ctxReq, log, []string{aservice_models.HostStatusDisabled})
+	// hostsInProgressMap := ainstaller_common.GetHostsInStatus(nodesMap, []string{aservice_models.HostStatusInstalled}, false)
+	// errNodesMap := ainstaller_common.GetHostsInStatus(nodesMap, []string{aservice_models.HostStatusError}, true)
+
+	// ainstaller_ctrl.logHostStatus(log, nodesMap)
 
 	return nil
 }


pkg/agent/zeroclient.go
```diff
@@ -0,0 +1,85 @@
...
+	infraenvResult, err := restClient.Installer.ListInfraEnvs(ctx, listInfraEnvParams)
+	if err != nil {
+		return nil, err
+	}
+	clusterZeroInfraEnv := infraenvResult.Payload[0].ID
+
+	return &nodeZeroClient{restClient, nodeZeroIP, clusterZeroID, clusterZeroInfraEnv}, nil
+}
+
+func isAgentAPILive(zero *nodeZeroClient, ctx context.Context) (bool, models.OpenshiftVersions, error) {
```
andfasano, line 61: It would be nice to have these methods as part of the `nodeZeroClient` type

pkg/agent/zeroclient.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	clusterZeroID := clusterResult.Payload[0].ID
+
+	// GET /v2/infraenvs
+	listInfraEnvParams := &installer.ListInfraEnvsParams{ClusterID: clusterZeroID}
+	infraenvResult, err := restClient.Installer.ListInfraEnvs(ctx, listInfraEnvParams)
+	if err != nil {
+		return nil, err
+	}
+	clusterZeroInfraEnv := infraenvResult.Payload[0].ID
```
andfasano, line 56: Keeping the two above GET calls within the factory method could be harder for the maintenance and tests, if possible I'd suggest moving them in a type method to be called separately after client instantiation

lranjbar: Yeah I agree with this. I think I can refactor this some. 🤔 

I would like to keep the metadata of ClusterZeroID and ClusterZeroInfraenv in nodeZeroClient to avoid having to call these calls more than once. The goal of storing that metadata was because other API calls ask for ClusterID and Infraenv a lot.

lranjbar: Also perhaps nodeZeroClient and clusterZero should be two different structs. Since I'm trying to solve two different problems there.

pkg/agent/waitfor.go
```diff
@@ -0,0 +1,107 @@
...
+	// ctx := context.Background()
+	// zeroClient, err := NewNodeZeroClient()
+	// if err != nil {
+	// 	return err
+	// }
+
+	return nil
+}
+
+func WaitForNodeZeroAgentRestAPIInit(ctx context.Context, zeroClient *nodeZeroClient, timeoutMins int) error {
```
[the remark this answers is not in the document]
lranjbar: 🤔 To expand this thought I wonder if it is NodeZeroChecker, ClusterZeroChecker or both?

`NodeZeroChecker` would be about nodeZero. The first half of `waitfor bootstrap` would be under this because we are inspecting nodeZero mostly. After nodeZero reboots we do a couple checks against the cluster api to validate bootstrap is complete.
`ClusterZeroChecker` would be about clusterZero which could include the `isKubeAPILive` and `doesKubeConfigExist` methods. This would also clarify that we are talking about the cluster we are installing, clusterZero.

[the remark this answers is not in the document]
lranjbar: or maybe these are on type `clusterZero` and the checkers are another layer of abstraction. 🤔 I will fiddle with it today

pkg/agent/zero/rest.go
```diff
@@ -0,0 +1,93 @@
...
+	if err := assetStore.Fetch(nmState); err != nil {
+		return nil, errors.Wrapf(err, "failed to fetch %s", nmState.Name())
+	}
+
+	nodeZeroIP, err := manifests.GetNodeZeroIP(nmState.Config)
+	if err != nil {
+		return nil, err
+	}
+	config := client.Config{}
+	*config.URL = url.URL{
```
andfasano, line 43:
```suggestion
	config.URL = &url.URL{
```

zaneb (commented):
This looks to be going fine, I put some minor comments inline.

cmd/openshift-install/agent/waitfor.go
```diff
@@ -18,17 +18,59 @@ func NewWaitForCmd() *cobra.Command {
...
 }
 
+func newWaitForClusterValidationSuccessCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "cluster-validated",
+		Short: "Wait until the cluster manifests are validated for install",
+		Args:  cobra.ExactArgs(0),
+		Run: func(cmd *cobra.Command, _ []string) {
+			assetDir := cmd.Flags().Lookup("dir").Value.String()
+			if len(assetDir) == 0 || assetDir == "" {
```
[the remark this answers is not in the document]
lranjbar: I'm terrible at naming. I think the most descriptive would be something like "ClusterToBeInstalled", "ClusterUnderInstall", "InstallingCluster" but I think AgentCluster is fine too.

pkg/agent/zero/cluster.go
```diff
@@ -0,0 +1,229 @@
...
+// be the models.Cluster perhaps it is an object from kubernetes instead.
+type ClusterZeroChecker interface {
+	IsInstalling(cluster *models.Cluster) (bool, string)
+	HasErrored(cluster *models.Cluster) (bool, string)
+	HasStoppedInstalling(cluster *models.Cluster) (bool, string)
+	ParseValidationInfo(cluster *models.Cluster) (bool, error)
+	PrintInstallStatus(cluster *models.Cluster) (string, error)
+}
+
+type ClusterZero struct {
```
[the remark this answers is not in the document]
lranjbar: I'm terrible at naming. I think the most descriptive would be something like "ClusterToBeInstalled", "ClusterUnderInstall", "InstallingCluster" but I think AgentCluster is fine too.

pkg/agent/waitfor.go
```diff
@@ -0,0 +1,179 @@
...
+
+	// Wait to see assisted service API for the first time
+	WaitForNodeZeroAgentRestAPIInit(clusterZero, 5)
+
+	// TODO(lranjbar)[AGENT-172]: Add wait for cluster validation
+	// Wait for cluster validations to succeed
+	// WaitForClusterZeroManifestsToValidate(clusterZero, 5)
+
+	// Wait for cluster Kube API to come up and kubeconfig to be created
+	WaitForClusterZeroKubeConfigToExist(clusterZero, 5)
```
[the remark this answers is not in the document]
lranjbar: This got refactored to just have one timeout of 40 mins. After talking with Andrea extensively I realized unlike the original wait-for we can't assume that the customer is using our wait-for in the beginning. So I refactored the checks to go "backwards" instead. Meaning we check for the end conditions first then back up through the installation process on the checks.

pkg/agent/waitfor.go
```diff
@@ -0,0 +1,179 @@
...
+
+	// Research notes: In installer main package create.go:
+	// waitForInitializedCluster()
+
+	// WaitForBootstrapComplete()
+
+	return nil
+}
+
+func WaitForNodeZeroAgentRestAPIInit(clusterZero *zero.ClusterZero, timeoutMins int) error {
```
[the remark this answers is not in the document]
lranjbar: Noted. In the current refactor I ended up taking out the parameter and just gave an overall timeout of 40 mins. That might still be too short maybe.

pkg/agent/waitfor.go
```diff
@@ -0,0 +1,179 @@
...
+	logrus.Info("WaitForBootstrapComplete")
+
+	ctx := context.Background()
+	clusterZero, err := zero.NewClusterZero(ctx, assetDir)
+	if err != nil {
+		return err
+	}
+
+	// Wait to see assisted service API for the first time
+	WaitForNodeZeroAgentRestAPIInit(clusterZero, 5)
```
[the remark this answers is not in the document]
lranjbar: Yeah there are a lot of edge cases to deal with on this sequential wait-for approach. Maybe there are still more edge cases to take into account. 🤔

pkg/agent/zero/kube.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+type clusterZeroKubeAPIClient struct {
+	Client     *kubernetes.Clientset
+	ctx        context.Context
+	config     *rest.Config
+	configPath string
+}
+
+func NewClusterZeroKubeAPIClient(ctx context.Context, assertDir string) (*clusterZeroKubeAPIClient, error) {
```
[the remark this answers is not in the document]
lranjbar: The original wait-for commands and their functions are private to the main package right now. When I started things were in the "agent" package already and I made another package called "zero". That can all be changed of course.

cmd/openshift-install/agent/waitfor.go
```diff
@@ -37,6 +79,14 @@ func newWaitForInstallCompleteCmd() *cobra.Command {
 
 }
 
-func runWaitForInstallCompleteCmd() error {
-	return agentcmd.WaitFor()
+func runWaitForClusterValidationSuccessCmd(directory string) error {
```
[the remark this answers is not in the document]
lranjbar: This was a misunderstanding I had of the tasks. I thought that because there was a third story we would make a third command. 😅 But the plan is to put the validation checks inside the bootstrap command.

cmd/openshift-install/agent/waitfor.go
```diff
@@ -18,17 +18,59 @@ func NewWaitForCmd() *cobra.Command {
...
 
+func newWaitForClusterValidationSuccessCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "cluster-validated",
+		Short: "Wait until the cluster manifests are validated for install",
+		Args:  cobra.ExactArgs(0),
+		Run: func(cmd *cobra.Command, _ []string) {
+			assetDir := cmd.Flags().Lookup("dir").Value.String()
+			if len(assetDir) == 0 || assetDir == "" {
+				logrus.Fatal("No cluster installation directory found")
```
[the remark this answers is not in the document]
lranjbar: I think we should sync up on the package stuff. 🤔

pkg/agent/cluster.go
```diff
@@ -0,0 +1,330 @@
...
+		models.ClusterStatusPendingForInput:             false,
+		models.ClusterStatusPreparingForInstallation:    false,
+		models.ClusterStatusReady:                       false,
+	}
+	return clusterErrorStates[status], status
+}
+
+// HasStoppedInstalling Determine if the cluster has stopped installing using the models from the Agent Rest API.
+func (czero *Cluster) HasStoppedInstalling(status string) (bool, string) {
+	clusterStoppedInstallingStates := map[string]bool{
```
andfasano, line 238: nit: couldn't be '!IsInstalling()`?

pkg/agent/cluster.go
```diff
@@ -0,0 +1,330 @@
...
+	return true, nil
+}
+
+// IsInstalling Determine if the cluster is still installing using the models from the Agent Rest API.
+func (czero *Cluster) IsInstalling(status string) (bool, string) {
+	clusterInstallingStates := map[string]bool{
+		models.ClusterStatusAddingHosts:                 true,
+		models.ClusterStatusCancelled:                   false,
+		models.ClusterStatusInstalling:                  true,
+		models.ClusterStatusInstallingPendingUserAction: false,
```
andfasano, line 208: Q: why this status is not considered an installing one? Couldn't be, for example, that the user has to fix something to let the installation process continue?

pkg/agent/cluster.go
```diff
@@ -0,0 +1,330 @@
...
+		models.ClusterStatusInsufficient:                false,
```
andfasano, line 209: Also this one is not recoverable?

pkg/agent/kube.go
```diff
@@ -0,0 +1,91 @@
...
+
+	return kubeClient, nil
+}
+
+// IsKubeAPILive Determine if the cluster under install has initailized the kubenertes API.
+func (kube *ClusterKubeAPIClient) IsKubeAPILive() (bool, error) {
+
+	discovery := kube.Client.Discovery()
+	version, err := discovery.ServerVersion()
+	if err != nil {
```
andfasano, line 52: Also here a simple bool could be returned

pkg/agent/cluster.go
```diff
@@ -0,0 +1,330 @@
...
+	clusterKubeAPILive, clusterKubeAPIErr := czero.API.Kube.IsKubeAPILive()
+	if clusterKubeAPIErr != nil {
+		logrus.Debug("cluster Kube API is not available")
+		logrus.Debug(clusterKubeAPIErr)
+	}
+
+	if clusterKubeAPILive {
+		// First time we see the cluster Kube API
+		if !czero.installHistory.ClusterKubeAPISeen {
+			logrus.Info("cluster Kube API Initialized")
```
andfasano, line 131: I'd keep it as a Debug trace, otherwise it'd clutter the logs

pkg/agent/cluster.go
```diff
@@ -0,0 +1,330 @@
...
+		czero.updateInstallHistoryClusterStatus(clusterState)
+
+		stopped, _ := czero.HasStoppedInstalling(*clusterState.Status)
+		if stopped {
+			logrus.Error("cluster has stopped installing")
+			errored, _ := czero.HasErrored(*clusterState.Status)
+			if errored {
+				return false, errors.New("cluster installation has stopped due to errors")
+			}
+			return false, errors.New("cluster has stopped installing and/or insallation was cancelled")
```
andfasano, line 172: I don't remember, but do we support (or plan to) the possibility to cancel the installation in the automated workflow?

```suggestion
			return false, errors.New("cluster has stopped installing and/or installation was cancelled")
```

pkg/agent/cluster.go
```diff
@@ -0,0 +1,330 @@
...
+		}
+	}
+
+	logrus.Debug("bootstrap is not complete. sleeping... ")
+	return false, nil
+}
+
+// IsInstallComplete Determine if the cluster has completed installation.
+func (czero *Cluster) IsInstallComplete() (bool, error) {
+	return true, nil
```
andfasano, line 199: I suppose that also this one should go in the other card (AGENT-173)

cmd/openshift-install/agent/waitfor.go
```diff
@@ -18,17 +18,40 @@ func NewWaitForCmd() *cobra.Command {
...
+		Short: "Wait until the cluster bootstrap is complete",
+		Args:  cobra.ExactArgs(0),
+		Run: func(cmd *cobra.Command, args []string) {
+			assetDir := cmd.Flags().Lookup("dir").Value.String()
+			if len(assetDir) == 0 {
+				logrus.Fatal("No cluster installation directory found")
+			}
+			err := runWaitForBootstrapCompleteCmd(assetDir)
+			if err != nil {
+				logrus.Fatal(err)
```
andfasano, line 38: In case of bootstrap failure it could be useful to log the cluster operators status via `logClusterOperatorConditions` and return the same error code as in https://github.com/openshift/installer/blob/e598e161c7142816fbfa39196d4f857047f1baab/cmd/openshift-install/waitfor.go#L49-L59

pkg/agent/cluster.go
```diff
@@ -0,0 +1,330 @@
...
+	return cluster, nil
+}
+
+// IsBootstrapComplete Determine if the cluster that agent installer
+// is installing has completed the bootstrap process.
+func (czero *Cluster) IsBootstrapComplete() (bool, error) {
+
+	agentRestAPILive, agentRestAPIErr := czero.API.Rest.IsRestAPILive()
+	if agentRestAPIErr != nil {
+		logrus.Debug("node zero Agent API is not available")
```
bfournie, line 118: Just a suggestion to use wrap here and at line 124 to save a line, e.g.
logrus.Debug(errors.Wrap(agentRestAPIErr, "node zero Agent API is not available"))

pkg/agent/cluster.go
```diff
@@ -0,0 +1,330 @@
...
+
+	}
+
+	// both API's are not available
+	if !agentRestAPILive && !clusterKubeAPILive {
+		logrus.Debug("current API Status: Node Zero Agent API: down, Cluster Kube API: down")
+		if !czero.installHistory.RestAPISeen && !czero.installHistory.ClusterKubeAPISeen {
+			logrus.Debug("nero zero Agent API never initialized. Cluster API never initialized")
+			logrus.Warn("unable to detect installation. Cluster install has either not initalized or was not started")
+			return false, nil
```
bfournie, line 183: Just wondering if its intentional to not return an error here. and also if its not an error that a Warning should be logged.

pkg/agent/cluster.go
bfournie: +1

pkg/agent/cluster.go
```diff
@@ -0,0 +1,330 @@
...
+		models.ClusterStatusPendingForInput:             "User input is required to continue cluster installation",
+		models.ClusterStatusPreparingForInstallation:    "Preparing cluster for installation",
+		models.ClusterStatusReady:                       "Cluster is ready for install",
+	}
+	return clusterStoppedInstallingStates[status]
+
+}
+
+// Update the install history struct when we see the status from the Agent Rest API
+func (czero *Cluster) updateInstallHistoryClusterStatus(cluster *models.Cluster) error {
```
bfournie, line 284: Since this function never returns an error could it not have a return value?

pkg/agent/kube.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// Get latest version of bootstrap configmap
+	bootstrap, err := kube.Client.CoreV1().ConfigMaps("kube-system").Get(kube.ctx, "bootstrap", v1.GetOptions{})
+
+	if err != nil {
+		logrus.Debug("bootstrap configmap not found")
+		return false, err
+	}
+	// Found a bootstrap configmap need to check its status
+	if bootstrap != nil && err == nil {
```
bfournie, line 80: Shouldn't need check for 'err == nil' as already checked at line 75

pkg/agent/rest.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	return true, nil
+}
+
+// GetRestAPIServiceBaseURL Return the url of the Agent Rest API on node zero
+func (rest *NodeZeroRestClient) GetRestAPIServiceBaseURL() *url.URL {
+	return rest.config.URL
+}
+
+// getClusterID Return the cluster ID assigned to the by the Agent Rest API
```
bfournie, line 77: Nit  typo - "to the by the"

pkg/agent/cluster.go
lranjbar: I don't know if its planned but someone could do it using the API on node zero before bootstrap reboots.

pkg/agent/cluster.go
lranjbar: Yes. This state happens when the customer has to interact with the machines to push the install forward but the installation has not reached an error that would stop the install completely. 

A common scenario is that they are installing on an old BIOS machine. The mechanism we use to reboot a machine isn't supported by BIOS but if the customer intervenes and reboots the BIOS machine the installation will resume.

pkg/agent/waitfor.go
```diff
@@ -0,0 +1,72 @@
...
+
+	wait.Until(func() {
+		bootstrap, err := cluster.IsBootstrapComplete()
+		if bootstrap && err == nil {
+			logrus.Info("cluster bootstrap is complete")
+			cancel()
+		}
+
+		if err != nil {
+			cancel()
```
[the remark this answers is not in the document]
lranjbar: Yeah.. this was not supposed to be there.

pkg/agent/cluster.go
lranjbar: It was.. I'm not sure if this is an error or a warning. More likely info..

pkg/agent/rest.go
```diff
@@ -0,0 +1,99 @@
...
+	return restClient, nil
+}
+
+// IsRestAPILive Determine if the Agent Rest API on node zero has initialized
+func (rest *NodeZeroRestClient) IsRestAPILive() (bool, error) {
+	// GET /v2/openshift-versions
+	listOpenshiftVersionsParams := versions.NewV2ListSupportedOpenshiftVersionsParams()
+	_, err := rest.Client.Versions.ListSupportedOpenshiftVersions(rest.ctx, (*versions.ListSupportedOpenshiftVersionsParams)(listOpenshiftVersionsParams))
+
+	if err != nil {
```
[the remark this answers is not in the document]
rwsu: The error could be useful for debugging.

rwsu (changes requested):
Nice work Lisa. I can see you put a lot of thought into this. I've found a few minor issues.

pkg/agent/cluster.go
andfasano: I missed the previous update, it seems a better choice to let always the loop naturally reach the timeout (in case of issue), following a similar behavior of the standard wait-for

cmd/openshift-install/agent/waitfor.go
lranjbar: I've added some verbosity to this yesterday using the logClusterOperatorConditions()

pkg/agent/cluster.go
lranjbar: Insufficient is recoverable in fact there is where it is before host validations complete.

Here is the state chart from the assisted-service docs: https://github.com/openshift/assisted-service/blob/master/docs/ClusterStatus.png

pkg/agent/rest.go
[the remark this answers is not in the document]
lranjbar: I went back and forth on this one. I ended up knocking these two errors down to logrus.Trace(). I think they will be useful for debugging but would be extremely noisy on logrus.Debug().

pkg/agent/cluster.go
```diff
@@ -0,0 +1,332 @@
...
+			logrus.Warn("Cluster has stopped installing... working to recover installation")
+			errored, _ := czero.HasErrored(*clusterMetadata.Status)
+			if errored {
+				return false, errors.New("cluster installation has stopped due to errors")
+			} else if *clusterMetadata.Status == models.ClusterStatusCancelled {
+				return false, errors.New("cluster installation was cancelled")
+			}
+		}
+
+		// Print most recent event assoicated with the clusterInfraEnvID
```
andfasano, line 200:
```suggestion
		// Print most recent event associated with the clusterInfraEnvID
```

pkg/agent/rest.go
```diff
@@ -0,0 +1,128 @@
...
+		return nil, err
+	}
+	if len(clusterList) == 1 {
+		clusterID := clusterList[0].ID
+		return clusterID, nil
+	} else if len(clusterList) == 0 {
+		logrus.Debug("cluster is not registered in rest API")
+		return nil, nil
+	} else {
+		logrus.Debugf("number of clusters found: %d", len(clusterList))
```
andfasano, line 104: IMHO this should be considered as a critical error. I couldn't figure out a case where this scenario could be possible, if not for a erroneous condition

pkg/agent/rest.go
```diff
@@ -0,0 +1,128 @@
...
+		return nil, err
+	}
+	if len(infraEnvList) == 1 {
+		clusterInfraEnvID := infraEnvList[0].ID
+		return clusterInfraEnvID, nil
+	} else if len(infraEnvList) == 0 {
+		logrus.Debug("infraenv is not registered in rest API")
+		return nil, nil
+	} else {
+		logrus.Debugf("number of infraenvs found: %d", len(infraEnvList))
```
andfasano, line 125: Same here, as above

pkg/agent/cluster.go
```diff
@@ -0,0 +1,332 @@
...
+		eventList, err := czero.API.Rest.GetInfraEnvEvents(czero.clusterInfraEnvID)
+		if err != nil {
+			return false, errors.Wrap(err, "unable to retrieve events about the cluster from the Agent Rest API")
+		}
+		if len(eventList) == 0 {
+			logrus.Trace("no cluster events detected from the Agent Rest API")
+		} else {
+			mostRecentEvent := eventList[len(eventList)-1]
+			// Don't print the same status message back to back
+			if mostRecentEvent.Message != &czero.installHistory.RestAPIPreviousEventMessage {
```
andfasano, line 210:
```suggestion
			if *mostRecentEvent.Message != czero.installHistory.RestAPIPreviousEventMessage {
```

pkg/agent/cluster.go
```diff
@@ -0,0 +1,332 @@
...
+				if *mostRecentEvent.Severity == models.EventSeverityInfo {
+					logrus.Info(mostRecentEvent.Message)
```
andfasano, line 212:
```suggestion
					logrus.Info(*mostRecentEvent.Message)
```
(otherwise the pointer address gets printed instead of the message)

pkg/agent/cluster.go
```diff
@@ -0,0 +1,332 @@
...
+				} else {
+					logrus.Warn(mostRecentEvent.Message)
```
andfasano, line 214:
```suggestion
					logrus.Warn(*mostRecentEvent.Message)
```

pkg/agent/cluster.go
andfasano: Thanks for the doc

pkg/agent/cluster.go
```diff
@@ -0,0 +1,332 @@
...
+
+	if clusterKubeAPILive {
+
+		// First time we see the cluster Kube API
+		if !czero.installHistory.ClusterKubeAPISeen {
+			logrus.Info("cluster Kube API Initialized")
+			czero.installHistory.ClusterKubeAPISeen = true
+		}
+
+		err := czero.API.OpenShift.LogClusterOperatorConditions()
```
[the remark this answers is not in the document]
lranjbar: Okay I think I just misunderstood where this is supposed to be called.

pkg/agent/waitfor.go
```diff
@@ -0,0 +1,56 @@
...
+			logrus.Info("cluster bootstrap is complete")
+			cancel()
+		}
+
+		current := time.Now()
+		elapsed := current.Sub(previous)
+		elapsedTotal := current.Sub(start)
+		logrus.Tracef("elapsed: %s, elapsedTotal: %s", elapsed.String(), elapsedTotal.String())
+		if elapsed >= 30*time.Second {
+			logrus.Infof("Waiting for cluster bootstrap to complete. Time elapsed: %s", elapsedTotal.String())
```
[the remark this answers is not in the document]
lranjbar: I removing the INFO I think this is more meant for a debugging thing in my head

PR #5955, opened 2022-06-02
Don't run create-cluster-and-infraenv in the background

zaneb:
This is a one-shot service, so we want the pod to run synchronously and
return its error code. When we allowed podman to detach, the service was
treated as successful as soon as the container started.


rwsu (approved):
/lgtm

PR #5956, opened 2022-06-02
Bug 2055612: Adding the report message when quotes are available but low

AnnaZivkovic:
If, in the unlikely case, a cluster runs out of quotas after checking if there are enough then the install will fail. In this case there is only a warring about which quotas are low. For more information we can add the number of the resource that is required to the output.


sadasu:
@AnnaZivkovic couple of nits: quotas and not quotes in the PR title.
And warning and probably not warring in the PR description.

PR #5957, opened 2022-06-02
AGENT-207: Assets for disconnected installs

bfournie:
Add raw files as assets to allow configuration of the registries.conf and ca-bundle.crt needed for disconnected installations. The contents of these files will eventually by populated from fields in install-config.yaml, for now the actual files must be provided in the mirror dir.


pkg/asset/agent/mirror/cabundle_test.go
```diff
@@ -0,0 +1,73 @@
...
+			data: `
+-----BEGIN CERTIFICATE-----
+MIIDvTCCAqWgAwIBAgICEAA....
+----- END CERTIFICATE -----`,
+			expectedFound: true,
+		},
+		{
+			name:          "empty",
+			data:          "",
+			expectedFound: true,
```
andfasano, line 35: As we started to do in other [similar tests](https://github.com/openshift/installer/blob/98bfce1602686f503fbcadf54b7011bddca0f129/pkg/asset/agent/manifests/agentpullsecret_test.go#L87), for the sake of clarity (and to avoid ignoring unexpected errors) I'd suggest to check explicitly the error message string for those tests where an error is expected

pawanpinjarkar: +1 to check the error message.

rwsu (approved):
/lgtm

PR #5959, opened 2022-06-03
AGENT-40: Generate kubeconfig alongside ISO

zaneb:
Generate a kubeconfig at the same time as the agent ISO. Pass the certificates used to generate this config into the ISO and to assisted-service to allow it to pass them to the installer, as described in https://github.com/openshift/assisted-service/pull/3836

Note that while this should be safe to merge on its own, the generated kubeconfig will not work until both:

1. https://github.com/openshift/assisted-service/pull/3836 has landed; and
2. we are no longer hard-coding the release image (i.e. openshift/installer#5954 has been reverted), but are instead using a (post-Feature Freeze) 4.11 image.

### DIFF
--- a/pkg/asset/kubeconfig/agent.go
+++ b/pkg/asset/kubeconfig/agent.go
@@ -1,0 +1,46 @@
+package kubeconfig
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/openshift/installer/pkg/asset"
+	agentmanifests "github.com/openshift/installer/pkg/asset/agent/manifests"
+	"github.com/openshift/installer/pkg/asset/tls"
+)
+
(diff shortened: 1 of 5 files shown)


pkg/asset/kubeconfig/agent.go
```diff
@@ -0,0 +1,48 @@
...
+
+	return k.kubeconfig.generate(
+		ca,
+		clientCertKey,
+		extAPIServerURL,
+		clusterName,
+		"admin",
+		kubeconfigAdminPath,
+	)
+}
```
bfournie, line 48: It looks like there isn't a Name() or Load() function for this asset, would it be possible to add for consistency.

zaneb: It's inherited from the regular `AdminClient`.

PR #5961, opened 2022-06-03
WIP: AGENT-139: Get nmstate builtin

celebdor:


patrickdillon:
Speaking with Toni on slack, I wanted to mention a few considerations:

1. Our builds are very iffy right now (failing frequently due to hitting the hour long timeout limit) after our terraform implementation. Each build builds the installer for 3 different architectures, and building and zipping each terraform provider takes a long time; this is why we are frequently hitting timeouts. This limitation needs to be considered and the pattern is not sustainable (long term).
2. It looks like some of the dependencies are not present in the builder image, so that would need to be worked out with ART.

/cc @jhixson74 who is working on profiling the installer build, but is out through early next week at a conference

celebdor:
@patrickdillon couldn't the builds be separated per architecture to avoid that particular problem?

patrickdillon:
> @patrickdillon couldn't the builds be separated per architecture to avoid that particular problem?

I don't _think_ so. Also I misspoke: for each architecture we're making three builds but it varies by os & arch: mac arm64, mac amd64, and linux amd64. 

But this discussion and the links should give more context: https://coreos.slack.com/archives/C68TNFWA2/p1648044301526129

bfournie:
I found that I needed to make this change to build.sh when building on rhel/Centos
```
-CGO_LDFLAGS="${CGO_LDFLAGS:--L${REPO_ROOT_DIR}/nmstate/rust/lib}"
+CGO_LDFLAGS="${CGO_LDFLAGS:--L${REPO_ROOT_DIR}/nmstate/rust/lib} -ldl -lrt"

```

PR #5964, opened 2022-06-03
Bug 2093368: OpenStack: Fix LoadBalancer FIP deletion on destroy

dulek:
d2630f299503305d6689bfcb50bfc04e20099c80 implemented deleting ports from
networks even if they're untagged. The motivation was to not block
destroy when some untagged ports were orphaned on the network (in
Neutron tagging is a separate operation that can fail).

As this was always done before deleting the network, we also deleted the
LoadBalancer Services VIP ports (untagged). This means that we couldn't
track down FIPs created for these Services and these were orphaned.

This commit makes sure that we only attempt to delete untagged ports on
409 failure to delete the network. We also do that only after successful
deletion of the LBs to make sure all the Service FIPs are already
tracked down and taken care of.

### DIFF
--- a/pkg/destroy/openstack/openstack.go
+++ b/pkg/destroy/openstack/openstack.go
@@ -846,17 +846,17 @@ func getRouterByPort(client *gophercloud.ServiceClient, allPorts []ports.Port) (
 	return empty, nil
 }
 
-func deleteLeftoverLoadBalancers(opts *clientconfig.ClientOpts, logger logrus.FieldLogger, networkID string) {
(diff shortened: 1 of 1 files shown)


dulek:
@mandre @pierreprinetti: I figured out why FIPs weren't deleted.

mandre (commented):
Nice catch!
/lgtm
/approve
/hold for CI

mandre (commented):
/lgtm
/hold cancel

PR #5965, opened 2022-06-03
Bug 2090049: GCP Destroy resources that are owned by cluster

barbacbd:
destroy/gcp: Clean target pools, Added Log statements

** Provide a bit more insight through log statements into the actions that the
installer is taking.

** Cleaned up the targetpools by finding all instances then using
the instances names (including the `owned` labeled instances) to
compare to values in the target pools. Valid target pools can be found
and destroyed, causing other resources to be released.

### DIFF
--- a/pkg/destroy/gcp/cloudcontroller.go
+++ b/pkg/destroy/gcp/cloudcontroller.go
@@ -38,10 +38,10 @@ func (o *ClusterUninstaller) listCloudControllerBackendServices(instanceGroups [
 	})
 }
 
-// listCloudControllerTargetPools returns target pools created by the cloud controller.
-// It list all target pools matching the cloud controller name convention that contain
-// only cluster instances.
-func (o *ClusterUninstaller) listCloudControllerTargetPools() ([]cloudResource, error) {
+// listCloudControllerTargetPools returns target pools created by the cloud controller or owned by the cloud controller.
+// It lists all target pools matching the cloud controller name convention that contain
+// only cluster instances or cluster instances that were owned by the cluster.
+func (o *ClusterUninstaller) listCloudControllerTargetPools(instances []cloudResource) ([]cloudResource, error) {
 	filter := "name eq \"a[0-9a-f]{30,50}\""
 	return o.listTargetPoolsWithFilter("items(name,instances),nextPageToken", filter, func(pool *compute.TargetPool) bool {
 		if len(pool.Instances) == 0 {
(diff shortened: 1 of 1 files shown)


pkg/destroy/gcp/instance.go
```diff
@@ -26,12 +26,12 @@ func (o *ClusterUninstaller) getInstanceNameAndZone(instanceURL string) (string,
 }
 
 func (o *ClusterUninstaller) listInstances() ([]cloudResource, error) {
-	byName, err := o.listInstancesWithFilter("items/*/instances(name,zone,status,machineType),nextPageToken", o.clusterIDFilter(), nil)
+	byName, err := o.listInstancesWithFilter("items/*/instances(labels,name,zone,status,machineType),nextPageToken", o.clusterIDFilter(), nil)
```
rna-afk, line 29: I feel like I'm missing something but why do we need the labels?

barbacbd: technically that is another way to identify Disks if created with the CSI. I may have left that in there as I was adding that to the return type but we can probably get rid of that if I am no longer using it as part of CloudResource objects that are returned

patrickdillon: Maybe this slipped in from #5976?

pkg/destroy/gcp/cloudcontroller.go
```diff
@@ -50,7 +50,18 @@ func (o *ClusterUninstaller) listCloudControllerTargetPools() ([]cloudResource,
 		for _, instanceURL := range pool.Instances {
 			name, _ := o.getInstanceNameAndZone(instanceURL)
```
patrickdillon, line 51: This approach seems reasonable, but I'm concerned we're complicating the existing approach when we might be able to solve the problem by simplifying it. The existing approach is to find all target pools that match the cloud controller name convention, and then iterate through all of the instances in those target pool; if all of those instances belong to the cluster we mark the target pool for delete.

We could have this approach for two reasons I see: 
1. We don't know how to identify cluster target pools
2. We don't want to delete a target pool if it has a non-cluster instance

For question 1, is it possible to identify the target pools by using `loadBalancerName`? Spot checking GCP it looks like the target pool matches the `loadBalancerName` exactly. 

For question 2, I find it unlikely that a non cluster VM would be added to the cluster load balancer, and moreoever what happens if we try to destroy a target pool with a non-cluster vm? We should test this. 

So my suggestion for an alternative approach would be to change the filter to match only on `loadBalancerName` and remove the instance checks entirely.

barbacbd: I might be a bit confused on this one, because the loadBalancerName is assumed to be the pool.Name. How can we use the loadBalancerName if we haven't found the target pool yet?

pkg/destroy/gcp/instance.go
barbacbd: squashed

PR #5941, opened 2022-05-24
Bug 2057582: For metal upi update source and versions of terraform providers

sadasu:
Update terraform provider source and versions for ignition, matchbox and metal

PR #5969, opened 2022-06-06
Bug 2057582: Remove all occurances of packet_device from upi/metal

sadasu:
With the move to equinix/metal terraform provider , resource names
have to be migrated to metal_device.

Is a remnant from the work done in https://github.com/openshift/installer/pull/5941


patrickdillon:
/retitle Bug 2057582: Remove all occurances of packet_device from upi/metal

PR #5973, opened 2022-06-07
Bug 2090836: Fixes CFE-489 - AWS installer should go through proxy for s3 bootstrap ignition call

sherine-k:
BZ-2090836 states:
Given An AWS Openshift Install
When I Configured HTTP Proxy for Openshift installer
Then bootstrap Ignition should grab assets in S3 through the configured HTTP proxy
And Not expect direct internet access to the assets
Unless VPC endpoints are added to the noProxy configuration to bypass those endpoints

This PR passes proxy configuration to the ignition by using `GenerateIgnitionShimWithCertBundleAndProxy`

### DIFF
--- a/pkg/asset/ignition/bootstrap/bootstrap_ignition.go
+++ b/pkg/asset/ignition/bootstrap/bootstrap_ignition.go
@@ -35,23 +35,9 @@ func parseCertificateBundle(userCA []byte) ([]igntypes.Resource, error) {
 	return carefs, nil
 }
 
-// GenerateIgnitionShimWithCertBundle is used to generate an ignition file that contains a user ca bundle
-// in its Security section.
-func GenerateIgnitionShimWithCertBundle(bootstrapConfigURL string, userCA string) ([]byte, error) {
-	ign, err := generateIgnitionShimWithCertBundle(bootstrapConfigURL, userCA)
-	if err != nil {
-		return nil, err
-	}
-
-	data, err := ignition.Marshal(ign)
-	if err != nil {
-		return nil, err
-	}
-
-	return data, nil
-}
-
-func generateIgnitionShimWithCertBundle(bootstrapConfigURL string, userCA string) (igntypes.Config, error) {
+// GenerateIgnitionShimWithCertBundleAndProxy is used to generate an ignition file that contains both a user ca bundle
+// in its Security section and proxy settings (if any).
+func GenerateIgnitionShimWithCertBundleAndProxy(bootstrapConfigURL string, userCA string, proxy *types.Proxy) ([]byte, error) {
 	ign := igntypes.Config{
 		Ignition: igntypes.Ignition{
 			Version: igntypes.MaxVersion.String(),
@@ -65,7 +51,7 @@ func generateIgnitionShimWithCertBundle(bootstrapConfigURL string, userCA string
 
 	carefs, err := parseCertificateBundle([]byte(userCA))
 	if err != nil {
-		return ign, err
+		return nil, err
 	}
 	if len(carefs) > 0 {
 		ign.Ignition.Security = igntypes.Security{
@@ -74,18 +60,10 @@ func generateIgnitionShimWithCertBundle(bootstrapConfigURL string, userCA string
 			},
 		}
 	}
-	return ign, nil
-}
 
-// GenerateIgnitionShimWithCertBundleAndProxy is used to generate an ignition file that contains both a user ca bundle
-// in its Security section and proxy settings.
-func GenerateIgnitionShimWithCertBundleAndProxy(bootstrapConfigURL string, userCA string, proxy *types.Proxy) ([]byte, error) {
-	ign, err := generateIgnitionShimWithCertBundle(bootstrapConfigURL, userCA)
-	if err != nil {
-		return nil, err
+	if proxy != nil {
+		ign.Ignition.Proxy = ignitionProxy(proxy)
 	}
-
-	ign.Ignition.Proxy = ignitionProxy(proxy)
 
 	data, err := ignition.Marshal(ign)
 	if err != nil {
--- a/pkg/asset/ignition/bootstrap/bootstrap_ignition.go
+++ b/pkg/asset/ignition/bootstrap/bootstrap_ignition.go
@@ -38,27 +38,9 @@ func parseCertificateBundle(userCA []byte) ([]igntypes.Resource, error) {
 // GenerateIgnitionShimWithCertBundle is used to generate an ignition file that contains a user ca bundle
 // in its Security section.
 func GenerateIgnitionShimWithCertBundle(bootstrapConfigURL string, userCA string) ([]byte, error) {
-	ign := igntypes.Config{
-		Ignition: igntypes.Ignition{
-			Version: igntypes.MaxVersion.String(),
-			Config: igntypes.IgnitionConfig{
-				Replace: igntypes.Resource{
-					Source: ignutil.StrToPtr(bootstrapConfigURL),
-				},
-			},
-		},
-	}
-
-	carefs, err := parseCertificateBundle([]byte(userCA))
+	ign, err := generateIgnitionShimWithCertBundle(bootstrapConfigURL, userCA)
 	if err != nil {
 		return nil, err
-	}
-	if len(carefs) > 0 {
-		ign.Ignition.Security = igntypes.Security{
-			TLS: igntypes.TLS{
-				CertificateAuthorities: carefs,
-			},
-		}
 	}
 
 	data, err := ignition.Marshal(ign)
@@ -69,9 +51,7 @@ func GenerateIgnitionShimWithCertBundle(bootstrapConfigURL string, userCA string
 	return data, nil
 }
 
-// GenerateIgnitionShimWithCertBundleAndProxy is used to generate an ignition file that contains both a user ca bundle
-// in its Security section and proxy settings.
-func GenerateIgnitionShimWithCertBundleAndProxy(bootstrapConfigURL string, userCA string, proxy *types.Proxy) ([]byte, error) {
+func generateIgnitionShimWithCertBundle(bootstrapConfigURL string, userCA string) (igntypes.Config, error) {
 	ign := igntypes.Config{
 		Ignition: igntypes.Ignition{
 			Version: igntypes.MaxVersion.String(),
@@ -80,13 +60,12 @@ func GenerateIgnitionShimWithCertBundleAndProxy(bootstrapConfigURL string, userC
 					Source: ignutil.StrToPtr(bootstrapConfigURL),
 				},
 			},
-			Proxy: ignitionProxy(proxy),
 		},
 	}
 
 	carefs, err := parseCertificateBundle([]byte(userCA))
 	if err != nil {
-		return nil, err
+		return ign, err
 	}
 	if len(carefs) > 0 {
 		ign.Ignition.Security = igntypes.Security{
@@ -95,6 +74,18 @@ func GenerateIgnitionShimWithCertBundleAndProxy(bootstrapConfigURL string, userC
 			},
 		}
 	}
+	return ign, nil
+}
+
+// GenerateIgnitionShimWithCertBundleAndProxy is used to generate an ignition file that contains both a user ca bundle
+// in its Security section and proxy settings.
+func GenerateIgnitionShimWithCertBundleAndProxy(bootstrapConfigURL string, userCA string, proxy *types.Proxy) ([]byte, error) {
+	ign, err := generateIgnitionShimWithCertBundle(bootstrapConfigURL, userCA)
+	if err != nil {
+		return nil, err
+	}
+
+	ign.Ignition.Proxy = ignitionProxy(proxy)
 
 	data, err := ignition.Marshal(ign)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


patrickdillon (commented):
There's a fair amount of duplication between `GenerateIgnitionShimWithCertBundle` & `GenerateIgnitionShimWithCertBundleAndProxy`.

One potential way to reduce that duplication could be to create a function that modifies the existing ignition config. 

For example, you could leave `stubIgn` to be created by the existing `GenerateIgnitionShimWithCertBundle` and then, if the proxy is populated, call another function that populates the values

sherine-k:
> There's a fair amount of duplication between `GenerateIgnitionShimWithCertBundle` & `GenerateIgnitionShimWithCertBundleAndProxy`.
> 
> One potential way to reduce that duplication could be to create a function that modifies the existing ignition config.
> 
> For example, you could leave `stubIgn` to be created by the existing `GenerateIgnitionShimWithCertBundle` and then, if the proxy is populated, call another function that populates the values

Agreed. Now that we've confirmed we want to share this function, I'll work on making it more DRY.
Thanks

yunjiang29:
Hello @sherine-k,
I did pre-merge tests:
* remove s3 VPC endpoint point from VPC
* create an IPI private cluster  (only private ip was allocated to bootstrap) in a disconnected network
* cluster was created successfully and worked well.

I also want to verify it on UPI process, but I got a cluster-bot issue and no available payload to verify, this PR will also work on  UPI installation, is it right?

sherine-k:
> 



> Hello @sherine-k, I did pre-merge tests:
> 
>     * remove s3 VPC endpoint point from VPC
> 
>     * create an IPI private cluster  (only private ip was allocated to bootstrap) in a disconnected network
> 
>     * cluster was created successfully and worked well.
> 
> 
> I also want to verify it on UPI process, but I got a cluster-bot issue and no available payload to verify, this PR will also work on UPI installation, is it right?
Yes

sherine-k:
/cc @mjpytlak 
we were wondering if any updates are necessary on docs side:
With this change, all installer calls will go through proxy, including the ones to S3, unless explicitly configured as non-proxy-host.
Previously, we used to instruct admins to create VPC endpoints in order to ensure the installer success. This is no longer necessary. 
Clients who want to keep using the VPC endpoints, and bypass proxy to reach them, need to explicitly set the S3 endpoint as a non proxy host. 
Please let me know what you think

pkg/asset/ignition/bootstrap/bootstrap_ignition.go
```diff
@@ -36,6 +38,20 @@ func parseCertificateBundle(userCA []byte) ([]igntypes.Resource, error) {
 // GenerateIgnitionShimWithCertBundle is used to generate an ignition file that contains a user ca bundle
 // in its Security section.
 func GenerateIgnitionShimWithCertBundle(bootstrapConfigURL string, userCA string) ([]byte, error) {
+	ign, err := generateIgnitionShimWithCertBundle(bootstrapConfigURL, userCA)
```
wking, line 41: Trying to figure out why we'd want to keep this non-proxy form around:

```console
$ git --no-pager log --oneline -1 origin/pr/5973
0defd86d01 (origin/pr/5973) Bootstrap: AWS - GenerateIgnitionShimWithCertBundleAndProxy should reuse function GenerateIgnitionShimWithCertBundle
$ git --no-pager grep 'GenerateIgnitionShimWithCertBundle(' origin/pr/5973
origin/pr/5973:pkg/asset/cluster/tfvars.go:                     shim, err := bootstrap.GenerateIgnitionShimWithCertBundle(bootstrapIgnURLPlaceholder, installConfig.Config.AdditionalTrustBundle)
origin/pr/5973:pkg/asset/ignition/bootstrap/bootstrap_ignition.go:func GenerateIgnitionShimWithCertBundle(bootstrapConfigURL string, userCA string) ([]byte, error) {
origin/pr/5973:pkg/tfvars/alibabacloud/alibabacloud.go: stubIgn, err := bootstrap.GenerateIgnitionShimWithCertBundle(sources.IgnitionPresignedURL, sources.AdditionalTrustBundle)
```

That's some `AzureStackCloud` thing, and an Alibaba thing.  Presumably they'd both want the proxy configuration too?  But I'm not an approver here, and it's fine if folks want to handle the providers one at a time, instead of all in a single pull request.

mjpytlak:
> With this change, all installer calls will go through proxy, including the ones to S3, unless explicitly configured as non-proxy-host.
> Previously, we used to instruct admins to create VPC endpoints in order to ensure the installer success. This is no longer necessary.
> Clients who want to keep using the VPC endpoints, and bypass proxy to reach them, need to explicitly set the S3 endpoint as a non proxy host.

@sherine-k @yunjiang29 Please correct me if I am wrong. Based on my understanding of this PR and the respective bug, this work is related to configuring a cluster-wide proxy during installation. If so, I believe we have two areas of the VPC doc that might need an update:

- [Requirements for using your VPC](https://docs.openshift.com/container-platform/4.10/installing/installing_aws/installing-restricted-networks-aws-installer-provisioned.html#installation-custom-aws-vpc-requirements_installing-restricted-networks-aws-installer-provisioned) (last paragraph in that section that states that the endpoints must be created) [1]
- [Configuring a cluster-wide proxy](https://docs.openshift.com/container-platform/4.10/installing/installing_aws/installing-restricted-networks-aws-installer-provisioned.html#installation-custom-aws-vpc-requirements_installing-restricted-networks-aws-installer-provisioned) [2]

[1] Are the `ec2`, `elasticloadbalancing`, and `s3` endpoints no longer required in a use case that requires a VPC or only a restricted (disconnected) installation?
[2] If I understand correctly, the 3rd bullet under Perquisites is no longer required, correct? Also - if customers wanted to use VPC endpoints, they would have to configure them as `noProxy` as detailed under step 1. Correct?

mjpytlak:
> Do we have the same content for managed vs self managed

@sherine-k If by "managed" you mean OCP and "self-managed" you mean OpenShift Dedicated, no the doc for configuring a cluster proxy only appears in OCP.

pkg/asset/ignition/bootstrap/bootstrap_ignition.go
sherine-k: Hi @wking 
I agree, especially that, looking at the documentation, we don't seem to distinguish Proxy Setup for each cloud provider...
I'm happy to extend this if we all agree on this.
Openstack implementation has its own `bootstrap_ignition.go` and its own `GenerateIgnitionShimWithCertBundleAndProxy`, so we could also reuse the code there (requires some changes).

sherine-k:
Hi @mjpytlak,
> [1] Are the `ec2`, `elasticloadbalancing`, and `s3` endpoints no longer required in a use case that requires a VPC or only a restricted (disconnected) installation?

The E2E test that I used (e2e-aws-proxy) did not remove the endpoints, nor declare VPC endpoints in the noProxy. 
I still could see that the call to the S3 bootstrap bucket passed through proxy.

Now, the proxy should allow cluster traffic to the internet. When VPC endpoints aren't available, calls to the S3 bucket would go through proxy -> internet gateway -> out of VPC -> S3 bucket. 

I would say that the VPC endpoints become optional for the customer.

[This PR ](https://github.com/openshift/release/pull/29445) was intended to play around with such a scenario where we create none of the endpoints. But I'm not sure if the installer team would like to go all the way with a e2e test workflow. @patrickdillon I'm happy to complete that work if needed. 

@yunjiang29, were you able to perform any similar tests to confirm:
* if the installer still succeeds without any of the endpoints
* if the installer still succeeds with all endpoints declared in noProxy

> [2] If I understand correctly, the 3rd bullet under Perquisites is no longer required, correct? Also - if customers wanted to use VPC endpoints, they would have to configure them as `noProxy` as detailed under step 1. Correct?

Yes, this is how I understand it

patrickdillon:
/retitle Bug 2090836: Fixes CFE-489 - AWS installer should go through proxy for s3 bootstrap ignition call

patrickdillon:
/approve

This LGTM. Keeping an eye on the tests...

mjpytlak:
@sherine-k @patrickdillon I will go ahead and create a doc bug based on https://github.com/openshift/installer/pull/5973#issuecomment-1163979628.

sherine-k:
/test e2e-aws-single-node
/test/e2e-aws-upi 
/test/e2e-azure-upi
/test/e2e-gcp-upi

yunjiang29:
@sherine-k test results:

> if the installer still succeeds without any of the endpoints

PASS: `Removed all EPs` + `no EPs in noProxy`

> if the installer still succeeds with all endpoints declared in noProxy

FAILED: `Removed all EPs` + `set all EPs in noProxy`, bootstrap machine can not get ignition file, this is expected, is it right?

yunjiang29:
@patrickdillon all the bootstrap machines created by IPI way was fetching ignition file via s3-presign, is it right?

sherine-k:
Thanks @yunjiang29 ! 
The results seem to reflect well the expected behavior.
Test#2 is expected to fail.

mjpytlak:
> I would say that the VPC endpoints become optional for the customer.

@patrickdillon @sherine-k Ahead of updating the doc, I want to make sure that I understand this correctly. Requirements for using a VPC [1] states that you must create the following endpoints:

`If you are working in a disconnected environment, you are unable to reach the public IP addresses for EC2 and ELB endpoints. To resolve this, you must create a VPC endpoint and attach it to the subnet that the clusters are using. The endpoints should be named as follows:`

- `ec2.<region>.amazonaws.com`
- `elasticloadbalancing.<region>.amazonaws.com`
- `s3.<region>.amazonaws.com`

This fix removes the requirement of the latter endpoints, correct? Is there a reason why would simply wouldn't remove the requirement from the documentation?

[1] https://docs.openshift.com/container-platform/4.10/installing/installing_aws/installing-restricted-networks-aws-installer-provisioned.html#installation-custom-aws-vpc-requirements_installing-restricted-networks-aws-installer-provisioned

sherine-k:
> > I would say that the VPC endpoints become optional for the customer.
> 
> @patrickdillon @sherine-k Ahead of updating the doc, I want to make sure that I understand this correctly. Requirements for using a VPC [1] states that you must create the following endpoints:

> This fix removes the requirement of the latter endpoints, correct? Is there a reason why would simply wouldn't remove the requirement from the documentation?
> 
> [1] https://docs.openshift.com/container-platform/4.10/installing/installing_aws/installing-restricted-networks-aws-installer-provisioned.html#installation-custom-aws-vpc-requirements_installing-restricted-networks-aws-installer-provisioned

Hi @mjpytlak 
Thanks for taking the time on the phone with me.
As we discussed, our customers working on disconnected environments, now have 3 options:
* I don't have a proxy at all, I must set these VPC endpoints in order to install the cluster
* I have a proxy to go out to the internet
  * I don't care if the traffic flows through the internet: I don't need to set VPC endpoints, traffic to the s3 bucket will flow by default through the proxy
  * I do care about protecting my traffic, and want to use VPC endpoints so that the traffic remains in private networks between my VPC and the bucket: I need to set up the VPC endpoints AND add them as non-proxy hosts

Asking @patrickdillon and @yunjiang29 what they think about this

yunjiang29:
Thank you @sherine-k @mjpytlak, I think we are already on the same page, I listed a table here

## Workable scenarios
| Item  | Endpoints  | Proxy | How network traffic goes (s3 as example) | 
| ------------- | ------------- | ------------- | ------------- |
| Scenario  1. Traditional way (official doc) |  Yes  | None  | VPC (with s3 EP) -> s3 |
| Scenario  2. All traffic goes through proxy  | None  | Yes  | proxy -> internet gateway -> outside of VPC -> s3  |
| Scenario  3. Traffic remains in private networks between my VPC and AWS service | Yes, EPs in noProxy  | Yes |  noProxy -> VPC (with s3 EP) -> s3 |
| Scenario  4. Set EPs and Proxy | Yes  | Yes |   |


@sherine-k, I got a technical question here, what do you think how the traffic goes in scenario 4? maybe this is an unusual configuration.

## Non-workable scenario
| Item  | Endpoints  | Proxy | How network traffic goes (s3 as example) | 
| ------------- | ------------- | ------------- | ------------- |
| Wrong scenario  1 |  None  | None  | VPC -> _BLOCKED_, since no EP or proxy available, can not fetch bootstrap ignition file from s3 |

sherine-k:
Hi @yunjiang29 
These scenarios are exactly what we need to verify this fix/feature. Great! Thanks!
To me scenario 4 depends on securityGroups and routes that apply to the proxy and EP: 
* If proxy has routes and is allowed to connect to the endpoints, installer traffic would go proxy -> EP ->S3.
* If for any weird reason, infra engineers applied securityGroup rules to the EP preventing it from accepting ingress traffic from the proxy/proxy secGroup/subnet or if the proxy is in a subnet with no route to the EP, installer will fail. [AWS doc](https://docs.aws.amazon.com/vpc/latest/privatelink/create-interface-endpoint.html#create-interface-endpoint-aws)

PS: Yesterday late in the evening, we decided to merge this to 4.11GA as well (https://github.com/openshift/installer/pull/6090). Would it be possible for you to validate that PR as a priority? ([Bugzilla bug 2104727](https://bugzilla.redhat.com/show_bug.cgi?id=2104727))

Thanks again so much

yunjiang29:
Thanks @sherine-k , good to know network traffic of scenario 4.

> Would it be possible for you to validate that PR as a priority? ([Bugzilla bug 2104727](https://bugzilla.redhat.com/show_bug.cgi?id=2104727))

Sure, I got the notification this morning, once nightly build get ready (maybe next Monday), I will verify it asap.

PR #5975, opened 2022-06-07
Improve user interface of agent installer

zaneb:
* ZTP manifests go in `cluster-manifests`
* Output ISO goes in work directory directly (i.e. same dir as install-config.yaml)
* Suppress debug logs by default


rwsu (commented):
Small unit test issue. In nmstateconfig_test.go, manifests directory should be changed to cluster-manifests. Looks good though.

PR #5977, opened 2022-06-08
AGENT-133: Remove double embedding of agent data

zaneb:
The code in from the fleeting prototype relied on the embed module to
embed the data in data/data/agent. Since that code has now been removed,
and replaced with native installer code, in #5945 the agent data is now
accessed through data.Assets. This second copy of it is no longer
needed.


rwsu (approved):
/approve

PR #5979, opened 2022-06-08
Bug 2090816: Make bootstrap timeout configurable

honza:
We add a new `bootstrap-timeout` flag to the `create cluster` command.

### DIFF
--- a/cmd/openshift-install/create.go
+++ b/cmd/openshift-install/create.go
@@ -40,10 +40,20 @@ import (
 	"github.com/openshift/library-go/pkg/route/routeapihelpers"
 )
 
+var (
+	createOpts struct {
+		bootstrapTimeout int
+	}
+)
+
 type target struct {
 	name    string
 	command *cobra.Command
 	assets  []asset.WritableAsset
+
+	// commandF can be used to alter the `command` above; this is necessary for
+	// things like persistent flags
+	commandF func(cmd *cobra.Command)
 }
 
 const (
@@ -166,6 +176,9 @@ var (
 				timer.LogSummary()
 			},
 		},
+		commandF: func(cmd *cobra.Command) {
+			cmd.PersistentFlags().IntVar(&createOpts.bootstrapTimeout, "bootstrap-timeout", 20, "Bootstrap timeout in minutes")
+		},
 		assets: targetassets.Cluster,
 	}
 
@@ -233,6 +246,9 @@ func newCreateCmd() *cobra.Command {
 	for _, t := range targets {
 		t.command.Args = cobra.ExactArgs(0)
 		t.command.Run = runTargetCmd(t.assets...)
+		if t.commandF != nil {
+			t.commandF(t.command)
+		}
 		cmd.AddCommand(t.command)
 	}
 
@@ -349,7 +365,7 @@ func waitForBootstrapComplete(ctx context.Context, config *rest.Config) *cluster
 
 	discovery := client.Discovery()
 
-	apiTimeout := 20 * time.Minute
+	apiTimeout := time.Duration(createOpts.bootstrapTimeout) * time.Minute
 	untilTime := time.Now().Add(apiTimeout)
 	logrus.Infof("Waiting up to %v (until %v) for the Kubernetes API at %s...",
 		apiTimeout, untilTime.Format(time.Kitchen), config.Host)


hardys:
As someone who's mostly used the installer for baremetal (where the delays related to the environment can vary a lot) this makes sense to me (although I suspect we'd ideally want to expose some way to control all of the timeouts, not only the bootstrap-complete one).

However previous feedback from @patrickdillon indicated there may be an installer team preference to just increasing the hard-coded timeout (just for baremetal) similar to [#2979](https://github.com/openshift/installer/pull/2979)

Personally I prefer the configurable timeout approach, since it avoids making installs fail-slow when there are not specific issues (like slow POST times, network bandwidth constraints etc), and also it's hard to choose any one hard-coded value that will always work on-prem, because the delays can vary so much unlike in typical cloud scenarios where things are more deterministic.

sadasu:
I could be wrong but do we have a default value for this configurable bootstrap-timeout flag?

honza:
> I could be wrong but do we have a default value for this configurable bootstrap-timeout flag?

[Yes](https://github.com/openshift/installer/pull/5979/files#diff-56276d5381d618d46ec8d35d93210662c8fdd4c9bcd90fd36afbc6a59227eb0bR180)

sadasu:
> > I could be wrong but do we have a default value for this configurable bootstrap-timeout flag?
> 
> [Yes](https://github.com/openshift/installer/pull/5979/files#diff-56276d5381d618d46ec8d35d93210662c8fdd4c9bcd90fd36afbc6a59227eb0bR180)

Cool! I missed that.

patrickdillon:
Thanks for putting this together with the alternative #5981. In general, this timeout should only be for baremetal (or perhaps other on-prem environs too)--not cloud. Hopefully the reason is clear, but I can explain further if needed. 

That cross-platform approach is one issue with this implementation. I'm also trying to check whether this flag applies to all `create` commands or just `create cluster`. I'm struggling to get this to work and the flag is undocumented:
```sh
$ ./openshift-install create cluster --help
Create an OpenShift cluster

Usage:
  openshift-install create cluster [flags]

Flags:
  -h, --help   help for cluster

Global Flags:
      --dir string         assets directory (default ".")
      --log-level string   log level (e.g. "debug | info | warn | error") (default "info")
$ ./openshift-install create cluster --dir gcp-test --bootstrap-timeout 30
FATA[0000] Error executing openshift-install: unknown flag: --bootstrap-timeout
```

These reasons suggest to me that a flag is not the right approach for this. 

> Personally I prefer the configurable timeout approach, since it avoids making installs fail-slow when there are not specific issues (like slow POST times, network bandwidth constraints etc), and also it's hard to choose any one hard-coded value that will always work on-prem, because the delays can vary so much unlike in typical cloud scenarios where things are more deterministic.

I take @hardys point here, and I agree that we don't want to degrade the general experience to enable potentially a handful of environments. I will point out that the downside for a configurable timeout is that it would still require configuration for those problematic environments to work, rather than working out of the box. But with the understanding those troublesome environments are the exception and not the norm, we agree it is better to require the extra configuration for those failing environments rather than potentially fail slow in all cases.

If the flag solution is not going to work, I see two potential solutions for allowing this to be configurable:

- environment variable - generally we avoid env vars and we don't document them, but I feel like this might fit the pattern where they have been applied in the past. I would need to double check (@sdodson do you have a sense?) whether we have restrictions about customers using them at all, or whether it would be suitable in cases like this
- add a field in the baremetal platform in the install config - I don't like this idea, but it is technically possible

sdodson:
> environment variable - generally we avoid env vars and we don't document them, but I feel like this might fit the pattern where they have been applied in the past. I would need to double check (@sdodson do you have a sense?) whether we have restrictions about customers using them at all, or whether it would be suitable in cases like this

Yes, we maintain a firm stance that environment variables are only suitable for installer development uses. Having read the discussion in this PR I guess a flag scoped to `cluster create` is OK.

patrickdillon:
> I'm not sure what happened there.

Definitely a mistake on my end!

> Yes, we maintain a firm stance that environment variables are only suitable for installer development uses. Having read the discussion in this PR I guess a flag scoped to cluster create is OK.

Thanks Scott. As environment variables are out of the question, I think the install config seems like the only possible option for providing configurable timeouts. 

Even if we do allow the timeout to be configurable across platforms (which I don't think is a good idea), we don't generally have install configuration through flags. At the minimum this would create a documentation problem.

hardys:
> Even if we do allow the timeout to be configurable across platforms (which I don't think is a good idea), we don't generally have install configuration through flags. At the minimum this would create a documentation problem.

Most configuration is done via the install-config, which describes the desired state of the cluster - but a flag seems like a reasonable approach for something which modifies the default behavior of the installer?

I guess we could make this platform-specific (e.g only enabled via [baremetal tag](https://github.com/openshift/installer/blob/master/images/baremetal/Dockerfile.ci#L8) for the openshift-baremetal-install binary?), but are we sure users will *never* need this escape-hatch option on other platforms? (for example I can imagine vsphere/openstack and any on-prem platforms having less deterministic performance than most clouds)

sdodson:
The rationale from @hardys is sound and with @sadasu LGTM we'll move forward with this.
/approve

cmd/openshift-install/create.go
```diff
@@ -40,10 +40,20 @@ import (
...
+)
+
 type target struct {
 	name    string
 	command *cobra.Command
 	assets  []asset.WritableAsset
+
+	// commandF can be used to alter the `command` above; this is necessary for
+	// things like persistent flags
+	commandF func(cmd *cobra.Command)
```
rna-afk, line 56: Not sure why we need this here? It doesn't look like we need this for all commands so why don't we add it only in the newCreateCmd() function?

```

func newCreateCmd() *cobra.Command {
	cmd := &cobra.Command{
		Use:   "create",
		Short: "Create part of an OpenShift cluster",
		RunE: func(cmd *cobra.Command, args []string) error {
			return cmd.Help()
		},
	}
cmd.PersistentFlags().IntVar(&createOpts.bootstrapTimeout, "bootstrap-timeout", 20, "Bootstrap timeout in minutes")
```

honza: This is a general solution: each subcommand can now modify the `cobra.Command` instance independently. Setting the bootstrap timeout only makes sense when creating the cluster.

rna-afk: Got it my bad. Was hoping to find a way to not add commandF for just one use case but I guess we can't.

honza:
Closing this as #5981 has merged

PR #5980, opened 2022-06-08
Bug 2057582: Removing all occurrences of Packet from Metal UPI

sadasu:
Packet.net has moved to Equinix Metal and these set of changes are
adjusting to this change in the Meatl Terraform Provider.


r4f4 (approved):

sadasu:
Metal UPI terraform templates are being removed so this is no longer relavent.

PR #5981, opened 2022-06-08
Bug 2090816: baremetal: wait longer for bootstrap to complete

honza:
This is a baremetal-specific alternative to #5979

### DIFF
--- a/cmd/openshift-install/create.go
+++ b/cmd/openshift-install/create.go
@@ -350,6 +350,16 @@ func waitForBootstrapComplete(ctx context.Context, config *rest.Config) *cluster
 	discovery := client.Discovery()
 
 	apiTimeout := 20 * time.Minute
+
+	// Wait longer for baremetal, due to length of time it takes to boot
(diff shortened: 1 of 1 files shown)


cmd/openshift-install/create.go
```diff
@@ -350,6 +350,16 @@ func waitForBootstrapComplete(ctx context.Context, config *rest.Config) *cluster
 	discovery := client.Discovery()
 
 	apiTimeout := 20 * time.Minute
+
+	// Wait longer for baremetal, due to length of time it takes to boot
+	if assetStore, err := assetstore.NewStore(rootOpts.dir); err == nil {
```
barbacbd, line 355: Same code is used in multiple places, should a function be used?

sadasu: Identical logic here https://github.com/openshift/installer/blob/master/cmd/openshift-install/create.go#L449-L455 while waiting for the cluster to be initialized.

PR #5986, opened 2022-06-09
Bug 2094896: ensure all the required agent manifests are available

andfasano:
This patch adds a check to ensure a valid configuration manifests is available. As per the current implementation this means that the related manifest has been correctly loaded from the disk.


pkg/asset/agent/manifests/clusterdeployment.go
```diff
@@ -40,7 +40,7 @@ func (*ClusterDeployment) Dependencies() []asset.Asset {
 // Generate generates the ClusterDeployment manifest.
 func (i *ClusterDeployment) Generate(dependencies asset.Parents) error {
```
pawanpinjarkar, line 41: instead of calling it just `i`, maybe change it to `cd`, an acronym for ClusterDeployment.

pkg/asset/agent/manifests/clusterimageset.go
```diff
@@ -36,7 +36,7 @@ func (*ClusterImageSet) Dependencies() []asset.Asset {
 
 // Generate generates the ClusterImageSet manifest.
 func (a *ClusterImageSet) Generate(dependencies asset.Parents) error {
```
pawanpinjarkar, line 38: Similar to other acronym comments above.

pawanpinjarkar (changes requested):
Some minor suggestions.

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -73,7 +74,7 @@ func (a *AgentClusterInstall) Generate(dependencies asset.Parents) error {
 	// 	Data:     agentClusterInstallData,
 	// }
 
-	return nil
+	return a.finish()
```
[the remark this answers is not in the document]
zaneb: Not a bad idea, but this would require changing it in every method, so I feel like that's a separate PR.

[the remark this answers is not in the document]
pawanpinjarkar: yes, I agree with you @zaneb

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -104,5 +105,17 @@ func (a *AgentClusterInstall) Load(f asset.FileFetcher) (bool, error) {
...
+	if err = a.finish(); err != nil {
+		return false, err
+	}
 	return true, nil
 }
+
+func (a *AgentClusterInstall) finish() error {
+
+	if a.Config == nil {
+		return errors.New("missing configuration or manifest file")
```
rwsu, line 117: With this PR I'm seeing this error when my cluster-manifests directory is missing:

`
FATAL failed to fetch Agent Installer ISO: failed to fetch dependency of "Agent Installer ISO": failed to fetch dependency of "Agent Installer Ignition": failed to fetch dependency of "Agent Manifests": failed to generate asset "Agent PullSecret": missing configuration or manifest file 
`

Would it be possible to include the path of the missing file(s) in the error message?

andfasano: I avoided explicitly to specify the exact file path mainly because as soon as we'll enable the `Generate()` code - for the all the agent manifests assets - this check will not be required anymore, and the `Load()` will become optional (either the `Load()` will find some file or the `Generate()` will be able to create a new resource, in any case the shared `finish()` method will receive non-nil data to validate/manipulate. But if you think it will be more clear, even in this transitory period, I could specify it as well.

rwsu (approved):
/approve

PR #5990, opened 2022-06-10
WIP: agent: allow assets to choose how they are persisted to disk

patrickdillon:
WritableAsset is too opinionated about how files are written to disk.
This enables assets to choose how they wish to be written to disk
while providing a default implementation to avoid duplication.

This is a proof of concept. I have done some nominal testing to make sure it writes the basic assets for the installer.

Also there is still some obvious cleanup, but hopefully the idea comes through and we can evaluate its merit.


pkg/asset/agent/image/agentimage.go
```diff
@@ -80,3 +80,8 @@ func (a *AgentImage) Files() []*asset.File {
 	// Return empty array because File will never be loaded.
 	return []*asset.File{}
 }
+
+func (a *AgentImage) PersistToFile(directory string) error {
+	// Do some or all of the stuff from Generate here
+	return nil
```
zaneb, line 86: :+1: It would be easy enough to store the io stream in the struct and just write it to disk (ll. 51-63) here.

PR #5991, opened 2022-06-11
Bug 2094694: Nutanix cluster name validation

thunderboltsid:
Ensure Nutanix is subject to the same cluster name validation as
other on prem platforms.


patrickdillon:
I think we want to do this for all platforms, but this is fine until we do that.

/approve

PR #5992, opened 2022-06-13
AGENT-214: Mount mirror registries.conf for assisted-service

bfournie:
Mount the registries.conf file (if created) that defines the mirror registry location when starting assisted-service, add the ca-bundle used to access the mirror registry, and add the mirror files to ignition. Set the assisted-service env variable for the release image mirror.


data/data/agent/files/usr/local/bin/start-assisted-service.sh.template
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+MIRROR_IMAGES_MOUNT=""
+
+if [ -f  '{{.MirrorRegistriesFilename}}' ]; then
```
andfasano, line 7: What about using a template directly for the `assisted-service.service` file, and using instead the template conditionals to optionally mount the registies.conf? In that way this file will not be required

pkg/asset/agent/mirror/mirror.go
```diff
@@ -55,3 +58,31 @@ func (m *AgentMirror) Load(f asset.FileFetcher) (bool, error) {
...
 }
+
+// GetReleaseImageMirror returns the mirror corresponding to the release image
+func (m *AgentMirror) GetReleaseImageMirror(releaseImage string) string {
+	// If registries.conf is defined, find mirror stripping the version out, e.g. for
+	// registry.ci.openshift.org/ocp/release:4.11.0-0.nightly-2022-06-06-201913, use
+	// registry.ci.openshift.org/ocp/release
+	source := strings.Split(":", releaseImage)
+
+	for _, file := range m.FileList {
```
andfasano, line 69: I'm not sure about the reason of having this portion code as a separate asset. Wouldn't be easier to move it into the  `RegistriesConf` asset and expose the mirror string as an asset member?

pkg/asset/agent/mirror/mirror.go
```diff
@@ -55,3 +58,31 @@ func (m *AgentMirror) Load(f asset.FileFetcher) (bool, error) {
...
+			}
+			mirrorIndex := bytes.Index([]byte("[[registry.mirror]]"), file.Data[imageIndex+1:])
+			if mirrorIndex == -1 {
+				return ""
+			}
+			// get mirror from "location = xxx"
+			fields := bytes.Fields(file.Data[mirrorIndex+1:])
+			if len(fields) > 2 && string(fields[0]) == "location" {
+				return string(fields[2])
+			}
```
andfasano, line 83: As a personal preference, I'd find easier maybe to develop/use a standalone parser specific for such file

data/data/agent/files/usr/local/bin/start-assisted-service.sh.template
bfournie: Good point. Changed to use template for the service and removed script.

pkg/asset/agent/mirror/mirror.go
bfournie: Changed to add function in registriesconfig asset that will build the source/mirror mapping and can be used with the registry  to find the mirror. Function was copied from assisted-service to avoid importing entire pkg.

pkg/asset/agent/mirror/mirror.go
bfournie: Yes the function will now parse the entire file and store it in an array of structs with source and mirror.

pkg/asset/agent/image/ignition.go
```diff
@@ -108,10 +116,28 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
...
+
+	mirrorRegistriesMount := ""
+	if _, err := os.Stat(mirror.RegistriesConfFilename); err == nil {
+		mirrorRegistriesMount = fmt.Sprintf("-v %s:/etc/containers/registries.conf:z", filepath.Join("/etc/assisted", mirror.RegistriesConfFilename))
+	}
+	releaseImageMirror := ""
+	for _, config := range agentMirror.MirrorConfig {
+		// Use the source location not including the version
+		source := strings.Split(":", agentManifests.ClusterImageSet.Spec.ReleaseImage)
+		if config.Location == source[0] {
```
rwsu, line 130: Could this also be a validation? To check that if MirrorConfigs are provided, at least one of the MirrorConfigs matches the ClusterImageSet's ReleaseImage?

bfournie: Good point, I can get the ClusterImageSet info in the RegistriesConf asset and add that check.

andfasano:
@bfournie I'd recommend to split at least the commit in two, one just for vendoring the new module, and the other one with the PR code

bfournie:
> @bfournie I'd recommend to split at least the commit in two, one just for vendoring the new module, and the other one with the PR code

I will split out the addition of the toml pkg,

pkg/asset/agent/image/ignition.go
bfournie: Actually I can't add the validation in the RegistriesConf asset since I can't get access to another asset (i.e. ClusterImageSet) in the Load() but I should be able to add the validation in the mirror asset as I can add ClusterImageSet as a dependency there. I will follow up in a separate PR with the validation.

pkg/asset/agent/image/ignition.go
```diff
@@ -108,10 +117,38 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
 		return err
 	}
 
+	agentMirror := &mirror.AgentMirror{}
+	dependencies.Get(agentMirror)
+
+	// Mount the mirror registries file for assisted-service
+	mirrorRegistriesMount := ""
+	if _, err := os.Stat(mirror.RegistriesConfFilename); err == nil {
```
andfasano, line 125: Since the mirror dependency it's a writable asset, wouldn't be better to rely just on some asset property (like the `File` or `FileList`, or another one created ad-hoc) to decide if some content it's available or not, and avoiding to access directly the disk from here?

bfournie: Yep, good point, thanks. Fixed.

pkg/asset/agent/image/ignition.go
```diff
@@ -108,10 +117,38 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
...
+	for _, config := range agentMirror.MirrorConfig {
+		if config.Location == source[0] {
+			// include the tag with the build release image
+			releaseImageMirror = fmt.Sprintf("%s:%s", config.Mirror, source[1])
+		}
+	}
+
+	// Mount the CA bundle for assisted-service
+	caBundleMount := ""
+	if _, err := os.Stat(mirror.CaBundleFilename); err == nil {
```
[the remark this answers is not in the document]
bfournie: Fixed.

bfournie:
> > @bfournie I'd recommend to split at least the commit in two, one just for vendoring the new module, and the other one with the PR code
> 
> I will split out the addition of the toml pkg,

Done

PR #5993, opened 2022-06-13
AGENT-218: Agent lock release version

andfasano:
This patch adds a new validation rule for the ClusterImageSet asset to ensure that the configured version is matching the current one supported by the installer

### DIFF
--- a/pkg/asset/agent/manifests/clusterimageset.go
+++ b/pkg/asset/agent/manifests/clusterimageset.go
@@ -7,7 +7,9 @@ import (
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	"github.com/openshift/installer/pkg/asset"
+	"github.com/openshift/installer/pkg/asset/releaseimage"
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"sigs.k8s.io/yaml"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/asset/agent/manifests/clusterimageset.go
```diff
@@ -67,5 +103,47 @@ func (a *ClusterImageSet) Load(f asset.FileFetcher) (bool, error) {
...
+func (a *ClusterImageSet) validateReleaseVersion() field.ErrorList {
+
+	var allErrs field.ErrorList
+
+	fieldPath := field.NewPath("Spec", "ReleaseImage")
+
+	releaseImage := &releaseimage.Image{}
+	releaseImage.Generate(asset.Parents{})
+
+	if a.Config.Spec.ReleaseImage != releaseImage.PullSpec {
```
zaneb, line 144: I don't know that we need to be _this_ strict. I am happy if the minor version (4.y) matches. It may be that we don't have enough information to verify in that case?

My guess is that this will be fine if you either download the installer from cloud.redhat.com and use the hosted payload, or if you extract the installer from a mirror. But it may prevent you from using the downloaded installer with a mirrored payload. I think that's acceptable iff this is already the case for IPI.

andfasano: This approach AFAIK continues to honor the already existing one in the standard installer, where it's not possible to install a release version different from the one hard-codeed in the binary (unless overridden through the env var). For the mirrored payload I guess that this approach should work with what @bfournie is implementing in https://github.com/openshift/installer/pull/5992.
(and yes, unfortunately we don't have enough information to verify if it's a minor version or not)

zaneb: OK, I think this is fine. The built-in pullspec appears to always be a hash anyway. Let's go ahead.

pkg/asset/agent/manifests/clusterimageset.go
```diff
@@ -31,11 +33,45 @@ func (*ClusterImageSet) Name() string {
...
-	return []asset.Asset{}
+	return []asset.Asset{
+		// &releaseimage.Image{},
+		// &installconfig.InstallConfig{},
+	}
 }
 
 // Generate generates the ClusterImageSet manifest.
 func (a *ClusterImageSet) Generate(dependencies asset.Parents) error {
+
```
rwsu, line 44: I'm guessing this will be turned on in the future when we are generating the ZTP manifests from install-config.yaml. When that is active the ClusterImageSet will be generated for the user if one is not provided.

rwsu (commented):
/lgtm

pkg/asset/agent/manifests/clusterimageset.go
andfasano: Yes

zaneb (commented):
/approve

PR #5994, opened 2022-06-13
[CFE-509] As a developer I want to update installer to use propagateUserFlags for GA

TrilokGeer:
Updated the experimental field for to use a formal field _propagateUserTag_

### DIFF
--- a/pkg/asset/installconfig/installconfig_test.go
+++ b/pkg/asset/installconfig/installconfig_test.go
@@ -249,6 +249,187 @@ pullSecret: "{\"auths\":{\"example.com\":{\"auth\":\"authorization value\"}}}"
 				Publish:    types.ExternalPublishingStrategy,
 			},
 		},
+		{
+			name: "experimentalPropagateUserTags takes precedence",
+			data: `
+apiVersion: v1
+metadata:
+  name: test-cluster
+baseDomain: test-domain
+platform:
+  aws:
+    region: us-east-1
+    experimentalPropagateUserTags: false
+    propagateUserTags: true
+pullSecret: "{\"auths\":{\"example.com\":{\"auth\":\"authorization value\"}}}"
+`,
+			expectedFound: true,
+			expectedConfig: &types.InstallConfig{
+				TypeMeta: metav1.TypeMeta{
+					APIVersion: types.InstallConfigVersion,
+				},
+				ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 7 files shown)


pkg/asset/installconfig/installconfig_test.go
```diff
@@ -249,6 +249,187 @@ pullSecret: "{\"auths\":{\"example.com\":{\"auth\":\"authorization value\"}}}"
...
+						Name:           "worker",
+						Replicas:       pointer.Int64Ptr(3),
+						Hyperthreading: types.HyperthreadingEnabled,
+						Architecture:   types.ArchitectureAMD64,
+					},
+				},
+				Platform: types.Platform{
+					AWS: &aws.Platform{
+						Region:                       "us-east-1",
+						ExperimentalPropagateUserTag: nil,
```
patrickdillon, line 365: I'm surprised `nil` is allowed here because the type is boolean:
https://github.com/openshift/installer/pull/5994/files#diff-7941d8da46384fc43b6eeae7d5c3319065b16940457822346c9ccb08cbcf47b4R55

patrickdillon: D'oh. My mistake I see you upgraded the type to pointer. just missed the asterisk

yunjiang29:
@TrilokGeer, I did a quick test for this PR, `experimentalPropagateUserTags` and `propagateUserTags` work as expected.

TrilokGeer:
Thanks for the tests, @yunjiang29

pkg/types/conversion/installconfig.go
```diff
@@ -137,3 +142,12 @@ func convertOpenStack(config *types.InstallConfig) error {
 
 	return nil
 }
+
+// convertAWS upconverts deprecated fields in the OpenStack platform.
```
patrickdillon, line 146: copypasta: `OpenStack` -> `AWS`

patrickdillon (commented):
This seems like it is in pretty good shape. It is good to have several small commits, but some of these need to be squashed: we don't want to introduce changes in one commit and change them in another.

Also it seems like the tests should be moved to the types pkg. 

Finally, can you add meaningful commit messages as according to the [contributing guidelines](https://github.com/openshift/installer/blob/master/CONTRIBUTING.md)?

jstuever (commented):
/lgtm

PR #5995, opened 2022-06-13
Bug 2070744: Fix infinite loop when failing to untag resources

rna-afk:
If AWS fails to untag a resource, it ends up going into an
infinite loop of querying tag client for resources and
retrying the untag process.

Removing the part that adds the tag client back to the list
and causes the retries.

### DIFF
--- a/pkg/destroy/aws/shared.go
+++ b/pkg/destroy/aws/shared.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi"
@@ -92,8 +93,14 @@ func (o *ClusterUninstaller) removeSharedTag(ctx context.Context, session *sessi
 				},
 			)
 			if err != nil {
-				err = errors.Wrap(err, "get tagged resources")
-				o.Logger.Info(err)
+				err2 := errors.Wrap(err, "get tagged resources")
+				o.Logger.Info(err2)
+				if aerr, ok := err.(awserr.Error); ok {
+					switch aerr.Code() {
+					case resourcegroupstaggingapi.ErrorCodeInvalidParameterException:
+						continue
+					}
+				}
 				nextTagClients = append(nextTagClients, tagClient)
 				continue
 			}


pkg/destroy/aws/shared.go
```diff
@@ -101,26 +101,32 @@ func (o *ClusterUninstaller) removeSharedTag(ctx context.Context, session *sessi
...
+					if err != nil {
+						err = errors.Wrap(err, "untag shared resources")
+						o.Logger.Info(err)
 						continue
 					}
-					o.Logger.WithField("arn", *arn).Infof("Removed tag %s: shared", key)
-					removed[*arn] = exists
+				}
+
+				if err == nil {
```
barbacbd, line 120: Can/should we consider adding the error to the return value if the previous statements were successful?

pkg/destroy/aws/shared.go
```diff
@@ -101,26 +101,32 @@ func (o *ClusterUninstaller) removeSharedTag(ctx context.Context, session *sessi
...
-				if err != nil {
-					err = errors.Wrap(err, "untag shared resources")
-					o.Logger.Info(err)
-					continue
-				}
-				for _, arn := range request.ResourceARNList {
-					if info, failed := result.FailedResourcesMap[*arn]; failed {
-						o.Logger.WithField("arn", *arn).Infof("Failed to remove tag %s: shared; error=%s", key, *info.ErrorMessage)
+				var result *resourcegroupstaggingapi.UntagResourcesOutput
+				for retry := 0; retry < 3; retry++ {
```
barbacbd, line 111: What is the significance behind 3 retries ?

rna-afk: It's arbitrary :). Just hoping to retry this piece of code for sometime before we decide the operation errored out.

pkg/destroy/aws/shared.go
rna-afk: Wasn't there before. Could add it.

rna-afk: I think the code below takes care of searching for the IAM roles that still exist and handle errors.

pkg/destroy/aws/shared.go
rna-afk: It's all async calls to the AWS APIs at this point so there's a possibility that all these will fail which is why the previous solution was to keep running this code until it's successful.

patrickdillon (commented):
I'm not fully understanding the bug and therefore the solution. 

Your comment says:

>  If AWS fails to untag a resource, it ends up going into an
> infinite loop of querying tag client for resources and
> retrying the untag process.

Is that really what's happening? How did you determine this? The logs in BZ are so abbreviated, I can't really tell what's going on. But from those two lines in the BZ and from reading the code, it looks like destroy does not get stuck where you have added the three retries.  

It looks like it is getting stuck after the log line for `Search for and remove tags...` 

It's not clear to me how adding three retries  is going to fix the problem.

When the logs look like:
```
INFO untag shared resources: InvalidParameterException: Invocation of UntagResources for this resource is not supported in this region
DEBUG Search for and remove tags in us-gov-west-1 matching kubernetes.io/cluster/test-cluster-bsmt4: shared
```
Are we just hanging indefinitely at that last call? Or is this just an excerpt, and we go into a loop where this is repeated over and over? If you have been able to reproduce the problem, can you provide more context? If not, I think we need to get better logs or reproduce this.

LMK if I am missing something.

patrickdillon:
@yunjiang29 do you have the ability to premerge test this? It would be very useful here.

patrickdillon:
> My basic understanding is that we run the outer loop until we have exhausted all the tag clients but we also add the current tag client back into the list [here](https://github.com/openshift/installer/blob/6daed68b9863a9b2ecebdf8a4056800aa5c60ad3/pkg/destroy/aws/shared.go#L104) which might cause the loop to run infinite times if the piece of code after it fails. I am trying to patch that by trying to run the latter piece of code 3 times before moving out of this loop.

It looks like the issue is actually in this block, where the client is added back in due to the error: https://github.com/openshift/installer/blob/6daed68b9863a9b2ecebdf8a4056800aa5c60ad3/pkg/destroy/aws/shared.go#L94-L98

So rather than add an arbitrary limit (of 3 tries), I think it would be better to skip adding the client back into  `nextTagClients` when we know the error is unrecoverable (in this case, I guess that would be `resourcegroupstaggingapi.InvalidParameterException`). cf https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/handling-errors.html

yunjiang29:
@patrickdillon sure, but I got an error while building payload using cluster bot today, will do pre-merge test once the issue gets fixed.

yunjiang29:
@patrickdillon sure, but I got an error while building payload using cluster bot today, will do pre-merge test once the issue gets fixed.

yunjiang29:
the [destroy log](https://privatebin.corp.redhat.com/?2c28e54470e05ead#FoSTdDVgikRwChhMgHbzmveeegWxWQjP9xAGTcgnBn3T) that reproduces the bug, the process was killed by automation framework b/c of timeout

sadasu:
/approve

Figuring out the actual error was helpful.

PR #6000, opened 2022-06-14
terraform: update azure provider to v3

r4f4:
If we do not upgrade, the old provider will continue to use the
deprecated ADAL endpoints. Those endpoints will continue to respond
according to the deprecation announcement, but they will not received
updates.

https://issues.redhat.com/browse/CORS-1911


jstuever:
/cc @patrickdillon 
I'm not sure the consequences of removing `resource_group_name` like this.

jstuever (commented):
/approve

r4f4:
> /cc @patrickdillon I'm not sure the consequences of removing `resource_group_name` like this.

There is an open issue regarding the parameter removal: https://github.com/hashicorp/terraform-provider-azurerm/issues/16106. It didn't seem to affect my local installs, but that's definitely not enough evidence.

jhixson74:
> /cc @patrickdillon I'm not sure the consequences of removing `resource_group_name` like this.

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/guides/3.0-upgrade-guide#data-source-azurerm_lb_rule indicates the resource group name is inferred from the loadbalance id.

PR #6002, opened 2022-06-14
Bug 2096905: Swap Nutanix Prism Client

thunderboltsid:
Use[ prism-go-client](https://github.com/nutanix-cloud-native/prism-go-client) instead of embedded client in nutanix terraform provider for the installer. The underlying client codebase is the same but Nutanix is moving the Go client out into its own repository (away from a package inside the terraform provider).


sgaoshang (commented):
lgtm

PR #6004, opened 2022-06-15
[WIP] AGENT-187: Asset for Agent CRD

rwsu:
The hostname, ignition, and installer args from the Agent CRD are
used for host customization.

Currently WIP, so it also contains the client code for AGENT-188. This will be separated into another PR.


rwsu:
#6022 replaces this PR.

PR #6014, opened 2022-06-15
Change defaultNetworkType to ovn-kubernetes

jluhrsen:
Signed-off-by: Jamo Luhrsen <jluhrsen@gmail.com>

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -1477,9 +1477,9 @@ spec:
                   type: object
                 type: array
               networkType:
+                default: OVNKubernetes
                 description: NetworkType is the type of network to install. The default
-                  value is OVNKubernetes for Single Node OpenShift and OpenShiftSDN
-                  for all other platforms.
+                  value is OVNKubernetes.
                 type: string
               serviceCIDR:
                 description: Deprecated way to configure an IP address pool for services.
--- a/pkg/types/conversion/installconfig.go
+++ b/pkg/types/conversion/installconfig.go
@@ -76,7 +76,7 @@ func convertNetworking(config *types.InstallConfig) {
 		netconf.NetworkType = netconf.DeprecatedType
 	}
 
-	// Recognize the default network plugin name regardless of capitalization, for
+	// Recognize the OpenShiftSDN network plugin name regardless of capitalization, for
 	// backward compatibility
 	if strings.ToLower(netconf.NetworkType) == strings.ToLower(string(operv1.NetworkTypeOpenShiftSDN)) {
 		netconf.NetworkType = string(operv1.NetworkTypeOpenShiftSDN)
--- a/pkg/types/defaults/installconfig.go
+++ b/pkg/types/defaults/installconfig.go
@@ -24,7 +24,7 @@ var (
 	defaultServiceNetwork = ipnet.MustParseCIDR("172.30.0.0/16")
 	defaultClusterNetwork = ipnet.MustParseCIDR("10.128.0.0/14")
 	defaultHostPrefix     = 23
-	defaultNetworkType    = string(operv1.NetworkTypeOpenShiftSDN)
+	defaultNetworkType    = string(operv1.NetworkTypeOVNKubernetes)
 )
 
 // SetInstallConfigDefaults sets the defaults for the install config.
@@ -48,11 +48,7 @@ func SetInstallConfigDefaults(c *types.InstallConfig) {
 		}
 	}
 	if c.Networking.NetworkType == "" {
-		if c.IsOKD() || c.IsSingleNodeOpenShift() {
-			c.Networking.NetworkType = string(operv1.NetworkTypeOVNKubernetes)
-		} else {
-			c.Networking.NetworkType = defaultNetworkType
-		}
+		c.Networking.NetworkType = defaultNetworkType
 	}
 	if len(c.Networking.ServiceNetwork) == 0 {
 		c.Networking.ServiceNetwork = []ipnet.IPNet{*defaultServiceNetwork}
--- a/pkg/types/validation/installconfig_test.go
+++ b/pkg/types/validation/installconfig_test.go
@@ -180,7 +180,7 @@ func validNutanixPlatform() *nutanix.Platform {
 
 func validIPv4NetworkingConfig() *types.Networking {
 	return &types.Networking{
-		NetworkType: "OpenShiftSDN",
+		NetworkType: "OVNKubernetes",
 		MachineNetwork: []types.MachineNetworkEntry{
 			{
 				CIDR: *ipnet.MustParseCIDR("10.0.0.0/16"),
@@ -455,7 +455,7 @@ func TestValidateInstallConfig(t *testing.T) {
 			name: "cluster network host prefix unset",
 			installConfig: func() *types.InstallConfig {
 				c := validInstallConfig()
-				c.Networking.NetworkType = "OpenShiftSDN"
+				c.Networking.NetworkType = "OVNKubernetes"
 				c.Networking.ClusterNetwork[0].CIDR = *ipnet.MustParseCIDR("192.168.1.0/24")
 				c.Networking.ClusterNetwork[0].HostPrefix = 0
 				return c
--- a/docs/user/customization.md
+++ b/docs/user/customization.md
@@ -44,7 +44,7 @@ The following `install-config.yaml` properties are available:
             The default is 10.0.0.0/16 for all platforms other than libvirt.
             For libvirt, the default is 192.168.126.0/24.
     * `networkType` (optional string): The type of network to install.
-        The default is [OVNKubernetes][ovn-kubernetes] for Single Node OpenShift and [OpenShiftSDN][openshift-sdn] for all other platforms.
+        The default is [OVNKubernetes][ovn-kubernetes].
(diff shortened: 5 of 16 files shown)


sadasu:
Is this being done for https://issues.redhat.com/browse/CORS-2085 ?

jluhrsen:
> Is this being done for https://issues.redhat.com/browse/CORS-2085 ?

well, I suppose, but I didn't really know about this Epic. This PR is initially to see it work (running tests with it). I'm sure
there is more work to be added to this.

patrickdillon:
Can you update the commit message(s) according to the format here: https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#commit-message-format

jluhrsen:
> Can you update the commit message(s) according to the format here: https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#commit-message-format

updated. LMK if it's not good enough and I'll clean it up further.

jluhrsen:
> Also it looks like some of the unit tests are failing. We can help with troubleshooting that if they prove problematic

should be good now.

dgoodwin:
/lgtm

From TRT perspective we've roughly matched or exceeded parity with SDN jobs across the handful we have been monitoring.

abhat:
@sadasu @patrickdillon - does this PR look good to go?

pkg/types/defaults/installconfig.go
```diff
@@ -24,7 +24,7 @@ var (
 	defaultServiceNetwork = ipnet.MustParseCIDR("172.30.0.0/16")
 	defaultClusterNetwork = ipnet.MustParseCIDR("10.128.0.0/14")
 	defaultHostPrefix     = 23
-	defaultNetworkType    = string(operv1.NetworkTypeOpenShiftSDN)
+	defaultNetworkType    = string(operv1.NetworkTypeOVNKubernetes)
```
sadasu, line 27: Logic around https://github.com/openshift/installer/blob/master/pkg/types/defaults/installconfig.go#L50-L56 also needs to be cleaned up.

sadasu (commented):
This PR can add kubebuilder default as OVNKubernetes. It was removed here https://github.com/openshift/installer/pull/5927/files#diff-ab4fb360ac4a8344c506c58e1b799894584bcd0daf8f1216c118b11cc079d375L279 when there was no default that was valid for all platforms.

pkg/types/validation/installconfig_test.go
```diff
@@ -1277,28 +1277,6 @@ func TestValidateInstallConfig(t *testing.T) {
 			}(),
 			expectedError: `Invalid value: "IPv6": single-stack IPv6 is not supported for this platform`,
 		},
-		{
```
barbacbd, line 1280: Don't these tests still apply? It is still possible to select `OpenShiftSDN` as the network type, it is just not the default.

pkg/types/defaults/installconfig.go
jluhrsen: It's updated now @sadasu, I think it is much simpler now right, and just set to ovnk if ""

jluhrsen:
> This PR can add kubebuilder default as OVNKubernetes. It was removed here https://github.com/openshift/installer/pull/5927/files#diff-ab4fb360ac4a8344c506c58e1b799894584bcd0daf8f1216c118b11cc079d375L279 when there was no default that was valid for all platforms.

I'm not exactly sure if I updated for this properly. I simply added a single line like:

`	// +kubebuilder:default=OVNKubernetes
`

please let me know if this is not correct, and what I need to do. I'll take care of it asap.

pkg/types/validation/installconfig_test.go
jluhrsen: good catch. I added them back.

abhat:
@jluhrsen you need to change the expected installconfigs.yml to add the default tag in the yaml too.

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -1478,8 +1478,7 @@ spec:
                 type: array
               networkType:
                 description: NetworkType is the type of network to install. The default
```
abhat, line 1480: Need to add `default: OVNKubernetes` here if we will add kubebuilder validation.

jluhrsen: thank you. updated.

sadasu:
> > This PR can add kubebuilder default as OVNKubernetes. It was removed here https://github.com/openshift/installer/pull/5927/files#diff-ab4fb360ac4a8344c506c58e1b799894584bcd0daf8f1216c118b11cc079d375L279 when there was no default that was valid for all platforms.
> 
> I'm not exactly sure if I updated for this properly. I simply added a single line like:
> 
> ` // +kubebuilder:default=OVNKubernetes `
> 
> please let me know if this is not correct, and what I need to do. I'll take care of it asap.

Right before this comment https://github.com/openshift/installer/blob/master/pkg/types/installconfig.go#L283, please add 
 // +kubebuilder:default=OVNKubernetes , as you correctly pointed out earlier.

pkg/types/conversion/installconfig.go
```diff
@@ -78,8 +78,8 @@ func convertNetworking(config *types.InstallConfig) {
 
 	// Recognize the default network plugin name regardless of capitalization, for
 	// backward compatibility
-	if strings.ToLower(netconf.NetworkType) == strings.ToLower(string(operv1.NetworkTypeOpenShiftSDN)) {
-		netconf.NetworkType = string(operv1.NetworkTypeOpenShiftSDN)
+	if strings.ToLower(netconf.NetworkType) == strings.ToLower(string(operv1.NetworkTypeOVNKubernetes)) {
+		netconf.NetworkType = string(operv1.NetworkTypeOVNKubernetes)
```
danwinship, line 82: This should not be changed. Well, the comment should be changed to not say "default", but the purpose of this code is not "whatever the default networkType value is, we should recognize it case-insensitively", it's "the correct value of this field used to be `OpenshiftSDN` but then we fixed it to be `OpenShiftSDN`, but we need to continue matching both values for backward-compatibility with (very) old install-configs.

pkg/types/conversion/installconfig_test.go
```diff
@@ -149,25 +149,6 @@ func TestConvertInstallConfig(t *testing.T) {
 			},
 			expectedError: "no version was provided",
 		},
-		{
-			name: "deprecated OpenShiftSDN spelling",
```
danwinship, line 153: and likewise this should be preserved

pkg/types/defaults/installconfig.go
```diff
@@ -48,11 +48,7 @@ func SetInstallConfigDefaults(c *types.InstallConfig) {
 		}
 	}
 	if c.Networking.NetworkType == "" {
-		if c.IsOKD() || c.IsSingleNodeOpenShift() {
-			c.Networking.NetworkType = string(operv1.NetworkTypeOVNKubernetes)
-		} else {
-			c.Networking.NetworkType = defaultNetworkType
-		}
+		c.Networking.NetworkType = defaultNetworkType
```
danwinship, line 51: is this even needed? with the `kubebuilder:default`, won't it always have gotten filled in if the user didn't specify it?

danwinship (changes requested):

pkg/types/defaults/installconfig.go
jluhrsen: is there no other way for c.Networking.NetworkType == ""

I don't know. this doesn't hurt, but if you know for certain then I can remove it.

pkg/types/conversion/installconfig_test.go
jluhrsen: fixed

pkg/types/conversion/installconfig.go
jluhrsen: fixed

sdodson:
/retest-required
4.12.0-0.ci-2022-07-28-142749 with fixes for the acl issue has now been accepted.

patrickdillon:
Looking at the history of e2e-vsphere runs, we have had several successful installs and they are all failing with:

>   [sig-network][Feature:Network Policy Audit logging] when using openshift ovn-kubernetes should ensure acl logs are created and correct [Suite:openshift/conformance/parallel] 

[example](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/6014/pull-ci-openshift-installer-master-e2e-vsphere/1552651967061299200)

sdodson:
@patrickdillon I think 4.12.0-0.ci-2022-07-28-142749 included https://github.com/openshift/cluster-policy-controller/pull/82
I was hoping to see a successful vsphere run after that fix landed but feel free to override as you wish.

patrickdillon:
> @patrickdillon I think 4.12.0-0.ci-2022-07-28-142749 included [openshift/cluster-policy-controller#82](https://github.com/openshift/cluster-policy-controller/pull/82) I was hoping to see a successful vsphere run after that fix landed but feel free to override as you wish.

doh I see your earlier comment now. thanks for clarifying.

jluhrsen:
> @sdodson, @patrickdillon: are we ready to lgtm again or do we think the network policy audit logging failure on vsphere should be gating this PR?

vsphere passed

pkg/types/defaults/installconfig.go
danwinship: yeah, dunno. it doesn't hurt

PR #6015, opened 2022-06-16
Power VS config.json changes

23TNC:
Removing zone and region from config.json, and updating calls to NewBxClient that only require the apikey to call GetSessionCredsFromAuthFIle directly.  This was done to bypass the additional survey prompts for region and zone that are no longer pulled from config.json.

### DIFF
--- a/docs/user/openstack/install_upi.md
+++ b/docs/user/openstack/install_upi.md
@@ -39,8 +39,7 @@ of this method of installation.
     - [Bootstrap Ignition](#bootstrap-ignition)
       - [Edit the Bootstrap Ignition](#edit-the-bootstrap-ignition)
       - [Upload the Boostrap Ignition](#upload-the-boostrap-ignition)
-        - [Example 1: Swift](#example-1-swift)
-        - [Example 2: Glance image service](#example-2-glance-image-service)
+        - [Example: Glance image service](#example-glance-image-service)
     - [Create the Bootstrap Ignition Shim](#create-the-bootstrap-ignition-shim)
       - [Ignition file served by server using self-signed certificate](#ignition-file-served-by-server-using-self-signed-certificate)
     - [Master Ignition](#master-ignition)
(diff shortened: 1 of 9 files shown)


barbacbd (commented):
Can you rebase your branch off of master? Some of your code may need to change after.

mjturek:
Rather than removing region & zone from `config.json`, we could uwse the values from `config.json` in two cases:

1) During the survey, could you use the values from `config.json` default for Region and Zone during the survey?
2) If not found in the install-config, use the values from `config.json`

Thoughts?

23TNC:
The install-config doesnt exist prior to the survey.  The config.json file is read to fill out the install-config
Something that may make sense is to read from environment variables first, thus overriding the config.json file.
But the ask that this is filling us to prompt for region and zone in all cases during the survey

mjturek:
> The install-config doesnt exist prior to the survey. The config.json file is read to fill out the install-config Something that may make sense is to read from environment variables first, thus overriding the config.json file. But the ask that this is filling us to prompt for region and zone in all cases during the survey

Sorry, 1 and 2 are two separate cases.

Case 1 - You do the survey to generate an install-config. In that case, provide the values in `config.json` as the default option for Region and Zone


Case 2 - You don't do the survey. If the install-config has Region and Zone, use those, otherwise use what's in `config.json`

23TNC:
Im pretty sure thats how it works.  During create install-config, first we read config.json, then env vars, and then the survey.  From this we populate the install-config.  During create manifests that install-config is read in and used.  During this process GetSession is called a few times, in our case, to get credentials for api stuff.  Those session objects do not override what was provided by the install-config.  Eitherway those GetSession calls are replaced by this PR to instead read that config.json which was written out during the install-config creation.  This replicates functionality of GetSession in these cases, however can cause a credential miss match if you modify the install-config and not the config.json prior to the create manifest step.  While a possible source of error, this is the same functionally as it sits currently.  

So I believe both 1 and 2 are already met.  Could you explain a bit more if I am incorrect?

sadasu:
@23TNC did the rebase go well? These merge conflicts need to be fixed.

sadasu:
@23TNC could you please rebase again?

PR #6017, opened 2022-06-16
[release-4.10] Bug 2097753: Bootstrap timeout increase

honza:


sdodson:
There was an ART build system outage which prevented the master branch change from arriving in a 4.11 nightly in a timely manner so I've overridden the bugzilla/valid-bug in order to speed this along based upon reviewing logs from CI jobs which show a change to the timeout message. The bug will still need lgtm, approved, backport-risk-assessed, and cherry-pick-approved from relevant parties and hopefully by the time this merges build systems will be back and QE can verify both 4.11 and 4.10 in a timely manner, at which point I'm fine merging the change to 4.9 and 4.8 in serial but without waiting for QE to verify those changes in nightly builds.

PR #6018, opened 2022-06-16
Bug 2100841: Print the "export KUBECONFIG=..." command on its own line for easier cut-and-paste

danwinship:
instead of saying

```
INFO To access the cluster as the system:admin user when using 'oc', run 'export KUBECONFIG=/home/danw/Desktop/cluster/auth/kubeconfig' 
```

say

```
INFO To access the cluster as the system:admin user when using 'oc', run
    export KUBECONFIG=/home/danw/Desktop/cluster/auth/kubeconfig
```

so you can just triple-click the line to select it rather than having to carefully select between the quote marks

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,8 @@ Example output:
 ```sh
 INFO Waiting 10m0s for the openshift-console route to be created...
 INFO Install complete!
-INFO To access the cluster as the system:admin user when using 'oc', run 'export KUBECONFIG=/path/to/installer/auth/kubeconfig'
+INFO To access the cluster as the system:admin user when using 'oc', run
+    export KUBECONFIG=/path/to/installer/auth/kubeconfig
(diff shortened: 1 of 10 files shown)


janosdebugs (approved):

danwinship:
do you want it for 4.11? I could create a bug I guess but it's not like this is super urgent...

janosdebugs:
@danwinship I have no strong opinion on this, I just thought it worth mentioning. I think this will be fine for 4.12.

patrickdillon:
Let's get this in 4.12 
/approve

janosdebugs:
@danwinship in this case, can you create a BZ and edit the title? :)

patrickdillon:
Or we can get it in 4.11 :)

janosdebugs:
Oops, I actually misread your comment @patrickdillon :D

PR #6019, opened 2022-06-16
Bug 2096605: vsphere: validate diskType if supplied

r4f4:
If the user inputs an invalid value for platform.vsphere.diskType in
install-config.yaml file, there is no validation checking that it's
valid and the installation continues, errorring much later. Instead,
let's check for valid diskType during install-config validation.


jinyunma:
Do we need to update the description of diskType in [install.openshift.io_installconfigs.yaml](https://github.com/openshift/installer/blob/master/data/data/install.openshift.io_installconfigs.yaml#L2914-L2923)? 
Empty value should not be supported, and default disk provision type is changed to thick, but not the default storage policy
of vsphere.

r4f4:
> Do we need to update the description of diskType in [install.openshift.io_installconfigs.yaml](https://github.com/openshift/installer/blob/master/data/data/install.openshift.io_installconfigs.yaml#L2914-L2923)? Empty value should not be supported, and default disk provision type is changed to thick, but not the default storage policy of vsphere.

@jinyunma since this is a doc issue, is it ok if I go ahead with the validation as solution to BZ#2096605 so it can be merged before FF and then create another bug for the doc fix?

jinyunma:
> since this is a doc issue, is it ok if I go ahead with the validation as solution to BZ#2096605 so it can be merged before FF and then create another bug for the doc fix?
sure, go ahead.

PR #6020, opened 2022-06-16
Bug 2097940: PowerVS: Handle optional VPCRegion

hamzy:
The VPCRegion is optional in the install-config.yaml file.  Therefore we
need to derive it from other data present when we destroy a cluster.


hamzy:
@patrickdillon Can we include this one in 4.11?

PR #6022, opened 2022-06-17
AGENT-187: AgentConfig asset

rwsu:
Adds a read only asset for AgentConfig.

Currently agent-config.yaml is written to the ISO. This will likely change in an update where the file will be parsed further and written as multiple files so that clients can consume the data more easily.


pkg/asset/agent/manifests/agent_config_type.go
```diff
@@ -0,0 +1,33 @@
...
+	// node0 hostname
+	Node0 string            `json:"node0,omitempty"`
+	Nodes []AgentConfigNode `json:"nodes"`
+}
+
+// AgentConfigNode defines per host configurations
+type AgentConfigNode struct {
+	Hostname        string                         `json:"hostname,omitempty"`
+	Role            string                         `json:"role"`
+	RootDeviceHints metal3v1alpha1.RootDeviceHints `json:"rootDeviceHints"`
```
zaneb, line 30: We should use the local copy of this type in pkg/types/baremetal/rootdevice.go

pkg/asset/agent/manifests/agent_config_type.go
```diff
@@ -0,0 +1,33 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec AgentConfigSpec `json:"spec,omitempty"`
+	// No status
+}
+
+// AgentConfigSpec contains additional configuration for the agent-based installer
+type AgentConfigSpec struct {
+	// node0 hostname
+	Node0 string            `json:"node0,omitempty"`
```
zaneb, line 22: (bikeshed alert) I have questions over whether `node0` is going to be the most meaningful name for users, even though it's been helpful for us.
I suggested `rendezvousIP` but I'm open to something easier to spell.

@celebdor are we confident that we can always accept a hostname here? I have a suspicion that there are circumstances where we will require an IP (that's certainly all we support at the moment), and perhaps we should lean in to this.

pkg/asset/agent/manifests/agent_config_type.go
```diff
@@ -0,0 +1,33 @@
...
+	Nodes []AgentConfigNode `json:"nodes"`
```
zaneb, line 23: (more bikeshed alert) I'm extremely reluctant to re-use the term 'node' here, when Node in k8s terms means a running kubelet. I think we should call this `hosts`.

zaneb (commented):
Looks great! Minor comments inline.

pkg/asset/agent/manifests/agent_config_type.go
rwsu: @zaneb Should agent_config_type.go be under pkg/types/agent?

pkg/asset/agent/manifests/agent_config_type.go
rwsu: I had an inkling that it should have been host. Thanks for the confirmation :)

pkg/asset/agent/manifests/agent_config_type.go
rwsu: I'm not tied to it being node0. It was defined as node0 in the example agent-config from a month ago.

pkg/asset/agent/manifests/agent_config_type.go
zaneb: Maybe. That sounds plausible now that you mention it :smile:

pkg/asset/agent/agentconfig/agent_config.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	return allErrs
+}
+
+func (a *Asset) validateNodesHaveAtLeastOneMacAddressDefined() field.ErrorList {
+	var allErrs field.ErrorList
+
+	if len(a.Config.Spec.Hosts) == 0 {
+		return allErrs
```
andfasano, line 94: Shouldn't be this case considered to be an error?

pkg/asset/agent/agentconfig/agent_config.go
```diff
@@ -0,0 +1,112 @@
...
+	allErrs := field.ErrorList{}
+
+	if err := a.validateNodesHaveAtLeastOneMacAddressDefined(); err != nil {
+		allErrs = append(allErrs, err...)
+	}
+
+	return allErrs
+}
+
+func (a *Asset) validateNodesHaveAtLeastOneMacAddressDefined() field.ErrorList {
```
andfasano, line 90: nit: the name of the rule looks a little bit misleading: the validation checks that _all_ the hosts have an interface with a macaddress configured

pkg/asset/agent/agentconfig/agent_config_test.go
```diff
@@ -0,0 +1,278 @@
...
+									Name:       "enp3s1",
+									MacAddress: "28:d2:44:d2:b2:1a",
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
```
andfasano, line 248: Not a blocking point for this specific PR, but I guess in future it would be useful to adopt a fluent interface approach for writing the cases under the tests, since especially for larger / more complex objects it could be difficult to spot/read the relevant parts

PR #6025, opened 2022-06-17
[release-4.9] Bug 2098158: Bootstrap timeout

honza:


sadasu:
We can achieve this backport with just 1 commit with the timeout added to the correct location. But, I notice there are 2. Any specific reason for this that I am missing?

tomassedovic:
/bugzilla refresh

Trying it again, not sure why the labels are still showing `bugzilla/invalid-bug`

PR #6026, opened 2022-06-17
[release-4.8] Bug 2098162: Bootstrap timeout

honza:


sadasu:
We can achieve this backport with just 1 commit with the timeout added to the correct location. But, I notice there are 2 commits. Any specific reason for this that I am missing?

sdodson:
Given the urgency and simplicity of this and the fact that it's been verified in 4.10 to resolve a customer issue I'm overriding bugzilla/valid-bug to accelerate the merge process.

PR #6027, opened 2022-06-17
Don't embed pre-network-manager-config script in a go string

zaneb:
Just put the script in the data directory.


bfournie:
So this looks good, just need to fix the unit test here https://github.com/openshift/installer/blob/agent-installer/pkg/asset/agent/image/ignition_test.go#L84

However with this change we will no longer be able to use the same file as assisted-service once it was moved out of their internal directory (which was what prevented us from doing that now) as tracked in https://issues.redhat.com/browse/AGENT-81.  Currently its a exact copy of that internal file.

zaneb:
> However with this change we will no longer be able to use the same file as assisted-service once it was moved out of their internal directory (which was what prevented us from doing that now) as tracked in https://issues.redhat.com/browse/AGENT-81. Currently its a exact copy of that internal file.

Thanks, I didn't know we had a plan to do this.

Once that package is available, I think I'd prefer to template the script in to an empty file rather than have what is really a static part of the host config appear out of nowhere.

zaneb:
/close
We'll defer this until we can do the full https://issues.redhat.com/browse/AGENT-81 fix.

PR #6028, opened 2022-06-17
Bug 2098299: Make new machinepool fields pointers+omitempty

2uasimojo:
install-configs generated with versions of installer libs since fdb10d991 / #5793 would fail to install on 4.10 due to the addition of optional-but-not-omitempty-pointer fields combined with strict unmarshalling that punts on unrecognized fields. Ensure new fields are a) pointer types, and b) `omitempty` so they are not emitted during marshaling if absent from the go object. cf. https://github.com/golang/go/issues/11939

closes: 2098299

### DIFF
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -532,6 +532,9 @@ func validateMarketplaceImage(client API, installConfig *types.InstallConfig) fi
 		if platform == nil {
 			continue
 		}
+		if platform.OSImage == nil {
+			continue
+		}
 		if platform.OSImage.Publisher == "" {
 			continue
 		}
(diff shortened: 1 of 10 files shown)


rna-afk:
/retitle Bug 2098299: Make AWS MachinePool EC2Metadata omitempty

2uasimojo:
/hold

This does *not* fix the problem. The field needs to be made a pointer (or we need to revert `UnmarshalStrict`).

See https://github.com/golang/go/issues/11939 (thanks for the pointer, @wking).

2uasimojo:
Updated. But let's keep the hold -- [discussing](https://coreos.slack.com/archives/C68TNFWA2/p1655842108884639) how to approach this class of problem more generally.

patrickdillon:
/close
in favor of #6039

2uasimojo:
> /close in favor of #6039

@patrickdillon the work in this PR is still valid though, isn't it?

PR #6029, opened 2022-06-18
AGENT-230: Enable multipathd in agent ISO

zaneb:
Enable the multipathd service with a default config file in the agent
ISO. This is equivalent to specifying the `rd.multipath=default` flag on
the kernel command line, but without the need to mess with the kernel
command line. Since the ISO itself will be attached as virtual or actual
media, there is no need for anything to run during the initramfs.

This allows us the agent to discover multipath Fibre Channel volumes if
they exist.

### DIFF
--- a/data/data/agent/files/etc/multipath.conf
+++ b/data/data/agent/files/etc/multipath.conf
@@ -1,0 +1,10 @@
+defaults {
+    user_friendly_names yes
+    find_multipaths yes
+    enable_foreign "^$"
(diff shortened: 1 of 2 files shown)


zaneb:
AWS test can be ignored.

andfasano:
As a next step it could be good to have a test at least on which units are being generated, even though it looks like the already existing code may require some refactoring to allow faking the fs

PR #6031, opened 2022-06-20
IBMCloud: Set regional URL for ibmcloud client

cjschaef:
Certain resources for IBM Cloud require that the region is set in
the Client. Adding support to set the regional URL for the IBM
Cloud client.


cjschaef:
Opening new PR with proper branch

PR #6033, opened 2022-06-21
Bug 1997704: [OpenStack] Document in-tree limitation for external LBs

MaysaMacedo:
By default, the cloud provider does not support creation of FIPs
using pre-defined addresses due to a policy that allows this operation
only for admins. This commit updates the documentation with possible
solutions for this issue.

### DIFF
--- a/docs/user/openstack/known-issues.md
+++ b/docs/user/openstack/known-issues.md
@@ -165,3 +165,18 @@ $ rm -f openshift/99_openshift-cluster-api_worker-machineset-*.yaml
 ```
 [bugzilla-upi]: https://bugzilla.redhat.com/show_bug.cgi?id=1927244
 [manifests-removal]:../openstack/install_upi.md#remove-machines-and-machinesets
+
+## Limitations of creating external load balancers using pre-defined FIPs
+
+On most clouds, the default policy prevents non-admin users from creating                                     
+a floating IP that has a specific address. Such policies cause the cloud provider
+to fail handling floating IP assignment to load balancers if a floating IP address
+is present in the service specification.
+
+You can pre-create a floating IP and pass the address of it in the service
+specification if using the external cloud provider. The in-tree cloud
+provider does not support this.
+Alternatively, you can [relax the Neutron policy][policy-change-steps]
+to allow non-admin users to create FIPs with a specific IP address.
+
+[policy-change-steps]: https://access.redhat.com/solutions/6069071


docs/user/openstack/known-issues.md
```diff
@@ -165,3 +165,13 @@ $ rm -f openshift/99_openshift-cluster-api_worker-machineset-*.yaml
...
+
+## Unable to create external load-balancers using pre-defined FIP (In-tree provider)
+
+Installations with in-tree cloud provider enabled does support usage of
+a Floating IP address that was pre-defined in the Service specification
+because of [a Neutron policy restriction][bugzilla-in-tree]. To create
+an external load-balancer, a Floating ip should not be specified in the
+Service to allow the cloud provider to handle its creation.
+
+[bugzilla-in-tree] https://bugzilla.redhat.com/show_bug.cgi?id=1997704
```
mandre, line 177: You're missing a `:` causing it to not render your link:
```suggestion
[bugzilla-in-tree]: https://bugzilla.redhat.com/show_bug.cgi?id=1997704
```

mandre (commented):
I find it strange that it would work for CCM and not in-tree if it's a neutron policy preventing non-admin users to create FIP with a given IP address. Can we try to understand how it works for CCM?

dulek:
> I find it strange that it would work for CCM and not in-tree if it's a neutron policy preventing non-admin users to create FIP with a given IP address. Can we try to understand how it works for CCM?

In case of CCM it allows using FIPs precreated by admin. In case of legacy cloud provider, it'll always try to create the FIP first, failing the policy check.

docs/user/openstack/known-issues.md
```diff
@@ -165,3 +165,13 @@ $ rm -f openshift/99_openshift-cluster-api_worker-machineset-*.yaml
 ```
 [bugzilla-upi]: https://bugzilla.redhat.com/show_bug.cgi?id=1927244
 [manifests-removal]:../openstack/install_upi.md#remove-machines-and-machinesets
+
+## Unable to create external load-balancers using pre-defined FIP (In-tree provider)
+
+Installations with in-tree cloud provider enabled does support usage of
+a Floating IP address that was pre-defined in the Service specification
+because of [a Neutron policy restriction][bugzilla-in-tree]. To create
+an external load-balancer, a Floating ip should not be specified in the
```
dulek, line 174: I'd capitalize it as "floating IP".

docs/user/openstack/known-issues.md
```diff
@@ -165,3 +165,13 @@ $ rm -f openshift/99_openshift-cluster-api_worker-machineset-*.yaml
 ```
 [bugzilla-upi]: https://bugzilla.redhat.com/show_bug.cgi?id=1927244
 [manifests-removal]:../openstack/install_upi.md#remove-machines-and-machinesets
+
+## Unable to create external load-balancers using pre-defined FIP (In-tree provider)
+
+Installations with in-tree cloud provider enabled does support usage of
+a Floating IP address that was pre-defined in the Service specification
+because of [a Neutron policy restriction][bugzilla-in-tree]. To create
```
dulek, line 173: Well, you can still use it, just make sure to change the policy and allow the user to create FIPs with specified address. It'd be real good to ask Neutron folks on why it's disallowed by default. I'd guess there's some security consideration here and we could explain it.

dulek (changes requested):

mandre:
> In case of CCM it allows using FIPs precreated by admin. In case of legacy cloud provider, it'll always try to create the FIP first, failing the policy check.

Oh I see. So if you provide a FIP that does not exist, it will fail the same, with a network policy check, independently of whether you're using in-tree provider or CCM, right?

dulek:
> > In case of CCM it allows using FIPs precreated by admin. In case of legacy cloud provider, it'll always try to create the FIP first, failing the policy check.
> 
> Oh I see. So if you provide a FIP that does not exist, it will fail the same, with a network policy check, independently of whether you're using in-tree provider or CCM, right?

Yeah, I think so! It's probably a known issue as well, thanks for phrasing it that way.

docs/user/openstack/known-issues.md
MaysaMacedo: You're right. I'm gonna ask them

MaysaMacedo: Just found the steps to update that policy https://access.redhat.com/solutions/6069071

docs/user/openstack/known-issues.md
```diff
@@ -165,3 +165,16 @@ $ rm -f openshift/99_openshift-cluster-api_worker-machineset-*.yaml
...
+## Limitations of creating external load-balancers using pre-defined FIP
+
+By default, installations with in-tree cloud provider enabled does support usage
+of a Floating IP address that was defined in the Service specification,
+since only admin users can create floating IP with specific address. It's
+possible to [change the policy to also allow non-admin users][policy-change-steps].
+If no policy change is wanted, the user should not provider a `loadBalancerIP` in
+the Service spec to create an external load-balancer with the in-tree provider.
+When using the out-of-tree cloud provider, a floating IP pre-created by the user
+can be used.
```
mandre, line 178: I still find the phrasing a little bit confusing as it kinda suggests the policy problem only affects in-tree cloud provider.
How about something like the following?
```suggestion
On most clouds, the default policy prevents non-admin users from creating                                     
a Floating IP with a specific address. This causes the cloud provider to fail
provisioning Load Balancers where a Floating IP address is present in the
Service specification.

It's possible to pre-create a floating IP and pass it in the Service
specification when using the external cloud provider. The in-tree cloud
provider does not support this.
Alternatively, it's also possible to [relax the policy][policy-change-steps]
and allow non-admin users to create FIPs with specific IP.
```

MaysaMacedo:
/cc @maxwelldb 
Can you have a quick look on this docs update?

docs/user/openstack/known-issues.md
```diff
@@ -165,3 +165,18 @@ $ rm -f openshift/99_openshift-cluster-api_worker-machineset-*.yaml
 ```
 [bugzilla-upi]: https://bugzilla.redhat.com/show_bug.cgi?id=1927244
 [manifests-removal]:../openstack/install_upi.md#remove-machines-and-machinesets
+
+## Limitations of creating external load-balancers using pre-defined FIP
```
maxwelldb, line 169:
```suggestion
## Limitations of creating external load balancers using pre-defined FIPs
```

docs/user/openstack/known-issues.md
```diff
@@ -165,3 +165,18 @@ $ rm -f openshift/99_openshift-cluster-api_worker-machineset-*.yaml
...
+
+On most clouds, the default policy prevents non-admin users from creating                                     
+a Floating IP with a specific address. This causes the cloud provider to fail
+provisioning Load Balancers where a Floating IP address is present in the
+Service specification.
```
maxwelldb, line 174:
```suggestion
service specification.
```

docs/user/openstack/known-issues.md
```diff
@@ -165,3 +165,18 @@ $ rm -f openshift/99_openshift-cluster-api_worker-machineset-*.yaml
...
+
+It's possible to pre-create a floating IP and pass it in the Service
+specification when using the external cloud provider. The in-tree cloud
+provider does not support this.
+Alternatively, it's also possible to [relax the Neutron policy][policy-change-steps]
+and allow non-admin users to create FIPs with specific IP.
```
maxwelldb, line 180:
```suggestion
You can pre-create a floating IP address and pass it in the service
specification [by using/if using] the external cloud provider. The in-tree cloud
provider does not support this.
Alternatively, you can [relax the Neutron policy][policy-change-steps]
to allow non-admin users to create FIPs with specific IP.
```

MaysaMacedo:
@dulek @mandre @maxwelldb thanks for the review! It seems ready for another look

docs/user/openstack/known-issues.md
```diff
@@ -165,3 +165,18 @@ $ rm -f openshift/99_openshift-cluster-api_worker-machineset-*.yaml
...
+
+## Limitations of creating external load balancers using pre-defined FIPs
+
+On most clouds, the default policy prevents non-admin users from creating                                     
+a floating IP that has a specific address. Such policies cause the cloud provider
+to fail handling floating IP assignment to load balancers if a floating IP address
+is present in the service specification.
+
+You can pre-create a floating IP and pass the address of it in the service
+specification [by using/if using] the external cloud provider. The in-tree cloud
```
mandre, line 177:
```suggestion
specification if using the external cloud provider. The in-tree cloud
```

mandre (commented):
/lgtm
/approve

PR #6037, opened 2022-06-21
AGENT-228: Systemd service to set hostname defined in agent-config

rwsu:
[AGENT-228](https://issues.redhat.com//browse/AGENT-228): Systemd service to set hostname defined in agent-config
    
Users may specify a hostname for a host by defining the hostname
and the host's macAddress in agent-config.yaml.
    
The agent Ignition then writes each macAddress to hostname mapping
to a file at /etc/assisted/hostnames/.
    
The file names in that path are MAC addresses. The desired hostname
is the file's content.
    
 A new set-hostname.service runs during boot and reads the macAddress
 to hostname mappings and if it finds one of the host's macAddress
 is in the mapping, it then updates the host's hostname using the
 mapped value.

~~Depends on: #6022~~

### DIFF
--- a/data/data/agent/systemd/units/set-hostname.service
+++ b/data/data/agent/systemd/units/set-hostname.service
@@ -6,7 +6,10 @@ After=network-online.target
 [Service]
 ExecStart=/usr/local/bin/set-hostname.sh
 ExecStartPost=/bin/sleep 5
-Restart=on-failure
+
+KillMode=none
+Type=oneshot
+RemainAfterExit=true
 
 [Install]
 WantedBy=multi-user.target
--- a/data/data/agent/systemd/units/set-hostname.service
+++ b/data/data/agent/systemd/units/set-hostname.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Agent-based installer hostname update service
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+ExecStart=/usr/local/bin/set-hostname.sh
+ExecStartPost=/bin/sleep 5
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target
--- a/data/data/agent/systemd/units/set-hostname.service
+++ b/data/data/agent/systemd/units/set-hostname.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Agent-based installer hostname update service
 Wants=network-online.target
-After=network-online.target
+After=local-fs.target
 
 [Service]
 ExecStart=/usr/local/bin/set-hostname.sh
--- a/data/data/agent/systemd/units/set-hostname.service
+++ b/data/data/agent/systemd/units/set-hostname.service
@@ -5,7 +5,6 @@ After=network-online.target
 
 [Service]
 ExecStart=/usr/local/bin/set-hostname.sh
-ExecStartPost=/bin/sleep 5
 
 KillMode=none
 Type=oneshot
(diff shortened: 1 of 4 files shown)


data/data/agent/systemd/units/set-hostname.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Agent-based installer hostname update service
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+ExecStart=/usr/local/bin/set-hostname.sh
+ExecStartPost=/bin/sleep 5
+Restart=on-failure
```
celebdor, line 9: Shouldn't it be:
```Type=oneshot```

File: data/data/agent/systemd/units/set-hostname.service [hunk 2]
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Agent-based installer hostname update service
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+ExecStart=/usr/local/bin/set-hostname.sh
+ExecStartPost=/bin/sleep 5
```
pawanpinjarkar, line 8: Why this sleep?

data/data/agent/systemd/units/set-hostname.service
pawanpinjarkar: It looks like we should be setting hostnames only once and hence this systemd service should run only once
```
RemainAfterExit=true
Type=oneshot
```

pkg/types/agent/agent_config_type.go
```diff
@@ -0,0 +1,34 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec Spec `json:"spec,omitempty"`
+	// No status
+}
+
+// Spec contains additional configuration for the agent-based installer
+type Spec struct {
+	// ip address of node0
+	RendezvousIP string `json:"rendezvousIP,omitempty"`
```
pawanpinjarkar, line 23: Can we not use `NodeZeroIP` instead of `RendezvousIP`? Easy to read and understand.

pawanpinjarkar: https://github.com/openshift/installer/pull/6022#discussion_r900664859

data/data/agent/systemd/units/set-hostname.service
rwsu: Yes, it should be oneshot.

data/data/agent/systemd/units/set-hostname.service
rwsu: It is not necessary. Will remove. Thanks.

zaneb (commented):
If we rebase then I think we'll be able to approve within the agent team, since this currently includes commits from #6022, which already merged.
/approve

rwsu:
rebased

data/data/agent/systemd/units/set-hostname.service
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Agent-based installer hostname update service
+Wants=network-online.target
+After=network-online.target
```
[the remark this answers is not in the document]
pawanpinjarkar: so `local-fs.target`?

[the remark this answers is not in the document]
pawanpinjarkar: or `network-pre.target`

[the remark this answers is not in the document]
zaneb: In principle `After=local-fs.target` should be safe.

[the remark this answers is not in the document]
rwsu: Thanks! Changed to After=local-fs.target.

PR #6038, opened 2022-06-22
Bug 2094923: Add the appropriate Tag for aws public subnets for workers

sadasu:
1. When creating new subnets, add the "kubernetes.io/role/elb" tag to the worker subnets. See https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L98-L100 for more information
2. When using existing subnets, add the tag with the value "shared". This is based on how the current tag "kubernetes.io/cluster/<cluster-id>" is added
4. Ensure the new tags are destroyed properly for both case 1 & 2. The tags are removed when the public subnets are destroyed because these public subnets could be shared among multiple clusters.

Subnets are auto-discovered using the following criteria:
// ServiceAnnotationLoadBalancerSubnets is the annotation used on the service to specify the
// Availability Zone configuration for the load balancer. The values are comma separated list of
// subnetID or subnetName from different AZs
// By default, the controller will auto-discover the subnets. If there are multiple subnets per AZ, auto-discovery
// will break the tie in the following order -
//   1. prefer the subnet with the correct role tag. kubernetes.io/role/elb for public and kubernetes.io/role/internal-elb for private access
//   2. prefer the subnet with the cluster tag kubernetes.io/cluster/<Cluster Name>
//   3. prefer the subnet that is first in lexicographic order

So, the goal of the fix is to make sure that the usable subnets in the AZ are tagged such that auto-discovery does not fall through to option 3 (and hence picking a subnet that is not owned by OpenShift)

### DIFF
--- a/data/data/aws/cluster/vpc/vpc-public.tf
+++ b/data/data/aws/cluster/vpc/vpc-public.tf
@@ -52,7 +52,8 @@ resource "aws_subnet" "public_subnet" {
 
   tags = merge(
     {
-      "Name" = "${var.cluster_id}-public-${var.availability_zones[count.index]}"
+      "Name"                   = "${var.cluster_id}-public-${var.availability_zones[count.index]}"
+      "kubernetes.io/role/elb" = ""
     },
     var.tags,
   )
--- a/pkg/asset/cluster/aws/aws.go
+++ b/pkg/asset/cluster/aws/aws.go
@@ -55,27 +55,48 @@ func tagSharedVPCResources(ctx context.Context, clusterID string, installConfig 
 		return err
 	}
 
-	ids := make([]*string, 0, len(privateSubnets)+len(publicSubnets))
-	for id := range privateSubnets {
-		ids = append(ids, aws.String(id))
-	}
-	for id := range publicSubnets {
-		ids = append(ids, aws.String(id))
-	}
-
 	session, err := installConfig.AWS.Session(ctx)
 	if err != nil {
 		return errors.Wrap(err, "could not create AWS session")
 	}
-
-	tagKey, tagValue := sharedTag(clusterID)
-
 	ec2Client := ec2.New(session, aws.NewConfig().WithRegion(installConfig.Config.Platform.AWS.Region))
+
+	ids := make([]*string, 0, len(publicSubnets))
+	for id := range publicSubnets {
+		ids = append(ids, aws.String(id))
+	}
+
+	// Add tags for Public Subnets
+	tagKey, tagValue := sharedTag(clusterID)
+	publicElbTagKey := "kubernetes.io/role/elb"
+	publicElbTagValue := ""
+
 	if _, err = ec2Client.CreateTagsWithContext(ctx, &ec2.CreateTagsInput{
 		Resources: ids,
-		Tags:      []*ec2.Tag{{Key: &tagKey, Value: &tagValue}},
+		Tags: []*ec2.Tag{
+			{Key: &tagKey, Value: &tagValue},
+			{Key: &publicElbTagKey, Value: &publicElbTagValue},
+		},
 	}); err != nil {
-		return errors.Wrap(err, "could not add tags to subnets")
+		return errors.Wrap(err, "could not add tags to public subnets")
+	}
+
+	ids = make([]*string, 0, len(privateSubnets))
+	for id := range privateSubnets {
+		ids = append(ids, aws.String(id))
+	}
+	// Add tags for Private Subnets
+	privateElbTagKey := "kubernetes.io/role/internal-elb"
+	privateElbTagValue := ""
+
+	if _, err = ec2Client.CreateTagsWithContext(ctx, &ec2.CreateTagsInput{
+		Resources: ids,
+		Tags: []*ec2.Tag{
+			{Key: &tagKey, Value: &tagValue},
+			{Key: &privateElbTagKey, Value: &privateElbTagValue},
+		},
+	}); err != nil {
+		return errors.Wrap(err, "could not add tags to private subnets")
 	}
 
 	if zone := installConfig.Config.AWS.HostedZone; zone != "" {


data/data/aws/cluster/vpc/vpc-public.tf
```diff
@@ -52,7 +52,8 @@ resource "aws_subnet" "public_subnet" {
 
   tags = merge(
     {
-      "Name" = "${var.cluster_id}-public-${var.availability_zones[count.index]}"
+      "Name"                   = "${var.cluster_id}-public-${var.availability_zones[count.index]}"
+      "kubernetes.io/role/elb" = ""
```
rna-afk, line 56: minor nit: Feels odd if this is not set to `shared` since that's the value mentioned below.

pkg/asset/cluster/aws/aws.go
```diff
@@ -69,11 +69,15 @@ func tagSharedVPCResources(ctx context.Context, clusterID string, installConfig
 	}
 
 	tagKey, tagValue := sharedTag(clusterID)
+	elbTagKey, elbTagValue := elbTag()
 
 	ec2Client := ec2.New(session, aws.NewConfig().WithRegion(installConfig.Config.Platform.AWS.Region))
 	if _, err = ec2Client.CreateTagsWithContext(ctx, &ec2.CreateTagsInput{
```
barbacbd, line 75: For the call to `ChangeTagsForResourceWithContext` we should be consistent with the contexts and make a child context before passing it to the API.

pkg/asset/cluster/aws/aws.go
```diff
@@ -95,3 +99,7 @@ func tagSharedVPCResources(ctx context.Context, clusterID string, installConfig
 func sharedTag(clusterID string) (string, string) {
 	return fmt.Sprintf("kubernetes.io/cluster/%s", clusterID), "shared"
 }
+
```
barbacbd, line 102: nit: Should we move these (sharedTag, elbTag) above in var/const section. At least above their usage in the function `tagSharedVPCResources`

data/data/aws/cluster/vpc/vpc-public.tf
sadasu: I was thinking about this myself. Similar tag for private VPCs are created without a set value : https://github.com/openshift/installer/blob/master/data/data/aws/cluster/vpc/vpc-private.tf#L39. But, the comment https://bugzilla.redhat.com/show_bug.cgi?id=2094923#c1 in the BZ, is talking about adding a value "shared" for pre-existing shared VPCs vs none for new ones. Maybe @patrickdillon can provide some context.

pkg/asset/cluster/aws/aws.go
sadasu: We could. I was just using the existing style in the code.

barbacbd: Simple nit picking. I really don't think it matters but I have seen in most of our code that the functions are C-Style where the caller sits below

data/data/aws/cluster/vpc/vpc-public.tf
patrickdillon: > a value "shared" for pre-existing shared VPCs vs none for new ones

+1 this is a new vpc so it doesn't need a shared tag

patrickdillon (commented):
Looking at the deprovision logs for the first run of `e2e-aws-shared-vpc` I don't see the elb tag getting deleted:

```
level=info msg=Deleted id=Z03657772V7RM2VECMK7C
level=info msg=Removed tag kubernetes.io/cluster/ci-op-cxdztb8p-c5f3a-w4h24: shared arn=arn:aws:ec2:us-west-1:460538899914:subnet/subnet-01145a2146611ee89
level=info msg=Removed tag kubernetes.io/cluster/ci-op-cxdztb8p-c5f3a-w4h24: shared arn=arn:aws:ec2:us-west-1:460538899914:subnet/subnet-07ba30af42f3a244c
level=info msg=Removed tag kubernetes.io/cluster/ci-op-cxdztb8p-c5f3a-w4h24: shared arn=arn:aws:ec2:us-west-1:460538899914:subnet/subnet-0fd9f10b61a30d137
level=info msg=Removed tag kubernetes.io/cluster/ci-op-cxdztb8p-c5f3a-w4h24: shared arn=arn:aws:ec2:us-west-1:460538899914:subnet/subnet-08a4758932ae1b042
```
Can we confirm the tag is getting added and deleted to/from existing subnets?

data/data/aws/cluster/vpc/vpc-public.tf
sadasu: We could add an "owned" value just to indicate that it is owned by this cluster because it was created by the cluster. But, "owned"/" " results in the same behavior with the current status of the code.

pkg/destroy/aws/shared.go
```diff
@@ -35,13 +35,14 @@ func (o *ClusterUninstaller) clusterOwnedKeys() []string {
 	var keys []string
 	for _, filter := range o.Filters {
 		for key, value := range filter {
-			if !strings.HasPrefix(key, "kubernetes.io/cluster/") {
-				continue
-			}
-			if value != "owned" {
-				o.Logger.Warnf("Ignoring non-owned cluster key %s: %s for shared-tag removal", key, value)
```
sadasu, line 42: Although this warning message says that a value other than "owned" would be ignored, it actually doesn't. So, this fix matches the implementation with what the warning message says. Hopefully that is the intention of this method also.

data/data/aws/cluster/vpc/vpc-public.tf
sadasu: Keeping it consistent with https://github.com/openshift/installer/blob/master/data/data/aws/cluster/vpc/vpc-private.tf#L39, leaving the value empty.

sadasu:
After discussing with @jstuever , we concluded that it would be better to not remove the "kubernetes.io/role/elb" tag in the deprovision path. In a multi cluster scenario, if one cluster removes the tag in its cleanup routine, the external LB would not find this public subnet which might cause failures in the other clusters.
As long as the public subnet exists, it can continue to have this tag and it will be deleted automatically when the subnet is deleted.

pkg/asset/cluster/aws/aws.go
```diff
@@ -95,3 +113,11 @@ func tagSharedVPCResources(ctx context.Context, clusterID string, installConfig
...
+func publicElbTag() (string, string) {
+	return "kubernetes.io/role/elb", ""
```
barbacbd, line 118: nit: should we consider making these constants as they have no value associated with the key?

barbacbd:
Is the installer supposed to be using these new tags on destruction ? Based on the comments it seemed like it was a possibility but there is nothing in the destroy code.

pkg/asset/cluster/aws/aws.go
```diff
@@ -55,27 +55,45 @@ func tagSharedVPCResources(ctx context.Context, clusterID string, installConfig
...
+	ec2Client := ec2.New(session, aws.NewConfig().WithRegion(installConfig.Config.Platform.AWS.Region))
+
+	ids := make([]*string, 0, len(publicSubnets))
+	for id := range publicSubnets {
+		ids = append(ids, aws.String(id))
+	}
 
+	// Add tags for Public Subnets
 	tagKey, tagValue := sharedTag(clusterID)
+	publicElbTagKey, publicElbTagValue := publicElbTag()
```
jstuever, line 71: As this is the only place publicElbTag() is used, perhaps these variables could be simplified by statically assigning the values here.

pkg/asset/cluster/aws/aws.go
```diff
@@ -55,27 +55,45 @@ func tagSharedVPCResources(ctx context.Context, clusterID string, installConfig
...
-		return errors.Wrap(err, "could not add tags to subnets")
+		return errors.Wrap(err, "could not add tags to public subnets")
+	}
+
+	ids = make([]*string, 0, len(privateSubnets))
+	for id := range privateSubnets {
+		ids = append(ids, aws.String(id))
+	}
+	// Add tags for Private Subnets
+	privateElbTagKey, privateElbTagValue := privateElbTag()
```
jstuever, line 88: As this is the only place privateElbTag() is used, perhaps these variables could be simplified by statically assigning the values here.

jstuever (approved):
I'm not sure we need to add the complexity with the publicElbTag() and privateElbTag() functions as there are no computations happening. With these only being used in one place, perhaps this can be simplified by assigning the strings directly to the variables. If possible, perhaps using the strings directly in the ec2.Tag{} declaration could further ease the complexity. Even so, I'm adding my approval because this is still a solid solution.
/approve

pkg/asset/cluster/aws/aws.go
sadasu: I just tried that. Key is expected to be a *string and a static string did not get converted. So, will have to keep this indirection for now.

pkg/asset/cluster/aws/aws.go
sadasu: Using a static string or a constant did not work because Tag and Value expect *string and the conversion is not automatic.

sadasu: Done.

pkg/destroy/aws/shared.go
sadasu: Not pursuing this portion of the fix anymore.

barbacbd (commented):
/lgtm

jstuever:
/hold
Continued discussion as to if this is a good idea.

jstuever:
/unapprove
There is valid concern about this creating a new cluster that will look different from an upgraded one. I originally approved this thinking it was a deprecation. Doing more research, it appears that the tag was removed prior to 4.0, and never shipped in any release after. Considering this does not directly fix the original bug, I'm less sure it is a good idea.

sadasu:
/close
The BZ associated with this PR has been deemed as not an issue. So, we donot need this PR at this time.

PR #6039, opened 2022-06-22
Bug 2098299: Switch to perform normal marshalling with unknown fields

rna-afk:
The hive team uses a lot of fields in the install config that might
not exist due to older versions of the installer missing a few fields
that may have been added in the newer versions, which causes the
installer to reject them all but is needed for proper functioning.

Changing the code a little to now provide a warning message to the
user in case of any unknown fields and then try to unmarshal it
normally instead of strict.

### DIFF
--- a/pkg/asset/installconfig/installconfig.go
+++ b/pkg/asset/installconfig/installconfig.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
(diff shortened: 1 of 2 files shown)


2uasimojo:
lgtm, thanks @rna-afk!

pkg/asset/installconfig/installconfig.go
```diff
@@ -131,11 +132,17 @@ func (a *InstallConfig) Load(f asset.FileFetcher) (found bool, err error) {
...
-			err = errors.Wrapf(err, "failed to parse first occurence of unknown field")
-		}
 		err = errors.Wrapf(err, "failed to unmarshal %s", installConfigFilename)
-		return false, errors.Wrap(err, asset.InstallConfigError)
+		if !strings.Contains(err.Error(), "unknown field") {
+			return false, errors.Wrap(err, asset.InstallConfigError)
+		}
+		err = errors.Wrapf(err, "failed to parse first occurence of unknown field")
+		logrus.Warnf(err.Error())
+		logrus.Warnf("trying unmarshal without unknown fields check which may cause errors")
```
patrickdillon, line 141: 1. Log messages--unlike error messages (which makes this hard to remember)--should be capitalized
2. Let's switch this to an Info
3. Maybe reword to something like
```suggestion
		logrus.Info("Attempting to unmarshal while ignoring unknown keys because strict unmarshaling failed")
```

pkg/asset/installconfig/installconfig.go
```diff
@@ -131,11 +132,17 @@ func (a *InstallConfig) Load(f asset.FileFetcher) (found bool, err error) {
...
+		if err = yaml.UnmarshalStrict(file.Data, config); err != nil {
```
patrickdillon, line 142:
```suggestion
		if err = yaml.Unmarshal(file.Data, config); err != nil {
```

patrickdillon:
/hold
for tests

rna-afk:
Turns out its due to rate limiting. Closing this then.
https://coreos.slack.com/archives/CBN38N3MW/p1656078735424419

rna-afk:
Errors in testing don't seem to be due to these changes. Removing hold.
/hold cancel

patrickdillon:
> This needs to be backported to 4.10. (Did this PR end up in 4.12? If so, it needs to be backported to 4.11 as well / first.)

@2uasimojo  Can this go into a 4.11.z stream or does it need to go into GA? If it needs to go into GA we need to set this as blocker+ and get staff engineer approval. I think doing that should be pretty clear cut with this change, if we need it.

PR #6042, opened 2022-06-23
Bug 2070318: Allow setting bootstrap kubelet ip

tsorya:
Bug 2070318: In assisted-installer flow bootstrap can have more than one ip in that case we should set speicific ip to bootstrap kubelet in order
to create kubeapi yaml with right ip.


data/data/bootstrap/files/usr/local/bin/kubelet.sh.template
```diff
@@ -11,6 +11,10 @@
     --runtime-request-timeout="${KUBELET_RUNTIME_REQUEST_TIMEOUT}" \
 {{- if .UseIPv6ForNodeIP }}
     --node-ip=:: \
+{{- end}}
+{{- if .BootstrapNodeIp }}
```
eranco74, line 15: UseIPv6ForNodeIP above is evaluted by the APIVIP
https://github.com/openshift/installer/blob/7cdf85d8df9a454c4de2297c5b5d4ae7b06fe96e/pkg/asset/ignition/bootstrap/common.go#L272

It seems that if the VIP is an IPv6 address and the bootstrapNodeIP is provided it will result in a malformed kubelet.sh with `--node-ip` flag provided twice.

tsorya: nope, look into the code. If you will provide bootstrap ip code will not set ipv6

tsorya: https://github.com/openshift/installer/pull/6042/files#diff-2a27cf9f710e504a0f8b3710fa10d24e403ff92e781cc8acbb425571427cc51bR278

data/data/bootstrap/files/usr/local/bin/kubelet.sh.template
```diff
@@ -11,6 +11,10 @@
...
+    --node-ip="{{.BootstrapNodeIp}}" \
+    --address="{{.BootstrapNodeIp}}" \
```
eranco74, line 17: Why not just set `--node-ip=0.0.0.0` and listen on all (as we currently do for IPv6)?

Also why do you set  --address="{{.BootstrapNodeIp}} ?

tsorya: If you set 0.0.0.0 it is the same as not setting anything.
The idea of setting "::" in ipv6 is to say to kubelet to choose ip from ipv6 and not to use ipv4

tsorya: kube-api always listen on 0.0.0.0 it just advertise itself on different one 
example :

[root@test-infra-cluster-86ad0a59-master-2 core]# oc get endpoints 
NAME         ENDPOINTS             AGE
kubernetes   192.168.127.12:6443   4m32s
[root@test-infra-cluster-86ad0a59-master-2 core]# netstat -tulpn | grep 6443
tcp6       0      0 :::6443                 :::*                    LISTEN      12231/kube-apiserve

eranco74: I'm not sure what the above means apart from seeing that kube-apiserver is listening on all IPv6 (as you would expect when it's configured with `::`)

tsorya: kube-api always listen on all interfaces, the issue is which one it set in entrypoint. 
Regarding the example:
I set machine cidr to 192.168.145.0/24 and set node-ip on kubelet service to 0.0.0.0, what we got is kubeapi endpoint set to 192.168.127.10 and kube-api (as always) listen on all

eranco74: So if you set node-ip (in kubelet) to the IP on that machine cidr that will be the IP for the kubeapi endpoint?

tsorya: exactly

pkg/asset/ignition/bootstrap/common.go
```diff
@@ -264,9 +265,16 @@ func (a *Common) getTemplateData(dependencies asset.Parents, bootstrapInPlace bo
 		platformData.VSphere = vsphere.GetTemplateData(installConfig.Config.Platform.VSphere)
 	}
 
+	bootstrapNodeIP := os.Getenv("OPENSHIFT_INSTALL_BOOTSTRAP_NODE_IP")
```
patrickdillon, line 268: We cannot have users set environment variables.

patrickdillon (commented):
This approach will not work. We cannot have users specifying the bootstrap ip address through environment variables. env vars are not an acceptable form of user input.

pkg/asset/ignition/bootstrap/common.go
```diff
@@ -264,9 +265,16 @@ func (a *Common) getTemplateData(dependencies asset.Parents, bootstrapInPlace bo
...
+	if bootstrapNodeIP != "" && net.ParseIP(bootstrapNodeIP) == nil {
+		logrus.Warnf("OPENSHIFT_INSTALL_BOOTSTRAP_NODE_IP must have valid ip address, given %s. Skipping it", bootstrapNodeIP)
+		bootstrapNodeIP = ""
+	}
+
 	var APIIntVIPonIPv6 bool
 	platformAPIVIP := apiVIP(&installConfig.Config.Platform)
-	if platformAPIVIP == "" {
+	// if specific ip was set for bootstrap we should use it
+	if platformAPIVIP == "" || bootstrapNodeIP != "" {
```
zaneb, line 277: It would seem more straightforward to me to use if/else in the template:

```
{{- if .BootstrapNodeIP }}
    --node-ip="{{.BootstrapNodeIP}}" \
    --address="{{.BootstrapNodeIP}}" \
{{- else}}
{{- if .UseIPv6ForNodeIP }}
    --node-ip=:: \
{{- end}}
{{- end}}
```

rather than trying to inhibit them both being true in the logic here, when the reason for it is not locally obvious.

flaper87:
> This approach will not work. We cannot have users specifying the bootstrap ip address through environment variables. env vars are not an acceptable form of user input.

Thanks for the feedback, @patrickdillon a couple of questions in the best interest of helping us move this forward:

- What would be a better approach instead of using environment variables?
- How is `UseIPv6ForNodeIP` set right now? 
- Ignoring the approach for a second, would setting the BootstrapIP during the deployment be accepted? 
- If setting the BootstrapIP during deployment is not accepted, would it be OK to have it as an env var that is set at runtime like `MACHINE_CONFIG_INFRA_IMAGE` is?

flaper87 (commented):
@patrickdillon @tsorya hey, I wasn't caught up on the discussion when I commented earlier today. It sounds like it was discussed and the agreement is to move forward with this approach and then work on a better long-term solution.

In that case, this PR lgtm

pkg/asset/ignition/bootstrap/common.go
```diff
@@ -264,6 +265,12 @@ func (a *Common) getTemplateData(dependencies asset.Parents, bootstrapInPlace bo
 		platformData.VSphere = vsphere.GetTemplateData(installConfig.Config.Platform.VSphere)
 	}
 
+	bootstrapNodeIP := os.Getenv("OPENSHIFT_INSTALL_BOOTSTRAP_NODE_IP")
+	if bootstrapNodeIP != "" && net.ParseIP(bootstrapNodeIP) == nil {
+		logrus.Warnf("OPENSHIFT_INSTALL_BOOTSTRAP_NODE_IP must have valid ip address, given %s. Skipping it", bootstrapNodeIP)
```
patrickdillon, line 270: Do you think an error is adequate here or should we return an error?

patrickdillon (commented):
/approve

Approving based on our discussion that we will follow this up with an enhancement for a long-term solution, which would allow passing in through the install config or another accepted method (install config is the only one we have identified).

I am happy to LGTM but I left one comment.

pkg/asset/ignition/bootstrap/common.go
tsorya: I thought not to block installation on it

PR #6044, opened 2022-06-23
Bug 2100496: oVirt VM creation fails on empty affinity group description

janosdebugs:
Also added an OWNERS file so this doesn't need constant approvals from the installer team.


engelmi (approved):
/lgtm

PR #6045, opened 2022-06-24
AGENT-144: Pre-generate InfraEnv ID

pawanpinjarkar:
Pre-generate infra env id and add as an env var in `assisted-service.env.template`. Assisted service then uses this ID to register infra env.
Depends on https://github.com/openshift/assisted-service/pull/4016


zaneb:
This is a good start, but (once your patch merges) we should also be able to remove all the stuff in the start-agent and start installation scripts that queries the API to get the InfraEnv ID.

pkg/asset/agent/image/ignition.go
```diff
@@ -145,14 +149,18 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
 		}
 	}
 
+	infraEnvID := strfmt.UUID(uuid.New().String())
+	log.Debug("Generated random infra-env id ", infraEnvID)
```
andfasano, line 153:
```suggestion
	logrus.Debug("Generated random infra-env id ", infraEnvID)
```

pkg/asset/agent/image/ignition_test.go
```diff
@@ -47,10 +48,11 @@ func TestIgnition_getTemplateData(t *testing.T) {
 	releaseImageMirror := "virthost.ostest.test.metalkube.org:5000/localimages/local-release-image"
 	mirrorRegistriesMount := "-v /etc/assisted/mirror/registries.conf:/etc/containers/registries.conf"
 	caBundleMount := "-v /etc/assisted/mirror/ca-bundle.crt:/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
+	infraEnvID := strfmt.UUID("random-infra-env-id")
 
 	releaseImageList, err := releaseImageList(clusterImageSet.Spec.ReleaseImage, "x86_64")
 	assert.NoError(t, err)
-	templateData := getTemplateData(pullSecret, nodeZeroIP, releaseImageList, releaseImageMirror, mirrorRegistriesMount, caBundleMount, agentClusterInstall)
+	templateData := getTemplateData(pullSecret, nodeZeroIP, releaseImageList, releaseImageMirror, mirrorRegistriesMount, caBundleMount, agentClusterInstall, infraEnvID)
 	assert.Equal(t, "http", templateData.ServiceProtocol)
```
andfasano, line 56: Maybe an assert to verify that templateData has the correct InfraEnvId?

data/data/agent/files/usr/local/bin/start-agent.sh.template
```diff
@@ -2,14 +2,5 @@
...
 
->&2 echo "Waiting for infra-env-id to be available"
-INFRA_ENV_ID=""
-until [[ $INFRA_ENV_ID != "" && $INFRA_ENV_ID != "null" ]]; do
-    sleep 5
-    >&2 echo "Querying assisted-service for infra-env-id..."
-    INFRA_ENV_ID=$(curl -s -S '{{.ServiceBaseURL}}/api/assisted-install/v2/infra-envs' | jq -r .[0].id)
-done
-echo "Fetched infra-env-id and found: $INFRA_ENV_ID"
-
```
andfasano, line 13: Maybe this script isn't required anymore, and the exec command moved directly in the ExecStart of the agent.service.template file?

pkg/asset/agent/image/ignition.go
pawanpinjarkar: Imported as `log "github.com/sirupsen/logrus"` still want to change :thinking: ?

data/data/agent/files/usr/local/bin/start-agent.sh.template
pawanpinjarkar: Done

pkg/asset/agent/image/ignition.go
zaneb: my 2c: if I see `log`, I expect it to be a `*Logger`, not the package,

pkg/asset/agent/image/ignition.go
```diff
@@ -48,6 +51,7 @@ type agentTemplateData struct {
 	ReleaseImageMirror    string
 	MirrorRegistriesMount string
 	CaBundleMount         string
+	InfraEnvID            strfmt.UUID
```
zaneb, line 54: Just make this a string, then we won't have to modify go.mod.
There's no actual advantage to using a strfmt.UUID (which is just a string), given that we are casting it to that type, and then using it as a string so there is no validation at any point.

zaneb (commented):
/approve

pkg/asset/agent/image/ignition.go
pawanpinjarkar: Done

PR #6048, opened 2022-06-27
Build optimization: Copy from installer

joepvd:


joepvd:
Resulting brew build for inspection: https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=2063423

yselkowitz:
nack, this is not multi-arch safe.
/hold

sosiouxme:
alas, if only we could copy FROM other arch builds of the same image.

PR #5802, opened 2022-04-12
Bug 2064170: Fixed Azure punctuation installconfig.controlPlane.platf…

AnnaZivkovic:
…orm.azure.osDisk explain

PR #6049, opened 2022-06-27
Bug 2082395: Make azure baseDomainResoureGroup optional for private c…

AnnaZivkovic:
…lusters

A private azure IPI cluster does not require any public base domain. If a base domain was not provided Terraform would through an error and the installation would fail. By removing omitempty the BaseDomainResourceGroupName becomes optional.

### DIFF
--- a/pkg/types/azure/platform.go
+++ b/pkg/types/azure/platform.go
@@ -34,7 +34,9 @@ type Platform struct {
 	// ClusterOSImage is the url of a storage blob in the Azure Stack environment containing an RHCOS VHD. This field is required for Azure Stack and not applicable to Azure.
 	ClusterOSImage string `json:"clusterOSImage,omitempty"`
 
-	// BaseDomainResourceGroupName specifies the resource group where the Azure DNS zone for the base domain is found.
+	// BaseDomainResourceGroupName specifies the resource group where the Azure DNS zone for the base domain is found. This field is optional when creating a private cluster, otherwise required.
+	//
+	// +optional
 	BaseDomainResourceGroupName string `json:"baseDomainResourceGroupName,omitempty"`
 
 	// DefaultMachinePlatform is the default configuration used when
--- a/data/data/azure/variables-azure.tf
+++ b/data/data/azure/variables-azure.tf
@@ -64,6 +64,7 @@ variable "azure_control_plane_ultra_ssd_enabled" {
 
 variable "azure_base_domain_resource_group_name" {
   type        = string
+  default     = ""
   description = "The resource group that contains the dns zone used as base domain for the cluster."
 }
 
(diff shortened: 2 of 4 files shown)


MayXuQQ:
pre-merge check with registry.build05.ci.openshift.org/ci-ln-drn5b3t/release:latest
- private cluster without 'baseDomainResoureGroup'  PASS
- public cluster, must has 'baseDomainResoureGroup'

- > ./openshift-install explain installconfig.platform.azure.baseDomainResourceGroupName
KIND:     InstallConfig
VERSION:  v1

RESOURCE: <string>
  BaseDomainResourceGroupName specifies the resource group where the Azure DNS zone for the base domain is found.

MayXuQQ (approved):
whether need change the the explain ?
> ./openshift-install explain installconfig.platform.azure.baseDomainResourceGroupName
> KIND: InstallConfig
> VERSION: v1
> 
> RESOURCE:
> BaseDomainResourceGroupName specifies the resource group where the Azure DNS zone for the base domain is found.

pkg/types/azure/platform.go
```diff
@@ -35,6 +35,8 @@ type Platform struct {
 	ClusterOSImage string `json:"clusterOSImage,omitempty"`
 
 	// BaseDomainResourceGroupName specifies the resource group where the Azure DNS zone for the base domain is found.
+	//
+	// +optional
 	BaseDomainResourceGroupName string `json:"baseDomainResourceGroupName,omitempty"`
```
jhixson74, line 40: You could probably write an explanation why this is parameter is optional and under what circumstances it should be set/not set.

jhixson74 (changes requested):

MayXuQQ:
ref https://github.com/openshift/installer/pull/5802#issuecomment-1097430220
You need to include [the generated code ](https://github.com/openshift/installer/blob/master/docs/dev/explain.md#generating-the-documentation)in your commit as well. That is why the verify-codegen test is failing.You need to include [the generated code ](https://github.com/openshift/installer/blob/master/docs/dev/explain.md#generating-the-documentation)in your commit as well. That is why the verify-codegen test is failing.You need to include [the generated code ](https://github.com/openshift/installer/blob/master/docs/dev/explain.md#generating-the-documentation)in your commit as well. That is why the verify-codegen test is failing.

MayXuQQ:
pre-merge checked with registry.build05.ci.openshift.org/ci-ln-x8q8qlk/release:latest

> ./openshift-install explain installconfig.platform.azure.baseDomainResourceGroupName
> KIND:     InstallConfig
> VERSION:  v1
> 
> RESOURCE: <string>
>   BaseDomainResourceGroupName specifies the resource group where the Azure DNS zone for the base domain is found. This field is optional when creating a private cluster, otherwise required.
>

patrickdillon:
/approve

The unit tests are failing because the tests here: https://github.com/openshift/installer/blob/master/pkg/explain/printer_test.go need to be updated. These tests are problematic in that way...

PR #6052, opened 2022-06-28
Bug 2101879: Use systemd path unit to monitor the node0 file

andfasano:
In some circumnstances the assisted-service-pod.service may fail to start since the node0 is not yet available.  This patch ensures that the `set-node-zero.sh` completes before systemd will activate follow up services.

### DIFF
--- a/data/data/agent/systemd/units/node-zero.service
+++ b/data/data/agent/systemd/units/node-zero.service
@@ -4,9 +4,9 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
+Type=oneshot
+RemainAfterExit=True
 ExecStart=/usr/local/bin/set-node-zero.sh
-ExecStartPost=/bin/sleep 5
-Restart=on-failure
 
 [Install]
 WantedBy=multi-user.target


data/data/agent/systemd/units/node-zero-monitor.path
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Description=Monitor the presence of the node0 file
+
+[Path]
+PathExists=/etc/assisted-service/node0
+Unit=assisted-service-pod.service
```
celebdor, line 6: Shouldn't this unit then not be enabled and only have the .path enabled?

andfasano: Thanks for the suggestion, initially I thought the `[Install]` section of `assisted-service-pod.service` was still required to let the dependencies work properly, but testing it (with a simulation of delayed node0 generation by a couple minutes) worked properly

pawanpinjarkar (approved):

zaneb (commented):
I think the problem is a race caused by the node-zero service having the wrong Type: https://bugzilla.redhat.com/show_bug.cgi?id=2101879#c2

andfasano:
> I think the problem is a race caused by the node-zero service having the wrong Type: https://bugzilla.redhat.com/show_bug.cgi?id=2101879#c2

Thanks @zaneb for the suggestion, after reviewing the possible environmental conditions I also agree that the service type is the cause of the race condition

data/data/agent/systemd/units/create-cluster-and-infraenv.service.template
```diff
@@ -3,7 +3,6 @@ Description=Service that creates initial cluster and infraenv
 Wants=network-online.target assisted-service.service
 PartOf=assisted-service-pod.service
 After=network-online.target assisted-service.service
-ConditionPathExists=/etc/assisted-service/node0
```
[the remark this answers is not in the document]
andfasano: I may be wrong, but I think that the dependencies on assisted-service prevented that, given that the path unit triggered the ancestor assisted-service-pod (but with that approach it would have been better to remove the install section on all the dependent services I guess).
Surprinsigly, the `ConditionPathExists=/etc/assisted-service/node0` is not present on assisted-service (even though its BindsTo+After on assisted-service-pod makes the dependency pretty strong)

data/data/agent/systemd/units/node-zero.service
```diff
@@ -4,9 +4,10 @@ Wants=network-online.target
 After=network-online.target
 
 [Service]
+Type=oneshot
+RemainAfterExit=True
 ExecStart=/usr/local/bin/set-node-zero.sh
 ExecStartPost=/bin/sleep 5
```
zaneb, line 10: I suspect that this sleep existed only to cover up the race condition, and could be removed now.

zaneb:
LGTM but one minor comment.

data/data/agent/systemd/units/node-zero.service
andfasano: Good point (I also removed it in the initial implementation)

mhanss:
@andfasano I have validated the changes, assisted-service-pod.service has started without any issue.
/label qe-approved

mhanss:
@andfasano Issue has not been fixed yet, I have added the latest logs on Bugzilla.

zaneb:
Hoping that https://github.com/openshift/release/pull/30062 will fix the bugzilla bot config.

PR #6053, opened 2022-06-28
WIP: Add network name to GCP cloud config

bertinatto:
This will be mostly used by the Filestore CSI Driver, who needs this information in order to provision shares.

Note: this is not ready for review as I need to figure out how we can get the network name in upgraded clusters.

CC @openshift/storage


bertinatto:
Since we will not create a StorageClass in 4.12, we don't need the shared VPC name.

PR #6054, opened 2022-06-29
AGENT-227: Use root device hints from agent-config.yaml

zaneb:
Pass the root device hints into the ISO, and create a service to apply them to the cluster once host inventory is available, but before installation starts.

### DIFF
--- a/data/data/agent/systemd/units/apply-host-config.service.template
+++ b/data/data/agent/systemd/units/apply-host-config.service.template
@@ -1,0 +1,24 @@
+[Unit]
+Description=Service that applies host-specific configuration
+Wants=network-online.target
+Requires=create-cluster-and-infraenv.service
+PartOf=assisted-service-pod.service
+After=network-online.target create-cluster-and-infraenv.service
+ConditionPathExists=/etc/assisted-service/node0
+
+[Service]
+Environment=PODMAN_SYSTEMD_UNIT=%n
+Environment=SERVICE_BASE_URL={{.ServiceBaseURL}}
+ExecStartPre=/bin/rm -f %t/%n.ctr-id
+ExecStartPre=/bin/mkdir -p %t/agent-installer /etc/assisted/hostconfig
+ExecStartPre=/usr/local/bin/wait-for-assisted-service.sh
+ExecStart=podman run --cidfile=%t/%n.ctr-id --cgroups=no-conmon --log-driver=journald --restart=on-failure:10 --pod-id-file=%t/assisted-service-pod.pod-id --replace --name=apply-host-config -v /etc/assisted/hostconfig:/etc/assisted/hostconfig -v %t/agent-installer:/var/run/agent-installer:z --env SERVICE_BASE_URL --env-file /etc/assisted/client_config quay.io/edge-infrastructure/assisted-service:latest /usr/local/bin/agent-installer-client configure
+ExecStop=/usr/bin/podman stop --ignore --cidfile=%t/%n.ctr-id
+ExecStopPost=/usr/bin/podman rm -f --ignore --cidfile=%t/%n.ctr-id
+
+KillMode=none
+Type=oneshot
+RemainAfterExit=true
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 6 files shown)


pkg/asset/agent/agentconfig/agent_config.go
```diff
@@ -94,19 +100,78 @@ func (a *Asset) validateNodesHaveAtLeastOneMacAddressDefined() field.ErrorList {
...
+	}
+
+	return allErrs
+}
+
+type hostConfigFileMap map[string][]byte
+
+// HostConfigFiles returns a map from filename to contents of the files used for
+// host-specific configuration by the agent installer client
+func (a *Asset) HostConfigFiles() (hostConfigFileMap, error) {
```
andfasano, line 147: I think that if an asset can produce additional entities (based on the ones natively managed) then this step should be placed in the `finish()` method, before the validation rules (and the entity exposed as fields). In this way the asset user will be ensured that when the dependency is retrieved all the validation rules will be applied, and in the positive case the fetched entities are good enough to be used without any additional check (for more complex cases the next refactoring step could be to create a new standalone asset for the additional entity)

data/data/agent/systemd/units/apply-host-config.service.template
```diff
@@ -0,0 +1,24 @@
...
+After=network-online.target create-cluster-and-infraenv.service
+ConditionPathExists=/etc/assisted-service/node0
+
+[Service]
+Environment=PODMAN_SYSTEMD_UNIT=%n
+Environment=SERVICE_BASE_URL={{.ServiceBaseURL}}
+ExecStartPre=/bin/rm -f %t/%n.ctr-id
+ExecStartPre=/bin/mkdir -p %t/agent-installer /etc/assisted/hostconfig
+ExecStartPre=/usr/local/bin/wait-for-assisted-service.sh
+ExecStart=podman run --cidfile=%t/%n.ctr-id --cgroups=no-conmon --log-driver=journald --restart=on-failure:10 --pod-id-file=%t/assisted-service-pod.pod-id --replace --name=apply-host-config -v /etc/assisted/hostconfig:/etc/assisted/hostconfig -v %t/agent-installer:/var/run/agent-installer:z --env SERVICE_BASE_URL --env-file /etc/assisted/client_config quay.io/edge-infrastructure/assisted-service:latest /usr/local/bin/agent-installer-client configure
```
andfasano, line 15: Given the approach we're taking to create multiple sub-commands for the `agent-installer-client`, what about orchestrating the workflow within a script executed by a single service, instead of having multiple different services with their own dependencies? Especially in those cases where the commands have a good level of cohesion (as it seems here) it could be an occasion to simplify the dependency web (and probably it could also be an help for the testing)

pkg/asset/agent/agentconfig/agent_config.go
zaneb: Everything _is_ validated when the asset is generated. Are you saying this function shouldn't return an `error`? I don't think it's possible for the yaml.Marshal to actually fail, given that it's marshalling data that was just unmarshalled from yaml.

data/data/agent/systemd/units/apply-host-config.service.template
zaneb: I'm sympathetic to that approach. However, having different services allows us to figure out where in the process we are stuck/failing. So I'd like to delay this at least until we have better status reporting.

pkg/asset/agent/agentconfig/agent_config.go
zaneb: There's no way to generate an agent-config (other than an empty one), it's the start of the line.

I'm struggling to see any difference between getting the data from a field vs. a method, except that the former requires 2 copies of everything to be kept around in memory and written to the asset store. This method could just as easily have been on the Ignition asset, or the agent.Config type itself.

> Placing this method in the `finish()` (better _after_ the validation rules) wil ensure to generate an `HostConfigFileMap` and eventually fail, if required, regardless of the fact that the asset was loaded from disk or generated.

I don't understand this part. There's no way this method can fail, so who cares when it runs?

andfasano: My comment was general for a (writable) asset, and we agree that for this specifc case the `Generate` it's not used.
An asset field it's to ensure that its content has been properly validated during the fetch phase (and if there was an error the whole process was terminated). A non-getter method with an error in its signature requires the consumer to handle it (or the method shouldn't have exposed it).
The point about the asset store storage it's fair one, and would probably justify moving the `HostConfigFiles` method into the ignition asset, especially if it's the only consumer of such artifact (ie no other assets are requiring such map).

data/data/agent/systemd/units/apply-host-config.service.template
zaneb: No conflict with updated #6502, so resolving this.

PR #6064, opened 2022-06-30
vSphere: Zonal Terraform

jcpowermac:
This PR adds an additional terraform workflow
and variables to allow for the provisioning
of the RHCOS OVA (template), bootstrap and
control plane nodes into a zonal installation.

### DIFF
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -759,41 +759,68 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		})
 
 	case vsphere.Name:
+		networkZoneMap := make(map[string]string)
+		var networkID string
 		controlPlanes, err := mastersAsset.Machines()
 		if err != nil {
 			return err
 		}
+
 		controlPlaneConfigs := make([]*machinev1beta1.VSphereMachineProviderSpec, len(controlPlanes))
 		for i, c := range controlPlanes {
 			controlPlaneConfigs[i] = c.Spec.ProviderSpec.Value.Object.(*machinev1beta1.VSphereMachineProviderSpec)
 		}
 
-		// Set this flag to use an existing folder specified in the install-config. Otherwise, create one.
-		preexistingFolder := installConfig.Config.Platform.VSphere.Folder != ""
-
-		// Must use the Managed Object ID for a port group (e.g. dvportgroup-5258)
-		// instead of the name since port group names aren't always unique in vSphere.
-		// https://bugzilla.redhat.com/show_bug.cgi?id=1918005
-		controlPlaneConfig := controlPlaneConfigs[0]
 		vim25Client, _, cleanup, err := vsphereconfig.CreateVSphereClients(context.TODO(),
-			controlPlaneConfig.Workspace.Server,
+			installConfig.Config.VSphere.VCenter,
 			installConfig.Config.VSphere.Username,
 			installConfig.Config.VSphere.Password)
 		if err != nil {
-			return errors.Wrapf(err, "unable to connect to vCenter %s. Ensure provided information is correct and client certs have been added to system trust.", controlPlaneConfig.Workspace.Server)
+			return errors.Wrapf(err, "unable to connect to vCenter %s. Ensure provided information is correct and client certs have been added to system trust.", installConfig.Config.VSphere.VCenter)
 		}
 		defer cleanup()
-
 		finder := vsphereconfig.NewFinder(vim25Client)
-		networkID, err := vsphereconfig.GetNetworkMoID(context.TODO(),
+
+		/* Each deployment zone requires a template to be imported.
+		 * The control plane machinepool might not have that zone assigned.
+		 * The zone could be unused or just defined for compute machines
+		 */
+		for _, deploymentZone := range installConfig.Config.VSphere.DeploymentZones {
+			var failureDomain vsphere.FailureDomain
+
+			for _, fd := range installConfig.Config.VSphere.FailureDomains {
+				if fd.Name == deploymentZone.FailureDomain {
+					failureDomain = fd
+				}
+			}
+
+			// Must use the Managed Object ID for a port group (e.g. dvportgroup-5258)
+			// instead of the name since port group names aren't always unique in vSphere.
+			// https://bugzilla.redhat.com/show_bug.cgi?id=1918005
+			networkZoneMap[deploymentZone.Name], err = vsphereconfig.GetNetworkMoID(context.TODO(),
+				vim25Client,
+				finder,
+				failureDomain.Topology.Datacenter,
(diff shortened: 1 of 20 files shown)


jcpowermac:
#5911 is required to be merged first.

pkg/destroy/bootstrap/bootstrap.go
```diff
@@ -41,6 +42,15 @@ func Destroy(dir string) (err error) {
 		platform = typesazure.StackTerraformName
 	}
 
+
+	// TODO: change metadata to include string to check if we are zoning
```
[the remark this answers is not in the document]
jcpowermac: Yep, thanks, let me remove.

pkg/destroy/vsphere/vsphere.go
```diff
@@ -127,8 +131,10 @@ func (o *ClusterUninstaller) deleteFolder() error {
 	// The installer should create at most one parent,
 	// the parent to the VirtualMachines.
 	// If there are more or less fail with error message.
-	if len(folderMoList) > 1 {
-		return errors.Errorf("Expected 1 Folder per tag but got %d", len(folderMoList))
+	if o.terraformPlatform != vspheretypes.ZoningTerraformName {
+		if len(folderMoList) > 1 {
+			return errors.Errorf("Expected 1 Folder per tag but got %d", len(folderMoList))
```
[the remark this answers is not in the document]
jcpowermac: right
N folder(s) per unique datacenter

rvanderp3:
govet is failing due to a dependency on https://github.com/openshift/installer/pull/5911 .  LGTM once 5911 merges

pkg/asset/cluster/tfvars.go
```diff
@@ -731,42 +731,70 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
+			controlPlaneConfigs[i] = c.Spec.ProviderSpec.Value.Object.(*machinev1beta1.VSphereMachineProviderSpec)
+			networkID, err = vsphereconfig.GetNetworkMoID(context.TODO(),
+				vim25Client,
+				finder,
+				controlPlaneConfigs[i].Workspace.Datacenter,
+				installConfig.Config.VSphere.Cluster,
+				controlPlaneConfigs[i].Network.Devices[0].NetworkName)
+			if err != nil {
+				return errors.Wrap(err, "failed to get vSphere network ID")
+			}
```
patrickdillon, line 791: I'm not following the purpose of this loop: is it just checking to make sure there is a valid `networkID` for each control plane node? It seems to change the logic, probably inconsequentially, to use the `networkID` from the last control plane node

pkg/tfvars/vsphere/vsphere.go
```diff
@@ -76,7 +104,99 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
...
+
+	for k, v := range deploymentZone {
+		tempFolder := new(folder)
+
+		tempFolder.Datacenter = failureDomainZones[v.FailureDomain].Topology.Datacenter
+		tempFolder.Name = v.PlacementConstraint.Folder
+
+		if tempFolder.Name == "" {
+			tempFolder.Name = infraID
+			deploymentZone[k].PlacementConstraint.Folder = infraID
```
patrickdillon, line 130: This does not seem like the right place to be setting this value. Should probably be in the machine pool defaults.

pkg/asset/cluster/tfvars.go
jcpowermac: I had changed this section a couple times, I think it was an oops. Will fix, thanks!

pkg/tfvars/vsphere/vsphere.go
```diff
@@ -2,8 +2,11 @@ package vsphere
 
 import (
 	"encoding/json"
+	"fmt"
 	"strings"
 
+	"github.com/openshift/installer/pkg/asset/installconfig"
```
rvanderp3, line 8: should this be at the bottom of the imports with the other installer packages?

pkg/tfvars/vsphere/vsphere.go
jcpowermac: We are using `Folder` directly in terraform which is why it needs to be defaulted:
https://github.com/openshift/installer/pull/6064/files#diff-d0b75770b7d6a46f012401b853e416a64c1dc1abbfdc5867698a93ebf098cbbbR75

Its optional for a user but not for terraform, is machinepool defaults where that should be? 
More Zones could be defined than used.

patrickdillon: > Its optional for a user but not for terraform, is machinepool defaults where that should be?

This is what I was thinking of... it seems to be consistent with what you are saying
https://github.com/openshift/installer/blob/master/pkg/types/vsphere/machinepool.go#L42-L65

jcpowermac: > This is what I was thinking of... it seems to be consistent with what you are saying

is it?

I need to set the a platform-spec defined struct parameter to the InfraID, which the machinepool doesn't include.
And is infraID available when that runs?

patrickdillon:
From a high-level perspective this code looks solid to me, and I trust @rvanderp3  and the rest of splat to cover any gaps in my vsphere knowledge. 

I am happy to give my `/approve` here, but I will reiterate: I think it would be really worthwhile to consider whether the multizone and normal vsphere terraform configs could be combined, to where the normal vsphere install is just a multizone install with a single availability zone. Once this goes in, I think it is unlikely that it would be refactored out, and would be a good idea to avoid the long-term maintenance of two largely overlapping configurations...

jcpowermac:
While researching if we can combine the existing terraform and zonal found a bug that the newest commit should fix...testing

/test e2e-vsphere-zones

jcpowermac:
upi image build failure :(

/test e2e-vsphere-zones

jcpowermac:
ugh upi
/test e2e-vsphere-zones

jcpowermac:
> From a high-level perspective this code looks solid to me, and I trust @rvanderp3 and the rest of splat to cover any gaps in my vsphere knowledge.
> 
> I am happy to give my `/approve` here, but I will reiterate: I think it would be really worthwhile to consider whether the multizone and normal vsphere terraform configs could be combined, to where the normal vsphere install is just a multizone install with a single availability zone. Once this goes in, I think it is unlikely that it would be refactored out, and would be a good idea to avoid the long-term maintenance of two largely overlapping configurations...

https://github.com/openshift/installer/pull/6279#issuecomment-1233277116

jcpowermac:
Issues with our zoning CI environment

tested in VMC bootstrap completed.

jcpowermac:
@rvanderp3 or @patrickdillon 
had to rebase for a conflict if I could get a lgtm please

jcpowermac:
ugh ci :(

/retest-required

PR #6066, opened 2022-06-30
AGENT-144: Pass pre-generated InfraEnv ID to apply-host-config

zaneb:
Since we now know the InfraEnv ID in advance, there's no need to record
it in create-cluster-and-infra-env and then pass it to
apply-host-config from the resulting environment file.


pawanpinjarkar (approved):
/lgtm

PR #6067, opened 2022-07-01
Bug 2103144: openstack: validate apiVIP and ingressVIP to be semantically different

pierreprinetti:
Before this patch, apiVIP and ingressVIP are checked to be different
based on a comparison of their string representation.

However, the same IP can have different string representation; this is
especially true with IPv6.

With this patch, IP comparison is delegated to the Go standard library.


pierreprinetti:
The test failure is unrelated; validation completed successfully.

mandre (commented):
/lgtm
Have you checked if the same fix should be ported more generally to the installer, outside of the openstack validation?

PR #6072, opened 2022-07-01
AGENT-234: Take node0 selection from agent-config if it exists and se…

pawanpinjarkar:
If node0 IP is specified in agentConfig, it takes precedence over the selection from NMStateConfigs, otherwise, we keep the same heuristic as we have now to choose.

### DIFF
--- a/pkg/asset/agent/image/ignition.go
+++ b/pkg/asset/agent/image/ignition.go
@@ -13,6 +13,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	hiveext "github.com/openshift/assisted-service/api/hiveextension/v1beta1"
+	"github.com/openshift/assisted-service/api/v1beta1"
 	"github.com/openshift/assisted-service/models"
 	"github.com/openshift/installer/pkg/asset"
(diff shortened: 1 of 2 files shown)


zaneb (commented):
/approve

pkg/asset/agent/image/ignition.go
```diff
@@ -112,9 +112,18 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
...
+	if agentConfigAsset.Config != nil && agentConfigAsset.Config.Spec.RendezvousIP != "" {
+		nodeZeroIP = agentConfigAsset.Config.Spec.RendezvousIP
+		logrus.Debug("RendezvousIP from the AgentConfig ", nodeZeroIP)
+
+	} else {
+		nodeZeroIP, err := manifests.GetNodeZeroIP(agentManifests.NMStateConfigs)
+		logrus.Debug("RendezvousIP from the NMStateConfig ", nodeZeroIP)
+		if err != nil {
+			return err
+		}
```
andfasano, line 126: It's be nice to have a test for this case

PR #6081, opened 2022-07-05
openstack UPI: clarify naming of RHCOS image

pierreprinetti:
This PR adds:
* explanation on how the name of the RHCOS image on Glance must match
  the content of inventory.yaml;
* instructions for removing the image after use.
* a scripted unique name for the RHCOS image in Glance, so that the UPI
  tests can be run in parallel.

Blocks https://github.com/openshift/release/pull/29672


mandre (commented):
/lgtm
/approve

PR #6087, opened 2022-07-06
AGENT-136: Add optional install-config asset for agent installation

zaneb:
With agent-based installation, we allow users to start directly from ZTP
CRs in cluster-manifests without first generating them from an
install-config. However, the latter will be an option, so resources will
need to depend on the InstallConfig without causing the survey to run if
it is not present.


pawanpinjarkar (commented):
/lgtm

PR #6091, opened 2022-07-07
openstack UPI: Remove Swift example, use Glance

pierreprinetti:
The UPI workflow needs the bootstrap ignition file to be hosted
somewhere. Prior to this patch, the first proposed example used Swift.
However, the example was using commands that are not available in the
openstack client shipped with the recommended operating systems and
repositories.

This patch removes the Swift example and focuses the workflow on using
Glance instead, along the lines of what the IPI installer does. The UPI
docs still mention that any hosting system can be used, including Swift.

### DIFF
--- a/docs/user/openstack/install_upi.md
+++ b/docs/user/openstack/install_upi.md
@@ -39,8 +39,7 @@ of this method of installation.
     - [Bootstrap Ignition](#bootstrap-ignition)
       - [Edit the Bootstrap Ignition](#edit-the-bootstrap-ignition)
       - [Upload the Boostrap Ignition](#upload-the-boostrap-ignition)
-        - [Example 1: Swift](#example-1-swift)
-        - [Example 2: Glance image service](#example-2-glance-image-service)
+        - [Example: Glance image service](#example-glance-image-service)
     - [Create the Bootstrap Ignition Shim](#create-the-bootstrap-ignition-shim)
       - [Ignition file served by server using self-signed certificate](#ignition-file-served-by-server-using-self-signed-certificate)
     - [Master Ignition](#master-ignition)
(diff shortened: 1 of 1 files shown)


docs/user/openstack/install_upi.md
```diff
@@ -617,34 +617,16 @@ In this guide, we will assume the file is at the following URL:
 
 https://static.example.com/bootstrap.ign
 
-##### Example 1: Swift
+##### Example: Glance image service
```
mandre, line 620: Please also update the Table of Contents. I think we could remove the example links from it.

docs/user/openstack/install_upi.md
```diff
@@ -692,6 +671,12 @@ Create a file called `$INFRA_ID-bootstrap-ignition.json` (fill in your `infraID`
     "config": {
       "merge": [
         {
+	  "httpHeaders: [
```
mandre, line 674:
```suggestion
	  "httpHeaders": [
```

docs/user/openstack/install_upi.md
```diff
@@ -941,7 +926,14 @@ The teardown playbook deletes the bootstrap port and server.
...
 
-If you haven't done so already, you should also disable the bootstrap Ignition URL.
+If you haven't done so already, you should also disable the bootstrap Ignition URL. If you are following the Glance example:
+
+<!--- e2e-openstack-upi: INCLUDE START --->
+```sh
+$ openstack image delete "bootstrap-ign-${INFRA_ID}"
+```
+<!--- e2e-openstack-upi: INCLUDE END --->
+
```
mandre, line 936: We should probably also document revoking the glance token.

pierreprinetti:
Updated the TOC, fixed JSON syntax and added token revocation. Thanks @mandre !

mandre (commented):
Looking good now, thanks!
/lgtm

PR #6098, opened 2022-07-07
AGENT-280: Generate mirror config from install-config

zaneb:
Generate the mirror configuration (`mirror/registries.conf` and `mirror/ca-bundle.crt`) from the install-config if it is supplied.

(Note: depends on #6087)


data/data/agent/systemd/units/assisted-service.service
```diff
@@ -11,7 +11,7 @@ Environment=PODMAN_SYSTEMD_UNIT=%n
 Restart=on-failure
 TimeoutStopSec=300
 ExecStartPre=/bin/rm -f %t/%n.ctr-id
-ExecStart=/usr/bin/podman run --cidfile=%t/%n.ctr-id --cgroups=no-conmon --log-driver=journald --rm --pod-id-file=%t/assisted-service-pod.pod-id --sdnotify=conmon --replace -d --name=service -v /opt/agent/tls:/opt/agent/tls:z {{.MirrorRegistriesMount}} {{.CaBundleMount}} --env-file=/usr/local/share/assisted-service/assisted-service.env --env-file=/usr/local/share/assisted-service/images.env --env-file=/etc/assisted-service/node0 quay.io/edge-infrastructure/assisted-service:latest
+ExecStart=/usr/bin/podman run --cidfile=%t/%n.ctr-id --cgroups=no-conmon --log-driver=journald --rm --pod-id-file=%t/assisted-service-pod.pod-id --sdnotify=conmon --replace -d --name=service -v /opt/agent/tls:/opt/agent/tls:z -v /etc/containers:etc/containers -v /etc/pki/ca-trust:/etc/pki/ca-trust --env-file=/usr/local/share/assisted-service/assisted-service.env --env-file=/usr/local/share/assisted-service/images.env --env-file=/etc/assisted-service/node0 quay.io/edge-infrastructure/assisted-service:latest
```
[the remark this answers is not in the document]
zaneb: :man_facepalming: thanks Toni.

pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -45,32 +126,44 @@ func (*RegistriesConf) Dependencies() []asset.Asset {
 
 // Generate generates the registries.conf file from install-config.
 func (i *RegistriesConf) Generate(dependencies asset.Parents) error {
+	installConfig := &agent.OptionalInstallConfig{}
+	dependencies.Get(installConfig)
+	if !installConfig.Supplied || len(installConfig.Config.ImageContentSources) == 0 {
```
rwsu, line 131: Actually I don't think we can add the default registries.conf file, at least not in its current form. Assisted-service isn't able to process it because it is expecting the [registry] table to be defined if the file is present: https://github.com/openshift/assisted-service/blob/master/pkg/mirrorregistries/generator.go#L69  

Consequently the cluster fails to install and I am seeing this in the event log:
Screen Shot 2022-07-11 at 11 13 20 PM

bfournie: I wonder if just uncommenting the 
`[[registry]]`
and 
`[[registry.mirror]]`

in the default file would allow to be considered as as valid TOML. Otherwise we'd need to avoid creating an asset from this default file if there is no valid registries info in the installConfig

zaneb: It looks like they also require location fields to be set. We should probably fix this in assisted-service in the long term. Just not creating the file in ignition won't help, because the default in CoreOS also has no mirrors configured (in fact it's essentially this same file). So we will have to make the mount conditional for now. Trying that out at the moment.

zaneb: Sorted. Thanks Richard!

data/data/agent/systemd/units/assisted-service.service.template
```diff
@@ -11,7 +11,7 @@ Environment=PODMAN_SYSTEMD_UNIT=%n
 Restart=on-failure
 TimeoutStopSec=300
 ExecStartPre=/bin/rm -f %t/%n.ctr-id
-ExecStart=/usr/bin/podman run --cidfile=%t/%n.ctr-id --cgroups=no-conmon --log-driver=journald --rm --pod-id-file=%t/assisted-service-pod.pod-id --sdnotify=conmon --replace -d --name=service -v /opt/agent/tls:/opt/agent/tls:z {{.MirrorRegistriesMount}} {{.CaBundleMount}} --env-file=/usr/local/share/assisted-service/assisted-service.env --env-file=/usr/local/share/assisted-service/images.env --env-file=/etc/assisted-service/node0 quay.io/edge-infrastructure/assisted-service:latest
+ExecStart=/usr/bin/podman run --cidfile=%t/%n.ctr-id --cgroups=no-conmon --log-driver=journald --rm --pod-id-file=%t/assisted-service-pod.pod-id --sdnotify=conmon --replace -d --name=service -v /opt/agent/tls:/opt/agent/tls:z {{ if .ReleaseImageMirror }}-v /etc/containers:/etc/containers{{ end }} -v /etc/pki/ca-trust:/etc/pki/ca-trust --env-file=/usr/local/share/assisted-service/assisted-service.env --env-file=/usr/local/share/assisted-service/images.env --env-file=/etc/assisted-service/node0 quay.io/edge-infrastructure/assisted-service:latest
```
rwsu, line 14: The conditional on ReleaseImageMirror isn't working as expected. When I tested the disconnected case using dev-scripts ("export MIRROR_IMAGES=true" in config_xxx.sh), ReleaseImageMirror has empty string as value. /etc/containers doesn't get mounted and the install-config.yaml for the cluster to be installed is missing additionalTrustBundle and imageContentSources.

pkg/asset/agent/image/ignition.go
```diff
@@ -127,25 +125,14 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
...
+	registriesConfig := &mirror.RegistriesConf{}
+	registryCABundle := &mirror.CaBundle{}
+	dependencies.Get(registriesConfig, registryCABundle)
 
 	// Get the mirror for release image
 	releaseImageMirror := ""
 	source := strings.Split(agentManifests.ClusterImageSet.Spec.ReleaseImage, ":")
-	for _, config := range agentMirror.MirrorConfig {
+	for _, config := range registriesConfig.MirrorConfig {
 		if config.Location == source[0] {
```
rwsu, line 136: releaseImageMirror is left as empty string because a match is never found. When dev-scripts is configured with MIRROR_IMAGES=true I see the following values for config.Location and source[0] in the for loop:
```
level=info msg=RWSU config.Location: registry.ci.openshift.org/ocp/release
level=info msg=RWSU source[0]: virthost.ostest.test.metalkube.org
level=info msg=RWSU config.Location: quay.io/openshift-release-dev/ocp-v4.0-art-dev
level=info msg=RWSU source[0]: virthost.ostest.test.metalkube.org
```
agentManifests.ClusterImageSet.Spec.ReleaseImage=virthost.ostest.test.metalkube.org:5000/localimages/local-release-image:4.11.0-0.nightly-2022-07-11-080250

rwsu (changes requested):

data/data/agent/systemd/units/assisted-service.service.template
bfournie: Yes, good find. The way it worked previously is MirrorRegistriesMount is set up based on the existence of the input registries.conf, not on the ReleaseImageMirror - https://github.com/openshift/installer/blob/89ebbd566cb7d7479e828f87f0248915f0eea946/pkg/asset/agent/image/ignition.go#L137

As you noted below, in dev-scripts the ReleaseImage is to the override location, i.e. the mirrored location.

pkg/asset/agent/image/ignition.go
zaneb: @bfournie could this be a limitation/problem of the dev-scripts config? I'd have expected a mirror to match here, and that some stuff in assisted-service might not work if none does.

bfournie: assisted-service should handle it fine, and does in my testing - it checks the mirror first and uses the releaseImage if not set. But we could set the releaseImageMirror [here](https://github.com/openshift/installer/blob/89ebbd566cb7d7479e828f87f0248915f0eea946/pkg/asset/agent/image/ignition.go#L149)
if its also in the mirror section of the config, then the conditional would work OK

bfournie: Also, using the release image set to the mirror is common setting for users, see for example Eran's comment #2 [here](https://issues.redhat.com/browse/MGMT-10209)

zaneb: OK, I realised I have to bite the bullet and pass another boolean in the config. I tested that both with and without MIRROR_IMAGES enabled in dev-scripts and it now works. Thanks everybody!

bfournie:
I tested this with https://github.com/openshift-metal3/dev-scripts/pull/1424 with MIRROR_IMAGES=true and verified it works as expected, up to the point where we hit the assisted-service ICSP issue (https://issues.redhat.com/browse/MGMT-10209)

rwsu (approved):
Verified that the issues have been fixed.

/approve

PR #6099, opened 2022-07-07
AGENT-246: Do not fail on nmstateconfig validation if agent-config is provided

pawanpinjarkar:
RendezvousIP is set either from agentconfig.spec.RendezvousIP or from nmstateconfig. If agent config does not have RendezvousIP defined then at least one nmstateconfig must be provided. Otherwise, the create image command will fail with an error `Missing Spec.RendezvousIP in agent-config or atleast one nmstateconfig manifest file"`

### DIFF
--- a/pkg/asset/agent/agentconfig/agent_config.go
+++ b/pkg/asset/agent/agentconfig/agent_config.go
@@ -7,7 +7,6 @@ import (
 	"strings"
 
 	"github.com/openshift/installer/pkg/asset"
-	"github.com/openshift/installer/pkg/asset/agent/common"
 	"github.com/openshift/installer/pkg/types/agent"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -62,12 +61,6 @@ func (a *AgentConfig) Load(f asset.FileFetcher) (bool, error) {
 	config := &agent.Config{}
 	if err := yaml.UnmarshalStrict(file.Data, config); err != nil {
 		return false, errors.Wrapf(err, "failed to unmarshal %s", agentConfigFilename)
-	}
-
-	if config.Spec.RendezvousIP != "" {
-		common.AgentConfigAvailable = true
-	} else {
-		return false, errors.New("missing Spec.RendezvousIP")
 	}
 
 	a.File, a.Config = file, config
--- a/pkg/asset/agent/agentconfig/agent_config.go
+++ b/pkg/asset/agent/agentconfig/agent_config.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/openshift/installer/pkg/asset"
+	"github.com/openshift/installer/pkg/asset/agent/common"
 	"github.com/openshift/installer/pkg/types/agent"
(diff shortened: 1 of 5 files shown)


pkg/asset/agent/common/common.go
```diff
@@ -0,0 +1,4 @@
+package common
+
+// AgentConfigAvaialble is used to check if agent-config.yaml was provided
+var AgentConfigAvaialble bool
```
zaneb, line 4: typo: Available

pkg/asset/agent/agentconfig/agent_config.go
```diff
@@ -63,6 +64,12 @@ func (a *Asset) Load(f asset.FileFetcher) (bool, error) {
 		return false, errors.Wrapf(err, "failed to unmarshal %s", agentConfigFilename)
 	}
 
+	if config.Spec.RendezvousIP != "" {
+		common.AgentConfigAvaialble = true
+	} else {
+		return false, errors.New("missing Spec.RendezvousIP")
```
zaneb, line 70: We can't return an error here, the RendezvousIP is not _required_ (but you must have _either_ that _or_ at least one NMState config with a static IP).

pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -124,15 +125,15 @@ func (n *NMStateConfig) Load(f asset.FileFetcher) (bool, error) {
 }
 
 func (n *NMStateConfig) finish() error {
+	if !common.AgentConfigAvaialble {
```
andfasano, line 128: I think that's another example of cross-assets validation (ie a validation rule that needs two or more inputs coming from different assets). In such case the usage of global variables IMHO is highly discouraged, whereas it would be recommended to push up the validation rule into _another_ asset depending on all the assets providing the inputs for the validation rule (a similar approach has been taken for AgentManifest for example, see [here](https://github.com/openshift/installer/blob/9b36612243bd08dbe5c32cd851741ffb5e9b9b61/pkg/asset/agent/manifests/agent.go#L125))

pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -124,15 +125,15 @@ func (n *NMStateConfig) Load(f asset.FileFetcher) (bool, error) {
...
+			return errors.New("missing configuration or manifest file")
+		}
 
-	if n.Config == nil {
-		return errors.New("missing configuration or manifest file")
-	}
-
-	if err := n.validateNMStateConfig().ToAggregate(); err != nil {
-		return errors.Wrapf(err, "invalid NMStateConfig configuration")
+		if err := n.validateNMStateConfig().ToAggregate(); err != nil {
```
zaneb, line 133: I think that the `}` landed in the wrong place here; we shouldn't skip validation of NMStateConfig data just because there is a RendezvousIP defined.

zaneb:
/approve
Looks good except for failing unit test.

rwsu (commented):
Unit test needs to be fixed.
/lgtm

rwsu (commented):
/lgtm

PR #6101, opened 2022-07-08
NO-ISSUE: Wait for InfraEnv to be created before running agent

rwsu:
PR#6045 removed the wait. The wait needs to be put back in
because host registration will fail if the InfraEnv has not been
registered with assisted-service.


pawanpinjarkar (approved):
/approve

PR #6103, opened 2022-07-08
Run bootstrap gather when bootstrap is preserved

r4f4:
/hold

This is just for the purpose of debugging https://bugzilla.redhat.com/show_bug.cgi?id=2102662


patrickdillon:
> There could be valuing in merging this if it proves fruitful.

On second thought, I'm not sure. We don't intend for customers to use preserve bootstrap and manually running gather bootstrap is not too much of a burden.

PR #6106, opened 2022-07-09
AGENT-268: Remove assisted-service UI container

bfournie:
We don't currently build openshift images for the assisted-service UI container.


rwsu (approved):
/approve

PR #6095, opened 2022-07-07
Bug 2095319: bump RHCOS 4.10 bootimage metadata

mike-nguyen:
These changes will update the RHCOS 4.10 boot image metadata in the
installer which includes the fixes for the following BZs:

BZ 2059305 [tracker] HPE Synergy 480 Gen10 Plus servers fail to boot control plane nodes with kernel panic

Changes generated with:
```
plume cosa2stream --target data/data/coreos/rhcos.json --distro rhcos --url https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases --no-signatures x86_64=410.84.202207061638-0 aarch64=410.84.202207051516-0 s390x=410.84.202207060728-0 ppc64le=410.84.202207061403-0
```

PR #6109, opened 2022-07-11
Bug 2102228: rhcos: Move to rhcos.mirror.openshift.com

vfreex:

### DIFF
--- a/data/data/coreos/rhcos.json
+++ b/data/data/coreos/rhcos.json
@@ -12,7 +12,7 @@
           "formats": {
             "vmdk.gz": {
               "disk": {
-                "location": "https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos-4.11-aarch64/411.85.202205040359-0/aarch64/rhcos-411.85.202205040359-0-aws.aarch64.vmdk.gz",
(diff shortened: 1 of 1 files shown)


cgwalters:
This seems sane to me, but can't we make `rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com` a [CNAME](https://en.wikipedia.org/wiki/CNAME_record) ?  That would avoid the need to change client systems for the near future.

mike-nguyen:
Is this for all RHCOS images (4.11 -> 4.7)? If so, we will use the updated mirror for any boot image changes we make going forward.  For example, we have this pending 4.10 boot image bump PR https://github.com/openshift/installer/pull/6095

miabbott:
> @vfreex Is this for all RHCOS images (4.11 -> 4.7)? If so, we will use the updated mirror for any boot image changes we make going forward. For example, we have this pending 4.10 boot image bump PR #6095

Yeah this will be applicable to all versions from 4.12 -> 4.8; see the first comment on the BZ

vfreex:
@cgwalters Yes, we will make rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com a CNAME once this CDN set up is tested (in case there are missing firewall configs). But that is just to ensure old installers to continue to work after the migration.

We would like to change this domain name baked in to the installer for future releases (including 4.6-4.11) because the domain name rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com will probably go away in 1-2 years.

PR #6114, opened 2022-07-12
AGENT-282: Create cluster deployment asset manifest from install config

pawanpinjarkar:
Creates cluster deployment yaml asset from the user-provided install-config.yaml.


pkg/asset/agent/manifests/clusterdeployment.go
```diff
@@ -39,21 +42,77 @@ func (*ClusterDeployment) Dependencies() []asset.Asset {
...
+				Name:      installConfig.Config.ObjectMeta.Name,
+				Namespace: installConfig.Config.ObjectMeta.Namespace,
+			},
+			Spec: hivev1.ClusterDeploymentSpec{
+				ClusterName: installConfig.Config.ObjectMeta.Name,
+				BaseDomain:  installConfig.Config.BaseDomain,
+				Platform: hivev1.Platform{
+					AgentBareMetal: &hivev1Agent.BareMetalPlatform{
+						AgentSelector: metav1.LabelSelector{
+							// What labels should be set here? Will the values come from install config?
```
pawanpinjarkar, line 65: @zaneb :point_up:

pkg/asset/agent/manifests/clusterdeployment.go
```diff
@@ -39,21 +42,77 @@ func (*ClusterDeployment) Dependencies() []asset.Asset {
...
+							MatchLabels: map[string]string{},
+						},
+					},
+				},
+				// What should be set as PullSecretRef name here? Will the values come from install config?
+				PullSecretRef: &v1.LocalObjectReference{Name: "pull-secret"},
```
pawanpinjarkar, line 71: @zaneb :point_up:

pkg/asset/agent/manifests/clusterdeployment.go
```diff
@@ -39,21 +42,77 @@ func (*ClusterDeployment) Dependencies() []asset.Asset {
...
+				ControlPlaneConfig: hivev1.ControlPlaneConfigSpec{
+					ServingCertificates: hivev1.ControlPlaneServingCertificateSpec{
+						// Default: "",
+						// Additional: []hivev1.ControlPlaneAdditionalCertificate{{
+						// 	Name:   clusterDeploymentFilename,
+						// 	Domain: "",
+						// }},
+					},
```
pawanpinjarkar, line 79: @zaneb :point_up:  From where will we read `ControlPlaneAdditionalCertificate` data?

pkg/asset/agent/manifests/clusterdeployment_test.go
```diff
@@ -25,8 +27,34 @@ func TestClusterDeployment_Generate(t *testing.T) {
 	}{
 		{
 			name:          "missing-config",
+			dependencies:  unittest.GetEmptyTestAssets(),
 			expectedError: "missing configuration or manifest file",
 		},
+		{
+			name:         "default",
+			dependencies: unittest.GetTestAssetsWithValidInstallConfig(),
```
pawanpinjarkar, line 35: The idea here is to read the valid/invalid inputs for install config from a single place as that will be needed by other asset unit tests when we enable them.

pkg/asset/agent/manifests/clusterdeployment.go
pawanpinjarkar: Fixed.

pkg/asset/agent/manifests/clusterdeployment.go
```diff
@@ -39,21 +41,61 @@ func (*ClusterDeployment) Dependencies() []asset.Asset {
...
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      installConfig.Config.ObjectMeta.Name,
+				Namespace: installConfig.Config.ObjectMeta.Namespace,
+			},
+			Spec: hivev1.ClusterDeploymentSpec{
+				ClusterName: installConfig.Config.ObjectMeta.Name,
+				BaseDomain:  installConfig.Config.BaseDomain,
+				PullSecretRef: &corev1.LocalObjectReference{
+					Name: "pull-secret",
```
zaneb, line 62: This needs to have the same name as the Secret in the AgentPullSecret manifest.

I think the safest way to do this might be to depend on AgentPullSecret and call AgentPullSecret.ResourceName(). That way if the user can created a PullSecret already with a different name, we will just use it. I'm not sure if this kind of approach can be used everywhere without circular dependencies though.

pkg/asset/agent/manifests/clusterdeployment.go
```diff
@@ -39,21 +41,61 @@ func (*ClusterDeployment) Dependencies() []asset.Asset {
...
+				},
+				ClusterInstallRef: &hivev1.ClusterInstallLocalReference{
+					Group:   "extensions.hive.openshift.io",
+					Version: "v1beta1",
+					Kind:    "AgentClusterInstall",
+					Name:    "agent-cluster-install",
```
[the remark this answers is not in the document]
pawanpinjarkar: Done

pkg/asset/agent/manifests/clusterdeployment.go
```diff
@@ -39,21 +41,62 @@ func (*ClusterDeployment) Dependencies() []asset.Asset {
 }
 
 // Generate generates the ClusterDeployment manifest.
-func (i *ClusterDeployment) Generate(dependencies asset.Parents) error {
-	// TODO: generate from install-config
-	return i.finish()
+func (cd *ClusterDeployment) Generate(dependencies asset.Parents) error {
+	installConfig := &agent.OptionalInstallConfig{}
+	agentPullSecret := &AgentPullSecret{}
```
zaneb, line 46: Have to add this to the return from `Dependencies()` also.

zaneb (commented):
/lgtm

rwsu (approved):
/approve

PR #6118, opened 2022-07-13
Bug 2106820: Convert all MAC addresses to lowercase

zaneb:
If MAC addresses are provided with uppercase hex digits in the
agent-config, convert them to lowercase before storing in the image. The
client program that searches for hosts just does a direct
(case-sensitive) comparison with the values obtained from inventory.


pawanpinjarkar (approved):
/lgtm

PR #6125, opened 2022-07-15
Bug 2095323: Create security group rules for each MachineNetwork CIDR

mandre:
Previously, only the first CIDR was taken into account when creating
the security group rules. This commit now create security group rules
for each one of them, allowing traffic from every possible
MachineNetwork.

### DIFF
--- a/data/data/openstack/masters/sg-master.tf
+++ b/data/data/openstack/masters/sg-master.tf
@@ -5,12 +5,13 @@ resource "openstack_networking_secgroup_v2" "master" {
 }
 
 resource "openstack_networking_secgroup_rule_v2" "master_mcs" {
+  count             = length(var.machine_v4_cidrs)
   direction         = "ingress"
   ethertype         = "IPv4"
   protocol          = "tcp"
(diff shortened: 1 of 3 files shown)


data/data/openstack/masters/sg-master.tf
```diff
@@ -264,13 +281,14 @@ resource "openstack_networking_secgroup_rule_v2" "master_ingress_https" {
 }
 
 resource "openstack_networking_secgroup_rule_v2" "master_ingress_router" {
+  count             = length(var.machine_v4_cidrs)
   count             = var.masters_schedulable ? 1 : 0
```
MaysaMacedo, line 285: Is this count needed?
It's conflicting with the new one defined as they have the same name.

MaysaMacedo (commented):
Besides the following remark it looks good!

data/data/openstack/masters/sg-master.tf
mandre: Good catch. I'll fix it.

PR #6126, opened 2022-07-15
Bug 2107151: use regional STS endpoints

r4f4:
By default the AWS SDK for Go V1 SDK uses the global STS access point.
Let's tell it to use the regional one instead.

https://github.com/aws/aws-sdk-go/issues/4112#issuecomment-941292922

### DIFF
--- a/pkg/asset/installconfig/aws/session.go
+++ b/pkg/asset/installconfig/aws/session.go
@@ -59,7 +59,10 @@ func GetSession() (*session.Session, error) { return GetSessionWithOptions() }
 // and, if no creds are found, asks for them and stores them on disk in a config file
 func GetSessionWithOptions(optFuncs ...SessionOptions) (*session.Session, error) {
 	options := session.Options{
-		Config:            aws.Config{MaxRetries: aws.Int(0)},
+		Config: aws.Config{
+			MaxRetries:          aws.Int(0),
+			STSRegionalEndpoint: endpoints.RegionalSTSEndpoint,
+		},
 		SharedConfigState: session.SharedConfigEnable,
 	}
 	for _, optFunc := range optFuncs {


patrickdillon:
@yunjiang29 would it be possible for you to do pre-merge testing on this? We are not setup for STS. Do you have the ability to do STS? If so, could you test this on ap-souteast-3 region?

yunjiang29:
@patrickdillon from installer side, STS cluster can be installed on `ap-southeast-3` successfully whether or not include this PR. Is there any way to determine if the regional endpoint is being used?

From ROSA side, the latest payload version is 4.11, but looks like this fix is for 4.12, so it can not be verified on ROSA at present.  cc @yasun1

oriAdler:
> @patrickdillon from installer side, STS cluster can be installed on `ap-southeast-3` successfully whether or not include this PR. Is there any way to determine if the regional endpoint is being used?
> 
> From ROSA side, the latest payload version is 4.11, but looks like this fix is for 4.12, so it can not be verified on ROSA at present. cc @yasun1

@yunjiang29 can you please elaborate on the fix in 4.12?
Were you able to install a ROSA STS cluster in **ap-southeast-3**?

A way to determine if the regional endpoint is used is to check the call to the AWS SDK, we did it using the logs in CS 
`2022-07-14T08:01:59Z INFO logging/transport.go:60 [opid='d680144e-e423-4cfe-9190-196f1077f6bf'] [cid='1tenusges9mp4u0s4nbhb94k24mq9io0'] Got back http 200 for POST https://sts.ap-southeast-3.amazonaws.com/`

We have an open bug for creating ROSA STS cluster on **ap-southeast-3** https://issues.redhat.com/browse/SDA-6243, this MR might solve it.

yunjiang29:
Hello @oriAdler 

> can you please elaborate on the fix in 4.12?

The target branch of this PR is master, it's for 4.12, is it right?

> Were you able to install a ROSA STS cluster in ap-southeast-3?

I thought this PR should be a patch for 4.12, can it be applied on 4.11? Since there is no 4.12 ROSA payload at present.

yasun1:
@oriAdler  This is the pre-merge test for ocp 4.12. We rosa can not validate the ocp pre-merge on ocm side directly, now the official version for rosa in the stable channel is still 4.10.

yasun1:
@oriAdler  This is the pre-merge test for ocp 4.12. We rosa can not validate the ocp pre-merge on ocm side directly, now the official version for rosa in the stable channel is still 4.10.

yasun1:
@oriAdler  This is the pre-merge test for ocp 4.12. We rosa can not validate the ocp pre-merge on ocm side directly, now the official version for rosa in the stable channel is still 4.10.

patrickdillon:
/hold
I'm putting a hold on this pr, because it does not seem like this is the direction we should be going. It's my understanding that for STS, global endpoints need to be enabled for all regions. 

The direction proposed by this PR would need to be part of a larger fix. We would need to enable regional endpoints in TF and all cluster operators that authenticate to aws.

jharrington22:
@patrickdillon we should be able to close this PR, the issue was with the way in which we get credentials during ROSA installs. 

The installer doesn't actually get any credentials from Hive. For ROSA/Managed OpenShift when Hive initiates an install for STS clusters it mounts a special AWS config into the installer pod[1]. The AWS SDK inside the installer is actually configured to call `hiveutil` to get credentials[2], `hiveutil` calls the AWS API AssumeRole and returns credentials when ever they are needed. This better explains why the errors we were seeing immediately as the installer started. The very first call to AWS that the installer makes the AWS SDK has to generate credentials using `hiveutil` wchich it failed immediately.

We've enabled the ap-southeast-3 region and set the STS configuration to "Valid in all AWS Regions" in the AWS account where the Hive was provisioning clusters, this fixed the problem.

r4f4:
@jharrington22 thanks for the investigation and update
/close

PR #6127, opened 2022-07-15
Bug 2102324: Fix validation errors for instance type

rna-afk:
Fixed an issue where the error message returned in case there are
no zones retrieved from the APIs.


jstuever (commented):
/lgtm

PR #6131, opened 2022-07-15
AWS: destroy: move helper functions to their own file

r4f4:
This should de-clutter the `aws.go` file a little bit.


barbacbd (commented):
nit: consider naming the file ec2.go and drop the helpers.

barbacbd (commented):
nit: consider naming the file iam.go and drop the helpers.

barbacbd (commented):
nit: consider naming the file elb.go and drop the helpers.

barbacbd:
suggestion: The PR is centered around breaking out AWS destroy code from a single file to multiple files based on the service. I don't know that it really needs other changes to me. However, The ability to add unit tests would be a strong plus. It is possible to make each of these files related to services an interface that can then be mocked. This does NOT need to be completed here, but it is a consideration.

patrickdillon:
> suggestion: The PR is centered around breaking out AWS destroy code from a single file to multiple files based on the service. I don't know that it really needs other changes to me. However, The ability to add unit tests would be a strong plus. It is possible to make each of these files related to services an interface that can then be mocked. This does NOT need to be completed here, but it is a consideration.

It is a great consideration, and we need to find a place to capture this work/research. I think we need something like an epic that would capture every area of the installer that is untested and we would create sub epics for, say, destroy/<platform>. We would want to investigate the best way to simulate/mock these tests.

patrickdillon:
I think this is a worthwhile refactor. Thanks!

/approve

r4f4:
@patrickdillon any opinion on dropping `helpers` from the filenames as @barbacbd suggested? I'm fine with either.

barbacbd (commented):
/lgtm

I am fine with the file names it was just a nit/suggestion.

PR #6134, opened 2022-07-18
AGENT-137: Convert the install-config to ZTP assets

pawanpinjarkar:
Create ZTP assets except for nmstateconfig from the user-provided install-config.yaml.

### DIFF
--- a/pkg/asset/agent/manifests/agent.go
+++ b/pkg/asset/agent/manifests/agent.go
@@ -130,7 +130,10 @@ func (m *AgentManifests) validateNMStateLabelSelector() field.ErrorList {
 
 	for _, networkConfig := range m.NMStateConfigs {
 		if !reflect.DeepEqual(m.InfraEnv.Spec.NMStateConfigLabelSelector.MatchLabels, networkConfig.ObjectMeta.Labels) {
-			allErrs = append(allErrs, field.Required(fieldPath, fmt.Sprintf("infraEnv and %s.NMStateConfig labels do not match", networkConfig.Name)))
+			allErrs = append(allErrs, field.Required(fieldPath, fmt.Sprintf("infraEnv and %s.NMStateConfig labels do not match. Expected: %s Found: %s",
+				networkConfig.Name,
+				m.InfraEnv.Spec.NMStateConfigLabelSelector.MatchLabels,
+				networkConfig.ObjectMeta.Labels)))
 		}
 
 	}
--- a/pkg/asset/agent/manifests/agentclusterinstall.go
+++ b/pkg/asset/agent/manifests/agentclusterinstall.go
@@ -12,8 +12,6 @@ import (
 	"github.com/openshift/installer/pkg/asset"
 	"github.com/openshift/installer/pkg/asset/agent"
 	"github.com/openshift/installer/pkg/ipnet"
-	baremetaltypes "github.com/openshift/installer/pkg/types/baremetal"
-	vspheretypes "github.com/openshift/installer/pkg/types/vsphere"
 	"github.com/openshift/installer/pkg/validate"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -108,15 +106,13 @@ func (a *AgentClusterInstall) Generate(dependencies asset.Parents) error {
 			},
 		}
 
-		// set APIVIP and IngressVIP only for non SNO cluster with Baremetal and Vsphere platforms
+		apiVIP, ingressVIP := getVIPs(&installConfig.Config.Platform)
+
+		// set APIVIP and IngressVIP only for non SNO cluster for Baremetal and Vsphere platforms
 		// SNO cluster is determined by number of ControlPlaneAgents which should be 1
-		if (int(*installConfig.Config.ControlPlane.Replicas) > 1) &&
-			(installConfig.Config.Platform.Name() == baremetaltypes.Name ||
-				installConfig.Config.Platform.Name() == vspheretypes.Name) {
-			apiVIP, ingressVIP := getVIPs(&installConfig.Config.Platform)
+		if int(*installConfig.Config.ControlPlane.Replicas) > 1 && apiVIP != "" && ingressVIP != "" {
 			agentClusterInstall.Spec.APIVIP = apiVIP
 			agentClusterInstall.Spec.IngressVIP = ingressVIP
-
 		}
 
 		a.Config = agentClusterInstall
--- a/pkg/asset/agent/manifests/agentclusterinstall.go
+++ b/pkg/asset/agent/manifests/agentclusterinstall.go
@@ -2,13 +2,19 @@ package manifests
 
 import (
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
 	"strings"
 
 	hiveext "github.com/openshift/assisted-service/api/hiveextension/v1beta1"
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	"github.com/openshift/installer/pkg/asset"
 	"github.com/openshift/installer/pkg/asset/agent"
+	"github.com/openshift/installer/pkg/ipnet"
+	baremetaltypes "github.com/openshift/installer/pkg/types/baremetal"
+	vspheretypes "github.com/openshift/installer/pkg/types/vsphere"
+	"github.com/openshift/installer/pkg/validate"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -46,27 +52,73 @@ func (a *AgentClusterInstall) Generate(dependencies asset.Parents) error {
 	dependencies.Get(installConfig)
 
 	if installConfig.Config != nil {
-		agentClusterInstall := &hiveext.AgentClusterInstall{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      "agent-cluster-install",
-				Namespace: installConfig.Config.Namespace,
-			},
-			Spec: hiveext.AgentClusterInstallSpec{
-				ClusterDeploymentRef: corev1.LocalObjectReference{
-					Name: installConfig.Config.ObjectMeta.Name,
-				},
-				SSHPublicKey: strings.Trim(installConfig.Config.SSHKey, "|\n\t"),
-				ProvisionRequirements: hiveext.ProvisionRequirements{
-					ControlPlaneAgents: int(*installConfig.Config.ControlPlane.Replicas),
-				},
-			},
-		}
-
 		var numberOfWorkers int = 0
 		for _, compute := range installConfig.Config.Compute {
 			numberOfWorkers = numberOfWorkers + int(*compute.Replicas)
 		}
-		agentClusterInstall.Spec.ProvisionRequirements.WorkerAgents = numberOfWorkers
+
+		clusterNetwork := []hiveext.ClusterNetworkEntry{}
+		for _, cn := range installConfig.Config.Networking.ClusterNetwork {
+			_, cidr, err := net.ParseCIDR(cn.CIDR.String())
+			if err != nil {
+				return errors.Wrap(err, "failed to parse ClusterNetwork CIDR")
+			}
+			err = validate.SubnetCIDR(cidr)
+			if err != nil {
+				return errors.Wrap(err, "failed to validate ClusterNetwork CIDR")
+			}
+
+			entry := hiveext.ClusterNetworkEntry{
+				CIDR:       cidr.String(),
+				HostPrefix: cn.HostPrefix,
+			}
+			clusterNetwork = append(clusterNetwork, entry)
+		}
+
+		serviceNetwork := []string{}
+		for _, sn := range installConfig.Config.Networking.ServiceNetwork {
+			cidr, err := ipnet.ParseCIDR(sn.String())
+			if err != nil {
+				return errors.Wrap(err, "failed to parse ServiceNetwork CIDR")
+			}
+			serviceNetwork = append(serviceNetwork, cidr.String())
+		}
+
+		agentClusterInstall := &hiveext.AgentClusterInstall{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      getAgentClusterInstallName(installConfig),
+				Namespace: getObjectMetaNamespace(installConfig),
+			},
+			Spec: hiveext.AgentClusterInstallSpec{
+				ImageSetRef: &hivev1.ClusterImageSetReference{
+					Name: "openshift-4.11",
+				},
+				ClusterDeploymentRef: corev1.LocalObjectReference{
+					Name: getClusterDeploymentName(installConfig),
+				},
+				Networking: hiveext.Networking{
+					ClusterNetwork: clusterNetwork,
+					ServiceNetwork: serviceNetwork,
+				},
+				SSHPublicKey: strings.Trim(installConfig.Config.SSHKey, "|\n\t"),
+				ProvisionRequirements: hiveext.ProvisionRequirements{
+					ControlPlaneAgents: int(*installConfig.Config.ControlPlane.Replicas),
+					WorkerAgents:       numberOfWorkers,
+				},
+			},
+		}
+
+		// set APIVIP and IngressVIP only for non SNO cluster with Baremetal and Vsphere platforms
+		// SNO cluster is determined by number of ControlPlaneAgents which should be 1
+		if (int(*installConfig.Config.ControlPlane.Replicas) > 1) &&
+			(installConfig.Config.Platform.Name() == baremetaltypes.Name ||
+				installConfig.Config.Platform.Name() == vspheretypes.Name) {
+			apiVIP, ingressVIP := getVIPs(&installConfig.Config.Platform)
+			agentClusterInstall.Spec.APIVIP = apiVIP
+			agentClusterInstall.Spec.IngressVIP = ingressVIP
+
+		}
+
 		a.Config = agentClusterInstall
 
 		agentClusterInstallData, err := yaml.Marshal(agentClusterInstall)
--- a/pkg/asset/agent/manifests/agentclusterinstall.go
+++ b/pkg/asset/agent/manifests/agentclusterinstall.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	hiveext "github.com/openshift/assisted-service/api/hiveextension/v1beta1"
 	"github.com/openshift/installer/pkg/asset"
@@ -54,7 +55,7 @@ func (a *AgentClusterInstall) Generate(dependencies asset.Parents) error {
 				ClusterDeploymentRef: corev1.LocalObjectReference{
 					Name: installConfig.Config.ObjectMeta.Name,
 				},
-				SSHPublicKey: installConfig.Config.SSHKey,
+				SSHPublicKey: strings.Trim(installConfig.Config.SSHKey, "|\n\t"),
 				ProvisionRequirements: hiveext.ProvisionRequirements{
 					ControlPlaneAgents: int(*installConfig.Config.ControlPlane.Replicas),
 				},
--- a/pkg/asset/agent/manifests/agentclusterinstall.go
+++ b/pkg/asset/agent/manifests/agentclusterinstall.go
@@ -91,7 +91,7 @@ func (a *AgentClusterInstall) Generate(dependencies asset.Parents) error {
 			},
 			Spec: hiveext.AgentClusterInstallSpec{
 				ImageSetRef: &hivev1.ClusterImageSetReference{
-					Name: "openshift-4.11",
+					Name: getClusterImageSetReferenceName(),
 				},
 				ClusterDeploymentRef: corev1.LocalObjectReference{
 					Name: getClusterDeploymentName(installConfig),
--- a/pkg/asset/agent/manifests/agentclusterinstall.go
+++ b/pkg/asset/agent/manifests/agentclusterinstall.go
@@ -9,6 +9,8 @@ import (
 	"github.com/openshift/installer/pkg/asset"
 	"github.com/openshift/installer/pkg/asset/agent"
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/yaml"
 )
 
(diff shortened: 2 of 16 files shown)


pawanpinjarkar:
/hold
Until testing the installer binary generated using install config and nmstateconfig yaml results in successful manual SNO cluster installation.

pawanpinjarkar:
/hold cancel
Successfully tested SNO cluster. Created agent.iso using the installer binary with cluster-manifests/nmstateconfig.yaml and install-config.yaml provided. Manually booted a VM using the agent.iso

pawanpinjarkar:
@zaneb @celebdor As we now generate the required assets(infraenv, clusterdeployment, pull-secret, etc) from install-config, do we need to save/create them on the disk as well for users to refer to?

pkg/asset/agent/manifests/clusterdeployment.go
```diff
@@ -37,53 +37,53 @@ func (*ClusterDeployment) Name() string {
 func (*ClusterDeployment) Dependencies() []asset.Asset {
 	return []asset.Asset{
 		&agent.OptionalInstallConfig{},
-		// &AgentPullSecret{},
+		&AgentPullSecret{},
```
zaneb, line 40: Good point about this causing the ClusterDeployment to get discarded, I forgot dependencies also have that effect :facepalm:

So I guess the best we can do is just generate both with the same name and it's the user's problem if they create them in such a way that the links don't match up.

zaneb:
> @zaneb @celebdor As we now generate the required assets(infraenv, clusterdeployment, pull-secret, etc) from install-config, do we need to save/create them on the disk as well for users to refer to?

They will be saved on the disk if the user runs `openshift-install agent create cluster-manifests`. If they go straight to `create image` they won't end up with any files on disk, and that's fine.

pkg/asset/agent/manifests/clusterdeployment.go
pawanpinjarkar: Thanks, fixed. Here is the link for dirty dependencies for ref https://github.com/openshift/installer/blob/master/docs/design/assetgeneration.md#dirty-detection

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -39,42 +42,43 @@ func (*AgentClusterInstall) Dependencies() []asset.Asset {
...
+		agentClusterInstall := &hiveext.AgentClusterInstall{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "agent-cluster-install",
+				Namespace: installConfig.Config.Namespace,
+			},
+			Spec: hiveext.AgentClusterInstallSpec{
+				ClusterDeploymentRef: corev1.LocalObjectReference{
+					Name: installConfig.Config.ObjectMeta.Name,
+				},
+				SSHPublicKey: strings.Trim(installConfig.Config.SSHKey, "|\n\t"),
```
andfasano, line 58: Why this trimming is required? It doesn't seem in other locations of the installer where `installConfig.Config.SSHKey` is used

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -39,42 +42,43 @@ func (*AgentClusterInstall) Dependencies() []asset.Asset {
...
-	// }
-
-	// a.File = &asset.File{
-	// 	Filename: agentClusterInstallFilename,
-	// 	Data:     agentClusterInstallData,
-	// }
+	installConfig := &agent.OptionalInstallConfig{}
+	dependencies.Get(installConfig)
+
+	if installConfig.Config != nil {
```
andfasano, line 48: How could happen this condition? If the dependency gets successfully resolved shouldn't be always valid?

pkg/asset/agent/manifests/agentpullsecret.go
```diff
@@ -43,33 +44,35 @@ func (*AgentPullSecret) Dependencies() []asset.Asset {
...
-	// }
-
-	// a.File = &asset.File{
-	// 	Filename: agentPullSecretFilename,
-	// 	Data:     secretData,
-	// }
+	installConfig := &agent.OptionalInstallConfig{}
+	dependencies.Get(installConfig)
+
+	if installConfig.Config != nil {
```
andfasano, line 50: Also this one

pkg/asset/agent/manifests/common.go
```diff
@@ -1,62 +1,78 @@
...
 
-// import (
-// 	"github.com/openshift/installer/pkg/asset/agent"
-// 	"github.com/openshift/installer/pkg/asset/installconfig"
-// 	"github.com/openshift/installer/pkg/types"
-// 	corev1 "k8s.io/api/core/v1"
-// 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-// 	"k8s.io/utils/pointer"
-// )
+import (
```
andfasano, line 3: If this file is meant to store shared test utils, for clarity I'd suggest to add the `_test.go` extension. If it's not possible due the imports, then please add a `test` suffix in the file name, ie `agenttest`.

pkg/asset/agent/manifests/infraenv.go
```diff
@@ -36,49 +39,49 @@ func (*InfraEnv) Name() string {
 func (*InfraEnv) Dependencies() []asset.Asset {
 	return []asset.Asset{
 		&agent.OptionalInstallConfig{},
-		// &AgentPullSecret{},
```
andfasano, line 39: Not sure if I missed some previous discussion, but why the pull secret and nmstateconfig are missing here?

pkg/asset/agent/manifests/clusterimageset_test.go
```diff
@@ -17,34 +18,44 @@ import (
...
+				GetValidOptionalInstallConfig(),
+				&releaseimage.Image{},
+			},
+			expectedError: "invalid ClusterImageSet configuration: Spec.ReleaseImage: Forbidden: value must be equal to registry.ci.openshift.org/origin/release:4.11",
+		},
+		{
+			name: "valid configuration",
+			dependencies: []asset.Asset{
+				GetValidOptionalInstallConfig(),
+				GetValidReleaseimage(),
```
andfasano, line 47:
```suggestion
				&releaseimage.Image{
					PullSpec: "registry.ci.openshift.org/origin/release:4.11",
				},
```
We should avoid mixing test and production code also, the test will break when `defaultReleaseImageOriginal` will get automatically updated in the next release

pkg/asset/agent/manifests/infraenv.go
pawanpinjarkar: This causes a dirty dependency issue and image generation fails with 
```
WARNING       Discarding the InfraEnv Config that was provided in the target directory
because its dependencies are dirty and it needs to be regenerated.
............
............
FATAL failed to fetch Agent Installer ISO: failed to fetch dependency of "Agent Installer ISO":
 failed to fetch dependency of "Agent Installer Ignition": 
failed to fetch dependency of "Agent Manifests": 
failed to generate asset "InfraEnv Config": missing configuration or manifest file
```
Ref https://github.com/openshift/installer/pull/6134#discussion_r925052704

pkg/asset/agent/manifests/agentclusterinstall.go
pawanpinjarkar: I also trimmed it here for infra-env https://github.com/pawanpinjarkar/installer/blob/convert-install-config-to-ztp-assets/pkg/asset/agent/manifests/infraenv.go#L64

If we don't and if the user provided install-config has multiline in the `ssh-key` like [this](https://gist.github.com/pawanpinjarkar/ff1186f4dd5d9294612ea4cf4ccc27e3#file-install-config-yaml-L31-L32), for example, then the `create-cluster-and-infraenv.service` fails and the assisted-service pod throws below error

```
time="2022-07-20T15:27:02Z" level=error 
msg="Failed to register InfraEnv infraEnv with id 55231fc2-bcfc-4414-89d3-78fd19a6e9ea. 
Error: SSH key is not valid" 
func="github.com/openshift/assisted-service/internal/bminventory.(*bareMetalInventory)
.RegisterInfraEnvInternal.func1" 
file="/src/internal/bminventory/inventory.go:3891" 
cluster_id=55231fc2-bcfc-4414-89d3-78fd19a6e9ea go-id=35
pkg=Inventory request_id=11d830d2-f13c-4d40-9ddb-6942466ee3ee
```

pkg/asset/agent/manifests/agentclusterinstall.go
pawanpinjarkar: This is added for this test case https://github.com/pawanpinjarkar/installer/blob/convert-install-config-to-ztp-assets/pkg/asset/agent/manifests/agentclusterinstall_test.go#L31-L35

In absence of this check the test case fails with below error -
```
--- FAIL: TestAgentClusterInstall_Generate (2.53s)
    --- FAIL: TestAgentClusterInstall_Generate/missing_install_config (2.53s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x2ffb8d3]

goroutine 68 [running]:
testing.tRunner.func1.2({0x33c9c20, 0x578af20})
	/usr/local/go/src/testing/testing.go:1209 +0x47e
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1212 +0x350
panic({0x33c9c20, 0x578af20})
	/usr/local/go/src/runtime/panic.go:1047 +0x262
github.com/openshift/installer/pkg/asset/agent/manifests.(*AgentClusterInstall).Generate(0xc00009ba08, 0xc00009ba68)
	/home/pawanpinjarkar/go/src/github.com/openshift/installer/pkg/asset/agent/manifests/agentclusterinstall.go:58 +0x113
github.com/openshift/installer/pkg/asset/agent/manifests.TestAgentClusterInstall_Generate.func1(0xc000603ba0)
	/home/pawanpinjarkar/go/src/github.com/openshift/installer/pkg/asset/agent/manifests/agentclusterinstall_test.go:66 +0x138
testing.tRunner(0xc000603ba0, 0xc0006a1d00)
	/usr/local/go/src/testing/testing.go:1259 +0x1db
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:1306 +0x673
```
Essentially, when the `Config` is empty and trying to retrieve values out of it.

zaneb: It's OptionalInstallConfig, so if the user doesn't provide one this will be nil. (You can also check the `installConfig.Supplied` flag, but the effect is that same and in retrospect I probably didn't need to create that.)

andfasano: Right, I didn't notice it initially :). Can you please remember me why it was required? I mean, as per the asset infrastructure during the fetch first of all it tries to load the file from the the disk, if not it goes to the Generate(). I'd say the if end up in a asset Generate() then the install-config must be present

andfasano: (so now that we're moving to fully support install-config maybe such test doesn't make sense anymore for an asset?)

pkg/asset/agent/manifests/infraenv.go
```diff
@@ -36,49 +39,49 @@ func (*InfraEnv) Name() string {
...
+				Namespace: installConfig.Config.Namespace,
+			},
+			Spec: aiv1beta1.InfraEnvSpec{
+				ClusterRef: &aiv1beta1.ClusterReference{
+					Name:      installConfig.Config.ObjectMeta.Name,
+					Namespace: installConfig.Config.ObjectMeta.Namespace,
+				},
+				SSHAuthorizedKey: strings.Trim(installConfig.Config.SSHKey, "|\n\t"),
+				PullSecretRef: &corev1.LocalObjectReference{
+					Name: "pull-secret",
```
zaneb, line 66: Should use `agentPullSecretName` instead of a literal.

pkg/asset/agent/manifests/infraenv.go
zaneb: The installer effectively doesn't allow dependencies between stuff at the same level (e.g. all stuff created by `agent create cluster-manifests`). You can depend on stuff generated by the previous command, but for stuff generated by the same command it will overwrite the user's changes if there is a dependency between them.
If it looked at file modification times or something (like `make` does) then it would be fine, but here we are.

pkg/asset/agent/manifests/infraenv.go
```diff
@@ -36,49 +39,49 @@ func (*InfraEnv) Name() string {
...
-	// 	Data:     infraEnvData,
-	// }
+	installConfig := &agent.OptionalInstallConfig{}
+	nmStateConfig := &NMStateConfig{}
+	dependencies.Get(installConfig, nmStateConfig)
+
+	if installConfig.Config != nil && nmStateConfig != nil {
+		infraEnv := &aiv1beta1.InfraEnv{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "infraEnv",
```
zaneb, line 56: We should use the cluster name for the InfraEnv name.

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -39,42 +42,43 @@ func (*AgentClusterInstall) Dependencies() []asset.Asset {
...
+				Name:      "agent-cluster-install",
+				Namespace: installConfig.Config.Namespace,
+			},
+			Spec: hiveext.AgentClusterInstallSpec{
+				ClusterDeploymentRef: corev1.LocalObjectReference{
+					Name: installConfig.Config.ObjectMeta.Name,
+				},
+				SSHPublicKey: strings.Trim(installConfig.Config.SSHKey, "|\n\t"),
+				ProvisionRequirements: hiveext.ProvisionRequirements{
+					ControlPlaneAgents: int(*installConfig.Config.ControlPlane.Replicas),
```
zaneb, line 60: Missing `workerAgents`.

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -39,42 +42,43 @@ func (*AgentClusterInstall) Dependencies() []asset.Asset {
...
+	installConfig := &agent.OptionalInstallConfig{}
+	dependencies.Get(installConfig)
+
+	if installConfig.Config != nil {
+		agentClusterInstall := &hiveext.AgentClusterInstall{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "agent-cluster-install",
+				Namespace: installConfig.Config.Namespace,
+			},
+			Spec: hiveext.AgentClusterInstallSpec{
```
zaneb, line 54: Missing `imageSetRef`, `networking`, `apiVip`, `ingressVIP`

pkg/asset/agent/manifests/infraenv.go
```diff
@@ -36,49 +39,49 @@ func (*InfraEnv) Name() string {
...
+				ClusterRef: &aiv1beta1.ClusterReference{
+					Name:      installConfig.Config.ObjectMeta.Name,
+					Namespace: installConfig.Config.ObjectMeta.Namespace,
+				},
+				SSHAuthorizedKey: strings.Trim(installConfig.Config.SSHKey, "|\n\t"),
+				PullSecretRef: &corev1.LocalObjectReference{
+					Name: "pull-secret",
+				},
+				NMStateConfigLabelSelector: metav1.LabelSelector{
+					MatchLabels: nmStateConfig.Config[0].ObjectMeta.Labels,
```
zaneb, line 69: We'll have to drop this dependency as well. For now at least, we'll need to just define some label that we always apply when we generate the NMState config and use that in the selector.

I'd suggest `infraenvs.agent-install.openshift.io = <name of infraenv>` (same as ZTP applies to BareMetalHosts) as a starting point.

zaneb:
Maybe it would make sense to define a bunch of functions like:

```
func infraEnvName(ic *agent.OptionalInstallConfig) {
    return ic.Config.Name
}
```

for all the places where references have to match up, so that there's no possibility of changing them in only one place such that they no longer match.

pkg/asset/agent/manifests/infraenv.go
pawanpinjarkar: Yes, that's the better solution atm.

pkg/asset/agent/manifests/agentclusterinstall.go
pawanpinjarkar: Good catch. I relied on the previously commented generated code and overlooked that it is missing the imp fields.

pkg/asset/agent/manifests/infraenv.go
pawanpinjarkar: We will go with `<clustername>-pull-secret` as discussed in slack with @zaneb

pkg/asset/agent/manifests/agentclusterinstall.go
zaneb: I guess we still need to decide what to do if the user doesn't supply an install-config _or_ ZTP manifests and they run "agent create cluster-manifests". Maybe the answer is to just return an error. (The answer is definitely not to panic though :laughing:)
It would be nice to generate a default template that users can fill in, but that might be a bit challenging for now (how good is our validation that they actually filled it in before creating the image?) so maybe that is a story for another day. I just created [AGENT-301](https://issues.redhat.com/browse/AGENT-301 to keep track of that.

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -39,42 +48,89 @@ func (*AgentClusterInstall) Dependencies() []asset.Asset {
...
+		}
+
+		agentClusterInstall := &hiveext.AgentClusterInstall{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      getAgentClusterInstallName(installConfig),
+				Namespace: getObjectMetaNamespace(installConfig),
+			},
+			Spec: hiveext.AgentClusterInstallSpec{
+				ImageSetRef: &hivev1.ClusterImageSetReference{
+					Name: "openshift-4.11",
```
rwsu, line 94: I don't think we should hard-code the version number here. I think there are ways to find the version number like: https://github.com/openshift/installer/blob/37ae964e24395c6fc23a349fe39e7037ac66a99d/pkg/asset/releaseimage/default.go#L23

andfasano: Should't be the name of the `ClusterImageSet` resource?

andfasano: (note: in case the current version will be required, I'd recommed to use [version.Version()](https://github.com/openshift/installer/blob/37ae964e24395c6fc23a349fe39e7037ac66a99d/pkg/version/version.go#L52))

zaneb: I agree we shouldn't hard-code it. I think it's fine to set the name to just the name of the cluster. If we're going to put a version then we should go with whatever Andrea suggests :)

zaneb: But either way, we should have a `getClusterImageSetName()` function to get it.
(Currently the ClusterImageSet name is set to `"openshift-" + version.Version()` as Andrea suggested.)

pawanpinjarkar: ok, so it will get the value as ` openshift-was not built correctly`

pkg/asset/agent/manifests/common.go
```diff
@@ -1,62 +1,50 @@
...
+	return ic.Config.ObjectMeta.Name + "pull-secret"
+}
+
+func getObjectMetaNamespace(ic *agent.OptionalInstallConfig) string {
+	return ic.Config.ObjectMeta.Namespace
+}
+
+func getNMStateConfigLabelSelector(ic *agent.OptionalInstallConfig) metav1.LabelSelector {
+	return metav1.LabelSelector{
+		MatchLabels: map[string]string{
```
zaneb, line 32: We'll want to apply the same labels to the NMStateConfig resources, so I'd suggest returning just this map and move the LabelSelector outside the function.

pkg/asset/agent/manifests/agent.go
```diff
@@ -130,7 +130,10 @@ func (m *AgentManifests) validateNMStateLabelSelector() field.ErrorList {
 
 	for _, networkConfig := range m.NMStateConfigs {
 		if !reflect.DeepEqual(m.InfraEnv.Spec.NMStateConfigLabelSelector.MatchLabels, networkConfig.ObjectMeta.Labels) {
```
zaneb, line 132: This is actually the wrong test, we should really only be checking that the label selector does actually select those labels. Don't need to fix that now though.

pkg/asset/agent/manifests/agentclusterinstall.go
andfasano: That happens only when you build from source (and without overriding the release) (ie see also https://github.com/openshift/installer/blob/6f836627e01f868c8ff62e6d19794d76b289d433/pkg/asset/agent/image/releaseimage_test.go#L44)

zaneb: lol, hopefully proper builds will have something more helpful.
Thinking a bit more...
Advantage of using the version: you can see the version just by looking at the AgentClusterInstall, without having to go look at the ClusterImageSet as well.
Advantage of using the cluster name: people will be less tempted to just change the name of the ImageSetRef and expect that to affect the version.

I think I'm tempted to just use the cluster name.

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -39,42 +48,89 @@ func (*AgentClusterInstall) Dependencies() []asset.Asset {
...
+					WorkerAgents:       numberOfWorkers,
+				},
+			},
+		}
+
+		// set APIVIP and IngressVIP only for non SNO cluster with Baremetal and Vsphere platforms
+		// SNO cluster is determined by number of ControlPlaneAgents which should be 1
+		if (int(*installConfig.Config.ControlPlane.Replicas) > 1) &&
+			(installConfig.Config.Platform.Name() == baremetaltypes.Name ||
+				installConfig.Config.Platform.Name() == vspheretypes.Name) {
```
zaneb, line 115: Isn't the check of the platform type effectively handled in `getVIPs()`?

pkg/asset/releaseimage/default.go
```diff
@@ -54,3 +54,8 @@ func Default() (string, error) {
 	}
 	return pullspec, nil
 }
+
+// GetDefaultReleaseImageOriginal returns the default release image
```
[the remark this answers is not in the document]
pawanpinjarkar: Let's keep it as is for now.

pkg/asset/agent/manifests/agentclusterinstall.go
pawanpinjarkar: yes, fixed it.

pkg/asset/releaseimage/default.go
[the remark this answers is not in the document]
pawanpinjarkar: ok, removed as it is not required.

pkg/asset/releaseimage/default.go
```diff
@@ -54,3 +54,4 @@ func Default() (string, error) {
 	}
 	return pullspec, nil
 }
+
```
pawanpinjarkar, line 57: My editor auto adds this newline. Need to remove.

PR #6136, opened 2022-07-19
Bug 2108579: [release-4.9] RHCOS: move to rhcos.mirror.openshift.com

vfreex:
Similar to https://github.com/openshift/installer/pull/6109.


sdodson:
/hold
I'd like to see us ship 4.10.z with this change and wait a week until we move forward with backports to 4.9 and 4.8. Delaying by a week or two won't change anything with regard to whether or not we can bring down the old infrastructure because that's going to be bound by the date at which 4.10 goes EOL.

sdodson:
/close
This got pulled into #6159

PR #6138, opened 2022-07-19
Bug 2108582: [release-4.7] RHCOS: move to rhcos.mirror.openshift.com

vfreex:
Similar to https://github.com/openshift/installer/pull/6109.


jstuever (approved):
/approve

sdodson:
/close
4.7 is now EOL

PR #6140, opened 2022-07-19
AWS: resolve subnet-related FIXMEs

r4f4:
When looking at the subnet code I noticed the comment `// FIXME: port to DescribeSubnetsPagesWithContext once we bump our vendored AWS package past v1.19.30`. So I did it.

### DIFF
--- a/pkg/destroy/aws/aws.go
+++ b/pkg/destroy/aws/aws.go
@@ -1297,8 +1297,8 @@ func deleteEC2Subnet(ctx context.Context, client *ec2.EC2, id string, logger log
 }
 
 func deleteEC2SubnetsByVPC(ctx context.Context, client *ec2.EC2, vpc string, failFast bool, logger logrus.FieldLogger) error {
-	// FIXME: port to DescribeSubnetsPages once we bump our vendored AWS package past v1.19.30
-	results, err := client.DescribeSubnetsWithContext(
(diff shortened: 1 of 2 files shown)


pkg/destroy/aws/aws.go
```diff
@@ -1308,26 +1308,27 @@ func deleteEC2SubnetsByVPC(ctx context.Context, client *ec2.EC2, vpc string, fai
 				},
 			},
 		},
+		func(results *ec2.DescribeSubnetsOutput, lastPage bool) bool {
+			for _, subnet := range results.Subnets {
+				err := deleteEC2Subnet(ctx, client, *subnet.SubnetId, logger.WithField("subnet", *subnet.SubnetId))
+				if err != nil {
+					err = errors.Wrapf(err, "deleting EC2 subnet %s", *subnet.SubnetId)
+					if lastError != nil {
```
barbacbd, line 1316: This could print the same error multiple times. Is this intended or should this only print new errors?

barbacbd (commented):
/lgtm

pkg/destroy/aws/aws.go
r4f4: We discussed this on Slack and we don't think the error will be printed multiple times. Also, this was the existent behaviour already.

PR #6144, opened 2022-07-20
BUG 1952739: Configurable PowerOn Timeout for vsphere

barbacbd:
** Use the poweron_timeout terrafrom value for vsphere. Currently setting the
value to 600 seconds (10 minutes, default is 5 minutes in the provider).

### DIFF
--- a/data/data/vsphere/master/main.tf
+++ b/data/data/vsphere/master/main.tf
@@ -22,6 +22,7 @@ resource "vsphere_virtual_machine" "vm_master" {
   folder               = var.folder
   enable_disk_uuid     = "true"
   annotation           = local.description
+  poweron_timeout      = var.vsphere_poweron_timeout
 
   wait_for_guest_net_timeout  = "0"
   wait_for_guest_net_routable = "false"
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -2965,6 +2965,12 @@ spec:
                     description: Password is the password for the user to use to connect
                       to the vCenter.
                     type: string
+                  powerOnTimeout:
+                    description: PowerOnTimeout is the number of seconds to allow
+                      the virtual machine to power on. When not specified, it will
+                      be set according to the vsphere terraform provider
+                    format: int64
+                    type: integer
                   resourcePool:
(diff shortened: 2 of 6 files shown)


patrickdillon:
Sorry for the confusion Brent but we don’t want to make this configurable by the user

patrickdillon:
e2e-vsphere is failing with 

```
evel=error msg=Error: expected poweron_timeout to be at least (300), got 0
```
So at least that shows us that `poweron_timeout` is indeed a thing.

data/data/vsphere/master/main.tf
```diff
@@ -22,6 +22,7 @@ resource "vsphere_virtual_machine" "vm_master" {
   folder               = var.folder
   enable_disk_uuid     = "true"
   annotation           = local.description
+  poweron_timeout      = var.vsphere_poweron_timeout
```
patrickdillon, line 25: I think this entire PR can be:
```suggestion
  poweron_timeout      =  600
```
Again, sorry for the confusion about making this configurable.

I don't see much of a downside for adding this change. All it means is that when we do hit this error it is going to take 5 minutes longer for the install to fail.

@jcpowermac Do you think it's possible this could help with the BZ? i.e. are there cases where a VM doesn't power on for 5 minutes but might power on in, say, 7 minutes?

jcpowermac: I was able to catch one of these virtual machines that experienced this issue and even hours later the allowed methods was missing poweron. So while I don't think 10 minutes will hurt anything I also don't think it will resolve anything either.

https://github.com/hashicorp/terraform-provider-vsphere/blob/fa3580820d1e27ff36419f87f8c2b487c6a167fa/vsphere/internal/helper/virtualmachine/virtual_machine_helper.go#L680

Occurrences the last 14 days in CI
https://search.ci.openshift.org/?search=PowerOnVM_Task&maxAge=336h&context=1&type=junit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

I don't think we have any user reports of this issue though, right?

patrickdillon:
/close
as @jcpowermac mentioned there are no user cases for this. We will not fix for reasons described here: https://bugzilla.redhat.com/show_bug.cgi?id=1952739#c23

PR #6145, opened 2022-07-20
[WIP] data/data/azure: Don't create NSG in existing network

jhixson74:
Allow the ability for the installer to specify an existing NSG


patrickdillon (commented):
Following where the output variable `nsg_name` is used, it ends up being used to create the security rule for SSH access to the bootstrap node: https://github.com/openshift/installer/blob/731ba91b9ed5ca28a6c3d07d9a1d546c5126901a/data/data/azure/bootstrap/main.tf#L248

So it seems like this implementation would break SSH access to the bootstrap node on existing networks. I will update the bug to indicate this is the use of the NSG.

But we would still have a useless NSG floating around after the bootstrap node is torn down:
Would it make sense to do something like: when `nsg_name` is null, we create a bootstrap nsg that would be torn down with the bootstrap node?

patrickdillon:
/lgtm
/assign @jstuever 

After today's conversation, my concerns about SSH appear to be unfounded; so I'm dropping the idea about testing that. 

Where did we land regarding documenting opening Port 22 in the docs for installing to a shared vnet? Do we think that is worthwhile? If so, I will open a bz.

patrickdillon:
/lgtm cancel 

This LGTM but I am cancelling my LGTM due to CI concerns:

1. All azure tests are failing, but when I try to access logs from prow, I get the k8s equivalent of a spinning beachball
2. These CI jobs do not seem to be required for this PR, in which case I would just leave the LGTM; but I am baffled after reviewing the [job](https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/installer/openshift-installer-master-presubmits.yaml#L939) [config](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L284-L288) as to why it is not required

patrickdillon:
/retest-required

e2e-azure* tests are failing due to failure to build the upi image, let's run again

MayXuQQ:
when verify the PR, whether must based on the 'publish: Internal' ?

MayXuQQ:
> > when verify the PR, whether must based on the 'publish: Internal' ? @jhixson74 @patrickdillon currently , with the pre-merge build 4.11.0-0.ci.test-2022-07-22-043801-ci-ln-f45y3yb-latest, IPI install failed : private cluster (no worker), public cluster (just bootstrap)
> 
> The main changes in this PR relate to `installconfig.platform.azure.virtualNetwork` rather than `publish`, but `publish: Internal` should not fail. What are the details for the failure?

test with public cluster, no more nsg created. cluster create succeed. but cluster failed 
$ oc describe co ingress
_The "default" ingress controller reports Available=False: IngressControllerUnavailable: One or more status conditions indicate unavailable: LoadBalancerReady=False (SyncLoadBalancerFailed: The service-controller component is reporting SyncLoadBalancerFailed events like: Error syncing load balancer: failed to ensure load balancer: nsg "maxu-nsgn-xbbzr-nsg" not found
The kube-controller-manager logs may contain more details.)_

actually, just has "[maxu-nsgn-nsg]" in the customer vpc. no nsg in the cluster group.

jstuever:
The CI job might not be setup for success here as it might be relying on the bits we removed.

MayXuQQ:
check more the cluster "infraID": "gpei-test-vhv5c", use the existed networkResourceGroupName: gpei-test-rg
1. ./manifests/cloud-provider-config.yaml:   "securityGroupName\": \"gpei-test-vhv5c-nsg\"
2. grep gpei-test-vhv5c-nsg . -rl -n | grep configmaps.yaml
./registry-build05-ci-openshift-org-ci-ln-f45y3yb-stable-sha256-fd1bb65180d169a046d3405bdef131a0af5cbc394258688d446702323d4d01b2/namespaces/openshift-cloud-controller-manager/core/configmaps.yaml
./registry-build05-ci-openshift-org-ci-ln-f45y3yb-stable-sha256-fd1bb65180d169a046d3405bdef131a0af5cbc394258688d446702323d4d01b2/namespaces/openshift-config-managed/core/configmaps.yaml
./registry-build05-ci-openshift-org-ci-ln-f45y3yb-stable-sha256-fd1bb65180d169a046d3405bdef131a0af5cbc394258688d446702323d4d01b2/namespaces/openshift-config/core/configmaps.yaml
./registry-build05-ci-openshift-org-ci-ln-f45y3yb-stable-sha256-fd1bb65180d169a046d3405bdef131a0af5cbc394258688d446702323d4d01b2/namespaces/openshift-kube-apiserver/core/configmaps.yaml
./registry-build05-ci-openshift-org-ci-ln-f45y3yb-stable-sha256-fd1bb65180d169a046d3405bdef131a0af5cbc394258688d446702323d4d01b2/namespaces/openshift-kube-controller-manager/core/configmaps.yaml

patrickdillon:
/close
We determined that the bug that sparked this is not really a bug. Please reopen if I am missing something.

PR #6149, opened 2022-07-21
Bug 2109538: nutanix: allow creating manifests without Prism Central connection

vrutkovs:
In Assisted Installer we create manifests and generate Ignition in the cloud and pass ignition files to the agents. For platform integration we can't request credentials from users, instead we fill these in with fake credentials and let them change these later.

This commit moves validations to ValidateForProvisioning, so that `openshift-install create manifests` phase would pass with fake credentials

### DIFF
--- a/pkg/asset/installconfig/nutanix/validation.go
+++ b/pkg/asset/installconfig/nutanix/validation.go
@@ -17,21 +17,6 @@ func Validate(ic *types.InstallConfig) error {
 		return field.Required(field.NewPath("platform", "nutanix"), "nutanix validation requires a nutanix platform configuration")
 	}
 
-	p := ic.Platform.Nutanix
-	nc, err := nutanixtypes.CreateNutanixClient(context.TODO(),
-		p.PrismCentral.Endpoint.Address,
-		strconv.Itoa(int(p.PrismCentral.Endpoint.Port)),
-		p.PrismCentral.Username,
-		p.PrismCentral.Password)
-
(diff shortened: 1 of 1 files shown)


thunderboltsid:
My only concern is that if someone has a misconfigured PE on install-config, they won’t get an error on running create manifests but only on create cluster .

thunderboltsid (approved):
LGTM

pkg/asset/installconfig/nutanix/validation.go
```diff
@@ -53,6 +38,14 @@ func ValidateForProvisioning(ic *types.InstallConfig) error {
...
 	}
 
+	// validate whether a prism element with the UUID actually exists
+	for _, pe := range p.PrismElements {
+		_, err = nc.V3.GetCluster(pe.UUID)
+		if err != nil {
+			return field.InternalError(field.NewPath("platform", "nutanix", "prismElements"), errors.Wrapf(err, "prism element UUID %s does not correspond to a valid prism element in Prism", pe.UUID))
+		}
+	}
+
```
sadasu, line 48: Now that you getting the "Name" of the PrismElement as an optional parameter, do you need to validate if that exists also? And I am assuming that the Name should correspond to the same UUID as within the Nutanix cluster?

vrutkovs: Yes, the name should match the UUID, it is being validated in `create cluster` phase - or filled in if its unset.

PR #6150, opened 2022-07-21
BUG 2109388: Add AWS S3 Bucket Permissions

barbacbd:
** Added the s3::GetBucketPolicy. This is require the user to have permissions to
return the policy of the specified bucket.


jstuever (commented):
/lgtm

PR #6154, opened 2022-07-21
AGENT-138: Get the base ISO from the release payload

bfournie:
If the 'oc' command is available, use it to extract the base ISO from the release payload instead of downloading it.


pkg/asset/agent/image/oc.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return image, nil
+}
+
+func (r *release) extractFileFromImage(log logrus.FieldLogger, image, cacheDir, pullSecret string, insecure bool, platformType models.PlatformType) (string, error) {
+
+	file := CoreOsFileName
```
zaneb, line 91: Can we add at least a TODO to use `platformType`?

pkg/asset/agent/image/oc.go
```diff
@@ -0,0 +1,133 @@
...
+	OcDefaltRetryDelay = time.Second * 5
+)
+
+type Config struct {
+	MaxTries   uint
+	RetryDelay time.Duration
+}
+
+type Release interface {
+	GetBaseIso(log logrus.FieldLogger, releaseImage string, pullSecret string, platformType models.PlatformType) (string, error)
```
zaneb, line 29: I don't see much value in using `models.PlatformType` as the type, since nothing is going to check it. Also this code is not actually agent-specific (I'd also like to use this ISO for the bootstrap in baremetal IPI), so using assisted types feels a bit odd. I'd be tempted to move this file to /pkg/coreos.

pkg/asset/agent/image/baseiso.go
```diff
@@ -88,17 +91,39 @@ func getIsoFromReleasePayload() (string, error) {
...
+
+		baseIsoFileName, err := ocRelease.GetBaseIso(log, releaseImage, pullSecret, "X86_64")
+		if err == nil {
+			log.Infof("got base iso image %s using oc command", baseIsoFileName)
+			i.File = &asset.File{Filename: baseIsoFileName}
+			return nil
+		}
+	}
+
+	// Download the Iso since it cannot be retrieved from the release payload
```
zaneb, line 126: I like the idea of falling back to downloading it from the Internet.
I guess this makes it even more important to have a disconnected CI job.

zaneb (commented):
This looks great!

pkg/asset/agent/image/oc.go
bfournie: Agree about the platform type. I'll update the PR and then investigate moving this file.

pkg/asset/agent/image/oc.go
bfournie: Updated to include the architecture in the coreos filename. I removed all the platformType stuff for now, we can add that back if the platform will determine which ISO to get.

pkg/asset/agent/image/oc.go
```diff
@@ -0,0 +1,220 @@
...
+		if err != nil {
+			return "", err
+		}
+		defer removeIcspFile(icspFile)
+		cmd = fmt.Sprintf(templateImageExtractWithIcsp, file, cacheDir, icspFile, image)
+	} else {
+		cmd = fmt.Sprintf(templateImageExtract, file, cacheDir, image)
+	}
+
+	log.Infof("extracting %s to %s, %s", file, cacheDir, cmd)
```
zaneb, line 116: It would be go to use [os/exec.LookPath](https://pkg.go.dev/os/exec@go1.17.12#LookPath) to check whether the `oc` executable exists; if it doesn't then there's no need to try, and we can report the reason.

pkg/asset/agent/image/baseiso.go
```diff
@@ -8,16 +8,24 @@ import (
...
+	"github.com/sirupsen/logrus"
 )
 
 // BaseIso generates the base ISO file for the image
 type BaseIso struct {
 	File *asset.File
 }
 
+const (
+	archName = "x86_64"
```
zaneb, line 26: Worth a TODO.

bfournie: Added

pkg/asset/agent/image/oc.go
bfournie: Added a check and output a warning

pkg/asset/agent/image/baseiso.go
```diff
@@ -88,17 +96,45 @@ func getIsoFromReleasePayload() (string, error) {
...
+		ocRelease := NewRelease(&executer.CommonExecuter{},
+			Config{MaxTries: OcDefaultTries, RetryDelay: OcDefaultRetryDelay})
+
+		log.Info("Extracting base ISO from release payload")
+		baseIsoFileName, err := ocRelease.GetBaseIso(log, releaseImage, pullSecret, registriesConf.MirrorConfig, archName)
+		if err == nil {
+			log.Debugf("retrieved base iso image %s from release payload", baseIsoFileName)
+			i.File = &asset.File{Filename: baseIsoFileName}
+			return nil
+		}
```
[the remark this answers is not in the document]
bfournie: Good point, I've added logging with the failure. Note that its at Info level as we still will attempt to get the iso by download

pkg/asset/agent/image/oc.go
```diff
@@ -113,14 +121,14 @@ func (r *release) extractFileFromImage(log logrus.FieldLogger, image, file, cach
...
-	log.Infof("extracting %s to %s, %s", file, cacheDir, cmd)
+	log.Debugf("extracting %s to %s, %s", file, cacheDir, cmd)
 	_, err := retry.Do(r.config.MaxTries, r.config.RetryDelay, execute, log, r.executer, pullSecret, cmd)
 	if err != nil {
 		return "", err
 	}
 	// set path
 	path := filepath.Join(cacheDir, file)
-	log.Infof("Successfully extracted %s binary from the release to: %s", file, path)
+	log.Infof("Successfully extracted %s ISO from the release to: %s", file, path)
```
zaneb, line 131: Should we be logging the path to the cache dir at INFO level? This seems like more of a debug thing.

pkg/asset/agent/image/baseiso.go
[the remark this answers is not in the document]
zaneb: Thinking along the lines of https://github.com/zaneb/openshift-installer/commit/baseiso-from-release-logging

pkg/asset/agent/image/baseiso.go
```diff
@@ -88,26 +97,58 @@ func getIsoFromReleasePayload() (string, error) {
...
+	baseIsoFileName, err2 := isoGetter.getter()
+	if err2 == nil {
+		log.Debugf("Using base ISO image %s", baseIsoFileName)
+		i.File = &asset.File{Filename: baseIsoFileName}
+		return nil
+	}
+	log.Debugf("Failed to download base ISO: %s", err2)
 
-	return nil
+	return errors.Wrap(err2, "failed to get base ISO image")
```
zaneb, line 151: This was intended to be `err`, not `err2` (see first paragraph of https://github.com/openshift/installer/pull/6154/files#r927768755)

zaneb (commented):
/lgtm

PR #6155, opened 2022-07-21
AGENT-231: agent extra manifests

andfasano:
This patch adds the capability to customize the cluster installation by providing additional manifests in the `/openshift` folder under the assets directory (ZTP manifests generation will follow in a subsequent PR)


pkg/asset/agent/image/ignition.go
```diff
@@ -323,6 +328,13 @@ func addHostConfig(config *igntypes.Config, agentConfig *agentconfig.Asset) erro
 	return nil
 }
 
+func addExtraManifests(config *igntypes.Config, extraManifests *manifests.ExtraManifests) {
```
rwsu, line 331: Would be great to unit test this function.

rwsu (commented):
I added an extra manifest to create a new namespace on the installed cluster. It worked, so this lgtm. I think it is probably already in progress; I suggested adding a unit test for validating the extra manifests are in the ignition.

pkg/asset/agent/image/ignition.go
andfasano: Indeed, the PR is still in WIP, I'm currently working on adding the unit tests (but this would give you a chance to look at the PR before it's ready)

andfasano: Note: I've added a more generic `TestIgnition_Generate` so that we could test properly the asset class just from the public methods (and avoid using the private ones), providing some support so that it could be easily extended in future also to cover the other cases

rwsu (approved):
/approve

pkg/asset/agent/manifests/extramanifests.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"path/filepath"
+
+	"github.com/openshift/installer/pkg/asset"
+	"github.com/pkg/errors"
+)
+
+const (
+	directory = "openshift"
```
zaneb, line 11: This is visible everywhere in the manifests package, so it could use a more specific name.

andfasano: I've reused the same one used elsewhere for consistency, but it seems a valid comment.

zaneb:
test failures look like prow problems
/retest

rwsu (commented):
/lgtm

PR #6157, opened 2022-07-22
Add support for private DNS server to PowerVS

hamzy:
This adds support for a private DNS server to PowerVS by
conditionally creating resources based on:
   publish: Internal
in the install-config.yaml

### DIFF
--- a/data/data/powervs/cluster/dns/dns.tf
+++ b/data/data/powervs/cluster/dns/dns.tf
@@ -20,3 +20,78 @@ resource "ibm_cis_dns_record" "kubernetes_api_internal" {
   content   = var.load_balancer_int_hostname
   ttl       = 60
 }
+
+locals {
+  proxy_count = var.publish_strategy == "Internal" ? 1 : 0
+}
+
+resource "ibm_is_ssh_key" "dns_ssh_key" {
+  count      = local.proxy_count
+  name       = "${var.cluster_id}-dns-ssh-key"
+  public_key = var.ssh_key
+}
+
+resource "ibm_is_security_group" "dns_vm_sg" {
+  count = local.proxy_count
+  name  = "${var.cluster_id}-dns-sg"
+  vpc   = var.vpc_id
+}
+
+# allow all outgoing network traffic
+resource "ibm_is_security_group_rule" "dns_vm_sg_outgoing_all" {
+  count     = local.proxy_count
+  group     = ibm_is_security_group.dns_vm_sg[0].id
+  direction = "outbound"
+  remote    = "0.0.0.0/0"
+}
+
+# allow all incoming network traffic on port 22
+resource "ibm_is_security_group_rule" "dns_vm_sg_ssh_all" {
+  count     = local.proxy_count
+  group     = ibm_is_security_group.dns_vm_sg[0].id
+  direction = "inbound"
+  remote    = "0.0.0.0/0"
+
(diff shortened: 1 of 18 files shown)


data/data/powervs/cluster/dns/dns.tf
```diff
@@ -20,3 +20,87 @@ resource "ibm_cis_dns_record" "kubernetes_api_internal" {
...
+
+# allow all outgoing network traffic
+resource "ibm_is_security_group_rule" "dns_vm_sg_outgoing_all" {
+  count     = local.proxy_count
+  group     = ibm_is_security_group.dns_vm_sg[0].id
+  direction = "outbound"
+  remote    = "0.0.0.0/0"
+}
+
+# allow all incoming network traffic on port 22
```
mjturek, line 48: Will we need SSH for anything other than debug? Might not be worth exposing the port

data/data/powervs/cluster/dns/dns.tf
```diff
@@ -20,3 +20,87 @@ resource "ibm_cis_dns_record" "kubernetes_api_internal" {
...
+    port_max = 22
+  }
+}
+
+#resource "ibm_is_subnet" "dns_vm_subnet" {
+#  name                     = "${var.cluster_id}-dns-subnet"
+#  vpc                      = var.vpc_id
+#  zone                     = var.vpc_zone
+#  total_ipv4_address_count = 256
+#}
```
mjturek, line 66: Can you remove?

mjturek:
Looking good so far! Left some comments. Also I think we should adjust commit message to something like:

"Power VS: Conditionally create VPC VM for proxy DNS"

data/data/powervs/cluster/dns/dns.tf
hamzy: No we don't need SSH for anything other than debug.  Which I am sure will be needed in the future.  Also, all of the master and worker nodes in an OpenShift cluster have ssh enabled.

mjturek: okay we'll leave it open for now

mjturek:
LGTM - can you remove WIP?

data/data/powervs/cluster/dns/dns.tf
hamzy: Done

clnperez:
@rna-afk can we get a review on this one?

data/data/powervs/variables-powervs.tf
```diff
@@ -39,6 +39,16 @@ variable "powervs_cloud_instance_id" {
   description = "The cloud instance ID of your account"
 }
 
+variable "powervs_publish_strategy" {
+  type        = string
+  description = "The cluster publishing strategy, either Internal or External"
+  default     = "External"
+  validation {
```
rna-afk, line 46: Why add validation here? Aren't we checking the value in the install config stage? If we are then this would be another code to maintain if we add more types(though unlikely but possible)

hamzy: Valid point. :) I removed that block

mjturek:
would you be opposed to moving VM creation into the `power_network`  module rather than `dns`?

It would simplify using the VM as a proxy DHCP service by allowing the dhcp service resouce directly read the IP from the proxy VM  after it's created rather than relying on an output

PR #6164, opened 2022-07-25
AGENT-281: Generate the NMStateConfig manifests from the agent-config

pawanpinjarkar:
Using the NMState data provided in the agent-config.yaml, generate the equivalent NMStateConfig manifests in cluster-manifests.


pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -57,14 +60,53 @@ func (*NMStateConfig) Name() string {
...
+	if agentConfig.Config != nil {
+		for _, host := range agentConfig.Config.Spec.Hosts {
+			nmStateConfig := aiv1beta1.NMStateConfig{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "NMStateConfig",
+					APIVersion: "agent-install.openshift.io/v1beta1",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      agentConfig.Config.ObjectMeta.Name,
+					Namespace: agentConfig.Config.ObjectMeta.Namespace,
```
zaneb, line 84: Would be good to user the convenience function that you defined in your previous PR.

pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -57,14 +60,53 @@ func (*NMStateConfig) Name() string {
...
+					Labels: map[string]string{
+						"infraenvs.agent-install.openshift.io": "ostest",
+					},
+				},
+				Spec: aiv1beta1.NMStateConfigSpec{
```
zaneb, line 89: Need to add `Interfaces` inside the Spec here.

pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -57,14 +60,53 @@ func (*NMStateConfig) Name() string {
...
+						Raw: []byte(host.NetworkConfig.Raw),
+					},
+				},
+			}
+			nmStateConfigs = append(nmStateConfigs, &nmStateConfig)
+		}
+
+		n.Config = nmStateConfigs
+		nmStateConfigData, err := k8syaml.Marshal(nmStateConfigs)
+		// os.WriteFile("/home/pawanpinjarkar/go/src/github.com/openshift/installer/nmstategenerate.yaml", nmStateConfigData, 0644)
```
zaneb, line 100: Can we remove this now? :)

pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -14,10 +14,13 @@ import (
 	"github.com/openshift/assisted-service/pkg/staticnetworkconfig"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/validation/field"
-	"k8s.io/apimachinery/pkg/util/yaml"
+	yaml "k8s.io/apimachinery/pkg/util/yaml"
```
zaneb, line 19: This one was already called `yaml` fwiw.

pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -57,14 +60,53 @@ func (*NMStateConfig) Name() string {
...
+					NetConfig: aiv1beta1.NetConfig{
+						Raw: []byte(host.NetworkConfig.Raw),
+					},
+				},
+			}
+			nmStateConfigs = append(nmStateConfigs, &nmStateConfig)
+		}
+
+		n.Config = nmStateConfigs
+		nmStateConfigData, err := k8syaml.Marshal(nmStateConfigs)
```
zaneb, line 99: I assume that k8syaml.Marshal treats the input as a list of separate yaml docs, to be concatenated using `---` separator lines, rather than as a list?

It would be nice to have a test to verify this. The current test data contains only one host, and we don't verify the output at all.

pkg/asset/agent/manifests/nmstateconfig.go
pawanpinjarkar: Yes, sorry :smile:

pkg/asset/agent/manifests/nmstateconfig.go
pawanpinjarkar: Also, I hardcoded the label. Need to fix it.

pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -57,14 +60,53 @@ func (*NMStateConfig) Name() string {
...
+
+	if agentConfig.Config != nil {
+		for _, host := range agentConfig.Config.Spec.Hosts {
+			nmStateConfig := aiv1beta1.NMStateConfig{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "NMStateConfig",
+					APIVersion: "agent-install.openshift.io/v1beta1",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      agentConfig.Config.ObjectMeta.Name,
```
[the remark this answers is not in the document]
pawanpinjarkar: Will use "agent-config-name`-`host.hostname". This should be unique.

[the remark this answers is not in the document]
zaneb: Sounds good, but we will need to substitute an alternative hostname when the hostname is not provided. See e.g. https://github.com/openshift/installer/blob/agent-installer/pkg/asset/agent/agentconfig/agent_config.go#L176-L179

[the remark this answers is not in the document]
pawanpinjarkar: or alternatively instead of hostname, maybe we can append the index of the hosts, e.g. `agent-config-name-0`, `agent-config-name-1`

pkg/asset/agent/manifests/nmstateconfig.go
pawanpinjarkar: Both are addressed.

pkg/asset/agent/manifests/nmstateconfig.go
pawanpinjarkar: I manually verified the file contents generated by `k8syaml.Marshal` with multiple hosts. When saved the generated  nmStateConfigData in a file, YAML validator verified that its a valid YAML although the separator `---` was not added to the file instead the contents were saved as a list.

pawanpinjarkar: Savings the contents as a list needs to be fixed as it causes `create-cluster-and-infraenv.service` to fail with 
```
Jul 26 01:11:46 control-0.example.org podman[2833]: time="2022-07-26T01:11:46Z" 
level=fatal msg="Failed to register infraenv with assisted-service: 
error unmarshalling contents of /manifests/nmstateconfig.yaml: 
error unmarshaling JSON: while decoding JSON: json:
 cannot unmarshal array into Go value of type v1beta1.NMStateConfig"
```
this is because the `/etc/assisted/manifests/nmstateconfig.yaml` file contents are marshalled into a list by `k8syaml.Marshal`

pawanpinjarkar: e.g. 
```
- apiVersion: agent-install.openshift.io/v1beta1
  kind: NMStateConfig
  ...................
  ..................
    - macAddress: 28:d2:44:d2:b2:1a
      name: enp3s1
- apiVersion: agent-install.openshift.io/v1beta1
  kind: NMStateConfig
  ...................
  ..................
```

zaneb: Yeah, we need something that you can `kubectl apply -f`, so

```
apiVersion:  agent-install.openshift.io/v1beta1
kind: NMStateConfig
spec:
    ...
---
apiVersion:  agent-install.openshift.io/v1beta1
kind: NMStateConfig
spec:
    ...
```

pawanpinjarkar: I am fixing the content manually [here](https://github.com/openshift/installer/pull/6164/files#diff-40f56313d8b98bce54d14f601b3096dfef80c415816eaed3dc270d5eee4d3535R104-R114) as I don't see any library that can do it

pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -57,14 +60,66 @@ func (*NMStateConfig) Name() string {
...
+			// Marshal the nmStateConfig one at a time
+			// and add a yaml seperator with new line
+			// so as not to marshal the nmStateConfigs
+			// as a yaml list in the generated nmstateconfig.yaml
+			nmStateConfigData, err := k8syaml.Marshal(nmStateConfig)
+
+			if err != nil {
+				return errors.Wrap(err, "failed to marshal agent installer NMStateConfig")
+			}
+			data = fmt.Sprint(data, fmt.Sprint(string(nmStateConfigData), "---\n"))
```
zaneb, line 113: Another option is to make a slice of marshalled configs and then use `strings.Join(configData, "\n---\n")`

pkg/asset/agent/manifests/nmstateconfig_test.go
```diff
@@ -41,6 +82,16 @@ func TestNMStateConfig_Generate(t *testing.T) {
...
 			} else {
 				assert.NoError(t, err)
+				assert.Equal(t, tc.expectedConfig, asset.Config)
+				assert.NotEmpty(t, asset.Files())
+
+				configFile := asset.Files()[0]
+				assert.Equal(t, "cluster-manifests/nmstateconfig.yaml", configFile.Filename)
+
+				var actualConfig []*aiv1beta1.NMStateConfig
+				err = yaml.Unmarshal(configFile.Data, &actualConfig)
```
zaneb, line 92: I think you'll have to use `strings.Split(configFile.Data, "\n---\n")` to split these.

Alternatively, here is the code we're actually using to read the file in the NMStateConfig asset:
https://github.com/openshift/installer/blob/agent-installer/pkg/asset/agent/manifests/nmstateconfig.go#L217-L245

pkg/asset/agent/manifests/nmstateconfig_test.go
```diff
@@ -21,12 +25,49 @@ func TestNMStateConfig_Generate(t *testing.T) {
 		name           string
 		dependencies   []asset.Asset
 		expectedError  string
-		expectedConfig *aiv1beta1.NMStateConfig
+		expectedConfig []*aiv1beta1.NMStateConfig
 	}{
 		{
-			name:          "missing-config",
+			name: "missing-config",
```
rwsu, line 31: Another test case to include would be a valid AgentConfig, but without networkConfig specified.

pawanpinjarkar: That causes a weird failure for that test 
```
expected: []*v1beta1.NMStateConfig{(*v1beta1.NMStateConfig)(0xc000774000)}
            	            	actual  : []*v1beta1.NMStateConfig{(*v1beta1.NMStateConfig)(0xc0007746e0)}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -44,4 +44,3 @@
            	            	    NetConfig: (v1beta1.NetConfig) {
            	            	-    Raw: (v1beta1.RawNetConfig) {
            	            	-    }
            	            	+    Raw: (v1beta1.RawNetConfig) <nil>
            	            	    }
```
I will add the test in a separate commit to this PR

pawanpinjarkar: @andfasano :point_up:

pawanpinjarkar: Discussed with @rwsu over the slack. When unmarshalled, the RawNetConfig gets set to a byte[] containing an empty interface causing this :point_up:  weird behavior. Removing this scenario.

zaneb (commented):
/approve

pkg/asset/agent/manifests/util_test.go
```diff
@@ -73,3 +77,103 @@ func getValidOptionalInstallConfig() *agent.OptionalInstallConfig {
...
+								Name:       "enp2s0",
+								MacAddress: "98:af:65:a5:8d:01",
+							},
+							{
+								Name:       "enp3s1",
+								MacAddress: "28:d2:44:d2:b2:1a",
+							},
+						},
+						NetworkConfig: v1beta1.NetConfig{
+							Raw: unmarshalJSON([]byte("interfaces:")),
```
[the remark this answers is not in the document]
pawanpinjarkar: No, that results into [this](https://github.com/openshift/installer/pull/6164#discussion_r930253227)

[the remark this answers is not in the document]
zaneb: Couldn't it just be `nil` then?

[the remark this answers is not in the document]
pawanpinjarkar: Nopes , still same issue

PR #6166, opened 2022-07-25
GCP: Add networkProjectID parameter to install-config.

jstuever:
This change adds the networkProjecID parameter to the GCP platform in
the install-config. When set, it also validates the project exists and
that the network and subnetworks belong to this project.

https://issues.redhat.com/browse/CORS-2036

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -2229,6 +2229,11 @@ spec:
                     description: Network specifies an existing VPC where the cluster
                       should be created rather than provisioning a new one.
                     type: string
+                  networkProjectID:
+                    description: NetworkProjectID is currently unsupported. NetworkProjectID
+                      specifies which project the network and subnets exist in when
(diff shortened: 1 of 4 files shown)


rna-afk:
LGTM do we need extra unit test cases for this?

patrickdillon:
LGTM too but holding off due to WIP label

nit: do we want static validation that hostproject != networkProject?

I do have one concern with merging this PR, unrelated to the PR itself. We do not plan on landing this feature in 4.12, but this PR will expose the fields in the install config. What is the best way to handle that (i.e. that we would have valid fields in the API for unstable features)? Potential ideas for handling this:

1. Don't worry about it
2. Slap a validation commit on this that will prevent users from using the feature
3. Use a different version of the install config for unstable features

jstuever:
I've added unit tests for the networkProjectID validation function. I also added wording in `oc explain` to indicate that this parameter is not (yet) supported. Hopefully this satisfies all of the concerns above.

PR #6167, opened 2022-07-26
AGENT-273 Installer command to create agent-config.yaml file

rwsu:
Introduces a new agent sub-command:

openshift-install agent create agent-config-template

The command creates a template of the agent-config.yaml file.
An interactive survey to generate the agent-config.yaml file will
be implemented in a future patch.


pkg/asset/agent/agentconfig/agent_config_test.go
```diff
@@ -12,9 +12,76 @@ import (
...
+					},
+				},
+			},
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+
+			parents := asset.Parents{}
+			asset := &Asset{}
```
lranjbar, line 62: Shouldn't asset be above parents here?

zaneb: The line above is referring to the package `assets`, not this object, which is shadowing the name. (It might pay to rename this to agentConfig though.)

lranjbar: Thanks was a just a quick look while I was making coffee and it looked off to me.

pkg/asset/agent/agentconfig/agent_config.go
```diff
@@ -34,9 +40,49 @@ func (*Asset) Dependencies() []asset.Asset {
...
+			Kind:       "AgentConfig",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "example-agent-config",
+			Namespace: "cluster0",
+		},
+		Spec: agent.Spec{
+			RendezvousIP: "your-node0-ip",
+			Hosts: []agent.Host{
```
pawanpinjarkar, line 56: Isn't `networkConfig` missing here?

pkg/asset/agent/agentconfig/agent_config_test.go
```diff
@@ -13,8 +13,75 @@ import (
...
+							Role:     "master",
+							RootDeviceHints: baremetal.RootDeviceHints{
+								DeviceName: "/dev/sda",
+							},
+							Interfaces: []*aiv1beta1.Interface{
+								{
+									Name:       "your-network-interface-name",
+									MacAddress: "00:00:00:00:00",
+								},
+							},
```
pawanpinjarkar, line 51: `networkConfig` is missing, isn't it?

pkg/asset/agent/agentconfig/agent_config.go
```diff
@@ -34,9 +40,49 @@ func (*Asset) Dependencies() []asset.Asset {
...
+				{
+					Hostname: "change-to-hostname",
```
zaneb, line 58: I have mixed feelings about whether we should fill in dummy values here, or just put empty fields. (The latter may mean we have to generate this as text directly, rather than Marshalling it from a struct.)
Either way, we will have to detect if we got the default config file and skip validation, and also not return any data from it from either `HostConfigFiles()` or any other way this asset's data might be accessed.
The implmentation of the latter might be more defensible if there are no values in the default.

pkg/asset/agent/agentconfig/agent_config.go
```diff
@@ -34,9 +40,49 @@ func (*Asset) Dependencies() []asset.Asset {
...
+							Name:       "your-network-interface-name",
+							MacAddress: "00:00:00:00:00",
+						},
+					},
+				},
+			},
+		},
+	}
+
+	a.Config = agentConfig
```
zaneb, line 74: We probably shouldn't do this. We want to create the file if nothing else is consuming this (i.e. if running `agent create agent-config`), but we don't actually want to pass the contents of the default file on to other assets in the graph.

We'd still have to detect the defaults in `Load()` as well, so that if the user does `agent create agent-config` and then does not customise the file, future steps (like `agent create cluster-manifests`/`image`) won't fail. Or would we?? Maybe it's OK to say that if you explicitly ask for the template for that file, it's an error if you don't replace the bogus values with real ones?

pkg/asset/agent/agentconfig/agent_config.go
zaneb: Yes, but I don't think that field exists until #6164 lands so it's not possible to add it yet.

pkg/asset/agent/agentconfig/agent_config.go
rwsu: Should we make it more obvious that "openshift agent create agent-config" is currently generating a template? Perhaps name it "agent-config-template"? And then change it when the interactive survey is ready.

I think it would be ok if the user realized that they generated a template, that later on when they run "create cluster-manifests" that they may see errors.

pkg/asset/agent/agentconfig/agent_config.go
rwsu: I've went with generate the text directly route. I don't know if we need to detect the template file. When "create cluster-manifests" is run, the installer should print out errors indicating there is an issue in the agent-config.yaml file. The user would correct and retry.

pkg/asset/agent/agentconfig/agent_config.go
```diff
@@ -34,8 +38,70 @@ func (*AgentConfig) Dependencies() []asset.Asset {
...
+  hosts:
+  # If a host is listed, then at least one interface
+  # needs to be specified.
+  - hostname: change-to-hostname
+    role: master
+    # For more information about rootDeviceHints:
+    # https://docs.openshift.com/container-platform/4.10/installing/installing_bare_metal_ipi/ipi-install-installation-workflow.html#root-device-hints_ipi-install-installation-workflow
+    rootDeviceHints:
+	  deviceName: /dev/sda
+    interfaces:
```
zaneb, line 68: Maybe put a comment here to the effect that:

```
# interfaces are used to identify the host to apply this configuration to
```

pkg/asset/agent/agentconfig/agent_config.go
```diff
@@ -34,8 +38,70 @@ func (*AgentConfig) Dependencies() []asset.Asset {
...
+      - macAddress: 00:00:00:00:00:00
+        name: host-network-interface-name
+    # networkConfig contains the NMStateConfig.Spec.Config.interfaces definition for this host.
+    # https://nmstate.io/examples.html
+    # NMStateConfig.Spec.Interfaces section is not included in networkConfig. It is specified
+    # in the above interfaces: node
```
zaneb, line 74: I feel like most people interacting with this file won't be familiar with the NMStateConfig resource at all. I'd be tempted to reduce this to just:

```
# networkConfig contains the network configuration for the host in NMState format.
# See https://nmstate.io/examples.html for examples.
```

pkg/asset/agent/agentconfig/agent_config.go
```diff
@@ -34,8 +38,70 @@ func (*AgentConfig) Dependencies() []asset.Asset {
...
+          mac-address: 00:00:00:00:00:00
+          ipv4:
+            enabled: true
+            address:
+              - ip: 192.168.122.2
+                prefix-length: 23
+            dhcp: false
+`
+
+	a.Template = agentConfigTemplate
```
zaneb, line 89: What happens if the user skips the `agent create agent-config-template` step and goes straight to `agent create image`? Won't this mean that we'll get the template passed to other assets as if it were real?

rwsu: Right now if the user doesn't create the agent-config template, then they see a message saying the nmstate config is missing:

rwsu: Ah, the improved error message will come from https://github.com/openshift/installer/pull/6099

pkg/asset/agent/agentconfig/agent_config.go
```diff
@@ -34,8 +38,69 @@ func (*AgentConfig) Dependencies() []asset.Asset {
...
 }
 
-// Generate generates the Agent manifest.
+// Generate generates the Agent Config manifest.
 func (a *AgentConfig) Generate(dependencies asset.Parents) error {
+
+	// TODO: We are temporarily generating a template of the agent-config.yaml
+	// Change this when its interactive survey is implemented.
+	agentConfigTemplate := `#
+# Note: This is an sample AgentConfig file showing
```
pawanpinjarkar, line 47: Nit: This is `a` sample

pawanpinjarkar (commented):
just a minor typo, otherwise /lgtm

pawanpinjarkar (approved):
/approve

lranjbar:
I was able to create the template with no problems. :+1:

PR #6170, opened 2022-07-26
AGENT-280: Use toml.Unmarshal to parse registries.conf

zaneb:
This means that validation will happen automatically, and it copes with
empty files.


pawanpinjarkar (commented):
/lgtm

PR #6171, opened 2022-07-26
AGENT-299: Set mode 0644 on containers.conf

zaneb:
If the file /etc/containers/containers.conf is not world-readable,
starting the podman API service fails. If nothing else, this leaves a
lot of misleading junk in the journal logs.

In practice, only the agent installation method is populating this file,
but in principle this is the right mode wherever it is used. We are
creating the file to work around a podman configuration bug, so it is
entirely possible that other installation methods will also use it in
the future. Therefore the fix is in the bootstrap ignition population
method.


jstuever (approved):
/approve

PR #6174, opened 2022-07-27
CORS-2167: Remove UPI Metal TF references

barbacbd:
** Removed the Example and terraform data in upi/metal
** Updated the docs in docs/user/metal to reflect the changes above. Pointing the user towards the official docs.


r4f4:
Is there a reason to keep the empty `metal` directory around?

barbacbd:
There is not. I dont know why it remained I deleted it.

barbacbd:
Git does not track directories that are empty. I think this is just a local copy from your system. I cannot delete something from git that git does not think exists at this point.

PR #6179, opened 2022-07-28
Alibaba: Fix security group rule

bd233:
The 'sg_rule_worker_ingress_kubelet_insecure' and 'sg_rule_master_ingress_kubelet_insecure' should restrict the CIDR of vpc, not another security group.


jstuever (commented):
/lgtm

PR #6185, opened 2022-07-29
GCP: Added user specified tags on compute instances in installconfig

AnnaZivkovic:
This is the first step to enabling an IPI cluster using Shared VPC where the service account does not have sufficient permissions to create firewall rules in the Host Project while maintaining cluster-specific firewall rules.


pkg/types/gcp/machinepools.go
```diff
@@ -17,6 +17,11 @@ type MachinePool struct {
 	//
 	// +optional
 	OSDisk `json:"osDisk"`
+
+	// Tags defines the set of character strings to be added to the MachineSet
```
patrickdillon, line 21:
```suggestion
	// Tags defines a set of network tags which will be added to instances in the machineset.
```

pkg/types/gcp/validation/machinepool.go
```diff
@@ -33,6 +34,15 @@ func ValidateMachinePool(platform *gcp.Platform, p *gcp.MachinePool, fldPath *fi
 		}
 	}
 
+	for i, tag := range p.Tags {
+		if tag == "" {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("tags").Index(i), tag, fmt.Sprintf("Tag can not be empty")))
+		} else if !regexp.MustCompile(`^[a-z0-9-]*$`).MatchString(tag) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("tags").Index(i), tag, fmt.Sprintf("Tag can only contain lowercase letters, numbers, and dashes")))
+		} else if len(tag) > 63 {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("tags").Index(i), tag, fmt.Sprintf("Maximum number of characters is 63")))
```
patrickdillon, line 43: error strings shouldn't be capitalized, cf: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

(although note that Log lines are capitalized)

patrickdillon (commented):
This looks good. There are a few small nits. Also it would be good to add some unit tests for the particular validation cases.

jstuever (approved):
/approve

jstuever:
Looks like the verify-codegen is failing.

jstuever:
Just needs to re-generate the installconfig explain bits....

pkg/asset/machines/gcp/machines.go
```diff
@@ -113,7 +113,7 @@ func provider(clusterID string, platform *gcp.Platform, mpool *gcp.MachinePool,
 			Email:  fmt.Sprintf("%s-%s@%s.iam.gserviceaccount.com", clusterID, role[0:1], platform.ProjectID),
 			Scopes: []string{"https://www.googleapis.com/auth/cloud-platform"},
 		}},
-		Tags:        []string{fmt.Sprintf("%s-%s", clusterID, role)},
+		Tags:        append(mpool.Tags, []string{fmt.Sprintf("%s-%s", clusterID, role)}...),
```
patrickdillon, line 116: nit: I think this line is doing too much. Would be better, imho, to move the consolidation of tags out of the struct literal. 

but I don't think this is necessary to change unless you want to or are making other changes.

pkg/types/gcp/validation/machinepool.go
```diff
@@ -33,6 +35,17 @@ func ValidateMachinePool(platform *gcp.Platform, p *gcp.MachinePool, fldPath *fi
 		}
 	}
 
+	for i, tag := range p.Tags {
+		if tag == "" {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("tags").Index(i), tag, fmt.Sprintf("tag can not be empty")))
+		} else if !unicode.IsLetter(rune(tag[0])) || (!unicode.IsLetter(rune(tag[len(tag)-1])) && !unicode.IsNumber(rune(tag[len(tag)-1]))) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("tags").Index(i), tag, fmt.Sprintf("tag can only start with a letter and must end with a letter or a number")))
+		} else if !regexp.MustCompile(`^[a-z0-9-]*$`).MatchString(tag) {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("tags").Index(i), tag, fmt.Sprintf("tag can only contain lowercase letters, numbers, and dashes")))
```
patrickdillon, line 44: Doesn't this case cover the preceding case?  If so, you could just remove the preceding case. If you need to update the error message, it is acceptable to include the regex in the message, if you think that's helpful.

patrickdillon (commented):
this LGTM but a couple of nitpicky comments

pkg/types/gcp/validation/machinepool.go
jstuever: Not exactly the same... as per the GCP limits for tags, this one check all characters for:
"lowercase letters, numbers, dashes"

Whereas the prior one checks for:

"must start with a lowercase letter"
"must end with either a number or a lowercase letter"

While I personally would have ordered these differently, they are logically the same.

https://cloud.google.com/vpc/docs/add-remove-network-tags#restrictions

AnnaZivkovic: Not quite. The previous case deals whit how the tags must start with a letter and must end with either a number or a letter. This case is just checking if all of the letters are lowercase and if the string only contains letters, numbers and -.

patrickdillon: Ah yeah you're right, this regex would allow starting and ending with numbers.

PR #6186, opened 2022-07-30
DO NOT MERGE: Aws no nat gateway

r4f4:
/hold
Just testing this configuration in Prow.


r4f4:
We've gone with a different approach https://github.com/openshift/installer/pull/6342.
/close

PR #6195, opened 2022-08-02
tls: only report error if one happened

r4f4:
Otherwise we'll see the red herring message

time="2022-08-02T14:15:46-04:00" level=debug msg="Failed to marshal PKIX public key: %!s(<nil>)"


barbacbd (commented):
/lgtm

PR #6198, opened 2022-08-03
[WIP] Bug 2096323: add bootstrap configuration to install config

rna-afk:
Adding bootstrap configuration to the install config for the
users to specify the instance type and other parameters that we
allow in the control plane configuration but currently only
using the instance type value during creation.


patrickdillon:
The code looks like a good start, but we shouldn’t consider this a bug (bug should be closed in favor of an rfe). I need to figure out why this slipped through bug triage as it has been opened for almost two months.

PR #6201, opened 2022-08-04
Bug 2114977: Populate staticNetworkConfig in NMStateConfig.Generate

rwsu:
The NMStateConfig.Generate was not populating the staticNetworkConfig
when the asset was being generated from the AgentConfig.

Because the staticNetworkConfig is missing, the Ignition asset does
does not add the /usr/local/bin/pre-network-manager-config.sh script
to the ignition config.

This causes the pre-network-manager-config.service to fail on
node0.


mhanss:
@rwsu changes have been verified. pre-network-manager-config service has started successfully.
/label qe-approved

PR #6202, opened 2022-08-04
Add SPLAT members to platform reviewer/approver aliases

sadasu:
Installer team is happy to collaborate with you!

### DIFF
--- a/OWNERS_ALIASES
+++ b/OWNERS_ALIASES
@@ -44,12 +44,14 @@ aliases:
     - Prashanth684
     - mjturek
   vsphere-approvers:
+    - bostrt
     - dav1x
     - jcpowermac
     - patrickdillon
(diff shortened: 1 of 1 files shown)


OWNERS_ALIASES
```diff
@@ -83,10 +87,12 @@ aliases:
   azure-approvers:
     - fabianofranz
```
patrickdillon, line 88: fabiano could probably be removed here as well, although he is always welcome :)

patrickdillon (commented):
Thanks sandhya!

Not sure if this has already been discussed but `bostrt` could be considered for the vsphere section I believe.

patrickdillon:
/approve

Do we need any more sign off on this or just waiting on tags (lgtm)?

PR #6203, opened 2022-08-04
bug 2055247: pkg/asset/installconfig/azure: standardDDv5Family is not currently supported

jhixson74:
Validate that the the VM family is not standardDDv5Family. This family of
confidential machines will be supported in a future release.

https://bugzilla.redhat.com/show_bug.cgi?id=2055247

### DIFF
--- a/pkg/asset/installconfig/azure/client.go
+++ b/pkg/asset/installconfig/azure/client.go
@@ -26,6 +26,7 @@ type API interface {
 	ListLocations(ctx context.Context) (*[]azsubs.Location, error)
 	GetResourcesProvider(ctx context.Context, resourceProviderNamespace string) (*azres.Provider, error)
 	GetVirtualMachineSku(ctx context.Context, name, region string) (*azsku.ResourceSku, error)
+	GetVirtualMachineFamily(ctx context.Context, name, region string) (string, error)
 	GetDiskSkus(ctx context.Context, region string) ([]azsku.ResourceSku, error)
 	GetGroup(ctx context.Context, groupName string) (*azres.Group, error)
 	ListResourceIDsByGroup(ctx context.Context, groupName string) ([]string, error)
@@ -276,6 +277,22 @@ func (c *Client) GetDiskEncryptionSet(ctx context.Context, subscriptionID, group
 	return &diskEncryptionSet, nil
 }
 
+// GetVirtualMachineFamily retrieves the VM family of an instance type.
+func (c *Client) GetVirtualMachineFamily(ctx context.Context, name, region string) (string, error) {
+	typeMeta, err := c.GetVirtualMachineSku(ctx, name, region)
+	if err != nil {
+		return "", fmt.Errorf("error connecting to Azure client: %v", err)
+	}
+	if typeMeta == nil {
+		return "", fmt.Errorf("not found in region %s", region)
+	}
+	if typeMeta.Family == nil {
+		return "", fmt.Errorf("error getting resource family")
+	}
+
+	return to.String(typeMeta.Family), nil
+}
+
 // GetVMCapabilities retrieves the capabilities of an instant type in a specific region. Returns these values
 // in a map with the capability name as the key and the corresponding value.
 func (c *Client) GetVMCapabilities(ctx context.Context, instanceType, region string) (map[string]string, error) {
@@ -286,11 +303,11 @@ func (c *Client) GetVMCapabilities(ctx context.Context, instanceType, region str
 	if typeMeta == nil {
 		return nil, fmt.Errorf("not found in region %s", region)
 	}
-
 	capabilities := make(map[string]string)
 	for _, capability := range *typeMeta.Capabilities {
 		capabilities[to.String(capability.Name)] = to.String(capability.Value)
 	}
+
 	return capabilities, nil
 }
 
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -155,6 +155,22 @@ func validateMininumRequirements(fieldPath *field.Path, req resourceRequirements
 	return allErrs
 }
 
+func validateFamily(fieldPath *field.Path, instanceType, family string) field.ErrorList {
+	confidentialVMFamilies := sets.NewString(
+		"standardDCASv5Family",
+		"standardDCADSv5Family",
+		"standardECASv5Family",
+		"standardECADSv5Family",
+	)
+	allErrs := field.ErrorList{}
+	if confidentialVMFamilies.Has(family) {
+		errMsg := fmt.Sprintf("%s is not currently supported but will be in a future release", family)
+		allErrs = append(allErrs, field.Invalid(fieldPath, instanceType, errMsg))
+	}
+
+	return allErrs
+}
+
 func validateAcceleratedNetworking(fieldPath *field.Path, vmNetworkingType string, instanceType string, capabilities map[string]string) field.ErrorList {
 	val, ok := capabilities[string(aztypes.AcceleratedNetworkingEnabled)]
 	if ok {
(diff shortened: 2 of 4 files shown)


MayXuQQ (changes requested):
Ref : https://docs.microsoft.com/en-us/azure/confidential-computing/confidential-vm-overview#os-support
Confidential VMs support the following VM sizes:
standardDCASv5Family
standardDCADSv5Family
standardECASv5Family
standardECADSv5Family

pkg/asset/installconfig/azure/client.go
```diff
@@ -286,7 +286,9 @@ func (c *Client) GetVMCapabilities(ctx context.Context, instanceType, region str
 	if typeMeta == nil {
 		return nil, fmt.Errorf("not found in region %s", region)
 	}
-
+	if typeMeta.Family != nil && to.String(typeMeta.Family) == "standardDCADSv5Family" {
+		return nil, fmt.Errorf("standardDCADSv5Family is not currently stupported but will be in a future release")
+	}
```
patrickdillon, line 291: I don't think this check should live here, because this function is just for grabbing the VM capabilities. We don't do any validation in this function, so we should move this validation to be with the rest. If this is the only place we get the `typeMeta`, which I suspect is the case, we can just do some minor refactoring to expose `typeMeta` to the calling validation code.

patrickdillon:
> Ref : https://docs.microsoft.com/en-us/azure/confidential-computing/confidential-vm-overview#os-support Confidential VMs support the following VM sizes: standardDCASv5Family standardDCADSv5Family standardECASv5Family standardECADSv5Family

Thanks MayXu. If we need to validate against all of these families, this is probably a good place to use k8s.io/apimachinery/pkg/util/sets

MayXuQQ:
> > Ref : https://docs.microsoft.com/en-us/azure/confidential-computing/confidential-vm-overview#os-support Confidential VMs support the following VM sizes: standardDCASv5Family standardDCADSv5Family standardECASv5Family standardECADSv5Family
> 
> Thanks MayXu. If we need to validate against all of these families, this is probably a good place to use k8s.io/apimachinery/pkg/util/sets

Fail to find the k8s.io/apimachinery/pkg/util/sets

patrickdillon:
> > > Ref : https://docs.microsoft.com/en-us/azure/confidential-computing/confidential-vm-overview#os-support Confidential VMs support the following VM sizes: standardDCASv5Family standardDCADSv5Family standardECASv5Family standardECADSv5Family
> > 
> > 
> > Thanks MayXu. If we need to validate against all of these families, this is probably a good place to use k8s.io/apimachinery/pkg/util/sets
> 
> Fail to find the k8s.io/apimachinery/pkg/util/sets

This note was intended as a suggestion for @jhixson74 for how to implement your suggestion in code

pkg/asset/installconfig/azure/client.go
jhixson74: I've stuffed the family type into the capabilities.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -136,6 +136,22 @@ func validateMininumRequirements(fieldPath *field.Path, req resourceRequirements
...
+func validateFamily(fieldPath *field.Path, instanceType string, capabilities map[string]string) field.ErrorList {
+	confidentialVMFamilies := sets.NewString(
+		"standardDCASv5Family",
+		"standardDCADSv5Family",
+		"standardECASv5Family",
+		"standardECADSv5Family",
+	)
+	allErrs := field.ErrorList{}
+	if confidentialVMFamilies.Has(capabilities["Family"]) {
+		errMsg := fmt.Sprintf("%s is not currently stupported but will be in a future release", capabilities["Family"])
```
sadasu, line 148: Nit: Typo in error message. Should be "supported".

jhixson74: Fixed.

pkg/asset/installconfig/azure/client.go
```diff
@@ -286,11 +286,15 @@ func (c *Client) GetVMCapabilities(ctx context.Context, instanceType, region str
...
 	}
-
+	if typeMeta.Family == nil {
+		return nil, fmt.Errorf("error getting resource family")
+	}
 	capabilities := make(map[string]string)
 	for _, capability := range *typeMeta.Capabilities {
 		capabilities[to.String(capability.Name)] = to.String(capability.Value)
 	}
+	capabilities["Family"] = to.String(typeMeta.Family)
```
r4f4, line 296: Should we worry that Azure could, in theory, add a capability called `"Family"` and we'd be overriding it here?

r4f4: I guess that won't be a problem unless we need to eventually validate it, by which point we can rename this one.
/lgtm

pkg/asset/installconfig/azure/client.go
```diff
@@ -286,11 +286,15 @@ func (c *Client) GetVMCapabilities(ctx context.Context, instanceType, region str
 	if typeMeta == nil {
 		return nil, fmt.Errorf("not found in region %s", region)
 	}
-
+	if typeMeta.Family == nil {
+		return nil, fmt.Errorf("error getting resource family")
```
patrickdillon, line 290: Should we fail/prevent an install in this case? If we can’t determine the family we don’t know it will fail, so perhaps a warning is better.

pkg/asset/installconfig/azure/client.go
patrickdillon, line 296: Why stuff family into capabilities? Capabilities corresponds to the Azure API and it seems counterintuitive to break with that model here. We see a use case here where we need the instance metadata and it seems likely that there will be future needs so I think it makes sense to expose it somehow

patrickdillon:
If confidential vms are possible in upi we should only run this validation for ipi (ValidateForProvisioning)

pkg/asset/installconfig/azure/client.go
jhixson74: I've rearranged the code so that the error is ignored.

pkg/asset/installconfig/azure/client.go
jhixson74: I've rearranged the code so that there is now a call to get the resource family.

jhixson74: I've removed this and reworked it.

patrickdillon:
/approve
/lgtm cancel

unit tests are failing. will lgtm once those get fixed

pkg/asset/installconfig/azure/validation.go
```diff
@@ -230,6 +246,11 @@ func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceTyp
 	allErrs = append(allErrs, validateMininumRequirements(fieldPath.Child("type"), req, instanceType, capabilities)...)
 	allErrs = append(allErrs, validateVMArchitecture(fieldPath.Child("type"), instanceType, architecture, capabilities)...)
 
+	family, _ := client.GetVirtualMachineFamily(context.TODO(), instanceType, region)
+	if family != "" {
+		allErrs = append(allErrs, validateFamily(fieldPath.Child("type"), instanceType, family)...)
+	}
+
```
r4f4, line 253: Couldn't it be useful here to at least log any errors from `GetVirtualMachineFamily` for debugging?
```suggestion
	family, err := client.GetVirtualMachineFamily(context.TODO(), instanceType, region)
	if family != "" {
		allErrs = append(allErrs, validateFamily(fieldPath.Child("type"), instanceType, family)...)
	}
	if err != nil {
	    logrus.Debugf("failed to retrieve instance family: %v", err)
	}

```

r4f4: Oh my bad. Disregard then.

patrickdillon: My previous suggestion was to log a warning but not throw an error: https://github.com/openshift/installer/pull/6203#discussion_r950123108

But I think existing code is fine.

patrickdillon:
I already approved this. what happened to the label? 

/approve

r4f4:
/override ci/prow/okd-e2e-gcp-ovn-upgrade
Not related to this PR.

PR #6205, opened 2022-08-05
manifests/gcp: Added Network Project ID to the manifest

barbacbd:
** Added the network project ID to match that of
https://github.com/openshift/origin/blob/57b52e0a9528027448422d678c4ef682dfc0b885/vendor/k8s.io/legacy-cloud-providers/gce/gce.go#L136

[CORS-2037](https://issues.redhat.com//browse/CORS-2037)

### DIFF
--- a/pkg/types/powervs/validation/machinepool.go
+++ b/pkg/types/powervs/validation/machinepool.go
@@ -1,9 +1,9 @@
 package validation
 
 import (
-	"math"
 	"regexp"
 	"strconv"
+	"strings"
 
 	"github.com/google/uuid"
 	"github.com/openshift/installer/pkg/types/powervs"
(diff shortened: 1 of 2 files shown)


barbacbd:
requires #6166

pkg/asset/manifests/gcp/cloudproviderconfig.go
```diff
@@ -61,5 +66,6 @@ node-tags       = {{$tag}}
 node-instance-prefix = {{.Global.NodeInstancePrefix}}
 external-instance-groups-prefix = {{.Global.ExternalInstanceGroupsPrefix}}
 subnetwork-name = {{.Global.SubnetworkName}}
+network-project-id = {{.Global.NetworkProjectID}}
```
jstuever, line 69: What are the implications of always passing network-project-id here? Should we put in some logic to only include it when network-project-id is not empty?

pkg/asset/manifests/gcp/cloudproviderconfig.go
```diff
@@ -51,7 +56,7 @@ func CloudProviderConfig(infraID, projectID, subnet string) (string, error) {
 	return buf.String(), nil
 }
 
-var configTmpl = `[global]
+var configTmpl = fmt.Sprintf(`[global]
```
patrickdillon, line 59: why wrap this with `fmt.Sprintf`? It seems like it is not needed.

barbacbd: Oh that is a mistake that snuck back in

patrickdillon:
/approve

LGTM with a successful GCP test run

PR #6206, opened 2022-08-08
Bug 2028474: oVirt: Removed cluster name length limitation

engelmi:
This PR fixes [OCPBUGSM-37760](https://issues.redhat.com/browse/OCPBUGSM-37760) (or [BZ 2028474](https://bugzilla.redhat.com/show_bug.cgi?id=2028474)) by removing the checks for the cluster name length on oVirt platform.

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -88,10 +88,6 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
 	if c.Platform.GCP != nil || c.Platform.Azure != nil {
 		nameErr = validate.ClusterName1035(c.ObjectMeta.Name)
 	}
-	if c.Platform.Ovirt != nil {
-		// FIX-ME: As soon bz#1915122 get resolved remove the limitation of 14 chars for the clustername
-		nameErr = validate.ClusterNameMaxLength(c.ObjectMeta.Name, 14)
-	}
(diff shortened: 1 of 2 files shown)


sadasu:
A quick look at https://github.com/openshift/mdns-publisher/pull/24 suggested maybe we should look for a max length < 64?

pkg/types/validation/installconfig.go
```diff
@@ -89,8 +89,7 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
 		nameErr = validate.ClusterName1035(c.ObjectMeta.Name)
 	}
 	if c.Platform.Ovirt != nil {
-		// FIX-ME: As soon bz#1915122 get resolved remove the limitation of 14 chars for the clustername
-		nameErr = validate.ClusterNameMaxLength(c.ObjectMeta.Name, 14)
+		nameErr = validate.ClusterName(c.ObjectMeta.Name)
 	}
```
r4f4, line 93: At line 87 we already check for `validate.ClusterName`, so this `if` can actually be removed.

r4f4 (changes requested):

pkg/types/validation/installconfig.go
engelmi: You are right, removing it. Thanks!

sadasu:
/retitle Bug 2028474: oVirt: Removed cluster name length limitation

PR #6207, opened 2022-08-08
gcp: Add validation for Network Project Data

barbacbd:
** Added checks in types to ensure that network, computeSubnet, and controlPlaneSubnet
are set when NetworkProjectID exists.
** Added checks in installconfig to ensure that CredentialsMode is set to either Passthrough or
Manual when NetworkProjectID exists.
    
[CORS-2048](https://issues.redhat.com//browse/CORS-2048)
Requires #6166

### DIFF
--- a/pkg/types/gcp/validation/platform.go
+++ b/pkg/types/gcp/validation/platform.go
@@ -7,8 +7,8 @@ import (
 
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
+	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/gcp"
-
 	"github.com/openshift/installer/pkg/validate"
 )
 
@@ -60,7 +60,7 @@ var (
 )
 
 // ValidatePlatform checks that the specified platform is valid.
-func ValidatePlatform(p *gcp.Platform, fldPath *field.Path) field.ErrorList {
+func ValidatePlatform(p *gcp.Platform, fldPath *field.Path, ic *types.InstallConfig) field.ErrorList {
 	allErrs := field.ErrorList{}
 	if p.Region == "" {
 		allErrs = append(allErrs, field.Required(fldPath.Child("region"), "must provide a region"))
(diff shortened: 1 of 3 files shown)


pkg/types/gcp/platform.go
```diff
@@ -37,4 +37,7 @@ type Platform struct {
 	// such as the current env OPENSHIFT_INSTALL_OS_IMAGE_OVERRIDE
 	// +optional
 	Licenses []string `json:"licenses,omitempty"`
+
+	// NetworkProjectID is the id of the pro
```
patrickdillon, line 41: Comment got cut off here.

pkg/types/gcp/validation/platform.go
```diff
@@ -1,6 +1,7 @@
 package validation
 
 import (
+	"github.com/openshift/installer/pkg/types"
```
patrickdillon, line 4: Can we move this to be in the final import group? Those installer imports should be grouped together. Imports look a little strange in this file.

pkg/types/gcp/validation/platform.go
```diff
@@ -69,7 +70,16 @@ func ValidatePlatform(p *gcp.Platform, fldPath *field.Path) field.ErrorList {
...
+	if p.NetworkProjectID != "" {
+		if p.Network == "" {
+			allErrs = append(allErrs, field.Required(fldPath.Child("network"), "must provide a network when a network id is specified"))
+		}
+		if ic.CredentialsMode != types.ManualCredentialsMode && ic.CredentialsMode != types.PassthroughCredentialsMode {
+			allErrs = append(allErrs, field.NotSupported(fldPath.Child("credentialsMode"),
+				ic.CredentialsMode, []string{string(types.ManualCredentialsMode), string(types.PassthroughCredentialsMode)}))
+		}
+	}
+	if p.Network != "" || p.NetworkProjectID != "" {
```
patrickdillon, line 82: I think we can just keep this condition as `p.Network != ""`. If users just specified p.NetworkProjectID, I think it's ok that these error messages would get skipped. I also suggest an extra line also:
```suggestion
	
	if p.Network != ""  {
```

pkg/types/gcp/validation/platform_test.go
```diff
@@ -1,6 +1,7 @@
 package validation
 
 import (
+	"github.com/openshift/installer/pkg/types"
```
patrickdillon, line 4: nit: can you move this to the final import block?

Imports should be

<stdlib>

<external>

<internal>

pkg/types/gcp/validation/platform.go
barbacbd: Absolutely, looks like goland automatic import resolution has got me again!

PR #6209, opened 2022-08-08
Power VS: Add Montreal as a supported region

mjturek:
Montreal was excluded initilally because it lacked an associated IBM Cloud region. We can instead map it to Toronto.

### DIFF
--- a/pkg/types/powervs/powervs_regions.go
+++ b/pkg/types/powervs/powervs_regions.go
@@ -38,6 +38,11 @@ var Regions = map[string]Region{
 			"lon06",
 		},
 	},
+	"mon": {
+		Description: "Montreal, Canada",
+		VPCRegion:   "ca-tor",
+		Zones:       []string{"mon01"},
+	},
 	"osa": {
 		Description: "Osaka, Japan",
 		VPCRegion:   "jp-osa",


clnperez (approved):
/lgtm

pkg/types/powervs/powervs_regions.go
```diff
@@ -38,6 +38,11 @@ var Regions = map[string]Region{
 			"lon06",
 		},
 	},
+	"tor": {
+		Description: "Montreal, Canada",
+		VPCRegion:   "ca-tor",
+		Zones:       []string{"mon01"},
+	},
```
Prashanth684, line 45: should this be `mon` ? there is already a `tor`

mjturek: whoops, yep you're right. Fixed

pkg/types/powervs/powervs_regions.go
```diff
@@ -38,6 +38,11 @@ var Regions = map[string]Region{
 			"lon06",
 		},
 	},
+	"mon": {
+		Description: "Montreal, Canada",
+		VPCRegion:   "ca-tor",
```
Prashanth684, line 43: i remember there being a `ca-mon` region but we don't push the rhcos vhds to that region. is `ca-tor` the correct vpc region ?

mjturek: We are only aware of `ca-tor` being available in IBM Cloud. I just took a quick look at the list of regions available and still do not see ca-mon being an option. The intent is to choose a data center that is close to Montreal as the default for mon01.

PR #6210, opened 2022-08-09
terraform: ibmcloud: bump provider version

clnperez:
to get `network_name` from ibm_pi_dhcps for the Power VS provider

Signed-off-by: Christy Norman <christy@linux.vnet.ibm.com>

PR #6217, opened 2022-08-10
powervs: allow VPC, Cloud connection, and NW re-use

clnperez:
because Power VS has some networking limitations, we have been restricted to one cluster/zone for IPI on Power VS. this PR allows re-using an existing VPC, Cloud connection (the GRE tunnel wrapper that allows us to communicate with IBM Cloud components from the Power VS DCs), an a Power Network. It should be noted that the intended use case is that these are from previously-installed OpenShift clusters, not created by hand.


clnperez:
Relies on PR https://github.com/openshift/installer/pull/6210

clnperez:
reconciling some doc/code comment descriptions and fixed the tf formatting. i think this is good to go for review. @r4f4 anyone else i need to get to look at it on the installer team?

clnperez:
weirdness with prow...

/retest-required

r4f4:
> i think this is good to go for review.

Do you mind if we stick a `/hold` on this PR to make it clear that #6210 has to merge first? `lgtm`s would be lost after the rebase anyway.

> @r4f4 anyone else i need to get to look at it on the installer team?

I'd say @rna-afk if he has some free cycles this sprint, given his background of reviewing PowerVS PRs.

clnperez:
@bkhadars @miyamotoh would appreciate a review from you both since this touches code you've submitted or modified in the past

miyamotoh:
Proposed changes make sense, and no obvious red flags I see by just looking at the diffs. So, LGTM, as long as it's tested working.

clnperez:
lots of `Pod pending timeout`s here. Hopefully those are fixed by now.

clnperez:
@r4f4 this seems stuck. any ideas?

r4f4:
> @r4f4 this seems stuck. any ideas?

It went through now. I guess it had to do with the Prow builder upgrade.

clnperez:
thanks @r4f4 

@rna-afk can we get this one in?

data/data/powervs/cluster/power_network/variables.tf
```diff
@@ -17,3 +17,15 @@ variable "vpc_crn" {
   type        = string
   description = "The CRN of the IBM Cloud VPC."
 }
+
+variable "pvs_network_name" {
+  type        = string
+  description = "The name of a pre-created Power VS DHCP Network."
+  default     = ""
```
rna-afk, line 24: lets remove the default fields if there is nothing to set.

clnperez: maybe this is a terraform knoweldge gap for me, but if these arent set to an empty string, they're null, and i thought i'd get a type mismatch later when trying to check if it equals `nil` or not. i do have to check those to set some resource `count`s later on

PR #6219, opened 2022-08-11
GCP: Remove firewall rules when providing network project id

barbacbd:
** Added the Network Project ID data to gcp terraform vars
** Added checks to GCP terraform to set count to 0 for firewall rules when
the Network Project ID is available.

CORS 2039
Requires #6166

### DIFF
--- a/data/data/gcp/variables-gcp.tf
+++ b/data/data/gcp/variables-gcp.tf
@@ -3,6 +3,12 @@ variable "gcp_project_id" {
   description = "The target GCP project for the cluster."
 }
 
+variable "gcp_network_project_id" {
+  type        = string
+  description = "The project that the network and subnets exist in when they are not in the main ProjectID."
+  default     = null
+}
+
 variable "gcp_service_account" {
   type        = string
   description = "The service account for authenticating with GCP APIs."
--- a/data/data/gcp/variables-gcp.tf
+++ b/data/data/gcp/variables-gcp.tf
@@ -6,7 +6,7 @@ variable "gcp_project_id" {
 variable "gcp_network_project_id" {
   type        = string
   description = "The project that the network and subnets exist in when they are not in the main ProjectID."
-  default     = null
+  default     = ""
 }
 
 variable "gcp_service_account" {
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -379,8 +379,9 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			return err
 		}
 		auth := gcptfvars.Auth{
-			ProjectID:      installConfig.Config.GCP.ProjectID,
-			ServiceAccount: string(sess.Credentials.JSON),
+			ProjectID:        installConfig.Config.GCP.ProjectID,
+			NetworkProjectID: installConfig.Config.GCP.NetworkProjectID,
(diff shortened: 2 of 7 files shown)


data/data/gcp/variables-gcp.tf
```diff
@@ -3,6 +3,12 @@ variable "gcp_project_id" {
   description = "The target GCP project for the cluster."
 }
 
+variable "gcp_network_project_id" {
+  type        = string
+  description = "The project that the network and subnets exist in when they are not in the main ProjectID."
+  default     = null
```
r4f4, line 9: You check for the empty string `""` when deciding whether to create a resource or not. Shouldn't this be 
```suggestion
  default     = ""
```

barbacbd: O believe that the value coming from install config would be an empty string. That may be the one that I need to change if any. The default here follows the convention used in the tf file.

data/data/gcp/variables-gcp.tf
```diff
@@ -3,6 +3,12 @@ variable "gcp_project_id" {
   description = "The target GCP project for the cluster."
 }
 
+variable "gcp_network_project_id" {
+  type        = string
+  description = "The project that the network and subnets exist in when they are not in the main ProjectID."
+  default     = ""
```
r4f4, line 9: I was able to deploy a cluster with `default = ""` instead of `default = null`.

pkg/tfvars/gcp/gcp.go
```diff
@@ -15,8 +15,9 @@ const (
 
 // Auth is the collection of credentials that will be used by terrform.
 type Auth struct {
-	ProjectID      string `json:"gcp_project_id,omitempty"`
-	ServiceAccount string `json:"gcp_service_account,omitempty"`
+	ProjectID        string `json:"gcp_project_id,omitempty"`
+	NetworkProjectID string `json:"gcp_network_project_id,omitempty"`
```
jstuever, line 19: This field doesn't really belong in the `Auth` structure.... I recommend placing it into the `config` structure below.

pkg/asset/cluster/tfvars.go
```diff
@@ -379,8 +379,9 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			return err
 		}
 		auth := gcptfvars.Auth{
-			ProjectID:      installConfig.Config.GCP.ProjectID,
-			ServiceAccount: string(sess.Credentials.JSON),
+			ProjectID:        installConfig.Config.GCP.ProjectID,
+			NetworkProjectID: installConfig.Config.GCP.NetworkProjectID,
```
jstuever, line 383: See my comment in `pkg/tfvars/gcp/gcp.go` about moving this out of the `auth` structure.

jstuever (changes requested):
Some minor changes as per my comments.

While making these changes, you might consider passing some sort of "disable firewall" parameter instead of relying on the "networkProjectID" directly in the Terraform templates. This will enable us to provision firewall rules in the future when networkProjectID is defined. However, I would not block this PR on that as enabling this functionality is technically out of scope.

pkg/tfvars/gcp/gcp.go
jstuever: I'm actually on the fence; I don't believe it's worth blocking on this.

r4f4 (approved):

PR #6221, opened 2022-08-11
AGENT-310: Check that cached base ISO matches hash in release payload

bfournie:
For the base ISO that is cached, it needs to be verified that it matches the one in the release payload. This change extracts the
coreos-x86-64.iso.sha256 file from the release payload and compares it to the hash of the cached file.

Note that this adds execution time to retrieve the sha256 file, but its less time than it would be to retrieve the actual iso from the
release payload.


rwsu (commented):
/lgtm

celebdor:
/title OCPBUGS-816: Check that cached base ISO matches hash in release payload

PR #6222, opened 2022-08-11
Bug 2098252: locations fix

RyanJenkins99:
Inside the rhcos-stream.json file the locations needed to be updated to the most current version


mike-nguyen:
All required tests have passed.

@patrickdillon can we get a backport-risk-assessed and approve label on this?  The only change here is the redirector to the RHCOS artifacts.

sdodson:
I've confirmed that the docs for 4.8 mention the new endpoint so moving this on.

PR #6223, opened 2022-08-11
Bug 2117687: compute.replicas must be 0 for SNO

pawanpinjarkar:
When controlPlane.replicas = 1 and compute.replicas is missing from install-config.yaml,  validation warns users that compute.replicas needs to be set to 0 when controlPlane.replicas = 1 for SNO cluster on none platform.

validation error: `invalid install-config configuration: Compute.Replicas: Required value: Installing a Single Node Openshift requires explicitly setting compute replicas to zero`

https://bugzilla.redhat.com/show_bug.cgi?id=2117687

### DIFF
--- a/pkg/asset/agent/installconfig.go
+++ b/pkg/asset/agent/installconfig.go
@@ -159,6 +159,9 @@ func (a *OptionalInstallConfig) validateSNOConfiguration(installConfig *types.In
 		if *installConfig.ControlPlane.Replicas != 1 {
 			fieldPath = field.NewPath("ControlPlane", "Replicas")
 			allErrs = append(allErrs, field.Required(fieldPath, fmt.Sprintf("control plane replicas must be 1 for %s platform. Found %v", none.Name, *installConfig.ControlPlane.Replicas)))
+		} else if len(installConfig.Compute) == 0 {
+			fieldPath = field.NewPath("Compute", "Replicas")
+			allErrs = append(allErrs, field.Required(fieldPath, "Installing a Single Node Openshift requires explicitly setting compute replicas to zero"))
 		}
 
 		var workers int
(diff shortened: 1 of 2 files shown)


rwsu (commented):
/lgtm

rwsu (approved):

pkg/asset/agent/installconfig.go
```diff
@@ -159,6 +159,9 @@ func (a *OptionalInstallConfig) validateSNOConfiguration(installConfig *types.In
 		if *installConfig.ControlPlane.Replicas != 1 {
 			fieldPath = field.NewPath("ControlPlane", "Replicas")
 			allErrs = append(allErrs, field.Required(fieldPath, fmt.Sprintf("control plane replicas must be 1 for %s platform. Found %v", none.Name, *installConfig.ControlPlane.Replicas)))
+		} else if len(installConfig.Compute) == 0 {
```
zaneb, line 162: Shouldn't this be `!= 0`?

zaneb: Sorry, I missed the `len()`.

Given that the user's intention is clear, shouldn't we just default this to a config section with no replicas rather than throw an error at them?

Where is it currently being defaulted to a higher number of compute replicas if no explicit config is provided?

andfasano: IINW the default for the worker is set [here](https://github.com/openshift/installer/blob/e1959c2931b2cee35c789958cc60989e10ee285d/pkg/types/defaults/installconfig.go#L74-L79).

Personally I'd like the defaulting if the install-config would have a field to specify explicitly the desired configuration (ie, `platformConfig: SingleNode`) - without it, the error message seems a less risky option.

zaneb:
After discussion, my position is that we are already [setting defaults for the Compute pools](https://github.com/openshift/installer/blob/e1959c2931b2cee35c789958cc60989e10ee285d/pkg/types/defaults/installconfig.go#L74-L79), we're just setting them on the assumption of an HA cluster. If we know that the cluster will be SNO, we should set different defaults (i.e. Compute replicas = 0 if the user does not specify anything).

That said, this patch is still an improvement on the status quo, so let's merge it and work on setting defaults separately.
/lgtm

PR #6227, opened 2022-08-15
[WIP] add bind config for PowerVS publish==internal IPI

miyamotoh:
This PR is to add `bind` config (`/etc/named.conf`) for forwarding DNS queries to the private DNS service when `publish==Internal` for IPI in PowerVS. This also assumes the "proxy VM" is created and SSH access enabled via #6157.

Essentially, this adds a .sed file to edit the default `named.conf` file with necessary changes so the `bind` listens to the right ports and forwards queries to the private DNS service where the custom zone and entries will be configured at a later point. The name servers' addresses are hardcoded as publicly documented at: https://cloud.ibm.com/docs/dns-svcs?topic=dns-svcs-getting-started#step-3-add-vpc-as-permitted-network-to-dns-zone


miyamotoh:
@rna-afk could you review/approve?

mjturek:
@jstuever could you take a look?

PR #6228, opened 2022-08-15
  Power VS: Enable IBM DNS service for private cluster support

mjturek:
If creating a public cluster, continue to use CIS DNS. If creating a private cluster, create DNS records in IBM DNS Services. This is controlled by the "PublishStrategy" install config element.


mjturek:
Requires #6227 and #6157

mjturek:
@clnperez @hamzy @miyamotoh @cs-zhang @23TNC

PTAL

miyamotoh:
The `count` control in TF with the hardcoded `Internal` looks a bit tedious, but I'm not sure if there's a standard/better way. LGTM.

mjturek:
> The `count` control in TF with the hardcoded `Internal` looks a bit tedious, but I'm not sure if there's a standard/better way. LGTM.

@miyamotoh yeah agreed. I could have a module for private and public DNS that would remove a lot of the `count` controls but want to consider doing that later since I know this currently works

pkg/asset/cluster/tfvars.go
```diff
@@ -748,10 +748,19 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
-		if err != nil {
-			return err
+		var cisCRN, dnsCRN string
+		if installConfig.Config.Publish == types.InternalPublishingStrategy {
+			// Get DNSInstanceCRN from InstallConfig metadata
+			dnsCRN, err = installConfig.PowerVS.DNSInstanceCRN(ctx)
+			if err != nil {
+				return err
+			}
+		} else {
```
clnperez, line 758: i'm not sure, but is there a possiblity that this could be set to anything other than one of the two valid publishing types? or that it could be null? wondering if instead of juts if/else it should be a case statement with a default that's an error

pkg/asset/installconfig/powervs/client.go
```diff
@@ -138,12 +201,16 @@ func (c *Client) GetDNSZoneIDByName(ctx context.Context, name string) (string, e
...
 
 // GetDNSZones returns all of the active DNS zones managed by CIS.
-func (c *Client) GetDNSZones(ctx context.Context) ([]DNSZoneResponse, error) {
+func (c *Client) GetDNSZones(ctx context.Context, publish types.PublishingStrategy) ([]DNSZoneResponse, error) {
 	_, cancel := context.WithTimeout(ctx, 1*time.Minute)
 	defer cancel()
 
 	options := c.controllerAPI.NewListResourceInstancesOptions()
-	options.SetResourceID(cisServiceID)
+	if publish == types.ExternalPublishingStrategy {
```
clnperez, line 209: same thing here (could this be someting else or nil?)

pkg/asset/installconfig/powervs/dns.go
```diff
@@ -27,21 +28,20 @@ func GetDNSZone() (*Zone, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
 	defer cancel()
 
-	publicZones, err := client.GetDNSZones(ctx)
+	// Default to offering only external DNS entries, as IBM Cloud does in their provider
+	// TODO(mjturek): Offer private zones (IBM DNS) when deploying a private cluster
```
clnperez, line 32: it might be good to make this a task for a future release vs a todo

pkg/asset/manifests/infrastructure.go
```diff
@@ -226,14 +226,24 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
...
-		if err != nil {
-			return errors.Wrapf(err, "failed to get instance CRN")
+		var cisInstanceCRN, dnsInstanceCRN string
+		var err error
+		if installConfig.Config.Publish == types.InternalPublishingStrategy {
+			dnsInstanceCRN, err = installConfig.PowerVS.DNSInstanceCRN(context.TODO())
+			if err != nil {
+				return errors.Wrapf(err, "failed to get instance CRN")
+			}
+		} else {
```
clnperez, line 236: same :D

pkg/destroy/powervs/dns.go
```diff
@@ -5,12 +5,16 @@ import (
...
+	"github.com/IBM-Cloud/bluemix-go/crn"
 	"github.com/IBM/go-sdk-core/v4/core"
-	"github.com/IBM/networking-go-sdk/dnsrecordsv1"
+	"github.com/IBM/networking-go-sdk/resourcerecordsv1"
 	"github.com/pkg/errors"
 )
 
-const dnsRecordTypeName = "dns record"
+const (
+	dnsRecordTypeName      = "cis dns record"
```
clnperez, line 15: these name types seem like one is dns and one is generic. i got a little confused when they were used in the case statement later. is there a way they could be a bit more specific and clear that one is CIS DNS and the other is IBM DNS?

clnperez (commented):
mostly lgtm. just a few little things :D

pkg/destroy/powervs/dns.go
mjturek: Definitely can see the confusion. I used "resource" because that is what they're referred to as in the `networking-go-sdk`.

pkg/asset/installconfig/powervs/dns.go
mjturek: Agreed. Should it be in the 4.13 epic?

pkg/asset/installconfig/powervs/client.go
mjturek: Will change.

pkg/asset/cluster/tfvars.go
mjturek: A third publishing strategy may be introduced one day, so will adjust

pkg/asset/installconfig/powervs/client.go
mjturek: Done

pkg/asset/manifests/infrastructure.go
mjturek: Done.

mjturek:
Thanks for the review @clnperez, took your advice in all comments

pkg/asset/installconfig/powervs/dns.go
clnperez: yah. not sure if we have one yet though

mjturek:
@rna-afk Could you take a look? We think this is ready to merge

mjturek:
@rna-afk fixed the unit test issues. Please take a look.

PR #6232, opened 2022-08-16
[CFE-465] Add additionalTrustBundlePolicy field to allow CA bundle propagation

TrilokGeer:
A new field _additionalTrustBundlePolicy_ is added to support transparent proxy scenario. 
_additionalTrustBundlePolicy_ defines conditional policy to add AdditionalTrustBundle to the nodes' trusted certificate store. Following are the valid values :
_default_ : use AdditionalTrustBundle if there is http/https proxy configured
_always_ : ignore all conditions and use AdditionalTrustBundle

### DIFF
--- a/pkg/types/installconfig.go
+++ b/pkg/types/installconfig.go
@@ -70,6 +70,17 @@ const (
 	InternalPublishingStrategy PublishingStrategy = "Internal"
 )
 
+// PolicyType is for usage polices that are applied to additionalTrustBundle.
+// +kubebuilder:validation:Enum="";Proxyonly;Always
+type PolicyType string
+
+const (
+	// PolicyProxyOnly  enables use of AdditionalTrustBundle when http/https proxy is configured.
+	PolicyProxyOnly PolicyType = "Proxyonly"
+	// PolicyAlways ignores all conditions and uses AdditionalTrustBundle.
+	PolicyAlways PolicyType = "Always"
+)
+
 //go:generate go run ../../vendor/sigs.k8s.io/controller-tools/cmd/controller-gen crd:crdVersions=v1 paths=. output:dir=../../data/data/
 
 // InstallConfig is the configuration for an OpenShift install.
@@ -84,6 +95,13 @@ type InstallConfig struct {
 	//
 	// +optional
 	AdditionalTrustBundle string `json:"additionalTrustBundle,omitempty"`
+
+	// AdditionalTrustBundlePolicy determines when to add the AdditionalTrustBundle
+	// to the nodes' trusted certificate store. "Proxyonly" is the default.
+	// The field can be set to following specified values.
+	// "Proxyonly" : adds the AdditionalTrustBundle to nodes when http/https proxy is configured.
+	// "Always" : always adds AdditionalTrustBundle.
+	AdditionalTrustBundlePolicy PolicyType `json:"additionalTrustBundlePolicy,omitempty"`
 
 	// SSHKey is the public Secure Shell (SSH) key to provide access to instances.
 	// +optional
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -81,6 +81,11 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
 			allErrs = append(allErrs, field.Invalid(field.NewPath("additionalTrustBundle"), c.AdditionalTrustBundle, err.Error()))
 		}
 	}
+	if c.AdditionalTrustBundlePolicy != "" {
+		if err := validateAdditionalCABundlePolicy(c); err != nil {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("additionalTrustBundlePolicy"), c.AdditionalTrustBundlePolicy, err.Error()))
+		}
+	}
 	nameErr := validate.ClusterName(c.ObjectMeta.Name)
 	if c.Platform.GCP != nil || c.Platform.Azure != nil {
 		nameErr = validate.ClusterName1035(c.ObjectMeta.Name)
@@ -741,6 +746,7 @@ func validateProxy(p *types.Proxy, c *types.InstallConfig, fldPath *field.Path) 
 	if p.HTTPProxy == "" && p.HTTPSProxy == "" {
 		allErrs = append(allErrs, field.Required(fldPath, "must include httpProxy or httpsProxy"))
 	}
+
 	if p.HTTPProxy != "" {
 		allErrs = append(allErrs, validateURI(p.HTTPProxy, fldPath.Child("httpProxy"), []string{"http"})...)
 		if c.Networking != nil {
@@ -960,4 +966,13 @@ func validateCapabilities(c *types.Capabilities, fldPath *field.Path) field.Erro
 		}
 	}
 	return allErrs
+}
+
+func validateAdditionalCABundlePolicy(c *types.InstallConfig) error {
+	switch c.AdditionalTrustBundlePolicy {
+	case types.PolicyProxyOnly, types.PolicyAlways:
+		return nil
+	default:
+		return fmt.Errorf("supported values \"Proxyonly\", \"Always\"")
+	}
 }
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -25,6 +25,17 @@ spec:
             description: AdditionalTrustBundle is a PEM-encoded X.509 certificate
               bundle that will be added to the nodes' trusted certificate store.
             type: string
+          additionalTrustBundlePolicy:
+            description: 'AdditionalTrustBundlePolicy determines when to add the AdditionalTrustBundle
+              to the nodes'' trusted certificate store. "Proxyonly" is the default.
+              The field can be set to following specified values. "Proxyonly" : adds
+              the AdditionalTrustBundle to nodes when http/https proxy is configured.
+              "Always" : always adds AdditionalTrustBundle.'
+            enum:
+            - ""
+            - Proxyonly
+            - Always
+            type: string
           apiVersion:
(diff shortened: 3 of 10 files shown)


pkg/types/validation/installconfig.go
```diff
@@ -752,3 +759,12 @@ func validateCapabilities(c *types.Capabilities, fldPath *field.Path) field.Erro
 	}
 	return allErrs
 }
+
+func validateAdditionalCABundlePolicy(c *types.InstallConfig) error {
+	switch strings.ToLower(c.AdditionalTrustBundlePolicy) {
+	case types.PolicyDefault, types.PolicyAlways:
+		return nil
+	default:
+		return fmt.Errorf("AdditionalTrustBundlePolicy must have valid policy defined")
```
barbacbd, line 768: nit: errors generally are formatted so that they begin with lower case letters.

pkg/types/installconfig.go
```diff
@@ -85,6 +90,12 @@ type InstallConfig struct {
 	// +optional
 	AdditionalTrustBundle string `json:"additionalTrustBundle,omitempty"`
 
+	// AdditionalTrustBundlePolicy defines conditional policy to add AdditionalTrustBundle
+	// to the nodes' trusted certificate store.
+	// default : use AdditionalTrustBundle if there is http/https proxy configured
+	// always : ignore all conditions and use AdditionalTrustBundle
+	AdditionalTrustBundlePolicy string `json:"additionalTrustBundlePolicy,omitempty"`
```
barbacbd, line 97: The code gen is failing. Please run hack/verify-codegen.sh

pkg/types/installconfig.go
```diff
@@ -85,6 +90,13 @@ type InstallConfig struct {
 	// +optional
 	AdditionalTrustBundle string `json:"additionalTrustBundle,omitempty"`
 
+	// AdditionalTrustBundlePolicy defines conditional policy to add AdditionalTrustBundle
+	// to the nodes' trusted certificate store. If omitted, "Default" is considered.
+	// The field can be set to following specified values.
+	// "Default" : uses AdditionalTrustBundle when http/https proxy is configured.
```
patrickdillon, line 96: I don't think `Default` is a good value here. Perhaps something like `Proxy` or `ProxyOnly` and then we declare that this value is the default.

pkg/types/installconfig.go
```diff
@@ -85,6 +90,13 @@ type InstallConfig struct {
 	// +optional
 	AdditionalTrustBundle string `json:"additionalTrustBundle,omitempty"`
 
+	// AdditionalTrustBundlePolicy defines conditional policy to add AdditionalTrustBundle
```
barbacbd, line 93: Can/should we move this to a boolean value? There are only two possible values that you have defined "default" which is technically not an additional trust bundle or "always". If you think that more values are possible then a string would definitely be correct but right not it seems like its True or False, right?

jhixson74:
Other than what @patrickdillon stated, this looks good to me.

pkg/types/installconfig.go
TrilokGeer: Yes makes sense. Updated the _Default_ to _ProxyOnly_. Thanks for the review.

pkg/types/installconfig.go
TrilokGeer: Thanks for the review, @barbacbd . The boolean does present a good option if the field is bound with strict yes/no meaning. The additionalTrustBundle is not strictly bound to only proxy, I think, it is better to have usage policy represented as explicit string for extendability and to describe well on the usage of the additionalTrustBundle.

patrickdillon: > Can/should we move this to a boolean value?

I agree with the question, but I have learned we try to avoid booleans in the API: c.f. https://github.com/openshift/api/pull/1057#discussion_r750562936

pkg/types/validation/installconfig.go
```diff
@@ -84,6 +84,11 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
 			allErrs = append(allErrs, field.Invalid(field.NewPath("additionalTrustBundle"), c.AdditionalTrustBundle, err.Error()))
 		}
 	}
+	if c.AdditionalTrustBundlePolicy != "" {
+		if err := validateAdditionalCABundlePolicy(c); err != nil {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("AdditionalTrustBundlePolicy"), c.AdditionalTrustBundlePolicy, err.Error()))
```
patrickdillon, line 89: nit:
```suggestion
			allErrs = append(allErrs, field.Invalid(field.NewPath("additionalTrustBundlePolicy"), c.AdditionalTrustBundlePolicy, err.Error()))
```

pkg/types/validation/installconfig.go
```diff
@@ -529,8 +534,10 @@ func validatePlatform(platform *types.Platform, fldPath *field.Path, network *ty
 func validateProxy(p *types.Proxy, c *types.InstallConfig, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
-	if p.HTTPProxy == "" && p.HTTPSProxy == "" {
-		allErrs = append(allErrs, field.Required(fldPath, "must include httpProxy or httpsProxy"))
+	if c.AdditionalTrustBundlePolicy != types.PolicyAlways {
+		if p.HTTPProxy == "" && p.HTTPSProxy == "" {
+			allErrs = append(allErrs, field.Required(fldPath, "must include httpProxy or httpsProxy"))
+		}
```
patrickdillon, line 540: I am confused by this check. Why does the additionalTrustBundlePolicy affect this requirement? What case is this protecting? To me it seems like we don't need an update here, but perhaps I'm missing something.

pkg/types/installconfig.go
```diff
@@ -85,6 +92,13 @@ type InstallConfig struct {
...
+	// to the nodes' trusted certificate store. If omitted, "ProxyOnly" is considered.
+	// The field can be set to following specified values.
+	// "ProxyOnly" : uses AdditionalTrustBundle when http/https proxy is configured.
+	// "Always" : ignores all conditions and uses AdditionalTrustBundle.
```
patrickdillon, line 99:
```suggestion
	// AdditionalTrustBundlePolicy determines when to add the AdditionalTrustBundle
	// to the nodes' trusted certificate store. "ProxyOnly" is the default.
	// The field can be set to following specified values.
	// "ProxyOnly" : adds the AdditionalTrustBundle to nodes when http/https proxy is configured.
	// "Always" : always adds AdditionalTrustBundle.
```

Suggesting some updated wording. 

Also, we are saying these are adds to the "nodes." Is that correct? I'm not sure how this works exactly. I thought it was added as a cluster object.

pkg/types/installconfig.go
```diff
@@ -85,6 +92,13 @@ type InstallConfig struct {
...
+	AdditionalTrustBundlePolicy string `json:"additionalTrustBundlePolicy,omitempty"`
```
patrickdillon, line 100: If we are going to set this as an enum, this should be typed similar to:

https://github.com/openshift/installer/blob/d10d64c9843e37f500bcb8d66caf339e940b918f/pkg/types/azure/platform.go#L93-L110

patrickdillon (commented):
This looks good, I think we just need to iron out some details.

The default is not being explicitly set, and I think that would be preferable. You can set a default here:
https://github.com/openshift/installer/blob/master/pkg/types/defaults/installconfig.go

pkg/types/installconfig.go
TrilokGeer: "Adds to nodes" is to say how the bundle is used. The configuration is added to cluster object.

pkg/types/validation/installconfig.go
TrilokGeer: In case of PolicyAlways, proxy configuration is not a mandatory field. For mitm proxy usecase, there might be no proxy details provided with additionalTrustBundle.

patrickdillon: This code is only executed when proxy is supplied:
https://github.com/openshift/installer/blob/4d3796fa29b7aa7767376718b318cec893ace4ba/pkg/types/validation/installconfig.go#L122-L124

This function just makes sure that when a proxy is supplied it is valid, the trust bundle policy shouldn't affect that. We shouldn't need any updates to this function. It seems like the current code would break the existing validation in the case of ProxyOnly

patrickdillon:
Also, it would be good to update the in-repo docs: https://github.com/openshift/installer/blob/master/docs/user/customization.md

pkg/types/validation/installconfig.go
TrilokGeer: User can provide proxy details with PolicyAlways configuration. As the policies only control the propagation, proxy details with PolicyAlways configuration will still be valid.

patrickdillon: Oops, my previous comment had a typo. it should have said "It seems like the current code would break the existing validation in the case of PolicyAlways". 

> User can provide proxy details with PolicyAlways configuration. As the policies only control the propagation, proxy details with PolicyAlways configuration will still be valid.

Right, but your code is saying we have different proxy validation rules based on the policy, which isn't true. Specifically your code says:  when `PolicyAlways`, it is ok to have `noProxy` without `httpProxy` or `httpsProxy`, but when `ProxyOnly` one of `httpProxy` or `httpsProxy` is required (when `noProxy` is set). 

If you removed your change, the logic would be correct: if `noProxy` is set, one of `httpProxy` or `httpsProxy` is required, regardless of policy.

TrilokGeer: Yes, you are right. The condition check is redundant and adds to confusing dependence on policy. Thanks :+1:

patrickdillon:
/approve

Can you squash the commits? (More than one commit is fine, but it seems like you have a bunch of fixups that should be squashed).

PR #6233, opened 2022-08-16
Bug 2098072: vsphere: fix default disk type when not speficied

r4f4:
During the terraform decoupling work [1], a fix for the default value
for the disk provisioning type [2] ended up being lost.

This commit bring those changes back, fixing a regression where the
default disk type would be set to "thick" instead of the default storage
policy of vsphere.

[1] https://github.com/openshift/installer/commit/9c77d6a437bcbf7289dc5affb8809bf3d17e018d
[2] https://github.com/openshift/installer/commit/f3c9bfade850662ab74ae4cc62a226d75da43819


patrickdillon:
/approve
/test e2e-vsphere

@rvanderp3 @jcpowermac PTAL. Wanted to make sure you're aware of this regression.

r4f4:
/test okd-unit
/override e2e-libvirt since it's failing after tests, when fetching results.

PR #6235, opened 2022-08-17
[OCPNODE-1258] Add imageDigestSources, deprecate imageContentSources

QiWang19:
The ImageDigestMirrorSet will be the new CRD that is used to configure a cluster-wide mirror, thus the imageContentSources of install-config will need to migrate to use this new CRD instead of ImageContentSourcePolicy.

It has been [discussed](https://coreos.slack.com/archives/C68TNFWA2/p1659533284752489) the install-config will add imageDigestSources, deprecate imageContentSources. The  
imageDigestSources configuration will generate imagediestmirrorset manifest

- deprecation process:
1) introduce new field
2) mark old field deprecated
3) make it an error to populate both
4) provide a warning if the old field is populated

- issue link: https://issues.redhat.com/browse/OCPNODE-1498, epic: https://issues.redhat.com/browse/OCPNODE-1258
- imagediestmirrorset epic: https://issues.redhat.com/browse/OCPNODE-521
- ImageDigestMirrorSet implemented: https://github.com/openshift/machine-config-operator/pull/3037
- ref discussion: https://coreos.slack.com/archives/C68TNFWA2/p1659533284752489

To verify this patch:
1. add below `imageDigestSources` to install-config.yaml
```
imageDigestSources:
- mirrors:
  - registry.access.redhat.com/ubi8/ubi-minimal
  source: registry.access.redhat.com/ubi8/ubi-minimal-1
```
2. launch the cluster using the installer built by this patch: `$ OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=quay.io/openshift-release-dev/ocp-release:4.13.0-ec.3-x86_64 ./openshift-install create cluster --dir install-dir/`
3. The ImageDigestMirrorSet is created by bootstrap 
```
$ oc get idms
NAME                     AGE
image-digest-mirror-0   91m
[qiwan@qiwan installer-cfg]$ oc describe idms/image--digest-mirror-0
Name:         image-digest-mirror-0
Namespace:    
Labels:       <none>
Annotations:  <none>
API Version:  config.openshift.io/v1
Kind:         ImageDigestMirrorSet
Metadata:
  Creation Timestamp:  2023-02-15T20:28:04Z
  Generation:          1
  Managed Fields:
    API Version:  config.openshift.io/v1
    Fields Type:  FieldsV1
    fieldsV1:
      f:spec:
        .:
        f:imageDigestMirrors:
    Manager:      cluster-bootstrap
    Operation:    Update
    Time:         2023-02-15T20:28:04Z
    API Version:  config.openshift.io/v1
    Fields Type:  FieldsV1
    fieldsV1:
      f:status:
    Manager:         cluster-bootstrap
    Operation:       Update
    Subresource:     status
    Time:            2023-02-15T20:28:04Z
  Resource Version:  1236
  UID:               b8933438-1a03-4735-964e-3413f0a44d4a
Spec:
  Image Digest Mirrors:
    Mirrors:
      registry.access.redhat.com/ubi8/ubi-minimal
    Source:  registry.access.redhat.com/ubi8/ubi-minimal-1
Status:
Events:  <none>

```
4. If set ImageContentSources in the install-config, it still works.
```
$ oc get imagecontentsourcepolicy
NAME             AGE
image-policy-0   23m

```
5. populate error if set both ImageContentSources and ImageDigestSources in the install-config.yaml
```
$ OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE=quay.io/openshift-release-dev/ocp-release:4.13.0-ec.3-x86_64 ./openshift-install create install-config --dir install-dir
WARNING ImageContentSource is deprecated, please use ImageDigestSource 
ERROR failed to fetch Install Config: failed to load asset "Install Config": failed to create install config: invalid "install-config.yaml" file: imageContentSources: Invalid value: "External": cannot set imageContentSources and imageDigestSources at the same time 

```
Signed-off-by: Qi Wang <qiwan@redhat.com>


QiWang19:
@patrickdillon @bparees We have discussed about this change https://redhat-internal.slack.com/archives/C68TNFWA2/p1659624385641969?thread_ts=1659533284.752489&cid=C68TNFWA2, could you review it?  I am working in to get this in 4.14.

pkg/types/validation/installconfig.go
```diff
@@ -131,6 +131,16 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
 	if _, ok := validPublishingStrategies[c.Publish]; !ok {
 		allErrs = append(allErrs, field.NotSupported(field.NewPath("publish"), c.Publish, validPublishingStrategyValues))
 	}
+	allErrs = append(allErrs, validateImageDigestSources(c.ImageDigestSources, field.NewPath("imageDigestSources"))...)
+	if _, ok := validPublishingStrategies[c.Publish]; !ok {
+		allErrs = append(allErrs, field.NotSupported(field.NewPath("publish"), c.Publish, validPublishingStrategyValues))
+	}
+	if len(c.ImageContentSources) != 0 {
+		logrus.Warningln("ImageContentSource is deprecated, please use ImageDigestSource")
```
bparees, line 139: I wouldn't print this warning when you're already going to error out because they supplied both fields.

If they supply both, just print the error.

QiWang19: fixed

QiWang19:
@bparees The branch is on 4.14 development phase. Could you review and help push this forward?

pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -139,14 +139,16 @@ func (i *RegistriesConf) Generate(dependencies asset.Parents) error {
...
+	if !installConfig.Supplied || len(installConfig.Config.ImageContentSources) == 0 && len(installConfig.Config.ImageDigestSources) == 0 {
 		return i.generateDefaultRegistriesConf()
 	}
 
 	registries := &sysregistriesv2.V2RegistriesConf{
 		Registries: []sysregistriesv2.Registry{},
 	}
-	for _, group := range bootstrap.MergedMirrorSets(installConfig.Config.ImageContentSources) {
+	digestMirrorSources := bootstrap.ContentSourceToDigestMirror(installConfig.Config.ImageContentSources)
+	digestMirrorSources = append(digestMirrorSources, installConfig.Config.ImageDigestSources...)
```
bparees, line 150: below you have validation logic that rejects installconfigs that set both fields, so why here do you have logic that merges the two fields?

is it to be treated as invalid, or not?

QiWang19: @bparees  It is invalid. In practice, it should fail before this line or one of the fields should be empty.

bparees: if it is invalid, then why does the code here treat it as valid?  i'd expect this code to throw an error if we manage to get here with both fields populated (which, as you say, we should not).

QiWang19:
/hold cancel 
I assume the 4.14 tree is open

bparees:
> I assume the 4.14 tree is open

yes, master is open for 4.14 development.

pkg/asset/agent/mirror/registriesconf.go
QiWang19: I thought the `ValidateInstallConfig` is called before this. I updated the validation code here anyway.

QiWang19:
@bparees cleaned up the reviews. Could you review it?

pkg/asset/ignition/bootstrap/common.go
```diff
@@ -248,7 +248,9 @@ func (a *Common) getTemplateData(dependencies asset.Parents, bootstrapInPlace bo
 	}
 
 	registries := []sysregistriesv2.Registry{}
-	for _, group := range MergedMirrorSets(installConfig.Config.ImageContentSources) {
+	digestMirrorSources := ContentSourceToDigestMirror(installConfig.Config.ImageContentSources)
+	digestMirrorSources = append(digestMirrorSources, installConfig.Config.ImageDigestSources...)
```
bparees, line 252: this is still merging content instead of erroring?

pkg/asset/manifests/imagedigestmirrorset.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Dependencies returns all of the dependencies directly needed to generate
+// the asset.
+func (*ImageDigestMirrorSet) Dependencies() []asset.Asset {
+	return []asset.Asset{
+		&installconfig.InstallConfig{},
+	}
+}
+
+// Generate generates the ImageDigestMirrorSet config and its CRD.
```
bparees, line 38:
```suggestion
// Generate generates the ImageDigestMirrorSet config and its CR.
```

?  you're generating a resource instance, not a resource definition, right?

bparees (commented):
@patrickdillon ultimately i expect your team will be the one to lgtm/approve this PR, so heads up

pkg/asset/ignition/bootstrap/common.go
QiWang19: fixed

bparees:
all my comments are addressed, final review/approval should come from the install team

patrickdillon:
> @patrickdillon ultimately i expect your team will be the one to lgtm/approve this PR, so heads up

ACK. Thanks Qi and Ben for getting this into such good shape. I will take a look at this withn a day or two.

cc @sadasu

patrickdillon (commented):
This looks really good to me. Just one small nit about some extra unneeded validation.

/approve

go.mod
```diff
@@ -31,6 +31,7 @@ require (
 	github.com/daixiang0/gci v0.9.0
 	github.com/diskfs/go-diskfs v1.2.1-0.20210727185522-a769efacd235
 	github.com/form3tech-oss/jwt-go v3.2.3+incompatible
+	github.com/ghodss/yaml v1.0.0
```
r4f4, line 34: This package doesn't seem to be mainted anymore. Can you replace it with `sigs.k8s.io/yaml`? See https://github.com/openshift/installer/pull/6935

pkg/types/validation/installconfig.go
```diff
@@ -131,6 +133,15 @@ func ValidateInstallConfig(c *types.InstallConfig, usingAgentMethod bool) field.
 	if _, ok := validPublishingStrategies[c.Publish]; !ok {
 		allErrs = append(allErrs, field.NotSupported(field.NewPath("publish"), c.Publish, validPublishingStrategyValues))
 	}
+	allErrs = append(allErrs, validateImageDigestSources(c.ImageDigestSources, field.NewPath("imageDigestSources"))...)
+	if _, ok := validPublishingStrategies[c.Publish]; !ok {
+		allErrs = append(allErrs, field.NotSupported(field.NewPath("publish"), c.Publish, validPublishingStrategyValues))
+	}
+	if len(c.ImageContentSources) != 0 && len(c.ImageDigestSources) != 0 {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("imageContentSources"), c.Publish, "cannot set imageContentSources and imageDigestSources at the same time"))
+	} else if len(c.ImageDigestSources) != 0 {
```
bfournie, line 142: Shouldn't this check be for  `len(c.ImageContentSources) != 0` not `c.ImageDigestSources` since the warning is indicating that ImageContentSources are being used

pkg/types/validation/installconfig.go
```diff
@@ -131,6 +133,15 @@ func ValidateInstallConfig(c *types.InstallConfig, usingAgentMethod bool) field.
...
+		logrus.Warningln("ImageContentSource is deprecated, please use ImageDigestSource")
```
bfournie, line 143: Warning text should use` Sources` not `Source` to match the field name

pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -163,7 +174,7 @@ func (i *RegistriesConf) Generate(dependencies asset.Parents) error {
 	i.setMirrorConfig(i.Config)
 
 	if !i.releaseImageIsSameInRegistriesConf(releaseImage.PullSpec) {
-		logrus.Warnf(fmt.Sprintf("The ImageContentSources configuration in install-config.yaml should have at least one source field matching the releaseImage value %s", releaseImage.PullSpec))
+		logrus.Warnf(fmt.Sprintf("The ImageContentSources or ImageDigestMirrorSet configuration in install-config.yaml should have at least one source field matching the releaseImage value %s", releaseImage.PullSpec))
```
zaneb, line 177: Should be `imageDigestSources` as that's the field name in install-config.
I understand why ImageContentSources is still mentioned here, but if it is deprecated then I'd be tempted to stop encouraging users to put more things in it.

pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -139,14 +140,24 @@ func (i *RegistriesConf) Generate(dependencies asset.Parents) error {
...
+		return fmt.Errorf("invalid install-config.yaml, cannot set imageContentSources and imageDigestSources at the same time")
+	}
+
 	registries := &sysregistriesv2.V2RegistriesConf{
 		Registries: []sysregistriesv2.Registry{},
 	}
-	for _, group := range bootstrap.MergedMirrorSets(installConfig.Config.ImageContentSources) {
+	digestMirrorSources := []types.ImageDigestSource{}
+	if len(installConfig.Config.ImageContentSources) > 0 {
+		digestMirrorSources = bootstrap.ContentSourceToDigestMirror(installConfig.Config.ImageContentSources)
```
zaneb, line 156: If this conversion were done in `ConvertInstallConfig()`, it would result in the cluster always getting an `ImageDigestMirrorSet` CR instead of an `ImageContentSourcePolicy` regardless of which field the user specified in the install-config. The decision to convert only here is presumably to avoid that.

In the agent installer, assisted-service generates a (second) install-config and it currently sets the `imageContentSources` field. This conversion means that we have lost any information as to whether the user specified `imageContentSources` or `imageDigestSources`, which I think is probably OK (on line 167 we set `MirrorByDigestOnly = true` anyway, so effectively we have always required digest-only mirrors in the agent installer). However, it will take an additional change to assisted-service to switch it to always setting `imageDigestSources` for `ImageDigestMirrorSet` CRs to start showing up in clusters deployed with the agent installer. Without that, it will always create `ImageContentSourcePolicy` CRs even if the user specifies `imageDigestSources` in the install-config.

That does make me wonder whether we could afford to always create ImageDigestMirrorSet` CRs on every type of deployment regardless of which field the user specifies (which wouldn't need a change in assisted-service). But I'm sure you've already spent a lot more time thinking about it than I.

pkg/types/validation/installconfig.go
```diff
@@ -131,6 +133,15 @@ func ValidateInstallConfig(c *types.InstallConfig, usingAgentMethod bool) field.
 	if _, ok := validPublishingStrategies[c.Publish]; !ok {
 		allErrs = append(allErrs, field.NotSupported(field.NewPath("publish"), c.Publish, validPublishingStrategyValues))
 	}
+	allErrs = append(allErrs, validateImageDigestSources(c.ImageDigestSources, field.NewPath("imageDigestSources"))...)
+	if _, ok := validPublishingStrategies[c.Publish]; !ok {
+		allErrs = append(allErrs, field.NotSupported(field.NewPath("publish"), c.Publish, validPublishingStrategyValues))
+	}
+	if len(c.ImageContentSources) != 0 && len(c.ImageDigestSources) != 0 {
```
andfasano, line 140: Minor point, but I'd nice to have all the validations related to the ImageDigestSource contained within one function.

pkg/types/validation/installconfig.go
andfasano: Also, the else could not be required (ie, always inform the user of the deprecation)

pkg/asset/agent/mirror/registriesconf.go
andfasano, line 1: (commenting here just because it's a file withing `pkg/asset`)

Shouldn't be changed also https://github.com/openshift/installer/blob/b2c4cb7bc5c166773ec423b57f1c2529c538fea5/pkg/asset/cluster/tfvars.go#L863?

bfournie:
I tested this PR with the agent installer in dev-scripts in a disconnected env and it installs fine with the existing `imageContentSources`.
I also tried it using `imageDigestSources` in install-config and that installs fine too but the `ImageDigestMirrorSet` isn't set in the installed cluster since assisted-service is still using `imageContentSources` as Zane indicated. Assisted would need to eventually be updated unless the installer does the translation.
```
[stack@host06 dev-scripts-latest]$ oc get idms
No resources found
```

andfasano:
cc @avishayt

pkg/asset/agent/mirror/registriesconf.go
QiWang19: I take a rough look at the assisted-service,  if we always create ImageDigestMirrorSet, the logic to generate install-config can stay the same, only the code writes the `ImageContentSourcePolicy` manifest needs to update.

@bparees @patrickdillon What do you think about only creating `ImageDigestMirrorSet` CRs on every type of deployment regardless of which field the user specifies? If we do it this way, adding the new field seems not necessary.

avishayt: FYI I opened [MGMT-14159](https://issues.redhat.com/browse/MGMT-14159) in case any adjustments need to be made in assisted installer.

pkg/asset/agent/mirror/registriesconf.go
QiWang19: changed this to `len(installConfig.Config.DeprecatedImageContentSources) == 0 && len(installConfig.Config.ImageDigestSources) == 0`

go.mod
QiWang19: fixed

pkg/types/validation/installconfig.go
QiWang19: Keep it as is. From the review https://github.com/openshift/installer/pull/6235#discussion_r1150947827 we only have to do the validation once in installconfig.go. 
I removed duplicates of this code from other files.

pkg/asset/ignition/bootstrap/common.go
```diff
@@ -247,8 +247,18 @@ func (a *Common) getTemplateData(dependencies asset.Parents, bootstrapInPlace bo
 		etcdEndpoints[i] = fmt.Sprintf("https://etcd-%d.%s:2379", i, installConfig.Config.ClusterDomain())
 	}
 
+	if len(installConfig.Config.ImageContentSources) != 0 && len(installConfig.Config.ImageDigestSources) != 0 {
+		logrus.Error("invalid install-config.yaml, cannot set imageContentSources and imageDigestSources at the same time")
+	}
```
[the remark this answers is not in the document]
QiWang19: fixed.

pkg/asset/agent/mirror/registriesconf.go
QiWang19: @avishayt thanks. I linked that card to the epic that tracks a set of ICSP migrations.

QiWang19: Does the assisted-service generate the install-config usong data from this registries.conf? So assisted service can not know if the mirror registry is from `ImageContentSources` or `ImageDigestSources` of the first install-config?

avishayt: @bfournie implemented most of the code in assisted-service that deals with ImageContentSources.  Bob, can you comment?

bfournie: @avishayt yes, I will make any changes to the assisted-service for this. However it looks like with the most recent change to use `ContentSourceToDigestMirror` that the ImageContentSources are converted to ImageDigestSources by the installer so no changes would be required in assisted-service

QiWang19:
@patrickdillon @r4f4 @andfasano @zaneb @bfournie Thanks for the review. I cleaned up the reviews. Could you take another round of review?

pkg/asset/agent/mirror/registriesconf.go
QiWang19: @bfournie  That convert function can ImageContent -> ImageDigest, and the converted contented will be written to `RegistriesConf` struct. Does this help? https://github.com/openshift/installer/pull/6235/files#diff-beb52b8fa67370e9332766c49cad67ee11cfe31bd15fdbbc7bf258bfdeadf632R156
or you mean the converted data should change the first install-config.yaml passed to the installer?

go.mod
r4f4: Thanks! It looks good now.

pkg/types/installconfig.go
```diff
@@ -138,8 +138,13 @@ type InstallConfig struct {
 	Proxy *Proxy `json:"proxy,omitempty"`
 
 	// ImageContentSources lists sources/repositories for the release-image content.
+	// The field is deprecated. Please use imagedigestsources.
```
zaneb, line 141: s/imagedigestsources/imageDigestSources/

pkg/asset/agent/mirror/registriesconf.go
zaneb: @bfournie as far as I can see it is still only translating at the point they're used to create the ignition file; the manifests are still generated based on the field actually set in the install-config.
To change that, the translation would have to happen in 'pkg/types/conversion/install-config.go`, so that everything sees the install-config as if the user had specified the new fields.

pkg/types/installconfig.go
QiWang19: fixed

pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -163,7 +174,7 @@ func (i *RegistriesConf) Generate(dependencies asset.Parents) error {
 	i.setMirrorConfig(i.Config)
 
 	if !i.releaseImageIsSameInRegistriesConf(releaseImage.PullSpec) {
-		logrus.Warnf(fmt.Sprintf("The ImageContentSources configuration in install-config.yaml should have at least one source field matching the releaseImage value %s", releaseImage.PullSpec))
+		logrus.Warnf(fmt.Sprintf("The imageDigestSources configuration in install-config.yaml should have at least one source field matching the releaseImage value %s", releaseImage.PullSpec))
```
bfournie, line 177: Changes look good. Just one question - previously you had this warning referring to `ImageContentSources or ImageDigestMirrorSet`
You changed it to be just `imageDigestSources` but it would be more accurate to also include `imageContentSources` so the warning is accurate for existing configurations.

QiWang19: oh, that's suggested by https://github.com/openshift/installer/pull/6235#discussion_r1151234220 to stop encouraging users to use ImageContentSources.

bfournie: Ah OK thanks.

PR #6236, opened 2022-08-17
Bug 2115770: Check for RendezvousIP in Agent config if NMStateConfig is not provided

lranjbar:
Digging around I found this function already implemented so I decided to use it.

### DIFF
--- a/pkg/agent/rest.go
+++ b/pkg/agent/rest.go
@@ -9,6 +9,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
+	"github.com/openshift/assisted-service/api/v1beta1"
 	"github.com/openshift/assisted-service/client"
 	"github.com/openshift/assisted-service/client/events"
 	"github.com/openshift/assisted-service/client/installer"
@@ -18,6 +19,7 @@ import (
 	"github.com/openshift/installer/pkg/asset/agent/image"
 	"github.com/openshift/installer/pkg/asset/agent/manifests"
 	assetstore "github.com/openshift/installer/pkg/asset/store"
+	"github.com/openshift/installer/pkg/types/agent"
 )
 
 // NodeZeroRestClient is a struct to interact with the Agent Rest API that is on node zero.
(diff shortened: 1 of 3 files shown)


pkg/agent/rest.go
```diff
@@ -30,31 +30,26 @@ type NodeZeroRestClient struct {
...
-	}
-	nmState := &manifests.NMStateConfig{}
-	if err := assetStore.Fetch(nmState); err != nil {
-		return nil, errors.Wrapf(err, "failed to fetch %s", nmState.Name())
-	}
+	agentManifests := &manifests.AgentManifests{}
+	agentConfigAsset := &agentconfig.AgentConfig{}
 
-	NodeZeroIP, err := manifests.GetNodeZeroIP(nmState.Config)
+	RendevousIP, err := image.RetrieveRendezvousIP(agentConfigAsset.Config, agentManifests.NMStateConfigs)
```
pawanpinjarkar, line 36: typo `rendezvous`

pawanpinjarkar: and other places once its corrected

pkg/agent/rest.go
```diff
@@ -30,31 +30,26 @@ type NodeZeroRestClient struct {
...
-	assetStore, err := assetstore.NewStore(assetDir)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to create asset store")
-	}
-	nmState := &manifests.NMStateConfig{}
-	if err := assetStore.Fetch(nmState); err != nil {
-		return nil, errors.Wrapf(err, "failed to fetch %s", nmState.Name())
-	}
+	agentManifests := &manifests.AgentManifests{}
+	agentConfigAsset := &agentconfig.AgentConfig{}
```
zaneb, line 34: These are just going to be empty structures unless we call `assetStore.Fetch()` to load them I think.

pkg/agent/rest.go
```diff
@@ -28,33 +30,40 @@ type NodeZeroRestClient struct {
...
 
 	assetStore, err := assetstore.NewStore(assetDir)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to create asset store")
 	}
-	nmState := &manifests.NMStateConfig{}
-	if err := assetStore.Fetch(nmState); err != nil {
-		return nil, errors.Wrapf(err, "failed to fetch %s", nmState.Name())
+
+	if agentConfigError := assetStore.Fetch(agentConfigAsset); agentConfigError != nil {
```
zaneb, line 43: It looks to me like `Load()` might be closer to what we want here:
https://github.com/lranjbar/installer/blob/63c6e934ee6e33a4f2a5c63665a959aab58ca2a0/pkg/asset/store.go#L18-L20

zaneb (commented):
/approve

pkg/agent/rest.go
```diff
@@ -34,27 +36,38 @@ func NewNodeZeroRestClient(ctx context.Context, assetDir string) (*NodeZeroRestC
...
 	}
-	nmState := &manifests.NMStateConfig{}
-	if err := assetStore.Fetch(nmState); err != nil {
-		return nil, errors.Wrapf(err, "failed to fetch %s", nmState.Name())
+
+	agentConfig, agentConfigError := assetStore.Load(&agentconfig.AgentConfig{})
+	agentManifests, manifestError := assetStore.Load(&manifests.AgentManifests{})
+
+	if agentConfigError != nil {
+		logrus.Debug(errors.Wrapf(agentConfigError, "failed to load %s", agentConfig.(*agentconfig.AgentConfig).Name()))
```
zaneb, line 44: If there's a non-nil error then agentConfig [will be nil](https://github.com/lranjbar/installer/blob/63c6e934ee6e33a4f2a5c63665a959aab58ca2a0/pkg/asset/store/store.go#L366). I think that should work, but since `Name()` doesn't depend on any of the data in the struct, it might be cleaner to just do `&agentconfig.AgentConfig{}.Name()`.

pkg/agent/rest.go
```diff
@@ -34,27 +36,38 @@ func NewNodeZeroRestClient(ctx context.Context, assetDir string) (*NodeZeroRestC
...
+	}
+	if manifestError != nil {
+		logrus.Debug(errors.Wrapf(manifestError, "failed to load %s", agentManifests.(*manifests.AgentManifests).Name()))
 	}
 
-	NodeZeroIP, err := manifests.GetNodeZeroIP(nmState.Config)
+	if agentConfigError != nil && manifestError != nil {
```
zaneb, line 50: This should be `||` I think.

pkg/agent/rest.go
```diff
@@ -34,27 +36,38 @@ func NewNodeZeroRestClient(ctx context.Context, assetDir string) (*NodeZeroRestC
...
+		return nil, errors.New("failed to load AgentConfig and NMStateConfig")
+	}
+
+	RendezvousIP, err := image.RetrieveRendezvousIP(agentConfig.(*agentconfig.AgentConfig).Config, agentManifests.(*manifests.AgentManifests).NMStateConfigs)
```
zaneb, line 54: We should check that agentConfig and agentManifests [are non-nil](https://github.com/lranjbar/installer/blob/63c6e934ee6e33a4f2a5c63665a959aab58ca2a0/pkg/asset/store/store.go#L370) before dereferencing them here.

zaneb (commented):
Looks good, just some error handling to tidy up.

pkg/agent/rest.go
lranjbar: We only need either the agent config or the NM state config but if we don't have either of them we won't have the IP address

pkg/agent/rest.go
andfasano, line 54: Just as a reminder note (not for this PR), I think in future it could be useful (at least for the maintenance) to deploy a (non-writable) RendezvousIP asset (with a dependency on AgentConfig and AgentManifests - if not directly NMStateConfig), to be used directly here and in the Ignition asset

lranjbar: I like this idea. After going through this exercise I think we can consolidate the logic for the RendezvousIP into it's own asset that will clean it up a lot for the future.

pkg/agent/rest.go
zaneb: Not sure I buy that. We only need one, but we don't know which one until we have them both.
This should never fail so any error here is an error I think.

lranjbar: I have tested this with only the ZTP manifests and it works as expected. Fails to load the non-existent Agent Config.

pawanpinjarkar: Can we add unit tests for this function to cover the error scenarios?

PR #6237, opened 2022-08-17
Collapse AgentConfig spec and add versioning

zaneb:
The InstallConfig schema does not have a Spec subresource, so
configuration items are added directly at the top level of the file. It
is desirable that the AgentConfig should behave the same. There is no
Status subresource for an AgentConfig either, and in fact not even any
plans to ever install AgentConfig into clusters as a CRD, so there is no
need for a Spec.

Set the initial version of AgentConfig to v1alpha1, since we are not yet
committing to not changing it. This will now be required.

This is a breaking change that will require users to change their
existing agent-config.yaml files, so better that it happen now.


pawanpinjarkar (approved):
/lgtm

PR #6239, opened 2022-08-18
AGENT-246: Temporarily require at least one NMStateConfig to be specified

zaneb:
Numerous places assume that there will be at least one NMStateConfig:

* https://bugzilla.redhat.com/show_bug.cgi?id=2116489
* https://bugzilla.redhat.com/show_bug.cgi?id=2115770
* https://bugzilla.redhat.com/show_bug.cgi?id=2115798
* https://bugzilla.redhat.com/show_bug.cgi?id=2115803
* https://bugzilla.redhat.com/show_bug.cgi?id=2117302

Since these are not easily discoverable by users, temporarily reinstate
the requirement that at least one NMState configuration be provided
(even if the RendezvousIP is specified). This will provide users with an
upfront error rather than a difficult-to-debug mystery.

### DIFF
--- a/pkg/asset/agent/image/ignition.go
+++ b/pkg/asset/agent/image/ignition.go
@@ -120,6 +120,10 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
 		},
 	}
 
+	if len(agentManifests.NMStateConfigs) == 0 {
+		return errors.New("at least one NMState configuration must be provided")
+	}
+
 	nodeZeroIP, err := retrieveRendezvousIP(agentConfigAsset.Config, agentManifests.NMStateConfigs)
 	if err != nil {
 		return err
--- a/pkg/asset/agent/image/ignition_test.go
+++ b/pkg/asset/agent/image/ignition_test.go
@@ -12,6 +12,7 @@ import (
 
 	hiveext "github.com/openshift/assisted-service/api/hiveextension/v1beta1"
 	"github.com/openshift/assisted-service/api/v1beta1"
+	aiv1beta1 "github.com/openshift/assisted-service/api/v1beta1"
 	"github.com/openshift/assisted-service/models"
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 	"github.com/openshift/installer/pkg/asset"
@@ -387,6 +388,18 @@ func buildIgnitionAssetDefaultDependencies() []asset.Asset {
 					ProvisionRequirements: hiveext.ProvisionRequirements{
 						ControlPlaneAgents: 3,
 						WorkerAgents:       5,
+					},
+				},
+			},
+			NMStateConfigs: []*aiv1beta1.NMStateConfig{
+				{
+					Spec: aiv1beta1.NMStateConfigSpec{
+						Interfaces: []*aiv1beta1.Interface{
+							{
+								Name:       "eth0",
+								MacAddress: "00:01:02:03:04:05",
+							},
+						},
 					},
 				},
 			},
--- a/pkg/asset/agent/manifests/nmstateconfig.go
+++ b/pkg/asset/agent/manifests/nmstateconfig.go
@@ -222,31 +222,41 @@ func (n *NMStateConfig) validateNMStateLabels() field.ErrorList {
 	return allErrs
 }
 
+func getFirstIP(nmStateConfig *nmStateConfig) string {
+	for _, intf := range nmStateConfig.Interfaces {
+		for _, addr4 := range intf.IPV4.Address {
+			if addr4.IP != "" {
+				return addr4.IP
+			}
+		}
+		for _, addr6 := range intf.IPV6.Address {
+			if addr6.IP != "" {
+				return addr6.IP
+			}
+		}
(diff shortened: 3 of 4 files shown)


pkg/asset/agent/image/ignition.go
```diff
@@ -120,6 +120,10 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
 		},
 	}
 
+	if len(agentManifests.NMStateConfigs) == 0 {
```
andfasano, line 123: I'd recommend to perform this check in the related asset, here: https://github.com/openshift/installer/blob/c70b4faafbe04f1d37d37465bb687c3204f34fd6/pkg/asset/agent/manifests/nmstateconfig.go#L193

zaneb: It's legal to e.g. use an empty agent-config to create the cluster-manifests and only fill in the NMStateConfigs after. So we don't want to validate too early, only when you go to create the image and you don't have any NMStateConfigs.

pawanpinjarkar: Should the message be more clear that either provide at least one nmstateconfig in cluster-manifests/nmstateconfig.yaml or if providing agent-config.yaml provide at least one nmstateconfig in the networkConfig?

andfasano: That's right, but it's just the NMStateConfig asset that loads the content of the `cluster-manifests/nmstateconfig.yaml` file if present on the disk. The AgentManifests it's just a collector, and it's resolved as a dependency in stack by the Ignition one, so not sure if I'm missing something but the validation location doesn't seem the right one?

andfasano: Screenshot from 2022-08-18 16-54-32

andfasano: Btw, if none of them (agent-config and nmstateconfig) are provided I'd expect that this [validation](https://github.com/openshift/installer/blob/c70b4faafbe04f1d37d37465bb687c3204f34fd6/pkg/asset/agent/manifests/nmstateconfig.go#L190-L192) will fail

rwsu: I think what Zane has currently implemented allows a user to first run "create cluster-manifests", edit the nmstateconfig to make any changes, and then run "create image".

If the validation is placed in nmstateconfig, then "create cluster-manifests" will immediately fail, thus forcing you to provide the nmstateconfig in the agent-config.

rwsu: > Btw, if none of them (agent-config and nmstateconfig) are provided I'd expect that this validation will fail

From my testing, the validation in ignition works.

rwsu (commented):
/lgtm

pkg/asset/agent/image/ignition.go
zaneb: > I think what Zane has currently implemented allows a user to first run "create cluster-manifests", edit the nmstateconfig to make any changes, and then run "create image".
> 
> If the validation is placed in nmstateconfig, then "create cluster-manifests" will immediately fail, thus forcing you to provide the nmstateconfig in the agent-config.

Bingo :+1:

andfasano: I'm not sure if the `create cluster-manifests` it's in good shape now, but if you run it with just the install-config.yaml (and without an agent-config.yaml, so that it will get generated, currently the [nil check ](https://github.com/openshift/installer/blob/c70b4faafbe04f1d37d37465bb687c3204f34fd6/pkg/asset/agent/manifests/nmstateconfig.go#L190-L192) will already fail

andfasano: | f the validation is placed in nmstateconfig, then "create cluster-manifests" will immediately fail, thus forcing you to provide the nmstateconfig in the agent-config.

So wouldn't be better in general to provide an agent-config template so that the NMStateConfig will not fail, rather than moving the check into another asset?

zaneb: > So wouldn't be better in general

This check is going away again in a couple of weeks, as soon as we have fixed all of the linked bugs. Let's not get _too_ general.

> to provide an agent-config template so that the NMStateConfig will not fail, rather than moving the check into another asset?

That wouldn't solve the case where you intentionally create an agent-config with no NMState data, with the intent to customise the manifests to add NMState data after `create cluster-manifests`.
The only way to catch that case is to make sure the validation doesn't run until you do `create image`, and the only way to do that is to put it in an asset that is only required by the image target.

bfournie:
Tested with dev-scripts
/approve

pkg/asset/agent/image/ignition.go
andfasano: Agree, especially considering if that check will go away soon. In general my concerns are about not making an asset self-sufficient, which could indicate a potential problem somewhere else

andfasano: Btw, the aforementioned scenario was still broken for me, I will re-test it with the latest changes as a double-check

PR #6241, opened 2022-08-18
IBMCloud: Use unique mutex for Client retrieval

cjschaef:
A shared mutex for all ibmcloud.metadata functions, including the
public Client method, deadlocks calls. Since the Client is a public
method, we should still use a mutex, but it should be a unique
mutex to prevent deadlocking.

### DIFF
--- a/pkg/asset/installconfig/ibmcloud/metadata.go
+++ b/pkg/asset/installconfig/ibmcloud/metadata.go
@@ -23,7 +23,8 @@ type Metadata struct {
 	computeSubnets      map[string]Subnet
 	controlPlaneSubnets map[string]Subnet
 
-	mutex sync.Mutex
+	mutex       sync.Mutex
+	clientMutex sync.Mutex
 }
 
 // NewMetadata initializes a new Metadata object.
@@ -131,20 +132,22 @@ func (m *Metadata) ControlPlaneSubnets(ctx context.Context) (map[string]Subnet, 
 
 // Client returns a client used for making API calls to IBM Cloud services.
 func (m *Metadata) Client() (*Client, error) {
-	m.mutex.Lock()
-	defer m.mutex.Unlock()
-
-	if m.client == nil {
-		client, err := NewClient()
-		if err != nil {
-			return nil, err
-		}
-		err = client.SetVPCServiceURLForRegion(context.TODO(), m.Region)
-		if err != nil {
-			return nil, err
-		}
-		m.client = client
+	if m.client != nil {
+		return m.client, nil
 	}
+
+	m.clientMutex.Lock()
+	defer m.clientMutex.Unlock()
+
+	client, err := NewClient()
+	if err != nil {
+		return nil, err
+	}
+	err = client.SetVPCServiceURLForRegion(context.TODO(), m.Region)
+	if err != nil {
+		return nil, err
+	}
+	m.client = client
 	return m.client, nil
 }
 


rvanderp3:
/hold
waiting for tests

pkg/asset/installconfig/ibmcloud/metadata.go
```diff
@@ -131,8 +132,8 @@ func (m *Metadata) ControlPlaneSubnets(ctx context.Context) (map[string]Subnet,
 
 // Client returns a client used for making API calls to IBM Cloud services.
 func (m *Metadata) Client() (*Client, error) {
-	m.mutex.Lock()
-	defer m.mutex.Unlock()
```
jstuever, line 135: I think this looks good, and am ready to approve it. However, I have one observation I'd like to throw out there...

If m.client is not null, from line 137, then is there value in locking the mutex at all? I'm assuming the reason for doing so is to handle the NewClient() and/or client.SetVPCServiceURLForRegion() calls. In the case where those are not being called because the client already exists, perhaps it's worth not locking at all?
```
if m.client != nil {
        return m.client, nil
}

m.clientMutex.Lock()
defer m.clientMutex.Unlock()

client,err := NewClient()
...
m.client = client
return m.client, nil
```

pkg/asset/installconfig/ibmcloud/metadata.go
```diff
@@ -131,8 +132,8 @@ func (m *Metadata) ControlPlaneSubnets(ctx context.Context) (map[string]Subnet,
...
+	m.clientMutex.Lock()
+	defer m.clientMutex.Unlock()
```
jstuever, line 136: I think this looks good, and am ready to approve it. However, I have one observation I'd like to throw out there...

If m.client is not null, from line 138, then is there value in locking the mutex at all? I'm assuming the reason for doing so is to handle the NewClient() and/or client.SetVPCServiceURLForRegion() calls. In the case where those are not being called because the client already exists, perhaps it's worth not locking at all?
```
if m.client != nil {
        return m.client, nil
}

m.clientMutex.Lock()
defer m.clientMutex.Unlock()

client,err := NewClient()
...
m.client = client
return m.client, nil
```

cjschaef: Yeah, I can make that check earlier, and lock later if necessary.

I'll make that change here, and redo some testing.

cjschaef:
`e2e-ibmcloud` install failure due to
`time="2022-08-18T20:33:42Z" level=error msg="failed to initialize the cluster: Cluster operators image-registry, storage are not available"`
Which might be a timing issue in regard to IBM Cloud LB's delaying the CCM from proper worker node labels for the Storage operator
`"msg":"Failed to initialize driver...","name":"ibm-vpc-block-csi-driver","CSIDriverName":"IBM VPC block driver","error":"Controller_Helper: Failed to initialize node metadata: error: One or few required node label(s) is/are missing`
and potentially the same reason for issues with image-registry
`http.response.status=503`

Will retest once I push up [changes recommended with mutex](https://github.com/openshift/installer/pull/6241#pullrequestreview-1077917887)

cjschaef:
Install worked as expected
```
INFO[2022-08-18T23:01:30Z] Running step e2e-ibmcloud-ipi-install-install. 
INFO[2022-08-18T23:47:00Z] Step e2e-ibmcloud-ipi-install-install succeeded after 45m30s. 
```
Failure due to conformance tests, which is currently expected
```
INFO[2022-08-19T00:42:52Z] Step e2e-ibmcloud-openshift-e2e-test failed after 55m32s. 
INFO[2022-08-19T00:42:52Z] Step phase test failed after 55m32s.
```

pkg/asset/installconfig/ibmcloud/metadata.go
cjschaef: Updated

jeffnowicki:
@jstuever this should be ready for your `approval`... thanks!

PR #6244, opened 2022-08-19
Bug 2116904: Base generated NMStateConfig on InstallConfig name

zaneb:
We don't want to require that the user provided the same (or, indeed,
any) name for the AgentConfig as for InstallConfig. So if NMStateConfig
is generated from the AgentConfig, use the cluster name from the
InstallConfig to do it.

Ideally if the user provided ZTP manifests *and* an AgentConfig, we
could generate just the NMStateConfig asset from the AgentConfig and
have the name match the one that appears in the InfraEnv. However, that
is not possible because it would entail a dependency between two
resources (NMStateConfig and InfraEnv) generated by the same command
(agent create cluster-manifests).

If no InstallConfig is provided, we will fall back to the default
cluster name "agent-cluster".

### DIFF
--- a/pkg/asset/agent/manifests/common.go
+++ b/pkg/asset/agent/manifests/common.go
@@ -8,23 +8,26 @@ import (
 )
 
 func getAgentClusterInstallName(ic *agent.OptionalInstallConfig) string {
-	return ic.Config.ObjectMeta.Name
+	return ic.ClusterName()
 }
 
 func getClusterDeploymentName(ic *agent.OptionalInstallConfig) string {
-	return ic.Config.ObjectMeta.Name
+	return ic.ClusterName()
 }
 
 func getInfraEnvName(ic *agent.OptionalInstallConfig) string {
-	return ic.Config.ObjectMeta.Name
+	return ic.ClusterName()
 }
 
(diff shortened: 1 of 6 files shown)


pkg/asset/agent/manifests/common.go
```diff
@@ -2,51 +2,43 @@ package manifests
...
 }
 
 func getObjectMetaNamespace(ic *agent.OptionalInstallConfig) string {
-	return ic.Config.Namespace
-}
-
-func getNMStateConfigName(a *agentconfig.AgentConfig) string {
-	return a.Config.ObjectMeta.Name
+	if ic.Config != nil {
+		return ic.Config.Namespace
```
celebdor, line 27: Why is it using namespace instead of ic.ClusterName()

zaneb: We still need to come up with an appropriate way to generate the namespace for ZTP manifests to go in. I had the thought that maybe we should generate a kustomize file to allow the user to override it for everything in one place. In any event, up to now we've used the namespace from the InstallConfig (which most people don't usually set), and this PR doesn't change it.

pawanpinjarkar (approved):
/approve

PR #6247, opened 2022-08-20
pkg/asset/ignition/bootstrap/cvoignore: Add group/kind/name(space) collision detection

wking:
[`docs/design/resource_dep.svg`][1] shows the complicated hierarchy of installer assets, and it is likely that not all users are familiar with all the manifests that the installer generates.  It's also possible that there is a complicated hierarchy on the caller's side, and not just a single actor, creating additional manifests to feed into a create-manifests flow.  With all of these inputs, it seems possible that folks occasionally call the installer with directory contents that result in multiple manifests for a single resource, as determined by the (group, kind, namespace, name) tuples.  This commit adds a check at the slightly-before-Ignition-serialization overrides asset, which was already in the business of iterating over the content and building ClusterVersion spec.overrides.  Failing fast before we build Ignition configs should allow for relatively cheap recovery, vs. a user's cluster failing to come up, or coming up with a config that diverges from their intention because some other manifest ended up taking precedence over the resource that they'd been trying to control.

Not covered in this commit is deduping between manifests generated on the bootstrap machine (e.g. by the various 'render' calls to operator containers).  But that can happen orthogonally, or not, and I think it's still worth having this cheap, pre-Ignition-config sanity check.

[1]: https://github.com/openshift/installer/blob/b08fe18f02768c2db33b043eba4925377b0b6cb7/docs/design/resource_dep.svg

### DIFF
--- a/pkg/asset/ignition/bootstrap/cvoignore.go
+++ b/pkg/asset/ignition/bootstrap/cvoignore.go
@@ -2,6 +2,7 @@ package bootstrap
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/ghodss/yaml"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


wking:
Lots [of][1] `Init container ... not ready`.

/retest

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/6247/pull-ci-openshift-installer-master-gofmt/1560888165072375808

pkg/asset/ignition/bootstrap/cvoignore.go
```diff
@@ -51,21 +52,34 @@ func (a *CVOIgnore) Generate(dependencies asset.Parents) error {
...
 		u := &unstructured.Unstructured{}
 		if err := yaml.Unmarshal(file.Data, u); err != nil {
 			return errors.Wrapf(err, "could not unmarshal %q", file.Filename)
 		}
+		group := u.GetObjectKind().GroupVersionKind().Group
+		kind := u.GetKind()
+		namespace := u.GetNamespace()
+		name := u.GetName()
+
+		key := fmt.Sprintf("%s |! %s |! %s |! %s", group, kind, namespace, name)
```
patrickdillon, line 67: Out of curiosity, why the `|!` separator?

patrickdillon (commented):
/lgtm
/assign @jhixson74 
/retest

pkg/asset/ignition/bootstrap/cvoignore.go
wking: I didn't spend much time poking around in allowed-characters for the four entries, and I wanted a delimiter that was unlikely to show up in any of them.

wking: it would perhaps be more robust to just use `%q%q%q%q`?

jhixson74:
This looks straight forward to me, I just want to see the tests pass first ;-)

wking:
[Confirmed][1] that this works as expected, with an error message that could be polished further, but which is already good enough to satisfy me ;)

[1]: https://github.com/openshift/release/pull/32306#issuecomment-1247636876

PR #6238, opened 2022-08-17
AGENT-308: Add 4.12 to OS_IMAGES

bfournie:
Add an entry for 4.12.

PR #6249, opened 2022-08-22
AGENT: use net host in containers for ipv6 support

karmab:


zaneb:
I would strongly prefer not to use host networking in the pod, as it may in future contain internal communication that we don't want to expose.

We should be able to detect whether we need IPv4 or IPv6 based on the address family of the Node 0 IP. It's not completely clear to me how we force podman to publish the port on the host's IPv6 addresses instead of IPv4 addresses - would `--ip6=::` do it? https://github.com/containers/podman/issues/14491#issuecomment-1148586811 suggests that we might also have to do https://developers.redhat.com/articles/2022/08/10/how-conifgure-podman-40-ipv6#create_an_interface_that_supports_the_dual_ipv4_ipv6_stack

bfournie:
I've tested this fix with https://github.com/openshift-metal3/dev-scripts/pull/1444 and verified it works to do an IPv6 installation. Note that OPENSHIFT_RELEASE_STREAM has to be 4.12 so that network type is OVNKubernetes by default, and this PR is also needed to use 4.12 images - https://github.com/openshift/installer/pull/6238

I also looked into not using `--net host` but instead usie the method in the [link](https://developers.redhat.com/articles/2022/08/10/how-conifgure-podman-40-ipv6#create_an_interface_that_supports_the_dual_ipv4_ipv6_stack) Zane has above. Currently CoreOS does not install the podman 'netavark' network driver and there are some concerns with using the netavark driver with openshift-sdn (in kubelet). The coreos team has recommended using --net host.

/lgtm

data/data/agent/systemd/units/create-cluster-and-infraenv.service.template
```diff
@@ -13,7 +13,7 @@ Environment=OPENSHIFT_INSTALL_RELEASE_IMAGE_MIRROR={{.ReleaseImageMirror}}
 EnvironmentFile=/usr/local/share/assisted-service/agent-images.env
 ExecStartPre=/bin/rm -f %t/%n.ctr-id
 ExecStartPre=/usr/local/bin/wait-for-assisted-service.sh
-ExecStart=podman run --cidfile=%t/%n.ctr-id --cgroups=no-conmon --log-driver=journald --rm --pod-id-file=%t/assisted-service-pod.pod-id --replace --name=create-cluster-and-infraenv -v /etc/assisted/manifests:/manifests -v /etc/assisted/extra-manifests:/extra-manifests -v /etc/pki/ca-trust:/etc/pki/ca-trust:z --env SERVICE_BASE_URL --env OPENSHIFT_INSTALL_RELEASE_IMAGE_MIRROR $SERVICE_IMAGE /usr/local/bin/agent-installer-client register
+ExecStart=podman run --net host --cidfile=%t/%n.ctr-id --cgroups=no-conmon --log-driver=journald --rm --pod-id-file=%t/assisted-service-pod.pod-id --replace --name=create-cluster-and-infraenv -v /etc/assisted/manifests:/manifests -v /etc/assisted/extra-manifests:/extra-manifests -v /etc/pki/ca-trust:/etc/pki/ca-trust:z --env SERVICE_BASE_URL --env OPENSHIFT_INSTALL_RELEASE_IMAGE_MIRROR $SERVICE_IMAGE /usr/local/bin/agent-installer-client register
```
zaneb, line 16: Do containers inside the pod need `--net=host` when the pod already has `--net=host`?

zaneb (commented):
I'm OK with this, but we will need to propose to master once the branch merges rather than merging to the branch here.

data/data/agent/systemd/units/create-cluster-and-infraenv.service.template
bfournie: I've tried this (adding `--net host` to assisted-service-pod.service and removing it from the containers) but it didn't work. It looks like the IPv6 address was not exposed as I get:
Sep 06 15:25:30 master-0.ostest.test.metalkube.org start-agent.sh[3644]: curl: (7) Failed to connect to fd2e:6f44:5dd8:c956::50 port 8090: Connection refused

karmab: then i guess we can merge it as is?

karmab:
opened https://github.com/openshift/installer/pull/6328 to merge in master branch instead

PR #6250, opened 2022-08-22
azure: Add client certification to terraform

rna-afk:
Adding the option for the users to authenticate using client
certificates instead of client secrets if they want. The certs
field is already in the osServicePrincipal.json and has to be set
for Azure SDK authentication. Terraform expects two extra fields
to be set for authentication and is added here.


patrickdillon:
This looks great. A couple of questions/notes:

1. How is the cert passed in `~/.azure/osServicePrincipal.json`: is it just the path or are the cert contents in the file?
2. We need to figure out the story for [the cluster secret](https://github.com/openshift/installer/blob/master/data/data/manifests/openshift/cloud-creds-secret.yaml.template#L24). The CCO does not support a cluster secret with a cert. So perhaps the best we can do is skip creating the cluster secret when authenticating with a certificate.

@2uasimojo @akhil-rane @abutcher @jharrington22 @wking have all been discussing how to handle setting the cluster secret and it will require some fiddling with manifests.

I have been trying to think of a way that the installer could authenticate with certs and still write the client secret without it being *too hacky*:

Perhaps we allow [environment-variable based authentication](https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authorization#use-environment-based-authentication) for authenticating the installer/terraform with a cert and then continue to pass the cluster secret in `~/.azure/osServicePrincipal.json`. What do you think of that idea?

pkg/asset/installconfig/azure/session.go
```diff
@@ -128,18 +132,27 @@ func getCredentials(fs auth.FileSettings) (*Credentials, error) {
...
-	}
 	tenantID := fs.Values[auth.TenantID]
 	if tenantID == "" {
 		return nil, errors.New("could not retrieve tenantId from auth file")
 	}
+	clientCertificatePassword := fs.Values[auth.CertificatePassword]
+	clientCertificatePath := fs.Values[auth.CertificatePath]
+	if tenantID == "" {
+		return nil, errors.New("could not retrieve tenantId from auth file")
+	}
```
s-amann, line 143: this conditional has already passed on line 136 and is duplicated here
```suggestion

```

pkg/asset/installconfig/azure/session.go
```diff
@@ -128,18 +132,27 @@ func getCredentials(fs auth.FileSettings) (*Credentials, error) {
...
+	if clientSecret == "" && (clientCertificatePassword == "" && clientCertificatePath == "") {
+		return nil, errors.New("could not retrieve either client secret or client certs from auth file")
+	} else if clientCertificatePath != "" {
```
s-amann, line 146: Is it possible to have `clientSecret` and only one of `certPath / certPassword` at this point, and should this be validated away with an error?

s-amann (changes requested):
please see comment regarding duplicate code, and my one open question.

pkg/asset/installconfig/azure/session.go
rna-afk: Fixed the if conditions. The installer/terraform needs at least one of the authentication methods for it to contact the APIs on azure so we need to return an error.

pkg/asset/installconfig/azure/session.go
```diff
@@ -128,18 +132,26 @@ func getCredentials(fs auth.FileSettings) (*Credentials, error) {
...
-		return nil, errors.New("could not retrieve clientSecret from auth file")
-	}
 	tenantID := fs.Values[auth.TenantID]
 	if tenantID == "" {
 		return nil, errors.New("could not retrieve tenantId from auth file")
 	}
+	clientCertificatePassword := fs.Values[auth.CertificatePassword]
+	clientCertificatePath := fs.Values[auth.CertificatePath]
+	if clientSecret == "" {
+		if clientCertificatePassword == "" || clientCertificatePath == "" {
```
bennerv, line 142: It's completely valid to have an empty `clientCertificatePassword`.  It's possible the certificate file is not encrypted at all. 

When we fetch an authorizer against the certificate, it calls the following function(s):
`c.Authorizer()` -> [calls](https://github.com/rna-afk/installer/blob/5c9a27ac82efb287c6e629bad2e1d7bcc97cdffb/vendor/github.com/Azure/go-autorest/autorest/azure/auth/auth.go#L622) `ServicePrincipalToken` -> [which attempts to decode the certificate file using the password](https://github.com/rna-afk/installer/blob/5c9a27ac82efb287c6e629bad2e1d7bcc97cdffb/vendor/github.com/Azure/go-autorest/autorest/azure/auth/auth.go#L613) -> `DecodeCertificatePfxData` [here](https://github.com/rna-afk/installer/blob/5c9a27ac82efb287c6e629bad2e1d7bcc97cdffb/terraform/providers/azurerm/vendor/github.com/Azure/go-autorest/autorest/adal/persist.go#L92) which has a comment at the top stating:
```
// If the private key is not password protected pass the empty string for password.
```

bennerv, line 142: It's completely valid to have an empty `clientCertificatePassword`.  It's possible the certificate file is not encrypted at all. 

When we fetch an authorizer against the certificate, it calls the following function(s):
`c.Authorizer()` -> [calls](https://github.com/rna-afk/installer/blob/5c9a27ac82efb287c6e629bad2e1d7bcc97cdffb/vendor/github.com/Azure/go-autorest/autorest/azure/auth/auth.go#L622) `ServicePrincipalToken` -> [which attempts to decode the certificate file using the password](https://github.com/rna-afk/installer/blob/5c9a27ac82efb287c6e629bad2e1d7bcc97cdffb/vendor/github.com/Azure/go-autorest/autorest/azure/auth/auth.go#L613) -> `DecodeCertificatePfxData` [here](https://github.com/rna-afk/installer/blob/5c9a27ac82efb287c6e629bad2e1d7bcc97cdffb/terraform/providers/azurerm/vendor/github.com/Azure/go-autorest/autorest/adal/persist.go#L92) which has a comment at the top stating:
```
// If the private key is not password protected pass the empty string for password.
```

bennerv (changes requested):

jharrington22:
@rna-afk thanks for this, I had some folks review this that are closer to the authorizer work in the ARO RP and they have a few comments that we should address before merging cc/ @patrickdillon

jharrington22 (commented):
@rna-afk thanks for this, I had some folks review this that are closer to the authorizer work in the ARO RP and they have a few comments that we should address before merging cc/ @patrickdillon

pkg/asset/installconfig/azure/session.go
rna-afk: Fixed it. Thanks for that!

patrickdillon:
@2uasimojo @akhil-rane @jharrington22 

Because we can't create a cluster secret from a cert, we have added validation to enforce that this can only be run with manual mode: 
https://github.com/openshift/installer/pull/6250/commits/352697fb93bff48b3857158ffd19a2eba38fa41e#diff-13678451d1af92fb770a62f31468daf0ca9e946ac5c707692c131a059c6ee767R95-R97

My understanding here is that this will only prevent the installer from laying down a cluster secret, users should still be able to edit manifests to provide a secret and set the CCO to the desired mode (i.e. I do not think this validatoin enforces the CCO to run in manual mode).

This seems like the simplest approach. I am still open to the idea of allowing both sets of creds to be supplied directly to the installer, but this seems like the simplest solution ATM.

PR #6254, opened 2022-08-23
Remove Washington DC from supported list

bkhadars:
No support for creating cloud connection in Wd06, Hence remove the DC from the supported list


bkhadars:
Powervs currently support for only WD04 as per the code, as discussed with @clnperez this patch is not needed.. Hence closing the PR

PR #6056, opened 2022-06-29
IBMCloud: Pull in latest openshift/api changes

cjschaef:
IBMCloud DNS support requires a change to PlatformStatus, which
is now available in the openshift/api.

PR #6258, opened 2022-08-24
Adds Insights, Console, and Storage capabilities

rluders:
This is a follow-up task for https://github.com/openshift/cluster-version-operator/pull/801


r4f4:
This looks like a duplicate of https://github.com/openshift/installer/pull/6056. Are you fine with closing this in favour of that one since it already has the `approved` label?

rluders:
> This looks like a duplicate of #6056. Are you fine with closing this in favour of that one since it already has the `approved` label?

Oh, I missed this one. Sure, lets keep the #6056

PR #6263, opened 2022-08-24
AGENT-318: Merge agent-based installer

zaneb:
Merge the `agent-installer` branch containing the agent-based installer implementation to master so that it will be part of the 4.12 installer.


jhixson74:
/cc @patrickdillon 

I think this is in good shape to merge, however I'd like to see what patrick thinks

zaneb:
I had to update because a change (#5798) merged on master that renames fields in the installConfig.

PR #6264, opened 2022-08-24
Power VS: Machine CIDR

clnperez:
Allow configuration of the machine CIDR. Currently the only option is the default.


clnperez:
@mjturek PTAL

data/data/powervs/cluster/power_network/pi_network.tf
```diff
@@ -1,7 +1,7 @@
 locals {
-  ids               = data.ibm_pi_dhcps.dhcp_services.servers[*].dhcp_id
-  names             = data.ibm_pi_dhcps.dhcp_services.servers[*].network_name
-  dhcp_id_from_name = var.pvs_network_name == "" ? "" : matchkeys(local.ids, local.names, [var.pvs_network_name])[0]
+  ids                = data.ibm_pi_dhcps.dhcp_services.servers[*].dhcp_id
```
barbacbd, line 2: these are causing tf-fmt to complain and can probably be squashed anyways since no changes other than whitespace were added.

data/data/powervs/cluster/power_network/variables.tf
```diff
@@ -29,3 +29,8 @@ variable "cloud_conn_name" {
   description = "The name of a pre-created Power VS Cloud connection."
   default     = ""
 }
+
+variable "machine_cidr" {
+  type        = string
+  description = "The machine network (IPv4 only)"
+}
```
barbacbd, line 36: nit: This passes tf tests locally, but adding the newline at the end is the suggestion

clnperez:
`Pod got deleted unexpectedly`

trying again

/retest

clnperez:
@barbacbd any idea what's going on? same issue again

clnperez:
@r4f4 @rna-afk any ideas? we probably do want the go-fmt test to run & pass, no matter what the openshift-ci bot says :D

r4f4:
> @r4f4 @rna-afk any ideas? we probably do want the go-fmt test to run & pass, no matter what theopenshift-ci bot says :D

It's something with Prow, so maybe the folks from testplatform might know. What seems to be working better for me lately it's to test a few jobs at a time.

/test gofmt

clnperez:
finally! 
@rna-afk can we get an approve now that CI has settled?

data/data/powervs/cluster/power_network/variables.tf
```diff
@@ -29,3 +29,8 @@ variable "cloud_conn_name" {
   description = "The name of a pre-created Power VS Cloud connection."
   default     = ""
 }
+
+variable "machine_cidr" {
```
rna-afk, line 33: Wouldn't this cause an error since it's not being set and has no default? I can't test it so relying on you guys

clnperez: the default is embedded in the functionality itself so this is safe. additionally, i'd rather not set a default here in case powervs changes it and they're out of sync

data/data/powervs/cluster/main.tf
```diff
@@ -36,6 +36,7 @@ module "pi_network" {
   cloud_instance_id = var.powervs_cloud_instance_id
   resource_group    = var.powervs_resource_group
   pvs_network_name  = var.powervs_network_name
+  machine_cidr      = var.machine_v4_cidrs[0]
```
clnperez, line 39: @rna-afk there's no default specified in that vars file b/c it's always set here ^

data/data/powervs/cluster/power_network/variables.tf
clnperez: sorry, this isn't entirely accurate and i realized it when i was rebasing. there is a default (pointed out in my other comment). that will always be set in the install-config file and passed to this module.

clnperez:
fixed the merge conflict. hopefully this time it doesn't take an entire week for all these tests to run without crashing 🤞🏻

PR #6265, opened 2022-08-25
download 'aliyun'

jianli-wei:
The alibabacloud client "aliyun" would be used when pre-configuring some resources (e.g. VPC, bastion host, etc.) before launching an OCP cluster with customization.


jianlinliu:
Suggest also update Dockerfile.upi.ci.rhel8 together.

PR #6268, opened 2022-08-25
Bug OCPBUGS-164: OpenStack: Set minimum disk of a flavor to 100 GB

dulek:
Other platforms require at least 100 GB of disk size and we've updated
openshift-docs to reflect that in OpenStack too. Seems like we forgot to
update flavor validation code and docs in the installer. This commit
fixes this.

### DIFF
--- a/docs/user/openstack/README.md
+++ b/docs/user/openstack/README.md
@@ -80,7 +80,7 @@ For a successful installation it is required:
 - Server Groups: 2, plus one per additional Availability zone in each machine-pool
 - RAM: 112 GB
 - vCPUs: 28
-- Volume Storage: 175 GB
+- Volume Storage: 700 GB
 - Instances: 7
 - Depending on the type of [image registry backend](#image-registry-requirements) either 1 Swift container or an additional 100 GB volume.
 - OpenStack resource tagging
@@ -97,13 +97,13 @@ Once you configure the quota for your project, please ensure that the user for t
 
 ### Master Nodes
 
-The default deployment stands up 3 master nodes, which is the minimum amount required for a cluster. For each master node you stand up, you will need 1 instance, and 1 port available in your quota. They should be assigned a flavor with at least 16 GB RAM, 4 vCPUs, and 25 GB Disk (or Root Volume). It is theoretically possible to run with a smaller flavor, but be aware that if it takes too long to stand up services, or certain essential services crash, the installer could time out, leading to a failed install.
+The default deployment stands up 3 master nodes, which is the minimum amount required for a cluster. For each master node you stand up, you will need 1 instance, and 1 port available in your quota. They should be assigned a flavor with at least 16 GB RAM, 4 vCPUs, and 100 GB Disk (or Root Volume). It is theoretically possible to run with a smaller flavor, but be aware that if it takes too long to stand up services, or certain essential services crash, the installer could time out, leading to a failed install.
 
 The master nodes are placed in a single Server group with "soft anti-affinity" policy by default; the machines will therefore be created on separate hosts when possible.
 
 ### Worker Nodes
 
-The default deployment stands up 3 worker nodes. Worker nodes host the applications you run on OpenShift. The flavor assigned to the worker nodes should have at least 2 vCPUs, 8 GB RAM and 25 GB Disk (or Root Volume). However, if you are experiencing `Out Of Memory` issues, or your installs are timing out, try increasing the size of your flavor to match the master nodes: 4 vCPUs and 16 GB RAM.
+The default deployment stands up 3 worker nodes. Worker nodes host the applications you run on OpenShift. The flavor assigned to the worker nodes should have at least 2 vCPUs, 8 GB RAM and 100 GB Disk (or Root Volume). However, if you are experiencing `Out Of Memory` issues, or your installs are timing out, try increasing the size of your flavor to match the master nodes: 4 vCPUs and 16 GB RAM.
 
 The worker nodes are placed in a single Server group with "soft anti-affinity" policy by default; the machines will therefore be created on separate hosts when possible.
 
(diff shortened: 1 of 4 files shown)


dulek:
/override ci/prow/e2e-openstack-kuryr

At least this one doesn't need to run here.

dulek:
cc @mandre

docs/user/openstack/README.md
```diff
@@ -97,21 +97,21 @@ Once you configure the quota for your project, please ensure that the user for t
 
 ### Master Nodes
 
-The default deployment stands up 3 master nodes, which is the minimum amount required for a cluster. For each master node you stand up, you will need 1 instance, and 1 port available in your quota. They should be assigned a flavor with at least 16 GB RAM, 4 vCPUs, and 25 GB Disk (or Root Volume). It is theoretically possible to run with a smaller flavor, but be aware that if it takes too long to stand up services, or certain essential services crash, the installer could time out, leading to a failed install.
+The default deployment stands up 3 master nodes, which is the minimum amount required for a cluster. For each master node you stand up, you will need 1 instance, and 1 port available in your quota. They should be assigned a flavor with at least 16 GB RAM, 4 vCPUs, and 100 GB Disk (or Root Volume). It is theoretically possible to run with a smaller flavor, but be aware that if it takes too long to stand up services, or certain essential services crash, the installer could time out, leading to a failed install.
```
mandre, line 100: Could you also update the total required storage (on line 83)? Github won't let me comment on the right line.

mandre (commented):
Well, ok the validation works, but now CI is failing because our flavors only have a 80GB disk (which should be enough for CI purpose).
How about we export the `OPENSHIFT_INSTALL_SKIP_PREFLIGHT_VALIDATIONS` environment variable in CI?

dulek:
Yup, that will be required, I'm on it.

pierreprinetti:
> How about we export the OPENSHIFT_INSTALL_SKIP_PREFLIGHT_VALIDATIONS environment variable in CI?

I think that we should test on flavors that reflect the minimum requirements exactly. I'm going to create such flavors on all our CI clouds and update [our CI configuration](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openstack/conf/resourcenames/openstack-conf-resourcenames-commands.sh). WDYT @mandre @dulek?

If instead we think that 80GB are enough, then why setting the requirement to 100GB in the docs?

mandre:
> > How about we export the OPENSHIFT_INSTALL_SKIP_PREFLIGHT_VALIDATIONS environment variable in CI?
> 
> I think that we should test on flavors that reflect the minimum requirements exactly. I'm going to create such flavors on all our CI clouds and update [our CI configuration](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openstack/conf/resourcenames/openstack-conf-resourcenames-commands.sh). WDYT @mandre @dulek?

Please confirm first that we have enough quota / resource in our clouds.

> If instead we think that 80GB are enough, then why setting the requirement to 100GB in the docs?

Because the product docs are intended for customers (you'll need to account for the growing size of data as your cluster ages) and not CI. Also, it aligns with other platforms.

pierreprinetti:
>you'll need to account for the growing size of data as your cluster ages

I thought that this was not in scope of _minimum_ requirements.

What does "minimum" in "minimum requirements for masters" mean here?

Either it means "the bare minimum for installing", and then we can set it to 80GB.
Or we want to give hints as to the cluster growth requirements, and then maybe we can go the extra mile and see if there's a metric (e.g. the number of pods) that can be a good proxy for calculating the required disk size.

As for the workers, I think we should really state what's the minimum requirements for standing up the cluster, because end users will need to figure out the extra resources based on their workloads.

dulek:
> > you'll need to account for the growing size of data as your cluster ages
> 
> I thought that this was not in scope of _minimum_ requirements.
> 
> What does "minimum" in "minimum requirements for masters" mean here?
> 
> Either it means "the bare minimum for installing", and then we can set it to 80GB. Or we want to give hints as to the cluster growth requirements, and then maybe we can go the extra mile and see if there's a metric (e.g. the number of pods) that can be a good proxy for calculating the required disk size.
> 
> As for the workers, I think we should really state what's the minimum requirements for standing up the cluster, because end users will need to figure out the extra resources based on their workloads.

I would not overstep here and try to be smarter than the core OpenShift team that recommends 100 GB for other platforms. 100 GB is in the official docs for a while now, we got to stick with it. The only question is if we can go away with 80 GB on the CI, which might be true as CI was working okay for a while.

mandre:
Then perhaps we want to have a warning when the flavors disk size is between 25 (the absolute minimum required to install a cluster) and 100 GB (the recommended minimum value for production clusters) and fail the validation when it's under 25GB like it is the case today. WDYT?

pierreprinetti:
> Then perhaps we want to have a warning when the flavors disk size is between 25 (the absolute minimum required to install a cluster) and 100 GB (the recommended minimum value for production clusters) and fail the validation when it's under 25GB like it is the case today. WDYT?

LGTM :+1:

dulek:
> > Then perhaps we want to have a warning when the flavors disk size is between 25 (the absolute minimum required to install a cluster) and 100 GB (the recommended minimum value for production clusters) and fail the validation when it's under 25GB like it is the case today. WDYT?
> 
> LGTM +1
> 
> and keep the docs at 100GB as per the other platforms, right?

Alright, I did so then. Seems like in other platforms warnings are just logged, so I haven't modified the whole validation system to be able to return both fatal errors and warnings.

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -11,23 +13,26 @@ import (
...
 
 type flavorRequirements struct {
-	RAM, VCPUs, Disk int
+	RAM, VCPUs, Disk, RecommendedDisk int
 }
 
 const (
-	minimumStorage = 25
+	minimumStorage    = 25
+	recommenedStorage = 100
```
pierreprinetti, line 21: typo

pierreprinetti (approved):

docs/user/openstack/README.md
dulek: Done.

pkg/asset/installconfig/openstack/validation/machinepool.go
dulek: Wow, nice catch.

PR #6271, opened 2022-08-26
Bug/rosa tag count

thatcherhubbard:
For ROSA|OSD-on-AWS installs the installer limits customer tags to 8. This has two issues:

- It's based on the limit of 10 for *object* tags (buckets are 50) -> https://docs.aws.amazon.com/AmazonS3/latest/userguide/CostAllocTagging.html
- The Managed ROSA or OSD installs actually use 4 of the available tags

ROSA/OSD doesn't tag objects, just the buckets, so the limit of 50 is a more appropriate place to start.

This PR includes simple changes to allow for up to 40 customer-defined tags to be applied.

### DIFF
--- a/pkg/types/aws/validation/platform.go
+++ b/pkg/types/aws/validation/platform.go
@@ -51,7 +51,7 @@ func validateUserTags(tags map[string]string, propagatingTags bool, fldPath *fie
 	if len(tags) == 0 {
 		return allErrs
 	}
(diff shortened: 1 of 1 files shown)


jstuever:
Also the same: https://github.com/openshift/installer/pull/6274

patrickdillon:
/close
In favor of #6274 and #6303 

Also note that the 10 tag limit for s3 objects is real. We create and tag an s3 object during install. Testing should have quickly revealed this.

PR #6273, opened 2022-08-29
vSphere Zonal: Force tag cat names

jcpowermac:
This PR changes tag category naming for zonal from
a definable option to forced defined

- openshift-region
- openshift-zone


jcpowermac:
Need to change the platform spec as well.

jcpowermac:
/close

A new pr will replace this.

PR #6274, opened 2022-08-29
OCPBUGS-1234 : [CFE-580] Extend user tags limit to 40 based on AWS limits

TrilokGeer:
Extend user tag limit to 40 based on AWS limits. 10 tags are reserved for OpenShift use. Previously, there has been a confusion on the tag limit for S3 bucket vs S3 bucket object tag limit. S3 bucket object tag limit is 10 while the requirement is for S3 bucket which has limit of 50 tags.

### DIFF
--- a/data/data/gcp/bootstrap/main.tf
+++ b/data/data/gcp/bootstrap/main.tf
@@ -50,10 +50,11 @@ resource "google_compute_address" "bootstrap" {
 }
 
 resource "google_compute_firewall" "bootstrap_ingress_ssh" {
-  count       = var.gcp_network_project_id != "" ? 0 : 1
+  count       = var.gcp_create_firewall_rules ? 1 : 0
   name        = "${var.cluster_id}-bootstrap-in-ssh"
   network     = var.network
   description = local.description
+  project     = var.gcp_network_project_id
 
   allow {
     protocol = "tcp"
--- a/data/data/gcp/cluster/main.tf
+++ b/data/data/gcp/cluster/main.tf
@@ -65,6 +65,8 @@ module "network" {
   master_subnet       = var.gcp_control_plane_subnet
   worker_subnet       = var.gcp_compute_subnet
   network_project_id  = var.gcp_network_project_id
+
+  create_firewall_rules = var.gcp_create_firewall_rules
 }
 
 module "dns" {
--- a/data/data/gcp/cluster/network/firewall.tf
+++ b/data/data/gcp/cluster/network/firewall.tf
@@ -1,8 +1,9 @@
 resource "google_compute_firewall" "api" {
-  count       = var.network_project_id != "" ? 0 : 1
+  count       = var.create_firewall_rules ? 1 : 0
   name        = "${var.cluster_id}-api"
   network     = local.cluster_network
   description = local.description
+  project     = var.network_project_id
 
   # API
   allow {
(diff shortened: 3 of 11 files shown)


data/data/aws/bootstrap/main.tf
```diff
@@ -61,10 +61,12 @@ resource "aws_s3_bucket_object" "ignition" {
   server_side_encryption = "AES256"
 
   tags = merge(
+    {
+      "kubernetes.io/cluster/${var.cluster_id}" = "owned"
+    },
```
patrickdillon, line 66: Why add the cluster owned tag? I'm not sure if we consider this to be owned by the cluster and it looks like none of the other bootstrap resources are tagged similarly.

data/data/aws/bootstrap/main.tf
```diff
@@ -61,10 +61,12 @@ resource "aws_s3_bucket_object" "ignition" {
...
     {
       "Name" = "${var.cluster_id}-bootstrap"
     },
-    local.tags,
```
patrickdillon, line 67: We discussed this in a meeting just now. s3 objects  have a limit of 10 tags. We have two options here:

1. Skip user tags.
2. Only take the first X number of tags which would end up totaling 10. So if we have 1 static tag, we would take 9 user-defined tags.

I lean toward option 2 and I believe we are in agreement, because otherwise removing tagging would generally be a regression in the product.

patrickdillon (commented):
This generally looks good. There are a few comments about tagging the s3 object and the commit structure needs to be cleaned up a bit. 

Can you add commit messages for the motivation behind these changes?

data/data/aws/bootstrap/main.tf
bparees: > otherwise removing tagging would generally be a regression in the product.

I talked to @TrilokGeer about this a bit and he explained the challenge of taking the "first" N tags given it's an unsorted map.  I suggested sorting the map alphanumerically and then taking the first N from the sorted list.  So that's one option to at least provide deterministic behavior when you have more than N tags provided.

But it also sounds like we shouldn't have tagged these objects in the first place, and they are ephemeral (being deleted post bootstrap).  Between those two things, I do have to wonder if it's really critical we preserve this behavior or if we can just remove the behavior and not have to add additional complexity around selecting which tags to use on the bucket objects.

I don't have much sense of why a customer would particularly care about the tagging of these resources, do you?

patrickdillon: There is also discussion in #6303, so the context is spread between the two PRs. Terraform already sorts map keys lexicographically, so that aspect should be handled without further effort.

I wouldn't consider maintaining this functionality to be critical, but I also don't have context for customer use cases. In general, I have gathered that customers come up with policies and automation that use tags. I don't have enough knowledge to say that those cases would never apply to s3 objects.

Hopefully this added clarification about terraform sorting clarifies the level of complexity. Given this info, which I learned yesterday, I don't think it is burdensome to implement, but I am definitely open to skipping.

patrickdillon (commented):
/approve

But I think we need to switch 9 -> 8

data/data/aws/bootstrap/main.tf
```diff
@@ -11,6 +11,18 @@ locals {
   volume_type      = "gp3"
   volume_size      = 30
   volume_iops      = local.volume_type == "io1" ? 100 : 0
+
+  // s3 object supports only 10 tags. The first 8 tags from
+  // the list of aws_extra_tags are used for s3 object
+  // slice function uses new_tag_len as excluding index
+  new_tag_len    = 9
```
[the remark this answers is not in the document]
TrilokGeer: Correct. :+1:

patrickdillon:
looks like a bad rebase or something happened here @TrilokGeer PTAL

TrilokGeer:
There is a metadata corruption for git in my local which created out-of-sync commits in local and server. Created a new PR for the follow-on : https://github.com/openshift/installer/pull/6345

PR #6272, opened 2022-08-27
[WIP] AGENT: Fix releaseImageMirror formatting

celebdor:
When the image is of the form:

    quay.io/openshift-release-dev/ocp-release@sha256:300bce8246cf880e792e106607925de0a404484637627edf5f517375517d54a4

Splitting by ':' means that the extracted location will be:

    quay.io/openshift-release-dev/ocp-release@sha256:

Instead of

    quay.io/openshift-release-dev/ocp-release

Whereas in registries.conf we'd have the location be:

    [[registry]]
      location = "quay.io/openshift-release-dev/ocp-release"

Thus. There wouldn't be a match and the release image will be empty in
places like:

    /usr/local/share/assisted-service/assisted-service.env

or

    /etc/systemd/system/create-cluster-and-infraenv.service

Signed-off-by: Antoni Segura Puimedon <antoni@redhat.com>

PR #6275, opened 2022-08-29
AGENT-326: Fix releaseImage mirror handling with unit test

bfournie:
Based on https://github.com/openshift/installer/pull/6272, this fixes the mirror generation to handle release images that include the sha256 checksum, e.g:
`quay.io/openshift-release-dev/ocp- release@sha256:300bce8246cf880e792e106607925de0a404484637627edf5f517375517d54a4`

### DIFF
--- a/pkg/asset/agent/image/ignition.go
+++ b/pkg/asset/agent/image/ignition.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/coreos/ignition/v2/config/util"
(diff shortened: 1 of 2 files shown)


pkg/asset/agent/image/ignition.go
```diff
@@ -387,3 +380,22 @@ func RetrieveRendezvousIP(agentConfig *agent.Config, nmStateConfigs []*v1beta1.N
...
+func getMirrorFromRelease(releaseImage string, registriesConfig *mirror.RegistriesConf) string {
+
+	releaseImageMirror := ""
+	source := regexp.MustCompile(`^(.+?)(@sha256)?:(.+)`).FindStringSubmatch(releaseImage)
+	for _, config := range registriesConfig.MirrorConfig {
+		if config.Location == source[1] {
+			// include the tag with the build release image
+			if len(source) == 4 {
+				// Has Sha256
+				releaseImageMirror = fmt.Sprintf("%s%s:%s", config.Mirror, source[2], source[3])
```
andfasano, line 393: nit: this could be slightly simplified to: `releaseImageMirror = fmt.Sprintf("%s@sha256:%s", config.Mirror, source[3])`

pkg/asset/agent/image/ignition.go
```diff
@@ -387,3 +380,22 @@ func RetrieveRendezvousIP(agentConfig *agent.Config, nmStateConfigs []*v1beta1.N
...
+			} else if len(source) == 3 {
+				releaseImageMirror = fmt.Sprintf("%s:%s", config.Mirror, source[2])
+			}
+		}
```
zaneb, line 397: Pre-existing issue, but it's a bit surprising that we don't return or break here, so we're always choosing the _last_ matching mirror on the list.

PR #6279, opened 2022-08-30
[WIP] Vsphere one terraform to rule them all

jcpowermac:
do not merge
ci testing only


jcpowermac:
/close

@patrickdillon 
after @rvanderp3 and I have worked on this for the past couple days we feel this is too risky to try to incorporate both terraform workflows

PR #6281, opened 2022-08-30
OCPBUGS-705: vSphere privilege checking failing when providing user-defined folder and/or resource pool

rvanderp3:


rvanderp3:
> I see the `e2e-vsphere` test failed for a reason unrelated to permissions and the actual install went through fine. Do you expect `e2e-vsphere-zones` to pass if we give it another try? If not, no worries and the changes lgtm.

Let's give it another run.  I'm hoping for a successful install, not necessarily passing e2e.  It looks like we might have had a flake with the spot instances on the build cluster.

/test e2e-vsphere.zones

rvanderp3:
@bostrt the install completed on e2e-vsphere-zones

bostrt:
Awesome!

/lgtm

PR #6284, opened 2022-08-30
gcp: machines to use networkProjectID when defined

jstuever:
This change adds the networkProjectID to the network configuration for
the compute and control plane machine manifests. It will be skipped
(omitempty) when the parameter is not specified in the install config.

https://issues.redhat.com/browse/CORS-2069


barbacbd (commented):
/lgtm

PR #6287, opened 2022-08-31
AGENT-325 Set NetworkType in AgentClusterInstall

rwsu:
The IgnitionConfig NetworkType value should be transferred to
the AgentClusterInstall. This will allow the installed cluster
to have the correct network type installed. Currently, the
installed cluster defaults to OpenShiftSDN even if the
NetworkType is set to OVNKubernetes.

If the NetworkType is unspecified, the value is defaulted to
the IgnitionConfig's default value: OpenShiftSDN. A message
is logged to the console indicating that the value was unspecified
and defaulted to OpenShiftSDN.

### DIFF
--- a/pkg/asset/agent/manifests/agentclusterinstall.go
+++ b/pkg/asset/agent/manifests/agentclusterinstall.go
@@ -12,6 +12,8 @@ import (
 	"github.com/openshift/installer/pkg/asset"
 	"github.com/openshift/installer/pkg/asset/agent"
 	"github.com/openshift/installer/pkg/ipnet"
+	"github.com/openshift/installer/pkg/types"
+	"github.com/openshift/installer/pkg/types/defaults"
 	"github.com/openshift/installer/pkg/validate"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -110,10 +112,8 @@ func (a *AgentClusterInstall) Generate(dependencies asset.Parents) error {
 		if installConfig.Config.NetworkType != "" {
 			agentClusterInstall.Spec.Networking.NetworkType = installConfig.Config.NetworkType
 		} else {
-			// default NetworkType to OpenShiftSDN if unspecified
-			// https://docs.openshift.com/container-platform/4.11/installing/installing_aws/installing-aws-network-customizations.html#installation-configuration-parameters-network_installing-aws-network-customizations
-			logrus.Info("NetworkType is not specified in install config. Defaulting to OpenShiftSDN.")
-			agentClusterInstall.Spec.Networking.NetworkType = "OpenShiftSDN"
+			agentClusterInstall.Spec.Networking.NetworkType = a.getDefaultNetworkType(installConfig.Config)
+			logrus.Infof("NetworkType is not specified in install config. Defaulting to %s.", agentClusterInstall.Spec.Networking.NetworkType)
 		}
 
 		// TODO: Handle the case where both IPv4 and IPv6 VIPs are specified
@@ -169,10 +169,9 @@ func (a *AgentClusterInstall) Load(f asset.FileFetcher) (bool, error) {
 	}
 
 	if agentClusterInstall.Spec.Networking.NetworkType == "" {
-		// default NetworkType to OpenShiftSDN if unspecified
-		// https://docs.openshift.com/container-platform/4.11/installing/installing_aws/installing-aws-network-customizations.html#installation-configuration-parameters-network_installing-aws-network-customizations
-		logrus.Info("NetworkType is not specified in AgentClusterInstall. Defaulting to OpenShiftSDN.")
-		agentClusterInstall.Spec.Networking.NetworkType = "OpenShiftSDN"
+		defaultInstallConfig := &types.InstallConfig{}
+		agentClusterInstall.Spec.Networking.NetworkType = a.getDefaultNetworkType(defaultInstallConfig)
+		logrus.Infof("NetworkType is not specified in AgentClusterInstall. Defaulting to %s.", agentClusterInstall.Spec.Networking.NetworkType)
 	}
 
 	a.Config = agentClusterInstall
@@ -198,4 +197,12 @@ func (a *AgentClusterInstall) finish() error {
 	}
 
 	return nil
+}
+
+func (a *AgentClusterInstall) getDefaultNetworkType(installConfig *types.InstallConfig) string {
+	// Returns default network type for a given install config.
+	// For 4.11, the default NetworkType is set to OpenShiftSDN if unspecified.
+	// https://docs.openshift.com/container-platform/4.11/installing/installing_aws/installing-aws-network-customizations.html#installation-configuration-parameters-network_installing-aws-network-customizations
+	defaults.SetInstallConfigDefaults(installConfig)
+	return installConfig.NetworkType
 }
--- a/pkg/asset/agent/manifests/agentclusterinstall.go
+++ b/pkg/asset/agent/manifests/agentclusterinstall.go
@@ -14,6 +14,7 @@ import (
 	"github.com/openshift/installer/pkg/ipnet"
 	"github.com/openshift/installer/pkg/validate"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/yaml"
@@ -106,6 +107,15 @@ func (a *AgentClusterInstall) Generate(dependencies asset.Parents) error {
 			},
 		}
 
+		if installConfig.Config.NetworkType != "" {
+			agentClusterInstall.Spec.Networking.NetworkType = installConfig.Config.NetworkType
+		} else {
+			// default NetworkType to OpenShiftSDN if unspecified
+			// https://docs.openshift.com/container-platform/4.11/installing/installing_aws/installing-aws-network-customizations.html#installation-configuration-parameters-network_installing-aws-network-customizations
+			logrus.Info("NetworkType is not specified in install config. Defaulting to OpenShiftSDN.")
+			agentClusterInstall.Spec.Networking.NetworkType = "OpenShiftSDN"
+		}
+
 		// TODO: Handle the case where both IPv4 and IPv6 VIPs are specified
 		apiVIP, ingressVIP := getVIPs(&installConfig.Config.Platform)
 
@@ -157,6 +167,14 @@ func (a *AgentClusterInstall) Load(f asset.FileFetcher) (bool, error) {
 		err = errors.Wrapf(err, "failed to unmarshal %s", agentClusterInstallFilename)
 		return false, err
 	}
+
+	if agentClusterInstall.Spec.Networking.NetworkType == "" {
+		// default NetworkType to OpenShiftSDN if unspecified
+		// https://docs.openshift.com/container-platform/4.11/installing/installing_aws/installing-aws-network-customizations.html#installation-configuration-parameters-network_installing-aws-network-customizations
+		logrus.Info("NetworkType is not specified in AgentClusterInstall. Defaulting to OpenShiftSDN.")
+		agentClusterInstall.Spec.Networking.NetworkType = "OpenShiftSDN"
+	}
+
 	a.Config = agentClusterInstall
 
 	if err = a.finish(); err != nil {
@@ -169,6 +187,14 @@ func (a *AgentClusterInstall) finish() error {
 
 	if a.Config == nil {
 		return errors.New("missing configuration or manifest file")
+	}
+
+	switch a.Config.Spec.Networking.NetworkType {
+	case
+		"OVNKubernetes",
+		"OpenShiftSDN":
+	default:
+		return errors.New("networkType has incorrect value. Expect value to be either 'OpenShiftSDN' or 'OVNKubernetes'")
 	}
 
 	return nil
--- a/pkg/asset/agent/manifests/agentclusterinstall.go
+++ b/pkg/asset/agent/manifests/agentclusterinstall.go
@@ -208,27 +208,41 @@ func setNetworkType(aci *hiveext.AgentClusterInstall, installConfig *types.Insta
 	aci.Spec.Networking.NetworkType = installConfig.NetworkType
 }
 
-func isIPv4(ipAddress string) bool {
-	return strings.Count(ipAddress, ":") < 2
+func isIPv4(ipAddress net.IP) bool {
+	ip := ipAddress.To4()
+	if ip == nil {
+		return false
+	}
+	return true
 }
 
-func isIPv6(ipAddress string) bool {
-	return strings.Count(ipAddress, ":") >= 2
+func isIPv6(ipAddress net.IP) bool {
+	ip := ipAddress.To16()
+	if ip == nil {
+		return false
+	}
+	return true
 }
 
 func (a *AgentClusterInstall) validateIPAddressAndNetworkType() field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	fieldPath := field.NewPath("spec", "networking", "networkType")
+	clusterNetworkPath := field.NewPath("spec", "networking", "clusterNetwork")
+	serviceNetworkPath := field.NewPath("spec", "networking", "serviceNetwork")
 
 	if a.Config.Spec.Networking.NetworkType == string(operv1.NetworkTypeOpenShiftSDN) {
 		hasIPv6 := false
 		hasIPv4 := false
 		for _, cn := range a.Config.Spec.Networking.ClusterNetwork {
-			if isIPv4(cn.CIDR) {
+			ip, _, errCIDR := net.ParseCIDR(cn.CIDR)
+			if errCIDR != nil {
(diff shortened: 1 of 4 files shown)


pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -156,6 +166,14 @@ func (a *AgentClusterInstall) Load(f asset.FileFetcher) (bool, error) {
 		err = errors.Wrapf(err, "failed to unmarshal %s", agentClusterInstallFilename)
 		return false, err
 	}
+
+	if agentClusterInstall.Spec.Networking.NetworkType == "" {
+		// default NetworkType to OpenShiftSDN if unspecified
+		// https://docs.openshift.com/container-platform/4.11/installing/installing_aws/installing-aws-network-customizations.html#installation-configuration-parameters-network_installing-aws-network-customizations
+		logrus.Info("NetworkType is not specified in AgentClusterInstall. Defaulting to OpenShiftSDN.")
+		agentClusterInstall.Spec.Networking.NetworkType = "OpenShiftSDN"
```
celebdor, line 174: Is there some way to retrieve the default from elsewhere in the code so that if it changes we don't get this left behind?

celebdor: Found it! It's in pkg/types/defaults/installconfig.go
```
defaultNetworkTYpe = string(operv1.NetworkTypeOpenShiftSDN)).
```
I believe if unset we should just retrieve it from installconfig defaults.

celebdor: Now that I look closer, in the same go file, if we look at SetInstallConfigDefaults, there's logic that changes the SDN depending on topology and upstrea/downstream. Being able to consume that logic would be useful.

celebdor: Now that I look closer, in the same go file, if we look at SetInstallConfigDefaults, there's logic that changes the SDN depending on topology and upstrea/downstream. Being able to consume that logic would be useful.

pkg/asset/agent/manifests/agentclusterinstall_test.go
```diff
@@ -34,6 +37,13 @@ func TestAgentClusterInstall_Generate(t *testing.T) {
 			},
 			expectedError: "missing configuration or manifest file",
 		},
+		{
+			name: "install config has bad network type value",
+			dependencies: []asset.Asset{
+				installConfigWithBadNetworkType,
+			},
+			expectedError: "networkType has incorrect value. Expect value to be either 'OpenShiftSDN' or 'OVNKubernetes'",
```
celebdor, line 45: Does InstallConfig only allow those? I wonder what is set for other SDNs.

rwsu: There are a few more network types that are available: Kuryr, Raw, and SimpleMacvlan. 

https://github.com/openshift/api/blob/master/operator/v1/types_network.go#L598-613

Should we be supporting these other types?

pkg/asset/agent/manifests/agentclusterinstall.go
rwsu: Good find. I will change it to use SetInstallConfigDefaults.

pkg/asset/agent/manifests/agentclusterinstall_test.go
celebdor: I think we should not prevent them from being set. Since people might also want to use things like calico and cilium

karmab: i think you should allow other network types, such as Calico or Contrail, so not block on specific values

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -156,6 +166,13 @@ func (a *AgentClusterInstall) Load(f asset.FileFetcher) (bool, error) {
 		err = errors.Wrapf(err, "failed to unmarshal %s", agentClusterInstallFilename)
 		return false, err
 	}
+
+	if agentClusterInstall.Spec.Networking.NetworkType == "" {
+		defaultInstallConfig := &types.InstallConfig{}
+		agentClusterInstall.Spec.Networking.NetworkType = a.getDefaultNetworkType(defaultInstallConfig)
```
karmab, line 172: default should be OVNKubernetes

pkg/asset/agent/manifests/agentclusterinstall_test.go
rwsu: Ok, I will update to allow other network types.

pkg/asset/agent/manifests/agentclusterinstall.go
rwsu: Thanks @karmab

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -157,6 +169,13 @@ func (a *AgentClusterInstall) Load(f asset.FileFetcher) (bool, error) {
...
+		logrus.Infof("NetworkType is not specified in AgentClusterInstall. Defaulting to %s.", agentClusterInstall.Spec.Networking.NetworkType)
+	}
+
```
andfasano, line 178: This block seems to be logically duplicated with the previous one, what about having just one "default" method in common to both the `Load()` and `Generate()`?

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -171,5 +190,66 @@ func (a *AgentClusterInstall) finish() error {
...
+func (a *AgentClusterInstall) getDefaultNetworkType(installConfig *types.InstallConfig) string {
+	// Returns default network type for a given install config.
+	// The default NetworkType is set to OVNKubernetes if unspecified.
+	defaults.SetInstallConfigDefaults(installConfig)
+	return installConfig.NetworkType
+}
+
+func isIPv4(ipAddress string) bool {
+	return strings.Count(ipAddress, ":") < 2
```
andfasano, line 208: What about using instead net.ParseCIDR()?

pkg/asset/agent/manifests/agentclusterinstall.go
rwsu: Sure. I will fold the log message into the getDefaultNetworkType function.

pkg/asset/agent/manifests/agentclusterinstall.go
rwsu: How does one use net.ParseCIDR() to determine if the IP address is v4 or v6?

andfasano: If there's no error, it is possible to reuse the first result from [net.ParseCIDR()](https://pkg.go.dev/net#ParseCIDR) to check if the [IP](https://pkg.go.dev/net#IP) is a v4, ie something like that

```
ip, _, err := net.ParseCIDR(cidrStr)
if err != nil {
  ...
}
ip4 := ip.To4()
if ip4 != nil {
   ... //Ok
}
```
(There are also other nice handy functions in that package to manipulate ip addresses, ie To16() for v6)

rwsu: Unless there is a problem with it, I prefer the current single line implementation.

andfasano: Well, given that the language offers a native library, which is well tested and robust, for that specific topic then IMHO I don't see a reason for not using it. Also, in other code portions of the installer where similar validations are applied the `net` package is tipically adopted

rwsu: Good point. I'll update to use the net package to perform the validation.

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -171,5 +181,97 @@ func (a *AgentClusterInstall) finish() error {
...
 	return nil
 }
+
+// Sets the default network type to OVNKubernetes if it is unspecified in the
+// AgentClusterInstall or InstallConfig
+func setNetworkType(aci *hiveext.AgentClusterInstall, installConfig *types.InstallConfig,
+	warningMessage string) {
+
+	if aci.Spec.Networking.NetworkType != "" {
+		return
```
bfournie, line 197: Just wondering about this check here. Should it not be changed if already set in AgentClusterInstall?

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -171,5 +181,97 @@ func (a *AgentClusterInstall) finish() error {
...
+				allErrs = append(allErrs, field.Required(clusterNetworkPath, "error parsing the clusterNetwork CIDR"))
+			}
+			if isIPv4(ip) {
+				hasIPv4 = true
+			}
+			if isIPv6(ip) {
+				hasIPv6 = true
+			}
+		}
+		if hasIPv6 && !hasIPv4 {
```
bfournie, line 249: Is OVN not required for dual stack?  I don't really know for sure but I can check.

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -171,5 +181,97 @@ func (a *AgentClusterInstall) finish() error {
...
+				allErrs = append(allErrs, field.Required(serviceNetworkPath, "error parsing the clusterNetwork CIDR"))
+			}
+			if isIPv4(ip) {
+				hasIPv4 = true
+			}
+			if isIPv6(ip) {
+				hasIPv6 = true
+			}
+		}
+		if hasIPv6 && !hasIPv4 {
```
bfournie, line 269: Same as above - need to use OVN for dual stack.

pkg/asset/agent/manifests/agentclusterinstall.go
rwsu: The validation returns an error if the network type is OpenShiftSDN and an IPv6 address has been entered. This should catch the dual stack case. The validation will fail if they selected OpenShiftSDN and would lead the user to change to a different network type.

pkg/asset/agent/manifests/agentclusterinstall.go
rwsu: I think it is working correctly here. If an IPv6 address is entered for the ServiceNetwork and the selected NetworkType is OpenShiftSDN, the validation will return an error.

bfournie: Right but I think it should still return an error in the dual stack case (one IPv6 address and one IPv4 address), at least that's how I interpret the InstallCfg validation. Now its only returning an error in the IPv6 only case.

rwsu: I see my blunder now. Thanks for pointing it out. Will update.

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -171,5 +181,83 @@ func (a *AgentClusterInstall) finish() error {
...
+		return
+	}
+
+	if installConfig != nil && installConfig.Networking != nil &&
+		installConfig.Networking.NetworkType != "" {
+		aci.Spec.Networking.NetworkType = installConfig.NetworkType
+		return
+	}
+
+	defaults.SetInstallConfigDefaults(installConfig)
```
andfasano, line 206: What's the reason to invoke such method here, since it seems that (potentially) could affect a completely different asset from the current one? If so, I'd find it confusing

lranjbar: My understanding is that there is a [relationship between these two settings](https://github.com/openshift-metal3/dev-scripts/blob/master/agent/docs/config-presets-and-options.md#ip-stack). I assume it is because of technical limitations but I don't know the limitations of OpenShiftSDN and OVNKubernetes.

lranjbar: My understanding is that there is a [relationship between these two settings](https://github.com/openshift-metal3/dev-scripts/blob/master/agent/docs/config-presets-and-options.md#ip-stack). I assume it is because of technical limitations but I don't know the limitations of OpenShiftSDN and OVNKubernetes.

lranjbar: Okay I did some hunting in the OCP docs. TLDR: [OpenShiftSDN](https://docs.openshift.com/container-platform/4.11/networking/openshift_sdn/about-openshift-sdn.html) doesn't support IPv6 but [OVNKubernetes](https://docs.openshift.com/container-platform/4.11/networking/ovn_kubernetes_network_provider/about-ovn-kubernetes.html) does. Maybe some other things too.. but that is the big one that sticks out to me.

lranjbar: In general... OVNKubernetes just has more features it seems. :) So if this is blank it would make sense to default to it.

camrossi:
@rwsu I am trying to use the AgentClusterInstall to deploy a Calico based cluster but I am not sure if this PR actually gives me this ability or not. 

If I set `networkType: Calico` the `openshift-install agent create image` works but 

- I have no idea on how to pass the Calico manifests to the installer
- The bootstrap fails with `Unable to retrieve cluster metadata from Agent Rest API: no clusterID known for the cluster`

From a few different places it seems it might be supported at least with ACM https://access.redhat.com/solutions/6994136 but I am really lost!

Could you point me in the right direction ?

rwsu:
Hi @camrossi. Hopefully these answers will help steer you towards the right direction.

>The bootstrap fails with `Unable to retrieve cluster metadata from Agent Rest API: no clusterID known for the cluster`

This error isn't as meaningful as it should be. I was able to reproduce that error by setting AgentClusterInstall.Spec.networking.networkType to Calico. The underlying issue is assisted-service REST-API doesn't support Calico as a networkType. So you can't specify Calico directory in the AgentClusterInstall Spec. There is a work around by the setting the networkType through the install-config override. The override is set through an annotation in AgentClusterInstall.

Here's an example:

```
apiVersion: extensions.hive.openshift.io/v1beta1
kind: AgentClusterInstall
metadata:
  annotations:
    agent-install.openshift.io/install-config-overrides: '{"networking": {"networkType": "Calico"}}'
```

> I have no idea on how to pass the Calico manifests to the installer

You can include additional manifests to be installed with the cluster by placing them in your <asset-dir>/openshift directory prior to running the create-image command:

`openshift-install agent create image --dir <asset dir>`

Here's an example of installing MCE by placing its manifests in the <asset-dir>/openshift directory.

camrossi:
Thank you @rwsu super clear! I will try to use this approach next time.

camrossi:
So I just tried but I am not managing to override the CNI... 
I create the 
```
apiVersion: extensions.hive.openshift.io/v1beta1
kind: AgentClusterInstall
metadata:
  annotations:
    agent-install.openshift.io/install-config-overrides: '{"networking": {"networkType": "Calico"}}'
```
config and placed it into `cluster-manifests\agent-cluster-install.yaml` and I can see is consumed as the file is deleted but then I still get this:

```
[root@rhel-1 install]# oc get  network.config -o yaml
apiVersion: v1
items:
- apiVersion: config.openshift.io/v1
  kind: Network
  metadata:
    creationTimestamp: "2023-02-03T13:00:38Z"
    generation: 1
    name: cluster
    resourceVersion: "551"
    uid: 8ce06ce8-46ab-4101-90ee-914bce25de31
  spec:
    clusterNetwork:
    - cidr: 10.214.0.0/16
      hostPrefix: 23
    externalIP:
      policy: {}
    networkType: OVNKubernetes <---
    serviceNetwork:
    - 172.30.0.0/16
  status: {}
kind: List
metadata:
  resourceVersion: ""

```

And the cluster brings up OVN and Caclio 😱  I guess I am doing something wrong

bfournie:
Hi @camrossi, I tried this and it seems to be working. I used the full set of ZTP manifests though - AgentClusterInstall, ClusterImageSet etc. and did not use install-config.yaml or agent-config.yaml. I'm not sure if you were doing the same thing, if not the annotations you added may have gotten overwritten.
In our dev-scripts setup I added to agent-cluster-install.yaml the metadata param:
```
  annotations:
    agent-install.openshift.io/install-config-overrides: '{"networking":{"networkType":"Calico"}}'
```
Then when I did the install I see:
```
$ oc get  network.config -o yaml
apiVersion: v1
items:
- apiVersion: config.openshift.io/v1
  kind: Network
  metadata:
    creationTimestamp: "2023-03-10T17:56:57Z"
    generation: 2
    name: cluster
    resourceVersion: "2643"
    uid: f5362c8c-5b52-47b3-b70e-87a90cdf4d4f
  spec:
    clusterNetwork:
    - cidr: 10.128.0.0/14
      hostPrefix: 24
    externalIP:
      policy: {}
    networkType: Calico
    serviceNetwork:
    - 172.30.0.0/16
  status:
    clusterNetwork:
    - cidr: 10.128.0.0/14
      hostPrefix: 24
    networkType: Calico
    serviceNetwork:
    - 172.30.0.0/16
kind: List
metadata:
  resourceVersion: ""
```
Note that this field is documented here - https://github.com/openshift/assisted-service/blob/96f0f2434fbadebf15b7edb697ec035488c3c872/docs/hive-integration/README.md#creating-installconfig-overrides
Note that the

camrossi:
Thanks @bfournie !
I didn't try with a full set oz ZTP manifests... might have to give it a try then. Perhaps that was the issue!

camrossi:
Using a full set of ZTP Manifest was the key! It worked fine!

PR #6288, opened 2022-08-31
CORS 2040: [gcp] Adding Public and Private Managed zones

barbacbd:
**Install config added the public and private managed zones under platform/gcp
**Install config validate the public and private managed zones (if provided) - when
empty or "None" the validation does not occur.
** Added tests for install config

https://issues.redhat.com/browse/CORS-2040

### DIFF
--- a/data/data/gcp/cluster/dns/base.tf
+++ b/data/data/gcp/cluster/dns/base.tf
@@ -1,8 +1,10 @@
 locals {
-  description = "Created By OpenShift Installer"
+  description       = "Created By OpenShift Installer"
+  private_zone_name = var.create_private_zone ? google_dns_managed_zone.int[0].name : var.private_zone_name
 }
 
 resource "google_dns_managed_zone" "int" {
+  count       = var.create_private_zone ? 1 : 0
   name        = "${var.cluster_id}-private-zone"
   description = local.description
   dns_name    = "${var.cluster_domain}."
@@ -18,27 +20,32 @@ resource "google_dns_managed_zone" "int" {
 }
 
 resource "google_dns_record_set" "api_external" {
-  count = var.public_endpoints ? 1 : 0
+  count = var.public_endpoints && var.create_public_zone_records ? 1 : 0
 
   name         = "api.${var.cluster_domain}."
   type         = "A"
   ttl          = "60"
-  managed_zone = var.public_dns_zone_name
+  managed_zone = var.public_zone_name
+  project      = var.public_zone_project
   rrdatas      = [var.api_external_lb_ip]
 }
 
 resource "google_dns_record_set" "api_internal" {
+  count        = var.create_private_zone_records ? 1 : 0
   name         = "api-int.${var.cluster_domain}."
   type         = "A"
   ttl          = "60"
-  managed_zone = google_dns_managed_zone.int.name
+  managed_zone = local.private_zone_name
+  project      = var.private_zone_project
   rrdatas      = [var.api_internal_lb_ip]
 }
 
 resource "google_dns_record_set" "api_external_internal_zone" {
+  count        = var.create_private_zone_records ? 1 : 0
   name         = "api.${var.cluster_domain}."
   type         = "A"
   ttl          = "60"
-  managed_zone = google_dns_managed_zone.int.name
+  managed_zone = local.private_zone_name
+  project      = var.private_zone_project
   rrdatas      = [var.api_internal_lb_ip]
 }
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -2286,10 +2286,39 @@ spec:
                       NetworkProjectID specifies which project the network and subnets
                       exist in when they are not in the main ProjectID.
                     type: string
+                  privateDNSZone:
+                    description: PrivateDNSZone Technology Preview. PrivateDNSZone
+                      contains the zone ID and project where the Private DNS zone
+                      will be created.
+                    properties:
+                      id:
+                        description: ID Technology Preview. ID or name of the zone.
+                        type: string
+                      project:
+                        description: ProjectID Technology Preview When the ProjectID
+                          is provided, the zone will be created in this project. When
+                          the ProjectID is empty, the DNS zone with this ID will be
+                          created and managed in the Service Project (GCP.ProjectID).
+                        type: string
+                    type: object
                   projectID:
                     description: ProjectID is the the project that will be used for
                       the cluster.
                     type: string
+                  publicDNSZone:
+                    description: PublicDNSZone Technology Preview. PublicDNSZone contains
+                      the zone ID and project where the Public DNS zone will be created.
+                    properties:
+                      id:
+                        description: ID Technology Preview. ID or name of the zone.
+                        type: string
+                      project:
+                        description: ProjectID Technology Preview When the ProjectID
+                          is provided, the zone will be created in this project. When
+                          the ProjectID is empty, the DNS zone with this ID will be
+                          created and managed in the Service Project (GCP.ProjectID).
+                        type: string
+                    type: object
                   region:
                     description: Region specifies the GCP region where the cluster
                       will be created.
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -375,7 +375,6 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			Data:     data,
 		})
 	case gcp.Name:
-		var publicZoneName string
 		sess, err := gcpconfig.GetSession(ctx)
 		if err != nil {
 			return err
@@ -417,18 +416,45 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		for i, w := range workers {
 			workerConfigs[i] = w.Spec.Template.Spec.ProviderSpec.Value.Object.(*machinev1beta1.GCPMachineProviderSpec)
 		}
-		if installConfig.Config.Publish == types.ExternalPublishingStrategy {
-			publicZone, err := gcpconfig.GetPublicZone(ctx, installConfig.Config.GCP.ProjectID, installConfig.Config.BaseDomain)
-			if err != nil {
-				return errors.Wrapf(err, "failed to get GCP public zone")
-			}
-			publicZoneName = publicZone.Name
-		}
 		preexistingnetwork := installConfig.Config.GCP.Network != ""
 
 		createFirewallRules := true
 		if installConfig.Config.GCP.CreateFirewallRules == gcp.CreateFirewallRulesDisabled {
 			createFirewallRules = false
+		}
+
+		// Setup defaults for public dns zone
+		createPublicZoneRecords := true
+		publicZoneName := ""
+		publicZoneProject := installConfig.Config.GCP.ProjectID
+		if installConfig.Config.GCP.PublicDNSZone != nil && installConfig.Config.GCP.PublicDNSZone.ProjectID != "" {
+			publicZoneProject = installConfig.Config.GCP.PublicDNSZone.ProjectID
+		}
+
+		switch {
+		case installConfig.Config.Publish != types.ExternalPublishingStrategy:
+			// Do not create public records when not publishing externally.
+			createPublicZoneRecords = false
+		case installConfig.Config.GCP.PublicDNSZone != nil && installConfig.Config.GCP.PublicDNSZone.ID != "":
+			publicZoneName = installConfig.Config.GCP.PublicDNSZone.ID
+		default:
+			// Search the project for a dns zone with the specified base domain.
+			publicZone, err := gcpconfig.GetPublicZone(ctx, publicZoneProject, installConfig.Config.BaseDomain)
+			if err != nil {
+				return errors.Wrapf(err, "failed to get GCP public zone")
+			}
+			publicZoneName = publicZone.Name
+		}
+
+		// Setup defaults for private DNS Zone
+		createPrivateZone := true
+		createPrivateZoneRecords := true
+		privateZoneName := ""
+		privateZoneProject := ""
+		if installConfig.Config.GCP.PrivateDNSZone != nil && installConfig.Config.GCP.PrivateDNSZone.ID != "" {
+			createPrivateZone = false
+			privateZoneName = installConfig.Config.GCP.PrivateDNSZone.ID
+			privateZoneProject = installConfig.Config.GCP.PrivateDNSZone.ProjectID
 		}
 
 		archName := coreosarch.RpmArch(string(installConfig.Config.ControlPlane.Architecture))
@@ -451,16 +477,22 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		imageURL := fmt.Sprintf("https://storage.googleapis.com/rhcos/rhcos/%s.tar.gz", img.Name)
 		data, err := gcptfvars.TFVars(
 			gcptfvars.TFVarsSources{
-				Auth:                   auth,
-				MasterConfigs:          masterConfigs,
-				WorkerConfigs:          workerConfigs,
-				CreateFirewallRules:    createFirewallRules,
-				ImageURI:               imageURL,
-				ImageLicenses:          installConfig.Config.GCP.Licenses,
-				InstanceServiceAccount: instanceServiceAccount,
-				PublicZoneName:         publicZoneName,
-				PublishStrategy:        installConfig.Config.Publish,
-				PreexistingNetwork:     preexistingnetwork,
+				Auth:                     auth,
+				MasterConfigs:            masterConfigs,
+				WorkerConfigs:            workerConfigs,
+				CreateFirewallRules:      createFirewallRules,
+				CreatePrivateZone:        createPrivateZone,
+				CreatePrivateZoneRecords: createPrivateZoneRecords,
+				CreatePublicZoneRecords:  createPublicZoneRecords,
+				ImageURI:                 imageURL,
+				ImageLicenses:            installConfig.Config.GCP.Licenses,
+				InstanceServiceAccount:   instanceServiceAccount,
+				PreexistingNetwork:       preexistingnetwork,
+				PrivateZoneName:          privateZoneName,
+				PrivateZoneProject:       privateZoneProject,
+				PublicZoneName:           publicZoneName,
+				PublicZoneProject:        publicZoneProject,
+				PublishStrategy:          installConfig.Config.Publish,
 			},
 		)
 		if err != nil {
--- a/pkg/asset/installconfig/gcp/client.go
+++ b/pkg/asset/installconfig/gcp/client.go
@@ -23,6 +23,7 @@ type API interface {
 	GetMachineType(ctx context.Context, project, zone, machineType string) (*compute.MachineType, error)
 	GetPublicDomains(ctx context.Context, project string) ([]string, error)
 	GetPublicDNSZone(ctx context.Context, project, baseDomain string) (*dns.ManagedZone, error)
+	GetDNSZoneByName(ctx context.Context, project, zoneName string) (*dns.ManagedZone, error)
 	GetSubnetworks(ctx context.Context, network, project, region string) ([]*compute.Subnetwork, error)
 	GetProjects(ctx context.Context) (map[string]string, error)
 	GetRegions(ctx context.Context, project string) ([]string, error)
@@ -109,6 +110,24 @@ func (c *Client) GetPublicDomains(ctx context.Context, project string) ([]string
 		return publicZones, err
 	}
 	return publicZones, nil
+}
+
+// GetDNSZoneByName returns a DNS zone matching the `zoneName` if the DNS zone exists
+// and can be seen (correct permissions for a private zone) in the project.
+func (c *Client) GetDNSZoneByName(ctx context.Context, project, zoneName string) (*dns.ManagedZone, error) {
+	ctx, cancel := context.WithTimeout(context.TODO(), 1*time.Minute)
+	defer cancel()
+
+	svc, err := c.getDNSService(ctx)
+	if err != nil {
+		return nil, err
+	}
+	returnedZone, err := svc.ManagedZones.Get(project, zoneName).Context(ctx).Do()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get DNS Zones")
+	}
+	return returnedZone, nil
+
 }
 
 // GetPublicDNSZone returns a public DNS zone for a basedomain.
--- a/pkg/asset/installconfig/gcp/validation.go
+++ b/pkg/asset/installconfig/gcp/validation.go
@@ -45,6 +45,8 @@ func Validate(client API, ic *types.InstallConfig) error {
 	allErrs = append(allErrs, validateNetworkProject(client, ic, field.NewPath("platform").Child("gcp"))...)
 	allErrs = append(allErrs, validateRegion(client, ic, field.NewPath("platform").Child("gcp"))...)
 	allErrs = append(allErrs, validateNetworks(client, ic, field.NewPath("platform").Child("gcp"))...)
+	allErrs = append(allErrs, validateZoneProjects(client, ic, field.NewPath("platform").Child("gcp"))...)
+	allErrs = append(allErrs, validateManagedZones(client, ic, field.NewPath("platform").Child("gcp"))...)
 	allErrs = append(allErrs, validateInstanceTypes(client, ic)...)
 	allErrs = append(allErrs, validateCredentialMode(client, ic)...)
 
@@ -108,6 +110,62 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList 
 		if compute.Platform.GCP != nil && compute.Platform.GCP.InstanceType != "" {
 			allErrs = append(allErrs, ValidateInstanceType(client, fieldPath.Child("platform", "gcp"), ic.GCP.ProjectID, zones[0].Name,
 				compute.Platform.GCP.InstanceType, computeReq)...)
+		}
+	}
+
+	return allErrs
+}
+
+// validateZoneProjects will validate the public and private zone projects when provided
+func validateZoneProjects(client API, ic *types.InstallConfig, fieldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	projects, err := client.GetProjects(context.TODO())
+	if err != nil {
+		return append(allErrs, field.InternalError(fieldPath.Child("project"), err))
+	}
+
+	// If the PublicZoneProject is empty, the value will default to ProjectID, and it won't be checked here
+	if ic.GCP.PublicDNSZone != nil && ic.GCP.PublicDNSZone.ProjectID != "" {
+		if _, found := projects[ic.GCP.PublicDNSZone.ProjectID]; !found {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("PublicDNSZone").Child("ProjectID"), ic.GCP.PublicDNSZone.ProjectID, "invalid public zone project"))
+		}
+	}
+
+	if ic.GCP.PrivateDNSZone != nil && ic.GCP.PrivateDNSZone.ProjectID != "" {
+		if _, found := projects[ic.GCP.PrivateDNSZone.ProjectID]; !found {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("PrivateDNSZone").Child("ProjectID"), ic.GCP.PrivateDNSZone.ProjectID, "invalid private zone project"))
+		}
+	}
+
+	return allErrs
+}
+
+// validateManagedZones will validate the public and private managed zones if they exist.
+func validateManagedZones(client API, ic *types.InstallConfig, fieldPath *field.Path) field.ErrorList {
+	commonZoneLookup := func(dnsZone, zoneProject, defaultProject string) error {
+		project := zoneProject
+		if project == "" {
+			project = defaultProject
+		}
+
+		if _, err := client.GetDNSZoneByName(context.TODO(), project, dnsZone); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	allErrs := field.ErrorList{}
+
+	if ic.GCP.PublicDNSZone != nil && ic.GCP.PublicDNSZone.ID != "" {
+		if err := commonZoneLookup(ic.GCP.PublicDNSZone.ID, ic.GCP.PublicDNSZone.ProjectID, ic.GCP.ProjectID); err != nil {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("PublicDNSZone").Child("ID"), ic.GCP.PublicDNSZone.ID, "invalid public managed zone"))
+		}
+	}
+
+	if ic.GCP.PrivateDNSZone != nil && ic.GCP.PrivateDNSZone.ID != "" {
+		if err := commonZoneLookup(ic.GCP.PrivateDNSZone.ID, ic.GCP.PrivateDNSZone.ProjectID, ic.GCP.ProjectID); err != nil {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("PrivateDNSZone").Child("ID"), ic.GCP.PrivateDNSZone.ID, "invalid private managed zone"))
 		}
 	}
 
--- a/pkg/asset/installconfig/gcp/validation_test.go
+++ b/pkg/asset/installconfig/gcp/validation_test.go
@@ -31,6 +31,8 @@ var (
 	validCPSubnet      = "valid-controlplane-subnet"
 	validCIDR          = "10.0.0.0/16"
 	validClusterName   = "valid-cluster"
+	validPublicZone    = "valid-short-public-zone"
+	invalidPublicZone  = "invalid-short-public-zone"
 
 	invalidateMachineCIDR = func(ic *types.InstallConfig) {
 		_, newCidr, _ := net.ParseCIDR("192.168.111.0/24")
@@ -70,6 +72,15 @@ var (
 	removeVPC                = func(ic *types.InstallConfig) { ic.GCP.Network = "" }
 	removeSubnets            = func(ic *types.InstallConfig) { ic.GCP.ComputeSubnet, ic.GCP.ControlPlaneSubnet = "", "" }
 	invalidClusterName       = func(ic *types.InstallConfig) { ic.ObjectMeta.Name = "testgoogletest" }
+	validManagedZone         = func(ic *types.InstallConfig) {
+		ic.GCP.PublicDNSZone.ID, ic.GCP.PublicDNSZone.ProjectID = validPublicZone, validProjectName
+	}
+	invalidManagedZone = func(ic *types.InstallConfig) {
+		ic.GCP.PublicDNSZone.ID, ic.GCP.PublicDNSZone.ProjectID = invalidPublicZone, validPublicZone
+	}
+	invalidZoneProject = func(ic *types.InstallConfig) {
+		ic.GCP.PublicDNSZone.ID, ic.GCP.PublicDNSZone.ProjectID = validPublicZone, invalidProjectName
+	}
 
 	machineTypeAPIResult = map[string]*compute.MachineType{
 		"n1-standard-1": {GuestCpus: 1, MemoryMb: 3840},
@@ -107,6 +118,14 @@ func validInstallConfig() *types.InstallConfig {
 				Network:                validNetworkName,
 				ComputeSubnet:          validComputeSubnet,
 				ControlPlaneSubnet:     validCPSubnet,
+				PublicDNSZone: &gcp.DNSZone{
+					ID:        validPublicZone,
+					ProjectID: validProjectName,
+				},
+				PrivateDNSZone: &gcp.DNSZone{
+					ID:        validPublicZone,
+					ProjectID: validProjectName,
+				},
 			},
 		},
 		ControlPlane: &types.MachinePool{
@@ -255,6 +274,23 @@ func TestGCPInstallConfigValidation(t *testing.T) {
 			expectedError:  true,
 			expectedErrMsg: "platform.gcp.region: Invalid value: \"us-east4\": invalid region",
 		},
+		{
+			name:          "Managed zone validated",
+			edits:         editFunctions{validManagedZone},
+			expectedError: false,
+		},
+		{
+			name:           "Valid project & invalid zone",
+			edits:          editFunctions{invalidManagedZone},
+			expectedError:  true,
+			expectedErrMsg: fmt.Sprintf("platform.gcp.PublicDNSZone.ID: Invalid value: \"%s\": invalid public managed zone", invalidPublicZone),
+		},
+		{
+			name:           "Valid zone & invalid project",
+			edits:          editFunctions{invalidZoneProject},
+			expectedError:  true,
+			expectedErrMsg: fmt.Sprintf("platform.gcp.PublicDNSZone.ProjectID: Invalid value: \"%s\": invalid public zone project", invalidProjectName),
+		},
 	}
 	mockCtrl := gomock.NewController(t)
 	defer mockCtrl.Finish()
@@ -295,6 +331,10 @@ func TestGCPInstallConfigValidation(t *testing.T) {
 
 	// Return fake credentials when asked
 	gcpClient.EXPECT().GetCredentials().Return(&googleoauth.Credentials{JSON: []byte("fake creds")}).AnyTimes()
+
+	// Expected results for the managed zone tests
+	gcpClient.EXPECT().GetDNSZoneByName(gomock.Any(), gomock.Any(), validPublicZone).Return(&dns.ManagedZone{}, nil).AnyTimes()
+	gcpClient.EXPECT().GetDNSZoneByName(gomock.Any(), gomock.Any(), invalidPublicZone).Return(nil, fmt.Errorf("no matching DNS Zone found")).AnyTimes()
 
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
--- a/pkg/asset/manifests/dns.go
+++ b/pkg/asset/manifests/dns.go
@@ -36,6 +36,10 @@ import (
 
 var (
 	dnsCfgFilename = filepath.Join(manifestDir, "cluster-dns-02-config.yml")
+
+	combineGCPZoneInfo = func(project, zoneName string) string {
+		return fmt.Sprintf("project/%s/managedZones/%s", project, zoneName)
+	}
 )
 
 // DNS generates the cluster-dns-*.yml files.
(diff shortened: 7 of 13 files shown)


pkg/asset/installconfig/gcp/dns.go
```diff
@@ -29,6 +31,49 @@ func GetPublicZone(ctx context.Context, project, baseDomain string) (*dns.Manage
...
+	if err != nil {
+		return "", "", err
+	} else {
+		if matched {
+			project, zoneName = splitZoneName(zoneName)
+		}
+	}
+
+	return project, zoneName, nil
+}
```
jstuever, line 56: The "https://" would be a URL as opposed to a "full resource name". Most users do not interact with the resource URLs... so I hesitate to use them here.

https://cloud.google.com/apis/design/resource_names#resource_name_vs_url

However, I don't see where DNS resources are defined to have full resource names.

https://cloud.google.com/iam/docs/full-resource-names

Perhaps we should instead expect a partial resource name. This removes the need for us to assume the format of a full resource name for a managedZone.

projects/project-id/managedZones/managed-zone-name

We should ensure the name matches this pattern precisely (has both the "projects/" and "/managedZones/" sections) before assuming it is a partial resource name. Otherwise, we just send the entire string to the gcp api for validation as-is.

pkg/asset/manifests/dns.go
```diff
@@ -134,14 +134,37 @@ func (d *DNS) Generate(dependencies asset.Parents) error {
...
+				config.Spec.PublicZone = &configv1.DNSZone{ID: zone.Name}
+			}
+		}
+
+		// Set the private zone
+		if installConfig.Config.GCP.PrivateManagedZone == "" {
+			config.Spec.PrivateZone = &configv1.DNSZone{ID: fmt.Sprintf("%s-private-zone", clusterID.InfraID)}
+		} else if !strings.EqualFold(installConfig.Config.GCP.PrivateManagedZone, "None") {
+			// When "None" manifest data is NOT set
+			_, zoneName, err := icgcp.ParseZone(installConfig.Config.Platform.GCP.ProjectID, installConfig.Config.GCP.PrivateManagedZone)
```
jstuever, line 161: No need to parse here, let's pass the `installConfig.Config.GCP.PrivateManagedZone` into the manifest
```
config.Spec.PublicZone = &configv1.DNSZone{ID: installConfig.Config.GCP.PrivateManagedZone}
```

pkg/asset/installconfig/gcp/dns.go
```diff
@@ -29,6 +32,50 @@ func GetPublicZone(ctx context.Context, project, baseDomain string) (*dns.Manage
...
+			if len(splitZoneName) < 4 {
+				return project, zoneName, fmt.Errorf("invalid zone format: %s", zoneName)
+			} else {
+				return splitZoneName[1], splitZoneName[3], nil
+			}
+		}
+	}
+
+	return project, zoneName, nil
+}
```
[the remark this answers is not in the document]
jstuever: Let's not pass a default project here... we can handle what to do when project is not found in this string outside of this function.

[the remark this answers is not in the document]
jstuever: And finally, let's take advantage of fmt.sscanf() to simplify things a bit...
```
func ParseZone(providedZoneName string) (string, string, error) {
	var zoneName string
	var project string
	found, err := fmt.Sscanf(providedZoneName, "projects/%s/managedZones/%s, &project, &zoneName)
	if err != nil {
		return null, null, err
	}
	if found != 2 {
		return nil, zone, nil
	}
	return project, zoneid, null
}
```

pkg/asset/installconfig/gcp/dns.go
```diff
@@ -29,6 +32,50 @@ func GetPublicZone(ctx context.Context, project, baseDomain string) (*dns.Manage
...
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+	defer cancel()
+
+	dnsZone, err := client.GetDNSZoneByName(ctx, project, zoneName)
+	if err != nil {
+		return nil, err
+	}
+
+	return dnsZone, nil
+}
```
jstuever, line 77: GetZoneByName will no longer be needed if we aren't parsing the zones in the manifests generation.

pkg/asset/installconfig/gcp/client.go
```diff
@@ -109,6 +110,40 @@ func (c *Client) GetPublicDomains(ctx context.Context, project string) ([]string
 	return publicZones, nil
 }
 
+// GetDNSZoneByName returns a DNS zone matching the `zoneName` if the DNS zone exists
+// and can be seen (correct permissions for a private zone) in the project.
+func (c *Client) GetDNSZoneByName(ctx context.Context, defaultProject, providedZoneName string) (*dns.ManagedZone, error) {
+	ctx, cancel := context.WithTimeout(context.TODO(), 1*time.Minute)
+	defer cancel()
+
+	project, zoneName, err := ParseZone(defaultProject, providedZoneName)
```
jstuever, line 119: If we remove the defaultProject argument, this would need to change
```
	project, zoneName, err := ParseZoneName(providedZoneName)
	if err != nil {
		return nil, err
	}
	if project == nil or project == "" {
		project = defaultProject
	}
```

data/data/gcp/cluster/dns/base.tf
```diff
@@ -18,7 +19,7 @@ resource "google_dns_managed_zone" "int" {
 }
 
 resource "google_dns_record_set" "api_external" {
-  count = var.public_endpoints ? 1 : 0
+  count = var.public_endpoints && ! var.skip_public_zone_records ? 1 : 0
```
jstuever, line 22: Do we need to specify the project here?

data/data/gcp/cluster/dns/base.tf
```diff
@@ -28,17 +29,19 @@ resource "google_dns_record_set" "api_external" {
 }
 
 resource "google_dns_record_set" "api_internal" {
+  count        = var.skip_private_zone_records ? 0 : 1
   name         = "api-int.${var.cluster_domain}."
   type         = "A"
   ttl          = "60"
-  managed_zone = google_dns_managed_zone.int.name
+  managed_zone = var.skip_private_zone_creation ? var.private_dns_zone_name : google_dns_managed_zone.int[0].name
```
jstuever, line 36: Project?

data/data/gcp/cluster/dns/base.tf
```diff
@@ -28,17 +29,19 @@ resource "google_dns_record_set" "api_external" {
...
   rrdatas      = [var.api_internal_lb_ip]
 }
 
 resource "google_dns_record_set" "api_external_internal_zone" {
+  count        = var.skip_private_zone_records ? 0 : 1
   name         = "api.${var.cluster_domain}."
   type         = "A"
   ttl          = "60"
-  managed_zone = google_dns_managed_zone.int.name
+  managed_zone = var.skip_private_zone_creation ? var.private_dns_zone_name : google_dns_managed_zone.int[0].name
```
jstuever, line 45: Project?

pkg/asset/installconfig/gcp/dns.go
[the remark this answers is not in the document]
jstuever: I'm going to call myself out here... as sscanf() requires space separated values, which is not what we have here.... so that's not going to work.

pkg/asset/installconfig/gcp/dns.go
```diff
@@ -29,6 +31,21 @@ func GetPublicZone(ctx context.Context, project, baseDomain string) (*dns.Manage
...
+// ParseZoneName will attempt to extract the project and zone from a long zone name that
+// includes the full path (i.e. projects/{project-id}/managedZones/{managed-zone-name}). If the data
+// cannot be parsed or does not include this information, the default (provided) values will be returned
+func ParseZoneName(providedZoneName string) (string, string, error) {
+
+	switch splitZoneName := strings.Split(providedZoneName, "/"); {
+	case len(splitZoneName) == 1:
+		return "", providedZoneName, nil
+	case len(splitZoneName) >= 4:
+		return splitZoneName[1], splitZoneName[3], nil
```
jstuever, line 43: What happens if the name is "fu/project-id/bar/managed-zone-name"? I recommend a slight tweak to the logic here...
```
case splitZoneName[0] == "projects" && splitZoneName[2] == "managedZones"
	return splitZoneName[1], splitZoneName[3], nil
```

data/data/gcp/cluster/dns/base.tf
```diff
@@ -18,27 +20,32 @@ resource "google_dns_managed_zone" "int" {
...
 
 resource "google_dns_record_set" "api_external" {
-  count = var.public_endpoints ? 1 : 0
+  count = var.public_endpoints && var.create_public_zone_records ? 1 : 0
 
   name         = "api.${var.cluster_domain}."
   type         = "A"
   ttl          = "60"
   managed_zone = var.public_dns_zone_name
+  project      = var.network_project_id == "" ? var.project_id : var.network_project_id
```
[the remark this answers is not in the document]
jstuever: FYI... when `project` is set to an empty string, it will default to the provider specified project.... so using empty strings to fall through to the "service project" will work when public_dns_zone_project and/or private_dns_zone_project are not specified.

jstuever (changes requested):
Two minor tweaks.... 1) I think we need to be very prescribed with which projects the terraform dns records exist, and 2) I think we should be more prescribed with the parsing of the relative path managedZone names.

jstuever (changes requested):
There are some functional changes needed here.... and a couple of nit pick items.

pkg/asset/installconfig/gcp/client.go
```diff
@@ -109,6 +110,42 @@ func (c *Client) GetPublicDomains(ctx context.Context, project string) ([]string
...
+// GetDNSZoneByName returns a DNS zone matching the `zoneName` if the DNS zone exists
+// and can be seen (correct permissions for a private zone) in the project.
+func (c *Client) GetDNSZoneByName(ctx context.Context, defaultProject, providedZoneName string) (*dns.ManagedZone, error) {
+	ctx, cancel := context.WithTimeout(context.TODO(), 1*time.Minute)
+	defer cancel()
+
+	project, zoneName, err := ParseZoneName(providedZoneName)
+	if err != nil {
+		return nil, err
+	} else if project == "" {
```
jstuever, line 122: There is no need to do an `else` here because the prior if is a return. You can flatten this and start a fresh `if project =="" {` on the next line.

```
if err != nil {
    return nil, err
}

if project == "" {
    project = defaultProject
}
```

pkg/asset/installconfig/gcp/dns.go
```diff
@@ -29,6 +31,21 @@ func GetPublicZone(ctx context.Context, project, baseDomain string) (*dns.Manage
...
 
+// ParseZoneName will attempt to extract the project and zone from a long zone name that
+// includes the full path (i.e. projects/{project-id}/managedZones/{managed-zone-name}). If the data
+// cannot be parsed or does not include this information, the default (provided) values will be returned
+func ParseZoneName(providedZoneName string) (string, string, error) {
+
+	switch splitZoneName := strings.Split(providedZoneName, "/"); {
+	case len(splitZoneName) == 1:
+		return "", providedZoneName, nil
+	case len(splitZoneName) >= 4 && splitZoneName[0] == "projects" && splitZoneName[2] == "managedZones":
```
jstuever, line 42: In this case, len(splitZoneName) can only be 4.... anything else would have gibberish on the end, and could be considered invalid.

jstuever (changes requested):
Looking great! Just a few more minor changes.

pkg/asset/installconfig/gcp/dns.go
barbacbd: It could be 3 couldn't it. We don't force this check in a specific sequence. What if someone entered something bad here like `projects/<project-id>/managedZones`. I hope they don't do this, but if they do then the splitZoneName[3] would cause an index out of bounds. 

Or am I just overthinking and adding too much error handling?

jstuever: Right, 3 would be too little... because there needs to be 4 segments... (where splitZoneName[3] is the 4th one).

jstuever:
```
case len(splitZoneName) == 4 && splitZoneName[0] == "projects" && splitZoneName[2] == "managedZones":
```

jstuever (approved):
/approve

patrickdillon (commented):
I have done a first pass reviewing. There's a lot going on here, so I would like to focus on `privateManagedZone: None` and hope a discussion of that may simplify subsequent discussions. The implementation here is that this setting will skip entirely creation of the private DNS zone.  I have a couple of concerns:

1. The API to me seems overloaded: the expected value is an existing hosted zone name which we use in the install, unless it's the magic string "None" in which case we skip private zones altogether. These seem to be different functionalities. Why not use a separate field for the latter use case?
2. More importantly, if we skip private zones (when the value is "None"). The install will fail. Always. Right? 

Here's my understanding of what we're doing: help me if I'm missing something. The issue we're trying to address is the case where we don't have permissions to create the zones in, I guess, either project. But, at the moment, I don't see the proposed solution.

A potential solution to this issue would be the one we have proposed/discussed for external DNS, which is to use on-cluster (baremetal & friends) DNS. I suggest for this effort, CORS-2040, we remove the idea of skipping the creation of private DNS zones (i.e. don't accept `None`). We can handle this use case in a way that aligns with whatever API we decide for on-cluster DNS.

If there is another path forward that would allow a successful install when skipping private DNS zones we could explore that too, but I'm not aware of one.

data/data/gcp/cluster/dns/base.tf
```diff
@@ -18,27 +19,32 @@ resource "google_dns_managed_zone" "int" {
...
   rrdatas      = [var.api_external_lb_ip]
 }
 
 resource "google_dns_record_set" "api_internal" {
+  count        = var.create_private_zone_records ? 1 : 0
   name         = "api-int.${var.cluster_domain}."
   type         = "A"
   ttl          = "60"
-  managed_zone = google_dns_managed_zone.int.name
+  managed_zone = var.create_private_zone ? google_dns_managed_zone.int[0].name : var.private_zone_name
```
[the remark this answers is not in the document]
barbacbd: If its None it would not have been created. The `count` should have been controlled before the managed_zone would ever be used.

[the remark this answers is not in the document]
barbacbd: The zone name already had to be parsed and validated before being passed to terraform. Does it make sense to do the same work again ?

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -2248,10 +2248,24 @@ spec:
                       specifies which project the network and subnets exist in when
                       they are not in the main ProjectID.
                     type: string
+                  privateManagedZone:
+                    description: PrivateManagedZone is currently unsupported. The PrivateManagedZone should be
+                      provided as "projects/{project-id}/managedZones/{managed-zone-name}". When the PrivateManagedZone
+                      is empty, the DNS zones will be created and managed in the Service Project (ProjectID). When
+                      "None", the DNS zones will be externally managed. When PrivateManagedZone is provided, the DNS
```
sadasu: Can we make "None" as "External" ? I feel that the semantics of an empty string and "None" are quite different but the values could be conflated easily while providing configuration. Using "external" makes the intent specific.

pkg/asset/installconfig/gcp/client.go
```diff
@@ -109,6 +110,40 @@ func (c *Client) GetPublicDomains(ctx context.Context, project string) ([]string
...
+	svc, err := c.getDNSService(ctx)
+	if err != nil {
+		return nil, err
+	}
+	req := svc.ManagedZones.List(project).Context(ctx)
+	var res *dns.ManagedZone
+	if err := req.Pages(ctx, func(page *dns.ManagedZonesListResponse) error {
+		for idx, v := range page.ManagedZones {
+			if v.Name == zoneName {
+				res = page.ManagedZones[idx]
```
sadasu, line 137: Can we break out of the for loop here after we found the matching ManagedZone?

sadasu:
/retitle CORS 2040: [gcp] Adding Public and Private Managed zones

data/data/gcp/cluster/dns/base.tf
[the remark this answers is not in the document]
barbacbd: Public zone name is always resolved in go and passed to terraform.

jstuever (approved):
/approve

pkg/asset/cluster/tfvars.go
```diff
@@ -400,15 +399,31 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
 		}
+
+		publicZoneName := ""
+		publicZoneProject := ""
 		if installConfig.Config.Publish == types.ExternalPublishingStrategy {
-			publicZone, err := gcpconfig.GetPublicZone(ctx, installConfig.Config.GCP.ProjectID, installConfig.Config.BaseDomain)
-			if err != nil {
-				return errors.Wrapf(err, "failed to get GCP public zone")
+
+			if installConfig.Config.GCP.PublicManagedZone == "" {
```
sadasu, line 407: Are we handling the case when installConfig.Config.GCP.PublicManagedZone == "None" ?

pkg/asset/cluster/tfvars.go
```diff
@@ -400,15 +399,31 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
+				publicZone, err := gcpconfig.GetPublicZone(ctx, installConfig.Config.GCP.ProjectID, installConfig.Config.BaseDomain)
+				if err != nil {
+					return errors.Wrapf(err, "failed to get GCP public zone")
+				}
+				publicZoneName = publicZone.Name
+			} else {
```
sadasu, line 413: Is this code taking care of the case where installConfig.Config.GCP.PublicManagedZone == "None"?

pkg/asset/cluster/tfvars.go
jstuever:
```
ERROR                                              
ERROR Error: Error retrieving record sets for "None": googleapi: Error 404: The 'parameters.managedZone' resource named 'None' does not exist., notFound 
ERROR                                              
ERROR   with module.dns.google_dns_record_set.api_external[0], 
ERROR   on dns/base.tf line 22, in resource "google_dns_record_set" "api_external": 
ERROR   22: resource "google_dns_record_set" "api_external" { 
ERROR                                              
ERROR failed to fetch Cluster: failed to generate asset "Cluster": failure applying terraform for "cluster" stage: failed to create cluster: failed to apply Terraform: exit status 1 
ERROR                                              
ERROR Error: Error retrieving record sets for "None": googleapi: Error 404: The 'parameters.managedZone' resource named 'None' does not exist., notFound 
ERROR                                              
ERROR   with module.dns.google_dns_record_set.api_external[0], 
ERROR   on dns/base.tf line 22, in resource "google_dns_record_set" "api_external": 
ERROR   22: resource "google_dns_record_set" "api_external" { 
ERROR                                              
ERROR 
```

jstuever: It also appears that using `privateManagedZone: None` didn't affect the terraform as expected. The zone was still created, as were the records.

barbacbd: The error above is point to a use for publicManagedZone

```
resource "google_dns_record_set" "api_external" {
  count = var.public_endpoints && var.create_public_zone_records ? 1 : 0

  name         = "api.${var.cluster_domain}."
  type         = "A"
  ttl          = "60"
  managed_zone = var.public_zone_name
  project      = var.public_zone_project
  rrdatas      = [var.api_external_lb_ip]
}
```

This was controlled in tfvars by 

```				
CreatePublicZoneRecords: !strings.EqualFold(installConfig.Config.GCP.PublicManagedZone, gcpconfig.SkipZoneCreationFlag),
```

Which should have controlled the count above and set to 0 since this was false. If count=0 is still attempting to resolve all other variables then this could be an issue.

jstuever: Ah... it's the omitempty on the tfvars... a boolean that is false is omitted. And, since the Terraform variables default to true, they become true in the variable's absence.

pkg/tfvars/gcp/gcp.go
```diff
@@ -21,41 +21,53 @@ type Auth struct {
...
+	ImageURI                 string   `json:"gcp_image_uri,omitempty"`
+	Image                    string   `json:"gcp_image,omitempty"`
+	PreexistingImage         bool     `json:"gcp_preexisting_image"`
+	InstanceServiceAccount   string   `json:"gcp_instance_service_account,omitempty"`
+	ImageLicenses            []string `json:"gcp_image_licenses,omitempty"`
+	VolumeType               string   `json:"gcp_master_root_volume_type"`
+	VolumeSize               int64    `json:"gcp_master_root_volume_size"`
+	VolumeKMSKeyLink         string   `json:"gcp_root_volume_kms_key_link"`
+	CreatePrivateZone        bool     `json:"gcp_create_private_zone,omitempty"`
+	CreatePrivateZoneRecords bool     `json:"gcp_create_private_zone_records,omitempty"`
```
jstuever, line 39: omitempty causes false values to be omitted.

pkg/tfvars/gcp/gcp.go
```diff
@@ -21,41 +21,53 @@ type Auth struct {
...
+	PrivateZoneName          string   `json:"gcp_private_zone_name,omitempty"`
+	PrivateZoneProject       string   `json:"gcp_private_zone_project,omitempty"`
+	CreatePublicZoneRecords  bool     `json:"gcp_create_public_zone_records,omitempty"`
```
jstuever, line 42: omitempty causes false values to be omitted.

pkg/tfvars/gcp/gcp.go
```diff
@@ -21,41 +21,53 @@ type Auth struct {
...
+	MasterAvailabilityZones  []string `json:"gcp_master_availability_zones"`
+	ImageURI                 string   `json:"gcp_image_uri,omitempty"`
+	Image                    string   `json:"gcp_image,omitempty"`
+	PreexistingImage         bool     `json:"gcp_preexisting_image"`
+	InstanceServiceAccount   string   `json:"gcp_instance_service_account,omitempty"`
+	ImageLicenses            []string `json:"gcp_image_licenses,omitempty"`
+	VolumeType               string   `json:"gcp_master_root_volume_type"`
+	VolumeSize               int64    `json:"gcp_master_root_volume_size"`
+	VolumeKMSKeyLink         string   `json:"gcp_root_volume_kms_key_link"`
+	CreatePrivateZone        bool     `json:"gcp_create_private_zone,omitempty"`
```
jstuever, line 38: omitempty causes false values to be omitted.

jstuever:
Let's see the e2e-gcp-ovn pass, then we can remove the hold.

jstuever:
/test e2e-gcp-ovn
I think it's just the normal flakes....

pkg/types/gcp/platform.go
```diff
@@ -59,4 +64,20 @@ type Platform struct {
 	// such as the current env OPENSHIFT_INSTALL_OS_IMAGE_OVERRIDE
 	// +optional
 	Licenses []string `json:"licenses,omitempty"`
+
+	// PrivateDNSZone Technology Preview.
+	// The PrivateDNSZone should be provided as "projects/{project-id}/managedZones/{managed-zone-name}".
+	// When the PrivateDNSZone is empty, the DNS zones will be created and managed in the Service
+	// Project (ProjectID). When "None", the DNS zones will be externally managed. When PrivateDNSZone
```
r4f4, line 71:
```suggestion
	// Project (ProjectID). When "Unmanaged", the DNS zones will be externally managed. When PrivateDNSZone
```
?

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -2275,10 +2275,28 @@ spec:
                       NetworkProjectID specifies which project the network and subnets
                       exist in when they are not in the main ProjectID.
                     type: string
+                  privateDNSZone:
+                    description: PrivateDNSZone Technology Preview. The PrivateDNSZone
+                      should be provided as "projects/{project-id}/managedZones/{managed-zone-name}".
+                      When the PrivateDNSZone is empty, the DNS zones will be created
+                      and managed in the Service Project (ProjectID). When "None",
```
sadasu, line 2282: We are checking if privateDNSZone is set to gcp.DNSZoneDisabled which is defined by the constant 
DNSZoneDisabled string = "Unmanaged".

So, should the description here be updated to reflect that?

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -2275,10 +2275,28 @@ spec:
...
+                    type: string
                   projectID:
                     description: ProjectID is the the project that will be used for
                       the cluster.
                     type: string
+                  publicDNSZone:
+                    description: PublicDNSZone Technology Preview. The PublicDNSZone
+                      should be provided as "projects/{project-id}/managedZones/{managed-zone-name}".
+                      When the PublicDNSZone is empty, the DNS zones will be created
+                      and managed in the Service Project (ProjectID). When "None",
```
sadasu, line 2295: See https://github.com/openshift/installer/pull/6288/files#r973002732.

pkg/asset/installconfig/gcp/validation_test.go
```diff
@@ -20,16 +20,20 @@ import (
...
+	invalidRegion          = "us-east4"
+	validZone              = "us-east1-b"
+	validComputeSubnet     = "valid-compute-subnet"
+	validCPSubnet          = "valid-controlplane-subnet"
+	validCIDR              = "10.0.0.0/16"
+	validClusterName       = "valid-cluster"
+	validShortPublicZone   = "valid-short-public-zone"
+	invalidShortPublicZone = "invalid-short-public-zone"
+	validLongPublicZone    = fmt.Sprintf("projects/%s/managedZones/%s", validProjectName, validShortPublicZone)
+	invalidLongPublicZone  = fmt.Sprintf("projects/%s/managedZones/%s", validProjectName, invalidShortPublicZone)
```
sadasu, line 36: Can we add a test where the LongPublicZone format is wrong?

pkg/types/gcp/platform.go
```diff
@@ -59,4 +64,20 @@ type Platform struct {
...
+	// is provided, the DNS zone will be created and managed in the Project where it was created.
+	// +optional
+	PrivateDNSZone string `json:"privateDNSZone,omitempty"`
+
+	// PublicDNSZone Technology Preview.
+	// The PublicDNSZone should be provided as "projects/{project-id}/managedZones/{managed-zone-name}".
```
patrickdillon, line 77: Having users pass the DNS zone in this format does not seem ergonomic to me; and is also inconsistent with other parts of the installconfig, like:
https://github.com/openshift/installer/blob/36e5ac36f6f1b796fdf94a3d4bef384e271d87bb/pkg/types/gcp/machinepools.go#L116-L132

I would suggest we have users pass just the managed-zone name with an optional project-id (which defaults to ic.GCP.ProjectID`). Then you can construct this formatted string when you need it.

This will allow you to get rid of the `ParseZoneName` function.

pkg/types/gcp/platform.go
```diff
@@ -11,6 +11,11 @@ const (
 	CreateFirewallRulesDisabled CreateFirewallRules = "Disabled"
 )
 
+const (
+	// DNSZoneDisabled is for when there is no DNS zone to manage.
+	DNSZoneDisabled string = "Unmanaged"
```
patrickdillon, line 16: This is jarring to me... why not either:
```go
	DNSZoneUnmanaged string = "Unmanaged"
```
or

```go
	DNSZoneDisabled string = "Disabled"
```

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -113,6 +115,26 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
 	return allErrs
 }
 
+// validateManagedZones will validate the public and private managed zones if they exist.
+func validateManagedZones(client API, ic *types.InstallConfig, fieldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if ic.GCP.PublicDNSZone != "" && !strings.EqualFold(ic.GCP.PublicDNSZone, gcp.DNSZoneDisabled) {
+		if _, err := client.GetDNSZoneByName(context.TODO(), ic.GCP.ProjectID, ic.GCP.PublicDNSZone); err != nil {
```
patrickdillon, line 123: Couldn't this DNS zone be in a different project other than ic.GCP.ProjectID?

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -113,6 +115,26 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("publicDNSZone"), ic.GCP.PublicDNSZone, "invalid public managed zone"))
+		}
+	}
+
+	if ic.GCP.PrivateDNSZone != "" && !strings.EqualFold(ic.GCP.PrivateDNSZone, gcp.DNSZoneDisabled) {
+		if _, err := client.GetDNSZoneByName(context.TODO(), ic.GCP.ProjectID, ic.GCP.PrivateDNSZone); err != nil {
```
patrickdillon, line 129: Couldn't this DNS zone be in a different project other than ic.GCP.ProjectID?

pkg/asset/installconfig/gcp/validation_test.go
patrickdillon, line 36: I am gathering from this that we are allowing users to pass strings in two formats. I think this is another example of where it would be better solved by structuring the install config field as an object to accept a zone name and optional project id.

patrickdillon (commented):
I took a pass through this, but my brain is a bit fried here at the end of the week. Still, I think I can give some substantive feedback, but I don't want to give the impression that this is a thorough review.

First it would be easier to review this if it were broken into logical commits, perhaps a structure like:
- add types
- edit dns manifests
- tfvars
- tf configs

Also, I highly recommend you structure the user inputs into the install config, something like:


```go
 type DNSZone struct { 

 	Name string `json:"name"` 
  
 	Project string `json:"project,omitempty"`
```

Structured data is always easier to work with, for both us in the code and the users. I think this would simplify some of the issues you've been working through in the code.

pkg/types/gcp/platform.go
sadasu: This is a great discussion to have in the enhancement proposal https://github.com/openshift/enhancements/pull/1134.

pkg/asset/installconfig/gcp/validation.go
barbacbd: ahh that is just a default place to look. If the full name which should be in PublicDNSZone (something in the format of /projects/{project}/managedZones/{managedZone} ). If the user does not supply the name in the long format then the ProjectID is used as default

pkg/asset/installconfig/gcp/validation.go
barbacbd: Same applies as the public zone

pkg/asset/cluster/tfvars.go
```diff
@@ -415,18 +414,61 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		for i, w := range workers {
 			workerConfigs[i] = w.Spec.Template.Spec.ProviderSpec.Value.Object.(*machinev1beta1.GCPMachineProviderSpec)
 		}
-		if installConfig.Config.Publish == types.ExternalPublishingStrategy {
+
+		preexistingnetwork := installConfig.Config.GCP.Network != ""
```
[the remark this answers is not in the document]
barbacbd: we can do that but that line wasn't new its only showing up on changes because it was moved.

pkg/asset/installconfig/gcp/validation.go
patrickdillon: Oops. Yeah this is correct. I realized this as I read through the code but forgot to delete my comment.

pkg/asset/cluster/tfvars.go
```diff
@@ -415,18 +414,61 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
+
+		createFirewallRules := true
+		if installConfig.Config.GCP.CreateFirewallRules == gcp.CreateFirewallRulesDisabled {
+			createFirewallRules = false
+		}
```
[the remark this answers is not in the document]
jstuever: Firewall rules shouldn't be part of this PR. I think it was just moved?

jstuever:
/unapprove
Still contention around some of the solutions here, need to regroup.

pkg/types/gcp/platform.go
jstuever: The format used is the `Name` format, as defined by the Google API. I can see where it could be considered technically overloaded by enabling the string to either be the `Name` or the `ID` values. Though, there are several decades of precedence for this; for example most modern operating system commands have accepted filenames given in short, relative, or fully qualified format. As such, I don't particularly see this as an issue. If we choose to break this up into a strict struct, I highly recommend using the `ID` identifier for both the specified managedZone and it's project.

https://cloud.google.com/apis/design/resource_names

jstuever: The Google API for Managed Zone also sets this precedence in that it allows both `Name` and `ID`:
https://cloud.google.com/dns/docs/reference/v1/managedZones/get#parameters

pkg/types/gcp/platform.go
```diff
@@ -59,4 +59,28 @@ type Platform struct {
...
+	// When the PrivateDNSZone is empty, the DNS zones will be created and managed in the Service
+	// Project (ProjectID). When PrivateDNSZone is provided, the DNS zone will be created and managed
+	// in the requested Project through PrivateZoneProject.
+	// +optional
+	PrivateDNSZone string `json:"privateDNSZone,omitempty"`
+
+	// PrivateZoneProject Technology Preview
+	// When the PrivateDNSZone is provided, the zone will be created in this project.
+	// +optional
+	PrivateZoneProject string `json:"privateZoneProject,omitempty"`
```
jstuever, line 73: If we are going to use multiple fields, we might want to consider joining them in a struct
```
type DNSZone struct {
    ID string `json:"id,omitempty"`
    ProjectID string `json:"project,omitempty"`
}
```

and
```
PublicDNSZone DNSZone `json:"publicDNSZone,omitempty"`
PrivateDNSZone DNSZone `json:"privateDNSZone,omitempty"`
```

jstuever (changes requested):
There are some functionality issues here that need to be resolved.

jstuever:
/unapprove
Not sure why the bot threw that back on there.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -114,6 +116,60 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+			if _, err := client.GetDNSZoneByName(context.TODO(), ic.GCP.ProjectID, ic.GCP.PublicDNSZone); err != nil {
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("publicDNSZone"), ic.GCP.PublicDNSZone, "invalid public managed zone"))
+			}
+		}
+	}
+
+	if ic.GCP.PrivateDNSZone != "" {
+		if ic.GCP.PrivateZoneProject == "" {
+			errMsg := "privateZoneProject cannot be empty when privateDNSZone is provided"
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("privateZoneProject"), ic.GCP.PrivateZoneProject, errMsg))
```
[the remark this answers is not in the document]
barbacbd: it defaults to ic.GCP.ProjectID in the next few lines. Is that what you are referring to ?

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -114,6 +116,60 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+}
+
+// validateManagedZones will validate the public and private managed zones if they exist.
+func validateManagedZones(client API, ic *types.InstallConfig, fieldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if ic.GCP.PublicDNSZone != "" {
+		if ic.GCP.PublicZoneProject == "" {
+			errMsg := "publicZoneProject cannot be empty when publicDNSZone is provided"
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("publicZoneProject"), ic.GCP.PublicZoneProject, errMsg))
```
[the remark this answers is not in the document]
barbacbd: it defaults to ic.GCP.ProjectID in the next few lines, is that what we want?

pkg/asset/installconfig/gcp/validation.go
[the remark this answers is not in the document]
jstuever: It doesn't matter if it defaults if you error out on validation prior to reaching that point, though, yeah?

pkg/asset/installconfig/gcp/validation.go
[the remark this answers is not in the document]
jstuever: I was referring to enabling the user to leave `PublicZoneProject` empty and have it default to the cluster project. This isn't possible with this validation in place, it errors out that the field is required.

pkg/asset/manifests/dns.go
```diff
@@ -36,6 +36,18 @@ import (
...
+	findGCPProject = func(zoneProject, defaultProject string) string {
+		project := zoneProject
+		if project == "" {
+			project = defaultProject
+		}
+		return project
+	}
+
+	combineGCPZoneInfo = func(project, zoneName string) string {
+		return fmt.Sprintf("/project/%s/managedZones/%s", project, zoneName)
```
jstuever, line 49: No leading slash on `project/`

pkg/asset/manifests/dns.go
```diff
@@ -134,14 +146,37 @@ func (d *DNS) Generate(dependencies asset.Parents) error {
...
-			zone, err := icgcp.GetPublicZone(context.TODO(), installConfig.Config.Platform.GCP.ProjectID, installConfig.Config.BaseDomain)
+
+		// Set the public zone
+		switch {
+		case installConfig.Config.Publish != types.ExternalPublishingStrategy:
+			// Do not use a public zone when not publishing externally.
+		case installConfig.Config.GCP.PublicDNSZone.ID != "":
+			// Use the provided zone if specified.
+			project := findGCPProject(installConfig.Config.GCP.PublicDNSZone.ProjectID, installConfig.Config.GCP.ProjectID)
+			config.Spec.PublicZone = &configv1.DNSZone{ID: combineGCPZoneInfo(project, installConfig.Config.GCP.PublicDNSZone.ID)}
```
jstuever, line 157: This is currently a breaking change in absence of modifications to the cluster-ingress-operator. It currently only accepts the short name (ID). By forcing all clusters to use the relative name (Name) here, we break the default installs. Instead, I recommend we only use the `combineGCPZoneInfo()` when `installConfig.Config.GCP.PublicDNSZone.ProjectID != "" && installConfig.Config.GCP.PublicDNSZone.ProjectID != installConfig.Config.GCP.ProjectID`

pkg/asset/manifests/dns.go
```diff
@@ -134,14 +146,37 @@ func (d *DNS) Generate(dependencies asset.Parents) error {
...
+		default:
+			// Search the project for a zone with the specified base domain.
+			project := findGCPProject(installConfig.Config.GCP.PublicDNSZone.ProjectID, installConfig.Config.GCP.ProjectID)
+			zone, err := icgcp.GetPublicZone(context.TODO(), project, installConfig.Config.BaseDomain)
 			if err != nil {
 				return errors.Wrapf(err, "failed to get public zone for %q", installConfig.Config.BaseDomain)
 			}
-			config.Spec.PublicZone = &configv1.DNSZone{ID: zone.Name}
+			config.Spec.PublicZone = &configv1.DNSZone{ID: combineGCPZoneInfo(project, zone.Name)}
```
jstuever, line 165: This is the same breaking change as above... More particular, here we know the project has not been overrided.

pkg/asset/manifests/dns.go
```diff
@@ -134,14 +146,37 @@ func (d *DNS) Generate(dependencies asset.Parents) error {
...
 		}
-		config.Spec.PrivateZone = &configv1.DNSZone{ID: fmt.Sprintf("%s-private-zone", clusterID.InfraID)}
+
+		// Set the private zone
+		switch {
+		case installConfig.Config.GCP.PrivateDNSZone.ID != "":
+			// Use the provided zone if specified.
+			project := findGCPProject(installConfig.Config.GCP.PrivateDNSZone.ProjectID, installConfig.Config.GCP.ProjectID)
+			config.Spec.PrivateZone = &configv1.DNSZone{ID: combineGCPZoneInfo(project, installConfig.Config.GCP.PrivateDNSZone.ID)}
```
jstuever, line 173: The same breaking change as above.

pkg/asset/manifests/dns.go
```diff
@@ -134,14 +146,37 @@ func (d *DNS) Generate(dependencies asset.Parents) error {
...
+		default:
+			// Use the installer created private zone.
+			project := findGCPProject(installConfig.Config.GCP.PrivateDNSZone.ProjectID, installConfig.Config.GCP.ProjectID)
+			config.Spec.PrivateZone = &configv1.DNSZone{ID: combineGCPZoneInfo(project, fmt.Sprintf("%s-private-zone", clusterID.InfraID))}
```
jstuever, line 177: The same breaking change as above.

pkg/types/gcp/platform.go
```diff
@@ -11,6 +11,22 @@ const (
 	CreateFirewallRulesDisabled CreateFirewallRules = "Disabled"
 )
 
+// DNSZone stores the information common and required to create DNS zones including
+// the project and id/name of the zone.
+type DNSZone struct {
+	// ID Technology Preview.
+	// When the ProjectID is empty, the DNS zone with this ID will be created and managed in the Service
+	// Project (GCP.ProjectID). When ProjectID is provided, the DNS zone will be created and managed
+	// in the requested ProjectID.
```
jstuever, line 20: Nit: The logic regarding ProjectID might be better served with the ProjectID parameter below.

pkg/asset/manifests/dns.go
jstuever: Actually, this only applies to the default: use case...

pkg/asset/manifests/dns.go
jstuever: Actually, this is only an issue for the default: case.

jstuever (commented):
This appears to function as desired. I'll leave it to the rest of the team to review and approve.

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -2286,10 +2286,39 @@ spec:
...
+                  privateDNSZone:
+                    description: PrivateDNSZone Technology Preview. PrivateDNSZone
+                      contains the zone ID and project where the Private DNS zone
+                      will be created.
+                    properties:
+                      id:
+                        description: ID Technology Preview. ID or name of the zone.
+                        type: string
+                      project:
+                        description: ProjectID Technology Preview When the ProjectID
```
sadasu, line 2298: Nit. Missing a period here Preview "." When.

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -2286,10 +2286,39 @@ spec:
...
                     type: string
+                  publicDNSZone:
+                    description: PublicDNSZone Technology Preview. PublicDNSZone contains
+                      the zone ID and project where the Public DNS zone will be created.
+                    properties:
+                      id:
+                        description: ID Technology Preview. ID or name of the zone.
+                        type: string
+                      project:
+                        description: ProjectID Technology Preview When the ProjectID
```
sadasu, line 2316: Ditto as https://github.com/openshift/installer/pull/6288/files#r982613655

pkg/asset/cluster/tfvars.go
```diff
@@ -417,18 +416,45 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
-		createFirewallRules := true
-		if installConfig.Config.GCP.CreateFirewallRules == gcp.CreateFirewallRulesDisabled {
-			createFirewallRules = false
+		// Setup defaults for private DNS Zone
+		createPrivateZone := true
+		createPrivateZoneRecords := true
+		privateZoneName := ""
+		privateZoneProject := ""
+		if installConfig.Config.GCP.PrivateDNSZone != nil && installConfig.Config.GCP.PrivateDNSZone.ID != "" {
+			createPrivateZone = false
```
sadasu, line 455: This seems counter intuitive to me. When PrivateDNSZone != nil and the ID is provided, we don't create private zones?

pkg/asset/manifests/dns.go
```diff
@@ -134,14 +138,41 @@ func (d *DNS) Generate(dependencies asset.Parents) error {
...
+			// Do not use a public zone when not publishing externally.
+		case installConfig.Config.GCP.PublicDNSZone != nil && installConfig.Config.GCP.PublicDNSZone.ID != "":
+			// Use the provided zone if specified.
+			zoneID := installConfig.Config.GCP.PublicDNSZone.ID
+			if installConfig.Config.GCP.PublicDNSZone.ProjectID != "" && installConfig.Config.GCP.ProjectID != installConfig.Config.GCP.PublicDNSZone.ProjectID {
+				zoneID = combineGCPZoneInfo(installConfig.Config.GCP.PublicDNSZone.ProjectID, installConfig.Config.GCP.PublicDNSZone.ID)
+			}
+			config.Spec.PublicZone = &configv1.DNSZone{ID: zoneID}
+		default:
+			// Search the project for a zone with the specified base domain.
```
sadasu, line 154: It appears that we would get to the default case when installConfig.Config.GCP.PublicDNSZone != nil && installConfig.Config.GCP.PublicDNSZone.ID == "". We can be in that case when installConfig.Config.GCP.PublicDNSZone != nil && installConfig.Config.GCP.PublicDNSZone.ProjectID != "". Basically, a projectID could potentially be different from installConfig.Config.GCP.ProjectID but the zone is not specified.
In that case, we ignore the installConfig.Config.GCP.PublicDNSZone.ProjectID and treat the projectID to be the same as installConfig.Config.GCP.ProjectID. 
Question is, should this case be caught in validation as an invalid config (i.e installConfig.Config.GCP.PublicDNSZone != nil and installConfig.Config.GCP.PublicDNSZone.ProjectID != "" and installConfig.Config.GCP.PublicDNSZone.ID == "") ?

pkg/asset/cluster/tfvars.go
barbacbd: no that means that one exists already we DO NOT want to create one. We are assuming it is already created and provided by the user. This option of adding the privateDNSZone.ID is not to name a new private zone but to use an existing one.

data/data/install.openshift.io_installconfigs.yaml
barbacbd: well its tech preview does it matter = )

pkg/asset/manifests/dns.go
barbacbd: In this case, no the project ID cannot be different than installConfig.Config.GCP.ProjectID if a project ID is not provided to the zone. It will fail validation.

pkg/asset/cluster/tfvars.go
sadasu: Ok. Got it. I think I was equating createPrivateZone with what createPrivateZoneRecords is doing.

pkg/asset/manifests/dns.go
sadasu: Agreed. I guess what I am trying to say is, if a zone is not provided, then the PrivateDNSZone and PublicDNSZone configs are really not useful. Should there be a check for that during validation?

sadasu: More validations are addressed as part of https://github.com/openshift/installer/pull/6300/commits/6537150a6438a2bd81767ab241bd33ad20d65cf9.

PR #6289, opened 2022-08-31
Power VS: Enable global routing for cloud connection and specify dns

mjturek:
Because the VPC will not always be in the same DC as the Power VS service instance, we want to enable global routing to allow for this. Additionally we want to specify the DNS server as it's required until private cluster support lands.

### DIFF
--- a/data/data/powervs/cluster/power_network/pi_network.tf
+++ b/data/data/powervs/cluster/power_network/pi_network.tf
@@ -12,15 +12,17 @@ resource "ibm_pi_dhcp" "new_dhcp_service" {
   count                  = var.pvs_network_name == "" ? 1 : 0
   pi_cloud_instance_id   = var.cloud_instance_id
   pi_cloud_connection_id = data.ibm_pi_cloud_connection.cloud_connection.id
+  pi_dns_server          = "1.1.1.1"
 }
 
 resource "ibm_pi_cloud_connection" "new_cloud_connection" {
-  count                           = var.cloud_conn_name == "" ? 1 : 0
-  pi_cloud_instance_id            = var.cloud_instance_id
-  pi_cloud_connection_name        = "cloud-con-${var.cluster_id}"
-  pi_cloud_connection_speed       = 50
-  pi_cloud_connection_vpc_enabled = true
-  pi_cloud_connection_vpc_crns    = [var.vpc_crn]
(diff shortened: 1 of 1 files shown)


data/data/powervs/cluster/power_network/pi_network.tf
```diff
@@ -12,15 +12,17 @@ resource "ibm_pi_dhcp" "new_dhcp_service" {
   count                  = var.pvs_network_name == "" ? 1 : 0
   pi_cloud_instance_id   = var.cloud_instance_id
   pi_cloud_connection_id = data.ibm_pi_cloud_connection.cloud_connection.id
+  pi_dns_server          = "1.1.1.1"
```
mkumatag, line 15: I prefer avoiding this for long run, if its workaround at the moment - lets mention in the comment as TODO to remove once the issue gets fixed in the infrastructure.

mkumatag (changes requested):

mjturek:
@rna-afk could you take a look at this? We need this to resume our CI efforts

data/data/powervs/cluster/power_network/pi_network.tf
mjturek: @mkumatag In the future, once Private Cluster Support lands, we will always be passing `pi_dns_server` and it will default to `1.1.1.1`.

Do you still think we need a `TODO`?

clnperez: what happens if we don't pass pi_dns_server? the api itself doesn't still default it to 1.1.1.1?

mkumatag: > @mkumatag In the future, once Private Cluster Support lands, we will always be passing `pi_dns_server` and it will default to `1.1.1.1`.
> 
> Do you still think we need a `TODO`?

nope, we should have if/else sort of a block where if user supplies that thens set this param or else drop(because this is an optional param and powervs code should take care of setting right dns server) to rely on what powervs dhcp service gives.

mkumatag: > what happens if we don't pass pi_dns_server? the api itself doesn't still default it to 1.1.1.1?

This is not happening with the latest code change but eventually that's the expectation

PR #6293, opened 2022-09-01
OCPBUGS-729: vSphere 6.7u3 failing privilege validation

rvanderp3:
Remove privileges which aren't available as determined via the vCenter API. Some privileges required for vSphere 7.0 aren't available in vSphere 6.7.  The intention of this PR is to filter the required privileges based on what privileges the authorization manager reports are available.


rvanderp3:
/hold

still working on unit tests but the fix is available and has been hand tested.

pkg/asset/installconfig/vsphere/permission_test.go
```diff
@@ -18,82 +23,149 @@ func buildPermissionGroup(authManagerMock *mock.MockAuthManager,
...
+			for _, permissionToExclude := range permissionToExcludeSet {
+				if permission == permissionToExclude {
+					include = false
+					break
+				}
+			}
+			if include {
+				filteredPermissionsToApply = append(filteredPermissionsToApply, permission)
+			}
+		}
```
r4f4, line 43: Maybe this can be simplified by using https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets ?

pkg/asset/installconfig/vsphere/permissions.go
```diff
@@ -162,14 +167,44 @@ var permissions = map[permissionGroup]PermissionGroupDefinition{
...
+			for _, availablePermission := range authManagerMo.PrivilegeList {
+				if availablePermission.PrivId == permission {
+					keep = true
+					break
+				}
+			}
+			if keep {
+				prunedPermissions = append(prunedPermissions, permission)
+			}
+		}
```
r4f4, line 191: What about
```suggestion
    availablePermissions := sets.NewString()
    for _, perm := range authManagerMo.PrivilegeList {
        availablePermissions.Insert(perm.PrivId)
    }
	for permissionGroupKey, permissionGroup := range permissions {
		prunedPermissions := sets.NewString(permissionGroup.Permissions...).Intersection(availablePermissions)
```

rvanderp3: great suggestion, i'll check that out :)

rvanderp3:
/hold

will rebase once https://github.com/openshift/installer/pull/6281 merges

rvanderp3:
e2e-vsphere-ovn and e2e-vsphere-zones both installed properly

PR #6294, opened 2022-09-01
Power VS: use the new name field for the dhcp nw

clnperez:
this will make the destroy logic more robust, as currently we have no way of knowing which networks belonged to which clusters when destroy errors occurred. the current logic is to look at the nw attached to a cluster vm and delete it, but if the cluster vms had been destroyed already, there would be an unused nw left behind.


clnperez:
@rna-afk can you review? i think all the tests pass now

r4f4 (commented):
/lgtm

clnperez:
the `approve` label isn't showing up so this one's not getting merged. @rna-afk do we need someone else's approval?

rna-afk:
Well that's weird. Let me try again.

/approve

PR #6296, opened 2022-09-02
Correct the GCP machine series name

jianli-wei:
1. [Google docs](https://cloud.google.com/compute/docs/machine-types)
2. The official name of the machine series is "Tau T2D", rather than "T2D".


jianli-wei:
The failure was after successful installation. Nothing to do with the changes of the PR, thanks! 

Test name | Installation Result / Failure
-- | --
ci/prow/e2e-gcp-shared-vpc | Step phase pre succeeded after 1h3m40s.

PR #6297, opened 2022-09-02
[azure] add tested family on aarch64

MayXuQQ:
ref [OCPBUGS#752 Change documentation from supported instance types to tested instance types #49799](https://github.com/openshift/openshift-docs/pull/49799)


MayXuQQ:
since just tested on aarch64 as worker, so close the PR.

MayXuQQ:
https://gitlab.cee.redhat.com/aosqe/flexy-templates/-/blob/master/functionality-testing/aos-4_12/upi-on-azure/versioned-installer-arm not just worker, master node support arm now.

MayXuQQ:
we don't know yet if we'll be able to officially support azure on arm 
there are some concerns and discussions between arm team and installer team
--aleskandro

PR #6298, opened 2022-09-02
azure: add support for arm64 instances (Dps_v5 instances)

Prashanth684:
As part of the overall effort to add IPI support for Azure on ARM
instances (https://issues.redhat.com/browse/ARMOCP-16), this PR enables
the creation of arm64 machines when the arm64 architecture is specified
in the install config. Also added validation and tests.

### DIFF
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -102,6 +102,25 @@ func validatePremiumDisk(fieldPath *field.Path, diskType string, instanceType st
 	return field.ErrorList{}
 }
 
+func validateVMArchitecture(fieldPath *field.Path, instanceType string, architecture types.Architecture, capabilities map[string]string) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	val, ok := capabilities["CpuArchitectureType"]
+	if ok {
+		if architecture != types.ArchitectureARM64 && architecture != types.ArchitectureAMD64 {
+			errMsg := fmt.Sprintf("invalid install config architecture %s", architecture)
+			allErrs = append(allErrs, field.Invalid(fieldPath, instanceType, errMsg))
+		} else if (architecture == types.ArchitectureARM64 && !strings.EqualFold(val, "Arm64")) || (architecture == types.ArchitectureAMD64 && !strings.EqualFold(val, "x64")) {
+			errMsg := fmt.Sprintf("instance type architecture '%s' does not match install config architecture %s", val, architecture)
+			allErrs = append(allErrs, field.Invalid(fieldPath, instanceType, errMsg))
+		}
+	} else {
+		logrus.Warnf("Could not determine VM's architecture from its capabilities. Assuming it is %v", architecture)
+	}
+
+	return allErrs
+}
+
 func validateMininumRequirements(fieldPath *field.Path, req resourceRequirements, instanceType string, capabilities map[string]string) field.ErrorList {
 	allErrs := field.ErrorList{}
 
@@ -200,7 +219,7 @@ func validateUltraSSD(client API, fieldPath *field.Path, icZones []string, regio
 }
 
 // ValidateInstanceType ensures the instance type has sufficient Vcpu and Memory.
-func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceType, diskType string, req resourceRequirements, ultraSSDEnabled bool, vmNetworkingType string, icZones []string) field.ErrorList {
+func ValidateInstanceType(client API, fieldPath *field.Path, region, instanceType, diskType string, req resourceRequirements, ultraSSDEnabled bool, vmNetworkingType string, icZones []string, architecture types.Architecture) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	capabilities, err := client.GetVMCapabilities(context.TODO(), instanceType, region)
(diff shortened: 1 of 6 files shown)


pkg/asset/installconfig/azure/validation.go
```diff
@@ -312,10 +335,22 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
 			ultraSSDEnabled := strings.EqualFold(ultraSSDCapability, "Enabled")
 			allErrs = append(allErrs, ValidateInstanceType(client, fieldPath.Child("platform", "azure"),
-				ic.Azure.Region, instanceType, diskType, computeReq, ultraSSDEnabled, vmNetworkingType, zones)...)
+				ic.Azure.Region, instanceType, diskType, computeReq, ultraSSDEnabled, vmNetworkingType, zones, architecture)...)
 		}
 	}
 
+	// we are going to use the defaultInstanceType so let's check it. The assumption here is that since cp and compute arches cannot differ today, it's ok to
+	// not check the default instance as long as it is used in any one place.
+	if !useDefaultInstanceType && defaultInstanceType != "" {
```
r4f4, line 344: Your comment says "we are going to use the defaultInstanceType". So shouldn't this be:
```suggestion
	if useDefaultInstanceType && defaultInstanceType != "" {
```

pkg/asset/installconfig/azure/validation_test.go
```diff
@@ -324,6 +344,16 @@ func TestAzureInstallConfigValidation(t *testing.T) {
...
+		{
+			name:     "Valid Arm64 instance types",
+			edits:    editFunctions{validArm64InstanceTypes},
+			errorMsg: "",
+		},
+		{
+			name:     "Invalid arch instance types",
+			edits:    editFunctions{invalidArchInstanceTypes},
+			errorMsg: `\[controlPlane.platform.azure.type: Invalid value: "Standard_D8ps_v5": instance type architecture does not match install config architecture amd64, compute\[0\].platform.azure.type: Invalid value: "Standard_D4ps_v5": instance type architecture does not match install config architecture amd64, platform.azure.defaultMachinePlatform.type: Invalid value: "Standard_D4ps_v5": instance type architecture does not match install config architecture amd64\]`,
+		},
```
r4f4, line 356: It would be good to have a test to cover the case when the `defaultMachinePlatform` is used for either controlPlane or compute so your last check in `validateInstanceTypes` is exercised.

r4f4 (changes requested):

pkg/asset/installconfig/azure/validation.go
```diff
@@ -102,6 +102,23 @@ func validatePremiumDisk(fieldPath *field.Path, diskType string, instanceType st
...
+	val, ok := capabilities["CpuArchitectureType"]
+	if ok {
+		if architecture != types.ArchitectureARM64 && architecture != types.ArchitectureAMD64 {
+			errMsg := fmt.Sprintf("invalid install config architecture %s", architecture)
+			allErrs = append(allErrs, field.Invalid(fieldPath, instanceType, errMsg))
+		} else if (architecture == types.ArchitectureARM64 && !strings.EqualFold(val, "Arm64")) || (architecture == types.ArchitectureAMD64 && !strings.EqualFold(val, "x64")) {
+			errMsg := fmt.Sprintf("instance type architecture does not match install config architecture %s", architecture)
+			allErrs = append(allErrs, field.Invalid(fieldPath, instanceType, errMsg))
+		}
+	}
```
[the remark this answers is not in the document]
Prashanth684: we could. i was just worried if some old instance type would not have this property and we don't want to block on this error in that case.

pkg/asset/installconfig/azure/validation.go
Prashanth684: no, i poorly worded it . what i  meant is - we are checking to see if defaultinstancetype has been provided but not been used above, in which case we need to check it specifically to see if it matches the correct architecture. i will change the wording.

pkg/asset/installconfig/azure/validation_test.go
Prashanth684: this test covers that case too - notice the last error `platform.azure.defaultMachinePlatform.type: Invalid value: "Standard_D4ps_v5": instance type architecture does not match install config architecture amd64`

pkg/asset/installconfig/azure/validation.go
Prashanth684: updated the comment.

pkg/asset/installconfig/azure/validation_test.go
r4f4: My comment was a result of my misunderstanding of what your code was doing.

pkg/asset/installconfig/azure/validation.go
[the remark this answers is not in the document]
r4f4: In that case, a debug message could help us diagnose a mismatched architecture on old instance types.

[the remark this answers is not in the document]
Prashanth684: added the message as a warning

pkg/asset/installconfig/azure/validation_test.go
Prashanth684: so are we all good on this then ?

PR #6300, opened 2022-09-02
CORS 2046: GCP Add base domain validation when DNS zone is specified

barbacbd:
** Skip base public zone validation against base domain when the public managed zone is set to "None"
** Log and skip base domain validation when error indications permissions forbidden.

### DIFF
--- a/pkg/asset/installconfig/gcp/validation.go
+++ b/pkg/asset/installconfig/gcp/validation.go
@@ -4,17 +4,18 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"net/http"
 	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	compute "google.golang.org/api/compute/v1"
+	"google.golang.org/api/dns/v1"
 	"google.golang.org/api/googleapi"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"github.com/openshift/installer/pkg/types"
+	"github.com/openshift/installer/pkg/types/gcp"
 	"github.com/openshift/installer/pkg/validate"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/asset/installconfig/gcp/validation.go
```diff
@@ -259,12 +263,9 @@ func ValidateEnabledServices(ctx context.Context, client API, project string) er
 	projectServices, err := client.GetEnabledServices(ctx, project)
 
 	if err != nil {
-		var gErr *googleapi.Error
-		if errors.As(err, &gErr) {
-			if gErr.Code == http.StatusForbidden {
-				logrus.Warn("Permission denied. Unable to fetch enabled services for project.")
-				return nil
-			}
+		if IsNotFound(err) {
```
jstuever, line 266: Shouldn't this be `if IsForbidden(err) {` ?

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -121,15 +119,21 @@ func ValidatePreExitingPublicDNS(client API, ic *types.InstallConfig) error {
...
-		if errors.As(err, &gErr) {
-			if gErr.Code == http.StatusNotFound {
-				return field.NotFound(field.NewPath("baseDomain"), fmt.Sprintf("DNS Zone (%s/%s)", ic.Platform.GCP.ProjectID, ic.BaseDomain))
-			}
+		if IsNotFound(err) {
+			return field.NotFound(field.NewPath("baseDomain"), fmt.Sprintf("DNS Zone (%s/%s)", ic.Platform.GCP.ProjectID, ic.BaseDomain))
+		}
+		if IsForbidden(err) {
+			logrus.Warn("Permission denied. Unable to fetch public dns zone information: %s", ic.BaseDomain)
+			return nil
```
jstuever, line 136: I think we want to return an error here. When we are planning on adding/removing records from a zone, we should at least have permissions to view the zone itself, right? So, if we don't have that, we should fail early.

jstuever (changes requested):
Looking good... A few minor tweaks.

patrickdillon:
Depends on #6288

pkg/asset/installconfig/gcp/client.go
```diff
@@ -155,11 +158,31 @@ func (c *Client) GetPublicDNSZone(ctx context.Context, project, baseDomain strin
...
 	return res, nil
 }
 
+// GetPrivateDNSZone returns a private DNS zone for a basedomain.
+func (c *Client) GetPrivateDNSZone(ctx context.Context, project, baseDomain string) (*dns.ManagedZone, error) {
+	ctx, cancel := context.WithTimeout(context.TODO(), 1*time.Minute)
+	defer cancel()
+
+	return c.GetDNSZoneWithVisibility(ctx, project, baseDomain, false)
+}
```
jstuever, line 176: This function isn't being used and can be removed from the code. As an extension, the `GetDNSZoneWithVisibility` function is also no longer required. As a result, all of this file can be reverted back into the original.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -172,35 +170,80 @@ func validateManagedZones(client API, ic *types.InstallConfig, fieldPath *field.
 	return allErrs
 }
 
-// ValidatePreExitingPublicDNS ensure no pre-existing DNS record exists in the public
-// DNS zone for cluster's Kubernetes API.
-func ValidatePreExitingPublicDNS(client API, ic *types.InstallConfig) error {
+// ValidatePreExistingPrivateDNS ensure that the PrivateZone exists in the cluster domain
+func ValidatePreExistingPrivateDNS(client API, ic *types.InstallConfig) error {
+
+	if ic.Platform.GCP.PrivateDNSZone != nil && ic.Platform.GCP.PrivateDNSZone.ID != "" {
```
jstuever, line 176: The logic in this validation can be flattened significantly by returning early when `ic.Platform.GCP.PrivateDNSZone == nil or ic.Platform.GCP.PrivateDNSZone.ID == ""` When this field is not specified, there are no validations necessary.

```
if ic.Platform.GCP.PrivateDNSZone != nil && ic.Platform.GCP.PrivateDNSZone.ID != "" {
    return nil
}
project := ic.Platform.GCP.PrivateDNSZone.ProjectID
...
```

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -172,35 +170,80 @@ func validateManagedZones(client API, ic *types.InstallConfig, fieldPath *field.
...
+		project := ic.Platform.GCP.PrivateDNSZone.ProjectID
```
jstuever, line 177: nit: the `ValidatePreExistingPublicDNS` function below defaults to the `ic.Platform.GCP.ProjectID`. You might want to choose one default project for both functions for consistency. My recommendation would be to use the `ic.Platform.GCP.ProjectID` as that would be congruent with the actual default behavior.
```
project := ic.Platform.GCP.PrivateDNSZone.ProjectID
if if.Platform.GCP.PublicDNSZone != nil && ic.Platform.GCP.PublicDNSZone.ProjectID != "" {
    project = ic.Platform.GCP.PublicDNSZone.ProjectID
}
```

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -172,35 +170,80 @@ func validateManagedZones(client API, ic *types.InstallConfig, fieldPath *field.
...
+	project := ic.Platform.GCP.ProjectID
+
+	switch {
+	case ic.Platform.GCP.PublicDNSZone == nil:
+		// Project and Zone can be skipped
+	case ic.Platform.GCP.PublicDNSZone != nil && ic.Platform.GCP.PublicDNSZone.ID != "":
+		// Find the project
+		if ic.Platform.GCP.PublicDNSZone.ProjectID != "" {
+			project = ic.Platform.GCP.PublicDNSZone.ProjectID
+		}
```
jstuever, line 221: The logic in this case should be very much parallel to the `ValidatePreExistingPrivateDNS` function. It should test that the specified zone in `Platform.GCP.PublicDNSZone.ID` exists (ie `getDNSZoneByName`) and contains the proper `DnsName`.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -172,35 +170,80 @@ func validateManagedZones(client API, ic *types.InstallConfig, fieldPath *field.
...
 		return nil
 	}
 
 	record := fmt.Sprintf("api.%s.", strings.TrimSuffix(ic.ClusterDomain(), "."))
+	zoneName := ""
+	project := ic.Platform.GCP.ProjectID
+
+	switch {
+	case ic.Platform.GCP.PublicDNSZone == nil:
+		// Project and Zone can be skipped
```
jstuever, line 216: This is functionally redundant with the following line in combination with the recommended `default:` case. The next case will test if != nil, and the only other option at this time will be the default. Or, in other words... this is really the default case.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -172,35 +170,80 @@ func validateManagedZones(client API, ic *types.InstallConfig, fieldPath *field.
...
+	case ic.Platform.GCP.PublicDNSZone != nil && ic.Platform.GCP.PublicDNSZone.ID != "":
+		// Find the project
+		if ic.Platform.GCP.PublicDNSZone.ProjectID != "" {
+			project = ic.Platform.GCP.PublicDNSZone.ProjectID
+		}
+	}
 
-	zone, err := client.GetPublicDNSZone(context.TODO(), ic.Platform.GCP.ProjectID, ic.BaseDomain)
+	zone, err := client.GetPublicDNSZone(context.TODO(), project, ic.BaseDomain)
```
jstuever, line 224: This would be the default: case for the switch statement. It would be used when ic.Platform.GCP.PublicDNSZone.ID != "".

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -172,35 +170,80 @@ func validateManagedZones(client API, ic *types.InstallConfig, fieldPath *field.
...
+// zone is verified against the BaseDomain. If no zone is provided, the base domain is
+// checked for any public zone that can be used.
+func ValidatePreExistingPublicDNS(client API, ic *types.InstallConfig) error {
 	// If this is an internal cluster, this check is not necessary
 	if ic.Publish == types.InternalPublishingStrategy {
 		return nil
 	}
 
 	record := fmt.Sprintf("api.%s.", strings.TrimSuffix(ic.ClusterDomain(), "."))
+	zoneName := ""
```
jstuever, line 211: nit: this is technically a `zoneID` at this point as it cannot include a relative resource name.

jstuever (changes requested):
There are some functional changes required here.

jstuever (changes requested):
This is really close. Just a few minor tweaks.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -141,66 +141,144 @@ func validateZoneProjects(client API, ic *types.InstallConfig, fieldPath *field.
...
-			project = defaultProject
-		}
+// internalFindProject finds the correct project to use during installation. If the project id is
+// provided in the zone use the project id, otherwise use the default project.
+func internalFindProject(zone *gcp.DNSZone, defaultProject string) string {
+	if zone != nil && zone.ProjectID != "" {
+		return zone.ProjectID
+	}
+	return defaultProject
+}
```
patrickdillon, line 151: I think this is fine, because I know this pattern is being used a lot from CORS-2040. But I would suggest an alternative which you could consider, which is to [set a default](https://github.com/openshift/installer/tree/master/pkg/types/gcp/defaults), so if `zone != nil && zone.ProjectID == ""` you set `zone.ProjectID` to the project used for auth. Then in all subsequent calls to `zone.ProjectID` you know it will be populated.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -141,66 +141,144 @@ func validateZoneProjects(client API, ic *types.InstallConfig, fieldPath *field.
...
-// validateManagedZones will validate the public and private managed zones if they exist.
-func validateManagedZones(client API, ic *types.InstallConfig, fieldPath *field.Path) field.ErrorList {
-	commonZoneLookup := func(dnsZone, zoneProject, defaultProject string) error {
-		project := zoneProject
-		if project == "" {
-			project = defaultProject
-		}
+// internalFindProject finds the correct project to use during installation. If the project id is
+// provided in the zone use the project id, otherwise use the default project.
+func internalFindProject(zone *gcp.DNSZone, defaultProject string) string {
```
patrickdillon, line 146: wdym by the `internal`- prefix?

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -141,66 +141,144 @@ func validateZoneProjects(client API, ic *types.InstallConfig, fieldPath *field.
...
-		if _, err := client.GetDNSZoneByName(context.TODO(), project, dnsZone); err != nil {
-			return err
-		}
-		return nil
+// internalFindDNSZone finds a zone in a project. If a project is provided in the zone, the project
+// is used otherwise the default project is used.
+func internalFindDNSZone(client API, zone *gcp.DNSZone, defaultProject string) (*dns.ManagedZone, string, error) {
+	if zone == nil {
+		return nil, defaultProject, fmt.Errorf("zone is not set")
 	}
```
patrickdillon, line 158: This check should not be necessary, the calling code is already ensuring zone != nil

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -141,66 +141,144 @@ func validateZoneProjects(client API, ic *types.InstallConfig, fieldPath *field.
...
+func validateManagedZones(client API, ic *types.InstallConfig, fieldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 
 	if ic.GCP.PublicDNSZone != nil && ic.GCP.PublicDNSZone.ID != "" {
-		if err := commonZoneLookup(ic.GCP.PublicDNSZone.ID, ic.GCP.PublicDNSZone.ProjectID, ic.GCP.ProjectID); err != nil {
-			allErrs = append(allErrs, field.Invalid(fieldPath.Child("PublicDNSZone").Child("ID"), ic.GCP.PublicDNSZone.ID, "invalid public managed zone"))
+		//_, _, err := internalFindDNSZone(client, ic.Platform.GCP.PublicDNSZone, ic.GCP.ProjectID)
+		//if err != nil {
+		//	return allErrs
+		//}
```
patrickdillon, line 176: remove commented code

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -141,66 +141,144 @@ func validateZoneProjects(client API, ic *types.InstallConfig, fieldPath *field.
...
+		returnedZone, project, err := internalFindDNSZone(client, ic.Platform.GCP.PublicDNSZone, ic.GCP.ProjectID)
```
patrickdillon, line 177: Passing the project in and then getting the project (or a different project) out I think makes this difficult to read. I think it would be better to either call `internalFindProject` separately or implement my previous suggestion of using a default value for the project.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -141,66 +141,144 @@ func validateZoneProjects(client API, ic *types.InstallConfig, fieldPath *field.
...
+		if err != nil {
+			switch {
+			case IsNotFound(err):
+				allErrs = append(allErrs, field.NotFound(field.NewPath("baseDomain"), fmt.Sprintf("DNS Zone (%s/%s)", project, ic.BaseDomain)))
+			case IsForbidden(err):
+				errMsg := fmt.Sprintf("Permission denied. Unable to fetch public dns zone information: %s", ic.BaseDomain)
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("publicManagedZone"), ic.GCP.PublicDNSZone.ID, errMsg))
```
patrickdillon, line 184: make sure that `err` is passed through. you can `wrap` with your error message (and that error message should use lowercase--don't start with caps

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -141,66 +141,144 @@ func validateZoneProjects(client API, ic *types.InstallConfig, fieldPath *field.
...
+			default:
+				allErrs = append(allErrs, field.InternalError(field.NewPath("baseDomain"), err))
+			}
+		} else {
+			// verify that the managed zone exists in the BaseDomain - Trim both values just in case
+			if !strings.EqualFold(strings.TrimSuffix(returnedZone.DnsName, "."), strings.TrimSuffix(ic.BaseDomain, ".")) {
+				errMsg := fmt.Sprintf("publicDNSZone does not exist in baseDomain %s", ic.BaseDomain)
+				allErrs = append(allErrs, field.Invalid(fieldPath.Child("publicDNSZone").Child("id"), ic.Platform.GCP.PublicDNSZone.ID, errMsg))
+			}
 		}
```
patrickdillon, line 194: I'm not sure if this validation should be behind an `else`

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -141,66 +141,144 @@ func validateZoneProjects(client API, ic *types.InstallConfig, fieldPath *field.
...
+		if err != nil {
+			return field.InternalError(field.NewPath("baseDomain"), err)
+		}
+
+		for _, r := range rrSets {
+			if strings.EqualFold(r.Name, record) {
+				errMsg := fmt.Sprintf("record %s already exists in DNS Zone (%s/%s) and might be in use by another cluster, please remove it to continue", record, project, zone.Name)
+				return field.Invalid(field.NewPath("metadata", "name"), ic.ObjectMeta.Name, errMsg)
+			}
+		}
```
patrickdillon, line 241: This is only checking for the api record... if api-int or *.apps exist, that could be problematic. I think it is fair to require it to be empty. I believe that is what AWS does:
https://github.com/openshift/installer/blob/ac9f1c19c96ebab17e67490cf8ccf6504166a566/pkg/asset/installconfig/aws/route53.go#L51-L69

EDIT: Although it looks like this may be using existing code, so perhaps this is "good enough."

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -141,66 +141,144 @@ func validateZoneProjects(client API, ic *types.InstallConfig, fieldPath *field.
...
-				return field.NotFound(field.NewPath("baseDomain"), fmt.Sprintf("DNS Zone (%s/%s)", ic.Platform.GCP.ProjectID, ic.BaseDomain))
-			}
+	if ic.GCP.PublicDNSZone != nil && ic.GCP.PublicDNSZone.ID != "" {
+		zoneName = ic.GCP.PublicDNSZone.ID
+	}
+
+	if zoneName == "" {
+		zone, err := client.GetPublicDNSZone(context.TODO(), project, ic.BaseDomain)
+		if err != nil {
+			return fmt.Errorf("failed to find public zone in project %s", project)
```
patrickdillon, line 267: can you `wrap` `err` with this message, so that err gets passed through to stdout

pkg/asset/installconfig/gcp/validation.go
patrickdillon: actually, LGTM, we need the `returnedZone` value for this.

jstuever:
/override ci/prow/e2e-aws-ovn
Not related to this PR.

jstuever:
Not related to this PR

jstuever:
/override ci/prow/e2e-gcp-ovn
Flaky unit tests, unrelated to this PR.

jstuever:
/override ci/prow/e2e-gcp-ovn
Flaky e2e tests, not related to this PR.

PR #6302, opened 2022-09-05
OCPBUGS-889: bump default channel to stable-4.12

evakhoni:
```
╰─ oc get clusterversion
NAME      VERSION                              AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.12.0-0.nightly-2022-09-02-154321   True        False         84m     Cluster version is 4.12.0-0.nightly-2022-09-02-154321

╰─ oc get clusterversion/version -ojson | jq .spec
{
  "channel": "stable-4.11",
  "clusterID": "f96fff8e-4905-4f9a-92e0-96a659fe8cc6"
}
```

### DIFF
--- a/data/data/manifests/bootkube/cvo-overrides.yaml.template
+++ b/data/data/manifests/bootkube/cvo-overrides.yaml.template
@@ -8,7 +8,7 @@ spec:
   upstream: https://amd64.origin.releases.ci.openshift.org/graph
   channel: stable-4
 {{- else }}
-  channel: stable-4.11
+  channel: stable-4.12
 {{- end }}
   clusterID: {{.CVOClusterID}}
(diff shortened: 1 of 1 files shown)


sadasu:
@evakhoni can this value be derived rather than a hard-coded value?

barbacbd (commented):
/lgtm

barbacbd:
@sadasu I added LGTM based on Trevor opening a PR that said this is how it is usually done. I think we could look changing the version in several locations at once (as a passed in value or something), but that could be later work.

evakhoni:
@sadasu @barbacbd looks like the failures are definitely orthogonal to my change. do you mind helping with an override?

PR #6303, opened 2022-09-05
OCPBUGS-759: Extend user tags limit to 25 based on AWS limits for 4.11 release

TrilokGeer:
Extend user tag limit to 15 based on AWS limits. 10 tags are reserved for OpenShift use. Previously, there has been a confusion on the tag limit for S3 bucket vs S3 bucket object tag limit. S3 bucket object tag limit is 10 while the requirement is for S3 bucket which has limit of 50 tags.

As agreed with customer scenario, 4.11 limit with 15 user tags works. 4.12 will have the limitation extended to support 40 user tags.

### DIFF
--- a/data/data/aws/bootstrap/main.tf
+++ b/data/data/aws/bootstrap/main.tf
@@ -11,6 +11,18 @@ locals {
   volume_type      = "gp3"
   volume_size      = 30
   volume_iops      = local.volume_type == "io1" ? 100 : 0
+
+  // s3 object supports only 10 tags. The first 8 tags from
+  // the list of aws_extra_tags are used for s3 object
+  // slice function uses new_tag_len as excluding index
+  new_tag_len    = 8
+  sliced_tag_map = length(var.aws_extra_tags) <= local.new_tag_len ? var.aws_extra_tags : { for k in slice(keys(var.aws_extra_tags), 0, local.new_tag_len) : k => var.aws_extra_tags[k] }
+  s3_object_tags = merge(
+    {
+      "kubernetes.io/cluster/${var.cluster_id}" = "owned"
+    },
+    local.sliced_tag_map,
+  )
 }
 
 provider "aws" {
@@ -64,7 +76,7 @@ resource "aws_s3_bucket_object" "ignition" {
     {
       "Name" = "${var.cluster_id}-bootstrap"
     },
-    local.tags,
+    local.s3_object_tags,
   )
 
   lifecycle {
--- a/pkg/types/aws/validation/platform.go
+++ b/pkg/types/aws/validation/platform.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"github.com/openshift/installer/pkg/types/aws"
@@ -49,7 +50,10 @@ func validateUserTags(tags map[string]string, propagatingTags bool, fldPath *fie
 		return allErrs
 	}
 	if len(tags) > 8 {
-		allErrs = append(allErrs, field.Invalid(fldPath, len(tags), "number of user tags cannot be more than 8"))
+		logrus.Warnf("Due to a limit of 10 tags on S3 Bucket Objects, only the first eight lexicographically sorted tags will be applied to the bootstrap ignition object, which is a temporary resource only used during installation")
+	}
+	if len(tags) > 25 {
+		allErrs = append(allErrs, field.Invalid(fldPath, len(tags), "number of user tags cannot be more than 25"))
 	}
 	for key, value := range tags {
 		if strings.EqualFold(key, "Name") {


patrickdillon:
> Previously, there has been a confusion on the tag limit for S3 bucket vs S3 bucket object tag limit. S3 bucket object tag limit is 10 while the requirement is for S3 bucket which has limit of 50 tags.

Can you help clarify the confusion? We create [an S3 bucket object here](https://github.com/openshift/installer/blob/master/data/data/aws/bootstrap/main.tf#L55) and apply the tags to it. 

https://bugzilla.redhat.com/show_bug.cgi?id=1958154 shows how this fails if more than 8 tags are supplied. 

What is different now? Were you able to test this with, say, 15 tags?

TrilokGeer:
> > Previously, there has been a confusion on the tag limit for S3 bucket vs S3 bucket object tag limit. S3 bucket object tag limit is 10 while the requirement is for S3 bucket which has limit of 50 tags.
> 
> Can you help clarify the confusion? We create [an S3 bucket object here](https://github.com/openshift/installer/blob/master/data/data/aws/bootstrap/main.tf#L55) and apply the tags to it.
> 
> https://bugzilla.redhat.com/show_bug.cgi?id=1958154 shows how this fails if more than 8 tags are supplied.
> 
> What is different now? Were you able to test this with, say, 15 tags?

The tags are added to S3 bucket resource and also object created under the resource. For S3 bucket resource, the tag limit is 50. For every object added to S3 bucket resource, there is a limitation of 10 tags. Post installation, the S3 object is removed from the S3 bucket created during bootstrap. Is there a scenario which requires individual S3 objects to be tagged as same as the S3 bucket resource in which they are created during bootstrap?  

This PR suggests to not add user tags to individual S3 objects and add user tags to S3 bucket resource only. For identification of objects, cluster tags are still maintained on the bootstrap S3 object. This way, >10 user tags are supported.

data/data/aws/bootstrap/main.tf
```diff
@@ -11,6 +11,24 @@ locals {
   volume_type      = "gp3"
   volume_size      = 30
   volume_iops      = local.volume_type == "io1" ? 100 : 0
+
+  // s3 object supports only 10 tags. The first 9 tags from 
+  // the list of aws_extra_tags are used for s3 object
+  new_tag_len = 9
+
+  keys  = flatten([
+      for key, value in local.tags: "${key}"
```
patrickdillon, line 20: To preserve the owned tag, we must get the slice of 8 user tags and then add in the owned tag.  If we add the owned tag to the user tags and then take the first 9, the owned tag may not be within the first 9 (based on the lexicographical ordering).

data/data/aws/bootstrap/main.tf
```diff
@@ -11,6 +11,24 @@ locals {
...
+  values = flatten([
+      for key, value in local.tags: "${value}"
+  ])
+
+  slicelen = length(local.keys) < local.new_tag_len ? length(local.keys) : local.new_tag_len
+
+  slicedkeys = slice(local.keys, 0, local.slicelen)
+  slicedvalues = slice(local.values, 0, local.slicelen)
+
+  s3_object_tags = zipmap(local.slicedkeys, local.slicedvalues)
```
patrickdillon, line 31: I think we can simplify how to get the first X items from the map. Using `terraform console` I tested out: `{for k in slice(keys(var.longMap), 0, 8): k => var.longMap[k]}`:

```
> var.longMap
tomap({
  "a" = 0
  "b" = 1
  "c" = 2
  "d" = 3
  "e" = 4
  "f" = 5
  "g" = 6
  "h" = 7
  "i" = 8
  "j" = 9
})
> {for k in slice(keys(var.longMap), 0, 8): k => var.longMap[k]}
{
  "a" = 0
  "b" = 1
  "c" = 2
  "d" = 3
  "e" = 4
  "f" = 5
  "g" = 6
  "h" = 7
}
```
So I think we could combine that with a ternary to get the shortened user tags, something like:

```
s3_object_user_tags = length(var.aws_extra_tags) <= 8 ? var.aws_extra_tags : {for k in slice(keys(var.aws_extra_tags), 0, 8): k => var.aws_extra_tags[k]}
```
(I did not test this last command)

Then we just need to ultimately make sure to add in the name tag and the owned tag.

pkg/types/aws/validation/platform.go
```diff
@@ -9,6 +9,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"github.com/openshift/installer/pkg/types/aws"
+	"github.com/sirupsen/logrus"
```
patrickdillon, line 12: nit: this should be in the middle import block. We try to follow the convention

```
<stdlib>

<external>

<internal>
```

We should enforce this by unit tests :/

pkg/types/aws/validation/platform.go
```diff
@@ -49,7 +50,10 @@ func validateUserTags(tags map[string]string, propagatingTags bool, fldPath *fie
 		return allErrs
 	}
 	if len(tags) > 8 {
-		allErrs = append(allErrs, field.Invalid(fldPath, len(tags), "number of user tags cannot be more than 8"))
+		logrus.Warnf("bootstrap s3 object tags will be random ordered as number of user  tags is more than 8")
```
patrickdillon, line 53: In terraform, maps are lexicographically ordered, so I believe the order would not be random...

```suggestion
		logrus.Warnf("Due to a limit of 10 tags on S3 Bucket Objects, only the first eight lexicographically sorted tags will be applied to the bootstrap ignition object, which is a temporary resource only used during installation")
```

Also, log messages--unlike error messages--should be capitalized.

patrickdillon (commented):
This looks really good, but the code could be improved.

pkg/types/aws/validation/platform.go
TrilokGeer: Updated the import section for the format.

data/data/aws/bootstrap/main.tf
TrilokGeer: True. Thanks for the check here. Updated the sliced map of tags creation.

sdodson:
/retitle OCPBUGS-759 : Extend user tags limit to 25 based on AWS limits for 4.11 release

patrickdillon:
/retitle OCPBUGS-759: Extend user tags limit to 25 based on AWS limits for 4.11 release

PR #6304, opened 2022-09-06
azure: create bootimages using image gallery

Prashanth684:
In order to support provisioning of arm64 bootimages, these need to be
created using the image gallery resource as they are not supported
through managed images. For this purpose, it would be good to unify the
implementation of x86 and arm64 to use image galleries as Microsoft have
said that it will be the supported way of creating bootimages going forward
for new instance types of new architectures.

This implementation would create one image gallery with two images: a
gen1 and a gen2 image which we would then pick appropriately based on
the VM generation.

xref: https://issues.redhat.com/browse/ARMOCP-16

### DIFF
--- a/data/data/azure/variables-azure.tf
+++ b/data/data/azure/variables-azure.tf
@@ -194,3 +194,8 @@ variable "random_storage_account_suffix" {
   type        = string
   description = "A random string generated to add a suffix to the storage account and blob"
 }
+
+variable "azure_vm_architecture" {
+  type        = string
+  description = "Architecture of the VMs - used when creating images in the image gallery"
+}
--- a/data/data/azure/variables-azure.tf
+++ b/data/data/azure/variables-azure.tf
@@ -199,3 +199,9 @@ variable "azure_vm_architecture" {
   type        = string
   description = "Architecture of the VMs - used when creating images in the image gallery"
 }
+
+variable "azure_image_release" {
+  type        = string
+  description = "RHCOS release image version - used when creating the image definition in the gallery"
+}
+
--- a/data/data/azure/vnet/main.tf
+++ b/data/data/azure/vnet/main.tf
@@ -128,7 +128,7 @@ resource "azurerm_shared_image" "clustergen2" {
 }
 
 resource "azurerm_shared_image_version" "cluster_image_version" {
-  name                = "0.0.1"
+  name                = var.azure_image_release
   gallery_name        = azurerm_shared_image.cluster.gallery_name
   image_name          = azurerm_shared_image.cluster.name
   resource_group_name = azurerm_shared_image.cluster.resource_group_name
@@ -144,7 +144,7 @@ resource "azurerm_shared_image_version" "cluster_image_version" {
 }
 
 resource "azurerm_shared_image_version" "clustergen2_image_version" {
-  name                = "0.0.1"
+  name                = var.azure_image_release
   gallery_name        = azurerm_shared_image.clustergen2.gallery_name
   image_name          = azurerm_shared_image.clustergen2.name
   resource_group_name = azurerm_shared_image.clustergen2.resource_group_name
--- a/data/data/azure/vnet/main.tf
+++ b/data/data/azure/vnet/main.tf
@@ -87,27 +87,75 @@ resource "azurerm_storage_blob" "rhcos_image" {
   metadata               = tomap({ source_uri = var.azure_image_url })
 }
 
-resource "azurerm_image" "cluster" {
+# Creates Shared Image Gallery
+# https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/shared_image_gallery
+resource "azurerm_shared_image_gallery" "sig" {
+  name                = "gallery_${replace(var.cluster_id, "-", "_")}"
+  resource_group_name = data.azurerm_resource_group.main.name
+  location            = var.azure_region
+}
+
+# Creates image definition
+# https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/shared_image
+resource "azurerm_shared_image" "cluster" {
   name                = var.cluster_id
+  gallery_name        = azurerm_shared_image_gallery.sig.name
   resource_group_name = data.azurerm_resource_group.main.name
   location            = var.azure_region
+  os_type             = "Linux"
 
-  os_disk {
-    os_type  = "Linux"
-    os_state = "Generalized"
-    blob_uri = azurerm_storage_blob.rhcos_image.url
+  identifier {
+    publisher = "RedHat"
+    offer     = "rhcos"
+    sku       = "basic"
   }
 }
 
-resource "azurerm_image" "clustergen2" {
+resource "azurerm_shared_image" "clustergen2" {
   name                = "${var.cluster_id}-gen2"
+  gallery_name        = azurerm_shared_image_gallery.sig.name
   resource_group_name = data.azurerm_resource_group.main.name
   location            = var.azure_region
+  os_type             = "Linux"
   hyper_v_generation  = "V2"
+  architecture        = var.azure_vm_architecture
 
-  os_disk {
-    os_type  = "Linux"
-    os_state = "Generalized"
-    blob_uri = azurerm_storage_blob.rhcos_image.url
+  identifier {
+    publisher = "RedHat-gen2"
+    offer     = "rhcos-gen2"
+    sku       = "gen2"
   }
 }
+
+resource "azurerm_shared_image_version" "cluster_image_version" {
+  name                = "0.0.1"
+  gallery_name        = azurerm_shared_image.cluster.gallery_name
+  image_name          = azurerm_shared_image.cluster.name
+  resource_group_name = azurerm_shared_image.cluster.resource_group_name
+  location            = azurerm_shared_image.cluster.location
+
+  blob_uri           = azurerm_storage_blob.rhcos_image.url
+  storage_account_id = azurerm_storage_account.cluster.id
+
+  target_region {
+    name                   = azurerm_shared_image.cluster.location
+    regional_replica_count = 1
+  }
+}
+
+resource "azurerm_shared_image_version" "clustergen2_image_version" {
+  name                = "0.0.1"
+  gallery_name        = azurerm_shared_image.clustergen2.gallery_name
+  image_name          = azurerm_shared_image.clustergen2.name
+  resource_group_name = azurerm_shared_image.clustergen2.resource_group_name
+  location            = azurerm_shared_image.clustergen2.location
+
+  blob_uri           = azurerm_storage_blob.rhcos_image.url
+  storage_account_id = azurerm_storage_account.cluster.id
+
+  target_region {
+    name                   = azurerm_shared_image.clustergen2.location
+    regional_replica_count = 1
+  }
+}
+
--- a/pkg/asset/machines/azure/machines.go
+++ b/pkg/asset/machines/azure/machines.go
@@ -9,6 +9,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"strings"
 
 	machineapi "github.com/openshift/api/machine/v1beta1"
 	icazure "github.com/openshift/installer/pkg/asset/installconfig/azure"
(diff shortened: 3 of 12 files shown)


data/data/azure/vnet/main.tf
```diff
@@ -91,27 +91,75 @@ resource "azurerm_storage_blob" "rhcos_image" {
...
-    blob_uri = azurerm_storage_blob.rhcos_image.url
+  identifier {
+    publisher = "RedHat-gen2"
+    offer     = "rhcos-gen2"
+    sku       = "gen2"
   }
 }
+
+resource "azurerm_shared_image_version" "cluster_image_version" {
+  name                = "0.0.1"
```
patrickdillon, line 135: I recognize that setting an arbitrary version is simpler, but we do actually know the rhcos version and it would be preferable to have the actual version here. 

I can think of two ways to get the version.

1. The RHCOS stream contains the `release`:
https://github.com/openshift/installer/blob/b8714c6d45d2700dcb44ddd86d2afb309e89fa35/pkg/asset/rhcos/image.go#L139
Similarly to the way we plumb the URL through the code base, we would also plumb the release.
2. The URL itself contains the version:
https://github.com/openshift/installer/blob/b8714c6d45d2700dcb44ddd86d2afb309e89fa35/data/data/coreos/rhcos.json#L768
It is a bit gnarly, but the version could be parsed from the URL

patrickdillon (commented):
In general this looks good. I can see that you're working through some errors in CI. The issue is not obvious, so feel free to provide an update or reach out if you want to discuss what's going on there.

I left a comment about setting the version. I think this is preferable, but I would not say i is a hard requirement. We can discuss this idea further. The first suggestion, in terms of piping the release through the codebase may be preferable to parsing the url, but may also be a bit more hairy to implement.

data/data/azure/variables-azure.tf
```diff
@@ -176,3 +176,9 @@ variable "azure_control_plane_vm_networking_type" {
   type        = bool
   description = "Whether to enable accelerated networking on control plane nodes."
 }
+
+variable "azure_vm_architecture" {
+  type        = string
+  description = "Architecture of the VMs - used when creating images in the image gallery"
+  default     = "x64"
```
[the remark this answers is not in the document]
Prashanth684: makes sense.

data/data/azure/vnet/main.tf
Prashanth684: ah..ok..i'll look into this. i am thinking I'll do a follow up PR for this if i find out it involves a bit of change..

Prashanth684:
> In general this looks good. I can see that you're working through some errors in CI. The issue is not obvious, so feel free to provide an update or reach out if you want to discuss what's going on there.
> 
> I left a comment about setting the version. I think this is preferable, but I would not say i is a hard requirement. We can discuss this idea further. The first suggestion, in terms of piping the release through the codebase may be preferable to parsing the url, but may also be a bit more hairy to implement.

yes the CI results concern  me a bit..the current run seems to be progressing fine but the failure was essentially:

```
time="2022-09-06T18:12:04Z" level=error msg="Error: creating/updating Shared Image: (Image Name \"ci-op-2h3vjhpr-2dc90-7tk4z\" / Gallery Name \"gallery_ci_op_2h3vjhpr_2dc90_7tk4z\" / Resource Group \"ci-op-2h3vjhpr-2dc90-7tk4z-rg\"): compute.GalleryImagesClient#CreateOrUpdate: Failure sending request: StatusCode=404 -- Original Error: Code=\"ParentResourceNotFound\" Message=\"Can not perform requested operation on nested resource. Parent resource 'gallery_ci_op_2h3vjhpr_2dc90_7tk4z' not found.\""
```

so the flow here is image gallery -> image definition(inside the gallery) -> image version (inside the image definition).

the error above indicates that the image gallery could not be found, but in the previous few lines i see:

```
time="2022-09-06T18:08:55Z" level=debug msg="azurerm_shared_image_gallery.sig: Creation complete after 34s [id=/subscriptions/d38f1e38-4bed-438e-b227-833f997adf6a/resourceGroups/ci-op-2h3vjhpr-2dc90-7tk4z-rg/providers/Microsoft.Compute/galleries/gallery_ci_op_2h3vjhpr_2dc90_7tk4z]"
time="2022-09-06T18:08:55Z" level=debug msg="azurerm_shared_image.clustergen2: Creating..."
time="2022-09-06T18:08:55Z" level=debug msg="azurerm_shared_image.cluster: Creating..."
```

so from a timeline perspective it seems ok. I'm still wondering if we need these resources to each have a `depends_on` argument just to eliminate the possibility of a race?

patrickdillon:
> so the flow here is image gallery -> image definition(inside the gallery) -> image version (inside the image definition).
> 
> the error above indicates that the image gallery could not be found, but in the previous few lines i see:
> 
> ```
> time="2022-09-06T18:08:55Z" level=debug msg="azurerm_shared_image_gallery.sig: Creation complete after 34s [id=/subscriptions/d38f1e38-4bed-438e-b227-833f997adf6a/resourceGroups/ci-op-2h3vjhpr-2dc90-7tk4z-rg/providers/Microsoft.Compute/galleries/gallery_ci_op_2h3vjhpr_2dc90_7tk4z]"
> time="2022-09-06T18:08:55Z" level=debug msg="azurerm_shared_image.clustergen2: Creating..."
> time="2022-09-06T18:08:55Z" level=debug msg="azurerm_shared_image.cluster: Creating..."
> ```
> 
> so from a timeline perspective it seems ok. I'm still wondering if we need these resources to each have a `depends_on` argument just to eliminate the possibility of a race?

This error is an example of a common class of errors we see in the upstream terraform providers. The terraform providers should handle these dependencies, but frequently there are bugs. We've had two similar bugs reported in the past week. 

Adding a `depends_on` may be a sufficient work around for this. It would also be good to open an issue in the upstream provider hashicorp/terraform-provider-azurerm. If the depends_on workaround relieves the symptoms, you could also open an installer bug for us to track the upstream work.

Digging into the upstream code: the creation of the image gallery is here:
https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/services/compute/shared_image_gallery_resource.go#L66

and the image itself is here:
https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/services/compute/shared_image_resource.go#L245

data/data/azure/vnet/main.tf
Prashanth684: well..i did try and make the changes but looks like the image gallery version names are pretty strict:
```
ERROR Error: Expected name to be in the format `1.2.3`, `latest`, or `recent` but got "412.86.202208101039". 
ERROR                                              
ERROR   with azurerm_shared_image_version.cluster_image_version, 
ERROR   on main.tf line 139, in resource "azurerm_shared_image_version" "cluster_image_version": 
ERROR  139:   name                = var.azure_image_release 

```
i removed the `-0` thinking that was the issue but still got this error. so looks like it only allows single digit versions

Prashanth684: we could use `latest` like the error says rather than an arbitrary string

Prashanth684: @patrickdillon thoughts ?

patrickdillon: Good idea: let's use latest.

Prashanth684: tried to use "latest", but then it wasn't happy with that either:
```
ERROR Error: creating Shared Image Version: (Version Name "latest" / Image Name "psundara-ocp-c99wz-gen2" / Gallery Name "gallery_psundara_ocp_c99wz" / Resource Group "psundara-ocp-c99wz-rg"): compute.GalleryImageVersionsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidParameter" Message="The image version name is invalid. The image version name should follow Major(int).Minor(int).Patch(int) format, where int is between 0 and 2,147,483,647 (inclusive). For e.g: 1.0.0, 2018.12.1 etc." Target="galleryImageVersionName"
``` 
but then the error message does say that the numbers must be between 0-2,147483647 so i just truncated the date in the rhcos string and was able to use the release image string

pkg/asset/machines/azure/machines.go
```diff
@@ -110,10 +111,13 @@ func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string
...
 	} else {
-		imageID := fmt.Sprintf("/resourceGroups/%s/providers/Microsoft.Compute/images/%s", rg, clusterID)
+		// image gallery names cannot have dashes
+		galleryName := strings.Replace(clusterID, "-", "_", -1)
+		id := clusterID
 		if hyperVGen == "V2" {
-			imageID += "-gen2"
+			id += "-gen2"
 		}
+		imageID := fmt.Sprintf("/resourceGroups/%s/providers/Microsoft.Compute/galleries/gallery_%s/images/%s/versions/0.0.1", rg, galleryName, id)
```
Prashanth684, line 120: oops! forgot this reference!

Prashanth684: fixed.

data/data/azure/variables-azure.tf
```diff
@@ -176,3 +176,15 @@ variable "azure_control_plane_vm_networking_type" {
...
 }
+
+variable "azure_vm_architecture" {
+  type        = string
+  description = "Architecture of the VMs - used when creating images in the image gallery"
+}
+
+variable "azure_image_release" {
+  type        = string
+  description = "Architecture of the VMs - used when creating images in the image gallery"
```
r4f4, line 187: Seems a c&p error from the above definition?

pkg/asset/rhcos/release.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+		azd := ext.AzureDisk
+		if azd == nil {
+			return "", fmt.Errorf("%s: No azure build found", st.FormatPrefix(archName))
+		}
+		return azd.Release, nil
+	default:
+		return "", nil
+	}
+}
```
r4f4, line 78: I'm wondering if it would make sense to add something like  
```suggestion
}

// image gallery image versions cannot have a "-" in the name and must be between 0-2147483647, so we have to truncate the hour and minutes of the date
func (r *Release) GetAzureReleaseVersion() string {
    imageVersion := string(r)
    if imageVersion != "" {
	    imageVersion = imageVersion[:len(imageVersion)-6]
	}
	return imageVersion
}
```
so we don't duplicate the logic and the explanation of why it's needed.

pkg/asset/machines/azure/machines.go
```diff
@@ -110,10 +111,20 @@ func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string
...
 		if hyperVGen == "V2" {
-			imageID += "-gen2"
+			id += "-gen2"
 		}
+		imageVersion := rhcosVersion
+		if imageVersion != "" {
+			// image gallery image versions cannot have a "-" in the name and must be between 0-2147483647, so had to truncate the hour and minutes of the date
+			imageVersion = imageVersion[:len(imageVersion)-6]
+		} else {
+			imageVersion = "0.0.1"
```
r4f4, line 125: Is this supposed to work or should it be an error scenario?

r4f4 (changes requested):

pkg/asset/rhcos/release.go
Prashanth684: hmm..yeah and it could be Azure specific...ok will do.

pkg/asset/machines/azure/machines.go
Prashanth684: i didn't want to error out but just use a default version...but i don't think this will ever be empty so might be worth erroring out.

data/data/azure/variables-azure.tf
Prashanth684: sigh..yeah

pkg/asset/machines/azure/machines.go
Prashanth684: the more i think about this, i don't think we need this default or to error out at all. the asset generation will error out if parsing fails and only other way it can be empty is if the json has no release string which i think we can safely assume it will. so i am not going to account for this. is that ok ?

pkg/tfvars/azure/azure.go
```diff
@@ -128,6 +130,10 @@ func TFVars(sources TFVarsSources) ([]byte, error) {
 		VMArchitecture:                  vmarch,
 	}
 
+	if sources.ImageRelease != "" {
+		cfg.ImageRelease = sources.ImageRelease
+	}
```
patrickdillon, line 135: Hm... do we need this or can you just add this to the literal above? 

Will sources.ImageRelease ever be ""? Even if it is I think the logic would be the same, because the empty value of cfg.ImageRelease is ""

pkg/asset/rhcos/release.go
Prashanth684: done.

pkg/tfvars/azure/azure.go
Prashanth684: yes removed the release image checks.

pkg/asset/machines/azure/machines.go
Prashanth684: cleaned up by removing the unnecessary checks.

aleskandro:
cc @lwan-wanglin

Prashanth684:
all azure tests are passing install phase now. looks like the issue is fixed somehow..will try again to confirm.

data/data/azure/vnet/main.tf
Prashanth684: used the rhcos release version by truncating the date

patrickdillon:
/override ci/prow/e2e-aws-ovn 
ci/prow/e2e-gcp-ovn 
ci/prow/e2e-vsphere-ovn

r4f4:
/override ci/prow/e2e-aws-ovn ci/prow/okd-scos-images ci/prow/okd-scos-unit ci/prow/okd-scos-verify-codegen ci/prow/e2e-agent-compact

PR #6306, opened 2022-09-06
fix agent-config-template emitted by `agent create`

joshgav:
The YAML emitted by `openshift-install agent create agent-config-template` is invalid, this fixes it as follows:

- indent rootDeviceHints.deviceName
- mark comment correctly


zaneb:
Thanks for this! LGTM but this branch has now merged back to master, so I think we'll want to re-propose this against master.

joshgav:
Thanks @zaneb, I switched this same PR to target master via the "Edit" option at the top.

zaneb:
> Thanks @zaneb, I switched this same PR to target master via the "Edit" option at the top.

I had no idea that was a thing you could do :laughing: 
/approve

joshgav:
@zaneb do you need to say `/lgtm`?
The test failures aren't related to the change.
LMK if anything further from me is needed, thanks!

PR #6307, opened 2022-09-06
data/manifests/bootkube/cvo-overrides: Default to stable-4.12

wking:
Like f995341c32 (#5621) and earlier, now that 4.11 has forked off from the development branch.


wking:
Woohoo, looks like #6302 is already in flight to do this 🎉

PR #6309, opened 2022-09-07
OCPBUGS-963: OpenStack: Lift validation for 14 chars cluster names

mandre:
The validation is no longer necessary now that we stopped using mDNS. Instead, rely on cluster name validation common to all platforms for length check and the on-prem cluster name validation for names containing dots.


mandre:
Because we're touching `pkg/types/validation/installconfig.go`, this change requires an approval from someone in the installer team.
/assign sadasu

PR #6311, opened 2022-09-07
OCPBUGS-927: Add depends to enforce order for azure terraform dependencies

barbacbd:
** Added depends for the azurerm_private_dns_zone_virtual_network_link.network to depend on "".private. This should enforce the dependencies.


patrickdillon (commented):
Code looks good and thanks for taking care of this so quickly.

This is an eventual consistentency bug, so we won't really be able to test except to merge this and then observe in CI.

`depends_on` should fix this issue unless there is a bug in the provider code for reading the resource. This is a fine workaround for the bug in the upstream provider, which is a bug upstream because the dependence between the two resources is explicit. Depends on should only be necessary when the dependency is hidden.

So I left a comment about adding a comment for the upstream bug, so that we can eventually remove this code when the fix merges upstream.

The commit message states what is changing, but in general it would be good for the messages to give context for why we are making the changes and can include details such as error message text and links.

PR #6318, opened 2022-09-08
OCPBUGS-816: Check that cached base ISP matches hash in release payload

bfournie:
For the base ISO that is cached, it needs to be verified that it matches the one in the release payload. This change first checks the sha256 hash in the rhcos.json file embedded in the installer. If that doesn't match the cached file it extracts the coreos-x86-64.iso.sha256 file from the release payload and compares that.  If neither match then a new iso will be retrieved from the release payload or downloaded.

### DIFF
--- a/pkg/asset/agent/image/oc.go
+++ b/pkg/asset/agent/image/oc.go
@@ -1,9 +1,13 @@
 package image
 
 import (
+	"bytes"
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"os"
 	"os/exec"
 	"path"
@@ -14,6 +18,7 @@ import (
 	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
 	"github.com/openshift/assisted-service/pkg/executer"
 	"github.com/openshift/installer/pkg/asset/agent/mirror"
+	"github.com/openshift/installer/pkg/rhcos"
 	"github.com/sirupsen/logrus"
 	"github.com/thedevsaddam/retry"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -21,8 +26,9 @@ import (
 )
 
 const (
-	machineOsImageName = "machine-os-images"
-	coreOsFileName     = "/coreos/coreos-%s.iso"
+	machineOsImageName   = "machine-os-images"
+	coreOsFileName       = "/coreos/coreos-%s.iso"
+	coreOsSha256FileName = "/coreos/coreos-%s.iso.sha256"
 	//OcDefaultTries is the number of times to execute the oc command on failues
 	OcDefaultTries = 5
 	// OcDefaultRetryDelay is the time between retries
@@ -37,7 +43,7 @@ type Config struct {
 
 // Release is the interface to use the oc command to the get image info
 type Release interface {
-	GetBaseIso(log logrus.FieldLogger, releaseImage string, pullSecret string, mirrorConfig []mirror.RegistriesConfig, architecture string) (string, error)
+	GetBaseIso(log logrus.FieldLogger, releaseImage, pullSecret, architecture string, mirrorConfig []mirror.RegistriesConfig) (string, error)
 }
 
 type release struct {
@@ -58,7 +64,7 @@ const (
 )
 
 // Get the CoreOS ISO from the releaseImage
-func (r *release) GetBaseIso(log logrus.FieldLogger, releaseImage string, pullSecret string, mirrorConfig []mirror.RegistriesConfig, architecture string) (string, error) {
+func (r *release) GetBaseIso(log logrus.FieldLogger, releaseImage, pullSecret, architecture string, mirrorConfig []mirror.RegistriesConfig) (string, error) {
 
 	// Get the machine-os-images pullspec from the release and use that to get the CoreOS ISO
 	image, err := r.getImageFromRelease(log, machineOsImageName, releaseImage, pullSecret, mirrorConfig)
@@ -73,19 +79,28 @@ func (r *release) GetBaseIso(log logrus.FieldLogger, releaseImage string, pullSe
 
 	filename := fmt.Sprintf(coreOsFileName, architecture)
 	// Check if file is already cached
-	filePath, err := GetFileFromCache(path.Base(filename), cacheDir)
+	cachedFile, err := GetFileFromCache(path.Base(filename), cacheDir)
 	if err != nil {
 		return "", err
 	}
-	if filePath != "" {
-		// Found cached file
-		return filePath, nil
+	if cachedFile != "" {
+		log.Info("Verifying cached file")
+		valid, err := r.verifyCacheFile(log, image, cachedFile, pullSecret, architecture, mirrorConfig)
+		if err != nil {
+			return "", err
+		}
+		if valid {
+			log.Infof("Using cached Base ISO %s", cachedFile)
+			return cachedFile, nil
+		}
 	}
 
+	// Get the base ISO from the payload
 	path, err := r.extractFileFromImage(log, image, filename, cacheDir, pullSecret, mirrorConfig)
 	if err != nil {
 		return "", err
 	}
+	log.Infof("Base ISO obtained from release and cached at %s", path)
 	return path, err
 }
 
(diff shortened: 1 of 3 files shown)


rwsu (approved):
I was able to verify that a new base ISO is extracted if its hash does not match the existing cached image. And also verified that the cached image is used if its hash matches the release image's sha256 hash.

pkg/asset/agent/image/oc.go
```diff
@@ -126,19 +138,54 @@ func (r *release) extractFileFromImage(log logrus.FieldLogger, image, file, cach
 		cmd = fmt.Sprintf(templateImageExtract, file, cacheDir, image)
 	}
 
-	log.Debugf("extracting %s to %s, %s", file, cacheDir, cmd)
+	logrus.Debugf("extracting %s to %s, %s", file, cacheDir, cmd)
```
zaneb, line 141: Why wouldn't we use `log` here?

pkg/asset/agent/image/oc.go
```diff
@@ -72,19 +75,28 @@ func (r *release) GetBaseIso(log logrus.FieldLogger, releaseImage string, pullSe
...
+			return cachedFile, nil
+		}
 	}
 
+	// Get the base ISO from the payload
 	path, err := r.extractFileFromImage(log, image, filename, cacheDir, pullSecret, mirrorConfig)
 	if err != nil {
 		return "", err
 	}
+	log.Infof("Base ISO obtained from release and cached at %s", path)
```
zaneb, line 99: What's the thinking behind printing the cache path in the info log (vs. debug before)?

pkg/asset/agent/image/oc.go
```diff
@@ -126,19 +138,54 @@ func (r *release) extractFileFromImage(log logrus.FieldLogger, image, file, cach
...
+	}
+
+	// Get the file containing the coreos sha256 hash
+	shaFilename := fmt.Sprintf(coreOsSha256FileName, architecture)
+	shaFile, err := r.extractFileFromImage(log, image, shaFilename, tempDir, pullSecret, mirrorConfig)
+	if err != nil {
+		return false, err
+	}
+
+	defer os.RemoveAll(tempDir)
```
zaneb, line 166: This should be called (er, deferred) before we extract the hash file to the tempdir.

pkg/asset/agent/image/oc.go
```diff
@@ -126,19 +138,54 @@ func (r *release) extractFileFromImage(log logrus.FieldLogger, image, file, cach
...
+
+	payloadSha256, err := ioutil.ReadFile(shaFile)
```
zaneb, line 168: Apparently `ioutil` has been deprecated, so for new code prefer `os.ReadFile`

pkg/asset/agent/image/oc.go
```diff
@@ -126,19 +138,54 @@ func (r *release) extractFileFromImage(log logrus.FieldLogger, image, file, cach
...
+	if err != nil {
+		return false, err
+	}
+
+	data, err := ioutil.ReadFile(file)
```
zaneb, line 173: This will read the entire 1GB file into memory, which is probably not what you want. Instead, create a checksum object with `sha256.New()` and then use `io.Copy()` to write the data to it. The `Sum` field contains the hash.

Example: https://pkg.go.dev/crypto/sha256@go1.19.1#example-New-File

pkg/asset/agent/image/oc.go
```diff
@@ -126,19 +138,54 @@ func (r *release) extractFileFromImage(log logrus.FieldLogger, image, file, cach
...
+	}
+	fileHash := sha256.Sum256(data)
+
+	// Compare the hash of cached file to the hash retrieved from payload
+	if hex.EncodeToString(fileHash[:]) == string(payloadSha256) {
+		// cached file is the same as in payload
+		return true, nil
+	}
+	logrus.Debugf("Cached file %s is not most recent", file)
+	return false, err
```
zaneb, line 185: `err` is always `nil` here though, right?

pkg/asset/agent/image/oc.go
```diff
@@ -126,19 +138,54 @@ func (r *release) extractFileFromImage(log logrus.FieldLogger, image, file, cach
...
+	}
+
+	data, err := ioutil.ReadFile(file)
+	if err != nil {
+		return false, err
+	}
+	fileHash := sha256.Sum256(data)
+
+	// Compare the hash of cached file to the hash retrieved from payload
+	if hex.EncodeToString(fileHash[:]) == string(payloadSha256) {
```
zaneb, line 180: It might pay to use `hex.DecodeString()` on `payloadSha256` instead, since there's multiple ways to encode bytes as a hex string.

pkg/asset/agent/image/oc.go
```diff
@@ -126,19 +138,54 @@ func (r *release) extractFileFromImage(log logrus.FieldLogger, image, file, cach
...
+	tempDir, err := ioutil.TempDir("", "cache")
+	if err != nil {
+		return false, err
+	}
+
+	// Get the file containing the coreos sha256 hash
+	shaFilename := fmt.Sprintf(coreOsSha256FileName, architecture)
+	shaFile, err := r.extractFileFromImage(log, image, shaFilename, tempDir, pullSecret, mirrorConfig)
+	if err != nil {
+		return false, err
```
zaneb, line 163: I think this should return `false, nil`, because as it stands if we fail to extract the checksum file from the image we will fail the entire operation, but we really want to continue to fall back to trying to download the image.

pkg/asset/agent/image/oc.go
```diff
@@ -126,19 +138,54 @@ func (r *release) extractFileFromImage(log logrus.FieldLogger, image, file, cach
...
 	path := filepath.Join(cacheDir, path.Base(file))
-	log.Info("Successfully extracted base ISO from the release")
-	log.Debugf("Base ISO %s cached at %s", file, path)
 	return path, nil
 }
 
-func execute(log logrus.FieldLogger, executer executer.Executer, pullSecret string, command string) (string, error) {
+// Check if there is a different base ISO in the release payload
+func (r *release) verifyCacheFile(log logrus.FieldLogger, image, file, pullSecret, architecture string, mirrorConfig []mirror.RegistriesConfig) (bool, error) {
+
```
zaneb, line 153: We could avoid the (expensive) extraction of the checksum file in most cases by comparing to the checksum in the rhcos metadata built in to the installer, and only checking the release payload if that doesn't match.

pkg/asset/agent/image/oc.go
bfournie: Its useful to see where the cache file is located in the non-debug case - especially when this file has had to be removed to get past this bug, so I think we should leave it here as Info.
As to why the entire path isn't in debug - we do have the file file there as part of the --path argument, it seemed like the entire path wasn't necessary if it was already being printed in the Info. I'm willing to change if there's a more useful way to do it.

pkg/asset/agent/image/oc.go
bfournie: Good catch, should be log, fixing.

pkg/asset/agent/image/oc.go
bfournie: Yep, moving.

pkg/asset/agent/image/oc.go
bfournie: I see... https://pkg.go.dev/io/ioutil - "Deprecated: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code."

Thanks for pointing it out, changing.

pkg/asset/agent/image/oc.go
bfournie: Yes, changed to make it obvious.

pkg/asset/agent/image/oc.go
bfournie: OK, used DecodeString()

pkg/asset/agent/image/oc.go
bfournie: Yeah, good point, fixed.

pkg/asset/agent/image/oc.go
bfournie: Yeah that saves time. Changed to check rhcos.json for hash first.

pkg/asset/agent/image/oc.go
bfournie: Good idea, changed.

pkg/asset/agent/image/oc.go
```diff
@@ -145,12 +160,93 @@ func (r *release) extractFileFromImage(log logrus.FieldLogger, image, file, cach
...
+	if err != nil {
+		return false, err
+	}
+
+	defer os.RemoveAll(tempDir)
+
+	shaFilename := fmt.Sprintf(coreOsSha256FileName, architecture)
+	shaFile, err := r.extractFileFromImage(log, image, shaFilename, tempDir, pullSecret, mirrorConfig)
+	if err != nil {
+		return false, nil
```
zaneb, line 233: A debug log here would be good.

pkg/asset/agent/image/oc.go
```diff
@@ -145,12 +160,93 @@ func (r *release) extractFileFromImage(log logrus.FieldLogger, image, file, cach
...
+			return true, format.Disk.Sha256
+		}
+	}
+
+	return false, ""
+}
+
+func matchingHash(imageSha []byte, sha string) bool {
+	decoded, err := hex.DecodeString(sha)
+	if err == nil && reflect.DeepEqual(imageSha, decoded) {
```
zaneb, line 193: nit: could use `bytes.Equal()` here rather than resorting to the reflect module.

zaneb (commented):
Looks good, thanks!

pkg/asset/agent/image/oc.go
bfournie: Done

pkg/asset/agent/image/oc.go
bfournie: Done

PR #6317, opened 2022-09-08
OCPBUGSM-47740: No networkconfig needed with DHCP agent config

pawanpinjarkar:
When creating DHCP only cluster with agent config ( i.e. only RendezvousIP), do not require any networkConfig to be provided.

PR #6323, opened 2022-09-09
OCPBUGSM-47737: Do not enable pre-network-manager-config.service

rwsu:
when there are no nmstateconfigs defined.

When DHCP is used, agent-config.yaml may have no networkConfigs defined, and therefore no nmstateconfigs will be generated. When nmstateconfigs are absent, the pre-network-manager-config systemd service should not be enabled.

### DIFF
--- a/pkg/asset/agent/image/ignition.go
+++ b/pkg/asset/agent/image/ignition.go
@@ -59,22 +59,6 @@ type agentTemplateData struct {
 	InfraEnvID          string
 }
 
-var (
-	agentEnabledServices = []string{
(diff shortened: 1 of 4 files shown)


rwsu:
Depends on #6317

rwsu:
The unit tests are failing because nmstatectl isn't available on some of the ci test environment. I will add a test skip to not run the Generate test if nmstatectl can't be found in the environment. We already have an existing test skip in TestIgnition_addStaticNetworkConfig for the same issue.

PR #6325, opened 2022-09-09
CORS-2280: IBMCloud: Add DNS Service permitted VPC

cjschaef:
Determine whether a VPC is already a permitted network
for Private/DNS (Internal) clusters for the DNS Service.
This can occur when using BYON and Private clusters.

Partial: https://issues.redhat.com/browse/CORS-2280

### DIFF
--- a/pkg/asset/installconfig/ibmcloud/metadata.go
+++ b/pkg/asset/installconfig/ibmcloud/metadata.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 
 	"github.com/IBM/go-sdk-core/v5/core"
+	"github.com/pkg/errors"
 
 	"github.com/openshift/installer/pkg/types"
 )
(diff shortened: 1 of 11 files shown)


pkg/asset/installconfig/ibmcloud/metadata.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/IBM/go-sdk-core/v5/core"
 
 	"github.com/openshift/installer/pkg/types"
+	"github.com/pkg/errors"
```
rvanderp3, line 11:
```suggestion
	"github.com/IBM/go-sdk-core/v5/core"
	"github.com/pkg/errors"
	
	"github.com/openshift/installer/pkg/types"
```

cjschaef: Ah, yep I can fix that

cjschaef: Fixed

PR #6330, opened 2022-09-11
GCP: allow environmental authentication

patrickdillon:
Enables environmental authentication, so that the Installer when running on a GCP VM can authenticate with the VM's service account without requiring a service account in a JSON file. 

The biggest challenge here is that the signed URLs used to enable accessing bootstrap ignition require the private key from a service account JSON file in order to sign the URL. To solve this, I updated terraform to create a service account specifically for bootstrapping, and use the private key from that. 

I explored the idea of granting explicit permission for the bootstrap node to access the bootstrap ignition object, but there is not a clear way to do this. Unless the object is public, the bootstrap node would need to pass an authentication token in the header when requesting access to the object, but the Terraform ignition provider does not support adding headers: so there is no way to pass the token when making the request.

The current solution is more in-line with the existing install procedure, which already uses signed urls. 

https://issues.redhat.com/browse/CORS-2050


patrickdillon:
Reworked the validation logic to allow better unit testing.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -284,3 +285,17 @@ func validateRegion(client API, ic *types.InstallConfig, fieldPath *field.Path)
...
 	return nil
 }
+
+// validateCredentialMode checks whether the credential mode is
+// compatible with the authentication mode.
+func validateCredentialMode(client API, ic *types.InstallConfig) field.ErrorList {
+	allErrs := field.ErrorList{}
+	creds := client.GetCredentials()
+
+	if creds.JSON == nil && ic.CredentialsMode != types.ManualCredentialsMode {
```
barbacbd, line 295: Should we check the credentials mode first? If the creds mode is not correct error immediately without calling `GetCredentials()`.

patrickdillon: The "correctness" of the credentials depends on their contents, so we have to grab them.

PR #6331, opened 2022-09-12
OCPBUGS-1247: Restrict CPMS from being created with single node clusters

rna-afk:
Single node clusters do not need the CPMS manifests and hence adding
a check to stop the creation of CPMS when the Replicas is set to
anything below 1.

Ideally, the replicas are set to 0 when the installer needs to create
single node clusters.


patrickdillon:
/approve

All eyes on single-node tests...

PR #6336, opened 2022-09-12
pkg/types: add feature set support

patrickdillon:
Adds basic support for feature sets in the install config, as well as validation for gated features.

See openshift/enhancements#1239


patrickdillon:
/wip

Needs further work for `explain`

patrickdillon:
Oops. Need to run go gen to bump explain

data/data/manifests/openshift/feature-gate.yaml.template
```diff
@@ -0,0 +1,6 @@
+apiVersion: config.openshift.io/v1
```
wking, line 1: We're vendoring this type, and also importing that directory in `cluster.go`.  It might be about as easy to set up this asset based on the Go type without going through the template system, and that will also get you nice compile-time warnings if an openshift/api vendor bump ever drops v1 support, without you needing to come touch a template string that isn't checked at compile-time.

wking: [The Proxy config asset][1] is an example of the "use the Go type" approach, in case that's useful.

[1]: https://github.com/openshift/installer/blob/503e90075a69ea891019ea2e3631806aa935bd0a/pkg/asset/manifests/proxy.go

pkg/asset/cluster/cluster.go
```diff
@@ -74,6 +75,10 @@ func (c *Cluster) Generate(parents asset.Parents) (err error) {
 	terraformVariables := &TerraformVariables{}
 	parents.Get(clusterID, installConfig, terraformVariables)
 
+	if fs := installConfig.Config.FeatureSet; fs != configv1.Default {
+		logrus.Warnf("FeatureSet %q is enabled. This may affect the supportability of the cluster.", fs)
```
wking, line 79: nit: are there any feature sets that do not affect supportability?  I realize you say `may` here, but it's also `warn`.  One generic approach would be `Infof` without the supportability sentence.  Another approach would be adding the supportability caveat only for an allow-list of support-impacting feature sets.  Probably other alternatives too.  But if there are supported feature sets, we don't want to spook users who enable them.

pkg/types/validation/installconfig.go
```diff
@@ -744,3 +746,30 @@ func validateCapabilities(c *types.Capabilities, fldPath *field.Path) field.Erro
...
+
+	validFeatureSets := func() []string {
+		v := []string{}
+		for n := range configv1.FeatureSets {
+			v = append(v, string(n))
+		}
+		return v
+	}()
+
+	if !sets.NewString(validFeatureSets...).Has(string(c.FeatureSet)) {
```
wking, line 762: nit: [`configv1.FeatureSets` is a `map`][1], so you can probably defer `validFeatureSets` and skip `sets.NewString` and use something like:

```go
if _, ok := configv1.FeatureSets[c.FeatureSet]; !ok {
  ...build validFeatureSets in here, for use in the error message...
  allErrs = append(allErrs, field.NotSupported(field.NewPath("featureSet"), c.FeatureSet, validFeatureSets))
}
```

[1]: https://github.com/openshift/api/blob/66f8283afa3b5761a9c26f76137d074d0a43bfc9/config/v1/types_feature.go#L105

wking: Also, after building `validFeatureSets`, you might want to `sort.Strings(validFeatureSets)` so the ordering in the error message was stable.

pkg/types/validation/installconfig.go
```diff
@@ -744,3 +746,30 @@ func validateCapabilities(c *types.Capabilities, fldPath *field.Path) field.Erro
...
+		allErrs = append(allErrs, field.NotSupported(field.NewPath("featureSet"), c.FeatureSet, validFeatureSets))
+	}
+
+	if c.FeatureSet != configv1.TechPreviewNoUpgrade {
+		errMsg := "the TechPreviewNoUpgrade feature set must be enabled to use this field"
```
wking, line 767: nit: you can defer this error message until you're inside the `NetworkProjectID` `if` block below.

wking: ah, never mind.  I see this is a genric error message that could be shared with future tech-preview properties.

pkg/asset/manifests/featuregate.go
```diff
@@ -0,0 +1,86 @@
...
+
+// Generate generates the FeatureGate CRD.
+func (f *FeatureGate) Generate(dependencies asset.Parents) error {
+	installConfig := &installconfig.InstallConfig{}
+	dependencies.Get(installConfig)
+
+	// A FeatureGate could be populated on every install,
+	// even for those using the default feature set, but for
+	// continuity let's only create a cluster feature gate
+	// when non-default feature gates are enabled.
```
wking, line 46: one potential wrinkle here is that implicit installer opinions will slip past #6247's opinion-collision guard.  That's probably fine in the "install-config didn't care" case, but may be surprising in the "install-config was explicitly...", oh, I guess [`Default` is the empty string][1].  So in all cases where the install-config set a non-empty string, we'll be creating this manifest, and getting opinion-collision guards.  Works for me :+1:

[1]: https://github.com/openshift/api/blob/66f8283afa3b5761a9c26f76137d074d0a43bfc9/config/v1/types_feature.go#L29-L30

pkg/asset/manifests/featuregate.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// Files returns the files generated by the asset.
+func (f *FeatureGate) Files() []*asset.File {
+	return f.FileList
+}
+
+// Load loads the already-rendered files back from disk.
+func (f *FeatureGate) Load(ff asset.FileFetcher) (bool, error) {
+	return false, nil
```
wking, line 85: I would expect folks interested in touching this would use the install-config knob.  But it might be worth implementing the `Load` step just in case they try to edit `99_feature-gate.yaml`?

pkg/asset/manifests/featuregate.go
patrickdillon: > So in all cases where the install-config set a non-empty string, we'll be creating this manifest, and getting opinion-collision guards. 

Correct

> Works for me +1 But creating the manifest every time would work for me too ;)

Yeah I debated this (which is why I included the comment). My opinion is pretty weakly held (i.e. I think either option is fine). I landed on not writing the manifest for the default feature set for two reasons:

1. maintain consistency
2. I have an assumption, that may be mistaken, that the existence of the featuregate resource may be a signal in and of itself. I would hope people wouldn't do an `oc get featuregate` without doing a `describe` too, but I highly doubt anyone is out there creating featuregate resources for a default featureset. So my reasoning was that omitting it in this case would be less confusing.

patrickdillon:
@jcpowermac @rvanderp3 https://github.com/openshift/installer/pull/6336/commits/0d9b518f42f24eea68231bfe3622d8929d1054a6 sets vsphere multizone as tech preview. PTAL

jcpowermac:
> @jcpowermac @rvanderp3 [0d9b518](https://github.com/openshift/installer/commit/0d9b518f42f24eea68231bfe3622d8929d1054a6) sets vsphere multizone as tech preview. PTAL

Looks great, thanks @patrickdillon !

rvanderp3:
> > @jcpowermac @rvanderp3 [0d9b518](https://github.com/openshift/installer/commit/0d9b518f42f24eea68231bfe3622d8929d1054a6) sets vsphere multizone as tech preview. PTAL
> 
> Looks great, thanks @patrickdillon !

very cool, lgtm.  thanks Patrick!

pkg/types/validation/installconfig.go
```diff
@@ -976,3 +978,52 @@ func validateAdditionalCABundlePolicy(c *types.InstallConfig) error {
...
+			sort.Strings(v)
+			return v
+		}()
+		allErrs = append(allErrs, field.NotSupported(field.NewPath("featureSet"), c.FeatureSet, sortedFeatureSets))
+	}
+
+	if c.FeatureSet != configv1.TechPreviewNoUpgrade {
+		errMsg := "the TechPreviewNoUpgrade feature set must be enabled to use this field"
+
+		if c.GCP != nil && len(c.GCP.NetworkProjectID) > 0 {
```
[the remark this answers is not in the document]
patrickdillon: Only if those features are "tech preview." I think those were going to be released as GA (not tech preview), but if they are tech preview we can add them (in a later PR).

wking (approved):
/lgtm

`docs/` entries to come in follow-up work.

pkg/asset/manifests/openshift.go
```diff
@@ -307,6 +309,7 @@ func (o *Openshift) Generate(dependencies asset.Parents) error {
 	}
 
 	o.FileList = append(o.FileList, openshiftInstall.Files()...)
+	o.FileList = append(o.FileList, featureGate.Files()...)
```
sadasu, line 312: Since FeatureGate's Generate() could return nil, do we want to check for that before calling the append()?

patrickdillon: `append` should gracefully handle this case:
https://go.dev/play/p/tOrMzBn_wS_I

pkg/types/validation/installconfig.go
```diff
@@ -976,3 +978,52 @@ func validateAdditionalCABundlePolicy(c *types.InstallConfig) error {
...
+				v = append(v, string(n))
+			}
+			sort.Strings(v)
+			return v
+		}()
+		allErrs = append(allErrs, field.NotSupported(field.NewPath("featureSet"), c.FeatureSet, sortedFeatureSets))
+	}
+
+	if c.FeatureSet != configv1.TechPreviewNoUpgrade {
+		errMsg := "the TechPreviewNoUpgrade feature set must be enabled to use this field"
```
sadasu, line 999: Could CustomNoUpgrade also be allowed? Could there be a scenario where a custome feature and a tech preview need to be tested together?

PR #6326, opened 2022-09-09
CORS-2043: GCP: Passsthrough installer service account to Terraform r…

AnnaZivkovic:
…esources

**skip creating service accounts in Terraform when using passthrough credentialsMode for control and compute plane. Instead, pass the installer service account to Terraform to be used as the service account for instances 
** when using passthrough credentialsMode the machinesets email is modified to use the installer service account

PR #6338, opened 2022-09-13
GCP: Add createFirewallRules parameter to installconfig.

jstuever:
The createFirewallRules parameter in the install config can be used to toggle the creation of the firewall rules. This enables users to provide their own rules in advance of the installer and skip the installer from attempting to create them.

https://issues.redhat.com/browse/CORS-2288

### DIFF
--- a/data/data/gcp/bootstrap/main.tf
+++ b/data/data/gcp/bootstrap/main.tf
@@ -50,10 +50,11 @@ resource "google_compute_address" "bootstrap" {
 }
 
 resource "google_compute_firewall" "bootstrap_ingress_ssh" {
-  count       = var.gcp_network_project_id != "" ? 0 : 1
+  count       = var.gcp_create_firewall_rules ? 1 : 0
   name        = "${var.cluster_id}-bootstrap-in-ssh"
   network     = var.network
   description = local.description
+  project     = var.gcp_network_project_id
 
   allow {
     protocol = "tcp"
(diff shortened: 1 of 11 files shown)


jstuever:
/hold
Wait for https://github.com/openshift/installer/pull/6326 to merge first as a priority...

r4f4 (commented):
I like the explicit `create_firewall_rules` instead of using `network_project_id`

/lgtm

jstuever:
/hold
I want to see e2e-gcp-ovn pass before this merges.

patrickdillon:
Should we restrict this to xpn only?

jstuever:
> Should we restrict this to xpn only?

I don't believe so. This can be beneficial any time a user pre-creates firewall rules and wants to use them (possibly with network tags) instead of the installer provisioned ones. It's an edge case, but no need to exclude it.

jstuever:
/unhold
Just the ususal flakes.

jstuever:
/override ci/prow/e2e-azure-ovn ci/prow/e2e-vsphere-ovn 
The only file where this overlaps with other platforms is in `pkg/asset/cluster/tfvars.go`, and the code changes there are safely in the gcp section of that file. It is highly unlikely this pr would affect these tests.

sadasu:
For the same reason as provided in https://github.com/openshift/installer/pull/6338#issuecomment-1246146109

PR #6339, opened 2022-09-13
OCPBUGS-1231: disconnected machine-os-images info

celebdor:
When in completely disconnected environments, we need to get the image pullspec information using the icsp file, otherwise it will attempt to reach out to quay to find out the image for the base rhcos image.

### DIFF
--- a/pkg/asset/agent/image/oc.go
+++ b/pkg/asset/agent/image/oc.go
@@ -52,6 +52,7 @@ func NewRelease(executer executer.Executer, config Config) Release {
 
 const (
 	templateGetImage             = "oc adm release info --image-for=%s --insecure=%t %s"
+	templateGetImageWithIcsp     = "oc adm release info --image-for=%s --insecure=%t --icsp-file=%s %s"
 	templateImageExtract         = "oc image extract --path %s:%s --confirm %s"
 	templateImageExtractWithIcsp = "oc image extract --path %s:%s --confirm --icsp-file=%s %s"
 )
@@ -60,7 +61,7 @@ const (
 func (r *release) GetBaseIso(log logrus.FieldLogger, releaseImage string, pullSecret string, mirrorConfig []mirror.RegistriesConfig, architecture string) (string, error) {
 
 	// Get the machine-os-images pullspec from the release and use that to get the CoreOS ISO
-	image, err := r.getImageFromRelease(log, machineOsImageName, releaseImage, pullSecret, len(mirrorConfig) > 0)
+	image, err := r.getImageFromRelease(log, machineOsImageName, releaseImage, pullSecret, mirrorConfig)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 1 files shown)


pkg/asset/agent/image/oc.go
```diff
@@ -88,19 +89,29 @@ func (r *release) GetBaseIso(log logrus.FieldLogger, releaseImage string, pullSe
...
+		if len(mirrorConfig) > 0 {
 			log.Warning("Unable to validate mirror config because \"oc\" command is not available")
 		} else {
 			log.Debug("Skipping ISO extraction; \"oc\" command is not available")
 		}
 		return "", err
 	}
 
-	cmd := fmt.Sprintf(templateGetImage, imageName, true, releaseImage)
+	if len(mirroConfig) > 0 {
```
andfasano, line 104: Any specific reason to create/delete twice the icsp file? It could be generated just once at the beginning of `GetBaseIso` IINW

pkg/asset/agent/image/oc.go
```diff
@@ -88,19 +89,29 @@ func (r *release) GetBaseIso(log logrus.FieldLogger, releaseImage string, pullSe
...
+		log.Debugf("Using mirror configuration")
+		icspFile, err := getIcspFileFromRegistriesConfig(log, mirrorConfig)
+		if err != nil {
+			return "", err
+		}
+		defer removeIcspFile(icspFile)
+		cmd := fmt.Sprintf(templateGetImageWithIcsp, imageName, true, icspFile, releaseImage)
+	} else {
+		cmd := fmt.Sprintf(templateGetImage, imageName, true, releaseImage)
+	}
```
rwsu, line 114:
```suggestion
	cmd := fmt.Sprintf(templateGetImage, imageName, true, releaseImage)
	if len(mirrorConfig) > 0 {
		log.Debugf("Using mirror configuration")
		icspFile, err := getIcspFileFromRegistriesConfig(log, mirrorConfig)
		if err != nil {
			return "", err
		}
		defer removeIcspFile(icspFile)
		cmd = fmt.Sprintf(templateGetImageWithIcsp, imageName, true, icspFile, releaseImage)
	}
```

rwsu: mirroConfig -> mirrorConfig
cmd needs to initialized outside of 
`if len(mirrorConfig) > 0 {`

celebdor: thanks. I just edited that and testing again

celebdor: I had the go mod cache gone. So I had to download the whole thing before the build (so pushed early)

pkg/asset/agent/image/oc.go
celebdor: and add it to the signature of the other two functions? It can be done.

PR #6340, opened 2022-09-13
OCPBUGS-1226: OpenStack UPI: Create server group for Computes

mandre:
For consistency with IPI, and so that the MachineSet documentation works the same for both installation methods, we should create a server group for Compute nodes.

The server group is named `<infra-ID>-worker`, and uses the `soft-anti-affinity`, similar to what IPI does.


pierreprinetti:
nice! thanks.
/lgtm

PR #6341, opened 2022-09-13
OCPBUGS-1242: Add PowerVS zone mon01 support

hamzy:
The 4.11 branch of the installer needs to support the zone mon01 in PowerVS for the CI

### DIFF
--- a/data/data/powervs/cluster/power_network/pi_network.tf
+++ b/data/data/powervs/cluster/power_network/pi_network.tf
@@ -4,9 +4,10 @@ resource "ibm_pi_dhcp" "dhcp_service" {
 }
 
 resource "ibm_pi_cloud_connection" "cloud_connection" {
-  pi_cloud_instance_id            = var.cloud_instance_id
-  pi_cloud_connection_name        = "cloud-con-${var.cluster_id}"
-  pi_cloud_connection_speed       = 50
-  pi_cloud_connection_vpc_enabled = true
-  pi_cloud_connection_vpc_crns    = [var.vpc_crn]
+  pi_cloud_instance_id               = var.cloud_instance_id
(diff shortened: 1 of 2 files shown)


hamzy:
@rna-afk can we get an approve?

hamzy:
@jhixson74 Hello.  How do I get the "backport-risk-assessed" label as this bug is not a cherry pick.

r4f4:
> @jhixson74 Hello. How do I get the "backport-risk-assessed" label as this bug is not a cherry pick.

Technically it is because you're applying a change in master to an older branch. Anyway you need a dev to give you the `backport-risk-assessed` and a QE to give `cherry-pick-approved` labels.

hamzy:
@r4f4 
>  Anyway you need a dev to give you the ...
I guess I am not a dev in this scenario?  If not, would that dev be in the approvers section?

Prashanth684:
> @r4f4
> 
> > Anyway you need a dev to give you the ...
> 
> I guess I am not a dev in this scenario? If not, would that dev be in the approvers section?

do we need this support in 4.11? given that powervs is not supported in 4.11

hamzy:
> do we need this support in 4.11? given that powervs is not supported in 4.11

Yes!  There is only one properly fixed zone so far in PowerVS and that is mon01.  While it may not be supported it is used by developers.  Which is why I created this PR.

hamzy:
@openshift-ci-robot Empty set [], nice! :)

r4f4:
> @r4f4
> 
> > Anyway you need a dev to give you the ...
> 
> I guess I am not a dev in this scenario? If not, would that dev be in the approvers section?

@jhixson74 or @patrickdillon should be able to set the label.

PR #6344, opened 2022-09-13
Revert "[release-4.11] OCPBUGS-1157: Add depends to enforce order for azure terraform dependencies"

sdodson:
Reverts openshift/installer#6333 which broke Azure installs but those jobs weren't blocking.


sdodson:
/hold
The followup fix looks like we should know whether or not it worked within the next 5-10 minutes.

PR #6345, opened 2022-09-14
OCPBUGS-1234 : [CFE-580] Extend user tags limit to 40 based on AWS limits

TrilokGeer:
Extend user tag limit to 40 based on AWS limits. 10 tags are reserved for OpenShift use. Previously, there has been a confusion on the tag limit for S3 bucket vs S3 bucket object tag limit. S3 bucket object tag limit is 10 while the requirement is for S3 bucket which has limit of 50 tags.


TrilokGeer:
Due to github metadata corruption, previous PR had issues, so recreated branch and opened new PR. Previous discussion and PR details can be found here : https://github.com/openshift/installer/pull/6274

PR #6355, opened 2022-09-15
OCPBUGS-1349: OpenStack: Lift validation for 14 chars cluster names

mandre:
Manual cherry-pick of https://github.com/openshift/installer/pull/6309/.

The validation is no longer necessary now that we stopped using mDNS. Instead, rely on cluster name validation common to all platforms for length check and the on-prem cluster name validation for names containing dots.

### DIFF
--- a/pkg/types/openstack/validation/platform.go
+++ b/pkg/types/openstack/validation/platform.go
@@ -3,7 +3,6 @@ package validation
 import (
 	"errors"
 	"net"
-	"strings"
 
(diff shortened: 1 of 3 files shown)


mandre:
@sadasu shouldn't you be in [the 4.11 approvers list](https://github.com/openshift/installer/blob/release-4.11/OWNERS_ALIASES#L4-L10)?

/assign @sdodson for approval, as the bot suggests.

sadasu:
@mandre this needs cherry-pick-approved label too.

PR #6360, opened 2022-09-15
[WIP] Increase agent wait-for bootstrap timeout

andfasano:
/draft

Just for testing, not to be merged

### DIFF
--- a/pkg/agent/waitfor.go
+++ b/pkg/agent/waitfor.go
@@ -22,7 +22,7 @@ func WaitForBootstrapComplete(assetDir string) (*Cluster, error) {
 
 	start := time.Now()
 	previous := time.Now()
-	timeout := 30 * time.Minute
+	timeout := 60 * time.Minute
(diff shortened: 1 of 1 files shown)


bfournie:
FYI - I've been running with this same change in my local dev-scripts setup for weeks now. It was the only way I could get the IPv6 test to pass. Without this change it was failing 100% of the time and with it, it passes every time.

lranjbar:
This could be hardware dependent. In my local testing bootstrap takes around 20 mins, but I had originally a longer timeout to account for some buffer. IMHO it is fine to increase it.

zaneb:
https://github.com/openshift/installer/pull/6362 seems to make it clear that the cause is a regression in assisted-service, but it's still surprising that this increase in timeout would cause the whole test to hit the 4 hour limit, so let's see if it happens again.
/test e2e-agent-compact

andfasano:
Looks like the problem was due a conflict check for schedulable master manifests(between assisted service and installer), so closing this PR

PR #6362, opened 2022-09-15
DNM: Test with 4.11 assisted-service image

zaneb:


zaneb:
Looks like the problem is an interaction between https://github.com/openshift/assisted-service/pull/4011 and https://github.com/openshift/installer/pull/6247 on topologies where the masters are schedulable (i.e. those with <5 nodes).
/close

PR #6365, opened 2022-09-16
powervs-delete-named-dhcp

hamzy:
Delete named DHCP service for PowerVS

The DHCP service went from:
DHCPSERVER2dc32880758344f08c8ff6933e87d27a_Private to:
DHCPSERVERrdr-maocp-lon06-hszgr_Private
in order to support multiple clusters in a zone.  Therefore, we need to delete only a specific DHCP service.


clnperez:
@rna-afk can we get an approve?

PR #6366, opened 2022-09-16
Fix generating Machine objects for PowerVS

hamzy:
This fixes the following error:
error getting network ID: failed to find an network ID with RegEx ^DHCPSERVER[0-9a-z]{32}_Private$ The DHCP service has changed the name it created to include the cluster infraID.


clnperez:
@rna-afk can you approve? this one is a fix we need asap. Thanks!

PR #6367, opened 2022-09-16
hack: Use golang-1.16 image of origin-release

eramoto:
go-lint.sh outputs the following errors:
```
$ hack/go-lint.sh $(go list -f '{{ .ImportPath }}' ./...)
./hack/go-lint.sh: 5: golint: not found
```

AFAIK, golint does not exist in docker.io/golang-1.15 or later. So, the script uses the image of origin-release before commit 2c09bc5.

### DIFF
--- a/hack/go-lint.sh
+++ b/hack/go-lint.sh
@@ -2,7 +2,8 @@
 # Example:  ./hack/go-lint.sh installer/... pkg/... tests/smoke
 
 if [ "$IS_CONTAINER" != "" ]; then
-  golint -set_exit_status "${@}"
+  go install golang.org/x/lint/golint@latest
+  ${GOPATH}/bin/golint -set_exit_status "${@}"
 else
   podman run --rm \
     --env IS_CONTAINER=TRUE \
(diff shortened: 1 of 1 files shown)


r4f4:
Another option would be:
```
diff --git a/hack/go-lint.sh b/hack/go-lint.sh
index 346f069dde..d1509cb3d0 100755
--- a/hack/go-lint.sh
+++ b/hack/go-lint.sh
@@ -2,7 +2,8 @@
 # Example:  ./hack/go-lint.sh installer/... pkg/... tests/smoke
 
-  golint -set_exit_status "${@}"
+  go install golang.org/x/lint/golint@latest
+  ${GOPATH}/bin/golint -set_exit_status "${@}"
 else
   podman run --rm \
     --env IS_CONTAINER=TRUE \
```

eramoto:
Thank you for your advice.  
Your option is better because it uses the same image as other `hack` scripts.  

i fixed PR, and the fixed script prints: 
```
$ hack/go-lint.sh $(go list -f '{{ .ImportPath }}' ./...)
go: downloading golang.org/x/lint v0.0.0-20210508222113-6edffad5e616
go: downloading golang.org/x/tools v0.0.0-20200130002326-2f3ba24bd6e7
```

PR #6368, opened 2022-09-16
aws user tags: set limit == openshift/api

patrickdillon:
openshift/api defines a limit for user tags, and this sets our validation to be equal to that. It would be nice if we could vendor the limit, but the limit is set within kubebuilder markers, which are not easily digestable in the installer. Therefore, let's just set a constant value with a reference.

Also cleans up the validation code and adds a test.

/cc @rna-afk @TrilokGeer

### DIFF
--- a/pkg/types/aws/validation/platform.go
+++ b/pkg/types/aws/validation/platform.go
@@ -21,6 +21,10 @@ var kubernetesNamespaceRegex = regexp.MustCompile(`^([^/]*\.)?kubernetes.io/`)
 // openshiftNamespaceRegex is used to check that a tag key is not in the openshift.io namespace.
 var openshiftNamespaceRegex = regexp.MustCompile(`^([^/]*\.)?openshift.io/`)
 
+// userTagLimit is defined in openshift/api
+// https://github.com/openshift/api/blob/1265e99256880f8679d1b74561c0bc7932067c43/config/v1/types_infrastructure.go#L370-L376
+const userTagLimit = 25
+
 // ValidatePlatform checks that the specified platform is valid.
 func ValidatePlatform(p *aws.Platform, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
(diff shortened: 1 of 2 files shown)


pkg/types/aws/validation/platform.go
```diff
@@ -74,13 +78,13 @@ func validateUserTags(tags map[string]string, propagatingTags bool, fldPath *fie
...
-//   OpenShift cannot accept empty tag values.
-// * The key is not in the kubernetes.io namespace.
-// * The key is not in the openshift.io namespace.
+//   - The key and value contain only valid characters.
+//   - The key is not empty and at most 128 characters.
+//   - The value is not empty and at most 256 characters. Note that, while many AWS services accept empty tag values,
+//     the additional tags may be applied to resources in services that do not accept empty tag values. Consequently,
+//     OpenShift cannot accept empty tag values.
+//   - The key is not in the kubernetes.io namespace.
+//   - The key is not in the openshift.io namespace.
```
patrickdillon, line 87: looks like this reformatting was applied by my IDE. looks fine/harmless to me, but will fix it if the tests complain

pkg/types/aws/validation/platform.go
```diff
@@ -21,6 +21,10 @@ var kubernetesNamespaceRegex = regexp.MustCompile(`^([^/]*\.)?kubernetes.io/`)
 // openshiftNamespaceRegex is used to check that a tag key is not in the openshift.io namespace.
 var openshiftNamespaceRegex = regexp.MustCompile(`^([^/]*\.)?openshift.io/`)
 
+// userTagLimit is defined in openshift/api
+// https://github.com/openshift/api/blob/1265e99256880f8679d1b74561c0bc7932067c43/config/v1/types_infrastructure.go#L370-L376
+const userTagLimit = 25
```
rna-afk, line 26: Installer also creates tags so I'm not sure if 25 is the limit and I think it is a little lower than that.

patrickdillon: The limit is 50, with 25 reserved for openshift and 25 for users.

patrickdillon: If https://github.com/openshift/api/pull/1272 merges, we will adjust this number too

TrilokGeer: Sounds good to me, thanks for the updates @patrickdillon

pkg/types/aws/validation/platform.go
sadasu: It would be nice if we did not merge these changes though.

PR #6369, opened 2022-09-16
Document design goals for the Installer

sadasu:
While designing new features it is important to use the Installer's design principles to guide the design and implementation of new feature so that it behaves consistently across the platforms it supports. This is expected to be a living document that we keep updated with our previous and current learnings.

### DIFF
--- a/docs/design/design.md
+++ b/docs/design/design.md
@@ -1,0 +1,30 @@
+#Introduction
+
+The installer gathers all user configuration and consolidates it in the install-config.yaml file. It then uses this configuration to generate a slew of manifests. The Installer is responsible for bootstrapping the cluster. The bootstrap process starts several OpenShift operators while providing configuration to them via these manifests. More information about the Bootstrap process can be found [here](https://github.com/openshift/installer/blob/master/docs/dev/bootstrap_services.md). The role of the Installer comes to an end after this point. These operators in turn use these manifests to generate the resulting cluster. It then becomes the responsibilty of these Operators to manage changes within the cluster on day-2.
+
+While designing new features for the Installer, it is important to adhere to some design principles so that we achieve the following outcomes in no specific order of their importance:
+1. the design of different Installer features are consistent
+2. the Installer is able to provide the best user experince
+3. communicating the Installer's design goals and principles while colloborating with other teams
+
+# Design Documentation
+
+This is an attempt to document the OpenShift Installer's existing design principles and articulate design goals to guide new feature development and in some cases inform bug fixes. This is expected to be a living document which is continually updated when we gain clarity around some of these principles.
+
+## Design Goals
+
+### Install methods
+
+The OpenShift Installer currently supports two installation types, UPI (User Provisioned Infrastructure) and IPI (Installer Provisionined Infrastructure). IPI is an opinionated method of installing OpenShift. The IPI installation method tries to automate a large portion of the OpenShift Install with the least amount of intervention from the user. While it tries to take care of supporting the most common install scenarios, it cannot be expected to support every install variation that a user wants to use for their cloud.
+
+This is where UPI offers greater flexibility required for highly specialized installs. In this install method, the user is expected to provide several Infrastructure components needed for the install on their own. For this reason, documentation surrounding the UPI flow are especially important.
+
+The OpenShift Installer has the responsibility to validate configuration provided to it. There is a validation step that is common to both the UPI and IPI install methods and additional validation is performed specially for the IPI install method.
+
+### Cluster Configuration
+
+A design goal for the Installer is to provide a user-friendly interface for collecting configuration details. One of the ways that the Installer accomplishes this goal is by minimizing the configuration details required to create an OpenShift cluster on the platform and with capabilities requested by the user. The Installer and OpenShift as a whole, have to strive to define sane defaults for configuration whenever possible and allow the user to override these defaults.
+
+Any information that can be learned from the install environment should not be requested as an input from the user. If another OpenShift entity is capable of learning this information, the Installer would defer this responsibility to this entity. This prevents the Installer from becoming a monolith and allows this component to keep its treatment of this environmental config consistent during the time of cluster bringup and day-2.
+
+Another important goal while adding new features to OpenShift is for a cluster upgraded to a version supporting this feature to behave identically to a freshly Installed cluster of that version. When adding configuration for enabling a new functionality within the OpenShift cluster, in addition to designing for a new Installation, consider how a Cluster upgraded to the version supporting this feature would behave. The Operator providing this new functionality should allow updating its API to enable this feature on day-2. Also, it may be helpful to design this feature to be disabled by default and enabled explicitly by changing this API. 


docs/design/design.md
```diff
@@ -0,0 +1,22 @@
+# Design Documentation
+
+This is an attempt to document the OpenShift Installer's existing design principles and also articulate design goals to guide with new feature development and in some cases inform bug fixes.
+
+## Design Goals
+
+### Install methods
+The OpenShift Installer currently supports 2 installation types, UPI (User Provisioned Infrastructure) and IPI (Installer Provisionined Infrastructure). IPI is a highly opinionated method of installing OpenShift. The IPI installation method tries to automate a large portion of the OpenShift Install with the least amount of intervention from the user. While it tries to take care of supporting the most common install scenarios, it cannot be expected to support every install variation that a user wants to use for their cloud.
```
patrickdillon, line 8:
```suggestion
The OpenShift Installer currently supports 2 installation types, UPI (User Provisioned Infrastructure) and IPI (Installer Provisionined Infrastructure). IPI is an opinionated method of installing OpenShift. The IPI installation method tries to automate a large portion of the OpenShift Install with the least amount of intervention from the user. While it tries to take care of supporting the most common install scenarios, it cannot be expected to support every install variation that a user wants to use for their cloud.
```
nit: to me "opinionated" has a technical meaning or at least connotation, i.e. valuing assumptions over configuration. when adding "highly" that seems like a value judgment.

docs/design/design.md
```diff
@@ -0,0 +1,22 @@
...
+
+This is where the UPI offers the greater flexibility required for these highly specialized installs. In this install method, the user is expected to provide and manage several non-OpenShift components needed for the install on their own. For this reason, documentation surrounding the UPI flow are especially important.
```
patrickdillon, line 10: > the user is expected to provide and manage several non-OpenShift components needed for the install on their own

I would nitpick on the words "manage" and "several non-OpenShift components." Manage sounds like an ongoing process, whereas day-2 UPI & IPI should be indistinguishable. 

"Infrastructure" would be a bit more precise, IMHO, than non-OpenShift components.

docs/design/design.md
```diff
@@ -0,0 +1,22 @@
...
+
+The Openshift Installer has the responsibility to validate configuration provided to it. This validation step is common to both the UPI and IPI install methods.
```
patrickdillon, line 12:
```suggestion
The OpenShift Installer has the responsibility to validate configuration provided to it. This validation step is common to both the UPI and IPI install methods.
```

I would suggest: 
> The Installer validates user-provided configuration in order to prevent failures during installation.

We do have special validation which is run only in the case of IPI.

docs/design/design.md
```diff
@@ -0,0 +1,22 @@
...
+
+### Cluster Configuration
+
+It is a design goal for the Installer to obtain the simplest and most minimal configuration required to fully create an OpenShift cluster on the platform and with the capabilities requested by the user. So, for that reason, any information that can be learned from the install environment should not be requested as an input configuration during the install.
```
patrickdillon, line 16: I like that we are stating a design goal, but I have two points here: 
1. I think it would be helpful to establish the motivation and guiding principles for the goal; which in the case of simplicity  and especially for discovering configuration, the principle would be something along the lines of single-responsibility; if we can defer the responsibility to another entity, we opt for that rather than having the installer become a monolith
2. I don't think "simple and minimal" is a good goal because it doesn't provide inherent value; "user-friendly" may be a suitable alternative

docs/design/design.md
```diff
@@ -0,0 +1,22 @@
...
+
+The installer gathers all user configuration and consolidates it the install-config.yaml file and uses this configuration to generated a slew of manifests. The Installer bootstraps the cluster by starting several OpenShift operators and providing configuration to them via these manifests. These operators in turn use these manifests to generate the resulting cluster. It then becomes the responsibilty of these Operators to manage this configuation data and the resulting changes within the cluster on day-2. Installer does not have a role to play in managing this configuration on day-2.
```
patrickdillon, line 18: > The Installer bootstraps the cluster by starting several OpenShift operators and providing configuration to them via these manifests.

The bootstrap process is already described in the design overview doc, and I think the same point, which is that the installer lays down configuration for cluster operators, without going into this detail, which could be a little confusing.

patrickdillon (commented):
This is cool. I have left some feedback. I hope it helps.

docs/design/design.md
sadasu: Tried rewording this.

docs/design/design.md
```diff
@@ -0,0 +1,32 @@
...
+
+Another guiding principle to consider regarding configuration during install, would be that any information that can be learned from the install environment should not be requested as an input from the user. And, if another OpenShift entity is capable of learning this information, the Installer would defer this responsibility to this entity. This prevents the Installer from becoming a monolith and allows this component to keep its treatment of this environmental config consistent during the time of cluster bringup and day-2.
+
+When adding any new configuration for enabling a new functionality within the OpenShift cluster, in addition to designing for a new Installation, consider how a Cluster upgraded to the version supporting this feature would behave. The design goal here would be for a cluster upgraded to a version supporting this feature behave identically to a fresh Installed cluster of that version.
```
barbacbd, line 30: Can we figure out a way to flip these sentences ?

docs/design/design.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Cluster Configuration
+
+It is a design goal for the Installer to obtain configuration from the user in the most user-friendly way. One of the ways to accomplish that goal would be to request a minimal amount of configuration to create an OpenShift cluster on the platform and with the capabilities requested by the user. The Installer and OpenShift as a whole, have to strive to define sane defaults for configuration whenever possible and allow the user to override these defaults.
+
+Another guiding principle to consider regarding configuration during install, would be that any information that can be learned from the install environment should not be requested as an input from the user. And, if another OpenShift entity is capable of learning this information, the Installer would defer this responsibility to this entity. This prevents the Installer from becoming a monolith and allows this component to keep its treatment of this environmental config consistent during the time of cluster bringup and day-2.
```
barbacbd, line 28: drop: "Another guiding principle to consider regarding configuration during install, would be that "

Remove the word "And" from the start of the second sentence

docs/design/design.md
```diff
@@ -0,0 +1,32 @@
...
+
+When adding any new configuration for enabling a new functionality within the OpenShift cluster, in addition to designing for a new Installation, consider how a Cluster upgraded to the version supporting this feature would behave. The design goal here would be for a cluster upgraded to a version supporting this feature behave identically to a fresh Installed cluster of that version.
+
+This above goal can be achieved by making sure that the Operator providing this new functionality allows updating its API enabling this feature on day-2. Also, it may be helpful to design this feature to be disabled by default and enabled explicitly by changing the API.
```
barbacbd, line 32: combine this with the information about the above goal if it goes with it

docs/design/design.md
```diff
@@ -0,0 +1,32 @@
...
+2. the Installer is able to provide the best user experince
+3. communicating the Installer's design goals and principles while colloborating with other teams
+
+# Design Documentation
+
+This is an attempt to document the OpenShift Installer's existing design principles and also articulate design goals to guide with new feature development and in some cases inform bug fixes. This is expected to be a living document which is continually updated when we gain clarity around some of these principles.
+
+## Design Goals
+
+### Install methods
```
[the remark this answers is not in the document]
sadasu: This is not meant to be user facing documentation only a quick  introduction to the 2 install methods within the context of how that would affect design decisions. Here is the corresponding user facing documentation: https://docs.openshift.com/container-platform/4.7/installing/index.html.

docs/design/design.md
```diff
@@ -0,0 +1,32 @@
...
+
+The OpenShift Installer currently supports 2 installation types, UPI (User Provisioned Infrastructure) and IPI (Installer Provisionined Infrastructure). IPI is an opinionated method of installing OpenShift. The IPI installation method tries to automate a large portion of the OpenShift Install with the least amount of intervention from the user. While it tries to take care of supporting the most common install scenarios, it cannot be expected to support every install variation that a user wants to use for their cloud.
+
+This is where the UPI offers the greater flexibility required for these highly specialized installs. In this install method, the user is expected to provide several Infrastructure components needed for the install on their own. For this reason, documentation surrounding the UPI flow are especially important.
```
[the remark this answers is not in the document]
sadasu: I don't know how I ended up with so many "the"s there!

docs/design/design.md
patrickdillon: Agreed, "user-friendly" is not great. I think the current wording:

> most minimal configuration required to fully create an OpenShift cluster on the platform and with the capabilities requested by the user 

is pretty good. I _think_ the objective for this document is so that someone can use it as a guide when answering the question: "should this knob go in the install config?" If we say the guiding principal is to keep the install config _minimal_, then the answer would always be no... in fact we should be removing fields!

The current wording is better in that it provides a little more wiggle room, but I do think we can do better, although I think it will be hard... (and also, again, assuming I am on the right track for what we are trying to accomplish).  

Another issue I have with minimal is that I actually don't think that should be a long-term goal. I think the idea that config should be "minimal" has been a defensive maneuver. to allow us to limit scope/focus; we have said we will limit our focus to a subset of features, power users can still go in and edit manifests, but there are no validations/guard rails in that case. But if we had the ability, I don't see why we wouldn't allow more appropriate config, assuming that it belong in the install config in the first place and we were able to validate the inputs. I think the area this becomes most apparent is for the machineset manifests: the machine API provides a set of features for configuring machines, and we mirror a subset of those in the install config. For example, in [AWS security groups](https://github.com/openshift/api/blob/master/machine/v1beta1/types_awsprovider.go#L60) are configurable. If we were able to safely allow that configuration, why wouldn't we include it in the install config? In fact, I think a fair amount of our work for the foreseeable future will be to expand our config in exactly this way. Ideally, IMHO, we would find a way to unify the efforts between the machine api operator and installer so that API definitions and validations are written once and then vendored in, but I get the sense that is a tall order.

Here are some starter criteria for things I think should belong in the install config, hopefully we will be able to think of more:
1. User provided - if the config can be inferred, it should be; also, general cluster config is generated by the [cluster config operator](https://github.com/openshift/cluster-config-operator)  on the bootstrap node, this has come up in PRs like https://github.com/openshift/installer/pull/5676
2. Input should be verifiable - we should be able to run validation to ensure that config, especially as a whole, will result in a valid cluster

docs/design/design.md
```diff
@@ -0,0 +1,32 @@
...
+
+The OpenShift Installer has the responsibility to validate configuration provided to it. There is a validation step that is common to both the UPI and IPI install methods and additional validation is performed specially for the IPI install method.
+
+### Cluster Configuration
+
+It is a design goal for the Installer to obtain configuration from the user in the most user-friendly way. One of the ways to accomplish that goal would be to request a minimal amount of configuration to create an OpenShift cluster on the platform and with the capabilities requested by the user. The Installer and OpenShift as a whole, have to strive to define sane defaults for configuration whenever possible and allow the user to override these defaults.
```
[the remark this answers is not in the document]
sadasu: Nice suggestion.

docs/design/design.md
sadasu: Done.

docs/design/design.md
sadasu: Attempted.

docs/design/design.md
sadasu: Done.

docs/design/design.md
```diff
@@ -0,0 +1,30 @@
...
+
+The OpenShift Installer has the responsibility to validate configuration provided to it. There is a validation step that is common to both the UPI and IPI install methods and additional validation is performed specially for the IPI install method.
+
+### Cluster Configuration
+
+A design goal for the Installer is to provide a user-friendly interface for collecting configuration details. One of the ways that the Installer accomplishes this goal is by minimizing the configuration details required to create an OpenShift cluster on the platform and with capabilities requested by the user. The Installer and OpenShift as a whole, have to strive to define sane defaults for configuration whenever possible and allow the user to override these defaults.
+
+Any information that can be learned from the install environment should not be requested as an input from the user. If another OpenShift entity is capable of learning this information, the Installer would defer this responsibility to this entity. This prevents the Installer from becoming a monolith and allows this component to keep its treatment of this environmental config consistent during the time of cluster bringup and day-2.
```
dhellmann, line 28: It would be good to cover the division of responsibilities between the installer, bootstrap, and operators. I don't know if it makes more sense to extend this paragraph or to talk about that separately.

docs/design/design.md
```diff
@@ -0,0 +1,30 @@
...
+
+This is an attempt to document the OpenShift Installer's existing design principles and articulate design goals to guide new feature development and in some cases inform bug fixes. This is expected to be a living document which is continually updated when we gain clarity around some of these principles.
+
+## Design Goals
+
+### Install methods
+
+The OpenShift Installer currently supports two installation types, UPI (User Provisioned Infrastructure) and IPI (Installer Provisionined Infrastructure). IPI is an opinionated method of installing OpenShift. The IPI installation method tries to automate a large portion of the OpenShift Install with the least amount of intervention from the user. While it tries to take care of supporting the most common install scenarios, it cannot be expected to support every install variation that a user wants to use for their cloud.
+
+This is where UPI offers greater flexibility required for highly specialized installs. In this install method, the user is expected to provide several Infrastructure components needed for the install on their own. For this reason, documentation surrounding the UPI flow are especially important.
```
dhellmann, line 20: Letting the user edit the manifests created by the installer is another way flexibility is managed and the install-config API is kept small. For example, when we added the workload partitioning feature, we did not add any flags to support it to install-config because it was only used by a small number of customers, at first. 

What heuristics does the team use to decide when to add a configuration option and when to rely on having the user edit manifests?

PR #6372, opened 2022-09-19
CORS-2170: Add backoff for GCP cluster destroy.

rna-afk:
There has been some rate limiting issues with the GCP destroy cluster step in CI so adding an exponential backoff to space the destroy requests and avoid the problem.

More details: https://issues.redhat.com/browse/CORS-2170

### DIFF
--- a/pkg/destroy/gcp/gcp.go
+++ b/pkg/destroy/gcp/gcp.go
@@ -127,8 +127,12 @@ func (o *ClusterUninstaller) Run() (*types.ClusterQuota, error) {
 		return nil, errors.Wrap(err, "failed to create resourcemanager service")
 	}
 
-	err = wait.PollImmediateInfinite(
-		time.Second*10,
+	err = wait.ExponentialBackoff(
+		wait.Backoff{
+			Duration: time.Second * 10,
+			Factor:   0.05,
+			Cap:      time.Minute * 5,
+		},
 		o.destroyCluster,
 	)
 	if err != nil {


patrickdillon:
This does not seem to be right. From the CI run:
```
time="2022-10-03T19:47:15Z" level=debug msg="OpenShift Installer unreleased-master-7060-g1bbaabc3f32009a84e3aa2b94a193a8e5e3f61ed-dirty"
time="2022-10-03T19:47:15Z" level=debug msg="Built from commit 1bbaabc3f32009a84e3aa2b94a193a8e5e3f61ed"
time="2022-10-03T19:47:15Z" level=info msg="Credentials loaded from environment variable \"GOOGLE_CLOUD_KEYFILE_JSON\", file \"/var/run/secrets/ci.openshift.io/cluster-profile/gce.json\""
time="2022-10-03T19:47:25Z" level=fatal msg="Failed to destroy cluster: failed to destroy cluster: timed out waiting for the condition"
```

rna-afk:
Backoff is way too small right now and it still times out. Is it even worth a backoff if we go below this?

patrickdillon:
> Backoff is way too small right now and it still times out. Is it even worth a backoff if we go below this?

It seems like a couple of things may be going wrong. First, we don't want to limit the amount of time the destroyCluster function takes: we just want to increase the amount of time in between runs. Second, the way you are using Exponential Backoff it seems to be decreasing the amount of time for each run, not increasing. 

Also another idea for improving rate limiting would be to not retry these functions if they have already destroyed all of their resources. I believe we are continuing to peg the list functions even when we have listed and removed all the resources.

patrickdillon:
/close
We're going to defer this work.

PR #6373, opened 2022-09-19
Agent: Don't pass AdminKubeConfigClientCertKey

zaneb:
This depends on the AdminKubeConfigSignerCertKey, which we also pass, with the result that AdminKubeConfigClientCertKey always gets regenerated by the installer when assisted-service runs it. Since this has no effect other than generating a spurious warning, just stop passing it.


zaneb:
/hold
We can't test this properly until the test job is fixed.

PR #6374, opened 2022-09-19
Bug 2104997: Add additional fields to proxy manifest

barbacbd:
** Add 168.63.129.16 IP to noProxy in Azure Stack Hub 
** Add ArmEndpoint (from install config) when performing azure installation

https://issues.redhat.com/browse/OCPBUGSM-46450

### DIFF
--- a/pkg/asset/manifests/proxy.go
+++ b/pkg/asset/manifests/proxy.go
@@ -158,6 +158,11 @@ func createNoProxy(installConfig *installconfig.InstallConfig, network *Networki
 
 	// TODO: IBM[#95]: proxy
 
+	if platform == azure.Name && installConfig.Azure.CloudName == azure.StackCloud {
+		set.Insert("168.63.129.16")
+		set.Insert(installConfig.Config.Azure.ARMEndpoint)
+	}
+
 	// From https://cloud.google.com/vpc/docs/special-configurations add GCP metadata.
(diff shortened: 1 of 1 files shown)


pkg/asset/manifests/proxy.go
```diff
@@ -158,6 +158,13 @@ func createNoProxy(installConfig *installconfig.InstallConfig, network *Networki
 
 	// TODO: IBM[#95]: proxy
 
+	if platform == azure.Name {
+		if installConfig.Azure.CloudName == azure.StackCloud {
+			set.Insert("168.63.129.16")
+		}
+		set.Insert(installConfig.Config.Azure.ARMEndpoint)
```
rna-afk, line 165: Isn't this only for Azure Stack? Right now it's adding this for Azure too

barbacbd: The ArmEndpoint can be for either

rna-afk: Code comments say otherwise. Mind changing them then?
https://github.com/openshift/installer/blob/c623779b6c2bbe300723475ad509757283844a5b/pkg/types/azure/platform.go#L31

rna-afk: Here's another one.
https://github.com/openshift/installer/blob/c623779b6c2bbe300723475ad509757283844a5b/pkg/asset/installconfig/azure/metadata.go#L22

PR #6375, opened 2022-09-19
CORS-2280: IBMCloud: Add DNS Record for internal public traffic

cjschaef:
Add a DNS Services DNS Record for Private clusters to allow traffic to the public server URL to be directed to the private LB.


jeffnowicki:
/retitle CORS-2280: IBMCloud: Add DNS Record for internal public traffic

PR #6379, opened 2022-09-20
OCPBUGS-891: aws: add explicit EIP dependency for nat gw

r4f4:
Hopefully by making the dependency between them explicit, we might alleviate the following error in CI

Error: error creating EC2 NAT Gateway: InvalidElasticIpID.NotFound: The elasticIp ID 'eipalloc-094ec9d0482d5b9f2' does not exist

### DIFF
--- a/data/data/aws/cluster/vpc/vpc-public.tf
+++ b/data/data/aws/cluster/vpc/vpc-public.tf
@@ -96,5 +96,5 @@ resource "aws_nat_gateway" "nat_gw" {
   )
 
   # https://issues.redhat.com/browse/OCPBUGS-891
-  depends_on = [aws_eip.nat_eip]
+  depends_on = [aws_eip.nat_eip, aws_subnet.public_subnet]
 }
--- a/data/data/aws/cluster/vpc/vpc-public.tf
+++ b/data/data/aws/cluster/vpc/vpc-public.tf
@@ -94,4 +94,7 @@ resource "aws_nat_gateway" "nat_gw" {
     },
     var.tags,
   )
+
+  # https://issues.redhat.com/browse/OCPBUGS-891
+  depends_on = [aws_eip.nat_eip]
 }


sadasu:
/approve
Good idea

data/data/aws/cluster/vpc/vpc-public.tf
```diff
@@ -94,4 +94,7 @@ resource "aws_nat_gateway" "nat_gw" {
     },
     var.tags,
   )
+
+  # https://issues.redhat.com/browse/OCPBUGS-891
+  depends_on = [aws_eip.nat_eip]
```
rna-afk, line 99: What if we add `aws_subnet.public_subnet` too to this to prevent future errors?

barbacbd: I agree with Aditya, we _could_ add that, but I am worried it may not work. Azure has/had something very similar which I initially added a depends_on, but ultimately it failed. I linked a PR below to what was completed for Azure. Not promising this works or is what we need but it is something to consider.

https://github.com/openshift/installer/pull/6385

rna-afk: That PR didn't fix it (basing it off of this comment https://coreos.slack.com/archives/C042E82T490/p1663855691063339) so might need to think of something else

patrickdillon:
/lgtm
We should open an upstream issue in https://github.com/hashicorp/terraform-provider-aws/issues for the eip error

r4f4:
/override ci/prow/okd-e2e-gcp-ovn-upgrade
it's timing out after 4h and it's unrelated to this PR

r4f4:
> /lgtm We should open an upstream issue in https://github.com/hashicorp/terraform-provider-aws/issues for the eip error

Done: https://github.com/hashicorp/terraform-provider-aws/issues/27087

PR #6384, opened 2022-09-21
Agent: Increase bootstrap-complete timeout

zaneb:
It appears that installs of 4.12 are taking consistently longer to bootstrap than for 4.11 - a local test run recorded 28 minutes, vs. 17-20 minutes in previous 4.11 testing. In CI testing, we are exceeding the 30 minute timeout.

Until such time as we are able to determine and address the reason for the slower bootstrapping, increase the timeout.


rwsu (commented):
/lgtm

PR #6386, opened 2022-09-21
AGENT-309 Display the rendezvous IP in motd and issue

rwsu:
motd and issue indicate whether the host is the rendezvous host.

The rendezvous host's IP address is displayed.


bfournie:
/lgtm

Tested:
"This host 192.168.111.80 is the rendezvous host."

and 
"This host is not the rendezvous host. The rendezvous host is at 192.168.111.80."

PR #6387, opened 2022-09-21
CORS-2073: Remove terraform provider "random"

rna-afk:
Currently, we import terraform "random" provider to generate a
random suffix for the azure storage account and storage blob
which is a needless provider import since it can be done with
a simple go code and a variable in terraform.

Adding logic to create a random string and removing the provider.


jhixson74:
Looks like this isn't quite complete. All the tests are failing since the provider is missing.

jhixson74:
Looks like you forgot to remove it in pkg/terraform/providers/providers.go

jhixson74:
There seems to be a problem with AzureStack. Is it related at all?

rna-afk:
It doesn't seem related to this PR. This error is happening on all PRs.

PR #6390, opened 2022-09-21
OCPBUGS-1490: [vsphere-zones] Handle no set zones

jcpowermac:
Fix the issue when there are no zones
set in the machinepool for control plane
and workers.

This should also handle if there are
more replicas required than FailureDomains.


jcpowermac:
/hold

Needs some work for compute.

jinyunma:
verified on pre-merged payload and passed, details can be found in bug comments.
/qe-approved

jinyunma:
verified on pre-merged payload and passed, details can be found in bug comments.
/label qe-approved

r4f4:
/override ci/prow/e2e-libvirt 
it's permafailing when gathering test results.

PR #6395, opened 2022-09-22
Agent based installer - assert that user provided release image architecture is supported

omertuc:
Right now the agent based installer hardcodes the x86 architecture and
this causes issues much further in the installation process that are
harder to track down.

I've added a couple of assertions to make the issue clearer for the
user.

This is obviously not robust as it's just parsing the pull-spec looking
for mentions of the arm architecture, they should be removed once
the agent based installer adds proper support for other architectures
and does away with hardcoding x86.


omertuc:
Force pushed for formatting purposes

lranjbar:
AGENT-397

zaneb:
Superseded by #6546
/close

PR #6398, opened 2022-09-22
vsphere: prevent duplicate validation check in multi-zone

bostrt:
`ValidateMultiZoneForProvisioning` is calling `ValidateForProvisioning` [here](https://github.com/openshift/installer/blob/2d0389c782601cd76e9b6454aee657c1c274a26b/pkg/asset/installconfig/vsphere/validation.go#L66-L72) already so it can be skipped in `platformprovisioncheck.go`


r4f4 (commented):
/lgtm

PR #6399, opened 2022-09-22
gcp: Fix text cases for GCP quota check.

rna-afk:
There is a customer case where the customer could not pass the GCP quota check with only the required services enabled and looked into the test cases. This one in particular seems to just check if the test fails or not and was not checking the actual error message, resulting in the test case with a wrong error message expected to pass.

Changing the error expected field from string to bool to just check if there is an error when the required services are not provided.

Support case: https://access.redhat.com/support/cases/#/case/03200212

### DIFF
--- a/pkg/asset/installconfig/gcp/validation_test.go
+++ b/pkg/asset/installconfig/gcp/validation_test.go
@@ -1,6 +1,7 @@
 package gcp
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"testing"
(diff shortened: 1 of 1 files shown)


rna-afk:
The reason I did not check for the exact error message is because the printed message prints a list and the order of the list is not always the same. I didn't want to run a sort function just to print the error message in the right format to test it later as I feel it's a time consuming sort just for an error message

If anyone feels it should be sorted, let me know and I'll do it.

sadasu:
> The reason I did not check for the exact error message is because the printed message prints a list and the order of the list is not always the same. I didn't want to run a sort function just to print the error message in the right format to test it later as I feel it's a time consuming sort just for an error message
> 
> If anyone feels it should be sorted, let me know and I'll do it.

Ok with the current fix. Could we have used err.Contains to check the type of error received back and continued to use error as string instead of the bool that it is now?

rna-afk:
> Ok with the current fix. Could we have used err.Contains to check the type of error received back and continued to use error as string instead of the bool that it is now?

The permissions do not show up in order and are random. I didn't want to sort the permissions just for a test.

patrickdillon:
/close
followed up by #6504

PR #6401, opened 2022-09-22
ARMOCP-417: enable arm64 for agent installer

jeffdyoung:
Enable arm64 iso builds for agent installer based on the control plane architecture set in install-config.yaml.

### DIFF
--- a/cmd/openshift-install/testdata/agent/image/manifests/default_manifests.txt
+++ b/cmd/openshift-install/testdata/agent/image/manifests/default_manifests.txt
@@ -111,6 +111,7 @@ spec:
   clusterRef:
     name: ostest
     namespace: cluster0
+  cpuArchitecture: x86_64
   ipxeScriptType: ""
   nmStateConfigLabelSelector:
     matchLabels:
--- a/pkg/asset/agent/image/baseiso.go
+++ b/pkg/asset/agent/image/baseiso.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"time"
 
+	"github.com/coreos/stream-metadata-go/arch"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
@@ -16,17 +17,13 @@ import (
 	"github.com/openshift/installer/pkg/asset/agent/manifests"
 	"github.com/openshift/installer/pkg/asset/agent/mirror"
 	"github.com/openshift/installer/pkg/rhcos"
+	"github.com/openshift/installer/pkg/types"
 )
 
 // BaseIso generates the base ISO file for the image
 type BaseIso struct {
 	File *asset.File
 }
-
-const (
-	// TODO - add support for other architectures
-	archName = "x86_64"
-)
 
 var (
 	baseIsoFilename = ""
@@ -40,7 +37,7 @@ func (i *BaseIso) Name() string {
 }
 
 // getIsoFile is a pluggable function that gets the base ISO file
-type getIsoFile func() (string, error)
+type getIsoFile func(archName string) (string, error)
 
 type getIso struct {
 	getter getIsoFile
@@ -54,8 +51,7 @@ func newGetIso(getter getIsoFile) *getIso {
 var GetIsoPluggable = downloadIso
 
 // Download the ISO using the URL in rhcos.json
-func downloadIso() (string, error) {
-
+func downloadIso(archName string) (string, error) {
 	ctx, cancel := context.WithTimeout(context.TODO(), 30*time.Second)
 	defer cancel()
 
@@ -64,9 +60,6 @@ func downloadIso() (string, error) {
 	if err != nil {
 		return "", err
 	}
-
-	// Defaults to using the x86_64 baremetal ISO for all platforms
-	// archName := arch.RpmArch(string(config.ControlPlane.Architecture))
 	streamArch, err := st.GetArchitecture(archName)
 	if err != nil {
 		return "", err
@@ -107,10 +100,17 @@ func (i *BaseIso) Generate(dependencies asset.Parents) error {
 	// use the GetIso function to get the BaseIso from the release payload
 	agentManifests := &manifests.AgentManifests{}
 	dependencies.Get(agentManifests)
-
 	var baseIsoFileName string
 	var err error
+
+	// Default iso archName to x86_64.
+	archName := arch.RpmArch(types.ArchitectureAMD64)
+
 	if agentManifests.ClusterImageSet != nil {
+		// If specified, use InfraEnv.Spec.CpuArchitecture for iso archName
+		if agentManifests.InfraEnv.Spec.CpuArchitecture != "" {
+			archName = agentManifests.InfraEnv.Spec.CpuArchitecture
+		}
 		releaseImage := agentManifests.ClusterImageSet.Spec.ReleaseImage
 		pullSecret := agentManifests.GetPullSecretData()
 		registriesConf := &mirror.RegistriesConf{}
@@ -135,7 +135,7 @@ func (i *BaseIso) Generate(dependencies asset.Parents) error {
 
 	logrus.Info("Downloading base ISO")
 	isoGetter := newGetIso(GetIsoPluggable)
-	baseIsoFileName, err2 := isoGetter.getter()
+	baseIsoFileName, err2 := isoGetter.getter(archName)
 	if err2 == nil {
 		logrus.Debugf("Using base ISO image %s", baseIsoFileName)
 		i.File = &asset.File{Filename: baseIsoFileName}
--- a/pkg/asset/agent/image/ignition.go
+++ b/pkg/asset/agent/image/ignition.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/coreos/ignition/v2/config/util"
 	igntypes "github.com/coreos/ignition/v2/config/v3_2/types"
+	"github.com/coreos/stream-metadata-go/arch"
 	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -29,6 +30,7 @@ import (
 	"github.com/openshift/installer/pkg/asset/password"
 	"github.com/openshift/installer/pkg/asset/tls"
 	"github.com/openshift/installer/pkg/rhcos"
+	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/agent"
 	"github.com/openshift/installer/pkg/version"
 )
@@ -126,8 +128,12 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
 	logrus.Infof("The rendezvous host IP (node0 IP) is %s", nodeZeroIP)
 
 	a.RendezvousIP = nodeZeroIP
+	// Default to x86_64
+	archName := arch.RpmArch(types.ArchitectureAMD64)
+	if infraEnv.Spec.CpuArchitecture != "" {
+		archName = infraEnv.Spec.CpuArchitecture
+	}
 
-	// TODO: don't hard-code target arch
 	releaseImageList, err := releaseImageList(agentManifests.ClusterImageSet.Spec.ReleaseImage, archName)
 	if err != nil {
 		return err
--- a/pkg/asset/agent/installconfig.go
+++ b/pkg/asset/agent/installconfig.go
@@ -96,8 +96,9 @@ func (a *OptionalInstallConfig) validateSupportedArchs(installConfig *types.Inst
 
 	switch string(installConfig.ControlPlane.Architecture) {
 	case types.ArchitectureAMD64:
+	case types.ArchitectureARM64:
 	default:
-		allErrs = append(allErrs, field.NotSupported(fieldPath, installConfig.ControlPlane.Architecture, []string{types.ArchitectureAMD64}))
+		allErrs = append(allErrs, field.NotSupported(fieldPath, installConfig.ControlPlane.Architecture, []string{types.ArchitectureAMD64, types.ArchitectureARM64}))
 	}
 
 	for i, compute := range installConfig.Compute {
@@ -105,8 +106,9 @@ func (a *OptionalInstallConfig) validateSupportedArchs(installConfig *types.Inst
 
 		switch string(compute.Architecture) {
 		case types.ArchitectureAMD64:
+		case types.ArchitectureARM64:
 		default:
-			allErrs = append(allErrs, field.NotSupported(fieldPath, compute.Architecture, []string{types.ArchitectureAMD64}))
+			allErrs = append(allErrs, field.NotSupported(fieldPath, compute.Architecture, []string{types.ArchitectureAMD64, types.ArchitectureARM64}))
 		}
 	}
 
--- a/pkg/asset/agent/installconfig_test.go
+++ b/pkg/asset/agent/installconfig_test.go
@@ -221,13 +221,13 @@ baseDomain: test-domain
 networking:
   networkType: OVNKubernetes
 compute:
-  - architecture: arm64
+  - architecture: s390x
     hyperthreading: Enabled
     name: worker
     platform: {}
     replicas: 0
 controlPlane:
-  architecture: arm64
+  architecture: s390x
   hyperthreading: Enabled
   name: master
   platform: {}
@@ -237,7 +237,7 @@ platform:
 pullSecret: "{\"auths\":{\"example.com\":{\"auth\":\"authorization value\"}}}"
 `,
 			expectedFound: false,
-			expectedError: "invalid install-config configuration: [ControlPlane.Architecture: Unsupported value: \"arm64\": supported values: \"amd64\", Compute[0].Architecture: Unsupported value: \"arm64\": supported values: \"amd64\"]",
+			expectedError: "invalid install-config configuration: [ControlPlane.Architecture: Unsupported value: \"s390x\": supported values: \"amd64\", \"arm64\", Compute[0].Architecture: Unsupported value: \"s390x\": supported values: \"amd64\", \"arm64\"]",
 		},
 		{
 			name: "valid configuration for none platform for sno",
--- a/pkg/asset/agent/manifests/infraenv.go
+++ b/pkg/asset/agent/manifests/infraenv.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/coreos/stream-metadata-go/arch"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -15,6 +16,7 @@ import (
 	"github.com/openshift/installer/pkg/asset"
 	"github.com/openshift/installer/pkg/asset/agent"
 	"github.com/openshift/installer/pkg/asset/agent/agentconfig"
+	"github.com/openshift/installer/pkg/types"
 )
 
 var (
@@ -70,6 +72,12 @@ func (i *InfraEnv) Generate(dependencies asset.Parents) error {
 				},
 			},
 		}
+
+		// Use installConfig.Config.ControlPlane.Architecture to determine cpuarchitecture for infraEnv.Spec.CpuArchiteture.
+		// installConfig.Config.ControlPlance.Architecture uses go/Debian cpuarchitecture values (amd64, arm64) so we must convert to rpmArch because infraEnv.Spec.CpuArchitecture expects x86_64 or aarch64.
+		if installConfig.Config.ControlPlane.Architecture != "" {
+			infraEnv.Spec.CpuArchitecture = arch.RpmArch(string(installConfig.Config.ControlPlane.Architecture))
+		}
 		if installConfig.Config.Proxy != nil {
 			infraEnv.Spec.Proxy = getProxy(installConfig)
 		}
@@ -116,7 +124,10 @@ func (i *InfraEnv) Load(f asset.FileFetcher) (bool, error) {
 	if err := yaml.UnmarshalStrict(file.Data, config); err != nil {
 		return false, errors.Wrapf(err, "failed to unmarshal %s", infraEnvFilename)
 	}
-
+	// If defined, convert to RpmArch amd64 -> x86_64 or arm64 -> aarch64
+	if config.Spec.CpuArchitecture != "" {
+		config.Spec.CpuArchitecture = arch.RpmArch(config.Spec.CpuArchitecture)
+	}
 	i.File, i.Config = file, config
 	if err = i.finish(); err != nil {
 		return false, err
@@ -131,5 +142,11 @@ func (i *InfraEnv) finish() error {
 		return errors.New("missing configuration or manifest file")
 	}
 
+	// Throw an error if CpuArchitecture isn't x86_64, aarch64, or ""
+	switch i.Config.Spec.CpuArchitecture {
+	case arch.RpmArch(types.ArchitectureAMD64), arch.RpmArch(types.ArchitectureARM64), "":
+	default:
+		return errors.Errorf("Config.Spec.CpuArchitecture %s is not supported ", i.Config.Spec.CpuArchitecture)
+	}
 	return nil
 }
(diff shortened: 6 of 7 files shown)


pkg/asset/agent/manifests/common.go
```diff
@@ -22,7 +22,9 @@ func getInfraEnvName(ic *agent.OptionalInstallConfig) string {
 func getPullSecretName(ic *agent.OptionalInstallConfig) string {
 	return ic.Config.ObjectMeta.Name + "-pull-secret"
 }
-
+func GetArchitecture(ic *agent.OptionalInstallConfig) string {
```
cs-zhang, line 25: For consistent, should use getArchitecture().

pkg/asset/agent/manifests/infraenv.go
```diff
@@ -54,6 +54,7 @@ func (i *InfraEnv) Generate(dependencies asset.Parents) error {
 				Namespace: getObjectMetaNamespace(installConfig),
 			},
 			Spec: aiv1beta1.InfraEnvSpec{
+				CpuArchitecture: GetArchitecture(installConfig),
```
cs-zhang: Need to use getArchitecture().

pkg/asset/agent/image/baseiso.go
```diff
@@ -117,7 +111,10 @@ func (i *BaseIso) Generate(dependencies asset.Parents) error {
 
 	var baseIsoFileName string
 	var err error
+	//agentManifests.ClusterImageSet = nil
 	if agentManifests.ClusterImageSet != nil {
+		rhcosArch := agentManifests.InfraEnv.Spec.CpuArchitecture
```
cs-zhang, line 116: The `rhcosArch := ` has to be change to `rhcosArch =` for global var assignment.

cs-zhang (changes requested):

pkg/asset/agent/image/baseiso.go
```diff
@@ -117,7 +111,10 @@ func (i *BaseIso) Generate(dependencies asset.Parents) error {
 
 	var baseIsoFileName string
 	var err error
+	//agentManifests.ClusterImageSet = nil
 	if agentManifests.ClusterImageSet != nil {
+		rhcosArch = agentManifests.InfraEnv.Spec.CpuArchitecture
```
cs-zhang, line 116: Need to be checked if it is defined, if not, default `x86_64` will be used.

pkg/asset/agent/image/baseiso.go
```diff
@@ -117,7 +111,12 @@ func (i *BaseIso) Generate(dependencies asset.Parents) error {
 
 	var baseIsoFileName string
 	var err error
+	//agentManifests.ClusterImageSet = nil
 	if agentManifests.ClusterImageSet != nil {
+		// Override rhcosArch = x86_64 if InfraEnv has Arch defined
+		if agentManifests.InfraEnv.Spec.CpuArchitecture != "" {
+			rhcosArch = agentManifests.InfraEnv.Spec.CpuArchitecture
```
zaneb, line 118: Let's define a local variable that defaults to "x86_64". I'm not comfortable with changing a global here, especially when the name is shadowed by local variables in lots of functions.

pkg/asset/agent/image/ignition.go
```diff
@@ -110,8 +110,14 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
...
 	logrus.Infof("The rendezvous host IP (node0 IP) is %s", nodeZeroIP)
 
-	// TODO: don't hard-code target arch
-	releaseImageList, err := releaseImageList(agentManifests.ClusterImageSet.Spec.ReleaseImage, "x86_64")
+	rhcosArch := infraEnv.Spec.CpuArchitecture
+	switch rhcosArch {
+	case "arm64":
+		rhcosArch = "aarch64"
+	case "amd64", "":
+		rhcosArch = "x86_64"
```
zaneb, line 118: can we do ppc64le at the same time?

pkg/asset/agent/image/baseiso.go
```diff
@@ -23,11 +23,9 @@ type BaseIso struct {
 	File *asset.File
 }
 
-const (
-	// TODO - add support for other architectures
-	archName = "x86_64"
+var (
+	rhcosArch = "x86_64"
```
zaneb, line 27: This repo is also used to build OKD (which uses FCOS, not RHCOS), so something like `isoArch` would be more appropriate.

pkg/asset/agent/image/oc.go
```diff
@@ -76,16 +76,21 @@ func (r *release) GetBaseIso(log logrus.FieldLogger, releaseImage, pullSecret, a
...
 		return "", err
 	}
-
-	filename := fmt.Sprintf(coreOsFileName, architecture)
+	switch rhcosArch {
+	case "arm64":
+		rhcosArch = "aarch64"
+	case "amd64":
+		rhcosArch = "x86_64"
+	}
```
zaneb, line 84: This code exists in multiple places now, so we should define a function to get the architecture name used by CoreOS given the name in the install-config.

pkg/asset/agent/image/oc.go
```diff
@@ -64,7 +64,7 @@ const (
 )
 
 // Get the CoreOS ISO from the releaseImage
-func (r *release) GetBaseIso(log logrus.FieldLogger, releaseImage, pullSecret, architecture string, mirrorConfig []mirror.RegistriesConfig) (string, error) {
+func (r *release) GetBaseIso(log logrus.FieldLogger, releaseImage, pullSecret, rhcosArch string, mirrorConfig []mirror.RegistriesConfig) (string, error) {
```
zaneb, line 67: Why rename this when the field that is passed in (amd64/arm64) isn't even in the format used by rhcos (x86_64/aarch64)?
We should define some constants and give them separate types so that we can't mix the two up.

pkg/asset/agent/manifests/common.go
```diff
@@ -22,7 +22,9 @@ func getInfraEnvName(ic *agent.OptionalInstallConfig) string {
 func getPullSecretName(ic *agent.OptionalInstallConfig) string {
 	return ic.Config.ObjectMeta.Name + "-pull-secret"
 }
-
+func getArchitecture(ic *agent.OptionalInstallConfig) string {
```
zaneb, line 25: Unless this is used in multiple places, I wouldn't bother defining a function for it.

pkg/asset/agent/image/ignition.go
jeffdyoung: We don't need to  do this translation with ppc64le (or s390x) because`$GOARCH` and `uname -m` are the same

data/data/agent/files/usr/local/share/assisted-service/images.env.template
```diff
@@ -1,3 +1,3 @@
...
....85.202203181601-0/x86_64/rhcos-411.85.202203181601-0-live.x86_64.iso","rootfs_url":"https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/rhcos-4.11/411.85.202203181601-0/x86_64/rhcos-411.85.202203181601-0-live-rootfs.x86_64.img","version":"411.85.202203181601-0"},{"openshift_version":"4.12","cpu_architecture":"x86_64","url":"https://releases-rhcos-art.cloud.privileged.psi.redhat.com/storage/releases/rhcos-4.12/412.86.202208170136-0/x86_64/rhcos-412.86.202208170136-0-live.x86_64.iso","rootfs_url":"https://releases-rhcos-art.cloud.privileged.psi.redhat.com/storage/releases/rhcos-4.12/412.86.202208170136-0/x86_64/rhcos-412.86.202208170136-0-live-rootfs.x86_64.img","version":"412.86.202208170136-0"},{"openshift_version":"4.12","cpu_architecture":"aarch64","url":"https://releases-rhcos-art.apps.ocp-virt.prod.psi.redhat.com/storage/releases/rhcos-4.12-aarch64/412.86.202209061132-0/aarch64/rhcos-412.86.202209061132-0-live.aarch64.iso","rootfs_url":"https://releases-rhcos-art.apps.ocp-virt.prod.psi.redhat.com/storage/releases/rhcos-4.12-aarch64/412.86.202209061132-0/aarch64/rhcos-412.86.202209061132-0-live-rootfs.aarch64.img","version":"412.86.202209061132-0"}]
```
zaneb, line 3: Just as a heads-up, we are going to stop hard-coding this and generate it on the fly in https://github.com/openshift/installer/pull/6441 - but the current implementation also hard-codes the architecture in the code. So you might want to rebase on that and add CPU Arch awareness there.

pkg/asset/agent/manifests/common.go
jeffdyoung: I kept this function so all of the arm64-> aarch64 translation can happen in one place, but I'm open to moving that logic elsewhere.

jeffdyoung: and amd64 -> x86_64 translation

jeffdyoung:
Hey @zaneb can you review while I track down some of the CI errors?  
This would need to merge in parallel: https://github.com/openshift/assisted-service/pull/4441

pkg/asset/agent/installconfig.go
```diff
@@ -94,6 +94,7 @@ func (a *OptionalInstallConfig) validateSupportedArchs(installConfig *types.Inst
 
 	switch string(installConfig.ControlPlane.Architecture) {
 	case types.ArchitectureAMD64:
+	case types.ArchitectureARM64:
 	default:
 		allErrs = append(allErrs, field.NotSupported(fieldPath, installConfig.ControlPlane.Architecture, []string{types.ArchitectureAMD64}))
```
rwsu, line 99: Add ARM64 to list of supported architectures in the error message.

```suggestion
		allErrs = append(allErrs, field.NotSupported(fieldPath, installConfig.ControlPlane.Architecture, []string{types.ArchitectureAMD64, types.ArchitectureARM64}))
```

pkg/asset/agent/installconfig.go
```diff
@@ -103,6 +104,7 @@ func (a *OptionalInstallConfig) validateSupportedArchs(installConfig *types.Inst
 
 		switch string(compute.Architecture) {
 		case types.ArchitectureAMD64:
+		case types.ArchitectureARM64:
 		default:
 			allErrs = append(allErrs, field.NotSupported(fieldPath, compute.Architecture, []string{types.ArchitectureAMD64}))
```
rwsu, line 109:
```suggestion
			allErrs = append(allErrs, field.NotSupported(fieldPath, compute.Architecture, []string{types.ArchitectureAMD64, types.ArchitectureARM64}))
```

pkg/asset/agent/installconfig_test.go
```diff
@@ -237,7 +237,7 @@ platform:
 pullSecret: "{\"auths\":{\"example.com\":{\"auth\":\"authorization value\"}}}"
 `,
 			expectedFound: false,
-			expectedError: "invalid install-config configuration: [ControlPlane.Architecture: Unsupported value: \"arm64\": supported values: \"amd64\", Compute[0].Architecture: Unsupported value: \"arm64\": supported values: \"amd64\"]",
+			expectedError: "invalid install-config configuration: [ControlPlane.Architecture: Unsupported value: \"s390x\": supported values: \"amd64\", Compute[0].Architecture: Unsupported value: \"s390x\": supported values: \"amd64\"]",
```
rwsu, line 240: Expected error will need to be updated because supported values list will  include arm64.

rwsu (changes requested):

cmd/openshift-install/testdata/agent/image/manifests/default_manifests.txt
```diff
@@ -111,6 +111,7 @@ spec:
   clusterRef:
     name: ostest
     namespace: cluster0
+  cpuArchitecture: x86_64
```
andfasano, line 114: I think it would be useful to have a new integration test for the arm64 arch. There could be a new `arch` folder to store the new tests, one for each arch, to ensure that ISO image gets properly created, with the expected name

pkg/asset/agent/image/baseiso.go
```diff
@@ -110,7 +101,13 @@ func (i *BaseIso) Generate(dependencies asset.Parents) error {
 
 	var baseIsoFileName string
 	var err error
+	archName := manifests.X86CPUArchitecture
```
andfasano, line 104: It could be useful to higlight somehow that's the default arch, at least with one comment.

pkg/asset/agent/image/baseiso.go
```diff
@@ -110,7 +101,13 @@ func (i *BaseIso) Generate(dependencies asset.Parents) error {
...
+
 	if agentManifests.ClusterImageSet != nil {
```
andfasano, line 106: Is this still valid? cc @rwsu @bfournie

cmd/openshift-install/testdata/agent/image/manifests/default_manifests.txt
jeffdyoung: @andfasano I added arm64_manifests.txt to the manifests folder, it seemed cleaner than creating subfolders. Happy to consider other options.

pkg/asset/agent/image/baseiso.go
jeffdyoung: @andfasano ended up using `DefaultCPUArchitecture` similar to AssistedInstaller.
https://github.com/openshift/assisted-service/blob/master/internal/common/common.go#L46

pkg/asset/agent/manifests/common.go
```diff
@@ -6,6 +6,13 @@ import (
 	"github.com/openshift/installer/pkg/version"
 )
 
+const (
+	AMD64CPUArchitecture   = "amd64"
+	X86CPUArchitecture     = "x86_64"
+	AARCH64CPUArchitecture = "aarch64"
+	ARM64CPUArchitecture   = "arm64"
+)
```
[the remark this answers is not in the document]
jeffdyoung: I removed `arm64` and `amd64`, but kept `aarch64` and `x86_64` here since they are specific to agent installer?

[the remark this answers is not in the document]
bfournie: Right, could those 2
 just be added to https://github.com/openshift/installer/blob/master/pkg/types/machinepools.go#L40-L49 to have all of the architecture defines in one place?

pkg/types/machinepools.go
```diff
@@ -34,18 +34,22 @@ const (
...
 
 // Architecture is the instruction set architecture for the machines in a pool.
-// +kubebuilder:validation:Enum="";amd64
+// +kubebuilder:validation:Enum="";amd64;arm64
 type Architecture string
 
 const (
 	// ArchitectureAMD64 indicates AMD64 (x86_64).
 	ArchitectureAMD64 = "amd64"
+	// X86CPUArchitecture indicates AMD64 (x86_64).
```
bfournie, line 43: Nit - the comment doesn't match the name, here and at line 51

cmd/openshift-install/testdata/agent/image/manifests/default_manifests.txt
andfasano: @jeffdyoung that sounds reasonable to me, not sure if you still have to push it?

pkg/asset/agent/manifests/common.go
```diff
@@ -51,3 +59,14 @@ func getClusterImageSetReferenceName() string {
 	versionString, _ := version.Version()
 	return "openshift-" + versionString
 }
+
+func NormalizeArch(arch string) string {
```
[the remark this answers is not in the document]
zaneb: Looks fairly trivial to add types, you're touching most of this code anyway: https://github.com/openshift/installer/commit/77ac2aa51655b29b472ed67c3f2ebbfc356f7054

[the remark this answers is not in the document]
zaneb: In fact, this function already exists and we should probably just use it: https://github.com/openshift/installer/blob/master/vendor/github.com/coreos/stream-metadata-go/arch/arch.go#L31

pkg/asset/agent/image/ignition.go
```diff
@@ -126,7 +126,8 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
 
 	a.RendezvousIP = nodeZeroIP
 
-	// TODO: don't hard-code target arch
+	archName := manifests.NormalizeArch(infraEnv.Spec.CpuArchitecture)
```
zaneb, line 129: IIUC this is the one case where we need to 'normalize', due to assisted's previous practice of using the CoreOS arch string for `x86_64` and the Docker string for `arm64`?

I think in this case I think it would be best to use one of the new functions you added in https://github.com/openshift/assisted-service/pull/4441 to do the conversion.

pkg/asset/agent/manifests/infraenv.go
```diff
@@ -70,6 +71,11 @@ func (i *InfraEnv) Generate(dependencies asset.Parents) error {
 				},
 			},
 		}
+
+		// Use RpmArch for infraEnv.Spec.CpuArchiteture. Default is amd64 or "", needs to be x86_64 or aarch64
```
zaneb, line 75: It seems like we might be halfway between two different approaches here. I think we should pick either:

1. Historically (pre https://github.com/openshift/assisted-service/pull/4441), the accepted values in ZTP were `x86_64` or `arm64`, so we should generate the manifest with those values and convert to RpmArch in ignition.go; or
2. We should be consistent in generating rpm-arch strings here, and also convert from docker-arch strings if necessary in `Load()` so that we only ever have rpm-arch in this field, and no conversion is required in ignition.go.

In either case we should have validation in `finish()` to check that only valid values have been provided in the case where we are `Load()`ing a user-provided manifest from disk. That makes me lead toward the second one.

pkg/asset/agent/manifests/infraenv.go
```diff
@@ -70,6 +71,11 @@ func (i *InfraEnv) Generate(dependencies asset.Parents) error {
...
+		if installConfig.Config.ControlPlane.Architecture != "" {
+			infraEnv.Spec.CpuArchitecture = arch.RpmArch(string(installConfig.Config.ControlPlane.Architecture))
```
zaneb, line 77: Elsewhere it looks like casting the argument to `RpmArch` to a string hasn't been required. Maybe we can drop it here too?
I know this is a nitpick but I am still a little traumatised from #6855.

pkg/asset/agent/image/baseiso.go
```diff
@@ -40,7 +37,7 @@ func (i *BaseIso) Name() string {
 }
 
 // getIsoFile is a pluggable function that gets the base ISO file
-type getIsoFile func() (string, error)
+type getIsoFile func(archName string) (string, error)
```
zaneb, line 40: It's tempting to make the argument here a `types.Architecture` and push the conversion down into `downloadIso()`. That way we get the compiler checking the types for us and in code reviews we only have to be on the lookout for dodgy casts.

pkg/asset/agent/manifests/infraenv.go
```diff
@@ -131,5 +145,9 @@ func (i *InfraEnv) finish() error {
 		return errors.New("missing configuration or manifest file")
 	}
 
+	// Throw an error if CpuArchitecture isn't RpmArch
+	if i.Config.Spec.CpuArchitecture != arch.RpmArch(i.Config.Spec.CpuArchitecture) {
```
zaneb, line 149: Given the [implementation](https://github.com/openshift/installer/blob/master/vendor/github.com/coreos/stream-metadata-go/arch/arch.go#L31-L38) of `RpmArch()`, I don't think this is a sufficient check. e.g. the hitherto-unknown CPU architecture `"fred"` would not cause an error here. In fact, there is no possible way to trigger an error here, given that we always do this conversion in `Load()`.

jeffdyoung: @zaneb I've updated/pushed the logic here: https://github.com/openshift/installer/pull/6401/files#diff-2d7fa80a467b94ef8ec0dd77511eeeb42630c9d1ccea2c5e21eaf041d6f1ed73R147

pkg/types/machinepools.go
```diff
@@ -34,17 +34,17 @@ const (
 )
 
 // Architecture is the instruction set architecture for the machines in a pool.
-// +kubebuilder:validation:Enum="";amd64
+// +kubebuilder:validation:Enum="";amd64;arm64
```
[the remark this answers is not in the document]
zaneb: The file `data/data/install.openshift.io_installconfigs.yaml` is (manually) generated from this one, so if you back out this change then the change to that file has to be backed out at the same time.

pkg/asset/agent/image/baseiso.go
zaneb: Ah, true, we'd have to cast it somewhere to get this to work, which kind of eliminates the benefit.

pkg/asset/agent/manifests/infraenv.go
```diff
@@ -131,5 +142,11 @@ func (i *InfraEnv) finish() error {
 		return errors.New("missing configuration or manifest file")
 	}
 
+	// Throw an error if CpuArchitecture isn't x86_64, aarch64, or ""
+	switch i.Config.Spec.CpuArchitecture {
+	case arch.RpmArch(types.ArchitectureAMD64), arch.RpmArch(types.ArchitectureARM64), "":
```
andfasano, line 147: @jeffdyoung I have a question: IIUC accepting the empty value `""` here will force any subsequent consumer of this field to check for a valid default (I noticed at least two dfferent ones, one in `ignition.go` and another one in `baseiso.go`). Wouldn't be easier (and less prone to error) to set the default directly here and just once for all, if the field is empty?

jeffdyoung: @andfasano Initially I did what you are saying, but I was getting a few unit test errors in https://github.com/openshift/installer/blob/master/pkg/asset/agent/manifests/infraenv_test.go
```
--- FAIL: TestInfraEnv_Generate (0.00s)
    --- FAIL: TestInfraEnv_Generate/valid_configuration (0.00s)
        infraenv_test.go:132: 
            	Error Trace:	/go/src/github.com/openshift/installer/pkg/asset/agent/manifests/infraenv_test.go:132
            	Error:      	Received unexpected error:
            	            	Config.Spec.CpuArchitecture  is not supported 
 .
 .
 .
      --- FAIL: TestInfraEnv_Generate/proxy_valid_configuration (0.00s)
        infraenv_test.go:132: 
            	Error Trace:	/go/src/github.com/openshift/installer/pkg/asset/agent/manifests/infraenv_test.go:132
            	Error:      	Received unexpected error:
            	            	Config.Spec.CpuArchitecture  is not supported 
```
Which made me think that there might be other applications that would expect to use `""` as an architecture. ([Hypershift](https://hypershift-docs.netlify.app/how-to/agent/create-agent-cluster/) may be one of those?)
Basically I did it this way so I wouldn't break other things 🤞

jeffdyoung: That said, I'm happy to make the changes if you all feel its safe to do: @zaneb @andfasano

andfasano: @jeffdyoung in general looks good to me, and I don't want to block on this specific point that could be potentially improved in a subsequent PR. Let's see if @zaneb may have any specific comment on that

jeffdyoung: I can imagine a scenario where install-config has `amd64` and InfraEnv being `multi`, but we aren't there yet.

jeffdyoung: @andfasano @zaneb we are good to resolve, and address this issue in the future when we start addressing issues with the `multi` payload?

jeffdyoung:
Lastly @zaneb @andfasano these 3 CI jobs seem to be flaky and are failing unrelated to this pr: 
`okd-e2e-aws-ovn-upgrade`
`okd-scos-e2e-aws-ovn`
`okd-scos-e2e-aws-upgrade`

Are you guys ok with an `override`?

andfasano:
> Lastly @zaneb @andfasano these 3 CI jobs seem to be flaky and are failing unrelated to this pr: `okd-e2e-aws-ovn-upgrade` `okd-scos-e2e-aws-ovn` `okd-scos-e2e-aws-upgrade`
> 
> Are you guys ok with an `override`?

Sounds fine for me, but cannot override them cc @patrickdillon

PR #6405, opened 2022-09-22
powervs: introduce manifest for OVNKubernetes configuration

clnperez:
we rely on an explicitly-configured host route, and the default OVNKube routing ignores this. as a result, packets aren't routed to the right internal gateway.

### DIFF
--- a/pkg/asset/manifests/network.go
+++ b/pkg/asset/manifests/network.go
@@ -11,13 +11,15 @@ import (
 	"github.com/openshift/installer/pkg/asset"
 	"github.com/openshift/installer/pkg/asset/installconfig"
 	"github.com/openshift/installer/pkg/asset/templates/content/openshift"
+	"github.com/openshift/installer/pkg/types/powervs"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


clnperez:
@mjturek @rna-afk @r4f4 fyi. 

i probably need to fix some linting and still need to test

pkg/asset/manifests/powervs/ovnkubeconfig.go
```diff
@@ -0,0 +1,47 @@
+package powervs
```
patrickdillon, line 1: This looks generic--not powervs specific. If that is true, could we move this to a more generic package or into network.go could be fine too

patrickdillon (commented):
Thanks for this. Ha

pkg/asset/manifests/powervs/ovnkubeconfig.go
clnperez: @patrickdillon i moved this over into just network.go. right now, the only thing configurable is what we want. i'm envisioning that as others need to tweak other params it can be built upon. lmk what you think

clnperez:
can we skip the azure-ovn test? it's not a related failure: ` Timed out waiting for node count (5) to equal or exceed machine count (6).`

Issue #6407, opened 2022-09-23
Systemd service been deactivated in limited network environment 


minatoaquaMK2:
# Version

```console
$ openshift-install version
openshift-install 4.11.0
built from commit 863cd1ea823559116e26de327705ed72ccdede8f
release image quay.io/openshift-release-dev/ocp-release@sha256:300bce8246cf880e792e106607925de0a404484637627edf5f517375517d54a4
release architecture amd64
```

# Platform:

baremetal
IPI

# What happened?

When using the agent based instller to zero-touch provision the cluster. If the network bandwidth is low, and the assisted-service or the assisted-service fails to pull the docker image within the timeout. The create-cluster-and-infraenv, apply-host-config, and start-cluster-installation services will be deactivated due to dependency failed. The process will be blocked, and require enable & start the service manually.

# What you expected to happen?

The assisted-service fails to start do not deactivate any other systemd service and the provisioning will not be blocked when assisted-service started successfully.

```
Sep 21 02:51:52 node0 systemd[1]: assisted-service-pod.service: start-pre operation timed out. Terminating.
Sep 21 02:51:52 node0 systemd[1]: Removed slice cgroup machine-libpod_pod_5c7270be24e916b503057926206884145d1a5b16252a17c24810e875f9a2175b.slice.
Sep 21 02:51:52 node0 systemd[1]: machine-libpod_pod_5c7270be24e916b503057926206884145d1a5b16252a17c24810e875f9a2175b.slice: Consumed 0 CPU time
Sep 21 02:51:52 node0 podman[2031]: 5c7270be24e916b503057926206884145d1a5b16252a17c24810e875f9a2175b
Sep 21 02:51:53 node0 systemd[1]: var-lib-containers-storage-overlay.mount: Succeeded.
Sep 21 02:51:53 node0 systemd[1]: assisted-service-pod.service: Failed with result 'timeout'.
Sep 21 02:51:53 node0 systemd[1]: Failed to start Assisted Service pod.
Sep 21 02:51:53 node0 systemd[1]: Dependency failed for Assisted Service container.
Sep 21 02:51:53 node0 systemd[1]: Dependency failed for Service that creates initial cluster and infraenv.
Sep 21 02:51:53 node0 systemd[1]: Dependency failed for Service that applies host-specific configuration.
Sep 21 02:51:53 node0 systemd[1]: Dependency failed for Service that starts cluster installation.
[...]

PR #6408, opened 2022-09-23
OCPBUGS-1824: enhance agent systemd service dependency

minatoaquaMK2:
The create-cluster-and-infraenv.service "Requires" assisted-service.service, this will make it be deactived when assisted-service.service fail to start. When assisted-service.service restarts, sinece create-cluster-and-infraenv.service was deactived, it will not restart automatically. Therefore, it is better to use "Wants=", it is a weaker version of Requires, and the dependency failure will not affect the dependent. https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Requires=

### DIFF
--- a/data/data/agent/systemd/units/create-cluster-and-infraenv.service.template
+++ b/data/data/agent/systemd/units/create-cluster-and-infraenv.service.template
@@ -1,7 +1,6 @@
 [Unit]
 Description=Service that creates initial cluster and infraenv
-Wants=network-online.target
-Requires=assisted-service.service
+Wants=network-online.target assisted-service.service
 PartOf=assisted-service-pod.service
 After=network-online.target assisted-service.service
 ConditionPathExists=/etc/assisted-service/node0


minatoaquaMK2:
#6407

minatoaquaMK2:
Given:
assisted-service-pod.service requires assisted-service-db.service assisted-service.service
assisted-service.service BindsTo=assisted-service-pod.service
create-cluster-and-infraenv.service Requires=assisted-service.service and PartOf=assisted-service-pod.service
apply-host-config.service Requires=create-cluster-and-infraenv.service
start-cluster-installation.service Requires=apply-host-config.service
Requires= "Configures requirement dependencies on other units. If this unit gets activated, the units listed here will be activated as well. If one of the other units gets deactivated or its activation fails, this unit will be deactivated."

When assisted-service-pod.service starts, assisted-service-db.service and assisted-service.service also be started,
Once assisted-service-pod.service fails to be started, assisted-service.service also fail to be started due to "BindsTo=assisted-service-pod.service".
Then dependency failed for create-cluster-and-infraenv.service due to Requires=assisted-service.service which activation fails, Therefore it will be deactived.
Then dependency failed for apply-host-config.service, due to Requires=create-cluster-and-infraenv.service which activation fails, Therefore it will be deactived.
Then  dependency failed for start-cluster-installation.service, due to Requires=apply-host-config.service which activation fails, Therefore it will be deactived.

Then assisted-service-pod.service restarts, assisted-service.service and assisted-service-db.service restarts as well, since they are binded to assisted-service-pod.service.
However, create-cluster-and-infraenv.service apply-host-config.service and start-cluster-installation.service was be deactivated, they requires to be activate mannully.

Eventually,  assisted-service started and hang with waitting for create infraenv. The provisioning is blocked.

Thus, is is better to use "Wants=assisted-service.service" rather than "Requires=", because "Wants" is weaker version of "Requires" when the dependencies gets deactivated or its activation fails, the dependant will not be deactivated.

zaneb:
Thanks for the patch. My concern with simply using Wants is that we don't actually want the scripts to start if the service is not running. Does this also have the effect of restarting the scripts if the services restart?

I had the thought of using BindsTo instead of PartOf, since that seems to work for the relationship between the services and the pod. I'll try that out and report back.

minatoaquaMK2:
> Thanks for the patch. My concern with simply using Wants is that we don't actually want the scripts to start if the service is not running. Does this also have the effect of restarting the scripts if the services restart?
> 
> I had the thought of using BindsTo instead of PartOf, since that seems to work for the relationship between the services and the pod. I'll try that out and report back.

Hi Zane, the start of the service is limited by the "After=x.service". When x.service is not started( for example pulling images), the service is not started and waiting x as its dependency. Therefore, the scripts will not start if the x service is not running. 

Using "BindsTo" has an issue similar to the "Requires", when x.service fail to start, the service "Bindsto" x.service will also be deactivated.

zaneb:
> When x.service is not started( for example pulling images), the service is not started and waiting x as its dependency. Therefore, the scripts will not start if the x service is not ready.

This is true, but once the assisted-service has failed, the next service can start. Still, the first thing it does is wait for the assisted-service API to be available, so it's not really a problem. I do wonder if maybe we only want this change in create-cluster-and-infraenv.service. None of these 3 services have `Restart:` set, so we're not actually protecting against a failure in one of those. It seems more important that we don't e.g. start the cluster installation when applying the host configs has exited with an error.

I've tried out what feels like every possible combination (including `RequiredBy=assisted-service-pod.service`) and this is the only thing that actually worked, so you are definitely on the right track.
/approve

minatoaquaMK2:
> I do wonder if maybe we only want this change in create-cluster-and-infraenv.service. None of these 3 services have `Restart:` set, so we're not actually protecting against a failure in one of those. It seems more important that we don't e.g. start the cluster installation when applying the host configs has exited with an error.

Yes, you are right. I have make this change only in the create-cluster-and-infraenv.service.

zaneb:
Thanks!
/lgtm

zaneb:
/rename OCPBUGS-1824: enhance agent systemd service dependency

zaneb:
/retitle OCPBUGS-1824: enhance agent systemd service dependency

PR #6410, opened 2022-09-23
vsphere: match username field names

bostrt:
This should match the username field used in legacy platform config and match the error message that's printed if the field is missing:

https://github.com/openshift/installer/blob/master/pkg/types/vsphere/validation/platform.go#L132-L134
`platform.vsphere.vcenters.username: Required value: must specify the username`


bostrt:
Worth noting, we don't use `vcenters` in release repo yet so this change isn't breaking anything else afaik. 

/cc @jcpowermac

jhixson74:
/lgtm

This looks fine to me, but I'd like to hear from @jcpowermac since I'm not familiar with the vsphere code.

PR #6352, opened 2022-09-15
go getter update for terraform & terraform providers

jhixson74:
This updates go-getter to v1.6.2 to address several vulnerabilities 

This can be cherry-picked into 4.13, 4.12, and 4.11.

PR #6411, opened 2022-09-23
[release-4.11] Bug 2101015: go getter update for terraform & terraform providers

jhixson74:
Backports https://github.com/openshift/installer/pull/6352


patrickdillon:
/approve

To get the valid bug label you'll need to set the bz to be dependent on the 4.12 bz

sdodson:
Adding `bugzilla/valid-bug` as this backports changes which merged into master which had no bug associated at the time. Since we can no longer create bugzillas we've created a JIRA requesting that QE test that https://issues.redhat.com/browse/OCPBUGS-2211

PR #6416, opened 2022-09-26
AGENT-319: Set agent-config version to v1beta1

zaneb:
We expect to add more fields to the agent-config file in the future (thus it is not time for v1 yet), but it will now appear in the 4.12 release and thus must be kept stable. Change the version to v1beta1.

Existing configs with version v1alpha1 will continue to work.


rwsu (commented):
/lgtm

PR #6418, opened 2022-09-26
AGENT-365: Set install invoker to "agent-installer"

zaneb:
Ensure that in telemetry data we can distinguish between the ephemeral agent-based installer and the hosted assisted installer.


pawanpinjarkar:
assisted service sets a default value so we need to set a different value to distinguish
```
InstallInvoker     string `envconfig:"INSTALL_INVOKER" default:"assisted-installer"`
```

https://github.com/openshift/assisted-service/blob/268172d230603597fe28a6de4c329a072833b2bc/pkg/generator/generator.go#L33

pawanpinjarkar (approved):
/lgtm

PR #6420, opened 2022-09-27
OCPBUGS-1737: OpenStack: Lift validation for 14 chars cluster names

mandre:
The validation is no longer necessary now that we stopped using mDNS.

Manual cherry-pick of https://github.com/openshift/installer/pull/6355/.


pierreprinetti (approved):

mandre:
The openstack job failed with an infra issue during deprovision. It has nothing to do with the PR.

PR #6423, opened 2022-09-27
OCPBUGS-2327: Add validation for releaseImage and registry.conf

pawanpinjarkar:
Ensure the releaseImage value used in the cluster-image-set.yaml matches with the registries.conf file contents


pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -208,6 +214,20 @@ func (i *RegistriesConf) finish() error {
...
 }
 
+func (i *RegistriesConf) validateReleaseImageIsSameInRegistriesConf() error {
+
+	releaseImage := &releaseimage.Image{}
+	releaseImage.Generate(asset.Parents{})
+
+	releaseImagePath := strings.Split(releaseImage.PullSpec, ":")[0]
+
+	if !strings.Contains(string(i.File.Data), releaseImagePath) {
```
zaneb, line 224: @bfournie is what we're looking for here really that [getMirrorFromRelease()](https://github.com/openshift/installer/blob/master/pkg/asset/agent/image/ignition.go#L381-L398) returns a non-empty value when a mirror config is set? I worry that the logic here is too simplistic.

Also is it really forbidden to add an ICSP that doesn't attempt to set up a mirror for the release? The user could just want to add an ICSP for their own images in the cluster on day 1, couldn't they? (Admittedly this seems unusual unless they are disconnected, in which case the release mirror really is needed.) Should this be a warning instead?

pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -208,6 +214,20 @@ func (i *RegistriesConf) finish() error {
...
+		return errors.New(fmt.Sprintf("%s should have an entry matching the releaseImage %s", RegistriesConfFilename, releaseImagePath))
```
zaneb, line 225: It's tricky to know what to put as the source here, since the user may have started with either an ICSP in the InstallConfig (when called from `Generate()`) or from a mirror/registries.conf file (when called from `Load()`). Ideally we could distinguish between the two cases.

pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -208,6 +214,20 @@ func (i *RegistriesConf) finish() error {
...
 
+func (i *RegistriesConf) validateReleaseImageIsSameInRegistriesConf(source string) error {
+
+	releaseImage := &releaseimage.Image{}
+	releaseImage.Generate(asset.Parents{})
+
+	releaseImagePath := strings.Split(releaseImage.PullSpec, ":")[0]
+
+	if !strings.Contains(string(i.File.Data), releaseImagePath) {
+		return errors.New(fmt.Sprintf("%s %s should have an entry matching the releaseImage %s", source, RegistriesConfFilename, releaseImagePath))
```
bfournie, line 225: I like the fact that the source of the setting is now identified but I'd suggest some way to point out where the faulty mirror setting was input, as "Generated" doesn't really give a clue. Maybe something like, for example:
```
if source == "Generated {
    return errors.New(fmt.Sprintf("The ImageContentSource configuration in install-config.yaml should have an entry matching the releaseImage %s", releaseImagePath)
} else {
   return errors.New(fmt.Sprintf("%s should have an entry matching the releaseImage %s", RegistriesConfFilename, releaseImagePath)
}
```

pawanpinjarkar: I had thought about a similar message earlier but later discarded it with my assumption that users know the difference between generated from install-config vs reading from the disk. I am happy to include this suggestion. Thank you.

pkg/asset/agent/mirror/registriesconf.go
andfasano: What's the reason to perform the validation against a string instead of working with a `sysregistriesv2.V2RegistriesConf` type? It would be easier to validate a struct. Moreover, I'd expect that the unmarshaling would happen in the load (a common habit for those assets that are loading manifests from the disk), also to immediately verify that a valid `registries.conf` has been passed.
This will make easier to apply the validations in the `finish()` method, and it will avoid specifying the source which I found somehow complex to manage: ie, the validations should be the same regardless of how the entity has been fetched, loaded or generated.

andfasano: Just to be clear, I think we should always follow this common pattern in all of our writable assets producing a type T:

* **Generate**:  create `t T` from InstallConfig(+AgentConfig) if present, otherwise provide a reasonable default. Then validate via `finish(t)`
* **Load**: create `t T` by unmarshaling the file if present, then validate via `finish(t)` (otherwise do nothing)

pawanpinjarkar: Currently, the unmarshalling is done in `extractLocationMirrorDataFromRegistries` from `finish()` which is called from `Generate()` as well as `Load()`. @bfournie Can you suggest why the registriesconf asset is handled a bit different here? All the other assets follow a different approach i.e. `finish()` only performs the validations and `Load()` does the unmarshalling.

@andfasano suggesting here to refactor this code such that all the assets follows the same approach.

pawanpinjarkar: @andfasano I would like to do the refactoring in a separate PR

pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -208,6 +214,23 @@ func (i *RegistriesConf) finish() error {
...
 
+func (i *RegistriesConf) validateReleaseImageIsSameInRegistriesConf(source string) error {
+
+	releaseImage := &releaseimage.Image{}
+	releaseImage.Generate(asset.Parents{})
+
+	releaseImagePath := strings.Split(releaseImage.PullSpec, ":")[0]
+
+	if !strings.Contains(string(i.File.Data), releaseImagePath) {
+		if source == "Generated" {
```
zaneb, line 225: nit: I'd prefer a boolean (or named constants) over magic strings.

pkg/asset/agent/mirror/registriesconf.go
zaneb: I agree with Andrea that it's deeply weird that we're just searching for the release string somewhere in the file without reference to the structure of it.
I still don't think I agree with Bob that we should prevent the user moving forward if they have mirrors but not of the release image. I think this should be a warning, not an error.

pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -208,6 +214,23 @@ func (i *RegistriesConf) finish() error {
 	return nil
 }
 
+func (i *RegistriesConf) validateReleaseImageIsSameInRegistriesConf(source string) error {
+
+	releaseImage := &releaseimage.Image{}
+	releaseImage.Generate(asset.Parents{})
```
zaneb, line 220: Is there a reason we can't depend on this asset and get it from the store instead of generating a new one?

pkg/asset/agent/mirror/registriesconf.go
andfasano: @pawanpinjarkar personally I don't see any valid reason to postpone such change, especially because it will greatly improve the readability and code maintenance by making this asset more similar to the other ones, with a few lines changed.
I'd also appreciate an improved test coverage on invalid config file, which apparently seems to be missing.

pkg/asset/agent/mirror/registriesconf.go
andfasano: I would even prefer not to have such distinction at all, and making the validation step completely unaware of the source

pawanpinjarkar: Restructured but I think the source field is needed especially if we want to deliver different error messages in the case of Generated Vs Loaded.

pkg/asset/agent/mirror/registriesconf.go
pawanpinjarkar: Not sure. I referred https://github.com/openshift/installer/blob/master/pkg/asset/agent/manifests/clusterimageset.go#L146-L147. I will look into getting it from store ( and what that even means :laughing: ) Probable we shoudl also change it in clusterimageset then.

pawanpinjarkar: @zaneb I am able to use the get the Image asset from the store in case of `Generate()` but in case of `Load()` I think, the asset needs to be generated.

zaneb: If we depend on it (i.e. it is returned in the list from `Dependencies()`) then it will get generated.

pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -128,17 +133,27 @@ func (*RegistriesConf) Dependencies() []asset.Asset {
 
 // Generate generates the registries.conf file from install-config.
 func (i *RegistriesConf) Generate(dependencies asset.Parents) error {
+
 	installConfig := &agent.OptionalInstallConfig{}
-	dependencies.Get(installConfig)
+	releaseImage := &releaseimage.Image{}
```
zaneb, line 138: need to also add `&releaseimage.Image{}` to `Dependencies()` at line 130.

pkg/asset/agent/mirror/registriesconf.go
zaneb: Oh, I see what you're saying. No list of parent dependencies get passed to `Load()`, so we don't have a way of fetching it. Yeah, this is unavoidable then.

pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -198,31 +239,33 @@ func (i *RegistriesConf) Load(f asset.FileFetcher) (bool, error) {
...
-	for i, reg := range registries.Registries {
-		registriesConfList[i] = RegistriesConfig{
-			Location: reg.Location,
-			Mirror:   reg.Mirrors[0].Location,
+	var found bool
+
+	for _, registry := range i.Config.Registries {
+		source := registry.Endpoint.Location
+		if !i.Generated && source == "" {
+			logrus.Warnf(fmt.Sprintf("%s should have an entry matching the releaseImage %s. Found \"%s\"", RegistriesConfFilename, i.ReleaseImagePath, source))
```
bfournie, line 256: As this check doesn't compare the ReleaseImage but just the key the warning should be something like "location key not found in %s

zaneb: We should only warn if the release image path isn't found in _any_ registry entry. By checking it inside the loop here, we require it to be found in _every_ registry entry to avoid triggering the warning.

pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -187,7 +210,25 @@ func (i *RegistriesConf) Load(f asset.FileFetcher) (bool, error) {
...
+		mirrorConfig[i] = RegistriesConfig{
+			Location: reg.Location,
+			Mirror:   reg.Mirrors[0].Location,
+		}
+	}
+	if err != nil {
+		return false, errors.Wrap(err, fmt.Sprintf("failed to parse mirrors in %s", RegistriesConfFilename))
+	}
+
+	i.MirrorConfig = mirrorConfig
```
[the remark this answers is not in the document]
pawanpinjarkar: @andfasano When I did the restructuring, my thought was we only do the unmarshalling in the `Load()` and hence when I refactored I moved the unmarshalling code into `Load()`. Originally here https://github.com/openshift/installer/blame/master/pkg/asset/agent/mirror/registriesconf.go#L199-L209 the unmarshalling is done in `finish()` which gets called in case of `Generate()` as well as `Load()`. Looking at it now makes me think why we even need to set `MirrorConfig`. I don't see it us being used anywhere else in the code. I think the public field `MirrorConfig` should be removed completely. @bfournie WDYT?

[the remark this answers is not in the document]
bfournie: The MirrorConfig is used when the registries.conf is read in ignition, e.g. https://github.com/openshift/installer/blob/master/pkg/asset/agent/image/ignition.go#L150. 

It should be set in the finish() for both Load and Generate (see line 201 and 206 below).  I thinks its a good catch that its no longer being set in the Generate.

celebdor:
/retitle OCPBUGS-2327: Add validation for releaseImage and registry.conf

pkg/asset/agent/mirror/registriesconf.go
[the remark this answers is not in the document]
pawanpinjarkar: Done

pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -187,7 +201,22 @@ func (i *RegistriesConf) Load(f asset.FileFetcher) (bool, error) {
...
-	i.File = file
+	registriesConf := &sysregistriesv2.V2RegistriesConf{}
+	if err := toml.Unmarshal(file.Data, registriesConf); err != nil {
+		return false, errors.Wrapf(err, "failed to unmarshal %s", RegistriesConfFilename)
+	}
+
+	i.File, i.Config = file, registriesConf
+	i.setMirrorConfig(i.Config)
+
+	if string(i.File.Data) != defaultRegistriesConf {
```
andfasano, line 212: In which case this could happen? The `Load` is invoked when something was found on the disk, and by assuming it was provided by the user then validations should be applied always

andfasano:
/approve

This asset had a lot of refactoring, which anyhow improved greatly its shape, so it seems a good enough incremental step

pkg/asset/agent/mirror/registriesconf.go
```diff
@@ -198,31 +227,52 @@ func (i *RegistriesConf) Load(f asset.FileFetcher) (bool, error) {
 
 func (i *RegistriesConf) finish() error {
 
-	config, err := extractLocationMirrorDataFromRegistries(i.File.Data)
-	if err != nil {
-		return errors.Wrap(err, fmt.Sprintf("failed to parse mirrors in %s", RegistriesConfFilename))
+	return nil
```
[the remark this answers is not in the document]
pawanpinjarkar: Nice catch!

pkg/asset/agent/mirror/registriesconf.go
zaneb: It could happen if the user ran `agent create cluster-manifests` and then didn't edit it, I think.
This check concerns me a bit, because we have had issues in the past where validating an empty registries.conf (which the default one is one example of) failed, and I suspect this check results in us not validating that any more in the unit tests.

PR #6432, opened 2022-09-29
bootkube: Drop cruft in MCO bootstrap

cgwalters:
These command line arguments *should* be dead code since https://github.com/openshift/installer/commit/322d4fef843ec849ab839007d71e134ae31d6e72 landed; we just didn't drop them out of conservatism.

But it turns out that hyperkube has a duplicate of this code path; it's time to clean it up.  Add a cross reference to that code.

To rephrase and highlight: A key thing we've done here is entirely remove the need to change openshift/installer (*and* openshift/hyperkube) when adding a new image reference.


cgwalters:
/reopen
This still makes sense

PR #6433, opened 2022-09-29
Bug 2117687: Check for compute.replicas must be 0 for SNO

mhanss:
When controlPlane.replicas = 1 and compute.replicas is missing from install-config.yaml, validation warns users that compute.replicas needs to be set to 0 when controlPlane.replicas = 1 for SNO cluster on none platform.

validation error: invalid install-config configuration: Compute.Replicas: Required value: Installing a Single Node Openshift requires explicitly setting compute replicas to zero

https://bugzilla.redhat.com/show_bug.cgi?id=2117687

### DIFF
--- a/pkg/asset/agent/installconfig.go
+++ b/pkg/asset/agent/installconfig.go
@@ -156,7 +156,12 @@ func (a *OptionalInstallConfig) validateSNOConfiguration(installConfig *types.In
 
 	var workers int
 	for _, worker := range installConfig.Compute {
-		workers = workers + int(*worker.Replicas)
+		if worker.Replicas == nil {
+			fieldPath = field.NewPath("Compute", "Replicas")
+			allErrs = append(allErrs, field.Required(fieldPath, "Installing a Single Node Openshift requires explicitly setting Compute.Replicas to 0"))
+		} else {
+			workers = workers + int(*worker.Replicas)
+		}
 	}
 
 	//  platform None always imply SNO cluster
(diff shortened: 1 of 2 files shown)


pkg/asset/agent/installconfig.go
```diff
@@ -170,7 +170,12 @@ func (a *OptionalInstallConfig) validateSNOConfiguration(installConfig *types.In
 
 		var workers int
 		for _, worker := range installConfig.Compute {
-			workers = workers + int(*worker.Replicas)
+			if worker.Replicas == nil {
+				fieldPath = field.NewPath("Compute", "Replicas")
```
andfasano, line 174: Isn't the check at line [166](https://github.com/openshift/installer/pull/6433/files#diff-ae1557c35cb29b1eb4fd7d286f859053055f437cb2177c8b10285a098aad5246L166) already being executed in such case?

mhanss: It will work when we don't have the `compute` in the file but if we have `compute` field and does not have a `replicas` under it then it wont work. Below is the example:

```
compute:
- architecture: amd64
  hyperthreading: Enabled
  name: worker
controlPlane:
  architecture: amd64
  hyperthreading: Enabled
  name: master
  replicas: 1
```

andfasano: Thanks @mhanss I missed the dereferencing problem initially. Even though I think that those kind of validations should be refactored into the `InstallConfig` asset, the fix looks reasonable.
It would be nice to add such test case [here](https://github.com/openshift/installer/blob/869b004e527207be564710e643498d01c16b580c/pkg/asset/agent/installconfig_test.go#L18) to ensure a proper coverage

mhanss: @andfasano I have added the test case, thanks.

mhanss:
> @mhanss looks like there is another conflicting PR in progress (#6462), where essentially the validation flow will be adjusted by applying first all the defaults and validation rules of `InstallConfig`, and then only in case of success agent specific validations will be subsequently applied. The good news it's that #6462 should already already cover this case.

Thanks @andfasano for the information. As the validation is already covered by (#6462) so closing this PR.

PR #6435, opened 2022-09-29
Add docs about IPv6 on the additional Networks

MaysaMacedo:
This doc includes limitation and steps to get additional IPv6 Networks to worker nodes and pods.

### DIFF
--- a/docs/user/openstack/connect_nodes_to_ipv6_network.md
+++ b/docs/user/openstack/connect_nodes_to_ipv6_network.md
@@ -1,0 +1,65 @@
+# Connect workers nodes and Pods to IPv6 network
+
+To connect your workers at the time of installation you can use [additionalNetworkIDs](https://github.com/openshift/installer/blob/master/docs/user/openstack/customization.md#additional-networks) parameter in the install config and set IPv6 network ID there:
+
+Example OpenShift install config:
+
+```yaml
+...
+compute:
+- name: worker
+  platform:
+    openstack:
+      additionalNetworkIDs:
+      - <ipv6_network_id>
+...
+```
+
+**Note**: To use Stateful IPv6 Networks, the arg `ip=dhcp,dhcpv6` needs to be included in the kernal args of the Worker nodes, otherwise the Nodes won't get an IPv6 address due to a [bug][dhcpv6-bug]. Use the [procedure][add-kernel-args] to add kernel argument to the Nodes.
+
+## Enable connectivity to the Pods
+
+To enable connectivity between Pods with additional Network on different Nodes, the Port security needs to be disabled for the IPv6 Port of the Server. This way it's possible to avoid adding an allowed-address-pairs with an IP and MAC address in the Server's IPv6 Port whenever a new Pod gets created.
+
+```sh
+openstack port set --no-security-group --disable-port-security <worker-ipv6-port>
+```
+
+## Add IPv6 connectivity to Pods
+
+Create a file named `network.yaml` and specify [the desired CNI config][configuring-an-additional-network]. Here is an example of CNI config used for slaac address mode with macvlan:
+
+```yaml
+spec:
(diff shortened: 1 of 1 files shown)


docs/user/openstack/connect_nodes_to_ipv6_network.md
```diff
@@ -0,0 +1,59 @@
...
+  platform:
+    openstack:
+      additionalNetworkIDs:
+      - <ipv6_network_id>
+...
+```
+
+**Note**: To use Stateful IPv6 Networks, the arg `ip=dhcp,dhcpv6` needs to be included in the kernal args of the Worker nodes, otherwise the Nodes won't get an IPv6 address. Use the [procedure][add-kernel-args] to add kernel argument to the Nodes.
+
+## Enable connectivity to the Pods
```
mandre, line 20: We should explain for what reason we recommend to disable the port security. That's because pods get random MAC addresses and this would force us to constantly change the allowed address pairs, after a new pod that is connected to this network is created. That also means that pods wouldn't be able to communicate on the network straight away, at least not until the port knew about the new allowed address pair.

docs/user/openstack/connect_nodes_to_ipv6_network.md
```diff
@@ -0,0 +1,59 @@
...
+compute:
+- name: worker
+  platform:
+    openstack:
+      additionalNetworkIDs:
+      - <ipv6_network_id>
+...
+```
+
+**Note**: To use Stateful IPv6 Networks, the arg `ip=dhcp,dhcpv6` needs to be included in the kernal args of the Worker nodes, otherwise the Nodes won't get an IPv6 address. Use the [procedure][add-kernel-args] to add kernel argument to the Nodes.
```
mandre, line 18: The fact that DHCPv6 doesn't work out of the box feels like a bug honestly. We should probably reference a jira, in addition to mentioning the workaround with modifying the kernel arguments.

MaysaMacedo: Done. I created a bug in the installer

docs/user/openstack/connect_nodes_to_ipv6_network.md
MaysaMacedo: Good point. I just added it.

mandre (commented):
LGTM, thanks. I'm adding a hold to give other folks a chance to review. Feel free to remove the hold if you think it's ready to go.
/lgtm
/hold

docs/user/openstack/connect_nodes_to_ipv6_network.md
```diff
@@ -0,0 +1,63 @@
...
+## Add IPv6 connectivity to Pods
+
+Create a file named `network.yaml` and specify [the desired CNI config][configuring-an-additional-network]. Here is an example of CNI config used for slaac address mode with macvlan:
+
+```yaml
+spec:
+  additionalNetworks:
+  - name: ipv6
+    namespace: ipv6
+    rawCNIConfig: '{ "cniVersion": "0.3.1", "name": "ipv6", "type": "macvlan", "master": "ens4"}'
```
dulek, line 37: I don't think `ens4` is set in stone, we should probably tell user that this will be different when there are more additionalNetworkIDs or when a different kernel driver is used (there were iface naming issues with PowerPC and Kuryr).

dulek (changes requested):
Comment inline.

docs/user/openstack/connect_nodes_to_ipv6_network.md
MaysaMacedo: Indeed. I just updated with it, thanks

dulek:
/lgtm

Looks good now!

mandre (commented):
/approve
/hold cancel

PR #6436, opened 2022-09-29
OCPBUGS-527: Improve error message in case of failure during the agent image generation

andfasano:
Currently whenever the `agent create image` command fails for any reason (usually a configuration error), a pretty technical and not very user-friendly message is reported.

This patch aims to provide a better message for the end user. The output will look like the following:
```
...
ERROR failed to write asset (Agent Installer ISO) to disk: cannot generate ISO image due some configuration errors 
...

### DIFF
--- a/pkg/asset/agent/image/agentimage.go
+++ b/pkg/asset/agent/image/agentimage.go
@@ -57,8 +57,10 @@ func (a *AgentImage) Generate(dependencies asset.Parents) error {
 
 // PersistToFile writes the iso image in the assets folder
 func (a *AgentImage) PersistToFile(directory string) error {
+	// If the imageReader is not set then it means that either one of the AgentImage
+	// dependencies or the asset itself failed for some reason
 	if a.imageReader == nil {
-		return errors.New("image reader not available")
+		return errors.New("cannot generate ISO image due to configuration errors")
 	}
 
 	defer a.imageReader.Close()

### CODE REVIEW

File: pkg/asset/agent/image/agentimage.go [hunk 1]
```diff
@@ -57,8 +57,10 @@ func (a *AgentImage) Generate(dependencies asset.Parents) error {
 
 // PersistToFile writes the iso image in the assets folder
 func (a *AgentImage) PersistToFile(directory string) error {
+	// If the imageReader is not set then it means that either one of the AgentImage
+	// dependencies or the asset itself failed for some reason
 	if a.imageReader == nil {
-		return errors.New("image reader not available")
+		return errors.New("cannot generate ISO image due some configuration errors")
```
bfournie, line 63: Maybe s/some/to/

zaneb:
It's still confusing to me why we are not just stopping when the configuration errors happen, but this is definitely an improvement.
/approve

andfasano:
> It's still confusing to me why we are not just stopping when the configuration errors happen, but this is definitely an improvement. /approve

I think the main reason it's due the fact that in any case the underlying infrastructure [tries to persist](https://github.com/openshift/installer/blob/c54534508191f01f99215aa18434ad506fc36c3c/cmd/openshift-install/create.go#L265-L272) the asset even though one of its dependencies failed for some reason

PR #6437, opened 2022-09-29
PowerVS: CIDR check against existing networks

23TNC:
Changing session.go's DHCP Validation check to check CIDR conflicts against all existing networks instead of just checking for an existing dhcp network as more then one may exist now.

Using net.IPNet objects to check if one cidr contains the other

### DIFF
--- a/pkg/asset/installconfig/powervs/session.go
+++ b/pkg/asset/installconfig/powervs/session.go
@@ -5,12 +5,14 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net"
 	gohttp "net/http"
 	"os"
 	"path/filepath"
 	"strings"
 	"time"
 
+	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/powervs"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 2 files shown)


pkg/asset/quota/quota.go
```diff
@@ -165,7 +165,7 @@ func (a *PlatformQuotaCheck) Generate(dependencies asset.Parents) error {
 		}
 		// @TODO: This needs to check the DHCP CIDR for a conflict, not just the existence of one
```
r4f4, line 166: Aren't you solving this TODO?

pkg/asset/installconfig/powervs/session.go
```diff
@@ -185,19 +187,31 @@ func (c *BxClient) ValidateAccountPermissions() error {
...
-	if len(dhcpServices) > 0 {
-		return fmt.Errorf("DHCP service already exists for provided cloud instance")
+		return errors.Wrap(err, "failed to get network references")
+	}
+	// Check each machineNetwork, typically one
+	for _, machineNetwork := range machineNetworks {
+                _, n1, _ := net.ParseCIDR(machineNetwork.CIDR.String())
+		// Use our networkReferences to get networks
+		for _, networkRef := range networkReferences.Networks {
+			network, err := networkClient.Get(*networkRef.NetworkID)
```
barbacbd, line 206: Is this `Get` necessary? You already got all of the network references above. You should already have that information you are looping over.

23TNC: I believe so, as the get all from the NetworkClient:
https://github.com/openshift/installer/blob/99b9d7b23a2308cddf3965de306fadfe60c6097b/vendor/github.com/IBM-Cloud/power-go-client/clients/instance/ibm-pi-network.go#L43

Returns a Networks (plural) Object
https://github.com/openshift/installer/blob/99b9d7b23a2308cddf3965de306fadfe60c6097b/vendor/github.com/IBM-Cloud/power-go-client/power/models/networks.go#L21

Which is just an array of NetworkReference objects, 
https://github.com/openshift/installer/blob/99b9d7b23a2308cddf3965de306fadfe60c6097b/vendor/github.com/IBM-Cloud/power-go-client/power/models/network_reference.go#L21

We need a cidr, which seems to come from a Network (singular) object
https://github.com/openshift/installer/blob/99b9d7b23a2308cddf3965de306fadfe60c6097b/vendor/github.com/IBM-Cloud/power-go-client/power/models/network.go#L22

To get that Network (Singular) object we can call a Get against our NetworkClient, which requires a Network id
https://github.com/openshift/installer/blob/99b9d7b23a2308cddf3965de306fadfe60c6097b/vendor/github.com/IBM-Cloud/power-go-client/clients/instance/ibm-pi-network.go#L28

I could be a complete idiot and missing something obvious, but I don't think we have our Network (singular) objects from our GetAll call against our NetworkClient.

pkg/asset/quota/quota.go
23TNC: Thank you, Ill remove it

pkg/asset/installconfig/powervs/session.go
```diff
@@ -185,19 +187,31 @@ func (c *BxClient) ValidateAccountPermissions() error {
...
+			if err != nil {
+				return errors.Wrap(err, "failed to get network")
+			}
+			_, n2, _ := net.ParseCIDR(*network.Cidr)
+			if ( n2.Contains(n1.IP) || n1.Contains(n2.IP) ) {
+				return fmt.Errorf("Cidr conflicts with existing network")
```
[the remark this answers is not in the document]
23TNC: I apologize for the oversight.

pkg/asset/installconfig/powervs/session.go
23TNC: @barbacbd Based on the above do you agree that a Get seems to be necessary here?

mjturek: @barbacbd I looked into this and I agree that the Get is necessary. The NetworkReference objects do not provide subnet CIDRs, so you have to get based on the ID to determine it.

clnperez:
we don't need to check all networks for CIDR conflicts. the only time a CIDR conflict is going to be a problem, is if that network is also attached to *a* (not the same, even), cloud connection. and we're safe just checking for cloud connections (created via API) vs direct link connections (set up manually through support), as the traffic through those are routed differently in the backend fabric.

clnperez:
i may have left a little ambiguity there. to be clear, we _should_ loop through all networks, not just dhcp-managed ones. those can also be attached to cloud connections.

hamzy:
@23TNC You can close this issue now.  It has been replaced by https://github.com/openshift/installer/pull/6798

PR #6438, opened 2022-09-29
AGENT-284: Allow user to choose platform type

pawanpinjarkar:
This PR reads the platform name from `install-config.yaml` and sets the appropriate and supported platform in the `AgentClusterInstall.Spec.PlatformType` . Assisted service uses the `AgentClusterInstall.Spec.PlatformType` as cluster params to create the cluster. When using the ZTP way, users can directly provide the supported platforms in `AgentClusterInstall.Spec.PlatformType`.

We already have a validation to throw an error if users provide unsupported platforms for the agent installer in the install-config.yaml. This PR also adds a similar validation when unsupported `PlatformType` is provided in the AgentClusterInstall.

Signed-off-by: Zane Bitter <zaneb@redhat.com>


zaneb:
Would be good to validate that for SNO specifically, only platform: none is allowed. Pawan is opening another Jira ticket to follow up with that.
/lgtm

pawanpinjarkar:
Opened https://issues.redhat.com/browse/AGENT-369 for additional validation for the SNO cluster

pkg/asset/agent/common.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"github.com/openshift/installer/pkg/types/baremetal"
+	"github.com/openshift/installer/pkg/types/none"
+	"github.com/openshift/installer/pkg/types/vsphere"
+)
+
+// SupportedPlatforms lists the supported platforms for agent installer
+var SupportedPlatforms = []string{baremetal.Name, vsphere.Name, none.Name}
+
+// IsSupportedPlatform returns true if provided platform is baremeral, vsphere or none.
```
bfournie, line 12: Typo (baremeral), but I'd recommend just removing this comment as it just reiterates the code 2 lines above and makes it harder to maintain when we support other platforms

zaneb: linter will complain if you remove it altogether, but agree that there's no need to list the platforms here.

PR #6440, opened 2022-09-30
openstack: Allow BGP to route traffic to API

pierreprinetti:
as an alternative to keepalived/haproxy


EmilienM:
@pierreprinetti do we want to rename `apiLoadbalancerType` to `loadbalancerType`? because we'll need it for Ingress as well, so I guess we need to decide whether or not we split them or use a single variable.

EmilienM:
https://github.com/openshift/installer/pull/6812

PR #6441, opened 2022-09-30
AGENT-347: generate OS_IMAGES dynamically

andfasano:
This patch uses the current rhcos.json information stored in the installer to craft properly the OS_IMAGES env var (arch is currently hard-coded to x86_64)


pkg/asset/agent/image/ignition.go
```diff
@@ -360,6 +373,53 @@ func addExtraManifests(config *igntypes.Config, extraManifests *manifests.ExtraM
 	}
 }
 
+func getOSImagesInfo() (*models.OsImage, error) {
+	ctx, cancel := context.WithTimeout(context.TODO(), 30*time.Second)
+	defer cancel()
+
+	st, err := rhcos.FetchCoreOSBuild(ctx)
```
zaneb, line 380: fwiw I think it's fine to use `context.Background()` here. It should be getting the stream info from the file built in to the binary. (TBH I don't understand why this requires passing a context at all.)

pkg/asset/agent/image/ignition.go
```diff
@@ -360,6 +373,53 @@ func addExtraManifests(config *igntypes.Config, extraManifests *manifests.ExtraM
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Architecture currently hard-coded to x86_64
+	cpurArch := archName
```
zaneb, line 386: Let's make this a parameter, since we know it will have to be one eventually. Multi-arch folks are already working on ARM support in #6401.

pkg/asset/agent/image/ignition.go
```diff
@@ -360,6 +373,53 @@ func addExtraManifests(config *igntypes.Config, extraManifests *manifests.ExtraM
...
+	if !ok {
+		return nil, fmt.Errorf("unable to retrieval coreos metal info for architecture %s", archName)
+	}
+	osImage.Version = &artifacts.Release
+
+	isoFormat, ok := artifacts.Formats["iso"]
+	if !ok {
+		return nil, fmt.Errorf("unable to retrieval coreos ISO info for architecture %s", archName)
+	}
+	osImage.URL = &isoFormat.Disk.Location
```
zaneb, line 412: We already have code to retrieve this (in the download fallback if we can't extract the image from the payload), I wonder if we could reuse it here.

zaneb (commented):
/approve

lranjbar (commented):
Non blocking question/comment: This PR seems to assume that there will only be one version and CPU architecture. Do we think that OCP will have multi arch images in the future? LGTM mostly just thinking ahead to multi arch support.

pkg/asset/agent/image/ignition.go
pawanpinjarkar: Ditto. There isn't any need to create the `ctx` instead use `context.Background()` directly. `pkg.rhcos.builds.FetchCoreOSBuild()` calls `FetchRawCoreOSStream()` which even then does not use the `ctx` object. I will create a seperate PR to remove the parameter but till then we can just pass `nil` or `context.Background()`

pkg/asset/agent/image/ignition.go
```diff
@@ -360,6 +373,53 @@ func addExtraManifests(config *igntypes.Config, extraManifests *manifests.ExtraM
...
+		return nil, err
+	}
+
+	// Architecture currently hard-coded to x86_64
+	cpurArch := archName
+	osImage := &models.OsImage{
+		CPUArchitecture: &cpurArch,
+	}
+
+	ver, err := version.Version()
```
pawanpinjarkar, line 391: for a moment, it looked like `var`. 
`version, err := version.Version()`

pkg/asset/agent/image/ignition.go
```diff
@@ -360,6 +373,53 @@ func addExtraManifests(config *igntypes.Config, extraManifests *manifests.ExtraM
...
+
+	artifacts, ok := streamArch.Artifacts["metal"]
+	if !ok {
+		return nil, fmt.Errorf("unable to retrieval coreos metal info for architecture %s", archName)
+	}
+	osImage.Version = &artifacts.Release
+
+	isoFormat, ok := artifacts.Formats["iso"]
+	if !ok {
+		return nil, fmt.Errorf("unable to retrieval coreos ISO info for architecture %s", archName)
```
pawanpinjarkar, line 410: Nit: `failed to retrieve`

pkg/asset/agent/image/ignition.go
```diff
@@ -360,6 +373,53 @@ func addExtraManifests(config *igntypes.Config, extraManifests *manifests.ExtraM
...
+	}
+	osImage.URL = &isoFormat.Disk.Location
+
+	pxeFormat, ok := artifacts.Formats["pxe"]
+	if !ok {
+		return nil, fmt.Errorf("unable to retrieval coreos rootfs info for architecture %s", archName)
```
pawanpinjarkar, line 416: Nit: `failed to retrieve`

pawanpinjarkar (changes requested):
Minor changes TBD

pkg/asset/agent/image/ignition.go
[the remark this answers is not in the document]
celebdor: Yes, we'll need to change this code when we tackle the multi arch epic

pkg/asset/agent/image/ignition.go
andfasano: fair point :)

andfasano:
/test okd-scos-images

(this job is not related at all with this patch)

PR #6443, opened 2022-09-30
Add initial support for OKD/SCOS

LorbusChris:
These changes add initial support for OKD on CentOS Stream CoreOS 9
aka SCOS.

Initially, and until bootimages are available, OKD/SCOS will provision
nodes with Fedora CoreOS, and then rebase them to OKD/SCOS' OSTree.

In order to differentiate between OKD/FCOS and
OKD/SCOS builds, the new `fcos` and `scos` build tags are added.
The existing `okd` build tag remains usable for OKD/FCOS builds, and
can be removed at a later time.

/cc @vrutkovs
/cc @cgwalters 
/cc @aleskandro 
/cc @sherine-k 
/cc @lmzuccarelli


data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
```diff
@@ -0,0 +1,42 @@
...
+        record_service_stage_success
+        break
+    else
+        record_service_stage_failure
+        echo "Pull failed. Retrying ${MACHINE_OS_IMAGE}..."
+    fi
+  done
+
+  record_service_stage_start "rebase-to-okd-os-image"
+  mnt="$(podman image mount "${MACHINE_OS_IMAGE}")"
```
cgwalters, line 33: Looking at this, it seems like we're going through a lot of *extra* complexity to pull the code added in https://github.com/openshift/os/pull/990 and it'd be way simpler to just include that code here too.

LorbusChris:
@cgwalters I think that's a good idea, but I'd like to not mix these things up right now and tackle that as a follow up.

I've updated this PR per discussion in https://github.com/openshift/os/pull/990

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
cgwalters: OK, understood.  How often does it change though?  

(One argument for not having it in the OS repository is that we don't actually test cluster installs there yet (embarrassingly enough, it shouldn't be hard to fix but it's kind of pointless because 95% of the changes that could break cluster installs are RPM changes, which *aren't* gated on PRs))

We could in theory of course move some of this logic to a separate (privileged) container image that we pull and run, but I'm not sure which github repo that'd be, and it'd have the same problem that it would really become heavily load-bearing and need e2e tests there too.
(In theory for example some of this logic could live in Go code in the MCO binary too, dunno if that's better or not)

LorbusChris: We did have a couple of changes there from time to time, especially recently with the migration to the ostree container rebase, and from time to time also regarding the OKD-specific manifests shipped there.

For now this PR doesn't change those mechanics in OKD/FCOS, but going forward I think doing this directly here in the installer would be OK.

I definitely want to loop in @vrutkovs on this.

data/data/manifests/bootkube/cvo-overrides.yaml.template
```diff
@@ -4,9 +4,12 @@ metadata:
   namespace: openshift-cluster-version
   name: version
 spec:
-{{- if .IsOKD }}
+{{- if .IsFCOS }}
   upstream: https://amd64.origin.releases.ci.openshift.org/graph
   channel: stable-4
+{{- else if .IsSCOS }}
+  upstream: https://amd64.okd-scos.releases.ci.openshift.org/graph
```
LorbusChris, line 11: Note: This doesn't exist yet.

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
vrutkovs: >How often does it change though?

Rarely.

>(In theory for example some of this logic could live in Go code in the MCO binary too, dunno if that's better or not)

I think it would be preferable. We already have okd-specific build tag there and the team is actively working on OKD/SCOC

LorbusChris:
/hold
let's get https://github.com/openshift/release/pull/32780 and https://github.com/openshift/release/pull/32864 in first

data/data/manifests/bootkube/cvo-overrides.yaml.template
```diff
@@ -4,9 +4,12 @@ metadata:
   namespace: openshift-cluster-version
   name: version
 spec:
-{{- if .IsOKD }}
+{{- if .IsFCOS }}
   upstream: https://amd64.origin.releases.ci.openshift.org/graph
   channel: stable-4
+{{- else if .IsSCOS }}
+  upstream: https://amd64.origin.releases.ci.openshift.org/graph
+  channel: scos-stable-4
```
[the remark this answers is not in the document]
bradmwilliams: Unfortunately, that channel will not work:
https://origin-release.ci.openshift.org/graph?channel=scos-stable-4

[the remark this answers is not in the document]
bradmwilliams: The release-controller is expecting a channel starting with `stable-`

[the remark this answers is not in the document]
LorbusChris: Thanks! I've changed this to `stable-scos-4` now. We can fiddle with the release controller further to get working later as we don't need upgrades to work for OKD/SCOS initially.

LorbusChris:
We have to wait for the next OCP CI release to land. The images will then be mirrored to the `origin/scos-4.12` imagestream, which will in turn enable e2e testing here.

/hold cancel

LorbusChris:
The origin/scos-4.12 ImageStream has been populated manually, tests should at least start to run now

/retest

travier:
It might be a consequence of the FCOS -> SCOS "downgrade"/rebase

PR #6445, opened 2022-09-30
Increase start timeouts for assisted services

zaneb:
Testing in a real lab shows that it takes 333s for assisted-service-pod and 293s for assisted-service and assisted-service-db to start up (this is due to the time taken to pull the images). Set the timeouts so that this generally should not happen.


pawanpinjarkar (commented):
/lgtm

PR #6450, opened 2022-10-04
WIP: CORS-2291: Log terraform messages to file.

rna-afk:
Terraform logs are currently being added to the installer logs even if the TF_LOG_PATH variable is set. This causes the logs to become large and unreadable.

Moving the terraform logs to a separate file determined by the TF_LOG_PATH variable.


r4f4:
@rna-afk the change looks good to me, but it would be nice if you added a throwaway commit to set `TF_LOG_PATH` so we can verifiy it in action in the CI infra.

rna-afk:
Closing in favor of #6532

PR #6453, opened 2022-10-04
OCPBUGS-1896: Validate Credentials Mode After Install Config step

barbacbd:
** The install config would fail immediately when the credentials were nil. There was no chance to set the Credentials Mode to Manual. Now the install config can be generated then the user can manually set the credentials mode to manual. ** More tests were generated for this functionality

### DIFF
--- a/pkg/asset/installconfig/gcp/validation.go
+++ b/pkg/asset/installconfig/gcp/validation.go
@@ -366,15 +366,32 @@ func validateRegion(client API, ic *types.InstallConfig, fieldPath *field.Path) 
 	return nil
 }
 
-// validateCredentialMode checks whether the credential mode is
+// ValidateCredentialMode checks whether the credential mode is
 // compatible with the authentication mode.
-func validateCredentialMode(client API, ic *types.InstallConfig) field.ErrorList {
-	allErrs := field.ErrorList{}
+func ValidateCredentialMode(client API, ic *types.InstallConfig) error {
 	creds := client.GetCredentials()
 
 	if creds.JSON == nil && ic.CredentialsMode != types.ManualCredentialsMode {
 		errMsg := "environmental authentication is only supported with Manual credentials mode"
-		allErrs = append(allErrs, field.Forbidden(field.NewPath("credentialsMode"), errMsg))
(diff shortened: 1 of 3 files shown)


patrickdillon:
This looks good. Two suggestions:

1. we could still keep validation that if `credentialMode != ""` we ensure it is manual mode in this case. This would prevent an incorrectly configured install config from slipping through.
2. It might be good to have a WARNING in the install config validation when the credentialMode is unset and using environmental auth. Something to say "You are currently using Environmental Auth. You either need to use manual mode or provide a service account json file." 

The reason I suggest point two is because this new validation will be run when we get to create cluster. If the user runs create install-config and everything seems fine, they run `create cluster` and it spits out an error, they can just edit the credentialMode rerun create cluster and the install will fail. Hopefully they will think twice when they switch credentialMode, and there is only so much validation we can do but it might be nice to have a little more warning up front, although I don't think it's strictly necessary.

patrickdillon:
/approve

Looks like gofmt may be failing on whitespace

patrickdillon:
LGTM but just need to fix the failing test.

barbacbd:
/override ci/prow/e2e-gcp-ovn

Overriding as the issues do not look related.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -366,15 +366,32 @@ func validateRegion(client API, ic *types.InstallConfig, fieldPath *field.Path)
...
+
+func validateCredentialMode(client API, ic *types.InstallConfig) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	creds := client.GetCredentials()
+	if creds.JSON == nil {
+		if ic.CredentialsMode == "" {
+			logrus.Warn("Currently using GCP Environmental Authentication. Please set credentialsMode to manual, or provide a service account json file.")
+		} else {
+			if ic.CredentialsMode != "" && ic.CredentialsMode != types.ManualCredentialsMode {
```
sadasu, line 390: The else here already implies that ic.CredentialsMode != "". So, just checking if the valus is not types.ManualCredentialsMode should be sufficient.

PR #6454, opened 2022-10-04
OCPBUGS-1936: bump RHCOS 4.10 bootimage metadata

mike-nguyen:
These changes will update the RHCOS 4.10 boot image metadata in the installer which includes the fixes for the following:

OCPBUGS-1938 Booting live ISO: /dev/sr0 already mounted or mount point busy

```
plume cosa2stream --target data/data/coreos/rhcos.json --distro rhcos --no-signatures --url https://rhcos.mirror.openshift.com/art/storage/releases x86_64=410.84.202210040010-0 aarch64=410.84.202210040011-0 s390x=410.84.202210040017-0 ppc64le=410.84.202210040023-0
```

### DIFF
--- a/data/data/coreos/rhcos.json
+++ b/data/data/coreos/rhcos.json
@@ -1,83 +1,83 @@
 {
   "stream": "rhcos-4.10",
   "metadata": {
-    "last-modified": "2022-07-11T21:13:21Z",
-    "generator": "plume cosa2stream 0.14.0+95-gbedcef257-dirty"
+    "last-modified": "2022-10-04T12:44:03Z",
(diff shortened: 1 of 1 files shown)


jlebon (approved):
This was all tested by CI, but I've also manually taken the x86_64 ISO for a spin (in fact, multiple spins :cd:) as a sanity-check and it works well.

/lgtm

ashcrow:
Looks like all required tests have passed

ashcrow:
cc @sdodson

sdodson:
The 4.11 version will merge shortly but given the install time only concern here and the fact that we're fixing just one bug accept this out of order.

PR #6455, opened 2022-10-04
Require the user to specify the `none` platform when performing bootstrap-in-place

omertuc:
Currently the installer allows users to do bootstrap in place
(`./openshift-install create single-node-ignition-config`) even when
they're using platforms other than `none`.

This is unsupported and will probably not work, yet the manifests get
generated without any explicit error about it.

To prevent users from doing this, the installer will now give an error
when the user runs `single-node-ignition-config` with a platform that is
not the none-platform.

(The installer already correctly gives an error when the user doesn't
set any platform or when the user sets more than one platform, so it's
sufficient to just additionaly check that the none platform is set)


omertuc:
Won't do

PR #6456, opened 2022-10-04
OCPBUGS-1881: vsphere/destroy: improve logs when folder is not empty

r4f4:
If the folder is not empty and it contains anything other than "virtualmachine" objects, the installer might be stuck in a loop and the only message we get about it is:

level=debug msg="Folder: Expected Folder $folder to be empty"

Let's improve that message by expanding on which objects in the folder are preventing the installer from destroying the cluster.

### DIFF
--- a/pkg/destroy/vsphere/vsphere.go
+++ b/pkg/destroy/vsphere/vsphere.go
@@ -145,22 +145,25 @@ func (o *ClusterUninstaller) deleteFolder() error {
 	// If there are no children in the folder, go ahead and remove it
 
 	for _, f := range folderMoList {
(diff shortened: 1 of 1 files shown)


sadasu:
/approve
@r4f4 nice fix! Do you think this needs to be back ported to older releases?

r4f4:
> /approve @r4f4 nice fix! Do you think this needs to be back ported to older releases?

That's a good question. I think it would be useful and it's a low risk change. The only caveat is that there was a big rewrite on the destroy code in 4.11, so 4.10 and earlier will most likely require manual cherry-picking. But I have no problem with doing that.

PR #6458, opened 2022-10-04
OCPBUGS-1856: IBMCloud: Allow traffic to kube-api-lb

cjschaef:
Allow traffic from any source to the kube-api-lb apiserver for IBM Cloud. Private cluster traffic should already be restricted to be from within the VPC, via DNS Services, so removing the restriction it must also be from within the cluster's network as well.

Related: https://issues.redhat.com/browse/OCPBUGS-1856


MayXuQQ (approved):
checked with registry.build05.ci.openshift.org/ci-ln-sk55dy2/release:latest PASS

PR #6461, opened 2022-10-05
OCPBUGS-2029: Fix agent installation InstallConfig proxy processing

celebdor:
In order for the proxy config to make it to the installed cluster, it must make it into the infraenv that we then register in the assisted service API. This commit sees to it that if installConfig has the proxy section provided, it makes into the infraenv.

### DIFF
--- a/pkg/asset/agent/manifests/common.go
+++ b/pkg/asset/agent/manifests/common.go
@@ -4,6 +4,8 @@ import (
 	"github.com/openshift/installer/pkg/asset/agent"
 	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/version"
+
+	aiv1beta1 "github.com/openshift/assisted-service/api/v1beta1"
 )
 
 func getAgentClusterInstallName(ic *agent.OptionalInstallConfig) string {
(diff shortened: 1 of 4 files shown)


pkg/asset/agent/manifests/common.go
```diff
@@ -22,6 +24,14 @@ func getPullSecretName(ic *agent.OptionalInstallConfig) string {
 	return ic.ClusterName() + "-pull-secret"
 }
 
+func getProxy(ic *agent.OptionalInstallConfig) *aiv1beta1.Proxy {
```
andfasano, line 27: I think we should really stop with this approach, for a couple of reasons:

* Defining a method far away from the original type is not maintenance-friendly and detrimental for the readability
* The consumer code should have a reference only to `types.InstallConfig`. The syntethic type `OptionalInstallConfig` should be used only to extract the underlying installConfig if present, and no other responsibility should be given to that type

celebdor: Agreed. I think making a refactor after the branching is a good idea

zaneb: The reason for most of these functions to exist is that they are used in multiple ZTP manifests to produce data that has to match up for the ZTP manifests to work correctly, and their implementations may change in the future. e.g. the PullSecret name has to be the same in the Secret, the InfraEnv, and the ClusterDeployment - hence `getPullSecretName()`.

I don't know at what point this became a cargo-cult thing, but there is absolutely no reason for this function to exist. We can just get the data out of the InstallConfig where we need it.

PR #6464, opened 2022-10-05
AGENT-349: Allow console login to Agent ISO

zaneb:
Allow the 'core' user to log in to the agent ISO via the host console using the kubeadmin password for the cluster.

sshd in CoreOS is configured not to allow password authentication, so this doesn't allow the user to login via ssh without supplying their public key when building the image.

The password hash is not persisted to disk, so this does not allow users to log in to the installed OpenShift node after rebooting.


bfournie:
Tested with dev-scripts and confirmed it didn't affect ssh login.

/lgtm

zaneb:
There is no possible mechanism by which this patch could be affecting the AWS build.
/hold cancel
/retest-required

PR #6469, opened 2022-10-06
OCPBUGS-881: fail to create install-config.yaml as apiVIP and ingress VIP are not in machine networks

rvanderp3:
validate VIPs are in machineNetwork in survey.  validation was added for vsphere, ovirt, and nutanix.

### DIFF
--- a/pkg/asset/installconfig/nutanix/nutanix.go
+++ b/pkg/asset/installconfig/nutanix/nutanix.go
@@ -11,14 +11,14 @@ import (
 	"github.com/AlecAivazis/survey/v2"
 	nutanixclient "github.com/nutanix-cloud-native/prism-go-client"
 	nutanixclientv3 "github.com/nutanix-cloud-native/prism-go-client/v3"
-	"github.com/openshift/installer/pkg/types"
-	"github.com/openshift/installer/pkg/types/defaults"
-	"github.com/openshift/installer/pkg/types/validation"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
+	"github.com/openshift/installer/pkg/types"
+	"github.com/openshift/installer/pkg/types/defaults"
 	"github.com/openshift/installer/pkg/types/nutanix"
 	nutanixtypes "github.com/openshift/installer/pkg/types/nutanix"
+	"github.com/openshift/installer/pkg/types/validation"
 	"github.com/openshift/installer/pkg/validate"
 )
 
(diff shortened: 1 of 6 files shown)


rvanderp3:
> /test e2e-vsphere /test e2e-nutanix

hmmm, these tests don't really test the behavior since these are install-config survey changes.  We'll run them anyway though.

rvanderp3:
cc: @patrickdillon

pkg/asset/installconfig/nutanix/nutanix.go
```diff
@@ -265,6 +268,9 @@ func getSubnet(ctx context.Context, client *nutanixclientv3.Client, peUUID strin
 
 func getVIPs() (string, string, error) {
 	var apiVIP, ingressVIP string
+	var installConfig types.InstallConfig
+	installConfig.Platform = types.Platform{Nutanix: &nutanix.Platform{}}
+	defaults.SetInstallConfigDefaults(&installConfig)
```
patrickdillon, line 273: nit: It might be simpler/less confusing to export `defaultMachineCIDR` from package `types/defaults`; because I wonder if this implies "we're creating the install config here", which we're not.

patrickdillon (commented):
/approve

pkg/asset/installconfig/nutanix/nutanix.go
rvanderp3: yeah, I was torn on that tbh.  My concern was that based on the platform, that defaultMachineCIDR might not be the actual defaultMachineCIDR.  While this doesn't impact any current onprem platforms, I was concerned this might cause some confusion down the line.  another option here might be to pull the logic out of SetInstallConfigDefaults that determines the final defaultMachineCIDR of the platform.

rvanderp3:
/hold 
placing a hold to address comment

pkg/asset/installconfig/nutanix/nutanix.go
rvanderp3: thinking about this a little more, I think i'll go ahead and revert to your suggestion here.

rvanderp3:
/hold cancel

made update per https://github.com/openshift/installer/pull/6469#discussion_r990098652

/assign @rna-afk

patrickdillon:

This is just failing on a flake unrelated to this PR.

rvanderp3:
hmmmm ... something has changed.  taking a look.

rvanderp3:
looks like linting rules changed.  ive pushed an update.

r4f4:
> looks like linting rules changed. ive pushed an update.

Correct. We've standardized the import order [1] and replaced the deprecated linter by a new one.

[1] https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#import-order

PR #6470, opened 2022-10-06
OCPBUGS-2086: Detect failure in wait-for on transition back to ready

bfournie:
If a cluster moves from "preparing-for-installation" back to "ready" it indicates that a failure has occurred and installation will not continue. Note that in order to check the status the return from the validation check is moved to after the handling of the status errors.

### DIFF
--- a/pkg/agent/cluster.go
+++ b/pkg/agent/cluster.go
@@ -114,11 +114,11 @@ func NewCluster(ctx context.Context, assetDir string) (*Cluster, error) {
 }
 
 // IsBootstrapComplete Determine if the cluster has completed the bootstrap process.
-func (czero *Cluster) IsBootstrapComplete() (bool, error) {
+func (czero *Cluster) IsBootstrapComplete() (bool, bool, error) {
 
 	if czero.installHistory.ClusterBootstrapComplete {
 		logrus.Info("Bootstrap is complete")
-		return true, nil
+		return true, false, nil
 	}
 
 	clusterKubeAPILive, clusterKubeAPIErr := czero.API.Kube.IsKubeAPILive()
@@ -138,7 +138,7 @@ func (czero *Cluster) IsBootstrapComplete() (bool, error) {
 		if configmap {
 			logrus.Info("Bootstrap configMap status is complete")
 			czero.installHistory.ClusterBootstrapComplete = true
-			return true, nil
+			return true, false, nil
 		}
 		if err != nil {
 			logrus.Debug(err)
@@ -162,7 +162,7 @@ func (czero *Cluster) IsBootstrapComplete() (bool, error) {
 		if czero.clusterID == nil {
 			clusterID, err := czero.API.Rest.getClusterID()
 			if err != nil {
-				return false, errors.Wrap(err, "Unable to retrieve clusterID from Agent Rest API")
+				return false, false, errors.Wrap(err, "Unable to retrieve clusterID from Agent Rest API")
 			}
 			czero.clusterID = clusterID
 		}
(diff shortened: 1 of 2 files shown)


pkg/agent/cluster.go
```diff
@@ -185,11 +185,14 @@ func (czero *Cluster) IsBootstrapComplete() (bool, error) {
...
-			return false, errors.New("cluster host validations failed")
-		}
-
 		czero.PrintInstallStatus(clusterMetadata)
+
+		// If the status changes back to Ready from Installing it indicates an error
+		if *clusterMetadata.Status == models.ClusterStatusReady &&
+			czero.installHistory.RestAPIPreviousClusterStatus == models.ClusterStatusPreparingForInstallation {
+			logrus.Warn("Cluster has stopped installing... state has changed back to Ready from Preparing")
+			return false, errors.New("cluster installation has changed to Ready from Preparing")
```
zaneb, line 194: The error message should be something like "Failed to prepare cluster installation".

Unfortunately, returning an err doesn't actually do anything because the return value is not being checked. We need another mechanism that will result in (a) the error message being printed (instead of the warning above), and (b) the process exiting.

That could mean replacing all of the other errors in this function with nil, since they don't do anything, and then actually checking the return value. Or we could add another return value. Or something else.

zaneb:
We also need to handle the case where `wait-for bootstrap-complete` is run only after assisted has already failed to prepare the installation. See the Jira ticket for more details. That could be in another PR, since it's quite a bit more complicated.

bfournie:
> We also need to handle the case where `wait-for bootstrap-complete` is run only after assisted has already failed to prepare the installation. See the Jira ticket for more details. That could be in another PR, since it's quite a bit more complicated.

Yes I was planning on handling that after we got agreement on the this first one.

celebdor:
/title OCPBUGS-2086: Detect failure in wait-for on transition back to ready

pkg/agent/cluster.go
bfournie: I changed to return an additional value to indicate whether the process should exit. This allows logs to be output which will not result in an exit, I've also removed some logs (like the validation) that happen every cycle.

When the error occurs the wait-for now exits with:
level=error msg=Bootstrap failed to complete: : bootstrap process returned error: Failed to prepare cluster installation

pkg/agent/waitfor.go
```diff
@@ -26,13 +26,22 @@ func WaitForBootstrapComplete(assetDir string) (*Cluster, error) {
...
 	wait.Until(func() {
-		bootstrap, err := cluster.IsBootstrapComplete()
+		bootstrap, err, exit_on_err := cluster.IsBootstrapComplete()
 		if bootstrap && err == nil {
 			logrus.Info("cluster bootstrap is complete")
 			cancel()
 		}
 
+		if err != nil {
+			logrus.Info(err)
```
zaneb, line 38: I think we should only have this Info log if exit_on_err is false.

zaneb:
linter is not happy: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/6470/pull-ci-openshift-installer-master-golint/1578161321176731648#1:build-log.txt%3A32-37

pkg/agent/waitfor.go
bfournie: Changed to log here if not exiting.

bfournie:
> linter is not happy: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/6470/pull-ci-openshift-installer-master-golint/1578161321176731648#1:build-log.txt%3A32-37

Fixed

pkg/agent/cluster.go
```diff
@@ -156,60 +157,79 @@ func (czero *Cluster) IsBootstrapComplete() (bool, error) {
...
+			if czero.installHistory.RestAPIPreviousClusterStatus == models.ClusterStatusReady {
+				current := time.Now()
+				elapsed := current.Sub(czero.installHistory.NotReadyTime)
+				if elapsed > 1*time.Minute {
+					return false, true, errors.New("failed to progress after all hosts available")
+				}
+			}
+		} else {
+			czero.installHistory.NotReadyTime = time.Now()
 		}
```
lranjbar, line 208: Lets move this to its own function instead of adding a second boolean to the overall function. The second boolean is confusing to me coming out of `IsBootstrapComplete()`. Maybe something like `RegressionInClusterStatus()`

lranjbar (changes requested):

pkg/agent/cluster.go
bfournie: Moved to function

PR #6479, opened 2022-10-12
OCPBUGS-2259: Fix soft-anti-affinity known issue

pierreprinetti:
Before this documentation patch, the known issue about soft-anti-affinity had several issues:
* it was in the UPI section, when it is not a UPI-specific issue
* it mentioned Control plane scale-out, when OCP only supports exactly 3 masters
* it is now possible to set strict anti-affinity from the install-config.yaml, and that should be the recommended solution when VM distribution across hosts is required.

### DIFF
--- a/docs/user/openstack/known-issues.md
+++ b/docs/user/openstack/known-issues.md
@@ -14,6 +14,19 @@ In order to validate your OpenStack infrastructure prior to installation or upgr
 
 Since the installer requires the *Name* of your external network and Red Hat Core OS image, if you have other networks or images with the same name, it will choose one randomly from the set. This is not a reliable way to run the installer. We highly recommend that you resolve this with your cluster administrator by creating unique names for your resources in openstack.
 
+## Soft-anti-affinity
+
+A long-standing OpenStack Compute issue prevents the soft anti-affinity
+policy to correctly apply when instances are created in parallel. To maximize
+the chances that the default soft-anti-affinity setting applies, the Installer
+generates the Control plane instances sequentially. However, there is no
+guarantee that other machines, and in particular Compute nodes, are created
+sequentially. This applies both to install-time machine building, and day 2
+operations (e.g. MachineSet scale-out).
+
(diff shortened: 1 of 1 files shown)


docs/user/openstack/known-issues.md
```diff
@@ -14,6 +14,13 @@ In order to validate your OpenStack infrastructure prior to installation or upgr
 
 Since the installer requires the *Name* of your external network and Red Hat Core OS image, if you have other networks or images with the same name, it will choose one randomly from the set. This is not a reliable way to run the installer. We highly recommend that you resolve this with your cluster administrator by creating unique names for your resources in openstack.
 
+## Soft-anti-affinity
+
+A long-standing OpenStack Compute issue prevents the **soft anti-affinity**
+policy to be correctly applied when instances are created in parallel. When
+machines must land in distinct Compute hosts, explicitly set
+`serverGroupPolicy` to `anti-affinity` for the machine pool.
```
mandre, line 22: Perhaps add a comment that the IPI installer works around this problem by creating instances sequentially.

pierreprinetti: Creating instances sequentially didn't prove to be effective in 100% of the cases, so I'd prefer not giving that sort of guarantees

mandre: OK, just out of curiosity, do you have specific example in mind where creating the instances sequentially caused them to end up on the same hypervisor?

mandre: Just realized, what you're saying here contradicts the first sentence.

pierreprinetti: I have rephrased. Can you please check that it makes more sense now?

mandre (commented):
/lgtm
/approve

docs/user/openstack/known-issues.md
mandre: Yes, thanks a lot :+1:

PR #6481, opened 2022-10-12
OCPBUGS-2290: Power VS Check for existing DNS permitted network and public gateway

mjturek:
In the private and disconnected scenarios, if you deploy with either an existing public gateway or the VPC already in the permitted networks of the DNS service, terraform will fail. This PR adds logic to create these resources only when needed. Much of the client code is lifted from IBM Cloud's client.go and adapted to work with our minor differences.

### DIFF
--- a/data/data/powervs/cluster/dns/dns.tf
+++ b/data/data/powervs/cluster/dns/dns.tf
@@ -38,7 +38,7 @@ data "ibm_dns_zones" "dns_zones" {
 }
 
 resource "ibm_dns_permitted_network" "permit_vpc_network_for_dns" {
-  count       = var.publish_strategy == "Internal" ? 1 : 0
+  count       = var.publish_strategy == "Internal" && ! var.vpc_permitted ? 1 : 0
   instance_id = var.service_id
   zone_id     = local.dns_zone.zone_id
   vpc_crn     = var.vpc_crn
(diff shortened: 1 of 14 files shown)


deepsm007:
/retitle OCPBUGS-2290: Power VS Check for existing DNS permitted network and public gateway

jhixson74:
/lgtm

This looks good to me. What tests need to pass for this?

mjturek:
@jhixson74 I believe this is ready to merge. It's failing some of the -ovn prow jobs but the failures should be unrelated as all the changes are to Power VS

mjturek:
@rna-afk could you take a look? We need this patch for disconnected cluster support.

clnperez:
/lgtm 
mike and i chatted about my concerns via slack. looks like it's already got everything it needs though so 🎉

clnperez:
@jhixson74 @Prashanth684 this seems not to have auto-merged. is there a way we can still get it into 4.12 or do we need to do a backport now from the 4.13 branch?

Prashanth684:
> @jhixson74 @Prashanth684 this seems not to have auto-merged. is there a way we can still get it into 4.12 or do we need to do a backport now from the 4.13 branch?

given that we are past code freeze, yes this would need a backport after this merges

r4f4:
/override ci/prow/e2e-azure-ovn The few e2e test failures are not affected by the changes in this PR.

PR #6483, opened 2022-10-13
OCPBUGS-2262: Update gcp explain for DNS zones

barbacbd:
** Update the comments/explain docs for the public and private DNS zone in the install config.


r4f4 (commented):
/lgtm

PR #6488, opened 2022-10-16
OCPBUGS-3164: hold bootkube service until bootstrap has pivoted

vrutkovs:
This solves a race between release-image-pivot and bootkube on firstboot. While manifests are being generated pivot may reboot a machine, which may cause corrupted manifests. Instead in OKD bootkube should wait for pivot to be completed.

This service doesn't run on OCP and has no effect in OCP installs


vrutkovs:
`okd-e2e-gcp` passes install phase

vrutkovs:
vsphere-ovn tests failing due to https://issues.redhat.com/browse/OCPBUGS-2927, please override

LorbusChris:
@vrutkovs do we need a `WantedBy` and `RequiredBy` on this too?

PR #6492, opened 2022-10-17
OCPBUGS-1033: support multiple documents in the same extra manifest file

andfasano:
This patch allows to specify multiple documents in the same yaml file when when used as an extra manifest for the agent installer.

### DIFF
--- a/pkg/asset/agent/image/ignition.go
+++ b/pkg/asset/agent/image/ignition.go
@@ -14,6 +14,7 @@ import (
 	igntypes "github.com/coreos/ignition/v2/config/v3_2/types"
 	"github.com/google/uuid"
 	"github.com/sirupsen/logrus"
+	"gopkg.in/yaml.v2"
 
 	hiveext "github.com/openshift/assisted-service/api/hiveextension/v1beta1"
 	"github.com/openshift/assisted-service/api/v1beta1"
@@ -228,7 +229,10 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
 
 	addHostConfig(&config, agentConfigAsset)
 
-	addExtraManifests(&config, extraManifests)
+	err = addExtraManifests(&config, extraManifests)
+	if err != nil {
+		return err
+	}
 
 	a.Config = &config
 	return nil
@@ -361,7 +365,7 @@ func addHostConfig(config *igntypes.Config, agentConfig *agentconfig.AgentConfig
 	return nil
 }
 
-func addExtraManifests(config *igntypes.Config, extraManifests *manifests.ExtraManifests) {
+func addExtraManifests(config *igntypes.Config, extraManifests *manifests.ExtraManifests) error {
 
 	user := "root"
 	mode := 0644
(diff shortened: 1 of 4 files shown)


pkg/asset/agent/image/ignition.go
```diff
@@ -380,9 +386,38 @@ func addExtraManifests(config *igntypes.Config, extraManifests *manifests.ExtraM
...
 	for _, file := range extraManifests.FileList {
-		extraFile := ignition.FileFromBytes(filepath.Join(extraManifestPath, filepath.Base(file.Filename)), user, mode, file.Data)
-		config.Storage.Files = append(config.Storage.Files, extraFile)
+
+		r := bytes.NewReader(file.Data)
+		decoder := yaml.NewDecoder(r)
+
+		var manifest map[string]interface{}
+		for n := 0; ; n++ {
+
```
celebdor, line 395: Why the empty line here?

pkg/asset/agent/image/ignition.go
```diff
@@ -380,9 +386,38 @@ func addExtraManifests(config *igntypes.Config, extraManifests *manifests.ExtraM
...
+			err := decoder.Decode(&manifest)
+			if err != nil {
+				if err != io.EOF {
+					return err
+				}
+				return nil
```
celebdor, line 401: Shouldn't this be a break instead?

pkg/asset/agent/image/ignition.go
andfasano: Exit condition(s) are defined below (either the decoder stops finding another document, or an error during decoding/marshaling happens)

pkg/asset/agent/image/ignition_test.go
```diff
@@ -296,6 +296,37 @@ func TestAddHostConfig_Roles(t *testing.T) {
 
 }
 
+func defaultGeneratedFiles() []string {
```
zaneb, line 299: Will we have to update this list every time we add a file to the image?

I think we should only test for files in the extra-manifests directory so that we can avoid that.

zaneb (commented):
/approve

pkg/asset/agent/image/ignition_test.go
andfasano: TBH I did it intentionally, as I noticed that we didn't have any coverage on the files we're going to add in the ISO (the "core" business for this assets) to catch any regression, which is critical point.
In this way, if in future someone will remove one portion of the code related to one or more files, then at least it would be signaled by a failing test (even better to check the content, but that would require more effort)

pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -287,37 +286,27 @@ func GetNMIgnitionFiles(staticNetworkConfig []*models.HostStaticNetworkConfig) (
...
-	err := yamlDecoder.Decode(&decodedData)
-
-	return decodedData, err
-}
-
-// Read a YAML file containing multiple YAML definitions of the same format
+// GetMultipleYamls reads a YAML file containing multiple YAML definitions of the same format
 // Each specific format must be of type DecodeFormat
-func getMultipleYamls(contents []byte, decoder decodeFormat) ([]interface{}, error) {
+func GetMultipleYamls[T any](contents []byte) ([]T, error) {
```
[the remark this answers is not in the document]
andfasano: That's a fair point, I was postponing such refactoring to see if maybe some other asset may needed it, we could follow up on that on the next occasion

lranjbar:
Created OCPBUGS-3035 for the 4.12 backport.

PR #6495, opened 2022-10-19
OCPBUGS-2627: agent-config: Generate missing AdditionalNTPSources in InfraEnv

celebdor:
In Agent based installation ZTP input, one can provide AdditionalNTPSources in InfraEnv. The sources configured there are merged by assisted-service with those that are provided by DHCP option 42 (According to RFC 2132) and provided to the openshift-install that assisted-service runs as MachineConfigs for the different roles defined for the cluster nodes.

This commit closes the gap we had when installing from install-config + agent-config.

### DIFF
--- a/pkg/asset/agent/manifests/infraenv.go
+++ b/pkg/asset/agent/manifests/infraenv.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/openshift/installer/pkg/asset"
 	"github.com/openshift/installer/pkg/asset/agent"
+	"github.com/openshift/installer/pkg/asset/agent/agentconfig"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


pkg/asset/agent/manifests/infraenv.go
```diff
@@ -70,6 +73,10 @@ func (i *InfraEnv) Generate(dependencies asset.Parents) error {
 		if installConfig.Config.Proxy != nil {
 			infraEnv.Spec.Proxy = getProxy(installConfig)
 		}
+
+		if agentConfig.Config != nil {
+			infraEnv.Spec.AdditionalNTPSources = agentConfig.Config.AdditionalNTPSources
```
andfasano, line 78: Do you think that any additional validation (on the content of `AdditionalNTPSources`) may be required/useful?

celebdor:
/retitle OCPBUGS-2627: agent-config: Generate missing AdditionalNTPSources in InfraEnv

PR #6496, opened 2022-10-19
WIP Automate Power VS service instance creation

AshwinHIBM:
The installer should create a Power VS service instance in case the service instance id is unspecified by the user while running the installer.


data/data/powervs/iaas/variables.tf
```diff
@@ -1,29 +1,11 @@
...
-  description = "The ID created by the installer to uniquely identify the created cluster."
-  default     = ""
-}
-
 variable "service_tags" {
   type        = list(string)
   description = "A list of tags for our resource instance."
   default     = []
 }
+
```
clnperez, line 6: actually we don't need this entire file. you can just add cloud_instance_id (your output var) to the data/data/powervs/variables-powervs.tf and it will be inherited by the other stages.

data/data/powervs/iaas/variables.tf
```diff
@@ -1,29 +1,11 @@
...
+variable "service_instance_id" {
```
clnperez, line 7: instead of this, you can use the inherited powervs_cloud_instance_id (in powervs-variables.tf)

data/data/powervs/cluster/variables.tf
```diff
@@ -0,0 +1,4 @@
+variable "cloud_instance_id" {
```
clnperez, line 1: this can be deleted and just use the output from the iaas stage after you put it into data/data/powervs/variables-powervs.tf

clnperez (changes requested):
some suggestions but also check out https://github.com/clnperez/installer/commit/d022af283666db50905ad2f9b3d9d811953dc76e for examples and some others

PR #6500, opened 2022-10-19
Agent-321: Cluster and Host validations rework

lranjbar:
This rework is to change the output for the cluster and host validations.

After querying the API, for each validation ID check the message against the most recent message received for that ID:

- If the message is unchanged, do nothing
- If the new message is a failure, log it at warning level
- If the new message is a success but there has previously been a failure, log it at info level
- If the new message is a success and this is the first report for that validation, log it at debug level

We won't log anything at error level because no validation in assisted prior to the install starting is ever final. The only error comes when we hit our timeout. 

To help with review [the struct coming from the Assisted API](https://github.com/openshift/assisted-service/blob/master/api/common/common_types.go#L5-L10) is the following:
```
type ValidationResult struct {
	ID      string `json:"id"`
	Status  string `json:"status"`
	Message string `json:"message"`
}
```
An example of this looks like:
```
{
    "id": "pull-secret-set",
    "status": "success",
    "message": "The pull secret is set."
}
```
[AGENT-321](https://issues.redhat.com//browse/AGENT-321)

### DIFF
--- a/pkg/agent/cluster.go
+++ b/pkg/agent/cluster.go
@@ -55,6 +55,7 @@ type clusterInstallStatusHistory struct {
 	ClusterConsoleRouteURLCreated                       bool
 	ClusterInstallComplete                              bool
 	NotReadyTime                                        time.Time
+	ValidationResults                                   *validationResults
 }
 
 // NewCluster initializes a Cluster object
@@ -105,6 +106,11 @@ func NewCluster(ctx context.Context, assetDir string) (*Cluster, error) {
 		ClusterInstallComplete:                              false,
 	}
 
+	cvalidationresults := &validationResults{
+		ClusterValidationHistory: make(map[string]*validationResultHistory),
+		HostValidationHistory:    make(map[string]map[string]*validationResultHistory),
+	}
+
 	czero.Ctx = ctx
 	czero.API = capi
 	czero.clusterID = nil
@@ -112,6 +118,7 @@ func NewCluster(ctx context.Context, assetDir string) (*Cluster, error) {
 	czero.assetDir = assetDir
 	czero.clusterConsoleRouteURL = ""
 	czero.installHistory = cinstallstatushistory
+	czero.installHistory.ValidationResults = cvalidationresults
 	return czero, nil
 }
 
@@ -214,8 +221,10 @@ func (czero *Cluster) IsBootstrapComplete() (bool, bool, error) {
 			}
 		}
 
-		if !checkHostsValidations(clusterMetadata, logrus.StandardLogger()) {
-			return false, false, nil
+		validationsErr := checkValidations(clusterMetadata, czero.installHistory.ValidationResults, logrus.StandardLogger())
+		if validationsErr != nil {
+			return false, false, errors.Wrap(validationsErr, "cluster host validations failed")
+
 		}
 
 		// Print most recent event associated with the clusterInfraEnvID
@@ -239,25 +248,24 @@ func (czero *Cluster) IsBootstrapComplete() (bool, bool, error) {
 			czero.installHistory.RestAPIInfraEnvEventList = eventList
 		}
 
-	}
+		// both API's are not available
+		if !agentRestAPILive && !clusterKubeAPILive {
+			logrus.Trace("Current API Status: Node Zero Agent API: down, Cluster Kube API: down")
+			if !czero.installHistory.RestAPISeen && !czero.installHistory.ClusterKubeAPISeen {
+				logrus.Debug("Node zero Agent Rest API never initialized. Cluster API never initialized")
+				logrus.Info("Waiting for cluster install to initialize. Sleeping for 30 seconds")
+				time.Sleep(30 * time.Second)
+				return false, false, nil
+			}
 
-	// both API's are not available
-	if !agentRestAPILive && !clusterKubeAPILive {
-		logrus.Trace("Current API Status: Node Zero Agent API: down, Cluster Kube API: down")
-		if !czero.installHistory.RestAPISeen && !czero.installHistory.ClusterKubeAPISeen {
-			logrus.Debug("Node zero Agent Rest API never initialized. Cluster API never initialized")
-			logrus.Info("Waiting for cluster install to initialize. Sleeping for 30 seconds")
-			time.Sleep(30 * time.Second)
-			return false, false, nil
+			if czero.installHistory.RestAPISeen && !czero.installHistory.ClusterKubeAPISeen {
+				logrus.Debug("Cluster API never initialized")
+				logrus.Debugf("Cluster install status from Agent Rest API last seen was: %s", czero.installHistory.RestAPIPreviousClusterStatus)
+				return false, false, errors.New("cluster bootstrap did not complete")
+			}
 		}
 
-		if czero.installHistory.RestAPISeen && !czero.installHistory.ClusterKubeAPISeen {
-			logrus.Debug("Cluster API never initialized")
-			logrus.Debugf("Cluster install status from Agent Rest API last seen was: %s", czero.installHistory.RestAPIPreviousClusterStatus)
-			return false, false, errors.New("cluster bootstrap did not complete")
-		}
 	}
-
 	logrus.Trace("cluster bootstrap is not complete")
 	return false, false, nil
 }
--- a/pkg/agent/kube.go
+++ b/pkg/agent/kube.go
@@ -52,7 +52,7 @@ func (kube *ClusterKubeAPIClient) IsKubeAPILive() (bool, error) {
 	if err != nil {
 		return false, err
 	}
-	logrus.Debugf("cluster API is up and running %s", version)
+	logrus.Tracef("cluster API is up and running %s", version)
 	return true, nil
 }
 
@@ -73,7 +73,8 @@ func (kube *ClusterKubeAPIClient) IsBootstrapConfigMapComplete() (bool, error) {
 	bootstrap, err := kube.Client.CoreV1().ConfigMaps("kube-system").Get(kube.ctx, "bootstrap", v1.GetOptions{})
 
 	if err != nil {
-		return false, errors.Wrap(err, "bootstrap configmap not found")
+		logrus.Trace(errors.Wrap(err, "bootstrap configmap not found"))
+		return false, nil
 	}
 	// Found a bootstrap configmap need to check its status
 	if bootstrap != nil {
--- a/pkg/agent/validations.go
+++ b/pkg/agent/validations.go
@@ -2,138 +2,112 @@ package agent
 
 import (
 	"encoding/json"
-	"fmt"
-	"reflect"
-	"sort"
 
 	"github.com/openshift/assisted-service/api/common"
 	"github.com/openshift/assisted-service/models"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
+// These validation status strings are defined to match the ones from assisted-service/internal.
+// These statuses are the statuses that mattered to us for this implementation.
+// Cluster: https://github.com/openshift/assisted-service/blob/master/internal/cluster/validator.go
+// Host: https://github.com/openshift/assisted-service/blob/master/internal/host/validator.go
 const (
 	validationFailure string = "failure"
 	validationError   string = "error"
+	validationSuccess string = "success"
 )
 
-// Re-using Assisted UI host validation labels (see https://github.com/openshift-assisted/assisted-ui-lib)
-// for logging human-friendly messages in case of validation failures
-var hostValidationLabels = map[string]string{
-	"odf-requirements-satisfied":                      "ODF requirements",
-	"disk-encryption-requirements-satisfied":          "Disk encryption requirements",
-	"compatible-with-cluster-platform":                "",
-	"has-default-route":                               "Default route to host",
-	"sufficient-network-latency-requirement-for-role": "Network latency",
-	"sufficient-packet-loss-requirement-for-role":     "Packet loss",
-	"has-inventory":                                   "Hardware information",
-	"has-min-cpu-cores":                               "Minimum CPU cores",
-	"has-min-memory":                                  "Minimum Memory",
-	"has-min-valid-disks":                             "Minimum disks of required size",
-	"has-cpu-cores-for-role":                          "Minimum CPU cores for selected role",
-	"has-memory-for-role":                             "Minimum memory for selected role",
-	"hostname-unique":                                 "Unique hostname",
-	"hostname-valid":                                  "Valid hostname",
-	"connected":                                       "Connected",
-	"media-connected":                                 "Media Connected",
-	"machine-cidr-defined":                            "Machine CIDR",
-	"belongs-to-machine-cidr":                         "Belongs to machine CIDR",
-	"ignition-downloadable":                           "Ignition file downloadable",
-	"belongs-to-majority-group":                       "Belongs to majority connected group",
-	"valid-platform-network-settings":                 "Platform network settings",
-	"ntp-synced":                                      "NTP synchronization",
-	"container-images-available":                      "Container images availability",
-	"lso-requirements-satisfied":                      "LSO requirements",
-	"ocs-requirements-satisfied":                      "OCS requirements",
-	"sufficient-installation-disk-speed":              "Installation disk speed",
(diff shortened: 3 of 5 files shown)


lranjbar:
Added unit tests and remove WIP label.

pkg/agent/validations.go
```diff
@@ -63,6 +62,129 @@ type validationTrace struct {
...
+			numFailures:     make(map[string]int),
+			seen:            make(map[string]bool),
+			currentStatus:   make(map[string]string),
+			currentMessage:  make(map[string]string),
+			previousStatus:  make(map[string]string),
+			previousMessage: make(map[string]string),
+		}
+	}
+	clusterLogPrefix := "Cluster validation: "
+	clusterValidationHistory, err := updateValidationsResultHistory(clusterLogPrefix, cluster.ValidationsInfo, validationHistory["cluster"], log)
```
andfasano, line 86: As an opportunity to simplify the code, the function signature could be modified to 

`func updateValidationsResultHistory(logPrefix string, validationsInfoString string, historyMap map[string]*validationResultHistory, key string, log *logrus.Logger) error {`

so that the content of the map could  be update directly within the function (and the element created if it doesn't exist)

pkg/agent/validations.go
```diff
@@ -63,6 +62,129 @@ type validationTrace struct {
...
+	numFailures     map[string]int
+	seen            map[string]bool
+	currentStatus   map[string]string
+	currentMessage  map[string]string
+	previousStatus  map[string]string
+	previousMessage map[string]string
+}
+
+func checkValidations(cluster *models.Cluster, validationHistory map[string]*validationResultHistory, log *logrus.Logger) error {
+	if validationHistory["cluster"] == nil {
```
andfasano, line 75: nit: for readability, I'd prefer the `comma ok` idiom, ie `if _, ok := validationHistory["cluster"]; !ok {` for checking the existence of the element in the map

pkg/agent/validations.go
```diff
@@ -63,6 +62,129 @@ type validationTrace struct {
```
andfasano, line 62: if the `validationTrace` is not used anymore then it could be removed

pkg/agent/validations.go
```diff
@@ -63,6 +62,129 @@ type validationTrace struct {
...
+			log.Warning(logPrefix + history.currentMessage[id])
+			loglevel = "warning"
+			history.seen[id] = true
+		case validationSuccess:
+			log.Debug(logPrefix + history.currentMessage[id])
+			loglevel = "debug"
+			history.seen[id] = true
+		}
+
+	} else {
```
andfasano, line 169: For the happy path a return could be used here

pkg/agent/validations.go
```diff
@@ -63,6 +62,129 @@ type validationTrace struct {
...
+				}
+			}
+
+			if _, ok := history.previousMessage[r.ID]; !ok {
+				history.previousMessage[r.ID] = ""
+				history.currentMessage[r.ID] = r.Message
+			} else {
+				history.previousMessage[r.ID] = history.currentMessage[r.ID]
+				history.currentMessage[r.ID] = r.Message
+			}
```
andfasano, line 146: I think that `previousMessage` and `previousStatus` could be unconditionally updated if moved after ` logValidationResultHistory`

andfasano:
/approve

The overall logic and approach seems fine, I left just some comments for cleaning up / simplifying the code

pkg/agent/validations_test.go
```diff
@@ -2,114 +2,271 @@ package agent
...
+	validationsInfoSuccess string = "{\"test\":[{\"id\":\"test\",\"status\":\"success\",\"message\":\"The validation succeeded\"}]}"
+	validationsInfoFailure string = "{\"test\":[{\"id\":\"test\",\"status\":\"failure\",\"message\":\"The validation failed\"}]}"
+)
+
+func TestUpdateValidationHistory(t *testing.T) {
 	tests := []struct {
-		name           string
-		hosts          []*models.Host
-		expectedResult bool
-		expectedLogs   []string
```
andfasano, line 19: Would be possible to keep the `expectedLogs` field (if not here, at leasr in another test)? Its main intent was to capture the output seen by the user under different conditions, and I think it would be useful to keep it (also as a form of internal dev documentation)

pkg/agent/validations.go
```diff
@@ -2,138 +2,125 @@ package agent
...
-			return true
+func logValidationHistory(logPrefix string, history *validationResultHistory, log *logrus.Logger) string {
+	loglevel := ""
+	// Fist time we print anything
+	if !history.seen {
+		switch history.currentStatus {
+		case validationFailure, validationError:
+			log.Warning(logPrefix + history.currentMessage)
+			loglevel = "warning"
+			history.seen = true
```
bfournie, line 101: Nit - could set 'seen' before the switch and remove set at line 101 and 105

pkg/agent/validations_test.go
lranjbar: This logic actually wasn't checking for the output of the log clearly. It was just checking if any output log output on the desired log level. I spent a couple of hours fixing this test when I switched to using Debug over Error and it wasn't clear as to why. I disagree that it was of any use.

pkg/agent/validations.go
```diff
@@ -2,138 +2,125 @@ package agent
...
+				log.Info(logPrefix + history.currentMessage)
+				loglevel = "info"
+			}
+		case validationSuccess:
+			if history.currentStatus == validationError || history.currentStatus == validationFailure {
+				log.Warning(logPrefix + history.currentMessage)
+				loglevel = "warning"
+			}
 		}
+		return loglevel
```
bfournie, line 123: Nit - could remove as it falls through to return at line 125

pkg/agent/validations.go
lranjbar: I think it makes more sense to be after the switch... so I moved it there.

pkg/agent/validations.go
```diff
@@ -2,138 +2,123 @@ package agent
...
-
-		if len(previousValidations) == 0 {
-			log.Info("Pre-installation validations are OK")
-			return true
+func logValidationHistory(logPrefix string, history *validationResultHistory, log *logrus.Logger) string {
+	loglevel := ""
+	// Fist time we print anything
+	if !history.seen {
+		switch history.currentStatus {
+		case validationFailure, validationError:
```
zaneb, line 98: This is a duplicate of the code below at line 111. That could be eliminated by only returning early in the success case:

```
if !history.seen {
    history.seen = true
    if history.currentStatus = validationSuccess {
        log.Debug(...)
        return
    }
}
```

pkg/agent/validations.go
```diff
@@ -2,138 +2,123 @@ package agent
...
+	}
 
-		log.Debug("Checking for validation failures ----------------------------------------------")
-		for _, v := range previousValidations {
-			log.WithFields(logrus.Fields{
-				"category": v.category,
-				"label":    v.label,
-				"message":  v.message,
-			}).Debug(v.header)
+	if history.currentStatus != history.previousStatus {
```
zaneb, line 109: This check means that if the message changes but the status stays the same, we won't log anything. I'm not sure if this is a problem in practice, but if it happens I would have assumed we'd want to print the new message again. The algorithm I had in mind was quite simple:

```
if history.currentMessage != history.PreviousMessage {
    switch history.currentStatus {
    case validationFailure, validationError:
        log.Warning(...)
    case validationSuccess {
        if history.PreviousMessage != "" {
            log.Info(...)
        } else {
            log.Debug(...)
        }
    }
}
```

pkg/agent/cluster.go
```diff
@@ -54,6 +54,7 @@ type clusterInstallStatusHistory struct {
 	ClusterConsoleRouteCreated                          bool
 	ClusterConsoleRouteURLCreated                       bool
 	ClusterInstallComplete                              bool
+	ValidationHistory                                   map[string]map[string]*validationResultHistory
```
zaneb, line 57: I'd be tempted to also define a

```
type ValidationResults map[string]validationResultHistory
```

so that this can be a `map[string]ValidationResults`, and you can also use it when initialising the per-host result maps.

pkg/agent/validations.go
```diff
@@ -2,138 +2,123 @@ package agent
...
+					numFailures:     0,
+					seen:            false,
+					currentStatus:   "",
+					currentMessage:  "",
+					previousStatus:  "",
+					previousMessage: "",
+				}
+			}
+			// First time we see a validation ID
+			if !validationHistory[r.ID].seen {
```
zaneb, line 74: This check is unnecessary, we can execute the 'else' block unconditionally.

pkg/agent/validations_test.go
zaneb: It looks to me like the code on lines 101-112 was checking that there was at least one log message matching each of the `expectedLogs` (as a regex)?

pkg/agent/validations_test.go
```diff
@@ -2,114 +2,271 @@ package agent
...
+				currentMessage:  "The validation failed",
+				previousStatus:  "success",
+				previousMessage: "The validation succeeded",
+			},
+			expectedResult: "warning",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, tt.expectedResult, logValidationHistory(testLogPrefix, tt.inputHistory, logger))
```
zaneb, line 269: If we used the logrus hooks as in the previous test code, we could do `hook.LastEntry().Level` to get the level instead of having to return it as a string from the function for test purposes and hope they stay in sync.

pkg/agent/validations_test.go
lranjbar: It matches the regex thats true but what is the value of testing the matching regex? Are we writing a unit test for regex or our code?

zaneb: Ideally when we're checking the level we should also test that we're checking the level _of the right message_, so that we can't accidentally lose test coverage by adding another log. Probably no need to have an `expectedLogs` field though, since we know it should always report the message.

pkg/agent/validations.go
```diff
@@ -2,138 +2,123 @@ package agent
...
 
 	"github.com/openshift/assisted-service/api/common"
 	"github.com/openshift/assisted-service/models"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
 const (
 	validationFailure string = "failure"
 	validationError   string = "error"
```
zaneb, line 14: What is the difference between 'failure' and 'error'? Is one of these actually terminal?

pkg/agent/cluster.go
```diff
@@ -186,8 +188,9 @@ func (czero *Cluster) IsBootstrapComplete() (bool, error) {
 			return false, errors.New("cluster metadata returned nil from Agent Rest API")
 		}
 
-		if !checkHostsValidations(clusterMetadata, logrus.StandardLogger()) {
-			return false, errors.New("cluster host validations failed")
+		validationsErr := checkValidations(clusterMetadata, czero.installHistory.ValidationHistory, logrus.StandardLogger())
+		if validationsErr != nil {
+			return false, errors.Wrap(validationsErr, "cluster host validations failed")
```
zaneb, line 193: Previously `checkHostsValidations()` returned false (and we bailed out of this function) if _any_ validations were not passing. In practice I suspect this may have meant that we rarely get past this point, since _something_ is usually complaining.
With this change, we no longer return here unless there is an actual error. That's probably the right thing to do, but it results in dozens of warning messages in the logs saying `Cluster has stopped installing... working to recover installation` (from line 204).
Not sure of the right thing to do. It might be to just not log that warning.

pkg/agent/validations.go
```diff
@@ -2,138 +2,123 @@ package agent
...
+			// If validation ID does not exist create it
+			if _, ok := validationHistory[r.ID]; !ok {
+				validationHistory[r.ID] = &validationResultHistory{
+					numFailures:     0,
+					seen:            false,
+					currentStatus:   "",
+					currentMessage:  "",
+					previousStatus:  "",
+					previousMessage: "",
+				}
```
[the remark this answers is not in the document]
lranjbar: Once you place a struct in a map it becomes unaddressable. I found using pointers was cleaner than the other option of making temporary structs to update the map.

The error I initally ran into was like `cannot assign to struct field validationHistory[r.ID].currentMessage in map`

Without using pointers I had to keep making structs...and it just seemed a more sufficient solution to use pointers.

pkg/agent/validations.go
lranjbar: Changed to look for difference in message vs status

pkg/agent/validations.go
lranjbar: There is actually a [decent amount of statuses in assisted-service](https://github.com/openshift/assisted-service/blob/dac620ca40898913e5c847424aec5dc1b84d1a78/internal/host/validator.go#L40-L45) we don't act on. Does that change the logic here?

pkg/agent/validations_test.go
lranjbar: I did some refactoring on `TestLogValidationResultHistory()` that should address all these comments.

pkg/agent/validations.go
lranjbar: Digging through the assisted service code its the different between incorrect input and a failing validation.

Here is the code I reviewed to come to that conclusion: https://github.com/openshift/assisted-service/blob/master/internal/host/validator.go

lranjbar: As far as terminal.. it doesn't seem so.

pkg/agent/validations.go
```diff
@@ -2,138 +2,123 @@ package agent
...
+			}
+			// First time we see a validation ID
+			if !validationHistory[r.ID].seen {
+				validationHistory[r.ID].currentMessage = r.Message
+				validationHistory[r.ID].currentStatus = r.Status
+			} else {
+				validationHistory[r.ID].previousMessage = validationHistory[r.ID].currentMessage
```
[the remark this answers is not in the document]
lranjbar: Sure we could... but it makes it harder to test the log output. Take a look at the unit test refactor.

pkg/agent/cluster.go
lranjbar: Perhaps we don't need this message anymore now that we are printing the validations.

pkg/agent/validations.go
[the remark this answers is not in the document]
zaneb: Ah, true, you'd have to copy it out and in again. I don't think I'd encountered that before. Thanks!

pkg/agent/cluster.go
```diff
@@ -111,6 +112,8 @@ func NewCluster(ctx context.Context, assetDir string) (*Cluster, error) {
 	czero.assetDir = assetDir
 	czero.clusterConsoleRouteURL = ""
 	czero.installHistory = cinstallstatushistory
+	czero.installHistory.ValidationResults.ClusterValidationHistory = make(map[string]*validationResultHistory)
```
zaneb, line 115: Tests are failing with a nil pointer here, presumably because ValidationResults is not initialised.

lranjbar: Yep good catch

pkg/agent/kube.go
```diff
@@ -73,7 +73,8 @@ func (kube *ClusterKubeAPIClient) IsBootstrapConfigMapComplete() (bool, error) {
 	bootstrap, err := kube.Client.CoreV1().ConfigMaps("kube-system").Get(kube.ctx, "bootstrap", v1.GetOptions{})
 
 	if err != nil {
-		return false, errors.Wrap(err, "bootstrap configmap not found")
+		logrus.Tracef("bootstrap configmap not found: s%", err)
```
bfournie, line 76: Looks like the s and % need to be reversed, seeing this in unit tests:
"pkg/agent/kube.go:76:3: github.com/sirupsen/logrus.Tracef format % is missing verb at end of string"

lranjbar: Doh!

zaneb:
The `cluster has stopped installing... working to recover installation` message is back, but that's because #6470 moved that section above the call to `checkHostValidations()` (and changed the messages around). Unfortunately I failed to notice the significance of that change or check the output when merging that PR. We'll want to resolve that, but since the issue already exists independently of this PR, there's no reason to hold this up.
/lgtm

PR #6505, opened 2022-10-20
CORS-2147: Add documentation for the client certs authentication

rna-afk:
Added documentation for Azure client certificate authentication.

### DIFF
--- a/docs/dev/azure/azure_client_certs_auth.md
+++ b/docs/dev/azure/azure_client_certs_auth.md
@@ -1,0 +1,60 @@
+# Azure Authentication using Client certificates
+
+Azure accepts client certificates as a means of authentication in the service principal and terraform accepts it too. 
+As of 4.12, the Installer accepts certificate-based service principals in addition to secret-based service principals.
+
+### Pitfalls
+Although the installer can now use the certs to authenticate, CCO does not support this and hence the installer
+should create the cluster in manual credentials mode only.
+
+### Prerequisites
+- A certificate that is suitable for Azure is created. 
+- Register the certificate with Azure as part of App Registrations in the Azure AD.
+More information on how to do these steps is below.
+
+## Steps
+The installer takes the service principal for authentication and the current requirement is that these fields are
+populated.
+
+After [1], we no longer need to pass the client secret and can pass the clientCertificate and clientCertificatePassword fields.
+
+1. Populate the service principal with the following fields. The service principal is by default in ~/.azure/osServicePrincipal.json
+-- subscriptionId
+-- tenantId
+-- clientId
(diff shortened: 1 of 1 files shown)


docs/dev/azure/azure_client_certs_auth.md
```diff
@@ -0,0 +1,37 @@
+# Azure Authentication using Client certificates
+
+Azure accepts client certificates as a means of authentication in the service principal and terraform accepts it too. 
+The installer now supports authentication via the certs after [1] merged.
```
patrickdillon, line 4:
```suggestion
As of 4.12, the Installer accepts certificate-based service principals in addition to secret-based service principals.
```

docs/dev/azure/azure_client_certs_auth.md
```diff
@@ -0,0 +1,37 @@
...
+
+Azure accepts client certificates as a means of authentication in the service principal and terraform accepts it too. 
+The installer now supports authentication via the certs after [1] merged. 
+
+### Pitfalls
+Although the installer can now use the certs to authenticate, CCO does not support this and hence the installer
+should create the cluster in manual credentials mode only.
+
+### Prerequisites
+- A certificate that is suitable for Azure is created.
```
patrickdillon, line 11: I think this would be the helpful part to explain. Is there an easy way to do this?

For example, can we use `az ad sp create-for-rbac --create-cert` to achieve this? If so it would be helpful to include what is the easy way to get started doing this.

docs/dev/azure/azure_client_certs_auth.md
```diff
@@ -0,0 +1,37 @@
...
+- Register the certificate with Azure as part of App Registrations in the Azure AD. (Refer [2] for more details)
+
+## Steps
+The installer takes the service principal for authentication and the current requirement is that these fields are
+populated.
+
+- Tenant ID
+- Client ID
+- Client Secret
+- Subscription ID
```
patrickdillon, line 21: I don't think we need to include this. The list of what's required for the cert-based SP should suffice.

docs/dev/azure/azure_client_certs_auth.md
```diff
@@ -0,0 +1,37 @@
...
+
+After [1], we no longer need to pass the client secret and can pass the clientCertificate and clientCertificatePassword fields.
+
+1. Populate the service principal with the following fields.
```
patrickdillon, line 25: It would be good to indicate where the creds live (`~/.azure/osServicePrincipal.json`).

patrickdillon (commented):
A few suggestions. The main one is we should have sort of quick start steps which would quickly enable someone not familiar with this to produce their own cert service principal.

PR #6509, opened 2022-10-20
OCPBUGS-2124: network discovery,cluster path issue 

jcpowermac:
- Added a check in tfvars if zonal then
don't run legacy section to grab the network mob id

- Added a check in GetClusterNetworks if cluster is a path.
If it use the cluster string directly, if not
generate the path from the datacenter and cluster name.


jcpowermac:
/assign @sadasu @patrickdillon
I think the owners file needs to be fixed, can one of you approve? Thanks!

PR #6512, opened 2022-10-21
[WIP] prototype: enable static IPs for bootstrap and control plane nodes

rvanderp3:


data/data/vsphere/bootstrap/main.tf
```diff
@@ -41,7 +41,13 @@ resource "vsphere_virtual_machine" "vm_bootstrap" {
     template_uuid = var.template[0].uuid
   }
 
-  extra_config = {
+  extra_config = length(var.vsphere_bootstrap_network_kargs) > 0 ? {
```
jcpowermac, line 44: I don't _think_ we need this branch

https://stackoverflow.com/questions/71157611/making-an-argument-optional-in-terraform-resource-block

Based on ^ it seems if its null then it won't be set.

PR #6515, opened 2022-10-24
Bug 2070744: Fix infinite loop error

rna-afk:
There is an infinite loop if the resource is not untagged during AWS destroy. This needs to be checked for an InvalidParameterException and if it is, we should not retry the same client again.

### DIFF
--- a/pkg/destroy/aws/shared.go
+++ b/pkg/destroy/aws/shared.go
@@ -108,6 +108,8 @@ func (o *ClusterUninstaller) removeSharedTag(ctx context.Context, session *sessi
 				o.Logger.Debugf("No matches in %s for %s: shared, removing client", *tagClient.Config.Region, key)
 				continue
 			}
+			// appending the tag client here but it needs to be removed if there is a InvalidParameterException when trying to
+			// untag below since that only leads to an infinite loop error.
 			nextTagClients = append(nextTagClients, tagClient)
 
 			for i := 0; i < len(arns); i += 20 {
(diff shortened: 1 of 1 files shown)


pkg/destroy/aws/shared.go
```diff
@@ -117,6 +117,12 @@ func (o *ClusterUninstaller) removeSharedTag(ctx context.Context, session *sessi
 				}
 				result, err := tagClient.UntagResourcesWithContext(ctx, request)
 				if err != nil {
+					if aerr, ok := err.(awserr.Error); ok {
```
r4f4, line 120: We should probably be using `errors.As` so typecasting can work even with wrapped errors.

pkg/destroy/aws/shared.go
```diff
@@ -117,6 +117,12 @@ func (o *ClusterUninstaller) removeSharedTag(ctx context.Context, session *sessi
...
+						switch aerr.Code() {
+						case resourcegroupstaggingapi.ErrorCodeInvalidParameterException:
+							nextTagClients = nextTagClients[:len(nextTagClients)-1]
```
patrickdillon, line 123: I'm not following this line. Why do this? Can you add a comment in the code?

rna-afk: turns out I was right the first time. tagClient is added to the last of the list and has to be removed if IPE is returned.

r4f4 (commented):
/lgtm

PR #6516, opened 2022-10-24
OCPBUGS-1704: gcp: fail during validation if service usage is not enabled

r4f4:
We were only printing a warning in case Service Usage was missing and proceeding with the installation. However, that service is required by the CCO and the deployment would end up failing later on.

Instead, let's fail during install-config validation time so users can enabled the service before proceeding.


r4f4:
/override ci/prow/e2e-gcp-ovn
Just some e2e failures and the PR's functionality has been pre-merge tested by QE.

PR #6518, opened 2022-10-24
OCPBUGS-2743: skip sts credentials check in terraform

patrickdillon:
This is a potential fix for OCPBUGS-2743. I'm not 100% certain this is what we want to do, but I think it is good to put it up for discussion. 

When specifying an STS endpoint in the install config, the terraform provider assumes we are also using sts credentials, and the check fails validation:

level=error msg=Error: error configuring Terraform AWS Provider: error validating provider credentials: error calling sts:GetCallerIdentity: SignatureDoesNotMatch: Credential should be scoped to a valid region.

This skips the validation check, to allow us to use non-STS credentials.

The only downside I can see for this is if this would make failures for invalid STS creds less user-friendly. Considering that we don't document how to use STS creds for the installer, this seems like a small issue. Nevertheless, it would be good to test this with both valid and invalid STS credentials.

Also, I'm not completely up-to-date on the state of custom endpoints, but my understanding is that we may not even need to include custom endpoints, at least for STS any more.

### DIFF
--- a/data/data/aws/bootstrap/main.tf
+++ b/data/data/aws/bootstrap/main.tf
@@ -16,7 +16,8 @@ locals {
 provider "aws" {
   region = var.aws_region
 
-  skip_region_validation = true
+  skip_region_validation      = true
+  skip_credentials_validation = true
 
   endpoints {
     ec2     = lookup(var.custom_endpoints, "ec2", null)
--- a/data/data/aws/cluster/main.tf
+++ b/data/data/aws/cluster/main.tf
@@ -11,8 +11,8 @@ locals {
 provider "aws" {
   region = var.aws_region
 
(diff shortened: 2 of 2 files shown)


patrickdillon:
This issue only appears in 4.10 so I will have to remember/figure out how to merge directly to 4.10 without fixing in later versions. Probably just bug scott.

sdodson:
> Probably just bug scott.

Sorry if I missed you bugging me somewhere. Based on the bug reported I see QE mentioning that they're already confirming that this bug is only present in 4.10, therefore i've overridden the valud-bug label. You should be able to do the rest to get this merged.

r4f4:
In case we still want to get this in, LGTM except for the tf formatting issues.

PR #6519, opened 2022-10-24
OCPBUGS-2130: import ova resource cluster path fix

jcpowermac:
Zonal uses the full absolute path for cluster
in the OVA import. In the existing terraform
the relative path is used.

If `cluster` is a path use that string
otherwise generate it from the datacenter and cluster name (which was the existing method)


jcpowermac:
/assign @sadasu @patrickdillon
I think the owners file needs to be fixed, can one of you approve? Thanks!

PR #6520, opened 2022-10-25
OCPBUGS-2837: Reduce debug logs in agent wait-for commands

zaneb:


pkg/agent/kube.go
```diff
@@ -72,7 +73,7 @@ func (kube *ClusterKubeAPIClient) IsBootstrapConfigMapComplete() (bool, error) {
 	// Get latest version of bootstrap configmap
 	bootstrap, err := kube.Client.CoreV1().ConfigMaps("kube-system").Get(kube.ctx, "bootstrap", v1.GetOptions{})
 
-	if err != nil {
+	if err != nil && !k8serrors.IsNotFound(err) {
```
andfasano, line 76:
```suggestion
	if !k8serrors.IsNotFound(err) {
```
nit: `IsNotFound` should return false if error is nil

zaneb: It should, but we want this block to _not_ execute if the error is nil.

PR #6521, opened 2022-10-25
OCPBUGS-1791: add bootstrap configmap during the SNO reboot

andfasano:
The `bootstrap` configmap it's consumed by external consumers, such as the `openshift-install wait-for`-like commands, to detect the successfull completion of the bootstrap phase.

The task of creating is it's usually carried on by the [progress service](https://github.com/openshift/installer/blob/3d80765f1833f757acc1b1b644ba8661fb26e3be/data/data/bootstrap/files/usr/local/bin/report-progress.sh#L22-L33), but in case of a SNO deployment it won't work due the lack of a control plane when the action is issued.

This patch thus adds a specific extra step for creating the bootstrap cm during the bootstrap in place flow - which is enabled for every SNO deployment.

**Reviewers notes** 
This code can be tested using https://github.com/openshift-metal3/dev-scripts/pull/1472, with the following configuration:
```
export AGENT_E2E_TEST_SCENARIO=SNO_IPV4_DHCP
export MIRROR_IMAGES=true

export INSTALLER_LOCAL_REPO=<installer path, ie ~/go/src/github.com/openshift/installer>
export INSTALLER_DOCKERFILE=images/installer/Dockerfile.ci
```

### DIFF
--- a/data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
+++ b/data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
@@ -12,6 +12,18 @@ function wait_for_api {
   done
 }
 
+# This is required since the progress service (https://github.com/openshift/installer/blob/dd9047c4c119e942331f702a4b7da85c60042da5/data/data/bootstrap/files/usr/local/bin/report-progress.sh#L22-L33),
+# usually dedicated to creating the bootstrap ConfigMap, will fail to create this configmap in case of bootstrap-in-place single node deployment, 
+# due to the lack of a control plane when bootkube is complete
+function signal_bootstrap_complete {
+  until oc get cm bootstrap -n kube-system &> /dev/null
+  do
+    echo "Creating bootstrap configmap ..."
+    oc create cm bootstrap -n kube-system --from-literal status=complete || true
+    sleep 5
+  done
+}
+
 function restart_kubelet {
   echo "Restarting kubelet"
   until [ "$(oc get pod -n openshift-kube-apiserver-operator --selector='app=kube-apiserver-operator' -o jsonpath='{.items[0].status.conditions[?(@.type=="Ready")].status}' | grep -c "True")" -eq 1 ];
@@ -64,6 +76,7 @@ function clean {
 }
 
 wait_for_api
+signal_bootstrap_complete
 approve_csr
 restart_kubelet
 wait_for_cvo


omertuc:
Let's keep the loop / implementation similar to the original implementation:

```
while ! oc --kubeconfig="$KUBECONFIG" create -f - <<-EOF
	apiVersion: v1
	kind: ConfigMap
	metadata:
	  name: bootstrap
	  namespace: kube-system
	data:
	  status: complete
EOF
do
	sleep 5
done
```

omertuc:
Also a comment [linking](https://github.com/openshift/installer/blob/c7d810f497d0c6c3ad22e5c14f873a70b0586231/data/data/bootstrap/files/usr/local/bin/report-progress.sh#L22) to the original implementation that does this that explains why this is here and why it has to be differently on SNO could be nice

omertuc:
We should also probably remove the [starting](https://github.com/openshift/assisted-installer/blob/f0c9d8ceab9938cb4b079c3869647d1a8f3ac7ee/src/installer/installer.go#L324) of the `progress.service` because we learnt it doesn't make much sense

andfasano:
> ~We should also probably remove the [starting](https://github.com/openshift/assisted-installer/blob/f0c9d8ceab9938cb4b079c3869647d1a8f3ac7ee/src/installer/installer.go#L324) of the `progress.service` because we learnt it doesn't make much sense~
> 
> Nevermind I'm saying silly things. the `startBootstrap` method is used by multinode

Do you maybe mean that the `bootstrap-in-place-post-reboot.sh` could be invoked also for non-SNO cases? In such scenario then very likely the cm was already created so better to keep the current implementation (which does not try to create the cm if it alreayd exists). I noticed anyhow that the sleep times were slightly different, will fix it.

omertuc:
> Do you maybe mean that the `bootstrap-in-place-post-reboot.sh` could be invoked also for non-SNO cases?

No, it cannot, it's used only when you have to create the bootstrap control plane from scratch when none exists. Ignore my comment, it's irrelevant

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -12,6 +12,17 @@ function wait_for_api {
   done
 }
 
+# This is required since the progress service (https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/report-progress.sh#L22-L33),
+# usually dedicated to create the bootstrap ConfigMap, will fail in case of SingleNodeOpenShift deployment due the lack of a control plane
```
omertuc, line 16:
```suggestion
# usually dedicated to creating the bootstrap ConfigMap, will fail to create this configmap in case of bootstrap-in-place single node deployment, due to the lack of a control plane when bootkube is complete
```

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -12,6 +12,17 @@ function wait_for_api {
   done
 }
 
+# This is required since the progress service (https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/report-progress.sh#L22-L33),
```
omertuc, line 15: Line links to a branch are unstable, use the "Copy permalink" feature in GitHub instead

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
omertuc: (The pedantic edit is because technically a SNO installatoin with a bootstrap node will not have this problem. This problem only affects bootstrap-in-place)

omertuc:
> > Let's keep the loop / implementation similar to the original implementation:
> > ```shell
> > while ! oc --kubeconfig="$KUBECONFIG" create -f - <<-EOF
> > 	apiVersion: v1
> > 	kind: ConfigMap
> > 	metadata:
> > 	  name: bootstrap
> > 	  namespace: kube-system
> > 	data:
> > 	  status: complete
> > EOF
> > do
> > 	sleep 5
> > done
> > ```
> 
> ~Good idea. It wouldn't remove the current duplication, but at least it will keep the things more similar~ (see below)

See what below?

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
andfasano: Thanks, copied the wrong one

andfasano:
> > > Let's keep the loop / implementation similar to the original implementation:
> > > ```shell
> > > while ! oc --kubeconfig="$KUBECONFIG" create -f - <<-EOF
> > > 	apiVersion: v1
> > > 	kind: ConfigMap
> > > 	metadata:
> > > 	  name: bootstrap
> > > 	  namespace: kube-system
> > > 	data:
> > > 	  status: complete
> > > EOF
> > > do
> > > 	sleep 5
> > > done
> > > ```
> > 
> > 
> > ~Good idea. It wouldn't remove the current duplication, but at least it will keep the things more similar~ (see below)
> 
> See what below?

The comment, I'd prefer to keep the current implementation as it doesn't fail if, for any reason, the cm already exists (and it's also more concise). 
In future probably we could remove such duplication (to avoid any issue) and use a single script for the cm creation.

omertuc:
> > > > Let's keep the loop / implementation similar to the original implementation:
> > > > ```shell
> > > > while ! oc --kubeconfig="$KUBECONFIG" create -f - <<-EOF
> > > > 	apiVersion: v1
> > > > 	kind: ConfigMap
> > > > 	metadata:
> > > > 	  name: bootstrap
> > > > 	  namespace: kube-system
> > > > 	data:
> > > > 	  status: complete
> > > > EOF
> > > > do
> > > > 	sleep 5
> > > > done
> > > > ```
> > > 
> > > 
> > > ~Good idea. It wouldn't remove the current duplication, but at least it will keep the things more similar~ (see below)
> > 
> > 
> > See what below?
> 
> The comment, I'd prefer to keep the current implementation as it doesn't fail if, for any reason, the cm already exists (and it's also more concise). In future probably we could remove such duplication (to avoid any issue) and use a single script for the cm creation.

Ah okay. good point about the cm already existing

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -12,6 +12,18 @@ function wait_for_api {
...
 }
 
+# This is required since the progress service (https://github.com/openshift/installer/blob/dd9047c4c119e942331f702a4b7da85c60042da5/data/data/bootstrap/files/usr/local/bin/report-progress.sh#L22-L33),
+# usually dedicated to creating the bootstrap ConfigMap, will fail to create this configmap in case of bootstrap-in-place single node deployment, due to the lack of
+# a control plane when bootkube is complete
+function signal_bootstrap_complete {
+  until oc get cm bootstrap -n kube-system &> /dev/null
+  do
+    echo "Creating bootstrap configmap ..."
+    oc create cm bootstrap -n kube-system --from-literal status=complete
```
omertuc, line 22: This could trap (see line 2)

andfasano: Nice catch, thanks @omertuc

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -12,6 +12,18 @@ function wait_for_api {
   done
 }
 
+# This is required since the progress service (https://github.com/openshift/installer/blob/dd9047c4c119e942331f702a4b7da85c60042da5/data/data/bootstrap/files/usr/local/bin/report-progress.sh#L22-L33),
+# usually dedicated to creating the bootstrap ConfigMap, will fail to create this configmap in case of bootstrap-in-place single node deployment, 
+# due to the lack of a control plane when bootkube is complete
+function signal_bootstrap_complete {
+  until oc get cm bootstrap -n kube-system &> /dev/null
+  do
+    echo "Creating bootstrap configmap ..."
```
barbacbd, line 21: Should this be printed at the top of the function rather than every 5 seconds when we are waiting?

omertuc: Yeah that's probably the right thing to do. However, I don't expect us to ever hit this echo more than once (at-least not under usual circumstances), because at this point the API server will most likely be up

lranjbar:
Created [OCPBUGS-3055](https://issues.redhat.com/browse/OCPBUGS-3055) for the 4.12 backport

PR #6524, opened 2022-10-25
WIP AGENT-62: Add the ability to view and download logs from Agent API

lranjbar:
When the Agent API is available:
- Print out the event log to the screen
- Download the cluster logs to a file


lranjbar:
We will be going in a different direction than this.

PR #6525, opened 2022-10-26
DNM: Build me an installer please

mandre:
I'm not able to reproduce the issue I'm debugging when building the installer locally, even when using the
registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.17-openshift-4.10 image. So, I'm having the CI build me an image I can use for my tests.


mandre:
Not the failure I expected
/test e2e-openstack

PR #6528, opened 2022-10-27
OCPBUGS-2891: aws: destroy: delete ELB listeners

r4f4:
There can be situations in which the listeners are not deleted with the ELB. In such cases, the installer will print the following message

level=warning msg=unrecognized elastic load balancing resource type listener arn=arn:aws:elasticloadbalancing:us-west-2:460538899914:listener/net/a9ac9f1b3019c4d1299e7ededc92b42b/a6f0655da877ddd4/45e05ee69d99bab0

and the ELB listeners will be leaked. Let's make sure to delete them instead.

### DIFF
--- a/pkg/destroy/aws/elbhelpers.go
+++ b/pkg/destroy/aws/elbhelpers.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/elb"
 	"github.com/aws/aws-sdk-go/service/elbv2"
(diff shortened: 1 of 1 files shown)


pkg/destroy/aws/elbhelpers.go
```diff
@@ -105,6 +107,18 @@ func deleteElasticLoadBalancerTargetGroup(ctx context.Context, client *elbv2.ELB
 	return nil
 }
 
+func deleteElasticLoadBalancerListener(ctx context.Context, client *elbv2.ELBV2, arn arn.ARN, logger logrus.FieldLogger) error {
+	_, err := client.DeleteListenerWithContext(ctx, &elbv2.DeleteListenerInput{
+		ListenerArn: aws.String(arn.String()),
+	})
+	if err != nil {
+		return err
```
wking, line 115: The AWS issue is about danging tags after the listener and load balancer have been removed.  So if you go this route, you'll probably want logic like [this][1] to say "ok, I guess it's already gone, and we can ignore this tag going forward".

But that might expose us to a race where the listener actually does exist and has been tagged, but due to some eventual-consistency hiccup we get a not-found response when trying to delete it.  Although in that case, I suppose we'd fail to delete the VPC, and the user could launch a fresh deprovision attempt, and by then maybe the eventual consistency issue would have resolved, and we'd be able to delete the listener.

[1]: https://github.com/openshift/installer/blob/5a9ad4c024091c6658813ab3a04c8deb61d684cc/pkg/destroy/aws/aws.go#L786

r4f4: By that logic, wouldn't the `targetgroup` resource type potentially suffer from the same problem?

Is there another route you'd suggest?

r4f4: Done.

r4f4:
/override ci/prow/golint
I ran the linter locally with a longer timeout, no issues reported on these PR changes.

PR #6529, opened 2022-10-27
OCPBUGS-2795: Bump gophercloud

mandre:
Update the pin to gophercloud v0.18.0 so that we get a fix that came
with https://github.com/gophercloud/gophercloud/pull/2160.

Also use a replacement for tencentcloud-sdk-go as they removed their
tags and broke all downstream projects in the process
(https://github.com/hashicorp/terraform#29021). This is needed in
order to run `go mod vendor`.


mandre:
It fixed deprovisioning on openstack. We should consider merging this once I sort out the jira situation to make tide happy.

PR #6533, opened 2022-10-28
OCPBUGS-2966: Add GCP CreateFirewallRules to tech preview

barbacbd:
** Require the Tech preview feature to be set in the install config when the platform.gcp.createFirewallRules is set to any value other than the default ("Enabled").

Linked to #6522

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -998,8 +998,14 @@ func validateFeatureSet(c *types.InstallConfig) field.ErrorList {
 	if c.FeatureSet != configv1.TechPreviewNoUpgrade {
 		errMsg := "the TechPreviewNoUpgrade feature set must be enabled to use this field"
 
-		if c.GCP != nil && len(c.GCP.NetworkProjectID) > 0 {
-			allErrs = append(allErrs, field.Forbidden(field.NewPath("platform", "gcp", "networkProjectID"), errMsg))
+		if c.GCP != nil {
+			if len(c.GCP.NetworkProjectID) > 0 {
+				allErrs = append(allErrs, field.Forbidden(field.NewPath("platform", "gcp", "networkProjectID"), errMsg))
+			}
+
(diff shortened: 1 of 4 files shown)


pkg/types/validation/installconfig.go
```diff
@@ -998,8 +998,14 @@ func validateFeatureSet(c *types.InstallConfig) field.ErrorList {
...
 		errMsg := "the TechPreviewNoUpgrade feature set must be enabled to use this field"
 
-		if c.GCP != nil && len(c.GCP.NetworkProjectID) > 0 {
-			allErrs = append(allErrs, field.Forbidden(field.NewPath("platform", "gcp", "networkProjectID"), errMsg))
+		if c.GCP != nil {
+			if len(c.GCP.NetworkProjectID) > 0 {
+				allErrs = append(allErrs, field.Forbidden(field.NewPath("platform", "gcp", "networkProjectID"), errMsg))
+			}
+
+			if len(c.GCP.CreateFirewallRules) > 0 && c.GCP.CreateFirewallRules != gcp.CreateFirewallRulesEnabled {
```
patrickdillon, line 1006:
```suggestion
			if len(c.GCP.CreateFirewallRules) > 0 {
```
nit: Any usage of this field, even when it's consistent with the existing functionality, should be protected by the flag. The main concern is we want the freedom to change the field, so we want users to be warned that we may break them in the future.

pkg/types/validation/installconfig_test.go
```diff
@@ -1999,6 +1999,19 @@ func TestValidateInstallConfig(t *testing.T) {
...
+				c := validInstallConfig()
+				c.Platform = types.Platform{
+					GCP: validGCPPlatform(),
+				}
+				c.GCP.CreateFirewallRules = gcp.CreateFirewallRulesDisabled
+
+				return c
+			}(),
+			expectedError: "platform.gcp.createFirewallRules: Forbidden: the TechPreviewNoUpgrade feature set must be enabled to use this field",
+		},
```
sadasu, line 2014: Do we need a test case here verify that this feature is not blocked when  c.FeatureSet is set to "TechPreviewNoUpgrade" ?

PR #6534, opened 2022-10-28
OCPBUGS-2881: Destroy the service and host project dns records

barbacbd:
** Search the host and service projects for the DNS records that were created during the install (when using the a networkProjectID).

### DIFF
--- a/pkg/destroy/gcp/dns.go
+++ b/pkg/destroy/gcp/dns.go
@@ -11,31 +11,40 @@ import (
 )
 
 type dnsZone struct {
-	name   string
-	domain string
+	name    string
+	domain  string
+	project string
 }
 
 func (o *ClusterUninstaller) listDNSZones() (private *dnsZone, public []dnsZone, err error) {
 	o.Logger.Debugf("Listing DNS Zones")
 	ctx, cancel := o.contextWithTimeout()
 	defer cancel()
-	req := o.dnsSvc.ManagedZones.List(o.ProjectID).Fields("managedZones(name,dnsName,visibility),nextPageToken")
-	err = req.Pages(ctx, func(response *dns.ManagedZonesListResponse) error {
-		for _, zone := range response.ManagedZones {
-			switch zone.Visibility {
-			case "private":
-				if o.isClusterResource(zone.Name) {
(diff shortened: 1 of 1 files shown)


jianli-wei (changes requested):
Please see my comment at [the bug](https://issues.redhat.com/browse/OCPBUGS-2881), thanks!

barbacbd:
@jianli-wei I commented back on the Jira bug. I think that your concerns have explanations. Also this PR is only meant to fix the destroy bug. Any other issues should be talked through or a bug would be opened.

pkg/destroy/gcp/dns.go
```diff
@@ -105,17 +114,19 @@ func possibleZoneParents(dnsDomain string) []string {
 	return result
 }
 
-func getParentDNSZone(dnsDomain string, publicZones []dnsZone, logger logrus.FieldLogger) *dnsZone {
+func getParentDNSZones(dnsDomain string, publicZones []dnsZone, logger logrus.FieldLogger) []*dnsZone {
+	possibleParentZones := []*dnsZone{}
```
r4f4, line 118: Nit: This name is very close to `possibleZoneParents`. Since you're checking in the loop below, can this just be `parentZones`?

pkg/destroy/gcp/dns.go
```diff
@@ -105,17 +114,19 @@ func possibleZoneParents(dnsDomain string) []string {
...
+
 	possibleParents := possibleZoneParents(dnsDomain)
 	for _, parentDomain := range possibleParents {
 		for _, zone := range publicZones {
 			if zone.domain == parentDomain {
-				logger.Debugf("Found parent dns zone: %s", zone.name)
-				return &dnsZone{name: zone.name, domain: parentDomain}
+				logger.Debugf("Found possible parent dns zone: %s", zone.name)
```
r4f4, line 124: Nit: Is the `possible` qualifier here necessary? From what I understand, the zones returned are later used as if they *were* parent zones, not *maybe let's double-check*.

r4f4 (commented):
Some nitpick comments but overall looks good.
/lgtm

PR #6540, opened 2022-10-31
OCPBUGS-2931: Azure Stack: use managed images for compute nodes

patrickdillon:
In 4.12 we switched Azure to use compute image galleries. ASH is still using managed images, but we forgot to account for that in the machine sets. This updates ASH to continue to use managed images in the machineset.

Still testing this locally, but want to post it to see some CI.

### DIFF
--- a/pkg/asset/machines/azure/machines.go
+++ b/pkg/asset/machines/azure/machines.go
@@ -3,13 +3,13 @@ package azure
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"strings"
 
 	machineapi "github.com/openshift/api/machine/v1beta1"
 	icazure "github.com/openshift/installer/pkg/asset/installconfig/azure"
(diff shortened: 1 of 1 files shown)


pkg/asset/machines/azure/machines.go
```diff
@@ -188,6 +188,9 @@ func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string
 
 	if platform.CloudName == azure.StackCloud {
 		spec.AvailabilitySet = fmt.Sprintf("%s-cluster", clusterID)
+
+		// Public Azure has moved to use image galleries, but ASH is still using managed images.
+		spec.Image.ResourceID = fmt.Sprintf("/resourceGroups/%s/providers/Microsoft.Compute/images/%s", rg, clusterID)
```
Prashanth684, line 193: ouch. my bad for missing this. is there an equivalent terraform that still creates managed images for ASH or do we need to add that as well?

Prashanth684: ah...i see azurestack has separate tf files with the managed image...

patrickdillon:
I'm going to blindly mash retest

/retest

But we can see from the failures that the machine api operator is happily creating the compute nodes, so it looks like this PR is fixing the intended issue.

jinyunma:
pre-merged validation is passed, details see bug comments.
/label qe-approved

r4f4:
Just some e2e failures.

PR #6544, opened 2022-11-01
OCPBUGS-3032: Report agent installation problems on the console

zaneb:
When a stage of agent-based installation has failed or is delayed, add a message to /etc/issues.d so that users can see immediately from the console (without logging in) what is happening.


data/data/agent/files/usr/local/bin/install-status.sh
```diff
@@ -1,5 +1,28 @@
 #!/bin/bash
 
+inactive_services() {
+    local services="assisted-service.service create-cluster-and-infraenv.service apply-host-config.service start-cluster-installation.service"
```
andfasano, line 4: This seems a little bit fragile. Everytime a developer will add/delete/remove an agent service there's the risk to miss updating this part. It could be useful to have a way to enumerate agent only service, ie by filtering for a fixed unique prefix

zaneb: I somewhat agree, and I thought about templating the list in, but I decided that what we're interested in here is really only a subset of the services - those in the critical chain leading up to start-cluster-installation. So somebody will have to maintain this list somewhere, and it might as well be here. Also, it should be vanishingly rare to add anything to this chain.

pawanpinjarkar (commented):
/lgtm

data/data/agent/files/usr/local/bin/install-status.sh
```diff
@@ -0,0 +1,39 @@
...
+            printf '\\e{cyan}Waiting for services:\\e{reset}\n'
+            systemctl --no-legend list-units --all "${show_services[@]}" | awk '{printf("[\\e{cyan}%s\\e{reset}]", $4); for (i=5; i<=NF; i++) {if (i>5 || $i != "start") printf(" %s", $i)}; print ""}'
+        } >"${services_issue}"
+    fi
+}
+
+check_host_config() {
+    local host_config_issue="/etc/issue.d/80_host-config.issue"
+    if [ -f /var/run/agent-installer/host-config-failures ]; then
+        printf '\n\\e{lightred}Installation cannot proceed:\\e{reset}\n%s\n' "$(cat /var/run/agent-installer/host-config-failures)" >"${host_config_issue}"
```
bfournie, line 29: I think probably should change "Installation cannot proceed" to "Installation validation in progress" or something. I tested this with a valid installation and this message pops up for about 30 seconds which would leave users to think their install failed, and maybe force them to abort, when in fact the next screen that pops up shows  that it has succeeded/

zaneb: Good point, it's highly ambiguous between "cannot proceed yet" or "will not proceed ever/until you do something".
I suspect it might not have been this one you saw though, but one of the ones in start-cluster-installation.sh. Can you confirm the whole message?

bfournie: You're right, its the one in start-cluster-installation.sh. I've attached a screenshot. So this one can stay, the other one we should indicate that it "may continue" or something, as opposed to "cannot proceed"
Screenshot from 2022-12-05 12-27-56

bfournie: Perhaps just change the message to "Installation waiting for hosts" so its clear that it hasn't failed yet

PR #6546, opened 2022-11-01
OCPBUGS-2841: (AGENT) only support amd64 archs

pawanpinjarkar:
With this fix, the agent installer's image creation command will fail if some other arch except amd64 is configured.

For example, when control plane and compute architecture in `install-config.yaml`
 is set to `arm64`, the command `openshift-install agent create image` will fail with
```
FATAL failed to fetch Agent Installer ISO: failed to load asset "Install Config":
invalid install-config configuration: 
[ControlPlane.Architecture: 
Unsupported value: "arm64": 
supported values: "amd64", 
Compute[0].Architecture: 
Unsupported value: "arm64": 
supported values: "amd64"]
```

### DIFF
--- a/pkg/asset/agent/installconfig.go
+++ b/pkg/asset/agent/installconfig.go
@@ -62,6 +62,10 @@ func (a *OptionalInstallConfig) validateInstallConfig(installConfig *types.Insta
 		allErrs = append(allErrs, err...)
 	}
 
+	if err := a.validateSupportedArchs(installConfig); err != nil {
+		allErrs = append(allErrs, err...)
+	}
+
 	warnUnusedConfig(installConfig)
 
 	if err := a.validateSNOConfiguration(installConfig); err != nil {
@@ -79,6 +83,26 @@ func (a *OptionalInstallConfig) validateSupportedPlatforms(installConfig *types.
 	if installConfig.Platform.Name() != "" && !IsSupportedPlatform(installConfig.Platform.Name()) {
 		allErrs = append(allErrs, field.NotSupported(fieldPath, installConfig.Platform.Name(), SupportedPlatforms))
 	}
+	return allErrs
+}
+
+func (a *OptionalInstallConfig) validateSupportedArchs(installConfig *types.InstallConfig) field.ErrorList {
+	var allErrs field.ErrorList
+
+	fieldPath := field.NewPath("ControlPlane", "Architecture")
+
+	if string(installConfig.ControlPlane.Architecture) != types.ArchitectureAMD64 {
+		allErrs = append(allErrs, field.NotSupported(fieldPath, installConfig.ControlPlane.Architecture, []string{"x86-64", "x86_64", types.ArchitectureAMD64}))
+	}
+
+	for i, compute := range installConfig.Compute {
+		fieldPath := field.NewPath(fmt.Sprintf("Compute[%d]", i), "Architecture")
+
+		if string(compute.Architecture) != types.ArchitectureAMD64 {
+			allErrs = append(allErrs, field.NotSupported(fieldPath, compute.Architecture, []string{"x86-64", "x86_64", types.ArchitectureAMD64}))
+		}
+	}
+
 	return allErrs
 }
 
@@ -138,11 +162,6 @@ func warnUnusedConfig(installConfig *types.InstallConfig) {
 			fieldPath := field.NewPath(fmt.Sprintf("Compute[%d]", i), "Hyperthreading")
 			logrus.Warnf(fmt.Sprintf("%s: %s is ignored", fieldPath, compute.Hyperthreading))
 		}
-		// kubebuilder:default=amd64. Set from generic install config code
-		if compute.Architecture != types.ArchitectureAMD64 && compute.Architecture != "x86_64" {
-			fieldPath := field.NewPath(fmt.Sprintf("Compute[%d]", i), "Architecture")
-			logrus.Warnf(fmt.Sprintf("%s: %s is ignored", fieldPath, compute.Architecture))
-		}
 
 		if compute.Platform != (types.MachinePoolPlatform{}) {
 			fieldPath := field.NewPath(fmt.Sprintf("Compute[%d]", i), "Platform")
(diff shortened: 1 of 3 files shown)


pkg/types/validation/installconfig.go
```diff
@@ -149,6 +150,19 @@ func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
...
 }
 
+func fixAMD64ArchSupport(controlPlane *types.MachinePool, compute []types.MachinePool) {
+	// "AMD64" is the name chosen by AMD for their 64-bit extension to the Intel x86 instruction set.
+	//  Before release, it was called "x86-64" or "x86_64", and some distributions still use these names
+	if controlPlane != nil && (controlPlane.Architecture == "x86-64" || controlPlane.Architecture == "x86_64") {
+		controlPlane.Architecture = types.ArchitectureAMD64
+	}
+	for i, c := range compute {
+		if c.Architecture == "x86-64" || c.Architecture == "x86_64" {
```
zaneb, line 160: I'm not sure we need this translation. In the docs (e.g. https://docs.openshift.com/container-platform/4.11/installing/installing_bare_metal/installing-bare-metal.html#installation-configuration-parameters-optional_installing-bare-metal) it appears we are pretty clear that the valid value is `amd64`, and we validate this [here](https://github.com/openshift/installer/blob/master/pkg/types/validation/machinepools.go#L74-L76).

pkg/asset/agent/installconfig.go
```diff
@@ -82,6 +86,26 @@ func (a *OptionalInstallConfig) validateSupportedPlatforms(installConfig *types.
...
+
+	if string(installConfig.ControlPlane.Architecture) != types.ArchitectureAMD64 {
+		allErrs = append(allErrs, field.NotSupported(fieldPath, installConfig.ControlPlane.Architecture, []string{"x86-64", "x86_64", types.ArchitectureAMD64}))
+	}
+
+	for i, compute := range installConfig.Compute {
+		fieldPath := field.NewPath(fmt.Sprintf("Compute[%d]", i), "Architecture")
+
+		if string(compute.Architecture) != types.ArchitectureAMD64 {
+			allErrs = append(allErrs, field.NotSupported(fieldPath, compute.Architecture, []string{"x86-64", "x86_64", types.ArchitectureAMD64}))
```
zaneb, line 102: Let's advertise only amd64 as supported, regardless of whether we translate internally.

pkg/asset/agent/installconfig_test.go
```diff
@@ -296,13 +324,13 @@ networking:
   serviceNetwork: 
   - 172.30.0.0/16
 compute:
-  - architecture: arm64
+  - architecture: x86_64
```
zaneb, line 327: `amd64`

pkg/asset/agent/installconfig.go
```diff
@@ -82,6 +86,26 @@ func (a *OptionalInstallConfig) validateSupportedPlatforms(installConfig *types.
 	return allErrs
 }
 
+func (a *OptionalInstallConfig) validateSupportedArchs(installConfig *types.InstallConfig) field.ErrorList {
+	var allErrs field.ErrorList
+
+	fieldPath := field.NewPath("ControlPlane", "Architecture")
+
+	if string(installConfig.ControlPlane.Architecture) != types.ArchitectureAMD64 {
```
zaneb, line 94: nit: we will shortly want to expand this to include at least arm64 and ppc64le, so to make that easier this could be a switch statement. (To make it even easier, it could be in a utility function so it only needs to be updated in one place.)

zaneb (commented):
Let's not try to translate it if people put `x86_64` imho.
The main thing is that we fail fast if somebody puts `arm64` or whatever, since we don't actually pay any attention to it yet. (@jeffdyoung this is another thing you'll need to change in your ARM support PR.)

pkg/asset/agent/installconfig.go
pawanpinjarkar: It looks like the info from Jira and the review comments are, are quite contradictory. For example, Jira says - 
```
Currently the agent installer supports only x86_64 arch. 
The image creation command must fail if some other arch is configured different from x86_64

We want to have an allowed list of architectures.

allowed = ['x86_64', 'amd64']
```

andfasano:
@pawanpinjarkar given that the current implementation looks good so far to address what will be just a temporary check, did you maybe also evaluate to move the check in the [ignition asset](https://github.com/openshift/installer/blob/3e6f11b8a8c579357fbcd0a8e5fe67ec7b4813ac/pkg/asset/agent/image/ignition.go#L145) to cover also the case of ztp manifests?
Also, it looks like the original [ticket](https://issues.redhat.com/browse/OCPBUGS-2841) requires to specify an allow list?

pawanpinjarkar:
@andfasano There is another PR https://github.com/openshift/installer/pull/6395/files which handles this. Is that sufficient or shall we include the changes in this? Also, looks like for the release image, only `x86_64` is allowed but the earlier discussion on this current PR is suggesting just checking for `amd64`.  If we have to check 2 different values for the arch in 2 diff places ( ZTP, install config) then a global allowed list should not be used.

pawanpinjarkar:
@andfasano and I discussed over slack and offline that we can continue with this PR as it is adding the required validation for the preferred way of the installation which covers 99% of the scenarios. For the remaining 1%, which is ZTP flow, a new task is created here https://issues.redhat.com/browse/AGENT-424

pawanpinjarkar:
Ref from slack discussion
>just amd64 for now, and we'll backport this to 4.12. In 4.13, the multi-arch team will likely add arm64 and ppc64le
https://coreos.slack.com/archives/C02TGJZHW81/p1668008340929209

pkg/asset/agent/installconfig.go
```diff
@@ -82,6 +86,30 @@ func (a *OptionalInstallConfig) validateSupportedPlatforms(installConfig *types.
...
+	switch string(installConfig.ControlPlane.Architecture) {
+	case types.ArchitectureAMD64:
+	default:
+		allErrs = append(allErrs, field.NotSupported(fieldPath, installConfig.ControlPlane.Architecture, []string{types.ArchitectureAMD64}))
+	}
+
+	for i, compute := range installConfig.Compute {
+		fieldPath := field.NewPath(fmt.Sprintf("Compute[%d]", i), "Architecture")
+
+		switch string(compute.Architecture) {
```
bfournie, line 103: What if the architecture is not defined? Will this now cause an error?

PR #6547, opened 2022-11-01
OCPBUGS-2758: Provide datastore as full path

rvanderp3:
Testing of https://github.com/openshift/cluster-config-operator/pull/269 revealed an issue with validation of the datastore path as provided by the installer in the infrastructure manifest.


r4f4 (commented):
/lgtm

PR #6551, opened 2022-11-02
azure: Add error message when authorizer fails.

rna-afk:
When the authorizer fails, the installer prints a generic message and continues with the collection of the service principal info. A clear error message as to why the authorizer fails would be helpful.

### DIFF
--- a/pkg/asset/installconfig/azure/session.go
+++ b/pkg/asset/installconfig/azure/session.go
@@ -89,7 +89,8 @@ func credentialsFromFileOrUser(cloudEnv *azureenv.Environment) (*Credentials, er
 	os.Setenv(azureAuthEnv, authFilePath)
 	_, err := auth.NewAuthorizerFromFileWithResource(cloudEnv.ResourceManagerEndpoint)
 	if err != nil {
-		logrus.Debug("Could not get an azure authorizer from file. Asking user to provide authentication info")
+		logrus.Infof("Could not get an azure authorizer from file: %s", err.Error())
+		logrus.Infof("Asking user to provide authentication info")
 		credentials, err := askForCredentials()
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to retrieve credentials from user")


pkg/asset/installconfig/azure/session.go
```diff
@@ -89,7 +89,8 @@ func credentialsFromFileOrUser(cloudEnv *azureenv.Environment) (*Credentials, er
 	os.Setenv(azureAuthEnv, authFilePath)
 	_, err := auth.NewAuthorizerFromFileWithResource(cloudEnv.ResourceManagerEndpoint)
 	if err != nil {
-		logrus.Debug("Could not get an azure authorizer from file. Asking user to provide authentication info")
+		logrus.Debugf("Could not get an azure authorizer from file: %s", err.Error())
```
r4f4, line 92: Could we make this an `Infof` so it's clear to users not running with debug logs that the certificate auth failed?

pkg/asset/installconfig/azure/session.go
```diff
@@ -89,7 +89,8 @@ func credentialsFromFileOrUser(cloudEnv *azureenv.Environment) (*Credentials, er
...
+		logrus.Debugf("Asking user to provide authentication info")
```
r4f4, line 93: Does it make sense to fallback to asking for auth info if the cert failed? Maybe we could just fail and let the user decide if they want to fix their certs or move over to client secret auth?

pkg/asset/installconfig/azure/session.go
rna-afk: If we are going to default to asking info from the user then keeping this in debug makes sense. I see the uses of making this Infof too

pkg/asset/installconfig/azure/session.go
rna-afk: This is a more general use case where if anything in the service principal is wrong, we default to creating a new one. I'm not sure if we need to change this for just certs use case.

r4f4:
We should create a bug for this so the fix can be backported to 4.12.

rna-afk:
/skip

This is just a log change

PR #6556, opened 2022-11-03
Enabling VPCZone setting in install-config.yaml for PowerVS

miyamotoh:
Adding support for `vpcZone` setting in `install-config.yaml` for `create manifests` operation. When specified;
- it'll be validated to be a valid VPC zone in IBM Cloud
- it'll be used to infer `vpcRegion` (in other words, specifying `vpcRegion` in `install-config.yaml` has no effects)
- the inferred region will be used to look up the `vpcName`, if also specified, and would fail if not found in the inferred region
- the inferred region will be used to look up the `vpcSubnets[0]`, if also specified, and would fail if not found in the inferred region
- it will _not_ affect the COS location for the bootstrap ignition file which will still be near the specified PowerVS region

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -3089,6 +3089,11 @@ spec:
                     items:
                       type: string
                     type: array
+                  vpcZone:
+                    description: VPCZone specifies the IBM Cloud zone in which to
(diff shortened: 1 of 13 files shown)


miyamotoh:
After discussing this work a bit further with @mjturek, we decided to rescind this one and create a new one with an easier to understand requirements and logic.

PR #6560, opened 2022-11-04
WIP agent: present questionnare in interactive mode

celebdor:
TODO:
* Disable the get-interactive-config.service in interactive mode
* Write node0 value and rendezvousIP in the right places

### DIFF
--- a/data/data/agent/files/usr/local/bin/get-interactive-config.sh
+++ b/data/data/agent/files/usr/local/bin/get-interactive-config.sh
@@ -1,0 +1,96 @@
+#!/bin/bash
+
+function connectivity_check() {
+	declare target
+
+	target="$1"
+
+	(>&2 echo "Pinging ${target} to check the validity of the current network configuration")
+	if ping -c 4 "$NODE_ZERO_IP"; then
+		(>&2 echo "Successfully pinged ${target}.")
+		return 0
+	else
+		(>&2 echo "Failed to ping ${target}.")
+		return 1
+	fi
+}
+
+function print_addressing_info() {
+	declare with_prefix
+
+	with_prefix="${1:-true}"
+	echo "Currently configured addresses"
+	if $with_prefix; then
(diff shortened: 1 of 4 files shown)


data/data/agent/files/usr/local/bin/get-interactive-config.sh
```diff
@@ -0,0 +1,96 @@
...
+	elif [[ "$NODE0_ANSWER" =~ ^[Nn](o)?$ ]]; then
+		(>&2 echo "This node will not run the installation service.")
+		break
+	else
+		(>&2 echo "Invalid answer.")
+	fi
+done
+
+if $NODE_ZERO; then
+	print_addressing_info
```
rwsu, line 45: The print_addressing_info is only displayed for Node0. The other nodes don't show their currently configured ip address. So it is unclear whether the nmtui should be used.

data/data/agent/files/usr/local/bin/get-interactive-config.sh
```diff
@@ -0,0 +1,96 @@
...
+
+	with_prefix="${1:-true}"
+	echo "Currently configured addresses"
+	if $with_prefix; then
+		ip -j a show  | jq -r '. | map({name: .ifname, addresses: [.addr_info[] | .local + "/" + (.prefixlen | tostring)]}) | .[] | select(.addresses | length > 0) | select( .name != "lo") | .name + ": " + (.addresses | join(" "))'
+	else
+		ip -j a show  | jq -r '. | map({name: .ifname, addresses: [.addr_info[] | .local]}) | .[] | select(.addresses | length > 0) | select( .name != "lo") | .name + ": " + (.addresses | join(" "))'
+	fi
+}
+
```
rwsu, line 29: I'm assuming we need to add something here that checks if node0 is defined in the ISO. If it is defined and connectivity check passes, then the following loops should be skipped after a predetermined amount timeout (according to https://issues.redhat.com/browse/AGENT-455)

rwsu (changes requested):

PR #6567, opened 2022-11-07
OCPBUGS-3280: Automatically retry install

bfournie:
In https://github.com/openshift/installer/pull/6470 we detected a failure when the cluster state moves back to Ready after an installation has been initiated. This adds an automatic retry when that condition occurs. It will help resolve issues like https://issues.redhat.com/browse/OCPBUGS-3280 and, in general, any problems that cause a cluster prepare failure.

### DIFF
--- a/data/data/agent/files/usr/local/bin/start-cluster-installation.sh.template
+++ b/data/data/agent/files/usr/local/bin/start-cluster-installation.sh.template
@@ -58,11 +58,13 @@ if [[ "${APIVIP}" != "" ]]; then
     fi
 fi
 
-while [[ "${cluster_status}" != "ready" ]]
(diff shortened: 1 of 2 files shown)


celebdor:
/lgtm

Maybe it would make sense to put a limit to the retrying so wait-for will not stay forever waiting for retries

bfournie:
> /lgtm
> 
> Maybe it would make sense to put a limit to the retrying so wait-for will not stay forever waiting for retries

So there's no change to the waitfor timeout and it will still terminate if the bootstrap doesn't complete (currently 45 minutes), but with this change the service will keep retrying, and the cluster could potentially come up after waitfor has timed out. We could put a timeout into the service to limit the time it keeps retrying.

PR #6570, opened 2022-11-08
Add tested instance types for secret regions

yunjiang29:
Add a separate table for tested instance types in C2S/SC2S regions.

More discussion can be found https://github.com/openshift/openshift-docs/pull/52199#issuecomment-1305435492


mtulio (commented):
@yunjiang29 lgtm, feel free to remove `/hold` when you are fine to merge and have more time for Patrick's feedback.

do we have the intention to fetch and parse those files from somewhere? If yes, maybe we could review the naming convention of those files. :)

/lgtm
/approve
/hold

PR #6571, opened 2022-11-08
OCPBUGS-3382: Fix cluster wide proxy

mhanss:
To create cluster wide proxy, proxy entries must be made in the agent-cluster-install.yaml. This commit will ensure that if the installConfig has the proxy section provided, it must be present into the agent-cluster-install.yaml.


lranjbar:
@mhanss We should backport this one to 4.12.

PR #6573, opened 2022-11-08
OCPBUGS-3278: (Agent) Do not require host data in platform baremetal section in installconfig

pawanpinjarkar:
For the agent-based installer, do not require the host data in the platform bare metal section in installconfig as it is unused by the agent installer. If still provided, a warning is issued to the user that the data is ignored.

### DIFF
--- a/pkg/types/baremetal/validation/platform.go
+++ b/pkg/types/baremetal/validation/platform.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/types/baremetal/validation/platform.go
```diff
@@ -404,23 +405,27 @@ func ValidatePlatform(p *baremetal.Platform, n *types.Networking, fldPath *field
 		}
 	}
 
-	if p.Hosts == nil {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("hosts"), p.Hosts, "bare metal hosts are missing"))
+	pathName, _ := os.Executable()
+	if !strings.Contains(pathName, "agent") {
```
pawanpinjarkar, line 409: This worked while debugging the test case but did not work when building the binary and running the create image command. Need to look more into it.

pkg/types/baremetal/validation/platform.go
```diff
@@ -404,23 +405,27 @@ func ValidatePlatform(p *baremetal.Platform, n *types.Networking, fldPath *field
...
+		if p.Hosts == nil {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("hosts"), p.Hosts, "bare metal hosts are missing"))
+		}
 	}
```
celebdor, line 413: I think it's better to use the args:
```
agentBasedInstallation := len(os.Args) > 1 && os.Args[1] == "agent"

if !agentBasedInstallation && p.Hosts == nil {
        allErrs = append(allErrs, field.Invalid(fldPath.Child("hosts"), p.Hosts, "bare metal hosts are missing"))
}

File: pkg/types/baremetal/validation/platform.go [hunk 3]
```diff
@@ -404,23 +405,27 @@ func ValidatePlatform(p *baremetal.Platform, n *types.Networking, fldPath *field
...
+		if err := validateHostsCount(p.Hosts, c); err != nil {
+			allErrs = append(allErrs, field.Required(fldPath.Child("Hosts"), err.Error()))
+		}
+		allErrs = append(allErrs, validateHostsWithoutBMC(p.Hosts, fldPath)...)
+		allErrs = append(allErrs, validateBootMode(p.Hosts, fldPath.Child("Hosts"))...)
+		allErrs = append(allErrs, validateNetworkConfig(p.Hosts, fldPath.Child("Hosts"))...)
 
-	allErrs = append(allErrs, validateHostsName(p.Hosts, fldPath.Child("Hosts"))...)
+		allErrs = append(allErrs, validateHostsName(p.Hosts, fldPath.Child("Hosts"))...)
+	}
```
celebdor, line 428: I think you could use the agentBasedInstallation I suggested in the snippet above

celebdor (changes requested):
With the changes I proposed, it works after compilation

pkg/types/baremetal/validation/platform.go
```diff
@@ -404,23 +405,26 @@ func ValidatePlatform(p *baremetal.Platform, n *types.Networking, fldPath *field
 		}
 	}
 
-	if p.Hosts == nil {
+	agentBasedInstallation := len(os.Args) > 1 && os.Args[1] == "agent"
```
andfasano, line 408: On a second thought, this approach seems to me too specific and fragile. 
Is there anything that prevents us from skipping _entirely_ the baremetal validation (with a proper validation msg) - until we're going to really use it?

zaneb: We need to validate the VIPs somehow.
Also, if we're hoping that people can use the same install-config for both IPI and agent installs, then it doesn't make sense to allow them to put invalid data in it, even if it's unused.

An alternative to looking at `os.Args` might be to put some flag in the InstallConfig struct that gets set by `OptionalInstallConfig.Load()`. That isn't any less horrible, but it does have the advantage of working the same way in unit tests.

zaneb:
I'm OK with this patch as it is, especially as it needs to be backported to 4.12. In the long term it would be good if we can make the InstallConfig code more modular so that we can run only the validations that we want, but that sounds like a potentially intrusive change to backport.

/lgtm
/cherry-pick release-4.12

PR #6589, opened 2022-11-14
AGENT-139: WIP - Embed nmstate

celebdor:
This requires:
* nmstate-static installed in the system
* glibc-static installed in the system (or equivalent static libc)
* Build with: CGO_ENABLED=1 GOARCH=amd64 GOOS=linux OUTPUT="bin/openshift-install" hack/build.sh

TODO:
* See what we do for mac to get the static libraries (they exist but we need to get them)
* Change nmstatectl usage to the bindings

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -200,6 +200,7 @@ require (
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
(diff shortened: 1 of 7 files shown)


cathay4t:
Hi @celebdor ,

We are working on including nmstate into host CoreOS via https://github.com/coreos/fedora-coreos-config/pull/2269 , do you still need to embed nmstate?

Does including `nmstate-libs` (C library of nmstate) also helps?

zaneb:
> We are working on including nmstate into host CoreOS via [coreos/fedora-coreos-config#2269](https://github.com/coreos/fedora-coreos-config/pull/2269) , do you still need to embed nmstate?

Yes, it's still needed in the installer binary so that we can use it to validate data when it is provided.

> Does including `nmstate-libs` (C library of nmstate) also helps?

No, the installer is a standalone binary that the user downloads, so it can't depend on dynamic libraries.

PR #6595, opened 2022-11-15
platformprovisioncheck: fix shadowing of `err` variable

r4f4:
Some switch branches were relying on the outer error variable but shadowing it with a local one whereas other branches were just returning right away in case of errors. Let's make them all consistent by returning the error and using `nil` explicitly at the end of the function to make it clear all validations passed at that point.


r4f4:
Known e2e failures.

PR #6596, opened 2022-11-15
OCPBUGS-3668: fully qualified username must be provided

rvanderp3:
The problem:
If a vCenter is configured with a default domain and the domain is not provided to the installer, the vCenter API is unable to validate privileges.

The fix:
Use the UserSession from session manager to determine fully qualified username


jcpowermac:
vSphere test failure unrelated to this change
/approve

rvanderp3:
just noticed golint failing, taking a look

PR #6599, opened 2022-11-18
WIP: Experimental: Extract plugins from release image

patrickdillon:
I have been working on this during shift week (or the past couple of shift weeks). 

The idea is to pull terraform and its providers from the release image rather than building them alongside the installer. I have this working locally with linux on amd64 (well... tested to install complete on aws, and a preliminary test on gcp), so I would like to see how it performs in CI.


patrickdillon:
ci/prow/images succeeded:
```
[36mINFO[0m[2022-11-18T17:44:02Z] Ran for 25m49s                               
```
Screenshot from 2022-11-18 13-31-30

First impression is that this may be shaving an hour off of average build time. 

e2e-aws had a successful install 

azure failure may be related to the fact that the provider artifacts I pushed to my personal quay repo (and being used in this test) may have been from an old branch...

PR #6608, opened 2022-11-21
OCPBUGS-3933: OpenStack: Ensure client does not initiate HTTP2 connections

mandre:
Workaround a gophercloud issue [1] where it fails to retrieve object list when Swift is served over HTTP2.

[1] https://github.com/gophercloud/gophercloud/issues/2509


pkg/destroy/openstack/openstack.go
```diff
@@ -90,6 +91,15 @@ func New(logger logrus.FieldLogger, metadata *types.ClusterMetadata) (providers.
 func (o *ClusterUninstaller) Run() (*types.ClusterQuota, error) {
 	opts := openstackdefaults.DefaultClientOpts(o.Cloud)
 
+	// Gophercloud may fail to parse swift response when served over HTTP2
+	// https://issues.redhat.com/browse/OCPBUGS-3933
+	opts.HTTPClient = &http.Client{
+		Transport: &http.Transport{
+			Proxy:             http.ProxyFromEnvironment,
+			ForceAttemptHTTP2: false,
```
pierreprinetti, line 99: The documentation does not say that setting `ForceAttemptHTTP2` to `false` will disable HTTP2.

[The `Transport` docs](https://pkg.go.dev/net%2Fhttp#Transport) say:

```
ForceAttemptHTTP2 controls whether HTTP/2 is enabled when a non-zero
Dial, DialTLS, or DialContext func or TLSClientConfig is provided.
By default, use of any those fields conservatively disables HTTP/2.
To use a custom dialer or TLS config and still attempt HTTP/2
upgrades, set this to true.
```

This is reflected in code [here](https://cs.opensource.google/go/go/+/refs/tags/go1.19.3:src/net/http/transport.go;l=387): `ForceAttemptHTTP2` is only relevant if any of those properties of `Transport` is non-nil.

---

In order to disable HTTP2 in the least impactful way, I would suggest to set `TLSNextProto` to an empty map. This will disable HTTP2 because [an empty map is not nil](https://go.dev/play/p/mTv4wQAhC1M), but it should not impact any protocol because there can be no match.

```Go
opts.HTTPClient = &http.Client{
	Transport: &http.Transport{
		Proxy:             http.ProxyFromEnvironment,
		TLSNextProto: map[string]func(string, *tls.Conn) RoundTripper{},
	},
}
```

pierreprinetti (changes requested):

pkg/destroy/openstack/openstack.go
mandre: Thanks, I'm using your suggestion and moved the code inside the deleteContainers() function so that I can modify the existing client without having to reimplement the gophercloud logic to handle self-signed certificates.

mandre:
/close
@pierreprinetti found a better solution in https://github.com/openshift/installer/pull/6615

PR #6609, opened 2022-11-21
Build installer-artifacts in one stage

yselkowitz:
This is to test if all the binaries can be built in one stage, and if it is faster to do so.

https://coreos.slack.com/archives/C68TNFWA2/p1669056726528059

/cc @patrickdillon


yselkowitz:
This doesn't seem to have any noticeable effect on build times; closing.

PR #6610, opened 2022-11-21
GCP: remove private zone manipulation

barbacbd:
** Should not specify the privatezone for GCP installations.

https://issues.redhat.com/browse/CORS-2409


r4f4 (commented):
/lgtm

PR #6612, opened 2022-11-22
[SPLAT-877] vSphere Zonal: validate dns and external load balancer

jcpowermac:
The PR:
https://github.com/openshift/installer/pull/5798
Introduced the ability to install with
the VIPs unset.

This PR adds validation to the vSphere
zonal installation for DNS and LB
existence.

It will query DNS for api and api-int.
If either fail a error will be provided.

It will perform a TCP/6443 port connection
to the `api.ClusterDomain()`. If not
available will emit a warning only.

### DIFF
--- a/pkg/asset/installconfig/vsphere/validation.go
+++ b/pkg/asset/installconfig/vsphere/validation.go
@@ -3,6 +3,9 @@ package vsphere
 import (
 	"context"
 	"fmt"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"net"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


jcpowermac:
/hold

breaks upi

pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -411,3 +429,56 @@ func validateVcenterPrivileges(validationCtx *validationContext, fldPath *field.
...
+	dialTimeout := time.Second
+	tcpTimeout := time.Second * 10
+	tcpContext, cancel := context.WithTimeout(context.TODO(), tcpTimeout)
+	defer cancel()
+
+	errList := field.ErrorList{}
+	var lastErr error
+	errorCount := 0
+
+	apiURI := fmt.Sprintf("api.%s.%s", installConfig.ObjectMeta.Name, installConfig.BaseDomain)
```
romfreiman, line 443: what about api-int?

bostrt, line 443: This can be simplified some using `installConfig.ClusterDomain()` https://github.com/openshift/installer/blob/master/pkg/types/installconfig.go#L195-L198

bostrt:
@jcpowermac I think only warning is a good idea :+1:

jcpowermac:
cc: @rvanderp3

bostrt:
The test failures don't look related to the changes in this PR

/lgtm
/approve

PR #6615, opened 2022-11-23
OCPBUGS-3933: OpenStack: Force JSON content-type in Swift

pierreprinetti:

With this change, we work around the issue by forcing Gophercloud to request a JSON response from the object storage when listing containers. When passed an `Accept: application/json` header, the server responds with `200 Ok` and a `content-type` header in our tests.

Note that in my local tests, I didn't find the 'full' listing to take more time than the short, name-only response we were requesting Swift prior to this change.

### DIFF
--- a/pkg/destroy/openstack/openstack.go
+++ b/pkg/destroy/openstack/openstack.go
@@ -1044,7 +1044,21 @@ func deleteContainers(opts *clientconfig.ClientOpts, filter Filter, logger logru
 		logger.Error(err)
 		return false, nil
 	}
-	listOpts := containers.ListOpts{Full: false}
+
+	// Listing with `Full == true` so that Gophercloud sets the `Accept`
+	// header to `application/json`. If `Full` is not true, Gophercloud
+	// sets `Accept: text/plain`, to which some objectstorage instances may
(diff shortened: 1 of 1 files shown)


pierreprinetti:
An alternative approach to https://github.com/openshift/installer/pull/6608. Let's see how the tests go.

pierreprinetti:
[The e2e-openstack test has passed](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/6615/pull-ci-openshift-installer-master-e2e-openstack/1595340954909282304). I will now force-push a better commit message, and no change in the code.

pkg/destroy/openstack/openstack.go
```diff
@@ -1044,7 +1044,7 @@ func deleteContainers(opts *clientconfig.ClientOpts, filter Filter, logger logru
 		logger.Error(err)
 		return false, nil
 	}
-	listOpts := containers.ListOpts{Full: false}
```
mandre, line 1047: Sweet. Could you perhaps add a comment to explain what you're doing here?

pierreprinetti: Good point.

PR #6616, opened 2022-11-23
Set ip=dhcp,dhcp6 for master nodes on dualstack

derekhiggins:
We were previously only seting this for workers but its also needed on the master nodes to ensure both networks are ready before ovnkube-node starts.


derekhiggins:
Closed in favour of a backport of the 4.13 PR 
https://github.com/openshift/installer/pull/6706

PR #6617, opened 2022-11-23
Bug OCPBUGS-3921: OpenStack: fix bootstrap destroy cmd

MaysaMacedo:
The installer bootstrap destroy command does not locate the clouds.yaml file containing the credentials when it's present in the current directory. This commit fixes the issue by ensuring that prior to destroying the resources, the installer attempts to look for the OS_CLIENT_CONFIG_FILE env var and if unset looks for any clouds.yaml file in the current directory.

### DIFF
--- a/pkg/destroy/bootstrap/bootstrap.go
+++ b/pkg/destroy/bootstrap/bootstrap.go
@@ -9,6 +9,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/openshift/installer/pkg/asset/cluster"
+	openstackasset "github.com/openshift/installer/pkg/asset/cluster/openstack"
 	osp "github.com/openshift/installer/pkg/destroy/openstack"
 	"github.com/openshift/installer/pkg/terraform"
 	platformstages "github.com/openshift/installer/pkg/terraform/stages/platform"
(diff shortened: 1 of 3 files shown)


MaysaMacedo:
This bug looks similar to https://github.com/openshift/installer/pull/5947
/hold for more testing
/cc @mandre

MaysaMacedo:
/hold cancel
`opensthift-install destroy bootrap` cmd works fine

pkg/destroy/bootstrap/bootstrap.go
```diff
@@ -31,6 +31,20 @@ func Destroy(dir string) (err error) {
 	}
 
 	if platform == openstack.Name {
+		if path := os.Getenv("OS_CLIENT_CONFIG_FILE"); path == "" {
```
mandre, line 34: This duplicates what we have in [PreTerraform()](https://github.com/openshift/installer/blob/e6e2400/pkg/asset/cluster/openstack/openstack.go#L19)... Should this be wrapped in a function?

MaysaMacedo: Done. I ended up using that same function and removing the parameters that were not necessary.

mandre (commented):
/lgtm

MaysaMacedo:
/assign @patrickdillon
Can you have a look on this PR? Thank you

mandre (commented):
/lgtm
/hold cancel

PR #6626, opened 2022-11-24
OCPBUGS-3196: Set ip=dhcp,dhcp6 for master nodes on dualstack

derekhiggins:
We were previously only seting this for workers but its also needed on the master nodes to ensure both networks are ready before ovnkube-node starts.


r4f4 (approved):
/approve

r4f4:
Known e2e failures

derekhiggins:
Images are failing to build in some of the jobs, looks like its because of a recent base image update
https://github.com/openshift/installer/pull/6657

derekhiggins:
/unhold
failing image builds should now be fixed

PR #6629, opened 2022-11-28
CORS-2367: Gather bootstrap logs using proxy

EmilienM:
Implements: [CORS-2367](https://issues.redhat.com//browse/CORS-2367)

If SSH port is authorized in the list of rules to configure to which ports the proxy can establish a connection, it's possible to gather the bootstrap logs
through the proxy, for example in a disconnected environment.

```sh
openshift-install gather bootstrap --bootstrap ${BOOTSTRAP_HOST_IP} --proxy https://foo:mysecret@proxy.host
```

Co-Authored-By: Maysa De Macedo Souza <mdemaced@redhat.com>

### DIFF
--- a/pkg/gather/ssh/ssh.go
+++ b/pkg/gather/ssh/ssh.go
@@ -2,6 +2,12 @@
 package ssh
 
 import (
+	"bufio"
+	"crypto/tls"
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -11,21 +17,37 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/crypto/ssh"
 	"golang.org/x/crypto/ssh/agent"
+	"golang.org/x/net/proxy"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"github.com/openshift/installer/pkg/lineprinter"
 )
 
+type httpProxy struct {
+	host     string
+	haveAuth bool
+	username string
+	password string
+	forward  proxy.Dialer
+}
+
+type httpsDialer struct{}
+
+// Dialer for HTTPS.
+var HTTPSDialer = httpsDialer{}
+
+// TLS configuration needed for HTTPS dialer.
+var TLSConfig = &tls.Config{}
+
 // NewClient creates a new SSH client which can be used to SSH to address using user and the keys.
-//
 // if keys list is empty, it tries to load the keys from the user's environment.
-func NewClient(user, address string, keys []string) (*ssh.Client, error) {
+func NewClient(user, address string, keys []string, proxyURL string) (*ssh.Client, error) {
 	ag, agentType, err := getAgent(keys)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to initialize the SSH agent")
 	}
 
-	client, err := ssh.Dial("tcp", address, &ssh.ClientConfig{
+	sshConfig := &ssh.ClientConfig{
 		User: user,
 		Auth: []ssh.AuthMethod{
 			// Use a callback rather than PublicKeys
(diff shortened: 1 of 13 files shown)


pkg/gather/ssh/ssh.go
```diff
@@ -13,10 +19,20 @@ import (
...
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
+// httpProxy is a HTTP/HTTPS connect proxy.
+type httpProxy struct {
+	host     string
+	haveAuth bool
+	username string
+	password string
+	forward  proxy.Dialer
```
MaysaMacedo, line 33: can we use proxy.FromEnvironment? It seems to return a Dialer as well and might avoid the getHttpProxyUrl func

EmilienM:
TODO:
* Make it work with the new patch made from Maysa's comments
* Test with `HTTPS_PROXY` (and a squid HTTPS endpoint)
* Add support for `NO_PROXY`
* Figure out if we need to add CLI args for proxy vars or should we read from env.

EmilienM:
@patrickdillon could you look at this one please -- and tell us if we need to add CLI args for proxy vars or should we read from env, for UX.

@MaysaMacedo HTTPS doesn't work, we need to investigate.

patrickdillon:
> @patrickdillon could you look at this one please -- and tell us if we need to add CLI args for proxy vars or should we read from env, for UX.

Typically we do not allow any env vars for user input. There are some exceptions to this, especially if we are using external packages that enable those env vars. In this case, I think env vars would be acceptable and sensible. 

As to env vars vs cli flags; I can't say I have a strong opinion there. My knee-jerk reaction is env vars, but that is pretty weakly held. Are you leaning one way or the other?

EmilienM:
> As to env vars vs cli flags; I can't say I have a strong opinion there. My knee-jerk reaction is env vars, but that is pretty weakly held. Are you leaning one way or the other?

Both options have advantages and disadvantages:
* without CLI options, a user can just re-use exported environment
* exported environment can be forgotten when running CLI and not wanted at that time, so having CLI options makes it clear and wanted for the user that a proxy is needed to SSH the bootstrap.

No strong opinion here either.

EmilienM:
@MaysaMacedo @patrickdillon I ended up adding a new argument for the CLI. This PR is ready for review now.

pkg/gather/ssh/ssh.go
```diff
@@ -141,3 +191,84 @@ func LoadPrivateSSHKeys(paths []string) (map[string]interface{}, error) {
...
+// newHTTPSProxy creates a new httpProxy populated with credentials using HTTPS dialer
+func newHTTPSProxy(uri *url.URL, forward proxy.Dialer) (proxy.Dialer, error) {
+	return newHTTPProxy(uri, HttpsDialer)
+}
+
+// Dial is a custom dialer for the ssh client that will use HTTP CONNECT method via TLS handshake
+func (d httpsDialer) Dial(network, addr string) (c net.Conn, err error) {
+	c, err = tls.Dial("tcp", addr, TlsConfig)
+	if err != nil {
+		fmt.Println(err)
```
MaysaMacedo, line 219: we probably want to wrap the error here with:

`return nil, errors.Wrap(err, "failed to connect to proxy")`

pkg/gather/ssh/ssh.go
```diff
@@ -141,3 +191,84 @@ func LoadPrivateSSHKeys(paths []string) (map[string]interface{}, error) {
...
+	}
+	return
```
MaysaMacedo, line 221: the default return should probably be:

`return c, nil`

pkg/gather/ssh/ssh.go
MaysaMacedo: Also, should we close the connection the same way is done with httpProxy?

pkg/gather/ssh/ssh.go
```diff
@@ -141,3 +191,84 @@ func LoadPrivateSSHKeys(paths []string) (map[string]interface{}, error) {
...
+}
+
+// Dial is a custom dialer for the ssh client that will use HTTP CONNECT method
+// to establish a connection to the proxy. This is needed because the net/proxy module
+// only supports SOCKS5 proxies.
+func (s *httpProxy) Dial(network, addr string) (net.Conn, error) {
+	c, err := s.forward.Dial("tcp", s.host)
+	if c == nil {
```
MaysaMacedo, line 229: is there a case when the connection is nil and no error was raised? I wonder if we can just check for the error

pkg/gather/ssh/ssh.go
```diff
@@ -141,3 +191,84 @@ func LoadPrivateSSHKeys(paths []string) (map[string]interface{}, error) {
...
+	}
+
+	resp, err := http.ReadResponse(bufio.NewReader(c), req)
+	if err != nil {
+		c.Close()
+		return nil, err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != 200 {
+		logrus.Debugf("connect server using proxy, StatusCode [%d]", resp.StatusCode)
```
MaysaMacedo, line 267: since we're returning the error, maybe there is no need for the log?

pkg/gather/ssh/ssh.go
```diff
@@ -141,3 +191,84 @@ func LoadPrivateSSHKeys(paths []string) (map[string]interface{}, error) {
...
+}
+
+// newHTTPSProxy creates a new httpProxy populated with credentials using HTTPS dialer
+func newHTTPSProxy(uri *url.URL, forward proxy.Dialer) (proxy.Dialer, error) {
+	return newHTTPProxy(uri, HttpsDialer)
+}
+
+// Dial is a custom dialer for the ssh client that will use HTTP CONNECT method via TLS handshake
+func (d httpsDialer) Dial(network, addr string) (c net.Conn, err error) {
+	c, err = tls.Dial("tcp", addr, TlsConfig)
```
MaysaMacedo, line 217: Is there a need for TlsConfig to contain RootCAs?

pkg/gather/ssh/ssh.go
EmilienM: So now we are providing the URL directly, so we can't use the Environment.
However I checked if we could use `proxy.FromURL` to return their Dialer, but this doesn't support HTTP Proxy (only Socks) - which is why we have to have the custom struct and function:
https://cs.opensource.google/go/x/net/+/refs/tags/v0.2.0:proxy/proxy.go;l=81-110

I'll close the comment but please re-open if you still think we can re-use something.

pkg/gather/ssh/ssh.go
EmilienM: Why would we need it? Do you refer to `additionalTrustBundle` somehow?

pkg/gather/ssh/ssh.go
EmilienM: done

pkg/gather/ssh/ssh.go
EmilienM: done

pkg/gather/ssh/ssh.go
EmilienM: right. i'll put the status code however.

pkg/gather/ssh/ssh.go
EmilienM: yes in fact if the proxy is unreachable (no route to host) we get no error.

sadasu:
@EmilienM still interested in this?

EmilienM:
I'll close it for now as none of us are actively working on that.
/close

PR #6633, opened 2022-11-28
Update error message

pawanpinjarkar:
Updated error message


cmd/openshift-install/agent/waitfor.go
```diff
@@ -76,7 +78,7 @@ func newWaitForInstallCompleteCmd() *cobra.Command {
 				cluster.PrintInfraEnvRestAPIEventList()
 				err2 := cluster.API.OpenShift.LogClusterOperatorConditions()
 				if err2 != nil {
-					logrus.Error("Attempted to gather ClusterOperator status after wait failure: ", err2)
+					logrus.Error(fmt.Sprintf("Attempted to gather ClusterOperator status after wait failure: %s. Agent-based installer succeeded with the cluster installation but the error is caused due to Cluster Operator's status availability conditions.", err2))
```
r4f4, line 81: Why not https://pkg.go.dev/github.com/sirupsen/logrus#Errorf

cmd/openshift-install/agent/waitfor.go
```diff
@@ -76,7 +76,7 @@ func newWaitForInstallCompleteCmd() *cobra.Command {
 				cluster.PrintInfraEnvRestAPIEventList()
 				err2 := cluster.API.OpenShift.LogClusterOperatorConditions()
 				if err2 != nil {
-					logrus.Error("Attempted to gather ClusterOperator status after wait failure: ", err2)
+					logrus.Errorf("Attempted to gather ClusterOperator status after wait failure: %s. Agent-based installer succeeded with the cluster installation but the error is caused due to Cluster Operator's status availability conditions.", err2)
```
andfasano, line 79: We've tried to keep a behavior as much as possible similar to the default `wait-for` command, so I'd not recommend to have a different error message for [the same cases](https://github.com/openshift/installer/blob/c45c58edee95a632ad3eedf55c2d0e7afa563f5e/cmd/openshift-install/waitfor.go#L51) (ie, in case let's change them both, as the problem would be the same also in the non-agent command)

cmd/openshift-install/agent/waitfor.go
```diff
@@ -76,7 +76,7 @@ func newWaitForInstallCompleteCmd() *cobra.Command {
 				cluster.PrintInfraEnvRestAPIEventList()
 				err2 := cluster.API.OpenShift.LogClusterOperatorConditions()
 				if err2 != nil {
-					logrus.Error("Attempted to gather ClusterOperator status after wait failure: ", err2)
+					logrus.Errorf("Attempted to gather ClusterOperator status after wait failure: %s. Openshift installer succeeded with the cluster installation but the error is caused due to Cluster Operator's status availability conditions.", err2)
```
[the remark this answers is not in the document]
zaneb: So IIUC, this error - `err2` - is one we get when attempting to get the list of ClusterOperators for debugging purposes. i.e. when the API is _not_ up. So I don't think this new message is right.

It's the `err` that tells us the cluster is up but the operators are not ready, and the error message for that below (ll. 81-83) looks correct.

pawanpinjarkar:
We will try increasing the timeout for bare-metal instead.

PR #6634, opened 2022-11-29
Confidential computing

eranco74:
Allow users to install OCP cluster on GCP confidential VMs

### Changes
pkg: Add EnableConfidentialCompute and OnHostMaintenance to gcp machines
terraform: Add enable_confidential_compute and on_host_maintenance configurations to google_compute_instance TF
api: go.mod: Replace github.com/openshift/api with github.com/eranco74/api@enableConfidentialCompute + vendor

### constraints:
1. Note that the ConfidentialCompute service requires an image with SEV_CAPABLE and gVNIC guest OS feature tags
    The 4.12 rhcos image meet the requirement for these tags and just need
    to be created with the above tags, see https://issues.redhat.com/browse/COS-1794.
    Until that issue is resolved you can upload the rhcos image with the
    required tags and set OPENSHIFT_INSTALL_OS_IMAGE_OVERRIDE to point to the image you created
2. The openshift/api module update should get updated back to upstream once https://github.com/openshift/api/pull/1361 merge
3. GCP enforces that OnHostMaintenance is set to "Terminate" while the Confidential VM service is enabled.


pkg/types/gcp/machinepools.go
```diff
@@ -22,6 +24,15 @@ type MachinePool struct {
 	//
 	// +optional
 	Tags []string `json:"tags,omitempty"`
+
+	// OnHostMaintenance determines the behavior when a maintenance event occurs that might cause the instance to reboot.
+	// +optional
+	OnHostMaintenance v1beta1.GCPHostMaintenanceType `json:"onHostMaintenance,omitempty"`
```
romfreiman, line 30: secure boot? disk encryption?

eranco74: disk encription is already implemented under `GCPDisk`
secure boot is something we will need to add to the GCPMachineProviderSpec

romfreiman: @patrickdillon do we even support secure boot on gcp?
@fabiendupont

data/data/gcp/cluster/master/main.tf
```diff
@@ -61,6 +61,19 @@ resource "google_compute_instance" "master" {
     kms_key_self_link = var.root_volume_kms_key_link
   }
 
+  dynamic "confidential_instance_config" {
+    for_each = var.enable_confidential_compute ? [1] : []
+    content {
+      enable_confidential_compute = true
+    }
+  }
```
patrickdillon, line 69: I'm not sure I'm following this syntax, but it seems overly complex. I suspect we don't need a dynamic block here at all. Can we simplify this to just:

```
  name         = "${var.cluster_id}-master-${count.index}"
  machine_type = var.machine_type
  zone         = element(var.zones, count.index)
  enable_confidential_compute = var.enable_confidential_compute
  on_host_maintenance = var.on_host_maintenance
```

data/data/gcp/cluster/main.tf
```diff
@@ -100,6 +100,15 @@ resource "google_compute_image" "cluster" {
...
   }
+  guest_os_features {
+    type = "GVNIC"
+  }
+  guest_os_features {
+    type = "VIRTIO_SCSI_MULTIQUEUE"
+  }
+  guest_os_features {
+    type = "SEV_CAPABLE"
+  }
```
patrickdillon, line 114: I know we talked about this in slack, but it would be good to discuss in the PR. In most, if not all cases, GCP is not creating a new compute image--instead it uses a global image specified in the coreos stream. The changes here will only be applied when terraform is creating a new image. So we need either:

1. Additional  code in the tfvars that will set `gcp_preexisting_image` to false when we want to use a confidential VM
2. To update the global coreos image and rm this commit. If we want to go this route, can you update the commit message to mark it as WIP and describe the dependency.

patrickdillon (commented):
In general this looks very good. The commit structure is good but it would be helpful to add some commit messages according to our contributing guidelines: https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#commit-message-format

data/data/gcp/cluster/main.tf
eranco74: Agree, this should be part of the default coreos image upload.
I'll remove this commit

data/data/gcp/cluster/master/main.tf
eranco74: Hmm, I added the dynamic block because the TF should look like this:
```
confidential_instance_config {
   enable_confidential_compute = var.enable_confidential_compute
}

scheduling {
   on_host_maintenance = var.on_host_maintenance
}
```
Also, I'd like to add these blocks only in case the user specified the configuration in the install config, otherwise, I'd like to  let the GCP platform choose the default

PR #6636, opened 2022-11-29
OCPBUGS-2739: Uplift terraform-provider-ironic to 0.3.0

elfosardo:
Use GO 1.18
Update to terraform SDK to 2.0
Remove go-getter dependency


r4f4:
@elfosardo This won't work on release-4.11 since we're not using Golang-1.18:
```
note: module requires Go 1.18
# github.com/zclconf/go-cty/cty/set
vendor/github.com/zclconf/go-cty/cty/set/iterator.go:3:17: syntax error: unexpected any, expecting ] 
```

elfosardo:
> @elfosardo This won't work on release-4.11 since we're not using Golang-1.18:
> 
> ```
> note: module requires Go 1.18
> # github.com/zclconf/go-cty/cty/set
> vendor/github.com/zclconf/go-cty/cty/set/iterator.go:3:17: syntax error: unexpected any, expecting ] 
> ```

@r4f4 thanks for the  comment, I've noticed from the errors that indeed we're on go 1.17 in 4.11 and earlier
this precludes the possibility to fix this CVE related bug as go 1.18 is required to run some libraries
how's the issue has been treated for other providers to remove go-getter dependency in 4.11 and older versions?

elfosardo:
@r4f4 great, thanks!

elfosardo:
this was already fixed as stated above

PR #6637, opened 2022-11-29
OCPBUGS-2124: network discovery,cluster path issue

jcpowermac:
- Added a check in tfvars if zonal then don't run legacy section to grab the network mob id

- Added a check in GetClusterNetworks if cluster is a path. If it use the cluster string directly, if not
generate the path from the datacenter and cluster name.


rna-afk:
LGTM. golint is failing on the MustCompile line.

jcpowermac:
@rna-afk can you please approve

PR #6643, opened 2022-11-30
Fix import order using the gci tool

r4f4:
The imports are ordered in 5 sections:
- standard lib
- default imports
- prefix(github.com/openshift)
- blank imports (_ <import_path>)
- dot imports (. <import_path>)

Generated with:
```
find . -name '*.go' ! -path '*/vendor/*' ! -path '*/.build/*' -exec ~/go/bin/gci write -s standard -s default -s "prefix(github.com/openshift)" -s blank -s dot --skip-generated {} \+
```


patrickdillon:
I prefer this ordering over #6644. 

I think #6644 looks better for large import blocks, e.g. 
6644: https://github.com/openshift/installer/blob/c39bb1c0540fae5560ca51dc9c3b0b7cebfa9f85/pkg/asset/machines/worker.go
vs
https://github.com/openshift/installer/blob/1ad08e527986588207ddcfa327afe839d5203dac/pkg/asset/machines/worker.go


but for small import blocks, which I suspect are more common, I think this approach makes more sense than 6644: https://github.com/openshift/installer/blob/c39bb1c0540fae5560ca51dc9c3b0b7cebfa9f85/pkg/asset/machines/vsphere/machinesets_test.go

And overall I think this approach is simpler, and easier to understand...

r4f4:
> I prefer this ordering over #6644.
> 
> I think #6644 looks better for large import blocks, e.g. 6644: https://github.com/openshift/installer/blob/c39bb1c0540fae5560ca51dc9c3b0b7cebfa9f85/pkg/asset/machines/worker.go vs https://github.com/openshift/installer/blob/1ad08e527986588207ddcfa327afe839d5203dac/pkg/asset/machines/worker.go
> 
> but for small import blocks, which I suspect are more common, I think this approach makes more sense than 6644: https://github.com/openshift/installer/blob/c39bb1c0540fae5560ca51dc9c3b0b7cebfa9f85/pkg/asset/machines/vsphere/machinesets_test.go
> 
> And overall I think this approach is simpler, and easier to understand...

@patrickdillon Got it. We can simplify it even further by not separating `_` and `.` imports into their own sections. Although I like the idea of having `_` separated, since they are used for the side effects.

So we can make it 4 sections:

-   standard lib
-   default imports
-   prefix(github.com/openshift)
-   blank imports (_ <import_path>)

Or 3 sections only:

-   standard lib
-   default imports
-   prefix(github.com/openshift)

barbacbd (commented):
The linter looks like it moved the imports that have a `_` before the package. Should we keep these imports in the same order they were before or should they be grouped with other packages that are only brought in for their side-effects ?

r4f4:
> The linter looks like it moved the imports that have a `_` before the package. Should we keep these imports in the same order they were before or should they be grouped with other packages that are only brought in for their side-effects ?

I'm not opposed to 3 sections, although I do like all `_` grouped for their meaning and the eye-pleasing effect.

r4f4:
Fixed the docs, added `gci` to the `tools.go` file so it can be vendored, added the gci command to `hack/go-fmt.sh` and made some changes to appease shellcheck.

r4f4:
/override ci/prow/e2e-gcp-ovn
Install and e2e succeeded, job timed out during deprovision after 4h

r4f4:
/override ci/prow/golint
Ran it locally with longer timeout, no problems reported

patrickdillon:
/override ci/prow/golint
based on https://github.com/openshift/installer/pull/6643#issuecomment-1351561052

r4f4:
Install succeeded, e2e timed out.

r4f4:
/override ci/prow/e2e-gcp-ovn
Install succeeded, just a few e2e failures

r4f4:
Install succeeded, timed out during e2e after 4h

r4f4:
It has installed successfully before.

PR #6646, opened 2022-12-01
OCPBUGS-4355: Fix return value from execute()

bfournie:
The execute() function was not returning an error properly, which allowed failures of the 'oc' command, such was when using an older version of 'oc' that did not support the --icsp-file param.

### DIFF
--- a/pkg/asset/agent/image/oc.go
+++ b/pkg/asset/agent/image/oc.go
@@ -272,7 +272,7 @@ func execute(log logrus.FieldLogger, executer executer.Executer, pullSecret, com
 		return strings.TrimSpace(stdout), nil
 	}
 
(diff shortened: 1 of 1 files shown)


zaneb:
/retitle AGENT-452: Fix return value from execute()

We should probably also use the real global logger instead of [creating a new one](https://github.com/openshift/installer/blob/2698cbb0ec7e96433a958ab6b864786c0c503c0b/pkg/asset/agent/image/baseiso.go#L109) with the default config that ignores the --log-level flag and prints weird `[0001]` stuff in the output for some reason. (The NMStateConfig manifests logging suffers from the same problem.)

celebdor:
/retitle OCPBUGS-4355: Fix return value from execute()

pawanpinjarkar (approved):
/approve

PR #6647, opened 2022-12-01
[release-4.11] OCPBUGS-4325: gcp: fail during validation if service usage is not enabled

r4f4:
We were only printing a warning in case Service Usage was missing and proceeding with the installation. However, that service is required by the CCO and the deployment would end up failing later on.

Instead, let's fail during install-config validation time so users can enable the service before proceeding.

### DIFF
--- a/pkg/asset/installconfig/gcp/validation.go
+++ b/pkg/asset/installconfig/gcp/validation.go
@@ -2,12 +2,12 @@ package gcp
 
 import (
 	"context"
-	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


r4f4:
/cc @gpei 
This is a continuation of https://github.com/openshift/installer/pull/6642. It's a separate PR because this could not be automatically cherry-picked.

r4f4:
> @r4f4 thanks for the notification, added the cherry-pick-approved label.

Thank you

PR #6649, opened 2022-12-01
OCPBUGS-4125: Enable CVO unmanage overrides in bootstrap-in-place installations

omertuc:
# Background

#5258 introduced a mechanism to tell CVO to not manage some cluster resources,
as that management was creating races between bootkube and CVO.

In #5271, this mechanism was removed from bootstrap-in-place
installations.

The reason for the removal was because attempting to apply
`original_cvo_overrides.patch` after the cluster-bootstrap podman
container was done wasn't working, as in BIP by the time that that
container is down, there's no longer a running kube-apiserver apply the
patch on (unlike regular installations with a bootstrap node where the
API server keeps running on the actual nodes). The justification for
simply disabling the mechanism instead of finding a solution was based
on an incorrect assumption that "CVO doesn't even run in BIP pre-pivot
so there's no need to perform overrides anyway as there's no CVO to race
with".

# Issue

Naturally, since CVO *is* actually running in the pre-pivot phase of
bootstrap-in-place, not applying CVO overrides to disable the CVO
management of bootkube-created resources led to the same races #5258 was
aiming to solve in bootstrap-in-place installations.

One manifestation of that is OCPBUGS-4125.

# Solution

This commit re-introduces the mechanism that was originally introduced in #5258
(and removed in #5271).

To solve the issue that motivated #5271, this commit doesn't attempt to
apply the `original_cvo_overrides.patch` in bootstrap-in-place
pre-reboot, and instead will transfer `original_cvo_overrides.patch` through
`master-update.fcc` to the post-pivot phase and the bootstrap-in-place
post-reboot script will make sure to apply the patch after the API
server is running again.

### DIFF
--- a/data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
+++ b/data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
@@ -63,6 +63,18 @@ function wait_for_cvo {
   done
 }
 
+function restore_cvo_overrides {
+    echo "Restoring CVO overrides"
+    until \
+        oc patch clusterversion.config.openshift.io version \
+            --type=merge \
+            --patch-file=/opt/openshift/original_cvo_overrides.patch
+    do
+        echo "Trying again to restore CVO overrides ..."
+        sleep 10
+    done
+}
+
 function clean {
   if [ -d "/etc/kubernetes/bootstrap-secrets" ]; then
      rm -rf /etc/kubernetes/bootstrap-*
@@ -77,6 +89,7 @@ function clean {
 
 wait_for_api
 signal_bootstrap_complete
+restore_cvo_overrides
 approve_csr
 restart_kubelet
 wait_for_cvo
--- a/data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/master-update.fcc
+++ b/data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/master-update.fcc
@@ -45,6 +45,9 @@ storage:
       contents:
         local: bin/installer-gather.sh
       mode: 0555
+    - path: /opt/openshift/original_cvo_overrides.patch
+      contents:
+        local: original_cvo_overrides.patch
 systemd:
   units:
     - name: bootkube.service
--- a/pkg/asset/ignition/bootstrap/bootstrap.go
+++ b/pkg/asset/ignition/bootstrap/bootstrap.go
@@ -2,7 +2,6 @@ package bootstrap
 
 import (
 	"github.com/openshift/installer/pkg/asset"
-	"github.com/openshift/installer/pkg/asset/ignition"
 )
 
 const (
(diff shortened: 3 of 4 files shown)


omertuc:
CCing everyone involved in the discussions:

/cc @eranco74 
/cc @vrutkovs 
/cc @wking 
/cc @tsorya 
/cc @romfreiman

vrutkovs (approved):

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -63,6 +63,18 @@ function wait_for_cvo {
...
 
+function restore_cvo_overrides {
+    echo "Restoring CVO overrides"
+    until \
+        oc patch clusterversion.config.openshift.io version \
+            --type=merge \
+            --patch-file=/opt/openshift/original_cvo_overrides.patch
+    do
+        echo "Trying again to restore CVO overrides ..."
+        sleep 10
```
wking, line 74: `bootkube.sh` uses `*.done` markers and expects the whole unit to get another pass if an intermediate command fails.  This `bootstrap-in-place-post-reboot.sh` looks like it prefers wrapping each command that has a reasonable chance of failing in its own retry loop.  No need to try to unify approaches in this pull request, but I'm just pointing out that the divergent approaches surprised me, in case anyone want to poke at motivation there in follow-up work.

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/master-update.fcc
```diff
@@ -45,6 +45,10 @@ storage:
       contents:
         local: bin/installer-gather.sh
       mode: 0555
+    - path: /opt/openshift/original_cvo_overrides.patch
+      contents:
+        local: original_cvo_overrides.patch
+      mode: 0555
```
wking, line 51: nit: no need for the executable bit on this `*.patch` file.  Looks like the precedent here is for `*.sh` to get the executable bit (makes sense), but for non-executables to leave off `mode` entirely, so perhaps we can just drop this line?

omertuc: Right, just copy paste mistake

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
omertuc: Yeah it would be best if we used the marker approach here as well. Not just for consistency, but because the marker approach is better as it ensures that when the entire script goes down (for whatever reason. shouldn't happen in any obvious happy path but could still theoretically happen) and then gets restarted, it doesn't repeat what it has already done

pkg/asset/ignition/bootstrap/bootstrap.go
```diff
@@ -16,29 +15,13 @@ type Bootstrap struct {
...
 		return err
 	}
 
-	// replace cvo-overrides.yaml so that CVO does not try to create resources for the manifests that the installer
-	// is supplying to cluster-bootstrap
-	cvoIgnore := &CVOIgnore{}
-	dependencies.Get(cvoIgnore)
-	for _, file := range ignition.FilesFromAsset(rootDir, "root", 0644, cvoIgnore) {
-		a.Config.Storage.Files = replaceOrAppend(a.Config.Storage.Files, file)
-	}
```
patrickdillon, line 40: Hm... I understand that this code was introduced in #5271 and that PR was created to skip cvo overrides for bootstrap in place, but it is unclear to me how this specific code relates to bootstrap in place. I can spend some more time on trying to grok this tomorrow, but I wanted to mention it in case it is obvious to anyone else and they might be able to educate me.

At first blush, it seems to me like we might still need this code and only the changes in bootkube.sh need to be addressed. I hope my question/observation makes sense.

eranco74: You are right, we still need it, it just moved to [pkg/asset/ignition/bootstrap/common.go](https://github.com/openshift/installer/pull/6649/files#diff-2a27cf9f710e504a0f8b3710fa10d24e403ff92e781cc8acbb425571427cc51bR518).
It seems that now bootstrap/bootstrap.go has nothing to add on top of bootstrap/common.go and we can do some refactoring (delete bootstrap.go, rename common.go to bootstrap.go and `Common` to `Bootstrap`, and embedding `Bootstrap` in `SingleNodeBootstrapInPlace` (instead of Common), that might make things clearer but will require the reverse refactoring the next time bootstrap and bootstrap in place diverge

eranco74 (commented):
/lgtm

pkg/asset/ignition/bootstrap/bootstrap.go
patrickdillon: Thanks for clarifying that common.go already had the equivalent code.

patrickdillon (commented):
/approve

romfreiman:
what's going on here?

patrickdillon:
> what's going on here?

vsphere postsubmit job is failing. We can override it, but the retests will often lose the override.

PR #6662, opened 2022-12-03
OCPBUGS-4367: Fix missing debug messages when getting baseISO

bfournie:
A new logger is currently created for the baseISO and oc commands. Since the default log level is not debug, this results in the messages not being output. It is cleaner to use the global logger which prints the debug log messages by default.

In addition, a log message was added for a failure running the oc command due to the --icspFile option as this indicates that an older oc binary is being used which doesn't support this option.

### DIFF
--- a/pkg/asset/agent/manifests/nmstateconfig.go
+++ b/pkg/asset/agent/manifests/nmstateconfig.go
@@ -180,7 +180,9 @@ func (n *NMStateConfig) Load(f asset.FileFetcher) (bool, error) {
 		nmStateConfigList = append(nmStateConfigList, &nmStateConfig)
 	}
 
-	staticNetworkConfigGenerator := staticnetworkconfig.New(logrus.WithField("pkg", "manifests"), staticnetworkconfig.Config{MaxConcurrentGenerations: 2})
+	log := logrus.New()
+	log.Level = logrus.WarnLevel
+	staticNetworkConfigGenerator := staticnetworkconfig.New(log.WithField("pkg", "manifests"), staticnetworkconfig.Config{MaxConcurrentGenerations: 2})
 
 	// Validate the network config using nmstatectl
 	if err = staticNetworkConfigGenerator.ValidateStaticConfigParams(context.Background(), staticNetworkConfig); err != nil {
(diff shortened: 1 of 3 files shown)


zaneb:
Two more to fix:

* https://github.com/openshift/installer/blob/master/pkg/asset/agent/manifests/nmstateconfig.go#L183
* https://github.com/openshift/installer/blob/master/pkg/asset/agent/manifests/nmstateconfig.go#L271

bfournie:
> Two more to fix:
> 
> * https://github.com/openshift/installer/blob/master/pkg/asset/agent/manifests/nmstateconfig.go#L183
> * https://github.com/openshift/installer/blob/master/pkg/asset/agent/manifests/nmstateconfig.go#L271

Fixed

pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -180,9 +180,7 @@ func (n *NMStateConfig) Load(f asset.FileFetcher) (bool, error) {
 		nmStateConfigList = append(nmStateConfigList, &nmStateConfig)
 	}
 
-	log := logrus.New()
-	log.Level = logrus.WarnLevel
```
zaneb, line 184: I'm remembering now the below call is to a library that logs a bunch of stuff at INFO level that we don't want to see: https://github.com/openshift/installer/commit/166434db40d0d399b336f1994d2677f7b59da8f8
We might need to save and restore the level.

bfournie: Yeah now we get:
```
INFO[0000] Start configuring static network for 5 hosts  pkg=manifests
INFO[0000] Adding NMConnection file <eno1.nmconnection>  pkg=manifests
INFO[0000] Adding NMConnection file <enp1s0.nmconnection>  pkg=manifests
INFO[0001] Adding NMConnection file <enp1s0.nmconnection>  pkg=manifests
INFO[0002] Adding NMConnection file <enp1s0.nmconnection>  pkg=manifests
INFO[0002] Adding NMConnection file <enp1s0.nmconnection>  pkg=manifests
```

Using this doesn't seem to work in hiding those Info messages though, I may need to revert it to use the prior way.
```
        level := logrus.GetLevel()
        logrus.SetLevel(logrus.WarnLevel)
        staticNetworkConfigGenerator := staticnetworkconfig.New(logrus.WithField("pkg", "manifests"), staticnetworkconfig.Config{MaxConcurrentGenerations: 2})
        logrus.SetLevel(level)
```

zaneb: Oh, the logging is done by the actual calls to the returned structure, not by `New()`. So we need to do `defer logrus.SetLevel(level)` instead.

bfournie: Yeah using defer works, thanks. Updated.

PR #6665, opened 2022-12-05
Better logic to derive vpcRegion/Zone from vpcName/Subnets

miyamotoh:
This change would allow users to specify which region to create a new or locate an existing VPC instance and its subnet. With this, users would no longer be tied to the installer-picked region for VPC resources that is hardcoded to be close to the PowerVS region, essentially allowing reuse of an existing VPC infrastructure the user always has, regardless of where the PowerVS instance is located.

The relevant params for VPC resources in question are; 1) `vpcRegion`, 2) `vpcName`, and 3) `vpcSubnets`.

Whether `publish: External` (default) or `publish: Internal`, when `vpcRegion` is specified, the VPC instance the installer works with will be in that region. If `vpcName` is specified additionally, a VPC instance with that name has to be found in that region. If not, one will be created in that region. 

If `vpcSubnets` is specified additionally, the subnet must be found in and attached to the VPC specified by `vpcName`. Custom `vpcSubnets` without `vpcName` will not be supported and thus error outcl. 

If `vpcName` is provided without `vpcRegion`, the VPC by the name will be searched in the nearby region of the PowerVS region, and would be accepted/reused if found, error out otherwise.

The `validation[_test].go` changes would validate all these combos of `vpc*` params.

### DIFF
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -4,10 +4,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
+	"github.com/IBM/vpc-go-sdk/vpcv1"
 	igntypes "github.com/coreos/ignition/v2/config/v3_2/types"
 	coreosarch "github.com/coreos/stream-metadata-go/arch"
 	"github.com/ghodss/yaml"
@@ -30,6 +33,7 @@ import (
 	aztypes "github.com/openshift/installer/pkg/asset/installconfig/azure"
 	gcpconfig "github.com/openshift/installer/pkg/asset/installconfig/gcp"
 	ovirtconfig "github.com/openshift/installer/pkg/asset/installconfig/ovirt"
+	powervsconfig "github.com/openshift/installer/pkg/asset/installconfig/powervs"
 	vsphereconfig "github.com/openshift/installer/pkg/asset/installconfig/vsphere"
 	"github.com/openshift/installer/pkg/asset/machines"
 	"github.com/openshift/installer/pkg/asset/manifests"
@@ -819,6 +823,46 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			masterConfigs[i] = m.Spec.ProviderSpec.Value.Object.(*machinev1.PowerVSMachineProviderConfig)
 		}
 
+		client, err := powervsconfig.NewClient()
+		if err != nil {
+			return err
+		}
+		var (
+			vpcRegion, vpcZone string
+		)
+		vpcName := installConfig.Config.PowerVS.VPCName
+		if vpcName != "" {
+			var vpc *vpcv1.VPC
+			vpc, err = client.GetVPCByName(ctx, vpcName)
+			if err != nil {
+				return err
+			}
+			var crnElems = strings.SplitN(*vpc.CRN, ":", 8)
+			vpcRegion = crnElems[5]
+		} else {
+			specified := installConfig.Config.PowerVS.VPCRegion
+			if specified != "" {
+				if powervs.ValidateVPCRegion(specified) {
+					vpcRegion = specified
+				} else {
+					return errors.New("unknown VPC region")
+				}
+			} else if vpcRegion, err = powervs.VPCRegionForPowerVSRegion(installConfig.Config.PowerVS.Region); err != nil {
+				return err
+			}
+		}
+		if vpcSubnet != "" {
+			var sn *vpcv1.Subnet
+			sn, err = client.GetSubnetByName(ctx, vpcSubnet, vpcRegion)
+			if err != nil {
+				return err
+			}
+			vpcZone = *sn.Zone.Name
+		} else {
+			rand.Seed(time.Now().UnixNano())
+			vpcZone = fmt.Sprintf("%s-%d", vpcRegion, rand.Intn(2)+1) //nolint:gosec // we don't need a crypto secure number
+		}
+
 		osImage := strings.SplitN(string(*rhcosImage), "/", 2)
 		data, err = powervstfvars.TFVars(
 			powervstfvars.TFVarsSources{
@@ -831,7 +875,9 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 				ImageBucketName:      osImage[0],
 				ImageBucketFileName:  osImage[1],
 				NetworkName:          installConfig.Config.PowerVS.PVSNetworkName,
-				VPCName:              installConfig.Config.PowerVS.VPCName,
+				VPCRegion:            vpcRegion,
+				VPCZone:              vpcZone,
+				VPCName:              vpcName,
 				VPCSubnetName:        vpcSubnet,
 				VPCPermitted:         vpcPermitted,
 				VPCGatewayName:       vpcGatewayName,
--- a/pkg/asset/installconfig/powervs/client.go
+++ b/pkg/asset/installconfig/powervs/client.go
@@ -36,6 +36,7 @@ type API interface {
 	GetAuthenticatorAPIKeyDetails(ctx context.Context) (*iamidentityv1.APIKey, error)
 	GetAPIKey() string
 	SetVPCServiceURLForRegion(ctx context.Context, region string) error
+	GetVPCs(ctx context.Context, region string) ([]vpcv1.VPC, error)
 }
 
 // Client makes calls to the PowerVS API.
@@ -515,4 +516,22 @@ func (c *Client) GetAuthenticatorAPIKeyDetails(ctx context.Context) (*iamidentit
 // GetAPIKey returns the PowerVS API key
 func (c *Client) GetAPIKey() string {
 	return c.APIKey
+}
+
+// GetVPCs gets all VPCs in a region.
+func (c *Client) GetVPCs(ctx context.Context, region string) ([]vpcv1.VPC, error) {
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+
+	err := c.SetVPCServiceURLForRegion(ctx, region)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to set vpc api service url")
+	}
+
+	vpcs, _, err := c.vpcAPI.ListVpcs(c.vpcAPI.NewListVpcsOptions())
+	if err != nil {
+		return nil, err
+	}
+
+	return vpcs.Vpcs, nil
 }
--- a/pkg/asset/installconfig/powervs/validation.go
+++ b/pkg/asset/installconfig/powervs/validation.go
@@ -7,6 +7,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"github.com/openshift/installer/pkg/types"
+	powervstypes "github.com/openshift/installer/pkg/types/powervs"
 )
 
 // Validate executes platform specific validation/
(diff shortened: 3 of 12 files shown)


pkg/asset/cluster/tfvars.go
```diff
@@ -62,6 +66,7 @@ import (
 	"github.com/openshift/installer/pkg/types/openstack"
 	"github.com/openshift/installer/pkg/types/ovirt"
 	"github.com/openshift/installer/pkg/types/powervs"
+	powervstypes "github.com/openshift/installer/pkg/types/powervs"
```
barbacbd, line 69: This is the same import as the line above. Can we consolidate them?

pkg/asset/installconfig/powervs/client.go
```diff
@@ -515,3 +516,24 @@ func (c *Client) GetAuthenticatorAPIKeyDetails(ctx context.Context) (*iamidentit
 func (c *Client) GetAPIKey() string {
 	return c.APIKey
 }
+
+// GetVPCs gets all VPCs in a region
+func (c *Client) GetVPCs(ctx context.Context, region string) ([]vpcv1.VPC, error) {
+	_, cancel := context.WithTimeout(ctx, 1*time.Minute)
```
barbacbd, line 522: The installer recreates the context:
`ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)`

pkg/asset/installconfig/powervs/client.go
```diff
@@ -515,3 +516,24 @@ func (c *Client) GetAuthenticatorAPIKeyDetails(ctx context.Context) (*iamidentit
...
+	err := c.SetVPCServiceURLForRegion(ctx, region)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to set vpc api service url")
+	}
+
+	allVPCs := []vpcv1.VPC{}
+	if vpcs, detailedResponse, err := c.vpcAPI.ListVpcs(c.vpcAPI.NewListVpcsOptions()); err != nil {
+		if detailedResponse.GetStatusCode() != http.StatusNotFound {
+			return nil, err
+		}
```
barbacbd, line 534: An error occurred, is there a reason to skip it here ?

miyamotoh: That check came from IBM Cloud code, and turns out it's rather misleading (giving the idea of potentially skipping), and in fact, it virtually never happens. So, I'm dropping the check in the next force-push. Thanks!!

miyamotoh:
All checks passed. @barbacbd PTAL.

pkg/asset/installconfig/powervs/validation.go
```diff
@@ -109,3 +110,75 @@ func validatePreExistingPrivateDNS(fldPath *field.Path, client API, ic *types.In
...
+		}
+	}
+	if !found {
+		allErrs = append(allErrs, field.NotFound(path.Child("vpcName"), name))
+	}
+
+	return allErrs
+}
+
+func findSubnetinVPC(client API, subnets []string, region string, name string, path *field.Path) field.ErrorList {
```
barbacbd, line 169: nit: The i in In should be capitalized

pkg/asset/installconfig/powervs/client.go
```diff
@@ -515,3 +516,22 @@ func (c *Client) GetAuthenticatorAPIKeyDetails(ctx context.Context) (*iamidentit
...
+
+	err := c.SetVPCServiceURLForRegion(ctx, region)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to set vpc api service url")
+	}
+
+	allVPCs := []vpcv1.VPC{}
+	if vpcs, _, err := c.vpcAPI.ListVpcs(c.vpcAPI.NewListVpcsOptions()); err != nil {
+		return nil, err
+	} else if vpcs != nil {
```
barbacbd, line 533: technically this else if is not needed.

pkg/asset/cluster/tfvars.go
```diff
@@ -819,6 +824,46 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
+		}
+		if vpcSubnet != "" {
+			var sn *vpcv1.Subnet
+			sn, err = client.GetSubnetByName(ctx, vpcSubnet, vpcRegion)
+			if err != nil {
+				return err
+			}
+			vpcZone = *sn.Zone.Name
+		} else {
+			rand.Seed(time.Now().UnixNano())
```
barbacbd, line 863: I am not familiar with powervs but looking at this I see some potentials for errors. What if there are other options for sub nets. Should we search for valid subnets and if you still want to use a random, randomly select from the list of valid subnets ?

clnperez: this is the way we've done it for a while. we (@mjturek and I) made a lot of time making sure it wasn't going to cause problems. however, we know it's not ideal not randomly select the vpc zone this way. but it's fine for now :)

miyamotoh: Not officially documented, but this knowledge of "valid zones" with the number 1 to 3 as suffix per region is gained from internal discussion with the IBM Cloud team. Therefore, this bit is basically allowing user to reuse his/her existing subnet, making sure it exists and deriving the zone from it, if specified, or we just pick one randomly from existing ones for the installer.

pkg/asset/installconfig/powervs/validation.go
miyamotoh: Addressed in the latest force-push.

pkg/asset/installconfig/powervs/client.go
miyamotoh: Addressed in the latest force-push.

barbacbd (commented):
/lgtm

pkg/asset/cluster/tfvars.go
barbacbd: thanks for the explanations !

r4f4:
Install succeeded. Timed out during test after 4h.

r4f4:
/override ci/prow/golint
Ran the linter locally with a longer timeout, no problems reported.

pkg/asset/installconfig/powervs/client.go
```diff
@@ -516,3 +517,21 @@ func (c *Client) GetAuthenticatorAPIKeyDetails(ctx context.Context) (*iamidentit
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to set vpc api service url")
+	}
+
+	allVPCs := []vpcv1.VPC{}
+	if vpcs, _, err := c.vpcAPI.ListVpcs(c.vpcAPI.NewListVpcsOptions()); err != nil {
+		return nil, err
+	}
+	allVPCs = append(allVPCs, vpcs.Vpcs...)
+	return allVPCs, nil
```
r4f4, line 536: pkg/asset/installconfig/powervs/client.go:532:5: vpcs declared but not used
pkg/asset/installconfig/powervs/client.go:535:28: undefined: vpcs 

`vpcs` does not live beyound the scope of `if`.

```suggestion
	vpcs, _, err := c.vpcAPI.ListVpcs(c.vpcAPI.NewListVpcsOptions())
	if err != nil {
		return nil, err
	}
	return vpcs.Vpcs, nil
```

miyamotoh: Whoa, sorry about that. I thought I had force-pushed my latest commit, but apparently not. With that now pushed, it passes `go-lint`, `go-fmt`, and `go-vet` (tested locally). Let's see...

pkg/asset/installconfig/powervs/client.go
```diff
@@ -516,3 +517,23 @@ func (c *Client) GetAuthenticatorAPIKeyDetails(ctx context.Context) (*iamidentit
...
+	}
+
+	vpcs, _, err := c.vpcAPI.ListVpcs(c.vpcAPI.NewListVpcsOptions())
+	if err != nil {
+		return nil, err
+	}
+
+	allVPCs := []vpcv1.VPC{}
+	allVPCs = append(allVPCs, vpcs.Vpcs...)
+	return allVPCs, nil
```
r4f4, line 538:
```suggestion
	return vpcs.Vpcs, nil
```
Any reason to create a new slice instead of returning the existing one?

miyamotoh: Will be addressed in the next force-push.

pkg/asset/cluster/tfvars.go
r4f4: It seems the double import was brought back.

pkg/asset/cluster/tfvars.go
```diff
@@ -819,6 +824,46 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
+			vpcZone = fmt.Sprintf("%s-%d", vpcRegion, rand.Intn(2)+1)
```
r4f4, line 864:
```suggestion
			vpcZone = fmt.Sprintf("%s-%d", vpcRegion, rand.Intn(2)+1) //nolint:gosec // we don't need a crypto secure number
```

pkg/asset/installconfig/powervs/client.go
```diff
@@ -516,3 +517,23 @@ func (c *Client) GetAuthenticatorAPIKeyDetails(ctx context.Context) (*iamidentit
 func (c *Client) GetAPIKey() string {
 	return c.APIKey
 }
+
+// GetVPCs gets all VPCs in a region
```
r4f4, line 521:
```suggestion
// GetVPCs gets all VPCs in a region.
```

pkg/asset/installconfig/powervs/validation.go
```diff
@@ -110,3 +111,75 @@ func validatePreExistingPrivateDNS(fldPath *field.Path, client API, ic *types.In
 	}
 	return allErrs
 }
+
+// ValidateCustomVPCSetup ensures optional VPC settings, if specified, are all legit
```
r4f4, line 115:
```suggestion
// ValidateCustomVPCSetup ensures optional VPC settings, if specified, are all legit.
```

pkg/types/powervs/powervs_regions.go
```diff
@@ -97,3 +97,15 @@ func RegionShortNames() []string {
 	}
 	return keys
 }
+
+// ValidateVPCRegion validates that given VPC region is known/tested
```
r4f4, line 101:
```suggestion
// ValidateVPCRegion validates that given VPC region is known/tested.
```

pkg/types/powervs/platform.go
```diff
@@ -17,15 +17,15 @@ type Platform struct {
...
+
 	// VPCRegion specifies the IBM Cloud region in which to create VPC resources.
 	// Leave unset to allow installer to select the closest VPC region.
 	//
 	// +optional
 	VPCRegion string `json:"vpcRegion,omitempty"`
 
-	// UserID is the login for the user's IBM Cloud account.
-	UserID string `json:"userID"`
-
```
r4f4, line 28: Any reason for this change? If so, it should go in its own commit.

pkg/asset/cluster/tfvars.go
r4f4, line 69: Linter is still complaining about this double import.

r4f4: There was a recent commit adding linters to the config, so you might not have it if you haven't rebased on top of master recently

r4f4: Instead of removing the previous import, I'd suggest removing the one you're introducing and replacing `powervstypes` by `powervs` in your code. The build is broken right now.

r4f4:
Install and e2e succeeded, timed out during gather

r4f4:
/override ci/prow/e2e-azure-ovn ci/prow/e2e-gcp-ovn
Install succeeded, timed out after 4h.
/hold cancel

r4f4:
Unrelated e2e failures

PR #6666, opened 2022-12-05
OCPBUGS-1769: Check for AWS STS installation before trying to get all IAM Roles

sadasu:
In the AWS cluster destroy/uninstall path, the Installer tries to find all the IAM Roles in the cluster and then attempts to delete the resources with the tag `kubernetes.io/cluster/<cluster-name>`. In the case of STS clusters, all IAM Roles are cleared outside the cluster (not by the Installer) and even trying to find them in the cluster results in errors because the Installer does not have the privileges to do that, let alone deleting them.

The goal of the fix is not attempt to delete these IAM Roles that Installer does not have access to.

### DIFF
--- a/pkg/destroy/aws/iamhelpers.go
+++ b/pkg/destroy/aws/iamhelpers.go
@@ -13,6 +13,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
+const (
+	ErrCodeAccessDeniedException = "AccessDeniedException"
+)
+
 type iamRoleSearch struct {
 	client    *iam.IAM
 	filters   []Filter
@@ -39,13 +43,21 @@ func (search *iamRoleSearch) find(ctx context.Context) (arns []string, names []s
 				// Unfortunately role.Tags is empty from ListRoles, so we need to query each one
 				response, err := search.client.GetRoleWithContext(ctx, &iam.GetRoleInput{RoleName: role.RoleName})
 				if err != nil {
-					if err.(awserr.Error).Code() == iam.ErrCodeNoSuchEntityException {
-						search.unmatched[*role.Arn] = exists
-					} else {
-						if lastError != nil {
-							search.logger.Debug(lastError)
+					var awsErr awserr.Error
+					if errors.As(err, &awsErr) {
+						switch awsErr.Code() {
+						case ErrCodeAccessDeniedException, iam.ErrCodeNoSuchEntityException:
+							// Installer does not have access to this IAM role or the
+							// the role does not exist.
+							// Ignore this IAM Role and donot report this error via
+							// lastError
+							search.unmatched[*role.Arn] = exists
+						default:
+							if lastError != nil {
+								search.logger.Debug(lastError)
+							}
+							lastError = errors.Wrapf(err, "get tags for %s", *role.Arn)
 						}
-						lastError = errors.Wrapf(err, "get tags for %s", *role.Arn)
 					}
 				} else {
 					role = response.Role


pkg/destroy/aws/aws.go
```diff
@@ -289,12 +294,14 @@ func (o *ClusterUninstaller) findResourcesToDelete(
 		}
 	}
 
-	// Find IAM roles
-	iamRoleResources, err := findIAMRoles(ctx, iamRoleSearch, deleted, o.Logger)
-	if err != nil {
-		errs = append(errs, err)
+	// Find IAM roles if this is not an STS cluster
+	if o.CredentialsMode != string(types.ManualCredentialsMode) {
```
r4f4, line 298: `CredentialsMode` can be set to `Manual` without implying STS use: https://docs.openshift.com/container-platform/4.11/installing/installing_aws/manually-creating-iam.html#manually-creating-iam-aws. Do we also want to skip deleting IAM roles in that case?

pkg/destroy/aws/aws.go
```diff
@@ -289,12 +294,14 @@ func (o *ClusterUninstaller) findResourcesToDelete(
...
+		iamRoleResources, err := findIAMRoles(ctx, iamRoleSearch, deleted, o.Logger)
+		if err != nil {
```
r4f4, line 300: Instead of skipping the finding of IAM roles entirely, we could also check the error for `AccessDenied` and ignore it in that case, either here or inside `findIAMRoles` since it's the `GetRoleWithContext` call that is failing.

patrickdillon: Also, the installer can still create IAM roles for the control plane nodes in manual mode: https://github.com/openshift/installer/blob/master/data/data/aws/cluster/master/main.tf#L18

Manual mode specifically disables IAM role creation by the cloud credential operator. So if we gate based on manual mode we will be leaking the IAM roles created by the installer.

pkg/destroy/aws/iamhelpers.go
```diff
@@ -39,6 +41,13 @@ func (search *iamRoleSearch) find(ctx context.Context) (arns []string, names []s
 				// Unfortunately role.Tags is empty from ListRoles, so we need to query each one
 				response, err := search.client.GetRoleWithContext(ctx, &iam.GetRoleInput{RoleName: role.RoleName})
 				if err != nil {
+					if quotaaws.IsUnauthorized(err) {
+						// Installer does not have access to this IAM role
+						// Ignore this IAM Role and donot report this error via
+						// lastError
+						search.unmatched[*role.Arn] = exists
+						continue
+					}
```
r4f4, line 50: It looks a bit weird to import a `quota` package to check if an error is `AccessDenied`. Either `IsUnauthorized` should be moved out of the quota package so it can be shared with multiple packages, or we should just have our own `isUnauthorized` function in this file. Actually, I think this whole block would look better with a `switch`

```suggestion
                    var awsErr awserr.Error
                    if errors.As(err, &awsErr) {
                        switch awsErr.Code() {
                        case "AccessDeniedException":
						    // Installer does not have access to this IAM role
						    // Ignore this IAM Role and do not report this error via
						    // lastError
						    search.logger.Debugf("AccessDenied to role %s. Expected if this is an STS install", *role.Arn)
						    fallthrough
					    case iam.ErrCodeNoSuchEntityException:
						    search.unmatched[*role.Arn] = exists
						    continue
						default:
						}
			        }
					if lastError != nil {
					    search.logger.Debug(lastError)
					}
					lastError = errors.Wrapf(err, "get tags for %s", *role.Arn)
				}
```

sadasu: It would be nice to have one helper function which identifies unauthorized access for any AWS API's return code and that was the reason behind using the pre-existing method (unfortunately it exists in quota/aws).  
Yes, this lends itself to using a switch statement.  
I am concerned if logging a statement for the AccessDenied case would make the logs noisy hence stayed away from it. Happy to add it if it provides value to the customer (in non STS cases maybe?)

patrickdillon:
/approve

Overall approach looks good. I agree that we could simplify the error checking a bit: we can just string match on the unauthorized error without bringing in another package.

pkg/destroy/aws/iamhelpers.go
```diff
@@ -13,6 +13,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
+const (
+	ErrCodeAccessDeniedException = "AccessDeniedException"
```
r4f4, line 17: Nit: probably doesn't need to be exported outside the package

r4f4 (commented):
Just one nitpicking comment but not worth blocking the merge on it.
/lgtm

PR #6670, opened 2022-12-06
OCPBUGS-11721:  Enable cluster destroy in AWS us-iso and us-isob regions

dmc5179:
Remove the block on cluster destroy in the us-iso and us-isob regions now that the AWS tagging API has been GA'd by AWS for these regions. I have built the openshift-installer code base with these changes and tested them in both the us-iso and us-isob regions. With these changes, cluster destroy works in these regions as expected of any other AWS region.

### DIFF
--- a/pkg/destroy/aws/aws.go
+++ b/pkg/destroy/aws/aws.go
@@ -28,7 +28,6 @@ import (
 	awssession "github.com/openshift/installer/pkg/asset/installconfig/aws"
 	"github.com/openshift/installer/pkg/destroy/providers"
 	"github.com/openshift/installer/pkg/types"
-	awstypes "github.com/openshift/installer/pkg/types/aws"
 	"github.com/openshift/installer/pkg/version"
 )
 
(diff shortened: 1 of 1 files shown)


montaguethomas:
@dmc5179 Happy to see this change! Wanted to call out that you're missing part of the fix to support C2S isolated regions. We've been using the following patch enable support and have tested it on both `us-iso` and `us-isob` partitions.

https://github.com/openshift/installer/commit/dbb4be7f91466c589da011ad29565d690cfcc733

andykrohg:
> @dmc5179 Happy to see this change! Wanted to call out that you're missing part of the fix to support C2S isolated regions. We've been using the following patch enable support and have tested it on both `us-iso` and `us-isob` partitions.
> 
> [montaguethomas@dbb4be7](https://github.com/montaguethomas/openshift-installer/commit/dbb4be7f91466c589da011ad29565d690cfcc733)

Hi! 👋🏼  I'm working with some folks who'd like to see this change get merged too. Is the holdup on this one that the only invocation of `isSecretRegion()` is being removed and the function definition is orphaned?

montaguethomas:
Noting I had opened another PR with additional change required. This PR is incomplete to support destroy in AWS isolated partitions: https://github.com/openshift/installer/pull/7086/files

r4f4:
/close
in favor of https://github.com/openshift/installer/pull/7086

PR #6003, opened 2022-06-14
Azure: use azidentity with adapter

r4f4:
Microsoft is in the process of upgrading the Azure SDK for Go from V1 to V2. Our clients and authentication are on V1. The V1 authentication utilizes ADAL which will be deprecated June 30, 2022. All V2 clients, except the V2 auth client azidentity, are in beta (azidentity is scheduled to be stable in Q2 2022). [0]
    
These changes remove the dependency on the ADAL API, replace the authentication with azidentity, and use an adapter so the auth will work with V1 clients.
    
 [0] https://azure.github.io/azure-sdk/releases/latest/index.html#go
    
 https://issues.redhat.com/browse/CORS-1910

PR #6614, opened 2022-11-22
OCPBUGS-4549: azure: replace deprecated AD Graph API

r4f4:
Use the new MS Graph API instead.
    
Please note that `Applications.ReadWrite.OwnedBy` permissions must be granted in Azure for MSGraph.
    
https://github.com/azure-deprecation/dashboard/issues/60
    
https://issues.redhat.com/browse/CORS-1897

PR #6671, opened 2022-12-06
[release-4.12] OCPBUGS-4542: Azure: use azidentity with an adapter

r4f4:
This is a manual cherry-pick of https://github.com/openshift/installer/pull/6003


barbacbd (commented):
This looks good, but I thought we try to not go through with major version updates

```
github.com/Azure/azure-sdk-for-go/sdk/internal v1.0.1
```

r4f4:
> This looks good, but I thought we try to not go through with major version updates
> 
> ```
> github.com/Azure/azure-sdk-for-go/sdk/internal v1.0.1
> ```

In this case we have to since Stack and GovCloud support are present in the official release (v1.0). `azure-sdk-for-go/sdk` is where the new SDK V2 clients live and we're not yet using any of them. However that's about to change when we backport https://github.com/openshift/installer/pull/6614 to replace `graphrbac`  (also being deprecated) with the new `msgraphsdk` API.

r4f4:
/hold
Not sure yet if we want this sooner or later

patrickdillon:
Given that usptream kubernetes still depends on ADAL in 4.12, I don't think we should backport this. Contrary to the messaging regarding ADAL EOL at the end of this month, it seems like we should begin to think about this as future proofing instead.

PR #6679, opened 2022-12-07
GCP: Tfvars will determine the user has create firewall permissions

barbacbd:
** Create Firewall permissions option is removed from the install config. The terraform vars checks the user permissions in the project to determine if the user can create firewall rules. If the user does not have permissions to create firewall rules then the firewall rules are skipped.

### DIFF
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -66,6 +66,10 @@ import (
 )
 
 const (
+	// GCPFirewallPermission is the role/permission to create or skip the creation of
+	// firewall rules for GCP during an xpn installation.
+	GCPFirewallPermission = "compute.firewalls.create"
+
 	// TfVarsFileName is the filename for Terraform variables.
 	TfVarsFileName = "terraform.tfvars.json"
 
@@ -407,6 +411,22 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			ServiceAccount:   string(sess.Credentials.JSON),
 		}
 
+		// In the case of a shared vpn, the firewall rules should only be created if the user has permissions to do so
+		createFirewallRules := true
+		if installConfig.Config.GCP.NetworkProjectID != "" {
+			client, err := gcpconfig.NewClient(context.Background())
+			if err != nil {
+				return err
+			}
+			permissions, err := client.GetProjectPermissions(context.Background(), installConfig.Config.GCP.NetworkProjectID, []string{
+				GCPFirewallPermission,
+			})
+			if err != nil {
+				return err
+			}
+			createFirewallRules = permissions.Has(GCPFirewallPermission)
+		}
+
 		masters, err := mastersAsset.Machines()
 		if err != nil {
 			return err
@@ -424,11 +444,6 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			workerConfigs[i] = w.Spec.Template.Spec.ProviderSpec.Value.Object.(*machinev1beta1.GCPMachineProviderSpec)
 		}
 		preexistingnetwork := installConfig.Config.GCP.Network != ""
-
-		createFirewallRules := true
-		if installConfig.Config.GCP.CreateFirewallRules == gcp.CreateFirewallRulesDisabled {
-			createFirewallRules = false
-		}
 
 		// Setup defaults for public dns zone
 		createPublicZoneRecords := true
--- a/pkg/asset/installconfig/gcp/client.go
+++ b/pkg/asset/installconfig/gcp/client.go
@@ -13,9 +13,16 @@ import (
 	dns "google.golang.org/api/dns/v1"
 	"google.golang.org/api/option"
 	"google.golang.org/api/serviceusage/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 //go:generate mockgen -source=./client.go -destination=./mock/gcpclient_generated.go -package=mock
+
+var (
+	// RequiredBasePermissions is the list of permissions required for an installation.
+	// A list of valid permissions can be found at https://cloud.google.com/iam/docs/understanding-roles.
+	RequiredBasePermissions = []string{}
+)
 
 // API represents the calls made to the API.
 type API interface {
@@ -31,6 +38,8 @@ type API interface {
 	GetZones(ctx context.Context, project, filter string) ([]*compute.Zone, error)
 	GetEnabledServices(ctx context.Context, project string) ([]string, error)
 	GetCredentials() *googleoauth.Credentials
+	GetProjectPermissions(ctx context.Context, project string, permissions []string) (sets.String, error)
+	ValidateServiceAccountHasPermissions(ctx context.Context, project string, permissions []string) (bool, error)
 }
 
 // Client makes calls to the GCP API.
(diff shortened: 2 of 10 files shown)


pkg/asset/cluster/gcp/permissions.go
```diff
@@ -0,0 +1,50 @@
...
+	service, err := api.NewService(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create resource manager service")
+	}
+
+	projectsService := api.NewProjectsService(service)
+	rb := &api.TestIamPermissionsRequest{Permissions: searchPermissions}
+	response, err := projectsService.TestIamPermissions(project, rb).Context(ctx).Do()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to retrieve permissions for project %s", project)
```
patrickdillon, line 40: I think this code (creation of the service and the API call) should live in https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/gcp/client.go and follow the patterns there. It's unclear to me how the existing code is authenticating, but moving to client.go the pattern with using the session credentials should work fine.

pkg/asset/cluster/gcp/permissions.go
```diff
@@ -0,0 +1,50 @@
...
+// value where the permission either exists or does not for the Service account in regard
+// to a specific project
+type CloudPermissions struct {
+	CreateFirewallRules bool `json:"compute.firewalls.create"`
+}
+
+// ValidatePermissions will pass a set of permissions to the cloud resource manager api and
+// determine which permissions were found from the original list. When a permission is not listed
+// in the returned data, it is not a permission of the current user in the supplied project.
+func ValidatePermissions(project string) (*CloudPermissions, error) {
```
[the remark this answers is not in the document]
barbacbd: I was putting together a model of what we do for AWS. AWS appears to have a list of permissions that have to be met. Do we want this to be configurable?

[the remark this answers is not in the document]
patrickdillon: It would be good and it seems like not too much trouble to keep in mind the use case for general permissions checking when designing this function. The general requirement seems to be the same for both cases, so essentially one function should be able to handle both. The `CloudPermissions` data structure is an interesting one and it's not so obvious to me how this would work if we were comparing a larger set of permissions. 

The permissions that are returned by the `TestIAMPermissions` are in a `[]string`, which could lend itself to converting to a set and using [difference](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#String.Difference) to compare with the original.

pkg/asset/installconfig/gcp/client.go
```diff
@@ -343,3 +352,56 @@ func (c *Client) getServiceUsageService(ctx context.Context) (*serviceusage.Serv
...
+	permissionsMap := make(map[string]bool)
+	for _, perm := range permissions {
+		permissionsMap[perm] = false
+	}
+
+	for _, validPerm := range validPermissions {
+		permissionsMap[validPerm] = true
+	}
+
+	logrus.Warnf("%+v", permissionsMap)
```
r4f4, line 394: Maybe
```suggestion
	logrus.Debugf("%+v", permissionsMap)
```
is more appropriate? Or is this temporary just to check it in CI?

pkg/asset/installconfig/gcp/client.go
```diff
@@ -343,3 +352,56 @@ func (c *Client) getServiceUsageService(ctx context.Context) (*serviceusage.Serv
...
+	}
+
+	return response.Permissions, nil
+
+}
+
+// GetProjectPermissions will consume a set of permissions as strings and return
+// a map of of whether or not the permission was found for the service account in the provided project.
+// A list of valid permissions can be found at https://cloud.google.com/iam/docs/understanding-roles.
+func (c *Client) GetProjectPermissions(ctx context.Context, project string, permissions []string) (map[string]bool, error) {
```
r4f4, line 379: Any reason to work with a `map` and not a `set`?

barbacbd: a map provides faster lookup when trying to determine what is there and what isn't. Technically we probably will never see the difference in results.

pkg/asset/installconfig/gcp/client.go
barbacbd: This was temporary and actually supposed to have been removed already ! I used Warnf as it alerts in Yellow and is easier to find, but this can be removed.

pkg/asset/cluster/tfvars.go
```diff
@@ -407,6 +407,22 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
+		if err != nil {
+			return err
+		}
+		permissions, err := client.GetProjectPermissions(context.Background(), credsProject, []string{
+			"compute.firewalls.create",
+		})
+		if err != nil {
+			return err
+		}
+		createFirewallRules := permissions.Has("create.firewalls.create")
```
r4f4, line 424: Shouldn't this be
```suggestion
		createFirewallRules := permissions.Has("compute.firewalls.create")
```

pkg/asset/installconfig/gcp/client.go
```diff
@@ -13,10 +13,17 @@ import (
...
 	"google.golang.org/api/serviceusage/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
 //go:generate mockgen -source=./client.go -destination=./mock/gcpclient_generated.go -package=mock
 
+var (
+	// RequiredBasePermissions is the list of permissions required for an installation.
+	// A list of valid permissions can be found at https://cloud.google.com/iam/docs/understanding-roles.
+	RequiredBasePermissions = []string{}
```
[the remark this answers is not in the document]
barbacbd: technically it doesn't need to be exported but I thought that someone may want to use/access it in the future. This is really about a future feature. This is not modified by the API call. The API consumes this list and returns a new list of the permissions that were allowed.

pkg/asset/cluster/tfvars.go
```diff
@@ -407,6 +409,22 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
+		if err != nil {
+			return err
+		}
+		permissions, err := client.GetProjectPermissions(context.Background(), credsProject, []string{
+			tfVarsGCPFirewallPermission,
+		})
+		if err != nil {
+			return err
+		}
+		createFirewallRules := permissions.Has(tfVarsGCPFirewallPermission)
```
patrickdillon, line 426: We should only be doing this check (and only disabling createFirewallRules) for XPN installs. 

It would be good to explain that we're doing it because the user may not have permissions in the host project... either in a comment here or in the commit message

pkg/asset/installconfig/gcp/client.go
```diff
@@ -343,3 +352,44 @@ func (c *Client) getServiceUsageService(ctx context.Context) (*serviceusage.Serv
...
+	defer cancel()
+
+	service, err := c.getCloudResourceService(ctx)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get cloud resource manager service")
+	}
+
+	projectsService := cloudresourcemanager.NewProjectsService(service)
+	rb := &cloudresourcemanager.TestIamPermissionsRequest{Permissions: permissions}
+	response, err := projectsService.TestIamPermissions(project, rb).Context(ctx).Do()
```
patrickdillon, line 367: I wonder if `TestIamPermissions` requires a new permission in order to run?

patrickdillon (commented):
A couple of small comments, but we need to fix this so it only runs the check for an XPN install, otherwise default to create firewalls

pkg/asset/installconfig/gcp/client.go
barbacbd: According to https://cloud.google.com/resource-manager/reference/rest/v1/projects/testIamPermissions, there are no permissions required to run this API call

patrickdillon:
lgtm'ing in addition to approving, based on rafael's earlier lgtm. minimal changes since that previous review.

pkg/asset/installconfig/gcp/client.go
```diff
@@ -343,3 +352,44 @@ func (c *Client) getServiceUsageService(ctx context.Context) (*serviceusage.Serv
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get Iam permissions")
+	}
+
+	return response.Permissions, nil
+
```
r4f4, line 373: The linter doesn't like this extra newline here

PR #6684, opened 2022-12-08
OCPBUGS-4654: azure: upi: use Image Gallery in ARM templates

r4f4:
When we moved to using Image Galleries in Azure/IPI, we also changed the
location that MAO uses for the rhcos bootimages. These changes update
our ARM templates so that the bootimages are picked from the right
location. It fixes the following error when scaling compute nodes:

```
Error Message: failed to reconcile machine "maxu-upi2-gc7n8-worker-eastus3-68gdx": failed to create vm maxu-upi2-gc7n8-worker-eastus3-68gdx: failure sending request for machine maxu-upi2-gc7n8-worker-eastus3-68gdx: cannot create vm: compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=404 -- Original Error: Code="GalleryImageNotFound" Message=""The gallery image /subscriptions/$sub_id/resourceGroups/$resource_group/providers/Microsoft.Compute/galleries/$gallery_name/images/maxu-upi2-gc7n8-gen2/versions/412.86.20220930 is not available in eastus region. Please contact image owner to replicate to this region, or change your requested region."" Target="imageReference"
```

It is also one step closer to getting Azure UPI for aarch64.

### DIFF
--- a/docs/user/azure/install_upi.md
+++ b/docs/user/azure/install_upi.md
@@ -195,8 +195,9 @@ Given the size of the RHCOS VHD, it's not possible to run the deployments with t
 We must copy and store it in a storage container instead. To do so, first create a blob storage container and then copy the VHD.
 
 ```sh
+export OCP_ARCH="x86_64" # or "aarch64"
 az storage container create --name vhd --account-name ${CLUSTER_NAME}sa
-export VHD_URL=$(openshift-install coreos print-stream-json | jq -r '.architectures.x86_64."rhel-coreos-extensions"."azure-disk".url')
+export VHD_URL=$(openshift-install coreos print-stream-json | jq -r --arg arch "$OCP_ARCH" '.architectures[$arch]."rhel-coreos-extensions"."azure-disk".url')
 az storage blob copy start --account-name ${CLUSTER_NAME}sa --account-key $ACCOUNT_KEY --destination-blob "rhcos.vhd" --destination-container vhd --source-uri "$VHD_URL"
 ```
 
@@ -288,11 +289,15 @@ Create the deployment using the `az` client:
 
 ```sh
 export VHD_BLOB_URL=`az storage blob url --account-name ${CLUSTER_NAME}sa --account-key $ACCOUNT_KEY -c vhd -n "rhcos.vhd" -o tsv`
+export STORAGE_ACCOUNT_ID=`az storage account show -g ${RESOURCE_GROUP} --name ${CLUSTER_NAME}sa --query id -o tsv`
+export AZ_ARCH=`echo $OCP_ARCH | sed 's/x86_64/x64/;s/aarch64/Arm64/'`
 
 az deployment group create -g $RESOURCE_GROUP \
   --template-file "02_storage.json" \
   --parameters vhdBlobURL="$VHD_BLOB_URL" \
-  --parameters baseName="$INFRA_ID"
+  --parameters baseName="$INFRA_ID" \
+  --parameters storageAccount="${CLUSTER_NAME}sa" \
+  --parameters architecture="$AZ_ARCH"
 ```
 
 ## Deploy the load balancers
--- a/upi/azure/02_storage.json
+++ b/upi/azure/02_storage.json
@@ -1,41 +1,153 @@
 {
-  "$schema" : "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
-  "contentVersion" : "1.0.0.0",
-  "parameters" : {
-    "baseName" : {
-      "type" : "string",
-      "minLength" : 1,
-      "metadata" : {
-        "description" : "Base name to be used in resource names (usually the cluster's Infra ID)"
+  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "architecture": {
+      "type": "string",
+      "metadata": {
+        "description": "The architecture of the Virtual Machines"
+      },
+      "defaultValue": "x64",
+      "allowedValues": [
+        "Arm64",
+        "x64"
+      ]
+    },
+    "baseName": {
+      "type": "string",
+      "minLength": 1,
+      "metadata": {
+        "description": "Base name to be used in resource names (usually the cluster's Infra ID)"
       }
     },
-    "vhdBlobURL" : {
-      "type" : "string",
-      "metadata" : {
-        "description" : "URL pointing to the blob where the VHD to be used to create master and worker machines is located"
+    "storageAccount": {
+      "type": "string",
+      "metadata": {
+        "description": "The Storage Account name"
+      }
+    },
+    "vhdBlobURL": {
+      "type": "string",
+      "metadata": {
+        "description": "URL pointing to the blob where the VHD to be used to create master and worker machines is located"
       }
     }
   },
-  "variables" : {
-    "location" : "[resourceGroup().location]",
-    "imageName" : "[concat(parameters('baseName'), '-image')]"
+  "variables": {
+    "location": "[resourceGroup().location]",
+    "galleryName": "[concat('gallery_', replace(parameters('baseName'), '-', '_'))]",
+    "imageName": "[parameters('baseName')]",
+    "imageNameGen2": "[concat(parameters('baseName'), '-gen2')]",
+    "imageRelease": "1.0.0"
   },
-  "resources" : [
+  "resources": [
     {
-      "apiVersion" : "2018-06-01",
-      "type": "Microsoft.Compute/images",
-      "name": "[variables('imageName')]",
-      "location" : "[variables('location')]",
-      "properties": {
-        "storageProfile": {
-          "osDisk": {
-            "osType": "Linux",
+      "apiVersion": "2021-10-01",
+      "type": "Microsoft.Compute/galleries",
+      "name": "[variables('galleryName')]",
+      "location": "[variables('location')]",
+      "resources": [
+        {
+          "apiVersion": "2021-10-01",
+          "type": "images",
+          "name": "[variables('imageName')]",
+          "location": "[variables('location')]",
+          "dependsOn": [
+            "[variables('galleryName')]"
+          ],
+          "properties": {
+            "architecture": "[parameters('architecture')]",
+            "hyperVGeneration": "V1",
+            "identifier": {
+              "offer": "rhcos",
+              "publisher": "RedHat",
+              "sku": "basic"
+            },
             "osState": "Generalized",
-            "blobUri": "[parameters('vhdBlobURL')]",
-            "storageAccountType": "Standard_LRS"
-          }
+            "osType": "Linux"
+          },
+          "resources": [
+            {
+              "apiVersion": "2021-10-01",
+              "type": "versions",
+              "name": "[variables('imageRelease')]",
+              "location": "[variables('location')]",
+              "dependsOn": [
+                "[variables('imageName')]"
+              ],
+              "properties": {
+                "publishingProfile": {
+                  "storageAccountType": "Standard_LRS",
+                  "targetRegions": [
+                    {
+                      "name": "[variables('location')]",
+                      "regionalReplicaCount": "1"
+                    }
+                  ]
+                },
+                "storageProfile": {
+                  "osDiskImage": {
+                    "source": {
+                      "id": "[resourceId('Microsoft.Storage/storageAccounts', parameters('storageAccount'))]",
+                      "uri": "[parameters('vhdBlobURL')]"
+                    }
+                  }
+                }
+              }
+            }
+          ]
+        },
+        {
+          "apiVersion": "2021-10-01",
+          "type": "images",
+          "name": "[variables('imageNameGen2')]",
+          "location": "[variables('location')]",
+          "dependsOn": [
+            "[variables('galleryName')]"
+          ],
+          "properties": {
+            "architecture": "[parameters('architecture')]",
+            "hyperVGeneration": "V2",
+            "identifier": {
+              "offer": "rhcos-gen2",
+              "publisher": "RedHat-gen2",
+              "sku": "gen2"
+            },
+            "osState": "Generalized",
+            "osType": "Linux"
+          },
+          "resources": [
+            {
+              "apiVersion": "2021-10-01",
+              "type": "versions",
+              "name": "[variables('imageRelease')]",
+              "location": "[variables('location')]",
+              "dependsOn": [
+                "[variables('imageNameGen2')]"
+              ],
+              "properties": {
+                "publishingProfile": {
+                  "storageAccountType": "Standard_LRS",
+                  "targetRegions": [
+                    {
+                      "name": "[variables('location')]",
+                      "regionalReplicaCount": "1"
+                    }
+                  ]
+                },
+                "storageProfile": {
+                  "osDiskImage": {
+                    "source": {
+                      "id": "[resourceId('Microsoft.Storage/storageAccounts', parameters('storageAccount'))]",
+                      "uri": "[parameters('vhdBlobURL')]"
+                    }
+                  }
+                }
+              }
+            }
+          ]
         }
-      }
+      ]
     }
   ]
 }
--- a/upi/azure/04_bootstrap.json
+++ b/upi/azure/04_bootstrap.json
@@ -36,6 +36,17 @@
       "metadata" : {
         "description" : "The size of the Bootstrap Virtual Machine"
       }
+    },
+    "hyperVGen": {
+      "type": "string",
+      "metadata": {
+        "description": "VM generation image to use"
+      },
+      "defaultValue": "V2",
+      "allowedValues": [
+        "V1",
+        "V2"
+      ]
     }
   },
   "variables" : {
@@ -51,7 +62,7 @@
     "vmName" : "[concat(parameters('baseName'), '-bootstrap')]",
     "nicName" : "[concat(variables('vmName'), '-nic')]",
     "galleryName": "[concat('gallery_', replace(parameters('baseName'), '-', '_'))]",
-    "imageName" : "[parameters('baseName')]",
+    "imageName" : "[concat(parameters('baseName'), if(equals(parameters('hyperVGen'), 'V2'), '-gen2', ''))]",
     "clusterNsgName" : "[concat(if(not(empty(parameters('vnetBaseName'))), parameters('vnetBaseName'), parameters('baseName')), '-nsg')]",
     "sshPublicIpAddressName" : "[concat(variables('vmName'), '-ssh-pip')]"
   },
--- a/upi/azure/04_bootstrap.json
+++ b/upi/azure/04_bootstrap.json
@@ -50,7 +50,8 @@
     "identityName" : "[concat(parameters('baseName'), '-identity')]",
     "vmName" : "[concat(parameters('baseName'), '-bootstrap')]",
     "nicName" : "[concat(variables('vmName'), '-nic')]",
-    "imageName" : "[concat(parameters('baseName'), '-image')]",
+    "galleryName": "[concat('gallery_', replace(parameters('baseName'), '-', '_'))]",
+    "imageName" : "[parameters('baseName')]",
     "clusterNsgName" : "[concat(if(not(empty(parameters('vnetBaseName'))), parameters('vnetBaseName'), parameters('baseName')), '-nsg')]",
     "sshPublicIpAddressName" : "[concat(variables('vmName'), '-ssh-pip')]"
   },
@@ -132,7 +133,7 @@
         },
         "storageProfile" : {
           "imageReference": {
-            "id": "[resourceId('Microsoft.Compute/images', variables('imageName'))]"
+            "id": "[resourceId('Microsoft.Compute/galleries/images', variables('galleryName'), variables('imageName'))]"
           },
           "osDisk" : {
             "name": "[concat(variables('vmName'),'_OSDisk')]",
--- a/upi/azure/05_masters.json
+++ b/upi/azure/05_masters.json
@@ -58,6 +58,17 @@
       "metadata" : {
         "description" : "Size of the Master VM OS disk, in GB"
       }
+    },
+    "hyperVGen": {
+      "type": "string",
+      "metadata": {
+        "description": "VM generation image to use"
+      },
+      "defaultValue": "V2",
+      "allowedValues": [
+        "V1",
+        "V2"
+      ]
     }
   },
   "variables" : {
@@ -71,7 +82,7 @@
     "sshKeyPath" : "/home/core/.ssh/authorized_keys",
     "identityName" : "[concat(parameters('baseName'), '-identity')]",
     "galleryName": "[concat('gallery_', replace(parameters('baseName'), '-', '_'))]",
-    "imageName" : "[parameters('baseName')]",
+    "imageName" : "[concat(parameters('baseName'), if(equals(parameters('hyperVGen'), 'V2'), '-gen2', ''))]",
     "copy" : [
       {
         "name" : "vmNames",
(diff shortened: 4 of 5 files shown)


docs/user/azure/install_upi.md
```diff
@@ -288,11 +288,15 @@ Create the deployment using the `az` client:
 
 ```sh
 export VHD_BLOB_URL=`az storage blob url --account-name ${CLUSTER_NAME}sa --account-key $ACCOUNT_KEY -c vhd -n "rhcos.vhd" -o tsv`
+export RELEASE_VERSION=`openshift-install coreos print-stream-json | jq -r '.architectures.x86_64."rhel-coreos-extensions"."azure-disk".release'`
```
Prashanth684, line 291: probably also good to call out that you specify this based on architecture: `architectures.x86_64` or `architectures.aarch64`.

r4f4: I was thinking about doing that as a separate PR to add `aarch64` support.

r4f4:
CI Azure UPI jobs are expected to fail until https://github.com/openshift/release/pull/34678 is merged.

upi/azure/02_storage.json
```diff
@@ -1,41 +1,159 @@
...
   },
-  "variables" : {
-    "location" : "[resourceGroup().location]",
-    "imageName" : "[concat(parameters('baseName'), '-image')]"
+  "variables": {
+    "location": "[resourceGroup().location]",
+    "galleryName": "[concat('gallery_', replace(parameters('baseName'), '-', '_'))]",
+    "imageName": "[parameters('baseName')]",
+    "imageNameGen2": "[concat(parameters('baseName'), '-gen2')]",
+	"imageVersion": "[take(parameters('imageRelease'), len(parameters('imageRelease'))-6)]"
```
Prashanth684, line 47: i'm now thinking if this should not be done here but we should give control to the user. like IPI, we are not making the decision here, but it depends on who is using the template. sure, if they use the release version, this is valid, but a user may choose differently and then it might break.

r4f4:
/hold
We need https://github.com/openshift/release/pull/34678 merged first so this can be tested in CI

upi/azure/02_storage.json
```diff
@@ -1,41 +1,159 @@
...
+            "[variables('galleryName')]"
+          ],
+          "properties": {
+            "architecture": "[parameters('architecture')]",
+            "hyperVGeneration": "V2",
+            "identifier": {
+              "offer": "rhcos-gen2",
+              "publisher": "RedHat-gen2",
+              "sku": "gen2"
+            },
```
MayXuQQ, line 121: just use the market-place image?

docs/user/azure/install_upi.md
r4f4: I decided to just do it in this PR.

upi/azure/02_storage.json
r4f4: Users are free to edit the templates, but if you think there is more value in exposing this choice, I can change it. 

It had to be like this at first because of MAO expectations from https://github.com/openshift/installer/blob/master/pkg/asset/machines/azure/machines.go#L120 but with https://github.com/openshift/installer/pull/6694 we're now moving to using `latest` in the VM, so the exact value of this variable doesn't matter anymore.

upi/azure/02_storage.json
Prashanth684: marketplace images are not supported for arm64

docs/user/azure/install_upi.md
```diff
@@ -288,11 +289,17 @@ Create the deployment using the `az` client:
 
 ```sh
 export VHD_BLOB_URL=`az storage blob url --account-name ${CLUSTER_NAME}sa --account-key $ACCOUNT_KEY -c vhd -n "rhcos.vhd" -o tsv`
+export RELEASE_VERSION=`openshift-install coreos print-stream-json | jq -r '.architectures.$OCP_ARCH."rhel-coreos-extensions"."azure-disk".release'`
+export STORAGE_ACCOUNT_ID=`az storage account show -g ${RESOURCE_GROUP} --name ${CLUSTER_NAME}sa --query id -o tsv`
```
MayXuQQ, line 293: | jq -r ".architectures.$OCP_ARCH.\"rhel-coreos-extensions\".\"azure-disk\".release"

upi/azure/02_storage.json
```diff
@@ -1,41 +1,159 @@
...
       }
     },
-    "vhdBlobURL" : {
-      "type" : "string",
-      "metadata" : {
-        "description" : "URL pointing to the blob where the VHD to be used to create master and worker machines is located"
+    "storageAccountId": {
+      "type": "string",
+      "metadata": {
+        "description": "The ID of the Storage Account"
```
MayXuQQ, line 32: whether can use storage account name, instead of the id ?

docs/user/azure/install_upi.md
r4f4: Tks, I need to update this. This is how I did it in CI https://github.com/openshift/release/pull/34678/files#diff-5483a633ffe74be033f28def4e9b0c9adbfb0ec96b53e43ef314e22672b64ae9R162

upi/azure/02_storage.json
r4f4: It just so happens we can! I have updated the code.

upi/azure/02_storage.json
r4f4: I changed it to use the `imageRelease` directly and dropped the `imageVersion` variable.

docs/user/azure/install_upi.md
r4f4: Fixed

upi/azure/05_masters.json
```diff
@@ -70,7 +70,8 @@
     "internalLoadBalancerName" : "[concat(parameters('baseName'), '-internal-lb')]",
     "sshKeyPath" : "/home/core/.ssh/authorized_keys",
     "identityName" : "[concat(parameters('baseName'), '-identity')]",
-    "imageName" : "[concat(parameters('baseName'), '-image')]",
+    "galleryName": "[concat('gallery_', replace(parameters('baseName'), '-', '_'))]",
+    "imageName" : "[parameters('baseName')]",
```
MayXuQQ, line 74: how about use the previous image name "[concat(parameters('baseName'), '-image')]" ?

r4f4: We can but then we also need to update https://github.com/openshift/installer/blob/master/pkg/asset/machines/azure/machines.go#L120. I made it consistent with the terraform changes https://github.com/openshift/installer/blob/master/data/data/azure/vnet/main.tf#L101-L130 but have no strong feelings about it.

MayXuQQ:
now the VM is generate V1, shall we add parameter to support V2 VM ?

upi/azure/06_workers.json
```diff
@@ -44,6 +44,17 @@
       "metadata" : {
         "description" : "The size of the each Node Virtual Machine"
       }
+    },
+    "hyperVGen": {
+      "type": "string",
+      "metadata": {
+        "description": "VM generation image to use"
+      },
+      "defaultValue": "V1",
```
MayXuQQ, line 53: the defaultValue is V1 or V2 ?  Now IPI default value is V2.

r4f4: IPI default to V2 if the VM supports it. Unfortunately, we cannot check which HyperVGen the VM supports in the ARM templates, so I'm defaulting to "V1" to keep it consistent with the previous behaviour.

docs/user/azure/install_upi.md
```diff
@@ -288,11 +289,17 @@ Create the deployment using the `az` client:
...
+export RELEASE_VERSION=`openshift-install coreos print-stream-json | jq -r --arg arch "$OCP_ARCH" '.architectures[$arch]."rhel-coreos-extensions"."azure-disk".release'`
+export STORAGE_ACCOUNT_ID=`az storage account show -g ${RESOURCE_GROUP} --name ${CLUSTER_NAME}sa --query id -o tsv`
+export AZ_ARCH=`echo $OCP_ARCH | sed 's/x86_64/x64/;s/aarch64/Arm64/'`
 
 az deployment group create -g $RESOURCE_GROUP \
   --template-file "02_storage.json" \
   --parameters vhdBlobURL="$VHD_BLOB_URL" \
-  --parameters baseName="$INFRA_ID"
+  --parameters baseName="$INFRA_ID" \
+  --parameters imageRelease="${RELEASE_VERSION::15}" \
```
jianlinliu, line 300: Why have to limit to 15 length?

r4f4: That's a good question. It's because the RHCOS version is in the format "412.86.202209302317-0" and the image version [1]:
 * Can only have numbers and periods on it
 * Each segment has to fit into a `int32`

So we use the RHCOS version up to the day (in the sample above it would be "412.86.20220930").

[1] https://learn.microsoft.com/en-us/azure/templates/microsoft.compute/2021-10-01/galleries/images/versions?pivots=deployment-language-arm-template#galleriesimagesversions-1

MayXuQQ:
- with hyperVGeneration: "V2", create cluster succeed, then scale one more worker succeed. 
```
$ az sig image-definition list -g $RG -r gallery_maxu_u12_2_hvwfj | jq '.[]|.id +" " + .hyperVGeneration'
"/subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-2-hvwfj-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_2_hvwfj/images/maxu-u12-2-hvwfj V1"
"/subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-2-hvwfj-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_2_hvwfj/images/maxu-u12-2-hvwfj-gen2 V2"
```
```
$ az vm list -g $RG | jq '.[]|.name + " " + .storageProfile.imageReference.id'
"maxu-u12-2-hvwfj-master-0 /subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-2-hvwfj-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_2_hvwfj/images/maxu-u12-2-hvwfj-gen2"
"maxu-u12-2-hvwfj-master-1 /subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-2-hvwfj-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_2_hvwfj/images/maxu-u12-2-hvwfj-gen2"
"maxu-u12-2-hvwfj-master-2 /subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-2-hvwfj-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_2_hvwfj/images/maxu-u12-2-hvwfj-gen2"
"maxu-u12-2-hvwfj-worker-eastus-1 /subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-2-hvwfj-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_2_hvwfj/images/maxu-u12-2-hvwfj-gen2"
"maxu-u12-2-hvwfj-worker-eastus-2 /subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-2-hvwfj-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_2_hvwfj/images/maxu-u12-2-hvwfj-gen2"
"maxu-u12-2-hvwfj-worker-eastus3-h26x5 /subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-2-hvwfj-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_2_hvwfj/images/maxu-u12-2-hvwfj-gen2/versions/latest"
```

- without specify hyperVGeneration create cluster succeed. and scale up worker succeed. 
```
$ az sig image-definition list -g $RG -r gallery_maxu_u12_1_h2t5r | jq '.[]|.id +" " + .hyperVGeneration'
"/subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-1-h2t5r-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_1_h2t5r/images/maxu-u12-1-h2t5r V1"
"/subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-1-h2t5r-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_1_h2t5r/images/maxu-u12-1-h2t5r-gen2 V2"
```
```
$ az vm list -g $RG | jq '.[]|.name + " " + .storageProfile.imageReference.id'
"maxu-u12-1-h2t5r-master-0 /subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-1-h2t5r-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_1_h2t5r/images/maxu-u12-1-h2t5r"
"maxu-u12-1-h2t5r-master-1 /subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-1-h2t5r-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_1_h2t5r/images/maxu-u12-1-h2t5r"
"maxu-u12-1-h2t5r-master-2 /subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-1-h2t5r-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_1_h2t5r/images/maxu-u12-1-h2t5r"
"maxu-u12-1-h2t5r-worker-eastus-1 /subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-1-h2t5r-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_1_h2t5r/images/maxu-u12-1-h2t5r"
"maxu-u12-1-h2t5r-worker-eastus-2 /subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-1-h2t5r-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_1_h2t5r/images/maxu-u12-1-h2t5r"
"maxu-u12-1-h2t5r-worker-eastus2-fvm28 /subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/maxu-u12-1-h2t5r-rg/providers/Microsoft.Compute/galleries/gallery_maxu_u12_1_h2t5r/images/maxu-u12-1-h2t5r-gen2/versions/latest"
```

docs/user/azure/install_upi.md
jianlinliu: How about introduce a function to convert the string inside ARM template, but only ask users to input the whole release version string gotten from installer. That will be more straight-forward.

upi/azure/06_workers.json
aleskandro: Hi @r4f4, V2 is the only one supported for ARM64. Which AMD64 machine types do not support V2?

Is it worth giving those machine types "more weight" than defaulting to V2 and supporting both architectures by default? 

I don't know how many they are, or their usage frequency, and it could be better to just document `if you're deploying on arm64, let's use V2` than `the following *list* of machine types needs V1`.

That's just a bit odd that we need to treat arm64 as "special" again and wanted you to reconsider the choice from this perspective as well.

docs/user/azure/install_upi.md
r4f4: That's what I did initially but then changed based on https://github.com/openshift/installer/pull/6684#discussion_r1043872859

r4f4: @Prashanth684 what do you think?

r4f4:
I have changed the code to default to HyperV Gen2, rebased on top of master and squash !fixup commits.

upi/azure/06_workers.json
r4f4: The default is now "V2".

docs/user/azure/install_upi.md
r4f4: After some discussions in Slack, we've decided to drop this parameter and use a static value instead. Asking the user to pass in the value creates some work for them with honestly very little value that we're aware of. This was a consequence of us not knowing until recently you can just use `latest` when referring to an image.

r4f4:
Updated the PR based on the review feedback: 
* use "1.0.0" for the version and remove the parameter input; 
* squash both doc commits into 1;
* default to HyperV gen 2.

`e2e-azure-ovn-upi` is expected to fail until https://github.com/openshift/release/pull/34978 merges.

upi/azure/02_storage.json
```diff
@@ -1,41 +1,153 @@
...
+        "description": "URL pointing to the blob where the VHD to be used to create master and worker machines is located"
       }
     }
   },
-  "variables" : {
-    "location" : "[resourceGroup().location]",
-    "imageName" : "[concat(parameters('baseName'), '-image')]"
+  "variables": {
+    "location": "[resourceGroup().location]",
+    "galleryName": "[concat('gallery_', replace(parameters('baseName'), '-', '_'))]",
```
MayXuQQ, line 38: How about contact 'architecture' to the galleryName? since 'arm64' and 'amd64' workers can be added to a cluster.  @r4f4

r4f4: If we do that, then we need to pass the `architecture` parameter to the VM templates (bootstrap, masters, and workers) since they reference the gallery by name.

barbacbd (commented):
/lgtm

upi/azure/02_storage.json
Prashanth684: i would not do this since in the future when we support installing a multi-arch compute cluster through installer, we would be creating arm images under the same gallery. we don't want to be creating a separate gallery for arm instances/x86 instances. maybe the image inside the gallery could be denoted with `-arch` suffix.

docs/user/azure/install_upi.md
jianlinliu: @r4f4 though I prefer the idea, will that become a problem when users is scaling up via machine-api in a UPI install?

upi/azure/04_bootstrap.json
```diff
@@ -50,7 +50,8 @@
     "identityName" : "[concat(parameters('baseName'), '-identity')]",
     "vmName" : "[concat(parameters('baseName'), '-bootstrap')]",
     "nicName" : "[concat(variables('vmName'), '-nic')]",
-    "imageName" : "[concat(parameters('baseName'), '-image')]",
+    "galleryName": "[concat('gallery_', replace(parameters('baseName'), '-', '_'))]",
+    "imageName" : "[parameters('baseName')]",
```
MayXuQQ, line 54: when try to create arm cluster, 
`"The selected VM size 'Standard_D4ps_v5' cannot boot Hypervisor Generation '1'. `
so the bootstrap.json need 
`    "imageName" : "[concat(parameters('baseName'), if(equals(parameters('hyperVGen'), 'V2'), '-gen2', ''))]",`

docs/user/azure/install_upi.md
r4f4: @jianlinliu It shouldn't be since we just use `latest` as the version for MAO: https://github.com/openshift/installer/blob/master/pkg/asset/machines/azure/machines.go#L120

upi/azure/02_storage.json
r4f4: Another problem with changing the gallery name is that we also need to change https://github.com/openshift/installer/blob/master/pkg/asset/machines/azure/machines.go#L120 otherwise we break scaling with MAO.

upi/azure/04_bootstrap.json
r4f4: :facepalm: how did I miss that?

r4f4:
Changed bootstrap to also default to HyperVGen2 which I missed in the previous updated.

upi/azure/04_bootstrap.json
MayXuQQ: after changed the 04_bootstrap.json, I can create UPI cluster on arm64.

MayXuQQ:
/remove-label qe-approved
when test on upi arm64_x86_heterogeneous_workers, need deploy arm and amd image together, with the same image name and same gallery name, got the following error: 

> "code":"Conflict","message":"{\r\n  \"error\": {\r\n    \"code\": \"PropertyChangeNotAllowed\",\r\n    \"message\": \"Changing property 'galleryImage.properties.architecture' is not allowed.

MayXuQQ:
how about the arm file for [azurestack]( https://github.com/openshift/installer/tree/master/upi/azurestack), whether plan use image gallery?

r4f4:
> how about the arm file for [azurestack](https://github.com/openshift/installer/tree/master/upi/azurestack), whether plan use image gallery?

I don't think so, but will confirm.

r4f4:
> /remove-label qe-approved when test on upi arm64_x86_heterogeneous_workers, need deploy arm and amd image together, with the same image name and same gallery name, got the following error:
> 
> > "code":"Conflict","message":"{\r\n  "error": {\r\n    "code": "PropertyChangeNotAllowed",\r\n    "message": "Changing property 'galleryImage.properties.architecture' is not allowed.

I'd say this is out-of-scope for OCPBUGS-4654 and I propose we do it in a separate PR since it'll involve making changes to IPI as well.

MayXuQQ:
> > /remove-label qe-approved when test on upi arm64_x86_heterogeneous_workers, need deploy arm and amd image together, with the same image name and same gallery name, got the following error:
> > > "code":"Conflict","message":"{\r\n  "error": {\r\n    "code": "PropertyChangeNotAllowed",\r\n    "message": "Changing property 'galleryImage.properties.architecture' is not allowed.
> 
> I'd say this is out-of-scope for OCPBUGS-4654 and I propose we do it in a separate PR since it'll involve making changes to IPI as well.

@aleskandro how about your suggestion? if your are agreed, I will approve this PR

aleskandro:
> > > /remove-label qe-approved when test on upi arm64_x86_heterogeneous_workers, need deploy arm and amd image together, with the same image name and same gallery name, got the following error:
> > > > "code":"Conflict","message":"{\r\n  "error": {\r\n    "code": "PropertyChangeNotAllowed",\r\n    "message": "Changing property 'galleryImage.properties.architecture' is not allowed.
> > 
> > 
> > I'd say this is out-of-scope for OCPBUGS-4654 and I propose we do it in a separate PR since it'll involve making changes to IPI as well.
> 
> @aleskandro how about your suggestion? if your are agreed, I will approve this PR

Hi @MayXuQQ , I agree with @r4f4: moreover, Azure UPI Multi-arch clusters are not supported yet.
We only run some jobs downstream to anticipate future efforts. I will take care of adapting that code according to the changes here.

barbacbd (commented):
/lgtm

r4f4:
The same tests
```
Failing tests:
[sig-network-edge][Conformance][Area:Networking][Feature:Router] The HAProxy router should pass the gRPC interoperability tests [apigroup:config.openshift.io][apigroup:route.openshift.io][apigroup:operator.openshift.io] [Suite:openshift/conformance/parallel/minimal]
[sig-network-edge][Conformance][Area:Networking][Feature:Router][apigroup:route.openshift.io][apigroup:config.openshift.io] The HAProxy router should pass the http2 tests [apigroup:image.openshift.io][apigroup:operator.openshift.io] [Suite:openshift/conformance/parallel/minimal]
```
are also perma-failing in the [periodic UPI job](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.13-e2e-azure-ovn-upi/1612794731241148416) so not introduced in this PR. So there is no point in keeping retesting.

/override ci/prow/e2e-azure-ovn-upi
/hold cancel

r4f4:
/override ci/prow/e2e-azure-ovn-upi
same 2 e2e tests failing

r4f4:
This change only touches the UPI ARM files for Azure, so IPI is not affected.

PR #6601, opened 2022-11-18
CORS-2340: hack: replace deprecated go-lint

r4f4:
Use golangci-lint instead, which is an aggregator of linters.

Resolves https://issues.redhat.com/browse/CORS-2340

PR #6688, opened 2022-12-09
OCPBUGS-3706: Wait longer for baremetal

pawanpinjarkar:


pkg/agent/waitfor.go
```diff
@@ -77,6 +81,14 @@ func WaitForInstallComplete(assetDir string) (*Cluster, error) {
 	}
 
 	timeout := 90 * time.Minute
+	// Wait longer for baremetal, due to length of time it takes to boot
+	if assetStore, err := assetstore.NewStore(assetDir); err == nil {
+		if installConfig, err := assetStore.Load(&installconfig.InstallConfig{}); err == nil && installConfig != nil {
+			if installConfig.(*installconfig.InstallConfig).Config.Platform.Name() == baremetal.Name {
+				timeout = 120 * time.Minute
```
zaneb, line 88: At least in the bug report it was timing out in wait-for bootstrap-complete, so I'm not sure this is required. 90 minutes already seems long compared to 60 minutes used in metal IPI.
I'm not convinced we have any more work to do than IPI. We do have to reboot the last control plane node and let it come up, but OTOH by that point we have already finished installing all of the workers in parallel with the other control plane hosts, whereas IPI doesn't start installing the workers until the CVO starts the baremetal-operator on the control plane.

andfasano: Agree, I'd also think that the timeout increase just for the bootstrap phase may be sufficient to cover this case

r4f4:
/override ci/prow/golint
Fix for linter timeout is on its way https://github.com/openshift/installer/pull/6601

PR #6694, opened 2022-12-13
OCPBUGS-4809: Azure: toggle image in machinesets

patrickdillon:
Hive vendors the installer and uses the asset package to generate machinesets for scaleup. Because Hive is using the latest code version but installing multiple previous versions, the machinesets--particularly the values--need to be backward compatible.

In this particular case, the installer switched from using Azure managed images to image galleries in 4.12. In 4.12+ Azure machinesets expect an image referencing an image gallery, while prior to this change the machinesets looked for a managed image.

This commit updates the machineset code to allow a toggle which will allow Hive to generate Azure machinesets utilizing managed images, which should be done with 4.11 and earlier clusters.

This change also future proofs the 4.12+ by switching the machinesets to use the latest version, rather than tying them to a particular RHCOS version.

### DIFF
--- a/pkg/asset/machines/azure/machines.go
+++ b/pkg/asset/machines/azure/machines.go
@@ -23,7 +23,7 @@ const (
 )
 
 // Machines returns a list of machines for a machinepool.
-func Machines(clusterID string, config *types.InstallConfig, pool *types.MachinePool, osImage, role, userDataSecret string, capabilities map[string]string, rhcosVersion string) ([]machineapi.Machine, error) {
+func Machines(clusterID string, config *types.InstallConfig, pool *types.MachinePool, osImage, role, userDataSecret string, capabilities map[string]string, useImageGallery bool) ([]machineapi.Machine, error) {
 	if configPlatform := config.Platform.Name(); configPlatform != azure.Name {
 		return nil, fmt.Errorf("non-Azure configuration: %q", configPlatform)
 	}
@@ -50,7 +50,7 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 		if len(azs) > 0 {
 			azIndex = int(idx) % len(azs)
 		}
-		provider, err := provider(platform, mpool, osImage, userDataSecret, clusterID, role, &azIndex, capabilities, rhcosVersion)
+		provider, err := provider(platform, mpool, osImage, userDataSecret, clusterID, role, &azIndex, capabilities, useImageGallery)
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to create provider")
 		}
@@ -82,7 +82,7 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 	return machines, nil
 }
 
-func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string, userDataSecret string, clusterID string, role string, azIdx *int, capabilities map[string]string, rhcosVersion string) (*machineapi.AzureMachineProviderSpec, error) {
+func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string, userDataSecret string, clusterID string, role string, azIdx *int, capabilities map[string]string, useImageGallery bool) (*machineapi.AzureMachineProviderSpec, error) {
 	var az *string
 	if len(mpool.Zones) > 0 && azIdx != nil {
 		az = &mpool.Zones[*azIdx]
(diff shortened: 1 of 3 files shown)


patrickdillon:
cc @2uasimojo @abutcher @dlom

pkg/asset/machines/azure/machines.go
```diff
@@ -110,15 +110,17 @@ func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string
...
 		galleryName := strings.Replace(clusterID, "-", "_", -1)
 		id := clusterID
 		if hyperVGen == "V2" {
 			id += "-gen2"
 		}
-		imageID := fmt.Sprintf("/resourceGroups/%s/providers/Microsoft.Compute/galleries/gallery_%s/images/%s/versions/%s", rg, galleryName, id, rhcosVersion)
+		imageID := fmt.Sprintf("/resourceGroups/%s/providers/Microsoft.Compute/galleries/gallery_%s/images/%s/versions/latest", rg, galleryName, id)
 		image.ResourceID = imageID
+	} else {
+		image.ResourceID = fmt.Sprintf("/resourceGroups/%s/providers/Microsoft.Compute/images/%s", rg, clusterID)
```
r4f4, line 123: You need to append `"-gen2"` here to the image name if the VM supports `hyperVGen == "V2"`

pkg/asset/machines/azure/machines.go
```diff
@@ -110,15 +110,17 @@ func provider(platform *azure.Platform, mpool *azure.MachinePool, osImage string
...
-	} else {
+	} else if useImageGallery {
 		// image gallery names cannot have dashes
 		galleryName := strings.Replace(clusterID, "-", "_", -1)
 		id := clusterID
 		if hyperVGen == "V2" {
 			id += "-gen2"
 		}
-		imageID := fmt.Sprintf("/resourceGroups/%s/providers/Microsoft.Compute/galleries/gallery_%s/images/%s/versions/%s", rg, galleryName, id, rhcosVersion)
+		imageID := fmt.Sprintf("/resourceGroups/%s/providers/Microsoft.Compute/galleries/gallery_%s/images/%s/versions/latest", rg, galleryName, id)
```
r4f4, line 120: Does using `latest` here work?

r4f4: @Prashanth684 I remember you had issues when using anything other than an actual version here

Prashanth684: yes it did - looking back at this comment: https://github.com/openshift/installer/pull/6304#discussion_r968799213, it did not like `latest` when used as a version

pkg/asset/machines/azure/machinesets.go
```diff
@@ -41,7 +41,8 @@ func MachineSets(clusterID string, config *types.InstallConfig, pool *types.Mach
 		if int64(idx) < total%numOfAZs {
 			replicas++
 		}
-		provider, err := provider(platform, mpool, osImage, userDataSecret, clusterID, role, &idx, capabilities, rhcosVersion)
+		useImageGallery := platform.CloudName != azure.StackCloud
```
Prashanth684, line 44: this "toggle" here seems to be hardcoded to say that if the platform is ASH use managed images, if not use image gallery? does this mean hive uses ASH? i don't understand how this toggle helps with backwards compatibility in this case.

pkg/asset/machines/azure/machines.go
r4f4: See `name` here https://learn.microsoft.com/en-us/azure/templates/microsoft.compute/2021-10-01/galleries/images/versions?pivots=deployment-language-arm-template. This is the ARM template but iirc the same restrictions apply 
```
Character limit: 32-bit integer

Valid characters:
Numbers and periods.
(Each segment is converted to an int32. So each segment has a max value of 2,147,483,647.)
```

patrickdillon: > yes it did - looking back at this comment: https://github.com/openshift/installer/pull/6304#discussion_r968799213, it did not like latest when used as a version

It doesn't like `latest` when creating an image gallery (version), but it is ok when specifying the image to be used.

pkg/asset/machines/azure/machinesets.go
patrickdillon: I am just learning about this myself... Within the installer, there is no backward compatibility and Azure Stack using the bool is more of a coincidence. Hive vendors the asset package and calls the function to create machinesets: https://github.com/openshift/hive/blob/33e68f241d3c784c5362b48febea14020107dd39/pkg/controller/machinepool/azureactuator.go#L139-L146

So basically they will set the bool based on whether the version is 4.12+.

pkg/asset/machines/azure/machines.go
patrickdillon: should be fixed in e26a4783415d32420778300ac5322f6953ca2511

pkg/asset/machines/azure/machinesets.go
Prashanth684: ah..ok..i see..good to know..thanks

2uasimojo:
This looks like it does what we talked about yesterday ✓
Thanks @patrickdillon!

patrickdillon:
/override ci/prow/golint 

golint job is timing out. Fix is in progress. Overriding here after running the test locally.

patrickdillon:
/hold cancel 
/override ci/prow/e2e-gcp-ovn

These actually didn't fail... hitting timeouts. Probably due to long image build times.

r4f4:
Existing linter issues are being fixed in https://github.com/openshift/installer/pull/6712 and running the linter locally for this PR looks good:
```sh
$: ~/go/bin/golangci-lint run --timeout=10m --new-from-rev=HEAD~2
$: 
```

patrickdillon:
/override ci/prow/e2e-azure-ovn ci/prow/e2e-gcp-ovn ci/prow/e2e-vsphere-ovn ci/prow/golint

Timeouts and a known vsphere issue... overriding

dlom:
@patrickdillon any chance we can get this cherry-picked to 4.12?

dlom:
let's try it... 

/cherry-pick release-4.12

PR #6697, opened 2022-12-14
OCPBUGS-3032: Report status on the console immediately

zaneb:
When updating the console debug messages, don't wait for the next refresh. Issue a reload, which will cause the console to be redisplayed unless the user has already started logging in.

Improve reporting of status so that the user can see what is happening when there are host-specific configurations that cannot be applied, and so that an installation that needs user intervention will report when it has recovered.

### DIFF
--- a/data/data/agent/files/usr/local/bin/install-status.sh
+++ b/data/data/agent/files/usr/local/bin/install-status.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# shellcheck disable=SC1091
+source "issue_status.sh"
+
 inactive_services() {
     local services="assisted-service.service create-cluster-and-infraenv.service apply-host-config.service start-cluster-installation.service"
     for s in ${services}; do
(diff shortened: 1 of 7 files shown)


bfournie:
/lgtm
Tested with successful install and saw the message transition on the console
Tested with various failures, for example by breaking the connection to quay and no mirror was defined and verified got "Unable to pull Openshift release image" message

patrickdillon:
/approve

There was a typo in my original approve post. not sure if the webhook or whatever will pick up the edit, so sending again

zaneb:
@hawkowl I guess it depends on what level of debugging you want. This stuff is about telling the user in broad terms about what's happening and what they can fix in their physical environment; it's not a substitute for us having the whole logs so we can debug issues in our code. This is also mostly targeting baremetal, where it is often challenging to read the logs as the scroll by as all you may have is a KVM switch to plug in.
FWIW CoreOS refreshes the console every 1 minute (I think) anyway, so this is just hurrying it up. The login prompt works fine with a serial console in libvirt, but I don't know anything about how Azure handles it. If there's no getty process running on the console (thus no login prompt) then the reload wouldn't have any effect.

hawkowl:
@zaneb Ah, fair. In our case, there's instances where networking and such means that bootstrap logging doesn't go anywhere (e.g. they blackhole the logging endpoints or the logging containers can't be pulled) and the machine is short lived (destroyed after failure). In that instance, the serial console is the only source we have to determine what caused it.

I guess in that case, we would need to log the output of podman, to determine why. Hmm.

zaneb:
/test ci/prow/e2e-gcp-ovn
Azure job is permafailing, but GCP might be fixed again.

PR #6698, opened 2022-12-14
MGMT-12839: Add Shielded VMs options to gcp mpool and TF

eranco74:
Add support to Shielded VMs on GCP
Feature link: https://issues.redhat.com/browse/OCPBU-158


data/data/gcp/cluster/master/main.tf
```diff
@@ -61,6 +61,15 @@ resource "google_compute_instance" "master" {
     kms_key_self_link = var.root_volume_kms_key_link
   }
 
+
+  dynamic "shielded_instance_config" {
+    for_each = var.secure_boot != "" ? [1] : []
+
+    content {
+      enable_secure_boot = var.secure_boot == Enabled ? true : false
```
barbacbd, line 69: Shouldn't Enabled be a string comparison "Enabled"?

nit: The ternary is a bit explicit, could it be dropped?

pkg/types/gcp/machinepools.go
```diff
@@ -76,6 +83,9 @@ func (a *MachinePool) Set(required *MachinePool) {
 		}
 		a.EncryptionKey.Set(required.EncryptionKey)
 	}
+	if required.SecureBoot != "" {
```
barbacbd, line 86: nit: If it happens that it is changed to default to Disabled, this will break. Should we check that this == "Enabled" rather than nothing?

pkg/types/gcp/machinepools.go
```diff
@@ -22,6 +22,13 @@ type MachinePool struct {
 	//
 	// +optional
 	Tags []string `json:"tags,omitempty"`
+
```
barbacbd, line 25: With these additions it looks like you need to run `hack/verify-codegen.sh`

barbacbd:
Looks like you need to check our terraform linting as well as golint.

patrickdillon (commented):
In general, this looks good. The commit message should be fleshed out according to our contributing guidelines. Generally I would expect this to be several commits, probably along the lines of: adding type to machine pools, add support in machinesets, and terraform; but this is small enough to be manageable as is.

pkg/types/gcp/machinepools.go
eranco74: The current default is disabled.
Checking for the empty string should be compatible with changing over time default.
If the user specified secureBoot configuration we add it to the machine pool, otherwise we do nothing and get the default.

data/data/gcp/variables-gcp.tf
```diff
@@ -163,3 +163,9 @@ variable "gcp_private_zone_project" {
   default = ""
   description = "Project where the private managed zone will exist."
 }
+
+variable "gcp_master_secure_boot" {
+  type = string
```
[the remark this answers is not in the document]
eranco74: I'd like to be able to add the shileded_instance_config only in case the user provided it
For that reason, I added the dynamic block in [data/data/gcp/cluster/master/main.tf](https://github.com/openshift/installer/pull/6698/files#diff-d2a68514e186572b544fef566ba0a25ad5c4540d4df56732d75937432d7fb486)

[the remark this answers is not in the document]
patrickdillon: I should have added more context/reasoning for my previous comment:

As a rule of thumb, I try to remove as much logic as possible from the terraform configs. If it is possible to move logic into go code, I will do that so that the terraform configs are as simple, declarative configs as possible. That said, a lot of logic still ends up in the configs; we just try to minimize it.

For this particular case, based on your comment, you want to keep three states in the context of the terraform configs:

1. User-specified enabled
2. User-specified disabled
3. Not user-specfiied: default

My comment was based on the idea that we would only have two states in the config: enabled/disabled, and we would determine the default in go code. Following this idea of _our_ determining the default behavior would have pros and cons: behavior would remain explicit, but if the default behavior changes on the cloud provider side we would need to (minimally) update our code to conform. 

If my reasoning about the three states is correct, then I think the existing code makes sense and I'm happy to keep moving it forward. Was this what you had in mind for why we're conditionally creating the dynamic block?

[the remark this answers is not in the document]
eranco74: That is accurate, In case the user didn't specify a configuration we should get the default behavior from the cloud provider. The dynamic block allows it + in case the default changes we don't need to update our code.

eranco74:
Example usage, update the install-config compute or controlPlane platform field: 
```
compute:
- architecture: amd64
  hyperthreading: Enabled
  name: worker
  platform:
    gcp:
      secureBoot: Enabled
  replicas: 0
controlPlane:
  architecture: amd64
  hyperthreading: Enabled
  name: master
  platform:
    gcp:
       secureBoot: Enabled
  replicas: 1
```

pkg/types/gcp/machinepools.go
eranco74: Updated the commit message to make it clearer

PR #6699, opened 2022-12-14
AGENT-438: Create PXE files

pawanpinjarkar:
Supports a new command for agent installer to create pxe image files.

Run `openshift-install agent create pxe-files` command which will consume the general YAML config files ( install-config, agent-config) similar to the `agent create image` command. The `create pxe-files` command first extracts the `initrd.img`, `rootfs.img` and the kernal `vmlinuz` files from the base RHCOS image into `/tmp` directory. Then, wraps the agent installer ignition in `initrd.img` file and save into `agent-initrd.x86_64.img`.  3 new files with appropriate arch name are created in `pxe` directory viz `agent-initrd.x86_64.img`, `agent-rootfs.x86_64.img`, and `agent-vmlinuz.x86_64` 

Example:
```
./bin/openshift-install agent create pxe-files
WARNING Platform.Baremetal.ClusterProvisioningIP: 172.22.0.3 is ignored 
WARNING Platform.Baremetal.ExternalBridge: baremetal is ignored 
WARNING Platform.Baremetal.ExternalMACAddress: 52:54:00:5a:33:c4 is ignored 
WARNING Platform.Baremetal.ProvisioningBridge: provisioning is ignored 
WARNING Platform.Baremetal.ProvisioningMACAddress: 52:54:00:1e:11:da is ignored 
WARNING Platform.Baremetal.ProvisioningNetworkCIDR: 172.22.0.0/24 is ignored 
WARNING Platform.Baremetal.ProvisioningDHCPRange: 172.22.0.10,172.22.0.254 is ignored 
WARNING Capabilities: %!!(MISSING)s(*types.Capabilities=<nil>) is ignored 
INFO The rendezvous host IP (node0 IP) is 192.168.111.20 
INFO Extracting base ISO from release payload     
WARNING Failed to extract base ISO from release payload - check registry configuration 
INFO Downloading base ISO                         
INFO Consuming Agent Config from target directory 
INFO PXE-files created in: pxe                    
WARNING Discarding the Kubeconfig Admin Client that was provided in the target directory because its dependencies are dirty and it needs to be regenerated
```

The pxe images are created in `pxe` directory

```
root@pawan /home/ppinjark/installer -  (create-pxe-image) $ ll pxe/
total 1123872
-rw-r--r--. 1 root root   91667208 Dec 14 07:42 agent-initrd.x86_64.img
-rw-r--r--. 1 root root 1048706560 Dec 14 07:42 agent-rootfs.x86_64.img
-rw-r--r--. 1 root root   10468720 Dec 14 07:42 agent-vmlinuz.x86_64
```

Run the specific integration test with `go test -v -run TestAgentIntegration/agent_pxe/sno ./cmd/...`

Signed-off: Pawan Pinjarkar(ppinjark@redhat.com)


cmd/openshift-install/agent.go
```diff
@@ -69,7 +69,21 @@ var (
 		},
 	}
 
-	agentTargets = []target{agentConfigTarget, agentManifestsTarget, agentImageTarget}
+	agentPXEImageTarget = target{
+		name: "PXE-Image",
+		command: &cobra.Command{
+			Use:   "pxe-image",
```
pawanpinjarkar, line 75: Should the command be `pxe-images` because more than 1 .img file is created?

pawanpinjarkar: We will go with `pxe-files`

pkg/asset/agent/image/agentpxeimage.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = isoeditor.Extract(isoPath, workDir)
+	if err != nil {
+		return err
+	}
+
+	irfsPath := filepath.Join(workDir, "images/pxeboot/initrd.img")
```
zaneb, line 68: Once we are extracting the files we want individually, we should write the base initrd to a tempfile, rather than putting it somewhere that could cause confusion if it got left behind after an error.
Even better would if we could stream it out of the ISO without writing it to disk, but the `NewInitRamFSStreamReader()` API currently takes a filesystem path so that isn't yet possible.

pkg/asset/agent/image/agentpxeimage.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return err
+	}
+
+	workDir, err := GetCacheDir(imageDataType)
+	if err != nil {
+		return err
+	}
+
+	err = isoeditor.Extract(isoPath, workDir)
```
bfournie, line 63: Could you use isoeditor.GetFileFromISO(isoPath, filePath) to get individual files?

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -0,0 +1,187 @@
...
+	if a.imageReader == nil {
+		return errors.New("cannot generate PXE assets due to configuration errors")
+	}
+
+	defer a.imageReader.Close()
+
+	os.RemoveAll(pxeAssetsPath)
+
+	pxeAssetsPath := filepath.Join(directory, pxeAssetsPath)
+	err := os.Mkdir(pxeAssetsPath, 0750)
```
pawanpinjarkar, line 99: Kept same permissions as `auth` dir

andfasano:
Can we move https://github.com/openshift/installer/pull/6721 directly here? It will make easier also the life for the reviewers, given that they will have directly the test to try out the code in the same PR

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -0,0 +1,187 @@
...
+	dependencies.Get(ignition)
+
+	baseImage := &BaseIso{}
+	dependencies.Get(baseImage)
+
+	isoGetter := newGetIso(GetIsoPluggable)
+	isoPath, err := isoGetter.getter()
+	if err != nil {
+		return err
+	}
```
andfasano, line 54: This part is a little bit tricky, as it exercises a whole set of code that is not necessarily required and also potentially problematic (worth probably to be reviewed into another PR @bfournie).
Essentially, it re-retriggers the download of the coreos image from the internet, even thought the `BaseIso` dependency already extracted it from the release payload:

```
level=debug msg=  Generating BaseIso Image...
level=info msg=Extracting base ISO from release payload
...
level=debug msg=Extracted base ISO image ~/.cache/agent/image_cache/coreos-x86_64.iso from release payload
```
and immediately after:
```
level=debug msg=Generating Agent Installer PXE Files...
level=debug msg=Obtaining RHCOS image file from 'https://rhcos.mirror.openshift.com/art/storage/prod/streams/4.13/builds/413.86.202212131234-0/x86_64/rhcos-413.86.202212131234-0-live.x86_64.iso'
level=debug msg=Unpacking file into "~/.cache/agent/image_cache/rhcos-413.86.202212131234-0-live.x86_64.iso"...
```

By debugging the code, that seems to happen because the image gets initially saved as `coreos-x86_64.iso` in the cache, thus the `GetIsoPluggable` re-downloads it again since it was looking for a `rhcos-413.86.202212131234-0-live.x86_64.iso` in the cache.

Apart this issue (and as I said I think we should review that portion of the code in the BaseIso), in the current asset the only piece of information required is the location of the base ISO. I think it would be safer than to expose this info directly as a member of the `BaseIso` and use it here

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -0,0 +1,187 @@
...
+func (a *AgentPXEFiles) PersistToFile(directory string) error {
+	// If the imageReader is not set then it means that either one of the AgentPXEFiles
+	// dependencies or the asset itself failed for some reason
+	if a.imageReader == nil {
+		return errors.New("cannot generate PXE assets due to configuration errors")
+	}
+
+	defer a.imageReader.Close()
+
+	os.RemoveAll(pxeAssetsPath)
```
andfasano, line 98: if the effective path is calculated the line below, shouldn't this one be moved after that?

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -0,0 +1,187 @@
...
+
+	pxeAssetsPath := filepath.Join(directory, pxeAssetsPath)
```
andfasano, line 100: I personally find this approach complex to read and prone to error, keeping two distinct variables (one for the relative path, and one for the absolute) will be safer IMHO

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	defer output.Close()
+
+	_, err = io.Copy(output, a.imageReader)
+	if err != nil {
+		return err
+	}
+
+	isoGetter := newGetIso(GetIsoPluggable)
+	isoPath, err := isoGetter.getter()
```
andfasano, line 119: Same as before

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+}
+
+// Generate generates the image files for PXE asset.
+func (a *AgentPXEFiles) Generate(dependencies asset.Parents) error {
+	ignition := &Ignition{}
+	dependencies.Get(ignition)
+
+	baseImage := &BaseIso{}
+	dependencies.Get(baseImage)
```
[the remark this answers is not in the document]
pawanpinjarkar: `BaseIso` is not required at all. Removing it.

pkg/asset/agent/image/agentpxefiles.go
pawanpinjarkar: indeed

pkg/asset/agent/image/agentpxefiles.go
pawanpinjarkar: I kept it the same as in `agentimage` asset https://github.com/openshift/installer/blob/1e05df75d18b93f4dd35ab3f7145cce3e1523741/pkg/asset/agent/image/agentimage.go#L71

pkg/asset/agent/image/agentpxefiles.go
pawanpinjarkar: done

pkg/asset/agent/image/agentpxefiles.go
pawanpinjarkar: Done

pkg/asset/agent/image/agentpxefiles.go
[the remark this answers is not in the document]
pawanpinjarkar: With `isoPath`, it is now required.

pkg/asset/agent/image/agentpxefiles.go
andfasano: `agentIsoFile` and `agentISOFilename` are distinct vars, so they are ok. Just to avoid the confusion, something like that could be used:

```
pxeAssetsFullPath := filepath.Join(directory, pxeAssetsPath)
```

andfasano:
Thanks @pawanpinjarkar for the patch, apart a minor point on the var name now it looks fine to me!

/lgtm

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -0,0 +1,178 @@
...
+
+	a.isoPath = baseImage.baseISOPath
```
bfournie, line 51: Could you use baseImage.File.Filename here instead of baseImage.baseISOPath?

pkg/asset/agent/image/baseiso.go
```diff
@@ -131,6 +132,7 @@ func (i *BaseIso) Generate(dependencies asset.Parents) error {
 		if err == nil {
 			logrus.Debugf("Extracted base ISO image %s from release payload", baseIsoFileName)
 			i.File = &asset.File{Filename: baseIsoFileName}
+			i.baseISOPath = baseIsoFileName
```
bfournie, line 135: Not needed if you access the Filename through i.File.

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -0,0 +1,178 @@
...
+
+	os.RemoveAll(pxeAssetsFullPath)
+
+	err := os.Mkdir(pxeAssetsFullPath, 0750)
+	if err != nil {
+		return err
+	}
+
+	agentInitrdFile := filepath.Join(pxeAssetsFullPath, fmt.Sprintf("agent-%s.%s.img", initrdimg, a.cpuArch))
+	output, err := os.Create(agentInitrdFile)
```
bfournie, line 104: You could use the copy() function below that creates the file and copies data to it.

pkg/asset/agent/image/agentpxefiles.go
pawanpinjarkar: Yes, that seems to work. For some reason, I thought it won't give me the full path from the cache, which is not the case.

pkg/asset/agent/image/baseiso.go
```diff
@@ -20,7 +20,7 @@ import (
 
 // BaseIso generates the base ISO file for the image
 type BaseIso struct {
-	File *asset.File
+	File        *asset.File
```
bfournie, line 23: Since no changes anymore file shouldn't show up.

cmd/openshift-install/create.go
```diff
@@ -299,7 +299,7 @@ func runTargetCmd(targets ...asset.WritableAsset) func(cmd *cobra.Command, args
 			logrus.Fatal(err)
 		}
 		switch cmd.Name() {
-		case "cluster", "image":
+		case "cluster", "image", "pxe-files":
```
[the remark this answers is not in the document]
pawanpinjarkar: For some reason, it says `INFO PXE-files created in: auth`. As an alternative, https://github.com/openshift/installer/pull/6699/files#diff-38f93f2da042069046ec25c748458e4aeb5457951a609feec974083616298f42R121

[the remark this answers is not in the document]
zaneb: Ah, right. It prints a list of directories (so we want it to say `auth, pxe`) **but** we use `PersistToFile()` instead of the regular mechanism for writing files, so maybe it doesn't get added to the list.

r4f4:
/override ci/prow/e2e-azure-ovn ci/prow/e2e-gcp-ovn
Not related to the changes of this PR and timing out after 4h (install succeeded)

PR #6700, opened 2022-12-14
PowerVS: Defer cancel when calling contextWithTimeout

hamzy:
Every place where this occurs:
        ctx, _ := o.contextWithTimeout()
Change to:
        ctx, cancel := o.contextWithTimeout()
        defer cancel()

NOTE:
For the functions instance.NewIBMPI.*Client(, we now use context.Background() because o.contextWithTimeout() and then cancel() will cause future calls to these functions to error out with "context canceled."

### DIFF
--- a/pkg/destroy/powervs/powervs.go
+++ b/pkg/destroy/powervs/powervs.go
@@ -240,7 +240,9 @@ func (o *ClusterUninstaller) Run() (*types.ClusterQuota, error) {
 	var ok bool
 	var err error
 
-	ctx, _ = o.contextWithTimeout()
+	ctx, cancel := o.contextWithTimeout()
+	defer cancel()
+
 	if ctx == nil {
 		return nil, errors.Wrap(err, "powervs.Run: contextWithTimeout returns nil")
 	}
@@ -369,7 +371,9 @@ func (o *ClusterUninstaller) executeStageFunction(f struct {
 	var ok bool
 	var err error
 
-	ctx, _ = o.contextWithTimeout()
+	ctx, cancel := o.contextWithTimeout()
+	defer cancel()
+
 	if ctx == nil {
 		return errors.Wrap(err, "executeStageFunction contextWithTimeout returns nil")
 	}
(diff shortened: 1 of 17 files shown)


hamzy:
/assign @r4f4 
FYI, I ran the new golint locally with no issues.

barbacbd (commented):
/lgtm cancel

pkg/destroy/powervs/powervs.go
```diff
@@ -520,34 +524,32 @@ func (o *ClusterUninstaller) loadSDKServices() error {
 		return fmt.Errorf("loadSDKServices: loadSDKServices: o.piSession is nil")
 	}
 
-	ctx, _ := o.contextWithTimeout()
-
-	o.instanceClient = instance.NewIBMPIInstanceClient(ctx, o.piSession, o.ServiceGUID)
+	o.instanceClient = instance.NewIBMPIInstanceClient(context.Background(), o.piSession, o.ServiceGUID)
```
r4f4, line 527: I see this client is weirdly designed. So the context you pass here when creating the client will be used for all the subsequent client calls?

pkg/destroy/powervs/powervs.go
```diff
@@ -628,7 +630,7 @@ func (o *ClusterUninstaller) loadSDKServices() error {
 
 		// Get the Zone ID
 		zoneOptions := o.zonesSvc.NewListZonesOptions()
-		zoneResources, detailedResponse, err := o.zonesSvc.ListZonesWithContext(ctx, zoneOptions)
+		zoneResources, detailedResponse, err := o.zonesSvc.ListZonesWithContext(context.Background(), zoneOptions)
```
r4f4, line 633: This seems to be a different case from before. Are you sure there isn't a sane value to pass here for a timeout when listing zones?

pkg/destroy/powervs/powervs.go
hamzy: Yep!

r4f4: Nice job hunting that one down.

pkg/destroy/powervs/powervs.go
hamzy: Good catch!

r4f4:
/approve
```sh
$: git branch
* (HEAD detached at hamzy/powervs-cleanup-context-timeout)
$: rg 'ctx, _' pkg/destroy/powervs/
$: 
```
Thank you for doing this.

pkg/destroy/powervs/powervs.go
hamzy: Thanks!  I was worried I would never find the issue...

r4f4:
/override ci/prow/golint
Ran locally with a longer timeout, no problems reported.

PR #6714, opened 2022-12-16
AGENT-494: Use 0600 for perms of created files

bfournie:
Gosec requires permissions of 0600 or more restrictive.

### DIFF
--- a/pkg/asset/agent/agentconfig/agent_config.go
+++ b/pkg/asset/agent/agentconfig/agent_config.go
@@ -97,7 +97,7 @@ func (a *AgentConfig) PersistToFile(directory string) error {
 	templatePath := filepath.Join(directory, agentConfigFilename)
 	templateByte := []byte(a.Template)
 
-	err := os.WriteFile(templatePath, templateByte, 0644)
+	err := os.WriteFile(templatePath, templateByte, 0600)
 	if err != nil {
 		return err
 	}
--- a/pkg/asset/agent/image/agentimage.go
+++ b/pkg/asset/agent/image/agentimage.go
@@ -84,7 +84,7 @@ func (a *AgentImage) PersistToFile(directory string) error {
 		return err
 	}
 
-	err = os.WriteFile(filepath.Join(directory, "rendezvousIP"), []byte(a.rendezvousIP), 0644)
+	err = os.WriteFile(filepath.Join(directory, "rendezvousIP"), []byte(a.rendezvousIP), 0600)
 	if err != nil {
 		return err
 	}


pkg/asset/agent/agentconfig/agent_config.go
```diff
@@ -97,7 +97,7 @@ func (a *AgentConfig) PersistToFile(directory string) error {
 	templatePath := filepath.Join(directory, agentConfigFilename)
 	templateByte := []byte(a.Template)
 
-	err := os.WriteFile(templatePath, templateByte, 0644)
+	err := os.WriteFile(templatePath, templateByte, 0600)
```
zaneb, line 100: I'd be in favour of disabling G306, because this seems overly conservative for writing to a local directory to me. It also doesn't actually catch much, because e.g. calling `os.Create()` doesn't trigger any warning, despite using `0666`.

r4f4: Would you rather disable G306 or raise its minimum from `0600` to `0644`? Notice we can always `//nolint:gosec` to disable the warning for that particular case, but I think it's still good to have the linter make us double check these kind of things.

bfournie:
/hold
Holding this one for now as we'll most likely close it and change the linter instead.

pkg/asset/agent/agentconfig/agent_config.go
bfournie: As Zane pointed out, there are many places in the installer where 0666 is used for local files. I would think you would need to raise the minimum to that if you want to use the linter for these checks. E.g.
```
cmd/openshift-install/log.go:	logfile, err := os.OpenFile(filepath.Join(baseDir, ".openshift_install.log"), os.O_WRONLY|os.O_APPEND|os.O_CREATE, 0666)
data/unpack.go:	return UnpackWithFilePermissions(base, uri, 0666)
pkg/asset/installconfig/alibabacloud/client.go:	file, err := os.OpenFile(filePath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0666)
pkg/asset/store/assetcreate_test.go:				if err := os.WriteFile(filepath.Join(tempDir, name), []byte(contents), 0666); err != nil {
pkg/asset/store/filefetcher_test.go:				err := os.WriteFile(filepath.Join(tempDir, filename), data, 0666)
pkg/asset/store/filefetcher_test.go:		err := os.WriteFile(filepath.Join(tempDir, path), data, 0666)
pkg/destroy/bootstrap/bootstrap.go:	return os.WriteFile(to, data, 0666)
pkg/terraform/init.go:	if err := os.WriteFile(filepath.Join(dir, name), data, 0666); err != nil {

```

celebdor:
/close

The decision was to instead do a change in the linter config

PR #6717, opened 2022-12-19
OCPBUGS-4549: destroy: azure: handle `nil` responses from msgraph sdk

r4f4:
It seems the SDK can return `nil` as a response even if no errors occurred. If we don't check for that, we get the following error on Gov cloud:

```
12-19 13:21:52.860  level=debug msg=deleting application registrations
12-19 13:21:52.860  E1219 05:21:48.969224     743 runtime.go:79] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
12-19 13:21:52.860  goroutine 1 [running]:
12-19 13:21:52.860  k8s.io/apimachinery/pkg/util/runtime.logPanic({0x6132160?, 0x22371b20})
12-19 13:21:52.860  	/go/src/github.com/openshift/installer/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:75 +0x99
12-19 13:21:52.860  k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0x0?})
12-19 13:21:52.860  	/go/src/github.com/openshift/installer/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:49 +0x75
12-19 13:21:52.861  panic({0x6132160, 0x22371b20})
12-19 13:21:52.861  	/usr/lib/golang/src/runtime/panic.go:884 +0x212
12-19 13:21:52.861  github.com/openshift/installer/pkg/destroy/azure.getServicePrincipalsByTag({0x1d9580b0, 0xc0008443c0}, 0xc000cec420, {0xc000792570, 0x29}, {0xc000dcc650, 0xd})
12-19 13:21:52.861  	/go/src/github.com/openshift/installer/pkg/destroy/azure/azure.go:632 +0x322
12-19 13:21:52.861  github.com/openshift/installer/pkg/destroy/azure.deleteApplicationRegistrations({0x1d9580b0, 0xc0008443c0}, 0xc000cec420, {0x1d98a0d0, 0xc0001da000}, {0xc000dcc650, 0xd})
```

This is a fallout of https://github.com/openshift/installer/pull/6614

### DIFF
--- a/pkg/destroy/azure/azure.go
+++ b/pkg/destroy/azure/azure.go
@@ -578,14 +578,18 @@ func extractODataError(err error) error {
 }
 
 func deleteApplicationRegistrations(ctx context.Context, graphClient *msgraphsdk.GraphServiceClient, logger logrus.FieldLogger, infraID string) error {
-	var errorList []error
-
 	tag := fmt.Sprintf("kubernetes.io_cluster.%s=owned", infraID)
 	servicePrincipals, err := getServicePrincipalsByTag(ctx, graphClient, tag, infraID)
 	if err != nil {
 		return errors.Wrap(err, "failed to gather list of Service Principals by tag")
 	}
+	// msgraphsdk can return a `nil` response even if no errors occurred
+	if servicePrincipals == nil {
+		logger.Debug("Empty response from API when listing Service Principals by tag")
+		return nil
+	}
 
+	var errorList []error
 	for _, sp := range servicePrincipals {
 		appID := *sp.GetAppId()
 		logger := logger.WithField("appID", appID)
@@ -600,6 +604,11 @@ func deleteApplicationRegistrations(ctx context.Context, graphClient *msgraphsdk
 		resp, err := graphClient.Applications().Get(ctx, &listQuery)
 		if err != nil {
 			errorList = append(errorList, err)
+			continue
+		}
+		// msgraphsdk can return a `nil` response even if no errors occurred
+		if resp == nil {
+			logger.Debugf("Empty response getting Application from Service Principal %s", *sp.GetDisplayName())
 			continue
 		}
 		apps := resp.GetValue()
@@ -626,7 +635,7 @@ func getServicePrincipalsByTag(ctx context.Context, graphClient *msgraphsdk.Grap
 		},
 	}
 	resp, err := graphClient.ServicePrincipals().Get(ctx, &listQuery)
-	if err != nil {
+	if err != nil || resp == nil {
 		return nil, err
 	}
 	return resp.GetValue(), nil


pkg/destroy/azure/azure.go
```diff
@@ -629,5 +633,9 @@ func getServicePrincipalsByTag(ctx context.Context, graphClient *msgraphsdk.Grap
 	if err != nil {
 		return nil, err
 	}
+	// msgraphsdk can return a `nil` response even if no errors occurred
+	if resp == nil {
```
barbacbd, line 637: nit: Since the response can be nil, couldn't we simplify the return to 

```
if err != nil || resp == nil {
    return nil, err
}
```

r4f4: Indeed. Will do.

pkg/destroy/azure/azure.go
```diff
@@ -602,6 +602,10 @@ func deleteApplicationRegistrations(ctx context.Context, graphClient *msgraphsdk
 			errorList = append(errorList, err)
 			continue
 		}
+		// msgraphsdk can return a `nil` response even if no errors occurred
+		if resp == nil {
+			continue
+		}
```
patrickdillon, line 608: Do we know what a nil response means? Was the deletion successful? API not supported? 

If we don't know, or aren't certain, I would suggest do a little more here in terms of logging, perhaps a warning or or adding an error to the error list.

pkg/destroy/azure/azure.go
patrickdillon: We only want to return `err` if it is not nil. As the code stands now, if `resp != nil` but `err == nil` we would return the nil err and calling code would panic.

pkg/destroy/azure/azure.go
r4f4: In this case it's just a GET (all the apps with ID == `appID`), the delete happens a few lines later. I cannot easily tell why, would have to dig dipper into the msgraph code: https://github.com/microsoftgraph/msgraph-sdk-go/blob/main/serviceprincipals/service_principal_item_request_builder.go#L267-L273

r4f4: I think this is the `SendAsync` function: https://github.com/microsoft/kiota-http-go/blob/main/nethttp_request_adapter.go#L286-L287

r4f4: Anyway, it's probably a good idea to at least add some logging in those cases. Thanks.

r4f4:
/hold
while we check what happens if we return error instead of ignoring a `nil` response

barbacbd (commented):
/lgtm

pkg/destroy/azure/azure.go
```diff
@@ -578,14 +578,18 @@ func extractODataError(err error) error {
...
 
 func deleteApplicationRegistrations(ctx context.Context, graphClient *msgraphsdk.GraphServiceClient, logger logrus.FieldLogger, infraID string) error {
-	var errorList []error
-
 	tag := fmt.Sprintf("kubernetes.io_cluster.%s=owned", infraID)
 	servicePrincipals, err := getServicePrincipalsByTag(ctx, graphClient, tag, infraID)
 	if err != nil {
 		return errors.Wrap(err, "failed to gather list of Service Principals by tag")
 	}
+	// msgraphsdk can return a `nil` response even if no errors occurred
```
sadasu, line 586: This logic can be moved within https://github.com/openshift/installer/blob/master/pkg/destroy/azure/azure.go#L621. Also, looking at getServicePrincipalsByTag(), it appears that this method can set resp to nil even if the original API did not return a nil.

pkg/destroy/azure/azure.go
```diff
@@ -602,6 +606,11 @@ func deleteApplicationRegistrations(ctx context.Context, graphClient *msgraphsdk
 			errorList = append(errorList, err)
 			continue
 		}
+		// msgraphsdk can return a `nil` response even if no errors occurred
+		if resp == nil {
+			logger.Debugf("Empty response getting Application from Service Principal %s", *sp.GetDisplayName())
+			continue
+		}
 		apps := resp.GetValue()
```
sadasu, line 614: Do we have to chck for resp == nil for https://github.com/openshift/installer/blob/master/pkg/destroy/azure/azure.go#L611?

pkg/destroy/azure/azure.go
r4f4: I preferred to do it here so I don't have to change the `getServicePrincipalsByTag` to receive a `logger` just so I can log a message in case that happens. I can change it if you think there is an advantage in doing that.

pkg/destroy/azure/azure.go
r4f4: Hard to say since the API is not very clear about it. I don't think so since in the docs they don't mention any error handling https://learn.microsoft.com/en-us/graph/sdks/create-requests?tabs=Go#delete-an-entity. Notice that `ApplicationsById` does not return an `Application` object, but an `ApplicationItemRequestBuilder` instead. I'd be very surprised if a `nil` response was returned for that.

pkg/destroy/azure/azure.go
r4f4: > Also, looking at getServicePrincipalsByTag(), it appears that this method can set resp to nil even if the original API did not return a nil.

That's OK because in that case `err != nil` and the error condition will (and should) be evaluated first.

PR #6723, opened 2022-12-20
AGENT-491: Support pxe base url

pawanpinjarkar:
AgentConfig accepts a new optional field `pxeBaseUrl`.

When provided, an iPXE script file is created along with other PXE files in the `pxe` directory.

### DIFF
--- a/cmd/openshift-install/agent_internal_integration_test.go
+++ b/cmd/openshift-install/agent_internal_integration_test.go
@@ -84,6 +84,8 @@ func runIntegrationTest(t *testing.T, testFolder string) {
 
 	testscript.Run(t, testscript.Params{
 		Dir: testFolder,
+		// Uncomment below line to help debug the testcases
+		// TestWork: true,
 
 		Setup: func(e *testscript.Env) error {
 			// This is required to allow proper
--- a/pkg/asset/agent/image/agentpxefiles.go
+++ b/pkg/asset/agent/image/agentpxefiles.go
@@ -7,11 +7,13 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 
 	"github.com/openshift/assisted-image-service/pkg/isoeditor"
 	"github.com/openshift/installer/pkg/asset"
+	config "github.com/openshift/installer/pkg/asset/agent/agentconfig"
 )
 
 const (
@@ -27,6 +29,7 @@ type AgentPXEFiles struct {
 	imageReader isoeditor.ImageReader
 	cpuArch     string
 	isoPath     string
+	ipxeBaseURL string
 }
 
 var _ asset.WritableAsset = (*AgentPXEFiles)(nil)
@@ -35,6 +38,7 @@ var _ asset.WritableAsset = (*AgentPXEFiles)(nil)
 func (a *AgentPXEFiles) Dependencies() []asset.Asset {
 	return []asset.Asset{
 		&Ignition{},
+		&config.AgentConfig{},
 		&BaseIso{},
 	}
 }
@@ -43,6 +47,9 @@ func (a *AgentPXEFiles) Dependencies() []asset.Asset {
 func (a *AgentPXEFiles) Generate(dependencies asset.Parents) error {
 	ignition := &Ignition{}
 	dependencies.Get(ignition)
+
+	agentconfig := &config.AgentConfig{}
+	dependencies.Get(agentconfig)
 
 	baseImage := &BaseIso{}
 	dependencies.Get(baseImage)
@@ -76,6 +83,7 @@ func (a *AgentPXEFiles) Generate(dependencies asset.Parents) error {
 
 	a.imageReader = custom
 	a.cpuArch = ignition.CPUArch
+	a.ipxeBaseURL = strings.Trim(agentconfig.Config.IPxeBaseURL, "/")
 
 	return nil
 }
@@ -118,6 +126,13 @@ func (a *AgentPXEFiles) PersistToFile(directory string) error {
 		return err
 	}
 
+	if a.ipxeBaseURL != "" {
+		err = a.createiPXEScript(a.cpuArch, a.ipxeBaseURL, pxeAssetsFullPath)
+		if err != nil {
+			return err
+		}
+	}
+
 	logrus.Infof("PXE-files created in: %s", pxeAssetsFullPath)
 
 	return nil
@@ -141,7 +156,7 @@ func (a *AgentPXEFiles) Files() []*asset.File {
 	return []*asset.File{}
 }
 
-func (a *AgentPXEFiles) extractPXEFileFromISO(isoPath string, srcfilename string, dstfilename string) error {
+func (a *AgentPXEFiles) extractPXEFileFromISO(isoPath, srcfilename, dstfilename string) error {
 	fileReader, err := isoeditor.GetFileFromISO(isoPath, srcfilename)
 	if err != nil {
 		return err
@@ -166,6 +181,29 @@ func (a *AgentPXEFiles) copy(filepath string, src io.Reader) error {
 	if err != nil {
 		return err
 	}
+
+	return nil
+}
+
+func (a *AgentPXEFiles) createiPXEScript(cpuArch, ipxeBaseURL, pxeAssetsFullPath string) error {
+	iPXEScriptTemplate := `#!ipxe
+initrd --name initrd %s/%s
+kernel %s/%s initrd=initrd coreos.live.rootfs_url=%s/%s ignition.firstboot ignition.platform.id=metal
+boot
+`
+
+	iPXEScript := fmt.Sprintf(iPXEScriptTemplate, ipxeBaseURL,
+		fmt.Sprintf("agent-%s.%s.img", initrdimg, cpuArch), ipxeBaseURL,
+		fmt.Sprintf("agent-%s", vmlinuz), ipxeBaseURL,
+		fmt.Sprintf("agent-%s.%s", rootfsimg, cpuArch))
+
+	iPXEFile := fmt.Sprintf("agent.%s.ipxe", a.cpuArch)
+
+	err := os.WriteFile(filepath.Join(pxeAssetsFullPath, iPXEFile), []byte(iPXEScript), 0600)
+	if err != nil {
+		return err
+	}
+	logrus.Infof("Created iPXE script %s in %s directory", iPXEFile, pxeAssetsFullPath)
 
 	return nil
 }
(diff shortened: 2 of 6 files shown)


pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -76,6 +77,7 @@ func (a *AgentPXEFiles) Generate(dependencies asset.Parents) error {
 
 	a.imageReader = custom
 	a.cpuArch = ignition.CPUArch
+	a.pxeBaseURL = ignition.PxeBaseURL
```
zaneb, line 80: Is there a reason to pass this down through Ignition instead of depending on the AgentConfig directly?

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -169,3 +178,35 @@ func (a *AgentPXEFiles) copy(filepath string, src io.Reader) error {
...
+random.trust_cpu=on 
+rd.luks.options=discard 
+ignition.firstboot 
+ignition.platform.id=metal 
+console=tty1 
+console=ttyS1,115200n8 
+coreos.inst.persistent-kargs="console=tty1 console=ttyS1,115200n8" 
+boot`
+
+	iPXEScript := fmt.Sprintf(iPXEScriptTemplate, pxeBaseURL,
```
zaneb, line 198: nit: the way this is formatted, with the line breaks occurring between the strings that are on the same line in the output, is less readable than it could be.

pkg/types/agent/agent_config_type.go
```diff
@@ -25,6 +25,7 @@ type Config struct {
 	AdditionalNTPSources []string `json:"additionalNTPSources,omitempty"`
 	// ip address of node0
 	RendezvousIP string `json:"rendezvousIP,omitempty"`
+	PxeBaseURL   string `json:"pxeBaseURL,omitempty"`
```
zaneb, line 28: On reflection, perhaps this should be `ipxeBaseURL`

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -118,6 +120,13 @@ func (a *AgentPXEFiles) PersistToFile(directory string) error {
 		return err
 	}
 
+	if a.pxeBaseURL != "" {
+		err = a.createiPXEScript(a.cpuArch, a.pxeBaseURL, pxeAssetsFullPath)
```
zaneb, line 124: It probably wouldn't hurt to strip any trailing `/`s from the URL the user gives us, so we don't double up on them.

zaneb (commented):
This looks basically there.
If we're adding a field to the agent-config, I'm not sure if we're supposed to bump the version to v1beta2? Might be worth checking with the installer team how they have managed additions to the install-config.

pkg/asset/agent/image/agentpxefiles.go
pawanpinjarkar: Similar to the way we are doing for other fields from AgentConfig like `RendezvousIP` 🤔

zaneb: Those are actually used in the Ignition though, this isn't.

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -169,3 +184,22 @@ func (a *AgentPXEFiles) copy(filepath string, src io.Reader) error {
 
 	return nil
 }
+
+func (a *AgentPXEFiles) createiPXEScript(cpuArch, ipxeBaseURL, pxeAssetsFullPath string) error {
+	iPXEScriptTemplate := `#!ipxe initrd --name initrd %s/%s kernel %s/%s  initrd=initrd coreos.live.rootfs_url=%s/%sg random.trust_cpu=on rd.luks.options=discard ignition.firstboot ignition.platform.id=metal console=tty1 console=ttyS1,115200n8 coreos.inst.persistent-kargs="console=tty1 console=ttyS1,115200n8" boot`
```
zaneb, line 189: This is missing some newlines I think.

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -169,3 +184,24 @@ func (a *AgentPXEFiles) copy(filepath string, src io.Reader) error {
 
 	return nil
 }
+
+func (a *AgentPXEFiles) createiPXEScript(cpuArch, ipxeBaseURL, pxeAssetsFullPath string) error {
+	iPXEScriptTemplate := `#!ipxe 
+	initrd --name initrd %s/%s kernel %s/%s  initrd=initrd coreos.live.rootfs_url=%s/%sg random.trust_cpu=on rd.luks.options=discard ignition.firstboot ignition.platform.id=metal console=tty1 console=ttyS1,115200n8 coreos.inst.persistent-kargs="console=tty1 console=ttyS1,115200n8"
```
zaneb, line 190: Still needs a newline before `kernel`
Where do the `console=` args come from? I don't see them in the CoreOS ISO:

```
$ coreos-installer iso kargs show ~/.cache/agent/image_cache/coreos-x86_64.iso
coreos.liveiso=rhcos-414.92.202304131328-0 ignition.firstboot ignition.platform.id=metal
```

pawanpinjarkar: The `console=` args were referred from https://github.com/openshift/assisted-service/blob/b633a3ab4b9207aeff323cafcd748442e551d39d/docs/user-guide/deploy-on-bare-metal.md?plain=1#L55 but look like not necessary.

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -169,3 +184,25 @@ func (a *AgentPXEFiles) copy(filepath string, src io.Reader) error {
 
 	return nil
 }
+
+func (a *AgentPXEFiles) createiPXEScript(cpuArch, ipxeBaseURL, pxeAssetsFullPath string) error {
+	iPXEScriptTemplate := `#!ipxe 
+initrd --name initrd %s/%s 
+kernel %s/%s  initrd=initrd coreos.live.rootfs_url=%s/%sg random.trust_cpu=on rd.luks.options=discard ignition.firstboot ignition.platform.id=metal
```
andfasano, line 191: Where are coming from the kargs `random.trust_cpu=on rd.luks.options=discard`?

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -169,3 +184,25 @@ func (a *AgentPXEFiles) copy(filepath string, src io.Reader) error {
...
+initrd --name initrd %s/%s 
+kernel %s/%s  initrd=initrd coreos.live.rootfs_url=%s/%sg random.trust_cpu=on rd.luks.options=discard ignition.firstboot ignition.platform.id=metal 
+boot`
+
+	iPXEScript := fmt.Sprintf(iPXEScriptTemplate, ipxeBaseURL,
+		fmt.Sprintf("agent-%s.%s.img", initrdimg, cpuArch), ipxeBaseURL,
+		fmt.Sprintf("agent-%s", vmlinuz), ipxeBaseURL,
+		fmt.Sprintf("agent-%s.%s", rootfsimg, cpuArch))
+
+	iPXEFile := fmt.Sprintf("iPXE.%s", a.cpuArch)
```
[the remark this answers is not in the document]
pawanpinjarkar: I think we can go with `agent.x86_64.ipxe`

pkg/asset/agent/image/agentpxefiles.go
pawanpinjarkar: We can remove it. They were originally referred from [here](https://github.com/openshift/assisted-service/blob/b633a3ab4b9207aeff323cafcd748442e551d39d/docs/user-guide/deploy-on-bare-metal.md?plain=1#L55)

andfasano: If not an issue, I'd keep the bare minimum amount of strictly required kargs in the config file

zaneb: We should keep the same kargs as we use in the ISO (which don't include these two), modulo the ones that are pxe-specific, so that we minimise behavioural differences.
It's possible that those kargs are added even to the ISO in the assisted discovery image, and if that's the case maybe we should consider adding them (in both places) too. But that isn't part of this work.

zaneb: Also the `g` at the end of `coreos.live.rootfs_url=%s/%sg` looks suspect.

cmd/openshift-install/testdata/agent/pxe/configurations/sno.txt
```diff
@@ -44,3 +45,4 @@ metadata:
   name: ostest
   namespace: cluster0
 rendezvousIP: 192.168.111.20
+ipxeBaseURL: http://user-specified-pxe-infra.com
```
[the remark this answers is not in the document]
pawanpinjarkar: Done

cmd/openshift-install/agent_internal_integration_test.go
```diff
@@ -84,6 +84,8 @@ func runIntegrationTest(t *testing.T, testFolder string) {
 
 	testscript.Run(t, testscript.Params{
 		Dir: testFolder,
+		// Uncomment below line to help debug the testcases
+		// TestWork: true,
```
bfournie, line 88: Not really sure what this does...

pawanpinjarkar: It helps to save the temp work directory thereby one can debug the files and other directories in the work dir.

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -186,10 +186,11 @@ func (a *AgentPXEFiles) copy(filepath string, src io.Reader) error {
...
 func (a *AgentPXEFiles) createiPXEScript(cpuArch, ipxeBaseURL, pxeAssetsFullPath string) error {
-	iPXEScriptTemplate := `#!ipxe 
-initrd --name initrd %s/%s 
-kernel %s/%s  initrd=initrd coreos.live.rootfs_url=%s/%s ignition.firstboot ignition.platform.id=metal 
-boot`
+	iPXEScriptTemplate := `#!ipxe
+initrd --name initrd %s/%s
+kernel %s/%s initrd=initrd coreos.live.rootfs_url=%s/%s ignition.firstboot ignition.platform.id=metal
+boot
+`
```
patrickdillon: This is updating changes introduced in the previous commit and should be squashed.

patrickdillon (commented):
It looks like the commit structure needs to be a cleaned up a little and a commit message added, like this:
https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#commit-message-format

It may make sense to just squash the entire second commit.

deepsm007:
@pawanpinjarkar Looks like its making changes in agent installer and agent installer has presubmit failures . Please retest to confirm its not breaking

pawanpinjarkar:
> @pawanpinjarkar Looks like its making changes in agent installer and agent installer has presubmit failures . Please retest to confirm its not breaking

This particular PR although related to the agent installer and may look like some of the existing agent presubmits are failing, those failures are not related to the new functionality being developed via this PR.

PR #6727, opened 2022-12-21
CORS-2469: upi: Document removal of CPMS for UPI installation

rna-afk:
Documenting the extra step of removing the control plane machine set files created for Azure and GCP which need to be removed for UPI installation.


rna-afk:
Depends on #6632 and #6695

patrickdillon:
/approve

Why is this running tests????

PR #6730, opened 2022-12-22
OCPBUGS-3706: Improve error reporting from agent wait-for install-complete

zaneb:
When running the `agent wait-for install-complete` command, we first check that bootstrapping is complete (by running the equivalent of `agent wait-for bootstrap-complete`. However, if this failed because the bootstrapping timed out, we would report it as an install failure along with the corresponding debug messages (stating that the problem is with the cluster operators, and inevitably failing to fetch data about which).

If the failure occurs during bootstrapping, report it as a bootstrap error the same as you would get from `agent wait-for bootstrap-complete`. If the failure occurs later, don't print the assisted-installer event list. If the command failed because of a timeout, log that.

### DIFF
--- a/cmd/openshift-install/agent/waitfor.go
+++ b/cmd/openshift-install/agent/waitfor.go
@@ -94,6 +94,7 @@ func newWaitForInstallCompleteCmd() *cobra.Command {
 			}
 
 			if err = agentpkg.WaitForInstallComplete(cluster); err != nil {
+				logrus.Error(err)
 				err2 := cluster.API.OpenShift.LogClusterOperatorConditions()
 				if err2 != nil {
 					logrus.Error("Attempted to gather ClusterOperator status after wait failure: ", err2)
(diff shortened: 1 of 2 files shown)


cmd/openshift-install/agent/waitfor.go
```diff
@@ -85,9 +89,11 @@ func newWaitForInstallCompleteCmd() *cobra.Command {
 				logrus.Exit(exitCodeBootstrapFailed)
 			}
 
+			if err := agentpkg.WaitForBootstrapComplete(cluster); err != nil {
+				handleBootstrapError(cluster, err)
+			}
+
 			if err = agentpkg.WaitForInstallComplete(cluster); err != nil {
-				logrus.Debug("Printing the event list gathered from the Agent Rest API")
-				cluster.PrintInfraEnvRestAPIEventList()
```
rwsu, line 76: I'm curious why we would stop printing out the event list if the install failed to complete. Wouldn't it be useful information?

zaneb: The theory is that at this point we've succeeded in bootstrapping, and it's just that the CVO isn't complete, so the events that happened before/during bootstrapping seem less relevant at this point.
I think this was previously here because it was trying to cover all cases, since there was previously no way of telling if it failed during bootstrapping or after.

rwsu (approved):

PR #6734, opened 2023-01-02
OCPBUGS-5184: azure: validate Windows-only VM types

r4f4:
From azure doc
https://learn.microsoft.com/en-us/azure/virtual-machines/nvv4-series, looks like the NVv4 series only supports Windows.


jinyunma:
pre-merged verification is passed, see details in bug comments.
/label qe-approved

PR #6737, opened 2023-01-04
Expose Azure useImageGallery parameter in the MachineSets() call

dlom:
As a followup to #6694, Hive needs to be able to select whether to use the image gallery setting or not.  This change just forwards the `useImageGallery` further up the chain, and also removes the unused `rhcosVersion` parameter.

/cc @patrickdillon

### DIFF
--- a/pkg/asset/machines/azure/machinesets.go
+++ b/pkg/asset/machines/azure/machinesets.go
@@ -13,7 +13,7 @@ import (
 )
 
 // MachineSets returns a list of machinesets for a machinepool.
-func MachineSets(clusterID string, config *types.InstallConfig, pool *types.MachinePool, osImage, role, userDataSecret string, capabilities map[string]string, rhcosVersion string) ([]*clusterapi.MachineSet, error) {
+func MachineSets(clusterID string, config *types.InstallConfig, pool *types.MachinePool, osImage, role, userDataSecret string, capabilities map[string]string, useImageGallery bool) ([]*clusterapi.MachineSet, error) {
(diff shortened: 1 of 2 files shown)


2uasimojo:
This looks good to me. <bonk self> for missing it the first time around.

r4f4:
It would be nice if the commit message was more descriptive as suggested in our contributing guidelines: https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#commit-message-format

r4f4:
`golint` is failing with
```
 pkg/asset/machines/worker.go:415:58: G601: Implicit memory aliasing in for loop. (gosec)
			sets, err := azure.MachineSets(clusterID.InfraID, ic, &pool, string(*rhcosImage), "worker", workerUserDataSecretName, capabilities, useImageGallery) 
```
Even though this was not introduced by this PR, the golint tool is setup in such a way it'll report existing problems when new changes touch them. There are 2 ways to appease the linter here:
 1. At the beginning of the loop we do the following so the issue is fixed for every place that uses `&pool` in this loop
```
	for _, pool := range ic.Compute {
            pool := pool
```
 2. We ignore the warning since we know it's fine in this case
```
sets, err := azure.MachineSets(clusterID.InfraID, ic, &pool, string(*rhcosImage), "worker", workerUserDataSecretName, capabilities, useImageGallery) //nolint:gosec
```

dlom:
@r4f4 hopefully this appeases the golint CI... thanks for the review

dlom:
:tada: thanks everyone

PR #6739, opened 2023-01-04
OCPBUGS-4874: Remove order dependency for agent CLI string

bfournie:
In order to remove the dependency on providing baremetal hosts, the "agent" string is checked in the CLI command. This fixes the requirement that it always be after the "create".

(Note that this solution is an improvement, but not perfect, as it can result in a false positive if "agent" is used elsewhere in the command line. A more complete solution will come in a later release).


zaneb:
/lgtm
I think this is the best we can do for now.
/cherry-pick release-4.12

PR #6741, opened 2023-01-05
OCPBUGS-5324: CVE-2021-4235 go-yaml: Denial of Service in go-yaml

r4f4:
A flaw was found in go-yaml. This issue occurs due to unbounded alias chasing, where a maliciously crafted YAML file can cause the system to consume significant system resources. If parsing user input, this may be used as a denial of service vector.


barbacbd (commented):
/lgtm but this also begs the question, should the installer have so many different yaml packages? The installer should probably try to drop to a common yaml package, suggesting the k8s package.

PR #6746, opened 2023-01-05
ARO-1796: Update Azure SDK for go to v63.1.0+incompatible

lranjbar:
Update the Azure SDK library to be compatible with ARO.

[ARO-1796](https://issues.redhat.com//browse/ARO-1796)


SudoBrendan (approved):
LGTM

r4f4:
We usually prefer go mod changes in their own commit: `For the sake of your fellow reviewers, commit vendored code separately from any other changes` [1]. Not sure if it's needed in this case since the change touches only one line.

[1] https://github.com/openshift/installer/blob/master/docs/dev/dependencies.md#go

patrickdillon:
> We usually prefer go mod changes in their own commit: `For the sake of your fellow reviewers, commit vendored code separately from any other changes` [1]. Not sure if it's needed in this case since the change touches only one line.
> 
> [1] https://github.com/openshift/installer/blob/master/docs/dev/dependencies.md#go

+1. Ideally, the changes to go.mod & go.sum would be in their own commit and the vendoring would be in another commit. The code changes for compatibility could be in the first or a separate commit. That said, these are just recommendations/best practices and this PR is already easy enough so let's move it along...

/lgtm
/approve

lranjbar:
> We usually prefer go mod changes in their own commit: `For the sake of your fellow reviewers, commit vendored code separately from any other changes` [1]. Not sure if it's needed in this case since the change touches only one line.
> 
> [1] https://github.com/openshift/installer/blob/master/docs/dev/dependencies.md#go

👍🏼 Ah okay in the future I will split it out 😄

PR #6749, opened 2023-01-06
[WIP] OCPBUGSM-36892: Azure: sorage account created with IPI should restric…

AnnaZivkovic:
…t network access using virtual network rules

Storage Access should restrict network access using virtual network rules to avoid potential security issues from being raised

### DIFF
--- a/data/data/azure/vnet/main.tf
+++ b/data/data/azure/vnet/main.tf
@@ -6,9 +6,6 @@ locals {
     var.azure_extra_tags,
   )
   description = "Created By OpenShift Installer"
-  # At this time min_tls_version is only supported in the Public Cloud and US Government Cloud.
-  environments_with_min_tls_version = ["public", "usgovernment"]
-
 }
 
 provider "azurerm" {
@@ -42,17 +39,6 @@ data "azurerm_resource_group" "network" {
   name = var.azure_network_resource_group_name
 }
 
-resource "azurerm_storage_account" "cluster" {
-  name                            = "cluster${var.random_storage_account_suffix}"
-  resource_group_name             = data.azurerm_resource_group.main.name
-  location                        = var.azure_region
-  account_tier                    = "Standard"
-  account_replication_type        = "LRS"
-  min_tls_version                 = contains(local.environments_with_min_tls_version, var.azure_environment) ? "TLS1_2" : null
-  allow_nested_items_to_be_public = false
-  tags                            = var.azure_extra_tags
-}
-
 resource "azurerm_user_assigned_identity" "main" {
   resource_group_name = data.azurerm_resource_group.main.name
   location            = data.azurerm_resource_group.main.location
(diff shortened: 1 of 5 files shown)


data/data/azure/vnet/main.tf
```diff
@@ -50,6 +50,12 @@ resource "azurerm_storage_account" "cluster" {
   account_replication_type        = "LRS"
   min_tls_version                 = contains(local.environments_with_min_tls_version, var.azure_environment) ? "TLS1_2" : null
   allow_nested_items_to_be_public = false
+
+  network_rules {
+    default_action             = "Allow"
+    virtual_network_subnet_ids = [azurerm_subnet.master_subnet[0].id]
+    bypass                     = ["Logging", "AzureServices"]
```
AnnaZivkovic, line 57: I am unsure what options are necessary to bypass if any.
The list of options can be found [here](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account_network_rules#bypass)

data/data/azure/vnet/main.tf
```diff
@@ -50,6 +50,12 @@ resource "azurerm_storage_account" "cluster" {
   account_replication_type        = "LRS"
   min_tls_version                 = contains(local.environments_with_min_tls_version, var.azure_environment) ? "TLS1_2" : null
   allow_nested_items_to_be_public = false
+
+  network_rules {
+    default_action             = "Allow"
```
patrickdillon, line 55: I think we want to set this to Deny
```suggestion
    default_action             = "Deny"
```
The [docs](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#default_action) are a little confusing, but it says:

> If specifying network_rules, one of either ip_rules or virtual_network_subnet_ids must be specified and default_action must be set to Deny.

So this means the default action will be to deny traffic, unless it is in one of the designated subnets.

data/data/azure/vnet/main.tf
```diff
@@ -50,6 +50,12 @@ resource "azurerm_storage_account" "cluster" {
...
+    virtual_network_subnet_ids = [azurerm_subnet.master_subnet[0].id]
```
patrickdillon, line 56: Let's include the compute subnet too.

data/data/azure/vnet/main.tf
patrickdillon: They're trusted services so let's just include them all--i.e. add `metrics`

data/data/azure/vnet/vnet.tf
```diff
@@ -17,6 +17,7 @@ resource "azurerm_subnet" "master_subnet" {
   ] : cidr.value if cidr.include]
   virtual_network_name = local.virtual_network
   name                 = var.azure_control_plane_subnet
+  service_endpoints    = ["Microsoft.Storage"]
```
patrickdillon, line 20: Hm... why do we need this? What is it doing?

patrickdillon (commented):
> I am looking on some feedback on whether or not we should make using the virtual network customizable in the install config instead of the default?

I don't think we want to make this customizable. The defaults we are setting right now are unnecessarily relax, so we just want to tighten them.

data/data/azure/vnet/vnet.tf
AnnaZivkovic: We need that to link the subnet to the storage account. 
If we don't have it we get 
ERROR Error: retrieving Storage Account: (Name "*" / Resource Group "*-rg"): storage.AccountsClient#GetProperties: Failure responding to request: StatusCode=404 -- Original Error: autorest/azure: Service returned an error. Status=404 Code="StorageAccountNotFound" Message="The storage account clusterocksi was not found."

data/data/azure/vnet/main.tf
AnnaZivkovic: After changing it to Deny, I am running into the following error 
https://github.com/hashicorp/terraform-provider-azurerm/issues/2977

data/data/azure/vnet/main.tf
```diff
@@ -50,6 +50,15 @@ resource "azurerm_storage_account" "cluster" {
   account_replication_type        = "LRS"
   min_tls_version                 = contains(local.environments_with_min_tls_version, var.azure_environment) ? "TLS1_2" : null
   allow_nested_items_to_be_public = false
+
+  dynamic "network_rules" {
```
jhixson74, line 54: This can be done with a single network rule using multiple virtual network ID's rather than multiple network rules. This will also work if somehow there are no subnets. The issue here is figuring out how to get the public IP address of the machine doing the installation. There isn't a clean way of doing it. 

```suggestion
  network_rules {
    default_action             = "Deny"
    virtual_network_subnet_ids = concat(
      [for subnet in azurerm_subnet.master_subnet: subnet.id],
      [for subnet in azurerm_subnet.worker_subnet: subnet.id]
    )
    ip_rules                   = ["<Public IP Address>"]
    bypass                     = ["Logging", "AzureServices", "Metrics"]
  }

```

patrickdillon: +1 It doesn't seem like we need the dynamic block here. There should only be two subnets, so I think you can use the local ID variables:
https://github.com/openshift/installer/blob/3d92c1e6dd1fa1588810ac57223261ff2c74a047/data/data/azure/vnet/common.tf#L35-L36

> The issue here is figuring out how to get the public IP address of the machine doing the installation. There isn't a clean way of doing it.

I'm not certain the installer host needs to access the storage account--at least directly. The storage account does contain boot diagnostic logs, but I assume those are accessed indirectly through azure. If the diagnostic log access is an issue, there has been discussion from the cloud team about moving those to an azure managed service account. There is benefit in doing that because we could then delete this storage account at bootstrap destroy. But that's probably beyond the scope of this bug fix.

jhixson74: It appears as though it is required to specify the IP or IP range of the machine doing the installation. It looks like we can dodge this bullet by using a private endpoint. That looks like the proper route to go here.

patrickdillon: > It appears as though it is required to specify the IP or IP range of the machine doing the installation

I'm not following where or why it is required to get the installer host ip. Can you explain?

`ip_rules` is an optional field: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account_network_rules#ip_rules

data/data/azure/vnet/main.tf
jhixson74: > I think we want to set this to Deny
> 
> The [docs](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_account#default_action) are a little confusing, but it says:
> 
> > If specifying network_rules, one of either ip_rules or virtual_network_subnet_ids must be specified and default_action must be set to Deny.
> 
> So this means the default action will be to deny traffic, unless it is in one of the designated subnets.

In practice, when setting the default to deny without a whitelist IP, this block will fail. After diving into this, it appears we must either whitelist an IP address or range using ip_rules, or create a private endpoint. Creating a private endpoint requires allocating a private IP address and an associated private DNS record. I'm not sure how realistic that will be for some folks.

patrickdillon:
Ah, ok, I see now that several of the issues we are seeing are when we are trying to write to the storage account--that explains why the installer host would need access: to write bootstrap ignition. When thinking about this, I was only considering the reads.

I think it would be perfectly acceptable to just implement the security rule after we have populated the storage account. I tested a [poc here](https://github.com/AnnaZivkovic/installer/pull/1) where I moved the creation of the rules later. The testing generally worked, but I accidentally tested while preserving the bootstrap node. I'm running a test now to see whether bootstrap destroy will work properly.

*BUT* looking at this more closely, I am wondering whether it would make more sense to just get rid of the storage account entirely at bootstrap destroy. At which point, if the storage account is only a temporary resource, we probably don't even need to bother with the security rules... 

The installation storage account contains:
* bootstrap ignition - this is already destroy at bootstrap complete
* rhcos blob 
* boot diagnostic logs

Boot diagnostic logs can be managed by Azure, discussed here: https://redhat-internal.slack.com/archives/C68TNFWA2/p1673372235466479?thread_ts=1673370323.978369&cid=C68TNFWA2

That leaves the RHCOS VHD blob, which we use to create the VM image. I have two questions, which I think if we could answer, would enable us to ditch the storage account:
1. Is it fine to delete the storage blob once the image is created? 
2. If so, how would we delete this along with the bootstrap resources? The storage account would need to be created in advance of the bootstrap stage, so it would not necessarily be related to those resources. @jhixson74 do you have an idea? I'm not that familiar with the terraform destroy mechanism.

PR #6754, opened 2023-01-09
GCP XPN: remove tech preview feature gate

patrickdillon:
GCP XPN, with DNS zones in the service projects, is intended to GA in the next release. Let's remove the feature gate requirement, especially because enabling the feature gate enables the CCM; so cluster behavior is different than what we would be seeing in actual deployments.


patrickdillon:
cc @AnnaZivkovic

barbacbd (commented):
/lgtm

PR #6756, opened 2023-01-10
AGENT-453: Create interactive console service for agent installer

rwsu:
Users would like the ability to change their network configuration at the console if network connectivity problems are detected.

To achieve this goal, this patch adds a new service called agent-interactive-console.service to block the login prompt and the agent services that pulls an image from the registry.

The service will execute the agent TUI to allow users to update their network configuration. The TUI will check there is connectivity to the registry and to the rendezvous host. If the connectivity checks pass, the TUI exits, which also lead the interactive console service to exit, and this unblocks the login prompt and agent services waiting for pull from the registry, allowing the agent-based installer to proceed.

The agent TUI will be added in a future patch.

For now, the service executes a script that logs its presence, sleeps for 60 seconds, and exits. This should not block the automated flow.

Most of the service definition was lifted from @celebdor's POC: https://github.com/openshift/installer/pull/6560/

### DIFF
--- a/data/data/agent/systemd/units/agent-interactive-console.service
+++ b/data/data/agent/systemd/units/agent-interactive-console.service
@@ -2,6 +2,7 @@
 Description=Get interactive user configuration at boot
 After=network-pre.target NetworkManager.service pre-network-manager-config.service selinux.service
 Before=network.target network.service agent.service NetworkManager-wait-online.service
+ConditionPathExists=/usr/local/bin/agent-tui
 
 [Service]
 Type=oneshot
--- a/data/data/agent/systemd/units/agent-interactive-console.service
+++ b/data/data/agent/systemd/units/agent-interactive-console.service
@@ -1,0 +1,19 @@
+[Unit]
+Description=Get interactive user configuration at boot
+After=network-pre.target NetworkManager.service pre-network-manager-config.service selinux.service
+Before=network.target network.service agent.service NetworkManager-wait-online.service
+
+[Service]
+Type=oneshot
+TTYPath=/dev/tty15
+ExecStartPre=/usr/bin/chvt 15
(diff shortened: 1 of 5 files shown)


pkg/asset/agent/image/ignition.go
```diff
@@ -215,6 +215,7 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
 
 	agentEnabledServices := []string{
 		"agent.service",
+		"agent-interactive-console.service",
```
andfasano, line 218: Thanks for naming it `agent-<...>`, in the long term I think it'd be easier to rename all of our services with a known fixed prefix (for the sake of troubleshooting)

data/data/agent/systemd/units/agent-interactive-console.service
```diff
@@ -0,0 +1,19 @@
...
+ExecStart="/usr/local/bin/agent-interactive-console.sh"
+ExecStartPost=/usr/bin/chvt 1
+TimeoutStartSec=0
+StandardInput=tty
+TTYVHangup=yes
+TTYVTDisallocate=yes
+
+[Install]
+WantedBy=default.target
+RequiredBy=sshd.service systemd-logind.service getty@tty1.service
```
andfasano, line 19: Since the service will require the presence of the agent-tui executable (and we already know that there would another service responsible for fetching it from the ISO), what about adding a [systemd Path](https://www.freedesktop.org/software/systemd/man/systemd.path.html) for delaying the activation of the service until the file is available in a given local, ie `/etc/assisted/agent-tui`?

rwsu: I like it and will add the ConditionPathExists.

rwsu: Done. See [Unit] section.

PR #6757, opened 2023-01-10
Shorten SNO installation duration by releasing CVO lease 

eranco74:
When installing SNO with bootstrap in place it takes CVO 6 minutes to acquire the leader lease
This code delete the cvo lease that was created during bootstrapping, allowing cvo to start faster without waiting for the lease to end.

### DIFF
--- a/data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
+++ b/data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
@@ -25,12 +25,15 @@ function signal_bootstrap_complete {
 }
 
 function restart_kubelet {
-  echo "Restarting kubelet"
-  until [ "$(oc get pod -n openshift-kube-apiserver-operator --selector='app=kube-apiserver-operator' -o jsonpath='{.items[0].status.conditions[?(@.type=="Ready")].status}' | grep -c "True")" -eq 1 ];
+  echo "Waiting for kube-apiserver-operator"
+  until [ "$(oc get pod -n openshift-kube-apiserver-operator --selector='app=kube-apiserver-operator' -o jsonpath='{.items[*].status.conditions[?(@.type=="Ready")].status}' | grep -c "True")" -eq 1 ];
   do
     echo "Waiting for kube-apiserver-operator ready condition to be True"
+    oc get --raw='/readyz' &> /dev/null || echo "Api is not available"
     sleep 10
   done
+
+  echo "Restarting kubelet"
   # daemon-reload is required because /etc/systemd/system/kubelet.service.d/20-nodenet.conf is added after kubelet started
   systemctl daemon-reload
   systemctl restart kubelet
@@ -40,26 +43,31 @@ function restart_kubelet {
     echo "Waiting for kube-apiserver to apply the new static pod configuration"
     sleep 10
   done
+
+  echo "Restarting kubelet"
   systemctl restart kubelet
 }
 
 function approve_csr {
-  echo "Approving csrs ..."
+  echo "Waiting for node to report ready status"
   # use [*] and not [0] in the jsonpath because the node resource may not have been created yet
   until [ "$(oc get nodes --selector='node-role.kubernetes.io/master' -o jsonpath='{.items[*].status.conditions[?(@.type=="Ready")].status}' | grep -c "True")" -eq 1 ];
   do
     echo "Approving csrs ..."
     oc get csr -o go-template='{{range .items}}{{if not .status}}{{.metadata.name}}{{"\n"}}{{end}}{{end}}' | xargs --no-run-if-empty oc adm certificate approve &> /dev/null || true
-    sleep 30
+    sleep 10
   done
+  echo "node is ready"
 }
 
 function wait_for_cvo {
   echo "Waiting for cvo"
-  until [ "$(oc get clusterversion -o jsonpath='{.items[0].status.conditions[?(@.type=="Available")].status}')" == "True" ];
+  until [ "$(oc get clusterversion -o jsonpath='{.items[*].status.conditions[?(@.type=="Available")].status}')" == "True" ];
   do
     echo "Still waiting for cvo ..."
-    sleep 30
+    # print the not ready operators names and message
+    oc get clusteroperator -o jsonpath='{range .items[*]}{@.metadata.name}: {range @.status.conditions[?(@.type=="Available")]}{@.type}={@.status} {@.message}{"\n"}{end}{end}' | grep -v "Available=True" || true
+    sleep 20
   done
 }
 
@@ -83,6 +91,7 @@ function clean {
   rm -rf /usr/local/bin/installer-gather.sh
   rm -rf /usr/local/bin/installer-masters-gather.sh
   rm -rf /var/log/log-bundle-bootstrap.tar.gz
+  rm -rf /opt/openshift
 
   systemctl disable bootkube.service
 }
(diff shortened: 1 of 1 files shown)


data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -24,41 +24,73 @@ function signal_bootstrap_complete {
   done
 }
 
+function expedite_bootstrapping {
```
vrutkovs, line 27: lets break it down in several commits? That would allow us reverting specific bits if it gets fixed upstream

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -24,13 +24,33 @@ function signal_bootstrap_complete {
...
+}
+
 function restart_kubelet {
-  echo "Restarting kubelet"
-  until [ "$(oc get pod -n openshift-kube-apiserver-operator --selector='app=kube-apiserver-operator' -o jsonpath='{.items[0].status.conditions[?(@.type=="Ready")].status}' | grep -c "True")" -eq 1 ];
+  echo "Waiting for kube-apiserver-operator"
+  until [ "$(oc get pod -n openshift-kube-apiserver-operator --selector='app=kube-apiserver-operator' -o jsonpath='{.items[*].status.conditions[?(@.type=="Ready")].status}' | grep -c "True")" -eq 1 ];
   do
     echo "Waiting for kube-apiserver-operator ready condition to be True"
+    oc get --raw='/readyz' &> /dev/null || echo "Api is not available"
```
wking, line 49: Is this temporary debugging, or are  you intending to dump `/readyz` output to stdout long-term?

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -24,13 +24,33 @@ function signal_bootstrap_complete {
   done
 }
 
+function release_cvo_lease {
+  if [ ! -f /opt/openshift/release_cvo_lease.done ]
+  then
+    until [ "$(oc get leases -n openshift-cluster-version version | grep -c "version")" -eq 0 ];
```
wking, line 30: I don't understand the need to loop here.  By the time we're at this stage, can we just run a single-shot deletion attempt?

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -25,12 +25,15 @@ function signal_bootstrap_complete {
 }
 
 function restart_kubelet {
-  echo "Restarting kubelet"
-  until [ "$(oc get pod -n openshift-kube-apiserver-operator --selector='app=kube-apiserver-operator' -o jsonpath='{.items[0].status.conditions[?(@.type=="Ready")].status}' | grep -c "True")" -eq 1 ];
+  echo "Waiting for kube-apiserver-operator"
+  until [ "$(oc get pod -n openshift-kube-apiserver-operator --selector='app=kube-apiserver-operator' -o jsonpath='{.items[*].status.conditions[?(@.type=="Ready")].status}' | grep -c "True")" -eq 1 ];
```
wking, line 46: The 55a53adb910e2 commit message says:

> Use [*] and not [0] in the jsonpath

But doesn't explain the benefit of that pivot.  If there  happen to be multiple operator pods,isn't that ok?  I guess we only get one line out of your new JSON Path,  and the logic is moving from "is the randomly-sorted-first pod ready?" to "are there any ready operator pods?"?

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -83,6 +91,7 @@ function clean {
   rm -rf /usr/local/bin/installer-gather.sh
   rm -rf /usr/local/bin/installer-masters-gather.sh
   rm -rf /var/log/log-bundle-bootstrap.tar.gz
+  rm -rf /opt/openshift
```
wking, line 111: This is another change where the 55a53adb910e commit message points out that this is happening, but doesn't explain why the pivot is useful.

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -40,26 +43,31 @@ function restart_kubelet {
...
 
 function wait_for_cvo {
   echo "Waiting for cvo"
-  until [ "$(oc get clusterversion -o jsonpath='{.items[0].status.conditions[?(@.type=="Available")].status}')" == "True" ];
+  until [ "$(oc get clusterversion -o jsonpath='{.items[*].status.conditions[?(@.type=="Available")].status}')" == "True" ];
   do
     echo "Still waiting for cvo ..."
-    sleep 30
+    # print the not ready operators names and message
+    oc get clusteroperator -o jsonpath='{range .items[*]}{@.metadata.name}: {range @.status.conditions[?(@.type=="Available")]}{@.type}={@.status} {@.message}{"\n"}{end}{end}' | grep "Available=False" || true
```
wking, line 86: nit: because the CVO is waiting for (among other things) `Available=True`, you might want `grep -v Available=True` here.  That way any stray `Available=Unknown`, etc. would get included in the output too.

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
eranco74: I'd like to keep it, it's not dumping the  /readyz output to stdout (`&> /dev/null`) it will just output the message as long the API isn't available while we wait for kube-apiserver-operator

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
eranco74: The loop is necessary for cases where the delete command fails, when it just works a single-shot delete would work, in case it fails this loop will check if the lease exists and keep trying to delete it.

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
eranco74: I'll update the commit message

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
eranco74: I made this change because if the pod doesn't exist yet we get a nasty output.
I updated the commit message.

vrutkovs (approved):
/lgtm

PR #6767, opened 2023-01-12
AGENT-455: Check registry and rendezvous host access at startup

rwsu:
Update the interactive console service to have it check the current
host can access the release image and rendezvous host. Display
this connectivity information to the user before prompting to run
the network configuration tui.

If there is connectivity, then the prompt times out after 60s and
the automation flow is allowed to continue.

If there is a connectivity issue, then the network configuration
tui will be executed to allow users to update. The tui, however,
has not been  integrated with the interactive console service, so
in the meantime, there is a 60s sleep in place of executing the tui.
This will be changed in a future patch.

Depends on: https://github.com/openshift/installer/pull/6756

### DIFF
--- a/data/data/agent/files/usr/local/bin/agent-interactive-console.sh
+++ b/data/data/agent/files/usr/local/bin/agent-interactive-console.sh
@@ -1,86 +1,91 @@
 #!/bin/bash
 
 # shellcheck disable=SC1091
-source /etc/assisted/agent-installer.env
+source common.sh
 
 function has_registry_connectivity() {
-    (>&2 echo "INFO: Checking OpenShift release image at $RELEASE_IMAGE is pullable.")
+    (>&2 echo "INFO: Checking that OpenShift release image at $RELEASE_IMAGE is pullable.")
     (>&2 podman pull "$RELEASE_IMAGE")
     registry_check_status_code=$?
     if [[ "$registry_check_status_code" -ne 0 ]]; then
-      (>&2 echo "WARNING: Unable to pull release image.")
-      echo 0
+        (>&2 echo "WARNING: Unable to pull release image.")
+        echo 0
     else
-      (>&2 echo "INFO: Release image is pullable.")
-      echo 1
+        (>&2 echo "INFO: Release image is pullable.")
+        echo 1
     fi
 }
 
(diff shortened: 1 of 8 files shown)


data/data/agent/files/usr/local/bin/agent-interactive-console.sh
```diff
@@ -0,0 +1,86 @@
...
+      (>&2 echo "WARNING: Unable to pull release image.")
+      echo 0
+    else
+      (>&2 echo "INFO: Release image is pullable.")
+      echo 1
+    fi
+}
+
+function has_rendezvous_host_connectivity() {
+    if (>&2 ping -c 4 "$NODE_ZERO_IP"); then
```
andfasano, line 20: Q: Isn't expected to be running on the node zero host?

data/data/agent/files/usr/local/bin/agent-interactive-console.sh
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
+
+# shellcheck disable=SC1091
+source /etc/assisted/agent-installer.env
+
+function has_registry_connectivity() {
+    (>&2 echo "INFO: Checking OpenShift release image at $RELEASE_IMAGE is pullable.")
```
bfournie, line 7: Nit - maybe "Checking *that* Openshift release ..."

bfournie (commented):
Looks good. I tested the normal, success path and I also changed the node's IP to force a ping failure and verified it entered the path to call the agent-tui.

data/data/agent/files/usr/local/bin/agent-interactive-console.sh
```diff
@@ -0,0 +1,86 @@
...
+    (>&2 podman pull "$RELEASE_IMAGE")
+    registry_check_status_code=$?
+    if [[ "$registry_check_status_code" -ne 0 ]]; then
+      (>&2 echo "WARNING: Unable to pull release image.")
```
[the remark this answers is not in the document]
andfasano: I guess we should plan more sophisticated checks in agent-tui only. There was a discussion to keep these initial checks in bash here for the moment, but to gradually move them in the agent-tui code

PR #6769, opened 2023-01-12
OCPBUGS-5782: CVE-2021-4235: Denial of Service in go-yaml

r4f4:
Manual cherry-pick of https://github.com/openshift/installer/pull/6741


barbacbd (commented):
I know this it the fix for the CVE, but there are also newer versions of this software. It looks like if we do decide to move to newer versions though there are some considerations and possible differences in function calls based on the comments here: https://github.com/go-yaml/yaml/commit/7649d4548cb53a614db133b2a8ac1f31859dda8c

However this looks good to me
/lgtm

PR #6771, opened 2023-01-12
GCP: Remove the BYOH code

barbacbd:
** If this is needed then it an always be added back. The gcp byoh code is causing bugs and isn't a requirement. It was added at the same time as the shared vpn code, and the two became entangled, but there was no need for the byoh code without a requirement.

[CORS-2474](https://issues.redhat.com/browse/CORS-2474)


r4f4 (commented):
/lgtm

PR #6776, opened 2023-01-16
OCPBUGS-5461: Allow NO_PROXY to contain IPs

derekhiggins:
Update the validation on NO_PROXY to allow IP addresses to be include.

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -770,9 +770,10 @@ func validateProxy(p *types.Proxy, c *types.InstallConfig, fldPath *field.Path) 
 			v = strings.TrimSpace(v)
 			errDomain := validate.NoProxyDomainName(v)
 			_, _, errCIDR := net.ParseCIDR(v)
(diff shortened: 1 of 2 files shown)


pkg/types/validation/installconfig.go
```diff
@@ -770,9 +770,10 @@ func validateProxy(p *types.Proxy, c *types.InstallConfig, fldPath *field.Path)
 			v = strings.TrimSpace(v)
 			errDomain := validate.NoProxyDomainName(v)
 			_, _, errCIDR := net.ParseCIDR(v)
-			if errDomain != nil && errCIDR != nil {
+			ip := net.ParseIP(v)
+			if errDomain != nil && errCIDR != nil && ip == nil{
```
barbacbd, line 774: shouldn't this be `or` comparison as the error message states?

derekhiggins: No, we enter here (and fail validation) if all 3 conditions report an error, i.e. the value being tested isn't any of the 3 supported types.

PR #6782, opened 2023-01-18
SPLAT-820: multi-zone is no longer tech preview

rvanderp3:
Removing tech preview feature gate from multi-zone install-config spec.


sgaoshang:
@rvanderp3 Looks like we also need to change verbiage here
https://github.com/openshift/installer/blob/610a121833c72b31e8155db476afc5c9f7b11bf2/data/data/install.openshift.io_installconfigs.yaml#L740
https://github.com/openshift/installer/blob/610a121833c72b31e8155db476afc5c9f7b11bf2/data/data/install.openshift.io_installconfigs.yaml#L1395
https://github.com/openshift/installer/blob/610a121833c72b31e8155db476afc5c9f7b11bf2/data/data/install.openshift.io_installconfigs.yaml#L3170

PR #6787, opened 2023-01-20
OCPBUGS-6086: OpenStack: Force JSON content-type in Swift

pierreprinetti:
This is a combined backport of https://github.com/openshift/installer/pull/6628 and https://github.com/openshift/installer/pull/6726

---

With this change, we work around the issue by forcing Gophercloud to request a JSON response from the object storage when listing containers. When passed an `Accept: application/json` header, the server responds with `220 OK` and a `content-type` header in our tests.


mandre (commented):
/lgtm
/approve

PR #6793, opened 2023-01-23
OCPBUGS-6270: Don't require vSphere details for agent installer

zaneb:
Copy the same fix as for the baremetal platform (OCPBUGS-3278) to
vSphere to prevent validation of the presence of fields in the
install-config that are, in fact, ignored by the agent installer.


rwsu (commented):
I verified that it does fix the issue where vSphere details are not provided.

It stuck me that we should add an integration test for this particular vsphere scenario. Here's an example of one we have for the base baremetal compact cluster case: https://github.com/openshift/installer/commit/55d450ad3772998970e9a23b3067a8cd8d0cfda7#diff-400e139b300a043192bc206f891ccf2742aa78301b9a31147fd62bfc519981b3

What do you think?

PR #6794, opened 2023-01-23
AGENT-506: Copy additional files on startup

andfasano:
This patch adds a new agent service responsible to extract and copy the required additional files from the agent ISO.

Requires #6786

### DIFF
--- a/data/data/agent/files/usr/local/bin/agent-copy-iso-files.sh
+++ b/data/data/agent/files/usr/local/bin/agent-copy-iso-files.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# ignition.img is a compressed cpio archive usually containing just the
+# config.ign file. In case of agent-based installation, it could be 
+# enriched with additional files
+
+AGENT_FILES_TEMP="$(mktemp -d)"
+
+cd "${AGENT_FILES_TEMP}" || { echo "Temp folder creation failed"; exit 1; }
+zcat /run/media/iso/images/ignition.img | cpio -idmv
+
+# agent-tui is required by the agent-interactive-console.service
+cp agent-tui /usr/local/bin/
+
+rm -rf "${AGENT_FILES_TEMP}"
--- a/data/data/agent/systemd/units/agent-copy-iso-files.service
+++ b/data/data/agent/systemd/units/agent-copy-iso-files.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Copy the additional files stored in the agent ISO during the boot
+After=network-pre.target NetworkManager.service pre-network-manager-config.service selinux.service
+Before=network.target network.service agent.service NetworkManager-wait-online.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStart=/usr/local/bin/agent-copy-iso-files.sh
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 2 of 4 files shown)


data/data/agent/systemd/units/agent-copy-iso-files.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Copy the additional files stored in the agent ISO during the boot
+Wants=network-online.target
+After=network-online.target
+
```
rwsu, line 5: I think there should be a Before on agent-tui.path.
https://github.com/openshift/installer/blob/master/data/data/agent/systemd/units/agent-tui.path

data/data/agent/files/usr/local/bin/agent-copy-iso-files.sh
```diff
@@ -0,0 +1,15 @@
...
+# config.ign file. In case of agent-based installation, it could be 
+# enriched with additional files
+
+AGENT_FILES_TEMP="$(mktemp -d)"
+
+cd "${AGENT_FILES_TEMP}" || { echo "Temp folder creation failed"; exit 1; }
+zcat /run/media/iso/images/ignition.img | cpio -idmv
+
+# agent-tui is required by the agent-interactive-console.service
+cp agent-tui /usr/local/bin/
```
rwsu, line 13: Is agent-tui already marked executable?

data/data/agent/systemd/units/agent-copy-iso-files.service
andfasano: Not sure I got the rationale. The synchronization point will be in any case set on the presence of `/usr/local/bin/agent-tui`, regardless of the execution ordering of the service - it's not an explicit dependency

rwsu: I was worried that it wouldn't trigger the path dependency in the right time. But it sounds like it ought to work.

data/data/agent/systemd/units/agent-copy-iso-files.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Copy the additional files stored in the agent ISO during the boot
+Wants=network-online.target
+After=network-online.target
```
rwsu, line 4: Is there a reason it needs to wait for network-online? Aren't we just copying files on the file system and not from the network?

andfasano: Not really, just wanted to be sure to be "late" enough in the [boot sequence](https://www.freedesktop.org/software/systemd/man/systemd.special.html), so I guess that in this caso also `basic.target` would work. Do you have any specific suggestion?

rwsu: I'm wondering how it will interact with the agent-interactive-console.service dependencies: 
```
Before=network.target network.service agent.service NetworkManager-wait-online.service
```

andfasano: It seems a fair point, thanks @rwsu. Since it's a service essentially required for making work `agent-interactive-console.service` it seems a good point at least to start them together.

rwsu (commented):
/lgtm

andfasano:
/hold

testing another approach

andfasano:
Not required due https://github.com/openshift/installer/pull/6842

PR #6795, opened 2023-01-24
integration: Don't pass host data in the install-config

zaneb:
This data is no longer required since the fix for OCPBUGS-4874, so don't add it to the install-config in the integration tests. This will allow us to pick up any regression.


bfournie:
Should it also be removed here? https://github.com/openshift/installer/blob/master/cmd/openshift-install/testdata/agent/image/validations/compact_not_enough_hosts.txt#L38-L40

andfasano:
> Should it also be removed here? https://github.com/openshift/installer/blob/master/cmd/openshift-install/testdata/agent/image/validations/compact_not_enough_hosts.txt#L38-L40

It seems so

PR #6796, opened 2023-01-24
AGENT-526: Refactor Agent InstallConfig embedding

zaneb:
Invert the dependencies of the Agent installer's InstallConfig and the IPI/UPI InstallConfig, so that both embed an InstallConfigBase struct, rather the the agent installer overriding parts of the regular InstallConfig.

This allows us to call the validation code from separate places, passing different arguments so that we can run different validations for the agent-based installer (which supports different platform-specific options). Previously this was achieved by parsing the command-line arguments looking for `agent`, which is unreliable without access to the full CLI library and command definitions, and doesn't work in unit tests.

/hold
Depends on #6793 (which will be backported to 4.12, while this will not).


pkg/types/validation/installconfig.go
```diff
@@ -54,7 +54,7 @@ import (
 var pluginsUsingHostPrefix = sets.NewString(string(operv1.NetworkTypeOpenShiftSDN), string(operv1.NetworkTypeOVNKubernetes))
 
 // ValidateInstallConfig checks that the specified install config is valid.
-func ValidateInstallConfig(c *types.InstallConfig) field.ErrorList {
+func ValidateInstallConfig(c *types.InstallConfig, agentMethod bool) field.ErrorList {
```
andfasano, line 57: I understand that this boolean has been kept to maintain the same previous behavior (cherry-picking the specific validations to enable/disable). This refactoring could open the possibility in the future to change approach and prepare the installconfig accordingly before the validation step.

nit: `agentMethod` is a little bit confusing, since it's not a method... I'd suggest to keep the previous `agentBasedInstallation`

PR #6797, opened 2023-01-24
OpenStack: Dual stack support with BYON

MaysaMacedo:
This commit adds dual stack support with bring your own network
for OpenStack platform. The new `ControlPlanePort` field accepts IPv4/IPv6
subnets and the network in the install config, while the `machinesSubnet`
that only supports IPv4 Subnets is deprecated.

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -3057,6 +3057,49 @@ spec:
                       use for instances in this cluster. Deprecated: use FlavorName
                       in DefaultMachinePlatform to define default flavor.'
                     type: string
+                  controlPlanePort:
+                    description: ControlPlanePort contains details of the network
+                      attached to the control plane port, with the network either
+                      containing one openstack subnet for IPv4 or two openstack subnets
+                      for dualstack clusters. Providing this configuration will prevent
+                      OpenShift from managing or updating this network and its subnets.
+                      The network and its subnets will be used during creation of
+                      all nodes. This is a TechPreview feature and requires setting
+                      featureSet to TechPreviewNoUpgrade.
+                    properties:
+                      fixedIPs:
+                        description: Specify subnets of the network where control
+                          plane port will be discovered.
+                        items:
+                          description: FixedIP identifies a subnet defined by a subnet
+                            filter.
+                          properties:
+                            subnet:
+                              description: SubnetFilter defines a subnet by ID and/or
+                                name.
+                              properties:
+                                id:
+                                  type: string
+                                name:
+                                  type: string
+                              type: object
+                          required:
+                          - subnet
+                          type: object
+                        type: array
+                      network:
+                        description: Network is a query for an openstack network that
+                          the port will be discovered on. This will fail if the query
+                          returns more than one network.
+                        properties:
+                          id:
+                            type: string
+                          name:
+                            type: string
+                        type: object
+                    required:
+                    - fixedIPs
+                    type: object
                   defaultMachinePlatform:
                     description: DefaultMachinePlatform is the default configuration
                       used when installing on OpenStack for machine pools which do
@@ -3190,11 +3233,12 @@ spec:
                         type: string
                     type: object
                   machinesSubnet:
-                    description: MachinesSubnet is the UUIDv4 of an openstack subnet.
-                      This subnet will be used by all nodes created by the installer.
-                      By setting this, the installer will no longer create a network
-                      and subnet. The subnet and network specified in MachinesSubnet
-                      will not be deleted or modified by the installer.
+                    description: 'DeprecatedMachinesSubnet is a string of the UUIDv4
+                      of an openstack subnet. This subnet will be used by all nodes
+                      created by the installer. By setting this, the installer will
+                      no longer create a network and subnet. The subnet and network
+                      specified in MachinesSubnet will not be deleted or modified
+                      by the installer. Deprecated: Use ControlPlanePort'
                     type: string
                   octaviaSupport:
                     description: 'OctaviaSupport holds a `0` or `1` value that indicates
--- a/pkg/asset/installconfig/openstack/validation/cloudinfo.go
+++ b/pkg/asset/installconfig/openstack/validation/cloudinfo.go
@@ -27,23 +27,25 @@ import (
 
 	"github.com/openshift/installer/pkg/quota"
 	"github.com/openshift/installer/pkg/types"
+	"github.com/openshift/installer/pkg/types/openstack"
 	openstackdefaults "github.com/openshift/installer/pkg/types/openstack/defaults"
 	"github.com/openshift/installer/pkg/types/openstack/validation/networkextensions"
 )
 
 // CloudInfo caches data fetched from the user's openstack cloud
 type CloudInfo struct {
-	APIFIP            *floatingips.FloatingIP
-	ExternalNetwork   *networks.Network
-	Flavors           map[string]Flavor
-	IngressFIP        *floatingips.FloatingIP
-	MachinesSubnet    *subnets.Subnet
-	OSImage           *images.Image
-	ComputeZones      []string
-	VolumeZones       []string
-	VolumeTypes       []string
-	NetworkExtensions []extensions.Extension
-	Quotas            []quota.Quota
+	APIFIP                  *floatingips.FloatingIP
+	ExternalNetwork         *networks.Network
+	Flavors                 map[string]Flavor
+	IngressFIP              *floatingips.FloatingIP
+	ControlPlanePortSubnets []*subnets.Subnet
+	ControlPlanePortNetwork *networks.Network
+	OSImage                 *images.Image
+	ComputeZones            []string
+	VolumeZones             []string
+	VolumeTypes             []string
+	NetworkExtensions       []extensions.Extension
+	Quotas                  []quota.Quota
 
 	clients *clients
 }
@@ -117,7 +119,7 @@ func GetCloudInfo(ic *types.InstallConfig) (*CloudInfo, error) {
 func (ci *CloudInfo) collectInfo(ic *types.InstallConfig, opts *clientconfig.ClientOpts) error {
 	var err error
 
-	ci.ExternalNetwork, err = ci.getNetwork(ic.OpenStack.ExternalNetwork)
+	ci.ExternalNetwork, err = ci.getNetworkByName(ic.OpenStack.ExternalNetwork)
 	if err != nil {
 		return fmt.Errorf("failed to fetch external network info: %w", err)
 	}
@@ -177,10 +179,16 @@ func (ci *CloudInfo) collectInfo(ic *types.InstallConfig, opts *clientconfig.Cli
 			}
 		}
 	}
-
-	ci.MachinesSubnet, err = ci.getSubnet(ic.OpenStack.MachinesSubnet)
-	if err != nil {
-		return fmt.Errorf("failed to fetch machine subnet info: %w", err)
+	if ic.OpenStack.ControlPlanePort != nil {
+		controlPlanePort := ic.OpenStack.ControlPlanePort
+		ci.ControlPlanePortSubnets, err = ci.getSubnets(controlPlanePort)
+		if err != nil {
+			return err
+		}
+		ci.ControlPlanePortNetwork, err = ci.getNetwork(controlPlanePort.Network.Name, controlPlanePort.Network.ID)
+		if err != nil {
+			return err
+		}
 	}
 
 	ci.APIFIP, err = ci.getFloatingIP(ic.OpenStack.APIFloatingIP)
@@ -226,20 +234,24 @@ func (ci *CloudInfo) collectInfo(ic *types.InstallConfig, opts *clientconfig.Cli
 
 	return nil
 }
-
-func (ci *CloudInfo) getSubnet(subnetID string) (*subnets.Subnet, error) {
-	if subnetID == "" {
-		return nil, nil
-	}
-	subnet, err := subnets.Get(ci.clients.networkClient, subnetID).Extract()
-	if err != nil {
-		if isNotFoundError(err) {
-			return nil, nil
+func (ci *CloudInfo) getSubnets(controlPlanePort *openstack.PortTarget) ([]*subnets.Subnet, error) {
+	controlPlaneSubnets := make([]*subnets.Subnet, 0, len(controlPlanePort.FixedIPs))
+	for _, fixedIP := range controlPlanePort.FixedIPs {
+		page, err := subnets.List(ci.clients.networkClient, subnets.ListOpts{ID: fixedIP.Subnet.ID, Name: fixedIP.Subnet.Name}).AllPages()
+		if err != nil {
+			return controlPlaneSubnets, err
 		}
-		return nil, err
+		subnetList, err := subnets.ExtractSubnets(page)
+		if err != nil {
+			return controlPlaneSubnets, err
+		}
+		if len(subnetList) == 1 {
+			controlPlaneSubnets = append(controlPlaneSubnets, &subnetList[0])
+		} else if len(subnetList) > 1 {
+			return controlPlaneSubnets, fmt.Errorf("found multiple subnets")
+		}
 	}
-
-	return subnet, nil
+	return controlPlaneSubnets, nil
 }
 
 func isNotFoundError(err error) bool {
@@ -285,7 +297,7 @@ func (ci *CloudInfo) getFlavor(flavorName string) (Flavor, error) {
 	}, nil
 }
 
-func (ci *CloudInfo) getNetwork(networkName string) (*networks.Network, error) {
+func (ci *CloudInfo) getNetworkByName(networkName string) (*networks.Network, error) {
 	if networkName == "" {
 		return nil, nil
 	}
@@ -303,6 +315,30 @@ func (ci *CloudInfo) getNetwork(networkName string) (*networks.Network, error) {
 	}
 
 	return network, nil
+}
+
+func (ci *CloudInfo) getNetwork(networkName, networkID string) (*networks.Network, error) {
+	opts := networks.ListOpts{
+		ID:   networkID,
+		Name: networkName,
+	}
+	allPages, err := networks.List(ci.clients.networkClient, opts).AllPages()
+	if err != nil {
+		return nil, err
+	}
+
+	allNetworks, err := networks.ExtractNetworks(allPages)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(allNetworks) == 0 {
+		return nil, nil
+	} else if len(allNetworks) > 1 {
+		return nil, fmt.Errorf("found multiple networks")
+	}
+
+	return &allNetworks[0], nil
 }
 
 func (ci *CloudInfo) getFloatingIP(fip string) (*floatingips.FloatingIP, error) {
--- a/pkg/asset/installconfig/openstack/validation/platform.go
+++ b/pkg/asset/installconfig/openstack/validation/platform.go
@@ -2,13 +2,14 @@ package validation
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"net"
 	"net/url"
 
 	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	utilsslice "k8s.io/utils/strings/slices"
 
 	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/openstack"
@@ -19,8 +20,10 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
 	var allErrs field.ErrorList
 	fldPath := field.NewPath("platform", "openstack")
 
-	// validate BYO machinesSubnet usage
-	allErrs = append(allErrs, validateMachinesSubnet(p, n, ci, fldPath)...)
+	// validate BYO controlPlanePort usage
+	if p.ControlPlanePort != nil {
+		allErrs = append(allErrs, validateControlPlanePort(p, n, ci, fldPath)...)
+	}
 
 	// validate the externalNetwork
 	allErrs = append(allErrs, validateExternalNetwork(p, ci, fldPath)...)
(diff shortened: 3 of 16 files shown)


data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -2815,12 +2815,22 @@ spec:
...
-                    description: MachinesSubnet is the UUIDv4 of an openstack subnet.
+                    description: 'MachinesSubnet is the UUIDv4 of an openstack subnet.
                       This subnet will be used by all nodes created by the installer.
                       By setting this, the installer will no longer create a network
                       and subnet. The subnet and network specified in MachinesSubnet
-                      will not be deleted or modified by the installer.
+                      will not be deleted or modified by the installer. Deprecated:
+                      Use machinesSubnets'
                     type: string
+                  machinesSubnets:
```
EmilienM, line 2825: See my comments here: https://github.com/openshift/enhancements/pull/1319/files#r1085865860

I think we need to discuss of the future of `machineSubnet` and eventually use FailureDomains for BYON advanced config, instead of adding `machineSubnets`. To me they both achieve BYON, so we need to converge I think.

pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -37,26 +38,63 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
...
-				allErrs = append(allErrs, field.InternalError(fldPath.Child("machinesSubnet"), fmt.Errorf("the first CIDR in machineNetwork, %s, doesn't match the CIDR of the machineSubnet, %s", n.MachineNetwork[0].CIDR.String(), ci.MachinesSubnet.CIDR)))
+// validateControlPlanePort validates the machines subnets and network, while enforcing proper byo subnets usage and returns a list of all validation errors.
+func validateControlPlanePort(p *openstack.Platform, n *types.Networking, ci *CloudInfo, fldPath *field.Path) (allErrs field.ErrorList) {
+	networkID := ""
+	if p.ControlPlanePort != nil {
+		if len(p.ControlPlanePort.FixedIPs) <= 2 {
+			if len(p.ExternalDNS) > 0 {
+				allErrs = append(allErrs, field.Invalid(fldPath.Child("externalDNS"), p.ExternalDNS, "externalDNS is set, externalDNS is not supported when ControlPlanePort is set"))
+			}
+			for _, fixedIP := range p.ControlPlanePort.FixedIPs {
```
dulek, line 49: Feels like we should also validate if the subnets are IPv4 and IPv6. I think it's possible to have two IPv4 addresses in a port and I don't think we would want to support that?

pkg/types/conversion/installconfig.go
```diff
@@ -176,6 +176,18 @@ func convertOpenStack(config *types.InstallConfig) error {
 		return err
 	}
 
+	// machinesSubnet has been deprecated in favor of ControlPlanePort
+	controlPlanePort := config.Platform.OpenStack.ControlPlanePort
+	deprecatedMachinesSubnet := config.Platform.OpenStack.DeprecatedMachinesSubnet
+	if deprecatedMachinesSubnet != "" && controlPlanePort == nil {
+		fixedIPs := make([]openstack.FixedIP, 1)
+		fixedIPs[0].Subnet.ID = deprecatedMachinesSubnet
+		config.Platform.OpenStack.ControlPlanePort = &openstack.PortTarget{FixedIPs: fixedIPs}
```
dulek, line 185: Should this emit some warning?

MaysaMacedo: I was following the same pattern done for upconvertVIP, but I can add a warning if you think it's worth. Let me know what you think

dulek: Fair enough, it can be a separate PR if this is a pattern already.

MaysaMacedo:
@dulek Can you take another look? I had to rebase and update a bit the validation to include the Name of the Subnet.

pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -37,26 +38,74 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
...
+				if fixedIP.Subnet.ID != "" && !validUUIDv4(fixedIP.Subnet.ID) {
+					allErrs = append(allErrs, field.InternalError(fldPath.Child("controlPlanePort").Child("fixedIPs"), errors.New("invalid subnet ID")))
+				}
+				subnet := getSubnet(ci.ControlPlanePortSubnets, fixedIP.Subnet.ID, fixedIP.Subnet.Name)
+				if subnet == nil {
+					allErrs = append(allErrs, field.NotFound(fldPath.Child("controlPlanePort").Child("fixedIPs"), fixedIP.Subnet.ID))
```
mandre, line 57: This error message could be confusing if the user specified the subnet by name and not by id.

mandre (commented):
We've already chatted offline about it, but I'm going to mention it here too that we'll need a doc patch to replace `platform.openstack.machinesSubnet` with the new `platform.openstack.controlPlanePort` field.

https://github.com/search?q=repo%3Aopenshift%2Finstaller+path%3Adocs+machinesSubnet&type=code

MaysaMacedo:
@mandre here is the PR for the docs update https://github.com/openshift/installer/pull/7269

pkg/tfvars/openstack/ports.go
```diff
@@ -32,9 +32,13 @@ func portTargetToTerraformPort(networkClient *gophercloud.ServiceClient, portTar
...
-	terraformFixedIPs := make([]terraformFixedIP, len(portTarget.FixedIPs))
-	for i := range terraformFixedIPs {
-		resolvedSubnetID, resolvedNetworkID, err := resolveSubnetFilter(networkClient, networkID, portTarget.FixedIPs[i].Subnet)
+	terraformFixedIPs := make([]terraformFixedIP, 0, len(portTarget.FixedIPs))
+	for _, fixedIP := range portTarget.FixedIPs {
+		subnetFilter := machinev1alpha1.SubnetFilter{
+			ID:   fixedIP.Subnet.ID,
+			Name: fixedIP.Subnet.Name,
+		}
+		resolvedSubnetID, resolvedNetworkID, err := resolveSubnetFilter(networkClient, networkID, subnetFilter)
```
pierreprinetti, line 41: I'd rather change the signature of `resolveSubnetFilter` to match the new subnet filter you defined here in Installer. This helper function is meant to translate to Terraform whatever we have here.

pkg/tfvars/openstack/openstack.go
```diff
@@ -178,31 +177,26 @@ func TFVars(
...
 	}
 
-	// defaultMachinesPort carries the machinesSubnet (and its resolved
-	// network) if provided.
+	// defaultMachinesPort carries the machine subnets and the network.
 	var defaultMachinesPort *terraformPort
-	if machinesSubnet := installConfig.Config.Platform.OpenStack.MachinesSubnet; machinesSubnet != "" {
-		networkID, err := getNetworkFromSubnet(networkClient, machinesSubnet)
+	if controlPlanePort := installConfig.Config.Platform.OpenStack.ControlPlanePort; controlPlanePort != nil {
+		port, err := portTargetToTerraformPort(networkClient, *controlPlanePort)
```
pierreprinetti, line 183: Good call! That `getNetworkFromSubnet` had no reason to exist any more

pierreprinetti (commented):
I have a couple comments on the validation logic, and a question on controlPlanePort.Network: we should be clear about whether the NetworkFilter is required. If it is not required, we must set on a clear rule for when it's allowed. I seem to remember @mdbooth having opinions on this topic.

Good job working around the existing PortTarget implementation. You changed an error message to be specifically about ControlPlanePort in `tfvars/ports.go`, but I think that's fine since we'll be merging https://github.com/openshift/installer/pull/7239 soon anyway.

Thank you for working on this!

pkg/asset/machines/openstack/machines.go
```diff
@@ -141,8 +153,8 @@ func generateProvider(clusterID string, platform *openstack.Platform, mpool *ope
 				Name: portTarget.Network.Name,
 			},
 		}
-		for i := range portTarget.FixedIPs {
-			networkParam.Subnets = append(networkParam.Subnets, machinev1alpha1.SubnetParam{Filter: portTarget.FixedIPs[i].Subnet})
+		for _, fixedIP := range portTarget.FixedIPs {
+			networkParam.Subnets = append(networkParam.Subnets, machinev1alpha1.SubnetParam{Filter: machinev1alpha1.SubnetFilter{ID: fixedIP.Subnet.ID, Name: fixedIP.Subnet.Name}})
```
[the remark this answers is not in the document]
MaysaMacedo: Before the Subnet would be a SubnetFilter, which is not the case at the moment. So, I needed to create one filter and fill it with the appropriate data to get added to NetworkParam.

pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -37,26 +38,78 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
...
+		}
+	}
+	return false
+}
+
+func getSubnet(controlPlaneSubnets []*subnets.Subnet, subnetID, subnetName string) *subnets.Subnet {
+	for _, subnet := range controlPlaneSubnets {
+		if subnet.ID == subnetID {
+			return subnet
+		} else if subnet.Name != "" && subnet.Name == subnetName {
```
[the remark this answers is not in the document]
pierreprinetti: If I'm not mistaken, `subnet.Name` is a property of one of the `subnets.Subnet` in the CloudInfo object, which is fetched from the cloud. Can it be the empty string?

pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -37,26 +38,78 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
...
-		} else {
-			if n.MachineNetwork[0].CIDR.String() != ci.MachinesSubnet.CIDR {
-				allErrs = append(allErrs, field.InternalError(fldPath.Child("machinesSubnet"), fmt.Errorf("the first CIDR in machineNetwork, %s, doesn't match the CIDR of the machineSubnet, %s", n.MachineNetwork[0].CIDR.String(), ci.MachinesSubnet.CIDR)))
+// validateControlPlanePort validates the machines subnets and network, while enforcing proper byo subnets usage and returns a list of all validation errors.
+func validateControlPlanePort(p *openstack.Platform, n *types.Networking, ci *CloudInfo, fldPath *field.Path) (allErrs field.ErrorList) {
+	networkID := ""
+	hasIPv4Subnet := false
+	hasIPv6Subnet := false
+	if p.ControlPlanePort != nil {
+		if len(p.ControlPlanePort.FixedIPs) <= 2 {
```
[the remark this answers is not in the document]
MaysaMacedo: yes.

pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -37,26 +38,78 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
...
+					if subnetDetail == "" {
+						subnetDetail = fixedIP.Subnet.Name
+					}
+					allErrs = append(allErrs, field.NotFound(fldPath.Child("controlPlanePort").Child("fixedIPs"), subnetDetail))
+				} else {
+					if subnet.IPVersion == 4 {
+						hasIPv4Subnet = true
+					} else if subnet.IPVersion == 6 {
+						hasIPv6Subnet = true
+					}
```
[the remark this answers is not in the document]
MaysaMacedo: Given we're checking the IPVersion of the Subnet that was retrieved from OpenStack, I don't think it makes sense to check if another IPVersion was found besides 4 or 6 as this is a validation already done by OpenStack API.
https://docs.openstack.org/api-ref/network/v2/?expanded=create-subnet-detail#subnets

pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -37,26 +38,78 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
...
+						allErrs = append(allErrs, field.InternalError(fldPath.Child("controlPlanePort").Child("fixedIPs"), errors.New("controlPlanePort CIDRs does not match machineNetwork")))
+					}
+					if networkID != "" && networkID != subnet.NetworkID {
+						allErrs = append(allErrs, field.InternalError(fldPath.Child("controlPlanePort").Child("fixedIPs"), errors.New("fixedIPs subnets must be on the same Network")))
+					}
+					networkID = subnet.NetworkID
+				}
 			}
+			if !hasIPv4Subnet && hasIPv6Subnet {
+				allErrs = append(allErrs, field.InternalError(fldPath.Child("controlPlanePort").Child("fixedIPs"), errors.New("one IPv4 subnet must be specified")))
```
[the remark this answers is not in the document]
MaysaMacedo: In this case and for the one bellow I will leave `InternalError`  because `Invalid` requires to specify the details of the exact field that is invalid and in this case if the hole fixedIPs, which does not format really well in the error msg

pkg/asset/installconfig/openstack/validation/cloudinfo.go
```diff
@@ -226,20 +228,24 @@ func (ci *CloudInfo) collectInfo(ic *types.InstallConfig, opts *clientconfig.Cli
...
-		return nil, nil
-	}
-	subnet, err := subnets.Get(ci.clients.networkClient, subnetID).Extract()
-	if err != nil {
-		if isNotFoundError(err) {
-			return nil, nil
+func (ci *CloudInfo) getSubnets(controlPlanePort *openstack.PortTarget) ([]*subnets.Subnet, error) {
+	controlPlaneSubnets := make([]*subnets.Subnet, 0, len(controlPlanePort.FixedIPs))
+	for _, fixedIP := range controlPlanePort.FixedIPs {
+		page, err := subnets.List(ci.clients.networkClient, subnets.ListOpts{ID: fixedIP.Subnet.ID, Name: fixedIP.Subnet.Name}).AllPages()
```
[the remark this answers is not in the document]
MaysaMacedo: Yes. In case the user is still using MachinesSubnet, we want to allow it to be find and the user does not provide any info about the Network with MachinesSubent.

pkg/asset/installconfig/openstack/validation/platform.go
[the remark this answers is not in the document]
MaysaMacedo: yes, if the user created a Port with no Name then Neutron returns the subnet with no name.

pkg/types/openstack/validation/techpreview.go
```diff
@@ -18,5 +18,9 @@ func FilledInTechPreviewFields(installConfig *types.InstallConfig) (fields []*fi
 		fields = append(fields, field.NewPath("controlPlane", "platform", "openstack", "failureDomains"))
 	}
 
+	if installConfig.OpenStack.ControlPlanePort != nil && installConfig.OpenStack.DeprecatedMachinesSubnet == "" {
```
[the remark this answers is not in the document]
MaysaMacedo: The MachinesSubnet is part of the stable API and shouldn't require the `TechPreviewNoUpgrade` in that case.

pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -37,26 +38,78 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
...
-			allErrs = append(allErrs, field.InternalError(fldPath.Child("machinesSubnet"), errors.New("invalid subnet ID")))
-		} else {
-			if n.MachineNetwork[0].CIDR.String() != ci.MachinesSubnet.CIDR {
-				allErrs = append(allErrs, field.InternalError(fldPath.Child("machinesSubnet"), fmt.Errorf("the first CIDR in machineNetwork, %s, doesn't match the CIDR of the machineSubnet, %s", n.MachineNetwork[0].CIDR.String(), ci.MachinesSubnet.CIDR)))
+// validateControlPlanePort validates the machines subnets and network, while enforcing proper byo subnets usage and returns a list of all validation errors.
+func validateControlPlanePort(p *openstack.Platform, n *types.Networking, ci *CloudInfo, fldPath *field.Path) (allErrs field.ErrorList) {
+	networkID := ""
+	hasIPv4Subnet := false
+	hasIPv6Subnet := false
+	if p.ControlPlanePort != nil {
```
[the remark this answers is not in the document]
MaysaMacedo: I'm adding a check to guarantee the fixedIPs exist in the context of NetworkFilter

pkg/types/openstack/platform.go
```diff
@@ -113,11 +113,17 @@ type Platform struct {
 	// +optional
 	IngressVIPs []string `json:"ingressVIPs,omitempty"`
 
-	// MachinesSubnet is the UUIDv4 of an openstack subnet. This subnet will be used by all nodes created by the installer.
+	// DeprecatedMachinesSubnet is a string of the UUIDv4 of an openstack subnet. This subnet will be used by all nodes created by the installer.
 	// By setting this, the installer will no longer create a network and subnet.
 	// The subnet and network specified in MachinesSubnet will not be deleted or modified by the installer.
 	// +optional
-	MachinesSubnet string `json:"machinesSubnet,omitempty"`
+	DeprecatedMachinesSubnet string `json:"machinesSubnet,omitempty"`
```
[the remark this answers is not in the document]
MaysaMacedo: Good point!
But given we're are converting the MachinesSubnet to ControlPlanePort because they represent the same thing, I believe it does not make sense to keep MachinesSubnet without mentioning deprecation of it even when in controlPlanePort is TP, but I'm no specialist here.
Note that with this PR the user can continue to define the MachinesSubnet in install-config normally.

@mandre @EmilienM any thoughts?

pkg/asset/machines/openstack/machines.go
```diff
@@ -113,14 +113,25 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
...
-			Subnets: []machinev1alpha1.SubnetParam{{
-				UUID: platform.MachinesSubnet,
-			}},
+			Subnets: subnets,
+			Filter: machinev1alpha1.Filter{
+				Name: controlPlanePort.Network.Name,
+				ID:   controlPlanePort.Network.ID,
+			},
 		}
+		primarySubnet = controlPlanePort.FixedIPs[0].Subnet.ID
```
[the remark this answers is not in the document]
MaysaMacedo: hmm, I tested one install with that and didn't have any issues.

dulek:
> @dulek Can you take another look? I had to rebase and update a bit the validation to include the Name of the Subnet.

/lgtm

MaysaMacedo:
/cc @patrickdillon

Would you mind to take a look on this PR? Thanks in advance

pkg/types/openstack/platform.go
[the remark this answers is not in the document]
mandre: Depends if we consider the API to be the json key `machinesSubnet`, or the field name `MachinesSubnet`. Ideally users will only interact with the json, but I suspect we also expect to fiddle with the structure directly as the fields are public?

pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -19,8 +19,8 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
 	var allErrs field.ErrorList
 	fldPath := field.NewPath("platform", "openstack")
 
-	// validate BYO machinesSubnet usage
-	allErrs = append(allErrs, validateMachinesSubnet(p, n, ci, fldPath)...)
+	// validate BYO controlPlanePort usage
```
sadasu, line 22: Adding a check to see if ControlPlanePort config will avoid a calll to validate if the config is not present.
```
if p.ControlPlanePort != nil {
        allErrs = append(allErrs, validateControlPlanePort(p, n, ci, fldPath)...)
}
```

pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -37,26 +37,74 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
...
 		}
+		if !hasIPv4Subnet && hasIPv6Subnet {
+			allErrs = append(allErrs, field.InternalError(fldPath.Child("controlPlanePort").Child("fixedIPs"), fmt.Errorf("one IPv4 subnet must be specified")))
+		} else if hasIPv4Subnet && !hasIPv6Subnet && len(p.ControlPlanePort.FixedIPs) == 2 {
+			allErrs = append(allErrs, field.InternalError(fldPath.Child("controlPlanePort").Child("fixedIPs"), fmt.Errorf("multiple IPv4 subnets is not supported")))
+		}
+		controlPlaneNetwork := p.ControlPlanePort.Network
+		if controlPlaneNetwork.ID != "" && controlPlaneNetwork.ID != networkID {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("controlPlanePort").Child("network"), controlPlaneNetwork.ID, "network must contain subnets"))
+		} else if controlPlaneNetwork.Name != "" && ci.ControlPlanePortNetwork.ID != networkID {
```
sadasu, line 80: This check is unclear. Could you please add a comment explaining what is being validated here? 
Also, if controlPlaneNetwork.ID is set, you will not validate controlPlaneNetwork.Name even if it is set? Is that intentional?

pkg/types/conversion/installconfig.go
```diff
@@ -176,6 +176,19 @@ func convertOpenStack(config *types.InstallConfig) error {
...
 
+	// machinesSubnet has been deprecated in favor of ControlPlanePort
+	controlPlanePort := config.Platform.OpenStack.ControlPlanePort
+	deprecatedMachinesSubnet := config.Platform.OpenStack.DeprecatedMachinesSubnet
+	if deprecatedMachinesSubnet != "" && controlPlanePort == nil {
+		fixedIPs := []openstack.FixedIP{{Subnet: openstack.SubnetFilter{ID: deprecatedMachinesSubnet}}}
+		config.Platform.OpenStack.ControlPlanePort = &openstack.PortTarget{FixedIPs: fixedIPs}
+	} else if deprecatedMachinesSubnet != "" &&
+		controlPlanePort != nil {
+		if !(len(controlPlanePort.FixedIPs) == 1 && controlPlanePort.FixedIPs[0].Subnet.ID == deprecatedMachinesSubnet) {
```
sadasu, line 187: Is this check really required? Should we be reporting an error when `deprecatedMachinesSubnet != "" &&
controlPlanePort != nil`? If so, can this test be a part of installconfig validation? We can report an error much earlier in the case of this mis-configuration.

pkg/asset/installconfig/openstack/validation/platform.go
MaysaMacedo: I was trying to follow the pattern of the other validate functions. Is the intention to move this check out of the `validateControlPlanePort` func to reduce the numbers indentations present at `validateControlPlanePort`?

pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -37,26 +37,74 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
...
+					subnetDetail = fixedIP.Subnet.Name
+				}
+				allErrs = append(allErrs, field.NotFound(fldPath.Child("controlPlanePort").Child("fixedIPs"), subnetDetail))
+			} else {
+				if subnet.IPVersion == 6 {
+					hasIPv6Subnet = true
+				} else {
+					hasIPv4Subnet = true
+				}
+				if !isSubnetCIDRInMachineNetworkCIDR(n.MachineNetwork, subnet) {
```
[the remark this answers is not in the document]
MaysaMacedo: go does not provide a built-in method for `Contains` and slice, so I would need to reshape the `isSubnetCIDRInMachineNetworkCIDR`  to that instead. Is that what you had in mind?

pkg/asset/installconfig/openstack/validation/platform.go
MaysaMacedo: `ControlPlanePortNetwork ` retrieves the network from openstack based on the name only. If the ID of the retrieved network does not match the ID of the network of the subnets it means the network provided is a wrong one.
I will add a comment explaining that

pkg/asset/installconfig/openstack/validation/platform.go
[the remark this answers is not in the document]
MaysaMacedo: ah, I just noticed the `utilsslice.Contains`

pkg/types/conversion/installconfig.go
MaysaMacedo: I believe it's required because if the user provided `machinesSubnet` then ran a `openshift-install --dir "$dir" create install-config` the install-config would get the `controlPlanePort`  added.

Similar check also happens for [the VIPs conversion](https://github.com/openshift/installer/blob/master/pkg/types/conversion/installconfig.go#L227-L231).

pkg/asset/installconfig/openstack/validation/platform.go
MaysaMacedo: I ended up reshaping it a bit.

pkg/asset/installconfig/openstack/validation/platform.go
MaysaMacedo: I added a check for existence of `ControlPlanePort`. Regarding the ExternelDNS, if it would be present and we skip the validation, no error would be displayed as no validation would happen, I think it would be better to avoid that.
Let me know if you have a different thought.

MaysaMacedo:
@sadasu thanks for your review. It's ready for another look when you have time

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -3057,6 +3057,50 @@ spec:
                       use for instances in this cluster. Deprecated: use FlavorName
                       in DefaultMachinePlatform to define default flavor.'
                     type: string
+                  controlPlanePort:
+                    description: ControlPlanePort contains the control plane port
```
sadasu, line 3061: Nit: ControlPlanePort contains details of the network attached to the control plane port, with the network either containing one openstack subnet for IPv4 or two openstack subnets for dualstack clusters. Providing this configuration will prevent OpenShift from managing or updating this network and its subnets. The network and its subnets will be used to during creation of all nodes. This is a TechPreview feature and requires setting featureSet to TechPreviewNoUpgrade.

Can we modify this to something like the above? I don't think we want to call out the specific OpenShift component (Installer) in this user facing description.

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -3057,6 +3057,50 @@ spec:
...
+                        type: array
+                      network:
+                        description: Network is a query for an openstack network that
+                          the port will be discovered on. This will fail if the query
+                          returns more than one network.
+                        properties:
+                          id:
+                            type: string
+                          name:
+                            type: string
```
sadasu, line 3099: Is the network required here?

pkg/asset/installconfig/openstack/validation/cloudinfo.go
```diff
@@ -305,6 +317,30 @@ func (ci *CloudInfo) getNetwork(networkName string) (*networks.Network, error) {
...
+
+	if len(allNetworks) == 0 {
+		return nil, nil
+	} else if len(allNetworks) > 1 {
+		return nil, fmt.Errorf("found multiple networks")
+	}
+
+	return &allNetworks[0], nil
+}
+
```
sadasu, line 343: Should we check if the subnets extracted above belong to the networks extracted here?

data/data/install.openshift.io_installconfigs.yaml
MaysaMacedo: No, because `machinesSubnet` only contains info about the subnet and that would be the only info added to `controlPlanePort` during conversion.

pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -37,26 +40,78 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
...
-		} else if !validUUIDv4(p.MachinesSubnet) {
-			allErrs = append(allErrs, field.InternalError(fldPath.Child("machinesSubnet"), errors.New("invalid subnet ID")))
+// validateControlPlanePort validates the machines subnets and network, while enforcing proper byo subnets usage and returns a list of all validation errors.
+func validateControlPlanePort(p *openstack.Platform, n *types.Networking, ci *CloudInfo, fldPath *field.Path) (allErrs field.ErrorList) {
+	networkID := ""
+	hasIPv4Subnet := false
+	hasIPv6Subnet := false
+	if len(p.ExternalDNS) > 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("externalDNS"), p.ExternalDNS, "externalDNS is set, externalDNS is not supported when ControlPlanePort is set"))
+	}
```
sadasu, line 50: return allErrs here.

MaysaMacedo: The idea of not returning `allErrs` here is to allow all the validation of the `controlPlanePort` field to happen and we can return all the errors at once. The same pattern is followed by the validation of the other fields.

Let me know if you would still prefer to return it here.

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -3190,11 +3234,12 @@ spec:
...
-                      This subnet will be used by all nodes created by the installer.
-                      By setting this, the installer will no longer create a network
-                      and subnet. The subnet and network specified in MachinesSubnet
-                      will not be deleted or modified by the installer.
+                    description: DeprecatedMachinesSubnet is a string of the UUIDv4
+                      of an openstack subnet. This subnet will be used by all nodes
+                      created by the installer. By setting this, the installer will
+                      no longer create a network and subnet. The subnet and network
+                      specified in MachinesSubnet will not be deleted or modified
+                      by the installer.
```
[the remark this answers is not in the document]
MaysaMacedo: done

pkg/asset/installconfig/openstack/validation/platform.go
```diff
@@ -37,26 +40,79 @@ func ValidatePlatform(p *openstack.Platform, n *types.Networking, ci *CloudInfo)
 	return allErrs
 }
 
-// validateMachinesSubnet validates the machines subnet and enforces proper byo subnet usage and returns a list of all validation errors
-func validateMachinesSubnet(p *openstack.Platform, n *types.Networking, ci *CloudInfo, fldPath *field.Path) (allErrs field.ErrorList) {
```
sadasu, line 41: This validation function is being removed because the machineSubnets config would be converted to controlPlanePort's network config before it is validated (i.e when the deprecated config is provided by the user and not the new config)?

MaysaMacedo: Yes.
The conversion happens when [loading the install-config](https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/installconfig.go#L118), which happens before [the platform validation.](https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/installconfig.go#L175)

PR #6798, opened 2023-01-24
WIP PowerVS: Check for potential network conflict in existing CCs

hamzy:
When we check the DHCP subnets now, we just check for their existence, because it was only possible to create a network with 192.168.X.0/24.  Now that we can specify a subnet and share a cloud connection, we need to make the quota check more granular. IOW, it should only fail if there is an existing DHCP network with the same subnet we're going to create.

https://issues.redhat.com/browse/MULTIARCH-2752


mjturek:
Looks good, but might warrant a test or two added to `installer/pkg/asset/installconfig/powervs`?

PR #6800, opened 2023-01-25
HIVE-2090: Add clusterID to the end of the resource pool for vsphere

dlom:
Similar to the folder parameter, if a full resource pool path is not provided, it should be constructed with the clusterID as the last part of the path.

xref: [HIVE-2090](https://issues.redhat.com//browse/HIVE-2090)


dlom:
/hold
/cc @abutcher 

Based on the example given in the card, the expected `folder` is `"/VxRail-Datacenter/vm/Carya-Pool-01"`.  The `folder` [is constructed](https://github.com/openshift/installer/blob/8c032f1ab9ae5b5b98e5b8c57b26853654988427/pkg/asset/machines/vsphere/machines.go#L104) using the `clusterID` as the last part of the path.  With this knowledge and the example `resourcePool` given in the card, I have inferred that the `resourcePool` should also have the `clusterID` as the last part of the path.

jcpowermac:
This PR is not possible as currently written. We can't require the clusterID in a resource pool without it being created.

jcpowermac:
cc: @rvanderp3

PR #6806, opened 2023-01-26
WIP: Add service to create registry and push container images

bfournie:
Create a registry and push container images to it so that all mirrored images are resolves locally, removing the need for an external registry.

This is strictly a POC as it relies on files (oc-mirror, registry tarfile) being put to the host, eventually they will be extracted from this ISO. See https://github.com/openshift-metal3/dev-scripts/pull/1509.


bfournie:
Closing this as implementation has been done using appliance
/close

PR #6808, opened 2023-01-26
OCPBUGS-6697: destroy/gcp: Use min length for destroying disks

barbacbd:
** The Length of the disks can be less than the 22 char max. In the case that the cluster name is very small and a minimal name for disks is generated an error was thrown, this will fix that error.

** Altered comment for uses to destroying from creating.


patrickdillon:
/retitle OCPBUGS-6697: destroy/gcp: Use min length for destroying disks

jianli-wei (approved):

PR #6809, opened 2023-01-26
OCPBUGS-6663: Check platform baremetal settings against default values

bfournie:
Spurious warning are being generated for platform baremetal values. Compare the values to default in order to detect the values were set in install-config.

Since the default values for MAC addresses are generated a valid comparison can't be done, so these checks have been removed.

### DIFF
--- a/pkg/asset/agent/installconfig.go
+++ b/pkg/asset/agent/installconfig.go
@@ -11,6 +11,7 @@ import (
 	"github.com/openshift/installer/pkg/asset/installconfig"
 	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/baremetal"
+	baremetaldefaults "github.com/openshift/installer/pkg/types/baremetal/defaults"
 	"github.com/openshift/installer/pkg/types/none"
 	"github.com/openshift/installer/pkg/types/vsphere"
 )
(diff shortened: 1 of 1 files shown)


pkg/asset/agent/installconfig.go
```diff
@@ -187,58 +188,53 @@ func warnUnusedConfig(installConfig *types.InstallConfig) {
...
+		if baremetal.BootstrapProvisioningIP != defaultBM.BootstrapProvisioningIP {
 			fieldPath := field.NewPath("Platform", "Baremetal", "BootstrapProvisioningIP")
 			logrus.Debugf(fmt.Sprintf("%s: %s is ignored", fieldPath, baremetal.BootstrapProvisioningIP))
 		}
-		if baremetal.ExternalBridge != "" {
+		if baremetal.ExternalBridge != defaultBM.ExternalBridge {
 			fieldPath := field.NewPath("Platform", "Baremetal", "ExternalBridge")
 			logrus.Warnf(fmt.Sprintf("%s: %s is ignored", fieldPath, baremetal.ExternalBridge))
 		}
-		if baremetal.ExternalMACAddress != "" {
```
rwsu, line 212: Good catch. I would have assumed the MAC address would have been blank instead of randomized.

rwsu (commented):
/lgtm

bfournie:
Fixed lint issue

rwsu (approved):

pkg/asset/agent/installconfig.go
```diff
@@ -187,58 +188,53 @@ func warnUnusedConfig(installConfig *types.InstallConfig) {
 	switch installConfig.Platform.Name() {
 
 	case baremetal.Name:
+		defaultIc := &types.InstallConfig{Platform: types.Platform{BareMetal: &baremetal.Platform{}}}
```
zaneb, line 191: In the case where the provisioning network is disabled and the ProvisioningIP is set, the default ProvisioningNetworkCIDR is derived from network CIDRs in the install-config. We could avoid a warning in that case by starting with a copy of the InstallConfig, instead of a blank one. (However, there will already be warnings about ProvisioningNetwork=disabled and ProvisioningIP in that case, so this is not a big deal.)

PR #6811, opened 2023-01-27
OCPBUGS-6722: bootimage: move secure execution artifact to separate artifact

mike-nguyen:
Having two artifacts under qemu was causing a failure in s390x CI.
It was not possible to keep the original RHCOS version because the
pipeline already produced a release.json with the qemu-secex nested
under qemu.

Also resolves the following bug for 4.13 (see comments):
OCPBUGS-6509 - Agent service fails to start since cannot pull assisted-installer-agent image

This change was generated using:
```
plume cosa2stream --target data/data/coreos/rhcos.json --distro rhcos --no-signatures --url https://rhcos.mirror.openshift.com/art/storage/prod/streams x86_64=413.86.202301302144-0 aarch64=413.86.202301302144-0 s390x=413.86.202301302144-0 ppc64le=413.86.202301302144-0

### DIFF
--- a/data/data/coreos/rhcos.json
+++ b/data/data/coreos/rhcos.json
@@ -1,95 +1,95 @@
 {
   "stream": "rhcos-4.13",
   "metadata": {
-    "last-modified": "2023-01-20T16:40:08Z",
-    "generator": "plume cosa2stream 0.14.0+445-gaad1da5ac-dirty"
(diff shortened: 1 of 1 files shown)


jlebon (approved):
Looks sane to me!
/lgtm
/approve

PR #6812, opened 2023-01-27
OSASINFRA-3090: External LB support for on-prem platforms

EmilienM:
Support for External LB for BM, vsphere, OpenStack, ovirt and Nutanix platforms

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -2225,6 +2225,28 @@ spec:
                       must be reachable from the host where the installer is run.
                       Default is qemu:///system
                     type: string
+                  loadBalancer:
+                    description: LoadBalancer defines how the load balancer used by
+                      the cluster is configured. LoadBalancer is available in TechPreview.
+                    properties:
+                      type:
+                        default: OpenShiftManagedDefault
+                        description: type defines the type of load balancer used by
+                          the cluster on BareMetal platform which can be a user-managed
+                          or openshift-managed load balancer that is to be used for
+                          the OpenShift API and Ingress endpoints. When set to OpenShiftManagedDefault
+                          the static pods in charge of API and Ingress traffic load-balancing
+                          defined in the machine config operator will be deployed.
+                          When set to UserManaged these static pods will not be deployed
+                          and it is expected that the load balancer is configured
+                          out of band by the deployer. When omitted, this means no
+                          opinion and the platform is left to choose a reasonable
+                          default. The default value is OpenShiftManagedDefault.
+                        enum:
+                        - OpenShiftManagedDefault
+                        - UserManaged
+                        type: string
+                    type: object
                   provisioningBridge:
                     description: Provisioning bridge is used for provisioning nodes,
                       on the host that will run the bootstrap VM.
@@ -2627,6 +2649,28 @@ spec:
                     maxItems: 2
                     type: array
                     uniqueItems: true
+                  loadBalancer:
+                    description: LoadBalancer defines how the load balancer used by
+                      the cluster is configured. LoadBalancer is available in TechPreview.
+                    properties:
+                      type:
+                        default: OpenShiftManagedDefault
+                        description: type defines the type of load balancer used by
+                          the cluster on Nutanix platform which can be a user-managed
+                          or openshift-managed load balancer that is to be used for
+                          the OpenShift API and Ingress endpoints. When set to OpenShiftManagedDefault
+                          the static pods in charge of API and Ingress traffic load-balancing
+                          defined in the machine config operator will be deployed.
+                          When set to UserManaged these static pods will not be deployed
+                          and it is expected that the load balancer is configured
+                          out of band by the deployer. When omitted, this means no
+                          opinion and the platform is left to choose a reasonable
+                          default. The default value is OpenShiftManagedDefault.
+                        enum:
+                        - OpenShiftManagedDefault
+                        - UserManaged
+                        type: string
+                    type: object
                   prismCentral:
                     description: PrismCentral is the endpoint (address and port) and
                       credentials to connect to the Prism Central.
@@ -2871,6 +2915,28 @@ spec:
                       IP in your OpenStack cluster to associate with the OpenShift
                       load balancer. Deprecated: this value has been renamed to apiFloatingIP.'
                     type: string
+                  loadBalancer:
+                    description: LoadBalancer defines how the load balancer used by
+                      the cluster is configured. LoadBalancer is available in TechPreview.
+                    properties:
+                      type:
+                        default: OpenShiftManagedDefault
+                        description: type defines the type of load balancer used by
+                          the cluster on OpenStack platform which can be a user-managed
+                          or openshift-managed load balancer that is to be used for
+                          the OpenShift API and Ingress endpoints. When set to OpenShiftManagedDefault
+                          the static pods in charge of API and Ingress traffic load-balancing
+                          defined in the machine config operator will be deployed.
+                          When set to UserManaged these static pods will not be deployed
+                          and it is expected that the load balancer is configured
+                          out of band by the deployer. When omitted, this means no
+                          opinion and the platform is left to choose a reasonable
+                          default. The default value is OpenShiftManagedDefault.
+                        enum:
+                        - OpenShiftManagedDefault
+                        - UserManaged
+                        type: string
+                    type: object
                   machinesSubnet:
                     description: MachinesSubnet is the UUIDv4 of an openstack subnet.
                       This subnet will be used by all nodes created by the installer.
@@ -3065,6 +3131,28 @@ spec:
                     maxItems: 2
                     type: array
                     uniqueItems: true
+                  loadBalancer:
+                    description: LoadBalancer defines how the load balancer used by
+                      the cluster is configured. LoadBalancer is available in TechPreview.
+                    properties:
+                      type:
+                        default: OpenShiftManagedDefault
+                        description: type defines the type of load balancer used by
+                          the cluster on Ovirt platform which can be a user-managed
+                          or openshift-managed load balancer that is to be used for
+                          the OpenShift API and Ingress endpoints. When set to OpenShiftManagedDefault
+                          the static pods in charge of API and Ingress traffic load-balancing
+                          defined in the machine config operator will be deployed.
+                          When set to UserManaged these static pods will not be deployed
+                          and it is expected that the load balancer is configured
+                          out of band by the deployer. When omitted, this means no
+                          opinion and the platform is left to choose a reasonable
+                          default. The default value is OpenShiftManagedDefault.
+                        enum:
+                        - OpenShiftManagedDefault
+                        - UserManaged
+                        type: string
+                    type: object
                   ovirt_cluster_id:
                     description: The target cluster under which all VMs will run
                     type: string
@@ -3372,6 +3460,28 @@ spec:
                     maxItems: 2
                     type: array
                     uniqueItems: true
+                  loadBalancer:
+                    description: LoadBalancer defines how the load balancer used by
+                      the cluster is configured. LoadBalancer is available in TechPreview.
+                    properties:
+                      type:
+                        default: OpenShiftManagedDefault
+                        description: type defines the type of load balancer used by
+                          the cluster on VSphere platform which can be a user-managed
+                          or openshift-managed load balancer that is to be used for
+                          the OpenShift API and Ingress endpoints. When set to OpenShiftManagedDefault
+                          the static pods in charge of API and Ingress traffic load-balancing
+                          defined in the machine config operator will be deployed.
+                          When set to UserManaged these static pods will not be deployed
+                          and it is expected that the load balancer is configured
+                          out of band by the deployer. When omitted, this means no
+                          opinion and the platform is left to choose a reasonable
+                          default. The default value is OpenShiftManagedDefault.
+                        enum:
+                        - OpenShiftManagedDefault
+                        - UserManaged
+                        type: string
+                    type: object
                   network:
                     description: 'Network specifies the name of the network to be
                       used by the cluster. Deprecated: Use FailureDomains.Topology.Network'
--- a/pkg/asset/manifests/infrastructure.go
+++ b/pkg/asset/manifests/infrastructure.go
@@ -160,6 +160,7 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
 			IngressIP:            installConfig.Config.Platform.BareMetal.IngressVIPs[0],
 			APIServerInternalIPs: installConfig.Config.Platform.BareMetal.APIVIPs,
 			IngressIPs:           installConfig.Config.Platform.BareMetal.IngressVIPs,
+			LoadBalancer:         installConfig.Config.Platform.BareMetal.LoadBalancer,
 		}
 	case gcp.Name:
 		config.Spec.PlatformSpec.Type = configv1.GCPPlatformType
@@ -210,6 +211,7 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
 			IngressIP:            installConfig.Config.OpenStack.IngressVIPs[0],
 			APIServerInternalIPs: installConfig.Config.OpenStack.APIVIPs,
 			IngressIPs:           installConfig.Config.OpenStack.IngressVIPs,
+			LoadBalancer:         installConfig.Config.OpenStack.LoadBalancer,
 		}
 	case vsphere.Name:
 		config.Spec.PlatformSpec.Type = configv1.VSpherePlatformType
@@ -219,8 +221,10 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
 				IngressIP:            installConfig.Config.VSphere.IngressVIPs[0],
 				APIServerInternalIPs: installConfig.Config.VSphere.APIVIPs,
 				IngressIPs:           installConfig.Config.VSphere.IngressVIPs,
+				LoadBalancer:         installConfig.Config.VSphere.LoadBalancer,
 			}
 		}
+
 		config.Spec.PlatformSpec.VSphere = vsphereinfra.GetInfraPlatformSpec(installConfig)
 
 		if _, exists := cloudproviderconfig.ConfigMap.Data["vsphere.conf"]; exists {
@@ -234,6 +238,7 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
 			IngressIP:            installConfig.Config.Ovirt.IngressVIPs[0],
 			APIServerInternalIPs: installConfig.Config.Ovirt.APIVIPs,
 			IngressIPs:           installConfig.Config.Ovirt.IngressVIPs,
+			LoadBalancer:         installConfig.Config.Ovirt.LoadBalancer,
 		}
 	case powervs.Name:
 		config.Spec.PlatformSpec.Type = configv1.PowerVSPlatformType
@@ -303,6 +308,7 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
 				IngressIP:            installConfig.Config.Nutanix.IngressVIPs[0],
 				APIServerInternalIPs: installConfig.Config.Nutanix.APIVIPs,
 				IngressIPs:           installConfig.Config.Nutanix.IngressVIPs,
+				LoadBalancer:         installConfig.Config.Nutanix.LoadBalancer,
 			}
 		}
 	default:
--- a/pkg/tfvars/openstack/openstack.go
+++ b/pkg/tfvars/openstack/openstack.go
@@ -12,6 +12,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/subnets"
 	"github.com/gophercloud/utils/openstack/clientconfig"
 
+	configv1 "github.com/openshift/api/config/v1"
 	machinev1alpha1 "github.com/openshift/api/machine/v1alpha1"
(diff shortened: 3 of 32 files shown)


EmilienM:
/uncc AnnaZivkovic bostrt
/cc rvanderp3 jcpowermac  patrickdillon 

This is highly VIP now but I'll use it as a baseline and apply changes from the enhancement in this PR.

pkg/asset/ignition/machine/node.go
```diff
@@ -35,17 +35,13 @@ func pointerIgnitionConfig(installConfig *types.InstallConfig, rootCA []byte, ro
 		// way to configure DNS before Ignition runs.
 		ignitionHost = net.JoinHostPort(installConfig.BareMetal.APIVIPs[0], "22623")
 	case nutanixtypes.Name:
-		if len(installConfig.Nutanix.APIVIPs) > 0 {
```
mandre, line 38: This conditional seems correct, we should probably add the same check to other on-prem platforms, in case we allow for empty APIVIPs.

pkg/tfvars/openstack/openstack.go
```diff
@@ -211,6 +225,7 @@ func TFVars(
 		MachinesNetwork                   string                            `json:"openstack_machines_network_id,omitempty"`
 		MasterAvailabilityZones           []string                          `json:"openstack_master_availability_zones,omitempty"`
 		MasterRootVolumeAvailabilityZones []string                          `json:"openstack_master_root_volume_availability_zones,omitempty"`
+		ExternalLoadBalancer              bool                              `json:"openstack_external_loadbalancer,omitempty"`
```
mandre, line 228: Oh, I see. You have a typo, `openstack_external_loadbalancer` vs `openstack_external_load_balancer`.

mandre (commented):
Seems like a good start. We'll need to see what type of validation we want to add.

I suggest we iterate on the patch with the OpenStack code first until we're happy with the result then port the same logic to other platforms.

jcpowermac:
@EmilienM if loadBalancerType is external we can check for the existance of the LB and warn if not found.
I would rather you not remove at least that check for vsphere.

I thought this was tech preview? If so it should be added here:
https://github.com/openshift/installer/blob/master/pkg/types/validation/installconfig.go#L985-L1027

A little concerned we are changing how vSphere UPI is installed, @patrickdillon @rvanderp3 wdyt?
The positive for the change is all the onprem platforms will be the same, negative we are adding fields and CoreDNS that isn't required for vSphere UPI.

EmilienM:
> @EmilienM if loadBalancerType is external we can check for the existance of the LB and warn if not found. I would rather you not remove at least that check for vsphere.

This was a mistake, I removed the revert.

> I thought this was tech preview? If so it should be added here: https://github.com/openshift/installer/blob/master/pkg/types/validation/installconfig.go#L985-L1027

Ok, I didn't know that interface. I'll add it here of course.

> A little concerned we are changing how vSphere UPI is installed, @patrickdillon @rvanderp3 wdyt? The positive for the change is all the onprem platforms will be the same, negative we are adding fields and CoreDNS that isn't required for vSphere UPI.

I agree it'll require your users to add `apiVIPs` and `ingresVIPs` fields. And that CoreDNS will be deployed, but I thought it was something we agreed in the enhancement...
I'll defer to @patrickdillon and his team but I think we should make the implementation consistent across all on-prem platforms...

rvanderp3:
> > Right...but as currently implemented you are requiring this change in 4.13 even though this is tech preview.
> 
> I understand the concern and I don't know how to handle it, I'll defer to installer folks to make a decision. I'm ok with whatever option.

I think it makes sense to keep this behind the tech preview feature gate for 4.13.  @jcpowermac and I were under the impression this wasn't targeting 4.13 for GA.

EmilienM:
> > > Right...but as currently implemented you are requiring this change in 4.13 even though this is tech preview.
> > 
> > 
> > I understand the concern and I don't know how to handle it, I'll defer to installer folks to make a decision. I'm ok with whatever option.
> 
> I think it makes sense to keep this behind the tech preview feature gate for 4.13. @jcpowermac and I were under the impression this wasn't targeting 4.13 for GA.

I updated my patch to reflect your proposal, so it's 100% backward compatible with the old behaviour.

EmilienM:
There is still the `sets.String` error (see lint job) to fix , but I'll keep that for later, unless it's done by someone else. It's needed because apimachinery was updated in this PR.

pkg/types/vsphere/platform.go
```diff
@@ -103,6 +107,11 @@ type Platform struct {
 	// FailureDomains is available in TechPreview.
 	// +kubebuilder:validation:Optional
 	FailureDomains []FailureDomain `json:"failureDomains,omitempty"`
+
+	// LoadBalancer defines how the load balancer used by the cluster is configured.
+	// LoadBalancer is available in TechPreview.
+	// +optional
+	LoadBalancer configv1.VSpherePlatformLoadBalancer `json:"loadBalancer,omitempty"`
```
EmilienM, line 114: should this be using a pointer for `VSpherePlatformLoadBalancer` ?

pkg/types/validation/installconfig.go
```diff
@@ -1020,6 +1022,34 @@ func validateFeatureSet(c *types.InstallConfig) field.ErrorList {
 			if len(c.Compute) != 0 && c.Compute[0].Platform.VSphere != nil && len(c.Compute[0].Platform.VSphere.Zones) > 0 {
 				allErrs = append(allErrs, field.Forbidden(field.NewPath("compute", "platform", "vsphere", "zones"), errMsg))
 			}
+
+			if c.VSphere.LoadBalancer.Type != "" {
```
EmilienM, line 1026: I wanted to do:

```
if c.VSphere.LoadBalancer != nil {
  ...
}
```

But this doesn't work, I probably missed something obvious or is it because I'm not using a pointer?

pkg/types/vsphere/platform.go
patrickdillon: Yes, a pointer is preferred for an optional field.

pkg/types/validation/installconfig.go
patrickdillon: > because I'm not using a pointer

+1

EmilienM: thanks

pkg/types/vsphere/platform.go
EmilienM: thanks

EmilienM:
Note for myself, add doc

pkg/asset/manifests/infrastructure.go
```diff
@@ -213,6 +219,10 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
 				IngressIPs:           installConfig.Config.VSphere.IngressVIPs,
 			}
 		}
+		if installConfig.Config.VSphere.LoadBalancer != nil && installConfig.Config.VSphere.LoadBalancer.Type != "" {
```
mandre, line 222: We should set a default value for `installConfig.Config.VSphere.LoadBalancer.Type` to  `OpenShiftManagedDefault` in [SetPlatformDefaults()](https://github.com/openshift/installer/blob/master/pkg/types/vsphere/defaults/platform.go#L9) (or [installconfig.go](https://github.com/openshift/installer/blob/master/pkg/types/defaults/installconfig.go) if you want to do it globally for all on-prem platforms) so that this value is never empty, right? Or is there a reason to check for empty values?

pkg/tfvars/openstack/openstack.go
```diff
@@ -95,6 +96,23 @@ func TFVars(
...
+	if installConfig.Config.Platform.OpenStack.LoadBalancer == nil {
+		userManagedLoadBalancer = false
+	} else {
+		loadBalancerType := installConfig.Config.Platform.OpenStack.LoadBalancer.Type
+		switch loadBalancerType {
+		case configv1.LoadBalancerTypeUserManaged:
+			userManagedLoadBalancer = true
+		case configv1.LoadBalancerTypeOpenShiftManagedDefault:
+			userManagedLoadBalancer = false
+		case "":
```
mandre, line 109: How is empty a valid value? We should have validated install-config at this point, and set a valid default too.

That way, you only need to check for `installConfig.Config.Platform.OpenStack.LoadBalancer.Type == configv1.LoadBalancerTypeUserManaged` to get `userManagedLoadBalancer`.

pkg/asset/manifests/infrastructure.go
```diff
@@ -153,6 +153,9 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
 			APIServerInternalIPs: installConfig.Config.Platform.BareMetal.APIVIPs,
 			IngressIPs:           installConfig.Config.Platform.BareMetal.IngressVIPs,
 		}
+		if installConfig.Config.BareMetal.LoadBalancer != nil {
+			config.Status.PlatformStatus.BareMetal.LoadBalancer = installConfig.Config.BareMetal.LoadBalancer
```
mandre, line 157: Shouldn't the installer always fill the load balancer config into the PlatformStatus? We should set a default for installConfig.Config.BareMetal.LoadBalancer so that we it's always set and remove the above test.

EmilienM: Since the feature is in TechPreview and it's not in the stable CRD, we have a validation later that checks for the install-config field to be set only if we are deploying with TechPreview featureSet. Therefore, I think we need this check for now and we should be able to remove it once it's promoted to stable.

pkg/asset/manifests/infrastructure.go
EmilienM: Like I explained later, this field doesn't exist on stable CRD (see my API PR), so I don't think we can set a default here usable by both stable & TechPreview deployments. I think we should leave it that way for now.

Maybe we can confirm with @patrickdillon whether it's good practice or not.

pkg/tfvars/openstack/openstack.go
EmilienM: Right. I'll remove the empty case. Good point.

pkg/asset/manifests/infrastructure.go
mandre: Oh good point, it didn't occur to me that because it was tech preview we didn't want to set platform status for clusters unless they enabled the tech preview. That said, allowing empty values is confusing as hell - can we use a default _when_ the user enables tech preview?

EmilienM: The default is at the API level, I'm not sure we want/need a default here.

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -3365,6 +3453,28 @@ spec:
                     maxItems: 2
                     type: array
                     uniqueItems: true
+                  loadBalancer:
+                    description: LoadBalancer defines how the load balancer used by
+                      the cluster is configured. LoadBalancer is available in TechPreview.
+                    properties:
+                      type:
+                        default: OpenShiftManagedDefault
+                        description: type defines the type of load balancer used by
```
rvanderp3, line 3462: Perhaps we want to mention in the description that this is tech preview

EmilienM: can we? I literally copy/pasted from the original CRD.

pierreprinetti (commented):
IIUC, this removes the API VIP and the FIPs when you set the external LB in install-config?

LGTM, with a couple nits on the testing part.

scripts/openstack/manifest-tests/user-managed-lb/test_cluster-infra.py
```diff
@@ -0,0 +1,33 @@
...
+class TestClusterInfraObject(unittest.TestCase):
+    def setUp(self):
+        """Parse the Cluster Infrastructure object into a Python data structure."""
+        self.machines = []
+        cluster_infra = f'{ASSETS_DIR}/manifests/cluster-infrastructure-02-config.yaml'
+        with open(cluster_infra) as f:
+            self.cluster_infra = yaml.load(f, Loader=yaml.FullLoader)
+
+
+    def failure_domains(self):
```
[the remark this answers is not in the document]
EmilienM: a copy/paste, I'll fix it.

EmilienM:
> IIUC, this removes the API VIP and the FIPs when you set the external LB in install-config?

This addresses what is described in https://github.com/openshift/enhancements/blob/master/enhancements/network/external-lb-vips.md

> LGTM, with a couple nits on the testing part.

:pray:

scripts/openstack/manifest-tests/user-managed-lb/test_cluster-infra.py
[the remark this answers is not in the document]
pierreprinetti: also: your test must be called `test_*`

pierreprinetti (approved):

pkg/types/validation/installconfig.go
```diff
@@ -996,5 +1008,91 @@ func validateFeatureSet(c *types.InstallConfig) field.ErrorList {
...
+				}
+			}
+		}
+		if c.Ovirt != nil {
+			if c.Ovirt.LoadBalancer != nil {
+				if !validateLoadBalancer(c.Ovirt.LoadBalancer.Type) {
+					allErrs = append(allErrs, field.Invalid(field.NewPath("platform", "ovirt", "loadBalancer", "type"), c.Ovirt.LoadBalancer.Type, "invalid load balancer type"))
+				}
+				if c.Ovirt.LoadBalancer.Type == configv1.LoadBalancerTypeOpenShiftManagedDefault && len(c.Ovirt.APIVIPs) == 0 {
+					allErrs = append(allErrs, field.Invalid(field.NewPath("platform", "ovirt", "apiVIPs"), c.Ovirt.APIVIPs, "must be set when using the OpenShiftManaged default load balancer"))
```
pierreprinetti, line 1049: What do you think of `field.Required` instead of `Invalid`? Invalid seems to be about a malformed value, while here it's empty and that's the problem.

EmilienM: very good idea!

data/data/install.openshift.io_installconfigs.yaml
patrickdillon: The description for `loadBalancer` includes that this is TechPreview, so I think we're good here :+1:

pkg/types/validation/installconfig.go
```diff
@@ -996,5 +1008,91 @@ func validateFeatureSet(c *types.InstallConfig) field.ErrorList {
...
+		if c.VSphere != nil {
+			if c.VSphere.LoadBalancer != nil {
+				if !validateLoadBalancer(c.VSphere.LoadBalancer.Type) {
+					allErrs = append(allErrs, field.Invalid(field.NewPath("platform", "vsphere", "loadBalancer", "type"), c.VSphere.LoadBalancer.Type, "invalid load balancer type"))
+				}
+				if c.VSphere.LoadBalancer.Type == configv1.LoadBalancerTypeOpenShiftManagedDefault && len(c.VSphere.APIVIPs) == 0 {
+					allErrs = append(allErrs, field.Required(field.NewPath("platform", "vsphere", "apiVIPs"), "must be set when using the OpenShiftManaged default load balancer"))
+				}
+			}
+		}
```
patrickdillon, line 1062: These validations should probably be moved to the platform-specific validations ([example](https://github.com/openshift/installer/blob/master/pkg/types/openstack/validation/platform.go) or at least they shouldn't be coupled with the featureset validations. 

There should be one validation to enforce that the field can only be set when tech preview is enabled, which is what you have in the following `else` block. The rest of the validations don't need to worry about tech preview, because they should only be run when a value is set.

patrickdillon (commented):
This looks great. I did leave a lot of nits, which you are free to ignore. I also didn't go into depth on the openstack specific bits. I think there are a couple of changes needed, but I think they should be rather straightforward. Thanks again!

patrickdillon:
> > > Right...but as currently implemented you are requiring this change in 4.13 even though this is tech preview.
> > 
> > 
> > I understand the concern and I don't know how to handle it, I'll defer to installer folks to make a decision. I'm ok with whatever option.
> 
> I think it makes sense to keep this behind the tech preview feature gate for 4.13. @jcpowermac and I were under the impression this wasn't targeting 4.13 for GA.

@jcpowermac @rvanderp3 @EmilienM  sorry I missed that you were waiting for me to weigh in here. Reading through this, it is unclear to me whether we're on the same page about vsphere UPI running coreDNS. Is there a resolution to this?

From an install config API point of view, we can't start requiring a field like `APIVIPs` that wasn't required before. Although we could conditionally require it when a new optional field is set. 

It seems like the default value for vSphere (at least upi, but there isn't really a distinction) must be user managed. And the validation only enforces APIVIPs when the openshift managed default is selected. So if we're ok with that default, it seems like we're fine or am I missing something?

pkg/types/validation/installconfig.go
EmilienM: I've put them here because vsphere had some validations for the Zones work (that have been removed in the meantime).
I'll move them per platform then.

patrickdillon:
/lgtm cancel 

I think @EmilienM is going to push some new changes, so this would get lost anyway.

/approve

pkg/asset/manifests/infrastructure.go
```diff
@@ -154,6 +154,9 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
...
+		}
```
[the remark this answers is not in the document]
EmilienM: That seems to work indeed, thanks!

pkg/asset/manifests/infrastructure.go
```diff
@@ -204,6 +207,9 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
 			APIServerInternalIPs: installConfig.Config.OpenStack.APIVIPs,
 			IngressIPs:           installConfig.Config.OpenStack.IngressVIPs,
 		}
+		if installConfig.Config.OpenStack.LoadBalancer != nil {
```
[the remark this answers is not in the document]
EmilienM: That seems to work indeed, thanks!

pkg/tfvars/openstack/openstack.go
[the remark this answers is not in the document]
EmilienM: the default being OpenShiftManaged, I'll set `userManagedLoadBalancer = false` for ""

pkg/tfvars/openstack/openstack.go
```diff
@@ -95,6 +96,23 @@ func TFVars(
...
+			panic("Load balancer type is empty and should be set to a valid value")
+		default:
+			panic("Invalid load balancer type: " + loadBalancerType)
```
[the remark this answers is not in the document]
EmilienM: I'll leave the code as is for this one:
* An extra validation wouldn't hurt I think
* We have code above that which calls `panic` already, so I followed the pattern.

pkg/tfvars/openstack/openstack.go
```diff
@@ -95,6 +96,23 @@ func TFVars(
...
+		}
+	}
```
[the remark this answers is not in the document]
EmilienM: nice!

pkg/tfvars/openstack/openstack.go
```diff
@@ -211,6 +229,7 @@ func TFVars(
 		MachinesNetwork                   string                            `json:"openstack_machines_network_id,omitempty"`
 		MasterAvailabilityZones           []string                          `json:"openstack_master_availability_zones,omitempty"`
 		MasterRootVolumeAvailabilityZones []string                          `json:"openstack_master_root_volume_availability_zones,omitempty"`
+		UserManagedLoadBalancer           bool                              `json:"openstack_user_managed_load_balancer,omitempty"`
```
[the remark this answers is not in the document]
EmilienM: :+1:

EmilienM:
Note for reviewers, I'm still working on moving the validations into the platform directories. I did BM already but I'm not done yet.

EmilienM:
> > > > > Right...but as currently implemented you are requiring this change in 4.13 even though this is tech preview.
> 
> > @jcpowermac @rvanderp3 @EmilienM sorry I missed that you were waiting for me to weigh in here. Reading through this, it is unclear to me whether we're on the same page about vsphere UPI running coreDNS. Is there a resolution to this?
> 
> We decided that for now we don't change anything with CoreDNS:
> 
>     * vsphere UPI will still live without CoreDNS because the API VIPs are empty in PlatformStatus.
> 
>     * When using a User Managed LB (on any on-prem), CoreDNS will be deployed, except for vsphere as explained in the first point.
> 
> 
> > From an install config API point of view, we can't start requiring a field like `APIVIPs` that wasn't required before. Although we could conditionally require it when a new optional field is set.
> 
> ack.
> 
> > It seems like the default value for vSphere (at least upi, but there isn't really a distinction) must be user managed. And the validation only enforces APIVIPs when the openshift managed default is selected. So if we're ok with that default, it seems like we're fine or am I missing something?
> 
> I'm fine with it. I'll apply that change now.

In fact, since we had no validations for the VIPs before, I won't add it here in this PR. This needs to be discussed and I think it's out of scope for this effort.

EmilienM:
/test e2e-metal-ipi
/test e2e-vsphere-upi
/test e2e-nutanix-sdn 

The OpenStack failures are not related.
BM, VSphere and Nutanix seems flacky, but let's retest in case.

EmilienM:
/hold cancel
I saw installs working on vsphere, BM and openstack, I think this is good to go.

EmilienM:
> /retest-required

we have to wait for https://github.com/openshift/origin/pull/27736

EmilienM:
/hold cancel
no need to hold, CI jobs are voting anyway.

PR #6814, opened 2023-01-30
CORS-2496: Default `2xlarge` instead of `xlarge` in AWS when the control-plane is single-node

omertuc:
# Background

* Starting with version 4.13 OCP is going to officially support Single
  Node clusters on AWS.

* The minimum documented OCP requirement for single-node control plane
  nodes is 8-cores and 16GiB of RAM

* The current default instance type chosen for AWS clusters by the
  installer is `xlarge` which is 4 cores and 16GiB of RAM

# Issue

The default instance type the installer currently chooses for Single
Node Openshift clusters doesn't follow our documented minimum
requirements

# Solution

When the number of replicas of the ControlPlane pool is 1, the installer
will now choose `2xlarge` instead of `xlarge`.

# Caveat

`2xlarge` has 32GiB of RAM, which is twice as much as we need, but it's
the best we can do to meet the minimum single-node requirements, because
AWS doesn't offer a 16GiB RAM instance type with 8 cores.

### DIFF
--- a/pkg/asset/machines/worker.go
+++ b/pkg/asset/machines/worker.go
@@ -182,11 +182,12 @@ func defaultNutanixMachinePoolPlatform() nutanixtypes.MachinePool {
 	}
 }
 
-func awsDefaultMachineTypes(region string, arch types.Architecture) []string {
+func awsDefaultMachineTypes(region string, arch types.Architecture, instanceSize string) []string {
 	classes := awsdefaults.InstanceClasses(region, arch)
+
 	types := make([]string, len(classes))
 	for i, c := range classes {
-		types[i] = fmt.Sprintf("%s.xlarge", c)
+		types[i] = fmt.Sprintf("%s.%s", c, instanceSize)
 	}
 	return types
 }
@@ -340,11 +341,13 @@ func (w *Worker) Generate(dependencies asset.Parents) error {
 					zoneDefaults = true
 				}
 			}
+
 			if mpool.InstanceType == "" {
-				mpool.InstanceType, err = aws.PreferredInstanceType(ctx, installConfig.AWS, awsDefaultMachineTypes(installConfig.Config.Platform.AWS.Region, installConfig.Config.ControlPlane.Architecture), mpool.Zones)
+				instanceSize := defaultAWSInstanceSize
+				mpool.InstanceType, err = aws.PreferredInstanceType(ctx, installConfig.AWS, awsDefaultMachineTypes(installConfig.Config.Platform.AWS.Region, installConfig.Config.ControlPlane.Architecture, instanceSize), mpool.Zones)
 				if err != nil {
 					logrus.Warn(errors.Wrap(err, "failed to find default instance type"))
-					mpool.InstanceType = awsDefaultMachineTypes(installConfig.Config.Platform.AWS.Region, installConfig.Config.ControlPlane.Architecture)[0]
+					mpool.InstanceType = awsDefaultMachineTypes(installConfig.Config.Platform.AWS.Region, installConfig.Config.ControlPlane.Architecture, instanceSize)[0]
 				}
 			}
 			// if the list of zones is the default we need to try to filter the list in case there are some zones where the instance might not be available
--- a/pkg/asset/machines/master.go
+++ b/pkg/asset/machines/master.go
@@ -112,6 +112,12 @@ const (
 	controlPlaneMachineSetFileName = "99_openshift-machine-api_master-control-plane-machine-set.yaml"
 )
 
+// AWS specific constants.
+const (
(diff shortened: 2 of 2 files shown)


barbacbd (commented):
This may only require a docs update for the new feature. The install config may have the ability to alter these fields.

pkg/asset/machines/worker.go
```diff
@@ -182,11 +182,22 @@ func defaultNutanixMachinePoolPlatform() nutanixtypes.MachinePool {
...
-func awsDefaultMachineTypes(region string, arch types.Architecture) []string {
+func awsDefaultMachineTypes(region string, arch types.Architecture, singleNodeControlPlane bool) []string {
 	classes := awsdefaults.InstanceClasses(region, arch)
+
+	// If the control plane is single node, we need to use a larger
+	// instance type for that node, as the minimum requirement for
+	// single-node control-plane nodes is 8 cores, and xlarge only has 4.
+	// Unfortunately 2xlarge has twice as much RAM as we need, but we
+	// default to it because AWS doesn't offer an 8-core 16GiB instance type
+	singleNodeTopologySizes := map[bool]string{
```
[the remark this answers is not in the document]
omertuc: of course, don't know what I was thinking

omertuc:
> This may only require a docs update for the new feature. 

What kind of docs are you thinking of? The docs team are already aware they need to be formally documenting AWS single-node, if that what you had in mind

>The install config may have the ability to alter these fields.

What do you mean by that? Is that related to your docs remark?

barbacbd:
> > This may only require a docs update for the new feature.
> 
> What kind of docs are you thinking of? The docs team are already aware they need to be formally documenting AWS single-node, if that what you had in mind
> 
> > The install config may have the ability to alter these fields.
> 
> What do you mean by that? Is that related to your docs remark?

Yes the docs team update could be all that we need here and my comment was related to the docs remark. @patrickdillon  may know more but there is a field here @omertuc does this field help ?

omertuc:
> > > This may only require a docs update for the new feature.
> > 
> > 
> > What kind of docs are you thinking of? The docs team are already aware they need to be formally documenting AWS single-node, if that what you had in mind
> > > The install config may have the ability to alter these fields.
> > 
> > 
> > What do you mean by that? Is that related to your docs remark?
> 
> Yes the docs team update could be all that we need here and my comment was related to the docs remark. @patrickdillon may know more but there is a field here @omertuc does this field help ?

Oh I see what you mean now, missed the word "only" in your original message.

This could be a simple docs update, and yes the user could simply specify the machine type in the install-config.yaml, but on the other hand it's a bit odd that our official installer chooses defaults that don't meet our documented requirements, and this is what this PR aims to solve.

I don't insist on including it, but I personally think it's nice to have

r4f4:
I think it makes sense to have a default that at least satisfies the minimum requirements. And in this case it won't affect the multi-node workflow, so it should be safe.

pkg/asset/machines/worker.go
```diff
@@ -182,11 +182,22 @@ func defaultNutanixMachinePoolPlatform() nutanixtypes.MachinePool {
 	}
 }
 
-func awsDefaultMachineTypes(region string, arch types.Architecture) []string {
+func awsDefaultMachineTypes(region string, arch types.Architecture, singleNodeControlPlane bool) []string {
```
r4f4, line 185: Hmm. I wonder if it doesn't make more sense to do:
```suggestion
func awsDefaultMachineTypes(region string, arch types.Architecture, instanceSize string) []string {
```
and then pass different sizes for controlPlane single node, controlPlane multi node, and compute. Practically it makes no difference, but I find it more clear semantically.
That way you could define in `pkg/asset/machines/master.go`:
```
const (
    defaultInstanceSize string = "xlarge"
    defaultSingleNodeInstanceSize string = "2xlarge"
)

[...]

instanceSize := defaultInstanceSize
if pool.Replicas != nil && *pool.Replicas == 1 {
    instanceSize = defaultSingleNodeInstanceSize
}
[...]
awsDefaultMachineType(..., instanceSize)
```

and for the worker in `pkg/asset/machines/worker.go`:
```
const defaultInstanceSize string = "xlarge"

[...]

awsDefaultMachineTypes(..., defaultInstanceSize)
```

omertuc: I had to make some adjustments to your suggestion for two reasons:

* worker.go and master.go share the same package so they would have to have a different prefix given to them
* worker.go and master.go apply to all clouds, while these constants are AWS specific

r4f4 (commented):
/lgtm

r4f4:
/lgtm cancel
There is a small lint issue to be fixed.

PR #6816, opened 2023-01-31
OCPBUGS-6486: fixing ovirt installation on iSCSI

janosdebugs:
This PR adds an update to go-ovirt-client, which fixes a bug when installing on iSCSI on non-preallocated disks.

Fixes OCPBUGS-6486


r4f4:
/lgtm
e2e-ovirt-sdn is green (I was just waiting for at least a successful install)

PR #6817, opened 2023-01-31
OCPBUGS-6840: Get hostname if not configured in manifests

bfournie:
When using ZTP manifests there is no mechanism for manually configuring the hostname, the hostname must be retrieved via address lookup. Run the address lookp in the set-hostname script once the interfaces are available and configure the hostname if it can be retrieved.


data/data/agent/files/usr/local/bin/set-hostname.sh
```diff
@@ -35,7 +36,29 @@ do
         echo "Host has matching MAC address: ${filename}" 1>&2
         echo "Setting hostname to ${HOSTNAME}" 1>&2
         hostnamectl set-hostname "${HOSTNAME}"
+	hostname_set=1
     else
         echo "MAC address, ${filename}, does not exist on this host" 1>&2
     fi
 done
+
+if [[ $hostname_set -eq 0 ]]; then
```
andfasano, line 45: Just for the sake of clarity, it could be useful to move this piece into a separate function,  and add some comments about the workaround

data/data/agent/files/usr/local/bin/set-hostname.sh
```diff
@@ -35,7 +36,29 @@ do
...
+      ips=$(hostname -I)
+   done
+
+   for ip in ${ips}
+   do
+      echo "Reverse dns for $ip" 1>&2
+      hostname=$(dig -x $ip +short)
+      # If hostname returned, configure it
+      if [[ $hostname != "" ]]; then
+         hostnamectl set-hostname "${hostname}"
```
andfasano, line 60: what if multiple names are returned?

data/data/agent/files/usr/local/bin/set-hostname.sh
```diff
@@ -35,7 +36,29 @@ do
...
+         echo "Setting hostname to ${hostname}" 1>&2
```
andfasano, line 61: If the name was set, could be the loop interrupted (instead of trying renaming again the host with remaining entries)?

data/data/agent/files/usr/local/bin/set-hostname.sh
bfournie: Changed to handle multiple names and use the first one.

data/data/agent/files/usr/local/bin/set-hostname.sh
bfournie: Done

bfournie:
> @bfournie is the sno-ipv6 still failing due another reason maybe?

Yes this fix doesn't help with the agent-sno-ipv6 test, that's failing due to an MCO bug tracked as critical - https://issues.redhat.com/browse/OCPBUGS-5888.

This fixes the validation failure due the the hostname being 'localhost' when one doesn't get assigned. In 4.13 this validation failure is masked because of the assisted-installer-agent change https://github.com/openshift/assisted-installer-agent/pull/449 that assigns hostnames based on the mac address, but we end up with an unexpected hostname in 4.13 and a validation failure in 4.12 since the assisted-installer-agent fix isn't backported.

bfournie:
/close
The NetworkManager fix for https://bugzilla.redhat.com/show_bug.cgi?id=2166711 should resolve this IPv6 name resolution problem so closing this

PR #6818, opened 2023-01-31
OCPBUGS-6725: In wait-for bootstrap, prevent repeating same error message

bfournie:
If the Kube API takes a while to come up we can end up with the message 
'INFO cluster bootstrap did not complete'
repeated every 2 seconds as shown in the bug.

Prevent the same message from being output continually in the wait-for bootstrap

### DIFF
--- a/pkg/agent/waitfor.go
+++ b/pkg/agent/waitfor.go
@@ -18,7 +18,7 @@ func WaitForBootstrapComplete(cluster *Cluster) error {
 	waitContext, cancel := context.WithTimeout(cluster.Ctx, timeout)
 	defer cancel()
 
-	var lastErr error
+	var lastErr, lastErrOnExit error
 	wait.Until(func() {
 		bootstrap, exitOnErr, err := cluster.IsBootstrapComplete()
 		if bootstrap && err == nil {
@@ -28,10 +28,13 @@ func WaitForBootstrapComplete(cluster *Cluster) error {
 
 		if err != nil {
 			if exitOnErr {
-				lastErr = err
+				lastErrOnExit = err
 				cancel()
 			} else {
-				logrus.Info(err)
+				if errors.Is(err, lastErr) {
+					logrus.Info(err)
+					lastErr = err
+				}
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/agent/cluster.go
```diff
@@ -134,18 +134,17 @@ func (czero *Cluster) IsBootstrapComplete() (bool, bool, error) {
...
-		if !czero.installHistory.RestAPISeen && !czero.installHistory.ClusterKubeAPISeen {
-			logrus.Debug("Agent Rest API never initialized. Bootstrap Kube API never initialized")
-			logrus.Info("Waiting for cluster install to initialize. Sleeping for 30 seconds")
+		if !czero.installHistory.ClusterKubeAPISeen {
+			if !czero.installHistory.RestAPISeen {
+				logrus.Debug("Agent Rest API never initialized. Bootstrap Kube API never initialized")
+				logrus.Info("Waiting for cluster install to initialize. Sleeping for 30 seconds")
+			} else {
+				logrus.Debug("Bootstrap Kube API never initialized. Last cluster install status from Agent Rest API: %s", czero.installHistory.RestAPIPreviousClusterStatus)
+				logrus.Info("Waiting for Bootstrap Kube API to initialize. Sleeping for 30 seconds")
```
zaneb, line 143: I assume the idea here was that if the ReST API has been seen but is not currently live, then that is an error (the ReST API should only go away once the bootstrap is up, at which point we should have at least seen it, even though it may not be live because it bounces several times during booting).

Is the issue here that we're connecting to the KubeAPI only once every 30s, and somehow managing to sample it at exactly the times when it is down?

bfournie: So the bug report is that we're getting dozens of these messages output every 2 seconds:
INFO cluster bootstrap did not complete

Prior to:
INFO Bootstrap Kube API Initialized
INFO Bootstrap configMap status is complete

The msg is generated at https://github.com/openshift/installer/blob/master/pkg/agent/cluster.go#L147 and logged at https://github.com/openshift/installer/blob/master/pkg/agent/waitfor.go#L34

So this is an attempt to throttle these log messages, similar to what we currently do with https://github.com/openshift/installer/blob/master/pkg/agent/cluster.go#L140.

An alternative way would be not output the log the error here https://github.com/openshift/installer/blob/master/pkg/agent/waitfor.go#L34 if it was the same message as last time. That may be less of a risky change.

zaneb: The log message is annoying (and should be fixed), but as I understood it the main complaint is that we seem to be permanently stuck in the part where it complains about the bootstrap not being complete, when in fact the cluster is happily deploying.

bfournie: Getting back to this.. I changed to just prevent duplicate log messages. Since we may be redoing the wait messages as part of https://issues.redhat.com/browse/OCPBUGS-5000 I wanted to just fix the case where we blast out the same message continually.

bfournie:
/close
Closing this as its now implemented as part of https://github.com/openshift/installer/pull/7060

PR #6820, opened 2023-02-01
CFE-757 : Fix issues reported by QE

bharath-b-rh:
- Public DNS records and Private Zone does not have user-defined tags where tag name contains @, +, = special characters.
- Missing validation for TechPreview feature

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -996,5 +996,14 @@ func validateFeatureSet(c *types.InstallConfig) field.ErrorList {
 		}()
 		allErrs = append(allErrs, field.NotSupported(field.NewPath("featureSet"), c.FeatureSet, sortedFeatureSets))
 	}
+
+	if c.FeatureSet != configv1.TechPreviewNoUpgrade {
+		errMsg := "the TechPreviewNoUpgrade feature set must be enabled to use this field"
+
+		if c.Azure != nil && len(c.Azure.UserTags) > 0 {
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("platform", "azure", "userTags"), errMsg))
+		}
+	}
+
 	return allErrs
 }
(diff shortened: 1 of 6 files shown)


barbacbd:
It looks like `@` is intended to work. Should we consider supporting this character?

https://github.com/MicrosoftDocs/azure-docs/issues/18460

bharath-b-rh:
> It looks like `@` is intended to work. Should we consider supporting this character?
> 
> [MicrosoftDocs/azure-docs#18460](https://github.com/MicrosoftDocs/azure-docs/issues/18460)

Yes, but except for Azure DNS resource([Azure tag limitations](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/tag-resources?tabs=json#limitations)), which does not allow special characters for tag keys.
Console form validator displays "The name must begin with a letter or number, end with a letter, number or underscore, and may contain only letters, numbers, underscores, periods, or hyphens." warning, when trying to add a tag key with special characters. Hence chose least allowed character set for all the resources created by openshift.

pkg/types/validation/installconfig.go
```diff
@@ -1021,6 +1021,12 @@ func validateFeatureSet(c *types.InstallConfig) field.ErrorList {
 				allErrs = append(allErrs, field.Forbidden(field.NewPath("compute", "platform", "vsphere", "zones"), errMsg))
 			}
 		}
+
+		if c.Azure != nil {
```
barbacbd, line 1025: nit: You _could_ combine the two if statements but this looks good.

bharath-b-rh: @barbacbd This feature is marked for TechPreview for 4.13 and @jinyunma during pre-merge tests [mentioned in the findings](https://issues.redhat.com/browse/CORS-2416?focusedCommentId=21635681&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21635681), featureSet validation is not present for this TechPreview feature. Is this check required?

barbacbd: @bharath-b-rh that is correct you should add the techPreview Feature set. You _should_  add a check in `pkg/types/validation/installconfig.go` in the `validateFeatureSet` function. If its not already, you should also change the install config comments to note that this is a TechPreview feature.

bharath-b-rh: incorporated comment.

PR #6821, opened 2023-02-01
OCPBUGS-6777: Fix file check for loading openshift manifests

rna-afk:
When generating the openshift manifests, it tries to load the dependency control plane asset that creates files in the openshift directory. It then checks to see if the openshift manifest is actually already generated before this by checking the openshift directory for files that are not generated by the machine asset.

The newly created CPMS file was missing from the check, hence the asset loader always thought the asset was generated since there are files that exist other than the machine files and exits but it actually had to be newly generated with 4 files that were crucial to installation.

Adding the CPMS files to the list of machine asset files.

### DIFF
--- a/pkg/asset/machines/master.go
+++ b/pkg/asset/machines/master.go
@@ -621,6 +621,15 @@ func (m *Master) Load(f asset.FileFetcher) (found bool, err error) {
 	}
 	m.MachineFiles = fileList
 
+	file, err = f.FetchByName(filepath.Join(directory, controlPlaneMachineSetFileName))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	m.ControlPlaneMachineSet = file
+
 	return true, nil
 }
 
@@ -644,6 +653,7 @@ func (m *Master) Machines() ([]machinev1beta1.Machine, error) {
 		&machinev1.AlibabaCloudMachineProviderConfig{},
 		&machinev1.NutanixMachineProviderConfig{},
 		&machinev1.PowerVSMachineProviderConfig{},
+		&machinev1.ControlPlaneMachineSet{},
 	)
 
 	machinev1beta1.AddToScheme(scheme)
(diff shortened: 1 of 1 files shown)


patrickdillon:
LGTM but waiting for tests

jianli-wei:
@rna-afk Please see my bug comment, i.e. no the original issue any more, but, `create manifests` and then `create cluster` got the controlplanemachineset in **Inactive** state, please tell if it's expected, thanks!

rna-afk:
>@rna-afk Please see my bug comment, i.e. no the original issue any more, but, create manifests and then create cluster got the controlplanemachineset in Inactive state, please tell if it's expected, thanks!

#6752 should have solved that. Is it still inactive?

jianli-wei:
> > @rna-afk Please see my bug comment, i.e. no the original issue any more, but, create manifests and then create cluster got the controlplanemachineset in Inactive state, please tell if it's expected, thanks!
> 
> #6752 should have solved that. Is it still inactive?

If `create cluster` directly does get it **Active**, but `create manifests` and then `create cluster` would get it **Inactive**. Any idea?

rna-afk:
> If create cluster directly does get it Active, but create manifests and then create cluster would get it Inactive. Any idea?

New commit should solve this

jianli-wei:
@rna-afk I tested a bit and told the testing results in bug comment, please take a look and advise, thanks!

jianli-wei (approved):

pkg/asset/machines/master.go
```diff
@@ -621,6 +621,15 @@ func (m *Master) Load(f asset.FileFetcher) (found bool, err error) {
 	}
 	m.MachineFiles = fileList
 
+	file, err = f.FetchByName(filepath.Join(directory, controlPlaneMachineSetFileName))
+	if err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
```
patrickdillon, line 627: should this be 
```suggestion
			return true, nil
```
It seems like one of the return statements in the conditional should be opposite the other. I think this is the one that should be true but not sure ;)

patrickdillon: oh wait, nm, I see that the difference is in the second return argument, err vs nil

PR #6822, opened 2023-02-02
OCPBUGS-6919: vsphere: fix validation of CPUS and CoresPerSocket

r4f4:
I think the original intention behind the validation was to check that, in case `CoresPerSocket` was not set by the user, it would be checked against a default value (currently set to 4). However, that's not what is happening in practice, and both the check against the user-inputed value and the default were being performed.

That means that using CPUS=2 and CoresPerSocket=2 as exemplified in the official docs [1] would result in a validation error:
```
failed to create install config: invalid "install-config.yaml" file: controlPlane.platform.vsphere.cpus:
Invalid value: 2: numCPUs specified should be a multiple of cores per socket which is by default 4]
```
even though it's a valid configuration (albeit a pretty tiny compute node).

I'm changing the code a bit to make it more clear what the defaults values are representing and fixing the checks so they are made only once with the correct values for CPUs and CoresPerSocket.

[1] https://docs.openshift.com/container-platform/4.11/installing/installing_vsphere/installing-vsphere-installer-provisioned-customizations.html#installation-installer-provisioned-vsphere-config-yaml_installing-vsphere-installer-provisioned-customizations

### DIFF
--- a/pkg/types/vsphere/validation/machinepool.go
+++ b/pkg/types/vsphere/validation/machinepool.go
@@ -25,20 +25,20 @@ func ValidateMachinePool(platform *vsphere.Platform, machinePool *types.MachineP
 	if vspherePool.MemoryMiB < 0 {
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("memoryMB"), vspherePool.MemoryMiB, "memory size must be positive"))
 	}
-	if vspherePool.NumCPUs < 0 {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("cpus"), vspherePool.NumCPUs, "number of CPUs must be positive"))
+	numCPUs := vspherePool.NumCPUs
+	if numCPUs < 0 {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("cpus"), numCPUs, "number of CPUs must be positive"))
 	}
-	if vspherePool.NumCoresPerSocket < 0 {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("coresPerSocket"), vspherePool.NumCoresPerSocket, "cores per socket must be positive"))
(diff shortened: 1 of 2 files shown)


pkg/types/vsphere/validation/machinepool.go
```diff
@@ -25,18 +32,27 @@ func ValidateMachinePool(platform *vsphere.Platform, machinePool *types.MachineP
...
-		} else if vspherePool.NumCoresPerSocket > 0 && vspherePool.NumCPUs%vspherePool.NumCoresPerSocket != 0 {
-			allErrs = append(allErrs, field.Invalid(fldPath.Child("cpus"), vspherePool.NumCPUs, "numCPUs specified should be a multiple of cores per socket"))
-		} else if vspherePool.NumCPUs%defaultCoresPerSocket != 0 {
-			allErrs = append(allErrs, field.Invalid(fldPath.Child("cpus"), vspherePool.NumCPUs, "numCPUs specified should be a multiple of cores per socket which is by default 4"))
-		}
-	} else if vspherePool.NumCoresPerSocket > defaultNumCPUs {
-		allErrs = append(allErrs, field.Invalid(fldPath.Child("coresPerSocket"), vspherePool.NumCoresPerSocket, "cores per socket must be less than number of CPUs which is by default 4"))
+		numCPUs = vspherePool.NumCPUs
+	} else {
+		numCPUs = defaultNumCPUs
```
jcpowermac, line 40: nit: since you are checking with `vspherePool.NumCPUs` you could drop this else and assign before the if.

same suggestion to the below `vspherePool.NumCoresPerSocket`

r4f4:
/hold
I forgot about the unit tests

r4f4:
/hold cancel
unit tests should be fixed

r4f4:
/skip
Only vsphere jobs are required for this change

PR #6824, opened 2023-02-02
Add Adam

Adam0Brien:
@ashcrow


patrickdillon:
/approve
@Adam0Brien it says you are not part of the OpenShift GitHub org. prow bot may require it when using these owners rights

PR #6827, opened 2023-02-03
OCPBUGS-4997: Set the configured proxy settings for agent installer

bfournie:
When the registry is only reachable via a proxy, the settings configured in install-config.yaml need to be used for podman. Save the proxy settings in systemd config.


rwsu (commented):
/lgtm

PR #6829, opened 2023-02-03
WIP: Provide a way to enable a custom DNS solution via Install Config

sadasu:
Based on Enhancement Proposal : https://github.com/openshift/enhancements/pull/1276


pkg/types/installconfig.go
```diff
@@ -83,6 +83,17 @@ const (
...
 
+// UserConfiguredDNS enables use of an external user configured DNS solution in place of the default DNS.
+// +kubebuilder:validation:Enum="";Enabled;Disabled
+type UserConfiguredDNS string
+
+const (
+	// EnableUserDNS allows use of user configured DNS that is not the default Cloud DNS.
+	EnableUserDNS UserConfiguredDNS = "Enabled"
+	// DisableUserDNS continues use of the default Cloud DNS that will be configured by OpenShift.
+	DisableUserDNS UserConfiguredDNS = "Disabled"
```
barbacbd, line 94: Do we need to know explicitly if it is "Disabled" or can we just assume that it is disabled when Enabled is not true ?

pkg/asset/cluster/tfvars.go
```diff
@@ -286,6 +286,12 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			awsMP.Set(mp.Platform.AWS)
 			masterIAMRoleName = awsMP.IAMRole
 		}
+		apiIntLBName := ""
+		apiLBName := ""
+		if installConfig.Config.AWS.UserConfiguredDNSLB.UserDNS == "Enabled" {
```
jianlinliu, line 291: @sadasu per the code here, only BYO LB is not in this feature's scope yet, right?

sadasu:
We are moving to a solution that depends on a self-hosted DNS. In that case, there would be no need for the customer to pre-create their LBs, thus making these changes obsolete.

PR #6832, opened 2023-02-07
PowerVS: Check whether Machine Pool CIDR is /24

AshwinHIBM:
PowerVS supports machine pool CIDR of /24 for deployment.

### DIFF
--- a/pkg/asset/installconfig/powervs/validation.go
+++ b/pkg/asset/installconfig/powervs/validation.go
@@ -1,185 +1,196 @@
 package powervs
 
 import (
-	"context"
-	"fmt"
+        "context"
+        "fmt"
+        "strings"
 
-	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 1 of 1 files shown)


pkg/asset/installconfig/powervs/validation.go
```diff
@@ -25,6 +25,12 @@ func Validate(config *types.InstallConfig) error {
 			fldPath := field.NewPath("compute").Index(idx)
 			allErrs = append(allErrs, validateMachinePool(fldPath, &compute)...)
 		}
+		//Check whether Machine Pool CIDR is /24 by checking whether its suffix is "/24".
+		if !(strings.HasSuffix((&config.Networking.MachineNetwork[0].CIDR).String(),"/24")){
```
mjturek, line 29: Definitely iterate over all MachineNetworks. We don't handle more than one currently but will be more robust in case we ever decide to.

AshwinHIBM:
Modified and tried to iterate over all machine pools.

PR #6834, opened 2023-02-08
OpenStack: Fix distribution of zones on 32-bit systems

pierreprinetti:
On 32-bit systems, Go `int` values are 32-bit integers. Given that we accept an explicit int64 number of replicas in the machine-pool, the index we use to distribute zones could end up negative when casting to int on 32-bit systems.

As a consequence, using availability zones when the replica number is greater than 2^31-1 on 32-bit systems makes the Installer panic.

With this change, we coerce the wrapping to be positive by casting to an unsigned integer. The automatic casting to int only happens after a modulo operation that is guaranteed to result in a number within int range.

While this fix may look risible, hitting this bug is just marginally less likely than encountering clusters of two billion workers, occurrence that the existing code goes out of its way to cover.


mandre (commented):
/approve

PR #6837, opened 2023-02-08
OpenStack Failure domains

pierreprinetti:
* Add failureDomains to install-config's machine-pools
* Generate Machines and Machinesets with Failure domains
* Create Control plane nodes in Terraform according to Failure domains

depends on https://github.com/openshift/api/pull/1399


pierreprinetti:
This is now ready for review; however, it can't be merged until https://github.com/openshift/api/pull/1399 is.
/cc mdbooth EmilienM

pierreprinetti:
Todo:
- in manifest-test, turn portSet into hasPorts and return a bool
- in tfvars/openstack.go, handle the case where no fd is passed

pkg/tfvars/openstack/openstack.go
```diff
@@ -211,6 +233,7 @@ func TFVars(
 		MachinesNetwork                   string                            `json:"openstack_machines_network_id,omitempty"`
 		MasterAvailabilityZones           []string                          `json:"openstack_master_availability_zones,omitempty"`
 		MasterRootVolumeAvailabilityZones []string                          `json:"openstack_master_root_volume_availability_zones,omitempty"`
+		MasterPortsJSON                   []string                          `json:"openstack_master_ports_json,omitempty"`
```
EmilienM, line 236: I think `omitempty` can be removed. This was suggested to me on another PR because `masterPortsJSON` is always going to be set and will never be empty.

pkg/types/openstack/validation/machinepool.go
```diff
@@ -24,5 +26,29 @@ func ValidateMachinePool(_ *openstack.Platform, machinePool *openstack.MachinePo
 	default:
 		errs = append(errs, field.NotSupported(fldPath.Child("serverGroupPolicy"), machinePool.ServerGroupPolicy, validServerGroupPolicies))
 	}
+
```
EmilienM, line 29: Innocent question but aren't we setting a guard to test if `featureSet` has `TechPreviewNoUpgrade` ? Or this is not TechPreview and directly stable?

I understand that API hasn't been changed in OpenShift API, however the user experience in the installer is changing a bit, so I wonder if the plan is to make these news params as Tech Preview or not.

I'm fine either way, I'm just making sure that this was discussed.

EmilienM (commented):
This is a solid start, I like how it's well tested!

* The Terrraform bits are quite a thing, it's a shame we can't loop over the 3 masters to avoid code duplication but I guess we have no choice.
* I have an inline question about `TechPreviewNoUpgrade` or not.

Otherwise LGTM

pkg/tfvars/openstack/openstack.go
pierreprinetti: For sure it’s irrelevant.

pkg/types/openstack/validation/machinepool.go
pierreprinetti: It’s just below! Unless I’m missing something?

EmilienM: Oops, didn't see it!

pkg/tfvars/openstack/openstack.go
pierreprinetti: Removed!

pierreprinetti:
Closing in favour of #6917

PR #6839, opened 2023-02-09
openstack-manifests: Add a debug option

pierreprinetti:
Add the `-p` (Persist) option to prevent the automated cleanup of generated manifests.


hack/openstack/test-manifests.sh
```diff
@@ -28,12 +29,13 @@ print_help() {
...
 	echo
 	echo -e "Use:"
-	echo -e "\\t${0} -c <cloud> -e <external network> -f <flavor> [-a <fip>] [-i <openshift-install>] [-t <test-dir>]"
+	echo -e "\\t${0} -c <cloud> -e <external network> -f <flavor> [-a <fip>] [-i <openshift-install>] [-t <test-dir>] [-p]"
 	echo
 	echo -e 'Additional arguments:'
 	echo -e "\\t-a\\tapiFloatingIP"
 	echo -e "\\t-i\\tpath to openshift-install (defaults to '${openshift_install}')"
 	echo -e "\\t-t\\tpath to the tests to be run (defaults to '${tests_dir}')"
+	echo -e "\\t-p\\tpersist artifacts: do not delete files generated by openshift-install"
```
mandre, line 38: Are the artifacts sanitized before being published, to remove potentially sensitive information?

mandre: But if it were?

pierreprinetti: If someone was to enable debugging, just like it they were using `-x` in their script, they would have to either remember to censor secrets, or rely on Prow's ability to mask values contained in the Vault.

What are you suggesting?

mandre: I'm not suggesting at the moment, I'm trying to understand the risks associated with this PR. I want to avoid secret leaks if someone one day decides that saving the artifacts is useful in CI to help debugging.

pierreprinetti: Without intentional changes, manifests are generated in a tmp directory that is not shared in the CI. If you enable `-p` in the CI without additional changes, nothing gets published.

mandre (commented):
/lgtm
/approve

PR #6840, opened 2023-02-09
OCPBUGS-7261: Update AgentConfig template

celebdor:
Due to it being a string, it had become out of date in a way that it impacted users. This commit syncs its content back to what is in pkg/types


bfournie:
Probably should update the apiVersion string in the test file too, e.g.:
https://github.com/openshift/installer/blob/master/pkg/asset/agent/agentconfig/agent_config_test.go#L33

celebdor:
> Probably should update the apiVersion string in the test file too, e.g.: https://github.com/openshift/installer/blob/master/pkg/asset/agent/agentconfig/agent_config_test.go#L33

Good idea. Thanks

PR #6843, opened 2023-02-09
AWS: Add wildcard apps domain DNS records for AWS

Shanedell:
AWS: Add wildcard apps domain DNS records for AWS

In the User-Provisioned Install there are instructions to create a wildcard apps (*.apps.<cluster_domain>). However, in the Infrastructure-Provisioned code there was no wildcard records being created. This caused errors in the authentication and console ClusterOperators getting deployed.

To be more clear this is the link [AWS UPI Create Ingress Endpoints](https://docs.openshift.com/container-platform/4.8/installing/installing_aws/installing-restricted-networks-aws.html#installation-create-ingress-dns-records_installing-restricted-networks-aws). In retrospect, if this is merged the section linked of the documentation could actually be removed since like the records for `api` and `api-int` this would be made by in Terraform in both instances I believe.

### DIFF
--- a/data/data/aws/cluster/route53/base.tf
+++ b/data/data/aws/cluster/route53/base.tf
@@ -67,7 +67,7 @@ resource "aws_route53_record" "api_internal_alias" {
   }
 }
 
-resource "aws_route53_record" "api_internal_alias" {
+resource "aws_route53_record" "apps_internal_alias" {
   count = local.use_alias ? 1 : 0
 
   zone_id = data.aws_route53_zone.int.zone_id
@@ -117,7 +117,7 @@ resource "aws_route53_record" "api_internal_cname" {
   records = [var.api_internal_lb_dns_name]
 }
 
-resource "aws_route53_record" "api_internal_cname" {
+resource "aws_route53_record" "apps_internal_cname" {
   count = local.use_cname ? 1 : 0
 
   zone_id = data.aws_route53_zone.int.zone_id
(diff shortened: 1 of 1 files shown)


Shanedell:
@mtulio @patrickdillon Would you guys happen to know why when I build the `openshift-install` executable locally and try to run it inside of AWS my cluster now fails on waiting for the Kubernetes API? It does this even when I remove my changes. Is there a different way I need to build for it actually work properly? I was using `./hack/build.sh`.

patrickdillon:
/hold

*.apps is created by [the cluster-ingress-operator](https://github.com/openshift/cluster-ingress-operator/)--not the installer. 

*.apps can be managed by the cluster ingress operator in UPI but we give instructions. which you linked to, in case users want to use a different DNS provider outside of the cluster's cloud platform, which is only supported in the UPI case.

patrickdillon:
> Is there a different way I need to build for it actually work properly? 

When you build the installer, you will need to set `OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE` and point to a release image that matches the installer you built. By default the installer will point to OKD.

Shanedell:
Oh okay. So my issue is coming from something not happening on the ingress controller side not the installer side? That would make sense, I appreciate the response. Also would you want this PR to be closed then?

patrickdillon:
/close

Yeah, the responsibility for creating that record is in a different component. Upon the cluster install failure, it should report which operators are degraded and that should give you an idea of what's going wrong. The cluster ingress operator depends on compute nodes being available. If there was a problem and compute nodes did not come up, that operator will fail and the records wouldn't be created.

PR #6846, opened 2023-02-13
AGENT-456: Remove connectivity checks from bash script

rwsu:
The connectivity checks will be performed by agent-tui.

Updated /etc/assisted/agent-installer.env to include AGENT_TUI_LOG_PATH=/var/log/agent/agent-tui.log.

Depends on: https://github.com/openshift/agent-installer-utils/pull/4


data/data/agent/systemd/units/agent-interactive-console.service
```diff
@@ -7,6 +7,7 @@ Before=network.target network.service agent.service NetworkManager-wait-online.s
 Type=oneshot
 TTYPath=/dev/tty15
 ExecStartPre=/usr/bin/chvt 15
+ExecStartPre=mkdir -p /var/log/agent
 ExecStart="/usr/local/bin/agent-interactive-console.sh"
```
zaneb, line 11: Could we add `EnvironmentFile=etc/assisted/agent-installer.env` and then ExecStart the agent-tui binary directly and remove the script?

rwsu: Done

data/data/agent/files/etc/assisted/agent-installer.env.template
```diff
@@ -9,3 +9,4 @@
 #   
 NODE_ZERO_IP={{.NodeZeroIP}}
 RELEASE_IMAGE={{.ReleaseImage}}
+AGENT_TUI_LOG_PATH=/var/log/agent/agent-tui.log
```
andfasano, line 12: I'm not sure if this log will be captured by the `agent-gather` scripts, in case we could follow up in a subsequent PR

PR #6850, opened 2023-02-13
PowerVS: Add capacity checks before installation

hamzy:
Make sure before creating a cluster that we have enough capacity in the system pool.

@TODO I cannot currently find documentation which describes what
PowerVS uses internally for the "Shared" option.  Therefore, we
do not check for capacity if masters or workers use it.

https://issues.redhat.com/browse/MULTIARCH-2115

### DIFF
--- a/pkg/asset/installconfig/powervs/session.go
+++ b/pkg/asset/installconfig/powervs/session.go
@@ -8,6 +8,7 @@ import (
 	gohttp "net/http"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


mjturek:
Overall looks good, but yeah we need to check with Power VS team on how we should calculate shared processors. I think the logic for the memory won't change though.

pkg/asset/installconfig/powervs/session.go
```diff
@@ -273,6 +278,153 @@ func (c *BxClient) ValidateCloudConnectionInPowerVSRegion(ctx context.Context, s
...
+	case "Dedicated":
+	case "Shared":
+		// @TODO I would think we should reduce the number of cores by some factor.
+		// However, I cannot currently find documentation which describes what
+		// PowerVS uses internally.
+	default:
+		return errors.Errorf("Unknown worker processor type (%v)", workerProcessorType)
+	}
+
+	for _, systemPool := range systemPools {
```
mjturek, line 383: I think we should not check cores at all in the Shared case for now. It's the most common case and we shouldn't overestimate how many cores we're using.

mjturek (commented):
If you can also update the commit message to indicate that "Shared" core validation isn't implemented that would be appreciated.

pkg/asset/installconfig/powervs/session.go
hamzy: Setting the number of CPUs to 0 in this case essentially disables the check,

hamzy:
@mjturek updated

PR #6853, opened 2023-02-14
GCP: conditionally create bootstrap service account

patrickdillon:
#6330, which introduced environmental auth, changed the default behavior of all installs to create a new service account for signing the ignition URL. Instead, we should only create that service account when using environmental auth (or for some other reason don't have a service account with a key).

### DIFF
--- a/data/data/gcp/bootstrap/main.tf
+++ b/data/data/gcp/bootstrap/main.tf
@@ -30,26 +30,29 @@ resource "google_storage_bucket_object" "ignition" {
 }
 
 resource "google_service_account" "bootstrap-node-sa" {
+  count        = var.gcp_create_bootstrap_sa ? 1 : 0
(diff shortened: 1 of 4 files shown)


data/data/gcp/bootstrap/main.tf
```diff
@@ -9,6 +9,7 @@ locals {
 
   public_endpoints = var.gcp_publish_strategy == "External" ? true : false
   external_ip      = local.public_endpoints ? [google_compute_address.bootstrap.address] : []
+  signed_url_creds = var.gcp_create_bootstrap_sa ? base64decode(google_service_account_key.bootstrap[0].private_key) : null
```
barbacbd, line 12: This is only used once, should we bother making a local variable?

barbacbd (commented):
/lgtm

PR #6855, opened 2023-02-15
OCPBUGS-7495: Convert platform type for AgentClusterInstall

zaneb:
The platform types in AgentClusterInstall are in mixed case. Previously we were just casting the lowercase platform name to set the PlatformType field. This resulted in the platform type being silently ignored when creating the Cluster in the assisted API. Clusters were getting the default type for their topology (None for SNO, otherwise BareMetal), but there was no way to set VSphere as the platform.

### DIFF
--- a/pkg/asset/agent/common.go
+++ b/pkg/asset/agent/common.go
@@ -1,18 +1,49 @@
 package agent
 
 import (
+	hiveext "github.com/openshift/assisted-service/api/hiveextension/v1beta1"
+	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/baremetal"
 	"github.com/openshift/installer/pkg/types/none"
 	"github.com/openshift/installer/pkg/types/vsphere"
 )
 
-// SupportedPlatforms lists the supported platforms for agent installer
-var SupportedPlatforms = []string{baremetal.Name, vsphere.Name, none.Name}
+// SupportedInstallerPlatforms lists the supported platforms for agent installer.
+func SupportedInstallerPlatforms() []string {
+	return []string{baremetal.Name, vsphere.Name, none.Name}
+}
+
+var supportedHivePlatforms = []hiveext.PlatformType{
+	hiveext.BareMetalPlatformType,
+	hiveext.VSpherePlatformType,
+	hiveext.NonePlatformType,
+}
+
+// SupportedHivePlatforms lists the supported platforms for AgentClusterInstall.
+func SupportedHivePlatforms() []string {
+	platforms := []string{}
+	for _, p := range supportedHivePlatforms {
+		platforms = append(platforms, string(p))
+	}
+	return platforms
+}
+
+func HivePlatformType(platform types.Platform) hiveext.PlatformType {
+	switch platform.Name() {
+	case baremetal.Name:
+		return hiveext.BareMetalPlatformType
+	case none.Name:
+		return hiveext.NonePlatformType
+	case vsphere.Name:
+		return hiveext.VSpherePlatformType
+	}
+	return ""
+}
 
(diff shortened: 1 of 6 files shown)


pkg/asset/agent/common.go
```diff
@@ -1,18 +1,50 @@
...
 )
 
-// SupportedPlatforms lists the supported platforms for agent installer
-var SupportedPlatforms = []string{baremetal.Name, vsphere.Name, none.Name}
+// SupportedInstallerPlatforms lists the supported platforms for agent installer
+func SupportedInstallerPlatforms() []string {
+	return []string{baremetal.Name, vsphere.Name, none.Name}
+}
+
+var supportedHivePlatforms = []hiveext.PlatformType{
```
bfournie, line 17: I see that these are also defined for infrastructure types here: https://github.com/openshift/installer/blob/master/vendor/github.com/openshift/api/config/v1/types_infrastructure.go#L177
Would it make sense to use the infrastructure defines instead of the the Hive defines (same strings)?

pkg/asset/agent/common.go
```diff
@@ -1,18 +1,50 @@
 package agent
 
 import (
+	"github.com/openshift/installer/pkg/types"
```
bfournie, line 4: Need to run gofmt on this file.

pkg/asset/agent/common.go
```diff
@@ -1,18 +1,50 @@
...
+}
+
+var supportedHivePlatforms = []hiveext.PlatformType{
+	hiveext.BareMetalPlatformType,
+	hiveext.VSpherePlatformType,
+	hiveext.NonePlatformType,
+}
+
+// SupportedHivePlatforms lists the supported platforms for AgentClusterInstall
+func SupportedHivePlatforms() []string {
```
bfournie, line 24: Lint had some complaints
pkg/asset/agent/common.go:12:81: Comment should end in a period (godot)
// SupportedInstallerPlatforms lists the supported platforms for agent installer
                                                                                ^
pkg/asset/agent/common.go:23:80: Comment should end in a period (godot)
// SupportedHivePlatforms lists the supported platforms for AgentClusterInstall
                                                                               ^
pkg/asset/agent/common.go:25:2: Consider preallocating `platforms` (prealloc)
	var platforms []string

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -239,6 +242,18 @@ func (a *AgentClusterInstall) Load(f asset.FileFetcher) (bool, error) {
...
+	// even though that is incorrect. Rewrite to the correct mixed case names
+	// for backward compatibility.
+	switch string(agentClusterInstall.Spec.PlatformType) {
+	case baremetal.Name:
+		agentClusterInstall.Spec.PlatformType = hiveext.BareMetalPlatformType
+	case none.Name:
+		agentClusterInstall.Spec.PlatformType = hiveext.NonePlatformType
+	case vsphere.Name:
+		agentClusterInstall.Spec.PlatformType = hiveext.VSpherePlatformType
+	}
```
bfournie, line 255: agentclusterinstall_test.go also needs updating with the new strings, e.g.
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -69,3 +69,3 @@
            	            	   Proxy: (*v1beta1.Proxy)(<nil>),
            	            	-  PlatformType: (v1beta1.PlatformType) (len=9) "baremetal"
            	            	+  PlatformType: (v1beta1.PlatformType) (len=9) "BareMetal"

pkg/asset/agent/common.go
zaneb: No, because they're different types and casting is how we got into this :smile:

PR #6856, opened 2023-02-15
Generate CSI driver manifests

jhixson74:
https://issues.redhat.com/browse/CORS-2090
https://issues.redhat.com/browse/CORS-2321
https://issues.redhat.com/browse/CORS-2320


pkg/asset/manifests/clustercsidriver.go
```diff
@@ -0,0 +1,127 @@
...
+		if platform == nil || platform.OSDisk.DiskEncryptionSet == nil {
+			return nil
+		}
+		session, err := installConfig.Azure.Session()
+		if err != nil {
+			return errors.Wrap(err, "could not get azure session")
+		}
+		configData, err := azure.ClusterCSIDriverConfig{
+			SubscriptionID:        session.Credentials.SubscriptionID,
+			ResourceGroupName:     installConfig.Config.Azure.ClusterResourceGroupName(clusterID.InfraID),
```
patrickdillon, line 82:
```suggestion
			ResourceGroupName:     platform.OSDisk.DiskEncryptionSet.ResourceGroup,
```

patrickdillon (commented):
This looks good. There's just a bug with the Azure Resource Group and then a couple of nits. 

Can you squash the go fmt commit and add some detail in the other messages? 

I am totally fine with this structure, but it seems like we could have all of the logic in `clustercsidriver.go` rather than using platform-specific packages. IMHO it is not worth refactoring, but thought I would mention it.

Hopefully API PR will merge soon.

pkg/asset/manifests/clustercsidriver.go
```diff
@@ -0,0 +1,89 @@
...
+// Dependencies returns all of the dependencies directly needed to generate the
+// asset.
+func (*ClusterCSIDriverConfig) Dependencies() []asset.Asset {
+	return []asset.Asset{
+		&installconfig.InstallConfig{},
+		&installconfig.ClusterID{},
+		// PlatformCredsCheck just checks the creds (and asks, if needed)
+		// We do not actually use it in this asset directly, hence
+		// it is put in the dependencies but not fetched in Generate
+		&installconfig.PlatformCredsCheck{},
```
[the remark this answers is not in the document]
jhixson74: Removed.

pkg/asset/manifests/clustercsidriver.go
jhixson74: Fixed.

gpei:
Hi @jhixson74 I noticed PR https://github.com/openshift/api/pull/1401 was merged and we removed the `[WIP]` on this PR, so tried this feature on AWS first, but seems something wrong with the generated `manifests/cluster-csi-driver-config.yaml` file.

For `managementState` field it's `null` currently, but which seems to be invalid. 
```
The ClusterCSIDriver "ebs.csi.aws.com" is invalid: spec.managementState: Invalid value: "": spec.managementState in body should match '^(Managed|Unmanaged|Force|Removed)$'
```
I added the detailed test result in https://issues.redhat.com/browse/CORS-2563, could you help to take a look and point out if anything I was missing? Thanks.

dobsonj:
The spec for the ClusterCSIDriver should have these fields by default:
```
spec:
  logLevel: Normal
  managementState: Managed
  operatorLogLevel: Normal
```
https://github.com/openshift/cluster-storage-operator/blob/master/assets/csidriveroperators/aws-ebs/standalone/10_cr.yaml

dobsonj:
Ah, `logLevel` and `operatorLogLevel` are optional and default to Normal [in the operator API](https://github.com/openshift/api/blob/d170fcdc0fa638b664e4f35f2daf753cb4afe36b/operator/v1/types.go#L55-L71) so I guess it's just `managementState` that needs to be set explicitly.

jhixson74:
> FWIW, if `managementState` has any value other than `Managed` then the SC will not get reconciled [here](https://github.com/openshift/library-go/blob/38724ca8e236fbd938f9c0b737b35c9d6f5091f6/pkg/operator/csi/csistorageclasscontroller/csi_storageclass_controller.go#L112).

I've updated the code to have the mangementState

patrickdillon:
gofmt is failing. Besides that this LGTM. Will tag after the tests run

PR #6858, opened 2023-02-15
OSASINFRA-3105: Convert username/password to app credentials

stephenfin:
[Application credentials](https://docs.openstack.org/keystone/latest/user/application_credentials.html) have a number of advantage over username/password combinations, not least the ease of rotation. However, the `clouds.yaml` files generated by various OpenStack tools like Horizon default to username/password combinations and users tend to pass these through unmodified. We don't want to outright reject these, but we can indicate an explicit preference for application credentials. We do this by modifying the `clouds.yaml` file we store as a secret before storing it. This requires generating a new application secret, which all users should be able to do by default.

### DIFF
--- a/pkg/asset/manifests/openshift.go
+++ b/pkg/asset/manifests/openshift.go
@@ -162,7 +162,7 @@ func (o *Openshift) Generate(dependencies asset.Parents) error {
 		}
 
 		// We also replace a username/password combination with an application credential
-		if cloud.AuthInfo != nil && (cloud.AuthInfo.UserID != "" || cloud.AuthInfo.Username != "") && cloud.AuthInfo.Password != "" {
+		if cloud.AuthInfo != nil && ( cloud.AuthType == clientconfig.AuthPassword || cloud.AuthType == clientconfig.AuthV2Password || cloud.AuthType == clientconfig.AuthV3Password) {
 			conn, err := clientconfig.NewServiceClient("identity", opts)
 			if err != nil {
 				return err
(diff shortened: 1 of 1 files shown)


pkg/asset/manifests/openshift.go
```diff
@@ -159,6 +161,46 @@ func (o *Openshift) Generate(dependencies asset.Parents) error {
...
+			// Gimme gimme gimme 🎶
+			createOpts := applicationcredentials.CreateOpts{
+				// TODO: Generate/retrieve a unique, per-install name
+				Name: "openshift-installer",
+			}
+			applicationCredential, err := applicationcredentials.Create(conn, userID, createOpts).Extract()
+
+			cloud.AuthInfo.Username = ""
+			cloud.AuthInfo.UserID = ""
+			cloud.AuthInfo.Password = ""
```
pierreprinetti, line 199: I believe you also have to remove project (or tenant) name, id and domain name, user domain name and set `auth_type` to `v3applicationcredential`

pierreprinetti:
I don't think this is a good idea.
With this change, we would be silently changing the credentials the user gives us before using them.

Consider this scenario:
1. I install OpenShift
2. Something goes really wrong with my cluster
3. In an effort to regain control, I rotate my OpenStack password. Production payloads are not expected to be affected, but at least OpenShift can't operate hardware any more
4. except OpenShift still can operate hardware because it's not using the password I installed it with.

I understand that it's an unlikely scenario. However, as light as my argument weights, I can't see any counterweight on the pros weighting plate.

On a high level, I think that credentials handling should be extremely explicit to the user and it's the last context where I want this kind of magic to happen.

pkg/asset/manifests/openshift.go
stephenfin: Done. I just recreate it now.

stephenfin:
Let's just emit a warning instead.

PR #6859, opened 2023-02-15
linter: bump version to sync with CI

r4f4:
CI now has golangci-lint 1.48 which supports Golang 1.19. Let's make sure our hack/ tool uses the same version.


r4f4:
/override ci/prow/e2e-aws-ovn ci/prow/okd-scos-images
This change doesn't affect any e2e jobs.

PR #6860, opened 2023-02-15
vSphere: ignore all bootstrap disk changes

jcpowermac:
In some vSphere infrastructures the disk type
could change underneath terraform state causing
bootstrap delete to fail.

Ignore all bootstrap disk changes.


jcpowermac:
odd failure with gather extra...

/retest-required

Issue #6862, opened 2023-02-16
support GOOGLE_APPLICATION_CREDENTIALS env var 


tosmi:
openshift-install should also consider the env var  GOOGLE_APPLICTION_CREDENTIALS. this seems to be the current standard according to: https://cloud.google.com/docs/authentication/provide-credentials-adc#local-key

# Version

```console
$ openshift-install version
./openshift-install 4.12.2
built from commit 7fea1c4fc00312fdf91df361b4ec1a1a12288a97
release image quay.io/openshift-release-dev/ocp-release@sha256:31c7741fc7bb73ff752ba43f5acf014b8fadd69196fc522241302de918066cb1
release architecture amd64
```

# Platform:
GCP IPI installation

# What happened?

we only had GOOGLE_APPLICATION_CREDENTIALS set. took us a while to figure out which env vars are supported by the installer.

PR #6863, opened 2023-02-16
OCPBUGS-17866: check GOOGLE_APPLICATION_CREDENTIALS env var

tosmi:
according to
https://cloud.google.com/docs/authentication/provide-credentials-adc#local-key the default for application credentials is to set
GOOGLE_APPLICATION_CREDENTIALS. currently this var is missing from the list of environment variables checked.

this should fix issue #6862

### DIFF
--- a/pkg/asset/installconfig/gcp/session.go
+++ b/pkg/asset/installconfig/gcp/session.go
@@ -16,7 +16,7 @@ import (
 )
 
 var (
-	authEnvs            = []string{"GOOGLE_CREDENTIALS", "GOOGLE_CLOUD_KEYFILE_JSON", "GCLOUD_KEYFILE_JSON"}
+	authEnvs            = []string{"GOOGLE_APPLICATION_CREDENTIALS", "GOOGLE_CREDENTIALS", "GOOGLE_CLOUD_KEYFILE_JSON", "GCLOUD_KEYFILE_JSON"}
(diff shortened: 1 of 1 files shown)


patrickdillon:
/ok-to-test

Thanks for sending this in.
/cc @barbacbd

pkg/asset/installconfig/gcp/session.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 var (
-	authEnvs            = []string{"GOOGLE_CREDENTIALS", "GOOGLE_CLOUD_KEYFILE_JSON", "GCLOUD_KEYFILE_JSON"}
+	authEnvs            = []string{"GOOGLE_APPLICATION_CREDENTIALS", "GOOGLE_CREDENTIALS", "GOOGLE_CLOUD_KEYFILE_JSON", "GCLOUD_KEYFILE_JSON"}
```
barbacbd, line 19: nit: This should be added to the comment in GetSession as well so users know the order that environment variables are selected.

barbacbd (changes requested):

barbacbd (commented):
/lgtm

barbacbd:
/retitle OCPBUGS-17866: check GOOGLE_APPLICATION_CREDENTIALS env var

PR #6868, opened 2023-02-16
OCPBUGS-7063: vSphere - Remove regexs in terraform ova import

jcpowermac:
In the progression of execution when
the terraform private resource has been
reached we should have already validated the
paths completely.

This PR removes the regular expressions
that is causing the terraform import failure.


r4f4 (commented):
/approve

PR #6869, opened 2023-02-16
OpenStack: Add dual-stack support with stateful IPv6

MaysaMacedo:
This PR enables Dual-Stack support for OpenStack cluster using Stateful IPv6 Subnets.

- [ ] Add required IPv6 security groups rules


MaysaMacedo:
We decided to go with BYON only. If things change this PR will be re-opened

PR #6871, opened 2023-02-17
OCPBUGS-7579: azure: fix certificate-based auth with passpharse

r4f4:
Fallout from https://github.com/openshift/installer/pull/6003

If we pass `nil` to `azidentity.ParseCertificates`, it assumes the private key isn't encrypted. Let's pipe the password through instead.


r4f4:
/retitle OCPBUGS-7579: azure: fix certificate-based auth with passpharse

barbacbd (commented):
/lgtm

PR #6873, opened 2023-02-17
OCPBUGS-7529: bump RHCOS 4.12 bootimage metadata

mike-nguyen:
These changes will update the RHCOS 4.12 boot image metadata in the installer which includes the fixes for the following:

OCPBUGS-7525 - [4.12] Add new AWS regions for ROSA

Changes generated with:
```
plume cosa2stream --target data/data/coreos/rhcos.json --distro rhcos --no-signatures --url https://rhcos.mirror.openshift.com/art/storage/prod/streams x86_64=412.86.202301311551-0 aarch64=412.86.202301311551-0 s390x=412.86.202301311551-0 ppc64le=412.86.202301311551-0
```

### DIFF
--- a/data/data/coreos/rhcos.json
+++ b/data/data/coreos/rhcos.json
@@ -1,8 +1,8 @@
 {
   "stream": "rhcos-4.12",
   "metadata": {
-    "last-modified": "2023-02-02T19:16:07Z",
-    "generator": "plume cosa2stream 465b8e2"
+    "last-modified": "2023-02-17T19:03:43Z",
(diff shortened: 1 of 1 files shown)


mike-nguyen:
@gpei can I get a `/label cherry-pick-approved` on this?  We are only adding additional AWS regions so it should not affect anything.

mike-nguyen:
@sdodson Can you override the valid bug label? we agreed that adding new regions can be done in parallel a while back. There's no change here except for adding additional regions.

PR #6874, opened 2023-02-17
gcp: let load balancer access node proxy healthz on port 10256

dcbw:
kube-proxy and ovn-kubernetes serve healthz on node port 10256 to allow the GCE load balancer to check node health for services with "Cluster" traffic-policy. GCE cloud provider

staging/src/k8s.io/legacy-cloud-providers/gce/gce_healthchecks.go:

```
const (
	nodesHealthCheckPath = "/healthz"
	// NOTE: Please keep the following port in sync with ProxyHealthzPort in pkg/cluster/ports/ports.go
	// ports.ProxyHealthzPort was not used here to avoid dependencies to k8s.io/kubernetes in the
	// GCE cloud provider which is required as part of the out-of-tree cloud provider efforts.
	// TODO: use a shared constant once ports in pkg/cluster/ports are in a common external repo.
	lbNodesHealthCheckPort = 10256
)

// GetNodesHealthCheckPort returns the health check port used by the GCE load
// balancers (l4) for performing health checks on nodes.
func GetNodesHealthCheckPort() int32 {
	return lbNodesHealthCheckPort
}
```

And it calls GetNodesHealthCheckPort() in a bunch of places to health-check services:

```
	// Ensure health check exists before creating the backend service. The health check is shared
	// if externalTrafficPolicy=Cluster.
	sharedHealthCheck := !servicehelpers.RequestsOnlyLocalTraffic(svc)
	hcName := makeHealthCheckName(loadBalancerName, clusterID, sharedHealthCheck)
	hcPath, hcPort := GetNodesHealthCheckPath(), GetNodesHealthCheckPort()
	if !sharedHealthCheck {
		// Service requires a special health check, retrieve the OnlyLocal port & path
		hcPath, hcPort = servicehelpers.GetServiceHealthCheckPathPort(svc)
	}
	hc, err := g.ensureInternalHealthCheck(hcName, nm, sharedHealthCheck, hcPath, hcPort)
	if err != nil {
		return nil, err
	}
```

See also https://issues.redhat.com/browse/OCPBUGS-7158

### DIFF
--- a/data/data/gcp/cluster/network/firewall.tf
+++ b/data/data/gcp/cluster/network/firewall.tf
@@ -34,6 +34,28 @@ resource "google_compute_firewall" "health_checks" {
   target_tags = ["${var.cluster_id}-master"]
 }
 
+resource "google_compute_firewall" "node_health_checks" {
+  count       = var.create_firewall_rules ? 1 : 0
+  name        = "${var.cluster_id}-node-health-checks"
+  network     = local.cluster_network
+  description = local.description
+  project     = var.network_project_id
+
+  # proxy healthz
+  allow {
+    protocol = "tcp"
+    ports    = ["10256"]
+  }
+
+  # Add the public load balancer ips when the cluster is public/external
+  source_ranges = concat(["35.191.0.0/16", "130.211.0.0/22"], var.public_endpoints ? ["209.85.152.0/22", "209.85.204.0/22"] : [])
+
+  target_tags = [
+    "${var.cluster_id}-master",
+    "${var.cluster_id}-worker"
+  ]
+}
+
 resource "google_compute_firewall" "etcd" {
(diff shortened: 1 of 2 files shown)


cblecker:
/hold
I don't think this is needed:
https://issues.redhat.com/browse/OCPBUGS-7158?focusedCommentId=21790400&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21790400
https://github.com/openshift/installer/blob/bb7e59b91738c73113f97247cef56fc0a44e1ebc/pkg/destroy/gcp/cloudcontroller.go#L211-L217

dcbw:
@cblecker fair enough, thanks for the pointer.

As to why this change makes clusters fail to install nodes though? That's really odd.

PR #6881, opened 2023-02-20
OPNET-199: Allow v6-primary dual stack on baremetal

cybertron:
Previously we only supported ipv4-primary dual stack clusters due to some limitations in our infrastructure. We are now making the necessary changes to run ipv6-primary dual stack clusters so we no longer need to enforce the ordering on baremetal. Note also that although we are adding support for dual stack to other platforms too, at this time they will not support v6-primary dual stack so I'm limiting this change to baremetal.

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -231,11 +231,14 @@ func validateNetworkingIPVersion(n *types.Networking, p *types.Platform) field.E
 			allErrs = append(allErrs, field.Invalid(field.NewPath("networking", "serviceNetwork"), strings.Join(ipnetworksToStrings(n.ServiceNetwork), ", "), "when installing dual-stack IPv4/IPv6 you must provide two service networks, one for each IP address type"))
 		}
 
+		allowV6Primary := false
 		experimentalDualStackEnabled, _ := strconv.ParseBool(os.Getenv("OPENSHIFT_INSTALL_EXPERIMENTAL_DUAL_STACK"))
 		switch {
 		case p.Azure != nil && experimentalDualStackEnabled:
 			logrus.Warnf("Using experimental Azure dual-stack support")
 		case p.BareMetal != nil:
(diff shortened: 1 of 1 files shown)


cybertron:
/hold

This should probably go in last once we have the other patches for v6-primary dual stack merged. Until then it doesn't make sense to allow it.

barbacbd:
> /hold
> 
> This should probably go in last once we have the other patches for v6-primary dual stack merged. Until then it doesn't make sense to allow it.

Any details on when this will be?

cybertron:
We're still hoping to get this in for 4.13. Patches are proposed to all of the necessary repos (the parent epic for OPNET-199 has stories for all of them), but we still need to get agreement from the various teams involved.

cybertron:
/hold cancel

There's one patch in MCO that still needs to land, but I think this is ready for review now.

pkg/types/validation/installconfig.go
```diff
@@ -231,11 +231,14 @@ func validateNetworkingIPVersion(n *types.Networking, p *types.Platform) field.E
 			allErrs = append(allErrs, field.Invalid(field.NewPath("networking", "serviceNetwork"), strings.Join(ipnetworksToStrings(n.ServiceNetwork), ", "), "when installing dual-stack IPv4/IPv6 you must provide two service networks, one for each IP address type"))
 		}
 
+		allowV6Primary := false
```
mkowalski, line 234: I would consider renaming variable to something like `allowDualStackV6Primary` to remove confusion with single-stack IPv6 configurations. We know today it's about dual-stack, but who knows what we are going to think about this variable in the future. Anyway your final call, for me lgtm anyway

cybertron: Eventually I'm hoping this variable can go away completely. Once we have upstream support for dual stack we should be able to allow v6-primary on any platform that supports dual stack.

PR #6887, opened 2023-02-21
OCPBUGS-7622: Ignore CPMS file if not exist

rna-afk:
For the machine asset to be generated, the CPMS asset is not required and for platforms outside of AWS, Azure and GCP, it is not necessary. This also affects UPI installation.

### DIFF
--- a/pkg/asset/machines/master.go
+++ b/pkg/asset/machines/master.go
@@ -624,9 +624,11 @@ func (m *Master) Load(f asset.FileFetcher) (found bool, err error) {
 	file, err = f.FetchByName(filepath.Join(directory, controlPlaneMachineSetFileName))
 	if err != nil {
 		if os.IsNotExist(err) {
-			return false, nil
+			// Choosing to ignore the CPMS file if it does not exist since UPI does not need it.
+			logrus.Debugf("CPMS file missing. Ignoring it while loading machine asset.")
+			return true, nil
 		}
-		return false, err
+		return true, err
 	}
 	m.ControlPlaneMachineSet = file
 


r4f4:
Is this something we could cover with a unit test?

pkg/asset/machines/master.go
```diff
@@ -624,9 +624,9 @@ func (m *Master) Load(f asset.FileFetcher) (found bool, err error) {
 	file, err = f.FetchByName(filepath.Join(directory, controlPlaneMachineSetFileName))
 	if err != nil {
 		if os.IsNotExist(err) {
-			return false, nil
+			return true, nil
```
sadasu, line 627: From https://github.com/openshift/installer/blob/master/pkg/asset/machines/master.go#L599-L600, returning true here when the file is clearly not found is probably incorrect.

The function calling Load() should be modified to not throw out an error (identified in the bug) when `found` is false but `err` is nil. That approach seems consistent with the description provided in this PR that, CPMS files is not mandatory.

rna-afk: UPI needs CPMS files to be deleted to work. So it's more like mandatory but not mandatory.

rna-afk:
> Is this something we could cover with a unit test?

I'm not sure if we have a filechecker unit test and if we need to set it up, we need to make it more comprehensive.

r4f4:
> > Is this something we could cover with a unit test?
> 
> I'm not sure if we have a filechecker unit test and if we need to set it up, we need to make it more comprehensive.

Right. Even if it's a follow-up/Spike, I think it's a good idea we look into it since we could easily break jobs when touching this part of the code in the future without understanding all the ramifications.

rna-afk:
> Right. Even if it's a follow-up/Spike, I think it's a good idea we look into it since we could easily break jobs when touching this part of the code in the future without understanding all the ramifications.

I have something in mind to fix CI. Our tests are not reliable and we need to at least make them faster if not succeed more. I agree we can create unit tests for this but it would take more time to cover everything than CI.

pkg/asset/machines/master.go
r4f4: Is there a way we can identify if the manifest is required or not and skip this block instead of relying on a combination of return values?

rna-afk: All these run agnostic of cloud platform or installation type so adding these checks will be real hard. Ignoring CPMS if it doesn't exist for now seems like an easier choice.

r4f4: So at least a comment explaining why we're returning `true` even though the file doesn't exist or there was an error loading it? Maybe also a debug log message for cases when we expect those files to be loaded?

rna-afk: Sounds good. Added comment and log.

r4f4:
/approve
Let's see if we can get UPI CI jobs back to normal. Any other comments can be addressed as a follow-up.

PR #6892, opened 2023-02-22
OCPBUGS-7699: bump terraform for go-getter CVE fix

r4f4:
This is a partial fix to the CVE bug report since just trying to bump go-getter results in the following error:

```
$: go mod tidy
go: finding module for package google.golang.org/grpc/naming
terraform imports
	github.com/hashicorp/terraform imports
	github.com/hashicorp/terraform/internal/backend/init imports
	github.com/hashicorp/terraform/internal/backend/remote-state/etcdv2 imports
	go.etcd.io/etcd/client tested by
	go.etcd.io/etcd/client.test imports
	go.etcd.io/etcd/integration imports
	go.etcd.io/etcd/proxy/grpcproxy imports
	google.golang.org/grpc/naming: module google.golang.org/grpc@latest found (v1.53.0), but does not contain package google.golang.org/grpc/naming
```

because the `naming` package has been deprecated/removed from grpc and replaced by `resolver`. Only by updating terraform we can get rid of that broken dependency.


andfasano:
Just noticed that https://github.com/openshift/installer/pull/6977 didn't land yet, and that's causing the failure.

PR #6895, opened 2023-02-22
WIP CORS-2501: Parse Load Balancer Data from TF

barbacbd:
AWS: parse the load balancer IP addresses from terraform output.

### DIFF
--- a/pkg/asset/cluster/cluster.go
+++ b/pkg/asset/cluster/cluster.go
@@ -3,14 +3,7 @@ package cluster
 import (
 	"context"
 	"fmt"
-	"os"
-	"path/filepath"
-	"strings"
-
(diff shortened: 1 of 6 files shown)


pkg/asset/cluster/cluster.go
```diff
@@ -101,7 +95,7 @@ func (c *Cluster) Generate(parents asset.Parents) (err error) {
 
 	stages := platformstages.StagesForPlatform(platform)
 
-	terraformDir := filepath.Join(InstallDir, "terraform")
+	terraformDir := filepath.Join(types.InstallDir, "terraform")
```
jhixson74, line 98: Same as above.

jhixson74 (changes requested):

PR #6898, opened 2023-02-23
AGENT-502: Enable agent tui

andfasano:
This patch finalizes the integration of the agent-tui binary within the agent ISO

This patch also bumps up the default release image to `registry.ci.openshift.org/origin/release:4.13`, as it's required by the integration tests (as pointed out by @zane in the comment, there's also a ticket for it https://issues.redhat.com/browse/CORS-2235)

### DIFF
--- a/data/data/agent/systemd/units/agent-interactive-console.service
+++ b/data/data/agent/systemd/units/agent-interactive-console.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Get interactive user configuration at boot
 After=network-pre.target NetworkManager.service pre-network-manager-config.service selinux.service
-Before=network.target network.service agent.service NetworkManager-wait-online.service
+Before=network.target network.service agent.service node-zero.service NetworkManager-wait-online.service
 
 [Service]
 Type=oneshot
--- a/pkg/asset/agent/image/agentimage.go
+++ b/pkg/asset/agent/image/agentimage.go
@@ -53,14 +53,12 @@ func (a *AgentImage) Generate(dependencies asset.Parents) error {
 		return err
 	}
 
-	additionalFiles := []string{}
-	agentTuiFile, err := a.fetchAgentTuiBinary(agentManifests.ClusterImageSet.Spec.ReleaseImage, agentManifests.GetPullSecretData(), registriesConf.MirrorConfig)
+	agentTuiFiles, err := a.fetchAgentTuiFiles(agentManifests.ClusterImageSet.Spec.ReleaseImage, agentManifests.GetPullSecretData(), registriesConf.MirrorConfig)
 	if err != nil {
 		return err
 	}
-	additionalFiles = append(additionalFiles, agentTuiFile)
 
-	err = a.prepareAgentISO(baseImage.File.Filename, ignitionByte, additionalFiles)
+	err = a.prepareAgentISO(baseImage.File.Filename, ignitionByte, agentTuiFiles)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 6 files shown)


zaneb:
openshift/agent-installer-utils#12 has merged
/retest

pkg/asset/agent/image/agentimage.go
```diff
@@ -181,7 +182,10 @@ func (a *AgentImage) appendAgentFilesToInitrd(additionalFiles []string) error {
 	// Add a dracut hook to copy the files. The $NEWROOT environment variable is exported by
 	// dracut during the startup and it refers the mountpoint for the root filesystem.
 	dracutHookScript := `#!/bin/sh
-cp -R /agent-files/* $NEWROOT/usr/local/bin/`
+cp -R /agent-files/* $NEWROOT/usr/local/bin/
+# Fix the selinux label
+for i in $(find /agent-files/ -printf "%P\n"); do chcon system_u:object_r:bin_t:s0 $NEWROOT/usr/local/bin/$i; done`
```
zaneb, line 187: If `restorecon -F` worked and avoided having to explicitly specify the type, that seems preferable.

pkg/asset/releaseimage/default.go
```diff
@@ -20,7 +20,7 @@ import (
 
 var (
 	// defaultReleaseImageOriginal is the value served when defaultReleaseImagePadded is unmodified.
-	defaultReleaseImageOriginal = "registry.ci.openshift.org/origin/release:4.12"
+	defaultReleaseImageOriginal = "registry.ci.openshift.org/origin/release:4.13"
```
zaneb, line 23: There is a ticket for this: https://issues.redhat.com/browse/CORS-2235
We can at least reference it in the patch headline.

zaneb (commented):
/approve

data/data/agent/systemd/units/agent-interactive-console.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=Get interactive user configuration at boot
 After=network-pre.target NetworkManager.service pre-network-manager-config.service selinux.service
-Before=network.target network.service agent.service NetworkManager-wait-online.service
+Before=network.target network.service agent.service node-zero.service NetworkManager-wait-online.service
```
zaneb, line 4: I'm surprised `getty@tty1.service` is not on this list

pkg/asset/agent/image/agentimage.go
andfasano: Not sure if I've missed something, but in my tests `restorecon` wasn't available

pkg/asset/agent/image/agentimage.go
```diff
@@ -181,7 +182,10 @@ func (a *AgentImage) appendAgentFilesToInitrd(additionalFiles []string) error {
 	// Add a dracut hook to copy the files. The $NEWROOT environment variable is exported by
 	// dracut during the startup and it refers the mountpoint for the root filesystem.
 	dracutHookScript := `#!/bin/sh
-cp -R /agent-files/* $NEWROOT/usr/local/bin/`
+cp -R /agent-files/* $NEWROOT/usr/local/bin/
```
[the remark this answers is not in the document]
andfasano: In general I agree, in the sense that this is just a first rough mechanism to copy over stuff from the initramfs to the disk to have an initial integration, and it could be definitely improved as our requirements increase. We could follouwup in a subsequent pr

data/data/agent/systemd/units/agent-interactive-console.service
zaneb: Actually, we should use `getty-pre.target` here.
https://www.freedesktop.org/software/systemd/man/systemd.special.html#getty-pre.target

pkg/asset/agent/image/agentimage.go
zaneb: It's in CoreOS, but maybe not in the initramfs :slightly_frowning_face:

data/data/agent/systemd/units/agent-interactive-console.service
zaneb: Is `network.service` a real thing?

I thought for a minute that we should use `Before=getty-pre.target` (https://www.freedesktop.org/software/systemd/man/systemd.special.html#getty-pre.target), but that will block _all_ consoles and we are only displaying on one console, so I think `Before=getty@tty1.service` is the right thing.

zaneb:
The HA and SNO jobs are failing with sshd not coming up. They are both disconnected jobs, so the likely cause is https://issues.redhat.com/browse/OCPBUGS-8107. This should be resolved by https://issues.redhat.com/browse/AGENT-565 once it is revendored into the a-i-a repo.

PR #6899, opened 2023-02-23
[wip] vSphere: add cns volume delete to destroy

jcpowermac:


jcpowermac:
@gnufied is this what you were thinking?

gnufied:
@jcpowermac yes that is what I was thinking. It would allow us to reuse stuff from CI and then developers can use this for use-throw environments..

cmd/openshift-install/destroy.go
```diff
@@ -58,11 +62,13 @@ func newDestroyClusterCmd() *cobra.Command {
 			logrus.Infof("Uninstallation complete!")
 		},
 	}
+	destroyCmd.Flags().BoolVar(&deleteVolumes, "delete-volumes", false, "Delete CNS volumes on destroy.")
```
jcpowermac, line 65: @patrickdillon installer doesn't have flags usually, any recommendation for alternative here?

PR #6900, opened 2023-02-24
MULTIARCH-3212: Update resource group in PowerVSPlatformStatus

dharaneeshvrd:
Supporting IBM COS as storage for PowerVS platform in image registry operator requires ResourceGroup to be present in PowerVSPlatformStatus.

### DIFF
--- a/pkg/asset/manifests/infrastructure.go
+++ b/pkg/asset/manifests/infrastructure.go
@@ -259,6 +259,7 @@ func (i *Infrastructure) Generate(dependencies asset.Parents) error {
 			return errors.New("unknown publishing strategy")
 		}
 		config.Status.PlatformStatus.PowerVS = &configv1.PowerVSPlatformStatus{
+			ResourceGroup:  installConfig.Config.Platform.PowerVS.PowerVSResourceGroup,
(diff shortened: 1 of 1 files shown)


mkumatag (approved):
/lgtm

clnperez:
@mjturek can you make sure we require that the resource group be set in the install config? i thought i remembered us using `Default` if it wasn't. So then it would be empty and the registry operator would blow up.

dharaneeshvrd:
@clnperez @mjturek 
I have put a hold on this PR for the same reason. As it's not prompted from user and user needs to fill the install-config.yaml manually. 
I think we might need to ask resource group from user like region and zone? Please share your opinion on this, so that we can avoid it from being set to empty.

mjturek:
Sorry I missed this, the MCO and registry operator will fail without resourcegroup set. @hamzy has a PR up to ask users for reosurcegroup https://github.com/openshift/installer/pull/6885/commits/31869cde417ac176fa7103868b422585f8fe23a6 but I don't think we can guarantee that 'Default' is an exists. New accounts can have no resourcegroups as far as I know.

PR #6901, opened 2023-02-24
OCPBUGS-7860: azure: session: fix unclear auth error messages

r4f4:
If you tried to authenticate with a certificate and provided the wrong password, the installer error message didn't actually say what the error was. It just said

```
$ ./openshift-install create install-config --dir ipi-test
INFO Could not get an azure authorizer from file: auth file missing client and certificate credentials
INFO Asking user to provide authentication info
? azure subscription id [? for help]
```

That happens because the autorest lib would just ignore the auth errors in favor of a generic message [1]. Since the `autorest/auth` module is going out of support by March 31, 2023, let's get rid of it and load the auth file ourselves.

Now when doing cert authentication with the wrong password we get:

```
INFO Credentials loaded from file "/root/.azure/osServicePrincipal.json"
WARNING Using client certs to authenticate. Please be warned cluster does not support certs and only the installer does.
ERROR failed to fetch Kubeconfig Admin Client: failed to load asset "Install Config": failed to create install config: creating Azure session: failed to parse client certificate: pkcs12: decryption password incorrect
```

[1] https://github.com/Azure/go-autorest/blob/main/autorest/azure/auth/auth.go#L264-L270


r4f4:
/hold
Ops, forgot to commit the vendor changes. Will do so once the Azure jobs finish.

pkg/asset/installconfig/azure/session.go
```diff
@@ -43,8 +43,8 @@ type Credentials struct {
 	ClientID                  string `json:"clientId,omitempty"`
 	ClientSecret              string `json:"clientSecret,omitempty"`
 	TenantID                  string `json:"tenantId,omitempty"`
-	ClientCertificatePath     string `json:"certificatePath,omitempty"`
-	ClientCertificatePassword string `json:"certificatePassword,omitempty"`
+	ClientCertificatePath     string `json:"clientCertificate,omitempty"`
```
rna-afk, line 46: Wasn't `certificatePath and certificatePassword` how azure expects certificates to be populated in the `osServicePrincipal.json` file? 

Even if the installer does not need it with that key name, we should have the same name as azure expects so it makes it easier for users to use the json file for both azure auth(like az client) and the installer without change.

r4f4: autorest/auth was the lib that was loading that file, but I'm removing it with this PR since it's going out of support by the end of March. I've actually just changed it to the same values the lib was expecting https://github.com/Azure/go-autorest/blob/9038e4a609b1899f0eb382d03c3e823b70537125/autorest/azure/auth/auth.go#L348-L349

patrickdillon: We do need to keep backwards compatibility for existing users though; and this change would break anyone who is currently authenticating with certs. 

We could potentially support both fields, but it may not be worth it considering the limited scope of this feature.

rna-afk: Just users using the old json file need to change the key name for these fields. If they choose to update the installer version they have to also change this key which they won't know of.

r4f4: If they do, then this never worked in the first place. `autorest/auth` was loading the json file, and it expects the `clientCertificate`  and `clientCertificatePassword` names.

r4f4: See https://github.com/Azure/go-autorest/blob/main/autorest/azure/auth/auth.go#L248-L249 which calls https://github.com/Azure/go-autorest/blob/main/autorest/azure/auth/auth.go#L321-L349

r4f4: These fields in Credentials would have been used if we ever asked for Certificate details during the survey. That's the only time when we use that json annotation to save the Credentials object to a file.

rna-afk: You're right, looks like before this PR, it could be anything since we don't use it to write to file since we don't collect certs in survey. Shouldn't have a problem with backward compatibility.

r4f4: Given how confusing this change is, maybe I should put it in its own commit with a reason why it's needed?

r4f4:
Added a new commit to isoloate and explain why the json annotations were changed. No code changes were introduced.

pkg/asset/installconfig/azure/session.go
patrickdillon: I was mistaken/confused in my original comment and I agree this does not break backward compatibility. I missed that these JSON tags are only used for serialization--not deserialization. By that same token, it seems like we could just remove the json tags for the cert fields, right? 

But this is definitely fine with me as it is.

r4f4: We cannot remove them because with this PR they **are** used for deserialization now. That was the point of changing the annotation to be compatible with what autorest expected.

patrickdillon: :man_facepalming:

patrickdillon:
/lgtm
e2e-azure-ovn was successful install. just one e2e test failure re registry.

PR #6905, opened 2023-02-27
OCPBUGS-7991: vSphere - only create RP when cluster is not empty

jcpowermac:


petr-balogh (approved):

jcpowermac:
vsphere install variants installed and are on to tests. Should be ready to go.

PR #6906, opened 2023-02-27
OCPBUGS-7993: hack/build-coreos-manifest: support OKD

vrutkovs:
If TAGS env var contains "fcos" (its an alias to OKD) CoreOS manifest script should use
FCOS json. This is necessary for OKD clusters to be installable via
Hypershift


LorbusChris (approved):
/lgtm

PR #6907, opened 2023-02-27
[release-4.12] OCPBUGS-7481: Fix file check for loading openshift manifests

rna-afk:
When generating the openshift manifests, it tries to load the dependency control plane asset that creates files in the openshift directory. It then checks to see if the openshift manifest is actually already generated before this by checking the openshift directory for files that are not generated by the machine asset.

The newly created CPMS file was missing from the check, hence the asset loader always thought the asset was generated since there are files that exist other than the machine files and exits but it actually had to be newly generated with 4 files that were crucial to installation.

Adding the CPMS files to the list of machine asset files.

### DIFF
--- a/pkg/asset/machines/master.go
+++ b/pkg/asset/machines/master.go
@@ -617,6 +617,17 @@ func (m *Master) Load(f asset.FileFetcher) (found bool, err error) {
 	}
 	m.MachineFiles = fileList
 
+	file, err = f.FetchByName(filepath.Join(directory, controlPlaneMachineSetFileName))
+	if err != nil {
+		if os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


r4f4:
So this is a manual cherry-pick of https://github.com/openshift/installer/pull/6821 + https://github.com/openshift/installer/pull/6887 ?
We are missing the comment and the warning in https://github.com/openshift/installer/pull/6887. Is that on purpose?

rna-afk:
> We are missing the comment and the warning in https://github.com/openshift/installer/pull/6887. Is that on purpose?

Knew I missed something. Thanks it's fixed.

r4f4 (commented):
/lgtm

jianli-wei (approved):

PR #6908, opened 2023-02-28
openstack-manifests: Option to run selected test cases

pierreprinetti:
Add a `-r <regexp>` to `test-manifests.sh` to optionally only run test cases with a matching name.


mandre (commented):
/lgtm
/approve

PR #6911, opened 2023-02-28
OSASINFRA-3111: Warn if user/pass auth used in clouds.yaml

stephenfin:
[Application credentials][1] have a number of advantage over username/password combinations, not least the ease of rotation. However, the `clouds.yaml` files generated by various OpenStack tools like Horizon default to username/password combinations and users tend to pass these through unmodified.

We don't want to outright reject these - at least, not yet - but we can indicate an explicit preference for application credentials. We do this by inspecting the `clouds.yaml` file that the user has provided before we store it as a secret for use in the deployment. If the user has provided password-based credentials, we will emit a log encouraging them to migrate and warning about the potential removal of support for password-based auth in a future release.

This is the less magical replacement for #6858.

[1]: https://docs.openstack.org/keystone/latest/user/application_credentials.html


pierreprinetti:
Are we really planning on deprecating password-based authentication?

stephenfin:
I've said we're "warning" users but I've actually logged at `info` level because this doesn't appear to be sufficiently important to warrant a warning (i.e. things will not explode right now if they don't do something about it). Happy to change this if needed though.

stephenfin:
> Are we really planning on deprecating password-based authentication?

I don't think we'd remove it unless e.g. some standards body suggests we do. We might want to deprecate it though (note: deprecation != deprecation for removal). I've said ["may be removed"](https://en.wikipedia.org/wiki/Weasel_word) because of this bit of uncertainty though.

stephenfin:
Agree with both suggestions. Added

mandre (commented):
/lgtm

pkg/asset/manifests/openshift.go
```diff
@@ -159,6 +160,15 @@ func (o *Openshift) Generate(dependencies asset.Parents) error {
 			cloud.CACertFile = "/etc/kubernetes/static-pod-resources/configmaps/cloud-config/ca-bundle.pem"
 		}
 
+		// Application credentials are easily rotated in the event of a leak and should be preferred. Encourage their use.
+		if cloud.AuthInfo != nil && (cloud.AuthType == clientconfig.AuthPassword || cloud.AuthType == clientconfig.AuthV2Password || cloud.AuthType == clientconfig.AuthV3Password) {
```
barbacbd, line 164: This check is relatively early in install process, if the user catches it in time then they could easily change it. Should we consider `logrus.Warnf` so that the message is easier to find?

mandre: Changed the log level to a warning. PTAL.

MaysaMacedo:
@barbacbd hello, can you take another look on this PR? Thanks

pkg/asset/manifests/openshift.go
rna-afk, line 164: Let's not do the chained if conditions.

```
authTypes := sets.NewString(clientconfig.AuthPassword, clientconfig.AuthV2Password, clientconfig.AuthV3Password)
if cloud.AuthInfo != nil && authTypes.Has(cloud.AuthType) {
```

stephenfin: Done

mandre (commented):
/lgtm

PR #6913, opened 2023-02-28
validation: Allow machine-pool validation based on FeatureSet

pierreprinetti:
After this change, validations of the OpenStack machine-pool will be able to depend on the `installConfig.FeatureSet` value to allow or reject TechPreview-only configurations.

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -433,19 +433,19 @@ func validateControlPlane(featureSet configv1.FeatureSet, platform *types.Platfo
 func validateCompute(featureSet configv1.FeatureSet, platform *types.Platform, control *types.MachinePool, pools []types.MachinePool, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 	poolNames := map[string]bool{}
-	for i, p := range pools {
+	for i := range pools {
 		poolFldPath := fldPath.Index(i)
-		if p.Name != types.MachinePoolComputeRoleName {
-			allErrs = append(allErrs, field.NotSupported(poolFldPath.Child("name"), p.Name, []string{"worker"}))
+		if pools[i].Name != types.MachinePoolComputeRoleName {
+			allErrs = append(allErrs, field.NotSupported(poolFldPath.Child("name"), pools[i].Name, []string{"worker"}))
 		}
-		if poolNames[p.Name] {
-			allErrs = append(allErrs, field.Duplicate(poolFldPath.Child("name"), p.Name))
+		if poolNames[pools[i].Name] {
+			allErrs = append(allErrs, field.Duplicate(poolFldPath.Child("name"), pools[i].Name))
 		}
-		poolNames[p.Name] = true
-		if control != nil && control.Architecture != p.Architecture {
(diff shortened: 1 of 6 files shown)


pkg/types/validation/installconfig.go
```diff
@@ -418,34 +418,34 @@ func validateClusterNetwork(n *types.Networking, cn *types.ClusterNetworkEntry,
...
-	allErrs = append(allErrs, ValidateMachinePool(platform, pool, fldPath)...)
+	allErrs = append(allErrs, ValidateMachinePool(featureSet, platform, pool, fldPath)...)
 	return allErrs
 }
 
-func validateCompute(platform *types.Platform, control *types.MachinePool, pools []types.MachinePool, fldPath *field.Path) field.ErrorList {
+func validateCompute(featureSet configv1.FeatureSet, platform *types.Platform, control *types.MachinePool, pools []types.MachinePool, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 	poolNames := map[string]bool{}
-	for i, p := range pools {
```
stephenfin, line 436: This doesn't seem crucial and probably would have been better left to a separate commit/PR to minimise the diff

stephenfin (commented):
/lgtm just plumbing for future changes

pkg/types/validation/installconfig.go
pierreprinetti: The linter didn’t let me commit without this change. The static analysis on this file was probably triggered by my change, which appears, as you note, unrelated.

pierreprinetti: It could have been a separate commit indeed.

r4f4: It's a "feature" of the new linter: it only analyses changes after a specific revision (`--new-from-rev=75173a17cf`) so we don't have to fix all the warnings in the entire codebase at once. The downside is that "old" warnings are triggered when new changes touch the area. A separate commit is a good compromise here, otherwise you'd have to wait for the linting fix PR to be merged.

patrickdillon:
Adding FeatureSet validation at this level seems like it would make it more difficult to graduate a feature from tech preview to GA, because the validation would be distributed throughout the code base. 

The idea for featureset validation was that it would be all in one place, currently that place would be here https://github.com/openshift/installer/blob/9a5ef710b67a8478b7b9cb1daa5b005d41c659f9/pkg/types/validation/installconfig.go#L1000-L1006
But that location is definitely not set in stone, the point is that it would seem better for developer experience if we could keep all tech preview installconfig validation in one place. So in this case, the change I would recommend is that you move validation of the machine pools to this location, and if a certain value is set you also ensure the appropriate tech preview feature set.

Reading this code, it's actually hard for me to tell what specific validation is being added. Is this just plumbing to allow future validation or is there a specific value you're trying to gate?

mandre:
> Reading this code, it's actually hard for me to tell what specific validation is being added. Is this just plumbing to allow future validation or is there a specific value you're trying to gate?

Correct, it's refactoring in preparation for an upcoming patch we're still working on.
This is going to change machinepool validation, hence why we did it like that. Let me check with @pierreprinetti if it's possible to move everything under one place.

patrickdillon:
/close
in favor of #6916

PR #6915, opened 2023-03-01
OCPBUGS-7015: Relax MachineCIDR check for vSphere, Nutanix

patrickdillon:
a58ac55e8a07e442c73747ffcbdbbc3f584b276d introduced many network validations for on-prem platforms. We want to relax one of those, the check that Ingress & API VIPs are in the machinenetwork because the check prevents using the TUI survey to generate the install config and this (mismatched) configuration has been allowed in all of OpenShift 4 without issue.

Fixes OCPBUGS-7015

/cc creydr jcpowermac sadasu cybertron rvanderp3


patrickdillon:
/retitle OCPBUGS-7015: Relax MachineCIDR check for vSphere, Nutanix

PR #6916, opened 2023-03-01
openstack: consistent TechPreview-only feature validation

pierreprinetti:
With this change, filling in TechPreview-only configuration flags in OpenStack throws similar errors as to what happens on other platforms.

Supersedes #6913

/cc mandre patrickdillon

### DIFF
--- a/pkg/types/validation/installconfig_test.go
+++ b/pkg/types/validation/installconfig_test.go
@@ -1931,6 +1931,20 @@ func TestValidateInstallConfig(t *testing.T) {
 			expectedError: "platform.openstack.apiVIPs: Invalid value: \"foobar\": \"foobar\" is not a valid IP",
 		},
 		{
+			name: "should reject load balancer on OpenStack if not TechPreviewNoUpgrade",
+			installConfig: func() *types.InstallConfig {
+				c := validInstallConfig()
+				c.Platform = types.Platform{
+					OpenStack: validOpenStackPlatform(),
+				}
+				c.Platform.OpenStack.LoadBalancer = &configv1.OpenStackPlatformLoadBalancer{
+					Type: configv1.LoadBalancerTypeOpenShiftManagedDefault,
+				}
+				return c
+			}(),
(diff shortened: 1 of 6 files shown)


mandre:
/lgtm cancel
golint and unit tests are unhappy.

pkg/types/validation/installconfig_test.go
```diff
@@ -1930,6 +1930,20 @@ func TestValidateInstallConfig(t *testing.T) {
...
+				c := validInstallConfig()
+				c.Platform = types.Platform{
+					OpenStack: validOpenStackPlatform(),
+				}
+				c.Platform.OpenStack.LoadBalancer = &configv1.OpenStackPlatformLoadBalancer{
+					Type: configv1.LoadBalancerTypeOpenShiftManagedDefault,
+				}
+				return c
+			}(),
+			expectedError: `platform.openstack.loadBalancer: Forbidden: the TechPreviewNoUpgrade feature set must be enabled to use this field`,
```
pierreprinetti, line 1945: Please do not test OpenStack's platform-specific logic here. We have our check in a separate directory so that we can operate autonomously on platform-specific fields; if the test is in the shared logic, it means that we again need Installer-team's approval for a platform-specific matter.

pierreprinetti, line 1945: @mandre 
Please do not test OpenStack's platform-specific logic here. We have our check in a separate directory so that we can operate autonomously on platform-specific fields; if the test is in the shared logic, it means that we again need Installer-team's approval for a platform-specific matter.

patrickdillon:
/approve

cool! this seems like a good pattern for platforms going forward. thank you.

pkg/types/validation/installconfig_test.go
mandre: Testing FilledInTechPreviewFields() like you did is good, but I'd also like to ensure `ValidateInstallConfig()` runs the openstack specific validations. Without some refactoring, I think the easiest way to achieve that is to restore the test in `pkg/types/validation/installconfig_test.go`.
I'll do that and fix the gofmt error at the same time.

mandre (commented):
/lgtm

PR #6917, opened 2023-03-01
OpenStack Failure domains

pierreprinetti:
Distribute Control plane machines across user-defined failure domains.

This feature is being release under a TechPreviewNoUpgrade FeatureSet.

Failure domains can be defined in the `controlPlane` machine-pool of `install-config.yaml` as follows:

```yaml
controlPlane:
  name: master
  platform:
    openstack:
      type: ${CONTROL_PLANE_FLAVOR}
      failureDomains:
      - computeAvailabilityZone: 'nova-1'
        storageAvailabilityZone: 'cinder-1'
        portTargets:
        - id: storage
          network:
            id: 8db6a48e-375b-4caa-b20b-5b9a7218bfe6
      - computeAvailabilityZone: 'nova-2'
        storageAvailabilityZone: 'cinder-2'
        portTargets:
        - id: storage
          network:
            id: 39a7b82a-a8a4-45a4-ba5a-288569a6edd1
      - computeAvailabilityZone: 'nova-3'
        storageAvailabilityZone: 'cinder-3'
        portTargets:
        - id: storage
          network:
            id: 8e4b4e0d-3865-4a9b-a769-559270271242
```

Each `failureDomains` entry can take an optional `computeAvailabilityZone` string, an optional `storageAvailabilityZone` string, and an optional `portTargets` array.

Each `portTargets` entry requires an arbirtary `id`, which must be unique per `failureDomain`. If `id` is exactly `control-plane`, then that `portTarget` is used instead of the default primary subnet (or instead of `machinesSubnet` if defined) as the first machine network.

Each `portTargets` entry takes an optional `network` object and an optional `fixedIPs` array (not represented in the example).

The `network` object taks an optional `name` string and an optional `id` string. `name` is ignored if `id` is passed.

Each `fixedIPs` entry takes a `subnet` object which syntax is [defined in the `machinev1alpha1` spec as `SubnetFilter`](https://github.com/openshift/api/blob/d170fcdc0fa638b664e4f35f2daf753cb4afe36b/machine/v1alpha1/types_openstack.go#L230-L281).

Note that unless an external load balancer is used, `portTargets` with id `control-plane` must all have one single subnet and must all refer to the same OpenStack subnet. As a consequence, the result will be similar as setting a `machinesSubnet`, except that Compute nodes will not follow.

### DIFF
--- a/data/data/openstack/masters/main.tf
+++ b/data/data/openstack/masters/main.tf
@@ -67,7 +67,7 @@ resource "openstack_blockstorage_volume_v3" "master_volume" {
   volume_type = var.openstack_master_root_volume_type
   image_id = data.openstack_images_image_v2.base_image.id
 
-  availability_zone = var.openstack_master_root_volume_availability_zones[count.index % length(var.openstack_master_root_volume_availability_zones)]
+  availability_zone = var.openstack_master_root_volume_availability_zones[count.index]
 }
 
 resource "openstack_compute_servergroup_v2" "master_group" {
@@ -99,7 +99,7 @@ resource "openstack_compute_instance_v2" "master_conf_0" {
   flavor_id = data.openstack_compute_flavor_v2.masters_flavor.id
   image_id = var.openstack_master_root_volume_size == null ? data.openstack_images_image_v2.base_image.id : null
   security_groups = local.master_sg_ids
-  availability_zone = var.openstack_master_availability_zones[0 % length(var.openstack_master_availability_zones)]
+  availability_zone = var.openstack_master_availability_zones[0]
   user_data = element(
     data.ignition_config.master_ignition_config.*.rendered,
     0,
@@ -125,6 +125,14 @@ resource "openstack_compute_instance_v2" "master_conf_0" {
   }
 
   dynamic "network" {
+    for_each = [for port in openstack_networking_port_v2.master_0_failuredomain : port.id]
+
+    content {
+      port = network.value
+    }
+  }
+
+  dynamic "network" {
     for_each = var.openstack_additional_network_ids
 
     content {
@@ -147,7 +155,7 @@ resource "openstack_compute_instance_v2" "master_conf_1" {
   flavor_id = data.openstack_compute_flavor_v2.masters_flavor.id
   image_id = var.openstack_master_root_volume_size == null ? data.openstack_images_image_v2.base_image.id : null
   security_groups = local.master_sg_ids
-  availability_zone = var.openstack_master_availability_zones[1 % length(var.openstack_master_availability_zones)]
+  availability_zone = var.openstack_master_availability_zones[1]
   user_data = element(
     data.ignition_config.master_ignition_config.*.rendered,
     1,
@@ -170,6 +178,14 @@ resource "openstack_compute_instance_v2" "master_conf_1" {
 
   scheduler_hints {
     group = openstack_compute_servergroup_v2.master_group.id
+  }
+
+  dynamic "network" {
+    for_each = [for port in openstack_networking_port_v2.master_1_failuredomain : port.id]
+
+    content {
+      port = network.value
+    }
   }
 
   dynamic "network" {
@@ -197,7 +213,7 @@ resource "openstack_compute_instance_v2" "master_conf_2" {
   flavor_id = data.openstack_compute_flavor_v2.masters_flavor.id
   image_id = var.openstack_master_root_volume_size == null ? data.openstack_images_image_v2.base_image.id : null
   security_groups = local.master_sg_ids
-  availability_zone = var.openstack_master_availability_zones[2 % length(var.openstack_master_availability_zones)]
+  availability_zone = var.openstack_master_availability_zones[2]
   user_data = element(
     data.ignition_config.master_ignition_config.*.rendered,
     2,
@@ -223,6 +239,14 @@ resource "openstack_compute_instance_v2" "master_conf_2" {
   }
 
   dynamic "network" {
+    for_each = [for port in openstack_networking_port_v2.master_2_failuredomain : port.id]
+
+    content {
+      port = network.value
+    }
+  }
+
+  dynamic "network" {
     for_each = var.openstack_additional_network_ids
 
     content {
@@ -245,4 +269,61 @@ resource "openstack_compute_servergroup_v2" "server_groups" {
   for_each = var.openstack_worker_server_group_names
   name = each.key
   policies = [var.openstack_master_server_group_policy]
+}
+
+resource "openstack_networking_port_v2" "master_0_failuredomain" {
+  count = var.master_count > 0 ? length(var.openstack_additional_ports[0]) : 0
+
+  name = "${var.cluster_id}-master-0-${count.index}"
+  description = local.description
+  network_id = var.openstack_additional_ports[0][count.index].network_id
+  security_group_ids = concat(var.openstack_master_extra_sg_ids, [openstack_networking_secgroup_v2.master.id])
+  tags = ["openshiftClusterID=${var.cluster_id}"]
+
+  dynamic "fixed_ip" {
+    for_each = var.openstack_additional_ports[0][count.index].fixed_ips
+
+    content {
+      subnet_id = fixed_ip.value["subnet_id"]
+      ip_address = fixed_ip.value["ip_address"]
+    }
+  }
+}
+
+resource "openstack_networking_port_v2" "master_1_failuredomain" {
+  count = var.master_count > 1 ? length(var.openstack_additional_ports[1]) : 0
+
+  name = "${var.cluster_id}-master-1-${count.index}"
+  description = local.description
+  network_id = var.openstack_additional_ports[1][count.index].network_id
+  security_group_ids = concat(var.openstack_master_extra_sg_ids, [openstack_networking_secgroup_v2.master.id])
+  tags = ["openshiftClusterID=${var.cluster_id}"]
+
+  dynamic "fixed_ip" {
+    for_each = var.openstack_additional_ports[1][count.index].fixed_ips
+
+    content {
+      subnet_id = fixed_ip.value["subnet_id"]
+      ip_address = fixed_ip.value["ip_address"]
+    }
+  }
+}
+
+resource "openstack_networking_port_v2" "master_2_failuredomain" {
+  count = var.master_count > 2 ? length(var.openstack_additional_ports[2]) : 0
+
+  name = "${var.cluster_id}-master-2-${count.index}"
+  description = local.description
+  network_id = var.openstack_additional_ports[2][count.index].network_id
+  security_group_ids = concat(var.openstack_master_extra_sg_ids, [openstack_networking_secgroup_v2.master.id])
+  tags = ["openshiftClusterID=${var.cluster_id}"]
+
+  dynamic "fixed_ip" {
+    for_each = var.openstack_additional_ports[2][count.index].fixed_ips
+
+    content {
+      subnet_id = fixed_ip.value["subnet_id"]
+      ip_address = fixed_ip.value["ip_address"]
+    }
+  }
 }
--- a/data/data/openstack/masters/private-network.tf
+++ b/data/data/openstack/masters/private-network.tf
@@ -1,9 +1,12 @@
 locals {
   # Create subnet for the first MachineNetwork CIDR if we need to
   nodes_cidr_block = var.machine_v4_cidrs[0]
-  nodes_subnet_id  = var.openstack_machines_subnet_id != "" ? var.openstack_machines_subnet_id : openstack_networking_subnet_v2.nodes[0].id
-  nodes_network_id = var.openstack_machines_network_id != "" ? var.openstack_machines_network_id : openstack_networking_network_v2.openshift-private[0].id
-  create_router    = (var.openstack_external_network != "" && var.openstack_machines_subnet_id == "") ? 1 : 0
+  nodes_default_port = var.openstack_default_machines_port != null ? var.openstack_default_machines_port : {
+    network_id = openstack_networking_network_v2.openshift-private[0].id,
+    fixed_ips  = [{ subnet_id = openstack_networking_subnet_v2.nodes[0].id, ip_address = "" }],
+  }
+  nodes_ports   = [for port in var.openstack_machines_ports : port != null ? port : local.nodes_default_port]
+  create_router = (var.openstack_external_network != "" && var.openstack_default_machines_port == null) ? 1 : 0
 }
 
 data "openstack_networking_network_v2" "external_network" {
@@ -14,7 +17,7 @@ data "openstack_networking_network_v2" "external_network" {
 }
 
 resource "openstack_networking_network_v2" "openshift-private" {
-  count          = var.openstack_machines_subnet_id == "" ? 1 : 0
+  count          = var.openstack_default_machines_port == null ? 1 : 0
   name           = "${var.cluster_id}-openshift"
   admin_state_up = "true"
   description    = local.description
@@ -22,12 +25,12 @@ resource "openstack_networking_network_v2" "openshift-private" {
 }
 
 resource "openstack_networking_subnet_v2" "nodes" {
-  count           = var.openstack_machines_subnet_id == "" ? 1 : 0
+  count           = var.openstack_default_machines_port == null ? 1 : 0
   name            = "${var.cluster_id}-nodes"
   description     = local.description
   cidr            = local.nodes_cidr_block
   ip_version      = 4
-  network_id      = local.nodes_network_id
+  network_id      = openstack_networking_network_v2.openshift-private[0].id
   tags            = ["openshiftClusterID=${var.cluster_id}"]
   dns_nameservers = var.openstack_external_dns
 
@@ -46,7 +49,7 @@ resource "openstack_networking_port_v2" "masters" {
   description = local.description
 
   admin_state_up = "true"
-  network_id     = local.nodes_network_id
+  network_id     = local.nodes_ports[count.index].network_id
   security_group_ids = concat(
     var.openstack_master_extra_sg_ids,
     [openstack_networking_secgroup_v2.master.id],
@@ -58,8 +61,13 @@ resource "openstack_networking_port_v2" "masters" {
     value = var.cluster_domain
   }
 
-  fixed_ip {
-    subnet_id = local.nodes_subnet_id
+  dynamic "fixed_ip" {
+    for_each = local.nodes_ports[count.index].fixed_ips
+
+    content {
+      subnet_id  = fixed_ip.value["subnet_id"]
+      ip_address = fixed_ip.value["ip_address"]
+    }
   }
 
   dynamic "allowed_address_pairs" {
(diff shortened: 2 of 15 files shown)


data/data/openstack/variables-openstack.tf
```diff
@@ -341,16 +353,52 @@ variable "openstack_master_server_group_policy" {
...
+/*   type        = list(string) */
+/*   default     = [] */
+/*   description = "ID of each master's first subnet." */
+/* } */
+
+/* variable "openstack_machines_network_ids" { */
+/*   type        = list(string) */
+/*   default     = [] */
+/*   description = "ID of each master's first network." */
+/* } */
```
pierreprinetti, line 378: Leftover

data/data/openstack/masters/main.tf
```diff
@@ -147,7 +155,7 @@ resource "openstack_compute_instance_v2" "master_conf_1" {
   flavor_id = data.openstack_compute_flavor_v2.masters_flavor.id
   image_id = var.openstack_master_root_volume_size == null ? data.openstack_images_image_v2.base_image.id : null
   security_groups = local.master_sg_ids
-  availability_zone = var.openstack_master_availability_zones[1 % length(var.openstack_master_availability_zones)]
+  availability_zone = var.openstack_master_availability_zones[1]
```
pierreprinetti, line 158: We used to pass a slice with the declared zones. Terraform was responsible for distributing them.

Now tfvars builds a slice of length(number-of-masters) containing the zones ready to be used, so Terraform just has to pick the one corresponding to the current instance.

data/data/openstack/masters/main.tf
```diff
@@ -99,7 +99,7 @@ resource "openstack_compute_instance_v2" "master_conf_0" {
   flavor_id = data.openstack_compute_flavor_v2.masters_flavor.id
   image_id = var.openstack_master_root_volume_size == null ? data.openstack_images_image_v2.base_image.id : null
   security_groups = local.master_sg_ids
-  availability_zone = var.openstack_master_availability_zones[0 % length(var.openstack_master_availability_zones)]
+  availability_zone = var.openstack_master_availability_zones[0]
```
pierreprinetti, line 102: We used to pass a slice with the declared zones. Terraform was responsible for distributing them.

Now tfvars builds a slice of length(number-of-masters) containing the zones ready to be used, so Terraform just has to pick the one corresponding to the current instance.

data/data/openstack/masters/main.tf
```diff
@@ -197,7 +213,7 @@ resource "openstack_compute_instance_v2" "master_conf_2" {
   flavor_id = data.openstack_compute_flavor_v2.masters_flavor.id
   image_id = var.openstack_master_root_volume_size == null ? data.openstack_images_image_v2.base_image.id : null
   security_groups = local.master_sg_ids
-  availability_zone = var.openstack_master_availability_zones[2 % length(var.openstack_master_availability_zones)]
+  availability_zone = var.openstack_master_availability_zones[2]
```
pierreprinetti, line 216: (see above: "We used to pass a slice with the declared zones")

data/data/openstack/masters/main.tf
```diff
@@ -132,6 +132,14 @@ resource "openstack_compute_instance_v2" "master_conf_0" {
     }
   }
 
+  dynamic "network" {
+    for_each = [for port in openstack_networking_port_v2.master_0_failuredomain : port.id]
```
pierreprinetti, line 136: these are the ports we create based on the non-control-plane portTargets in the corresponding failureDomain.

data/data/openstack/masters/private-network.tf
```diff
@@ -1,9 +1,12 @@
...
   nodes_cidr_block = var.machine_v4_cidrs[0]
-  nodes_subnet_id  = var.openstack_machines_subnet_id != "" ? var.openstack_machines_subnet_id : openstack_networking_subnet_v2.nodes[0].id
-  nodes_network_id = var.openstack_machines_network_id != "" ? var.openstack_machines_network_id : openstack_networking_network_v2.openshift-private[0].id
-  create_router    = (var.openstack_external_network != "" && var.openstack_machines_subnet_id == "") ? 1 : 0
+  nodes_default_port = var.openstack_default_machines_port != null ? var.openstack_default_machines_port : {
+    network_id = openstack_networking_network_v2.openshift-private[0].id,
+    fixed_ips  = [{ subnet_id = openstack_networking_subnet_v2.nodes[0].id, ip_address = "" }],
+  }
+  nodes_ports   = [for port in var.openstack_machines_ports : port != null ? port : local.nodes_default_port]
+  create_router = (var.openstack_external_network != "" && var.openstack_default_machines_port == null) ? 1 : 0
```
pierreprinetti, line 9: I have removed two values that we used to pass two values to Terraform:
* `openstack_machines_network_id` was the network of `machinesSubnet`, or empty
* `openstack_machines_subnet_id` was `machinesSubnet`, or it was empty

If empty, Terraform created a new network.

Now the logic is similar, but I have started using this new `port` structure instead. The reason for this change is that we need to be able to pass multiple subnets per network. This is irrelevant for machinesSubnet and for the Installer-generated network, but it's essential when we have a control-plane portTarget. In that case, we want to be able to define a port with multiple subnets (most notably, for IPv6).

pierreprinetti: The old variables are replaced with:
* `openstack_default_machines_port`: machinesSubnet, or null
* `openstack_machines_ports`: array of length(number-of-masters) defining the primary port for the master. If null, it's going to be the network created by Terraform.

Here we replace the null values, if present, with data from the network created by Terraform.

data/data/openstack/masters/private-network.tf
```diff
@@ -58,8 +61,13 @@ resource "openstack_networking_port_v2" "masters" {
...
 
-  fixed_ip {
-    subnet_id = local.nodes_subnet_id
+  dynamic "fixed_ip" {
+    for_each = local.nodes_ports[count.index].fixed_ips
+
+    content {
+      subnet_id  = fixed_ip.value["subnet_id"]
+      ip_address = fixed_ip.value["ip_address"]
+    }
```
pierreprinetti, line 70: This becomes a dynamic bloc, because with ports we are not limited any more to just one subnet per port. This array will have length!=1 if a failuredomain  defines a control-plane portTarget, AND that portTarget has multiple fixedIPs.

data/data/openstack/variables-openstack.tf
```diff
@@ -306,6 +306,18 @@ variable "openstack_additional_network_ids" {
...
+    network_id = string
+    fixed_ips = list(object({
+      subnet_id  = string
+      ip_address = string
+    }))
+  })))
+  description = "Additional ports for each master node."
+  default     = [[], [], []]
+}
+
```
pierreprinetti, line 320: Here, and below (`openstack_default_machines_port`, `openstack_machines_ports`) we define a Terraform data structure that mimics the input of a Terraform port. It allows multiple fixed_ips. It's the structure as defined in `tfvars/openstack/ports.go`

pkg/asset/machines/openstack/machines.go
```diff
@@ -56,31 +56,33 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 		total = *pool.Replicas
 	}
 	machines := make([]machineapi.Machine, 0, total)
-	providerConfigs := map[string]*machinev1alpha1.OpenstackProviderSpec{}
```
pierreprinetti, line 59: This map was used to accumulate identical providerSpecs so that they could be reused. This optimisation is now problematic because we distribute differently; I have removed it.

pkg/asset/machines/openstack/machines.go
```diff
@@ -108,26 +110,49 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 	return machines, nil
 }
 
-func generateProvider(clusterID string, platform *openstack.Platform, mpool *openstack.MachinePool, osImage string, az string, role, userDataSecret string, trunkSupport bool, rootVolumeAZ string) (*machinev1alpha1.OpenstackProviderSpec, error) {
-	var networks []machinev1alpha1.NetworkParam //nolint:prealloc // declared here for the conditional initialization
+func generateProvider(clusterID string, platform *openstack.Platform, mpool *openstack.MachinePool, osImage string, role, userDataSecret string, trunkSupport bool, failureDomain openstack.FailureDomain) (*machinev1alpha1.OpenstackProviderSpec, error) {
+	var controlPlaneNetwork machinev1alpha1.NetworkParam
+	additionalNetworks := make([]machinev1alpha1.NetworkParam, 0, len(failureDomain.PortTargets)+len(mpool.AdditionalNetworkIDs))
```
pierreprinetti, line 115: The final slice of `networkParam` that we put in the providerSpec will be:
```
controlPlaneNetwork + additionalNetworks...
```

I keep them separate until the end just for readability.

pkg/asset/machines/openstack/machinesets.go
```diff
@@ -47,7 +47,19 @@ func MachineSets(clusterID string, config *types.InstallConfig, pool *types.Mach
...
+			clusterID,
+			platform,
+			mpool,
+			osImage,
+			role,
+			userDataSecret,
+			trunkSupport,
+			openstack.FailureDomain{
+				ComputeAvailabilityZone: az,
+				StorageAvailabilityZone: volumeAZs[idx%len(volumeAZs)],
```
pierreprinetti, line 60: We pass exactly the same data as before (no failure domains for Compute nodes!). We just pack them in an `openstack.FailureDomain` for consistency with masters. I think it's better than passing regular strings.

pkg/tfvars/openstack/openstack.go
```diff
@@ -179,19 +174,46 @@ func TFVars(
 		additionalNetworkIDs = mastermpool.AdditionalNetworkIDs
 	}
 
-	var additionalSecurityGroupIDs []string
-	if mastermpool != nil {
-		additionalSecurityGroupIDs = mastermpool.AdditionalSecurityGroupIDs
+	// defaultMachinesPort carries the machinesSubnet (and its resolved
+	// network) if provided.
+	var defaultMachinesPort *terraformPort
```
mdbooth, line 179: I see we're substituting MachinesSubnet here, but what about the case where it isn't defined and we're using a terraform-created network?

Assuming there's some terraform logic to substitute that later, couldn't we just put it all here? The split logic is harder to follow.

pkg/tfvars/openstack/openstack.go
```diff
@@ -33,12 +33,16 @@ func TFVars(
...
-		machinesSubnet = installConfig.Config.Platform.OpenStack.MachinesSubnet
+		cloud        = installConfig.Config.Platform.OpenStack.Cloud
+		mastermpool  = installConfig.Config.ControlPlane.Platform.OpenStack
+		defaultmpool = installConfig.Config.OpenStack.DefaultMachinePlatform
 	)
 
+	networkClient, err := clientconfig.NewServiceClient("network", openstackdefaults.DefaultClientOpts(cloud))
+	if err != nil {
+		return nil, fmt.Errorf("failed to build an OpenStack service client: %w", err)
+	}
```
pierreprinetti, line 44: We use `networkClient` to resolve some missing information, like machinesSubnet's network. Since it's used more than once, I thought I'd instantiate the client once. That also makes more evident where we're using Gophercloud in function calls.

pkg/tfvars/openstack/openstack.go
```diff
@@ -101,24 +105,15 @@ func TFVars(
...
-		seen := make(map[string]struct{})
-		for _, config := range masterSpecs {
-			if _, ok := seen[config.AvailabilityZone]; !ok {
-				zones = append(zones, config.AvailabilityZone)
-				seen[config.AvailabilityZone] = struct{}{}
-			}
-		}
+	computeAvailabilityZones := make([]string, len(masterSpecs))
+	for i := range computeAvailabilityZones {
+		computeAvailabilityZones[i] = masterSpecs[i].AvailabilityZone
```
pierreprinetti, line 110: We used to pick the availability zone from the install-config. Instead, this function now takes zones from machine generation. No need to duplicate logic here.

The interesting side effect is that `zones` is no longer the list of the available zones: it's now a list where each index corresponds to a specific machine. This means less computation on the Terraform side.

pkg/tfvars/openstack/openstack.go
```diff
@@ -101,24 +105,15 @@ func TFVars(
...
-	{
-		if defaultmpool != nil && defaultmpool.RootVolume != nil {
-			masterRootVolumeAvailabilityZones = defaultmpool.RootVolume.Zones
-		}
-		if mastermpool != nil && mastermpool.RootVolume != nil && mastermpool.RootVolume.Zones != nil {
-			masterRootVolumeAvailabilityZones = mastermpool.RootVolume.Zones
+	storageAvailabilityZones := make([]string, len(masterSpecs))
+	for i := range storageAvailabilityZones {
+		if masterSpecs[i].RootVolume != nil {
+			storageAvailabilityZones[i] = masterSpecs[i].RootVolume.Zone
```
pierreprinetti, line 116: Again: no need to recompute storage zones. We just pick what we have in Machines.

mandre, line 116: Perhaps add a comment explaining that maintaining the index is important. I was confused at first that `storageAvailabilityZones` could have holes in it.

pkg/tfvars/openstack/ports.go
```diff
@@ -0,0 +1,103 @@
...
+		CIDR:            subnetFilter.CIDR,
+		IPv6AddressMode: subnetFilter.IPv6AddressMode,
+		IPv6RAMode:      subnetFilter.IPv6RAMode,
+		ID:              subnetFilter.ID,
+		SubnetPoolID:    subnetFilter.SubnetPoolID,
+		Tags:            subnetFilter.Tags,
+		TagsAny:         subnetFilter.TagsAny,
+		NotTags:         subnetFilter.NotTags,
+		NotTagsAny:      subnetFilter.NotTagsAny,
+		Limit:           1,
```
mdbooth, line 82: Adding a Limit isn't consistent here: we normally return all results and throw an error if there's more than one. Otherwise the behaviour is non-deterministic.

data/data/openstack/variables-openstack.tf
```diff
@@ -341,16 +353,52 @@ variable "openstack_master_server_group_policy" {
...
+/* } */
+
+variable "openstack_default_machines_port" {
+  type = object({
+    network_id = string
+    fixed_ips = list(object({
+      subnet_id  = string
+      ip_address = string
+    }))
+  })
```
mdbooth, line 387: Is there a DRY opportunity here? Does terraform have typedefs?

data/data/openstack/masters/main.tf
```diff
@@ -132,6 +132,14 @@ resource "openstack_compute_instance_v2" "master_conf_0" {
     }
   }
 
+  dynamic "network" {
```
mdbooth, line 135: Does terraform add these in the order they're defined? I hope so, and these need to be above openstack_additional_network_ids.

mdbooth (commented):
I understand there is some mechanism to prevent certain parts of install-config being used unless the TechPreviewNoUpgrade flag is set. We need to do that. If it's done in this patch I've missed it.

mandre (commented):
I'm currently in the middle of reviewing the patch but need to drop soon. I'll leave the comments I have so far and will continue review later when I'm back.

pkg/tfvars/openstack/openstack.go
```diff
@@ -179,19 +174,46 @@ func TFVars(
...
 
-	var machinesNetwork string
-	if machinesSubnet != "" {
-		machinesNetwork, err = getNetworkFromSubnet(cloud, machinesSubnet)
-		if err != nil {
-			return nil, err
+	machinesPorts := make([]*terraformPort, len(masterSpecs))
+	additionalPorts := make([][]terraformPort, len(masterSpecs))
+	for i := range masterSpecs {
+		additionalPorts[i] = []terraformPort{}
```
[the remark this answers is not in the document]
pierreprinetti: If no terraformPort is added, and this line is removed, the content of the json array in this position is `null`, which Terraform does not expect.

pkg/tfvars/openstack/openstack.go
pierreprinetti: We can only leave blanks where the Terraform-created network should be: we don't have that data. I will try to make it more explicit that `nil` means that.

data/data/openstack/variables-openstack.tf
pierreprinetti: No idea.

scripts/openstack/manifest-tests/failure-domains/test_machines.py
```diff
@@ -0,0 +1,127 @@
...
+        for machine in self.machines:
+            actual_first_network = machine["spec"]["providerSpec"]["value"]["networks"][0]
+            actual_first_network_uuid = actual_first_network.get("uuid")
+            actual_first_network_subnet_uuids = [fixed_ip.get("filter", {}).get("id") for fixed_ip in actual_first_network.get("subnets", [{}])]
+            actual_first_networks[actual_first_network_uuid] = actual_first_network_subnet_uuids
+
+        # Assert that all expected networks and subnets are represented in machines
+        for expected_control_plane_network, expected_subnet_uuids in expected_control_plane_networks.items():
+            self.assertIn(expected_control_plane_network, actual_first_networks)
+            actual_subnet_uuids = actual_first_networks.get(expected_control_plane_network, [])
```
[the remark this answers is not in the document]
pierreprinetti: I keep forgetting that assertions in `unittest` stop the execution of the method.

data/data/openstack/masters/main.tf
pierreprinetti: No idea and OK

scripts/openstack/manifest-tests/failure-domains/test_machines.py
```diff
@@ -0,0 +1,127 @@
...
+                self.expected_failure_domains.append(failure_domain)
+            self.expected_machine_replicas = installconfig["controlPlane"]["replicas"]
+
+
+    def test_total_instance_number(self):
+        """Assert that there are as many Machines as required ControlPlane replicas."""
+        self.assertEqual(len(self.machines), self.expected_machine_replicas)
+
+
+    def test_control_plane_ports(self):
```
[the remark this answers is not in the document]
pierreprinetti: You suggestion makes complete sense. I am just unsure whether I have the strength to do that now.

pierreprinetti:
I have the changes you suggest locally. I'll wait a bit for the CI to complete before pushing.

pierreprinetti:
proxy and kuryr jobs succeeded!
I am now pushing a rebased commit which addresses your comments and adds validation.

pkg/types/openstack/validation/machinepool_test.go
```diff
@@ -47,29 +63,91 @@ func TestValidateMachinePool(t *testing.T) {
...
+		},
+		{
+			"failure domains together with zones",
+			testMachinePool(
+				withZones(),
+				withFailureDomain(openstack.FailureDomain{
+					PortTargets: []openstack.NamedPortTarget{{ID: "one"}, {ID: "two"}},
+				}),
+			),
+			"worker",
```
mandre, line 128: Should be master, otherwise we can't differentiate with the `failure domains on worker` test.

pkg/types/openstack/validation/machinepool_test.go
```diff
@@ -47,29 +63,91 @@ func TestValidateMachinePool(t *testing.T) {
...
+		},
+		{
+			"failure domains together with root volume zones",
+			testMachinePool(
+				withVolumeZones(),
+				withFailureDomain(openstack.FailureDomain{
+					PortTargets: []openstack.NamedPortTarget{{ID: "one"}, {ID: "two"}},
+				}),
+			),
+			"worker",
```
mandre, line 142: ditto

mandre (commented):
Looking at the code, this looks good to me. And I've seen successful runs of this PR to trust it's not breaking deployment in the most fundamental way.
Holding off my LGTM until the CI comes back.

patrickdillon:
/approve
for the generated install config. I will leave the openstack reviewing to the experts.

mandre (commented):
/lgtm
/cherry-pick release-4.13

mandre:
The 4.13 cherry-pick is going to depend on https://github.com/openshift/installer/pull/6934.

PR #6919, opened 2023-03-01
OCPBUGS-8094: In agent 'wait-for bootstrap' command, test ssh to Node0

bfournie:
If the Rest API and Kube API are not reachable, it may be because network connectivity checks are preventing the install from progressing (which will also prevent the Node0 SSH server from starting). Attempt to SSH to the node and provide instructions for further debug to the user upon failure.


zaneb:
SSHing into a node once the cluster is up has [side effects](https://github.com/openshift/machine-config-operator/blob/master/docs/MachineConfigDaemon.md#annotating-on-ssh-access) that I don't think we want to trigger.
/hold

pkg/agent/cluster.go
```diff
@@ -136,7 +142,14 @@ func (czero *Cluster) IsBootstrapComplete() (bool, bool, error) {
 		logrus.Trace("Current API Status: Agent Rest API: down, Bootstrap Kube API: down")
 		if !czero.installHistory.RestAPISeen && !czero.installHistory.ClusterKubeAPISeen {
 			logrus.Debug("Agent Rest API never initialized. Bootstrap Kube API never initialized")
-			logrus.Info("Waiting for cluster install to initialize. Sleeping for 30 seconds")
+			elapsedSinceInit := time.Since(czero.installHistory.ClusterInitTime)
+			// After allowing time for the interface to come up, check if Node0 can be accessed via ssh
+			if elapsedSinceInit > 2*time.Minute && !czero.CanSSHToNodeZero() {
+				logrus.Info("Cannot access Node0. There may be a network configuration problem, check console for additional info")
```
andfasano, line 148: For consistency I think it'd be better to use "rendezvous host" instead of node0

zaneb:
I was thinking that this could run at any time, but if it only happens before assisted comes up it's ok.
There are definitely times that the k8s API goes down temporarily after assisted-service has gone away, but you'd have to be fairly unlucky to (re-)run the command at that exact moment.
/hold cancel

pkg/agent/cluster.go
```diff
@@ -425,6 +438,25 @@ func (czero *Cluster) PrintInstallStatus(cluster *models.Cluster) error {
...
 
+// CanSSHToNodeZero checks if ssh to NodeZero succeeds.
+func (czero *Cluster) CanSSHToNodeZero() bool {
+
+	ip := czero.API.Rest.NodeZeroIP
+	port := 22
+
+	_, err := ssh.NewClient("core", net.JoinHostPort(ip, strconv.Itoa(port)), czero.API.Rest.NodeSSHKey)
+	if err != nil {
+		if errors.Is(err, syscall.ECONNREFUSED) || errors.Is(err, syscall.ETIMEDOUT) {
```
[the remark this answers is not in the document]
bfournie: Sure

pkg/agent/cluster.go
bfournie: Fixed

PR #6921, opened 2023-03-01
OCPBUGS-8109: AWS Control Plane MachineSet: Set PublicIP False 

patrickdillon:
When PublicIP is not set on the machine provider spec, we default to the subnet's settings.[0] We don't want control plane nodes to get public IPs so let's explicilty set PublicIP to false.

[0] https://github.com/openshift/api/blob/master/machine/v1beta1/types_awsprovider.go#L43-L46

### DIFF
--- a/pkg/asset/machines/aws/machines.go
+++ b/pkg/asset/machines/aws/machines.go
@@ -98,7 +98,7 @@ func Machines(clusterID string, region string, subnets map[string]string, pool *
 			}}
 		} else {
 			domain.Subnet.Type = machinev1.AWSIDReferenceType
-			domain.Subnet.ID = pointer.StringPtr(subnet)
+			domain.Subnet.ID = pointer.String(subnet)
 		}
 		failureDomains = append(failureDomains, domain)
 	}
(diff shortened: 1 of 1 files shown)


patrickdillon:
In the bug, we only see control plane nodes getting public IPs, but I wonder if we should set this to false for all: https://github.com/openshift/installer/blob/master/pkg/asset/machines/aws/machines.go#L161-L189?

pkg/asset/machines/aws/machines.go
```diff
@@ -247,6 +247,7 @@ func ConfigMasters(machines []machineapi.Machine, controlPlane *machinev1.Contro
 	for _, machine := range machines {
 		providerSpec := machine.Spec.ProviderSpec.Value.Object.(*machineapi.AWSMachineProviderConfig)
 		providerSpec.LoadBalancers = lbrefs
+		providerSpec.PublicIP = pointer.Bool(false)
```
r4f4, line 250: Could this break the public-only-subnets case?

patrickdillon: Yeah: that implementation may depend on this default. That job would need to edit these manifests to set public ip to true.

patrickdillon:
/close

This looks like a bug in a PR (#6371), not in our existing code.

PR #6925, opened 2023-03-02
OCPBUGS-7262: Disable systemd status while TUI showing

zaneb:
Bringing interfaces up in nmtui can cause systemd services to start, and
since the agent-tui runs before startup is completed (i.e. the runlevel
target is reached), this will cause systemd to print the service status
in random places on the console.

To avoid this, tell systemd to disable status messages before running
the agent-tui, and enable them again after it exits. The API to do this
is to send systemd the realtime signals SIGRTMIN+21 and SIGRTMIN+20,
respectively.


andfasano:
Thanks @zaneb I've tested it and it works like a charm!

/approve

PR #6928, opened 2023-03-03
OCPBUGS-8305: Power VS: Add resourceGroup to infrastructure manifest

mjturek:
After https://github.com/openshift/api/commit/925f75a26305c13db2d89c0e993060ab60c267ce ResourceGroup is needed by the machine-config-operator.


dharaneeshvrd:
LGTM as well.

But we need to address this https://github.com/openshift/installer/pull/6900#issuecomment-1449266619
We can take it up as a separate issue. Will close the other PR.

PR #6940, opened 2023-03-06
OpenStack: support multiple API and ingress VIPs

MaysaMacedo:
This commit enables the usage of more than one API and ingress VIP, which is required by Dual-Stack clusters.

### DIFF
--- a/data/data/openstack/masters/private-network.tf
+++ b/data/data/openstack/masters/private-network.tf
@@ -71,16 +71,16 @@ resource "openstack_networking_port_v2" "masters" {
   }
 
   dynamic "allowed_address_pairs" {
-    for_each = var.openstack_user_managed_load_balancer ? [] : [1]
+    for_each = var.openstack_user_managed_load_balancer ? [] : var.openstack_api_int_ips
     content {
-      ip_address = var.openstack_api_int_ip
+      ip_address = allowed_address_pairs.value
     }
   }
 
   dynamic "allowed_address_pairs" {
-    for_each = var.openstack_user_managed_load_balancer ? [] : [1]
+    for_each = var.openstack_user_managed_load_balancer ? [] : var.openstack_ingress_ips
     content {
-      ip_address = var.openstack_ingress_ip
+      ip_address = allowed_address_pairs.value
     }
   }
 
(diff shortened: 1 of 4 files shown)


MaysaMacedo:
Note that the Ports for the VIPs only uses the first VIP for now. The reason behind it is that it needs to be handled differently when using either BYON or basic installation, and it will be handled in separate PR.

MaysaMacedo:
/cc @dulek @mandre 
would you mind to have a look? Thanks

mandre (commented):
/lgtm

dulek:
Looks good to me too! We're not changing an API here as APIVIPs are there for a while now.

MaysaMacedo:
/override e2e-openstack-ovn

Attempting to override. Other openstack jobs have passed

data/data/openstack/masters/private-network.tf
```diff
@@ -102,7 +102,7 @@ resource "openstack_networking_port_v2" "api_port" {
 
     content {
       subnet_id  = fixed_ip.value["subnet_id"]
-      ip_address = var.openstack_api_int_ip
+      ip_address = var.openstack_api_int_ips[0]
```
pierreprinetti, line 105: why just the first?

data/data/openstack/masters/private-network.tf
```diff
@@ -122,7 +122,7 @@ resource "openstack_networking_port_v2" "ingress_port" {
 
     content {
       subnet_id  = fixed_ip.value["subnet_id"]
-      ip_address = var.openstack_ingress_ip
+      ip_address = var.openstack_ingress_ips[0]
```
pierreprinetti, line 125: why just the first?

data/data/openstack/masters/private-network.tf
pierreprinetti: I should have read your comment first!

> Note that the creation of the Neutron Ports for the VIPs only uses the first VIP for now. The reason behind it is that it needs to be handled differently when using either BYON or basic installation, and it will be handled in separate PR.

data/data/openstack/masters/private-network.tf
pierreprinetti: I should have read your comment first!

> Note that the creation of the Neutron Ports for the VIPs only uses the first VIP for now. The reason behind it is that it needs to be handled differently when using either BYON or basic installation, and it will be handled in separate PR.

PR #6944, opened 2023-03-07
OCPBUGS-8035: IBMCloud: Fix SSH Private bootstrap

cjschaef:
Fix IBM Cloud SecurityGroupRule for the bootstrap node, to allow SSH access to the node from any node within the private network, across all VPC availability zone subnets (10.x), such as a bastion host running IPI.

Related: https://issues.redhat.com/browse/OCPBUGS-8035

### DIFF
--- a/data/data/ibmcloud/bootstrap/main.tf
+++ b/data/data/ibmcloud/bootstrap/main.tf
@@ -2,6 +2,25 @@ locals {
   prefix              = var.cluster_id
   port_kubernetes_api = 6443
   port_machine_config = 22623
+
+  # If we need to setup SecurityGroupRules to SSH to bootstrap, for non-public clusters (no Floating IP)
+  # combine the Control Plane and Compute subnet CIDRs, to create rules for ingress on those CIDR's
+  all_subnet_cidrs = local.public_endpoints ? [] : concat(data.ibm_is_subnet.control_plane_subnets[*].ipv4_cidr_block, data.ibm_is_subnet.compute_subnets[*].ipv4_cidr_block)
+}
+
+############################################
+# Subnet lookup
+############################################
+data "ibm_is_subnet" "control_plane_subnets" {
+  count = local.public_endpoints ? 0 : length(var.control_plane_subnet_id_list)
+
+  identifier = var.control_plane_subnet_id_list[count.index]
+}
+
+data "ibm_is_subnet" "compute_subnets" {
+  count = local.public_endpoints ? 0 : length(var.compute_subnet_id_list)
+
+  identifier = var.compute_subnet_id_list[count.index]
 }
 
 ############################################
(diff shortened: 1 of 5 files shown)


rvanderp3:
/lgtm
/hold
/retest-required

placing a hold until required tests pass.  feel free to remove the hold when ready.

jianlinliu:
In customer vpc case, users are allowed to use any CIDR,  not have to be `10.0.0.0/8`, right?

jeffnowicki:
> In customer vpc case, users are allowed to use any CIDR, not have to be `10.0.0.0/8`. Shall we hard code it here?

@jianlinliu for 4.13 this may be our easiest approach to unblock. We can note a limitation (treating this as tactical) and then looking into a more strategic/proper fix to allow any CIDR.

cjschaef:
@jianlinliu In Private or Disconnected mode, the customer could use any machine to perform the installation. However, if they wish to see the installation complete successfully (even though the installer is in monitor mode once bootstrapping completes), they would need access to the Private network the cluster's apiserver is hosted on, meaning they likely are running the installer on the Private network, which means they should be using a machine on the 10.0.0.0/8 network.

If we'd rather open ingress up to any IP (`0.0.0.0/0`), we can do that, but I think we'd need to also add a FloatingIP back to the bootstrap node in Private mode, which I didn't think seemed appropriate. Otherwise, at this time, I think the primary way to access the bootstrap node would be from within the VPC Private network, hence the `10.0.0.0/8`.

But, if there are other methods, we can surely update what I have now to accommodate them too, or add additional SecurityGroupRules as well.

jianlinliu:
> In Private or Disconnected mode

Just like what I explained in slack, the issue does not only happened in Private or Disconnected mode, also in BYON (customer vpc) mode. The Private or Disconnected mode has a prerequisite, bring your own network firstly, then run the installer inside the customer vpc to complete the install process. 

Sometimes, especially when troubleshooting, or collect cluster log, we do not only need to access boostrap, also master, even compute nodes.  But this PR,  seem like only bootstrap is accessible, I am looking forwards some solution the machine where installer is running on can access any node from this cluster via internal IP.

>  but I think we'd need to also add a FloatingIP back to the bootstrap node in Private mode, which I didn't think seemed appropriate.

Totally agree. 

My last comment is about if it is appropriate to use a hard code to limit the BYON vpc's CIDR to `10.0.0.0/8`. On the other  hand, I agree with @jeffnowicki 's proposal. We can note the hardcode CIDR as a limitation.

cjschaef:
Per the discussions, I am putting up new changes to restrict SSH access to only machine within the same VPC subnets.

Once tested, I will update the PR with those changes.

/hold

cjschaef:
Made changes to SecurityGroupRules to use Subnet CIDR's for SSH access to bootstrap, control plane, and compute nodes.
Attempts from a machine within the VPC and on one of the Control Plane or Compute Subnets to SSH to one of the cluster nodes (Control Plane and Compute and bootstrap when it is running) should now work.

Subnet CIDR's are used to keep some form of isolation from clusters sharing the same VPC, but on different Subnets.

This PR should now be ready for review once again, I hope
/unhold

rvanderp3 (commented):
/lgtm

rvanderp3:
/test e2e-ibmcloud-ovn 

e2e-ibmcloud-ovn appears to be an infra flake

PR #6946, opened 2023-03-07
OCPBUGS-5201: Error message should be more user-friendly when "openshift-install create cluster" fails due to unexpected VMware configuration

vr4manta:
Enhanced installer for vsphere to verify ComputeResources existed in ComputeClusterResource before progressing to network selection.

### DIFF
--- a/pkg/asset/installconfig/vsphere/validation.go
+++ b/pkg/asset/installconfig/vsphere/validation.go
@@ -673,12 +673,11 @@ func validateClusterComputeResourceHasHosts(cluster string, hosts []*object.Host
 // users know what to address.  This only returns an error if unable to verify or if no ComputeResources
 // are found.  This method is used mostly from interactive installation to make sure ClusterComputeResource
 // has HostSystems to select a network.
-func validateClusterComputeResourceHasHostsWithLookup(ctx context.Context, datacenter string, cluster string, finder Finder) error {
+func validateClusterComputeResourceHasHostsWithLookup(ctx context.Context, cluster string, finder Finder) error {
 	ctx, cancel := context.WithTimeout(ctx, 60*time.Second)
 	defer cancel()
 
-	path := fmt.Sprintf("/%s/host/%s", datacenter, cluster)
-	ccr, err := finder.ClusterComputeResource(ctx, path)
+	ccr, err := finder.ClusterComputeResource(ctx, cluster)
 	if err != nil {
 		return err
 	}
--- a/pkg/asset/installconfig/vsphere/validation.go
+++ b/pkg/asset/installconfig/vsphere/validation.go
@@ -13,6 +13,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/vmware/govmomi/find"
+	"github.com/vmware/govmomi/object"
 	vapitags "github.com/vmware/govmomi/vapi/tags"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/mo"
@@ -173,16 +174,10 @@ func validateFailureDomain(validationCtx *validationContext, failureDomain *vsph
 	}
 	computeClusterName := clusterPathParts[2]
 
-	allErrs = append(allErrs, validateESXiVersion(validationCtx, computeCluster, vsphereField, topologyField.Child("computeCluster"))...)
+	allErrs = append(allErrs, validateClusterComputeResources(validationCtx, computeCluster, vsphereField, topologyField.Child("computeCluster"), checkComputeClusterPrivileges, checkTags)...)
 	allErrs = append(allErrs, validateVcenterPrivileges(validationCtx, topologyField.Child("server"))...)
-	allErrs = append(allErrs, computeClusterExists(validationCtx, computeCluster, topologyField.Child("computeCluster"), checkComputeClusterPrivileges, checkTags)...)
 	allErrs = append(allErrs, datacenterExists(validationCtx, failureDomain.Topology.Datacenter, topologyField.Child("datacenter"), checkDatacenterPrivileges)...)
 	allErrs = append(allErrs, datastoreExists(validationCtx, failureDomain.Topology.Datacenter, failureDomain.Topology.Datastore, topologyField.Child("datastore"))...)
-
-	err := validateClusterComputeResourceHasHosts(context.TODO(), failureDomain.Topology.Datacenter, computeClusterName, validationCtx.Finder)
-	if err != nil {
-		allErrs = append(allErrs, field.InternalError(topologyField.Child("computeCluster"), err))
-	}
 
 	for _, network := range failureDomain.Topology.Networks {
 		allErrs = append(allErrs, validateNetwork(validationCtx, failureDomain.Topology.Datacenter, computeClusterName, network, topologyField)...)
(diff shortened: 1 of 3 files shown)


pkg/asset/installconfig/vsphere/vsphere.go
```diff
@@ -170,6 +178,36 @@ func getClients() (*vCenterClient, error) {
 	}, nil
 }
 
+// checkComputeResource checks the ClusterComputeResource to see if there are any attached hosts.  Without
+// hosts, the check for networks will fail.  We want the message to be clear if this is the cause so
+// users know what to address.  This only returns an error if unable to verify or if no ComputeResources
+// are found.
+func checkComputeResource(ctx context.Context, datacenter string, cluster string, finder Finder) error {
```
rvanderp3, line 185: Typically for validation that resources exist and/or are valid, we perform that check in the validation [validation phase](https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/vsphere/validation.go).  We could probably add it to https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/vsphere/validation.go#L347.

rvanderp3 (commented):
Added a few comments, overall it looks good.

pkg/asset/installconfig/vsphere/vsphere.go
```diff
@@ -170,6 +178,36 @@ func getClients() (*vCenterClient, error) {
...
+	ctx, cancel := context.WithTimeout(ctx, 60*time.Second)
+	defer cancel()
+
+	path := fmt.Sprintf("/%s/host/%s", datacenter, cluster)
+	ccr, err := finder.ClusterComputeResource(context.TODO(), path)
```
r4f4, line 190: You just created a child context above, shouldn't you use it here?

pkg/asset/installconfig/vsphere/vsphere.go
```diff
@@ -4,6 +4,7 @@ package vsphere
 import (
 	"context"
 	"fmt"
+	"github.com/vmware/govmomi/vim25/mo"
```
r4f4, line 7: Please use `hack/go-fmt.sh` to fix the import order

pkg/asset/installconfig/vsphere/vsphere.go
```diff
@@ -170,6 +178,36 @@ func getClients() (*vCenterClient, error) {
...
+	ccr, err := finder.ClusterComputeResource(context.TODO(), path)
+	if err != nil {
+		return err
+	}
+
+	var ccrMo mo.ComputeResource
+	logrus.Debugf("CCR: %v", ccr)
+
+	err = ccr.Properties(ctx, ccr.Reference(), []string{"host"}, &ccrMo)
+	logrus.Debugf("Hosts: %v", ccrMo.Host)
```
r4f4, line 199: This and the log above look more like dev logs during testing. If you intend this to appear when customers run the installer with `--log-level debug` then I recommend wording them in a more friendly way.

pkg/asset/installconfig/vsphere/vsphere.go
```diff
@@ -170,6 +178,36 @@ func getClients() (*vCenterClient, error) {
...
+	if err != nil {
+		return err
+	}
+	if len(ccrMo.Host) == 0 {
+		logrus.Errorf("No hosts found in cluster %v. Please make sure Compute Resources are assigned to the cluster before installation.", cluster)
+		return fmt.Errorf("no hosts found in cluster %v", cluster)
```
r4f4, line 205: Why do we need both a message logged and an error here saying basically the same thing?

r4f4 (changes requested):

rvanderp3:
/retitle OCPBUGS-5201: Error message should be more user-friendly when "openshift-install create cluster" fails due to unexpected VMware configuration

pkg/asset/installconfig/vsphere/vsphere.go
vr4manta: Thanks, I 'll fix that.

pkg/asset/installconfig/vsphere/vsphere.go
vr4manta: Ok, I'll have to look into moving it over there.  If we do this, is it still valid to use it as part of the automated install-config creation steps?

pkg/asset/installconfig/vsphere/vsphere.go
```diff
@@ -57,6 +58,13 @@ func Platform() (*vsphere.Platform, error) {
 		return nil, err
 	}
 
+	// Check for CRs under the CCR.  If none, then getNetwork will fail.  Return helpful error message
```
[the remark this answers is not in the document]
vr4manta: Will do thanks.

pkg/asset/installconfig/vsphere/vsphere.go
vr4manta: My original thought was to make sure we had enough info to instruct user when to have this fix.  I didn't want to return the longer string because it makes the overall error really long and unsightly.  I can remove this message.

pkg/asset/installconfig/vsphere/vsphere.go
vr4manta: Good catch.  I'll update those.  Very useful for debugging what is coming back from vsphere.  Intention is more for support when issues occur to make sure we are able to see what vsphere is returning.

rvanderp3 (commented):
overall, lgtm, found a nit in a comment

pkg/asset/installconfig/vsphere/vsphere.go
```diff
@@ -58,9 +57,9 @@ func Platform() (*vsphere.Platform, error) {
 		return nil, err
 	}
 
-	// Check for CRs under the CCR.  If none, then getNetwork will fail.  Return helpful error message
-	// for installer user
-	err = checkComputeResource(ctx, dc, cluster, finder)
+	// Check for ComputeResourcess under the ClusterComputeResource.  If none, then getNetwork
```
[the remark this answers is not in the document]
vr4manta: Good catch.  Fixed.

rvanderp3:
/lgtm
/assign @jcpowermac 
/hold

We just addressed an issue in the IBM vCenter.  I'd atleast like to see an installation complete.  I'll place a hold while the tests rerun.

jcpowermac:
@vr4manta sorry I didn't see this before...

We are already finding the hosts that reside in a cluster:
https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/vsphere/validation.go#L275-L279

Do you think it would be possible to modify that function and just return an error if `len(hosts)` is 0?

jcpowermac:
> @vr4manta sorry I didn't see this before...
> 
> We are already finding the hosts that reside in a cluster: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/vsphere/validation.go#L275-L279
> 
> Do you think it would be possible to modify that function and just return an error if `len(hosts)` is 0?

You could also break out that section that gathers the hosts and still keep the multiple checks

pkg/asset/installconfig/vsphere/validation.go
```diff
@@ -633,3 +638,32 @@ func validateTagAttachment(validationCtx *validationContext, reference vim25type
...
+
+// validateClusterComputeResourceHasHosts checks the ClusterComputeResource to see if there are any attached hosts.  Without
+// hosts, the check for networks will fail.  We want the message to be clear if this is the cause so
+// users know what to address.  This only returns an error if unable to verify or if no ComputeResources
+// are found.
+func validateClusterComputeResourceHasHosts(ctx context.Context, datacenter string, cluster string, finder Finder) error {
+	ctx, cancel := context.WithTimeout(ctx, 60*time.Second)
+	defer cancel()
+
+	path := fmt.Sprintf("/%s/host/%s", datacenter, cluster)
```
jcpowermac, line 650: Cluster in this case is a path already. We shouldn't be making paths

jcpowermac: take a look at this unmerged pr: https://github.com/openshift/installer/pull/6973/files

vr4manta:
> @vr4manta looks good! still need to remove creating the path: https://github.com/openshift/installer/pull/6946/files#diff-b9fefa06ad2cecc56dc1451447e6ecb1ab155af23a0cd5914aa6df90ef2d5901R680

@jcpowermac So that one is needed for the interactive installer.  Would you prefer I move that specific function to the vsphere.go?  https://github.com/openshift/installer/blob/70de7f7bc42e5f2677b3abb0aaa24272ab12f9ef/pkg/asset/installconfig/vsphere/vsphere.go#L62

PR #6947, opened 2023-03-08
OCPBUGS-8658: Pass Capabilites from install-config to cluster

zaneb:


zaneb:
@andfasano mentioned that he approved this but evidently the comment didn't get posted, so...
/approve

PR #6949, opened 2023-03-08
OCPBUGS-8119: CVE-2023-25173: bump containerd 

r4f4:
A bug was found in containerd prior to versions 1.6.18 and 1.5.18 where supplementary groups are not set up properly inside a container.


barbacbd:
Looks like there is a newer version or several but the same patches for the CVE's were applied to all. 
/lgtm

cgwalters:
But surely we're not running the vulnerable code here...openshift/installer isn't creating containers itself...

r4f4:
> But surely we're not running the vulnerable code here...openshift/installer isn't creating containers itself...

Yeah but I figured there was no harm in bumping the dep version.

PR #6951, opened 2023-03-08
terraform: google: bump stringset dependency

r4f4:
bitbucket.org/creachadair/stringset < 0.0.10 depends on https://github.com/creachadair/staticfile which seems to have been deleted once it was not used by stringset anymore.


barbacbd (commented):
/lgtm

PR #6959, opened 2023-03-09
CORS-2525: Azure: remove storage account with bootstrap destroy

AnnaZivkovic:
The work is being done to address the issue of Azure storage accounts being persisted for the life of an OpenShift cluster, which incurs costs and management effort. The goal is to destroy the storage account along with other bootstrap resources during the bootstrap destroy process. Additionally, the default creation of two Azure storage accounts has raised security issues, which the work aims to address by restricting network access using virtual network rules.

Using azurerm_storage_account_network_rules on the storage account did not work as it blocks other resources from using the storage accounts.

### DIFF
--- a/data/data/azure/bootstrap/outputs.tf
+++ b/data/data/azure/bootstrap/outputs.tf
@@ -1,3 +1,4 @@
 output "bootstrap_ip" {
   value = var.azure_private ? azurerm_network_interface.bootstrap.private_ip_address : azurerm_public_ip.bootstrap_public_ip_v4[0].ip_address
 }
+
--- a/data/data/azure/bootstrap/variables.tf
+++ b/data/data/azure/bootstrap/variables.tf
@@ -59,12 +59,11 @@ variable "outbound_udr" {
   description = <<EOF
 This determined whether User defined routing will be used for egress to
 Internet.
-When false, Standard LB will be used for egress to the Internet.
 
+When false, Standard LB will be used for egress to the Internet.
 This is required because terraform cannot calculate counts during plan phase
 completely and therefore the `vnet/public-lb.tf`
 conditional need to be recreated. See
 https://github.com/hashicorp/terraform/issues/12570
 EOF
 }
-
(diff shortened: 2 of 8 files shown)


r4f4:
I don't think we can do this. If you remove the storage account where the VHD was uploaded, then when autoscaling machinesets they won't be able to find the VM image. Until we move to Marketplace images by default we cannot get rid of the RHCOS VHD (therefore the storage account that contains it). This could even break a normal deploy if it can happen the bootstrap is destroyed before some of the compute nodes are provisioned.

r4f4:
> I don't think we can do this. If you remove the storage account where the VHD was uploaded, then when autoscaling machinesets they won't be able to find the VM image. Until we move to Marketplace images by default we cannot get rid of the RHCOS VHD (therefore the storage account that contains it). This could even break a normal deploy if it can happen the bootstrap is destroyed before some of the compute nodes are provisioned.

Maybe the VHD blob itself can be destroyed, but the Gallery Image Versions  (gen1 and gen2) must live for the duration of the cluster.

AnnaZivkovic:
> 

I agree with your assessment that scaling issues may arise. Ill change to making a pre-vnet stage, which will eliminate the need to interact with the azurerm_shared_image_gallery and azurerm_shared_image_version resources. 
It appears that the creation of an image version involves copying the source VHD, without leaving any residual linkage between the image version and its source

r4f4:
> I agree with your assessment that scaling issues may arise. Ill change to making a pre-vnet stage, which will eliminate the need to interact with the azurerm_shared_image_gallery and azurerm_shared_image_version resources. It appears that the creation of an image version involves copying the source VHD, without leaving any residual linkage between the image version and its source

Right. When you specify a `blob_uri` to the `azurerm_shared_image_version` resource, you have to also specify the `storage_account_id` where that blob exists. That means you need to keep the following order of events:
1. The storage account is created
2. The rhcos vhd is uploaded to the storage account
3. shared_image_gallery, shared_image and shared_image_version are created.
4. Bootstrap node is privisioned
5. Storage account is deleted (perhaps together with the bootstrap resources)

PR #6960, opened 2023-03-10
OCPBUGS-9949: Verify output file exists when `oc image extract` is run

bfournie:
The command `oc image extract --path <file>:<dir> ...` will not return an error even if the file is not extracted. Add a check to ensure that the file was extracted from the release image. If not, then fail the `create image` command.

### DIFF
--- a/pkg/asset/agent/image/baseiso.go
+++ b/pkg/asset/agent/image/baseiso.go
@@ -3,6 +3,7 @@ package image
 import (
 	"context"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
 	"time"
@@ -127,6 +128,11 @@ func (i *BaseIso) Generate(dependencies asset.Parents) error {
 			logrus.Debugf("Extracted base ISO image %s from release payload", baseIsoFileName)
 			i.File = &asset.File{Filename: baseIsoFileName}
 			return nil
+		}
+
+		if errors.Is(err, fs.ErrNotExist) {
+			// if image extract failed to extract the iso that architecture may be missing from release image
+			return fmt.Errorf("base ISO for %s not found in release image, check release image architecture", archName)
 		}
 		if !errors.Is(err, &exec.Error{}) { // Already warned about missing oc binary
 			logrus.Warning("Failed to extract base ISO from release payload - check registry configuration")
--- a/pkg/asset/agent/image/oc.go
+++ b/pkg/asset/agent/image/oc.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path"
@@ -15,6 +16,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/thedevsaddam/retry"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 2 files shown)


pkg/asset/agent/image/oc.go
```diff
@@ -188,8 +188,13 @@ func (r *release) extractFileFromImage(image, file, cacheDir string) (string, er
 	if err != nil {
 		return "", err
 	}
-	// set path
 	path := filepath.Join(cacheDir, path.Base(file))
+	// Make sure file exists
```
zaneb, line 192: Could the file have existed before running the extraction? We may need to check and delete it first.

pkg/asset/agent/image/baseiso.go
```diff
@@ -129,6 +130,10 @@ func (i *BaseIso) Generate(dependencies asset.Parents) error {
 			return nil
 		}
 		if !errors.Is(err, &exec.Error{}) { // Already warned about missing oc binary
+			if strings.Contains(err.Error(), "no such file or directory") {
```
zaneb, line 133: This looks pretty risky, we don't even control that error message.

Wouldn't `os.IsNotExist(err)` (or I guess more properly `errors.Is(err, fs.ErrNotExist)`) actually do what we need?

pkg/asset/agent/image/baseiso.go
```diff
@@ -129,6 +130,10 @@ func (i *BaseIso) Generate(dependencies asset.Parents) error {
...
+				// if image extract failed to extract the iso that architecture may be missing from release image
+				return errors.Wrapf(err, "failed to extract a base ISO image matching %s", archName)
```
zaneb, line 135: We can be more explicit about the problem here I think. e.g. "Base ISO for %s not found in release image. Check release image architecture."

pkg/asset/agent/image/oc.go
andfasano: IINW the check if there was already the iso in the cache is performed before reaching it, so it should be ok

bfournie: GetFileFromCache() https://github.com/openshift/installer/blob/master/pkg/asset/agent/image/oc.go#L111 does already check for the file.

pkg/asset/agent/image/baseiso.go
bfournie: Yep, that's better. Changed.

pkg/asset/agent/image/baseiso.go
bfournie: Good point, updated.

bfournie:
Resulting output now...
```
ERROR failed to write asset (Agent Installer ISO) to disk: cannot generate ISO image due to configuration errors 
FATAL failed to fetch Agent Installer ISO: failed to fetch dependency of "Agent Installer ISO": failed to generate asset "BaseIso Image": Base ISO for aarch64 not found in release image. Check release image architecture. 
```

pkg/asset/agent/image/baseiso.go
```diff
@@ -129,6 +130,10 @@ func (i *BaseIso) Generate(dependencies asset.Parents) error {
 			return nil
 		}
 		if !errors.Is(err, &exec.Error{}) { // Already warned about missing oc binary
+			if errors.Is(err, fs.ErrNotExist) {
```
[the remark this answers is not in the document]
bfournie: Yes. Updated.

pkg/asset/agent/image/oc.go
zaneb: But if `verifyCacheFile()` returns false then we will proceed to this code despite the existence of an old file.
And in fact it will always return false when we have the wrong release payload, because it calls this function to get the SHA sum, which will return an error: https://github.com/openshift/installer/blob/master/pkg/asset/agent/image/oc.go#L260-L264

bfournie: That's true, we could get a false positive with new existence check.
Add a change to remove the file in case it exists before extraction.
Tested with both failing and passing cases.

pkg/asset/agent/image/oc.go
```diff
@@ -182,14 +182,21 @@ func (r *release) extractFileFromImage(image, file, cacheDir string) (string, er
 	} else {
 		cmd = fmt.Sprintf(templateImageExtract, file, cacheDir, image)
 	}
+	// Remove file if it exists
+	path := filepath.Join(cacheDir, path.Base(file))
+	os.Remove(path)
```
zaneb, line 187: We should definitely ignore `fs.ErrNotExist`, but it would be better if we handled other errors. e.g. a permissions problem here could result in us using an old ISO if the release payload has only the wrong arch.

PR #6961, opened 2023-03-10
Remove m6i* family from 4.10-4.11 tested instance type list

yunjiang29:
Due to EBS restrictions on AWS (https://issues.redhat.com/browse/OCPBUGS-3964), the embedded lists for documentation are different for 4.10-4.11 and 4.12+, so split them into two lists.


yunjiang29:
@patrickdillon we need to back-port this PR to 4.10, do we need a new BZ to attach to this PR?

mtulio:
> @yunjiang29 will the CSI driver fix backported to <4.12?

I read the answer on Jira, and commented there.

PR #6965, opened 2023-03-10
OCPBUGS-9982: bootstrap-pivot: skip pivot in SCOS Live ISO

vrutkovs:
If bootstrap is running in a Live ISO of SCOS `rpm-ostree pivot` will fail as /usr is not writable. Live ISO is used in assisted-installer flow and it doesn't need to pivot machine-os-content during bootstrap (it would write SCOS image on disk and run pivot via machine-config-daemon-pivot service). 

This change unblocks OKD Stream isntall using assisted-installer

### DIFF
--- a/data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
+++ b/data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
@@ -1,6 +1,16 @@
 {{if .IsOKD -}}
 #!/usr/bin/env bash
 set -euo pipefail
+
+# Exit early if pivot is attempted on SCOS Live ISO
+{{if .IsSCOS -}}
+source /etc/os-release
+if [[ ! $(touch /usr/.test) ]] && [[ ${ID} =~ ^(centos)$ ]]; then
+  touch /opt/openshift/.pivot-done
+  exit 0
+fi
+{{end -}}
+
 # Rebase to OKD's OSTree container image.
 # This is required in OKD as the node is first provisioned with plain Fedora CoreOS.
 


patrickdillon:
retesting, hopefully we'll see some scos tests not get rate limited

vrutkovs:
Needs unit test updates

vrutkovs:
okd-scos-e2e-aws-ovn is looking good now
/test e2e-aws-ovn

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
```diff
@@ -1,6 +1,16 @@
 {{if .IsOKD -}}
 #!/usr/bin/env bash
 set -euo pipefail
+
+# Exit early if pivot is attempted on SCOS Live ISO
+{{if .IsSCOS -}}
+source /etc/os-release
+if [[ ! $(touch /usr/.test) ]] && [[ ${ID} =~ ^(centos)$ ]]; then
```
cgwalters, line 8: It's not going to work on any live ISO right?  As we move towards potentially doing https://github.com/openshift/os/issues/799 we'll also want to have this change for OCP/RHEL.

That said, I think the Live ISO case is going to have interesting issues here.  What we probably want to support is a case where we do dynamic overlays of things like kubelet and other tools, without a reboot.

vrutkovs: Right, that script would fail on any LiveISO, but we run boostrap-pivot on OKD only. The CentOS guard here is because OKD SCOS can start via:
* FCOS in IPI
* LiveISO SCOS in Assisted Installer

cgwalters: Right, but I'm saying we're at least potentially going to make the same change for OCP to ship disk images that are pure RHEL content.  (enhancement etc. to be written; no action required now, just noting)

vrutkovs: Ah, yes, in that case we'll need something similar - and it would be a great opportunity to rework bootstrap-pivot entirely

PR #6967, opened 2023-03-13
OCPBUGS-7015: vsphere, nutanix survey: relax vip in machine cidr

patrickdillon:
This commit removes validation to ensure the ingress and API vips are in the default machine network cidr. This validation was added to all on-prem platforms. Installs have worked without this validation since the release of openshift 4 and the new requirements seems to be causing user issues. So we are relaxing the validation to be consistent with previous releases and because there is not a clear need for the validation.

### DIFF
--- a/pkg/asset/installconfig/nutanix/nutanix.go
+++ b/pkg/asset/installconfig/nutanix/nutanix.go
@@ -14,11 +14,8 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
-	"github.com/openshift/installer/pkg/types"
-	"github.com/openshift/installer/pkg/types/defaults"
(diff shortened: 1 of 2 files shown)


patrickdillon:
@rvanderp3 added this validation in [ocpbugs-881](https://issues.redhat.com/browse/OCPBUGS-881). The motivation for that fix was in order to make the TUI consisitent with the install config validation. #6915 removed the install config validation for the VIP in machine cidr check. This PR is a follow up. My original PR missed removing this validation from the survey.

patrickdillon:
Rich's PR also added validation for ovirt. My PR is only removing validation for vsphere and nutanix based on the slack conversation. If we think ovirt should be relaxed as well, please LMK.

patrickdillon:

Single e2e test failure, looks like a flake. Certainly not related to this pr.

PR #6972, opened 2023-03-14
OCPBUGS-5931: destroy/gcp: Disk names are filtered using kubernetes name format

barbacbd:
** Disks are now ALSO filtered using the kubernetes name format which currently can be found https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/util/util.go function GenerateVolumeName(). The max name length comes out to be 22 characters as the kubernetes volume name or pv name appears to take 40 characters. Search for disks matching the similar name structure in the project on GCP.

Removed labels from returned types

** Altering var to const for the variables as they are not altered during execution


r4f4 (commented):
/lgtm

PR #6973, opened 2023-03-14
OCPBUGS-2130: vSphere - finding networks use full path cluster

jcpowermac:
In previous releases we didn't use full paths for
cluster. Since we are now using full paths generating the path is incorrect. This PR removes that.


rvanderp3:
/lgtm
/approve
/hold

placing a hold pending CI

jinyunma:
pre-merge test is passed, details can be found from comments in bug.
/label qe-approved

PR #6974, opened 2023-03-14
azure: session: support additional authentication methods

r4f4:
By using the DefaultAzureCredential available in azidentity we can support all methods officially supported by Azure to authenticate [1]

[1] https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/azidentity#defaultazurecredential

/wip


r4f4:
Unfortunately we're limited to the auth methods supported by the terraform provider.

PR #6975, opened 2023-03-15
OCPBUGS-8449: pkg/asset/installconfig: set subscriptionID

jhixson74:
Set the subscriptionID in the Azure defaultMachinePlatform.

https://issues.redhat.com/browse/OCPBUGS-8449

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -258,6 +258,7 @@ spec:
                               required:
                               - name
                               - resourceGroup
+                              - subscriptionId
                               type: object
                             diskSizeGB:
                               description: DiskSizeGB defines the size of disk in
@@ -1164,6 +1165,7 @@ spec:
                             required:
                             - name
                             - resourceGroup
+                            - subscriptionId
                             type: object
                           diskSizeGB:
                             description: DiskSizeGB defines the size of disk in GB.
@@ -2373,6 +2375,7 @@ spec:
                             required:
                             - name
                             - resourceGroup
+                            - subscriptionId
                             type: object
                           diskSizeGB:
                             description: DiskSizeGB defines the size of disk in GB.
--- a/pkg/types/azure/disk.go
+++ b/pkg/types/azure/disk.go
@@ -36,7 +36,7 @@ type OSDisk struct {
 type DiskEncryptionSet struct {
 	// SubscriptionID defines the Azure subscription the disk encryption
 	// set is in.
-	SubscriptionID string `json:"subscriptionId,omitempty"`
+	SubscriptionID string `json:"subscriptionId"`
 	// ResourceGroup defines the Azure resource group used by the disk
 	// encryption set.
 	ResourceGroup string `json:"resourceGroup"`


pkg/asset/installconfig/installconfig.go
```diff
@@ -157,6 +157,13 @@ func (a *InstallConfig) finish(filename string) error {
 	}
 	if a.Config.Azure != nil {
 		a.Azure = icazure.NewMetadata(a.Config.Azure.CloudName, a.Config.Azure.ARMEndpoint)
+		if a.Config.Platform.Azure.DefaultMachinePlatform.OSDisk.DiskEncryptionSet != nil && a.Config.Platform.Azure.DefaultMachinePlatform.OSDisk.DiskEncryptionSet.SubscriptionID == "" {
+			session, err := a.Azure.Session()
+			if err != nil {
+				return errors.Wrap(err, "failed to fetch session")
+			}
+			a.Config.Platform.Azure.DefaultMachinePlatform.OSDisk.DiskEncryptionSet.SubscriptionID = session.Credentials.SubscriptionID
+		}
```
r4f4, line 166: So we are loading the value correctly when using the DiskEncryptionSet [for master](https://github.com/openshift/installer/blob/master/pkg/asset/machines/master.go#L354-L357) and [for worker](https://github.com/openshift/installer/blob/master/pkg/asset/machines/worker.go#L439-L442) but the problem was not doing that for the validation?

pkg/asset/installconfig/installconfig.go
```diff
@@ -157,6 +157,13 @@ func (a *InstallConfig) finish(filename string) error {
 	}
 	if a.Config.Azure != nil {
 		a.Azure = icazure.NewMetadata(a.Config.Azure.CloudName, a.Config.Azure.ARMEndpoint)
+		if a.Config.Platform.Azure.DefaultMachinePlatform.OSDisk.DiskEncryptionSet != nil && a.Config.Platform.Azure.DefaultMachinePlatform.OSDisk.DiskEncryptionSet.SubscriptionID == "" {
```
r4f4, line 160: This will fail if there is no `defaultMachinePlatform` specified in the install-config.yaml (or `DefaultMachinePlatform.OSDisk`):
```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x3486976]

goroutine 1 [running]:
github.com/openshift/installer/pkg/asset/installconfig.(*InstallConfig).finish(0xc00073fdc0, {0x7be442d, 0x13})
	~/prjs/upstream/openshift/installer/pkg/asset/installconfig/installconfig.go:160 +0x3b6
github.com/openshift/installer/pkg/asset/installconfig.(*InstallConfig).Load(0xc00073fdc0, {0x1f640c60?, 0xc00073cbc0?})
	~/prjs/upstream/openshift/installer/pkg/asset/installconfig/installconfig.go:120 +0x5b
github.com/openshift/installer/pkg/asset/store.(*storeImpl).load(0xc0008a7b60, {0x1f648930, 0xc00073eb40}, {0x7ba4cf0, 0x2})
	~/prjs/upstream/openshift/installer/pkg/asset/store/store.go:263 +0x2b2
github.com/openshift/installer/pkg/asset/store.(*storeImpl).load(0xc0008a7b60, {0x7f54f0c6f718, 0x241959d0}, {0x0, 0x0})
	~/prjs/upstream/openshift/installer/pkg/asset/store/store.go:246 +0xc05
github.com/openshift/installer/pkg/asset/store.(*storeImpl).fetch(0xc0008a7b60, {0x7f54f0c6f718, 0x241959d0}, {0x0, 0x0})
	~/prjs/upstream/openshift/installer/pkg/asset/store/store.go:200 +0x1a9
github.com/openshift/installer/pkg/asset/store.(*storeImpl).Fetch(0x7ffed232f883?, {0x7f54f0c6f718, 0x241959d0}, {0x241768c0, 0x6, 0x6})
	~/prjs/upstream/openshift/installer/pkg/asset/store/store.go:76 +0x48
main.runTargetCmd.func1({0x7ffed232f883, 0x2})
	~/prjs/upstream/openshift/installer/cmd/openshift-install/create.go:260 +0x125
main.runTargetCmd.func2(0x2417dee0?, {0xc00073e740?, 0x4?, 0x4?})
	~/prjs/upstream/openshift/installer/cmd/openshift-install/create.go:290 +0xe7
github.com/spf13/cobra.(*Command).execute(0x2417dee0, {0xc00073e700, 0x4, 0x4})
	~/prjs/upstream/openshift/installer/vendor/github.com/spf13/cobra/command.go:920 +0x847
github.com/spf13/cobra.(*Command).ExecuteC(0xc0004bac00)
	~/prjs/upstream/openshift/installer/vendor/github.com/spf13/cobra/command.go:1040 +0x3bd
github.com/spf13/cobra.(*Command).Execute(...)
	~/prjs/upstream/openshift/installer/vendor/github.com/spf13/cobra/command.go:968
main.installerMain()
	~/prjs/upstream/openshift/installer/cmd/openshift-install/main.go:61 +0x2b0
main.main()
	~/prjs/upstream/openshift/installer/cmd/openshift-install/main.go:38 +0xff
```

r4f4:
What about doing this as part of the `GetDiskEncryptionSet` client call?
```
// GetDiskEncryptionSet retrieves the specified disk encryption set.
func (c *Client) GetDiskEncryptionSet(ctx context.Context, subscriptionID, groupName, diskEncryptionSetName string) (*azenc.DiskEncryptionSet, error) {
        if subscriptionID == "" {
                subscriptionID = c.ssn.Credentials.SubscriptionID
        }
	client := azenc.NewDiskEncryptionSetsClientWithBaseURI(c.ssn.Environment.ResourceManagerEndpoint, subscriptionID)
	client.Authorizer = c.ssn.Authorizer
	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
	defer cancel()

	diskEncryptionSet, err := client.Get(ctx, groupName, diskEncryptionSetName)
	if err != nil {
		return nil, errors.Wrap(err, "failed to get disk encryption set")
	}

	return &diskEncryptionSet, nil
}
```

pkg/asset/installconfig/installconfig.go
jhixson74: Fixed.

pkg/asset/installconfig/installconfig.go
jhixson74: Yes.

jhixson74:
> What about doing this as part of the `GetDiskEncryptionSet` client call?
> 
> ```
> // GetDiskEncryptionSet retrieves the specified disk encryption set.
> func (c *Client) GetDiskEncryptionSet(ctx context.Context, subscriptionID, groupName, diskEncryptionSetName string) (*azenc.DiskEncryptionSet, error) {
>         if subscriptionID == "" {
>                 subscriptionID = c.ssn.Credentials.SubscriptionID
>         }
> 	client := azenc.NewDiskEncryptionSetsClientWithBaseURI(c.ssn.Environment.ResourceManagerEndpoint, subscriptionID)
> 	client.Authorizer = c.ssn.Authorizer
> 	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
> 	defer cancel()
> 
> 	diskEncryptionSet, err := client.Get(ctx, groupName, diskEncryptionSetName)
> 	if err != nil {
> 		return nil, errors.Wrap(err, "failed to get disk encryption set")
> 	}
> 
> 	return &diskEncryptionSet, nil
> }
> ```

I don't have a problem with this approach, however we use the diskEncryptionSet SubscriptionID elsewhere, not just to get the diskEncryptionSet.

pkg/asset/installconfig/installconfig.go
```diff
@@ -157,6 +157,14 @@ func (a *InstallConfig) finish(filename string) error {
...
 	if a.Config.Azure != nil {
 		a.Azure = icazure.NewMetadata(a.Config.Azure.CloudName, a.Config.Azure.ARMEndpoint)
+		defaultMachinePlatform := a.Config.Platform.Azure.DefaultMachinePlatform
+		if defaultMachinePlatform != nil && defaultMachinePlatform.OSDisk.DiskEncryptionSet != nil && defaultMachinePlatform.OSDisk.DiskEncryptionSet.SubscriptionID == "" {
+			session, err := a.Azure.Session()
+			if err != nil {
+				return errors.Wrap(err, "failed to fetch session")
+			}
+			a.Config.Platform.Azure.DefaultMachinePlatform.OSDisk.DiskEncryptionSet.SubscriptionID = session.Credentials.SubscriptionID
+		}
```
patrickdillon, line 167: This code block looks good to me, but this doesn't look like the right place to set this value: adding this code here makes it less readable and it would only get worse if we build on this pattern. 

If this is the best place in the code to have this value set, then I think we just need one layer of abstraction on top of this (such as a general function for setting defaults that other platforms could use as well) . 

I couldn't find any existing code that sets defaults in this manner, so I think it would be good to build it out. If that's too much for this bug fix, we could either set the field to required or put this in place temporarily.

r4f4: +1 to the idea of filling the install-config.yaml with defaults as its own step in the install process. Once filled we can use the install-config for validation and for creating the resources knowing that both operations are seeing exactly the same config.

PR #6977, opened 2023-03-15
OCPBUGS-10313: fix agent tui showing up multiple times

andfasano:


andfasano:
cc @bfournie

bfournie:
cc @zaneb

data/data/agent/systemd/units/agent-interactive-console.service
```diff
@@ -2,6 +2,8 @@
 Description=Get interactive user configuration at boot
 After=network-pre.target NetworkManager.service pre-network-manager-config.service selinux.service
 Before=getty@tty1.service network.target network.service agent.service node-zero.service NetworkManager-wait-online.service
+ConditionPathExists=/usr/local/bin/agent-tui
+ConditionPathExists=!/etc/assisted/node0
 
 [Service]
 Type=oneshot
```
aleskandro, line 9: Hi, I just hit this while testing the agent on arm64. Shouldn't `RemainAfterExit=yes` be also considered here?

bfournie: `RemainAfterExit=yes` is used in example here https://www.redhat.com/sysadmin/systemd-oneshot-service

aleskandro: > I don't know the exact cause of it appearing twice,

in my case, continuously restarting without explicit errors after the users quit or after waiting for the non-action-needed timeout. I couldn't debug further as neither ssh nor TTYs were available at that stage.

andfasano: I suspect that the problem was related to the path and the burst limits/freq, even though the builds didn't show it up until one day ago, so there could some additional enviromental condition changed.
In any case, given the consolidated approach we adopted to make available the agent-tui binary, the systemd path is not effectively required anymore. The additional condition on node0 should prevent any further restarts, thus the `RemainAfterExit=yes` could be worth to be used, adding it.
Note that [last job failed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_installer/6977/pull-ci-openshift-installer-master-e2e-agent-compact-ipv4/1636084652940726272/artifacts/e2e-agent-compact-ipv4/baremetalds-devscripts-setup/artifacts/root/dev-scripts/logs/06_agent_create_cluster-2023-03-15-205617.log) but due a later problem in the installation, the initial boot phase was ok

data/data/agent/systemd/units/agent-interactive-console.service
```diff
@@ -2,6 +2,8 @@
 Description=Get interactive user configuration at boot
 After=network-pre.target NetworkManager.service pre-network-manager-config.service selinux.service
 Before=getty@tty1.service network.target network.service agent.service node-zero.service NetworkManager-wait-online.service
+ConditionPathExists=/usr/local/bin/agent-tui
```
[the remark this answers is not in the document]
andfasano: The dependencies were not changed, just dropped the path for a direct unit condition(s)

andfasano:
> So we got a report on [internal chat](https://gist.github.com/cgwalters/536737496d01a8bf3c641b844f0cd75c) that this PR was failing for space reasons that may be related to [coreos/fedora-coreos-config#2188 (comment)](https://github.com/coreos/fedora-coreos-config/pull/2188#issuecomment-1398406088)
> 
> At the time I didn't do an analysis of space usage for assisted installer; almost certainly should have CC'd the team on that PR. Internal report said the scenario here was 16GiB of RAM - how big are the containers being pulled?
> 
> If we're seeing divergence between 4.14 and 4.13 it may actually be dependent on when the bootimages landed in openshift/installer. I suspect what happened here is we regenerated 4.13 bootimages a bit later after that fix had already landed.
> 
> If that's the case, we can do another PR to update 4.14 bootimages now.

Opened the same PR on the release 4.13 branch (https://github.com/openshift/installer/pull/6989) to verify through the CI what we experienced locally

cgwalters:
OK here's how to do this, plug the version number of the 4.14 metal version and we get https://url.corp.redhat.com/9026029 which if you look at the `meta.json` :arrow_right: `coreos-assembler.container-config-git` :arrow_right: https://github.com/openshift/os/commit/6e4377e7ad16cb4afff5840262e6d67c279c30ae

Which is unfortunately right *before* https://github.com/openshift/os/commit/1168220fcd021fde1ee60b4c80d3ac93c9078de9 which has the fix.

So we need to respin bootimages for 4.14 and 4.13.  If this needs to go into 4.13 then we need a tracker bug I think.

eranco74:
> So we need to respin bootimages for 4.14 and 4.13. If this needs to go into 4.13 then we need a tracker bug I think.

We have [this](https://issues.redhat.com/browse/MGMT-13937) issue in the MGMT project, we can move it or open another bug, I'm just not sure to which project.

andfasano:
> > So we need to respin bootimages for 4.14 and 4.13. If this needs to go into 4.13 then we need a tracker bug I think.
> 
> https://issues.redhat.com/browse/OCPBUGS-8036

That bug affects just version 4.13, which seems to be fixed and working (even though the bug is not yet marked as closed?). I guess  that for 4.14 we need a new one by cloning it and amending the affect version

andfasano:
/retest-required 

Since Retesting since https://github.com/openshift-metal3/dev-scripts/pull/1519 landed

andfasano:
> > > So we need to respin bootimages for 4.14 and 4.13. If this needs to go into 4.13 then we need a tracker bug I think.
> > 
> > 
> > https://issues.redhat.com/browse/OCPBUGS-8036
> 
> That bug affects just version 4.13, which seems to be fixed and working (even though the bug is not yet marked as closed?). I guess that for 4.14 we need a new one by cloning it and amending the affect version

Just double-checked and it looks like the 4.13 is affected as well (the CI job was passing due using an outdated version still stored in the `machine-os-images` (but the installer [rhcos.json](https://github.com/openshift/installer/blob/f11c21e5e98c0f19516a0c0b13b8350a8f636b36/data/data/coreos/rhcos.json#L11) is still effectively using version `413.92.202303011445-0` as previously pointed by @cgwalters ) and needs to be updated

andfasano:
The failure on `ci/prow/e2e-aws-ovn` does not seem to be related to the current patch

PR #6981, opened 2023-03-16
OCPBUGS-6727: Nutanix Hostname of the VM is not set when using DHCP network config

yanhua121:
Fix OCPBUGS-6727: Nutanix Hostname of the VM is not set when using DHCP network config

### DIFF
--- a/data/data/nutanix/cluster/main.tf
+++ b/data/data/nutanix/cluster/main.tf
@@ -38,14 +38,38 @@ resource "nutanix_image" "rhcos" {
   }
 }
 
+data "ignition_file" "hostname" {
+  count = var.master_count
+  mode  = "420" // 0644
+  path  = "/etc/hostname"
+
+  content {
+    content = <<EOF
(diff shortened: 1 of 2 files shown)


rvanderp3:
@yanhua121 the tf-fmt test needs some attention.  My question is, while not necessarily related to this PR, how does the machine API handle this for the compute nodes?

yanhua121:
> @yanhua121 the tf-fmt test needs some attention. My question is, while not necessarily related to this PR, how does the machine API handle this for the compute nodes?

@rvanderp3 Addressed the tf-fmt and the test now passed. 
The mapi-nutanix-controller change for the worker nodes is here: https://github.com/openshift/machine-api-provider-nutanix/pull/43/files#diff-335b80e6abb1a41cdc4ae4413b95cfd8740dadb1d00753e0e00e11e94f948dbb. Basically, the fix is to add the "/etc/hostname" file with content of the Machine name to the ignition config data, before passing it to create the worker node VM.

PR #6991, opened 2023-03-19
OCPBUGS-7966: Do not remove host default configuration unless network configuration is provided for it

paul-maidment:
Presently, in the PreNetworkConfig script, we are removing the default network manager configuration for a host in every case. This should be altered so that the default configuration is deleted only if custom nmstate configuration has been provided for the host.

### DIFF
--- a/pkg/asset/agent/manifests/network-scripts.go
+++ b/pkg/asset/agent/manifests/network-scripts.go
@@ -60,10 +60,6 @@ else
   ETC_NETWORK_MANAGER="/etc/NetworkManager/system-connections"
 fi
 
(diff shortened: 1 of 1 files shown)


bfournie:
@paul-maidment Thanks for making this change in the installer. I had created a bug for this https://issues.redhat.com/browse/OCPBUGS-7966. Could you change the title to preface it with "OCPBUGS-7966" so it gets tracked properly?

/lgtm

patrickdillon:
/retitle OCPBUGS-7966: Do not remove host default configuration unless network configuration is provided for it

bfournie:
Thanks @patrickdillon 

/jira refresh
/cherry-pick release-4.13

PR #6996, opened 2023-03-20
OCPBUGS-14076: PowerVS: Remove ClusterOSImage

hamzy:
Remove changing the image name for a MachineSet if ClusterOSImage is set

Terraform has already created an image bucket based on OPENSHIFT_INSTALL_OS_IMAGE_OVERRIDE
for us. So worker nodes should not use OPENSHIFT_INSTALL_OS_IMAGE_OVERRIDE directly and
instead use the image bucket.

### DIFF
--- a/pkg/asset/machines/powervs/machinesets.go
+++ b/pkg/asset/machines/powervs/machinesets.go
@@ -26,9 +26,6 @@ func MachineSets(clusterID string, config *types.InstallConfig, pool *types.Mach
 	mpool := pool.Platform.PowerVS
 	var network string
 	image := fmt.Sprintf("rhcos-%s", clusterID)
-	if platform.ClusterOSImage != "" {
-		image = platform.ClusterOSImage
-	}
 	if platform.PVSNetworkName != "" {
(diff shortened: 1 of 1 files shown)


mjturek:
@mkumatag We mentioned this patch in Slack but interested in your opinion here. I think this is the right way to use the override.

r4f4:
@hamzy it would be good to have more context in the commit message about why this change is needed.

hamzy:
@r4f4 how about if I add the following to the commit message?  Will that be sufficient?

"Terraform has already created an image bucket based on OPENSHIFT_INSTALL_OS_IMAGE_OVERRIDE
for us.  So worker nodes should not use OPENSHIFT_INSTALL_OS_IMAGE_OVERRIDE directly and
instead use the image bucket."

r4f4:
@hamzy Yes, that's good. I understand now what this PR is trying to do.

hamzy:
@r4f4 Cool, thanks!  I wanted to verify via comments before doing the push effort. :)

PR #6997, opened 2023-03-20
OCPBUGS-10585: rhcos: Update to 413.92.202303190222-0

cgwalters:
This is to fix OCPBUGS-10585

### DIFF
--- a/data/data/coreos/rhcos.json
+++ b/data/data/coreos/rhcos.json
@@ -1,95 +1,95 @@
 {
   "stream": "rhcos-4.13",
   "metadata": {
-    "last-modified": "2023-03-01T22:32:48Z",
-    "generator": "plume cosa2stream 0.15.0+113-gf35e2c8a7-dirty"
+    "last-modified": "2023-03-21T23:49:18Z",
+    "generator": "plume cosa2stream a1eedef"
   },
   "architectures": {
     "aarch64": {
       "artifacts": {
         "aws": {
-          "release": "413.92.202303011445-0",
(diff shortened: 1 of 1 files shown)


mike-nguyen:
We'll need to run the cloud-replicate job if we want the cloud images replicated to all the regions

zaneb:
> Note: It'd be cleaner to land a 4.14 bump first, but we're still in the process of branching and I'd like to just consolidate things there for master/4.14 so we both fix this bug and correctly switch to a 4.14 version, etc.

The Jira bot will complain about that, but I can manually set the labels if necessary, given that there is no danger of failing bumping the version at any time during 4.14 development.
I would note that CI for the agent installer is broken right now on 4.14 though, so we'd prefer not to delay too long.

patrickdillon:
/retest

Most cloud platforms were borked due to the dns migration issues, which should be fixed now.

/approve
/label backport-risk-assessed

cgwalters:
I'm tentatively planning to take a look at this at the same time as we do a final bootimage respin, which will happen sometime soon.

mike-nguyen:
The cloud replication job is pending. I will update here when it is completed (usually about one more hour)

mike-nguyen:
`cloud-replication` job completed.  I don't think I can push to this PR so I have a branch in my repo with the changes if needed: https://github.com/mike-nguyen/installer/tree/413_92_bib

cgwalters:
I pulled in your changes, and also rebased :surfer: for good measure.

PR #6999, opened 2023-03-21
OCPBUGS#10306:[vSphere]vSphere Upi installation failed due to VMs for master and worker node creation failed.

WenXinWei:
Terraform will not create VMs for master and worker for upi vsphere when unset var.control_plane_ip_addresses and var.compute_ip_addresses. When users are using IPAM (as before) to reserve IPs instead of setting static IPs directly into var.control_plane_ip_addresses and var.compute_ip_addresses, Based on upstream code #1 and #2. The count of master and worker is always 0, then terraform will not create any VMs for master and worker nodes. If we changed code as below, it works in IPAM case as before.  
control_plane_fqdns = [for idx in range(length(var.control_plane_ip_addresses)) : "control-plane-${idx}.${var.cluster_domain}"]  
compute_fqdns = [for idx in range(length(var.compute_ip_addresses)) : "compute-${idx}.${var.cluster_domain}"] ==>>
control_plane_fqdns = [for idx in range(var.control_plane_count) : "control-plane-${idx}.${var.cluster_domain}"]
compute_fqdns = [for idx in range(var.compute_count) : "compute-${idx}.${var.cluster_domain}"]


WenXinWei:
also cc  @rvanderp3  for review. thanks

r4f4:
/retitle OCPBUGS-10306: [vSphere] Upi installation failed due to VMs for master and worker node creation failed

r4f4:
@WenXinWei the terraform configs look broken for the `e2e-vsphere-upi-zones` job:
```
 2023-03-27 07:28:54+00:00 - terraform apply...
Error: Error in function call
  on main.tf line 163, in module "compute_a_records":
 163:    records = zipmap(local.compute_fqdns, module.ipam_compute.ip_addresses)
    ├────────────────
    │ local.compute_fqdns is tuple with 3 elements
    │ module.ipam_compute.ip_addresses is list of string with 4 elements
Call to function "zipmap" failed: number of keys (3) does not match number of
values (4). 
```

jcpowermac:
Why do we still need to use IPAM? We should be using the static range in each segment like all of other CI jobs.
IPAM is another piece of infrastructure we should not be managing still.

WenXinWei:
> Why do we still need to use IPAM? We should be using the static range in each segment like all of other CI jobs. IPAM is another piece of infrastructure we should not be managing still.

We still have some jobs , The manual installation needs to be executed in qe-segment/discon-segment. In this case, the ci method cannot be reused, because multiple clusters may need to be run at the same time, So the ipam is still required

WenXinWei:
@jcpowermac @rvanderp3  Could you please help to merge this pr first? now only step 'e2e-vsphere-upi-zones' failed,  the issue  can refer to  [comments](https://github.com/openshift/release/pull/38241#issuecomment-1517477217) but in the merged PR [38241](https://github.com/openshift/release/pull/38241), I missed to assign MASTER_REPLICA/WORKER_REPLICA  to control_plane_count/compute_count.  if i create a new pr to update, it will blocked by this pr(control_plane_count and compute_count not updated yet), so i think this pr need to be approved first.

jcpowermac:
I am still not fond of the use of IPAM and an alternate method of IP allcation should be used.

/approve
/lgtm

WenXinWei:
> @WenXinWei this broke zones

yes, because i have another change need to be merged. after [pr](https://github.com/openshift/release/pull/39916)  merged , it should be ok.

PR #7001, opened 2023-03-21
openstack: Remove version info, update lb FIP -> API FIP

stephenfin:
Our indicated supported version was incorrect. Rather than having to remember to update it for each new OSP version, simply remove this snippet.

The LB FIP is now called the API FIP.


pierreprinetti:
/approve
/lgtm
/skip

nit: Consider replacing instances of the literal "Apps FIP" with `ingressFloatingIP` consistently with [the install-config property](https://github.com/openshift/installer/blob/master/docs/user/openstack/customization.md#cluster-scoped-properties)
/hold
Feel free to remove the hold if you don't want to make this change.

EmilienM:
Yes please use ingress instead of apps

PR #7002, opened 2023-03-21
gcp: add confidential compute support for boostrap TF

mresvanis:
This change adds the [confidential compute configuration options](https://github.com/openshift/installer/pull/6799) for the GCP bootstrap machine as well.

The confidential compute configuration options for the bootstrap instance are defined by the respective control-plane instance options.

The scheduling and confidential_instance_config options are defined as dynamic blocks and they will be applied when the user specifies the respective config options. This allows us to respect the cloud provider defaults, when the user does not provide any values. At the same time, it does not require code updates when the cloud provider defaults change.

Feature link: https://issues.redhat.com/browse/OCPBU-142

/cc @eranco74

### DIFF
--- a/data/data/gcp/bootstrap/main.tf
+++ b/data/data/gcp/bootstrap/main.tf
@@ -113,6 +113,27 @@ resource "google_compute_instance" "bootstrap" {
     network_ip = local.public_endpoints ? null : google_compute_address.bootstrap.address
   }
 
+  dynamic "shielded_instance_config" {
+    for_each = var.gcp_master_secure_boot != "" ? [1] : []
+    content {
+      enable_secure_boot = var.gcp_master_secure_boot == "Enabled"
+    }
+  }
+
+  dynamic "confidential_instance_config" {
+    for_each = var.gcp_master_confidential_compute != "" ? [1] : []
+    content {
+      enable_confidential_compute = var.gcp_master_confidential_compute == "Enabled"
+    }
+  }
+
+  dynamic "scheduling" {
+    for_each = var.gcp_master_on_host_maintenance != "" ? [1] : []
+    content {
+      on_host_maintenance = var.gcp_master_on_host_maintenance
+    }
+  }
+
   metadata = {
     user-data = data.ignition_config.redirect.rendered
   }


mresvanis:
> You should probably update the machine_type as well. 

If I understand correctly, the bootstrap machine type is set to the control-plane's machine type [by default](https://github.com/openshift/installer/blob/master/pkg/tfvars/gcp/gcp.go#L77). Should I try to make this clear in any way, like with a comment?

> Consider adding a comment explaining that the bootstrap will use the same confidential compute configuration as the master nodes

Thank you for the suggestion! I've added a comment in the commit message, do you think it would be better to be somewhere else?

eranco74:
> If I understand correctly, the bootstrap machine type is set to the control-plane's machine type [by default](https://github.com/openshift/installer/blob/master/pkg/tfvars/gcp/gcp.go#L77). Should I try to make this clear in any way, like with a comment?
You're correct, I think it's all good.
> Thank you for the suggestion! I've added a comment in the commit message, do you think it would be better to be somewhere else?
The commit message is great.

mresvanis:
Hi Patrick, I apologise for involving you directly in this PR. But I thought since you're already familiar with the change introduced because of https://github.com/openshift/installer/pull/6799, you would be an appropriate reviewer.

/cc @patrickdillon

r4f4:
In the e2e-gcp-secureboot job, the control plane nodes seem to be failing after reboot with: 
```
error: ../../grub-core/kern/efi/sb.c:182:bad shim signature.

error: ../../grub-core/kern/efi/sb.c:182:bad shim signature.

error: ../../grub-core/loader/i386/efi/linux.c:258:you need to load the kernel

first.

error: ../../grub-core/loader/i386/efi/linux.c:258:you need to load the kernel

first.
```

mresvanis:
> In the e2e-gcp-secureboot job, the control plane nodes seem to be failing after reboot with:
> 
> ```
> error: ../../grub-core/kern/efi/sb.c:182:bad shim signature.
> 
> error: ../../grub-core/kern/efi/sb.c:182:bad shim signature.
> 
> error: ../../grub-core/loader/i386/efi/linux.c:258:you need to load the kernel
> 
> first.
> 
> error: ../../grub-core/loader/i386/efi/linux.c:258:you need to load the kernel
> 
> first.
> ```

hi @r4f4, I'm not quite sure why this happens, as I was able to run the respective QEMU image successfully with

```sh
$ cosa run --qemu-firmware uefi-secure --qemu-image ./rhcos-414.92.202303241545-0-qemu.x86_64.qcow2
+cosa:16> podman run --rm -ti --security-opt 'label=disable' --privileged '--uidmap=1000:0:1' '--uidmap=0:1:1000' --uidmap 1001:1001:64536 -v /home/mresvani/Downloads:/srv/ --device /dev/kvm -
-device /dev/fuse --tmpfs /tmp -v /var/tmp:/var/tmp --name cosa coreos-assembler:latest run --qemu-firmware uefi-secure --qemu-image ./rhcos-414.92.202303241545-0-qemu.x86_64.qcow2
Red Hat Enterprise Linux CoreOS 414.92.202303241545-0
  Part of OpenShift 4.14, RHCOS is a Kubernetes native operating system
  managed by the Machine Config Operator (`clusteroperator/machine-config`).

WARNING: Direct SSH access to machines is not recommended; instead,
make configuration changes via `machineconfig` objects:
  https://docs.openshift.com/container-platform/4.14/architecture/architecture-rhcos.html

---
Last login: Wed Mar 29 13:18:37 2023
[core@cosa-devsh ~]$ rpm-ostree status
State: idle
Deployments:
● f224455ca12f8480eab736feb0304d01b83879c448144971a93b4a163fd1f292
                  Version: 414.92.202303241545-0 (2023-03-24T15:48:37Z)
[core@cosa-devsh ~]$ dmesg |grep -i secure
[    0.000000] secureboot: Secure boot enabled
[    0.000000] Kernel is locked down from EFI Secure Boot mode; see man kernel_lockdown.7
[    0.013459] secureboot: Secure boot enabled
[    1.555726] integrity: Loaded X.509 cert 'Red Hat Secure Boot CA 5: cc6fa5e72868ba494e939bbd680b9144769a9f8f'
[    3.717701] systemd[1]: CoreOS Secex Ignition Config Decryptor was skipped because of an unmet condition check (ConditionPathExists=/run/coreos/secure-execution).
[core@cosa-devsh ~]$ rpm -q kernel
kernel-5.14.0-284.4.1.el9_2.x86_64
```
This issue should be also affecting other open GCP PRs, but I'm guessing that as soon as [this](https://github.com/openshift/installer/pull/7038) is merged we won't see such issues again. 

Would you propose to wait until this is resolved in order to move forward with this PR?

r4f4:
> This issue should be also affecting other open GCP PRs, but I'm guessing that as soon as [this](https://github.com/openshift/installer/pull/7038) is merged we won't see such issues again.
> 
> Would you propose to wait until this is resolved in order to move forward with this PR?

I think `e2e-gcp-secureboot` is the only CI job exercising the functionality introduced in this PR. I think it would be good if we could at least se a cluster up in that job before merging this PR. Is it a problem if we wait for #7038 to merge first? It already has the needed labels so should merge soon.

mresvanis:
> Is it a problem if we wait for #7038 to merge first? It already has the needed labels so should merge soon.

Not at all, I agree. Let's wait and see this job green before moving forward :+1:

r4f4:
/approve
e2e-gcp-secureboot succeeded in the install step and the few e2e failures shouldn't be affected by the bootstrap node.

r4f4:
Let's also make sure that the !secureboot workflow is not affected:
/test e2e-gcp-ovn

r4f4:
e2e-gcp-ovn also looks good. Install succeeded and the e2e failures are not related to the PR changes. :+1:

mresvanis:
Hi @patrickdillon, would you maybe have some time to check out this PR (eranco had already lgtm-ed, but we had to force push)?

r4f4:
/retitle OCPBUGS-15855: gcp: add confidential compute support for boostrap TF

PR #6848, opened 2023-02-13
Azure: use managed storage account for boot logs

patrickdillon:
Update installer-created VMs to use managed-storage accounts rather than custom storage accounts, created by the installer. We're doing this in order to have consistency with the control-plane machinesets and to decrease our dependence on the installer-created storage account, which we would like to remove at bootstrap destroy.

PR #7012, opened 2023-03-22
OCPBUGS-10478: gather: azure: fix collecting VM serial logs

r4f4:
With the change to use managed storage accounts for boot diagnostics [1], we were getting empty filenames for the boot diagnostics files:
```
INFO Pulling VM console logs
DEBUG attemping to download
DEBUG attemping to download
DEBUG attemping to download
DEBUG attemping to download
DEBUG attemping to download
DEBUG attemping to download
DEBUG attemping to download
DEBUG attemping to download
INFO Failed to gather VM console logs: unable to download file: /c/serial-log-bundle-20230320085802
```

After adding some more debug information, we've found out the blob URIs are just null:
```
DEBUG appending boot diags for vm rdossant-03200931-5rld5-bootstrap: &{<nil> <nil>}
DEBUG appending boot diags for vm rdossant-03200931-5rld5-master-0: &{<nil> <nil>}
DEBUG appending boot diags for vm rdossant-03200931-5rld5-master-1: &{<nil> <nil>}
DEBUG appending boot diags for vm rdossant-03200931-5rld5-master-2: &{<nil> <nil>}
```

That is because boot diagnostics data are not present in the InstanceView of the VirtualMachine but actually need to be retrieved separately when using managed storage. And support for doing that was added to compute profile version >= 2020-09-01. Once we have the diagnostics data URIs, they are generated with a SAS token and can be be downloaded directly.

So this commit introduces the following changes:
 * Update the compute profile used to 2020-09-01
 * If no diagnostics data is found in the InstanceView, then check the BootDiagnosticsData object
 * Check the file URI to figure out how to download it: if it starts with a owned storage account host, then use shared keys; otherwise, just download it with a simple request
 * Skip shared keys not matching the blob's storage account

[1] https://github.com/openshift/installer/pull/6848


r4f4:
/uncc @jhixson74 @MayXuQQ @AnnaZivkovic 
/hold
We'll probably prefer to go with the new SDK (https://github.com/openshift/installer/pull/6992) but I'll keep this up for now for reference.

r4f4:
/close
in favor of #6992

PR #7018, opened 2023-03-23
OCPBUGS-10728 add project filter to gcp usage api requests

kaovilai:
… to usages

timeseries returned by 
serviceruntime.googleapis.com/quota/allocation/usage can return other project_ids than intended.
 
With the filters added in the PR, only usages in the intended project will be used to match with limits of the same project.
 
Previously, it is possible to by chance have limits from intended project, and usage from an incorrect project accessible by the account that can lead to incorrect quota calculation (including potentially negative quota) that prevents an otherwise capable account from launching a cluster.

Negative quota can occurs if selected project has lower limits than other projects and the quota code happen to match it up with usage from a higher limit projects.

Example occurrence at breakpoint reference
https://github.com/kaovilai/installer/blob/b20f28a6dd09f949676ab15f1caafb7124f28612/pkg/quota/gcp/usage.go#L36
watch variables.
image

Co-authored-by: @r4f4

### DIFF
--- a/pkg/quota/gcp/usage.go
+++ b/pkg/quota/gcp/usage.go
@@ -18,7 +18,7 @@ import (
 func loadUsage(ctx context.Context, client *monitoring.MetricClient, project string) ([]record, error) {
 	req := &monitoringpb.ListTimeSeriesRequest{
 		Name:   fmt.Sprintf("projects/%s", project),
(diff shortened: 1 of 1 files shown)


kaovilai:
Without filtering by project, the query used previously may return more than one results per service.
image

PR #7023, opened 2023-03-24
SPLAT-847: Generate machine manifests for compute nodes (Static IP IPI)

vr4manta:
Enhanced installer to create worker machine manifests when static IPs are configured.

### DIFF
--- a/pkg/asset/machines/vsphere/machines.go
+++ b/pkg/asset/machines/vsphere/machines.go
@@ -3,14 +3,16 @@ package vsphere
 
 import (
 	"fmt"
-	machineapi "github.com/openshift/api/machine/v1beta1"
-	"github.com/openshift/installer/pkg/types"
-	"github.com/openshift/installer/pkg/types/vsphere"
+
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


pkg/asset/machines/vsphere/machines.go
```diff
@@ -88,6 +105,26 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
...
+				provider.Network.Devices[0].IPAddrs = networkDevice.IPAddrs
+				provider.Network.Devices[0].Nameservers = networkDevice.Nameservers
+				provider.Network.Devices[0].Gateway4 = networkDevice.Gateway4
+				provider.Network.Devices[0].Gateway6 = networkDevice.Gateway6
+			}
+			hosts = append(hosts[:index], hosts[index+1:]...)
+			break
+		}
+	}
+	return hosts
```
rvanderp3, line 125: Should we be returning an error if there isn't an IP address assigned?  I ran in to a situation where a given zone didn't have an associated host and no address was assigned.  Another way we could handle this would be an update to the platform validation to catch this situation before the installer even gets to the machine asset generation.

vr4manta: Ideally this type of check should be done during validation in my opinion.  I can update SPLAT-828 to include that validation step.  Looking at SPLAT-828, we do check that network.IPAddrs != 0 for static IP assignment.  This may be a defect related to dhcp assignment of IPs.  Would OCPBUGS-5201 catch this?

PR #7025, opened 2023-03-27
OCPBUGS-10845: Use 100 GB as minimum disk size in validations

bfournie:
The minimum disk size is currently set to 120 GB but for consistency with documentation and with assisted it should be 100 GB.

dev-scripts change openshift-metal3/dev-scripts#1521


bfournie:
dev-scripts change https://github.com/openshift-metal3/dev-scripts/pull/1521

PR #7033, opened 2023-03-28
GCP: add europe-west12 region to the survey as supported region

makentenza:
Add the europe-west12 GCP region that was not previously listed as a supported region for the installer to use.

[CORS-2593](https://issues.redhat.com/browse/CORS-2593)


r4f4:
/retitle OCPBUGS-11038: GCP: add europe-west12 region to the survey as supported region

PR #7036, opened 2023-03-28
OCPBUGS-11039: remove container-runtime flag from kubelet config

rphillips:
removes the container-runtime=remote flag to correlate with:

```
CHANGELOG/CHANGELOG-1.27.md
478:- Kubelet: remove deprecated flag `--container-runtime` 
```

ref: https://github.com/kubernetes/kubernetes/pull/114017

### DIFF
--- a/data/data/bootstrap/files/usr/local/bin/kubelet.sh.template
+++ b/data/data/bootstrap/files/usr/local/bin/kubelet.sh.template
@@ -6,7 +6,6 @@
 /usr/bin/hyperkube \
   kubelet \
     --anonymous-auth=false \
-    --container-runtime=remote \
     --container-runtime-endpoint=/var/run/crio/crio.sock \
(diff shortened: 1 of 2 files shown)


data/data/bootstrap/files/usr/local/bin/kubelet.sh.template
```diff
@@ -6,7 +6,6 @@
 /usr/bin/hyperkube \
   kubelet \
     --anonymous-auth=false \
-    --container-runtime=remote \
```
bertinatto, line 9: Isn't the line right below related to this one? Shouldn't it be removed too?

bertinatto: `user/customization.md` has this entry as well, might be worth removing too.

rphillips:
flake

/test ci/prow/okd-images

PR #7041, opened 2023-03-29
rhcos: Bump to 413.92.202303281804-0

cgwalters:
This is the 4.13 version of https://github.com/openshift/installer/pull/7038


sdodson:
Reviewed all of the jobs, most completed the install and those that failed appear to be flakes on infrastructure.

PR #7042, opened 2023-03-29
[azure] update tested instance type on 4.13

jinyunma:
Docs update:
1. remove `standardDCSFamily` due to unable to verify it with insufficient quota
2. add two new families `standardDLDSv5Family` `standardDLSv5Family` which are tested passed on 4.13 nightly build.

### DIFF
--- a/docs/user/azure/tested_instance_types_x86_64.md
+++ b/docs/user/azure/tested_instance_types_x86_64.md
@@ -3,11 +3,12 @@
 * `standardDASv4Family`
 * `standardDASv5Family`
 * `standardDCSv3Family`
-* `standardDCSFamily`
 * `standardDCSv2Family`
 * `standardDDCSv3Family`
 * `standardDDSv4Family`
 * `standardDDSv5Family`
+* `standardDLDSv5Family`
+* `standardDLSv5Family`
 * `standardDSFamily`
 * `standardDSv2Family`
 * `standardDSv2PromoFamily`


jinyunma:
cc: @jianlinliu @gpei @MayXuQQ to review

gpei:
@jinyunma for `standardDCSFamily` type, could we apply for more quotas to unblock the testing

jianlinliu:
Can we add one instance type in our coverage - `Edv5`, because https://issues.redhat.com/browse/ARO-1350 mentioned the vm type.

jinyunma:
>  for `standardDCSFamily` type, could we apply for more quotas to unblock the testing

I did request when testing on 4.12, here is the response from Azure engineer on that time.

> Due to both capacity constraints in our DC Series regions and the availability of our new VM series that is in GA, we are recommending all new DC Series deployments, to deploy using the newer DCsv2 Series VMs available in UK South, UK West, US East, US Southcentral, US West, US West 2, Europe North, Europe West, Asia Southeast, Canada Central, and Canada East. The DCsv2 Series is in GA and allows you to configure new Confidential Computing VMs with a larger selection of sizes, higher enclave page cache memory, and a greater level of support

Current cpu for `standardDCSFamily` type is 8 in our environment, and it was okay to provision one worker nodes with this type when testing 4.12, but always failed this time when checking on 4.13.

gpei:
@jinyunma thanks for the info, I see `standardDCSv2Family` was covered then dropping `standardDCSFamily` is ok for me

jinyunma:
> Can we add one instance type in our coverage - `Edv5`, because https://issues.redhat.com/browse/ARO-1350 mentioned the vm type.

Family `Edv5` does not support PremiumIO, installer will get error when using vm size under family `Edv5`
```
$ az vm list-skus -l southcentralus --size Standard_E4d_v5 --query "[].capabilities[?name=='PremiumIO'].value" -otsv
False

$ ./openshift-install create cluster --dir ipi/
INFO Credentials loaded from file "/home/fedora/.azure/osServicePrincipal.json" 
ERROR failed to fetch Metadata: failed to load asset "Install Config": failed to create install config: controlPlane.platform.azure.osDisk.diskType: Invalid value: "Premium_LRS": PremiumIO not supported for instance type Standard_E4d_v5 

```

jianlinliu:
> Family Edv5 does not support PremiumIO, installer will get error when using vm size under family Edv5

Good to know, can we add unsupported vm type in the doc?

jinyunma:
> > Family Edv5 does not support PremiumIO, installer will get error when using vm size under family Edv5
> 
> Good to know, can we add unsupported vm type in the doc?

In [azure installation doc](https://docs.openshift.com/container-platform/4.12/installing/installing_azure/installing-azure-customizations.html#installation-minimum-resource-requirements_installing-azure-customizations) , there is note to describe that `"You are required to use Azure virtual machines with premiumIO set to true."` so I think that vm types which does not support premiumIO are all not supported. 

Here we only list the supported vm types which we have verified.

jianlinliu:
> Here we only list the supported vm types which we have verified.

Got it.

jinyunma:
cc @patrickdillon @r4f4 to review and merge, thanks.

PR #7052, opened 2023-03-30
openstack: Bump CI base image to v4.14

pierreprinetti:
Note: these are rhel-8 images for 4.14.

cf https://github.com/openshift/installer/blob/master/images/installer/Dockerfile.ci


mandre (commented):
/lgtm
/approve

PR #7057, opened 2023-03-31
OCPBUGS-7954: openstack: Only check HTTPS certs on public endpoints

pierreprinetti:
Before this change the validation steps, and the script, assumed that `internal` and `admin` OpenStack endpoints were always reachable. With this change, the manual steps and the script are amended to only check the validity of HTTPS certificates on the `public` endpoints of the OpenStack catalog.

In a separate commit, this PR fixes a syntax issue in the script that could cause it to crash when the OpenStack catalog contained an URL that is not HTTP and not HTTPS.


pierreprinetti:
/skip
This is a docs-only change.

gryf (approved):
/lgtm

PR #7062, opened 2023-04-04
Agent: run shellcheck on start-cluster-installation.sh

zaneb:
Templating in shell scripts makes it impossible for us to run shellcheck
on them. Pass the data we need through environment variables and
template in the systemd unit.


rwsu (commented):
/lgtm

PR #7066, opened 2023-04-04
AGENT-275: Add new agent graph command to output agent internal dependency graph

andfasano:
This patch adds the equivalent installer `graph`  command for the `agent` specific targets

### DIFF
--- a/cmd/openshift-install/graph.go
+++ b/cmd/openshift-install/graph.go
@@ -26,13 +26,15 @@ func newGraphCmd() *cobra.Command {
 		Short: "Outputs the internal dependency graph for installer",
 		Long:  "",
 		Args:  cobra.ExactArgs(0),
-		RunE:  runGraphCmd,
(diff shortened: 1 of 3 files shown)


cmd/openshift-install/graph.go
```diff
@@ -42,7 +44,7 @@ func runGraphCmd(cmd *cobra.Command, args []string) error {
 		string(gographviz.Shape): "box",
 		string(gographviz.Style): "filled",
 	}
-	for _, t := range targets {
+	for _, t := range cmdTargets {
```
jhixson74, line 47: What happens to targets? This appears to throw it out.

zaneb: It's now passed as an argument on l. 30.

zaneb (commented):
/lgtm

It might be worth also generating the graph and adding it to the docs like the main one is in https://github.com/openshift/installer/blob/master/docs/design/assetgeneration.md#dependency-graph

PR #7068, opened 2023-04-05
AGENT-556: Wait for rendezvous host configuration

zaneb:
Block in the node-zero service for the rendezvous host configuration to
exist. In future this will allow the configuration to be added later,
either interactively or by introducing a config image.


rwsu (approved):
/approve

PR #7070, opened 2023-04-05
OCPBUGS-10767: Fix and improve locking session and AWS Metadata access

sadasu:
This fix eliminates the need for mutexSubnets to update subnet
information within AWS metadata. It also updates populateSubnets
to take care of getting VPC and subnets once for the installation
eliminating duplicate code that could be prone to errors.

### DIFF
--- a/pkg/asset/installconfig/aws/metadata.go
+++ b/pkg/asset/installconfig/aws/metadata.go
@@ -26,8 +26,7 @@ type Metadata struct {
 	Subnets  []string                   `json:"subnets,omitempty"`
 	Services []typesaws.ServiceEndpoint `json:"services,omitempty"`
 
-	mutex        sync.Mutex
-	mutexSubnets sync.Mutex
+	mutex sync.Mutex
 }
 
 // NewMetadata initializes a new Metadata object.
@@ -66,10 +65,9 @@ func (m *Metadata) AvailabilityZones(ctx context.Context) ([]string, error) {
 		if err != nil {
 			return nil, err
 		}
-
 		m.availabilityZones, err = availabilityZones(ctx, session, m.Region)
 		if err != nil {
-			return nil, errors.Wrap(err, "creating AWS session")
+			return nil, errors.Wrap(err, "error retrieving Availability Zones")
 		}
 	}
 
@@ -82,9 +80,8 @@ func (m *Metadata) AvailabilityZones(ctx context.Context) ([]string, error) {
 func (m *Metadata) EdgeSubnets(ctx context.Context) (map[string]Subnet, error) {
 	err := m.populateSubnets(ctx)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "error retrieving Edge Subnets")
 	}
-
 	return m.edgeSubnets, nil
 }
 
@@ -94,9 +91,8 @@ func (m *Metadata) EdgeSubnets(ctx context.Context) (map[string]Subnet, error) {
 func (m *Metadata) PrivateSubnets(ctx context.Context) (map[string]Subnet, error) {
 	err := m.populateSubnets(ctx)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "error retrieving Private Subnets")
 	}
-
 	return m.privateSubnets, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/asset/installconfig/aws/metadata.go
```diff
@@ -61,15 +60,15 @@ func (m *Metadata) AvailabilityZones(ctx context.Context) ([]string, error) {
...
-		session, err := m.unlockedSession(ctx)
-		if err != nil {
-			return nil, err
-		}
+	session, err := m.unlockedSession(ctx)
+	if err != nil {
+		return nil, err
+	}
 
+	if len(m.availabilityZones) == 0 {
```
r4f4, line 68: We don't need this change. `session` is not used unless `availabilityZones` hasn't been set yet.

pkg/asset/installconfig/aws/metadata.go
```diff
@@ -161,12 +158,12 @@ func (m *Metadata) InstanceTypes(ctx context.Context) (map[string]InstanceType,
...
-		session, err := m.unlockedSession(ctx)
-		if err != nil {
-			return nil, err
-		}
+	session, err := m.unlockedSession(ctx)
+	if err != nil {
+		return nil, err
+	}
 
+	if len(m.instanceTypes) == 0 {
```
r4f4, line 166: As before, we don't need this change.

pkg/asset/installconfig/aws/metadata.go
```diff
@@ -106,25 +105,26 @@ func (m *Metadata) PrivateSubnets(ctx context.Context) (map[string]Subnet, error
...
 	}
 
 	if len(m.Subnets) == 0 {
 		return errors.New("no subnets configured")
 	}
 
-	m.mutexSubnets.Lock()
-	defer m.mutexSubnets.Unlock()
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
```
r4f4, line 125: I'd rather have this function be dumber and just populate the subnets. Then we can drop the check if the subnets are populated and the mutex locking and move that to the calling functions. For example:
```golang
func (m *Metadata) PublicSubnets(ctx context.Context) (map[string]Subnet, error) {
    m.mutex.Lock()
    defer m.mutex.Unlock()

    if len(m.publicSubnets) == 0 {
	    err := m.populateSubnets(ctx)
	    if err != nil {
		    return nil, errors.Wrap(err, "retrieving Public Subnets")
        }
	}

	return m.publicSubnets, nil
}
```
With that change, all exported functions (InstanceTypes, VPC, *Subnets, AvailabilityZones, Session) will behave and look consistently. I think that's more desirable than a bit of code "duplication".

pkg/asset/installconfig/aws/metadata.go
```diff
@@ -134,22 +134,19 @@ func (m *Metadata) populateSubnets(ctx context.Context) error {
...
 	m.edgeSubnets = sb.Edge
-	return err
+	return nil
 }
 
 // VPC retrieves the VPC ID containing PublicSubnets and PrivateSubnets.
 func (m *Metadata) VPC(ctx context.Context) (string, error) {
-	m.mutex.Lock()
-	defer m.mutex.Unlock()
-
```
r4f4, line 144: I don't think we should get rid of this. We need to make sure we have exclusive access to the object before reading any value, otherwise we can read a value while it's being set by another routine and therefore get inconsistent results.

pkg/asset/installconfig/aws/metadata.go
```diff
@@ -134,22 +134,19 @@ func (m *Metadata) populateSubnets(ctx context.Context) error {
 	m.privateSubnets = sb.Private
 	m.publicSubnets = sb.Public
 	m.edgeSubnets = sb.Edge
-	return err
+	return nil
```
r4f4, line 137: If are going to return `nil` here, then we need to check `err` after calling `subnets`, otherwise we're just ignoring the errors.

r4f4 (changes requested):

pkg/asset/installconfig/aws/metadata.go
sadasu: I had considered that. This would also need an accompanying change (not just with mutexes) within populateSubnets().

r4f4:
@sadasu since you updated the error messages, the unit tests also need to be updated.

pkg/asset/installconfig/aws/metadata.go
```diff
@@ -106,25 +102,32 @@ func (m *Metadata) PrivateSubnets(ctx context.Context) (map[string]Subnet, error
...
+	}
+	return m.vpc, nil
+}
+
 func (m *Metadata) populateSubnets(ctx context.Context) error {
-	if len(m.publicSubnets) > 0 || len(m.privateSubnets) > 0 {
+	m.mutex.Lock()
+	defer m.mutex.Unlock()
+
+	if m.vpc != "" || len(m.privateSubnets) > 0 || len(m.publicSubnets) > 0 {
```
sadasu, line 123: @mtulio adding m.edgeSubnets to this tests causes the unit tests to fail. Looks like you specifically added a test to prevent that. I don't see why we can't check for len(m.edgeSubnets) > 0 here. Some special way EdgeSubnets are handled?

patrickdillon:
/approve

Collapsing the mutexes into a single mutex in the common `populateSubnets` function makes sense to me.

 It would be good to include to a commit message explaining the change according to the contributing guidelines: https://github.com/openshift/installer/blob/master/CONTRIBUTING.md#commit-message-format

pkg/asset/installconfig/aws/metadata.go
mtulio: @sadasu , sorry by missed that message. I am on it validating it to recall the scenario that drove me to create that unit blocking. Ideally, calling `subnets()` at once should populate all `*Subnets`, including `edge`.

I am [looking at the failed unit job on this PR history](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/7070/pull-ci-openshift-installer-master-unit/1644019735538438144#1:build-log.txt%3A78-83), overall we required that edge subnets exist only when public or private have been provided, we should never accept the edge subnets only (as CP and regular workers must exist alongside LZ nodes) - this is a required defined on EP due AWS limitations of network resources (NLB/NGW, etc)

Would you mind sharing the current unit failure to help to dig in? I can see changes after the job I checked.

mtulio: also sharing on slack that I have installed successfully the cluster with Local Zones/edge + PHZ on this version
```
commit 1fced156cc5540a540a20ac208d79812ef72eae2 (HEAD -> pr_7070)
Author: Sandhya Dasu <sdasu@redhat.com>
Date:   Wed Apr 5 12:42:59 2023 -0400
```

install log:
```
$ OPENSHIFT_INSTALL_RELEASE_IMAGE_OVERRIDE="$RELEASE" $INSTALLER version
./openshift-install_pr7070 unreleased-master-7977-g1fced156cc5540a540a20ac208d79812ef72eae2-dirty
built from commit 1fced156cc5540a540a20ac208d79812ef72eae2
release image registry.ci.openshift.org/origin/release:4.13
release architecture amd64
(...)
INFO Time elapsed: 26m46s     
(...)                    
$ oc --kubeconfig $INSTALL_DIR/auth/kubeconfig get machines -n openshift-machine-api
NAME                                         PHASE     TYPE          REGION      ZONE               AGE
ocp-lz14-h9rmj-edge-us-east-1-nyc-1a-x9pkz   Running   c5d.2xlarge   us-east-1   us-east-1-nyc-1a   50m

```

mtulio (commented):
validated with PHZ + LZ subnet
/lgtm

PR #7074, opened 2023-04-06
pkg: rhcos: use Errorf instead of Error

r4f4:
Otherwise the log message looks like
```
 time="2023-04-06T17:03:26Z" level=error msg="could not fetch the rhcos stream data: %wfailed to read embedded CoreOS stream metadata: open data/coreos/rhcos.json: no such file or directory"
 ```

### CODE REVIEW

**Comment by r4f4** [comment]
Fixed the following error:
```
 pkg/rhcos/ami_regions.go:17:3: printf: github.com/sirupsen/logrus.Errorf does not support error-wrapping directive %w (govet)
		logrus.Errorf("could not fetch the rhcos stream data: %w", err)
		^ 
```

PR #7075, opened 2023-04-06
OCPBUGS-11479: Upgrade libnmstate version used

bfournie:
The version of libnmstate in the release image changed from 1.x to 2.x. Use a wildcard match to handle the upgrade which can be used in a generic way with other libraries.

Tested with both libnmstate.so.1.3.3 (currently 4.14 until a CI release passes) and 2.2.9 (4.13)

$ ls -al /home/stack/.cache/agent/files_cache/
total 15592
drwxr-xr-x. 2 stack stack      73 Apr 10 13:40 .
drwxr-xr-x. 4 stack stack      44 Apr  6 09:05 ..
-r-xr-xr-x. 1 stack stack 9284768 Mar 30 16:29 agent-tui
lrwxrwxrwx. 1 stack stack      19 Aug 15  2022 libnmstate.so.1 -> libnmstate.so.1.3.3
-r-xr-xr-x. 1 stack stack 6678840 Aug 15  2022 libnmstate.so.1.3.3

$ ls -al /home/stack/.cache/agent/files_cache
total 16872
drwxr-xr-x. 2 stack stack      73 Apr 10 12:33 .
drwxr-xr-x. 4 stack stack      44 Apr  6 09:05 ..
-r-xr-xr-x. 1 stack stack 9284472 Apr  5 17:40 agent-tui
lrwxrwxrwx. 1 stack stack      19 Apr  5 11:07 libnmstate.so.2 -> libnmstate.so.2.2.9
-r-xr-xr-x. 1 stack stack 7990432 Apr  5 11:08 libnmstate.so.2.2.9

### DIFF
--- a/pkg/asset/agent/image/oc.go
+++ b/pkg/asset/agent/image/oc.go
@@ -8,7 +8,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/fs"
 	"os"
 	"os/exec"
 	"path"
@@ -16,7 +15,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/thedevsaddam/retry"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -184,10 +182,9 @@ func (r *release) extractFileFromImage(image, file, cacheDir string) (string, er
 	} else {
 		cmd = fmt.Sprintf(templateImageExtract, file, cacheDir, image)
 	}
-	// Remove file if it exists
 	path := filepath.Join(cacheDir, path.Base(file))
-	if err := os.Remove(path); err != nil && !errors.Is(err, fs.ErrNotExist) {
-		logrus.Debugf("Could not remove existing baseISO %s", path)
+	// Remove file if it exists
+	if err := removeCacheFile(path); err != nil {
 		return "", err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/asset/agent/image/oc.go
```diff
@@ -288,6 +295,52 @@ func (r *release) verifyCacheFile(image, file, architecture string) (bool, error
 	return false, nil
 }
 
+// Remove any existing files in the cache.
+func removeCacheFile(path string) error {
+	var err error
+	if strings.HasSuffix(path, "*") {
+		err = filepath.Walk(filepath.Dir(path), func(file string, info fs.FileInfo, err error) error {
```
zaneb, line 302: nit: I think we could also use `filepath.Glob(path)` for this, which might save a bit of code.

zaneb (commented):
/approve

pkg/asset/agent/image/oc.go
bfournie: Yeah, using Glob is a bit cleaner, changed this function and getExtractedFileFromWildcard()

pkg/asset/agent/image/oc.go
```diff
@@ -288,6 +293,40 @@ func (r *release) verifyCacheFile(image, file, architecture string) (bool, error
...
+	return nil
+}
+
+func getExtractedFileFromWildcard(path string) (string, error) {
+	matches, err := filepath.Glob(path)
+	if err != nil {
+		return "", err
+	}
+
+	for _, file := range matches {
```
andfasano, line 318: I think in this case that having more than one match could be potentially a problem, so at least a warning could be useful under such condition

bfournie: Sure, added a check for more than one files that aren't symlinks

PR #7080, opened 2023-04-10
Write storage manifests

gnufied:


data/data/bootstrap/files/usr/local/bin/bootkube.sh.template
```diff
@@ -295,6 +297,28 @@ then
...
+    record_service_stage_start "storage-operator-bootstrap"
+	echo "Rendering Storage Operator core manifests..."
+
+	rm --recursive --force storage-operator-bootstrap
+
+	bootkube_podman_run \
+		--name storage-render \
+		--volume "$PWD:/assets:z" \
+		"${CLUSTER_STORAGE_OPERATOR_IMAGE}" \
+		render \
```
dobsonj, line 311: There is no render command for CSO :(

patrickdillon:
@gnufied can you provide some more context behind the motivation for this change?

dobsonj:
We explored this as an option to resolve some rendering issues we had with vsphere migration in 4.13, but it's no longer needed.
/close

PR #7081, opened 2023-04-11
OCPBUGS-11636: AWS - Remove ACLs from s3 ign

patrickdillon:
AWS has recently disabled ACL support by default. See https://aws.amazon.com/blogs/aws/heads-up-amazon-s3-security-changes-are-coming-in-april-of-2023/ for further information.

This commit removes our usage of ACLs. By default, S3 buckets will have public access blocked.

/wip
Investigating whether this is a sufficient solution.

### DIFF
--- a/data/data/aws/bootstrap/main.tf
+++ b/data/data/aws/bootstrap/main.tf
@@ -59,16 +59,10 @@ resource "aws_s3_bucket" "ignition" {
   }
 }
 
-resource "aws_s3_bucket_acl" ignition {
-  bucket = aws_s3_bucket.ignition.id
-  acl    = "private"
-}
-
 resource "aws_s3_object" "ignition" {
   bucket = aws_s3_bucket.ignition.id
   key    = "bootstrap.ign"
   source = var.ignition_bootstrap_file
-  acl    = "private"
 
   server_side_encryption = "AES256"
 


patrickdillon:
/hold

local testing is consistently failing with masters failing to pull ignition 
```
[  341.844136] ignition[768]: GET error: Get "https://api-int.padillon-04111054.devcluster.openshift.com:22623/config/master": dial tcp: lookup api-int.padillon-04111054.devcluster.openshift.com on 10.0.0.2:53: no such host
```

unclear to me how these changes could be affecting masters pulling ignition...

patrickdillon:
> @patrickdillon should we be reverting [65acad2](https://github.com/openshift/installer/commit/65acad22b7f9ee238c2509eb8323b5809456cbf8) (as you are doing in this PR) or adjusting how we apply ACL's? By rolling this out we may go back to having to deal with the errors described in [#5719 (comment)](https://github.com/openshift/installer/pull/5719#issue-1171859775) however this depend if we are using the 3.x or 4.x version of terraform.

This PR is removing ACLs altogether. #5719 needed to change the way ACLs were applied due to a terraform upgrade, but we shouldn't have that issue if we don't use ACLs at all. The new defaults set by AWS will be sufficient.

sdodson:
I've reviewed the CI results and the OKD images are failing on repo errors and I believe they're both unrelated and unlikely to block relevant OCP testing. Lets merge this.

PR #7086, opened 2023-04-11
Allow destroy for C2S isolated (us-iso and us-isob) partitions.

montaguethomas:
Tagging APIs are now supported in the C2S isolated partitions (us-iso and us-isob). We have tested these changes on clusters installed with IPI in both partitions.


montaguethomas:
poke.

montaguethomas:
@patrickdillon would it be possible to cherry pick this improvement to 4.12 and later? or at least to 4.14 and 4.15. We are tracking several fixes making their way into 4.14 and hopefully 4.12 for C2S region support.

r4f4:
/retitle OCPBUGS-11721:  Allow destroy for C2S isolated (us-iso and us-isob) partitions

PR #7087, opened 2023-04-12
OCPBUGS-11658: Handle libnmstate version in integration tests

bfournie:
Use a wildcard to handle the libnmstate version in integration tests as its just been upgraded from 1 to 2.

### DIFF
--- a/cmd/openshift-install/agent_internal_integration_test.go
+++ b/cmd/openshift-install/agent_internal_integration_test.go
@@ -322,7 +322,8 @@ func checkFileFromInitrdImg(isoPath string, fileName string) error {
 
 			// check if the current cpio files match the required ones
 			for _, f := range files {
-				if f == fileName {
+				matched, err := filepath.Match(fileName, f)
+				if matched && err == nil {
 					return nil
 				}
 			}
--- a/cmd/openshift-install/testdata/agent/image/assets/default_assets.txt
+++ b/cmd/openshift-install/testdata/agent/image/assets/default_assets.txt
@@ -6,7 +6,7 @@ exists $WORK/agent.x86_64.iso
 
 ignitionImgContains agent.x86_64.iso config.ign
 initrdImgContains agent.x86_64.iso /agent-files/agent-tui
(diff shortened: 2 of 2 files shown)


cmd/openshift-install/testdata/agent/image/assets/default_assets.txt
```diff
@@ -40,4 +40,4 @@ apiVersion: v1alpha1
 metadata:
   name: ostest
   namespace: cluster0
-rendezvousIP: 192.168.111.20
\ No newline at end of file
+rendezvousIP: 192.168.111.20
```
andfasano, line 43: small typo?

bfournie: I actually didn't make that change so I assume its the editor that removed the newline but I think its correct as these files should not have a newline at the end of file

cmd/openshift-install/agent_internal_integration_test.go
```diff
@@ -322,7 +322,8 @@ func checkFileFromInitrdImg(isoPath string, fileName string) error {
 
 			// check if the current cpio files match the required ones
 			for _, f := range files {
-				if f == fileName {
+				matched, err := filepath.Match(filepath.Base(fileName), filepath.Base(f))
```
andfasano, line 325: It's not clear to me the reason for stripping out the path from the comparison. The location is relevant, so I think it's be better to avoid using filepath.Base.

cmd/openshift-install/testdata/agent/image/assets/default_assets.txt
```diff
@@ -1,12 +1,12 @@
 # Verify that the most relevant assets are properly generated in the ISO
 
-exec openshift-install agent create image --dir $WORK
+exec openshift-install agent create image --log-level debug --dir $WORK
```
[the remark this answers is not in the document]
bfournie: Whoops, that was unintentional as I was using it during debugging. Removed.

PR #7094, opened 2023-04-14
OCPBUGS-11796: azure: skip NSG creation when BYO vnet

r4f4:
In an install where users bring their networks they also bring their own NSGs. However, the installer still creates NSG. In Azure environments using the rule [1] below, users are prohibited from installing cluster, as the apiserver_in rule has the rule set as 0.0.0.0. Having a rule in place where the users could define this before install would allow them to set this connectivity without having the inbound access.

[1] - Rule: Network Security Groups shall not allow rule with 0.0.0.0/Any Source/Destination IP Addresses - Custom Deny

### DIFF
--- a/data/data/azure/vnet/nsg.tf
+++ b/data/data/azure/vnet/nsg.tf
@@ -20,6 +20,7 @@ resource "azurerm_subnet_network_security_group_association" "worker" {
 }
 
 resource "azurerm_network_security_rule" "apiserver_in" {
+  count                       = var.azure_preexisting_network ? 0 : 1
   name                        = "apiserver_in"
   priority                    = 101
   direction                   = "Inbound"


patrickdillon:
e2e-azure-ovn-shared-vpc is the relevant test here and that is passing. 

Let's see if the e2e-azure-ovn failure was a flake

patrickdillon:
This LGTM. 
/approve

@jhixson74 PTAL I know you have a lot of experience in this area

/assign @jhixson74

barbacbd (commented):
Discussing something similar with AWS. If the security group exists don't we want to skip its creation not the creation of the rule? We would still want to create the rule here but add it to the security group.

r4f4:
> Discussing something similar with AWS. If the security group exists don't we want to skip its creation not the creation of the rule? We would still want to create the rule here but add it to the security group.

Replicating here what we discussed in Slack: we cannot skip the creation of the security group because the cloud provider [requires a NSG](https://github.com/openshift/installer/blob/master/pkg/asset/manifests/azure/cloudproviderconfig.go#L49). So if we simply skip creating the NSG we will have an issue (cloud provider will throw an error).

PR #7095, opened 2023-04-14
OCPBUGS-11801: Fix agent-tui libnmstate dependency name

andfasano:
This patch fixes the https://issues.redhat.com/browse/OCPBUGS-11801 bug by:
* removing the hard-coded version when extracting libnmstate
*  copying also the symlinks in the agent cpio archive

Due a recent bump in the repos, the agent-tui (currently shipped in the `agent-installer-node-agent` image) is now comprised of the following files:
```
agent-tui
libnmstate.so.2 -> libnmstate.so.2.2.9 (symlink)
libnmstate.so.2.2.9
```

### DIFF
--- a/pkg/asset/agent/image/agentimage.go
+++ b/pkg/asset/agent/image/agentimage.go
@@ -78,16 +78,19 @@ func (a *AgentImage) fetchAgentTuiFiles(releaseImage string, pullSecret string, 
 	files := []string{}
 
 	for _, srcFile := range agentTuiFilenames {
-		f, err := release.ExtractFile("agent-installer-node-agent", srcFile)
+		extracted, err := release.ExtractFile("agent-installer-node-agent", srcFile)
 		if err != nil {
 			return nil, err
 		}
-		// Make sure it could be executed
-		err = os.Chmod(f, 0o555)
-		if err != nil {
-			return nil, err
+
+		for _, f := range extracted {
+			// Make sure it could be executed
+			err = os.Chmod(f, 0o555)
+			if err != nil {
+				return nil, err
+			}
+			files = append(files, f)
 		}
-		files = append(files, f)
 	}
 
 	return files, nil
--- a/pkg/asset/agent/image/cpio.go
+++ b/pkg/asset/agent/image/cpio.go
@@ -3,6 +3,7 @@ package image
 import (
 	"bytes"
 	"compress/gzip"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
@@ -70,13 +71,7 @@ func (ca *CpioArchive) StorePath(path string) error {
 
 // StoreFile appends to the current archive the specified file.
 func (ca *CpioArchive) StoreFile(filename string, dstPath string) error {
-	f, err := os.Open(filename)
-	if err != nil {
-		return err
-	}
-	defer f.Close()
-
-	fileInfo, err := f.Stat()
+	fileInfo, err := os.Lstat(filename)
 	if err != nil {
 		return err
 	}
@@ -91,9 +86,32 @@ func (ca *CpioArchive) StoreFile(filename string, dstPath string) error {
 		return err
 	}
 
-	_, err = io.Copy(ca.cpioWriter, f)
-	if err != nil {
-		return err
+	fm := fileInfo.Mode()
+	switch {
+	case fm.IsRegular():
+		// Copy the file content
+		f, err := os.Open(filename)
+		if err != nil {
+			return err
+		}
+		defer f.Close()
+
+		_, err = io.Copy(ca.cpioWriter, f)
+		if err != nil {
+			return err
+		}
+	case fm&os.ModeSymlink != 0:
+		// In case of a symbolic link, copy the link text
+		s, err := os.Readlink(filename)
+		if err != nil {
+			return err
+		}
+		_, err = ca.cpioWriter.Write([]byte(s))
+		if err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("unsupported file mode %v for file %s", fm, filename)
 	}
 
 	return nil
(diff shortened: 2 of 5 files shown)


pkg/asset/agent/image/agentimage.go
```diff
@@ -74,7 +75,9 @@ func (a *AgentImage) fetchAgentTuiFiles(releaseImage string, pullSecret string,
 		Config{MaxTries: OcDefaultTries, RetryDelay: OcDefaultRetryDelay},
 		releaseImage, pullSecret, mirrorConfig)
 
-	agentTuiFilenames := []string{"/usr/bin/agent-tui", "/usr/lib64/libnmstate.so.*"}
+	libnmstateFull := "libnmstate.so.2.2.9"
+	libnmstate := "libnmstate.so.2"
+	agentTuiFilenames := []string{"/usr/bin/agent-tui", filepath.Join("/usr/lib64/", libnmstateFull)}
```
bfournie, line 80: I think it can be done in a more generic way and still keep the wildcard at line 77. We should still support libnmstate.1 until the CI builds are updated and also libnmstate.so.2.2.10 is available soon.

For example, perhaps use
libnmstateFull := "libnmstate.so.2.*"
libnmstateOveride := "libnmstate.so.2"

And keep agentTuiFilenames the same at line 77 with the wildcard.
Then at line 94 do a Glob match on libnmstateFull and only override in that case.

andfasano: I'm not convinced that the effort to support libnmstate.1 is worth the game, considering that it's present only in the CI payload for mistake (given that it's  hasn't been yet updated).
The glob on the minor version seems a good idea I'll update it.

bfournie:
I've tested the fix in dev-scripts and it looks good.  I was able to install the latest nightly, see the TUI during boot, and ssh in in bootstrap.

$ ls -al /usr/local/bin/libnm*
-r-xr-xr-x. 1 root root 7990432 Apr 14 14:25 /usr/local/bin/libnmstate.so.2

Note that the agent CI tests will continue to fail until `agent-installer-utils` is updated to use libnmstate.2 in CI.

/approve

andfasano:
/hold 

We decided to update assisted-installer-agent before landing this patch, so that libnmstate.2 will be available also in the CI stream

andfasano:
Requires https://github.com/openshift/assisted-installer-agent/pull/531

pkg/asset/agent/image/agentimage.go
```diff
@@ -87,6 +89,21 @@ func (a *AgentImage) fetchAgentTuiFiles(releaseImage string, pullSecret string,
...
+		if err != nil {
+			return nil, err
+		}
+		if matched {
+			old := f
+			f = filepath.Join(filepath.Dir(f), libnmstateOverride)
+			err = os.Rename(old, f)
+			if err != nil {
+				return nil, err
+			}
```
patrickdillon, line 104: nit: IMHO it is a little easier to read:

```go
		override := filepath.Join(filepath.Dir(f), libnmstateOverride)
		err = os.Rename(f, override)
		if err != nil {
			return nil, err
		}
```
or even

```go
		override := filepath.Join(filepath.Dir(f), libnmstateOverride)
		if err = os.Rename(f, override); err != nil {
			return nil, err
		}
```

pkg/asset/agent/image/agentimage.go
```diff
@@ -87,6 +89,21 @@ func (a *AgentImage) fetchAgentTuiFiles(releaseImage string, pullSecret string,
 		if err != nil {
 			return nil, err
 		}
+
+		// Rename libnmstate shared lib
+		matched, err := filepath.Match(filepath.Base(libnmstateFull), filepath.Base(f))
+		if err != nil {
+			return nil, err
```
patrickdillon, line 96: In general we should consider [wrapping these errors](https://pkg.go.dev/github.com/pkg/errors#Wrapf) to make it easier to discern failure points within the same function

patrickdillon (commented):
The code I here LGTM, but I'm still not as familiar with the agent code as I would like it and it is unclear to me why renaming the file to `libnmstate.so.2` resolves this bug.

rwsu:
> The code I here LGTM, but I'm still not as familiar with the agent code as I would like it and it is unclear to me why renaming the file to `libnmstate.so.2` resolves this bug.

I think it is because agent-tui is expecting to load libnmstate.so.2, while in the release image the file is actually named libnmstate.so.2.2.9. Normally there would be a symlink from libnmstate.so.2 to libnmstate.so.2.2.9 when the nmstate-libs package is installed, but here we are copying the library file from the release image and inserting it into the agent ISO with the "correct" name. Perhaps a note in `fetchAgentTuiFiles` explaining why the file is being renamed would help.

rwsu:
lgtm. I suggest adding a comment to explain the file renaming.

andfasano:
After a discussion with @zaneb, I've switched the implementation so that now also the symbolic links extracted are copied into the agent ISO (while before they were ignored)

pkg/asset/agent/image/cpio.go
```diff
@@ -91,9 +92,26 @@ func (ca *CpioArchive) StoreFile(filename string, dstPath string) error {
 		return err
 	}
 
-	_, err = io.Copy(ca.cpioWriter, f)
-	if err != nil {
-		return err
+	fm := fileInfo.Mode()
+	switch {
+	case fm.IsRegular():
+		// Copy the file content
```
zaneb, line 98: nit: could move the `Open()` call to here.

zaneb (commented):
/lgtm

pkg/asset/agent/image/cpio.go
andfasano: Thanks, good point

pawanpinjarkar:
/lgtm
Please update the description mentioning the details about the fix.

rwsu (approved):

PR #7096, opened 2023-04-14
ic: gcp: validate instances against user-configured zones

r4f4:
We have assumed that instance types are available in all zones, therefore we just checked for the instance in one of them. However that's not true for some instance types like `t2a-standard-*`:

```
$: gcloud compute zones list --filter="region:(asia-southeast1)"
NAME               REGION           STATUS  NEXT_MAINTENANCE  TURNDOWN_DATE
asia-southeast1-b  asia-southeast1  UP
asia-southeast1-a  asia-southeast1  UP
asia-southeast1-c  asia-southeast1  UP

$: gcloud compute machine-types list --filter="zone:(asia-southeast1) AND name=t2a-standard-4"
NAME            ZONE               CPUS  MEMORY_GB  DEPRECATED
t2a-standard-4  asia-southeast1-b  4     16.00
t2a-standard-4  asia-southeast1-c  4     16.00
```

In the above example, the install would fail if zone `asia-southeast1-a` was used.

To fix this issue, this PR introduces the following changes:
1. Validate instance types against all zones, not just the first one in the list;
2. Validate user-configured zones, if any;
3. Validate instance types against user-configured zones. If not set, defaults to all the zones in the project/region.


r4f4:
Using the following install-config excerpt:
```yaml
apiVersion: v1
baseDomain: <domain>.openshift.com
platform:
  gcp:
    projectID: <project>
    region: us-central1
    defaultMachinePlatform:
      type: t2a-standard-4
```
results in
```bash
# ./openshit-install create ignition-configs --log-level debug --dir /c
DEBUG OpenShift Installer unreleased-master-8005-g270b48f47bac4d61fa844ddf0ea0be330d6b4267
DEBUG Built from commit 270b48f47bac4d61fa844ddf0ea0be330d6b4267
[...]
INFO Credentials loaded from file "/root/.gcp/osServiceAccount.json"
ERROR failed to fetch Kubeconfig Admin Client: failed to load asset "Install Config": failed to create install config: platform.gcp.defaultMachinePlatform.type: Invalid value: "t2a-standard-4": googleapi: Error 404: The resource 'projects/openshift-dev-installer/zones/us-central1-c/machineTypes/t2a-standard-4' was not found, notFound
```
because
```bash
$: gcloud compute machine-types list --filter="zone:(us-central1) AND name=t2a-standard-4" | awk '{print $2}'
ZONE
us-central1-a
us-central1-b
us-central1-f
us-central1-d
```
then limiting the zones to `zones: ["us-central1-a", "us-central1-b", "us-central1-d", "us-central1-f"]`
```bash
ERROR failed to fetch Kubeconfig Admin Client: failed to load asset "Install Config": failed to create install config: platform.gcp.defaultMachinePlatform.zones: Invalid value: []string{"us-central1-d"}: zones not found in region
```
and finally `zones: ["us-central1-a", "us-central1-b", "us-central1-f"]` passes validation.

r4f4:
Update: added one more commit to fix a linting issue.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -45,24 +45,31 @@ func Validate(client API, ic *types.InstallConfig) error {
...
-	typeMeta, err := client.GetMachineType(context.TODO(), project, zone, instanceType)
-	if err != nil {
-		if _, ok := err.(*googleapi.Error); ok {
-			return append(allErrs, field.Invalid(fieldPath.Child("type"), instanceType, err.Error()))
+	var typeMeta *compute.MachineType
+	for zone := range zones {
+		var err error
+		typeMeta, err = client.GetMachineType(context.TODO(), project, zone, instanceType)
+		if err != nil {
+			var gErr *googleapi.Error
```
Prashanth684, line 64: shouldn't this be the other way round? i.e, if there is no match for any instance, then return error, but if there is a match, exit out of the loop and continue?

r4f4: The Installer spreads the nodes through the availability zones: https://github.com/openshift/installer/blob/master/pkg/asset/machines/gcp/machines.go#L44-L46 so I find it safer to make sure that the instance is available in all zones. It was happening to me that the Installer was assigning a node to `us-central1-c` which is exactly the one that doesn't contain a `t2a-standard-*` type, even though `-a`, and `-b` do. With your suggestion I would've exited the loop on the match of zone `us-central1-a` but still would've failed during install.

Another possibility would be to replicate the way the machines are assigned zones by sorting the zones and then just checking for `len(nodes)` zones.

Prashanth684: so with the current implementation ,it would mean if you use `t2a-standard`, you'll have to specify the zone in the install-config, else it will fail the validation if it uses the default set of zones?

r4f4:
> don't we also need some logic similar to https://github.com/openshift/installer/blob/master/pkg/asset/machines/aws/instance_types.go#L43? today, we just use all zones and round robin through them for assignment: https://github.com/openshift/installer/blob/master/pkg/asset/machines/gcp/machines.go#L134. but now we need to pass in the filtered set of zones which support the instance types and cycle through that list?

I couldn't find a way to do that but I think I just managed by using aggregatedList + filter. I'll update the PR.

pkg/asset/installconfig/gcp/validation.go
r4f4: Yes. But we might be able to change that with https://github.com/openshift/installer/pull/7096#issuecomment-1511004714

r4f4:
/hold
While I address Prashanth's comments.

r4f4:
/hold cancel
Reworked to include Prashanth's suggestion of only using zones where the instance is available.

r4f4:
Update: fixed an autocomplete typo that would cause the following failure:
```
level=fatal msg=failed to fetch Master Machines: failed to generate asset "Master Machines": failed to fetch availability zones: failed to get zones for instance type: googleapi: Error 400: Invalid field selection MachineType
level=fatal msg=Details:
level=fatal msg=[
level=fatal msg=  {
level=fatal msg=    "@type": "type.googleapis.com/google.rpc.BadRequest",
level=fatal msg=    "fieldViolations": [
level=fatal msg=      {
level=fatal msg=        "description": "Error processing the 'fields' param value 'items/*/MachineType(zone),nextPageToken': Invalid field selection MachineType",
level=fatal msg=        "field": "items.*.MachineType(zone),nextPageToken"
level=fatal msg=      }
level=fatal msg=    ]
level=fatal msg=  }
level=fatal msg=]
level=fatal msg=, invalidParameter 
```

barbacbd (commented):
/lgtm

pkg/asset/installconfig/gcp/client.go
```diff
@@ -80,6 +81,40 @@ func (c *Client) GetMachineType(ctx context.Context, project, zone, machineType
...
+
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+
+	var machine *compute.MachineType
+	zones := sets.New[string]()
+	filter := fmt.Sprintf("name = \"%s\" AND zone : %s-*", machineType, region)
+	req := svc.MachineTypes.AggregatedList(project).Filter(filter).Context(ctx)
+	err = req.Pages(ctx, func(page *compute.MachineTypeAggregatedList) error {
+		for _, scopedList := range page.Items {
```
Prashanth684, line 99: i'm guessing there'll be only 1 "MachineTypesScopedList" item because this is just for one instance type right? wondering if we can get away with just referencing the first item ([0])

pkg/asset/installconfig/gcp/client.go
```diff
@@ -80,6 +81,40 @@ func (c *Client) GetMachineType(ctx context.Context, project, zone, machineType
...
+			for _, item := range scopedList.MachineTypes {
+				if machine == nil {
+					machine = item
```
Prashanth684, line 102: maybe if we do what i  suggested in the previous comment, this can also be referenced to the first one?

pkg/asset/machines/gcp/zones.go
```diff
@@ -48,3 +49,39 @@ func AvailabilityZones(project, region string) ([]string, error) {
...
+		return nil, errors.Wrap(err, "failed to get session")
+	}
+
+	svc, err := compute.NewService(ctx, option.WithCredentials(ssn.Credentials))
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create compute service")
+	}
+
+	found := sets.New[string]()
+	filter := fmt.Sprintf("name = \"%s\" AND zone : %s-*", instanceType, region)
```
Prashanth684, line 72: can we get this routine in a common place and ref that function here?

pkg/asset/installconfig/gcp/client.go
barbacbd: We could technically access it like that yes, but you still need to have checks to ensure that items were returned. I believe that it is still possible to have 0 returned in which a static access would fail.

pkg/asset/machines/gcp/zones.go
r4f4: Addressed the the latest update https://github.com/openshift/installer/pull/7096/commits/c7e91d546f4c963fe8efe82ef585544a84110016

Prashanth684:
lgtm!

r4f4:
Update: aargh. I always forget the dots in the exported symbol comments.

pkg/asset/installconfig/gcp/client.go
r4f4: We're still going through the whole list but now because it's a shared code. It should address this comment.

barbacbd (commented):
/lgtm

r4f4:
/hold
@AnnaZivkovic just found out that some zones might not be available for some projects. So the list of zones where the machineType is available needs to be intersected with the list of zones available for the project:
```
$: gcloud --project "openshift-dev-installer" compute zones list --filter="name=us-central1"
NAME           REGION       STATUS  NEXT_MAINTENANCE  TURNDOWN_DATE
us-central1-c  us-central1  UP
us-central1-a  us-central1  UP
us-central1-f  us-central1  UP
us-central1-b  us-central1  UP

$: gcloud --project=openshift-dev-installer compute machine-types list --filter="zone:us-central1 AND name=t2a-standard-4"
NAME            ZONE           CPUS  MEMORY_GB  DEPRECATED
t2a-standard-4  us-central1-a  4     16.00
t2a-standard-4  us-central1-b  4     16.00
t2a-standard-4  us-central1-f  4     16.00
t2a-standard-4  us-central1-d  4     16.00
```
In the above output, note that:
* us-central1-c: available for the project but doesn't contain t2a-standard instances
* us-central1-d: contains t2a-standard but is not available for the project.

r4f4:
Update: rebased on top of current master and fixed to filter by zones available in project.
/hold cancel

pkg/asset/installconfig/gcp/client.go
```diff
@@ -80,6 +82,62 @@ func (c *Client) GetMachineType(ctx context.Context, project, zone, machineType
...
 }
 
+// GetMachineTypeList retrieves the machine type with the specified fields.
+func GetMachineTypeList(ctx context.Context, svc *compute.Service, project, region, machineType, fields string) ([]*compute.MachineType, error) {
+	var machines []*compute.MachineType
+
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+
+	filter := fmt.Sprintf("name = \"%s\" AND zone : %s-*", machineType, region)
```
rna-afk, line 92: text says zone but variable is region? Mismatch or is that how the query works?

rna-afk: ah got it. this would work.

r4f4: Yeah, the zone name is prefixed by the region's name.

PR #7098, opened 2023-04-17
OCPBUGS-9081: openstack: Bump Gophercloud

pierreprinetti:
Version v1.3.0 contains an improved handling of the bulk deletion of Swift objects, leading to a faster deprovisioning of clusters.


mandre (commented):
/lgtm

PR #7099, opened 2023-04-17
OpenStack: enable ingress traffic for dual-stack installations

MaysaMacedo:
Enable ingress traffic for api and ingress. Note that no internal communication over IPv6 is enabled because OpenStack has no support yet for IPv6 primary clusters, consequently IPv4 is always enforced.

### DIFF
--- a/data/data/openstack/masters/sg-master.tf
+++ b/data/data/openstack/masters/sg-master.tf
@@ -78,6 +78,18 @@ resource "openstack_networking_secgroup_rule_v2" "master_ingress_api" {
   description       = local.description
 }
 
+resource "openstack_networking_secgroup_rule_v2" "master_ingress_api_v6" {
+  count             = length(var.machine_v6_cidrs)
+  direction         = "ingress"
+  ethertype         = "IPv6"
+  protocol          = "tcp"
+  port_range_min    = 6443
+  port_range_max    = 6443
+  remote_ip_prefix  = "::/0"
+  security_group_id = openstack_networking_secgroup_v2.master.id
+  description       = local.description
+}
+
 resource "openstack_networking_secgroup_rule_v2" "master_ingress_vxlan" {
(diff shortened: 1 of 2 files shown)


MaysaMacedo:
@mandre Yes, but I would prefer to add that once I start taking a look on UPI given that at the moment I'm focused on IPI. Would that work for you?

mandre:
> @mandre Yes, but I would prefer to add that once I start taking a look on UPI given that at the moment I'm focused on IPI. Would that work for you?

Fine by me. In that case, could you make sure the same change is tracked in Jira for UPI, linking to that PR?
/lgtm

MaysaMacedo:
/cc @dulek
would you mind taking a look at this PR as well?

MaysaMacedo:
@mandre here is the card https://issues.redhat.com/browse/OSASINFRA-3142

dulek:
I'm late to the party, but I wonder if it's a good idea to only add IPv6 rules when we have a dual stack setup when at some point we might want to enable a way to convert an IPv4 env to dual stack. Then users will need to add these rules manually as we don't have a way to rerun terraform.

MaysaMacedo:
@dulek Excellent point. The rules I opened here are related to access to the api and ingress, so I used the api and ingress VIPs when testing that connectivity. When you convert from single stack to dual stack the VIPs remain the same on an already installed cluster, the only thing that changes is the internal connectivity of the cluster.

dulek:
> @dulek Excellent point. The rules I opened here are related to access to the api and ingress, so I used the api and ingress VIPs when testing that connectivity. When you convert from single stack to dual stack the VIPs remain the same on an already installed cluster, the only thing that changes is the internal connectivity of the cluster. I think it would makes sense to always open if we allowed ipv6 first subnet in dual-stack clusters with openstack, but at the moment that is not the target.

Great, thanks!

PR #7100, opened 2023-04-17
OCPBUGS-6759: Fix discrepancy with disk size master

rna-afk:
The master disk size default is set to 1024 when the max size allowed in azurestack is 1024. Futhermore, the check in the tfvars sets the minimum value for disk size to 1023 which should not be done since the users can configure their own size. Changing these values.

### DIFF
--- a/data/data/azurestack/cluster/master/master.tf
+++ b/data/data/azurestack/cluster/master/master.tf
@@ -54,7 +54,7 @@ resource "azurestack_virtual_machine" "master" {
   storage_os_disk {
     name              = "${var.cluster_id}-master-${count.index}_OSDisk" # os disk name needs to match cluster-api convention
     create_option     = "FromImage"
-    disk_size_gb      = min(var.os_volume_size, 1023)
+    disk_size_gb      = var.os_volume_size
     managed_disk_type = var.os_volume_type
   }
 
--- a/pkg/asset/machines/master.go
+++ b/pkg/asset/machines/master.go
@@ -344,6 +344,9 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 			installConfig.Config.ControlPlane.Architecture,
 		)
 		mpool.OSDisk.DiskSizeGB = 1024
+		if installConfig.Config.Platform.Azure.CloudName == azuretypes.StackCloud {
+			mpool.OSDisk.DiskSizeGB = azuredefaults.AzurestackMinimumDiskSize
+		}
 		mpool.Set(ic.Platform.Azure.DefaultMachinePlatform)
 		mpool.Set(pool.Platform.Azure)
 
--- a/pkg/types/azure/validation/machinepool.go
+++ b/pkg/types/azure/validation/machinepool.go
@@ -5,6 +5,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
 	"github.com/openshift/installer/pkg/types/azure"
+	"github.com/openshift/installer/pkg/types/azure/defaults"
 )
 
 // ValidateMachinePool checks that the specified machine pool is valid.
@@ -13,6 +14,8 @@ func ValidateMachinePool(p *azure.MachinePool, poolName string, platform *azure.
 
 	if p.OSDisk.DiskSizeGB < 0 {
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("diskSizeGB"), p.OSDisk.DiskSizeGB, "Storage DiskSizeGB must be positive"))
+	} else if platform.CloudName == azure.StackCloud && p.OSDisk.DiskSizeGB != 0 && (p.OSDisk.DiskSizeGB < defaults.AzurestackMinimumDiskSize || p.OSDisk.DiskSizeGB > defaults.AzurestackMaximumDiskSize) {
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("diskSizeGB"), p.OSDisk.DiskSizeGB, "Storage DiskSizeGB must be between 128 and 1023 inclusive for Azure Stack"))
 	}
 
 	if p.OSDisk.DiskType != "" {
--- a/pkg/types/azure/defaults/platform.go
+++ b/pkg/types/azure/defaults/platform.go
@@ -7,6 +7,11 @@ import (
 var (
 	// Overrides
 	defaultMachineClass = map[string]string{}
+
+	// AzurestackMinimumDiskSize is the minimum disk size value for azurestack.
+	AzurestackMinimumDiskSize int32 = 128
+	// AzurestackMaximumDiskSize is the maximum disk size value for azurestack.
+	AzurestackMaximumDiskSize int32 = 1023
 )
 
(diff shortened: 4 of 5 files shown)


barbacbd (commented):
based on the comments in the linked jira card /lgtm

data/data/azurestack/cluster/master/master.tf
```diff
@@ -54,7 +54,7 @@ resource "azurestack_virtual_machine" "master" {
   storage_os_disk {
     name              = "${var.cluster_id}-master-${count.index}_OSDisk" # os disk name needs to match cluster-api convention
     create_option     = "FromImage"
-    disk_size_gb      = min(var.os_volume_size, 1023)
+    disk_size_gb      = max(var.os_volume_size, 1023)
```
patrickdillon, line 57: Hm I think my comment about max vs min in the jira card was mistaken. `min` seems correct here. If we set this to `max` and `1023` is the maximum value according to the bug, then the disk size will always be set to 1023. The default value of 128 would effectively be ignored.
https://developer.hashicorp.com/terraform/language/functions/max

patrickdillon: Changing this to 
```
max(var.os_volume_size, 128)
```
would work, but we would need to make sure there is validation to cap the disk size at 1023, at which point it would seem to make more sense to just add validation to make sure disk size is between 128 and 1023

rna-afk: How about current change? Not sure if it would even work

barbacbd (commented):
/lgtm

data/data/azurestack/cluster/master/master.tf
```diff
@@ -54,7 +54,7 @@ resource "azurestack_virtual_machine" "master" {
   storage_os_disk {
     name              = "${var.cluster_id}-master-${count.index}_OSDisk" # os disk name needs to match cluster-api convention
     create_option     = "FromImage"
-    disk_size_gb      = min(var.os_volume_size, 1023)
+    disk_size_gb      = max(128, min(var.os_volume_size, 1023))
```
patrickdillon, line 57: I suspect we don't really need logic here at all, we can simply:
```suggestion
    disk_size_gb      = var.os_volume_size
```

jinyunma:
@patrickdillon @rna-afk I did some testing on payload including this PR.

When setting controlPlane.platform.azure.osDisk.diskSizeGB to 1024, original issue reported in bug still exist, details see test scenario 2 in my [comments](https://issues.redhat.com/browse/OCPBUGS-6759?focusedId=22223787&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22223787) from bug OCPBUGS-6759.

```
disk_size_gb = max(128, min(var.os_volume_size, 1023))
```
Do you think is it necessary to do some validation for disk size? make sure that disk size setting is between 128 and 1023.

jinyunma:
@rna-afk tested with new commit:
when there is no disk size setting on control plane and compute node in install-config:
```
controlPlane:
  architecture: amd64
  hyperthreading: Enabled
  name: master
  platform:
    azure: {}
  replicas: 3
compute:
- architecture: amd64
  hyperthreading: Enabled
  name: worker
  platform:
    azure: {}
  replicas: 3
```
Running openshift-installer, got error as below, I think it should be set to default value 128GB if no setting, wdyt?
```
07-25 15:13:07.335  level=error msg=failed to fetch Master Machines: failed to load asset "Install Config": failed to create install config: invalid "install-config.yaml" file: [controlPlane.platform.azure.diskSizeGB: Invalid value: 0: Storage DiskSizeGB must be between 128 and 1023 inclusive for Azure Stack, compute[0].platform.azure.diskSizeGB: Invalid value: 0: Storage DiskSizeGB must be between 128 and 1023 inclusive for Azure Stack]
```

rna-afk:
@jinyunma Thanks I fixed the validation to exclude the zero check. Should work now.

jinyunma:
pre-merge testing is passed, details can be found from https://issues.redhat.com/browse/OCPBUGS-6759?focusedId=22684784&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22684784.

/label qe-approved

pkg/types/azure/validation/machinepool.go
```diff
@@ -13,6 +13,8 @@ func ValidateMachinePool(p *azure.MachinePool, poolName string, platform *azure.
 
 	if p.OSDisk.DiskSizeGB < 0 {
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("diskSizeGB"), p.OSDisk.DiskSizeGB, "Storage DiskSizeGB must be positive"))
+	} else if platform.CloudName == azure.StackCloud && p.OSDisk.DiskSizeGB != 0 && (p.OSDisk.DiskSizeGB < 128 || p.OSDisk.DiskSizeGB > 1023) {
```
r4f4, line 16: Can we define these numbers as contants somewhere with a link to the Azure docs where they are specified?

pkg/types/azure/validation/machinepool.go
```diff
@@ -13,6 +13,8 @@ func ValidateMachinePool(p *azure.MachinePool, poolName string, platform *azure.
...
+		allErrs = append(allErrs, field.Invalid(fldPath.Child("diskSizeGB"), p.OSDisk.DiskSizeGB, "Storage DiskSizeGB must be between 128 and 1023 inclusive for Azure Stack"))
```
r4f4, line 17: What about some unit tests?

pkg/asset/machines/master.go
```diff
@@ -344,6 +344,9 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 			installConfig.Config.ControlPlane.Architecture,
 		)
 		mpool.OSDisk.DiskSizeGB = 1024
+		if installConfig.Config.Platform.Azure.CloudName == azuretypes.StackCloud {
+			mpool.OSDisk.DiskSizeGB = 128
```
r4f4, line 348: Nit: can we use the constant you defined here?

r4f4 (commented):
/lgtm

pkg/types/azure/validation/machinepool.go
```diff
@@ -13,6 +14,8 @@ func ValidateMachinePool(p *azure.MachinePool, poolName string, platform *azure.
 
 	if p.OSDisk.DiskSizeGB < 0 {
 		allErrs = append(allErrs, field.Invalid(fldPath.Child("diskSizeGB"), p.OSDisk.DiskSizeGB, "Storage DiskSizeGB must be positive"))
+	} else if platform.CloudName == azure.StackCloud && p.OSDisk.DiskSizeGB != 0 && (p.OSDisk.DiskSizeGB < defaults.AzurestackMinimumDiskSize || p.OSDisk.DiskSizeGB > defaults.AzurestackMaximumDiskSize) {
```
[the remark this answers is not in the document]
rna-afk: The validation should not run if the value is 0. 0 means the user did not provide a value in the install config. If I leave it to `< MinDiskSize`, it throws error.

PR #7101, opened 2023-04-17
OCPBUGS-7410: Reject active VPC connections before service destroy

rna-afk:
The installer fails to destroy VPC endpoint service with active VPC connections to it. It needs to reject all active connections before attempting to destroy the service.

### DIFF
--- a/pkg/destroy/aws/ec2helpers.go
+++ b/pkg/destroy/aws/ec2helpers.go
@@ -92,7 +92,7 @@ func deleteEC2(ctx context.Context, session *session.Session, arn arn.ARN, logge
 	if err != nil {
 		return err
 	}
-	logger = logger.WithField("id", id)
+	logger = logger.WithField("id", id).WithField("resourceType", resourceType)
 
 	switch resourceType {
 	case "dhcp-options":
@@ -786,7 +786,40 @@ func deleteEC2VPCPeeringConnection(ctx context.Context, client *ec2.EC2, id stri
 }
 
 func deleteEC2VPCEndpointService(ctx context.Context, client *ec2.EC2, id string, logger logrus.FieldLogger) error {
-	_, err := client.DeleteVpcEndpointServiceConfigurationsWithContext(ctx, &ec2.DeleteVpcEndpointServiceConfigurationsInput{
+	output, err := client.DescribeVpcEndpointConnectionsWithContext(ctx, &ec2.DescribeVpcEndpointConnectionsInput{
+		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("service-id"),
+				Values: aws.StringSlice([]string{id}),
+			},
+		},
+	})
+
+	if err != nil {
+		logger.Warn("Unable to get the list of VPC endpoint connections connected to service: ", err)
+		logger.Warn("Attempting to delete the VPC Endpoint Service")
+	} else {
+		endpointList := make([]*string, len(output.VpcEndpointConnections))
+		for _, endpoint := range output.VpcEndpointConnections {
+			if aws.StringValue(endpoint.VpcEndpointState) != "rejected" {
+				endpointList = append(endpointList, endpoint.VpcEndpointId)
+			}
+		}
+
+		_, err = client.RejectVpcEndpointConnectionsWithContext(ctx, &ec2.RejectVpcEndpointConnectionsInput{
+			ServiceId:      &id,
+			VpcEndpointIds: endpointList,
+		})
+
+		if err != nil {
+			logger.Warn("Unable to reject VPC endpoint connections for service: ", err)
+			logger.Warn("Attempting to delete the VPC Endpoint Service")
+		} else {
+			logger.WithField("resourceType", "VPC Endpoint Connection").Info("Rejected")
+		}
+	}
+
+	_, err = client.DeleteVpcEndpointServiceConfigurationsWithContext(ctx, &ec2.DeleteVpcEndpointServiceConfigurationsInput{
 		ServiceIds: aws.StringSlice([]string{id}),
 	})
 	if err != nil {


pkg/destroy/aws/ec2helpers.go
```diff
@@ -786,7 +786,16 @@ func deleteEC2VPCPeeringConnection(ctx context.Context, client *ec2.EC2, id stri
 }
 
 func deleteEC2VPCEndpointService(ctx context.Context, client *ec2.EC2, id string, logger logrus.FieldLogger) error {
-	_, err := client.DeleteVpcEndpointServiceConfigurationsWithContext(ctx, &ec2.DeleteVpcEndpointServiceConfigurationsInput{
+	_, err := client.RejectVpcEndpointConnections(&ec2.RejectVpcEndpointConnectionsInput{
+		ServiceId: &id,
```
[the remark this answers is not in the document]
rna-afk: Was going off of this comment in their code. I'll pull the list and pass it.

```
// Rejects VPC endpoint connection requests to your VPC endpoint service.
```

pkg/destroy/aws/ec2helpers.go
```diff
@@ -786,7 +786,38 @@ func deleteEC2VPCPeeringConnection(ctx context.Context, client *ec2.EC2, id stri
...
+			{
+				Name:   aws.String("service-id"),
+				Values: aws.StringSlice([]string{id}),
+			},
+		},
+	})
+
+	if err != nil {
+		logrus.Warn("Unable to get the list of VPC endpoint connections connected to service: ", err)
+		logrus.Warn("Attempting to delete the VPC Endpoint Service")
```
r4f4, line 800: So any specific reason for not using `logger` passed in as argument? That way we could track this error back to the specific resource `id` that caused it.

pkg/destroy/aws/ec2helpers.go
```diff
@@ -786,7 +786,38 @@ func deleteEC2VPCPeeringConnection(ctx context.Context, client *ec2.EC2, id stri
...
+		}
+
+		_, err = client.RejectVpcEndpointConnectionsWithContext(ctx, &ec2.RejectVpcEndpointConnectionsInput{
+			ServiceId:      &id,
+			VpcEndpointIds: endpointList,
+		})
+
+		if err != nil {
+			logrus.Warn("Unable to reject VPC endpoint connections for service: ", err)
+			logrus.Warn("Attempting to delete the VPC Endpoint Service")
```
r4f4, line 816: Same applies here.

r4f4 (commented):
I don't see any failures in the e2e-aws-ovn deprovision but it's also hard to say if there were any active Service Endpoint connections or not.

pkg/destroy/aws/ec2helpers.go
rna-afk: Was putting more focus on if this actually works. Fixed it.

rna-afk:
>I don't see any failures in the e2e-aws-ovn deprovision but it's also hard to say if there were any active Service Endpoint connections or not.

I don't think we test for that.

patrickdillon (commented):
Awesome thanks for pushing this forward.

/approve

A couple of small changes otherwise LGTM

pkg/destroy/aws/ec2helpers.go
```diff
@@ -786,7 +786,40 @@ func deleteEC2VPCPeeringConnection(ctx context.Context, client *ec2.EC2, id stri
...
+		_, err = client.RejectVpcEndpointConnectionsWithContext(ctx, &ec2.RejectVpcEndpointConnectionsInput{
+			ServiceId:      &id,
+			VpcEndpointIds: endpointList,
+		})
+
+		if err != nil {
+			logger.Warn("Unable to reject VPC endpoint connections for service: ", err)
+			logger.Warn("Attempting to delete the VPC Endpoint Service")
+		}
+		logger.Info("Rejected")
```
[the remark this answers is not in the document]
rna-afk: Almost all logs are just this one word statements like `deleted`, `disassociated` etc. Want me to change this too?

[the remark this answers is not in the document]
r4f4: The logger already has the `id` of the resource. If we were to change this, maybe it would be good to include `resourceType` in https://github.com/openshift/installer/blob/master/pkg/destroy/aws/ec2helpers.go#L95

pkg/destroy/aws/ec2helpers.go
```diff
@@ -786,7 +786,39 @@ func deleteEC2VPCPeeringConnection(ctx context.Context, client *ec2.EC2, id stri
...
+		_, err = client.RejectVpcEndpointConnectionsWithContext(ctx, &ec2.RejectVpcEndpointConnectionsInput{
+			ServiceId:      &id,
+			VpcEndpointIds: endpointList,
+		})
+
+		if err != nil {
+			logger.Warn("Unable to reject VPC endpoint connections for service: ", err)
+			logger.Warn("Attempting to delete the VPC Endpoint Service")
+		}
+		logger.WithField("resourceType", "VPC Endpoint Connection").Info("Rejected")
```
r4f4, line 818: This is fine for this PR but maybe in the future we can do: 
```go
    logger = logger.WithField("id", id, "resourceType", resourceType)
```
at 
https://github.com/openshift/installer/blob/54a78f56dd2564993bbd876ed9ade69cd2b162cc/pkg/destroy/aws/ec2helpers.go#L95

so the logs for all the resource types clean up functions have both `id` and `type` and the log output is more informative.

r4f4, line 818:
```suggestion
		} else {
		    logger.WithField("resourceType", "VPC Endpoint Connection").Info("Rejected")
		}
```
otherwise you're going to log that you could not reject the connection and that it was rejected.

pkg/destroy/aws/ec2helpers.go
```diff
@@ -92,7 +92,7 @@ func deleteEC2(ctx context.Context, session *session.Session, arn arn.ARN, logge
 	if err != nil {
 		return err
 	}
-	logger = logger.WithField("id", id)
+	logger = logger.WithField("id", id).WithField("resourceType", resourceType)
```
r4f4, line 95: Awesome!

pkg/destroy/aws/ec2helpers.go
```diff
@@ -786,7 +786,40 @@ func deleteEC2VPCPeeringConnection(ctx context.Context, client *ec2.EC2, id stri
...
+		_, err = client.RejectVpcEndpointConnectionsWithContext(ctx, &ec2.RejectVpcEndpointConnectionsInput{
+			ServiceId:      &id,
+			VpcEndpointIds: endpointList,
+		})
+
+		if err != nil {
+			logger.Warn("Unable to reject VPC endpoint connections for service: ", err)
+			logger.Warn("Attempting to delete the VPC Endpoint Service")
+		} else {
+			logger.WithField("resourceType", "VPC Endpoint Connection").Info("Rejected")
```
r4f4, line 818: `resourceType` is duplicated now. You can drop `WithField` from here :)

rna-afk: So the actual resourceType getting deleted is the VPC Endpoint but this one is for the Endpoint connection.

r4f4: Aaaah, I see. My bad.

PR #7106, opened 2023-04-18
OCPBUGS-9066: Retry image download on failure

dulek:
Sometimes in the CI we see transient failures to download the image by the installer. This shouldn't make the whole installation to fail. This commit attempts to solve it by retrying the download on failures.


pkg/asset/agent/image/cache.go
```diff
@@ -249,7 +249,8 @@ func (u *urlWithIntegrity) download(dataType string) (string, error) {
 
 // DownloadImageFile is a helper function that obtains an image file from a given URL,
 // puts it in the cache and returns the local file path.  If the file is compressed
-// by a known compressor, the file is uncompressed prior to being returned.
+// by a known compressor, the file is uncompressed prior to being returned.  The download
+// is retried 3 times in case of an error.
```
mandre, line 253: Nit, it's only retried 2 times (but tried 3 times) at most. We should update the comment.

dulek: Done!

pkg/asset/agent/image/cache.go
```diff
@@ -266,5 +267,16 @@ func DownloadImageFile(baseURL string) (string, error) {
 	}
 	u.location = *parsedURL
 
-	return u.download(imageDataType)
+	for i := 0; i < 3; i++ {
+		path := ""
+		path, err = u.download(imageDataType)
+		if err == nil {
+			return path, nil
+		} else if i < 2 {
```
andfasano, line 275: I don't think this check is required. If the download had an error, it would be useful to log it always (at info level)

pkg/tfvars/internal/cache/cache.go
```diff
@@ -243,5 +244,16 @@ func DownloadImageFile(baseURL string) (string, error) {
 	}
 	u.location = *parsedURL
 
-	return u.download(imageDataType)
+	for i := 0; i < 3; i++ {
```
andfasano, line 247: I understand that this part was already duplicated, it would be nice to find a way at least to avoid increasing it (cc @bfournie that may have some better insight for that)

pkg/asset/agent/image/cache.go
```diff
@@ -266,5 +267,16 @@ func DownloadImageFile(baseURL string) (string, error) {
...
+			// Only log during first 2 tries.
+			logrus.Warnf("Error when obtaining RHCOS image file, retrying: %s", err)
+		}
```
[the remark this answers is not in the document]
dulek: Alright, I'll add a 5 seconds of sleep here.

pkg/tfvars/internal/cache/cache.go
dulek: I'm open to suggestions. Could we e.g. move this under rhcos module?

bfournie: Sorry I was on PTO when the cc above was added and did not notice it.

The reason the function was duplicated, as you probably know, was because its in an `internal` dir and can't be accessed from `pkg/asset/agent/`.
I think moving it to `pkg/rhcos` may be an option as that's already included by files both in `pkg/tfvars/` and in `pkg/asset/agent/`

bfournie:
Hi @dulek, is this PR still valid? Is there some evidence that it resolves the issue identified in https://issues.redhat.com//browse/OCPBUGS-9066?

dulek:
> Hi @dulek, is this PR still valid? Is there some evidence that it resolves the issue identified in https://issues.redhat.com//browse/OCPBUGS-9066?

I guess it could be valid, but given its low priority I don't have time to work on this if it requires me to deduplicate the code first. I also don't have any real proof this helps, but it should.

dulek:
> > > Hi @dulek, is this PR still valid? Is there some evidence that it resolves the issue identified in https://issues.redhat.com//browse/OCPBUGS-9066?
> > 
> > 
> > I guess it could be valid, but given its low priority I don't have time to work on this if it requires me to deduplicate the code first. I also don't have any real proof this helps, but it should.
> 
> OK, I just wanted to confirm that you are using the agent-based installer, i.e. `openshift-install agent create image `? This fix is only relevant for ABI

Hm, weird. I believe we're doing regular installation and found the logs indicating that the error originates here. Can you point me where the `openshift-install create cluster` would download the image?

bfournie:
> > > > Hi @dulek, is this PR still valid? Is there some evidence that it resolves the issue identified in https://issues.redhat.com//browse/OCPBUGS-9066?
> > > 
> > > 
> > > I guess it could be valid, but given its low priority I don't have time to work on this if it requires me to deduplicate the code first. I also don't have any real proof this helps, but it should.
> > 
> > 
> > OK, I just wanted to confirm that you are using the agent-based installer, i.e. `openshift-install agent create image `? This fix is only relevant for ABI
> 
> Hm, weird. I believe we're doing regular installation and found the logs indicating that the error originates here. Can you point me where the `openshift-install create cluster` would download the image?

Actually, never mind, you got the right place since you changed the download in 2 places. The code in pkg/asset/agent/image isn't relevant here for this bug but its useful to change that too.

sadasu:
@dulek, @andfasano and @bfournie are we still interested in seeing this fix through?

mandre:
> @dulek, @andfasano and @bfournie are we still interested in seeing this fix through?

Most definitely. We're still seeing errors dowloading the RHCOS image from time to time in our CI and would like a fix.
For instance https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-shiftstack-shiftstack-ci-main-periodic-4.15-upgrade-from-stable-4.14-e2e-openstack-ovn-upgrade/1713949752652992512#1:build-log.txt%3A203

andfasano:
> > @dulek, @andfasano and @bfournie are we still interested in seeing this fix through?
> 
> Most definitely. We're still seeing errors dowloading the RHCOS image from time to time in our CI and would like a fix. For instance https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-shiftstack-shiftstack-ci-main-periodic-4.15-upgrade-from-stable-4.14-e2e-openstack-ovn-upgrade/1713949752652992512#1:build-log.txt%3A203

Also for agent related jobs (this patch will impact just them)?

pkg/asset/agent/image/cache.go
```diff
@@ -266,5 +267,16 @@ func DownloadImageFile(baseURL string) (string, error) {
...
 
-	return u.download(imageDataType)
+	for i := 0; i < 3; i++ {
+		path := ""
+		path, err = u.download(imageDataType)
+		if err == nil {
+			return path, nil
+		} else if i < 2 {
+			// Only log during first 2 tries.
+			logrus.Warnf("Error when obtaining RHCOS image file, retrying: %s", err)
```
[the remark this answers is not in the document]
andfasano: On a second thought, by reviewing this code, I think that using a retry.Do to wrap that portion of code:

```
resp, err := http.Get(u.location.String())
	if err != nil {
		return "", err
	}
	defer resp.Body.Close()

	// Check server response
	if resp.StatusCode != http.StatusOK {
		return "", errors.Errorf("bad status: %s", resp.Status)
	}
```

Will allow to implement a more generic retry mechanism, will require fewer lines of code and it will also help in reducing the code duplication (instead of apply the retry outside of the `download` function). @bfournie wdyt?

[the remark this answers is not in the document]
dulek: Done!

dulek:
/hold

Wait, this is wrong, we'd like to catch and retry the invalid checksum errors too.

dulek:
Okay, this has two commits. One is deduplicating the code by moving it to pkt/rhcos. The second one is adding a `retry.DoFunc()` to the download.

dulek:
@andfasano, @mandre, @bfournie, @sadasu: This should be ready for reviews now.

mandre (commented):
/lgtm

dulek:
I addressed a merged conflict.

mandre (commented):
/lgtm

PR #7108, opened 2023-04-18
CORS-2372: Azure: auth Installer with Managed Identity from VM

r4f4:
Use the managed identity from the installer host VM (running in Azure) so that a user can install a cluster without copying credentials to the installer host.

### DIFF
--- a/pkg/asset/installconfig/azure/session.go
+++ b/pkg/asset/installconfig/azure/session.go
@@ -29,6 +29,16 @@ var (
 	onceLoggers         = map[string]*sync.Once{}
 )
 
+// AuthenticationType identifies the authentication method used.
+type AuthenticationType int
+
+// The authentication types supported by the installer.
+const (
+	ClientSecretAuth AuthenticationType = iota
+	ClientCertificateAuth
+	ManagedIdentityAuth
+)
+
 // Session is an object representing session for subscription
 type Session struct {
 	Authorizer   autorest.Authorizer
@@ -37,6 +47,7 @@ type Session struct {
 	AuthProvider *azurekiota.AzureIdentityAuthenticationProvider
 	TokenCreds   azcore.TokenCredential
 	CloudConfig  cloud.Configuration
+	AuthType     AuthenticationType
 }
 
 // Credentials is the data type for credentials as understood by the azure sdk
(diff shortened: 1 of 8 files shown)


r4f4:
I have added 2 more commits:
1. https://github.com/openshift/installer/pull/7108/commits/be6caf19c371351ac0f47c6a68f4d0ee99bc963b: validate that credential mode is set to "Manual" when using Managed Identities
1. https://github.com/openshift/installer/pull/7108/commits/921840cc7fde2712345e57daac1e39ce24cf2053: cosmetic change to print a warning about managed identities only once. I wouldn't mind dropping this if reviewers think it doesn't bring any benefits.

pkg/asset/installconfig/azure/session.go
```diff
@@ -98,11 +109,31 @@ func GetSessionWithCredentials(cloudName azure.CloudEnvironment, armEndpoint str
...
+		onceLoggers[authKey].Do(func() {
+			logrus.Warnf("Using client certs to authenticate. Please be warned cluster does not support certs and only the installer does.")
+		})
 		cred, err = newTokenCredentialFromCertificates(credentials, cloudConfig)
-	} else {
+		authType = ClientCertificateAuth
+	case credentials.ClientSecret != "":
 		cred, err = newTokenCredentialFromCredentials(credentials, cloudConfig)
+		authType = ClientSecretAuth
+	default:
```
barbacbd, line 127: The default behavior appears to have changed but the original card states no docs changes should be necessary. If this is true should we change the default or docs in this case?

pkg/asset/installconfig/azure/session.go
```diff
@@ -169,15 +201,13 @@ func checkCredentials(creds Credentials) error {
...
 	}
-	if creds.ClientID == "" {
-		return errors.New("could not retrieve clientId from auth file")
-	}
 	if creds.TenantID == "" {
 		return errors.New("could not retrieve tenantId from auth file")
 	}
-	if creds.ClientSecret == "" && creds.ClientCertificatePath == "" {
-		return errors.New("could not retrieve either client secret or client certs from auth file")
+	if (creds.ClientSecret != "" || creds.ClientCertificatePath != "") && creds.ClientID == "" {
```
barbacbd, line 207: nit: technically you can have fewer checks if you check `creds.ClientID == ""` first

r4f4: `creds.ClientID` _can_ be `""` if using system Managed Identity. That's why I had to remove the check here https://github.com/openshift/installer/pull/7108/files#diff-da950a4ddabbede621d9d3b1058bb34f8931c89179306ee88a0e4d76a4cf0b13L172-L174. However, `ClientID` is still required for `ClientSecret` and `ClientCertificatePath`.

pkg/asset/installconfig/azure/session.go
r4f4: We might have to review the need for docs.

barbacbd (commented):
lgtm

patrickdillon:
/approve

Tagging on Brent's lgtm:
/lgtm

PR #7111, opened 2023-04-19
OpenStack: add support to multiple subnets in the bootstrap

MaysaMacedo:
Installations with dual-stack requires the bootstrap node to have one IPv4 and IPv6 Subnet. While the masters already have support for it derived from failure domains work, the bootstrap this not. This commit add support for multiple subnets and ip addresses.


MaysaMacedo:
/cc @pierreprinetti @mdbooth 

Can you have a look? This is just adding the same that currently exists for the masters

dulek:
/lgtm

Makes sense to me.

MaysaMacedo:
/cc @EmilienM
can you have a look?

PR #7112, opened 2023-04-20
WIP: AGENT-601: support for okd-scos

andfasano:
This is PoC to explore the required changes to support okd-scos for agent-based installer.

Notes:
* The `scos.json` is identical to the previous `fcos.json` except for x86_64 metal iso
* The extraction of the base ISO from `machine-os-images` has been excluded because not yet available for scos, so using a direct download (that would be required for disconnected installs)

### DIFF
--- a/hack/build-coreos-manifest.go
+++ b/hack/build-coreos-manifest.go
@@ -10,7 +10,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


hack/build-coreos-manifest.go
```diff
@@ -20,15 +19,23 @@ import (
 const (
 	streamRHCOSJSON = "data/data/coreos/rhcos.json"
 	streamFCOSJSON  = "data/data/coreos/fcos.json"
+	streamSCOSJSON  = "data/data/coreos/scos.json"
 	fcosTAG         = "okd"
+	scosTAG         = "scos"
```
LorbusChris, line 24:
```suggestion
        okdTag          = "okd"
	fcosTAG         = "fcos"
	scosTAG         = "scos"
```
the `okd` and `fcos` Go build tags used elsewhere in the installer are synonymous. Let's add an `okdTAG` here and treat it like `fcosTAG`.

LorbusChris:
We're not going to add an `scos.json` that references SCOS bootimages in the near term, as we don't currently have any distribution channels for these boot images.
We might revisit this once:
- tooling to derive bootimages from ostree containers has been built out (https://github.com/coreos/fedora-coreos-tracker/issues/1151, https://gitlab.com/CentOS/cloud/sagano-builder)
- bootimages are distributed as OCI artifacts through OCI registries

Voting to close this PR.

andfasano:
> We're not going to add an `scos.json` that references SCOS bootimages in the near term, as we don't currently have any distribution channels for these boot images. We might revisit this once:
> 
>     * tooling to derive bootimages from ostree containers has been built out ([Create tooling to take a FCOS-derived container and generate bootable images from it coreos/fedora-coreos-tracker#1151](https://github.com/coreos/fedora-coreos-tracker/issues/1151), https://gitlab.com/CentOS/cloud/sagano-builder)
> 
>     * bootimages are distributed as OCI artifacts through OCI registries
> 
> 
> Voting to close this PR.

Not a problem, as this PR was meant just to explore the ABI gaps to support OKD scos. Is there any documentation available about the current distribution?

PR #1232, opened 2019-02-11
Create separate chains of trust for separate actors

abhinavdahiya:
fixes https://jira.coreos.com/browse/CORS-991

The new chains are based on requirements from the master team

```
etcd:
etcd-signer.crt/key - self-signed, 10 year expiry
etcd-ca-bundle.crt - contains the etcd-signer.crtEtcd-serving-ca-bundle.crt - contains etcd-signer.crt
etcd-client.crt/key - signed by etcd-signer.crt, 10 year expiry

kubelet-csr:
kubelet-csr-signer.crt/key - self-signed 1 day expiry
kubelet-client-ca-bundle.crt - contains kubelet-csr-signer.crt
Individual kubelet client certs - signed by kubelet-csr-signer.crt, 1 day expiry

aggregator:
aggregator-signer.crt/key - self-signed 1 day expiry
aggregator-ca-bundle.crt - contains aggregator-signer.crt
aggregator-client.crt/key - signed by aggregator-signer.crt, 1 day expiry

service-ca - removed

admin-kubeconfig:
admin-kubeconfig-signer.crt/key - self-signed 10 year expiry
Admin-kubeconfig-ca-bundle.crt - contains admin-kubeconfig-signer.crt
admin-kubeconfig-client.crt/key - signed by admin-kubeconfig-signer.crt, 10 year expiry

kube-apiserver-to-kubelet:
kube-apiserver-to-kubelet-signer.crt/key - self-signed X year expiry
Kube-apiserver-to-kubelet-ca-bundle.crt - contains kube-apiserver-to-kubelet-signer.crt (This goes to the kubelet somehow.  I bet it’s the MCD and painful to rotate)
kube-apiserver-to-kubelet-client.crt/key - signed by kube-apiserver-to-kubelet-signer.crt, 1 day expiry. (Keep this subject consistent with what you sign for now.)

kube-control-plane:
kube-control-plane-signer.crt/key - self-signed 1 day expiry
Kube-control-plane-ca-bundle.crt - contains kube-control-plane-signer.crt
kube-control-plane-kube-controller-manager-client.crt/key - signed by kube-control-plane-signer.crt, 1 day expiry. (Keep this subject consistent with what you sign for now.)
kube-control-plane-kube-scheduler-client.crt/key - signed by kube-control-plane-signer.crt, 1 day expiry  (Keep this subject consistent with what you sign for now.)

kube-apiserver-localhost (signs for localhost, 127.0.0.1, and the ipv6 one):
[...]

PR #7115, opened 2023-04-20
SPLAT-1034: Allow CIDR expansion on single zone deployment in AWS

mtulio:
The goal of this change is to provide flexibility for subnet expansion when deploying a cluster in AWS with default IPI in single-zone on AWS.

Currently the installer tries a best-effort to fill all the CIDR blocks from VPC spliting into two blocks: public and private. The problem when creating in single-zone, it could prevent VPC expansion as it will create two subnets /17, when using MachineCIDR (VPC) with /16.

This change checks the length of zones, and when the user-provided only one Availability Zone on install-config.yaml, it will increase one bit on the available blocks for public and private, leaving free at least 50% of the VPC for future expansions, like creating manually subnets into different locations (Local Zones) when the user need it. It also allows the user to create a cluster cheaper then expand it without needing to create a new cluster with new configuration.

Examples:

- Using MachineCIDR /16, two subnets /18 is created, with empty block /17.
- Using MachineCIDR /20, two subnets /22 created, with empty block /21.

Use case:

- As a User, I would like to evaluate OCP minimally in AWS, single zone, then be able to increase to more availability zones
- As a User extending nodes in Local Zones, I would like to create new subnets on recently announced AWS Local Zones locations, and MachineSet manifests, so I can extend my nodes, and workloads, to the metropolitan regions in Local Zones. (supported on 4.12+)

References:

- [1] https://issues.redhat.com/browse/SPLAT-1034
- [2] https://issues.redhat.com/browse/SPLAT-657
- [3] Open Questions section on [Local Zones EP #1232](https://github.com/openshift/enhancements/pull/1232/files#diff-f2590494c9138551ab5b3d5632d11170fe8742a128256f9e1d8ebc63baa19f41R832-R850)
- [4] Document with tested scenarios covering Local Zones implementation is available here:
 https://docs.google.com/document/d/1ePnQCeTsX0UUhkL8aMWeQPsRw_W84xwz-nW20yNzEko/edit#
- [5] Node limits on cluster https://docs.openshift.com/container-platform/4.12/scalability_and_performance/planning-your-environment-according-to-object-maximums.html#cluster-maximums-major-releases_object-limits 

## Examples
1) Install-config with single-zone and default MachineCIDR mask `/20`:

```yaml
apiVersion: v1
publish: External
baseDomain: devcluster.openshift.com
metadata:
  name: "lz-p2-69"

platform:
  aws:
    region: us-east-1

controlPlane:   
  hyperthreading: Enabled 
  name: master
  platform:
    aws:
      zones:
      - us-east-1a

compute:
- name: worker
  platform:
    aws:
      zones:
      - us-east-1a
```

Results:

Screenshot from 2023-04-20 15-04-17

2) Install-config with single-zone with custom MachineCIDR mask `/20`:

```yaml
apiVersion: v1
publish: External
baseDomain: devcluster.openshift.com
metadata:
  name: "lz-p2-68"

platform:
  aws:
    region: us-east-1

networking:
  machineNetwork:
  - cidr: 10.0.0.0/20

controlPlane:   
  hyperthreading: Enabled 
  name: master
  platform:
    aws:
      zones:
      - us-east-1a

compute:
- name: worker
  platform:
    aws:
      zones:
      - us-east-1a

```

Results:

Screenshot from 2023-04-20 14-20-46

### DIFF
--- a/data/data/aws/cluster/vpc/vpc.tf
+++ b/data/data/aws/cluster/vpc/vpc.tf
@@ -1,6 +1,18 @@
 locals {
-  new_private_cidr_range = cidrsubnet(data.aws_vpc.cluster_vpc.cidr_block, 1, 1)
-  new_public_cidr_range  = cidrsubnet(data.aws_vpc.cluster_vpc.cidr_block, 1, 0)
+  # The VPC CIDR block is split into two main blocks for private and public subnets.
+  # Both blocks is sliced depending the number of zones to create subnets on it,
+  # using the maximum it can increase bits from the base MachineCIDR (VPC) for each sub-block.
+  # When the IPI deployment is created with a single zone, the allow_expansion_zones will be
+  # triggered to prevent the expansion to consume all the CIDR blocks, allowing the VPC
+  # to be expanded creating new subnets (on regular AZs, Local or other zone types) in Day-2
+  # Operations, then creating manifests to run workloads on it.
+  # For example: On the deployment with single-zone using MachineCIDR 10.0.0.0/16, the terraform creates:
+  # - 2 * subnets /18: 10.0.0.0/18 and 10.0.128.0/18
+  # - keep two block ranges for expansion: 10.0.64.0-10.0.127.255/18 and 10.0.192.0-10.0.255.255/18
+
+  allow_expansion_zones  = length(var.availability_zones) == 1 ? 1 : 0
+  new_private_cidr_range = cidrsubnet(cidrsubnet(data.aws_vpc.cluster_vpc.cidr_block, 1, 0), local.allow_expansion_zones, 0)
+  new_public_cidr_range  = cidrsubnet(cidrsubnet(data.aws_vpc.cluster_vpc.cidr_block, 1, 1), local.allow_expansion_zones, 0)
 }
 
 resource "aws_vpc" "new_vpc" {


mtulio:
Awaiting for tests
/hold

mtulio:
Scott and Patrick,

I am addressing this issue separately from Local Zones considering it is something I've observed a long time ago in single-zone deployments on OCP. I am also cc'ing colleagues that I already mentioned in that subject [also asking for feedback].

I am not sure if we have, or have already, customers raised this issue. But I am considering this as an improvement, a kind of protection that provides flexibility to users to expand their subnets as Day-2 operations without needing to raise questions/support.

This issue will be addressed in Phase II of Local Zones. The preview of tests is attached to the docs, and I have also had this open question on EP. On LZ implementation I will derivate that implementation on terraform, using public subnets CIDR block to allocate Local Zone subnets, considering expansion too (similar implementation here). I can see more importance on LZ as AWS is constantly launching new zones, which requires subnets and free CIDR blocks to allow users to expand their cluster wherever they need. In the test document, I used all Local Zones in us-east-1 using that logic of expansion on terraform, it resulted in a number relatively high of nodes per zone, considering currently we support ~2k nodes on the entire cluster[2].


I would like to hear you'll feedback about that approach.

/assign @sdodson @patrickdillon 
/cc @wking @jupierce @rvanderp3 @jcpowermac @ffranz @yunjiang29

mtulio:
fyi PR #7137 incorporates changes in this PR, adding Local Zone/edge CIDR blocks.

mtulio:
I am closing this in favor of of https://github.com/openshift/installer/pull/7137 . It will be incorporated there with Local Zones feature.

PR #7123, opened 2023-04-24
CORS-2525: Azure: remove storage account with bootstrap destroy

AnnaZivkovic:
The work is being done to address the issue of Azure storage accounts being persisted for the life of an OpenShift cluster, which incurs costs and management effort. The goal is to destroy the storage account along with other bootstrap resources during the bootstrap destroy process.


r4f4 (commented):
Do we need a new terraform config file? I wonder if it wouldn't be better to just move things between `bootstrap/main.tf` and `vnet/main.tf`.

PR #7124, opened 2023-04-24
OPNET-298: Allow primary-v6 dual-stack on vSphere

mkowalski:
Previously we only supported primary-v6 dual-stack installations only on baremetal. As we are now extending full support of dual-stack to vSphere, we no longer need to enforce the order of networks on those clusters.

Closes: [OPNET-298](https://issues.redhat.com//browse/OPNET-298)
Contributes-to: [OPNET-230](https://issues.redhat.com//browse/OPNET-230)


barbacbd (commented):
/lgtm

dgoodwin:
Needs approved before we can label for merge.

mkowalski:
@dgoodwin all the labels are in place now, can we get into the queue?

PR #7126, opened 2023-04-24
CFE-688: Update install-config CRD to support gcp labels and tags

bharath-b-rh:
PR is for adding new fields userLabels and userTags for gcp platform type, to add user defined labels and tags to all those gcp resources created by installer.

### DIFF
--- a/pkg/types/gcp/validation/platform.go
+++ b/pkg/types/gcp/validation/platform.go
@@ -1,7 +1,9 @@
 package validation
 
 import (
+	"fmt"
 	"os"
+	"regexp"
 	"sort"
 
 	"k8s.io/apimachinery/pkg/util/validation/field"
@@ -65,7 +67,33 @@ var (
 		sort.Strings(validValues)
 		return validValues
 	}()
+
+	// userLabelKeyRegex is for verifying that the label key contains only allowed characters.
+	userLabelKeyRegex = regexp.MustCompile(`^[a-z][0-9a-z_-]{1,62}$`)
+
+	// userLabelValueRegex is for verifying that the label value contains only allowed characters.
+	userLabelValueRegex = regexp.MustCompile(`^[0-9a-z_-]{1,63}$`)
+
+	// userLabelKeyPrefixRegex is for verifying that the label key does not contain restricted prefixes.
+	userLabelKeyPrefixRegex = regexp.MustCompile(`^(?i)(kubernetes\-io|openshift\-io)`)
+
+	// userTagKeyRegex is for verifying that the tag key contains only allowed characters.
+	userTagKeyRegex = regexp.MustCompile(`^[a-zA-Z0-9]([0-9A-Za-z_.-]{0,61}[a-zA-Z0-9])?$`)
+
+	// userTagValueRegex is for verifying that the tag value contains only allowed characters.
+	userTagValueRegex = regexp.MustCompile(`^[a-zA-Z0-9]([0-9A-Za-z_.@%=+:,*#&()\[\]{}\-\s]{0,61}[a-zA-Z0-9])?$`)
+
+	// userTagParentIDRegex is for verifying that the tag parentID contains only allowed characters.
+	userTagParentIDRegex = regexp.MustCompile(`(^[1-9][0-9]{0,31}$)|(^[a-z][a-z0-9-]{4,28}[a-z0-9]$)`)
 )
+
+// maxUserLabelLimit is the maximum userLabels that can be configured as defined in openshift/api.
+// https://github.com/openshift/api/commit/ae73a19d05c35068af16c9aeff375d0b7c936a8a#diff-07b264a49084976b670fb699badaca1795027d6ea732a99226a5388104f6174fR592-R602
+const maxUserLabelLimit = 32
+
+// maxUserTagLimit is the maximum userTags that can be configured as defined in openshift/api.
+// https://github.com/openshift/api/commit/ae73a19d05c35068af16c9aeff375d0b7c936a8a#diff-07b264a49084976b670fb699badaca1795027d6ea732a99226a5388104f6174fR604-R613
+const maxUserTagLimit = 50
 
 // ValidatePlatform checks that the specified platform is valid.
 func ValidatePlatform(p *gcp.Platform, fldPath *field.Path, ic *types.InstallConfig) field.ErrorList {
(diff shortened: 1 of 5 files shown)


pkg/featuregates/featuregate.go
```diff
@@ -0,0 +1,101 @@
...
+func getKnownFeatures() sets.Set[configv1.FeatureGateName] {
+	knownFeatures := sets.New[configv1.FeatureGateName]()
+
+	for _, featureSet := range configv1.FeatureSets {
+		for _, feature := range featureSet.Enabled {
+			knownFeatures.Insert(feature.FeatureGateAttributes.Name)
+		}
+		for _, feature := range featureSet.Enabled {
+			knownFeatures.Insert(feature.FeatureGateAttributes.Name)
+		}
```
r4f4, line 73: Are these duplicates on purpose?

bharath-b-rh: No, sorry, it's an error from my side, second loop is for Disabled features. I will correct it.

pkg/types/validation/installconfig.go
```diff
@@ -1053,8 +1054,29 @@ func validateFeatureSet(c *types.InstallConfig) field.ErrorList {
...
+					allErrs = append(allErrs, field.Forbidden(field.NewPath("platform", "gcp", "userTags"), errMsg))
+				}
+				if len(c.GCP.UserLabels) > 0 {
+					allErrs = append(allErrs, field.Forbidden(field.NewPath("platform", "gcp", "userLabels"), errMsg))
+				}
+			}
+		}
+	}
+
+	if c.FeatureSet != configv1.TechPreviewNoUpgrade && c.FeatureSet != configv1.CustomNoUpgrade {
```
patrickdillon, line 1078: We shouldn't need separate checks for TechPreviewNoUpgrade or CustomNoUpgrade. Either of these feature sets should populate the feature gates, and then we only need a single check against the feature gates. Currently this doesn't work, though, because featureGates is not being populated when TechPreview is set.

pkg/types/gcp/validation/platform.go
```diff
@@ -65,8 +67,34 @@ var (
...
+	userLabelKeyPrefixRegex = regexp.MustCompile(`^(?i)(kubernetes\-io|openshift\-io)`)
+
+	// userTagKeyRegex is for verifying that the tag key contains only allowed characters.
+	userTagKeyRegex = regexp.MustCompile(`^[a-zA-Z0-9]([0-9A-Za-z_.-]{0,61}[a-zA-Z0-9])?$`)
+
+	// userTagValueRegex is for verifying that the tag value contains only allowed characters.
+	userTagValueRegex = regexp.MustCompile(`^[a-zA-Z0-9]([0-9A-Za-z_.@%=+:,*#&()\[\]{}\-\s]{0,61}[a-zA-Z0-9])?$`)
+
+	// userTagParentIDRegex is for verifying that the tag parentID contains only allowed characters.
+	userTagParentIDRegex = regexp.MustCompile(`(^[1-9][0-9]{0,31}$)|(^[a-z][a-z0-9-]{4,28}[a-z0-9]$)`)
```
patrickdillon, line 87: As user tags are pre-existing resources, can we validate the user input by passing it directly to a GCP API call to get the user tag? Unless these regexes are documented/supported by GCP, I think they provide more risk than benefit. 
I'm not sure if it is present in a separate PR, but I think we need validation to ensure the specified user tags indeed exist.

pkg/featuregates/featuregate.go
```diff
@@ -0,0 +1,101 @@
+package featuregates
```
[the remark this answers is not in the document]
bharath-b-rh: Comment incorporated. Source file moved to suggested path.

pkg/types/validation/installconfig.go
bharath-b-rh: Simplified the code by moving GCP tag feature checks to gcp package.

pkg/types/gcp/validation/platform.go
bharath-b-rh: Explicit check on the tag key or tag value existence is not done, but the code [here](https://github.com/openshift/installer/pull/7279/files#diff-fe206eadc78bb637c8b5ebbf96a7fe49eb5eab26b5ce28153f793d8d4c1f72a7R175) fails if the either of them doesn't exist.

Regex created based on the requirements defined by GCP for tag keys and values, same as labels. I think this will be helpful in catching any errors with tags saving the API invocations. Please let me know WDYT?

pkg/types/validation/installconfig.go
patrickdillon: > Currently this doesn't work, though, because featureGates is not being populated when TechPreview is set.

Was this addressed? Is my concern here clear/accurate?

bharath-b-rh: @patrickdillon I think I misunderstood the feedback and I assumed if TechPreviewNoUpgrade is set, then all features marked for TechPreview are by default enabled and updated the code accordingly.

Please correct me, suggestion is to populate featureGates for both TechPreviewNoUpgrade and CustomNoUpgrade, and to make use of featureGates instead of featureSets for checking feature status, correct?

bharath-b-rh: @patrickdillon I have updated featureSet handling to rely on enabled featureGates based on the featureSet configured. Please let me know if the changes are as expected.

pkg/types/gcp/validation/platform.go
patrickdillon: If you want to leave the regexes in, that works for me. I'm not a regex expert but I looked over them carefully and they look good. My preference/advice is still as originally stated, but as far as I can tell this is fine.

pkg/types/gcp/validation/platform.go
```diff
@@ -65,8 +67,34 @@ var (
...
 	}()
+
+	// userLabelKeyRegex is for verifying that the label key contains only allowed characters.
+	userLabelKeyRegex = regexp.MustCompile(`^[a-z][0-9a-z_-]{1,62}$`)
+
+	// userLabelValueRegex is for verifying that the label value contains only allowed characters.
+	userLabelValueRegex = regexp.MustCompile(`^[0-9a-z_-]{1,63}$`)
+
+	// userLabelKeyPrefixRegex is for verifying that the label key does not contain restricted prefixes.
+	userLabelKeyPrefixRegex = regexp.MustCompile(`^(?i)(kubernetes\-io|openshift\-io)`)
```
patrickdillon, line 78: I'm surprised these are the restricted prefixes the hyphens seem weird to me. Do you recall the source of how these were determined?

pkg/types/gcp/validation/platform_test.go
```diff
@@ -175,3 +175,301 @@ func TestValidatePlatform(t *testing.T) {
...
+				{Key: "key20", Value: "value20"}, {Key: "key4", Value: "value4"},
+				{Key: "key23", Value: "value23"}, {Key: "key26", Value: "value26"},
+				{Key: "key12", Value: "value12"}, {Key: "key33", Value: "value33"},
+				{Key: "key1", Value: "value1"}, {Key: "key2", Value: "value2"},
+				{Key: "key5", Value: "value5"}, {Key: "key8", Value: "value8"},
+				{Key: "key30", Value: "value30"}, {Key: "key6", Value: "value6"},
+				{Key: "key13", Value: "value13"}, {Key: "key16", Value: "value16"},
+				{Key: "key24", Value: "value24"},
+			},
+			wantErr: true,
```
patrickdillon, line 223: nit: using an expected errMsg provides more insight into how the code is working. With a simple boolean like this the test could be satisified for the wrong reasons.

patrickdillon (commented):
The featuregate validation looks better but it is still not quite right (the default feature set is not populating the feature gates). But I think the feature gate validation is turning out to be a big lift and we should remove it from this PR, so that this PR can stay focused on the GCP labels and tags. I have started #7413 to handle the feature gate validation. That implementation stays pretty close to what is in the cluster-config-operator. Let's iterate on the feature gates in a separate PR and focus this one solely on the GCP labels and tags.

pkg/types/gcp/validation/platform.go
bharath-b-rh: A label key can contain only hyphen and underscore special characters, unlike period character allowed in AWS, Azure. The OCP default label we add is of the form `kubernetes-io-cluster-<cluster_id>:owned`. Hence`kubernetes-io` and `openshift-io` namespaces are reserved for Openshift's usage and labels having these as prefixes are not allowed.

bharath-b-rh:
> Let's iterate on the feature gates in a separate PR and focus this one solely on the GCP labels and tags.
I think keeping just the TechPreviewNoUpgrade condition for the feature and a separate PR to handle other FeatureSets should be fine.

pkg/types/gcp/validation/platform_test.go
bharath-b-rh: suggestion incorporated.

PR #7130, opened 2023-04-26
OCPBUGS-12751: [Alibaba] update the bandwidth value of EIP

bd233:
Solve the issue of interrupted image downloads from quay.io by increasing the bandwidth value of the EIP bound to the gateway


jianli-wei (approved):
/label qe-approved

PR #7134, opened 2023-04-26
OCPBUGS-12776: GCP XPN Private Cluster Fails with no Public Zone

barbacbd:
** Do no look for a public zone if this is a private cluster. It would not be used anyway, but the look up would always fail if there was no public zone with the same DNS name.


barbacbd:
/retitle OCPBUGS-12776: GCP XPN Private Cluster Fails with no Public Zone

PR #7136, opened 2023-04-26
OSASINFRA-3155 - OpenStack: Create ControlPlaneMachineSet CRDs

EmilienM:
- vendoring
- [OSASINFRA-3155](https://issues.redhat.com//browse/OSASINFRA-3155) - OpenStack: Create ControlPlaneMachineSet CRDs


pkg/asset/machines/openstack/machines.go
```diff
@@ -107,7 +108,70 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 		machines = append(machines, machine)
 	}
 
-	return machines, nil
+	controlPlaneMachineSet := &machinev1.ControlPlaneMachineSet{}
+	replicas := int32(total)
+	failureDomains := []machinev1.OpenStackFailureDomain{}
+
+	if len(mpool.FailureDomains) > 0 {
```
EmilienM, line 115: @pierreprinetti you'll notice how I consume both the TechPreview FailureDomain API that you introduced in 4.13 and the CPMS FailureDomain that we're working on now.

EmilienM: @JoelSpeed @damdo
talked offline to @pierreprinetti and he wondered if we could call a function from CCPMSO here to build the `controlPlaneMachineSet` instead of having to implement a way here, that might be different from CPMS.

is that even possible? is it in the roadmap? or should we just do it like this for now.

pkg/asset/machines/openstack/machines.go
```diff
@@ -107,7 +108,69 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
...
+			Spec: machinev1.ControlPlaneMachineSetSpec{
+				Replicas: &replicas,
+				Selector: metav1.LabelSelector{
+					MatchLabels: map[string]string{
+						"machine.openshift.io/cluster-api-cluster":      clusterID,
+						"machine.openshift.io/cluster-api-machine-role": role,
+						"machine.openshift.io/cluster-api-machine-type": role,
+					},
+				},
+				Template: machinev1.ControlPlaneMachineSetTemplate{
```
pierreprinetti, line 148: We still are in the `if len(mpool.FailureDomains) > 0` block here. Shouldn't we add the machine template to CPMS regardless of the failureDomains?

pierreprinetti:
/uncc pierreprinetti JoelSpeed damdo mdbooth AnnaZivkovic mandre

pierreprinetti:
I have rebased onto master in a separate branch to make comparing easier: https://github.com/openshift/installer/compare/master...shiftstack:installer:openstack_cpms?expand=1

pierreprinetti:
/close
in favour of #7280

PR #7139, opened 2023-04-27
openstack image: Remove oc cli

pierreprinetti:
Since the `oc` tool is injected in the test steps directly, we can save one image layer in the Containerfile.


pierreprinetti:
It is not! Let's see if there is a workable solution for our conf-clouds step.
/hold

pierreprinetti:
Depends on https://github.com/openshift/release/pull/38847

pierreprinetti:
I am less and less convinced that removing `oc` from the base Installer image is a good idea.
The current mechanism might be better for us, as it injects `oc` of the same version as the Installer, rather than just the latest available.

There may be good reasons to merge something like this PR in the future, but for now, it doesn't seem to advance us.

/close

PR #7145, opened 2023-04-28
OCPBUGS-14077: MULTIARCH-3492: Avoid conflicting subnets

hamzy:
Use a random number for DefaultMachineCIDR: 192.168.%d.0/24 This should significantly reduce the chances for collisions.

### DIFF
--- a/pkg/types/powervs/defaults/platform.go
+++ b/pkg/types/powervs/defaults/platform.go
@@ -1,6 +1,10 @@
 package defaults
 
 import (
+	"crypto/rand"
+	"fmt"
+	"math/big"
+
 	"github.com/openshift/installer/pkg/ipnet"
 	"github.com/openshift/installer/pkg/types/powervs"
 )
@@ -13,4 +17,11 @@ var (
 
 // SetPlatformDefaults sets the defaults for the platform.
 func SetPlatformDefaults(p *powervs.Platform) {
+	n, err := rand.Int(rand.Reader, big.NewInt(253))
+	if err != nil {
+		panic(err)
+	}
+	subnet := n.Int64()
+	// MustParseCIDR parses a CIDR from its string representation. If the parse fails, the function will panic.
+	DefaultMachineCIDR = ipnet.MustParseCIDR(fmt.Sprintf("192.168.%d.0/24", subnet))
 }
--- a/pkg/types/defaults/installconfig.go
+++ b/pkg/types/defaults/installconfig.go
@@ -42,11 +42,6 @@ func SetInstallConfigDefaults(c *types.InstallConfig) {
 				{CIDR: *libvirtdefaults.DefaultMachineCIDR},
 			}
 		}
-		if c.Platform.PowerVS != nil {
-			c.Networking.MachineNetwork = []types.MachineNetworkEntry{
-				{CIDR: *powervsdefaults.DefaultMachineCIDR},
-			}
-		}
 	}
 	if c.Networking.NetworkType == "" {
 		c.Networking.NetworkType = defaultNetworkType
@@ -122,6 +117,9 @@ func SetInstallConfigDefaults(c *types.InstallConfig) {
 		}
 	case c.Platform.PowerVS != nil:
 		powervsdefaults.SetPlatformDefaults(c.Platform.PowerVS)
+		c.Networking.MachineNetwork = []types.MachineNetworkEntry{
+			{CIDR: *powervsdefaults.DefaultMachineCIDR},
+		}
 	case c.Platform.None != nil:
 		nonedefaults.SetPlatformDefaults(c.Platform.None)
 	case c.Platform.Nutanix != nil:


clnperez:
@alishaIBM is this default CIDR in the documentation somewhere? IIRC it is, and we'll need to remove the fact that there's a default if not done already.

pkg/types/powervs/defaults/platform.go
```diff
@@ -1,16 +1,27 @@
 package defaults
 
 import (
+	"crypto/rand"
```
r4f4, line 4: Shouldn't a pseudo-random number from `math/rand` be enough here?

pkg/types/powervs/defaults/platform.go
```diff
@@ -1,16 +1,27 @@
...
 
 // SetPlatformDefaults sets the defaults for the platform.
 func SetPlatformDefaults(p *powervs.Platform) {
+	n, err := rand.Int(rand.Reader, big.NewInt(253))
+	if err != nil {
+		panic(err)
+	}
+	subnet := n.Int64()
+	// MustParseCIDR parses a CIDR from its string representation. If the parse fails, the function will panic.
+	DefaultMachineCIDR = ipnet.MustParseCIDR(fmt.Sprintf("192.168.%d.0/24", subnet))
```
r4f4, line 26:
```suggestion
    r := rand.New(rand.NewSource(time.Now().UnixNano()))
	subnet := r.Intn(255)
	// MustParseCIDR parses a CIDR from its string representation. If the parse fails, the function will panic.
	DefaultMachineCIDR = ipnet.MustParseCIDR(fmt.Sprintf("192.168.%d.0/24", subnet))
```

pkg/types/powervs/defaults/platform.go
```diff
@@ -1,16 +1,27 @@
...
+
 	"github.com/openshift/installer/pkg/ipnet"
 	"github.com/openshift/installer/pkg/types/powervs"
 )
 
 var (
 	// DefaultMachineCIDR is the PowerVS default IP address space from
 	// which to assign machine IPs.
-	DefaultMachineCIDR = ipnet.MustParseCIDR("192.168.0.0/24")
+	DefaultMachineCIDR *ipnet.IPNet
```
r4f4, line 15: What about we still set the `192.168.0.0/24` value here in case this value is used before `SetPlatformDefaults`? Worst case scenario we get a subnet conflict instead of dereferencing a `nil` value.

r4f4 (changes requested):

pkg/types/powervs/defaults/platform.go
hamzy: The reason why I went with crypto/rand was because of the following error from lint:
Use of weak random number generator (math/rand instead of crypto/rand) (gosec)

r4f4: Ok. If you decide to go back to `math/rand` it's totally fine to ignore the linter's comment since this is not security-sensitive: `//nolint:gosec <reason here>`. Otherwise, `crypto/rand` is also fine and there should be negligible performance impact.

hamzy: Thanks for the ignore tip!  I'll leave it as it is for now :)

pkg/types/powervs/defaults/platform.go
r4f4: Yes. Ok.

hamzy:
@r4f4 Are there any other open issues for a review?

mjturek:
> 



> @alishaIBM is this default CIDR in the documentation somewhere? IIRC it is, and we'll need to remove the fact that there's a default if not done already.

Yep it is documented so this patch will require an update to the documentation explaining the behavior.

hamzy:
> > @alishaIBM is this default CIDR in the documentation somewhere? IIRC it is, and we'll need to remove the fact that there's a default if not done already.
> 
> Yep it is documented so this patch will require an update to the documentation explaining the behavior.

@mjturek which file contains this documentation?

pkg/types/powervs/defaults/platform.go
```diff
@@ -13,4 +17,11 @@ var (
```
mjturek, line 17: Worth keeping this as a sane default, but If I'm understanding correctly this value will always be overwritten by [this call to SetPlatformDefaults ](https://github.com/openshift/installer/blob/master/pkg/types/defaults/installconfig.go#L124)

Not worth submitting a new revision, but if you wind up fixing anything else here worth mentioning in a comment that this value will be overwritten before deployment.

mjturek:
> > > @alishaIBM is this default CIDR in the documentation somewhere? IIRC it is, and we'll need to remove the fact that there's a default if not done already.
> > 
> > 
> > Yep it is documented so this patch will require an update to the documentation explaining the behavior.
> 
> @mjturek which file contains this documentation?

https://github.com/openshift/openshift-docs/blob/main/modules/installation-configuration-parameters.adoc#LL544C7-L544C7

hamzy:
> > > > @alishaIBM is this default CIDR in the documentation somewhere? IIRC it is, and we'll need to remove the fact that there's a default if not done already.
> > > 
> > > 
> > > Yep it is documented so this patch will require an update to the documentation explaining the behavior.
> > 
> > 
> > @mjturek which file contains this documentation?
> 
> https://github.com/openshift/openshift-docs/blob/main/modules/installation-configuration-parameters.adoc#LL544C7-L544C7

openshift-docs is a separate component and would need a separate PR.

PR #7150, opened 2023-05-01
AGENT-498: Get iPXE script template kernel parameters from ISO

pawanpinjarkar:
Read the default  kernel args from the `coreos/kargs.json` file from the base iso 
such as `ignition.firstboot ignition.platform.id=metal` and add it to the iPXE script.

### DIFF
--- a/pkg/asset/agent/image/agentpxefiles.go
+++ b/pkg/asset/agent/image/agentpxefiles.go
@@ -1,6 +1,7 @@
 package image
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -26,6 +27,11 @@ type AgentPXEFiles struct {
 	cpuArch     string
 	tmpPath     string
 	ipxeBaseURL string
+	kernelArgs  string
+}
+
+type coreOSKargs struct {
+	DefaultKernelArgs string `json:"default"`
 }
 
 var _ asset.WritableAsset = (*AgentPXEFiles)(nil)
@@ -65,6 +71,11 @@ func (a *AgentPXEFiles) Generate(dependencies asset.Parents) error {
 	a.cpuArch = agentArtifacts.CPUArch
 	a.ipxeBaseURL = strings.Trim(agentconfig.Config.IPxeBaseURL, "/")
 
+	kernelArgs, err := getKernelArgs(filepath.Join(a.tmpPath, "coreos", "kargs.json"))
+	if err != nil {
+		return err
+	}
+	a.kernelArgs = kernelArgs
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -118,6 +121,32 @@ func (a *AgentPXEFiles) PersistToFile(directory string) error {
...
 	}
 
+	tmpdir, err := os.MkdirTemp("", pxeAssetsPath)
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tmpdir)
+
+	dstfilename := filepath.Join(tmpdir, "config.cfg")
+	err = a.extractPXEFileFromISO(a.isoPath, "EFI/redhat/grub.cfg", dstfilename)
```
andfasano, line 131: What about (logically) post-poning this PR after https://github.com/openshift/installer/pull/7185, so that it could be possible to read directly the artifacts already extracted in the temp dir

andfasano: Another point: wouldn't be recommended to read `/coreos/kargs.json` for extracting the kernel parameters? It should be also simpler to parse, being a json file, ie something like: 

```
{
  "default": "coreos.liveiso=rhcos-414.92.202305090606-0 ignition.firstboot ignition.platform.id=metal",
  "files": [
    {
      "offset": 968,
      "path": "EFI/redhat/grub.cfg"
    },
    {
      "offset": 1865,
      "path": "isolinux/isolinux.cfg"
    }
  ],
  "size": 1112
}
```

pawanpinjarkar: Done

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -118,6 +131,51 @@ func (a *AgentPXEFiles) PersistToFile(directory string) error {
 		return err
 	}
 
+	tmpdir, err := os.MkdirTemp("", pxeAssetsPath)
```
andfasano, line 134: A couple of points about this block:
* For the sake of clarity, it'd be recommended to extract the code specific to the kernel args into its own function
* Most of the current logic should be moved into the `Generate` method (at least up to the extraction of the relevant data, in this case I'd to the unmarshaling point). The `PersistToFile` method should contain only the logic relevant for writing the current asset data to the disk

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -118,6 +131,51 @@ func (a *AgentPXEFiles) PersistToFile(directory string) error {
...
+
+	data, err := ioutil.ReadAll(fileReader)
+	if err != nil {
+		return err
+	}
+
+	var configFile CoreOSKargs
+	json.Unmarshal(data, &configFile)
+
+	var file filesystem.File
```
andfasano, line 154: I'm not sure that this whole part is really needed: the `default` field of `kargs.json` already contains the desired data (except for the `coreos.liveiso`)

pawanpinjarkar: Discussed and confirmed parsing the default value from kargs.json

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -179,3 +194,26 @@ boot
 
 	return nil
 }
+
+func getKernelArgs(isoPath, file string) (string, error) {
+	fileReader, err := isoeditor.GetFileFromISO(isoPath, file)
```
andfasano, line 199: Couldn't be retrieved also this one from the tmp path?

pawanpinjarkar: Done

PR #7151, opened 2023-05-01
CORS-2631: Add additional security group ids in AWS

barbacbd:
Updates support [CORS-2631](https://issues.redhat.com//browse/CORS-2631) and [CORS-2632](https://issues.redhat.com//browse/CORS-2632)

- [x] Add install config option for additionalSecurityGroupIDs
- [x] Validate Security Groups - check that subnet ids exist
- [x] Custom Security Group Install config validation by querying API
- [x] Tag User Supplied Security Groups with Shared
- [x] Add (sorted) list of Security Group IDs provided by the user for compute node(s)
- [x] Add (sorted) list of Security Group IDs provided by the user for control plane node(s)

### DIFF
--- a/pkg/asset/installconfig/aws/ec2.go
+++ b/pkg/asset/installconfig/aws/ec2.go
@@ -7,17 +7,16 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
-	"k8s.io/apimachinery/pkg/util/sets"
 )
 
-// GetVpcsFromSecurityGroups will get the list of VPCs that are attached to the security groups with the ids provided.
-func GetVpcsFromSecurityGroups(ctx context.Context, session *session.Session, securityGroupIDs []string, region string) ([]string, error) {
+// DescribeSecurityGroups returns the list of ec2 Security Groups that contain the group id and vpc id.
+func DescribeSecurityGroups(ctx context.Context, session *session.Session, securityGroupIDs []string, region string) ([]*ec2.SecurityGroup, error) {
 	client := ec2.New(session, aws.NewConfig().WithRegion(region))
 
 	sgIDPtrs := []*string{}
 	for _, sgid := range securityGroupIDs {
-		sgidAlias := sgid
-		sgIDPtrs = append(sgIDPtrs, &sgidAlias)
+		sgid := sgid
+		sgIDPtrs = append(sgIDPtrs, &sgid)
 	}
 
 	cctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
@@ -27,10 +26,5 @@ func GetVpcsFromSecurityGroups(ctx context.Context, session *session.Session, se
 	if err != nil {
 		return nil, err
 	}
-
-	vpcIds := sets.New[string]()
-	for _, sg := range sgOutput.SecurityGroups {
-		vpcIds.Insert(*sg.VpcId)
-	}
-	return sets.List(vpcIds), nil
+	return sgOutput.SecurityGroups, nil
 }
--- a/pkg/asset/installconfig/aws/ec2.go
+++ b/pkg/asset/installconfig/aws/ec2.go
@@ -1,0 +1,36 @@
+package aws
+
+import (
+	"context"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// GetVpcsFromSecurityGroups will get the list of VPCs that are attached to the security groups with the ids provided.
+func GetVpcsFromSecurityGroups(ctx context.Context, session *session.Session, securityGroupIDs []string, region string) ([]string, error) {
+	client := ec2.New(session, aws.NewConfig().WithRegion(region))
+
+	sgIDPtrs := []*string{}
+	for _, sgid := range securityGroupIDs {
+		sgidAlias := sgid
+		sgIDPtrs = append(sgIDPtrs, &sgidAlias)
+	}
+
+	cctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+
+	sgOutput, err := client.DescribeSecurityGroupsWithContext(cctx, &ec2.DescribeSecurityGroupsInput{GroupIds: sgIDPtrs})
+	if err != nil {
+		return nil, err
+	}
+
+	vpcIds := sets.New[string]()
+	for _, sg := range sgOutput.SecurityGroups {
+		vpcIds.Insert(*sg.VpcId)
+	}
+	return sets.List(vpcIds), nil
+}
--- a/pkg/asset/installconfig/aws/validation.go
+++ b/pkg/asset/installconfig/aws/validation.go
@@ -255,6 +255,36 @@ func validateMachinePool(ctx context.Context, meta *Metadata, fldPath *field.Pat
 			allErrs = append(allErrs, field.Invalid(fldPath.Child("type"), pool.InstanceType, errMsg))
 		}
 	}
+
+	allErrs = append(allErrs, validateSecurityGroupIDs(ctx, meta, fldPath.Child("additionalSecurityGroupIDs"), platform, pool)...)
+
+	return allErrs
+}
+
+func validateSecurityGroupIDs(ctx context.Context, meta *Metadata, fldPath *field.Path, platform *awstypes.Platform, pool *awstypes.MachinePool) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if len(pool.AdditionalSecurityGroupIDs) == 0 {
+		return allErrs
+	}
+
+	vpcs, err := GetVpcsFromSecurityGroups(ctx, meta.session, pool.AdditionalSecurityGroupIDs, platform.Region)
+	if err != nil {
+		return append(allErrs, field.Invalid(fldPath, pool.AdditionalSecurityGroupIDs, err.Error()))
+	}
+
+	if len(vpcs) != 1 {
+		return append(allErrs, field.Invalid(fldPath, vpcs, "exactly 1 vpc should be found"))
+	}
+
+	vpcFromSecurityGroups := vpcs[0]
+	vpc, err := meta.VPC(ctx)
+	if err == nil {
+		if vpc != vpcFromSecurityGroups {
+			errMsg := fmt.Sprintf("vpcs do not match %s, %s", vpcFromSecurityGroups, vpc)
+			allErrs = append(allErrs, field.Invalid(fldPath, vpcFromSecurityGroups, errMsg))
+		}
+	}
 	return allErrs
 }
 
--- a/pkg/asset/installconfig/aws/validation.go
+++ b/pkg/asset/installconfig/aws/validation.go
@@ -256,7 +256,9 @@ func validateMachinePool(ctx context.Context, meta *Metadata, fldPath *field.Pat
 		}
 	}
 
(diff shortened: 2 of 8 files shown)


pkg/types/aws/machinepool.go
```diff
@@ -74,6 +80,10 @@ func (a *MachinePool) Set(required *MachinePool) {
 	if required.IAMRole != "" {
 		a.IAMRole = required.IAMRole
 	}
+
+	if required.AdditionalSecurityGroupIDs != nil {
+		a.AdditionalSecurityGroupIDs = append(required.AdditionalSecurityGroupIDs[:0:0], required.AdditionalSecurityGroupIDs...)
```
r4f4, line 85: Interesting. So this guarantees that `a.AdditionalSecurityGroupIDs` won't be `nil` if `required.AdditionalSecurityGroupIDs` has 0 elements which would be the case if we did
 ```go
a.AdditionalSecurityGroupIDs = append([]string(nil), required.AdditionalSecurityGroupIDs...)
```
We could use `copy` since we know `required.AdditionalSecurityGroupIDs` is not `nil` but it'd result in 2 lines instead of 1:
```go
a.AdditionalSecurityGroupIDs = make([]string, len(required.AdditionalSecurityGroupIDs))
copy(a.AdditionalSecurityGroupIDs, required.AdditionalSecurityGroupIDs)
```

pkg/types/aws/validation/machinepool.go
```diff
@@ -47,6 +47,17 @@ func ValidateMachinePool(platform *aws.Platform, p *aws.MachinePool, fldPath *fi
...
+
+	return allErrs
+}
+
+func validateSecurityGroups(platform *aws.Platform, p *aws.MachinePool, fldPath *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if len(p.AdditionalSecurityGroupIDs) > 0 && len(platform.Subnets) <= 0 {
+		allErrs = append(allErrs, field.Required(fldPath.Child("additionalSecurityGroupID"), "subnets must be provided for security groups"))
+	}
```
r4f4, line 60: Do we also want to validate that the IDs provided are in the format `sg-xxx`? Maybe not here but in the install-config validation?

r4f4 (commented):
/lgtm

pkg/asset/installconfig/aws/ec2.go
```diff
@@ -0,0 +1,36 @@
...
+
+	cctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+
+	sgOutput, err := client.DescribeSecurityGroupsWithContext(cctx, &ec2.DescribeSecurityGroupsInput{GroupIds: sgIDPtrs})
+	if err != nil {
+		return nil, err
+	}
+
+	vpcIds := sets.NewString()
```
r4f4, line 31: String sets is deprecated [1]. We should be using the Generics syntax now:
```suggestion
	vpcIds := sets.New[string]()
```

[1] https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#String

pkg/asset/installconfig/aws/validation.go
```diff
@@ -255,6 +255,34 @@ func validateMachinePool(ctx context.Context, meta *Metadata, fldPath *field.Pat
...
+
+	allErrs = append(allErrs, validateSecurityGroupIDs(ctx, meta, fldPath.Child("additionalSecurityGroupIDs"), platform, pool)...)
+
+	return allErrs
+}
+
+func validateSecurityGroupIDs(ctx context.Context, meta *Metadata, fldPath *field.Path, platform *awstypes.Platform, pool *awstypes.MachinePool) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if len(pool.AdditionalSecurityGroupIDs) > 0 {
```
r4f4, line 267: Nit: If we invert the check we can reduce the indent level by 1
```suggestion
    // Nothing to validate
	if len(pool.AdditionalSecurityGroupIDs) == 0 {
	    return allErrs
	}
```

r4f4 (commented):
/lgtm

pkg/asset/installconfig/aws/ec2.go
```diff
@@ -0,0 +1,36 @@
...
+	defer cancel()
+
+	sgOutput, err := client.DescribeSecurityGroupsWithContext(cctx, &ec2.DescribeSecurityGroupsInput{GroupIds: sgIDPtrs})
+	if err != nil {
+		return nil, err
+	}
+
+	vpcIds := sets.New[string]()
+	for _, sg := range sgOutput.SecurityGroups {
+		vpcIds.Insert(*sg.VpcId)
```
r4f4, line 33: Any chance VpcId could be `nil` and we'd be dereferencing a `nil` pointer?

pkg/asset/installconfig/aws/validation.go
```diff
@@ -255,6 +255,36 @@ func validateMachinePool(ctx context.Context, meta *Metadata, fldPath *field.Pat
...
+		return allErrs
+	}
+
+	vpcs, err := GetVpcsFromSecurityGroups(ctx, meta.session, pool.AdditionalSecurityGroupIDs, platform.Region)
+	if err != nil {
+		return append(allErrs, field.Invalid(fldPath, pool.AdditionalSecurityGroupIDs, err.Error()))
+	}
+
+	if len(vpcs) != 1 {
+		return append(allErrs, field.Invalid(fldPath, vpcs, "exactly 1 vpc should be found"))
```
r4f4, line 277: Maybe we can expand the error message a bit more:
```suggestion
        errMsg := fmt.Sprintf("exactly 1 vpc expected; found %d instead", len(vpcs))
		return append(allErrs, field.Invalid(fldPath, vpcs, errMsg))
```

r4f4, line 277: Maybe we can expand the error message a bit more:
```suggestion
        errMsg := fmt.Sprintf("exactly 1 vpc expected; found %d instead", len(vpcs))
		return append(allErrs, field.Invalid(fldPath, vpcs, errMsg))
```

pkg/asset/machines/aws/machines.go
```diff
@@ -170,11 +170,30 @@ func Machines(clusterID string, region string, subnets map[string]string, pool *
...
+
+	securityGroups := []machineapi.AWSResourceReference{{
+		Filters: []machineapi.Filter{{
+			Name:   "tag:Name",
+			Values: []string{fmt.Sprintf("%s-%s-sg", in.clusterID, in.role)},
+		}},
+	}}
+	securityGroupsFromPool := []machineapi.AWSResourceReference{}
+	for _, sgID := range pool.AdditionalSecurityGroupIDs {
+		tmpID := sgID
```
r4f4, line 187: Nit: btw, this is also valid and makes the linter happy: 
```suggestion
		sgID := sgID
```
it creates a new `sgID` variable with a different scope therefore it's safe to take its address.

r4f4 (commented):
/lgtm

pkg/asset/installconfig/aws/validation.go
```diff
@@ -255,6 +255,36 @@ func validateMachinePool(ctx context.Context, meta *Metadata, fldPath *field.Pat
...
+	}
+
+	vpcFromSecurityGroups := vpcs[0]
+	vpc, err := meta.VPC(ctx)
+	if err == nil {
+		if vpc != vpcFromSecurityGroups {
+			errMsg := fmt.Sprintf("vpcs do not match %s, %s", vpcFromSecurityGroups, vpc)
```
patrickdillon: Error message could be slightly more clear I think. Something like `security group %q is not associated with provided VPC`

pkg/asset/machines/aws/machines.go
```diff
@@ -66,7 +66,7 @@ func Machines(clusterID string, region string, subnets map[string]string, pool *
 			imds:           mpool.EC2Metadata,
 			userTags:       userTags,
 			publicSubnet:   false,
```
sadasu, line 68: Why not add these additional security groups to the MachineProviderInput?

pkg/asset/machines/aws/machines.go
```diff
@@ -170,11 +170,30 @@ func Machines(clusterID string, region string, subnets map[string]string, pool *
...
+
+	securityGroups := []machineapi.AWSResourceReference{{
+		Filters: []machineapi.Filter{{
+			Name:   "tag:Name",
+			Values: []string{fmt.Sprintf("%s-%s-sg", in.clusterID, in.role)},
+		}},
+	}}
+	securityGroupsFromPool := []machineapi.AWSResourceReference{}
+	for _, sgID := range pool.AdditionalSecurityGroupIDs {
+		sgID := sgID
```
sadasu, line 187: Do we need this temporary variable sgID (the one on the left of := ) ?

sadasu (commented):
Commits look great hence wondering why the 2nd one is still in WIP. Some comments inline...nothing major.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -274,17 +274,21 @@ func validateSecurityGroupIDs(ctx context.Context, meta *Metadata, fldPath *fiel
 	}
 
 	if len(vpcs) != 1 {
-		return append(allErrs, field.Invalid(fldPath, vpcs, "exactly 1 vpc should be found"))
+		errMsg := fmt.Sprintf("exactly 1 vpc expected; found %d instead", len(vpcs))
```
sadasu, line 277: Can changes to pkg/asset/installconfig/aws/validation.go be part of commit 2?

pkg/asset/machines/aws/machines.go
r4f4: @sadasu yes because we're storing the address of the `sgID` in the following lines and the outer `sgID` pointed to value changes at each iteration of the loop. Without this, the linter will complain as well.

pkg/asset/installconfig/aws/validation.go
```diff
@@ -255,6 +255,36 @@ func validateMachinePool(ctx context.Context, meta *Metadata, fldPath *field.Pat
...
+	if len(pool.AdditionalSecurityGroupIDs) == 0 {
+		return allErrs
+	}
+
+	vpcs, err := GetVpcsFromSecurityGroups(ctx, meta.session, pool.AdditionalSecurityGroupIDs, platform.Region)
+	if err != nil {
+		return append(allErrs, field.Invalid(fldPath, pool.AdditionalSecurityGroupIDs, err.Error()))
+	}
+
+	if len(vpcs) != 1 {
```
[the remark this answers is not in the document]
patrickdillon: Oh, I see. When I asked the question, I thought there was a one-to-many relationship from sg to vpc.

This seems like a slightly roundabout way of saying the provided security groups do not belong to the provided vpc. I think the error message would be more helpful to indicate to the user "this sg x does not belong to vpc y"

pkg/asset/installconfig/aws/validation.go
```diff
@@ -255,6 +255,41 @@ func validateMachinePool(ctx context.Context, meta *Metadata, fldPath *field.Pat
...
+	}
+
+	vpcFromSecurityGroups := vpcs[0]
+	vpc, err := meta.VPC(ctx)
+	if err != nil {
+		errMsg := fmt.Sprintf("could not determine if securityGroups VPC matches the cluster's: %s", err.Error())
+		return append(allErrs, field.Invalid(fldPath, vpcFromSecurityGroups, errMsg))
+	}
+	if vpc != vpcFromSecurityGroups {
+		errMsg := fmt.Sprintf("vpcs do not match %s, %s", vpcFromSecurityGroups, vpc)
```
patrickdillon, line 289: on a similar note to my other comment, this error message could be a little more clear. I think the same suggestion as the other (it seems like these validations could be combined): "sg x is associated with vpc y not the provided vpc z"

r4f4 (commented):
/lgtm

r4f4 (commented):
/approve

PR #7153, opened 2023-05-02
CFE-687: Apply user defined labels on created gcp resources

bharath-b-rh:
PR is for applying OCP default label ("kubernetes-io-cluster-${cluster_id}" = "owned") and user defined labels to all gcp resources created.

### DIFF
--- a/data/data/gcp/bootstrap/main.tf
+++ b/data/data/gcp/bootstrap/main.tf
@@ -1,10 +1,4 @@
 locals {
-  labels = merge(
-    {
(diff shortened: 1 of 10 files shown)


barbacbd (commented):
I believe that you can add the labels to the boot_disk in cluster/master/main.tf. Can we look into that one?

bharath-b-rh:
> I believe that you can add the labels to the boot_disk in cluster/master/main.tf. Can we look into that one?

It's supported, I had missed it. Made the changes and tested it as well, will update the PR. Thank you!

PR #7155, opened 2023-05-02
WIP Ocpbugs 11736

barbacbd:
** Check for private zone. If one exists in the project then the private zone
    will be used.
** validate the private zone by checking for existing records.
** update manifests/dns to try to find the private zone and use the predefined name if that
could not be found
** update terraform variables to accept a prviate zone name
** updated tf code to only create the private zone if one was not provided
** updated tf code to create zone records in the correct private zone based
on whether a zone name was provided or one was created.


patrickdillon (commented):
The commit title for 3d057cd63d2df322790deb75f30dfffaf05f7299 "GCP Private Zone is no longer required" seems inaccurate. Should it be reworded?

I left a bunch of nitpicks. Besides that there are two main things to consider:
1. This functionality should, at least initially, only be enabled for gcp xpn installs
2. Consistent use of functions to get zones. Sometimes we're using `GetZones` and sometimes we're using `GetPublicZone`/GetPrivateZone`

jianli-wei:
> ** Check for private zone. If one exists in the project then the private zone
> will be used.
> ** validate the private zone by checking for existing records.

@barbacbd Could you tell how to specify the private zone in install-config.yaml? e.g. the field name, and its sample value? I tried with a build having the PR, `openshift-install explain installconfig.platform.gcp` doesn't tell such a field. Also tried with `platform.gcp.privateZoneDomain`, but there's warning like 'unknown field "privateZoneDomain"' when trying to "create cluster" from the updated install-config.yaml. Thanks!

barbacbd:
> > ** Check for private zone. If one exists in the project then the private zone
> > will be used.
> > ** validate the private zone by checking for existing records.
> 
> @barbacbd Could you tell how to specify the private zone in install-config.yaml? e.g. the field name, and its sample value? I tried with a build having the PR, `openshift-install explain installconfig.platform.gcp` doesn't tell such a field. Also tried with `platform.gcp.privateZoneDomain`, but there's warning like 'unknown field "privateZoneDomain"' when trying to "create cluster" from the updated install-config.yaml. Thanks!

@jianli-wei this will require docs changes, but the domain for the private cluster should be (if precreated) <cluster-name>.<base-domain>. If one exists matching this domain then it will be used, other wise the installer will create one for you.

jianli-wei:
> @jianli-wei this will require docs changes, but the domain for the private cluster should be (if precreated) .. If one exists matching this domain then it will be used, other wise the installer will create one for you.

@barbacbd please see my [comment](https://issues.redhat.com/browse/OCPBUGS-11736?focusedId=22226625&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22226625) in the bug, thanks!

jianli-wei (approved):
/label qe-approved
/lgtm

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -118,25 +118,49 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
+		return field.InternalError(field.NewPath("baseDomain"), err)
+	}
+	return checkRecordSets(client, ic, zone)
+}
+
+// ValidatePrivateDNSZone ensure no pre-existing DNS record exists in the private dns zone
+// matching the name that will be used for this installation.
+func ValidatePrivateDNSZone(client API, ic *types.InstallConfig) *field.Error {
+	if ic.GCP.Network == "" {
+		logrus.Warnf("Not searching for Private DNS Zone")
```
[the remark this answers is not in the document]
patrickdillon: This was marked resolved, but the original code remains with no comment.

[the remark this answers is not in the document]
patrickdillon: Nevermind, I think you just need to squash the commits. Let me take a closer look.

pkg/asset/installconfig/gcp/dns.go
```diff
@@ -9,36 +9,9 @@ import (
...
-	}
-	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
-	defer cancel()
-
-	dnsZone, err := client.GetDNSZone(ctx, project, baseDomain, isPublic)
-	if err != nil {
-		return nil, err
-	}
-	return dnsZone, nil
-}
```
patrickdillon: It looks like this code (or some of it) was introduced in a previous commit in this PR. So those changes should be squashed into at most one commit.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -131,20 +140,19 @@ func ValidatePreExistingPublicDNS(client API, ic *types.InstallConfig) *field.Er
...
-	return checkRecordSets(client, ic, zone)
+	return checkRecordSets(client, ic, zone, []string{apiRecordType(ic)})
 }
 
 // ValidatePrivateDNSZone ensure no pre-existing DNS record exists in the private dns zone
 // matching the name that will be used for this installation.
 func ValidatePrivateDNSZone(client API, ic *types.InstallConfig) *field.Error {
-	if ic.GCP.Network == "" {
-		logrus.Warnf("Not searching for Private DNS Zone")
+	if ic.GCP.Network == "" && ic.GCP.NetworkProjectID != "" {
```
patrickdillon, line 149: I think we want 
```go
if ic.GCP.Network == "" || ic.GCP.NetworkProjectID == ""
```
wdyt?

patrickdillon:
This is looking pretty close, but I think some of the commits may need to be squashed and I spotted a couple of small but significant errors that need a quick fix. I will also try to give a manual test soon.

pkg/asset/cluster/tfvars.go
```diff
@@ -436,13 +437,22 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
 
+		privateZoneName := ""
+		privateZone, err := client.GetDNSZone(ctx, installConfig.Config.GCP.ProjectID, installConfig.Config.ClusterDomain(), false)
+		if err != nil {
+			return errors.Wrapf(err, "failed to get GCP private zone")
+		}
+		if privateZone != nil {
+			privateZoneName = privateZone.Name
+		}
+
```
patrickdillon, line 455: In order to only enable this functionality with GCP XPN, I think we need to also wrap this code block with

```go
// Using an existing hosted zone is only supported in GCP XPN scenarios
if installconfig.Config.GCP.NetworkProjectID !== "" {
```

pkg/asset/installconfig/gcp/client.go
```diff
@@ -139,9 +139,9 @@ func (c *Client) GetDNSZoneByName(ctx context.Context, project, zoneName string)
 
 }
 
-// GetPublicDNSZone returns a public DNS zone for a basedomain.
-func (c *Client) GetPublicDNSZone(ctx context.Context, project, baseDomain string) (*dns.ManagedZone, error) {
-	ctx, cancel := context.WithTimeout(context.TODO(), 1*time.Minute)
+// GetDNSZone returns a DNS zone for a basedomain.
+func (c *Client) GetDNSZone(ctx context.Context, project, baseDomain string, isPublic bool) (*dns.ManagedZone, error) {
```
r4f4, line 143: NIT: a `bool` is fine but it would be nicer if we defined a proper enum type making it obvious in the code if the return matches the intention:
```suggestion
func (c *Client) GetDNSZone(ctx context.Context, project, baseDomain string, zoneType DNSZoneType) (*dns.ManagedZone, error) {
```
```golang
type DNSZoneType int
const (
    DNSZonePublic DNSZoneType = iota
    DNSZonePrivate
)

[...]
    publicZone, err := client.GetDNSZone(ctx, installConfig.Config.GCP.ProjectID, installConfig.Config.BaseDomain, DNSZonePublic)
[...]
    privateZone, err := client.GetDNSZone(ctx, installConfig.Config.GCP.ProjectID, installConfig.Config.ClusterDomain(), DNSZonePrivate)
```
Hopefully it'd make easier to catch a mistake like: 
```golang
    publicZone, err := client.GetDNSZone(ctx, installConfig.Config.GCP.ProjectID, installConfig.Config.BaseDomain, DNSZonePrivate)
```

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -118,39 +127,84 @@ func validateInstanceTypes(client API, ic *types.InstallConfig) field.ErrorList
...
-		if strings.EqualFold(r.Name, record) {
-			errMsg := fmt.Sprintf("record %s already exists in DNS Zone (%s/%s) and might be in use by another cluster, please remove it to continue", record, ic.GCP.ProjectID, zone.Name)
-			return field.Invalid(field.NewPath("metadata", "name"), ic.ObjectMeta.Name, errMsg)
-		}
+		setOfReturnedRecords.Insert(r.Name)
+	}
+	preexistingRecords := sets.New[string](records...).Intersection(setOfReturnedRecords)
+
+	if preexistingRecords.Len() > 0 {
+		errMsg := fmt.Sprintf("record(s) %q already exists in DNS Zone (%s/%s) and might be in use by another cluster, please remove it to continue", preexistingRecords.UnsortedList(), ic.GCP.ProjectID, zone.Name)
```
r4f4, line 182:
```suggestion
		errMsg := fmt.Sprintf("record(s) %q already exists in DNS Zone (%s/%s) and might be in use by another cluster, please remove it to continue", sets.List(preexistingRecords), ic.GCP.ProjectID, zone.Name)
```
It would be better to print a sorted list here so if we ever return more than one record in the unit tests mocking, the list will be deterministic and we can check for them in the regex. It's working fine now because we're just returning a single value.

patrickdillon:
/approve

This LGTM but leaving off the tag in case you want to update from @r4f4's review

pkg/asset/installconfig/gcp/client.go
barbacbd: I will keep this one as a boolean for now. We can readdress in a new PR if this is a problem. This creates a cyclic import for the installer mock client. This could reside in the types but it wont be used anywhere else.

PR #7061, opened 2023-04-04
AGENT-555: Move Rendezvous Host config to separate file

zaneb:
Create a separate environment file, `/etc/assisted/rendezvous-host.env`,
to contain the `NODE_ZERO_IP` and URLs based on it.

PR #7157, opened 2023-05-03
AGENT-563: Create configuration image

bfournie:
Adds a new installer command - `openshift-install agent create config-image` that will:
- generate an ignition from the provided manifests
- retrieve the files necessary for configuration from ignition
- create a cpio archive containing the config files
- generate an ISO (agentconfig.noarch.iso) with only the cpio archive (configiso.gz)


pkg/asset/agent/configimage/configiso.go
```diff
@@ -0,0 +1,159 @@
...
+	dependencies.Get(ignition)
+
+	ca := image.NewCpioArchive()
+
+	// Walk the list of ignition files.
+	// For the ones to be included in the configimage add to cpio archive
+	for _, file := range ignition.Config.Storage.Files {
+		filename := file.Node.Path
+		mode := *file.FileEmbedded1.Mode
+		if includeFileInImage(filename) {
```
andfasano, line 75: I have a mixed feeling about keeping the logic of which file should be included in the config ISO here, far away from the Ignition asset: it's in that asset where we effectively decide all the files that will land in the final ignition. Keeping a duplicated list here increases the risk that in future a person will add a new file in the `/data/data/agent/files` folder forgetting to add it here.

pkg/asset/agent/configimage/configiso.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/vincent-petithory/dataurl"
+
+	"github.com/openshift/assisted-image-service/pkg/isoeditor"
+	"github.com/openshift/installer/pkg/asset"
+	"github.com/openshift/installer/pkg/asset/agent/image"
+)
+
+const (
+	configImageFilename = "agentconfig.iso"
```
andfasano, line 18: Is it the content bound to a specifich arch? If not, I'd suggest to use `agentconfig.noarch.iso`, more in line with the naming convention adopted also for the standard ISO

pkg/asset/agent/configimage/configiso.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	// Save the cpi archive to a temp file
+	tmpPath, err := os.MkdirTemp("", "agent")
+	if err != nil {
+		return err
+	}
+	a.tmpPath = tmpPath
+
+	output := filepath.Join(tmpPath, "configiso.gz")
```
andfasano, line 100: This name seems to be slightly misleading, this is a compressed cpio archive, not an iso?

pkg/asset/agent/configimage/configiso.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Name returns the human-friendly name of the asset.
+func (a *ConfigImage) Name() string {
+	return "Config Image"
```
andfasano, line 136: I'd suggest a more speaking name, ie `Agent Config ISO`

pkg/asset/agent/configimage/configiso.go
bfournie: Yeah its a good point, I originally had this file in `pkg/asset/agent/image/` but all the files there are associated with building an agent image iso so I made a separate dir for this config image. The files that end up in ignition are dispersed though, all of the yamls come from the `manifests` dir.
I don't really like the duplication of files names as it is now but to that point - the files that get included in the config image are there for a reason so I would expect that adding one would be a conscious change and not by pulling in a new file added to ignition.

Let me see if I can find a common place for including this list of files.

pkg/asset/agent/configimage/configiso.go
bfournie: Good idea, will change.

pkg/asset/agent/configimage/configiso.go
bfournie: Yes I will fix that name.

pkg/asset/agent/configimage/configiso.go
bfournie: Sure, will fix.

pkg/asset/agent/configimage/configiso.go
```diff
@@ -0,0 +1,160 @@
...
+	if a.tmpPath == "" {
+		return errors.New("cannot generate Config image due to configuration errors")
+	}
+
+	configImageFile := filepath.Join(directory, configImageFilename)
+
+	// Remove symlink if it exists
+	os.Remove(configImageFile)
+
+	logrus.Infof("In PersistToFile, creating %s", configImageFile)
```
pawanpinjarkar, line 126: Let's remove the name of the internal function `PersistToFile` being logged here

pawanpinjarkar (changes requested):
Also, can you add an integration test?

pkg/asset/agent/configimage/configiso.go
bfournie: Will remove

bfournie:
> Also, can you add an integration test?

Coming soon...

pkg/asset/agent/configimage/configiso.go
```diff
@@ -0,0 +1,159 @@
...
+	extraManifests              string = "/etc/assisted/extra-manifests"
+	hostConfigFiles             string = "/etc/assisted/hostconfig"
+	hostNameFiles               string = "/etc/assisted/hostnames"
+	kubeadminPasswordHash       string = "/opt/agent/tls/kubeadmin-password.hash" //nolint:gosec // not hardcoded credentials
+	kubeconfigTLSCerts          string = "/opt/agent/tls/admin-kubeconfig-signer"
+	kubeapiserverTLSCerts       string = "/opt/agent/tls/kube-apiserver"
+)
+
+// TODO - need to add rendezvous host file when available.
+var fileFilters = []string{
```
rwsu, line 35: I think some of the files under data/agent/files like /usr/local/bin/start-cluster-installation.sh.template, /usr/local/share/assisted-service.env.template, and /usr/share/assisted-service/images.env.template would also need to be included. Also create-cluster-and-infraenv.service.template.

cmd/openshift-install/testdata/agent/configimage/manifests/default_manifests.txt
```diff
@@ -0,0 +1,145 @@
+# Verify that the basic manifests are properly generated within the /etc/assisted/manifests folder
+
+exec openshift-install agent create config-image --dir $WORK
+
+exists $WORK/agentconfig.noarch.iso
+
+isocmp agentconfig.noarch.iso /etc/assisted/manifests/agent-cluster-install.yaml expected/agent-cluster-install.yaml
```
rwsu, line 7: Would also be work checking that /etc/assisted/manifests/infraenv.yaml is not included in the config.iso.

rwsu (changes requested):

pkg/asset/agent/configimage/configiso.go
zaneb: #7062 and #7061 should take care of removing all the templating I think.
We'll need to make sure that this change is integrated with those ones though, so that the new environment files get passed through.

pkg/asset/agent/configimage/configiso.go
```diff
@@ -0,0 +1,159 @@
...
+
+	pullSecretManifest          string = "/etc/assisted/manifests/pull-secret.yaml" //nolint:gosec // not hardcoded credentials
+	nmStateConfigManifest       string = "/etc/assisted/manifests/nmstateconfig.yaml"
+	clusterDeploymentManifest   string = "/etc/assisted/manifests/cluster-deployment.yaml"
+	clusterImageSetManifest     string = "/etc/assisted/manifests/cluster-image-set.yaml"
+	agentClusterInstallManifest string = "/etc/assisted/manifests/agent-cluster-install.yaml"
+	extraManifests              string = "/etc/assisted/extra-manifests"
+	hostConfigFiles             string = "/etc/assisted/hostconfig"
+	hostNameFiles               string = "/etc/assisted/hostnames"
+	kubeadminPasswordHash       string = "/opt/agent/tls/kubeadmin-password.hash" //nolint:gosec // not hardcoded credentials
```
danielerez, line 29: IIUC, that's the hash of the generated kubeadmin [random pass](https://github.com/openshift/installer/blob/8d73989c1efe04448bf62bcbbf0b87b13dbc0a97/pkg/asset/password/password.go#LL44C29-L44C55)?
If so, how is the raw pass exposed to the user? on console?
Or is it something that can be specified by the user on config?

pkg/asset/agent/configimage/configiso.go
```diff
@@ -0,0 +1,159 @@
...
+
+const (
+	configImageFilename = "agentconfig.noarch.iso"
+	configCpioArchive   = "config.gz"
+
+	pullSecretManifest          string = "/etc/assisted/manifests/pull-secret.yaml" //nolint:gosec // not hardcoded credentials
+	nmStateConfigManifest       string = "/etc/assisted/manifests/nmstateconfig.yaml"
+	clusterDeploymentManifest   string = "/etc/assisted/manifests/cluster-deployment.yaml"
+	clusterImageSetManifest     string = "/etc/assisted/manifests/cluster-image-set.yaml"
+	agentClusterInstallManifest string = "/etc/assisted/manifests/agent-cluster-install.yaml"
```
danielerez, line 25: Is the sshKey specified by the user on install-config included here?

pkg/asset/agent/configimage/configiso.go
```diff
@@ -0,0 +1,159 @@
...
+	clusterDeploymentManifest,
+	clusterImageSetManifest,
+	agentClusterInstallManifest,
+	extraManifests,
+	hostConfigFiles,
+	hostNameFiles,
+	kubeadminPasswordHash,
+	kubeconfigTLSCerts,
+	kubeapiserverTLSCerts,
+}
```
danielerez, line 47: When is the kubeconfig file being created in this flow? How is it exposed to the user? Using console/ssh/etc?

pkg/asset/agent/configimage/configiso.go
bfournie: The sshKey is in the infraenv ZTP asset which was not planned on being included in the config ISO as its in the unconfigured ignition https://issues.redhat.com/browse/AGENT-558.

danielerez: But the sshKey is supposed to be specified by the user. I mean, the user won't be able to ssh into the machine without specifying the public ssh key.  So IIUC, it should be apart of the config-image API. At least that what I've understood from this explanation in the [ticket](https://issues.redhat.com/browse/AGENT-558):
`Configuration for the pull secret, image mirror, ssh public key, proxy, additional NTP sources, and rendezvous IP should be passed in as usual.`
Am I missing something? @zaneb

zaneb: There's _three_ places an ssh key can appear. They'll often contain the same key (and in the ABI we populate both from the key given in the install-config), but they don't have to.

1. The ssh key in the agent ISO's authorized_keys - this is used to allow debugging of the installation. This has to be in the ignition. In the factory-installed use case this will either not be possible, or a partner-supplied public key will have to be baked into the disk image.
2. The ssh key in the InfraEnv - this is used to generate a discovery image, but we don't do that in ABI so it is actually ignored.
3.  The ssh key in the AgentClusterInstall - this is used to allow debugging of the _cluster_ after installation.

So the answer to the original question is yes, the ssh key from the install-config will appear here in agent-cluster-install.yaml: https://github.com/openshift/installer/blob/master/pkg/asset/agent/manifests/agentclusterinstall.go#L145

cmd/openshift-install/agent.go
```diff
@@ -70,6 +71,18 @@ var (
...
 
+	agentConfigImageTarget = target{
+		name: "Agent Config Image",
+		command: &cobra.Command{
+			Use:   "config-image",
+			Short: "Generates an ISO containing configuration files only",
+			Args:  cobra.ExactArgs(0),
+		},
+		assets: []asset.WritableAsset{
+			&configimage.ConfigImage{},
```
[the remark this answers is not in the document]
danielerez: So how is kubeconfig exposed to the user after providing the config ISO. Are the details will be displayed to console? Or should the user ssh to the machine and look for it in a specific location?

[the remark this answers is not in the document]
zaneb: It's a file generated alongside the Config ISO

pkg/asset/agent/configimage/configiso.go
bfournie: Updated to generate the auth assets.

pkg/asset/agent/configimage/configiso.go
bfournie: In `auth/` (after last update)

pkg/asset/agent/configimage/configiso.go
bfournie: I've moved the file strings to the ignition asset. Having them there will also allow them to be used to template the load-config script in https://github.com/openshift/installer/pull/7200 so that only these files are copied from the config-image

cmd/openshift-install/agent.go
```diff
@@ -70,6 +71,21 @@ var (
 		},
 	}
 
+	agentConfigImageTarget = target{
+		name: "Agent Config Image",
+		command: &cobra.Command{
```
rwsu, line 76: @zaneb Should this command also be hidden?

bfournie: Per discussion with Zane, this command will be available for the customer to run and should not be hidden.

cmd/openshift-install/testdata/agent/configimage/manifests/default_manifests.txt
bfournie: Will add this as a follow-on as it looks like will be able to reuse ignitionContains() in https://github.com/openshift/installer/pull/7186 when that merges.

cmd/openshift-install/testdata/agent/configimage/manifests/default_manifests.txt
```diff
@@ -0,0 +1,145 @@
...
+status: {}
+-- expected/cluster-image-set.yaml --
+metadata:
+  creationTimestamp: null
+  name: openshift-was not built correctly
+  namespace: cluster0
+spec:
+  releaseImage: registry.ci.openshift.org/origin/release:4.13
+status: {}
+-- expected/infraenv.yaml --
```
rwsu, line 104: The infraenv.yaml shouldn't be in the config image. There isn't a check for it to see if it is in the config image, so I think this block can be removed to avoid confusion.

bfournie: Fixed

cmd/openshift-install/agent_internal_integration_test.go
```diff
@@ -121,12 +122,39 @@ func ignitionImgContains(ts *testscript.TestScript, neg bool, args []string) {
...
+
+	workDir := ts.Getenv("WORK")
+	isoPath, eFilePath := args[0], args[1]
+	isoPathAbs := filepath.Join(workDir, isoPath)
+
+	_, err := extractArchiveFile(isoPathAbs, eFilePath, "")
+	ts.Check(err)
+}
+
+// [!] archiveFileNames `isoPath` get the names of the archive files to use
```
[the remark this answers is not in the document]
bfournie: Fixed

rwsu (commented):
/lgtm

rwsu (commented):
/lgtm

rwsu (commented):
/lgtm

bfournie:
Tested in dev-scripts as part of https://github.com/openshift-metal3/dev-scripts/pull/1533

rwsu (commented):
/lgtm

PR #7159, opened 2023-05-03
images: decouple installer and installer-artifacts

patrickdillon:
The installer-artifacts image depends on the installer image, which means they cannot be built in parallel. The effect of this dependency is that the installer-artifacts image contains the linux amd64 builds in addition to the mac/linux amd64/arm64 builds present in the installer-artifacts image.

Breaking the dependency results in a more efficient build process because they can be built in parallel as well as no duplication of the linux amd64 binary in the release image (before this change the binary would be present in both images within the release image).

The risk, though, is that clients are inconsistently grabbing the linux amd64 binary from either image, rather than only the installer image, in which case we would break any client trying to get that binary from the installer-artifacts image.

Will open an accompanying PR in the release repo.

### DIFF
--- a/images/installer-artifacts/Dockerfile.rhel
+++ b/images/installer-artifacts/Dockerfile.rhel
@@ -30,10 +30,6 @@ COPY --from=macbuilder /go/src/github.com/openshift/installer/bin/openshift-inst
 COPY --from=macarmbuilder /go/src/github.com/openshift/installer/bin/openshift-install /usr/share/openshift/mac_arm64/openshift-install
 COPY --from=linuxbuilder /go/src/github.com/openshift/installer/bin/openshift-install /usr/share/openshift/linux_amd64/openshift-install
 COPY --from=linuxarmbuilder /go/src/github.com/openshift/installer/bin/openshift-install /usr/share/openshift/linux_arm64/openshift-install
-RUN mkdir /output && chown 1000:1000 /output
-USER 1000:1000
-ENV PATH /bin
-ENV HOME /output
-WORKDIR /output
+
 # We're not really an operator, we're just getting some data into the release image.
 LABEL io.openshift.release.operator=true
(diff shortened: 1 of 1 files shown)


images/installer-artifacts/Dockerfile.rhel
```diff
@@ -25,8 +25,15 @@ WORKDIR /go/src/github.com/openshift/installer
...
 COPY --from=macarmbuilder /go/src/github.com/openshift/installer/bin/openshift-install /usr/share/openshift/mac_arm64/openshift-install
 COPY --from=linuxbuilder /go/src/github.com/openshift/installer/bin/openshift-install /usr/share/openshift/linux_amd64/openshift-install
 COPY --from=linuxarmbuilder /go/src/github.com/openshift/installer/bin/openshift-install /usr/share/openshift/linux_arm64/openshift-install
+RUN mkdir /output && chown 1000:1000 /output
+USER 1000:1000
+ENV PATH /bin
+ENV HOME /output
+WORKDIR /output
+# We're not really an operator, we're just getting some data into the release image.
+LABEL io.openshift.release.operator=true
```
patrickdillon, line 39: This is mostly my cargo culting from the installer image: https://github.com/openshift/installer/blob/ba517b23497126614e6ea7aa2e23d4d735b7484b/images/installer/Dockerfile.ci#L15-L22

I'm not sure if this is actually necessary.

patrickdillon:
Regarding my concern
> we would break any client trying to get that binary from the installer-artifacts image.

It appears we would not break `oc`, which grabs the installer from the installer image.
https://github.com/openshift/oc/blob/master/pkg/cli/admin/release/extract_tools.go#L271-L274

patrickdillon:
Depends on https://github.com/openshift/release/pull/38975

/wip

This PR is intended for discussion.

patrickdillon:
/hold

as WIP. This PR is intended for discussion.

jhixson74:
I believe this is related: https://issues.redhat.com/browse/CORS-2606

patrickdillon:
> I believe this is related: https://issues.redhat.com/browse/CORS-2606

I don't see much connection here. This PR is about how we build and deliver the installer for different archs and OSes; and this is independent of the openshift release image. CORS-2606 is about breaking the dependency  between the installer and the release image, so this PR wouldn't affect that.

PR #7165, opened 2023-05-05
OCPBUGS-13107: openstack destroy: Limit Swift workers to 3

pierreprinetti:
down from 10. More than that killed RabbitMQ on a standalone OpenStack cloud.

Also add a smidge of logging information.


mandre:
> down from 10. More than that killed RabbitMQ on a standalone OpenStack cloud.

That's not good :grimacing:
Is there anything to report on openstack side to make it more robust?

mandre (commented):
/lgtm

stephenfin:
/lgtm

Yeah, it would be good to get logs from swift so we can report this. 10 connections isn't that excessive

PR #7168, opened 2023-05-08
OCPBUGS-9081: openstack destroy: account for BULK DELETE limits on object-storage

pierreprinetti:
Some object-storage instances may be set to have a limit to the LIST operation that is higher to the limit to the BULK DELETE operation. On those clouds, objects in the BULK DELETE call beyond the limit are silently ignored. As a consequence, the call to destroy the container fails and object deletion is re-queued after a growing waiting time, potentially dilating deletion by hours.

With this change, object bulk deletion is put in a loop. After checking that no errors were encountered, we reduce the BULK DELETE list by the number of processed objects, and send it back to the server. As a consequence, the object deletion routines should only complete when the container is emtpy, thus avoiding the 409 error that causes a retry.

### DIFF
--- a/pkg/destroy/openstack/openstack.go
+++ b/pkg/destroy/openstack/openstack.go
@@ -1099,21 +1099,31 @@ func deleteContainers(opts *clientconfig.ClientOpts, filter Filter, logger logru
 						return false, err
 					}
 					queue.Add(func() {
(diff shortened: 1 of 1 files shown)


gryf (approved):
/lgtm

PR #7169, opened 2023-05-08
CORS-2656: Remove context from cluster uninstaller struct

barbacbd:
** context structs should not be stored in another struct. Removed from GCP and altered the gcp destroy functionality to pass around the base context to all destroy functions and each will create a child off of the context.

### DIFF
--- a/pkg/destroy/gcp/gcp.go
+++ b/pkg/destroy/gcp/gcp.go
@@ -29,6 +29,7 @@ import (
 
 var (
 	defaultTimeout = 2 * time.Minute
+	longTimeout    = 10 * time.Minute
 )
 
 // ClusterUninstaller holds the various options for the cluster we want to delete
@@ -39,7 +40,6 @@ type ClusterUninstaller struct {
 	NetworkProjectID  string
 	PrivateZoneDomain string
 	ClusterID         string
-	Context           context.Context //nolint:containedctx
 
 	computeSvc *compute.Service
 	iamSvc     *iam.Service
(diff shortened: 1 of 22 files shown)


pkg/destroy/gcp/gcp.go
```diff
@@ -97,7 +95,7 @@ func (o *ClusterUninstaller) Run() (*types.ClusterQuota, error) {
 
 	o.cpusByMachineType = map[string]int64{}
 	req := o.computeSvc.MachineTypes.AggregatedList(o.ProjectID).Fields("items/*/machineTypes(name,guestCpus),nextPageToken")
-	if err := req.Pages(o.Context, func(list *compute.MachineTypeAggregatedList) error {
+	if err := req.Pages(ctx, func(list *compute.MachineTypeAggregatedList) error {
```
r4f4, line 98: This is going from a never-expiring context to a 2-minute-timeout context in which some time already passed. Since this is caching machine types by iterating through possible pages of data, maybe we should use a longer timeout context? Or maybe we create a child context with longer timeout just for this?

r4f4 (commented):
Thank you for working on this.

pkg/destroy/gcp/gcp.go
```diff
@@ -108,22 +108,25 @@ func (o *ClusterUninstaller) Run() (*types.ClusterQuota, error) {
 		return nil, errors.Wrap(err, "failed to cache machine types")
 	}
 
-	o.iamSvc, err = iam.NewService(ctx, options...)
+	// GCP suggests a context without a timeout for the new client, then API calls create a child context
+	serviceCtx := context.Background()
```
r4f4, line 112: Can't we just use `ctx` here and in the following clients creation?

r4f4 (commented):
/lgtm

barbacbd:
@deepsm007 I rebased looks like the issues are resolved but the LGTM was removed. Want me to see if someone can tag that really quick?

PR #7172, opened 2023-05-09
OCPBUGS-13300: masters on a single compute server group

EmilienM:
2.5 years ago we allowed to configure `serverGroupPolicy` in
install-config so a user could choose which Nova scheduling policy
to adopt for the machines.

However, if the masters were configured with AZ, Terraform would
create one ServerGroup in OpenStack (the one from master-0) but
configure the Machine providerSpec with different ServerGroups, one
per AZ. This was unwanted and now we want to use a single ServerGroup
for masters.

With compute AZ support, the users already have the possibility to
ensure that masters aren't on the same failure domain as others.
Also, even if there is less than 3 AZs (e.g. 2), the default
`soft-anti-affinity` server group policy would make Nova to
scheduling in best effort the machines on different hosts within a same
AZ.

Therefore, there is no need to configure the master machines with a
`serverGroup` per availability zone in their Machines.

Also, note that in OCP 4.14, CPMS will be enabled by default.
If a user has set multiple AZ for the controlPlane and upgrade from
4.13 to 4.14, CPMS will adopt the control plane and create a CPMS in
Inactive mode, with a single `serverGroup`. The `serverGroup` will
likely be the one from master-0, and this will be shared across all
control plane machines.
It'll be up to the user to set the CPMS to Active
and then the masters will be redeployed in the unique group for all
masters. They will never have a ServerGroup with "clusterID + role" name
because in previous releases we added the AZ name in it.

### DIFF
--- a/pkg/asset/machines/openstack/machines.go
+++ b/pkg/asset/machines/openstack/machines.go
@@ -174,9 +174,17 @@ func generateProvider(clusterID string, platform *openstack.Platform, mpool *ope
 	}
 
 	serverGroupName := clusterID + "-" + role
-	if failureDomain.ComputeAvailabilityZone != "" {
+	// We initially used the AZ name as part of the server group name for the masters
+	// but we realized that it was not useful. Whether or not the AZ is specified, the
+	// masters will be spread across multiple hosts by default by the Nova scheduler
+	// (the policy can be changed via `serverGroupPolicy` in install-config.yaml).
+	// For the workers, we still use the AZ name as part of the server group name
+	// so the user can control the scheduling policy per AZ and change the MachineSets
+	// if needed on a day 2 operation.
+	if role == "worker" && failureDomain.ComputeAvailabilityZone != "" {
 		serverGroupName += "-" + failureDomain.ComputeAvailabilityZone
 	}
+
 	spec := machinev1alpha1.OpenstackProviderSpec{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: machinev1alpha1.GroupVersion.String(),
--- a/scripts/openstack/manifest-tests/server-groups/install-config.yaml
+++ b/scripts/openstack/manifest-tests/server-groups/install-config.yaml
@@ -10,6 +10,7 @@ controlPlane:
       serverGroupPolicy: anti-affinity
       zones:
       - zoneone
+      - zonetwo
   replicas: 3
 compute:
 - name: worker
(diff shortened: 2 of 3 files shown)


pkg/asset/machines/openstack/machines.go
```diff
@@ -174,9 +174,6 @@ func generateProvider(clusterID string, platform *openstack.Platform, mpool *ope
 	}
 
 	serverGroupName := clusterID + "-" + role
```
EmilienM, line 176: if folks want I can also add a comment here

pierreprinetti (changes requested):

pierreprinetti:
Interestingly, what you are coding was actually declared in the `manifest-test` suite, except that it was not tested with multiple control-plane zones.

You may want to patch, to cover your changes:

```diff
diff --git a/scripts/openstack/manifest-tests/server-groups/install-config.yaml b/scripts/openstack/manifest-tests/server-groups/install-config.yaml
index 63e5613c6c..080a625d67 100644
--- a/scripts/openstack/manifest-tests/server-groups/install-config.yaml
+++ b/scripts/openstack/manifest-tests/server-groups/install-config.yaml
@@ -10,6 +10,7 @@ controlPlane:
       serverGroupPolicy: anti-affinity
       zones:
       - zoneone
+      - zonetwo
   replicas: 3
 compute:
 - name: worker
```

EmilienM:
/hold

I need to apply Pierre's suggestion and also fix the commit message because we were never creating three SG but only one for masters, in TF.

There were only unique per AZ in the Machine.

mdbooth (commented):
Does this now match what we actually created with terraform?

pierreprinetti (approved):

mdbooth:
Yep, looks like we are indeed using whatever the SG was set to in the provider spec for master-0: https://github.com/openshift/installer/blob/9e2ccd12ed7d0d7278558b2e3a965a67c721bfc2/pkg/tfvars/openstack/openstack.go#L158-L158

scripts/openstack/manifest-tests/server-groups/install-config.yaml
```diff
@@ -10,6 +10,7 @@ controlPlane:
       serverGroupPolicy: anti-affinity
       zones:
       - zoneone
+      - zonetwo
```
mdbooth, line 13: @pierreprinetti What does adding this do?

I would expect a change like this to include test changes to be input and expected output. What's the context here?

pierreprinetti: This change adds the failing test that the rest of the code fixes.

The test says "all masters have the same server group". With this change, that test is executed with an install-config that gives masters two different AZs (rather than just one).

pierreprinetti:
/lgtm
/hold

@mdbooth please `/hold cancel` when you're OK with merging.

PR #7173, opened 2023-05-09
OCPBUGS-14917: PowerVS: Cleanup service instances for destroy cluster

hamzy:
New code has been added to create a new PowerVS service instance during create cluster.  Therefore, we need to clean it up during destroy cluster.

### DIFF
--- a/pkg/destroy/powervs/serviceinstance.go
+++ b/pkg/destroy/powervs/serviceinstance.go
@@ -1,0 +1,281 @@
+package powervs
+
+import (
+	"context"
+	"fmt"
+	"math"
+	gohttp "net/http"
+	"strings"
+	"time"
+
+	"github.com/IBM/go-sdk-core/v5/core"
+	"github.com/IBM/platform-services-go-sdk/resourcecontrollerv2"
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+const (
+	serviceInstanceTypeName = "service instance"
+
+	// resource ID for Power Systems Virtual Server in the Global catalog.
+	virtualServerResourceID = "abd259f0-9990-11e8-acc8-b9f54a8f1661"
+)
+
+// convertResourceGroupNameToID converts a resource group name/id to an id.
+func (o *ClusterUninstaller) convertResourceGroupNameToID(resourceGroupID string) (string, error) {
+	listResourceGroupsOptions := o.managementSvc.NewListResourceGroupsOptions()
+
+	resourceGroups, _, err := o.managementSvc.ListResourceGroups(listResourceGroupsOptions)
+	if err != nil {
+		return "", err
+	}
+
+	for _, resourceGroup := range resourceGroups.Resources {
+		if *resourceGroup.Name == resourceGroupID {
+			return *resourceGroup.ID, nil
+		} else if *resourceGroup.ID == resourceGroupID {
+			return resourceGroupID, nil
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/destroy/powervs/serviceinstance.go
```diff
@@ -0,0 +1,314 @@
...
+			}
+			if response != nil && response.StatusCode == gohttp.StatusNotFound {
+				o.Logger.Debugf("listServiceInstances: gohttp.StatusNotFound")
+				continue
+			} else if response != nil && response.StatusCode == gohttp.StatusInternalServerError {
+				o.Logger.Debugf("listServiceInstances: gohttp.StatusInternalServerError")
+				continue
+			}
+
+			if resourceInstance.Type == nil {
```
AshwinHIBM, line 118: Is there a scenario where the resourceInstance.Type can be nil?

pkg/destroy/powervs/serviceinstance.go
```diff
@@ -0,0 +1,314 @@
...
+				o.Logger.Debugf("listServiceInstances: sub type: nil")
+			} else {
+				o.Logger.Debugf("listServiceInstances: sub type: %v", *resourceInstance.SubType)
+			}
+			if resourceInstance.ResourcePlanID == nil {
+				o.Logger.Debugf("listServiceInstances: plan: nil")
+			} else {
+				o.Logger.Debugf("listServiceInstances: plan: %v", *resourceInstance.ResourcePlanID)
+			}
+			o.Logger.Debugf("listServiceInstances: TEST1 = %v", resourceInstance.Type != nil && *resourceInstance.Type == "service_instance")
```
AshwinHIBM, line 133: Can these debug statements be removed prior to merge since they're adding up quite a large amount of command-line output?

AshwinHIBM:
The code destroys the installer-provisioned service instance (+all other assets associated with the cluster) and does not destroy any other service instance as expected.

pkg/destroy/powervs/serviceinstance.go
hamzy: As it's a pointer, yes.

pkg/destroy/powervs/serviceinstance.go
hamzy: I removed some of them.  Since it is debug output, I am not worried about taking up a lot of output with potentially helpful debugging output.

hamzy:
@mjturek Good catch!  Ready for a review now...

pkg/destroy/powervs/serviceinstance.go
```diff
@@ -0,0 +1,281 @@
...
+		err = wait.ExponentialBackoffWithContext(ctx, backoff, func(context.Context) (bool, error) {
+			err2 := o.destroyServiceInstance(item)
+			if err2 == nil {
+				return true, err2
+			}
+			o.errorTracker.suppressWarning(item.key, err2, o.Logger)
+			return false, err2
+		})
+		if err != nil {
+			o.Logger.Fatal("destroyServiceInstances: ExponentialBackoffWithContext (destroy) returns ", err)
```
r4f4, line 246: `Fatal` calls `os.Exit(1)` after logging ([source](https://github.com/sirupsen/logrus#level-logging)) is that what you intend here?

hamzy: Yes.  The destroy code everywhere expects to work perfectly or fail immediately and not continue destroying things.

r4f4 (commented):
/approve

PR #7177, opened 2023-05-10
OCPBUGS-6639: pkg/types/azure/validation: validate resource groups

jhixson74:
Validate that baseDomainResourceGroup is not the same as networkResourceGroupName. networkResourceGroupName needs to be a separate resource group so that the created resources can be tracked outside of existing vnet resources.

https://issues.redhat.com/browse/OCPBUGS-6639

### DIFF
--- a/pkg/types/azure/validation/platform.go
+++ b/pkg/types/azure/validation/platform.go
@@ -70,6 +70,9 @@ func ValidatePlatform(p *azure.Platform, publish types.PublishingStrategy, fldPa
 			allErrs = append(allErrs, field.Required(fldPath.Child("networkResourceGroupName"), "must provide a network resource group when a virtual network is specified"))
 		}
 	}
+	if p.BaseDomainResourceGroupName != "" && p.NetworkResourceGroupName != "" && strings.ToUpper(p.BaseDomainResourceGroupName) == strings.ToUpper(p.NetworkResourceGroupName) {
+		allErrs = append(allErrs, field.Required(fldPath.Child("baseDomainResourceGroupname"), "baseDomainResourceGroupName cannot be the same as networkResourceGroupName"))
+	}
 	if (p.ComputeSubnet != "" || p.ControlPlaneSubnet != "") && (p.VirtualNetwork == "" || p.NetworkResourceGroupName == "") {
 		if p.VirtualNetwork == "" {
 			allErrs = append(allErrs, field.Required(fldPath.Child("virtualNetwork"), "must provide a virtual network when supplying subnets"))


patrickdillon:
When I try to test an install with the `resourceGroupName` == `baseDomainResourceGroupName` I hit this existing validation:
```
FATAL failed to fetch Terraform Variables: failed to fetch dependency of "Terraform Variables": failed to generate asset "Platform Provisioning Check": platform.azure.resourceGroupName: Invalid value: "padillon-eg-rg-bug-test": resource group must be empty but it has 1 resources like /subscriptions/433715e6-37fe-4328-af75-3661e13b15fc/resourceGroups/padillon-eg-rg-bug-test/providers/Microsoft.Network/dnszones/padilloneqrgbug.installer.azure.devcluster.openshift.com ...
```

This validation appears to exist back in 4.8. So it looks like we may be missing a different edge case... perhaps we're letting this slip through if the DNS zone does not exist in the `baseDomainResourceGroup`

pkg/types/azure/validation/platform.go
```diff
@@ -70,6 +70,9 @@ func ValidatePlatform(p *azure.Platform, publish types.PublishingStrategy, fldPa
 			allErrs = append(allErrs, field.Required(fldPath.Child("networkResourceGroupName"), "must provide a network resource group when a virtual network is specified"))
 		}
 	}
+	if p.BaseDomainResourceGroupName != "" && p.NetworkResourceGroupName != "" && strings.ToUpper(p.BaseDomainResourceGroupName) == strings.ToUpper(p.NetworkResourceGroupName) {
```
barbacbd, line 73:
```suggestion
	if p.BaseDomainResourceGroupName != "" && p.NetworkResourceGroupName != "" && strings.EqualFold(p.BaseDomainResourceGroupName, p.NetworkResourceGroupName) {
```

jhixson74: Old habits die hard. Good find ;-)

jhixson74:
> When I try to test an install with the `resourceGroupName` == `baseDomainResourceGroupName` I hit this existing validation:
> 
> ```
> FATAL failed to fetch Terraform Variables: failed to fetch dependency of "Terraform Variables": failed to generate asset "Platform Provisioning Check": platform.azure.resourceGroupName: Invalid value: "padillon-eg-rg-bug-test": resource group must be empty but it has 1 resources like /subscriptions/433715e6-37fe-4328-af75-3661e13b15fc/resourceGroups/padillon-eg-rg-bug-test/providers/Microsoft.Network/dnszones/padilloneqrgbug.installer.azure.devcluster.openshift.com ...
> ```
> 
> This validation appears to exist back in 4.8. So it looks like we may be missing a different edge case... perhaps we're letting this slip through if the DNS zone does not exist in the `baseDomainResourceGroup`

If it existed then, it doesn't now. I can start an install without failure. The install eventually fails.

patrickdillon:
> > When I try to test an install with the `resourceGroupName` == `baseDomainResourceGroupName` I hit this existing validation:
> > ```
> > FATAL failed to fetch Terraform Variables: failed to fetch dependency of "Terraform Variables": failed to generate asset "Platform Provisioning Check": platform.azure.resourceGroupName: Invalid value: "padillon-eg-rg-bug-test": resource group must be empty but it has 1 resources like /subscriptions/433715e6-37fe-4328-af75-3661e13b15fc/resourceGroups/padillon-eg-rg-bug-test/providers/Microsoft.Network/dnszones/padilloneqrgbug.installer.azure.devcluster.openshift.com ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This validation appears to exist back in 4.8. So it looks like we may be missing a different edge case... perhaps we're letting this slip through if the DNS zone does not exist in the `baseDomainResourceGroup`
> 
> If it existed then, it doesn't now. I can start an install without failure. The install eventually fails.
> 
> What does your install-config look like?

I believe the baseDomainResourceGroup should be pointing to a resource group that contains the public DNS zone for the cluster. When specifying a pre existing resource group we validate that resource group is empty. So in theory it should not be possible for resourceGroup == baseDomainResourceGroup. That is the validation error I was hitting above.

But AFAIK we do not validate that the public DNS zone is in the baseDomainResourceGroup. So it seems it is possible to specify an “incorrect” baseDomainResource group that does not contain the public DNS zone.

I do not know whether this causes any practical problems—it would depend on how the baseDomainResourceGroup is used.

jhixson74:
> When I try to test an install with the `resourceGroupName` == `baseDomainResourceGroupName` I hit this existing validation:
> 
> ```
> FATAL failed to fetch Terraform Variables: failed to fetch dependency of "Terraform Variables": failed to generate asset "Platform Provisioning Check": platform.azure.resourceGroupName: Invalid value: "padillon-eg-rg-bug-test": resource group must be empty but it has 1 resources like /subscriptions/433715e6-37fe-4328-af75-3661e13b15fc/resourceGroups/padillon-eg-rg-bug-test/providers/Microsoft.Network/dnszones/padilloneqrgbug.installer.azure.devcluster.openshift.com ...
> ```
> 
> This validation appears to exist back in 4.8. So it looks like we may be missing a different edge case... perhaps we're letting this slip through if the DNS zone does not exist in the `baseDomainResourceGroup`

I am unable to duplicate this using the 4.8 installer.

patrickdillon:
/close

We will be addressing this differently, by checking to ensure that a public hosted zone exists in the baseDomainResourceGroup. Once that check is added, it would be impossible to set the base domain resource group and the cluster resource group to be equal (due to the existing check to make sure the cluster resource group is empty).

PR #7180, opened 2023-05-11
OCPBUGS-1769: Ignore IAM Roles that the Installer is not authorized to access

sadasu:
Instead of looking for a specific error code (unreliable), look for "AccessDenied" in the error message

### DIFF
--- a/pkg/destroy/aws/iamhelpers.go
+++ b/pkg/destroy/aws/iamhelpers.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"context"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
@@ -11,11 +12,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/sets"
-)
-
-const (
-	// ErrCodeAccessDeniedException is the access denied error code returned by IAM.
-	ErrCodeAccessDeniedException = "AccessDeniedException"
 )
 
 type iamRoleSearch struct {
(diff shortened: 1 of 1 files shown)


pkg/destroy/aws/iamhelpers.go
```diff
@@ -46,13 +42,16 @@ func (search *iamRoleSearch) find(ctx context.Context) (arns []string, names []s
...
+						switch {
+						case awsErr.Code() == iam.ErrCodeNoSuchEntityException:
+							// The role does not exist.
 							// Ignore this IAM Role and donot report this error via
 							// lastError
 							search.unmatched[*role.Arn] = exists
+						case strings.Contains(err.Error(), "AccessDenied"):
+							// Installer does not have access to this IAM role
+							// Ignore this IAM Role and donot report this error via
+							// lastError
```
r4f4, line 54: Shouldn't you 
```suggestion
							// lastError
							search.unmatched[*role.Arn] = exists
```
to ignore the role as the comment says?

r4f4: @sadasu If I'm reading it correctly, you can avoid this API call if you ever see this role again: https://github.com/openshift/installer/pull/7180/files#diff-2cf3547918c79b5cad1244c173df556f9664596bb88d8bc2a6c888e681d8a038R36-R38

patrickdillon:
X-posting: I left a question or two [in the bug](https://issues.redhat.com/browse/OCPBUGS-1769?focusedId=22298035&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22298035). It's not clear to me whether this PR is catching cases that were missed in the original PR or if the original fix was misunderstood.

sadasu:
@patrickdillon from the latest logs it appears that the error code for the AccessDenied error introduced in the 1st fix is not returned when that error is encountered. We decided early on that we weren't going to import the AccessDenied error code from https://github.com/aws/aws-sdk-go/blob/main/service/servicequotas/errors.go. So, the next best option seems to be to look foe "AccessDenied" in the error message.

patrickdillon:
> @patrickdillon from the latest logs it appears that the error code for the AccessDenied error introduced in the 1st fix is not returned when that error is encountered. 

Do we have logs for this? I haven't been able to find new logs in the bug.

patrickdillon:
> > @patrickdillon from the latest logs it appears that the error code for the AccessDenied error introduced in the 1st fix is not returned when that error is encountered.
> 
> Do we have logs for this? I haven't been able to find new logs in the bug.

I see 

```
grep -E "get tags for.*AccessDenied: User" instdir_yunjiang-role16b/.openshift_install.log | wc -l
41219
```
If we have anything more concrete LMK, but based on that line count, it does seem like they're repeating. 

/approve

Issue #721, opened 2018-11-23
Cluster deployment hangs at "Failed to connect events watcher: Get https://mycluster-api.openshift.testing:6443/apikube-system/events?


jihed:
# Version

```console
openshift-install version
openshift-install v0.3.0-250-g30bb25ac57d7c7d3dae519186cbfca9af8aeaca2
Terraform v0.11.8

Your version of Terraform is out of date! The latest version
is 0.11.10. You can update by downloading from www.terraform.io/downloads.html
```
```console
/usr/local/bin/terraform-provider-libvirt -version
/usr/local/bin/terraform-provider-libvirt was not built correctly
Compiled against library: libvirt 4.3.0
Using library: libvirt 4.3.0
Running hypervisor: QEMU 2.10.0
Running against daemon: 4.3.0
```

# Platform (aws|libvirt|openstack): 

Libvirt

# What happened?
The installation stuck at the check of the api. 
`level=warning msg="Failed to connect events watcher: Get https://mycluster-api.openshift.testing:6443/api/v1/namespaces/kube-system/events?resourceVersion=2176&watch=true: dial tcp 192.168.126.10:6443: connect: connection refused"
`
The etcd is running with some issue regarding the tls certificate:
```[root@mycluster-master-0 ~]# crictl logs $(sudo crictl ps --pod=$(sudo crictl pods --name=etcd-member --quiet) --quiet)
...
2018-11-23 09:52:56.673956 I | etcdserver: setting up the initial cluster version to 3.2
2018-11-23 09:52:56.674057 I | embed: ready to serve client requests
2018-11-23 09:52:56.674405 I | embed: serving client requests on [::]:2379
2018-11-23 09:52:56.677078 N | etcdserver/membership: set the initial cluster version to 3.2
2018-11-23 09:52:56.677356 I | etcdserver/api: enabled capabilities for version 3.2
WARNING: 2018/11/23 09:52:56 Failed to dial 0.0.0.0:2379: connection error: desc = "transport: authentication handshake failed: remote error: tls: bad certificate"; please retry.
2018-11-23 09:57:36.927049 W | etcdserver: apply entries took too long [136.407186ms for 1 entries]
2018-11-23 09:57:36.927368 W | etcdserver: avoid queries with large range/delete range!
2018-11-23 09:58:53.660410 W | etcdserver: apply entries took too long [368.197198ms for 1 entries]
2018-11-23 09:58:53.671038 W | etcdserver: avoid queries with large range/delete range!
[...]

PR #7186, opened 2023-05-15
AGENT-558 Generate unconfigured agent ignition

rwsu:
Adds hidden subcommand:

openshift-install agent create unconfigured-ignition

to generate a unconfigured ignition config to be used by factory
use cases.

The InfraEnv, ClusterImageSet, and AgentPullSecret are the only
ZTP assets used to build the unconfigured ignition.
ClusterDeployment, AgentClusterInstall, and NMStateConfigs are
excluded.

The apply-host-config and start-cluster-installation services
are also disabled. These services will be enabled by a new
service that will  detect that the config ISO has been mounted
and the configuration files have been copied onto the host.

A new IgnitionBaseAsset has been created and is used
to generate the unconfigured ignition. The existing Ignition
asset adds a dependency to IgnitionBaseAsset and builds
its ignition config using the IgnitionBase's Ignition
config as a starting point.

The existing Ignition includes configuration information from
ClusterDeployment, AgentClusterInstall, and NMStateConfig
manifests. This information is included in the agentTemplateData
to template files or systemd units and replace existing versions
inherited from the base Ignition config


pkg/asset/agent/image/ignitionbase.go
```diff
@@ -0,0 +1,269 @@
...
+	if err := a.generateFile(unconfiguredIgnitionFilename); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// PersistToFile writes the unconfigured ignition in the assets folder.
+func (a *IgnitionBase) PersistToFile(directory string) error {
+	unconfiguredIgnFile := filepath.Join(directory, a.File.Filename)
```
danielerez, line 183: I'm getting "nil pointer dereference" here when running 'openshift-install agent create unconfigured-ignition'.
I guess 'a.File' is missing?

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x3a325df]

goroutine 1 [running]:
github.com/openshift/installer/pkg/asset/agent/image.(*IgnitionBase).PersistToFile(0x229cd500, {0x7ffedeade49b?, 0x229cd500?})
	/home/git/installer/pkg/asset/agent/image/ignitionbase.go:183 +0x3f
main.runTargetCmd.func1({0x7ffedeade49b, 0xb})
	/home/git/installer/cmd/openshift-install/create.go:265 +0x24a
main.runTargetCmd.func2(0x229a1da0?, {0xc00059b020?, 0x2?, 0x2?})
	/home/git/installer/cmd/openshift-install/create.go:290 +0xe7
```

cmd/openshift-install/agent.go
```diff
@@ -85,7 +85,20 @@ var (
...
+	agentUnconfiguredIgnitionTarget = target{
+		name: "Agent unconfigured ignition",
+		command: &cobra.Command{
+			Use:    "unconfigured-ignition",
+			Short:  "Generates an agent ignition that excludes cluster configuration",
+			Args:   cobra.ExactArgs(0),
+			Hidden: true,
+		},
+		assets: []asset.WritableAsset{
+			&image.IgnitionBase{},
```
danielerez, line 97: I've tried to manually use the IgnitionBase asset and getting the following error:
```
failed to fetch dependency of "Agent Installer Ignition Base": failed to generate asset "InfraEnv Config": missing configuration or manifest file
```
The error is thrown from InfraEnv [finish](https://github.com/openshift/installer/blob/14f8ef527b8cb1d1556f1ca882a2344641fc5133/pkg/asset/agent/manifests/infraenv.go#L142) func I guess?

pkg/asset/agent/image/ignitionbase.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	for _, file := range filesToInclude {
+		manifestFile := ignition.FileFromBytes(filepath.Join(manifestPath, filepath.Base(file.Filename)),
+			"root", 0600, file.Data)
+		config.Storage.Files = append(config.Storage.Files, manifestFile)
+	}
+
+	disabledServices := []string{"set-hostname.service", "start-cluster-installation.service"}
+	enabledServices := StringArrayRemove(agentEnabledServices, disabledServices)
```
zaneb, line 166: It's strange to see services being removed from the list in ignitionbase rather than added in the fully configured ignition.

pkg/asset/agent/image/ignition.go
```diff
@@ -94,30 +104,17 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
...
-					Name: "core",
-					SSHAuthorizedKeys: []igntypes.SSHAuthorizedKey{
-						igntypes.SSHAuthorizedKey(infraEnv.Spec.SSHAuthorizedKey),
-					},
-					PasswordHash: &pwdHash,
-				},
-			},
-		},
-	}
+	config := ignitionBaseAsset.Config
```
zaneb, line 116: This doesn't do a deep copy, so I think it will end up modifying the content of the ignitionBase. I'm not sure how the asset store will respond to that, but I think things may get weird (so if you ran e.g. `agent create image` followed by `agent create unconfigured-ignition`, the resulting ignition file may in fact be a partly-configured one).

In the absence of a DeepCopy() method, you could dump to JSON and then reload.

An alternative you may already have considered is to use golang struct composition instead of the asset graph to incorporate the base implementation. (That would require adding a separate asset for the unconfigured ignition, since that one is a WritableAsset but we don't want this one to be.)

pkg/asset/agent/image/ignitionbase.go
rwsu: Yes a.File is missing. This could happen if one of the assets that IgnitionBase depends on is missing. I will add a check and additional integration test validations to check for these cases.

pkg/asset/agent/image/ignitionbase.go
```diff
@@ -0,0 +1,269 @@
...
+			"bootstrap/files/usr/local/bin/"+script+".template",
+			relImgData); err != nil {
+			return err
+		}
+	}
+
+	filesToInclude := [...]asset.File{
+		*infraEnvAsset.File,
+		*clusterImageSetAsset.File,
+		*pullSecretAsset.File,
```
danielerez, line 156: Do we need the pull secret here only for pulling the release image? If so, since in disconnected env we already provide the entire release payload, can we make it optional?

zaneb: Even assuming the payload is cached already, is it possible to create a cluster without _any_ pull secret?
It seems unlikely that assisted-service would accept this, and even less likely that the installer process it runs to generate the ignition would.

pkg/asset/agent/image/ignitionbase.go
rwsu: Ok, I'll switch it to having a default list in IgnitionBaseAsset and then in the IgnitionAsset add in the additional services.

pkg/asset/agent/image/ignition.go
rwsu: Ah I see where this may be confusing. Ignition.Config is a pointer while IgnitionBase.Config is not. So this line should create a copy instead of modifying the content of the IgnitionBase. I'll add a unit test to verify that the Ignition.Generate doesn't change the IgnitionBase.

zaneb: That is confusing, but the real problem is that it's only a shallow copy. So while any pointers, maps, or slices will get copied (along with other fields like strings/ints and structs), the stuff they are pointing to will not.
so e.g. on the line below this, `config.Passwd.Users` is a new copy of the slice, but setting `config.Passwd.Users[0].PasswordHash` modifies the original object in the underlying array.

rwsu: That is tricky. Ok I will copy function to dump and reload from yaml.

pkg/asset/agent/image/ignitionbase.go
danielerez: So the pull secret is required as part of the config-image, right? Which seems like the proper way to provide the pull secret for creating a cluster. So I wonder why is it needed _also_ in the unconfigured ignition?

pkg/asset/agent/manifests/clusterimageset.go
```diff
@@ -55,27 +55,30 @@ func (a *ClusterImageSet) Generate(dependencies asset.Parents) error {
 		return err
 	}
 
+	clusterImageSet := &hivev1.ClusterImageSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("openshift-%s", currentVersion),
```
danielerez, line 60: IIUC, it means that the unconfigured-ignition would be locked to this release image version. It could be problematic since the appliance supports specifying a specific OCP release version. Can we maybe make it configurable? I.e. allow to bypass the validation in 'validateReleaseVersion' func.

pkg/asset/agent/image/ignitionbase.go
```diff
@@ -0,0 +1,303 @@
...
+	a.File = &asset.File{
+		Filename: filename,
+		Data:     data,
+	}
+	return nil
+}
+
+// Load returns the IgnitionBase from disk.
+func (a *IgnitionBase) Load(f asset.FileFetcher) (bool, error) {
+	// The IgnitionBase will not be needed by another asset so load is noop.
```
danielerez, line 258: We actually want to try using this asset from the appliance code, so could the func be implemented?

pkg/asset/agent/manifests/clusterimageset.go
zaneb: > the appliance supports specifying a specific OCP release version

What? No. The agent installer does not and will not support that.
And it doesn't even make any sense. The release image is built in to the appliance, so of course it has to use that one.

pkg/asset/agent/image/ignitionbase.go
zaneb: This needs some explanation.

pkg/asset/agent/manifests/clusterimageset.go
danielerez: ok, I thought we need to supply the ClusterImageSet. If the the release image in ClusterImageSet should be exactly the one in binary, it can be just auto-generated for the unconfigured ignition. It means that we have to use the binary, just wondered if we can import the IgnitionBase asset.

pkg/asset/agent/image/ignitionbase.go
danielerez: Wanted to try importing the IgnitionBase asset in the appliance code. But nevermind, I understand it won't be supported. So we'll extract the installer image from the release and use the API.

zaneb: You want to shell out to the installer I think, otherwise you'll have to recompile your code every time you want to change the version to install.

pkg/asset/agent/image/ignitionbase.go
```diff
@@ -0,0 +1,269 @@
...
+	releaseImageList, err := releaseImageList(clusterImageSet.Spec.ReleaseImage, archName)
+	if err != nil {
+		return err
+	}
+
+	registriesConfig := &mirror.RegistriesConf{}
+	// registryCABundle := &mirror.CaBundle{}
+	// dependencies.Get(registriesConfig, registryCABundle)
+
+	publicContainerRegistries := getPublicContainerRegistries(registriesConfig)
```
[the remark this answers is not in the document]
rwsu: I've update it to include the mirror config if it is provided.

pkg/asset/agent/manifests/clusterimageset.go
rwsu: A recent update made the ClusterImageSet optional. If one isn't provided it will generate one inside the ignition using the installer's release version.

pkg/asset/agent/image/ignitionbase.go
zaneb: In every case except a factory-installed image with pre-cached containers (e.g. in the interactive use case) a pull secret is required to pull e.g. the assisted-service container in order to do anything at all. And if one is exists then we'll want to be able use it for the cluster (so the user is not required to provide it twice in the interactive flow).

I don't know how deep the validation we do of the pull secret goes, but I suspect in your case you could easily make up a bogus one that would keep the installer happy, like:

```
{"auths": {"example.com": {"email": "test@example.com", "auth": "this is not a real pull secret"}}}
```

pkg/asset/agent/image/ignitionbase.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+// Files returns the files generated by the asset.
+func (a *IgnitionBase) Files() []*asset.File {
+	// Return empty array because File will never be loaded.
+	return []*asset.File{}
+}
+
+func getOSImagesInfo(cpuArch string) (*models.OsImage, error) {
+	st, err := rhcos.FetchCoreOSBuild(context.Background())
```
andfasano, line 270: Recently I opened a PR to fix the lack of support for OPENSHIFT_INSTALL_OS_IMAGE_OVERRIDE (see https://github.com/openshift/installer/pull/721). Wondering if also this point should honor it?

rwsu:
The coupling between Ignition and IgnitionBase has made maintaining the unit tests to be very difficult. I am thinking about breaking up this coupling and not have a dependency between Ignition and the unconfigured ignition. Code reuse was one of the original reasons of having the dependency, but seeing how the implementation has gone, it appears to have made the code more difficult to follow/understand.

pkg/asset/agent/image/ignitionbase.go
```diff
@@ -0,0 +1,301 @@
...
+	"assisted-service-pod.service",
+	"assisted-service.service",
+	"create-cluster-and-infraenv.service",
+	"node-zero.service",
+	"multipathd.service",
+	"selinux.service",
+	"install-status.service",
+	// Services disabled in IgnitionBase
+	// "set-hostname.service",
+	// "start-cluster-installation.service",
```
bfournie, line 58: I've found that if the dependencies are set such that that node-zero.service doesn't start until load-config.iso.service has run and created the `/etc/assisted/rendezvous-host.env` file, then these services shouldn't be removed as the are necessary and will start after the node-zero.service

cmd/openshift-install/testdata/agent/unconfigured-ignition/configurations/from-ztp-manifests.txt
```diff
@@ -0,0 +1,82 @@
...
+exec openshift-install agent create unconfigured-ignition --dir $WORK
+
+exists $WORK/unconfigured-agent.ign
+! exists $WORK/auth/kubeconfig
+! exists $WORK/auth/kubeadmin-password
+! ignitionContains unconfigured-agent.ign /etc/assisted/manifests/agent-cluster-install.yaml
+! ignitionContains unconfigured-agent.ign /etc/assisted/manifests/cluster-deployment.yaml
+igncmp unconfigured-agent.ign /etc/assisted/manifests/cluster-image-set.yaml expected/cluster-image-set.yaml
+igncmp unconfigured-agent.ign /etc/assisted/manifests/infraenv.yaml expected/infraenv.yaml
+igncmp unconfigured-agent.ign /etc/assisted/manifests/pull-secret.yaml expected/pull-secret.yaml
```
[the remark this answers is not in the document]
rwsu: Thanks for the suggestion. It does read better. I've changed the two functions created to "unconfiguredIgnContains" and "unconfiguredIgnCmp". For both the ignition file path is implicit.

cmd/openshift-install/testdata/agent/unconfigured-ignition/configurations/sno.txt
```diff
@@ -0,0 +1,47 @@
+# Verify a default configuration for the SNO topology
```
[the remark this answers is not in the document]
rwsu: Different platform type, but it doesn't matter for unconfigured ignition. Will remove.

pkg/asset/agent/image/ignitionbase.go
rwsu: This function was moved from ignition.go. But yeah, we'll need to fix it. Probably not under the context of this PR.

pkg/asset/agent/image/unconfigured_ignition.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/openshift/installer/pkg/asset/ignition"
+	"github.com/openshift/installer/pkg/asset/ignition/bootstrap"
+	"github.com/openshift/installer/pkg/types"
+)
+
+const (
+	unconfiguredIgnitionFilename = "unconfigured-agent.ign"
+)
+
+// UnconfiguredIgnition is an asset that generates the agent installer base ignition filea
```
bfournie, line 25: nit typo

pkg/asset/agent/image/unconfigured_ignition.go
```diff
@@ -0,0 +1,199 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func (a *UnconfiguredIgnition) generateFile(filename string) error {
+	data, err := ignition.Marshal(a.Config)
+	if err != nil {
+		return errors.Wrap(err, "failed to Marshal Ignition config")
```
bfournie, line 179: nit - error should probably be `failed to Marshal Unconfigured Ignition` (to distinguish it from https://github.com/openshift/installer/blob/master/pkg/asset/ignition/bootstrap/common.go#L217

PR #7190, opened 2023-05-17
OCPBUGS-11727: WIP: baremetal: support very large ignition

dtantsur:
1) Update terraform-provider-libvirt to get rid of the RPC size limitation
2) Serve ignition via HTTP to avoid the very slow fw_cfg


dtantsur:
/retest

EPEL mirror problems

pkg/asset/cluster/tfvars.go
```diff
@@ -668,6 +669,12 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			imageCacheIP = installConfig.Config.Platform.BareMetal.BootstrapProvisioningIP
 		}
 
+		machineCIDR := &installConfig.Config.Networking.MachineNetwork[0].CIDR
+		virtHostIP, err := cidr.Host(&machineCIDR.IPNet, 1)
```
zaneb, line 673: I wouldn't expect this to work in general. I think we need to find an IP address of the local host that is in the machine network. There's no guarantee that this will be the first IP address in that CIDR. it could be anything.

dtantsur: Do you have any ideas how to detect that? I could not find anything in terraform...

dtantsur:
/test govet

Already fixed

derekhiggins:
/test e2e-metal-ipi-sdn-virtualmedia
This works for me locally, I think the problem might be due to the firewall blocking the port used to share the ignition, I'll ssh to CI job to confirm while its running

dtantsur:
Tigera has trimmed the Calico manifests, so I won't be chasing this change any more.

PR #7192, opened 2023-05-18
OCPBUGS-13764: Support /dev/disk/by-path root device hints

zaneb:
Update the installer/terraform implementation to match the Metal³ changes in [OCPBUGS-13080](https://issues.redhat.com/browse/OCPBUGS-13080) so that control plane and worker machines will have equivalent functionality.

This includes accepting /by-path/ paths and adding validation to ensure that invalid paths that can't ever match are rejected immediately. The code to construct the HardwareDetails for the status annotation of the control plane hosts runs in the baremetal-operator container so it always matches and no changes are needed here.

### DIFF
--- a/pkg/asset/agent/agentconfig/agent_config.go
+++ b/pkg/asset/agent/agentconfig/agent_config.go
@@ -13,6 +13,7 @@ import (
 	"github.com/openshift/installer/pkg/asset"
 	"github.com/openshift/installer/pkg/types/agent"
 	"github.com/openshift/installer/pkg/types/agent/conversion"
+	"github.com/openshift/installer/pkg/types/baremetal/validation"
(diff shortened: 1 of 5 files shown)


andfasano:
> > It would be nice also to have it exercised in one of our existing dev-scripts agent job
> 
> This change only affects the metal IPI control plane nodes (although there was _another_ change to assisted-service that also enables this for agent-based). It does need approval from an agent approver though because of the new validation in the agent code. /cc @bfournie

I understood the validation was related to the comments here https://github.com/openshift/assisted-service/pull/5185/files#r1188375081?

zaneb:
It's the same validation as in the Metal³ webhook, basically telling you up front if you specify a path that can't possibly match any device. We probably should always have had this.

PR #7193, opened 2023-05-18
[release-4.13] OCPBUGS-13765: Support /dev/disk/by-path root device hints

zaneb:
In baremetal IPI, we should support /dev/disk/by-path root device hints in the install-config to match the implementation in Metal³.


zaneb:
@pamoedom this needs cherry-pick-approved please

PR #7200, opened 2023-05-21
AGENT-562: Load config from config image

bfournie:
Add a service to run a script to load the contents of the archive created in AGENT-563 to the system. The script will check for a device with the config image label and copy its contents.

This service will eventually be added to the unconfigured ignition as part of AGENT-558.


data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -0,0 +1,75 @@
...
+       echo "The release $arch_release_image in archive does not match the current release ${release_image}"
+       cleanup_files
+       return 1
+    fi
+    echo "Archive on ${device} contains release ${arch_release_image}"
+
+    # Copy all files from archive
+    # The rendezvousIP file must be copied last as it triggers set-node-zero.sh to continue configuration
+    cpio -icvd --nonmatching "${AGENT_RENDEZVOUS_IP_FILE}" < "${unzipped_file}"
+    cpio -icvd "${RENDEZVOUS_HOST_ENV}" < "${unzipped_file}"
```
zaneb, line 40: I think we need to be careful about trusting that the supplied config file contains only the data we generate from the installer. I think we have to verify that it only contains the allowed files we are expecting before we unpack it into the filesystem.

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -0,0 +1,75 @@
...
+    fi
+
+    cp ${AGENT_CONFIG_MOUNT}/${AGENT_CONFIG_ARCHIVE_FILE} /tmp
+    gunzip -f /tmp/${AGENT_CONFIG_ARCHIVE_FILE}
+    unzipped_file=$(echo /tmp/${AGENT_CONFIG_ARCHIVE_FILE} | cut -d'.' -f1)
+
+    # Get the releaseImage in the archive and verify it matches the current cluster-image-set
+    release_image=$(grep releaseImage ${CLUSTER_IMAGE_SET} | sed -n -e 's/^.*releaseImage: //p')
+    arch_release_image=$(cpio -icv --to-stdout ${CLUSTER_IMAGE_SET} < "${unzipped_file}" | grep releaseImage | sed -n -e 's/^.*releaseImage: //p')
+    if [[ "${release_image}" != "${arch_release_image}" ]]; then
```
zaneb, line 30: Off the top of my head, I can't think of a reason not to compare the full contents of the files, since they're both supposed to be generated by the same version of the installer.

```
if diff "${CLUSTER_IMAGE_SET}" <(cpio -icv --to-stdout "${CLUSTER_IMAGE_SET}" <"${unzipped_file}"); then
```

zaneb (commented):
Great idea to label the image so we can identify it :+1:

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -0,0 +1,75 @@
...
+    mkdir -p ${AGENT_CONFIG_MOUNT}
+    mount ${device} ${AGENT_CONFIG_MOUNT}
+
+    if [[ ! -f ${AGENT_CONFIG_MOUNT}/${AGENT_CONFIG_ARCHIVE_FILE} ]]; then
+       echo "Could not find file ${AGENT_CONFIG_ARCHIVE_FILE} on ${device}"
+       cleanup_files
+       return 1
+    fi
+
+    cp ${AGENT_CONFIG_MOUNT}/${AGENT_CONFIG_ARCHIVE_FILE} /tmp
```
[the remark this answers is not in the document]
bfournie: Yep. Fixed.

data/data/agent/files/usr/local/bin/load-config-iso.sh
bfournie: Changed to compare file.

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -0,0 +1,75 @@
...
+    fi
+    if [[ -f ${unzipped_file} ]]; then
+       rm ${unzipped_file}
+    fi
+}
+
+while true
+do
+    # Look for devices matching config image label
+    device=$(lsblk -p -o NAME,LABEL | grep ${AGENT_CONFIG_IMAGE_LABEL} | cut -d ' ' -f1)
```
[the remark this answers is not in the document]
bfournie: Good idea, will investigate this.

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -0,0 +1,75 @@
...
+AGENT_CONFIG_ARCHIVE_FILE="config.gz"
+AGENT_CONFIG_MOUNT="/mnt/config_image"
+RENDEZVOUS_HOST_ENV="/etc/assisted/rendezvous-host.env"
+CLUSTER_IMAGE_SET="/etc/assisted/manifests/cluster-image-set.yaml"
+
+copy_archive_contents() {
+ 
+    # Mount device
+    mkdir -p ${AGENT_CONFIG_MOUNT}
+    mount ${device} ${AGENT_CONFIG_MOUNT}
```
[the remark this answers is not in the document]
bfournie: I've added a udev rule to start the service with the device name when it detects the device with that label. I wasn't able to get the mount created in the udev rule itself, I've also found [recommendations](https://unix.stackexchange.com/questions/464959/udev-rule-to-automount-media-devices-stopped-working-after-systemd-was-updated-t) to not mount within the rule but in the service itself. So I've added the systemd-mount to the service.

data/data/agent/files/usr/local/bin/load-config-iso.sh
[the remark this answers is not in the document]
bfournie: Added a udev rule to detect the label and start a service with the devname.

data/data/agent/files/usr/local/bin/load-config-iso.sh
bfournie: Working on a change to share the list of files between the load-config script and  create config (https://github.com/openshift/installer/pull/7157).

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -0,0 +1,67 @@
...
+          break
+       fi
+    else
+       echo "Could not find ${AGENT_CONFIG_ARCHIVE_FILE} in ${AGENT_CONFIG_MOUNT}"
+    fi
+
+    echo "Retrying to copy contents from ${AGENT_CONFIG_MOUNT}/${AGENT_CONFIG_ARCHIVE_FILE}"
+    sleep 5
+done
+
```
rwsu, line 67: Because the unconfigured ignition disables these services:
* apply-host-config.service
* start-cluster-installation.service

They should be enabled here.

rwsu (changes requested):

data/data/agent/systemd/units/load-config-iso@.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
```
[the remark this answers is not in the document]
bfournie: Yes I think so.

data/data/agent/files/usr/local/bin/load-config-iso.sh
bfournie: OK I will add these

bfournie: @rwsu Actually could we add those services in the unconfigured ignition with a dependency on the load-config-iso.service

zaneb: I like that idea, assuming it can be templated in to the unit easily enough.
I too was expecting that we would enable the services here, but calling systemctl from inside a systemd service is definitely a bit icky.

data/data/agent/files/usr/local/bin/load-config-iso.sh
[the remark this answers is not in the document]
zaneb: :+1: this is awesome

data/data/agent/files/etc/udev/rules.d/80-agent-config-image.rules
```diff
@@ -0,0 +1 @@
+ACTION=="add", SUBSYSTEM=="block", ENV{ID_FS_LABEL}=="agent_configimage", ENV{SYSTEMD_WANTS}+="load-config-iso@$devnode.service"
```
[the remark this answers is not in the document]
bfournie: > Just wondering, why is the load-config-iso defined as an instanced service unit? I.e. what's the reasoning for naming it with a '@' postfix?

This is necessary so that udev can start the host with the devname of the config-image. For example, the service below is created when the config-image is mounted on /dev/sr1
```
$ sudo systemctl status load-config-iso@-dev-sr1.service
● load-config-iso@-dev-sr1.service - If a config ISO is mounted, copy its files to the system
     Loaded: loaded (/etc/systemd/system/load-config-iso@.service; disabled; preset: disabled)
     Active: activating (start) since Tue 2023-06-06 13:00:21 UTC; 4min 3s ago
   Main PID: 1581 (load-config-iso)
      Tasks: 2 (limit: 204700)
     Memory: 1.0M
        CPU: 956ms
     CGroup: /system.slice/system-load\x2dconfig\x2diso.slice/load-config-iso@-dev-sr1.service
             ├─1581 /bin/bash /usr/local/bin/load-config-iso.sh /dev/sr1
             └─8764 sleep 5
```

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -0,0 +1,71 @@
...
+    fi
+}
+
+# This script will be invoked by a udev rule when it detects a device with the correct label
+devname="$1"
+systemd-mount --no-block --automount=yes --collect "$devname" "${AGENT_CONFIG_MOUNT}"
+echo "Mounted ${devname} on ${AGENT_CONFIG_MOUNT}"
+
+while true
+do
```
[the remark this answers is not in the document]
zaneb: I definitely think we need this message... somewhere.
Let's try to keep the text neutral as to whether the next step is the config image or interactive GUI. Something like "Waiting for cluster configuration input from user" maybe?

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -0,0 +1,87 @@
...
+
+    # Copy expected files from archive, overwriting the existing file
+    for file in "${array[@]}"
+    do
+       if [[ "${file}" = *.* ]]; then
+          cpio -icvdu "${file}" < "${unzipped_file}"
+          if [[ -f ${file} ]]; then
+             echo "Copied ${file}"
+          else
+             echo "Failure copying ${file}"
```
andfasano, line 39: q: what's the reason for just logging the copy error? Shouldn't be a blocker?

bfournie: Not necessarily, for example there may not be any files populated in `/etc/assisted/extra-manifests`.
I'll change the text to not log a failure.

data/data/agent/systemd/units/node-zero.service
```diff
@@ -1,7 +1,8 @@
 [Unit]
```
rwsu, line 1: Shouldn't node-zero.service wait for /etc/assisted/rendezvous-host.env to be present?

zaneb: It's done within the script (just merged 4 hours ago): #7068

rwsu: Ah thanks for reminding me.

bfournie: So, I didn't have the 7608 PR included but I set node-zero.service to be 
`After=network-online.target load-config-iso@.service`
Since load-config-iso@.service writes /etc/assisted/rendezvous-host.env it was there when node-zero.service ran

data/data/agent/files/usr/local/bin/load-config-iso.sh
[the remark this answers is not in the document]
rwsu: Good idea, we'll add it to the unconfigured ignition. 
Created a task to track it: https://issues.redhat.com/browse/AGENT-660

rwsu (commented):
/lgtm

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -0,0 +1,89 @@
...
+    cp ${AGENT_CONFIG_MOUNT}/${AGENT_CONFIG_ARCHIVE_FILE} "${tmpdir}"
+    gunzip -f "${tmpdir}"/${AGENT_CONFIG_ARCHIVE_FILE}
+    unzipped_file=$(echo "${tmpdir}/${AGENT_CONFIG_ARCHIVE_FILE}" | cut -d'.' -f1)
+    filelist=$(cpio --list < "${unzipped_file}")
+    echo -e "List of files to copy: ${filelist}"
+
+    # Get the releaseImage in the archive and verify it matches the current cluster-image-set
+    release_image=$(grep releaseImage ${CLUSTER_IMAGE_SET} | sed -n -e 's/^.*releaseImage: //p')
+    arch_image_set=$(cpio -icv --to-stdout ${CLUSTER_IMAGE_SET} < "${unzipped_file}")
+    if [[ $(< "${CLUSTER_IMAGE_SET}") != "${arch_image_set}" ]]; then
```
danielerez, line 21: Can we have this validation according to ClusterImageSet's name (i.e. the version in binary)?
That way, we could support also openshift-install binaries downloaded from: https://mirror.openshift.com/pub/openshift-v4/amd64/clients/ocp.
As the releaseImage url is different than the binary in release (registry.ci.openshift.org/origin/release vs quay.io/openshift-release-dev/ocp-release)

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+
+set -e
+
+AGENT_CONFIG_ARCHIVE_FILE="config.gz"
+AGENT_CONFIG_MOUNT="/media/config-image"
+CLUSTER_IMAGE_SET="/etc/assisted/manifests/cluster-image-set.yaml"
+AGENT_CONFIG_MOUNT="/media/config_image"
```
danielerez, line 8: Duplicated var? Line 6 is redundant I guess?

data/data/agent/files/usr/local/bin/load-config-iso.sh
bfournie: I think this is probably not what we want do, as we want the releaseImage of the config-image to exactly match the releaseImage of the unconfigured igntion. Wdyt @zaneb ?

data/data/agent/files/usr/local/bin/load-config-iso.sh
bfournie: Yes, thanks.

pkg/asset/agent/image/unconfigured_ignition.go
```diff
@@ -22,6 +23,34 @@ const (
...
+		"/etc/assisted/manifests/cluster-deployment.yaml",
+		"/etc/assisted/manifests/cluster-image-set.yaml",
+		"/etc/assisted/manifests/agent-cluster-install.yaml",
+		"/etc/assisted/extra-manifests", // all files in directory
+		"/etc/assisted/hostconfig",      // all files in directory
+		"/etc/assisted/hostnames",       // all files in directory
+		"/etc/issue",
+		"/etc/systemd/system.conf.d/10-default-env.conf",
+		"/root/.docker/config.json",
+		"/usr/local/bin/start-cluster-installation.sh",
```
danielerez, line 40: I think also start-cluster-installation.service should be included, as REQUIRED_MASTER_NODES/REQUIRED_WORKER_NODES env vars are now specified by the service.template: https://github.com/openshift/installer/commit/666dde5a975a2ed2c8d83c78abac2c9fc69d88be#diff-e185c956916cc81d4934f74f0d47973ba058cc5f52569ad91d4b9a131f38c84a.
I.e. otherwise, required_master_nodes/required_worker_nodes vars would be '0'.

data/data/agent/files/usr/local/bin/load-config-iso.sh
zaneb: We really want it to exactly match, otherwise we can never change anything in the installer.
What we need is to provide a way for users to get the installer from the release image that is preinstalled on the servers (or even run it on one of those servers).

pkg/asset/agent/image/unconfigured_ignition.go
zaneb: systemd services go in a different section of ignition from files.

data/data/agent/files/usr/local/bin/load-config-iso.sh
zaneb: Actually, does this work when we can't predict the service name (due to it being templated with the device path, which we can't predict)?
And did this change actually happen somewhere?

pkg/asset/agent/image/unconfigured_ignition.go
bfournie: Yep. I had noted that needed to be done in https://redhat-internal.slack.com/archives/C02SPBZ4GPR/p1686658183214679?thread_ts=1686057590.009719&cid=C02SPBZ4GPR when https://github.com/openshift/installer/pull/7062 merged but forget to change it after it did. Will fix it

data/data/agent/files/usr/local/bin/load-config-iso.sh
bfournie: I actually don't think that any additional service dependencies are needed. Since `set-node-zero.sh` is looping waiting for the  rendezvous host file to appear, when it does the `node-zero.service` completes, which will allow the `assisted-service-pod.service` (https://github.com/openshift/installer/blob/master/data/data/agent/systemd/units/assisted-service-pod.service#L4) to run, followed by create-cluster-and-infraenv.service

zaneb: That's true for the appliance if `apply-host-config.service` and `start-cluster-installation.service` are enabled in the unconfigured ignition.
But if we want the unconfigured ignition to work for an interactive workflow, we need to _disable_ those services by default and only enable them if the config drive is attached.
I think the only way to do this is to call systemctl, and we might as well implement it now.

pkg/asset/agent/image/unconfigured_ignition.go
bfournie: Updated to copy the start-cluster-installation.service. The correct  node numbers are now being used. 
un 27 21:53:29 master-0.registry.ci.openshift.org start-cluster-installation.sh[3209]: Number of required master nodes: 3
Jun 27 21:53:29 master-0.registry.ci.openshift.org start-cluster-installation.sh[3209]: Number of required worker nodes: 0
Jun 27 21:53:29 master-0.registry.ci.openshift.org start-cluster-installation.sh[3209]: Total number of required nodes: 3

Note that this required using `systemctl daemon-reload` after the cpio in order for the updated env variables to be used.

data/data/agent/files/usr/local/bin/load-config-iso.sh
bfournie: Added ability to enable the disabled services in `load-config-iso.sh`

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -0,0 +1,103 @@
...
+             echo "Could not copy directory ${file}, may be empty, attempting to create it"
+             mkdir -p "${file}"
+          fi
+       fi
+    done
+
+    echo "Successfully copied contents of ${AGENT_CONFIG_ARCHIVE_FILE} on ${devname}"
+
+    # For services that were updated a daemon-reload is required to pick up changes
+    systemctl daemon-reload
```
zaneb, line 56: OK, _this_ is the thing I've been told we must never do.
We need to move the variables out of the systemd unit file and into a separate environment file so that the config-drive never updates the systemd unit.

bfournie: I moved the `REQUIRED_MASTER_NODES` and `REQUIRED_WORKER_NODES` to assisted-service.env.template and added that file to list of files to copy. I was able to remove the `systemctl daemon-reload` and confirmed the values are picked up
```
Jun 28 14:20:49 master-0.ostest.test.metalkube.org start-cluster-installation.sh[3237]: Number of required master nodes: 3
Jun 28 14:20:49 master-0.ostest.test.metalkube.org start-cluster-installation.sh[3237]: Number of required worker nodes: 0
Jun 28 14:20:49 master-0.ostest.test.metalkube.org start-cluster-installation.sh[3237]: Total number of required nodes: 3

```

rwsu (commented):
/lgtm

bfournie:
Added infraenv.yaml to GetConfigImageFiles() for https://issues.redhat.com/browse/OCPBUGS-15637

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -2,74 +2,99 @@
...
+    echo -e "List of files to copy: ${filelist}"
 
     # Get the releaseImage in the archive and verify it matches the current cluster-image-set
-    release_image=$(cat ${CLUSTER_IMAGE_SET} | grep releaseImage | sed -n -e 's/^.*releaseImage: //p')
-    arch_release_image=$(cpio -icv --to-stdout ${CLUSTER_IMAGE_SET} < ${unzipped_file} | grep releaseImage | sed -n -e 's/^.*releaseImage: //p')
-    if [[ ${release_image} != ${arch_release_image} ]]; then
-       echo "The release $arch_release_image in archive does not match the current release ${release_image}"
+    release_image=$(grep releaseImage ${CLUSTER_IMAGE_SET} | sed -n -e 's/^.*releaseImage: //p')
+    arch_image_set=$(cpio -icv --to-stdout ${CLUSTER_IMAGE_SET} < "${unzipped_file}")
+    if [[ $(< "${CLUSTER_IMAGE_SET}") != "${arch_image_set}" ]]; then
```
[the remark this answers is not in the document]
bfournie: OK, changed to use diff per suggestion.

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -2,74 +2,99 @@
...
-       return 1 
-    fi
-
-    cp ${AGENT_CONFIG_MOUNT}/${AGENT_CONFIG_ARCHIVE_FILE} /tmp
-    gunzip -f /tmp/${AGENT_CONFIG_ARCHIVE_FILE}
-    unzipped_file=`echo /tmp/${AGENT_CONFIG_ARCHIVE_FILE} | cut -d'.' -f1`
+    tmpdir=$(mktemp --tmpdir -d "config-image--XXXXXXXXXX")
+    cp ${AGENT_CONFIG_MOUNT}/${AGENT_CONFIG_ARCHIVE_FILE} "${tmpdir}"
+    gunzip -f "${tmpdir}"/${AGENT_CONFIG_ARCHIVE_FILE}
+    unzipped_file=$(echo "${tmpdir}/${AGENT_CONFIG_ARCHIVE_FILE}" | cut -d'.' -f1)
```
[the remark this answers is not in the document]
bfournie: Changed

data/data/agent/systemd/units/start-cluster-installation.service.template
```diff
@@ -8,9 +8,8 @@ ConditionPathExists=/etc/assisted/node0
 
 [Service]
 EnvironmentFile=/usr/local/share/assisted-service/assisted-service.env
+EnvironmentFile=/usr/local/share/start-cluster/start-cluster.env
```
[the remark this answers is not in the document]
bfournie: Removed template

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -2,74 +2,99 @@
...
+       return 1
     fi
-    echo "Archive on ${device} contains release ${arch_release_image}"
+    echo "Archive on ${devname} contains release ${release_image}"
+
+    # Get array from string
+    IFS=',' read -r -a array <<< "${CONFIG_IMAGE_FILES}"
+
+    # Copy expected files from archive, overwriting the existing file
+    for file in "${array[@]}"
```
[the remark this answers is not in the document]
bfournie: Changed

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -2,74 +2,99 @@
...
 
     # Get the releaseImage in the archive and verify it matches the current cluster-image-set
-    release_image=$(cat ${CLUSTER_IMAGE_SET} | grep releaseImage | sed -n -e 's/^.*releaseImage: //p')
-    arch_release_image=$(cpio -icv --to-stdout ${CLUSTER_IMAGE_SET} < ${unzipped_file} | grep releaseImage | sed -n -e 's/^.*releaseImage: //p')
-    if [[ ${release_image} != ${arch_release_image} ]]; then
-       echo "The release $arch_release_image in archive does not match the current release ${release_image}"
+    release_image=$(grep releaseImage ${CLUSTER_IMAGE_SET} | sed -n -e 's/^.*releaseImage: //p')
+    arch_image_set=$(cpio -icv --to-stdout ${CLUSTER_IMAGE_SET} < "${unzipped_file}")
+    if [[ $(< "${CLUSTER_IMAGE_SET}") != "${arch_image_set}" ]]; then
+       echo "The cluster-image-set in archive does not match current release ${release_image}"
```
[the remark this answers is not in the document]
bfournie: Added console display (although using common code in issue_status.sh affected by https://issues.redhat.com/browse/OCPBUGS-14900 selinux issue).

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -2,74 +2,99 @@
...
+    # Copy expected files from archive, overwriting the existing file
+    for file in "${array[@]}"
+    do
+       if [[ "${file}" = *.* || "${file}" == "/etc/issue" ]]; then
+	  # Copy file
+          cpio -icvdu "${file}" < "${unzipped_file}"
+          if [[ -f ${file} ]]; then
+             echo "Copied ${file}"
+          else
+             echo "Expected file ${file} is not in archive"
```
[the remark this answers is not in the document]
bfournie: Changed to exit with failure if any of the files (nor dirs) are missing. I think that all files in list in required

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -2,74 +2,99 @@
...
-    echo "Successfully copied contents of ${AGENT_CONFIG_ARCHIVE_FILE} on ${device}"
+    # Enable any services which are not enabled by default
+    declare -a servicelist=("start-cluster-installation.service")
+
+    for service in "${servicelist[@]}"
+    do
+       is_enabled=$(systemctl is-enabled "$service")
+       if [[ "${is_enabled}" == "disabled" ]]; then
+          echo "Service ${service} is disabled, enabling it"
+          systemctl enable "${service}"
```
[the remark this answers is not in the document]
bfournie: So for the appliance the service(s) are enabled by default and we rely on the systemd dependencies. I believe that should also be adequate for the GUI case when we enable the services here, although we will need to test that when we implement the GUI.

pkg/asset/agent/image/unconfigured_ignition.go
```diff
@@ -22,6 +23,36 @@ const (
 	unconfiguredIgnitionFilename = "unconfigured-agent.ign"
 )
 
+// GetConfigImageFiles returns the list of files or file paths to be included in the config-image.
+func GetConfigImageFiles() []string {
+	return []string{
+		"/etc/assisted/manifests/pull-secret.yaml", //nolint:gosec // not hardcoded credentials
+		"/etc/assisted/manifests/nmstateconfig.yaml",
```
rwsu, line 30: Since the nmstateconfig.yaml could be added in the config image, shouldn't the static network config files under /etc/assisted/network also be copied? And then applied by the /usr/local/bin/pre-network-manager-config.sh script?

bfournie: Good catch Richard, we had missed this since dev-scripts was using the nmstateconfig info for the unconfigured ignition. Added the optional dir to the list of files. However there seems to be a dependency issue after that, marking this as hold
/hold

pawanpinjarkar (commented):
/lgtm

rwsu (approved):
/approve

PR #7201, opened 2023-05-22
CFE-858: Update google terraform provider to latest version

bharath-b-rh:
PR is for updating the version of google terraform provider required making use of the latest APIs related to GCP tags and labels. The changes are required for supporting enhancement proposed [here](https://github.com/openshift/enhancements/pull/1217)

Compute Instance and Cloud Storage GCP resources are not global but zonal and tagging these resources require providing the location where the said resources reside and the API for the same `google_tags_location_tag_binding` was made available in [terraform-provider-google/releases#v4.51.0](https://github.com/hashicorp/terraform-provider-google/releases/tag/v4.51.0)

Labels was added to ForwardingRule GCP resource and was made available in GA in [terraform-provider-google/releases#v4.64.0](https://github.com/hashicorp/terraform-provider-google/releases/tag/v4.64.0)

Now the google provider version has been updated to the latest version [terraform-provider-google/releases#v4.74.0](https://github.com/hashicorp/terraform-provider-google/releases/tag/v4.74.0), which has the last [commit](https://github.com/hashicorp/terraform-provider-google/commit/0584082dd66278550d3a604be00bcdd50d4ebea3)


terraform/providers/google/go.mod
```diff
@@ -76,34 +78,18 @@ require (
...
-	google.golang.org/grpc v1.53.0 // indirect
-	google.golang.org/protobuf v1.29.1 // indirect
+	google.golang.org/genproto v0.0.0-20230410155749-daa745c078e1 // indirect
+	google.golang.org/grpc v1.54.0 // indirect
+	google.golang.org/protobuf v1.30.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
 
-// https://issues.redhat.com//browse/OCPBUGS-7699
-replace github.com/hashicorp/go-getter => github.com/hashicorp/go-getter v1.7.0
```
bharath-b-rh, line 94: github.com/hashicorp/go-getter -- No longer used

terraform/providers/google/go.mod
```diff
@@ -76,34 +78,18 @@ require (
...
-
-// https://bugzilla.redhat.com/show_bug.cgi?id=2064702
-replace golang.org/x/crypto => golang.org/x/crypto v0.0.0-20220517005047-85d78b3ac167
-
-// https://bugzilla.redhat.com/show_bug.cgi?id=2100495
-replace golang.org/x/text => golang.org/x/text v0.8.0
-
-// https://issues.redhat.com/browse/OCPBUGS-5667
-replace github.com/Masterminds/goutils => github.com/Masterminds/goutils v1.1.1
```
bharath-b-rh, line 103: github.com/Masterminds/goutils -- no longer used

terraform/providers/google/go.mod
```diff
@@ -76,34 +78,18 @@ require (
...
-
-// https://bugzilla.redhat.com/show_bug.cgi?id=2045880
-replace github.com/prometheus/client_golang => github.com/prometheus/client_golang v1.12.1
```
bharath-b-rh, line 106: github.com/prometheus/client_golang -- no longer used

barbacbd:
There is already a newer version. There is a such a large queue to get merged, do you want to think about reopening with a newer version when we know items are being merged ?

bharath-b-rh:
> There is already a newer version. There is a such a large queue to get merged, do you want to think about reopening with a newer version when we know items are being merged ?

Yeah sure I can do that. Closing for now, to reopen once EP is merged.

terraform/providers/google/go.mod
```diff
@@ -1,41 +1,41 @@
...
 
-require github.com/hashicorp/terraform-provider-google v1.20.1-0.20230327171628-0dc3bde12208 // v4.59.0
+require github.com/hashicorp/terraform-provider-google v1.20.1-0.20230718215755-3edc574a3a8f // v4.74.0
 
 require (
-	bitbucket.org/creachadair/stringset v0.0.11 // indirect
-	cloud.google.com/go v0.110.0 // indirect
-	cloud.google.com/go/bigtable v1.17.0 // indirect
-	cloud.google.com/go/compute v1.18.0 // indirect
+	bitbucket.org/creachadair/stringset v0.0.8 // indirect
```
r4f4, line 8: So even with the `replace` this is being downgraded?

bharath-b-rh: v4.74.0 of terraform-provider-google is using v0.0.8, but we are replacing with v0.0.11 and go.sum has entry for v0.0.11. Please correct me, this is as expected isn't it?

terraform/providers/google/go.mod
```diff
@@ -1,41 +1,41 @@
 module github.com/openshift/installer/terraform/providers/google
 
-go 1.18
+go 1.19
```
pawanpinjarkar, line 3: The installer uses go version 1.20 [here](https://github.com/openshift/installer/blob/master/go.mod#L3) Should it not be the same here as well?

pawanpinjarkar (commented):
Can we keep the go versions same throughout the installer repository?

terraform/providers/google/go.mod
bharath-b-rh: Yeah, but the provider is still using 1.19, so thought to keep the same. Should I update the go version to 1.20 same as installer?

pawanpinjarkar: @patrickdillon please suggest.

r4f4: @jhixson74 is working on updating the go version of terraform providers to 1.20.

r4f4: @bharath-b-rh The Golang versions are supposed to be backward-compatible, so any version >= 1.19 should be safe to use.

bharath-b-rh: Thank you @r4f4 , I will make the changes for google provider.

bharath-b-rh: suggestion incorporated.

terraform/providers/google/go.mod
patrickdillon: if this is no longer used, can it be removed?

terraform/providers/google/go.mod
bharath-b-rh: I think, this can be removed, was being replaced with the higher version.

terraform/providers/google/go.mod
bharath-b-rh: Package was not in required block, and replace to a particular version added, not sure of the reason and the effects of removing it.Please suggest.

terraform/providers/google/go.mod
bharath-b-rh: I think, this can be removed, was being replaced with the higher version.

terraform/providers/google/go.mod
r4f4: The reason could be because it was pulled in as dep of a dep.

bharath-b-rh: Thank you @r4f4. Package is not in the vendor list too, so I think it's safe to remove.

terraform/providers/google/go.mod
bharath-b-rh: replace removed.

terraform/providers/google/go.mod
patrickdillon: It looks like you might have inadvertently removed: replace golang.org/x/text => golang.org/x/text v0.8.0

terraform/providers/google/go.mod
patrickdillon: It looks like you might have inadvertently removed: replace golang.org/x/text => golang.org/x/text v0.8.0

PR #7205, opened 2023-05-24
CORS-2628: Allow users to set ManagedBy tag to resource group

rna-afk:
Adding a pipe from the install config to terraform for enabling users to set the managedBy tag in the resource group. Defaulting to "openshift" if the field is not set.


bennerv (commented):
@rna-afk I think there's some confusion here.  

The managedBy field isn't a tag, it's an actual field on the resource group.  It looks like terraform doesn't have native support for this field yet, which was realized late last week.  

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/resource_group.

The below shows that there's a difference.  

```bash
$ az group show -n aro-dsfx17my 
{
  "id": "/subscriptions/<sub-id>/resourceGroups/aro-dsfx17my",
  "location": "eastus",
  "managedBy": "/subscriptions/<sub-id>/resourceGroups/shared-cluster/providers/Microsoft.RedHatOpenShift/openShiftClusters/shared-cluster",
  "name": "aro-dsfx17my",
  "properties": {
    "provisioningState": "Succeeded"
  },
  "tags": {
    "abd": "def",
    "createdAt": "2023-04-12T20:54:10.7804839Z",
    "hello": "world",
    "something": "something",
    "test": "test"
  },
  "type": "Microsoft.Resources/resourceGroups"
}

File: pkg/asset/installconfig/azure/metadata.go [hunk 1]
```diff
@@ -31,6 +31,10 @@ type Metadata struct {
 	// serve multiple users with different credentials via a web server.
 	Credentials *Credentials `json:"credentials,omitempty"`
 
+	// ManagedBy is a field that can be set as a tag to the azure
+	// resource group.
+	ManagedBy string `json:"managedBy,omitempty"`
```
patrickdillon, line 36: If you want this value to be user-supplied, it should be added to the install config in pkg/types. 

This is adding the field to the install config metadata, I suspect you don't need the field added here.

PR #7206, opened 2023-05-24
AGENT-627: Decompress kernel on ARM

pawanpinjarkar:
Decompress kernel for ARM architecture in case of PXE boot.

### DIFF
--- a/pkg/asset/agent/image/agentpxefiles.go
+++ b/pkg/asset/agent/image/agentpxefiles.go
@@ -1,6 +1,7 @@
 package image
 
 import (
+	"compress/gzip"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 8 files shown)


pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -100,9 +103,22 @@ func (a *AgentPXEFiles) PersistToFile(directory string) error {
...
-	err = a.copy(agentVmlinuzFile, fileReader)
-	if err != nil {
-		return err
+
+	if a.cpuArch == arch.RpmArch(types.ArchitectureARM64) {
+		gzipReader, err := gzip.NewReader(fileReader)
+		if err != nil {
+			panic(err)
+		}
+		defer gzipReader.Close()
```
zaneb, line 112: This is good, but it makes me notice that we're not actually closing all the fileReaders we open in this function.

pawanpinjarkar:
Closing this one because of a bad rebase that introduced unwanted changes.  
Created a new and clean one here https://github.com/openshift/installer/pull/7276

PR #7209, opened 2023-05-26
OCPBUGS-13108: Log additional host info at warning level

bfournie:
When the host fails to boot due to pending-user-action (this can occur when the disk boot order is set incorrectly) log the message at Warning level instead of Debug to make it obvious.

This also removes an additional spurious info message that was being logged.


rwsu (approved):
/approve

PR #7212, opened 2023-05-29
AGENT-596: use agent-installer-utils for agent-tui extraction

andfasano:
Since `agent-installer-utils` image is now shipped within the release payload, it's then possible to start using it (instead of the temporary solution provided by `agent-installer-node-agent`) for extracting the agent-tui related artifacts


andfasano:
Latest 4.14 okd release included agent-installer-utils, so bumping up the default release image

rwsu (commented):
/lgtm

PR #7213, opened 2023-05-30
OCPBUGS-14818: disable oVirt provider

engelmi:
This PR disables oVirt as a supported platform by 
- removing it as an option from the interactive install
- and returning an validation error when using an `install-config.yaml`

### DIFF
--- a/pkg/asset/installconfig/platform.go
+++ b/pkg/asset/installconfig/platform.go
@@ -18,7 +18,6 @@ import (
 	libvirtconfig "github.com/openshift/installer/pkg/asset/installconfig/libvirt"
 	nutanixconfig "github.com/openshift/installer/pkg/asset/installconfig/nutanix"
 	openstackconfig "github.com/openshift/installer/pkg/asset/installconfig/openstack"
-	ovirtconfig "github.com/openshift/installer/pkg/asset/installconfig/ovirt"
 	powervsconfig "github.com/openshift/installer/pkg/asset/installconfig/powervs"
 	vsphereconfig "github.com/openshift/installer/pkg/asset/installconfig/vsphere"
 	"github.com/openshift/installer/pkg/types"
@@ -101,10 +100,7 @@ func (a *platform) Generate(asset.Parents) error {
 			return err
 		}
 	case ovirt.Name:
-		a.Ovirt, err = ovirtconfig.Platform()
-		if err != nil {
-			return err
-		}
+		return fmt.Errorf("platform oVirt is no longer supported")
 	case vsphere.Name:
 		a.VSphere, err = vsphereconfig.Platform()
 		if err != nil {
--- a/pkg/types/ovirt/validation/platform.go
+++ b/pkg/types/ovirt/validation/platform.go
@@ -1,105 +1,20 @@
 package validation
 
 import (
-	"fmt"
-
-	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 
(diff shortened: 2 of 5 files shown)


pkg/asset/installconfig/platform.go
```diff
@@ -101,10 +100,7 @@ func (a *platform) Generate(asset.Parents) error {
 			return err
 		}
 	case ovirt.Name:
-		a.Ovirt, err = ovirtconfig.Platform()
-		if err != nil {
-			return err
-		}
+		return fmt.Errorf("platform oVirt is not supported anymore")
```
barbacbd, line 103:
```suggestion
		return fmt.Errorf("platform oVirt is no longer supported")
```

pkg/types/ovirt/validation/platform.go
```diff
@@ -1,105 +1,20 @@
...
-	case configv1.LoadBalancerTypeOpenShiftManagedDefault, configv1.LoadBalancerTypeUserManaged:
-		return true
-	default:
-		return false
+	return field.ErrorList{
+		&field.Error{
+			Type:     field.ErrorTypeForbidden,
+			Field:    fldPath.String(),
+			BadValue: "Unsupported platform",
+			Detail:   "Platform oVirt is not supported anymore",
```
barbacbd, line 17:
```suggestion
			Detail:   "Platform oVirt is no longer supported",
```

pkg/types/ovirt/validation/platform_test.go
```diff
@@ -51,107 +46,7 @@ func TestValidatePlatform(t *testing.T) {
...
-		},
-		{
-			name: "valid machine pool",
-			platform: func() *ovirt.Platform {
-				p := validPlatform()
-				p.DefaultMachinePlatform = &ovirt.MachinePool{}
-				return p
-			}(),
-			valid: true,
+			expectedError: `^test-path: Forbidden: Platform oVirt is not supported anymore`,
```
barbacbd, line 49:
```suggestion
			expectedError: `^test-path: Forbidden: Platform oVirt is no longer supported`,
```

barbacbd (commented):
This looks good just a couple of nit comments. If this is no longer supported do we require tests in a ovirt subdirectory?

pkg/types/validation/installconfig_test.go
```diff
@@ -590,7 +590,7 @@ func TestValidateInstallConfig(t *testing.T) {
 				c.Platform = types.Platform{}
 				return c
 			}(),
-			expectedError: `^platform: Invalid value: "": must specify one of the platforms \(alibabacloud, aws, azure, baremetal, gcp, ibmcloud, none, nutanix, openstack, ovirt, powervs, vsphere\)$`,
+			expectedError: `^platform: Invalid value: "": must specify one of the platforms \(alibabacloud, aws, azure, baremetal, gcp, ibmcloud, none, nutanix, openstack, powervs, vsphere\)$`,
```
jhixson74, line 593: Have these tests been validated? I don't see where the error message has been changed to reflect this.

engelmi: What do you mean by validated? The unit tests are being run by several steps in the pipeline. It is testing `ValidateInstallConfig`. I also calls `validatePlatform`, which throws that error [here](https://github.com/openshift/installer/blob/master/pkg/types/validation/installconfig.go#L713) and, in turn, uses a [slice of PlatformNames](https://github.com/openshift/installer/blob/master/pkg/types/installconfig.go#L37-L48) to build it.

PR #7216, opened 2023-05-31
OCPBUGS-14932: Update azure cli to 2.40.0+

jinyunma:
Due to [security vulnerability](https://github.com/Azure/azure-cli/security/advisories/GHSA-47xc-9rr2-q7p4) affecting Azure CLI versions previous to 2.40.0, it is recommended to update azure cli to higher version to avoid this issue.

Changes:
1. update azure cli yum repository to rhel8 based repository

### DIFF
--- a/images/installer/Dockerfile.upi.ci.rhel8
+++ b/images/installer/Dockerfile.upi.ci.rhel8
@@ -17,7 +17,7 @@ COPY --from=builder /go/src/github.com/openshift/installer/bin/openshift-install
 COPY --from=builder /go/src/github.com/openshift/installer/upi /var/lib/openshift-install/upi
 
 RUN rpm --import https://packages.microsoft.com/keys/microsoft.asc && \
-    sh -c 'echo -e "[azure-cli]\nname=Azure CLI\nbaseurl=https://packages.microsoft.com/yumrepos/azure-cli\nenabled=1\ngpgcheck=1\ngpgkey=https://packages.microsoft.com/keys/microsoft.asc" >/etc/yum.repos.d/azure-cli.repo' && \
+    sh -c 'echo -e "[packages-microsoft-com-prod]\nname=packages-microsoft-com-prod\nbaseurl=https://packages.microsoft.com/rhel/8/prod\nenabled=1\ngpgcheck=1\ngpgkey=https://packages.microsoft.com/keys/microsoft.asc" >/etc/yum.repos.d/microsoft-prod.repo' && \
     sh -c 'echo -e "[google-cloud-cli]\nname=Google Cloud CLI\nbaseurl=https://packages.cloud.google.com/yum/repos/cloud-sdk-el8-x86_64\nenabled=1\ngpgcheck=1\nrepo_gpgcheck=1\ngpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg\n       https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg" > /etc/yum.repos.d/google-cloud-sdk.repo'
 
 RUN yum update -y && \


r4f4 (commented):
We should also update `images/installer/Dockerfile.upi.ci`. In the Installer CI jobs, we don't even use `.upi.ci.rhel8`, only `.upi.ci`: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L98-L107

jinyunma:
> We should also update `images/installer/Dockerfile.upi.ci`. In the Installer CI jobs, we don't even use `.upi.ci.rhel8`, only `.upi.ci`: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L98-L107

thanks for pointing out this. The repository to install azure cli is different between rhel7 and rhel8. 
https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-linux?pivots=dnf&view=azure-cli-latest#install

The latest azure cli version on rhel7 is 2.38.0, which we use now. In order to use version 2.40.0+, need to install it on rhel8.

seems that base image for `.upi.ci` is rhel-7? could we change to use `.upi.ci.rhel8` or just use it only on azure? any suggestion?

r4f4:
> seems that base image for .upi.ci is rhel-7? could we change to use .upi.ci.rhel8 or just use it only on azure? any suggestion?

My suggestion is: 
1. change the `from: rhel7` to `from: base` in [1]. Note that the rhel7 image has an old golang version so we can't really use it to build the Installer [2]. It was `base` before but there is no explanation of why it was changed to `rhel7` [3]
2. Merge `.upi.ci.rhel8` into `.upi.ci` and get rid of `.upi.ci.rhel8`.

@patrickdillon What do you think?

[1] https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-master.yaml#LL99C15-L99C15
[2] https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L58-L61
[3] https://github.com/openshift/release/commit/a639f0c5ede5514c0894ab2f723cf3aceeec225e

r4f4:
@jupierce any reason we should still be using rhel7 to build `Dockerfile.upi.ci`?

patrickdillon:
> Due to [security vulnerability](https://github.com/Azure/azure-cli/security/advisories/GHSA-47xc-9rr2-q7p4) affecting Azure CLI versions previous to 2.40.0, it is recommended to update azure cli to higher version to avoid this issue.

Doesn't the package manager already install the latest version, so there is no risk of 2.40 being used? It seems like we don't really need this change in order to avoid the security vulnerability.

I am reluctant to pin our version of the azure cli to a specific version. I would rather the CLI be regularly updated so that we can catch issues if they appear with newer versions.

>   change the from: rhel7 to from: base in [1]. 

Would `base` still have a package manager? I'm not sure what is the underlying OS for the base image.

r4f4:
> > change the from: rhel7 to from: base in [1].
> 
> Would `base` still have a package manager? I'm not sure what is the underlying OS for the base image.

@patrickdillon it's just a RHEL 8:
```
$: podman run --rm -it registry.ci.openshift.org/ocp/4.14:base
[root@cea9b89a53d8 /]# cat /etc/redhat-release
Red Hat Enterprise Linux release 8.6 (Ootpa)
[root@cea9b89a53d8 /]# yum --version
4.7.0
  Installed: dnf-0:4.7.0-8.el8.noarch at Tue Sep  6 17:22:20 2022
  Built    : Red Hat, Inc. <http://bugzilla.redhat.com/bugzilla> at Fri Mar 18 15:21:28 2022

  Installed: rpm-0:4.14.3-24.el8_6.x86_64 at Thu Mar 30 19:07:22 2023
  Built    : Red Hat, Inc. <http://bugzilla.redhat.com/bugzilla> at Wed Sep 14 09:12:50 2022
```

r4f4:
> > Due to [security vulnerability](https://github.com/Azure/azure-cli/security/advisories/GHSA-47xc-9rr2-q7p4) affecting Azure CLI versions previous to 2.40.0, it is recommended to update azure cli to higher version to avoid this issue.
> 
> Doesn't the package manager already install the latest version, so there is no risk of 2.40 being used? It seems like we don't really need this change in order to avoid the security vulnerability.

@patrickdillon  Because the container is based on RHEL-7 [1], the latest azcli available is < 2.40. You can check it in the repo we use: https://packages.microsoft.com/yumrepos/azure-cli/Packages/a/. 
That's why Jinyun is trying to move to a RHEL8-based repo, so we can have access to newer versions.

[1] https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L98-L99

r4f4:
@patrickdillon We keep coming back to the same point: why do we have/need `Dockerfile.upi.ci` and `Dockerfile.upi.ci.rhel8`? I don't see `Dockerfile.upi.ci.rhel8` being used anywhere in the `openshift/release` repo and I don't understand why we are using rhel7 as a base for the `upi-installer` container instead of rhel8.

jinyunma:
> I would rather the CLI be regularly updated so that we can catch issues if they appear with newer versions.

make sense, will move the specific version, and always install the latest one.

As Rafael explained, current upi-installer is based on rhel-7, and the latest azure cli version is 2.38.0 in rhel7 [repository](https://packages.microsoft.com/yumrepos/azure-cli/Packages/a/). The version 2.40.0+ which we want to update is only available in rhel8 based [repository](https://packages.microsoft.com/rhel/8/prod/Packages/a/).

so can we change `from: rhel-7` to `from: base` as Rafael suggested and use `Dockerfile.upi.ci.rhel8` [here](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L98-L99)?

patrickdillon:
Excellent. Thanks for explaining (again). 

https://github.com/openshift/installer/pull/7216#issuecomment-1569765876 sounds good. We should move the images to rhel 8, and I agree specifically:

> Merge .upi.ci.rhel8 into .upi.ci and get rid of .upi.ci.rhel8.

And it sounds like we're on the same page to not pin the version. Thanks!

r4f4:
> Excellent. Thanks for explaining (again).
> 
> [#7216 (comment)](https://github.com/openshift/installer/pull/7216#issuecomment-1569765876) sounds good. We should move the images to rhel 8, and I agree specifically:
> 
> > Merge .upi.ci.rhel8 into .upi.ci and get rid of .upi.ci.rhel8.
> 
> And it sounds like we're on the same page to not pin the version. Thanks!

Great, thanks for your feedback. I'll work with Jinyun on this.

r4f4:
@jinyunma Ideally we would update `Dockerfile.upi.ci.rhel8` with some missing bits (the govc image and the ibmcli dir permissions), delete `Dockerfile.upi.ci`, rename `Dockerfile.upi.ci.rhel8` -> `Dockerfile.upi.ci` and then change `from: rhel7` -> `from: base` in the release repo.

However, since we haven't really tested the rhel8-based installer container in CI and the `upi-installer` image is depended on by most of the CI steps, I'm afraid we could cause a lot of breakage.

What if we did the change in parts? We create a new stanza like https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L98-L107 for, let's say, `upi-installer-rhel8` with `from: base` and then we use that image for the Azure steps?

Another option is to change https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L98-L107 to use `images/installer/Dockerfile.upi.ci.rhel8` and then deal with all issues at once.

WDYT?

jinyunma:
@r4f4 thanks for sharing your opinions.

Checked with @jianlinliu @gpei , we prefer to the second option, since it may need to be updated to rhel8 sooner or later and your also mentioned previously `Note that the rhel7 image has an old golang version so we can't really use it to build the Installer`
> to change https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L98-L107 to use images/installer/Dockerfile.upi.ci.rhel8 and then deal with all issues at once.

I will open new PR in release repository to update https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L98-L107, and see if any impact when using `Dockerfile.upi.ci.rhel8` with `from: base`.

jinyunma:
/retitle OCPBUGS-14932: Update azure cli to 2.40.0+

images/installer/Dockerfile.upi.ci.rhel8
```diff
@@ -17,12 +17,12 @@ COPY --from=builder /go/src/github.com/openshift/installer/bin/openshift-install
...
 
 RUN rpm --import https://packages.microsoft.com/keys/microsoft.asc && \
-    sh -c 'echo -e "[azure-cli]\nname=Azure CLI\nbaseurl=https://packages.microsoft.com/yumrepos/azure-cli\nenabled=1\ngpgcheck=1\ngpgkey=https://packages.microsoft.com/keys/microsoft.asc" >/etc/yum.repos.d/azure-cli.repo' && \
+    sh -c 'echo -e "[packages-microsoft-com-prod]\nname=packages-microsoft-com-prod\nbaseurl=https://packages.microsoft.com/rhel/8/prod\nenabled=1\ngpgcheck=1\ngpgkey=https://packages.microsoft.com/keys/microsoft.asc" >/etc/yum.repos.d/microsoft-prod.repo' && \
     sh -c 'echo -e "[google-cloud-cli]\nname=Google Cloud CLI\nbaseurl=https://packages.cloud.google.com/yum/repos/cloud-sdk-el8-x86_64\nenabled=1\ngpgcheck=1\nrepo_gpgcheck=1\ngpgkey=https://packages.cloud.google.com/yum/doc/yum-key.gpg\n       https://packages.cloud.google.com/yum/doc/rpm-package-key.gpg" > /etc/yum.repos.d/google-cloud-sdk.repo'
 
 RUN yum update -y && \
     yum install --setopt=tsflags=nodocs -y \
-      azure-cli \
+      azure-cli-2.49.0-1.el8 \
```
jianlinliu, line 25: Shall we always install the latest version instead of a specific version?

jinyunma: yeah, I plan to change back to always install the latest version once PR https://github.com/openshift/release/pull/39871 get merged, then triggered job can verify new azure cli.

jinyunma:
[job](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/7216/pull-ci-openshift-installer-master-e2e-azure-ovn-upi/1670611083884236800) `e2e-azure-ovn-upi` installation part is successful, az cli version is 2.49.0.
```
az version:
{
  "azure-cli": "2.49.0",
  "azure-cli-core": "2.49.0",
  "azure-cli-telemetry": "1.0.8",
  "extensions": {}
}
```

r4f4:
/lgtm
Install succeeded for all Azure jobs.

r4f4:
/override ci/prow/e2e-azure-ovn-upi
Just a few e2e failures not related to the PR.

r4f4:
/override ci/prow/e2e-vsphere-upi
Just a few e2e failures and not affected by this PR.

PR #7217, opened 2023-05-31
OCPCLOUD-2036: introduce External platform type

rvanderp3:
The intent of this pull request is to allow the definition of the external platform type so that partners are able to more easily integrate with the installer and other components that recognize the external platform type.

cc: @elmiko

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -2502,6 +2502,18 @@ spec:
                 required:
                 - hosts
                 type: object
+              external:
+                description: External is the configuration used when installing on
+                  an external cloud provider.
+                properties:
+                  platformName:
+                    default: Unknown
+                    description: PlatformName holds the arbitrary string representing
+                      the infrastructure provider name, expected to be set at the
+                      installation time. This field is solely for informational and
+                      reporting purposes and is not expected to be used for decision-making.
+                    type: string
+                type: object
               gcp:
                 description: GCP is the configuration used when installing on Google
                   Cloud Platform.
--- a/pkg/terraform/stages/platform/stages.go
+++ b/pkg/terraform/stages/platform/stages.go
@@ -20,6 +20,7 @@ import (
 	awstypes "github.com/openshift/installer/pkg/types/aws"
 	azuretypes "github.com/openshift/installer/pkg/types/azure"
 	baremetaltypes "github.com/openshift/installer/pkg/types/baremetal"
+	externaltypes "github.com/openshift/installer/pkg/types/external"
 	gcptypes "github.com/openshift/installer/pkg/types/gcp"
 	ibmcloudtypes "github.com/openshift/installer/pkg/types/ibmcloud"
 	libvirttypes "github.com/openshift/installer/pkg/types/libvirt"
@@ -62,6 +63,9 @@ func StagesForPlatform(platform string) []terraform.Stage {
 		return vsphere.PlatformStages
 	case nonetypes.Name:
 		// terraform is not used when the platform is "none"
+		return []terraform.Stage{}
+	case externaltypes.Name:
+		// terraform is not used when the platform is "external"
 		return []terraform.Stage{}
 	default:
 		panic(fmt.Sprintf("unsupported platform %q", platform))
--- a/pkg/types/external/doc.go
+++ b/pkg/types/external/doc.go
@@ -1,0 +1,6 @@
+// Package none contains generic structures for installer
+// configuration and management.
+package external
+
+// Name is name for the External platform.
+const Name string = "external"
--- a/pkg/types/external/platform.go
+++ b/pkg/types/external/platform.go
@@ -1,0 +1,12 @@
+package external
+
+// Platform stores configuration related to external cloud providers.
+type Platform struct {
+	// PlatformName holds the arbitrary string representing the infrastructure provider name, expected to be set at the installation time.
+	// This field is solely for informational and reporting purposes and is not expected to be used for decision-making.
+	// +kubebuilder:default:="Unknown"
+	// +default="Unknown"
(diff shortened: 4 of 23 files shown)


pkg/types/external/doc.go
```diff
@@ -0,0 +1,6 @@
+// Package none contains generic structures for installer
+// configuration and management.
+package external
+
+// Name is name for the None platform.
```
elmiko, line 5: should this read:
```suggestion
// Name is name for the External platform.
```
?

elmiko (commented):
i think this makes sense, just a question about one of the comments.

also, i'm assuming we need the unit tests passing?

pkg/types/external/doc.go
rvanderp3: Yep, it should.  I was troubleshooting the unit tests yesterday afternoon and am picking that back up today.

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -2773,6 +2773,10 @@ spec:
                 required:
                 - hosts
                 type: object
+              external:
+                description: External is the empty configuration used when an external
+                  platform provider is used.
+                type: object
```
mtulio, line 2779: Maybe we need to create the properties too?

```suggestion
              external:
                description: External is the configuration used when an external
                    platform provider is used.
                properties:
                  platformName:
                    description: defines the arbitrary string represented cloud
                      provider name.
                    type: string
                type: object
```

pkg/terraform/stages/platform/stages.go
```diff
@@ -63,6 +64,9 @@ func StagesForPlatform(platform string) []terraform.Stage {
 	case nonetypes.Name:
 		// terraform is not used when the platform is "none"
 		return []terraform.Stage{}
+	case external.Name:
```
mtulio, line 67: nit: maybe using the same file convention of `platformTypetypes.Name`:
```suggestion
	case externaltypes.Name:
```

data/data/install.openshift.io_installconfigs.yaml
adriengentil: we might want to add the [CloudControllerManagerState](https://github.com/openshift/api/pull/1434/files#diff-07b264a49084976b670fb699badaca1795027d6ea732a99226a5388104f6174fR347) as well otherwise we won't expect any CCM by default.

rvanderp3: > Maybe we need to create the properties too?

Platform none appears to use the same minimal platform spec.  Are there other properties we want to include?

rvanderp3: > we might want to add the [CloudControllerManagerState](https://github.com/openshift/api/pull/1434/files#diff-07b264a49084976b670fb699badaca1795027d6ea732a99226a5388104f6174fR347) as well otherwise we won't expect any CCM by default.

@adriengentil , we would include this in the infrastructure status, right?

adriengentil: I don't know if it's worth to make it configurable, but yes, it should be in the infrastructure object :+1:

mtulio: > > Maybe we need to create the properties too?
> 
> @mtulio Platform none appears to use the same minimal platform spec. Are there other properties we want to include?

I think the unique attribute we may need to capture in the config is the `platformName`.

I will defer to @adriengentil and @elmiko to double-check.

rvanderp3: yep, that makes sense, i missed that aspect of it

mtulio:
The current version has been successfully tested using external platform type in OCI. The results are detailed here: https://issues.redhat.com/browse/SPLAT-985?focusedId=22553648&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-22553648

data/data/install.openshift.io_installconfigs.yaml
```diff
@@ -2487,6 +2487,18 @@ spec:
                 required:
                 - hosts
                 type: object
+              external:
+                description: External is the configuration used when an external platform
+                  provider is used.
```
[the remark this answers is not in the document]
elmiko: +1 from me

elmiko (commented):
looks like we just need a couple minor fixups

pkg/types/external/platform.go
```diff
@@ -0,0 +1,12 @@
+package external
+
+// Platform stores configuration related to external cloud providers.
+type Platform struct {
+	// PlatformName holds the arbitrary string representing the infrastructure provider name, expected to be set at the installation time.
+	// This field is solely for informational and reporting purposes and is not expected to be used for decision-making.
+	// +kubebuilder:default:="Unknown"
+	// +default="Unknown"
```
patrickdillon, line 8: I don't see this value of "Unknown" being set.

If it's not being set (i.e. I am mistaken), you could set it in the `SetPlatformDefaults` func.

rvanderp3:
/hold

placing hold until feedback is addressed

elmiko (commented):
/lgtm

mtulio:
Linking the Platform External enhancement document: https://github.com/openshift/enhancements/blob/master/enhancements/cloud-integration/infrastructure-external-platform-type.md

rvanderp3:
/hold

we are discussing dropping the tech preview feature gate

elmiko:
> we are discussing dropping the tech preview feature gate

we talked on friday and again today, i think we will remove the FG from this PR before going forward. likewise, we will remove the FG from the related code in the API and CCO as well.

elmiko (commented):
thanks @rvanderp3 !

/lgtm

r4f4:
/override ci/prow/unit
The failure is not related to this PR and is being worked on.

PR #7218, opened 2023-06-01
OCPBUGS-13662: Ignore CPUPartitioning for ABI

pawanpinjarkar:
Print a warning if `CPUPartitioning` is set to a non-default value.


rwsu (commented):
/lgtm

PR #7219, opened 2023-06-01
OCPBUGS-14416: Shorten SNO installation duration by releasing CPC lease

mresvanis:
When installing Single Node OpenShift with bootstrap in place, it takes 6 minutes for the `cluster-policy-controller` to acquire the leader lease. This leads to some OpenShift namespaces not getting the `openshift.io/sa.scc.uid-range` annotation during that time, i.e. no pods can be scheduled in them.  Which in turn leads to delaying the installation completion.

This change deletes the cluster-policy-controller lease, which was created during bootstrapping. This allows the latter to start sooner, without waiting for the lease to expire.

/cc @eranco74 
/cc @tsorya

### DIFF
--- a/data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
+++ b/data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
@@ -24,20 +24,35 @@ function signal_bootstrap_complete {
   done
 }
 
+function release_lease {
+  local ns="$1"
+  local lease="$2"
+
+  until [ "$(oc get leases -n "${ns}" "${lease}" | grep -c "${lease}")" -eq 0 ];
+  do
+    echo "Deleting ${ns} ${lease} lease"
+    oc delete leases -n "${ns}" "${lease}" || sleep 5
+  done
+  until [ "$(oc get cm -n "${ns}" "${lease}" | grep -c "${lease}")" -eq 0 ];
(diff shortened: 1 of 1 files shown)


romfreiman:
Similar to https://github.com/openshift/installer/pull/6757

romfreiman:
@vrutkovs @wking

data/data/bootstrap/bootstrap-in-place/files/opt/openshift/bootstrap-in-place/bootstrap-in-place-post-reboot.sh
```diff
@@ -37,7 +37,24 @@ function release_cvo_lease {
...
     done
-    touch /opt/openshift/expedite_bootstrapping.done
+    touch /opt/openshift/release_cvo_lease.done
+  fi
+}
+
+function release_cpc_lease {
+  if [ ! -f /opt/openshift/release_cpc_lease.done ]
+  then
+    until [ "$(oc get leases -n kube-system cluster-policy-controller-lock | grep -c "cluster-policy-controller-lock")" -eq 0 ];
```
tsorya, line 47: seems like this part is very similar to cvo, maybe better to have a function and provide lease name and namespace?

mresvanis: I tried to move the relevant parts to a common function, it seems a lot cleaner now. Thanks for the pointer @tsorya!

eranco74 (commented):
/lgtm

PR #7223, opened 2023-06-01
AGENT-624: Allow override of networkType

bfournie:
Allow override of the networkType. If the networkType is set to a value other then OpenshiftSDN or OVNKubernetes then the installConfigOverrides will be set to the configured value.

### DIFF
--- a/pkg/asset/agent/manifests/agentclusterinstall.go
+++ b/pkg/asset/agent/manifests/agentclusterinstall.go
@@ -77,6 +77,8 @@ type agentClusterInstallInstallConfigOverrides struct {
 	Capabilities *types.Capabilities `json:"capabilities,omitempty"`
 	// AdditionalTrustBundle must be set here when mirroring not configured
 	AdditionalTrustBundle string `json:"additionalTrustBundle,omitempty"`
+	// Allow override of network type
+	Networking *types.Networking `json:"networking,omitempty"`
 }
 
 var _ asset.WritableAsset = (*AgentClusterInstall)(nil)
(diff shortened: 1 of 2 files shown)


pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -190,7 +192,12 @@ func (a *AgentClusterInstall) Generate(dependencies asset.Parents) error {
 			agentClusterInstall.Spec.IngressVIP = installConfig.Config.Platform.VSphere.IngressVIPs[0]
 		}
 
-		setNetworkType(agentClusterInstall, installConfig.Config, "NetworkType is not specified in InstallConfig.")
+		networkOverridden, networkTypeOverride := setNetworkType(agentClusterInstall, installConfig.Config, "NetworkType is not specified in InstallConfig.")
+		if networkOverridden {
+			icOverridden = true
+			icOverrides.Networking = installConfig.Config.Networking
+			icOverrides.Networking.NetworkType = networkTypeOverride
```
zaneb, line 199: This line is redundant I think?

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -298,21 +305,26 @@ func (a *AgentClusterInstall) finish() error {
...
 		installConfig.Networking.NetworkType != "" {
-		aci.Spec.Networking.NetworkType = installConfig.NetworkType
-		return
+		if installConfig.Networking.NetworkType == string(operv1.NetworkTypeOVNKubernetes) || installConfig.Networking.NetworkType == string(operv1.NetworkTypeOpenShiftSDN) {
+			aci.Spec.Networking.NetworkType = installConfig.NetworkType
+			return false, ""
+		}
+
+		aci.Spec.Networking.NetworkType = string(operv1.NetworkTypeOVNKubernetes)
+		return true, installConfig.NetworkType
```
zaneb, line 321: This is a bit confusing because `installConfig.NetworkType` and `installConfig.Networking.NetworkType` are actually the same field.
I don't think we need to return this here anyway though.

bfournie: Yeah I overcomplicated that a bit, fixed.

pkg/asset/agent/manifests/agentclusterinstall.go
bfournie: Yeah, fixed.

zaneb (commented):
/approve

PR #7232, opened 2023-06-05
tls/root: Document this more and change friendly name

cgwalters:
Reflect in this my current understanding of reality.

See also https://github.com/openshift/api/pull/1484


cgwalters:
This is part of https://issues.redhat.com/browse/MCO-642
I think since this is just basically changing comment text and the text output in debug logging we could just skip the jira/valid-bug step?

PR #7234, opened 2023-06-06
OCPBUGS-14121: Convert Rendezvous IPv6 address to canonical format

bfournie:
If the RendezvousIP is provided as a non-canonical (i.e. not compressed) IPv6 address then the IP comparison on the host will fail. Use ParseIp() address to convert this to canonical format. Note that this function will not affect the IPv4 address format.

### DIFF
--- a/pkg/asset/agent/image/ignition.go
+++ b/pkg/asset/agent/image/ignition.go
@@ -498,8 +498,16 @@ func RetrieveRendezvousIP(agentConfig *agent.Config, nmStateConfigs []*v1beta1.N
 		logrus.Debug("RendezvousIP from the NMStateConfig ", rendezvousIP)
 	} else {
 		err = errors.New("missing rendezvousIP in agent-config or at least one NMStateConfig manifest")
+		return "", err
 	}
-	return rendezvousIP, err
+
+	// Convert IPv6 address to canonical to match host format for comparisons
+	addr := net.ParseIP(rendezvousIP)
+	if addr == nil {
+		err = errors.New(fmt.Sprintf("invalid rendezvous IP: %s", rendezvousIP))
+		return "", err
+	}
+	return addr.String(), err
 }
 
 func getPublicContainerRegistries(registriesConfig *mirror.RegistriesConf) string {
--- a/pkg/asset/agent/image/ignition_test.go
+++ b/pkg/asset/agent/image/ignition_test.go
@@ -231,6 +231,32 @@ func TestRetrieveRendezvousIP(t *testing.T) {
 			},
 			expectedError: "missing rendezvousIP in agent-config or at least one NMStateConfig manifest",
 		},
+		{
+			Name: "non-canonical-ipv6-address",
+			agentConfig: &agent.Config{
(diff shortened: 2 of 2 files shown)


bfournie:
Output from test of agent create image
```
level=debug msg=RendezvousIP from the NMStateConfig **fd2e:6f44:5dd8:c956:0000:0000:0000:0050**
level=info msg=The rendezvous host IP (node0 IP) is **fd2e:6f44:5dd8:c956::50**
```

pkg/asset/agent/image/ignition.go
```diff
@@ -499,7 +499,10 @@ func RetrieveRendezvousIP(agentConfig *agent.Config, nmStateConfigs []*v1beta1.N
 	} else {
 		err = errors.New("missing rendezvousIP in agent-config or at least one NMStateConfig manifest")
 	}
-	return rendezvousIP, err
+
+	// Convert IPv6 address to canonical to match host format for comparisons
+	addr := net.ParseIP(rendezvousIP)
```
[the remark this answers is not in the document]
bfournie: Right, will add a check

pkg/asset/agent/image/ignition.go
```diff
@@ -499,7 +499,16 @@ func RetrieveRendezvousIP(agentConfig *agent.Config, nmStateConfigs []*v1beta1.N
 	} else {
 		err = errors.New("missing rendezvousIP in agent-config or at least one NMStateConfig manifest")
```
zaneb, line 500: nit: I'd be inclined to `return "", err` here rather than check for the error later.

zaneb (commented):
/approve

pkg/asset/agent/image/ignition_test.go
```diff
@@ -231,6 +231,19 @@ func TestRetrieveRendezvousIP(t *testing.T) {
...
+				RendezvousIP: "fd2e:6f44:5dd8:c956:0000:0000:0000:0050",
+				Hosts: []agent.Host{
+					{
+						Hostname: "control-0.example.org",
+						Role:     "master",
+					},
+				},
+			},
+			expectedRendezvousIP: "fd2e:6f44:5dd8:c956::50",
+		},
```
andfasano, line 246: Another case with an invalid ip could be useful

pkg/asset/agent/image/ignition.go
```diff
@@ -499,7 +499,16 @@ func RetrieveRendezvousIP(agentConfig *agent.Config, nmStateConfigs []*v1beta1.N
...
 	}
-	return rendezvousIP, err
+
+	// Convert IPv6 address to canonical to match host format for comparisons
+	addr := net.ParseIP(rendezvousIP)
+	if addr == nil {
+		if err == nil {
+			err = errors.New("invalid rendezvous IP")
```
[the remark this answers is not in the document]
bfournie: Added

pkg/asset/agent/image/ignition.go
bfournie: > nit: I'd be inclined to `return "", err` here rather than check for the error later.

Changed to return earlier.

PR #7236, opened 2023-06-07
AGENT-648: Remove validation check limiting None platform to SNO

pawanpinjarkar:
- For none platform with only one master and zero workers, only the network type `OVNKubernetes` is allowed.
- none platform can now have ControlPlane.Replicas>1 and workers>0
- Set UserManagedNetworking to true for none platform
- Remove redundant integration test when supporting non-SNO cluster topologies for none platform

### DIFF
--- a/pkg/asset/agent/installconfig.go
+++ b/pkg/asset/agent/installconfig.go
@@ -133,29 +133,21 @@ func (a *OptionalInstallConfig) validateSNOConfiguration(installConfig *types.In
 		workers = workers + int(*worker.Replicas)
 	}
 
-	//  platform None always imply SNO cluster
-	if installConfig.Platform.Name() == none.Name {
-		if installConfig.Networking.NetworkType != "OVNKubernetes" {
-			fieldPath = field.NewPath("Networking", "NetworkType")
-			allErrs = append(allErrs, field.Invalid(fieldPath, installConfig.Networking.NetworkType, "Only OVNKubernetes network type is allowed for Single Node OpenShift (SNO) cluster"))
-		}
-
-		if *installConfig.ControlPlane.Replicas != 1 {
-			fieldPath = field.NewPath("ControlPlane", "Replicas")
-			allErrs = append(allErrs, field.Required(fieldPath, fmt.Sprintf("ControlPlane.Replicas must be 1 for %s platform. Found %v", none.Name, *installConfig.ControlPlane.Replicas)))
-		}
-
-		if workers != 0 {
+	if installConfig.ControlPlane != nil && *installConfig.ControlPlane.Replicas == 1 {
+		if workers == 0 {
+			if installConfig.Platform.Name() == none.Name && installConfig.Networking.NetworkType != "OVNKubernetes" {
+				fieldPath = field.NewPath("Networking", "NetworkType")
+				allErrs = append(allErrs, field.Invalid(fieldPath, installConfig.Networking.NetworkType, "Only OVNKubernetes network type is allowed for Single Node OpenShift (SNO) cluster"))
+			}
+			if installConfig.Platform.Name() != none.Name {
+				fieldPath = field.NewPath("Platform")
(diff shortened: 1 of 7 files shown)


rwsu (commented):
/lgtm

cs-zhang:
Just tested the patch, got error to register the cluster for compact install:
Jun 13 21:07:22 master-1 create-cluster-and-infraenv[4682]: time="2023-06-13T21:07:22Z" level=fatal msg="Failed to register cluster with assisted-service: AssistedServiceError Code: 400 Href:  ID: 400 Kind: Error Reason: Can't set none platform with user-managed-networking disabled"
If platform set to none, user-managed-network should be set to true.

bfournie:
Looks like the integration test needs to change https://github.com/openshift/installer/blob/master/cmd/openshift-install/testdata/agent/image/validations/sno_invalid_workers.txt#L3

bfournie:
Also looks like will need to set UserManagedNetworking in the AgentClusterInstall networking, https://github.com/openshift/installer/blob/master/pkg/asset/agent/manifests/agentclusterinstall.go#L146 to get past the issue that @cs-zhang reported.

bfournie:
In the Load() for AgentClusterInstall I think it should also be set to `True` if its not set (as its a pointer) and platform is `None`

cmd/openshift-install/testdata/agent/image/configurations/sno.txt
```diff
@@ -13,10 +13,10 @@ apiVersion: v1
 baseDomain: test.metalkube.org
 controlPlane: 
   name: master
-  replicas: 1
+  replicas: 3
```
rwsu, line 16: Because this is named sno.txt, it is odd the masters replicas is changed to 3. I think we should leave this one unchanged and create another integration test for none platform and either compact or HA cluster configuration.

cmd/openshift-install/testdata/agent/image/validations/sno_invalid_workers.txt
```diff
@@ -1,40 +0,0 @@
-! exec openshift-install agent create image --dir $WORK
-
-stderr 'Total number of Compute.Replicas must be 0 for none platform. Found 1'
```
rwsu, line 3: Having master replicas = 1 and compute replicas = 1 is an invalid configuration. I think the validation should still heck this. Perhaps the error or validation message should be changed.

rwsu (changes requested):

rwsu (commented):
/lgtm

pkg/asset/agent/installconfig.go
```diff
@@ -133,29 +133,21 @@ func (a *OptionalInstallConfig) validateSNOConfiguration(installConfig *types.In
...
-		if workers != 0 {
+	if installConfig.ControlPlane != nil && *installConfig.ControlPlane.Replicas == 1 {
+		if workers == 0 {
+			if installConfig.Platform.Name() == none.Name && installConfig.Networking.NetworkType != "OVNKubernetes" {
+				fieldPath = field.NewPath("Networking", "NetworkType")
+				allErrs = append(allErrs, field.Invalid(fieldPath, installConfig.Networking.NetworkType, "Only OVNKubernetes network type is allowed for Single Node OpenShift (SNO) cluster"))
+			}
+			if installConfig.Platform.Name() != none.Name {
+				fieldPath = field.NewPath("Platform")
+				allErrs = append(allErrs, field.Invalid(fieldPath, installConfig.Platform.Name(), fmt.Sprintf("Platform should be set to %s if the ControlPlane.Replicas is %d and total number of Compute.Replicas is %d", none.Name, *installConfig.ControlPlane.Replicas, workers)))
```
andfasano, line 144:
```suggestion
				allErrs = append(allErrs, field.Invalid(fieldPath, installConfig.Platform.Name(), fmt.Sprintf("Only platform %s supports 1 ControlPlane and 0 Compute nodes", none.Name)))
```

rwsu (approved):
/approve

andfasano:
Tide seems to be stuck? Trying to refresh the PR

/jira refresh

PR #7239, opened 2023-06-09
OSASINFRA-3182: openstack: remove portTargets

EmilienM:
The FailureDomain implementation in 4.13 was experimental and delivered
as TechPreview.

In 4.14, we'll deliver a stable FailureDomain API for OpenStack that
will be used in CPMS.

The `porttargets` will have to be re-worked so we can't continue to have
it in the installer since its interface is very likely going to change
for something else in the future.

So we decided to remove it so we can switch the FailureDomain to use the
stable API that was merged in 4.14.


data/data/install.openshift.io_installconfigs.yaml
EmilienM, line 1: I didn't remove `failureDomains` from this file yet.

pierreprinetti: it's done

pkg/tfvars/openstack/ports.go
```diff
@@ -21,85 +9,3 @@ type terraformPort struct {
 	NetworkID string             `json:"network_id"`
 	FixedIP   []terraformFixedIP `json:"fixed_ips"`
 }
-
-func portTargetToTerraformPort(networkClient *gophercloud.ServiceClient, portTarget types_openstack.PortTarget) (terraformPort, error) {
```
MaysaMacedo, line 25: This is being used for the dual-stack support:
https://github.com/openshift/installer/pull/6797/files#diff-069c83ecc685bc97ef280d65021909c03ea67f51f67a998bb3a5cb647b8cf16bR183

pkg/types/openstack/machinepool.go
```diff
@@ -95,56 +85,3 @@ type RootVolume struct {
...
-// intended purpose. If ID is set to "control-plane", then the PortTarget will
-// replace the default cluster primary network (or the machinesSubnet if
-// defined) as the first network for the machine.
-type NamedPortTarget struct {
-	ID         string `json:"id"`
-	PortTarget `json:",inline"`
-}
-
-// PortTarget defines, directly or indirectly, one or more subnets where to attach a port.
-type PortTarget struct {
```
MaysaMacedo, line 132: PortTarget is being used for the dual-stack support:
https://github.com/openshift/installer/pull/6797/files#diff-0b0e430b5843d7e62d8138813bc0517be9584d7c77e3ad30686b3fc3dde02e21L132-L138

EmilienM:
@patrickdillon before requesting it to be in the OpenShift 4.14 Feature Merge Queue, could you or someone else in your team please look at this patch and approve it if it's fine.
Thanks

pkg/types/openstack/machinepool.go
pierreprinetti: Or even better if we can merge your PR first

patrickdillon:
/approve

nit: it would be best to separate vendoring into a separate commit. fwiw I have found the pattern described in [dependencies.md](https://github.com/openshift/installer/blob/master/docs/dev/dependencies.md), go.(mod,sum) in a separate commit, to be personally very useful.

pierreprinetti:
> /approve
> 
> nit: it would be best to separate vendoring into a separate commit. fwiw I have found the pattern described in [dependencies.md](https://github.com/openshift/installer/blob/master/docs/dev/dependencies.md), go.(mod,sum) in a separate commit, to be personally very useful.

So that would be
* one commit for go.mod, go.sum and vendor/; OR
* one commit for go.{mod,sum} and another for vendor/
?

Where it gets complicated, is when a dep bump requires you to update some function call. Would you commit that together with the bump? I personally would.

However I understand that this PR is the other way around: a change in production code depends on a bump, and not the opposite.

patrickdillon:
This: 
> one commit for go.{mod,sum} and another for vendor/

Typically we separate the vendor code to make it easier to review. Having go.(mod,sum) in a separate commit makes it easier to review those changes, and I have also found it easier for my development workflows

> Where it gets complicated, is when a dep bump requires you to update some function call. Would you commit that together with the bump? I personally would.

No. I would expect most vendoring to have accompanying code changes, but the general guidelines are to separate all vendoring and I have not seen any issues with that.

pierreprinetti:
> This:
> 
> > one commit for go.{mod,sum} and another for vendor/
> 
> Typically we separate the vendor code to make it easier to review. Having go.(mod,sum) in a separate commit makes it easier to review those changes, and I have also found it easier for my development workflows
> 
> > Where it gets complicated, is when a dep bump requires you to update some function call. Would you commit that together with the bump? I personally would.
> 
> No. I would expect most vendoring to have accompanying code changes, but the general guidelines are to separate all vendoring and I have not seen any issues with that.

OK! I'll try to comply. Thank you for the explanation!

PR #7250, opened 2023-06-15
Destroy startironic.sh forever

zaneb:
Explode the `startironic.sh` script - a tangled mess containing everything needed to do baremetal provisioning of the control plane, entwined with homegrown `rc`-style process management - into separate services managed by systemd.


zaneb:
Failures were in the conformance tests; the cluster is still installing successfully in both cases.
/retest-required

data/data/bootstrap/baremetal/files/etc/containers/systemd/image-customization.container
```diff
@@ -0,0 +1,36 @@
...
+Environment="IRONIC_BASE_URL=${IRONIC_BASE_URL}"
+Environment="IRONIC_RAMDISK_SSH_KEY=${IRONIC_RAMDISK_SSH_KEY}"
+Environment="IRONIC_AGENT_IMAGE=${IRONIC_AGENT_IMAGE}"
+Environment="IP_OPTIONS=${EXTERNAL_IP_OPTIONS}"
+Environment="REGISTRIES_CONF_PATH=/tmp/containers/registries.conf"
+
+[Service]
+EnvironmentFile=/etc/ironic.env
+Environment="NMSTATE_DIR=/tmp/nmstate"
+ExecStartPre=/usr/local/bin/setup-image-data.sh
```
dtantsur, line 29: I wonder why cannot this be a service in the same spirit as everything else.

data/data/bootstrap/baremetal/files/etc/containers/systemd/image-customization.container
```diff
@@ -0,0 +1,36 @@
+[Unit]
+Description=Customized Machine OS Image Server
+BindsTo=ironic-volume.service
+Requires=build-ironic-env.service extract-machine-os.service
+PartOf=ironic.service
+Wants=network-online.target
+After=network-online.target ironic-volume.service build-ironic-env.service extract-machine-os.service
+
+[Container]
```
dtantsur, line 9: Curious: is it some openshift or podman feature? I cannot find any references to the [Container] section.

data/data/bootstrap/baremetal/files/usr/local/bin/dhcp-filter.sh
```diff
@@ -0,0 +1,41 @@
...
+    done
+
+    $IPTABLES -t raw -A DHCP_IRONIC -j DROP
+}
+
+remove_filter() {
+    "$IPTABLES-save" -t raw | grep -v DHCP_IRONIC | "$IPTABLES-restore"
+}
+
+if [ -n "${DHCP_ALLOW_MACS}" ]; then
```
dtantsur, line 28: nit: this script can fail if DHCP_ALLOW_MACS is not defined (I guess it cannot happen in reality).

dtantsur: Well, actually... where is this even defined? I don't see it in ironic-dnsmasq...

data/data/bootstrap/baremetal/files/etc/containers/systemd/image-customization.container
```diff
@@ -0,0 +1,36 @@
...
+Image=$CUSTOMIZATION_IMAGE
+Exec=/image-customization-server --nmstate-dir=${NMSTATE_DIR} --images-publish-addr=http://0.0.0.0:8084
+Network=host
+PodmanArgs=--privileged --secret pull-secret,mode=400
+Volume=${NMSTATE_DIR}/:${NMSTATE_DIR}/
+Volume=ironic.volume:/shared:z
+Volume=/etc/containers:/tmp/containers:z
+Environment="DEPLOY_ISO=/shared/html/images/ironic-python-agent.iso"
+Environment="DEPLOY_INITRD=/shared/html/images/ironic-python-agent.initramfs"
+Environment="IRONIC_BASE_URL=${IRONIC_BASE_URL}"
```
dtantsur, line 20: This is not coming from ironic.env, is it? I find tracing the variables quite confusing...

data/data/bootstrap/baremetal/systemd/units/wait-iptables-init.service.template
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Wait for iptables to be initialised
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+# Create a symlink to the correct iptables binary for the provisioning network
+ExecStartPre=ln -fs /usr/sbin/ip{{if .PlatformData.BareMetal.ProvisioningIPv6}}6{{end}}tables /usr/local/bin/prov-iptables
```
dtantsur, line 8: OMG, please no. Let's create a templated script like dhcp-filter.sh.

data/data/bootstrap/baremetal/systemd/units/wait-iptables-init.service.template
```diff
@@ -0,0 +1,18 @@
...
+# Apparently network-online doesn't necessarily mean iptables is ready, so
+# wait until it is.
```
dtantsur, line 10: Sigh. SIGH.

data/data/bootstrap/baremetal/files/etc/ironic.env.template
```diff
@@ -0,0 +1,9 @@
+AUTH_DIR=/opt/metal3/auth
+HTTP_PORT=6180
+# This DHCP range is used by dnsmasq to serve DHCP to the cluster. If empty
+# dnsmasq will only serve TFTP, and DHCP will be disabled.
+DHCP_RANGE="{{.PlatformData.BareMetal.ProvisioningDHCPRange}}"
+DHCP_ALLOW_MACS="{{.PlatformData.BareMetal.ProvisioningDHCPAllowList}}"
+# Used by ironic to allow ssh to running IPA instances
+IRONIC_RAMDISK_SSH_KEY="{{.SSHKey}}"
+IPTABLES={{if .PlatformData.BareMetal.ProvisioningIPv6}}ip6tables{{else}}iptables{{end}}
```
dtantsur, line 9: I'd rather prefer everything to be handled by built-ironic-env.sh rather than having two places to keep in mind.

data/data/bootstrap/baremetal/files/etc/containers/systemd/ironic.container.template
```diff
@@ -0,0 +1,38 @@
+[Unit]
+Description=Ironic baremetal deployment service
+BindsTo=ironic-volume.service
+Requires=build-ironic-env.service wait-iptables-init.service image-customization.service
+{{ if .PlatformData.BareMetal.ProvisioningIP }}
+Requires=provisioning-interface.service
```
dtantsur, line 6: You probably want After= in addition to Requires=

data/data/bootstrap/baremetal/files/etc/containers/systemd/ironic-dnsmasq.container
```diff
@@ -0,0 +1,27 @@
+[Unit]
+Description=DHCP Service for Provisioning Network
+BindsTo=ironic-volume.service
+Requires=build-ironic-env.service provisioning-interface.service
```
dtantsur, line 4: You probably want After= in addition to Requires=

data/data/bootstrap/baremetal/files/etc/containers/systemd/ironic.container.template
```diff
@@ -0,0 +1,38 @@
...
+{{ end }}
+{{ if .PlatformData.BareMetal.ProvisioningDNSMasq }}
```
dtantsur, line 8: nit: IIUC you can leave Wants= against a non-existing service

data/data/bootstrap/baremetal/files/etc/containers/systemd/ironic.container.template
```diff
@@ -0,0 +1,38 @@
...
+{{ end }}
+Wants=network-online.target ironic-httpd.service ironic-inspector.service ironic-ramdisk-logs.service
+After=network-online.target ironic-volume.service build-ironic-env.service wait-iptables-init.service provisioning-interface.service ironic-dnsmasq.service image-customization.service
+
+[Container]
+ContainerName=ironic
+Image=$IRONIC_IMAGE
+Exec=/bin/runironic
+Network=host
+PodmanArgs=--privileged
```
dtantsur, line 19: Mental note: get rid of this, only dnsmasq needs to be privileged

dtantsur:
This is a great effort, thank you! I've left some comments inline, but in general I really like the direction.

data/data/bootstrap/baremetal/files/etc/containers/systemd/ironic-dnsmasq.container
zaneb: Fixed.

data/data/bootstrap/baremetal/files/etc/containers/systemd/ironic.container.template
zaneb: I've now removed all the --privileged args.

data/data/bootstrap/baremetal/systemd/units/wait-iptables-init.service.template
dtantsur: I don't see creating scripts in runtime as a good idea, especially if some other scripts are created in advance.

dtantsur: Even worse, the inspector service implicitly depends on this script...

data/data/bootstrap/baremetal/files/etc/ironic.env.template
dtantsur: Why would you have problems with shellcheck? From its perspective, `IRONIC_RAMDISK_SSH_KEY="{{.SSHKey}}"` is just setting a variable to a string. I don't think bash interprets {{ in strings?

zaneb: You have to rename the file to `.sh.template`

data/data/bootstrap/baremetal/systemd/units/wait-iptables-init.service.template
zaneb: Done.

data/data/bootstrap/baremetal/files/usr/local/bin/dhcp-filter.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# It is possible machine-api-operator comes up while the bootstrap is
+# online, meaning there could be two DHCP servers on the network. To
+# avoid bootstrap responding to a worker, which would cause a failed
+# deployment, we filter out requests from anyone else than the control
+# plane.  We are using iptables instead of dnsmasq's dhcp-host because
```
dtantsur, line 9: One day, we need to get rid of MAC address requirement for Redfish (but we're not there yet)

dtantsur:
/approve

I'm not fond of some details, but we can iterate (rather than try to make this already large patch perfect)

PR #7251, opened 2023-06-15
STOR-1353: Cleanup GCP Filestore instances on destroy

tsmetana:
Make use of https://github.com/openshift/gcp-filestore-csi-driver/pull/17 and https://github.com/openshift/gcp-filestore-csi-driver-operator/pull/38 to cleanup also GCP Filestore instances created dynamically by the CSI Filestore driver.


sadasu:
@tsmetana are you still interested in this PR?

tsmetana:
> @tsmetana are you still interested in this PR?

Sorry for the slow response: I failed to test the PR (I seem to be unable to install custom GCP cluster), but yes, we still would like to have this included. It's not super-urgent, but would be good to have to keep all the platforms behaving consistently.

tsmetana:
I seem to be unable to install GCP with my patched installer. However I think the patch should be safe enough not to break things.

pkg/destroy/gcp/filestore.go
```diff
@@ -0,0 +1,82 @@
...
+	return result, nil
+}
+
+func (o *ClusterUninstaller) deleteFilestore(ctx context.Context, item cloudResource) error {
+	o.Logger.Debug("Deleting filestore %s", item.name)
+
+	ctx, cancel := context.WithTimeout(ctx, defaultTimeout)
+	defer cancel()
+	instSvc := file.NewProjectsLocationsInstancesService(o.fileSvc)
+	_, err := instSvc.Delete(item.name).Context(ctx).Do()
```
r4f4, line 60: What about a debug log message about the filestore being deleted?
```suggestion
	_, err := instSvc.Delete(item.name).Context(ctx).Do()
	if err == nil {
	    o.Logger.Debug("Deleted filestore %s", item.name)
	}
```

pkg/destroy/gcp/filestore.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/pkg/errors"
+	"google.golang.org/api/file/v1"
+)
+
+func (o *ClusterUninstaller) filestoreParentPath() string {
+	return fmt.Sprintf("projects/%s/locations/-", o.ProjectID)
+}
+
+func (o *ClusterUninstaller) clusterFilestoreLabelFilter() string {
+	return fmt.Sprintf("labels.kubernetes-io-cluster-%s = \"owned\"", o.ClusterID)
```
bharath-b-rh, line 16: nit: Recently a new constant was introduced in `github.com/openshift/installer/pkg/constants/gcp` package for OCP default GCP label format `ClusterIDLabelFmt = "kubernetes-io-cluster-%s"`. I feel we could make use of it.

pkg/destroy/gcp/filestore.go
r4f4: Actually, it could even be `o.Logger.Infof` instead of debug.

pkg/destroy/gcp/filestore.go
```diff
@@ -0,0 +1,82 @@
+package gcp
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
tsmetana: Ack. Will fix that.

pkg/destroy/gcp/filestore.go
tsmetana: dtto

pkg/destroy/gcp/filestore.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (o *ClusterUninstaller) deleteFilestore(ctx context.Context, item cloudResource) error {
+	o.Logger.Debug("Deleting filestore %s", item.name)
+
+	ctx, cancel := context.WithTimeout(ctx, defaultTimeout)
+	defer cancel()
+	instSvc := file.NewProjectsLocationsInstancesService(o.fileSvc)
+	_, err := instSvc.Delete(item.name).Context(ctx).Do()
+	if err == nil {
+		o.Logger.Debug("Deleted filestore %s", item.name)
```
r4f4, line 62:
```suggestion
		o.Logger.Debugf("Deleted filestore %s", item.name)
```
my bad, I had a typo in my suggestion.

pkg/destroy/gcp/filestore.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		nextPageToken = instances.NextPageToken
+		lCall.PageToken(nextPageToken)
+	}
+	return result, nil
+}
+
+func (o *ClusterUninstaller) deleteFilestore(ctx context.Context, item cloudResource) error {
+	o.Logger.Debug("Deleting filestore %s", item.name)
```
r4f4, line 55:
```suggestion
	o.Logger.Debugf("Deleting filestore %s", item.name)
```

pkg/destroy/gcp/filestore.go
```diff
@@ -0,0 +1,85 @@
...
+			return nil, fmt.Errorf("error retrieving filestore instances: %w", err)
+		}
+
+		for _, activeInstance := range instances.Instances {
+			if len(activeInstance.FileShares) == 0 {
+				// skip multi-share instances
+				o.Logger.Debug("Skipping multi-share filestore %s", activeInstance.Name)
+				continue
+			}
+			o.Logger.Debug("Found filestore %s", activeInstance.Name)
```
r4f4, line 41:
```suggestion
			o.Logger.Debugf("Found filestore %s", activeInstance.Name)
```

pkg/destroy/gcp/filestore.go
```diff
@@ -0,0 +1,85 @@
...
+	for nextPageToken != "" {
+		instances, err := lCall.Do()
+		if err != nil {
+			return nil, fmt.Errorf("error retrieving filestore instances: %w", err)
+		}
+
+		for _, activeInstance := range instances.Instances {
+			if len(activeInstance.FileShares) == 0 {
+				// skip multi-share instances
+				o.Logger.Debug("Skipping multi-share filestore %s", activeInstance.Name)
```
r4f4, line 38:
```suggestion
				o.Logger.Debugf("Skipping multi-share filestore %s", activeInstance.Name)
```

pkg/destroy/gcp/filestore.go
```diff
@@ -0,0 +1,92 @@
...
+
+	ctx, cancel := context.WithTimeout(ctx, defaultTimeout)
+	defer cancel()
+	instSvc := file.NewProjectsLocationsInstancesService(o.fileSvc)
+	_, err := instSvc.Delete(item.name).Context(ctx).Do()
+	if err != nil && !isNoOp(err) {
+		o.resetRequestID(item.typeName, item.name)
+		return fmt.Errorf("failed to delete filestore %s", item.name)
+	}
+	if err != nil && isNoOp(err) {
```
r4f4, line 66: Shouldn't this be
```suggestion
	if err == nil || isNoOp(err) {
```
so you also remove the filestore from the pending items in case of successful `Delete`?

tsmetana: Yes, this looks weird at the first sight. I'm almost sure I copied that logic from some other place (disk.go maybe?). The `destroyFilestores` method gets called in a loop, which means the item would get removed from pending in the next attempt (essentially verifying the filestore really doesn't exist, since the `Delete` call should return 404).

I even wonder if this whole `if` block is needed: at this point it's clear the call either succeeded or the error is benign (`isNoOp`) so there's probably no point resetting the request and we can happily remove the item from pending and log success.

tsmetana:
The latest changes were successfully tested (thanks @r4f4 for the idea how to do it).

r4f4 (commented):
/approve

tsmetana:
Sorry for the further delays. I need to check the behaviour in case of missing permissions for Filestore editing.

tsmetana:
There's a problem with destroying the clusters when therea are missing premissions for Filestore. There is no timeout it seems so the uninstallation would keep trying to remove the Filestore instances indefinitely. Since the Filestore operator is an optional component it is more likely for this situation to happen and the UX is pretty bad in such a case.

r4f4:
> There's a problem with destroying the clusters when therea are missing premissions for Filestore. There is no timeout it seems so the uninstallation would keep trying to remove the Filestore instances indefinitely. Since the Filestore operator is an optional component it is more likely for this situation to happen and the UX is pretty bad in such a case.

Can you check the error returned and fail the uninstall in that case? This would give users the chance to add the permission and retry the uninstall.

tsmetana:
> > There's a problem with destroying the clusters when therea are missing premissions for Filestore. There is no timeout it seems so the uninstallation would keep trying to remove the Filestore instances indefinitely. Since the Filestore operator is an optional component it is more likely for this situation to happen and the UX is pretty bad in such a case.
> 
> Can you check the error returned and fail the uninstall in that case? This would give users the chance to add the permission and retry the uninstall.

Yes, this might be a possibility I think. I wasn't entirely sure about it since that would mean Filestore would be treated differently from all the other resources. Let's try it out.

r4f4:
> > > There's a problem with destroying the clusters when therea are missing premissions for Filestore. There is no timeout it seems so the uninstallation would keep trying to remove the Filestore instances indefinitely. Since the Filestore operator is an optional component it is more likely for this situation to happen and the UX is pretty bad in such a case.
> > 
> > 
> > Can you check the error returned and fail the uninstall in that case? This would give users the chance to add the permission and retry the uninstall.
> 
> Yes, this might be a possibility I think. I wasn't entirely sure about it since that would mean Filestore would be treated differently from all the other resources. Let's try it out.

Even if we don't fail, if we can at least display a message to the user, then it's up to them to stop the uninstall process, add the required perm and retry.

@patrickdillon WDYT regarding returning a failure when a perm issue is found for Filestore during cluster destroy?

barbacbd:
> > > > There's a problem with destroying the clusters when therea are missing premissions for Filestore. There is no timeout it seems so the uninstallation would keep trying to remove the Filestore instances indefinitely. Since the Filestore operator is an optional component it is more likely for this situation to happen and the UX is pretty bad in such a case.
> > > 
> > > 
> > > Can you check the error returned and fail the uninstall in that case? This would give users the chance to add the permission and retry the uninstall.
> > 
> > 
> > Yes, this might be a possibility I think. I wasn't entirely sure about it since that would mean Filestore would be treated differently from all the other resources. Let's try it out.
> 
> Even if we don't fail, if we can at least display a message to the user, then it's up to them to stop the uninstall process, add the required perm and retry.
> 
> @patrickdillon WDYT regarding returning a failure when a perm issue is found for Filestore during cluster destroy?

@tsmetana adding the filestore is a day-2 operation correct? If this is true, I don't think the installer should be the one to delete these resources. 

@r4f4 If this is true and the installer does not create these resources (but we still decide to delete them), maybe we should just warn and move on as long as the file stores do not have a link to another resource that stops the destroy process.

tsmetana:
> @tsmetana adding the filestore is a day-2 operation correct? If this is true, I don't think the installer should be the one to delete these resources.

Yes, CSI Filestore is a day-2 operator and installer does not create the resources, the installed cluster may. The idea behind this patch was to be consistent with AWS which does delete every resource with the owner tag, including those not created by installer (like EFS volumes which is an AWS equivalent of Filestore).
 
> @r4f4 If this is true and the installer does not create these resources (but we still decide to delete them), maybe we should just warn and move on as long as the file stores do not have a link to another resource that stops the destroy process.

The latest commit should just give up and warn in case the error returned is 403, telling the user some cluster resources would need to be cleaned manually. This sounded like a reasonable workaround.

patrickdillon (commented):
I agree with the conclusions in the PR:

- installer should destroy anything day-2 generated from a cluster; if we can identify it comes from openshift and belongs to the cluster we destroy it
- destroy should not stop based on a perms error

pkg/destroy/gcp/filestore.go
```diff
@@ -0,0 +1,96 @@
...
+func (o *ClusterUninstaller) deleteFilestore(ctx context.Context, item cloudResource) error {
+	o.Logger.Debugf("Deleting filestore %s", item.name)
+
+	ctx, cancel := context.WithTimeout(ctx, defaultTimeout)
+	defer cancel()
+	instSvc := file.NewProjectsLocationsInstancesService(o.fileSvc)
+	_, err := instSvc.Delete(item.name).Context(ctx).Do()
+	if err != nil && isForbidden(err) {
+		o.deletePendingItems(item.typeName, []cloudResource{item})
+		return fmt.Errorf("Insufficient permissions to delete filestore %s: giving up", item.name)
```
patrickdillon, line 64: the linter is complaining about the caps. If we're not failing on missing perms, should this be an error, or should this be a logrus Warning? (in which case, with logrus warning it should stay capitalized).

pkg/destroy/gcp/filestore.go
```diff
@@ -0,0 +1,96 @@
...
+	nextPageToken := "pageToken"
+	for nextPageToken != "" {
+		instances, err := lCall.Do()
+		if err != nil {
+			return nil, fmt.Errorf("error retrieving filestore instances: %w", err)
+		}
+
+		for _, activeInstance := range instances.Instances {
+			if len(activeInstance.FileShares) == 0 {
+				// skip multi-share instances
```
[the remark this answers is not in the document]
tsmetana: OpenShift doesn't support creation of multishare Filestore instances (yet). They allow for allocation of multiple volumes per one instance and the idea behind this bit of code was that if we encounter such an instance, we should simply ignore it since it's most likely being used by multiple clusters and it's safer to leak the resource than to lose data of the other Filestore users.

Documentation for the reference: https://cloud.google.com/filestore/docs/multishares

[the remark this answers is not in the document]
tsmetana: So I discussed the above with @jsafrane and the conclusion was that the reasoning I provided was not very clever, so I'll remove that check. It is a no-op, but in the case we add support for multi-shares, the "owned" label should mean the cluster owns the resource, so it should get removed along with the cluster.

pkg/destroy/gcp/filestore.go
tsmetana: I'm unsure myself about this part. The uninstallation can't fail as such, so we only need some way to inform the user the resource deletion failed. Perhaps logging a warning and returning an error would work.

patrickdillon: as the error doesn't stop the destroyer, this is fine. just drop the caps to appease the linter

```suggestion
		return fmt.Errorf("insufficient permissions to delete filestore %s: giving up", item.name)
```

There was one other failure in go-lint that needed to be resolved

tsmetana: I have added a log message (so the user is warned when the deletion fails because of missing permissions) and changed the character case for the error.

The other linter failure was caused by the new `isForbidden` function that now takes a different error type so copying the `isNoOp` was apparently incorrect.

r4f4:
/override ci/prow/okd-images
Known issue

tsmetana:
Thanks for the reviews and the help with this PR. After some discussion with the Storage team I need to keep this on hold: Filestore would behave differently from everything else, which is rather suboptimal, and we'd like to have it also properly documented and tested but there's no time for it in the upcoming release. :(

tsmetana:
/hold cancel

Let's try again in the next relase.

p0lyn0mial:
FYI: It looks like this PR broke [a few CI jobs](https://search.dptools.openshift.org/?search=ipi-deprovision-deprovision+container+test&maxAge=48h&context=1&type=bug%2Bissue%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job). 

These jobs are currently failing on `Filestores: error retrieving filestore instances: googleapi: Error 403: Cloud Filestore API has not been used in project`.

Are we okay with reverting this PR to unblock failing CI jobs?

tsmetana:
> Are we okay with reverting this PR to unblock failing CI jobs?

Yeah. I think it's the best solution. I will take a look a the failed jobs and see if I can fix the PR to make them happy.

p0lyn0mial:
> Yeah. I think it's the best solution. I will take a look a the failed jobs and see if I can fix the PR to make them happy.

@tsmetana thanks, are you also going to prepare a revert PR?

PR #7255, opened 2023-06-16
Use 600 for kubeconfig permissions in bootkube template

rhmdnd:


rhmdnd:
Note - this is the wrong place to update this since this is specific to the bootstrap node and not the actual cluster that gets turned over to the user.

sadasu:
/close

Can be re-opened if this change is needed in the future.

PR #7257, opened 2023-06-16
SPLAT-1094: warn users about deprecation of Alibaba Cloud

fabianofranz:
The Alibaba Cloud provider is soon going to be removed. This PR introduces a warning to users about the deprecation and a request for confirmation before to proceed with an installation in that provider. Doesn't affect the UX of other providers.

alibaba-deprecation-msg

### DIFF
--- a/pkg/asset/installconfig/alibabacloud/alibabacloud.go
+++ b/pkg/asset/installconfig/alibabacloud/alibabacloud.go
@@ -63,6 +63,11 @@ func Platform() (*alibabacloud.Platform, error) {
 		return nil, err
 	}
 
+	err = bypassDeprecation()
+	if err != nil {
+		return nil, err
+	}
+
 	region, err := selectRegion(client)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


fabianofranz:
@patrickdillon ptal, tks!

pkg/asset/installconfig/alibabacloud/alibabacloud.go
```diff
@@ -136,3 +141,25 @@ func selectRegion(client *Client) (string, error) {
 	}
 	return selectedRegion, nil
 }
+
+func bypassDeprecation() error {
+	confirmationMsg := "DEPRECATED. Alibaba Cloud is deprecated and will be removed in " +
+		"a future version of the installer. Please reach out to your support " +
```
julienlim, line 147: Can we update message to the following:

DEPRECATED. Alibaba Cloud is deprecated and will be removed in a future OpenShift version. Please reach out to your Red Hat Support or Technical Account Manager for more information.

fabianofranz: Done.

patrickdillon:
This log message will only appear if users are using the survey. Should we display a warning if users pass an install config?

fabianofranz:
@patrickdillon good catch. I updated it with a smaller version of the message in the survey, and the full version as a log warning in the place you suggested. Thank you!

r4f4:
Here is the output if you answer `N`:
```bash
]$: ./bin/openshift-install create install-config --dir /tmp/alino/
? SSH Public Key ~/.ssh/id_rsa.pub
? Platform alibabacloud
? DEPRECATED. Alibaba Cloud is deprecated and will be removed in a future OpenShift version. Would you still like to continue? No
FATAL failed to fetch Install Config: failed to fetch dependency of "Install Config": failed to fetch dependency of "Base Domain": failed to generate asset "Platform": deprecated platform
```

/lgtm

PR #7259, opened 2023-06-20
OpenStack: enable IPv6 primary dual-stack cluster

MaysaMacedo:
This commit removes the enforcement of the ordering IPv4 to api and ingress VIPs and adds IPv6 security group rules to allow ingress and egress traffic over IPv6.

### DIFF
--- a/data/data/openstack/masters/sg-master.tf
+++ b/data/data/openstack/masters/sg-master.tf
@@ -16,6 +16,18 @@ resource "openstack_networking_secgroup_rule_v2" "master_mcs" {
   description       = local.description
 }
 
+resource "openstack_networking_secgroup_rule_v2" "master_mcs_v6" {
+  count             = length(var.machine_v6_cidrs)
+  direction         = "ingress"
+  ethertype         = "IPv6"
+  protocol          = "tcp"
+  port_range_min    = 22623
+  port_range_max    = 22623
+  remote_ip_prefix  = element(var.machine_v6_cidrs, count.index)
+  security_group_id = openstack_networking_secgroup_v2.master.id
+  description       = local.description
+}
+
 # TODO(mandre) Explicitely enable egress
 
 resource "openstack_networking_secgroup_rule_v2" "master_ingress_icmp" {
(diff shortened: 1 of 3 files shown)


MaysaMacedo:
Nodes and Pods were listing IPv6 address first in the list of addresses.

data/data/openstack/masters/sg-master.tf
```diff
@@ -30,6 +42,19 @@ resource "openstack_networking_secgroup_rule_v2" "master_ingress_icmp" {
   description       = local.description
 }
 
+resource "openstack_networking_secgroup_rule_v2" "master_ingress_icmp_v6" {
+  count          = length(var.machine_v6_cidrs)
+  direction      = "ingress"
+  ethertype      = "IPv6"
+  protocol       = "icmp"
```
EmilienM, line 49: I'm not sure it's `icmp` that we want:
https://opendev.org/openstack/neutron-lib/src/branch/master/neutron_lib/constants.py#L168

EmilienM: https://docs.openstack.org/api-ref/network/v2/?expanded=create-security-group-rule-detail#create-security-group-rule

MaysaMacedo: done

mandre (commented):
Looks good. I was surprised at first that we didn't have V6 equivalent to `master_ingress_ike_nat_t` and `worker_ingress_ike_nat_t` rules, but then I remembered that NAT does not really make sense with IPv6.

/lgtm

data/data/openstack/masters/sg-worker.tf
```diff
@@ -206,6 +337,18 @@ resource "openstack_networking_secgroup_rule_v2" "worker_ingress_services_udp" {
...
+  count             = length(var.machine_v6_cidrs)
+  direction         = "ingress"
+  ethertype         = "IPv6"
+  protocol          = "udp"
+  port_range_min    = 30000
+  port_range_max    = 32767
+  remote_ip_prefix  = element(var.machine_v6_cidrs, count.index)
+  security_group_id = openstack_networking_secgroup_v2.worker.id
+  description       = local.description
+}
```
dulek, line 350: Same here.

data/data/openstack/masters/sg-worker.tf
```diff
@@ -194,6 +313,18 @@ resource "openstack_networking_secgroup_rule_v2" "worker_ingress_services_tcp" {
...
+  count             = length(var.machine_v6_cidrs)
+  direction         = "ingress"
+  ethertype         = "IPv6"
+  protocol          = "tcp"
+  port_range_min    = 30000
+  port_range_max    = 32767
+  remote_ip_prefix  = element(var.machine_v6_cidrs, count.index)
+  security_group_id = openstack_networking_secgroup_v2.worker.id
+  description       = local.description
+}
```
dulek, line 326: This feels like something that should have been here since dual stack support, i.e. IPv6 LoadBalancer Services need this.

dulek (approved):
I'm late to the party, but here's a comment about something that stands out to me.

data/data/openstack/masters/sg-worker.tf
MaysaMacedo: I thought about this, but when I tried connecting to a IPv6 LB without it in tech preview, it worked fine.

PR #7267, opened 2023-06-21
OCPBUGS-15238: GCP: ic: improve project validation

r4f4:
If we want to know if a project is valid, we can use the API to `Get` it directly, instead of `List`ing all projects and then filtering the one we want. Listing can be problematic for accounts with many associated projects.

### DIFF
--- a/pkg/asset/installconfig/gcp/client.go
+++ b/pkg/asset/installconfig/gcp/client.go
@@ -39,6 +39,7 @@ type API interface {
 	GetEnabledServices(ctx context.Context, project string) ([]string, error)
 	GetCredentials() *googleoauth.Credentials
 	GetProjectPermissions(ctx context.Context, project string, permissions []string) (sets.Set[string], error)
+	GetProjectByID(ctx context.Context, project string) (*cloudresourcemanager.Project, error)
 	ValidateServiceAccountHasPermissions(ctx context.Context, project string, permissions []string) (bool, error)
 }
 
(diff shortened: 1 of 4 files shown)


rna-afk:
LGTM but wouldn't this problem show up if we ever use `GetProjects` ? Looking at [survey](https://github.com/openshift/installer/blob/099a0d824f27f7a5b4db3714b743c50a158161c7/pkg/asset/installconfig/gcp/gcp.go#L49) we might end up with a lot of projects getting displayed?

r4f4:
> LGTM but wouldn't this problem show up if we ever use `GetProjects` ? Looking at [survey](https://github.com/openshift/installer/blob/099a0d824f27f7a5b4db3714b743c50a158161c7/pkg/asset/installconfig/gcp/gcp.go#L49) we might end up with a lot of projects getting displayed?

Yes. I don't know how many projects people have in their accounts, but the OP of the bug report had
```bash
> gcloud projects list | wc -l
  152793
```

So the Survey would show 152793 projects. I would say that is less of a problem because you can always generate the install-config yourself, bypassing the Survey. But I don't mind changing it if you have any suggestions of what we should do.

rna-afk:
Don't think there's a solution here that's simple. Guess the users have to bypass the survey.

patrickdillon:
/approve

It's good to think about the survey but I agree we shouldn't bother trying to fix that at this stage. Bumping the context timeout on the survey to be longer than 1 minute would perhaps allow the api call to complete, but that is a bad ux. nobody should wait a minute to list the projects.

PR #7269, opened 2023-06-21
OpenStack: Add steps to enable dual-stack clusters

MaysaMacedo:
The `MachinesSubnet` field has been reshaped as `controlPlanePort`,
this commit updates the docs to ensure `controlPlanePort` is used.
Also, this commit adds dual-stack documentation.


dulek:
/retest-required

Doesn't seem like these jobs would block merging this?

PR #7270, opened 2023-06-22
Update hack/go-test.sh to golang 1.20

rwsu:
It was previously set to 1.18 and needed to be bumped to match move to golang 1.20.


r4f4 (commented):
/lgtm

PR #7273, opened 2023-06-22
OCPBUGS-15041: Pass releaseImageMirror to assisted if mirror defined

bfournie:
In https://github.com/openshift/assisted-service/pull/5102 a new assisted-service validation was added to check DNS for the releaseImage. This fails in a disconnected environment when the releaseImage is the source.  If a mirror is defined, pass in the releaseImage set to the mirror.

Note that an assisted-service bug for the failure in a disconnected environment was created - https://issues.redhat.com/browse/MGMT-15056. If a fix for that merges we can revert this change.


bmanzari (approved):

bfournie:
The change to the validation in assisted-service resolves this issue so no need for this PR.
/close

PR #7274, opened 2023-06-22
OCPBUGS-2324: terraform: aws: bump version to 5.4.0

r4f4:
It contains bug fixes we need for tagging in ISO partitions.


barbacbd (commented):
talking with @r4f4 its a bit odd that the go version doesn't seem to match in the mod files. This may be a concern but it doesn't seem to cause errors. 

/lgtm

PR #7276, opened 2023-06-22
AGENT-627: Decompress kernel on ARM

pawanpinjarkar:
Decompress kernel for ARM architecture in case of PXE boot.

### DIFF
--- a/pkg/asset/agent/image/agentpxefiles.go
+++ b/pkg/asset/agent/image/agentpxefiles.go
@@ -1,6 +1,7 @@
 package image
 
 import (
+	"compress/gzip"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


cmd/openshift-install/testdata/agent/pxe/configurations/sno.txt
```diff
@@ -2,25 +2,27 @@
 
 exec openshift-install agent create pxe-files --dir $WORK
 
-stderr 'Created iPXE script agent.x86_64.ipxe'
+stderr 'Created iPXE script agent.aarch64.ipxe'
```
andfasano, line 5: What about having also another integration test specific form arm, ie `cmd/openshift-install/testdata/agent/pxe/configurations/sno-arm.txt`?

pkg/asset/agent/image/agentpxefiles.go
```diff
@@ -120,9 +123,22 @@ func (a *AgentPXEFiles) PersistToFile(directory string) error {
 		return err
 	}
 	defer kernelReader.Close()
-	err = a.copy(agentVmlinuzFile, kernelReader)
-	if err != nil {
-		return err
+
+	if a.cpuArch == arch.RpmArch(types.ArchitectureARM64) {
```
andfasano, line 127: It could be useful to have a switch capturing the known archs, and failing for all the others

cmd/openshift-install/testdata/agent/pxe/configurations/sno.txt
pawanpinjarkar: sure

PR #7279, opened 2023-06-25
CFE-857: Apply user defined tags on created gcp resources

bharath-b-rh:
PR is for adding user-defined tags to the GCP resources created by installer during cluster creation.
- Terraform APIs require the TagValue resource `Name` and it needs to be derived using the parameters `OrganizationID, UserTag.Key, UserTag.Value` defined by user in install-config.
- Logic for deriving the TagValues Name is implemented using golang, which happens during install-config validation or while generating the terraform variables.
- Terraform code has the logic for adding tags to the resources during (for compute resources) and after it's creation(for storage buckets).


patrickdillon:
You will want to separate the go vendoring into a separate commit (like this https://github.com/openshift/installer/blob/master/docs/dev/dependencies.md#go) to make it easier to review

bharath-b-rh:
> You will want to separate the go vendoring into a separate commit (like this https://github.com/openshift/installer/blob/master/docs/dev/dependencies.md#go) to make it easier to review

Thank you for reviewing @patrickdillon!

I have split the go mod changes into two for go.mod update and for adding new vendor code.

data/data/gcp/bootstrap/main.tf
```diff
@@ -61,6 +68,16 @@ data "ignition_config" "redirect" {
   }
 }
 
+resource "google_tags_location_tag_binding" "user_tag_binding_bucket" {
+  for_each = var.gcp_extra_tags
+
+  parent = format("//storage.googleapis.com/projects/_/buckets/%s",
+    google_storage_bucket.ignition.name,
+  )
```
patrickdillon, line 76: It seems strange that there is an underscore here in place for the project. Can you provide any more context for that?

data/data/gcp/cluster/main.tf
```diff
@@ -107,3 +108,20 @@ resource "google_compute_image" "cluster" {
 
   licenses = var.gcp_image_licenses
 }
+
+data "google_compute_image" "cluster_images" {
+  count = var.gcp_preexisting_image ? 0 : 1
+  name  = "${var.cluster_id}-rhcos-image"
+
+  depends_on = [google_compute_image.cluster]
+}
```
patrickdillon, line 117: Why create a data source here? I think you can just read image_id from the cluster image

data/data/gcp/cluster/main.tf
```diff
@@ -107,3 +108,20 @@ resource "google_compute_image" "cluster" {
...
+
+resource "google_tags_tag_binding" "image_tag_binding" {
+  for_each = { for key, val in var.gcp_extra_tags : key => val if ! var.gcp_preexisting_image }
+
+  parent = format("//compute.googleapis.com/projects/%s/global/images/%s",
+    var.gcp_project_id,
+    data.google_compute_image.cluster_images[0].image_id,
```
patrickdillon, line 124: Instead of using the data source here, you can just read from the resource, it would be something like:

```
google_compute_image.cluster[0].image_id
```

pkg/tfvars/gcp/gcp.go
```diff
@@ -137,3 +169,121 @@ func generateDiskEncryptionKeyLink(keyRef *machineapi.GCPEncryptionKeyReference,
 
 	return fmt.Sprintf(kmsKeyNameFmt, projectID, keyRef.KMSKey.Location, keyRef.KMSKey.KeyRing, keyRef.KMSKey.Name)
 }
+
+// getTagValuesNames returns list of tag values Names as required by terraform APIs
+// fetched using the user provided tag ParentID, Key shortname and Value shortname.
+func getTagValuesNames(cred, projID string, tags []gcp.UserTag) (map[string]string, error) {
```
patrickdillon, line 175: I don't think this logic belongs in tfvars. tfvars is primarily concerned with converting inputs into the proper terraform variable formats. One typical pattern, which applies here, is to read from the machine manifests.

patrickdillon (commented):
I did a first pass and left some feedback, but I still need to review the tag client logic (which you currently have in tfvars and as I noted should be moved elsewhere. I will try to review this soon, as the purpose of the logic itself is not immediately clear to me, but the purpose should determine where it would go). 

I do notice that it does not appear that the tags are being written to the machine manifests. I suspect the tags should be attached to the machines (I believe you are doing this in the terraform configs for the terraform-created control plane machines). You are attaching the labels to the machines in your labels pr here: https://github.com/openshift/installer/pull/7153/files#diff-2c8a4950a72f42e38fa822fcf3661543c1c584559cdb852a4e59f42855c61493R205

bharath-b-rh:
> I do notice that it does not appear that the tags are being written to the machine manifests. I suspect the tags should be attached to the machines (I believe you are doing this in the terraform configs for the terraform-created control plane machines). You are attaching the labels to the machines in your labels pr here: https://github.com/openshift/installer/pull/7153/files#diff-2c8a4950a72f42e38fa822fcf3661543c1c584559cdb852a4e59f42855c61493R205

Yeah, it should have been done the same way for Tags too like in labels. But MachineProviderSpec currently doesn't support tags and [new features are not being accepted for Machine API](https://github.com/openshift/api/pull/1506#issuecomment-1611754604), unless it's already supported in Cluster API(upstream). Hence I had to make use of [TFVarsSources](https://github.com/openshift/installer/pull/7279/files#diff-fe206eadc78bb637c8b5ebbf96a7fe49eb5eab26b5ce28153f793d8d4c1f72a7R77) model for tags.

data/data/gcp/bootstrap/main.tf
bharath-b-rh: GCP Buckets aren't project scoped but is global resource, hence I think API is not expecting the project ID. More details on the bucket scope can be found [here](https://cloud.google.com/storage/docs/json_api/v1/buckets) and details on API for adding tags to buckets can be found [here](https://cloud.google.com/storage/docs/tags-and-labels#tags)

data/data/gcp/cluster/main.tf
bharath-b-rh: Yeah, but `google_compute_image` API is not exporting `image_id` attribute. Hence had to make use of data source to fetch the `image_id` value. [Link](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_image.html#attributes-reference) has details on the list of attributes that are made available by the API.

data/data/gcp/cluster/main.tf
bharath-b-rh: Yeah, but google_compute_image API is not exporting image_id attribute. Hence had to make use of data source to fetch the image_id value. [Link](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_image.html#attributes-reference) has details on the list of attributes that are made available by the API.

pkg/tfvars/gcp/gcp.go
bharath-b-rh: Terraform tag APIs accept tag value in the form `tagValues/<tag_value_id>`, which can be obtained using the user provided input `ParentID`, `Tag_Key_Name`, `Tag_Value_Name` and this is required just for the terraform APIs, golang APIs accept the above form as well as the `<ParentID>/<Tag_Key_Name>/<Tag_Value_Name>`. so I thought to keep it in this source file.

Please suggest.

data/data/gcp/bootstrap/main.tf
```diff
@@ -48,6 +48,13 @@ resource "google_project_iam_member" "bootstrap-storage-admin" {
   member  = "serviceAccount:${google_service_account.bootstrap-node-sa[0].email}"
 }
 
+resource "google_project_iam_member" "bootstrap-tag-user" {
+  count   = var.gcp_create_bootstrap_sa ? 1 : 0
+  project = var.gcp_project_id
+  role    = "roles/resourcemanager.tagUser"
+  member  = "serviceAccount:${google_service_account.bootstrap-node-sa[0].email}"
+}
```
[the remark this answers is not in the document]
bharath-b-rh: Yeah it is not required. But earlier while on working the PR, operation was failing with permission denied error, and that was the reason I added the role to the account. I will update the code.

patrickdillon:
> > I do notice that it does not appear that the tags are being written to the machine manifests. I suspect the tags should be attached to the machines (I believe you are doing this in the terraform configs for the terraform-created control plane machines). You are attaching the labels to the machines in your labels pr here: https://github.com/openshift/installer/pull/7153/files#diff-2c8a4950a72f42e38fa822fcf3661543c1c584559cdb852a4e59f42855c61493R205
> 
> Yeah, it should have been done the same way for Tags too like in labels. But MachineProviderSpec currently doesn't support tags and [new features are not being accepted for Machine API](https://github.com/openshift/api/pull/1506#issuecomment-1611754604), unless it's already supported in Cluster API(upstream). Hence I had to make use of [TFVarsSources](https://github.com/openshift/installer/pull/7279/files#diff-fe206eadc78bb637c8b5ebbf96a7fe49eb5eab26b5ce28153f793d8d4c1f72a7R77) model for tags.

Oh, I see. The `tags` field I see in the machine provider spec is for network tags: https://github.com/openshift/api/pull/1506/files#diff-9bd66f01f63b7ce0455fe993aa03dfe93964ac62b31278a6dbf3db2fcb7f792eR105

got it!

pkg/tfvars/gcp/gcp.go
```diff
@@ -137,3 +169,121 @@ func generateDiskEncryptionKeyLink(keyRef *machineapi.GCPEncryptionKeyReference,
...
+
+// getTagValueClient returns the client to be used for fetching the tag value details.
+func getTagValueClient(cred string) (*rscmgr.TagValuesClient, error) {
+	opts := []option.ClientOption{
+		option.WithCredentialsJSON([]byte(cred)),
+		option.WithEndpoint(tagRequestEndPoint),
+	}
+
+	return rscmgr.NewTagValuesRESTClient(context.TODO(), opts...)
+}
```
patrickdillon, line 240: It would be good to follow the patterns we have for GCP clients: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/gcp/client.go

pkg/tfvars/gcp/gcp.go
```diff
@@ -137,3 +169,121 @@ func generateDiskEncryptionKeyLink(keyRef *machineapi.GCPEncryptionKeyReference,
...
+		limiter.AllowN(time.Now(), burst)
+	}
+
+	return limiter
+}
+
+// getTagValueClient returns the client to be used for fetching the tag value details.
+func getTagValueClient(cred string) (*rscmgr.TagValuesClient, error) {
+	opts := []option.ClientOption{
+		option.WithCredentialsJSON([]byte(cred)),
```
patrickdillon, line 235: We support multiple methods of authentication, so we can't assume that users will pass a JSON file. Hopefully following the client pattern in https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/gcp/client.go will naturally resolve this issue.

pkg/types/gcp/validation/platform.go
```diff
@@ -67,6 +67,16 @@ var (
...
 )
 
+const (
+	// maxUserLabelLimit is the maximum userLabels that can be configured as defined in openshift/api.
+	// https://github.com/openshift/api/commit/ae73a19d05c35068af16c9aeff375d0b7c936a8a#diff-07b264a49084976b670fb699badaca1795027d6ea732a99226a5388104f6174fR592-R602
+	maxUserLabelLimit = 32
+
+	// MaxUserTagLimit is the maximum userTags that can be configured as defined in openshift/api.
+	// https://github.com/openshift/api/commit/ae73a19d05c35068af16c9aeff375d0b7c936a8a#diff-07b264a49084976b670fb699badaca1795027d6ea732a99226a5388104f6174fR604-R613
+	MaxUserTagLimit = 50
```
bharath-b-rh, line 77: These constants were first defined in https://github.com/openshift/installer/pull/7126, MaxUserTagLimit is exported here, hence include in this PR as well, will resolve the conflicts, when 7126 is merged.

pkg/tfvars/gcp/gcp.go
bharath-b-rh: suggestion incorporated.

pkg/tfvars/gcp/gcp.go
bharath-b-rh: suggestion incorporated.

pkg/tfvars/gcp/gcp.go
bharath-b-rh: suggestion incorporated.

bharath-b-rh:
> Also, this is unfortunate that VMs will not be able to be tagged unless they are the control plane nodes created by the installer

Yeah, it will be limitation for TP. But I have created an issue upstream CAPI and will be working on adding the feature there and GA should have the feature hopefully.

pkg/asset/installconfig/gcp/resourcetags.go
```diff
@@ -0,0 +1,165 @@
...
+		tagValues[tag.Parent] = tag.Name
+	}
+
+	return tagValues, nil
+}
+
+// findTagConflicts returns error when the user-defined tags already existing at the
+// project in which the resources will be created. Tags defined at project will be
+// inherited by the child resources and adding same tag will result in conflict error.
+func (c *Client) findTagConflicts(project string, tags map[string]string) error {
```
patrickdillon, line 91: Conflicts and checking for the existence of tags should be done in the GCP installconfig validation: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/gcp/validation.go

Ideally this would include unit tests to verify this logic.  

Because this code path is so API intensive, it might be worth adding something like AWS's metadata to hold the tag data: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/aws/metadata.go

bharath-b-rh: > Conflicts and checking for the existence of tags should be done in the GCP installconfig validation: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/gcp/validation.go
>
> Ideally this would include unit tests to verify this logic.

Suggestion incorporated

> Because this code path is so API intensive, it might be worth adding something like AWS's metadata to hold the tag data: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/aws/metadata.go

I think I didn't understand why it would be required also I feel it's not required, since there no unknown/dynamic values to pass around, and it will be just for tags. Please let me know WDYT?

pkg/asset/installconfig/gcp/usertags.go
```diff
@@ -0,0 +1,212 @@
...
+	logrus.Debugf("user defined tags list: %v", userTags)
+
+	projectTags, err := getProjectTags(ctx, mgr, project)
+	if err != nil {
+		return nil, err
+	}
+
+	if dupTags := findDuplicateTags(userTags, projectTags); len(dupTags) != 0 {
+		return nil, fmt.Errorf("found duplicate tags, %v tags already exist on %s project", dupTags, project)
+	}
```
patrickdillon, line 82: This check should go in the installconfig asset validation

pkg/asset/installconfig/gcp/usertags.go
```diff
@@ -0,0 +1,212 @@
...
+	for _, tag := range userTags {
+		name := fmt.Sprintf("%s/%s/%s", tag.ParentID, tag.Key, tag.Value)
+		key, value, err := client.getTagKeyValueName(ctx, name)
+		if err != nil {
+			// check and return all non-existing tags at once
+			// for user to fix in one go.
+			var gErr *apierror.APIError
+			// google API returns StatusForbidden or StatusNotFound when the tag
+			// does not exist, since it could be because of permission issues
+			// or genuinely tag does not exist.
```
patrickdillon, line 101: this check should also go in the install config validation section

patrickdillon (commented):
/approve
/lgtm

The error checking in `ConvertUserTags` should run when validating the install config. We can fix that in a followup PR

patrickdillon:
/lgtm cancel
I see that we are still waiting on other upstream PRs. Let's get the validation fixed in this PR, then

pkg/asset/installconfig/gcp/usertags.go
bharath-b-rh: The error check here is when converting the tags to required form by querying for additional tag details, should we check for tags existence first and later convert the tags(call same API twice for each tag)? or maybe during the first run tags could be cached and same is returned when ConvertUserTags() is called.
Please let me know WDYT?

patrickdillon: > maybe during the first run tags could be cached and same is returned when ConvertUserTags() 

Considering that these API calls are somewhat computationally expensive, I think caching them makes sense. I would suggest making a helper function using a singleton pattern that could be used both by the validation and tfvars. 

We do something kind of like this with subnets in aws metatadata:
https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/aws/metadata.go

pkg/asset/installconfig/gcp/usertags.go
bharath-b-rh: comment incorporated.

pkg/asset/installconfig/gcp/usertags.go
bharath-b-rh: comment incorporated.

pkg/asset/installconfig/gcp/usertags.go
```diff
@@ -0,0 +1,298 @@
...
+	"context"
+	"fmt"
+	"net/http"
+	"sync"
+	"time"
+
+	rscmgr "cloud.google.com/go/resourcemanager/apiv3"
+	rscmgrpb "cloud.google.com/go/resourcemanager/apiv3/resourcemanagerpb"
+	"github.com/googleapis/gax-go/v2/apierror"
+	"github.com/pkg/errors"
```
r4f4, line 13: Let's not introduce this deprecated lib in new code. Use `fmt.Errorf` with the `%w` verb instead.

pkg/asset/installconfig/gcp/usertags.go
```diff
@@ -0,0 +1,298 @@
...
+	logrus.Debugf("user defined tags processed list: %v, took %s to finish", processedTags.tags, time.Since(start))
+	return nil
+}
+
+// getProjectTags returns of the list of tags attached to the passed project
+// resource and the tags inherited by it.
+func getProjectTags(ctx context.Context, mgr TagManager, project string) (effectiveTags, error) {
+	client, err := getTagBindingsClient(ctx, mgr)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create client to fetch effective tags on the %s project", project)
```
r4f4, line 212:
```suggestion
		return nil, fmt.Errorf("failed to create client to fetch effective tags on the %s project: %w", project, err)
```

pkg/asset/cluster/tfvars/tfvars.go
```diff
@@ -525,6 +525,11 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 			return fmt.Errorf("%s: No GCP build found", st.FormatPrefix(archName))
 		}
 
+		tags, err := gcpconfig.GetUserTags(gcpconfig.NewTagManager(client), installConfig.Config)
+		if err != nil {
+			return errors.Wrapf(err, "failed to fetch user-defined tags")
```
r4f4, line 530:
```suggestion
			return fmt.Errorf("failed to fetch user-defined tags: %w", err)
```

pkg/asset/installconfig/gcp/usertags.go
```diff
@@ -0,0 +1,298 @@
...
+// GetUserTags returns the processed list of user provided tags if already available,
+// else validates, stores and returns the processed tags.
+func GetUserTags(mgr TagManager, ic *types.InstallConfig) (map[string]string, error) {
+	return getProcessedTags(mgr, ic)
+}
+
+// getProcessedTags returns the list user-provided validated tags.
+func getProcessedTags(mgr TagManager, ic *types.InstallConfig) (map[string]string, error) {
+	if !processedTags.getState() {
+		if err := validateUserTags(mgr, ic); err != nil {
```
r4f4, line 134: It's a bit weird that the definition for `validateUserTags` lives in the `validation.go` file and it basically just calls `validateAndStoreUserTags` which is defined in this file. Also, I don't see why this function has to receive the `InstallConfig` object if it only needs the project name and the userTags to do its job.

pkg/asset/installconfig/gcp/usertags.go
```diff
@@ -0,0 +1,298 @@
...
+// updateState is for updating the tags processed status.
+func (p *processedUserTags) updateState() {
+	p.Lock()
+	defer p.Unlock()
+
+	p.processed = true
+}
+
+// getState is for getting the tags processed status.
+func (p *processedUserTags) getState() bool {
```
r4f4, line 101:
```suggestion
func (p *processedUserTags) IsProcessed() bool {
```
Maybe this is a better name? When using this, it's not clear what it's checking `if !processedTags.getState()`. I think it would be more clear with `if !processedTags.IsProcessed()` or something more descriptive.

pkg/asset/installconfig/gcp/usertags.go
```diff
@@ -0,0 +1,298 @@
...
+// addTag is for adding a tag-set to processedUserTags.
+func (p *processedUserTags) addTag(k, v string) {
+	p.Lock()
+	defer p.Unlock()
+
+	p.tags[k] = v
+}
+
+// updateState is for updating the tags processed status.
+func (p *processedUserTags) updateState() {
```
r4f4, line 93:
```suggestion
func (p *processedUserTags) markProcessed() {
```
Or something like that.

pkg/asset/installconfig/gcp/usertags.go
```diff
@@ -0,0 +1,298 @@
...
+		if err != nil {
+			// check and return all non-existing tags at once
+			// for user to fix in one go.
+			var gErr *apierror.APIError
+			// google API returns StatusForbidden or StatusNotFound when the tag
+			// does not exist, since it could be because of permission issues
+			// or genuinely tag does not exist.
+			if errors.As(err, &gErr) && (gErr.HTTPCode() == http.StatusNotFound ||
+				gErr.HTTPCode() == http.StatusForbidden) {
+				logrus.Debugf("does not have permission to access %s tag or does not exist", name)
```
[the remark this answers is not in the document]
bharath-b-rh: Error returned by the GCP APIs is also similar. I think it's because RBAC can be defined on each tag key and value, and with a given service account, it won't be possible to say whether the tag exists and is inaccessible or it doesn't exist at all, so thought to keep the same.

```
User [xyz] does not have permission to access tagValues instance [namespaced] (or it may not exist): Permission denied on resource 'openshift/key/value' (or it may not exist).
```

pkg/asset/installconfig/gcp/usertags.go
bharath-b-rh: Yeah, I missed this unnecessary call while incorporating the comments, I will update it.
I thought to pass the `InstallConfig` object reference and use what's required, instead of passing individual parameters. Please let me know, if needs to be changed.

pkg/asset/installconfig/gcp/mock/usertags_mock.go
```diff
@@ -0,0 +1,83 @@
+// Code generated by MockGen. DO NOT EDIT.
```
bharath-b-rh, line 1: This code will be merged with mock code in `pkg/asset/installconfig/gcp/mock/gcpclient_generated.go` after updating `google.golang.org/api/cloudresourcemanager` from `v1` to `v3` in the follow up PR.

pkg/asset/installconfig/gcp/usertags.go
```diff
@@ -0,0 +1,247 @@
...
+	for _, tag := range userTags {
+		tagNamespacedName := fmt.Sprintf("%s/%s/%s", tag.ParentID, tag.Key, tag.Value)
+		if parentTags.Has(tagNamespacedName) {
+			dupTags = append(dupTags, tagNamespacedName)
+		}
+	}
+	return dupTags
+}
+
+// getCloudResourceServiceForTags returns the client required for querying resource manager resources.
```
bharath-b-rh, line 192: [Code](https://github.com/openshift/installer/pull/7279/files#diff-c4351fb8fcc08cb33821af969413576f521211960eeeb7c0c299ffa2e66cefc1R192-R247) will be merged with client code in `pkg/asset/installconfig/gcp/client.go` after updating `google.golang.org/api/cloudresourcemanager` from `v1` to `v3` in the follow up PR.

pkg/asset/installconfig/gcp/usertags.go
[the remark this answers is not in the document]
bharath-b-rh: Code updated, new API of `cloudresourcemanager` package returns `StatusForbidden` for both non-existent and inaccessible tags.

r4f4:
LGTM in terms of the code but I haven't tested it myself. I'll leave it to @patrickdillon to give this a last look and tag it.

patrickdillon:
Generally this looks good to me (but needs a rebase). We are moving away from Terraform and implementing a CAPI-based install pattern (see [the enhancement](https://github.com/openshift/enhancements/pull/1528) for details). This code is already done so I am fine with merging it, but depending on the progress we make, it is possible we will not be using Terraform for GCP in 4.16. In order to support this feature going forward, it will be necessary for us to implement user tag support in the [cluster-api-provider-gcp](https://github.com/kubernetes-sigs/cluster-api-provider-gcp) as it currently does not have support.

bharath-b-rh:
> Generally this looks good to me (but needs a rebase). We are moving away from Terraform and implementing a CAPI-based install pattern (see [the enhancement](https://github.com/openshift/enhancements/pull/1528) for details). This code is already done so I am fine with merging it, but depending on the progress we make, it is possible we will not be using Terraform for GCP in 4.16. In order to support this feature going forward, it will be necessary for us to implement user tag support in the [cluster-api-provider-gcp](https://github.com/kubernetes-sigs/cluster-api-provider-gcp) as it currently does not have support.

Thank you! I have rebased the PR.
Tags support was added in capg as part of this [issue](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/992), where GCP Cluster and Machine Spec API were updated to support tags. And presently, tags are added to Compute Instance created by the provider, need to check for any other resources managed by it.

PR #7280, opened 2023-06-26
OSASINFRA-3155 - OpenStack: Create ControlPlaneMachineSet CRDs

pierreprinetti:


EmilienM:
really good work IMO
/lgtm

pkg/asset/machines/openstack/machines.go
```diff
@@ -105,7 +96,57 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 		machines = append(machines, machine)
 	}
 
-	return machines, nil
+	replicas := int32(total)
+
```
EmilienM, line 100: we need to remove the FD fields before, see https://github.com/openshift/installer/blob/master/pkg/asset/machines/aws/machines.go#L125-L126

EmilienM:
@pierreprinetti somehow something is wrong with the workers, I ran the tests and the generated manifest for the worker machineset doesn't have any AZ or rootVolume.

pierreprinetti:
> @pierreprinetti somehow something is wrong with the workers, I ran the tests and the generated manifest for the worker machineset doesn't have any AZ or rootVolume.

Yeah, the magic of pointers. Fortunately, the manifest tests catched that. Should be sorted out now

EmilienM:
/lgtm

Nice work @pierreprinetti

EmilienM:
/hold https://github.com/openshift/cluster-control-plane-machine-set-operator/pull/195 needs to land first.

At this point we can probably ping someone from the installer to get /approve.

pierreprinetti:
The CCPMSO patch merged. This is good to go!

/hold cancel
/assign sadasu patrickdillon

pkg/asset/machines/openstack/machines.go
```diff
@@ -105,7 +94,71 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
...
+					},
+				},
+			},
+		},
+	}
+
+	if CPMSFailureDomains := pruneFailureDomains(failureDomains); CPMSFailureDomains != nil {
+		controlPlaneMachineSet.Spec.Template.OpenShiftMachineV1Beta1Machine.FailureDomains = machinev1.FailureDomains{
+			Platform:  v1.OpenStackPlatformType,
+			OpenStack: pruneFailureDomains(failureDomains),
```
r4f4, line 158:
```suggestion
			OpenStack: CPMSFailureDomains,
```

docs/user/openstack/control-plane-machine-set.md
```diff
@@ -0,0 +1,324 @@
...
+            tags:
+            - openshiftClusterID=ocp1-2g2xs
+            trunk: true
+            userDataSecret:
+              name: master-user-data
+```
+
+In this case, the `spec.template.machines_v1beta1_machine_openshift_io.failureDomains` stanza does not contain the `openstack` property.
+
+Leaving `failureDomains` without an platform value means: do not substitute the values in the `providerSpec`. In this case, the `providerSpec` does not contain the `availabilityZone` property, because it's implicitly set to the empty string.
```
r4f4, line 80:
```suggestion
Leaving `failureDomains` without a platform value means: do not substitute the values in the `providerSpec`. In this case, the `providerSpec` does not contain the `availabilityZone` property, because it's implicitly set to the empty string.
```

pkg/asset/machines/openstack/machines.go
```diff
@@ -78,9 +68,8 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 			failureDomain,
 		)
 		if err != nil {
-			return nil, err
+			return nil, nil, fmt.Errorf("failed to generate the Machine providerSpec for replica %d: %w", idx, err)
```
r4f4, line 71:
```suggestion
			return nil, nil, errors.Wrapf(err, "failed to generate the Machine providerSpec for replica %d", idx)
```

pkg/asset/machines/openstack/machines.go
```diff
@@ -105,7 +94,71 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
...
+		platform,
+		mpool,
+		osImage,
+		role,
+		userDataSecret,
+		trunkSupport,
+		machinev1.OpenStackFailureDomain{RootVolume: &machinev1.RootVolume{}},
+	)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to generate the CPMS providerSpec: %w", err)
```
r4f4, line 108:
```suggestion
		return nil, nil, errors.Wrap(err, "failed to generate the CPMS providerSpec")
```

r4f4:
/approve

/hold
In case you decide to address the nit comments before merging.

pkg/asset/machines/openstack/machines.go
pierreprinetti: Dave Cheney's `pkg/errors` [has been archived](https://github.com/pkg/errors) and all its functionality has been incorporated into the standard library. I know the dependency is still available in the Installer's `go.mod`, however I prefer not to use it in new code.

pkg/asset/machines/openstack/machines.go
```diff
@@ -198,6 +251,97 @@ func generateProvider(clusterID string, platform *openstack.Platform, mpool *ope
...
+			}
+
+			if volumes := len(mpool.RootVolume.Zones); volumes > numberOfFailureDomains {
+				numberOfFailureDomains = volumes
+			}
+		}
+	}
+
+	// No failure domain is exactly like one failure domain with the default values.
+	if numberOfFailureDomains < 1 {
```
barbacbd, line 309:
```suggestion
	if numberOfFailureDomains == 0 {
```

pkg/types/openstack/defaults/machinepool.go
```diff
@@ -1,6 +1,11 @@
 package defaults
 
-// DefaultRootVolumeAZ returns the default value for Root Volume availability zones.
-func DefaultRootVolumeAZ() []string {
-	return []string{""}
+// DefaultRootVolumeAZ returns the default value for Root Volume availability zone.
+func DefaultRootVolumeAZ() string {
```
barbacbd, line 4: nit: Does this need to be a function or can it be a constant ?

pkg/types/openstack/defaults/machinepool.go
```diff
@@ -1,6 +1,11 @@
...
+	return ""
+}
+
+// DefaultComputeAZ returns the default value for Compute availability zone.
+func DefaultComputeAZ() string {
```
barbacbd, line 9: Same as above.

pkg/types/openstack/defaults/machinepool.go
pierreprinetti: A slice can't be const, and a variable is at risk of being polluted in the code. I thought that a function is the best way to guarantee a deterministic output, while not being expensive since we're only calling it a couple times

pierreprinetti: It can be a const! Would it be better as a const?

PR #7283, opened 2023-06-27
CORS-2658: azure: multi-zonal NAT gateways support 

r4f4:
Add multi-zone support for Azure NAT gateways. This means that there are
as many NAT gateways and subnets for controlPlane and compute machines
as there are availability zones.

Depends on https://github.com/openshift/installer/pull/6933

### DIFF
--- a/data/data/azure/vnet/common.tf
+++ b/data/data/azure/vnet/common.tf
@@ -6,7 +6,7 @@ data "azurerm_subnet" "preexisting_master_subnet" {
 
   resource_group_name  = var.azure_network_resource_group_name
   virtual_network_name = var.azure_virtual_network
-  name                 = var.azure_control_plane_subnet
+  name                 = var.azure_control_plane_subnet[0]
 }
 
 data "azurerm_subnet" "preexisting_worker_subnet" {
@@ -14,7 +14,7 @@ data "azurerm_subnet" "preexisting_worker_subnet" {
 
   resource_group_name  = var.azure_network_resource_group_name
   virtual_network_name = var.azure_virtual_network
-  name                 = var.azure_compute_subnet
+  name                 = var.azure_compute_subnet[0]
 }
 
 data "azurerm_virtual_network" "preexisting_virtual_network" {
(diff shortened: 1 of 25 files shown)


r4f4:
Update: rebased on master to fix merge conflicts (https://github.com/openshift/installer/pull/6933 has merged).

I will keep at as WIP because I still need to resolve the CPMS situation.

r4f4:
/hold
We still need to fix CPMS for the NAT workflow, but we can start checking the non-nat scenarios for regressions.

r4f4:
`e2e-azure-ovn-shared-vpc` is back to installing:
```bash
time="2023-07-18T18:01:09Z" level=debug msg="Time elapsed per stage:"
time="2023-07-18T18:01:09Z" level=debug msg="              vnet: 11m31s"
time="2023-07-18T18:01:09Z" level=debug msg="         bootstrap: 2m26s"
time="2023-07-18T18:01:09Z" level=debug msg="           cluster: 6m15s"
time="2023-07-18T18:01:09Z" level=debug msg="Bootstrap Complete: 14m37s"
time="2023-07-18T18:01:09Z" level=debug msg="               API: 32s"
time="2023-07-18T18:01:09Z" level=debug msg=" Bootstrap Destroy: 1m39s"
time="2023-07-18T18:01:09Z" level=debug msg=" Cluster Operators: 26m4s"
time="2023-07-18T18:01:09Z" level=info msg="Time elapsed: 1h2m43s"
```

/test e2e-azurestack

data/data/azure/vnet/common.tf
```diff
@@ -6,15 +6,15 @@ data "azurerm_subnet" "preexisting_master_subnet" {
 
   resource_group_name  = var.azure_network_resource_group_name
   virtual_network_name = var.azure_virtual_network
-  name                 = var.azure_control_plane_subnet
+  name                 = var.azure_control_plane_subnet[0]
```
jhixson74, line 9: Shouldn't this use an index instead of 0? How do you setup more than a single subnet in install-config and pass it into terraform ?

data/data/azure/vnet/common.tf
```diff
@@ -6,15 +6,15 @@ data "azurerm_subnet" "preexisting_master_subnet" {
...
 }
 
 data "azurerm_subnet" "preexisting_worker_subnet" {
   count = var.azure_preexisting_network ? 1 : 0
 
   resource_group_name  = var.azure_network_resource_group_name
   virtual_network_name = var.azure_virtual_network
-  name                 = var.azure_compute_subnet
+  name                 = var.azure_compute_subnet[0]
```
jhixson74, line 17: Same question here as described above for a pre-existing master subnet.

jhixson74 (changes requested):
This looks pretty good.  I just have the couple of questions I listed. Another thing I am curious about is in: 
https://github.com/openshift/installer/blob/9dfc08bc02dcce80af0d2f3dc63c6d0216247591/pkg/asset/machines/azure/machines.go#L264
If we can set that to the internal LB name indexed. Maybe then only the cloud config needs to learn about it. 
It would also be nice to have a DIA diagram documenting the architecture here.

data/data/azure/vnet/common.tf
r4f4: When bringing their own network, users were expected to provide only one subnet for control plane and one for compute. So I'm using `[0]` to keep the previous logic explicit here. I can change it to an index but the effect is the same given the `count` is either `0` or `1`.

r4f4:
> This looks pretty good. I just have the couple of questions I listed. Another thing I am curious about is in:
> 
> https://github.com/openshift/installer/blob/9dfc08bc02dcce80af0d2f3dc63c6d0216247591/pkg/asset/machines/azure/machines.go#L264
> 
> If we can set that to the internal LB name indexed. Maybe then only the cloud config needs to learn about it.
> It would also be nice to have a DIA diagram documenting the architecture here.

I have changed the setup to create only one internal LB with one frontend/backend for each subnet. So this name is fine as is. The problem now is in the backend pools attached to the internal LB. Somewhere outside the Installer, something is matching against the backend pool name and is creating a new backend, breaking `oc` commands run against private clusters.

r4f4:
Depends on https://github.com/openshift/api/pull/1526

r4f4:
Update: fix some issues found by QE:
1. because of my misunderstanding of the `master_availability_zones` variable, cluster deploy would fail in non-zonal regions (e.g `westus`)
2. fix deploys when the zones specified for controlPlane are different from those specified for compute. Before we were only using the controlPlane zones to decide how many NATs to create, but now we considers all distinct zones for all nodes.

r4f4:
@jhixson74 with the move to capi/capz, I think this PR doesn't make sense anymore. I'm going to close it, if that's ok with you.

r4f4:
/close

Azure is now capi/capz based, so this implementation doesn't make sense anymore.

PR #7284, opened 2023-06-27
OCPBUGS-15421: Allow different service account for xpn installs in gcp

barbacbd:
**Add service account field to the install config for control Planes
**Validation to ensure service account field added is only used for xpn installs
**Add Validation to ensure that the service account is available in the host project

** Machine manifests will use the provided service account for control planes.


pkg/asset/installconfig/gcp/client.go
```diff
@@ -368,6 +370,34 @@ func (c *Client) getServiceUsageService(ctx context.Context) (*serviceusage.Serv
...
+func (c *Client) GetServiceAccounts(ctx context.Context, project string) ([]string, error) {
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+
+	svc, err := iam.NewService(ctx)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed create IAM service")
+	}
+
+	rsp, err := svc.Projects.ServiceAccounts.List(fmt.Sprintf("projects/%s", project)).Do()
```
patrickdillon, line 383: I would think we should call `ServiceAccounts.Get` instead of `List`. I think in some of these cases the number of service accounts could be quite large. I think `Get` would cover our use case?

pkg/types/gcp/machinepools.go
```diff
@@ -43,6 +43,14 @@ type MachinePool struct {
 	// +kubebuilder:validation:Enum=Enabled;Disabled
 	// +optional
 	ConfidentialCompute string `json:"confidentialCompute,omitempty"`
+
+	// ServiceAccount is the email for a gcp service account to be used only during shared
+	// vpn installations. The installer does not create a service account for shared vpn installations. The
+	// service account is passed through with the permissions of the account used for installation. In order to provide
+	// the necessary permissions for installations, users may pass in the email of a preexisting service account.
```
patrickdillon, line 50:
```
// ServiceAccount is the email of a GCP service account to be used for shared
// vpc installations. The provided service account will be attached to control-plane nodes
// in order to provide the permissions required by the cloud provider in the host project.
```

pkg/asset/machines/gcp/machines.go
```diff
@@ -161,20 +161,27 @@ func provider(clusterID string, platform *gcp.Platform, mpool *gcp.MachinePool,
 	// service accounts with host project privileges. Instead, we can use the existing service account provided
 	// to the installer.
 	if len(platform.NetworkProjectID) > 0 {
-		sess, err := gcpconfig.GetSession(context.TODO())
-		if err != nil {
-			return nil, err
-		}
+		// The service account is only valid for Control Plane Nodes. However, we don't know if this
+		// block is executed for control plane nodes or compute nodes. But validation should have already
```
patrickdillon, line 165: The `role` variable will indicate compute vs control plane

pkg/asset/machines/gcp/machines.go
```diff
@@ -161,20 +161,27 @@ func provider(clusterID string, platform *gcp.Platform, mpool *gcp.MachinePool,
...
+		// occurred to ensure that the service account is not set for compute nodes.
+		if mpool.ServiceAccount != "" {
+			instanceServiceAccount = mpool.ServiceAccount
+		} else {
```
patrickdillon, line 169: I think we need to spell out the logic here a bit more. As this code stands right now, it is still problematic because in the case where `sess.Credentials.JSON` is `nil`, this will still throw an error for the compute nodes. 

My suggestion for an approach would be:
- in all cases, the installler should create the service account for the compute nodes (it seems like this would be fine and not cause any issues with existing installs
- add validation at some point that says if `sess.Credentials.JSON` is `nil` then controlplane.gcp.serviceAccount is required

I think that should work?

barbacbd: Is it possible for the `sess.Credentials.JSON` to be `nil`on normal installs or at least those that are not with a shared vpc

patrickdillon: Yes. Any time the installer is authenticated without the service account key file (so from a VM or the gcloud cli) it will be nil.

pkg/asset/installconfig/gcp/client.go
```diff
@@ -370,32 +370,30 @@ func (c *Client) getServiceUsageService(ctx context.Context) (*serviceusage.Serv
...
+	// The service accounts are listed in the format:
+	// projects/{project-id}/serviceAccounts/{ServiceAccountName}
+	// The use of the service account is by the name only when passed through manifests, so only
+	// the service account name is pulled here
+	saSplit := strings.Split(rsp.Name, "/")
+	if len(saSplit) == 0 {
+		return "", fmt.Errorf("no service account name found")
 	}
-	return serviceAccounts, nil
+	return saSplit[len(saSplit)-1], nil
```
patrickdillon: Hm I'm not sure I understand why we are splitting the name like this and returning this value. Do we need to do this?

patrickdillon:
The most recent changes look like the right direction to me. Left a question about the string manipulation in the service account check.

The commit structure needs to be cleaned up, which will make this easier for review. 

But again, looks good and thanks for burning through this.

pkg/asset/installconfig/gcp/client.go
```diff
@@ -368,6 +370,24 @@ func (c *Client) getServiceUsageService(ctx context.Context) (*serviceusage.Serv
 	return svc, nil
 }
 
+// GetServiceAccount retrieves a service account from a project if it exists.
+func (c *Client) GetServiceAccount(ctx context.Context, project, serviceAccount string) (string, error) {
+	ctx, cancel := context.WithTimeout(ctx, 1*time.Minute)
+	defer cancel()
+
+	svc, err := iam.NewService(ctx)
```
r4f4, line 378: I think we should not give a short timeout context when creating a client/service, and instead use it for the API call itself.
https://pkg.go.dev/cloud.google.com/go#hdr-Timeouts_and_Cancellation

data/data/gcp/cluster/dns/base.tf
```diff
@@ -9,6 +9,7 @@ resource "google_dns_managed_zone" "int" {
   description = local.description
   dns_name    = "${var.cluster_domain}."
   visibility  = "private"
+  project     = var.project_id
```
patrickdillon, line 12: Why do we need to add project_id for these resources (when we previously didn't need them)? Was the project being inferred in some way?

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -50,12 +50,19 @@ func Validate(client API, ic *types.InstallConfig) error {
 		allErrs = append(allErrs, field.Invalid(field.NewPath("clusterName"), ic.ObjectMeta.Name, err.Error()))
 	}
 
+	sess, err := GetSession(context.TODO())
+	if err != nil {
+		allErrs = append(allErrs, field.InternalError(field.NewPath("session"), fmt.Errorf("failed to get gcp session")))
+	}
+	allErrs = append(allErrs, validateServiceAccountPresent(sess, ic)...)
```
patrickdillon, line 57: You should be able to follow the same pattern as the following functions, particularly I think `client` embeds a `Session` so you don't need to create one or handle errors here

patrickdillon: You can probably keep the function signature to just take a session but pass the one in the client rather than creating a new one

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -84,6 +91,20 @@ func ValidateInstanceType(client API, fieldPath *field.Path, project, zone, inst
 	return allErrs
 }
 
+func validateServiceAccountPresent(sess *Session, ic *types.InstallConfig) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	if ic.GCP.NetworkProjectID != "" {
+		if sess.Credentials != nil && sess.Credentials.JSON == nil {
+			if ic.ControlPlane.Platform.GCP != nil && ic.ControlPlane.Platform.GCP.ServiceAccount == "" {
+				errMsg := "service account must be provided when credentials are missing"
```
patrickdillon, line 100: nit: perhaps

`service account must be provided when authentication credentials do not provide a service account`

patrickdillon (commented):
This is looking in good shape. Left a few nits.

patrickdillon:
/close
in favor of #7308

PR #7285, opened 2023-06-28
Add me-central1 GCP region to the survey as supported region

makentenza:
Add the me-central1 GCP region that was not previously listed as a supported region for the installer to use.

[CORS-2699](https://issues.redhat.com/browse/CORS-2699)


barbacbd (commented):
/lgtm

PR #7288, opened 2023-06-29
CORS-2700: Make bootstrap S3 bucket optional during bootstrap destroy

rna-afk:
Making a copy of the AWS bootstrap S3 bucket in the cluster stage and adding a field to the install config. If the user does not want to delete the S3 bucket created for the bootstrap, they need to set the new field to true.

This causes the S3 bucket to be created during the cluster stage and not the boostrap stage which will then be ignored when the bootstrap destroy code runs.


rna-afk:
Tested locally with the following steps:

1. Create install config and set `ignoreS3Deletion` field under `platform.aws` to `true`
2. Create cluster.
3. You can still see the bootstrap S3 bucket after the installation is complete using the `aws s3 ls` command.
4. Deleted the cluster which also deleted the S3 bucket.
5. Also set the field to false and created a cluster which works as usual.

pkg/types/aws/platform.go
```diff
@@ -101,6 +101,12 @@ type Platform struct {
 	//
 	// +optional
 	LBType configv1.AWSLBType `json:"lbType,omitempty"`
+
+	// IgnoreS3Deletion is an optional field that can be used to make the S3 deletion optional
+	// during bootstrap destroy.
+	// See: https://issues.redhat.com/browse/CORS-2661
+	// +optional
+	IgnoreS3Deletion bool `json:"ignoreS3Deletion,omitempty"`
```
mtulio, line 109: What about using a suggested name for the bootstrap bucket? Are we using the bucket for other things than generating pre-signed URLs for bootstrap user data?

rna-afk: Added a commit that allows users to change name

patrickdillon, line 109: "Ignore" seems a little negative and I think we should include "bootstrap" and/or "ignition" in the field name, so perhaps something like:

`PreserveBootstrapIgnition`
`SkipS3BootstrapDestroy`
`SkipBootstrapIgnitionDestroy`

I prefer the first one.

pkg/types/aws/platform.go
```diff
@@ -101,6 +101,12 @@ type Platform struct {
 	//
 	// +optional
 	LBType configv1.AWSLBType `json:"lbType,omitempty"`
+
+	// IgnoreS3Deletion is an optional field that can be used to make the S3 deletion optional
+	// during bootstrap destroy.
+	// See: https://issues.redhat.com/browse/CORS-2661
```
patrickdillon, line 107: Let's not include a Jira link directly in the code (unless maybe a bug), but including in the commit message would be fine. It would be helpful to include in the commit message the motivation for this change.

pkg/types/aws/platform.go
```diff
@@ -101,6 +101,16 @@ type Platform struct {
...
+
+	// PreserveBootstrapIgnition is an optional field that can be used to make the S3 deletion optional
+	// during bootstrap destroy.
+	// +optional
+	PreserveBootstrapIgnition bool `json:"preserveBootstrapIgnition,omitempty"`
+
+	// BootstrapS3Name is an optional field that can be set to change the name of the S3 bucket
+	// created during bootstrap stage. Defaults to "<cluster_id>-bootstrap"
+	// +optional
+	BootstrapS3Name string `json:"bootstrapS3Name,omitempty"`
```
patrickdillon, line 113: Unless we get an answer for https://github.com/openshift/installer/pull/7288#discussion_r1263702327 I don't think we should allow this name to be set.

@mtulio did you have a use case in mind for allowing bootstrap ignition buckets to be named?

patrickdillon:
@rna-afk I think drop 525d3a2c663b3bb8a5a9d11c21fff9a1814c85b8. you can add it in a separate pr if you need. having this in a separate commit makes this easy. nice!

pkg/types/aws/platform.go
mtulio: @patrickdillon The only benefit I can see providing a bucket name in IC is to reuse the existing one, then when provided the installer could create an object path `infraID/bootstrap.ign` or even `infraID-bootstrap.ign` and generate the presigned URL for this object (maybe failing when already exists), so the customer can keep a single bucket with their N deployments, instead of N buckets - when the flag `PreserveBootstrapIgnition` is set.

My other suggestion is the var name, instead of using "S3", use "Bucket" - considering what is deleted is the instance (Bucket), and not the service (S3): `IgnoreBucketDeletion`

pkg/types/aws/platform.go
mtulio: or `IgnoreBootstrapBucketDeletion`

data/data/aws/cluster/main.tf
```diff
@@ -5,7 +5,15 @@ locals {
     },
     var.aws_extra_tags,
   )
-  description = "Created By OpenShift Installer"
+  description    = "Created By OpenShift Installer"
+  new_tag_len    = 8
+  sliced_tag_map = length(var.aws_extra_tags) <= local.new_tag_len ? var.aws_extra_tags : { for k in slice(keys(var.aws_extra_tags), 0, local.new_tag_len) : k => var.aws_extra_tags[k] }
+  s3_object_tags = merge(
+    {
+      "kubernetes.io/cluster/${var.cluster_id}" = "owned"
```
patrickdillon, line 13: Is the bucket successfully deleted by the cluster destroy? If so, it should be easy to add an optional ci job for this and we could have some observable testing in CI.

patrickdillon (commented):
This generally looks good to me. 

/approve

If the bucket gets cleaned up with normal bootstrap destroy, it would be pretty lightweight (one line install config change) to add ci testing, which would be nice to have for confidence.

data/data/aws/cluster/main.tf
rna-afk: Yeah it did get destroyed when I tested it. I'll create a step for this.

barbacbd (commented):
/lgtm

PR #7289, opened 2023-06-29
create: add check for cluster operator stability

r4f4:
Adds a check to see whether each cluster operator has stopped
progressing for at least 30 seconds. There is a five minute period
where operators can meet this threshold.

This check prevents against a class of errors where operators report
themselves as Available=true but continue to progress and are not
fully functional.

### DIFF
--- a/cmd/openshift-install/create.go
+++ b/cmd/openshift-install/create.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"crypto/x509"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
@@ -14,8 +15,10 @@ import (
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
@@ -776,12 +779,64 @@ func getCOProgressingStatus(ctx context.Context, cc *configclient.Clientset, nam
 	return pStatus, err
 }
 
-func wrapWithReplay(w watch.Interface) watch.Interface {
+type replayingListWatch struct {
+	delegate cache.ListerWatcher
+
+	lastListItem *watch.Event
+	lastLock     sync.Mutex
+}
+
+func (r *replayingListWatch) List(options metav1.ListOptions) (runtime.Object, error) {
+	uncastList, err := r.delegate.List(options)
+	if err != nil {
+		return nil, err
+	}
+	items, err := meta.ExtractList(uncastList)
+	if err != nil {
+		return nil, fmt.Errorf("unable to understand list result %#v (%w)", uncastList, err)
+	}
+	if len(items) == 0 {
+		return uncastList, nil
+	}
+	lastItem := items[len(items)-1]
+	// we know this should be a clusteroperator, if testing fails on this, hardconvert it here.
+
+	r.lastLock.Lock()
+	defer r.lastLock.Unlock()
+	r.lastListItem = &watch.Event{
+		Type:   watch.Added,
+		Object: lastItem,
+	}
+
+	return uncastList, nil
+}
+
+// Watch is called strictly after List because it needs a resourceVersion.
+func (r *replayingListWatch) Watch(options metav1.ListOptions) (watch.Interface, error) {
+	w, err := r.delegate.Watch(options)
+	if err != nil {
+		return w, err
+	}
+
+	r.lastLock.Lock()
+	defer r.lastLock.Unlock()
+	return wrapWithReplay(w, r.lastListItem.DeepCopy()), nil
+}
+
+func wrapWithReplay(w watch.Interface, initialLastEvent *watch.Event) watch.Interface {
 	fw := &replayingWatch{
 		incoming: w,
 		result:   make(chan watch.Event),
 		closed:   make(chan struct{}),
 	}
+	if initialLastEvent != nil {
+		func() {
+			fw.lastLock.Lock()
+			defer fw.lastLock.Unlock()
+			fw.last = *initialLastEvent
+		}()
+	}
+
 	go fw.watchIncoming()
 	go fw.resendLast()
 	return fw
@@ -848,16 +903,8 @@ func (r *replayingWatch) resendLast() {
 }
 
 func replayingListWatcher(in cache.ListerWatcher) cache.ListerWatcher {
-	return &cache.ListWatch{
-		ListFunc: in.List,
-		WatchFunc: func(options metav1.ListOptions) (watch.Interface, error) {
-			w, err := in.Watch(options)
-			if err != nil {
-				return w, err
-			}
-			return wrapWithReplay(w), nil
-		},
-		DisableChunking: true,
+	return &replayingListWatch{
+		delegate: in,
 	}
 }
 
--- a/cmd/openshift-install/create.go
+++ b/cmd/openshift-install/create.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -52,6 +53,11 @@ const (
 	exitCodeInfrastructureFailed
 	exitCodeBootstrapFailed
 	exitCodeInstallFailed
+	exitCodeOperatorStabilityFailed
+
+	// coStabilityThreshold is how long a cluster operator must have Progressing=False
+	// in order to be considered stable. Measured in seconds.
+	coStabilityThreshold float64 = 30
 )
 
 // each target is a variable to preserve the order when creating subcommands and still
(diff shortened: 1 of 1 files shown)


r4f4:
/hold
while we check how this fares with some CI runs

deads2k:
Good catch.

/lgtm

r4f4:
Update: squashed commits, no code changes.
/hold cancel

sdodson:
A number of odd test infra looking failures in the last run.

cmd/openshift-install/create.go
```diff
@@ -549,6 +555,57 @@ func waitForInitializedCluster(ctx context.Context, config *rest.Config) error {
...
 }
 
+// waitForStableOperators ensures that each cluster operator is "stable", i.e. the
+// operator has not been in a progressing state for at least a certain duration,
+// 30 seconds by default. Returns an error if any operator does meet this threshold
+// after a deadline, 5 minutes by default.
+func waitForStableOperators(ctx context.Context, config *rest.Config) error {
+	timer.StartTimer("Cluster Operators Stable")
+
+	stabilityCheckDuration := 10 * time.Minute
```
deads2k, line 565: bump to 30.

deads2k:
This run failed on the wait for apparently legitimate reasons: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/7289/pull-ci-openshift-installer-master-e2e-vsphere-ovn/1676911592173735936

a node never came up.  That failure is correct and isn't a problem with this PR  per-se, but extending the timeout will make the "it never worked" more clear.  Suggested 30 minutes above.

cmd/openshift-install/create.go
r4f4: Sure. And I just noticed the function's comment doesn't match the actual timeout set. I'll fix that too.

patrickdillon:
LGTM
/approve

cmd/openshift-install/create.go
```diff
@@ -549,6 +555,58 @@ func waitForInitializedCluster(ctx context.Context, config *rest.Config) error {
...
+	// encounters a stability error
+	stabilityCheck := coStabilityChecker()
+
+	var wg sync.WaitGroup
+	for _, co := range coNames {
+		wg.Add(1)
+		go func(co string) {
+			defer wg.Done()
+			status, statusErr := getCOProgressingStatus(stabilityContext, cc, co)
+			err = stabilityCheck(co, status, statusErr)
```
wking, line 594: This looks like a single `err` for all of the ClusterOperators.  What happens if one is `Progressing=True` but another goes `Progressing=False` right on the `stabilityCheckDuration` threshold?  Isn't there a risk that the happy response sets a `nil` in `err`, and the installer exits happy when openshift/enhancements#1189 would have wanted it to exit sad?  Maybe you can use an `[]error` slice and then use `NewAggregate(errs)` or some such to aggregate?

cmd/openshift-install/create.go
```diff
@@ -655,3 +717,177 @@ func checkIfAgentCommand(assetStore asset.Store) {
 		logrus.Warning("An agent configuration was detected but this command is not the agent wait-for command")
 	}
 }
+
+func getClusterOperatorNames(ctx context.Context, cc *configclient.Clientset) ([]string, error) {
+	listCtx, cancel := context.WithTimeout(ctx, 1*time.Minute)
```
wking, line 722: I don't understand this 1m timeout.  It seems unlikely that the list takes longer than that, but if it does, and the wrapping `ctx` has more time available, wouldn't we want to wait longer?

cmd/openshift-install/create.go
wking: Ah, I see the `coStabilityChecker returns a closure which references a shared error variable...` comment below.  Feels a bit opaque to handle it that way, but :shrug:, looks like it will work.

cmd/openshift-install/create.go
```diff
@@ -655,3 +717,177 @@ func checkIfAgentCommand(assetStore asset.Store) {
...
+					logrus.Debugf("Cluster Operator %s progressing == nil", name)
+					return false, nil
+				}
+				pStatus = progressing
+
+				if meetsStabilityThreshold(pStatus) {
+					logrus.Debugf("Cluster Operator %s is stable", name)
+					return true, nil
+				}
+				logrus.Debugf("Cluster Operator %s is Progressing=%s LastTransitionTime=%v DurationSinceTransition=%.fs Reason=%s Message=%s", name, progressing.Status, progressing.LastTransitionTime.Time, time.Since(progressing.LastTransitionTime.Time).Seconds(), progressing.Reason, progressing.Message)
```
wking, line 771: [Install logs][1] from the [vSphere OVN run][2]:

```
time="2023-07-06T17:51:56Z" level=debug msg="Cluster Operator kube-apiserver is Progressing=False LastTransitionTime=2023-07-06 17:51:56 +0000 UTC DurationSinceTransition=0s Reason=AsExpected Message=NodeInstallerProgressing: 3 nodes are at revision 6"
time="2023-07-06T17:51:56Z" level=debug msg="Cluster Operator kube-apiserver is Progressing=False LastTransitionTime=2023-07-06 17:51:56 +0000 UTC DurationSinceTransition=0s Reason=AsExpected Message=NodeInstallerProgressing: 3 nodes are at revision 6"
time="2023-07-06T17:51:57Z" level=debug msg="Cluster Operator kube-apiserver is Progressing=False LastTransitionTime=2023-07-06 17:51:56 +0000 UTC DurationSinceTransition=1s Reason=AsExpected Message=NodeInstallerProgressing: 3 nodes are at revision 6"
time="2023-07-06T17:51:58Z" level=debug msg="Cluster Operator kube-apiserver is Progressing=False LastTransitionTime=2023-07-06 17:51:56 +0000 UTC DurationSinceTransition=2s Reason=AsExpected Message=NodeInstallerProgressing: 3 nodes are at revision 6"
time="2023-07-06T17:51:59Z" level=debug msg="Cluster Operator kube-apiserver is Progressing=False LastTransitionTime=2023-07-06 17:51:56 +0000 UTC DurationSinceTransition=3s Reason=AsExpected Message=NodeInstallerProgressing: 3 nodes are at revision 6"
```

Is this something we want to log each second?  Especially when we hit the `Progressing=False` target and are just waiting out the clock on `coStabilityThreshold`?  Maybe we could log once when it transitions to `Progressing=False`, and not until it changes after that?  On the other hand, it's debug-level logs, so maybe not worth the effort to denoise?

[1]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_installer/7289/pull-ci-openshift-installer-master-e2e-vsphere-ovn/1676982353563488256/artifacts/e2e-vsphere-ovn/ipi-install-install/artifacts/.openshift_install-1688665947.log
[2]: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/7289/pull-ci-openshift-installer-master-e2e-vsphere-ovn/1676982353563488256

cmd/openshift-install/create.go
```diff
@@ -655,3 +717,177 @@ func checkIfAgentCommand(assetStore asset.Store) {
...
+	var err error
+
+	return func(name string, status *configv1.ClusterOperatorStatusCondition, statusErr error) error {
+		if statusErr == nil {
+			return err
+		}
+		if !wait.Interrupted(statusErr) {
+			logrus.Errorf("Error checking cluster operator %s Progressing status: %q", name, statusErr)
+			logrus.Exit(exitCodeOperatorStabilityFailed)
+			err = errors.New("cluster operators are not stable")
```
wking, line 878: Is this reachable code?  It looks like [`logrus.Exit(...)`][1] is fatal, but maybe I'm missing some context?

[1]: https://pkg.go.dev/github.com/sirupsen/logrus#Exit

r4f4: @patrickdillon do you remember if `logrus.Exit` is just an artifact of a trial phase? It does look like we want to set and return the error, not exit the install immediately.

wking: `logrus.Exit(exitCode...)` looks like existing precedent to me.  But yes, I would expect to be returnning `err` here, and then using `logrus.Exit(exitCodeOperatorStabilityFailed)` right after the `wg.Wait()` if `err` wasn't `nil`.

patrickdillon: The pattern trevor suggested makes sense. I can't remember the motivation for originally writing this code. Perhaps the idea was to fail quickly if there were problems checking the status, but it definitely looks like it could be improved

patrickdillon:
/hold

I'm suspicious of the results in [the vsphere test](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/7289/pull-ci-openshift-installer-master-e2e-vsphere-ovn/1677158351949008896)

The round number gives me pause:
> time="2023-07-07T05:50:15Z" level=debug msg="   Cluster Operators Stable: 30m0s"

For some reason, in this run, it looks like the authentication operator is stable, but we are waiting quite a while (until the deadline?) to register that stability

```
time="2023-07-07T05:20:15Z" level=debug msg="Cluster Operator authentication is Progressing=False LastTransitionTime=2023-07-07 05:19:51 +0000 UTC DurationSinceTransition=25s Reason=AsExpected Message=AuthenticatorCertKeyProgressing: All is well"
[...]
time="2023-07-07T05:21:04Z" level=debug msg="Cluster Operator kube-apiserver is stable"
time="2023-07-07T05:50:15Z" level=debug msg="Cluster operator authentication is now stable: Progressing=False LastTransitionTime=2023-07-07 05:19:51 +0000 UTC DurationSinceTransition=1825s Reason=AsExpected Message=AuthenticatorCertKeyProgressing: All is well"
time="2023-07-07T05:50:15Z" level=info msg="All cluster operators have completed progressing"
```

cmd/openshift-install/create.go
r4f4: Done

r4f4:
Should be fixed now.
/hold cancel

PR #7290, opened 2023-06-30
OCPBUGS-15238: GCP: ic: client: use a higher context timeout

r4f4:
We are probably being too optimistic with timeouts like 30s and 1min for API calls. If your account happens to have access to thousands of projects, the API calls can easily exceed that. Let's settle in 2min for now.

I also used the opportunity to fix a few instances where the context passed as parameter was being ignored.

### DIFF
--- a/pkg/asset/installconfig/gcp/client.go
+++ b/pkg/asset/installconfig/gcp/client.go
@@ -52,9 +52,6 @@ type Client struct {
 
 // NewClient initializes a client with a session.
 func NewClient(ctx context.Context) (*Client, error) {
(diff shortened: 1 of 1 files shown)


barbacbd (commented):
/lgtm 

Let's hope that this is enough time. Only the QE project will let us know.

r4f4:
Update: removed spurious gofmt change. No change in functionality.

r4f4:
/retitle OCPBUGS-15238: GCP: ic: client: use a higher context timeout

jianli-wei (approved):

PR #7291, opened 2023-07-01
CORS-2719: Remove service account user permission

barbacbd:
** Removed the addition of the service account user permission for control plane nodes in terraform.

The service account user permission is passed to MAO as it is needed by that operator to access specific resources. It is not needed during the creation of master/control plane nodes.

### DIFF
--- a/data/data/gcp/cluster/master/main.tf
+++ b/data/data/gcp/cluster/master/main.tf
@@ -37,13 +37,6 @@ resource "google_project_iam_member" "master-storage-admin" {
   member  = "serviceAccount:${google_service_account.master-node-sa[0].email}"
 }
 
-resource "google_project_iam_member" "master-service-account-user" {
-  count   = var.service_account == "" ? 1 : 0
-  project = var.project_id
-  role    = "roles/iam.serviceAccountUser"
-  member  = "serviceAccount:${google_service_account.master-node-sa[0].email}"
-}
-
 resource "google_compute_instance" "master" {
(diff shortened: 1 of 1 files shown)


r4f4 (commented):
/lgtm

patrickdillon:
This was originally added in https://github.com/openshift/installer/commit/a402e9986229378e9d009245533224cef9523597 with the explanation 
> The service user role is required for masters to be able to attach a
disk to an instance running as a different role.
See https://cloud.google.com/compute/docs/disks/add-persistent-disk

Reviewing that link, I don't see anything about those permissions
>   compute.disks.create on the project before you can create a new disk
    compute.instances.attachDisk on the virtual machine (VM) instance
    compute.disks.use permission on the disk to attach

Looking at a historical copy of that page from around the time of the commit: https://web.archive.org/web/20190711121954/https://cloud.google.com/compute/docs/disks/add-persistent-disk

I see:
> If you are connecting to a VM instance that can run as a service account, you must also grant the roles/iam.serviceAccountUser role.

This is under the Permissions Required for Formatting and mounting a zonal persistent disk

I'm a little wary of Chesterton's Fence here.

I suspect the behavior in question would be exercised in e2e tetsts:
/retest

patrickdillon:
Speaking with @csrwng out-of-band he confirmed:

>  if there's an issue it would present itself when attempting to create a pod that references a pvc in a worker node.

which should be covered extensively in e2e tests.

PR #7293, opened 2023-07-05
OCPBUGS-15825: Fix agent gather tui logs

andfasano:
This patch adds the agent-tui log to the agent-gather artifact


pawanpinjarkar (approved):
/lgtm

PR #7296, opened 2023-07-05
WIP: Update assisted-service vendored code and use nmstate

bfournie:
This is a POC to use the nmstate lib that assisted-service includes and remove the need for nmstatectl in nmstateconfig.yaml. It requires the nmstate-devel pkg.


images/baremetal/Dockerfile.ci
```diff
@@ -6,7 +6,7 @@ ARG libvirt_version="8.0.0"
 FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.20-openshift-4.14 AS builder
 ARG libvirt_version
 ARG TAGS="libvirt baremetal"
-RUN dnf install -y libvirt-devel-$libvirt_version && \
+RUN dnf install --enablerepo=powertools -y libvirt-devel-$libvirt_version nmstate-devel && \
```
andfasano, line 9: What's the need to enable powertools? In agent-installer-utils we use just the base image [repos](https://github.com/openshift/agent-installer-utils/blob/main/images/Dockerfile.ocp)

hack/build.sh
```diff
@@ -74,6 +74,7 @@ if (echo "${TAGS}" | grep -q 'libvirt')
 then
 	export CGO_ENABLED=1
 fi
+export CGO_ENABLED=1
```
andfasano, line 77: If this is going to be effectively an always required setting, then the previous block could be removed

PR #7299, opened 2023-07-06
AGENT-660 Display insert config image message to console

rwsu:
The new service will only be added to the unconfigured ignition.

The purpose is to give customers some indication that in order for the hosts to proceed with cluster installation the config image need to be inserted or mounted.


pawanpinjarkar (commented):
/lgtm

PR #7304, opened 2023-07-09
OCPBUGS-15852: Single node cannot be installed if etcd appears in the hostname

eranco74:
Add filter by etcd container name regular expression pattern before the grep

https://issues.redhat.com/browse/OCPBUGS-15852

### DIFF
--- a/data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/bootstrap-in-place.sh
+++ b/data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/bootstrap-in-place.sh
@@ -16,7 +16,7 @@ if [ ! -f stop-etcd.done ]; then
   echo "Stop etcd static pod by moving the manifest"
   mv /etc/kubernetes/manifests/etcd-member-pod.yaml /etc/kubernetes || echo "already moved etcd-member-pod.yaml"
 
-  until ! crictl ps | grep etcd; do
+  until crictl ps --state running -o json | jq -e '[.containers[] | select(.metadata.name == "etcd")] | length == 0' > /dev/null; do
     echo "Waiting for etcd to go down"
     sleep 10
   done


data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/bootstrap-in-place.sh
```diff
@@ -16,7 +16,7 @@ if [ ! -f stop-etcd.done ]; then
   echo "Stop etcd static pod by moving the manifest"
   mv /etc/kubernetes/manifests/etcd-member-pod.yaml /etc/kubernetes || echo "already moved etcd-member-pod.yaml"
 
-  until ! crictl ps | grep etcd; do
+  until ! crictl ps --name etcd | grep etcd; do
```
omertuc, line 19: `--name etcd` might still capture `etcd` in container names which might contain the hostname, maybe we should work with structured output instead:

```suggestion
  until sudo crictl ps --state running -o json | jq -e '[.containers[] | select(.metadata.name == "etcd")] | length == 0'; do
```

eranco74: I'm sure that the container name during bootstrap is just etcd, yet this jq seems more robust

omertuc: The etcd container name is just etcd, but other containers might be running and might have the hostname in their name

data/data/bootstrap/bootstrap-in-place/files/usr/local/bin/bootstrap-in-place.sh
```diff
@@ -16,7 +16,7 @@ if [ ! -f stop-etcd.done ]; then
   echo "Stop etcd static pod by moving the manifest"
   mv /etc/kubernetes/manifests/etcd-member-pod.yaml /etc/kubernetes || echo "already moved etcd-member-pod.yaml"
 
-  until ! crictl ps | grep etcd; do
+  until sudo crictl ps --state running -o json | jq -e '[.containers[] | select(.metadata.name == "etcd")] | length == 0' > /dev/null; do
```
omertuc, line 19: Sorry didn't mean to suggest sudo

sadasu:
/retitle OCPBUGS-15852: Single node cannot be installed if etcd appears in the hostname

PR #7305, opened 2023-07-10
OCPBUGS-15861: Load libvirt library on demand 

zaneb:
The libvirt-go library can now dynamically load the libvirt client
sharted library on demand using dlopen(), instead of linking to it
dynamically (which requires the shared library to always be present at
both build and run time).

This means that the same Linux binaries can be used even on systems that
don't have the libvirt client installed, as long as they are needed only
for code that doesn't involve calling libvirt. (Currently this is
everything except the baremetal platform and the 'destroy cluster'
command on the libvirt platform. All other interaction with libvirt is
via Terraform.)


berrange:
FYI, you should no longer need to install `libvirt-devel` with this change, as the `libvirt-go` build is entirely self-contained when the `libvirt_dlopen` tag is set.

You'll only need `libvirt-daemon-kvm` (or `libvirt-daemon-qemu` for non-KVM hosts), as the runtime environment.

I see `libvirt-devel` mentioned in a couple of docs and dockerfiles.

zaneb:
Note that the e2e-crc and e2e-libvirt tests have been permafailing for a while.
The functionality that still uses this library is mostly not exercised by the e2e tests. (There's some validation code in metal ipi that uses it; I haven't checked if it actually gets run. Otherwise it's the `destroy cluster` command in metal and libvirt only.) Hence WIP until I (or somebody) get the chance to do some manual testing.

zaneb:
#7252 does more for us, and we don't really have any reason to backport this change at the moment, so it is not required.
/close

PR #7307, opened 2023-07-10
OCPBUGS-14900: Use correct SELinux label. Make rename atomic.

pawanpinjarkar:
Signed-off-by: Zane Bitter <zbitter@redhat.com > and Pawan Pinjarkar <ppinjark@redhat.com>

### DIFF
--- a/data/data/agent/files/usr/local/bin/issue_status.sh
+++ b/data/data/agent/files/usr/local/bin/issue_status.sh
@@ -8,7 +8,7 @@ set_issue() {
     local outfile
     outfile="$(issue_file "$1")"
     local tmp
-    tmp="$(mktemp)"
+    tmp="$(mktemp -p /etc/issue.d/)"
     {
         printf '\n'
         cat -


data/data/agent/files/usr/local/bin/issue_status.sh
```diff
@@ -1,25 +1,19 @@
...
 issue_file() {
-    printf "/etc/issue.d/%s.issue" "$1"
+    printf "/etc/issue.d/%s" "$1"
 }
 
 set_issue() {
     local outfile
     outfile="$(issue_file "$1")"
-    local tmp
-    tmp="$(mktemp)"
```
zaneb, line 11: One advantage of using mktemp is that there is no way for two processes running this code to interfere with each other. I think we could continue to get the benefits by doing

```
tmp=$(mktemp -p /etc/issue.d/)
```

and then none of the other code needs to change.

pawanpinjarkar:
After fix
Screenshot 2023-07-11 at 10 04 52 AM

rwsu (commented):
/lgtm

PR #7309, opened 2023-07-10
OCPBUGS-15997: openstack: add root volume AZ validation

EmilienM:
## Problem

When a machine is created with a compute availability zone (defined via `mpool.zones`) and a storage root volume (defined as `mpool.rootVolume`) and that `rootVolume` has no specified `zones`, CAPO will use the compute AZ for the volume AZ.
This can be problematic if the AZ doesn't exist in Cinder.

Source:
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/9d183bd479fe9aed4f6e7ac3d5eee46681c518e7/pkg/cloud/services/compute/instance.go#L439-L442

```golang
func (s *Service) getOrCreateRootVolume(eventObject runtime.Object, instanceSpec *InstanceSpec, imageID string) (*volumes.Volume, error) {

(...)

        availabilityZone := instanceSpec.FailureDomain
        if rootVolume.AvailabilityZone != "" {
                availabilityZone = rootVolume.AvailabilityZone
        }

(...)
```

## Proposed solution

If a compute AZ is provided alongside with a root volume, we now require
the root volume to have an AZ, so we force the user to make a choice on
which AZ the root volume is deployed on.

We are also enforcing it via CEL validation in https://github.com/openshift/api/pull/1521

## Alternatives

* Do nothing - at the risk of hitting this situation: a failure domain with a Compute AZ and a root volume with no AZ, CAPO using the compute AZ to create the volume but that AZ doesn't exist in Cinder, leading into Machine creation errors.
* Only do a validation in the CPMS - which will require CPMS manual
  edits from the user.
* Change logic in CAPO wrt how root volume AZ is picked - unlikely to happen

### DIFF
--- a/docs/user/openstack/customization.md
+++ b/docs/user/openstack/customization.md
@@ -49,6 +49,9 @@ Beyond the [platform-agnostic `install-config.yaml` properties](../customization
   * `size` (required integer): Size of the root volume in GB. Must be set to at least 25.
   * `type` (required string): The volume pool to create the volume from.
   * `zones` (optional list of strings): The names of the availability zones you want to install your root volumes on. If unset, the installer will use your default volume zone.
+    If compute `zones` contains at least one value, `rootVolume.zones` must also contain at least one value.
+    Indeed, when a machine is created with a compute availability zone and a storage root volume with no specified rootVolume.availabilityZone, [CAPO](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/9d183bd479fe9aed4f6e7ac3d5eee46681c518e7/pkg/cloud/services/compute/instance.go#L439-L442) will use the compute AZ for the volume AZ.
+    This can be problematic if the AZ doesn't exist in Cinder, therefore we enforce that `rootVolume.zones` to be set if `zones` is set.
 * `zones` (optional list of strings): The names of the availability zones you want to install your nodes on. If unset, the installer will use your default compute zone.
 
 > **Note**
--- a/pkg/types/openstack/validation/machinepool_test.go
+++ b/pkg/types/openstack/validation/machinepool_test.go
@@ -13,6 +13,14 @@ func withServerGroupPolicy(serverGroupPolicy string) func(*openstack.MachinePool
 	return func(mp *openstack.MachinePool) { mp.ServerGroupPolicy = openstack.ServerGroupPolicy(serverGroupPolicy) }
 }
 
+func withRootVolume(rootVolume openstack.RootVolume) func(*openstack.MachinePool) {
+	return func(mp *openstack.MachinePool) { mp.RootVolume = &rootVolume }
+}
+
+func withAvailabilityZones(zones []string) func(*openstack.MachinePool) {
+	return func(mp *openstack.MachinePool) { mp.Zones = zones }
+}
+
 func testMachinePool(options ...func(*openstack.MachinePool)) *openstack.MachinePool {
 	var mp openstack.MachinePool
 	for _, apply := range options {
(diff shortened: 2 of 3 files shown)


pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -49,6 +49,10 @@ func ValidateMachinePool(p *openstack.MachinePool, ci *CloudInfo, controlPlane b
 			logrus.Warnf("Volume size is recommended to be greater than %d GB to use root volumes, had %d GB", recommendedStorage, p.RootVolume.Size)
 		}
 
+		if len(p.Zones) > 0 && len(p.RootVolume.Zones) == 0 {
```
EmilienM, line 52: just a note for the reviewers, there is no CEL validation in the installer (e.g. `XValidation:rule`) which could have been handy too.

pkg/asset/installconfig/openstack/validation/machinepool.go
```diff
@@ -49,6 +49,10 @@ func ValidateMachinePool(p *openstack.MachinePool, ci *CloudInfo, controlPlane b
...
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("rootVolume").Child("zones"), p.RootVolume.Zones, "root volume availability zones must be specified when compute availability zones are specified"))
```
EmilienM, line 53: The question I've asked myself is whether or not we want to enforce it only for the control plane machines or for all.

To me the same problem applies to workers which is why I've not made the difference - however I'm open for feedback on this one.

dulek:
/lgtm

I'm convinced by the arguments in the PR description. I'm leaving the approval for someone else to express their opinion.

pkg/asset/installconfig/openstack/validation/machinepool.go
EmilienM: Since this is a static validation, I think I have to move it to `pkg/types/openstack/validation/machinepool.go`. I learnt that today thanks to @pierreprinetti .

I'll address that in a next patchset.

pierreprinetti: Nice, thanks

pkg/types/openstack/validation/machinepool_test.go
```diff
@@ -71,6 +79,21 @@ func TestValidateMachinePool(t *testing.T) {
...
 		},
+		{
+			"with availability zone and valid root volume",
+			testMachinePool(withAvailabilityZones([]string{"az0", "az1", "az2"}), withRootVolume(openstack.RootVolume{Zones: []string{"az0", "az1", "az2"}, Size: 100})),
+			"default",
+			check(noError),
+		},
+		{
+			"with availability zone and invalid root volume",
+			testMachinePool(withAvailabilityZones([]string{"az0", "az1", "az2"}), withRootVolume(openstack.RootVolume{Size: 0})),
```
pierreprinetti, line 90: Why isn't this also triggering an error with the volume size?

EmilienM: The Size validation exists but in the other file where "dynamic" validations are. I think this is fine as it is now?
Do you see anything I should change?

EmilienM: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/openstack/validation/machinepool.go#L44-L56

docs/user/openstack/customization.md
```diff
@@ -49,6 +49,9 @@ Beyond the [platform-agnostic `install-config.yaml` properties](../customization
   * `size` (required integer): Size of the root volume in GB. Must be set to at least 25.
   * `type` (required string): The volume pool to create the volume from.
   * `zones` (optional list of strings): The names of the availability zones you want to install your root volumes on. If unset, the installer will use your default volume zone.
+    If compute `zones` contains at least one value, `rootVolume.zones` must also contain at least one value.
+    Indeed, when a machine is created with a compute availability zone and a storage root volume with no specified rootVolume.availabilityZone, [CAPO](https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/9d183bd479fe9aed4f6e7ac3d5eee46681c518e7/pkg/cloud/services/compute/instance.go#L439-L442) will use the compute AZ for the volume AZ.
+    This can be problematic if the AZ doesn't exist in Cinder, therefore we enforce that `rootVolume.zones` to be set of `zones` is set.
```
MaysaMacedo, line 54: Nit:
This is a bit confusing. Should it be:
```suggestion
    This can be problematic if the AZ doesn't exist in Cinder, therefore we enforce that `rootVolume.zones` to be set when `zones` is set.
```

MaysaMacedo (commented):
/approve

I just had a nit, I won't hold the merging because of it.

PR #7319, opened 2023-07-12
prevent send to closed channel

deads2k:
Builds on top of the revert.

The channel closes could race, using a basic lock will not race.

/assign @r4f4


deads2k:
/hold

still concerned over possible false etcd progressing=true

r4f4:
Don't forget to update https://github.com/openshift/installer/pull/7319/files#diff-56276d5381d618d46ec8d35d93210662c8fdd4c9bcd90fd36afbc6a59227eb0bR830

patrickdillon:
/close

This has been solved in another PR.

PR #7323, opened 2023-07-12
AGENT: publish services diagrams

rwsu:
Includes service diagrams for automated and factory flows.


docs/user/agent/agent-services.md
andfasano, line 1: I feel it's a kind of doc more useful for dev than for a normal user

docs/user/agent/agent-services.md
```diff
@@ -0,0 +1,9 @@
+# Agent-based Installer Services
```
andfasano, line 1: It would be nice to add just a small introduction (few lines), just to describe what are these diagrams and in which moment of the lifecycle are triggered (the bootstrap)

docs/user/agent/agent-services.md
rwsu: I think end users could benefit from understanding the services. If their cluster fails to install this may provide some useful information as to where to look.

docs/user/agent/agent-services.md
rwsu: Good idea. Added.

docs/user/agent/agent-services.md
```diff
@@ -0,0 +1,35 @@
...
+![Automated Flow](./agent_installer_services-automated_flow.png)
+
+In the automated flow, all information needed to configure a cluster is included in the agent ISO.
+When the agent ISO is booted, agent-tui and nmstate libraries are copied to the host from initrd. This copy 
+step is required because the agent-tui and nmstate libraries are too big to fit into the Ignition config.
+
+* agent-interactive-console - checks the host has connectivity to the release image
+* agent - takes host inventory and performs validations and reports results back to assisted-service
+* selinux - updates the SELinux policy to allow chronyd access to container files and UNIX datagram socket 
+* pre-network-manager-config - applies any static network configs before NetworkManager is started
```
bfournie, line 14: One nit - since the network configs can also be used to define interfaces with DHCP, maybe remove "static"

docs/user/agent/agent-services.md
```diff
@@ -0,0 +1,35 @@
...
+## Appliance flow using unconfigured ignition and config image
+
+![Appliance Flow](./agent_installer_services-unconfigured_ignition_and_config_image_flow.png)
+
+In the appliance flow, most of the cluster deployment information is included in a config image that is mounted
+onto the host running the unconfigured-ignition. The appliance flow does not include the agent-tui or the
+agent-interactive-console service because the disk images generated in the factory contains the release payload
+so no connectivity checks to the release image is needed.
+
+* load-config-iso@ - detects the config image has been mounted and copies the cluster configuration files from the config image to the host. It also enables start-cluster-installation which is disabled by default in the unconfigured-ignition
```
bfournie, line 34: Might want to note something like - the actual service is created when the config image is mounted as a device and its name will be `load-config-iso@<dev-name>.service`  e.g `load-config-iso@-dev-sr1.service`

rwsu: Thanks Bob. Adding it.

andfasano:
tide seems to be stuck, let's try a refresh.

/retest-required

PR #7329, opened 2023-07-14
Bug 2108582 [4.7] RHCOS: move to rhcos.mirror.openshift.com

vfreex:
rhcos.mirror.openshift.com is the new formal location to download RHCOS boot images. It is backed by CloudFront CDN, which should be more reliable and faster than the rhcos-redirector.


vfreex:
It might not be required. But it would be ideal if we can move those urls to [rhcos.mirror.openshift.com](http://rhcos.mirror.openshift.com/) for 4.7 and 4.6 as well:

sdodson:
Looks like the known issue in 4.6-4.8 which will be fixed when https://github.com/openshift/kubernetes/pull/1633 and associated vendoring into openshift/origin land.

PR #7330, opened 2023-07-14
Bug 2108582 [4.6] RHCOS: move to rhcos.mirror.openshift.com

vfreex:
rhcos.mirror.openshift.com is the new formal location to download RHCOS boot images. It is backed by CloudFront CDN, which should be more reliable and faster than the rhcos-redirector.


vfreex:
It might not be required. But it would be ideal if we can move those urls to [rhcos.mirror.openshift.com](http://rhcos.mirror.openshift.com/) for 4.7 and 4.6 as well:

sdodson:
vSphere failure was on glibc issues associated with fips backport post install, lets override this failing test.

patrickdillon:
> vSphere failure was on glibc issues associated with fips backport post install, lets override this failing test.

+1. The other failures are either ci config errors, or e2e fails after a successful install. I don't see any problems related to this pr.

PR #7332, opened 2023-07-16
OCPBUGS-16380: Add /etc/containers volume on create-cluster-and-infraenv

danielerez:
As done in assisted-service.service.template, added '-v /etc/containers:/etc/containers' when HaveMirrorConfig flag is true to the create-cluster-and-infraenv service. This should allow propagation of registries.conf, which is required on disconnected environments when using OPENSHIFT_INSTALL_RELEASE_IMAGE_MIRROR.

### DIFF
--- a/data/data/agent/systemd/units/create-cluster-and-infraenv.service.template
+++ b/data/data/agent/systemd/units/create-cluster-and-infraenv.service.template
@@ -13,7 +13,7 @@ EnvironmentFile=/usr/local/share/assisted-service/agent-images.env
 EnvironmentFile=/usr/local/share/assisted-service/assisted-service.env
 ExecStartPre=/bin/rm -f %t/%n.ctr-id
 ExecStartPre=/usr/local/bin/wait-for-assisted-service.sh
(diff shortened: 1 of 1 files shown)


bfournie:
Is the same change needed for apply-host-config.service which also uses agent-installer-client?
https://github.com/openshift/installer/blob/master/data/data/agent/systemd/units/apply-host-config.service#L17

danielerez:
> Is the same change needed for apply-host-config.service which also uses agent-installer-client? https://github.com/openshift/installer/blob/master/data/data/agent/systemd/units/apply-host-config.service#L17

Seems it's unnecessary there as we don't pass the OPENSHIFT_INSTALL_RELEASE_IMAGE_MIRROR env var.

bfournie:
The change looks good but is it possible to create a bug to describe what this is fixing? We haven't seen this omission causing a problem with the general agent installer flow (yet). Was the issue specific to the Openshift appliance?

rwsu (commented):
/lgtm

PR #7333, opened 2023-07-17
OCPBUGS-16245: Make nmstateconfig.yaml optional in config-drive

zaneb:


pkg/asset/agent/image/unconfigured_ignition.go
```diff
@@ -27,11 +27,11 @@ const (
 func GetConfigImageFiles() []string {
 	return []string{
 		"/etc/assisted/manifests/pull-secret.yaml", //nolint:gosec // not hardcoded credentials
-		"/etc/assisted/manifests/nmstateconfig.yaml",
 		"/etc/assisted/manifests/cluster-deployment.yaml",
 		"/etc/assisted/manifests/cluster-image-set.yaml",
 		"/etc/assisted/manifests/agent-cluster-install.yaml",
 		"/etc/assisted/manifests/infraenv.yaml",
+		"/etc/assisted/manifests",       // optional nmstateconfig.yaml
```
andfasano, line 34: AFAIU the load-config-iso [script](https://github.com/openshift/installer/blob/c2d7db9d4eedf7b79fcf975f3cbd8042542982ca/data/data/agent/files/usr/local/bin/load-config-iso.sh#L37) performs the existence check only on files but not on the folders, something non immediately clear from here, it may be useful to add a note for that

rwsu: When I tested it, for directories, it copied all the files inside the directory. So this is a good way to include nmstateconfig.yaml and have it optional.

rwsu (commented):
/lgtm

PR #7338, opened 2023-07-18
OCPBUGS-15994: Update core password after loading config-image

bfournie:
The core user password should be set to the one stored in kubeadmin-password.hash after loading the config-image.

### DIFF
--- a/data/data/agent/files/usr/local/bin/load-config-iso.sh
+++ b/data/data/agent/files/usr/local/bin/load-config-iso.sh
@@ -13,6 +13,7 @@ EXTRA_MANIFESTS="/etc/assisted/extra-manifests"
 ASSISTED_NETWORK_DIR="/etc/assisted/network"
 NM_CONNECTION_DIR="/etc/NetworkManager/system-connections"
 PASSWORD_HASH="/opt/agent/tls/kubeadmin-password.hash"
+OVERRIDE_PASSWORD_SET="/etc/assisted/appliance-override-password-set"
 
 copy_archive_contents() {
     tmpdir=$(mktemp --tmpdir -d "config-image--XXXXXXXXXX")
@@ -66,8 +67,11 @@ copy_archive_contents() {
     echo "Successfully copied contents of ${AGENT_CONFIG_ARCHIVE_FILE} on ${devname}"
 
     # Update core password with one from config-image if not overridden by appliance
-    if [[ -z "${SKIP_CORE_PASSWORD_SET}" ]] || [[ "${SKIP_CORE_PASSWORD_SET}" == "false" ]]; then
+    if ! cpio --list < "${unzipped_file}" | grep -q -e "${OVERRIDE_PASSWORD_SET}"; then
+       echo "Setting core password"
        usermod --password "$(cat ${PASSWORD_HASH})" core
+    else
+       echo "Setting of core password is overridden"
     fi
 
     # Enable any services which are not enabled by default
--- a/data/data/agent/files/usr/local/bin/load-config-iso.sh
+++ b/data/data/agent/files/usr/local/bin/load-config-iso.sh
@@ -67,7 +67,7 @@ copy_archive_contents() {
     echo "Successfully copied contents of ${AGENT_CONFIG_ARCHIVE_FILE} on ${devname}"
 
     # Update core password with one from config-image if not overridden by appliance
-    if ! cpio --list < "${unzipped_file}" | grep -q -e "${OVERRIDE_PASSWORD_SET}"; then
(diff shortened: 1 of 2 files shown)


bfournie:
Verified that can successfully log in via the console using the kubeadmin-password.

/retest

bfournie:
> Any way to view the password without manually extracting the config image? Maybe worth adding it to the output of config-image API?

Its is available in auth/kubeadmin-password after building the config-image
```
$ ./bin/openshift-install agent create config-image --dir ./cluster-manifests/ --log-level debug
$ ls cluster-manifests/auth/kubeadmin-password 
cluster-manifests/auth/kubeadmin-password
```

rwsu (approved):
/approve

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -64,6 +65,9 @@ copy_archive_contents() {
 
     echo "Successfully copied contents of ${AGENT_CONFIG_ARCHIVE_FILE} on ${devname}"
 
+    # Update core password with one from config-image
+    usermod --password "$(cat ${PASSWORD_HASH})" core
```
danielerez, line 69: Won't it affect only the bootstrap step? I mean, after rebooting and start the installation it won't be possible to login using that password?
Also, kubeadmin-password is quite long, so it isn't ideal for connecting through console.
Can we maybe make this behaviour optional some how? I.e. some env var or file to override the password with a custom one. Or just disable the update of the password.

bfournie: So I guess I don't understand the questions. This fix brings the use of the config-image to parity with the normal installation flow. The config-image was missing the user that gets created when doing an installation and this fixes it. 

Yes this allows console login with the `kubeadmin-password` during bootstrap only, which can very useful if the hosts are not booting properly and network errors prevent ssh login.  Is there a reason why we'd want to disable this?

bfournie: I talked with Daniel at the same time and this was his response. Still thinking about the best way to handle this...

> So the issue with that change is that we won’t be able to override the core user pass. I mean, in the appliance we have an option to specify a custom ‘userCorePass’: https://github.com/openshift/appliance/blob/master/docs/user-guide.md#set-appliance-config
> So this is useful as some partners want to connect to console using some easy predefined password. Especially for the bootstrap step - to test the machine, etc. Not sure it’s crucial for after rebooting the machine to the ocp installation step.
> Any way, it would be great if you could just add some condition before setting the password. I.e. an env var, a file or any indication. Just so we could override it for the appliance flow.

data/data/agent/files/usr/local/bin/load-config-iso.sh
```diff
@@ -64,6 +65,11 @@ copy_archive_contents() {
 
     echo "Successfully copied contents of ${AGENT_CONFIG_ARCHIVE_FILE} on ${devname}"
 
+    # Update core password with one from config-image if not overridden by appliance
+    if [[ -z "${SKIP_CORE_PASSWORD_SET}" ]] || [[ "${SKIP_CORE_PASSWORD_SET}" == "false" ]]; then
```
danielerez, line 69: We also need a way to override it in load-config-iso@.service.template

E.g.
```
EnvironmentFile=/etc/assisted/load-config-iso.env
```

pkg/asset/agent/configimage/configiso.go
```diff
@@ -65,6 +65,14 @@ func (a *ConfigImage) Generate(dependencies asset.Parents) error {
 		}
 	}
 
+	// Add additional file to allow for override of setting the core password
+	if _, ok := os.LookupEnv("OPENSHIFT_APPLIANCE_OVERRIDE_PASSWORD_SET"); ok {
```
danielerez, line 69: This solution is problematic... IIUC, it means that we'll have to instruct all appliance users to manually set this env var - not a great user experience.
Instead, can you just add the file to [load-config-iso@.service.template](https://github.com/openshift/installer/blob/8cd8f79a20b326fb4bcb25f65580323b638ee326/data/data/agent/systemd/units/load-config-iso@.service.template#L6-L7)?
I.e.:
```
EnvironmentFile=/etc/assisted/appliance-override-password-set
```
Then, the appliance will just ensure to create this file on bootstrap ignition.

bfournie: OK, changed to just check the file directly that could be added in appliance flow

zaneb: Yeah, we don't use environment variables as an API to control behaviour of the installer.
Looking for a file in the image creates an API between the appliance builder and the installer that we have to keep stable, but this is still preferable.

PR #7345, opened 2023-07-19
OCPBUGS-16207: ic: aws: validate max security groups 

r4f4:
According to https://docs.aws.amazon.com/vpc/latest/userguide/amazon-vpc-limits.html, the default Security groups number per network interface is 5 and could be 16 at most, so we better have some pre-check on the number of provided custom security groups.

When it's more than 15 (since the maximum is 16, but installer will also create one (`${var.cluster_id}-master-sg/${var.cluster_id}-worker-sg`), installer should quit and warn user about this.

This PR also introduces a unit test for the validation. For now we only test that the validation fails when the number of
security groups exceeds the maximum. We should add more tests in the future when the AWS SDK client calls the validation uses are mockable.

### DIFF
--- a/pkg/types/aws/validation/machinepool.go
+++ b/pkg/types/aws/validation/machinepool.go
@@ -29,6 +29,9 @@ var (
 	validMetadataAuthValues = sets.NewString("Required", "Optional")
 )
 
+// https://docs.aws.amazon.com/vpc/latest/userguide/amazon-vpc-limits.html
+const maxSecurityGroupsCount = 16
+
 // ValidateMachinePool checks that the specified machine pool is valid.
 func ValidateMachinePool(platform *aws.Platform, p *aws.MachinePool, fldPath *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
@@ -58,6 +61,12 @@ func validateSecurityGroups(platform *aws.Platform, p *aws.MachinePool, fldPath 
 	if len(p.AdditionalSecurityGroupIDs) > 0 && len(platform.Subnets) == 0 {
 		allErrs = append(allErrs, field.Required(fldPath.Child("platform.subnets"), "subnets must be provided when additional security groups are present"))
 	}
+
+	// The installer also creates a security group: `${var.cluster_id}-master-sg/${var.cluster_id}-worker-sg`
+	if count := len(p.AdditionalSecurityGroupIDs); count > maxSecurityGroupsCount-1 {
+		allErrs = append(allErrs, field.TooMany(fldPath, count, maxSecurityGroupsCount-1))
+	}
+
 	return allErrs
 }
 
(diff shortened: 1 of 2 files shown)


pkg/asset/installconfig/aws/validation.go
```diff
@@ -266,6 +269,12 @@ func validateMachinePool(ctx context.Context, meta *Metadata, fldPath *field.Pat
 func validateSecurityGroupIDs(ctx context.Context, meta *Metadata, fldPath *field.Path, platform *awstypes.Platform, pool *awstypes.MachinePool) field.ErrorList {
 	allErrs := field.ErrorList{}
 
+	// The installer also creates a security group: `${var.cluster_id}-master-sg/${var.cluster_id}-worker-sg`
+	if len(pool.AdditionalSecurityGroupIDs) > maxSecurityGroupsCount-1 {
+		errMsg := fmt.Sprintf("the number of additional Security Group IDs exceeds the maximum of %d", maxSecurityGroupsCount-1)
+		return append(allErrs, field.Invalid(fldPath, pool.AdditionalSecurityGroupIDs, errMsg))
```
patrickdillon, line 275: nit: this could be a good opportunity to use `field.TooMany`!

pkg/asset/installconfig/aws/validation.go
```diff
@@ -266,6 +269,12 @@ func validateMachinePool(ctx context.Context, meta *Metadata, fldPath *field.Pat
 func validateSecurityGroupIDs(ctx context.Context, meta *Metadata, fldPath *field.Path, platform *awstypes.Platform, pool *awstypes.MachinePool) field.ErrorList {
 	allErrs := field.ErrorList{}
 
+	// The installer also creates a security group: `${var.cluster_id}-master-sg/${var.cluster_id}-worker-sg`
+	if len(pool.AdditionalSecurityGroupIDs) > maxSecurityGroupsCount-1 {
```
patrickdillon, line 273: this validation is static (does not depend on api calls) so it could and probably should go in pkg types.

pkg/asset/installconfig/aws/validation_test.go
```diff
@@ -1005,3 +1005,45 @@ func TestSkipRecords(t *testing.T) {
...
+		name        string
+		edits       []poolEdit
+		expectedErr string
+	}{
+		{
+			name:        "fails when security groups exceed maximum",
+			edits:       []poolEdit{exceedMaxSGs},
+			expectedErr: `^.*: Invalid value: \[\]string\{(?:"sg\-\d+", ){15,}"sg\-\d+"\}: the number of additional Security Group IDs exceeds the maximum of 15$`,
+		},
+		// FIXME: it would be good to also add the cases when the function succeeds, but that involves mocking the ec2.DescribeSecurityGroupsWithContext call
```
patrickdillon, line 1029: another reason to move this validation to pkg types

patrickdillon (commented):
thanks for fixing this so quickly

pkg/asset/installconfig/aws/validation.go
r4f4: That makes so much more sense.

r4f4:
Update: moved validation to `pkg/types/aws/` instead of `installconfig`.

PR #7347, opened 2023-07-19
OCPBUGS-16292: GCP XPN: clarify service account support

patrickdillon:
Clarify the wording in the service account field in the gcp machine pool to indicate the field is only supported in the control plane machine set.

Because compute and control plane machine pools are the same type, openshift-install explain will display the service account field in the compute machine pool. Hopefully this will make it slightly less confusing.

### DIFF
--- a/data/data/install.openshift.io_installconfigs.yaml
+++ b/data/data/install.openshift.io_installconfigs.yaml
@@ -436,10 +436,11 @@ spec:
                           type: string
                         serviceAccount:
                           description: ServiceAccount is the email of a gcp service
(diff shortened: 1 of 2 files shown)


r4f4:
Don't forget to re-run the codegen.

/approve

rna-afk:
Do we wanna change this too? 
https://github.com/openshift/installer/blob/9dfc08bc02dcce80af0d2f3dc63c6d0216247591/pkg/asset/cluster/tfvars.go#L427

pawanpinjarkar (commented):
/lgtm

PR #7349, opened 2023-07-20
OCPBUGS-15941: ABI - Validate release image arch, add cpu_architectures to RELEASE_IMAGES

jeffdyoung:
Adding additional validation to check that the release payload architecture matches the cluster's architecture.
Also fixes OCPBUGS-27787 - single arch  ABI installs w/multi payload, adding `cpu_architectures` to RELEASE_IMAGES

RELEASE_IMAGES: 
Single arch -
```  
'[{"openshift_version":"4.15","cpu_architecture":"x86_64","cpu_architectures":["x86_64"],"url":"quay.io/openshift-release-dev/ocp-release:4.15.0-rc.4-x86_64","support_level":"beta","version":"4.15.0-rc.4"}]
```
Multi -
```
[{"openshift_version":"4.15-multi","cpu_architecture":"multi","cpu_architectures":["x86_64","arm64","ppc64le","s390x"],"url":"quay.io/openshift-release-dev/ocp-release:4.15.0-rc.4-multi","support_level":"beta","version":"4.15.0-rc.4-multi"}]'
```

### DIFF
--- a/pkg/asset/agent/common.go
+++ b/pkg/asset/agent/common.go
@@ -1,6 +1,8 @@
 package agent
 
 import (
+	"github.com/sirupsen/logrus"
+
 	hiveext "github.com/openshift/assisted-service/api/hiveextension/v1beta1"
 	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/baremetal"
@@ -55,4 +57,26 @@ func IsSupportedPlatform(platform hiveext.PlatformType) bool {
 		}
 	}
 	return false
+}
+
+// DetermineReleaseImageArch returns the arch of the release image.
+func DetermineReleaseImageArch(pullSecret, pullSpec string) (string, error) {
+	templateFilter := "-o=go-template={{if and .metadata.metadata (index . \"metadata\" \"metadata\" \"release.openshift.io/architecture\")}}{{index . \"metadata\" \"metadata\" \"release.openshift.io/architecture\"}}{{else}}{{.config.architecture}}{{end}}"
+
+	var getReleaseArch = []string{
+		"oc",
+		"adm",
+		"release",
+		"info",
+		pullSpec,
+		templateFilter,
+	}
+
+	releaseArch, err := ExecuteOC(pullSecret, getReleaseArch)
+	if err != nil {
+		logrus.Errorf("Release Image arch could not be found: %s", err)
+		return "", err
+	}
+	logrus.Debugf("Release Image arch is: %s", releaseArch)
+	return releaseArch, nil
 }
--- a/pkg/asset/agent/image/ignition.go
+++ b/pkg/asset/agent/image/ignition.go
@@ -21,6 +21,7 @@ import (
 	"github.com/openshift/assisted-service/api/v1beta1"
 	"github.com/openshift/assisted-service/models"
 	"github.com/openshift/installer/pkg/asset"
+	agentcommon "github.com/openshift/installer/pkg/asset/agent"
 	"github.com/openshift/installer/pkg/asset/agent/agentconfig"
 	"github.com/openshift/installer/pkg/asset/agent/manifests"
 	"github.com/openshift/installer/pkg/asset/agent/mirror"
@@ -135,8 +136,20 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
 	if infraEnv.Spec.CpuArchitecture != "" {
 		archName = infraEnv.Spec.CpuArchitecture
 	}
-
-	releaseImageList, err := releaseImageList(agentManifests.ClusterImageSet.Spec.ReleaseImage, archName)
+	// Examine the release payload to see if its multi
+	releaseArch, err := agentcommon.DetermineReleaseImageArch(agentManifests.GetPullSecretData(), agentManifests.ClusterImageSet.Spec.ReleaseImage)
+	if err != nil {
+		logrus.Warnf("Unable to validate the release image architecture, using infraEnv.Spec.CpuArchitecture for the release image arch")
+		releaseArch = archName
+	} else {
+		releaseArch = arch.RpmArch(releaseArch)
+		logrus.Debugf("Found Release Image Architecture: %s", releaseArch)
+	}
+	releaseArchs := []string{releaseArch}
+	if releaseArch == "multi" {
+		releaseArchs = []string{arch.RpmArch(types.ArchitectureARM64), arch.RpmArch(types.ArchitectureAMD64), arch.RpmArch(types.ArchitecturePPC64LE), arch.RpmArch(types.ArchitectureS390X)}
+	}
+	releaseImageList, err := releaseImageList(agentManifests.ClusterImageSet.Spec.ReleaseImage, releaseArch, releaseArchs)
 	if err != nil {
 		return err
 	}
--- a/pkg/asset/agent/image/releaseextract.go
+++ b/pkg/asset/agent/image/releaseextract.go
@@ -24,6 +24,7 @@ import (
 	"sigs.k8s.io/yaml"
 
 	operatorv1alpha1 "github.com/openshift/api/operator/v1alpha1"
+	"github.com/openshift/installer/pkg/asset/agent"
 	"github.com/openshift/installer/pkg/asset/agent/mirror"
 	"github.com/openshift/installer/pkg/rhcos"
 	"github.com/openshift/installer/pkg/rhcos/cache"
@@ -34,19 +35,19 @@ const (
 	coreOsFileName       = "/coreos/coreos-%s.iso"
 	coreOsSha256FileName = "/coreos/coreos-%s.iso.sha256"
 	coreOsStreamFileName = "/coreos/coreos-stream.json"
-	//OcDefaultTries is the number of times to execute the oc command on failues
+	// OcDefaultTries is the number of times to execute the oc command on failures.
 	OcDefaultTries = 5
-	// OcDefaultRetryDelay is the time between retries
+	// OcDefaultRetryDelay is the time between retries.
 	OcDefaultRetryDelay = time.Second * 5
 )
 
-// Config is used to set up the retries for extracting the base ISO
+// Config is used to set up the retries for extracting the base ISO.
 type Config struct {
 	MaxTries   uint
 	RetryDelay time.Duration
 }
 
-// Release is the interface to use the oc command to the get image info
+// Release is the interface to use the oc command to the get image info.
 type Release interface {
 	GetBaseIso(architecture string) (string, error)
 	GetBaseIsoVersion(architecture string) (string, error)
@@ -60,7 +61,7 @@ type release struct {
 	mirrorConfig []mirror.RegistriesConfig
 }
 
-// NewRelease is used to set up the executor to run oc commands
+// NewRelease is used to set up the executor to run oc commands.
 func NewRelease(config Config, releaseImage string, pullSecret string, mirrorConfig []mirror.RegistriesConfig) Release {
 	return &release{
 		config:       config,
@@ -69,13 +70,6 @@ func NewRelease(config Config, releaseImage string, pullSecret string, mirrorCon
 		mirrorConfig: mirrorConfig,
 	}
 }
-
-const (
-	templateGetImage             = "oc adm release info --image-for=%s --filter-by-os=linux/%s --insecure=%t %s"
-	templateGetImageWithIcsp     = "oc adm release info --image-for=%s --filter-by-os=linux/%s --insecure=%t --icsp-file=%s %s"
-	templateImageExtract         = "oc image extract --path %s:%s --filter-by-os=linux/%s --confirm %s"
-	templateImageExtractWithIcsp = "oc image extract --path %s:%s --filter-by-os=linux/%s --confirm --icsp-file=%s %s"
-)
 
 // ExtractFile extracts the specified file from the given image name, and store it in the cache dir.
 func (r *release) ExtractFile(image string, filename string, architecture string) ([]string, error) {
@@ -171,7 +165,6 @@ func (r *release) GetBaseIsoVersion(architecture string) (string, error) {
 func (r *release) getImageFromRelease(imageName string, architecture string) (string, error) {
 	// This requires the 'oc' command so make sure its available
 	_, err := exec.LookPath("oc")
-	var cmd string
 	if err != nil {
 		if len(r.mirrorConfig) > 0 {
 			logrus.Warning("Unable to validate mirror config because \"oc\" command is not available")
@@ -182,7 +175,19 @@ func (r *release) getImageFromRelease(imageName string, architecture string) (st
 	}
 
 	archName := arch.GoArch(architecture)
+	imagefor := "--image-for=" + imageName
+	filterbyos := "--filter-by-os=linux/" + archName
+	insecure := "--insecure=true"
 
+	var cmd = []string{
+		"oc",
+		"adm",
+		"release",
+		"info",
+		imagefor,
+		filterbyos,
+		insecure,
+	}
 	if len(r.mirrorConfig) > 0 {
 		logrus.Debugf("Using mirror configuration")
 		icspFile, err := getIcspFileFromRegistriesConfig(r.mirrorConfig)
@@ -190,13 +195,12 @@ func (r *release) getImageFromRelease(imageName string, architecture string) (st
 			return "", err
 		}
 		defer removeIcspFile(icspFile)
-		cmd = fmt.Sprintf(templateGetImageWithIcsp, imageName, archName, true, icspFile, r.releaseImage)
-	} else {
-		cmd = fmt.Sprintf(templateGetImage, imageName, archName, true, r.releaseImage)
+		icspfile := "--icsp-file=" + icspFile
+		cmd = append(cmd, icspfile)
 	}
-
+	cmd = append(cmd, r.releaseImage)
 	logrus.Debugf("Fetching image from OCP release (%s)", cmd)
-	image, err := execute(r.pullSecret, cmd)
+	image, err := agent.ExecuteOC(r.pullSecret, cmd)
 	if err != nil {
 		if strings.Contains(err.Error(), "unknown flag: --icsp-file") {
 			logrus.Warning("Using older version of \"oc\" that does not support mirroring")
@@ -208,26 +212,36 @@ func (r *release) getImageFromRelease(imageName string, architecture string) (st
 }
 
 func (r *release) extractFileFromImage(image, file, cacheDir string, architecture string) ([]string, error) {
-	var cmd string
 	archName := arch.GoArch(architecture)
+	extractpath := "--path=" + file + ":" + cacheDir
+	filterbyos := "--filter-by-os=linux/" + archName
+
+	var cmd = []string{
+		"oc",
+		"image",
+		"extract",
+		extractpath,
+		filterbyos,
+		"--confirm",
+	}
+
 	if len(r.mirrorConfig) > 0 {
 		icspFile, err := getIcspFileFromRegistriesConfig(r.mirrorConfig)
 		if err != nil {
 			return nil, err
 		}
 		defer removeIcspFile(icspFile)
-		cmd = fmt.Sprintf(templateImageExtractWithIcsp, file, cacheDir, archName, icspFile, image)
-	} else {
-		cmd = fmt.Sprintf(templateImageExtract, file, cacheDir, archName, image)
+		icspfile := "--icsp-file=" + icspFile
+		cmd = append(cmd, icspfile)
 	}
 	path := filepath.Join(cacheDir, path.Base(file))
 	// Remove file if it exists
 	if err := removeCacheFile(path); err != nil {
 		return nil, err
 	}
-
+	cmd = append(cmd, image)
 	logrus.Debugf("extracting %s to %s, %s", file, cacheDir, cmd)
-	_, err := retry.Do(r.config.MaxTries, r.config.RetryDelay, execute, r.pullSecret, cmd)
+	_, err := retry.Do(r.config.MaxTries, r.config.RetryDelay, agent.ExecuteOC, r.pullSecret, cmd)
 	if err != nil {
 		return nil, err
 	}
@@ -244,7 +258,7 @@ func (r *release) extractFileFromImage(image, file, cacheDir string, architectur
 	return matches, nil
 }
 
-// Get hash from rhcos.json
+// Get hash from rhcos.json.
 func getHashFromInstaller(architecture string) (bool, string) {
 	// Get hash from metadata in the installer
 	ctx, cancel := context.WithTimeout(context.TODO(), 30*time.Second)
@@ -277,7 +291,7 @@ func matchingHash(imageSha []byte, sha string) bool {
 	return false
 }
 
-// Check if there is a different base ISO in the release payload
+// Check if there is a different base ISO in the release payload.
 func (r *release) verifyCacheFile(image, file, architecture string) (bool, error) {
 	// Get hash of cached file
 	f, err := os.Open(file)
@@ -343,44 +357,7 @@ func removeCacheFile(path string) error {
 	return nil
 }
 
-func execute(pullSecret, command string) (string, error) {
-	ps, err := os.CreateTemp("", "registry-config")
-	if err != nil {
-		return "", err
-	}
-	defer func() {
-		ps.Close()
-		os.Remove(ps.Name())
-	}()
-	_, err = ps.Write([]byte(pullSecret))
-	if err != nil {
-		return "", err
-	}
-	// flush the buffer to ensure the file can be read
-	ps.Close()
-	executeCommand := command[:] + " --registry-config=" + ps.Name()
-	args := strings.Split(executeCommand, " ")
-
-	var stdoutBytes, stderrBytes bytes.Buffer
-	cmd := exec.Command(args[0], args[1:]...) //nolint:gosec
-	cmd.Stdout = &stdoutBytes
-	cmd.Stderr = &stderrBytes
-
-	err = cmd.Run()
-	if err == nil {
-		return strings.TrimSpace(stdoutBytes.String()), nil
-	}
-
-	var exitErr *exec.ExitError
-	if errors.As(err, &exitErr) {
-		err = fmt.Errorf("command '%s' exited with non-zero exit code %d: %s\n%s", executeCommand, exitErr.ExitCode(), stdoutBytes.String(), stderrBytes.String())
-	} else {
-		err = fmt.Errorf("command '%s' failed: %w", executeCommand, err)
-	}
-	return "", err
-}
-
-// Create a temporary file containing the ImageContentPolicySources
+// Create a temporary file containing the ImageContentPolicySources.
 func getIcspFileFromRegistriesConfig(mirrorConfig []mirror.RegistriesConfig) (string, error) {
 	contents, err := getIcspContents(mirrorConfig)
 	if err != nil {
@@ -406,7 +383,7 @@ func getIcspFileFromRegistriesConfig(mirrorConfig []mirror.RegistriesConfig) (st
 	return icspFile.Name(), nil
 }
 
-// Convert the data in registries.conf into ICSP format
+// Convert the data in registries.conf into ICSP format.
 func getIcspContents(mirrorConfig []mirror.RegistriesConfig) ([]byte, error) {
 	icsp := operatorv1alpha1.ImageContentSourcePolicy{
 		TypeMeta: metav1.TypeMeta{
--- a/pkg/asset/agent/image/releaseimage.go
+++ b/pkg/asset/agent/image/releaseimage.go
@@ -10,18 +10,18 @@ import (
 )
 
 type releaseImage struct {
-	ReleaseVersion string `json:"openshift_version"`
-	Arch           string `json:"cpu_architecture"`
-	PullSpec       string `json:"url"`
-	Tag            string `json:"version"`
+	ReleaseVersion string   `json:"openshift_version"`
+	Arch           string   `json:"cpu_architecture"`
+	Archs          []string `json:"cpu_architectures"`
+	PullSpec       string   `json:"url"`
+	Tag            string   `json:"version"`
 }
 
 func isDigest(pullspec string) bool {
 	return regexp.MustCompile(`.*sha256:[a-fA-F0-9]{64}$`).MatchString(pullspec)
 }
 
-func releaseImageFromPullSpec(pullSpec, arch string) (releaseImage, error) {
-
+func releaseImageFromPullSpec(pullSpec, arch string, archs []string) (releaseImage, error) {
 	// When the pullspec it's a digest let's use the current version
 	// stored in the installer
 	if isDigest(pullSpec) {
@@ -33,6 +33,7 @@ func releaseImageFromPullSpec(pullSpec, arch string) (releaseImage, error) {
 		return releaseImage{
 			ReleaseVersion: versionString,
 			Arch:           arch,
+			Archs:          archs,
 			PullSpec:       pullSpec,
 			Tag:            versionString,
 		}, nil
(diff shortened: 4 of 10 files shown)


jeffdyoung:
This will also undo the work done in https://github.com/openshift/installer/pull/7276
because this test is creating an agent.aarch64.iso with the default x86 payload

hack/go-lint.sh
```diff
@@ -6,4 +6,5 @@ podman run --rm \
     --volume "${PWD}:/go/src/github.com/openshift/installer:z" \
     --workdir /go/src/github.com/openshift/installer \
     docker.io/golangci/golangci-lint:v1.48.0 \
-    golangci-lint run "${@}"
+    golangci-lint run "${@}" --new-from-rev=75173a17cf
+    #golangci-lint run "${@}" -c .golangci.yaml
```
r4f4, line 10: We don't want to add this here I think. You can run `./hack/go-lint.sh --new-from-rev=<commit>`. 
Happy to discuss in a new PR if you want. The idea was that devs might want to slowly work on the existing linting issues but for CI there was no way to solve all of them at once to make the test pass.

jeffdyoung: 🤦 I did not mean to commit this. Good catch @r4f4...removing..

cmd/openshift-install/testdata/agent/image/validations/sno_invalid_arch.txt
```diff
@@ -1,28 +1,22 @@
-# Verify a default configuration for the SNO topology for ARM architecture
+! exec openshift-install agent create image --dir $WORK
```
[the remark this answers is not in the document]
jeffdyoung: Hey @pawanpinjarkar the issue is the original test case is using a x86 release payload, and you can't create a valid agent.aarch64.iso from that. Since this PR checks to see if the arch of the release payload matches the target arch, the old test case will no longer pass. Unfortunately, there isn't an upstream ARM64 release payload available in CI.

pkg/asset/agent/image/oc.go
```diff
@@ -169,6 +177,62 @@ func (r *release) getImageFromRelease(imageName string) (string, error) {
...
+		err := json.Unmarshal([]byte(image), &releaseimagedata)
+		if err != nil {
+			logrus.Errorf("can't read oc adm release info for %s ", r.releaseImage)
+			return err
+		}
+
+		// Check the release image to see if it is multi.
+		if releaseimagedata.Metadata.Metadata != nil {
+			if releaseimagedata.Metadata.Metadata.ReleaseArch == "multi" {
+				err = fmt.Errorf("multi release image detected: %s.... agent installs with this release payload is not supported", r.releaseImage)
```
pawanpinjarkar, line 216: nit: `err = fmt.Errorf("multi architecture release image %s is not supported", r.releaseImage)`

pkg/asset/agent/image/oc.go
```diff
@@ -169,6 +177,62 @@ func (r *release) getImageFromRelease(imageName string) (string, error) {
...
+				return err
+			}
+		} else {
+			// if the release image isn't multi, then its single arch, and it must match the cpu architecture.
+			switch releaseimagedata.Config.Architecture {
+			case getarch.GoArch(architecture):
+				logrus.Debugf("The release image: %s supports this architecture: %s", r.releaseImage, architecture)
+			default:
+				err = fmt.Errorf("release image: %s doesn't support this architecture: %s detected this architecture: %s", r.releaseImage, architecture, releaseimagedata.Config.Architecture)
```
pawanpinjarkar, line 225: nit: `err = fmt.Errorf("Unsupported release image architecture. Expected: %s Found:%s ", r.releaseImage, architecture, releaseimagedata.Config.Architecture)`

pawanpinjarkar (commented):
Minor changes

pkg/asset/agent/image/oc.go
```diff
@@ -146,6 +148,11 @@ func (r *release) getImageFromRelease(imageName string) (string, error) {
 		return "", err
 	}
 
+	err = r.validReleaseImageArch(architecture)
```
pawanpinjarkar, line 151: maybe `validateReleaseImageArch`

pkg/asset/agent/image/oc.go
```diff
@@ -169,6 +176,62 @@ func (r *release) getImageFromRelease(imageName string) (string, error) {
...
+			return err
+		}
+
+		// Check the release image to see if it is multi.
+		if releaseimagedata.Metadata.Metadata != nil {
+			if releaseimagedata.Metadata.Metadata.ReleaseArch == "multi" {
+				err = fmt.Errorf("multi architecture release image %s is not supported", r.releaseImage)
+				return err
+			}
+		} else {
```
zaneb, line 218: Metadata is always nil except for multi-arch? Or should this `else` go with the second `if`?

pkg/asset/agent/image/oc.go
```diff
@@ -169,6 +176,62 @@ func (r *release) getImageFromRelease(imageName string) (string, error) {
...
+		err := json.Unmarshal([]byte(image), &releaseimagedata)
+		if err != nil {
+			logrus.Errorf("can't read oc adm release info for %s ", r.releaseImage)
+			return err
+		}
+
+		// Check the release image to see if it is multi.
+		if releaseimagedata.Metadata.Metadata != nil {
+			if releaseimagedata.Metadata.Metadata.ReleaseArch == "multi" {
+				err = fmt.Errorf("multi architecture release image %s is not supported", r.releaseImage)
```
zaneb, line 215: This is just because we know we need to make more changes to make this work, I assume? A comment to that effect might be helpful for future-us.

Specifically, IIRC, we have to make changes to make it work when the cluster cpu arch is different from the host the installer is running on. Potentially we could limit the error to just that case.

pkg/asset/agent/image/oc.go
```diff
@@ -169,6 +176,62 @@ func (r *release) getImageFromRelease(imageName string) (string, error) {
...
+				return err
+			}
+		} else {
+			// if the release image isn't multi, then its single arch, and it must match the cpu architecture.
+			switch releaseimagedata.Config.Architecture {
```
zaneb, line 220: nit: this is really an `if` statement disguised as a `switch` :slightly_smiling_face:

pkg/asset/agent/image/oc.go
```diff
@@ -169,6 +176,62 @@ func (r *release) getImageFromRelease(imageName string) (string, error) {
 
 	return image, nil
 }
+func (r *release) validateReleaseImageArch(architecture string) error {
```
andfasano, line 179: IIUC correctly the location of this validation doesn't seem ideal here, for different reasons:
* It could be invoked multiple times (since the `getImageFromRelease` is also part of the `ExtractFile`)
* The validation essentially checks that embedded installer pullspec arch (here retrieved via ClusterImageSet.Spec.ReleaseImage) with the one configured (if defined) in the install-config (here retrieved via InfraEnv.Spec.CpuArchitecture). The goal is thus to ensure that the current installer binary will not be used to generate an agent ISO of a different arch. So this validation, as long as many other validations, should be performed just once in a proper asset (for example in the `OptionalInstallConfig` one)

cmd/openshift-install/testdata/agent/image/validations/sno_invalid_arch.txt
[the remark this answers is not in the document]
andfasano: This kind of problem seems to become more pressing, and removing tests doesn't seem a good strategy in the long term. It could be an acceptable approach for the current PR, as well as if we could start discussing (in a separate thread) on a possible way to support it somehow.

pkg/asset/agent/image/oc.go
jeffdyoung: > This is just because we know we need to make more changes to make this work, I assume? A comment to that effect might be helpful for future-us.

Correct, we will tackle using the `multi` payload in 4.15. The changes are mostly adding '--filter-by-os' to the oc adm release command.

> Specifically, IIRC, we have to make changes to make it work when the cluster cpu arch is different from the host the installer is running on. Potentially we could limit the error to just that case.

This is true for BMIPI, but not agent. An x86 client can create a valid agent.aarch.iso as long as the release payload is arm64 (or in the future multi).

pkg/asset/agent/image/oc.go
jeffdyoung: > IIUC correctly the location of this validation doesn't seem ideal here, for different reasons:
> 
> * It could be invoked multiple times (since the `getImageFromRelease` is also part of the `ExtractFile`)
> * The validation essentially checks that embedded installer pullspec arch (here retrieved via ClusterImageSet.Spec.ReleaseImage) with the one configured (if defined) in the install-config (here retrieved via InfraEnv.Spec.CpuArchitecture). The goal is thus to ensure that the current installer binary will not be used to generate an agent ISO of a different arch. 

I think we're on the same page, but just to make sure. The binary of the current installer doesn't matter, the releaseimage architecture of the current installer does.
We can create an agent.aarch64.iso  from an amd64 host as long as the openshift-install binary uses the correct release payload:
https://mirror.openshift.com/pub/openshift-v4/aarch64/clients/ocp/4.13.4/openshift-install-linux-amd64-4.13.4.tar.gz

```
[jeyoung]$ uname -m 
x86_64
[jeyoung]$ ./openshift-install version
./openshift-install 4.13.4
built from commit 90acb3fa2990c35c9beeff4a188fb133fedba432
release image registry.ci.openshift.org/ocp-arm64/release-arm64@sha256:13b14f0514d24d241d40ebacac9f15f93acebc4a7849e4740df49e65e48af424
release architecture arm64
```

> So this validation, as long as many other validations, should be performed just once in a proper asset (for example in the `OptionalInstallConfig` one)

Fair, was hoping to keep in close to the other oc templates. Will reevaluate.

cmd/openshift-install/testdata/agent/image/validations/sno_invalid_arch.txt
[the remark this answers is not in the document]
jeffdyoung: @andfasano agree, happy to discuss offline. Unfortunately, there aren't a lot of options for multi-arch CI testing. We should be able to do validation in nightly's (if it isn't there already)

pkg/asset/agent/image/oc.go
andfasano: > I think we're on the same page, but just to make sure. The binary of the current installer doesn't matter, the releaseimage architecture of the current installer does.

Agree, it's the embedded pullspec that matters

pkg/asset/agent/image/oc.go
zaneb: right, I mean that's the case where `--filter-by-os` is needed. if the arch happens to match I think it will pull the right image even without `--filter-by-os`?

jeffdyoung: correct, but then the story is you can only create an arm cluster with an x86 client IF you use a single arch arm payload. If using the multi payload the client machine has to match the architecture of the target nodes. At this time, we'd rather say multi isn't supported at install time, and then bring the multi payload to 'parity'.

pkg/asset/agent/image/oc.go
jeffdyoung: @andfasano I've been working on moving the validation to assets. I can get unit/integration tests to pass on my local machine, but I'm worried it will fail in CI b/c `oc` won't be available?

pkg/asset/agent/image/oc.go
jeffdyoung: Yeap, this is bad logic. fixed.

pkg/asset/agent/manifests/releaseimagearch.go
```diff
@@ -0,0 +1,156 @@
+package manifests
```
jeffdyoung, line 1: @andfasano I think this is what you were going for, can I get a sanity check before I continue?

pkg/asset/agent/image/oc.go
andfasano: Currently the [integration tests in the installer repo](https://github.com/openshift/release/blob/6b9fd78a454b0e9408e1cf783b22fd8bec4009e0/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L688-L692) are run within a specially crafted image `src-oc` that includes the `oc` cli

pkg/asset/agent/manifests/releaseimagearch.go
```diff
@@ -0,0 +1,156 @@
...
+			return err
+		}
+	} else {
+		logrus.Error("Unable to validate the release image architecture")
+		return err
+	}
+
+	return nil
+}
+func execute(executer executer.Executer, pullSecret, command string) (string, error) {
```
andfasano, line 113: In general it would be recommend to avoid duplicating the code. https://github.com/openshift/installer/blob/e20821e3b213dbf6af19c2739dbab7c0c1e3c765/pkg/asset/agent/image/oc.go could be refactored to allow re-using it directly.

pkg/asset/agent/manifests/releaseimagearch.go
```diff
@@ -0,0 +1,156 @@
...
+	pullSpec            string
+	pullSecret          string
+	controlPlaneCPUArch string
+	executer            executer.Executer
+}
+
+var _ asset.WritableAsset = (*ReleaseImageArch)(nil)
+
+const (
+	templateGetReleaseImageArch = "oc adm release info %s -o=go-template='{{if and .metadata.metadata (index . \"metadata\" \"metadata\" \"release.openshift.io/architecture\")}}{{index . \"metadata\" \"metadata\" \"release.openshift.io/architecture\"}}{{else}}{{.config.architecture}}{{end}}'"
```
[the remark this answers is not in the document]
jeffdyoung: `config.architecture` will only have `amd64, arm64, ppc64le, s390x` as values (true for multi and single arch)

 `.metadata.metadata.release.openshift.io/architecture` will only be set `multi` (single arch payload don't have this field) when the release payload is manifest listed.

This is why my thinking is:
Look at `.metadata.metadata.release.openshift.io/architecture` to see if it is `multi`, if this isn't true, then this release must be a single arch payload and `config.architecture` will match the release architecture.

pkg/asset/agent/manifests/releaseimagearch.go
jeffdyoung: agree.. this was the main reason for putting this PR to WIP.. trying to get a better sense of how this should look :)

pkg/asset/agent/common.go
```diff
@@ -52,3 +58,48 @@ func IsSupportedPlatform(platform hiveext.PlatformType) bool {
 	}
 	return false
 }
+
+// OCexecute will execute an oc command.
+func OCexecute(executer executer.Executer, pullSecret, command string) (string, error) {
```
jeffdyoung, line 63: @andfasano When I tried to use the existing `Execute`  in oc.go I would get import cycle issues when trying to refernece from installconfig.go  
```
package github.com/openshift/installer/cmd/openshift-install
	imports github.com/openshift/installer/cmd/openshift-install/agent
	imports github.com/openshift/installer/pkg/agent
	imports github.com/openshift/installer/pkg/asset/agent/image
	imports github.com/openshift/installer/pkg/asset/agent
	imports github.com/openshift/installer/pkg/asset/agent/image: import cycle not allowed
```
Moving this function to pkg/asset/agent/common.go seems to be the best compromise? Please advise.

pkg/asset/agent/image/oc.go
jeffdyoung: yeah, integration tests are ok. I do see issues when running `hack/go-test.sh` locally because [ docker.io/golang:1.20](https://github.com/openshift/installer/blob/master/hack/go-test.sh#L11 ) doesn't have oc. 

@tvardema  @SaravanaStorageNetwork  Do you all have a suggestion on a go container w/oc that we could use here?

jeffdyoung: Yeap, these tests fail because they use the [ci-operator:latest](https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/installer/openshift-installer-master-presubmits.yaml#L26) container, which doesn't have `oc`

```                    
ci/prow/aro-unit                    
ci/prow/okd-scos-unit                   
ci/prow/okd-unit                    
ci/prow/unit
```
Basically this PR inadvertently adds a `check to make sure oc is in $PATH` for all agent validation.
Here are some options I see to move fwd:

1. Insert `oc` into `ci-operator:latest`, or find another container that has `oc`. Also find a suitable replacement for [docker.io/golang:1.20](https://github.com/openshift/installer/blob/master/hack/go-test.sh#L11) for local testing. 
2. Only throw a warning if `oc` isn't present during validation (basically skip release arch validation instead of erroring out). 
3. No fix, this kind of validation isn't worth the effort to implement properly. Seems like I'm doing more harm than good here? 
4. Vendor `oc`. 

Thoughts? @zaneb @andfasano @pawanpinjarkar 

FWIW, my preferences in this order are:
1 (only if the needed containers already exist), 2, and lastly 3. 
Not a fan of 4 just for this, but maybe there are other reasons you'd want to vendor that I don't know about?

Thanks again for being patience with this.

zaneb: I vote for 2.
Not opposed to doing 1 as well.
4 unfortunately comes with horrendous dependencies, otherwise we'd do it: https://issues.redhat.com/browse/AGENT-307

andfasano: As long as 2 could be a valid option, there is also the alternative to adopt the `src-oc` image as we already did for the [integration tests](https://github.com/openshift/release/blob/e7613ad8671c48cb9a8c5984fa6406c33da3d1e4/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L689-L692), an [image](https://github.com/openshift/release/blob/e7613ad8671c48cb9a8c5984fa6406c33da3d1e4/ci-operator/config/openshift/installer/openshift-installer-master.yaml#L108-L121) crafted just for including `oc`

jeffdyoung: @andfasano what do you mean by `As long as 2 could be a valid option`? Are you asking for something like `openshift-install agent create image --validate-release-image`

andfasano: I mean, if it's just a test problem then extending the usage of `src-oc` image also to the tests that are not currently using it (but just `src`) should  make them happy. Option 2 seems still a feasible alternative (IIUC, just impacting the test code?)

pkg/asset/agent/manifests/releaseimagearch.go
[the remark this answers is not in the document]
andfasano: I'm not following, if it's `multi` then how this info is going to be used?

pkg/asset/agent/installconfig.go
```diff
@@ -110,6 +113,29 @@ func (a *OptionalInstallConfig) validateSupportedArchs(installConfig *types.Inst
...
+		logrus.Warnf(fmt.Sprintf("Unable to validate the release image architecture"))
+	} else {
+		releaseArch = strings.Trim(releaseArch, "'")
+		// Validate that the release image supports the install-config architectures.
+		switch releaseArch {
+		// Check the release image to see if it is multi, if so fail. We don't support multi release images at install time (yet).
+		case "multi":
+			allErrs = append(allErrs, field.Forbidden(fieldPath, fmt.Sprintf("multi architecture release image %s is not supported", releaseImage.PullSpec)))
+		// if the release image isn't multi, then its single arch, and it must match the cpu architecture.
+		case string(installConfig.ControlPlane.Architecture):
```
jeffdyoung, line 133: @andfasano as of now.. 'multi' isn't supported for agent installer. If this release image is detected, then we fail. Enabling the use of multi will require additional work in openshift-install and assisted-service (something we've targeted for 4.15)

pkg/asset/agent/image/oc.go
jeffdyoung: yeah, its breaking ci tests and local test https://github.com/openshift/installer/blob/master/hack/go-test.sh#L11

pkg/asset/agent/common.go
andfasano: Sorry somehow I missed this one, let me check it better

andfasano: I think a reasonable approach could be the following:
* The `OCexecute` function could be moved to a new  `plg/asset/agent/oc.go` file. It depends only on system packages, and it will provide the pure functionality of the `oc` command, in its most generic form.
* The old `pkg/asset/agent/image/oc.go` must be renamed to something else (I'd suggest `releaseextract.go`) since it's currently offering just two funcs to extract stuff from a release: a generic `ExtractFile` and a more specialized `GetBaseIso`

pkg/asset/agent/common.go
```diff
@@ -56,3 +64,57 @@ func IsSupportedPlatform(platform hiveext.PlatformType) bool {
 	}
 	return false
 }
+
+// OCexecute will execute an oc command.
+func OCexecute(pullSecret, command string) (string, error) {
```
andfasano, line 69: nit: given that naming is usually difficult, my preference in this case would be for something like `RunOC` or `ExecuteOC`

aleskandro (commented):
I'll go testing this before merging... In the meantime, two comments, an 'architectural' one about the concern on the `oc` binary dependency for this particular case, and a minor one about the `executeOC` function.

pkg/asset/agent/installconfig.go
```diff
@@ -110,6 +113,29 @@ func (a *OptionalInstallConfig) validateSupportedArchs(installConfig *types.Inst
 	var allErrs field.ErrorList
 
 	fieldPath := field.NewPath("ControlPlane", "Architecture")
+	releaseImage := &releaseimage.Image{}
+	asseterr := releaseImage.Generate(asset.Parents{})
+	if asseterr != nil {
+		allErrs = append(allErrs, field.InternalError(fieldPath, asseterr))
+	}
+	cmd := fmt.Sprintf(templateGetReleaseImageArch, releaseImage.PullSpec)
+	releaseArch, err := ExecuteOC(installConfig.PullSecret, cmd)
```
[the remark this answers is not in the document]
andfasano: `oc` is used also for other tasks (ie image extraction), I think that to remove it completely it will not be straightforward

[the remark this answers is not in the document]
aleskandro: @andfasano yes I see... I'm talking of this validation... We can avoid relying on `oc` for that at least. The extraction I agree is not straightforward as the above snippet.

pkg/asset/agent/oc.go
```diff
@@ -0,0 +1,65 @@
...
+		case lastQuote != rune(0):
+			return false
+		case unicode.In(c, unicode.Quotation_Mark):
+			lastQuote = c
+			return false
+		default:
+			return unicode.IsSpace(c)
+		}
+	}
+	args := strings.FieldsFunc(executeCommand, f)
```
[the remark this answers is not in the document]
jeffdyoung: I agree that this can be better, but I think refactoring how OC executes commands is out of the scope of this PR.

pkg/asset/agent/image/oc.go
jeffdyoung: going with option 2, seems unreasonable to force all local testing for the installer to require `oc`?

pkg/asset/agent/installconfig.go
```diff
@@ -114,6 +120,36 @@ func (a *OptionalInstallConfig) validateSupportedPlatforms(installConfig *types.
 	return allErrs
 }
 
+func (a *OptionalInstallConfig) validateReleaseArch(installConfig *types.InstallConfig) field.ErrorList {
```
jeffdyoung, line 123: @andfasano I'm not sure if this is exactly what you meant by this [comment](https://github.com/openshift/installer/pull/7349#discussion_r1344319648)? Once I took a 2nd look, it made sense to move all of the releaseimage arch logic to its own validation.

pkg/asset/agent/installconfig.go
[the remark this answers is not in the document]
jeffdyoung: IMO, its much cleaner to use `oc`, but I'm good with changing to this if you prefer it @zaneb @andfasano 

This makes me wish we had a `multi-arch` helper library, so we could share this with [assisted-service](https://github.com/openshift/assisted-service/blob/master/internal/oc/release.go#L183).. 🤔

pkg/asset/agent/oc.go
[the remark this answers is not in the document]
zaneb: I would argue that rolling your own shell parsing is even _more_ out of scope than fixing the API to a function that you're already splitting out to a common place :slightly_smiling_face:

pkg/asset/agent/installconfig.go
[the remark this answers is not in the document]
zaneb: My 2c: if we can do this validation without depending on oc that's good, it means we'll always be able to run it and not have to skip if oc is not installed.
It's also one less thing we'd have to do later if we do work on getting rid of oc altogether.
In this case there doesn't seem to be any code that is specific to the release image or OpenShift, so there's no great danger of it drifting out of sync with oc.
I do like the idea of a multi-arch library though :slightly_smiling_face:

jeffdyoung:
Going to wait for [this](https://github.com/openshift/installer/pull/7595) to land before continuing.

pkg/asset/agent/image/releaseimage.go
```diff
@@ -33,6 +33,7 @@ func releaseImageFromPullSpec(pullSpec, arch string) (releaseImage, error) {
 		return releaseImage{
 			ReleaseVersion: versionString,
 			Arch:           arch,
+			Archs:          archs,
```
andfasano, line 36: @jeffdyoung this point it's not clear to me. What's the requirement to have separate fields `Archs` and `Arch`? Would make sense to have just one field `Archs` to cover all the cases?

pkg/asset/agent/image/ignition.go
```diff
@@ -135,8 +136,20 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
...
+	if err != nil {
+		logrus.Warnf("Unable to validate the release image architecture, using infraEnv.Spec.CpuArchitecture for the release image arch")
+		releaseArch = archName
+	} else {
+		releaseArch = arch.RpmArch(releaseArch)
+		logrus.Debugf("Found Release Image Architecture: %s", releaseArch)
+	}
+	releaseArchs := []string{releaseArch}
+	if releaseArch == "multi" {
+		releaseArchs = []string{arch.RpmArch(types.ArchitectureARM64), arch.RpmArch(types.ArchitectureAMD64), types.ArchitecturePPC64LE, types.ArchitectureS390X}
```
zaneb, line 150: Can we use arch.RpmArch on all of these please. The fact that the go and rpm arches happen to be the same strings on some platforms shouldn't mean that we treat them as semantically equivalent.

pkg/asset/agent/common.go
```diff
@@ -56,3 +58,24 @@ func IsSupportedPlatform(platform hiveext.PlatformType) bool {
...
+
+	var getReleaseArch = []string{
+		"oc",
+		"adm",
+		"release",
+		"info",
+		pullSpec,
+		templateFilter,
+	}
+	releaseArch, err := ExecuteOC(pullSecret, getReleaseArch)
```
zaneb, line 74: This is ok for this PR, but particularly with the heterogeneous cluster support coming in I'd like to see a follow-up switching to direct docker client calls as suggested in https://github.com/openshift/installer/pull/7349#discussion_r1349870697.
I think validation of the release payload arch is going to be critical for heterogeneous clusters in particular.

pkg/asset/agent/oc.go
```diff
@@ -0,0 +1,50 @@
...
+		ps.Close()
+		os.Remove(ps.Name())
+	}()
+	_, err = ps.Write([]byte(pullSecret))
+	if err != nil {
+		return "", err
+	}
+	// flush the buffer to ensure the file can be read
+	ps.Close()
+	registryConfig := " --registry-config=" + ps.Name()
```
zaneb, line 29: nit
```suggestion
	registryConfig := "--registry-config=" + ps.Name()
```

pkg/asset/agent/oc.go
```diff
@@ -0,0 +1,50 @@
...
+	command = append(command, registryConfig)
+
+	var stdoutBytes, stderrBytes bytes.Buffer
+	// squash command into a string, then run in a shell. We do this to handle nested spaces in args
+	cmd := exec.Command("sh", "-c", strings.Join(command, " ")) //nolint:gosec
```
zaneb, line 34: gosec is right to complain here.
```suggestion
	cmd := exec.Command(command...)
```

pkg/asset/agent/image/ignition.go
```diff
@@ -135,8 +136,20 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
 	if infraEnv.Spec.CpuArchitecture != "" {
 		archName = infraEnv.Spec.CpuArchitecture
 	}
-
-	releaseImageList, err := releaseImageList(agentManifests.ClusterImageSet.Spec.ReleaseImage, archName)
+	// Examine the release payload to see if its multi
+	releaseArch, err := agentcommon.DetermineReleaseImageArch(agentManifests.GetPullSecretData(), agentManifests.ClusterImageSet.Spec.ReleaseImage)
+	if err != nil {
+		logrus.Warnf("Unable to validate the release image architecture, using infraEnv.Spec.CpuArchitecture for the release image arch")
+		releaseArch = archName
```
zaneb, line 143: Do you know what happens inside assisted-service if we tell it we're giving it a single-arch image, but the actual image spec is for a multi-arch image? Does it just handle that transparently?

Also, this fallback won't work correctly for heterogeneous clusters (which require the `multi` release payload)... I assume that's because we're not planning to support those yet?

zaneb (commented):
This would be easier to review (and faster to re-review) if the logically separate changes were in separate commits. I can see at least these 5 logical changes:

- Use more realistic pull secret in unit tests
- Refactor execute() function to not do arg splitting
- Move execute() function to a common location
- Support multi-arch payload in releaseImage data
- Validate release arch when possible

pkg/asset/agent/image/releaseimage.go
jeffdyoung: Looks like we'll need both for now:
```
level=fatal msg="failed to create Versions handlers" func=main.main.func1 file="/src/cmd/main.go:237" error="Missing value in ReleaseImage for 'cpu_architecture' field"
```

pkg/asset/agent/common.go
jeffdyoung: @zaneb I agree, that something* needs to be done to validate the release image(s) for all IPI/assisted/agent installs. Not sure how that suggestion will work with sparse manifest list? But yeah, want to figure this out for heterogeneous in future releases.

pkg/asset/agent/image/ignition.go
jeffdyoung: I tested this, and was transparent on both x86 and arm. WRT heterogeneous installs, that will be a completely different effort. This PR is to ensure that the multi payload behaves the same way as a single arch payload.

pkg/asset/agent/image/releaseextract.go
```diff
@@ -208,26 +212,36 @@ func (r *release) getImageFromRelease(imageName string, architecture string) (st
 }
 
 func (r *release) extractFileFromImage(image, file, cacheDir string, architecture string) ([]string, error) {
-	var cmd string
 	archName := arch.GoArch(architecture)
+	extractpath := "--path " + file + ":" + cacheDir
```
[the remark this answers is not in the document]
jeffdyoung: It worked, but only because of the way I was previously squashing args in oc.go... fixing...

pkg/asset/agent/image/releaseimage.go
jeffdyoung: This isn't an exhaustive list, but I see both used in various parts of the assisted-service:
https://github.com/openshift/assisted-service/blob/master/models/infra_env.go#L39
https://github.com/openshift/assisted-service/blob/master/models/infra_env_create_params.go#L40
https://github.com/openshift/assisted-service/blob/master/models/os_image.go#L26
https://github.com/openshift/assisted-service/blob/master/models/release_image.go#L27-L30
https://github.com/openshift/assisted-service/blob/master/models/openshift_version.go#L25

pkg/asset/agent/image/ignition.go
zaneb: ack

pkg/asset/agent/oc.go
```diff
@@ -0,0 +1,48 @@
...
+	_, err = ps.Write([]byte(pullSecret))
+	if err != nil {
+		return "", err
+	}
+	// flush the buffer to ensure the file can be read
+	ps.Close()
+	registryConfig := "--registry-config=" + ps.Name()
+	command = append(command, registryConfig)
+	var stdoutBytes, stderrBytes bytes.Buffer
+	cmd := exec.Command(command[0], command[1:]...)
```
zaneb, line 32: gosec is still complaining, but now for spurious reasons.

zaneb (commented):
/approve

pkg/asset/agent/oc.go
jeffdyoung: yeap.. added `// #nosec G204` should fix it 🤞

PR #7355, opened 2023-07-20
OCPBUGS-16219: Fix timing issue between network services

bfournie:
This fixes a timing issue between pre-network-manager-config.service and NetworkManager.service when defining static IPs in nmstateconfig. The problem is not apparent when using a few network interfaces, but when many interfaces (e.g. more than 8) are defined then the pre-network-manager-config.sh does not complete copying the files to /etc/NetworkManager/system-connections/ BEFORE NetworkManager.service starts running. The result is that the static IPs are not configured.

The fix is to add a dependency on NetworkManager.service to ensure that the pre-network-manager-config.service completes before NetworkManager.service is run.

### DIFF
--- a/data/data/agent/systemd/units/pre-network-manager-config.service
+++ b/data/data/agent/systemd/units/pre-network-manager-config.service
@@ -1,13 +1,11 @@
 [Unit]
 Description=Prepare network manager config content
-Before=dracut-initqueue.service
-After=dracut-cmdline.service
+Before=NetworkManager.service
 DefaultDependencies=no
 
 [Service]
 User=root
 ExecStart=/usr/local/bin/pre-network-manager-config.sh
-
 TimeoutSec=60
 KillMode=none
 Type=oneshot


bfournie:
Testing of the fix...

Before the dependency was added
=============================
Script running:
Jul 14 00:04:40 localhost pre-network-manager-config.sh[3117]: /usr/local/bin/pre-network-manager-config.sh: line 135: //: Is a directory
Jul 14 00:04:40 localhost pre-network-manager-config.sh[3041]: Removing default connection files in '/etc/NetworkManager/system-connections'

NetworkManager starting:
Jul 14 00:04:40 localhost NetworkManager[3182]: <info> [1689307480.9885] NetworkManager (version 1.42.2-1.el9) is starting... (boot:5431ea31-b86a-47e5-a109-9e102fe18613)

Script completing:
Jul 14 00:04:41 c3-esx04.rackg10.local systemd[1]: pre-network-manager-config.service: Deactivated successfully.

After the dependency was added (including an artificial 1 second delay added in script)
========================================================================
Script running:
Jul 20 19:57:22 localhost pre-network-manager-config.sh[1147]: /usr/local/bin/pre-network-manager-config.sh: line 135: //: Is a directory
Jul 20 19:57:22 localhost pre-network-manager-config.sh[1109]: Removing default connection files in '/etc/NetworkManager/system-connections'

Script completing:
Jul 20 19:57:23 master-0 systemd[1]: pre-network-manager-config.service: Deactivated successfully.

NetworkManager starting:
Jul 20 19:57:23 master-0 NetworkManager[1297]: <info>  [1689883043.0877] NetworkManager (version 1.42.2-3.el9_2) is starting... (boot:5b9825ea-8b99-4e9b-bc23-05562faef511)

rwsu (approved):
/approve

data/data/agent/systemd/units/pre-network-manager-config.service
```diff
@@ -1,6 +1,7 @@
 [Unit]
 Description=Prepare network manager config content
 Before=dracut-initqueue.service
```
zaneb, line 3: I'm confused by this service (which is copied verbatim [from assisted](https://github.com/openshift/assisted-service/blob/master/internal/ignition/templates/pre-network-manager-config.service)). The Before/After here suggest that we're expecting it to run during the initramfs. But below we have `WantedBy=multi-user.target`, not `initrd.target`.
Not that making it wanted by `initrd.target` would work (I don't think that Ignition lets you add services to run in the initramfs). I'm not sure that anything short of a dracut module is going to make it actually work in the initramfs.
As far as I can see, this is going to run _after_ the pivot, and the Before/After on the dracut services will be ignored because they're not accompanied by Wants/Requires and those services are not enabled in the multi-user target.

Looking back at the history, the NetworkManager ordering used to exist, but was removed and replaced with the dracut stuff: https://github.com/openshift/assisted-service/pull/1870
This was based on [this recommendation](https://bugzilla.redhat.com/show_bug.cgi?id=1959961#c30), which in turn was predicated on the [assertion](https://bugzilla.redhat.com/show_bug.cgi?id=1959961#c29) that this service was already running during the initramfs. The reason given for doing that (so that the network config can be used to load the rootfs from the minimal ISO) make sense, and from day 1 the script included [code intended to handle](https://github.com/openshift/assisted-service/blob/a714f40bf720c30d8e491bdf54e847fb468c0654/internal/constants/scripts.go#L53-L54) this case, but when it was run in the initramfs it was done by writing the same script into a dracut hook, and then recently [changed](https://github.com/openshift/assisted-service/pull/5107) to a _separate_ systemd service. _This_ systemd service (the one used for the non-minimal ISO) has always been the one that runs after the pivot, and it should always have been constrained to run before NetworkManager.

I think we should clean up these spurious dracut references so that we don't confuse anybody else. And also make the same change in the assisted discovery ISO.

jhernand: Thanks for the detailed analysis @zaneb. I think this is what is needed in assisted service: https://github.com/openshift/assisted-service/pull/5375

bfournie: Thanks for tracking that down @zaneb ! Updated to remove the dracut references.

bfournie:
/test unit
Checking if updated nmstate is available yet

PR #7365, opened 2023-07-25
CFE-686 : Update openshift/api package to latest version

bharath-b-rh:
PR is for updating the openshift/api package to the latest version which contains the API changes for supporting GCP labels & tags made in https://github.com/openshift/api/pull/1445


patrickdillon:
#7401 is pulling in a newer API version. We should probably defer this PR for that one.

bharath-b-rh:
Closing in favor of https://github.com/openshift/installer/pull/7401

PR #7363, opened 2023-07-25
OCPBUGS-16692: OpenStack: fix crash with empty platform in machinepool

mandre:
Passing an empty platform in machinepools caused the installer to exit with an ugly stack trace. This commit fixes the convertOpenStack() function to account for cases where the platform is not defined in the machine pools.

PR #7367, opened 2023-07-25
Bug OCPBUGS-16249: Add ip=dhcp,dhcp6 option to Kernel args

MaysaMacedo:
On clusters configured with dual-stack network the
IPv4 and IPv6 addresses can be added to the main interface
at different time, which results in the openshift node addresses
not containing the IPv6 address. This commit fixes the issue
by including `ip=dhcp,dhcp6` to the kernel args of masters and works,
which sets `required-timeout` to an value that the IP configuration
will be tried before succeeds. This configuration is valid for day1
dual-stack clusters only.

Partially Implements: https://github.com/openshift/enhancements/pull/1365

### DIFF
--- a/pkg/asset/machines/machineconfig/ipv6.go
+++ b/pkg/asset/machines/machineconfig/ipv6.go
@@ -1,0 +1,42 @@
+package machineconfig
+
+import (
+	"fmt"
+
+	igntypes "github.com/coreos/ignition/v2/config/v3_2/types"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/installer/pkg/asset/ignition"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+)
+
+// ForDualStackAddresses creates the MachineConfig to tell kernel to configure the IP addresses with DHCP and DHCPV6.
+func ForDualStackAddresses(role string) (*mcfgv1.MachineConfig, error) {
+	ignConfig := igntypes.Config{
+		Ignition: igntypes.Ignition{
+			Version: igntypes.MaxVersion.String(),
+		},
+	}
+
+	rawExt, err := ignition.ConvertToRawExtension(ignConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	return &mcfgv1.MachineConfig{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: mcfgv1.SchemeGroupVersion.String(),
+			Kind:       "MachineConfig",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: fmt.Sprintf("99-dual-stack-%s", role),
+			Labels: map[string]string{
+				"machineconfiguration.openshift.io/role": role,
+			},
+		},
+		Spec: mcfgv1.MachineConfigSpec{
+			Config:          rawExt,
+			KernelArguments: []string{"ip=dhcp,dhcp6"},
+		},
+	}, nil
+}
--- a/pkg/asset/machines/master.go
+++ b/pkg/asset/machines/master.go
@@ -539,6 +539,17 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 		}
 		machineConfigs = append(machineConfigs, ignMultipath)
 	}
+	// The maximum number of networks supported on ServiceNetwork is two, one IPv4 and one IPv6 network.
+	// The cluster-network-operator handles the validation of this field.
+	// Reference: https://github.com/openshift/cluster-network-operator/blob/fc3e0e25b4cfa43e14122bdcdd6d7f2585017d75/pkg/network/cluster_config.go#L45-L52
+	if ic.Platform.Name() == openstacktypes.Name && len(installConfig.Config.ServiceNetwork) == 2 {
+		// Only configure kernel args for dual-stack clusters.
+		ignIPv6, err := machineconfig.ForDualStackAddresses("master")
+		if err != nil {
+			return errors.Wrap(err, "failed to create ignition to configure IPv6 for master machines")
+		}
+		machineConfigs = append(machineConfigs, ignIPv6)
+	}
 
 	m.MachineConfigFiles, err = machineconfig.Manifests(machineConfigs, "master", directory)
 	if err != nil {
--- a/pkg/asset/machines/worker.go
+++ b/pkg/asset/machines/worker.go
@@ -307,6 +307,17 @@ func (w *Worker) Generate(dependencies asset.Parents) error {
 			}
 			machineConfigs = append(machineConfigs, ignMultipath)
 		}
+		// The maximum number of networks supported on ServiceNetwork is two, one IPv4 and one IPv6 network.
+		// The cluster-network-operator handles the validation of this field.
+		// Reference: https://github.com/openshift/cluster-network-operator/blob/fc3e0e25b4cfa43e14122bdcdd6d7f2585017d75/pkg/network/cluster_config.go#L45-L52
+		if ic.Platform.Name() == openstacktypes.Name && len(installConfig.Config.ServiceNetwork) == 2 {
+			// Only configure kernel args for dual-stack clusters.
+			ignIPv6, err := machineconfig.ForDualStackAddresses("worker")
+			if err != nil {
+				return errors.Wrap(err, "failed to create ignition to configure IPv6 for worker machines")
+			}
+			machineConfigs = append(machineConfigs, ignIPv6)
+		}
 
 		switch ic.Platform.Name() {
 		case alibabacloudtypes.Name:
--- a/scripts/openstack/manifest-tests/dual-stack/install-config.yaml
+++ b/scripts/openstack/manifest-tests/dual-stack/install-config.yaml
@@ -1,0 +1,46 @@
+apiVersion: v1
+baseDomain: shiftstack.example.com
+featureSet: TechPreviewNoUpgrade
+controlPlane:
+  hyperthreading: Enabled
+  architecture: amd64
+  name: master
+  platform:
+    openstack:
+      type: ${COMPUTE_FLAVOR}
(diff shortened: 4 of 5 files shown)


MaysaMacedo:
[Here](https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/merge_requests/994) there are details on why `ip=dhcp,dhcp6` helps with the timing issue.

The following are the settings added with `ip=dhcp,dhcp6`:

`$ /usr/libexec/nm-initrd-generator -s -- ip=dhcp,dhcp6

*** Configuration '15-carrier-timeout.conf' ***
[device-15-carrier-timeout]
match-device=*
carrier-wait-timeout=10000
[.config]
enable=env:initrd

*** Connection 'default_connection' ***
[connection]
id=Wired Connection
uuid=9d45bc82-f499-4a60-829e-1e1a6342f8be
type=ethernet
autoconnect-priority=-100
autoconnect-retries=1
multi-connect=3
[ethernet]
[ipv4]
dhcp-timeout=90
method=auto
required-timeout=20000
[ipv6]
dhcp-timeout=90
method=auto
required-timeout=20000
[proxy]
[user]
org.freedesktop.NetworkManager.origin=nm-initrd-generator`

There was an issue with `systemd-network-generator.service` but it seems to be tracked by this [bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=2192573):

`[core@ostest-n7tzh-master-0 ~]$ sudo journalctl -u systemd-network-generator.service
Jul 25 13:19:26 ostest-n7tzh-master-0 systemd[1]: systemd-network-generator.service: Deactivated successfully.
Jul 25 13:19:26 ostest-n7tzh-master-0 systemd[1]: Stopped Generate network units from Kernel command line.
-- Boot 36e45f7edefd47df85a1239de2264c8e --
Jul 25 13:19:39 ostest-n7tzh-master-0 systemd-network-generator[766]: Failed to parse kernel command line: Invalid argument`

pkg/asset/machines/machineconfig/ipv6.go
```diff
@@ -0,0 +1,42 @@
...
+
+	igntypes "github.com/coreos/ignition/v2/config/v3_2/types"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/installer/pkg/asset/ignition"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+)
+
+// ForIPv6Addresses creates the MachineConfig to tell kernel to configure the IP addresses with DHCP and DHCPV6.
+func ForIPv6Addresses(role string) (*mcfgv1.MachineConfig, error) {
```
mandre, line 14: Shouldn't it be `ForDualStack()` then? Because the current name `ForIPv6Addresses()` and the name of the file suggest it configures the node for IPv6 only.

mandre (commented):
It would be nice if we had tests, in the form of unit tests and/or manifests test (like I did in https://github.com/openshift/installer/pull/7363) to verify we correctly generate the manifest adding the kernel option.

pkg/asset/machines/machineconfig/ipv6.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+
+	igntypes "github.com/coreos/ignition/v2/config/v3_2/types"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/installer/pkg/asset/ignition"
+	mcfgv1 "github.com/openshift/machine-config-operator/pkg/apis/machineconfiguration.openshift.io/v1"
+)
+
+// ForIPv6Addresses creates the MachineConfig to tell kernel to configure the IP addresses with DHCP and DHCPV6.
```
mkowalski, line 13: Just a side note, feel free to discard if not relevant - if you enable DHCPv6, make sure nodes don't get IPv6 address using SLAAC. We are noticing lots of issues for setups where you have both SLAAC and DHCPv6 and in principle we call such a configuration as unsupported

MaysaMacedo: @mkowalski Which issues are you seeing?
I tried this configuration multiple times on an install that was using slaac Subnet and didn't see any issues.

Is slaac supported with baremetal? I see baremetal configures dhcpv6 by default:
https://github.com/openshift/cluster-baremetal-operator/blob/master/provisioning/provisioning_info.go#L28-L29

mkowalski: Have you tried dual-stack VIPs with hosts that have DHCPv6 and SLAAC at the same time? I am getting numerous reports that in this case the 2nd VIP (from the IPv6 subnet) is missing. After some investigations it turns out that if a host has multiple IPv6 addresses from the Machine Network, undesired stuff happens

MaysaMacedo: Is your questions about if we have tried dual-stack installs where the OpenShift nodes have more than one IPv6 subnet for the main interface (the one node addresses are derived from)? or is it about the main interface having one v4 and v6 and an additional interface having another v6 type?

For the first case we don't support that, in the installer we validate that only one v4 and v6 exists.
For the second case we didn't have any issues.

mkowalski: 2nd case is not a problem indeed.

>OpenShift nodes have more than one IPv6 subnet for the main interface (the one node addresses are derived from)?

Yes, this, but it's not about a content of install-config.yaml because this indeed you validate in o/installer; it's about output of `ip a`. Look at the example below

```
$ ip -6 a
2: wlp0s20f3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 state UP qlen 1000
    inet6 fde4:425:6b66:a::628/128 scope global dynamic noprefixroute 
       valid_lft 42524sec preferred_lft 42524sec
    inet6 2001:1711:fa41:6a0a::628/128 scope global dynamic noprefixroute 
       valid_lft 42524sec preferred_lft 42524sec
    inet6 2001:1711:fa41:6a0a:15f4:a897:a3c0:e009/64 scope global noprefixroute 
       valid_lft forever preferred_lft forever
    inet6 fde4:425:6b66:a:a848:758a:a664:6deb/64 scope global noprefixroute 
       valid_lft forever preferred_lft forever
    inet6 fe80::84ca:4a24:ffbf:1778/64 scope link noprefixroute 
       valid_lft forever preferred_lft forever
```

`2001:1711:fa41:6a0a::628` comes from DHCPv6 and `2001:1711:fa41:6a0a:15f4:a897:a3c0:e009` comes from SLAAC. In install-config.yaml the v6 subnet is `2001:1711:fa41:6a0a::/64`. Such an attempt will fail because interface `wlp0s20f3` has 2 IP addresses that belong to the same subnet.

I am almost certain that o/installer doesn't have a validation for this because it's not something you see before you boot your node and let NetworkManager to settle

MaysaMacedo: I just tested one install with slaac and I see two addresses, but both came from the same slaac subnet. One correspond to the primary address assigned to the VM and the other correspond to the API VIP. So, no issues seen.

```
[core@ostest-jphwr-master-2 ~]$ ip a show br-ex |grep fd2
    inet6 fd2e:6f44:5dd8:c956:f816:3eff:fe78:cf36/128 scope global nodad deprecated 
    inet6 fd2e:6f44:5dd8:c956:f816:3eff:fe5e:9d93/64 scope global noprefixroute
```

In your case, could it be that users have been adding VIPs to the install-config with address type different from  the primary subnet of the nodes?

mkowalski: It's hard to say as whenever I fixed some part, we were discovering yet-another-edge-case that was leading to issues. Overall the conclusion was "the less IP addresses you have on a single interface, the better".

Having said all this, I don't expect that you change anything on your side. Just that once you start using it and spot some weird things happening, remember that we had them and can probably skip some initial debugging

MaysaMacedo: ack, thanks for bringing this to my attention!

MaysaMacedo:
Note that this configuration is only enforced for day-1 dual-stack clusters. Once we start working on conversion to dual-stack clusters we will need to figure out one additional way on how to apply this config.

pkg/asset/machines/master.go
```diff
@@ -539,6 +539,14 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 		}
 		machineConfigs = append(machineConfigs, ignMultipath)
 	}
+	if ic.Platform.Name() == openstacktypes.Name && len(installConfig.Config.ServiceNetwork) == 2 {
```
EmilienM, line 542: I'm surprised here that we do not check the ip version of strings in `installConfig.Config.ServiceNetwork`; and if there is a ipv6 match we enter in the condition.
What if the `installConfig.Config.ServiceNetwork` has two IPv4 service networks?

MaysaMacedo: I followed the same rationale from here https://github.com/openshift/installer/blob/master/pkg/asset/ignition/bootstrap/common.go#L291-L297

Also only one v4 and v6 CIDR max is allowed, according to this [code fragment](https://github.com/openshift/cluster-network-operator/blob/master/pkg/network/cluster_config.go#L50-L52).

EmilienM: Indeed, https://github.com/openshift/cluster-network-operator/blob/fc3e0e25b4cfa43e14122bdcdd6d7f2585017d75/pkg/network/cluster_config.go#L46-L52 validates it in CNO.
I guess this is fine, as we probably don't want to repeat the same validation logic in the installer; in case it changes in CNO at some point.

As someone who reads this code for the first time, I would have appreciated a comment explain that ^ with a permalink to the CNO code eventually.
So whoever comes back to it knows why we "only" validate that ServiceNetwork has 2 strings and the rest is done via CNO.

mandre (commented):
/lgtm
/approve

r4f4 (commented):
/approve

PR #7372, opened 2023-07-27
OCPBUGS-16915: [release-4.11] azure: skip LB creation when not needed

r4f4:
When one creates an IPI Azure cluster with an `internal` publishing method, it creates a standard load balancer with an empty definition. This load balancer doesn't serve a purpose since the configuration is completely empty. Because it doesn't have a public IP address and backend pools it's not providing any outbound connectivity, and there are no frontend IP configurations for ingress connectivity to the cluster.


pawanpinjarkar (commented):
/lgtm

PR #7378, opened 2023-07-28
OpenStack: Remove SGS created by CPO on destroy

dulek:
cloud-provider-openstack can be configured to create security groups for the NodePorts of the load balancers it is creating. These SGs are then attached to the nodes. On `cluster destroy` we're orphaning them. This commit makes sure that we're looking for them.

As they aren't tagged or have a proper cluster ID in the name, we will look at each of the ports, list its SGs and evaluate them comparing names with the pattern. If it matches, `destroy` will attempt to delete such SG.

### DIFF
--- a/pkg/destroy/openstack/openstack.go
+++ b/pkg/destroy/openstack/openstack.go
@@ -3,6 +3,7 @@ package openstack
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
@@ -43,6 +44,7 @@ const (
 	cinderCSIClusterIDKey           = "cinder.csi.openstack.org/cluster"
 	manilaCSIClusterIDKey           = "manila.csi.openstack.org/cluster"
 	minOctaviaVersionWithTagSupport = "v2.5"
+	cloudProviderSGNamePattern      = `^lb-sg-[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}`
 )
 
 // Filter holds the key/value pairs for the tags we will be matching
@@ -372,6 +374,27 @@ func getFIPsByPort(conn *gophercloud.ServiceClient, logger logrus.FieldLogger) (
 	return fipByPort, err
 }
 
+// getSGsByID prefetches a list of SGs and organizes it by ID for easy lookup.
+func getSGsByID(conn *gophercloud.ServiceClient, logger logrus.FieldLogger) (map[string]sg.SecGroup, error) {
+	sgByID := make(map[string]sg.SecGroup)
+	allPages, err := sg.List(conn, sg.ListOpts{}).AllPages()
+	if err != nil {
+		logger.Error(err)
+		return sgByID, nil
+	}
+	allSGs, err := sg.ExtractGroups(allPages)
+	if err != nil {
+		logger.Error(err)
+		return sgByID, nil
+	}
+
+	// Organize SGs for easy lookup
+	for _, group := range allSGs {
+		sgByID[group.ID] = group
+	}
+	return sgByID, err
+}
+
 func deletePorts(opts *clientconfig.ClientOpts, listOpts ports.ListOpts, logger logrus.FieldLogger) (bool, error) {
 	logger.Debug("Deleting openstack ports")
 	defer logger.Debugf("Exiting deleting openstack ports")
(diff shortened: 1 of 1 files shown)


dulek:
If any of reviewers have a better idea how to find these SGs and **not include any SG of other cluster** I'd be grateful for advice.

dulek:
/retest

`gzip: invalid checksum`.

pkg/destroy/openstack/openstack.go
```diff
@@ -422,6 +452,28 @@ func deletePorts(opts *clientconfig.ClientOpts, listOpts ports.ListOpts, logger
 				}
 			}
 
+			// If there is a security group created by cloud-provider-openstack we should find it and delete it.
+			// We'll look through the ones on each of the ports and attempt to remove it from the port and delete it.
+			// Most of the time it's a conflict, but last port should be guaranteed to allow deletion.
+			// TODO(dulek): Currently this is the only way to do it and if delete fails there's no way to get back to
+			//              that SG. This is bad and we should make groups created by CPO tagged by cluster ID ASAP.
```
MaysaMacedo, line 459: Is this issue being tracked somewhere so we don't forget about it?

pkg/destroy/openstack/openstack.go
```diff
@@ -422,6 +452,28 @@ func deletePorts(opts *clientconfig.ClientOpts, listOpts ports.ListOpts, logger
...
+				SecurityGroups: &[]string{}, // We can just detach all, we're deleting this port anyway.
+			})
+			for _, groupID := range port.SecurityGroups {
+				if group, ok := sgByID[groupID]; ok {
+					if cloudProviderSGNameRegexp.MatchString(group.Name) {
+						logger.Debugf("Deleting cloud-provider-openstack SG %q", groupID)
+						err := sg.Delete(conn, groupID).ExtractErr()
+						var err404 gophercloud.ErrDefault404
+						var err409 gophercloud.ErrDefault409
+						if !errors.As(err, &err404) && !errors.As(err, &err409) { // Ignore 404 Not Found and 409 Conflict
```
MaysaMacedo, line 470: If there was no error when deleting the sg, should we clean it up from the `sgByID` map? perhaps this way we avoid unnecessary calls to delete sgs when they're already gone

pkg/destroy/openstack/openstack.go
```diff
@@ -422,6 +452,28 @@ func deletePorts(opts *clientconfig.ClientOpts, listOpts ports.ListOpts, logger
...
 
+			// If there is a security group created by cloud-provider-openstack we should find it and delete it.
+			// We'll look through the ones on each of the ports and attempt to remove it from the port and delete it.
+			// Most of the time it's a conflict, but last port should be guaranteed to allow deletion.
+			// TODO(dulek): Currently this is the only way to do it and if delete fails there's no way to get back to
+			//              that SG. This is bad and we should make groups created by CPO tagged by cluster ID ASAP.
+			ports.Update(conn, port.ID, ports.UpdateOpts{
+				SecurityGroups: &[]string{}, // We can just detach all, we're deleting this port anyway.
+			})
+			for _, groupID := range port.SecurityGroups {
```
mandre, line 463: It's a bit confusing to use `port.SecurityGroups` after [resetting the port SGs](https://github.com/openshift/installer/pull/7378/files#diff-e480f6c0393b6788fe8352f4f8bdc63f021f9e84de632a21607aab8b414c6e3fR460). I suggest saving the SGs list into a variable.

pkg/destroy/openstack/openstack.go
dulek: Hm, not on 409, but on 404 for sure.

pkg/destroy/openstack/openstack.go
dulek: https://issues.redhat.com/browse/OSASINFRA-3200. This is a broader problem really, we don't have a good way to distinguish LBs between clusters too and currently we can have clashes by names in CPO. :/

pkg/destroy/openstack/openstack.go
```diff
@@ -422,6 +452,32 @@ func deletePorts(opts *clientconfig.ClientOpts, listOpts ports.ListOpts, logger
...
+				SecurityGroups: &[]string{}, // We can just detach all, we're deleting this port anyway.
+			})
+			for _, groupID := range assignedSGs {
+				if group, ok := sgByID[groupID]; ok {
+					if cloudProviderSGNameRegexp.MatchString(group.Name) {
+						logger.Debugf("Deleting cloud-provider-openstack SG %q", groupID)
+						err := sg.Delete(conn, groupID).ExtractErr()
+						var err404 gophercloud.ErrDefault404
+						var err409 gophercloud.ErrDefault409
+						if errors.As(err, &err404) {
```
MaysaMacedo, line 471: can't we clean the sgID from the map right after the successful operation of deleting the SG (in other words when there is no error)?

dulek: Ah, okay, you're right.

mandre (commented):
/lgtm
/approve

PR #7380, opened 2023-07-30
OCPBUGS-17064: always create a MachineSet

EmilienM:
... even if zero workers is requested from install-config (replicas ==
0).

This was a non backward compatible change introduced by https://github.com/openshift/installer/pull/7300

### DIFF
--- a/pkg/asset/machines/openstack/machinesets.go
+++ b/pkg/asset/machines/openstack/machinesets.go
@@ -31,9 +31,6 @@ func MachineSets(clusterID string, config *types.InstallConfig, pool *types.Mach
 	if poolPlatform := pool.Platform.Name(); poolPlatform != openstack.Name {
 		return nil, fmt.Errorf("non-OpenStack machine-pool: %q", poolPlatform)
 	}
-	if pool.Replicas == nil || *pool.Replicas < 1 {
-		return nil, nil
-	}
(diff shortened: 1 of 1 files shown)


EmilienM:
We need to watch for `e2e-openstack-nfv-intel`.

mandre (commented):
The testpmd pod is failing with the following error, that is unrelated to the fix here:
```
"message": "failed to run pre-start hook for container \"fec941d140c00ad95372c7c3c81a52c9a81a6e5ca997a67475c5f9dea44eb545\": set CPU load balancing: disabling CPU load balancing on cgroupv2 not yet supported"
```

At least, we can see it now deploys the compact cluster.

/lgtm
/approve

EmilienM:
> The testpmd pod is failing with the following error, that is unrelated to the fix here:
> 
> ```
> "message": "failed to run pre-start hook for container \"fec941d140c00ad95372c7c3c81a52c9a81a6e5ca997a67475c5f9dea44eb545\": set CPU load balancing: disabling CPU load balancing on cgroupv2 not yet supported"
> ```
> 
> At least, we can see it now deploys the compact cluster.

Yes this is another issue; I thought the cgroups thing was fixed, but I'll investigate that next. In the meantime I think we can land this one.

PR #7320, opened 2023-07-12
OSDOCS-6880: Adding 64-bit ARM GCP instance types to documentation

kelbrown20:
**Jira story:**  [OSDOCS-6880](https://issues.redhat.com/browse/OSDOCS-6880)
Subtask of [OSDOCS-6319](https://issues.redhat.com/browse/OSDOCS-6319) [Documentation enablement for GCP on ARM]

**Description:** Adding supported ARM64 GCP instance types to installer documentation. Will also use this file for OCP documentation

PR #7381, opened 2023-07-31
OSDOCS-6999: tested machine series 'C3', 'C2D'

jianli-wei:
We also tested machine series `C3` and `C2D` during 4.14, and machine series `Tau T2A` during 4.14 ARM testing. So add them to "Tested instance types for GCP".


jianli-wei:
I believe the failure `ci/prow/e2e-gcp-ovn-shared-vpc` has nothing to do with my changes, thanks! 

/assign @patrickdillon

kelbrown20:
Hello @jianli-wei! For the OCP documentation, we are using a separate `md` file for the ARM machine types, I created one in #7320. So the ARM instance types do not need to be in this `tested_instance_types.md` file

jianli-wei:
> Hello @jianli-wei! For the OCP documentation, we are using a separate `md` file for the ARM machine types, I created one in #7320. So the ARM instance types do not need to be in this `tested_instance_types.md` file

Thank you @kelbrown20 ! I removed `Tau T2A` here.

PR #7384, opened 2023-07-31
OCPBUGS-16912: Ensure DHCPv6 client sends Solicit with mac address

bfournie:
When using DHCPv6, ensure that the host sends a DHCP Solicitation with the Client ID containing the mac address, i.e. DUID-LL. See https://datatracker.ietf.org/doc/html/rfc3315#section-9.4 for more info.  This allows the DHCPv6 server to assign an IPv6 address that can be used as the rendezvousIP.


rwsu (commented):
/lgtm

PR #7317, opened 2023-07-12
OCPBUGS-16515: gcp: use zones available for both instance and project

r4f4:
In https://github.com/openshift/installer/pull/7096 we fixed the validation but the same logic needs to be done for the machine asset generation.

PR #7388, opened 2023-08-02
OCPBUGS-17227: gcp: fix validation of custom instance types

r4f4:
Users can create their own instance types in GCloud. Those types are not returned when listing the machine types for a given project/region. This would cause those installs with custom types to fail with the error:

```
Internal error: failed to fetch instance type, this error usually occurs if the region or the instance type is not found
```

after the validations introduced by
https://github.com/openshift/installer/pull/7317 and https://github.com/openshift/installer/pull/7096.

To fix that, let's fallback to the previous way of fetching the specific machine type whenever the aggregated list returns 0 elements.

### DIFF
--- a/pkg/asset/installconfig/gcp/client.go
+++ b/pkg/asset/installconfig/gcp/client.go
@@ -104,9 +104,6 @@ func GetMachineTypeList(ctx context.Context, svc *compute.Service, project, regi
 		}
 		return nil
 	})
-	if len(machines) == 0 {
-		return nil, errors.New("failed to fetch instance type, this error usually occurs if the region or the instance type is not found")
-	}
 
 	return machines, err
 }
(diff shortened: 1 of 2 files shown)


jianli-wei (approved):

barbacbd (commented):
LGTM

We may want to change the search to loop through all of the zones to make sure that the instances are found if they exist in any zone other than the first found.

r4f4:
> LGTM
> 
> We may want to change the search to loop through all of the zones to make sure that the instances are found if they exist in any zone other than the first found.

This is what was done before the zone validation was introduced. See https://github.com/openshift/installer/commit/ad63edf46bd525cb6a4462bfafab2b940158807b#diff-df0d06427cd5be6b8b56f21664d1f138086071766cb01af9d0ce7b8437396cf7L124-R162

PR #7392, opened 2023-08-03
OCPBUGS-13408: $ openshift-install agent create agent-config-template --dir=./foo gives nothing in the INFO log-level

marcbrown2024:
Fixed empty log message by adding the correct INFO log during 'agent-config-template' creation.

### DIFF
--- a/pkg/asset/agent/agentconfig/agent_config.go
+++ b/pkg/asset/agent/agentconfig/agent_config.go
@@ -94,6 +94,12 @@ hosts:
 
 	a.Template = agentConfigTemplate
 
+	// Set the File field correctly with the generated agent config YAML content
+	a.File = &asset.File{
+		Filename: agentConfigFilename,
+		Data:     []byte(a.Template),
+	}
+
 	// TODO: template is not validated
 	return nil
 }


rwsu (commented):
/lgtm

rwsu:
Nice fix @marcbrown2024

pkg/asset/agent/agentconfig/agent_config.go
```diff
@@ -94,6 +94,12 @@ hosts:
 
 	a.Template = agentConfigTemplate
 
+	// Set the File field correctly with the generated agent config YAML content
+	a.File = &asset.File{
+		Filename: agentConfigFilename,
+		Data:     []byte(a.Template),
+	}
+
```
andfasano, line 102: Thanks for fix @marcbrown2024 ! It would be really nice if you could add also an integration terst for the `create agent-config-template` command, so that we could verify both the output message and generated file (see for example https://github.com/openshift/installer/blob/master/cmd/openshift-install/testdata/agent/image/configurations/compact.txt) (the it framework was added later, so not all commands where still covered)

pawanpinjarkar (approved):
/approve

pkg/asset/agent/agentconfig/agent_config.go
pawanpinjarkar: Here is https://github.com/openshift/installer/pull/7402

pawanpinjarkar:
Moving the changes into a new PR https://github.com/openshift/installer/pull/7408
/closed

PR #7397, opened 2023-08-04
CORS-2660: GCP: deprecate the licenses field

AnnaZivkovic:
It has been at leas two years since the licence field has become obsolete due RHCOS GCP images already having nested virt licences by default. The licenses field is being kept for backwards compatibility

For a more detailed history reference https://github.com/openshift/installer/pull/4696

### DIFF
--- a/pkg/asset/machines/gcp/machines.go
+++ b/pkg/asset/machines/gcp/machines.go
@@ -132,9 +132,7 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 
 func provider(clusterID string, platform *gcp.Platform, mpool *gcp.MachinePool, osImage string, azIdx int, role, userDataSecret string, credentialsMode types.CredentialsMode) (*machineapi.GCPMachineProviderSpec, error) {
 	az := mpool.Zones[azIdx]
-	if len(platform.Licenses) > 0 {
-		osImage = fmt.Sprintf("%s-rhcos-image", clusterID)
-	} else if mpool.OSImage != nil {
+	if mpool.OSImage != nil {
 		osImage = fmt.Sprintf("projects/%s/global/images/%s", mpool.OSImage.Project, mpool.OSImage.Name)
 	}
 	network, subnetwork, err := getNetworks(platform, clusterID, role)
--- a/pkg/types/gcp/platform.go
+++ b/pkg/types/gcp/platform.go
@@ -35,14 +35,6 @@ type Platform struct {
 	// +optional
 	ComputeSubnet string `json:"computeSubnet,omitempty"`
 
-	// Licenses is a list of licenses to apply to the compute images
-	// The value should a list of strings (https URLs only) representing the license keys.
-	// When set, this will cause the installer to copy the image into user's project.
-	// This option is incompatible with any mechanism that makes use of pre-built images
-	// such as the current env OPENSHIFT_INSTALL_OS_IMAGE_OVERRIDE
-	// +optional
-	Licenses []string `json:"licenses,omitempty"`
-
 	// userLabels has additional keys and values that the installer will add as
 	// labels to all resources that it creates on GCP. Resources created by the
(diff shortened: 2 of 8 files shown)


pkg/types/gcp/validation/platform.go
```diff
@@ -104,8 +108,8 @@ func ValidatePlatform(p *gcp.Platform, fldPath *field.Path, ic *types.InstallCon
 	}
 
 	for i, license := range p.Licenses {
-		if validate.URIWithProtocol(license, "https") != nil {
-			allErrs = append(allErrs, field.Invalid(fldPath.Child("licenses").Index(i), license, "licenses must be URLs (https) only"))
+		if license != nestedVirtLicense {
+			allErrs = append(allErrs, field.Invalid(fldPath.Child("licenses").Index(i), license, "only accepted license is enable-vmx"))
```
r4f4, line 112: If you remove the creation of the image with the licenses in terraform, specifying any license in the install-config does nothing.

r4f4 (commented):
I was thinking about only marking the field as deprecated not removing the functionality itself (yet). If we can remove the functionality, great, but I'm afraid we might not be able to.

@patrickdillon Do you know?

pkg/types/gcp/validation/platform.go
AnnaZivkovic: Are you thinking we should just change the license field description and put out a warning if it is used?

r4f4: Yes, that's what I had in mind. We need to deprecate the field first, and then after a few releases we can remove.

pkg/types/gcp/platform.go
```diff
@@ -35,11 +35,7 @@ type Platform struct {
...
 	ComputeSubnet string `json:"computeSubnet,omitempty"`
 
-	// Licenses is a list of licenses to apply to the compute images
-	// The value should a list of strings (https URLs only) representing the license keys.
-	// When set, this will cause the installer to copy the image into user's project.
-	// This option is incompatible with any mechanism that makes use of pre-built images
-	// such as the current env OPENSHIFT_INSTALL_OS_IMAGE_OVERRIDE
+	// Licenses is no longer used and should be either empty or contain the nested virt license.
 	// +optional
 	Licenses []string `json:"licenses,omitempty"`
```
patrickdillon, line 40: Let's go ahead and remove `Licenses`. It has been deprecated long enough that we are safe to remove.

patrickdillon:
Once we remove licenses, please make sure to update the epic & alert docs:

- remove `licenses` field from docs
- A release note or something to indicate the removal

pkg/asset/machines/gcp/machines.go
```diff
@@ -132,9 +132,7 @@ func Machines(clusterID string, config *types.InstallConfig, pool *types.Machine
 
 func provider(clusterID string, platform *gcp.Platform, mpool *gcp.MachinePool, osImage string, azIdx int, role, userDataSecret string, credentialsMode types.CredentialsMode) (*machineapi.GCPMachineProviderSpec, error) {
 	az := mpool.Zones[azIdx]
-	if len(platform.Licenses) > 0 {
-		osImage = fmt.Sprintf("%s-rhcos-image", clusterID)
-	} else if mpool.OSImage != nil {
+	if mpool.OSImage != nil {
 		osImage = fmt.Sprintf("projects/%s/global/images/%s", mpool.OSImage.Project, mpool.OSImage.Name)
 	}
```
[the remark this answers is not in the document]
AnnaZivkovic: Then I think we have more duplicated code

r4f4:
@AnnaZivkovic when I try to build using your branch, no binary is produced and I get the following error:
```
# github.com/openshift/installer/pkg/types/gcp/validation
pkg/types/gcp/validation/platform.go:4:2: "os" imported and not used
```
Fixing that should fix the CI job failures.

r4f4 (commented):
/lgtm

PR #7400, opened 2023-08-07
openstack: Test zero replicas in worker machine-pool

pierreprinetti:
Assert that:
* setting zero workers does not crash the Installer
* there always is at least one MachineSet

A follow-up to #7380


EmilienM:
/lgtm
/approve

Thanks @pierreprinetti !

PR #7405, opened 2023-08-08
OCPBUGS-17860: OpenStack: Remove NodePorts range 0.0.0.0/0 rules

dulek:
With openshift/cluster-cloud-controller-manager-operator#264 we should no longer need the SG rules opening whole NodePorts range by default. cloud-provider-openstack will manage this on its own now.
    
We still need to keep the rules opening the traffic inside the cluster to make sure traffic redirections and regular NodePort services work.

### DIFF
--- a/data/data/openstack/masters/sg-master.tf
+++ b/data/data/openstack/masters/sg-master.tf
@@ -233,25 +233,25 @@ resource "openstack_networking_secgroup_rule_v2" "master_ingress_etcd" {
 }
 
 resource "openstack_networking_secgroup_rule_v2" "master_ingress_services_tcp" {
-  direction      = "ingress"
-  ethertype      = "IPv4"
-  protocol       = "tcp"
-  port_range_min = 30000
-  port_range_max = 32767
-  # For OVN LBs the traffic will have the *real* origin source-ip, so anything goes.
-  remote_ip_prefix  = "0.0.0.0/0"
+  count             = length(var.machine_v4_cidrs)
+  direction         = "ingress"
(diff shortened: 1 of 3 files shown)


dulek:
I've seen this working locally, let's see!

dulek:
/hold
/test e2e-openstack-ovn

I'd like to see a successful run of e2e-openstack-ovn here.

dulek:
Failures are because of https://github.com/kubernetes/cloud-provider-openstack/pull/2320.

mandre:
/hold until we have https://github.com/kubernetes/cloud-provider-openstack/pull/2320.

We'll probably need to update UPI scripts as well: https://github.com/openshift/installer/blob/aa258c0/upi/openstack/security-groups.yaml#L161-L175 and https://github.com/openshift/installer/blob/aa258c0/upi/openstack/security-groups.yaml#L275-L289

dulek:
/hold cancel

Okay, that's an MVP version of this that should pass the tests and also makes a significant improvement to the previous situation.

dulek:
UPI looks good, proxy job indicates no problems with NodePort Services tests either.

dulek:
OVN job looks good.

r4f4:
/override ci/prow/okd-scos-images
Pod doesn't seem to have enough resources

dulek:
/retest-required

Weird, seems like no required jobs are hanging here. e2e-openstack-ovn fails with a known issue getting fixes at openshift/openstack-test#118.

PR #7408, opened 2023-08-08
OCPBUGS-13408: Log message and add integration test

pawanpinjarkar:
Log message and add integration test for `openshift-install agent create agent-config-template` command

Signed-Off by:
@marcbrown2024
@pawanpinjarkar


rwsu (commented):
/lgtm

PR #7411, opened 2023-08-09
OpenStack: Set external network for cloud-provider

dulek:
We allow users to specify an external network to create the floating IPs from on the install-config.yaml. We use that to create the API and Ingress FIPs and forget about it. We can however also use it to configure the cloud-provider-openstack to make sure it'll choose the correct network when creating FIPs for the `type=LoadBalancer` Services too. This commit implements it.

### DIFF
--- a/pkg/asset/installconfig/openstack/session.go
+++ b/pkg/asset/installconfig/openstack/session.go
@@ -17,6 +17,7 @@ var onceLoggers = map[string]*sync.Once{}
 // Session is an object representing session for OpenStack.
 type Session struct {
 	CloudConfig *clientconfig.Cloud
+	ClientOpts  *clientconfig.ClientOpts
 }
 
 // GetSession returns an OpenStack session for a given cloud name in clouds.yaml.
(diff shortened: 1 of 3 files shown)


gryf (approved):
/lgtm

dulek:
> Remind me what is the current behavior of the cloud provider when omitting the `floating-network-id` setting? I agree that setting the default floating IP network to be the one that is specified as the external network is more user friendly and avoid unexpected behavior when there are multiple external networks in the environment. They can change the setting on day-2 if desired.

They can change the setting on day one too, just use `create manifests` and update config as you wish.

Current behavior is to take the first external network that has an IPv4 subnet: https://github.com/kubernetes/cloud-provider-openstack/blob/ae7fe54025b7b7c0b01ec039112f625859128c18/pkg/util/openstack/network.go#L118-L125. This is problematic if there are multiple. It's even worse as the networks are returned in alphabetical order based on UUID, meaning that often we're getting the correct one and doesn't realize there's a problem here.

dulek:
@gryf: Would you care to `/lgtm` again? Seems like this can get in once master is open for 4.15.

mandre (commented):
/lgtm
/approve

PR #7413, opened 2023-08-10
CORS-2877: Install Config Feature Gate Validation

patrickdillon:
Adds the ability to validate install config fields using feature gates.

Fields can be gated based on any feature set: Default, TechPreviewNoUpgrade, or CustomNoUpgrade.

This means that features can be promoted without changing validation code by vendoring the updated API, where the feature has been promoted to the Default feature set.

Also enables more targeted testing for technical quality.

### DIFF
--- a/pkg/types/featuregates/featuregate.go
+++ b/pkg/types/featuregates/featuregate.go
@@ -1,0 +1,57 @@
+package featuregates
+
+// source: https://github.com/openshift/library-go/blob/c515269de16e5e239bd6e93e1f9821a976bb460b/pkg/operator/configobserver/featuregates/featuregate.go#L23-L28
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	configv1 "github.com/openshift/api/config/v1"
+)
+
+// FeatureGate indicates whether a given feature is enabled or not
+// This interface is heavily influenced by k8s.io/component-base, but not exactly compatible.
+type FeatureGate interface {
+	// Enabled returns true if the key is enabled.
+	Enabled(key configv1.FeatureGateName) bool
+}
+
+type featureGate struct {
+	enabled  sets.Set[configv1.FeatureGateName]
+	disabled sets.Set[configv1.FeatureGateName]
+}
+
+// GatedInstallConfigFeature contains fields that will be used to validate
+// that required feature gates are enabled when gated install config fields
+// are used.
+// FeatureGateName: openshift/api feature gate required to enable the use of Field
+// Condition: the check which determines whether the install config field is used,
+// if Condition evaluates to True, FeatureGateName must be enabled  to pass validation.
+// Field: the gated install config field, Field is used in the error message.
+type GatedInstallConfigFeature struct {
+	FeatureGateName configv1.FeatureGateName
+	Condition       bool
+	Field           *field.Path
+}
+
+func newFeatureGate(enabled, disabled []configv1.FeatureGateName) FeatureGate {
+	return &featureGate{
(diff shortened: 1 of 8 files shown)


pkg/types/featuregates/featuregate.go
```diff
@@ -0,0 +1,55 @@
...
+
+type featureGate struct {
+	enabled  sets.Set[configv1.FeatureGateName]
+	disabled sets.Set[configv1.FeatureGateName]
+}
+
+// GatedInstallConfigFeature contains fields that will be used to validate
+// that required feature gates are enabled when gated install config fields
+// are used.
+// FeatureGateName: required openshift/api to enable the use of Field
```
patrickdillon, line 28: required openshift/api feature gate to enable the use of Field

pkg/types/vsphere/validation/featuregates.go
```diff
@@ -0,0 +1,22 @@
...
+// be validated to ensure that the proper featuregate is enabled when the field is used.
+func GatedFeatures(c *types.InstallConfig) []featuregates.GatedInstallConfigFeature {
+	v := c.VSphere
+	return []featuregates.GatedInstallConfigFeature{
+		{
+			FeatureGateName: configv1.FeatureGateVSphereStaticIPs,
+			Condition:       len(v.Hosts) > 0,
+			Field:           field.NewPath("platform", "vsphere", "hosts"),
+		},
+	}
```
patrickdillon, line 21: /cc @jcpowermac @rvanderp3

pkg/types/validation/installconfig.go
```diff
@@ -1117,3 +1099,47 @@ func validateCustomFeatureGates(c *types.InstallConfig) field.ErrorList {
...
+	// legacy feature gating
+	if c.FeatureSet != configv1.TechPreviewNoUpgrade {
+		errMsg := "the TechPreviewNoUpgrade feature set must be enabled to use this field"
+
+		if c.OpenStack != nil {
+			for _, f := range openstackvalidation.FilledInTechPreviewFields(c) {
+				allErrs = append(allErrs, field.Forbidden(f, errMsg))
+			}
+		}
+	}
```
patrickdillon, line 1117: Not updating this because it will be removed by #7570, which should merge soon.

patrickdillon: @MaysaMacedo @EmilienM @pierreprinetti @mandre PTAL for future feature gating needs

pkg/types/validation/featuregate_test.go
```diff
@@ -0,0 +1,115 @@
...
+		{
+			name: "GCP UserTags is not allowed without Feature Gates",
+			installConfig: func() *types.InstallConfig {
+				c := validInstallConfig()
+				c.GCP = validGCPPlatform()
+				c.GCP.UserTags = []gcp.UserTag{{ParentID: "a", Key: "b", Value: "c"}}
+				return c
+			}(),
+			expected: `^platform.gcp.userTags: Forbidden: this field is protected by a feature gate which must be enabled through either the TechPreviewNoUpgrade or CustomNoUpgrade feature set$`,
+		},
```
patrickdillon, line 39: I'm ambivalent about adding these negative test cases, because they will need to be removed when the api is bumped to add the feature to the default feature set. I did add them in this case because I wanted to ensure the gating actually works.

patrickdillon: > Could you structure the code in such a way that you can use a fake set of features/featuresets in testing to test the logic without the moving base from o/api?

Yes I was considering the same thing and think this is a good alternative. My own personal inclination is to implement the individual test as it protects against other mistakes that could happen (users select the wrong field). IMHO I think this is good for now, but perhaps something we could refine if needed.

PR #7414, opened 2023-08-10
OCPBUGS-15659: IPI pre-check for MachineAPI capability

rwsu:
Adds a validation check to ensure that MachineAPI capability is included for IPI deployments.


patrickdillon (commented):
/approve

jinyunma:
@rwsu FYI. MachineAPI is required on IPI including IPI SON, IPI compact and common IPI (3 masters + N workers). It only can be disabled in UPI cluster.
Discussion is [here](https://redhat-internal.slack.com/archives/C032HSVS71T/p1689850599305979).

PR #7417, opened 2023-08-14
OCPBUGS-17218: Warn when firewall rull missing.

barbacbd:
**GCP Xpn: The compute.firewalls.create rule is not required, but a warning should be provided when it is not found.

### DIFF
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -434,6 +434,10 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 				return err
 			}
 			createFirewallRules = permissions.Has(GCPFirewallPermission)
+
+			if !createFirewallRules {
+				logrus.Warnf("failed to find permission %s, skipping firewall rule creation", GCPFirewallPermission)
+			}
 		}
 
 		masters, err := mastersAsset.Machines()


pkg/asset/cluster/tfvars.go
```diff
@@ -434,6 +434,10 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 				return err
 			}
 			createFirewallRules = permissions.Has(GCPFirewallPermission)
+
+			if !createFirewallRules {
+				logrus.Warnf("failed to fin permission %s, skipping firewall rule creation", GCPFirewallPermission)
```
jianli-wei, line 439: "fin" => "find"?

jianli-wei (changes requested):

pkg/asset/cluster/tfvars.go
```diff
@@ -434,6 +434,10 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 				return err
 			}
 			createFirewallRules = permissions.Has(GCPFirewallPermission)
+
+			if !createFirewallRules {
+				logrus.Warnf("failed to find permission %s, skipping firewall rule creation", GCPFirewallPermission)
```
sadasu, line 439: Can GCPFirewallPermission be "" ?

rna-afk: `GCPFirewallPermission` is a constant set here
https://github.com/openshift/installer/blob/186cb916c388d29ed3f6ef4e71d9fda409f30bdf/pkg/asset/cluster/tfvars.go#L79

PR #7422, opened 2023-08-15
OCPBUGS-17757: check credentials type to handle different gcp authentication methods

barbacbd:
**Altered GCP validation to check the type of credentials that are set (if they are set). The installer cannot

count on GCP credentials being nil or unset when authenticating using the gcloud cli. Now the installer checks for nil credentials and if the credentials exist then check the type.

### DIFF
--- a/pkg/asset/installconfig/gcp/validation.go
+++ b/pkg/asset/installconfig/gcp/validation.go
@@ -2,6 +2,7 @@ package gcp
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net"
 	"strings"
@@ -57,7 +58,7 @@ func Validate(client API, ic *types.InstallConfig) error {
 	allErrs = append(allErrs, validateZones(client, ic)...)
 	allErrs = append(allErrs, validateNetworks(client, ic, field.NewPath("platform").Child("gcp"))...)
 	allErrs = append(allErrs, validateInstanceTypes(client, ic)...)
-	allErrs = append(allErrs, validateCredentialMode(client, ic)...)
+	allErrs = append(allErrs, ValidateCredentialMode(client, ic)...)
 	allErrs = append(allErrs, validatePreexistingServiceAccountXpn(client, ic)...)
 	allErrs = append(allErrs, validateServiceAccountPresent(client, ic)...)
 	allErrs = append(allErrs, validateMarketplaceImages(client, ic)...)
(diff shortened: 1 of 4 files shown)


pkg/asset/installconfig/gcp/validation.go
```diff
@@ -469,31 +470,37 @@ func validateRegion(client API, ic *types.InstallConfig, fieldPath *field.Path)
...
 
+		if credsType.(string) == string(gcp.AuthorizedUserMode) && ic.CredentialsMode != types.ManualCredentialsMode {
+			errMsg := "environmental authentication is only supported with Manual credentials mode"
+			return append(allErrs, field.Forbidden(field.NewPath("credentialsMode"), errMsg))
+		}
+	} else if creds.JSON == nil && ic.CredentialsMode != types.ManualCredentialsMode {
+		errMsg := "environmental authentication is only supported with Manual credentials mode"
+		return append(allErrs, field.Forbidden(field.NewPath("credentialsMode"), errMsg))
+	}
 	return allErrs
```
rna-afk, line 504: I'm worried about allErrs not being used in this function. How about we return just one error and append it in the calling function?

rna-afk: By this I mean we just use allErrs to append and return which is duplicating the append logic everytime we return it. Best if we do the append once in the calling function?

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -469,31 +470,37 @@ func validateRegion(client API, ic *types.InstallConfig, fieldPath *field.Path)
...
 func ValidateCredentialMode(client API, ic *types.InstallConfig) error {
-	creds := client.GetCredentials()
-
-	if creds.JSON == nil && ic.CredentialsMode != types.ManualCredentialsMode {
-		errMsg := "environmental authentication is only supported with Manual credentials mode"
-		return field.Forbidden(field.NewPath("credentialsMode"), errMsg)
-	}
-
-	return nil
+	allErrs := validateCredentialMode(client, ic)
```
sadasu, line 473: I don't see why have 2 methods ValidateCredentialMode() and validateCredentialMode()? Just for the different

pkg/asset/installconfig/gcp/validation.go
sadasu: Agreed. We are only returning 1 error at a time from this method.

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -469,31 +470,37 @@ func validateRegion(client API, ic *types.InstallConfig, fieldPath *field.Path)
...
+			return append(allErrs, field.NotFound(field.NewPath("credentials").Child("JSON").Child("type"), "failed to find credentials type"))
 		}
-	}
 
+		if credsType.(string) == string(gcp.AuthorizedUserMode) && ic.CredentialsMode != types.ManualCredentialsMode {
+			errMsg := "environmental authentication is only supported with Manual credentials mode"
+			return append(allErrs, field.Forbidden(field.NewPath("credentialsMode"), errMsg))
+		}
+	} else if creds.JSON == nil && ic.CredentialsMode != types.ManualCredentialsMode {
+		errMsg := "environmental authentication is only supported with Manual credentials mode"
```
sadasu, line 501: How about "Manual credentials mode needs to be enabled to use environmental authentication"? We want to have different error messages for different error conditions 1. to help the user get to the correct configuration 2. for better debugging.

sadasu: Thanks for making the change even if you weren't totally convinced. :-)

pkg/asset/installconfig/gcp/validation.go
```diff
@@ -466,34 +467,32 @@ func validateRegion(client API, ic *types.InstallConfig, fieldPath *field.Path)
 	return nil
 }
 
-// ValidateCredentialMode checks whether the credential mode is
-// compatible with the authentication mode.
-func ValidateCredentialMode(client API, ic *types.InstallConfig) error {
+// ValidateCredentialMode The presence of `authorized_user` in the credentials indicates that no service account
+// was used for authentication and requires Manual credential mode.
```
sadasu, line 471: Nit: 
// ValidateCredentialMode validates setting of Manual credential mode when `authorized_user` in the 
// credentials indicates that no service account is being used for authentication

PR #7426, opened 2023-08-16
OCPBUGS-17770: azure: use marketplace image plan's publisher

r4f4:
It seems the marketplace image publisher and the image plan's publisher can differ, even in character case. For example:
```
$ az vm image list --offer rh-ocp-worker --all -otable
Architecture    Offer          Publisher       Sku                 Urn                                                             Version
--------------  -------------  --------------  ------------------  --------------------------------------------------------------  --------------
x64             rh-ocp-worker  RedHat          rh-ocp-worker       RedHat:rh-ocp-worker:rh-ocp-worker:4.8.2021122100               4.8.2021122100

$ az vm image show --urn RedHat:rh-ocp-worker:rh-ocp-worker:4.8.2021122100 --query plan
{
  "name": "rh-ocp-worker",
  "product": "rh-ocp-worker",
  "publisher": "redhat"
}
```
When checking for the image plan, Azure will match the properties values exactly and they are case-sensitive. So setting `Publisher: RedHat` in the `install-config.yaml` for the image above will result in:
```
Unable to deploy from the Marketplace image or a custom image sourced from Marketplace image. The part number in the purchase information for VM '/subscriptions/<REDACTED>/resourceGroups/jima15image1-flg24-rg/providers/Microsoft.Compute/virtualMachines/jima15image1-flg24-bootstrap' is not as expected. Beware that the Plan object's properties are case-sensitive. Learn more about common virtual machine error codes.
```
where as using `redhat` works as expected.

To avoid such issues and simplify the user experience, let's query the image's plan properties and use the publisher value returned instead of the one set by the user in the install-config.

### DIFF
--- a/pkg/asset/machines/master.go
+++ b/pkg/asset/machines/master.go
@@ -369,6 +369,18 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 			}
 		}
 
+		if mpool.OSImage.Publisher != "" {
+			img, ierr := client.GetMarketplaceImage(context.TODO(), ic.Platform.Azure.Region, mpool.OSImage.Publisher, mpool.OSImage.Offer, mpool.OSImage.SKU, mpool.OSImage.Version)
+			if ierr != nil {
+				return fmt.Errorf("failed to fetch marketplace image: %w", ierr)
+			}
+			// Publisher is case-sensitive and matched against exactly. Also the
+			// Plan's publisher might not be exactly the same as the Image's
+			// publisher
+			if img.Plan != nil && img.Plan.Publisher != nil {
+				mpool.OSImage.Publisher = *img.Plan.Publisher
+			}
+		}
 		pool.Platform.Azure = &mpool
 
 		capabilities, err := client.GetVMCapabilities(context.TODO(), mpool.InstanceType, installConfig.Config.Platform.Azure.Region)
--- a/pkg/asset/machines/worker.go
+++ b/pkg/asset/machines/worker.go
@@ -488,6 +488,18 @@ func (w *Worker) Generate(dependencies asset.Parents) error {
 				}
 			}
 
+			if mpool.OSImage.Publisher != "" {
+				img, ierr := client.GetMarketplaceImage(context.TODO(), ic.Platform.Azure.Region, mpool.OSImage.Publisher, mpool.OSImage.Offer, mpool.OSImage.SKU, mpool.OSImage.Version)
+				if ierr != nil {
+					return fmt.Errorf("failed to fetch marketplace image: %w", ierr)
+				}
+				// Publisher is case-sensitive and matched against exactly. Also
+				// the Plan's publisher might not be exactly the same as the
+				// Image's publisher
+				if img.Plan != nil && img.Plan.Publisher != nil {
+					mpool.OSImage.Publisher = *img.Plan.Publisher
+				}
+			}
 			pool.Platform.Azure = &mpool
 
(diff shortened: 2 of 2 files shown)


r4f4:
Update: fixed an issue where the install would fail if the image has no plan.

MayXuQQ:
tested "RedHat:rh-ocp-worker:rh-ocp-worker:4.8.2021122100" , and "redhat:rh-ocp-worker:rh-ocp-worker:4.8.2021122100", failed. 
_"Error: waiting for creation of Linux Virtual Machine: (Name "maxu-plan2-dmz4q-master-2" / Resource Group "maxu-plan2-dmz4q-rg"): Code="OSProvisioningTimedOut" Message="OS Provisioning for VM 'maxu-plan2-dmz4q-master-2' did not finish in the allotted time. The VM may still finish provisioning successfully. Please check provisioning state later. For details on how to check current provisioning state of Windows VMs, refer to https://aka.ms/WindowsVMLifecycle and Linux VMs, refer to https://aka.ms/LinuxVMLifecycle."_
the image just can be used on workers,  and I fail to find market-image for masters.

without the PR, error message is : 
_Error: waiting for creation of Linux Virtual Machine: (Name "jima15image1-flg24-bootstrap" / Resource Group "jima15image1-flg24-rg"): Code="VMMarketplaceInvalidInput" Message="Unable to deploy from the Marketplace image or a custom image sourced from Marketplace image. The part number in the purchase information for VM '/subscriptions/53b8f551-f0fc-4bea-8cba-6d1fefd54c8a/resourceGroups/jima15image1-flg24-rg/providers/Microsoft.Compute/virtualMachines/jima15image1-flg24-bootstrap' is not as expected. Beware that the Plan object's properties are case-sensitive. "_

I think the PR fixed the bug https://issues.redhat.com/browse/OCPBUGS-17770
cc @jinyunma

rna-afk:
> To avoid such issues and simplify the user experience, let's query the image's plan properties and use the publisher value returned instead of the one set by the user in the install-config.

Wouldn't this revert the actual use case the users wanted? Shouldn't we just say something like we don't see the publisher you supplied and it might cause errors?

r4f4:
> > To avoid such issues and simplify the user experience, let's query the image's plan properties and use the publisher value returned instead of the one set by the user in the install-config.
> 
> Wouldn't this revert the actual use case the users wanted? Shouldn't we just say something like we don't see the publisher you supplied and it might cause errors?

From what I've seen, the Image's publisher is not compared for an exact match as it happens with the Image Plan's publisher. We could add another field to the `install-config.yaml` for users to input the Plan's publisher, but I think that just makes the user experience more confusing.

pkg/asset/machines/master.go
```diff
@@ -369,6 +369,18 @@ func (m *Master) Generate(dependencies asset.Parents) error {
 			}
 		}
 
+		if mpool.OSImage.Publisher != "" {
```
sadasu, line 372: So, what happens when mpool.OSImage.Publisher == "" ? Can it be empty/unset? Looking at https://github.com/openshift/installer/blob/master/data/data/install.openshift.io_installconfigs.yaml#L350, it seems like a required field.

r4f4: We cannot remove it, since we need that value to retrieve the image with the API: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/azure/client.go#L325-L334.

What would the INFO message accomplish in this case?

patrickdillon:
This solution is in-line with upstream code in the cluster api provider. Short of extending the machine API and reworking the cluster API provider, which would be far more work for little benefit, this seems like the best solution. 

/approve

PR #7428, opened 2023-08-16
OCPBUGS-13664: Add KMS encryption keys if provided

pawanpinjarkar:
If users have specified the  KMS encryption keys in the install config, 
either in the master machine pool or the aws machine pool, 
then add the required permission group to the permission check.

### DIFF
--- a/pkg/asset/installconfig/aws/permissions.go
+++ b/pkg/asset/installconfig/aws/permissions.go
@@ -38,6 +38,9 @@ const (
 
 	// PermissionDeleteHostedZone is a set of permissions required when the installer destroys a route53 hosted zone.
 	PermissionDeleteHostedZone PermissionGroup = "delete-hosted-zone"
+
+	// PermissionKMSEncryptionKeys is an additional set of permissions required when the installer uses user provided kms encryption keys.
+	PermissionKMSEncryptionKeys PermissionGroup = "kms-encryption-keys"
 )
 
 var permissions = map[PermissionGroup][]string{
@@ -243,6 +246,16 @@ var permissions = map[PermissionGroup][]string{
 	},
 	PermissionDeleteHostedZone: {
 		"route53:DeleteHostedZone",
+	},
+	PermissionKMSEncryptionKeys: {
+		"kms:Decrypt",
+		"kms:Encrypt",
+		"kms:GenerateDataKey",
+		"kms:GenerateDataKeyWithoutPlainText",
+		"kms:DescribeKey",
+		"kms:RevokeGrant",
+		"kms:CreateGrant",
+		"kms:ListGrants",
 	},
 }
 
--- a/pkg/asset/installconfig/platformpermscheck.go
+++ b/pkg/asset/installconfig/platformpermscheck.go
@@ -5,11 +5,13 @@ import (
 	"fmt"
 
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 
 	"github.com/openshift/installer/pkg/asset"
 	awsconfig "github.com/openshift/installer/pkg/asset/installconfig/aws"
 	gcpconfig "github.com/openshift/installer/pkg/asset/installconfig/gcp"
 	powervsconfig "github.com/openshift/installer/pkg/asset/installconfig/powervs"
+	"github.com/openshift/installer/pkg/types"
 	"github.com/openshift/installer/pkg/types/alibabacloud"
 	"github.com/openshift/installer/pkg/types/aws"
 	"github.com/openshift/installer/pkg/types/azure"
@@ -47,8 +49,10 @@ func (a *PlatformPermsCheck) Generate(dependencies asset.Parents) error {
 	dependencies.Get(ic)
 
 	if ic.Config.CredentialsMode != "" {
+		logrus.Debug("CredentialsMode is set. Skipping platform permissions checks before attempting installation.")
 		return nil
 	}
+	logrus.Debug("CredentialsMode is not set. Performing platform permissions checks before attempting installation.")
 
 	var err error
 	platform := ic.Config.Platform.Name()
@@ -64,6 +68,20 @@ func (a *PlatformPermsCheck) Generate(dependencies asset.Parents) error {
 
 		if !usingExistingPrivateZone {
 			permissionGroups = append(permissionGroups, awsconfig.PermissionCreateHostedZone)
+		}
+
+		var ec2RootVolume = aws.EC2RootVolume{}
+		var awsMachinePoolUsingKMS, masterMachinePoolUsingKMS bool
+		if ic.Config.AWS.DefaultMachinePlatform != nil && ic.Config.AWS.DefaultMachinePlatform.EC2RootVolume != ec2RootVolume {
+			awsMachinePoolUsingKMS = len(ic.Config.AWS.DefaultMachinePlatform.EC2RootVolume.KMSKeyARN) != 0
+		}
+		if ic.Config.ControlPlane.Name == types.MachinePoolControlPlaneRoleName && ic.Config.ControlPlane.Platform.AWS.EC2RootVolume != ec2RootVolume {
+			masterMachinePoolUsingKMS = len(ic.Config.ControlPlane.Platform.AWS.EC2RootVolume.KMSKeyARN) != 0
+		}
+		// Add KMS encryption keys, if provided.
+		if awsMachinePoolUsingKMS || masterMachinePoolUsingKMS {
+			logrus.Debugf("Adding %s to the group of permissions to validate", awsconfig.PermissionKMSEncryptionKeys)
+			permissionGroups = append(permissionGroups, awsconfig.PermissionKMSEncryptionKeys)
 		}
 
 		// Add delete permissions for non-C2S installs.


pkg/asset/installconfig/aws/permissions.go
```diff
@@ -38,6 +38,9 @@ const (
 
 	// PermissionDeleteHostedZone is a set of permissions required when the installer destroys a route53 hosted zone.
 	PermissionDeleteHostedZone PermissionGroup = "delete-hosted-zone"
+
+	// PermissionKMSEncryptionKeys is an additional set of permissions required when the installer uses user provided kms encryption keys.
+	PermissionKMSEncryptionKeys PermissionGroup = "kms-encryption-keys"
```
barbacbd, line 43: The permission group should cover a set of permissions. I don't see this one added to the permissions map below.

pkg/asset/installconfig/platformpermscheck.go
```diff
@@ -56,12 +56,17 @@ func (a *PlatformPermsCheck) Generate(dependencies asset.Parents) error {
...
 		usingExistingVPC := len(ic.Config.AWS.Subnets) != 0
+		usingKMS := len(ic.Config.AWS.DefaultMachinePlatform.EC2RootVolume.KMSKeyARN) != 0
 		usingExistingPrivateZone := len(ic.Config.AWS.HostedZone) != 0
 
 		if !usingExistingVPC {
 			permissionGroups = append(permissionGroups, awsconfig.PermissionCreateNetworking)
 		}
 
+		if usingKMS {
+			permissionGroups = append(permissionGroups, awsconfig.PermissionKMSEncryptionKeys)
```
barbacbd, line 67: Adding here will only check when the credentials mode is in default or for AWS that is mint mode. Do we also need to check passthrough mode ?

pkg/asset/installconfig/aws/permissions.go
pawanpinjarkar: The list of permissions is yet to be finalized.   https://issues.redhat.com/browse/OCPBUGS-13664?focusedId=22816098&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22816098

pawanpinjarkar: Added the permissions

pkg/asset/installconfig/platformpermscheck.go
```diff
@@ -56,12 +56,17 @@ func (a *PlatformPermsCheck) Generate(dependencies asset.Parents) error {
 	case aws.Name:
 		permissionGroups := []awsconfig.PermissionGroup{awsconfig.PermissionCreateBase}
 		usingExistingVPC := len(ic.Config.AWS.Subnets) != 0
+		usingKMS := len(ic.Config.AWS.DefaultMachinePlatform.EC2RootVolume.KMSKeyARN) != 0
```
patrickdillon, line 59: 1. I believe both `DefaultMachinePlatform` and `EC2RootVolume` could be `nil`, which is probably why the e2e tests are failing
2. You will probably need to check the other machine pools (master and worker)--not just the default machine pool.

patrickdillon (commented):
This approach looks good to me. I left a few comments for details that need to be resolved. Is there a way to test this?

pkg/asset/installconfig/platformpermscheck.go
pawanpinjarkar: @patrickdillon ^^

patrickdillon: That's a good point, but for the scope of this bug this is fine. Just make sure to mention to QE that the credential validation check is only triggered when using the default mode.

pkg/asset/installconfig/platformpermscheck.go
patrickdillon: > I believe both DefaultMachinePlatform and EC2RootVolume could be nil, which is probably why the e2e tests are failing

Isn't this still a problem? Won't we have a nil pointer derefence if `DefaultMachinePlatform` or `EC2RootVolume` is nil? 

> You will probably need to check the other machine pools (master and worker)--not just the default machine pool.

Still need to check this. Actually, we just need to check the master machine pool. Worker permissions would be handled by operators.

pkg/asset/installconfig/platformpermscheck.go
sadasu: > Adding here will only check when the credentials mode is in default or for AWS that is mint mode. Do we also need to check passthrough mode ?

@barbacbd , looking at https://github.com/openshift/installer/pull/7428/files#diff-8e36b5854a90a47985a78aa607f0ac89636aae3231ec21e55b690470e151739eL49-L51, it appears that the credentialsMode should be "".  `ic.Config.CredentialsMode` should be populated with its default value by now, correct?

sadasu: Nit: @pawanpinjarkar can you add this new code for determining value of usingKMS here: https://github.com/openshift/installer/blob/master/pkg/asset/installconfig/platformpermscheck.go#L68. Better to avoid interspersing with existing code for other permissionGroups.

barbacbd: Yes it would be populated with the default. In AWS the default ("") is the same as Mint Mode. That is fine, but I am curious if these checks also _should_ be run when using pass through. Manual mode being skipped seems fine.

barbacbd: The other issue would be that this check wouldn't be hit when the user explicitly adds `credentialsMode: Mint` to the install config

sadasu: > The other issue would be that this check wouldn't be hit when the user explicitly adds `credentialsMode: Mint` to the install config

From the description here: https://github.com/openshift/installer/blob/master/data/data/install.openshift.io_installconfigs.yaml#L1973, it appears, that is the intended behavior.

pkg/asset/installconfig/platformpermscheck.go
```diff
@@ -66,6 +70,22 @@ func (a *PlatformPermsCheck) Generate(dependencies asset.Parents) error {
 			permissionGroups = append(permissionGroups, awsconfig.PermissionCreateHostedZone)
 		}
 
+		var ec2RootVolume = aws.EC2RootVolume{}
+		var awsMachinePoolUsingKMS, masterMachinePoolUsingKMS bool
+		if ic.Config.AWS.DefaultMachinePlatform != nil && ic.Config.AWS.DefaultMachinePlatform.EC2RootVolume != ec2RootVolume {
+			logrus.Debug("Using KMS key from default machine pool")
```
sadasu, line 76: if len(ic.Config.AWS.DefaultMachinePlatform.EC2RootVolume.KMSKeyARN) == 0, we would not be using KMS key from this machine pool. So, I think it is OK to remove above debug msg.

pkg/asset/installconfig/platformpermscheck.go
```diff
@@ -66,6 +70,22 @@ func (a *PlatformPermsCheck) Generate(dependencies asset.Parents) error {
...
+			awsMachinePoolUsingKMS = len(ic.Config.AWS.DefaultMachinePlatform.EC2RootVolume.KMSKeyARN) != 0
+		}
+		if ic.Config.ControlPlane.Name == types.MachinePoolControlPlaneRoleName && ic.Config.ControlPlane.Platform.AWS.EC2RootVolume != ec2RootVolume {
+			logrus.Debug("Using KMS key from master machine pool")
```
sadasu, line 80: Similar to https://github.com/openshift/installer/pull/7428/files#r1354111804

sadasu:
Best practice would be to cherry pick to just the release prior and then issue another cherry-pick from that PR to the branch before that.

pkg/asset/installconfig/platformpermscheck.go
barbacbd: @sadasu, @patrickdillon and I discussed this already. There is actually an enhancement that "describes" this feature. Neither of us were 100% on it. I think this is worth discussing but that is up to you. If you believe that this is good then push it for merge.

pawanpinjarkar: @barbacbd I think we can go ahead with this given the above discussions( will need LGTM label) and someone from the installer team can prioritize the enhancement when feasible. Currently, I do not know what that enhancement is and when it is planned for.

barbacbd (commented):
/lgtm

pkg/asset/installconfig/platformpermscheck.go
sadasu: > @sadasu, @patrickdillon and I discussed this already. There is actually an enhancement that "describes" this feature. Neither of us were 100% on it. I think this is worth discussing but that is up to you. If you believe that this is good then push it for merge.

The enhancement referred to here is https://github.com/openshift/enhancements/blob/master/enhancements/installer/aws-permissions-check-bypass.md. It proposes adding a config object to allow modifying the runtime behavior of CCO. This fix takes care of the problem the enhancement is looking to solve without the ability to modify runtime behavior on day-2. So, there is value in merging this now.

PR #7429, opened 2023-08-16
OCPBUGS-17724: Graceful fail for AWS getUser on destroy

barbacbd:
**The getUser role validation can be skipped during manual mode installations. During the uninstall process, fail gracefully when the role is not found.

### DIFF
--- a/pkg/destroy/aws/iamhelpers.go
+++ b/pkg/destroy/aws/iamhelpers.go
@@ -112,13 +112,24 @@ func (search *iamUserSearch) arns(ctx context.Context) ([]string, error) {
 				// Unfortunately user.Tags is empty from ListUsers, so we need to query each one
 				response, err := search.client.GetUserWithContext(ctx, &iam.GetUserInput{UserName: aws.String(*user.UserName)})
 				if err != nil {
-					if err.(awserr.Error).Code() == iam.ErrCodeNoSuchEntityException {
-						search.unmatched[*user.Arn] = exists
-					} else {
-						if lastError != nil {
-							search.logger.Debug(lastError)
+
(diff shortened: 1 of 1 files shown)


pkg/destroy/aws/iamhelpers.go
```diff
@@ -112,13 +112,24 @@ func (search *iamUserSearch) arns(ctx context.Context) ([]string, error) {
...
+					var awsErr awserr.Error
+					if errors.As(err, &awsErr) {
+						switch {
+						case awsErr.Code() == iam.ErrCodeNoSuchEntityException:
+							// The role does not exist.
+							// Ignore this IAM Role and do not report this error via lastError.
+							search.unmatched[*user.Arn] = exists
+						case strings.Contains(err.Error(), "AccessDenied"):
+							// Installer does not have access to this IAM role.
+							// Ignore this IAM Role and do not report this error via lastError.
```
rna-afk, line 125: How about a log line here? What if they did provide a credentials in a normal install that was rejected.

sadasu: That log line will appear multiple times in the logs. This fix is based on https://github.com/openshift/installer/pull/7180 and the "Actual results" section of the associated issue https://issues.redhat.com//browse/OCPBUGS-1769 has sample output.

rna-afk: Some user will in the future definitely want to know why the creds they provided not work. I'm fine if you all are worried about log lines repeating.

sadasu: We can update code to make sure the message is logged only once. It just can't be within the switch block.

PR #7433, opened 2023-08-17
OCPBUGS-7690: azure: destroy: dns records leak if permissions missing

r4f4:
If the Service Principal destroying an Azure cluster lacks the permissions to list DNS records, the Installer will complete the destroy but leave public DNS records behind.

Unfortunately the Azure SDK doesn't return any errors when permissions are missing neither does it offer a way for us to check if the required permissions are present. As an in-between solution, let's log a warning when we can't find any DNS records so users know to check for any records being leaked.

### DIFF
--- a/pkg/destroy/azure/azure.go
+++ b/pkg/destroy/azure/azure.go
@@ -314,11 +314,13 @@ func deletePublicRecords(ctx context.Context, dnsClient dns.ZonesClient, records
 		}
 	}
 
+	pageCount := 0
 	for ; zonesPage.NotDone(); err = zonesPage.NextWithContext(ctx) {
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/destroy/azure/azure.go
```diff
@@ -314,11 +314,13 @@ func deletePublicRecords(ctx context.Context, dnsClient dns.ZonesClient, records
 		}
 	}
 
+	pageCount := 0
```
barbacbd, line 317: Can we separate the public and private records to create a warning for each?

r4f4:
/override ci/prow/e2e-aws-ovn
Failed due to lack of available aws instances. Skipping this since it's not affected by the PR and retesting it will just add strain to CI.

PR #7436, opened 2023-08-17
OCPBUGS-17861: cinder volume type/az validation

EmilienM:
For each rootVolume type specified that has zones configured in Cinder,
check that the zones were specified in the machine pool with the rootVolume zones field.

See https://docs.openstack.org/cinder/latest/admin/availability-zone-type.html for more information.

Note that this is not a complete check, as if more than one type is specified,
the zones could be specified in another index of the rootVolume zones field.
For that we should add some validation that zones found in a type at a
specific index, matches the zone name of that index in root volume zones
field.

### DIFF
--- a/pkg/asset/installconfig/openstack/validation/cloudinfo.go
+++ b/pkg/asset/installconfig/openstack/validation/cloudinfo.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"strings"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/availabilityzones"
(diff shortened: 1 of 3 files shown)


EmilienM:
I'm unsure about this PR anymore because it seems that volume types and volume backends are somewhat orthogonal. We can indeed specify a volume type be valid in a list of AZs, but we also need to ensure there's a volume service in each of those AZs; which is not possible to verify as of today.

Also see https://github.com/openstack/cinder-specs/blob/master/specs/xena/expose-cinder-user-visible-extra-specs-spec.rst

EmilienM:
/close

Closing this one for now, when setting an AZ to a volume type with RESKEY:availability_zones, there is no validation check in place as of now.
The validation happens later when we try to create a volume from it, the API checks if there is an intersection between the availability zone specified in the volume type and the the availability zone set in the active volume services.

I think for now this is not very useful to have more validations in the installer.

PR #7439, opened 2023-08-22
AGENT-692 OCPBUGS-3860: Update assisted-service dependencies

zaneb:
Depend only on the new `github.com/openshift/assisted-service/client` go module, along with the existing `/api` and `/models` modules, and drop the dependency on the main assisted-service module. This allows us to update to the latest versions of all of these. Previously this was not possible because assisted-service now depends on the rust nmstate bindings, which we don't yet have a way to build on all platforms.

In order to drop the dependency, we have to inline the old code for shelling out to `nmstatectl`, and stop relying on some library convenience functions.


pkg/asset/agent/image/oc.go
```diff
@@ -323,13 +320,21 @@ func execute(executer executer.Executer, pullSecret, command string) (string, er
...
-	stdout, stderr, exitCode := executer.Execute(args[0], args[1:]...)
+	var stdoutBytes, stderrBytes bytes.Buffer
+	cmd := exec.Command(args[0], args[1:]...)
+	cmd.Stdout = &stdoutBytes
+	cmd.Stderr = &stderrBytes
 
-	if exitCode == 0 {
-		return strings.TrimSpace(stdout), nil
+	err = cmd.Run()
+	if err != nil {
```
rwsu, line 329:
```suggestion
	if err == nil {
```

rwsu (changes requested):

pkg/asset/agent/image/oc.go
bfournie, line 329: I think this should be:
`if err == nil {`

According to example https://pkg.go.dev/os/exec#Command

bfournie: With that change I was able to successfully install with `4.14.0-0.ci-2023-08-22-121303`

bfournie: Duplicate comments :-)

pkg/asset/agent/manifests/staticnetworkconfig/generator.go
```diff
@@ -141,28 +155,32 @@ func (s *StaticNetworkConfigGenerator) createNMConnectionFiles(nmstateOutput, ho
...
-		if err != nil {
-			return nil, err
-		}
-		s.log.Infof("Adding NMConnection file <%s>", fileName)
-		newFile := StaticNetworkConfigData{
-			FilePath:     filepath.Join(hostDir, fileName),
-			FileContents: fileContents,
+		if connectionElems, ok := connection.([]interface{}); ok {
+			if fileName, ok := connectionElems[0].(string); ok {
+				if nmConnection, ok := connectionElems[1].(string); ok {
```
andfasano, line 165: Are these linter-driven changes?

rwsu (approved):
/approve

PR #7440, opened 2023-08-22
OCPBUGS-17851: Update Azure zone to non-pointer

RadekManak:
Azure's AzureMachineProviderSpec Zone field was updated to be a non-pointer. This PR updates its usage to be compatible again.


r4f4 (commented):
/lgtm

r4f4:
/retest-required
CI should be more stable now

RadekManak:
/hold

We have reverted https://github.com/openshift/machine-api-provider-azure/pull/72. We need to reconsider the implications of this change. This will probably not ship in 4.14.

PR #7442, opened 2023-08-22
AGENT-693: Support external platform

pawanpinjarkar:
Support the new `external` platform type

- Added `external` to the supported platforms
- When platform type is `external`, only valid platform name is `oci`
- Added integration tests for external platform (SNO and Compact)
- Added additional integration test for invalid network type in case of SNO
- Updated unit tests for newly added `external` platform


pawanpinjarkar:
depends on https://github.com/openshift/assisted-service/pull/5438
/hold
/cc @zaneb @andfasano @bfournie

pkg/asset/agent/common.go
```diff
@@ -34,6 +36,8 @@ func HivePlatformType(platform types.Platform) hiveext.PlatformType {
 	switch platform.Name() {
 	case baremetal.Name:
 		return hiveext.BareMetalPlatformType
+	case external.Name:
+		return hiveext.ExternalPlatformType
```
zaneb, line 40: We need something that will get converted to `oci` by this function: https://github.com/openshift/assisted-service/blob/master/internal/controller/controllers/clusterdeployments_controller.go#L691-L708

That means that:
1. we will have to modify that function. (this is a bit of a weird case, because ZTP does not - and probably will never - support oci as a platform, but we need a way to represent it in this ZTP manifest).
2. in the short term, this is more likely to be `OCI` than `External`, because to support a generic external platform in assisted we'd need to add another field to pass the `platformName` as well.

cmd/openshift-install/testdata/agent/image/validations/sno_invalid_platform.txt
```diff
@@ -0,0 +1,41 @@
+! exec openshift-install agent create image --dir $WORK
+
+stderr 'Only platform none supports 1 ControlPlane and 0 Compute nodes'
```
pawanpinjarkar, line 3: This needs to be fixed and the agent installer should allow SNO in case of external platform as well , similar to assisted installer

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -276,6 +277,8 @@ func (a *AgentClusterInstall) Load(f asset.FileFetcher) (bool, error) {
 	switch string(agentClusterInstall.Spec.PlatformType) {
 	case baremetal.Name:
 		agentClusterInstall.Spec.PlatformType = hiveext.BareMetalPlatformType
+	case external.Name:
+		agentClusterInstall.Spec.PlatformType = hiveext.ExternalPlatformType
```
zaneb, line 281: Based on the latest version of https://github.com/openshift/assisted-service/pull/5438 we'll need to set external.platformName here as well.

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -156,8 +157,8 @@ func (a *AgentClusterInstall) Generate(dependencies asset.Parents) error {
 			},
```
zaneb, line 157: Based on the latest version of https://github.com/openshift/assisted-service/pull/5438, we also need to set `agentClusterInstall.Spec.External.PlatformName = installConfig.Config.Platform.External.PlatformName` somewhere in `Generate()`.

pawanpinjarkar: Added https://github.com/openshift/installer/pull/7442/files#diff-8c25c4539e452f6075222c27d475ff8586f3843ea64d79ace077c3a3062b3f8aR160-R165

pawanpinjarkar (commented):
Done

zaneb:
This looks good to me as soon as we are able to revendor with https://github.com/openshift/assisted-service/pull/5438

rwsu (changes requested):

rwsu (approved):
/approve

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -156,8 +157,14 @@ func (a *AgentClusterInstall) Generate(dependencies asset.Parents) error {
...
 		}
 
-		if installConfig.Config.Platform.Name() == none.Name {
-			logrus.Debugf("Setting UserManagedNetworking to true for %s platform", none.Name)
+		if installConfig.Config.Platform.Name() == external.Name {
```
zaneb, line 160: This works if the user started with an install-config, but if they started with ZTP manifests it won't:

```suggestion
		if agentClusterInstall.Spec.PlatformType == hiveext.ExternalPlatformType {
```

pkg/asset/agent/manifests/agentclusterinstall_test.go
```diff
@@ -625,7 +625,7 @@ spec:
   sshPublicKey: |
     ssh-rsa AAAAmyKey`,
 			expectedFound: false,
-			expectedError: "invalid PlatformType configured: spec.platformType: Unsupported value: \"aws\": supported values: \"BareMetal\", \"VSphere\", \"None\"",
+			expectedError: "invalid PlatformType configured: spec.platformType: Unsupported value: \"aws\": supported values: \"BareMetal\", \"VSphere\", \"None\",\"External\"",
```
zaneb, line 628:
```suggestion
			expectedError: "invalid PlatformType configured: spec.platformType: Unsupported value: \"aws\": supported values: \"BareMetal\", \"VSphere\", \"None\", \"External\"",
```

cmd/openshift-install/testdata/agent/image/configurations/external_compact.txt
```diff
@@ -0,0 +1,43 @@
+# Verify a default configuration for the compact topology on external
+
+exec openshift-install agent create image --dir $WORK
```
rwsu, line 3: It would be nice to use isocmp to also check the agent-cluster-install.yaml file in the ISO also contains the external type and external platform name.

pawanpinjarkar: Done

PR #7443, opened 2023-08-22
OCPBUGS-5471: Try to select rendezvousIP among non-worker hosts

andfasano:
This patch enhances the rendezvousIP selection logic when it's not explicitly defined in the agent-config.yaml. In such case, it tries to look for an eligible IP within the non-worker hosts eventually configured in the agent-config.yaml first. If none is found (there couldn't be any host defined, or defined without nmstate config), it falls back to the previous behavior (by looking at the static configuration potentially loaded from the `nmstateconfig.yaml` file)

### DIFF
--- a/pkg/asset/agent/image/ignition.go
+++ b/pkg/asset/agent/image/ignition.go
@@ -518,12 +518,12 @@ func RetrieveRendezvousIP(agentConfig *agent.Config, nmStateConfigs []*v1beta1.N
 		rendezvousIP = agentConfig.RendezvousIP
 		logrus.Debug("RendezvousIP from the AgentConfig ", rendezvousIP)
 
-	} else if len(nmStateConfigs) > 0 {
-		rendezvousIP, err = manifests.GetNodeZeroIP(nmStateConfigs)
-		logrus.Debug("RendezvousIP from the NMStateConfig ", rendezvousIP)
 	} else {
-		err = errors.New("missing rendezvousIP in agent-config or at least one NMStateConfig manifest")
-		return "", err
+		rendezvousIP, err = manifests.GetNodeZeroIP(agentConfig, nmStateConfigs)
+		if err != nil {
+			return "", errors.Wrap(err, "missing rendezvousIP in agent-config, at least one host networkConfig in agent-config, or at least one NMStateConfig manifest")
+		}
+		logrus.Debug("RendezvousIP from the NMStateConfig ", rendezvousIP)
 	}
 
 	// Convert IPv6 address to canonical to match host format for comparisons
--- a/pkg/asset/agent/manifests/nmstateconfig.go
+++ b/pkg/asset/agent/manifests/nmstateconfig.go
@@ -23,6 +23,7 @@ import (
 	"github.com/openshift/installer/pkg/asset/agent/agentconfig"
 	"github.com/openshift/installer/pkg/asset/agent/manifests/staticnetworkconfig"
 	"github.com/openshift/installer/pkg/types"
+	agenttypes "github.com/openshift/installer/pkg/types/agent"
 )
 
 var (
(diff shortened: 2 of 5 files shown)


pkg/asset/agent/image/ignition.go
```diff
@@ -507,12 +507,12 @@ func RetrieveRendezvousIP(agentConfig *agent.Config, nmStateConfigs []*v1beta1.N
...
 
-	} else if len(nmStateConfigs) > 0 {
-		rendezvousIP, err = manifests.GetNodeZeroIP(nmStateConfigs)
-		logrus.Debug("RendezvousIP from the NMStateConfig ", rendezvousIP)
 	} else {
-		err = errors.New("missing rendezvousIP in agent-config or at least one NMStateConfig manifest")
-		return "", err
+		rendezvousIP, err = manifests.GetNodeZeroIP(agentConfig, nmStateConfigs)
+		if err != nil {
+			return "", errors.Wrap(err, "missing rendezvousIP in agent-config or at least one NMStateConfig manifest")
```
rwsu, line 513: I think the error should say: "missing rendezvousIP in agent-config, at least one host networkConfig in agent-config, or at least one NMStateConfig manifest"

rwsu (changes requested):
Looks good. I think one error message should be updated.

pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -242,41 +243,68 @@ func (n *NMStateConfig) validateNMStateLabels() field.ErrorList {
...
+// in the agent-config hosts - so that it would be possible to skip the worker nodes - and then
+// the NMStateConfig.
+func GetNodeZeroIP(agentConfig *agenttypes.Config, nmStateConfigs []*aiv1beta1.NMStateConfig) (string, error) {
+	rawConfigs := []aiv1beta1.RawNetConfig{}
+
+	// Select first the configs from the hosts, if defined
+	if agentConfig != nil {
+		for _, host := range agentConfig.Hosts {
+			// Skip worker hosts
+			if host.Role == "worker" {
```
zaneb, line 280: This is not enough, because say you have a 3+2 HA cluster and the assigned roles are: `"", "", "master", "master", "master"`. In this case the first two hosts _must_ be workers, but we will still select the IP from the first one.

pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -242,41 +243,68 @@ func (n *NMStateConfig) validateNMStateLabels() field.ErrorList {
...
+	}
+
+	// Then pick up the ones from nmStateConfigs, if defined
+	for _, nmStateConfig := range nmStateConfigs {
+		rawConfigs = append(rawConfigs, nmStateConfig.Spec.NetConfig.Raw)
+	}
+
+	// Try to look for an eligible IP
+	for _, raw := range rawConfigs {
+		nodeZeroIP, err := getFirstIP(raw)
```
zaneb, line 294: Arguably we should search all valid hosts for an IPv4 address before going back and searching for an IPv6 address only if none of them have IPv4.
In practice it probably doesn't matter though.

pkg/asset/agent/manifests/nmstateconfig.go
andfasano: Here I'd like to avoid here complex logic as we did in the `validateHostCount` (which I originally implemented in the baremetal part), in favor of a simpler best effort approach, oriented to promote the explicit configuration when available. 
We could change the implementation to include only those hosts having the `role` field equal to `master` (thus ignoring the `workers` and the ones not defined) - and it's fine for me, even though there would be always a partial configuration (the original sin) setup that wouldn't be fine (`"", "", "worker", "worker", "worker"`).
It'd be anyhow something easy to explain in the documentation, ie when the rendezvousIP is not defined then it will be choose one of the static IP configured among the master hosts only.

bfournie: I also think you should only include those hosts having role equal to `master` (ignoring `workers` and those not defined)

pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -242,41 +243,68 @@ func (n *NMStateConfig) validateNMStateLabels() field.ErrorList {
...
+		for _, host := range agentConfig.Hosts {
+			// Skip worker hosts
+			if host.Role == "worker" {
+				continue
+			}
+			rawConfigs = append(rawConfigs, host.NetworkConfig.Raw)
+		}
+	}
+
+	// Then pick up the ones from nmStateConfigs, if defined
```
bfournie, line 287: Although probably not a problem, it seems that you should have a check here to ignore the settings in nmstateconfig.yaml if agent-config.yaml had the nmstate configuration, in other words `if rawConfigs > 0`.

pkg/asset/agent/manifests/nmstateconfig.go
andfasano: Done

pkg/asset/agent/manifests/nmstateconfig.go
```diff
@@ -242,41 +243,68 @@ func (n *NMStateConfig) validateNMStateLabels() field.ErrorList {
...
+	// Select first the configs from the hosts, if defined
+	if agentConfig != nil {
+		for _, host := range agentConfig.Hosts {
+			// Skip worker hosts (or without an explicit role assigned)
+			if host.Role != "master" {
+				continue
+			}
+			rawConfigs = append(rawConfigs, host.NetworkConfig.Raw)
+		}
+	}
```
zaneb, line 285: This could still end up selecting a worker (due to falling back to the nmstateconfigs below) in the case [`"worker"`, `"worker"`, `""`,  `""`, `""`].

This could be solved by adding a second loop after this:

```
if agentConfig != nil {
		for _, host := range agentConfig.Hosts {
			// Skip worker hosts
			if host.Role != "" {
				continue
			}
			rawConfigs = append(rawConfigs, host.NetworkConfig.Raw)
		}
	}
```

That would cover almost all cases - everything except the case where the workers have implicit role assignment, but only workers have network configs with static IPs. Complete correctness could be achieved by doing this conditionally only when we didn't already find enough explicit masters to equal the number of control plane replicas.

pkg/asset/agent/manifests/nmstateconfig.go
zaneb: The tricky thing here is that nmstateconfig is both an alternative input to agent-config _and_ a sequential step in a series of transformations starting with agent-config that the user can edit at any time.

In an ideal world we would match each nmstateconfig back to it's corresponding host in agent-config (if there is one) and respect any role settings provided there. Otherwise we could get inconsistent behaviour where the user has edited the nmstateconfig but those changes aren't reflected in the final IP selection. In practice that would require a _lot_ of very gnarly logic.

I think a good compromise would be to treat the nmstateconfigs strictly as an alternative - i.e. only add them to the list when `agentConfig == nil || len(agentConfig.Hosts) == 0`.

There's still potential for inconsistency if the user does crazy stuff, but when combined with the change I suggested above we'd avoid selecting an IP from a host explicitly marked as a worker, without introducing any regressions.

pkg/asset/agent/manifests/nmstateconfig.go
zaneb: > I'd like to avoid here complex logic

The complexity doesn't go away, the question is whether we handle it or push it onto the user.

pkg/asset/agent/manifests/nmstateconfig.go
andfasano: Yeah I think it's working (I've added another test for that) (assuming you meant except for the case with the implict masters). I'm still even more convinced that if we had opted (initially) for mandatory roles the overall logic could have been simpler, also in the validations.

pawanpinjarkar (commented):
/lgtm

sadasu:
Is this waiting on https://github.com/openshift/installer/pull/7591?

andfasano:
> Is this waiting on #7591?

Not strictly (let me re-kick off another run), it'd require another pass of reviews.

/retest-required

PR #7445, opened 2023-08-24
Implement workaround to allow SNO installations for OKD/FCOS

JM1:
OKD/FCOS uses FCOS as its bootimage, i.e. when booting cluster nodes the first time during installation. FCOS does not provide tools such as OpenShift Client (`oc`) or hyperkube which are used during single-node cluster installation at first boot (e.g. `oc` in [bootkube.sh](https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template)). RHCOS and SCOS include these tools, but FCOS has to [pivot the root fs](https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template) to [okd-machine-os](https://github.com/openshift/okd-machine-os) first in order to make those tools available.

Pivoting uses `rpm-ostree rebase` but during SNO installation the node will be booted from a FCOS Live ISO where the root fs and `/sysroot` are mounted read-only. Thus `rpm-ostree rebase` fails and necessary tools for SNO installation will not be available, causing the setup to stall.

This patch works around this issue by mounting `/usr` and `/etc` as overlay filesystems which contains both the content of the live iso as well as from the okd-machine-os container image.

### DIFF
--- a/data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
+++ b/data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
@@ -42,10 +42,25 @@ if [ ! -f /opt/openshift/.pivot-done ]; then
   record_service_stage_start "rebase-to-okd-os-image"
 {{if .IsFCOS -}}
   mnt="$(podman image mount "${MACHINE_OS_IMAGE}")"
+{{- if .BootstrapInPlace }}
+  # SNO setup boots into Live ISO which cannot be rebased
+  # https://github.com/coreos/rpm-ostree/issues/4547
+  mkdir /var/mnt/{upper,worker}
+  mount -t overlay overlay -o "lowerdir=/usr:$mnt/usr" /usr
+  mount -t overlay overlay -o "lowerdir=/etc:$mnt/etc,upperdir=/var/mnt/upper,workdir=/var/mnt/worker" /etc
+  systemctl daemon-reload
+
+  # Workaround for SELinux denials when launching crio.service from overlayfs
+  setenforce Permissive
+
+  systemctl start crio.service
+  # No reboot necessary because SNO setup will reboot system
+{{ else }}
   pushd "${mnt}/bootstrap"
     # shellcheck disable=SC1091
     . ./pre-pivot.sh
   popd
+{{ end -}}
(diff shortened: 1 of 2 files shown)


LorbusChris:
Maybe add a comment in-code that this is supposed to be temporary, and ideally also add a link to an upstream issue.
Other than that, LGTM

JM1:
> Maybe add a comment in-code that this is supposed to be temporary, and ideally also add a link to an upstream issue. Other than that, LGTM

There is no upstream issue which is why i linked the [RFE about allowing rpm-ostree rebase on live iso](https://github.com/coreos/rpm-ostree/issues/4547). I dare to write temporary because a proper solution is not easy and it might turn into something which is perma-temporary 😅

JM1:
OpenShift's conformance test suite [fails with](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_installer/7445/pull-ci-openshift-installer-master-okd-e2e-aws-ovn/1694690863638122496/artifacts/e2e-aws-ovn/openshift-e2e-test/artifacts/e2e.log):
```
Flaky tests:

[sig-cli] oc idle [apigroup:apps.openshift.io][apigroup:route.openshift.io][apigroup:project.openshift.io][apigroup:image.openshift.io] by checking previous scale [Suite:openshift/conformance/parallel]
[sig-cli] oc idle [apigroup:apps.openshift.io][apigroup:route.openshift.io][apigroup:project.openshift.io][apigroup:image.openshift.io] by label [Suite:openshift/conformance/parallel]

Failing tests:

[sig-api-machinery] Aggregator Should be able to support the 1.17 Sample API Server using the current Aggregator [Conformance] [Suite:openshift/conformance/parallel/minimal] [Suite:k8s]
```
Seems to be unrelated to this change. Is there anything we can do except for `/retest-required`?

LorbusChris:
I'd like to see one of the OKD/FCOS E2Es go green.
/retest

LorbusChris:
LGTM, I'm in favor of merging this now as it only affects OKD, and enables SNO installs for OKD 1.14.

PR #7449, opened 2023-08-26
Master update terraform golang

jhixson74:
Update terraform go modules to use v1.20.

https://issues.redhat.com/browse/CORS-2695

### DIFF
--- a/terraform/providers/alicloud/go.mod
+++ b/terraform/providers/alicloud/go.mod
@@ -1,6 +1,6 @@
 module github.com/openshift/installer/terraform/providers/alicloud
 
-go 1.18
+go 1.20
 
 require github.com/aliyun/terraform-provider-alicloud v1.148.0
 
--- a/terraform/providers/aws/go.mod
+++ b/terraform/providers/aws/go.mod
@@ -1,6 +1,6 @@
 module github.com/openshift/installer/terraform/providers/aws
 
-go 1.18
+go 1.20
 
 require github.com/hashicorp/terraform-provider-aws v1.60.1-0.20230615210323-24881e9f4460 // v5.4.0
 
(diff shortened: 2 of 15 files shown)


terraform/Makefile
```diff
@@ -25,7 +23,7 @@ go-mod-tidy-vendor: $(GO_MOD_TIDY_TARGETS)
 # The -compat=1.18 is needed for the openstack provider. The provider uses golang.org/x/mod, which go 1.18 selects
 # as v0.3.0 but go 1.16 selects as v0.4.2.
 $(GO_MOD_TIDY_TARGETS): go-mod-tidy-vendor.%:
-	cd providers/$* && go mod tidy -compat=1.18 && go mod vendor
+	cd providers/$* && go mod edit -go=$(GOVERSION) && go mod tidy -compat=1.18 && go mod vendor
```
r4f4, line 26: Should we also bump the compat version?
```suggestion
	cd providers/$* && go mod edit -go=$(GOVERSION) && go mod tidy -compat=$(GOVERSION) && go mod vendor
```

jhixson74: No idea. I was going to ask the guys that did this if it should be bumped. I will find out. I am currently not familiar with what this actually does.

sadasu:
/override ci/prow/e2e-alibaba

Hasn't passed since April this year.

patrickdillon:
It's looking more and more likely that we will be trying to remove TF as quickly as possible. Considering that, would it still make sense to add the Makefile?

patrickdillon:
The top-level makefile warrants its own PR, because the Go version update should be a quick merge, but the makefile should probably be reviewed more carefully by the team.

sadasu:
@jhixson74 this work may have been completed as part of https://github.com/openshift/installer/pull/7590. Could you PTAL?

PR #6770, opened 2023-01-12
SPLAT-823: Zonal GA, single terraform, deprecate legacy platform spec

jcpowermac:
This PR does the following:

- Converts existing install-config vsphere platform spec for IPI, UPI and 4.12 Zonal IPI to 4.13 unified non-depreciated platform fields
- Replaces the existing non zonal terraform project with zonal as the unified
- vsphere-specific fields are now required to be paths 
- Remove OVA hardware version upgrade within terraform resource ova import (no longer required, OVA is HW15)
- Add check for vSphere 7 Update 2 or greater being required in 4.13

Left to do:

- [x] Update unit tests 
- [x] Clean up `todo: jcallen:` items
- [x] make sure TUI functions correctly

PR #7451, opened 2023-08-29
OCPBUGS-18304: for vsphere ipi add cluster domain to the uploaded vm configs so that…

vrutkovs:
… 30-local-dns-prepender can use it

Fixes https://github.com/okd-project/okd/issues/1715

This re-adds domain setting from https://github.com/openshift/installer/pull/5788 removed in https://github.com/openshift/installer/pull/6770


jcpowermac:
not quite sure why that failed...
/test e2e-vsphere-ovn

PR #7452, opened 2023-08-29
OCPBUGS-5728: Log "agent wait-for" commands to .openshift_install.log

rwsu:
Updated "agent wait-for" commands to also call
SetupFileHook(RootOpts.Dir) so that it logs to the common .openshift_install.log file.

Moved log.go and rootOpts to a command package so that it can be shared with agent wait-for.

### DIFF
--- a/pkg/agent/cluster.go
+++ b/pkg/agent/cluster.go
@@ -126,19 +126,13 @@ func (czero *Cluster) IsBootstrapComplete() (bool, bool, error) {
 		return true, false, nil
 	}
 
-	clusterKubeAPILive, clusterKubeAPIErr := czero.API.Kube.IsKubeAPILive()
-	if clusterKubeAPIErr != nil {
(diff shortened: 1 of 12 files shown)


pkg/agent/cluster.go
```diff
@@ -126,19 +126,13 @@ func (czero *Cluster) IsBootstrapComplete() (bool, bool, error) {
 		return true, false, nil
 	}
 
-	clusterKubeAPILive, clusterKubeAPIErr := czero.API.Kube.IsKubeAPILive()
-	if clusterKubeAPIErr != nil {
-		logrus.Trace(errors.Wrap(clusterKubeAPIErr, "Bootstrap Kube API is not available"))
-	}
+	clusterKubeAPILive, _ := czero.API.Kube.IsKubeAPILive()
```
bfournie, line 129: Looks like a lint error here

PR #7453, opened 2023-08-29
OCPBUGS-2339: Restart agent.service if assisted-service restarts

rwsu:
The agent service is unable to continue to next step if assisted-service restarts and clears its internal database.

A new service named agent-restart.service now checks if assisted-service goes offline. When it goes offline, the new service restarts agent.service and itself. This forces the agent to reregister with assisted-service and allows installation to continue.

This fix does not handle the situation where the cluster has started installation and assisted-service is then restarted. In that case, all hosts will need to be rebooted.


rwsu:
The e2e-agent-compact-ipv4 job failed in gather step:
```
  * could not run steps: step e2e-agent-compact-ipv4 failed: "e2e-agent-compact-ipv4" post steps failed: "e2e-agent-compact-ipv4" pod "e2e-agent-compact-ipv4-gather-extra" failed: could not watch pod: the pod ci-op-3x4cx19w/e2e-agent-compact-ipv4-gather-extra failed after 24s (failed containers: test): ContainerFailed one or more containers exited

Container test exited with code 1, reason Error
---
speedup is 0.00
[must-gather      ] OUT namespace/openshift-must-gather-5pw4x deleted
[must-gather      ] OUT clusterrolebinding.rbac.authorization.k8s.io/must-gather-lcqsf deleted
```
Retrying again.
/test e2e-agent-compact-ipv4

andfasano:
> The e2e-agent-compact-ipv4 job failed in gather step:
> 
> ```
>   * could not run steps: step e2e-agent-compact-ipv4 failed: "e2e-agent-compact-ipv4" post steps failed: "e2e-agent-compact-ipv4" pod "e2e-agent-compact-ipv4-gather-extra" failed: could not watch pod: the pod ci-op-3x4cx19w/e2e-agent-compact-ipv4-gather-extra failed after 24s (failed containers: test): ContainerFailed one or more containers exited
> 
> Container test exited with code 1, reason Error
> ---
> speedup is 0.00
> [must-gather      ] OUT namespace/openshift-must-gather-5pw4x deleted
> [must-gather      ] OUT clusterrolebinding.rbac.authorization.k8s.io/must-gather-lcqsf deleted
> ```
> 
> Retrying again. /test e2e-agent-compact-ipv4

Recently we had several failures in the gather steps, mostly due operators still progressing

data/data/agent/files/usr/local/bin/restart-agent.sh
```diff
@@ -0,0 +1,33 @@
...
+until [[ $INFRA_ENV_ID == "" || $INFRA_ENV_ID == "null" ]]; do
+    sleep 5
+    >&2 echo "Checking assisted-service is still online"
+    INFRA_ENV_ID=$(curl -s -S "${SERVICE_BASE_URL}/api/assisted-install/v2/infra-envs" | jq -r .[0].id)
+    echo "assisted-service is online with infra-env-id: $INFRA_ENV_ID"
+done
+echo "assisted-service interrupted, restarting agent.service"
+
+sudo systemctl restart agent.service
+sudo systemctl restart agent-restart.service
```
zaneb, line 33: This looks dangerous. Need to use `--no-block` at least, but preferably avoid the need for the script to restart itself using systemd at all.
Restart=always is set, isn't that enough to cause the script to be restarted?

data/data/agent/files/usr/local/bin/restart-agent.sh
```diff
@@ -0,0 +1,33 @@
...
+    >&2 echo "Querying assisted-service for infra-env-id..."
+    INFRA_ENV_ID=$(curl -s -S "${SERVICE_BASE_URL}/api/assisted-install/v2/infra-envs" | jq -r .[0].id)
+done
+echo "Fetched infra-env-id and found: $INFRA_ENV_ID"
+
+
+>&2 echo "Restart agent.service if REST-API becomes unavailable"
+INFRA_ENV_ID="tbd"
+until [[ $INFRA_ENV_ID == "" || $INFRA_ENV_ID == "null" ]]; do
+    sleep 5
```
zaneb, line 25: The idea here is we're hoping to catch it in the interval between when assisted-service restarts and when create-cluster-and-infraenv sets everything back up again?
This should be fine if the renedezvous host is rebooted. It feels a bit racy for just a service restart, although solving the problem for reboot is a huge step forward from today.

zaneb (commented):
I'm wondering if there's a way to exploit the unique properties of the ephemeral image (which are the cause of the problem) to fix the issue in the agent itself.
e.g. if the agent thinks it is registered, but assisted-service has no record of it, then exit with a return code and let systemd restart the agent?

PR #7455, opened 2023-08-30
OCPBUGS-17869: azure: fix setting outboundType

r4f4:
https://github.com/openshift/installer/pull/7312 inadvertently removed the setting of `OutboundType` in the terraform variables, causing the outbound to always be the default `LoadBalancer` even if `NatGateway` is set in the install-config.


barbacbd (commented):
/lgtm

PR #7458, opened 2023-08-31
OCPBUGS-18365: Fix defaulting of userManagedNetworking value

zaneb:
Set default value for `userManagedNetworking` in AgentClusterInstall ZTP manifest regardless of whether the user specified the platform type as `none` or `None`.
Also, return a validation error if the user explicitly specifies `false` when the value must be `true`, instead of overwriting it.

### DIFF
--- a/pkg/asset/agent/manifests/agentclusterinstall.go
+++ b/pkg/asset/agent/manifests/agentclusterinstall.go
@@ -287,7 +287,7 @@ func (a *AgentClusterInstall) Load(f asset.FileFetcher) (bool, error) {
 		agentClusterInstall.Spec.PlatformType = hiveext.ExternalPlatformType
 	case none.Name:
 		agentClusterInstall.Spec.PlatformType = hiveext.NonePlatformType
-		if agentClusterInstall.Spec.Networking.UserManagedNetworking != swag.Bool(true) {
+		if agentClusterInstall.Spec.Networking.UserManagedNetworking == nil || !*agentClusterInstall.Spec.Networking.UserManagedNetworking {
 			logrus.Debugf("Setting UserManagedNetworking to true for %s platform", none.Name)
 			agentClusterInstall.Spec.Networking.UserManagedNetworking = swag.Bool(true)
 		}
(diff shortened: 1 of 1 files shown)


pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -278,14 +278,20 @@ func (a *AgentClusterInstall) Load(f asset.FileFetcher) (bool, error) {
...
-		}
 	case vsphere.Name:
 		agentClusterInstall.Spec.PlatformType = hiveext.VSpherePlatformType
 	}
 
+	// Set the default value for userManagedNetworking, as would be done by the
+	// mutating webhook in ZTP.
+	if agentClusterInstall.Spec.Networking.UserManagedNetworking == nil {
+		switch agentClusterInstall.Spec.PlatformType {
+		case hiveext.NonePlatformType:
```
bfournie, line 289: The linter is complaining about this being a switch statement instead of an 'if'. The switch statement will make sense when adding the `external` platform here, so maybe disable the lint check.

pawanpinjarkar: Yes, this now also needs to be set in case of `external` platform

pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -394,10 +400,19 @@ func (a *AgentClusterInstall) validateIPAddressAndNetworkType() field.ErrorList
...
 
-	fieldPath := field.NewPath("spec", "platformType")
-
 	if a.Config.Spec.PlatformType != "" && !agent.IsSupportedPlatform(a.Config.Spec.PlatformType) {
+		fieldPath := field.NewPath("spec", "platformType")
 		allErrs = append(allErrs, field.NotSupported(fieldPath, a.Config.Spec.PlatformType, agent.SupportedHivePlatforms()))
 	}
+
+	switch a.Config.Spec.PlatformType {
+	case hiveext.NonePlatformType:
```
pawanpinjarkar, line 409: Also for `external`

pawanpinjarkar (changes requested):
UMN needs to be set to true in extenal as well

zaneb:
Rebased and updated to do the same for External platform.

PR #7462, opened 2023-09-03
integration tests: Swap order of diff arguments

zaneb:
Display the actual value as a diff from the expected value, as unit tests with the assert library do, instead of the other way around.


pawanpinjarkar (approved):
/lgtm

zaneb:
Only affects (failing) tests, so safe to merge.
/label acknowledge-critical-fixes-only

PR #7463, opened 2023-09-04
OCPBUGS-18457: Make extracting ISO kargs more robust

zaneb:
When getting the kernel arguments for PXE, we want to strip out the coreos.liveiso argument from the list in the ISO, since that one is specific to the ISO.

Previosuly we did this by returning all but the first arg. (Although the comment said we were returning the last 2, this did not match the code.)

Latest versions of FCOS have an additional arg before coreos.liveiso:
```
mitigations=auto,nosmt coreos.liveiso=fedora-coreos-38.20230609.3.0 ignition.firstboot ignition.platform.id=metal
```
Unlike latest versions of RHCOS:
```
coreos.liveiso=rhcos-414.92.202308032115-0 ignition.firstboot ignition.platform.id=metal
```
The result is that on FCOS we included an unwanted argument and dropped a wanted one.

To make this robust against both different platforms and potential future changes, don't rely on the ordering of arguments but instead remove the specific one that we don't want.


zaneb:
/override ci/prow/e2e-agent-sno-ipv4-pxe
Cluster installed ok, failure was at the must-gather step.

PR #7469, opened 2023-09-06
OCPBUGS-18387: Add Azure ConfidentialVM capability and DiskEncryptionSet validations

mresvanis:
This PR adds the following install config validations regarding [Azure Confidential VMs](https://learn.microsoft.com/en-us/azure/confidential-computing/confidential-vm-overview) and [Trusted Launch for VMs](https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch):

- Confidential VM (specifically `AMD SEV-SNP`) support for the selected VM size
- Trusted Launch for VMs support for the selected VM size
    - Trusted Launch is not supported on Confidential VMs
- when ConfidentialVM with `SecurityEncryptionType: DiskWithVMGuestState` and customer-managed key, the `DiskEncryptionSet.EncryptionType` should be `ConfidentialVMEncryptedWithCustomerKey`
- `Settings.SecurityType` (either `ConfidentialVM` or `TrustedLaunch`) is not currently supported on AzureStack

### DIFF
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -11,6 +11,7 @@ import (
 
 	azdns "github.com/Azure/azure-sdk-for-go/profiles/2018-03-01/dns/mgmt/dns"
 	aznetwork "github.com/Azure/azure-sdk-for-go/profiles/2018-03-01/network/mgmt/network"
+	azenc "github.com/Azure/azure-sdk-for-go/profiles/latest/compute/mgmt/compute"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


pkg/asset/installconfig/azure/validation.go
```diff
@@ -92,6 +93,58 @@ func ValidateDiskEncryptionSet(client API, ic *types.InstallConfig) field.ErrorL
...
+	allErrs := field.ErrorList{}
+
+	invalidDiskEncryptionSetTypeMsg := fmt.Sprintf("the disk encryption set should be created with type %s", azenc.DiskEncryptionSetTypeConfidentialVMEncryptedWithCustomerKey)
+
+	if ic.Platform.Azure.DefaultMachinePlatform != nil &&
+		ic.Platform.Azure.DefaultMachinePlatform.OSDisk.SecurityProfile != nil &&
+		ic.Platform.Azure.DefaultMachinePlatform.OSDisk.SecurityProfile.DiskEncryptionSet != nil {
+		desFieldPath := field.NewPath("platform").Child("azure", "defaultMachinePlatform", "osDisk", "securityProfile", "diskEncryptionSet")
+		diskEncryptionSet := ic.Platform.Azure.DefaultMachinePlatform.OSDisk.SecurityProfile.DiskEncryptionSet
+		resp, err := client.GetDiskEncryptionSet(context.TODO(), diskEncryptionSet.SubscriptionID, diskEncryptionSet.ResourceGroup, diskEncryptionSet.Name)
```
sadasu, line 107: Can this block of code from L107 - L112 can be moved to a new method? It appears 3 times in this validation code.

pkg/asset/installconfig/azure/validation.go
```diff
@@ -158,6 +158,29 @@ func validateMininumRequirements(fieldPath *field.Path, req resourceRequirements
...
+func validateSecurityType(fieldPath *field.Path, securityType aztypes.SecurityTypes, instanceType string, capabilities map[string]string) field.ErrorList {
+	allErrs := field.ErrorList{}
+
+	unsupportedSecurityType := false
+	_, hasTrustedLaunchDisabled := capabilities["TrustedLaunchDisabled"]
+	confidentialComputingType, hasConfidentialComputingType := capabilities["ConfidentialComputingType"]
+	isConfidentialComputingTypeSNP := confidentialComputingType == "SNP"
+
+	switch securityType {
+	case aztypes.SecurityTypesConfidentialVM:
```
sadasu, line 223: Nit: Is it at all possible to make this code represent cases that are supported and the default represent the unsupported cases?

mresvanis: Good point, I also think it's more readable this way and updated this code accordingly PTAL

pkg/asset/installconfig/azure/validation.go
mresvanis: Valid point and it kind of smells to me too. But it's not exactly the same piece of code, it varies between the `default`, `control-plane` and `compute` definitions and I could not find a good way to refactor these 3 lines to a method. Any ideas that could help make this code more readable and sustainable?

pkg/asset/installconfig/azure/validation.go
```diff
@@ -158,6 +211,29 @@ func validateMininumRequirements(fieldPath *field.Path, req resourceRequirements
...
+	supportedSecurityType := true
+	switch securityType {
+	case aztypes.SecurityTypesConfidentialVM:
+		supportedSecurityType = hasConfidentialComputingType && isConfidentialComputingTypeSNP
+	case aztypes.SecurityTypesTrustedLaunch:
+		supportedSecurityType = !(hasTrustedLaunchDisabled || hasConfidentialComputingType)
+	}
+
+	if !supportedSecurityType {
+		errMsg := fmt.Sprintf("this security type is not supported for instance type %s", instanceType)
```
sadasu, line 230: Is it possible for you to formulate this errMsg within the case statement, so that the message can be tailored to to the unsupported case?

pkg/asset/installconfig/azure/validation.go
sadasu: It appears that` desFieldPath` and `diskEncryptionSet` differ for `default`, `control-plane` and `compute`. Provide those as input parameters to a method that does the rest . You could move the definition of `invalidDiskEncryptionSetTypeMsg` into this method that returns error.
You could add the returned err from this method to allErrs using a single call before the return at L145.
`allErrs = append(allErrs, field.Invalid(desFieldPath, diskEncryptionSet, err.Error()))`

pkg/asset/installconfig/azure/validation.go
```diff
@@ -92,6 +93,58 @@ func ValidateDiskEncryptionSet(client API, ic *types.InstallConfig) field.ErrorL
...
+		diskEncryptionSet := ic.ControlPlane.Platform.Azure.OSDisk.SecurityProfile.DiskEncryptionSet
+		resp, err := client.GetDiskEncryptionSet(context.TODO(), diskEncryptionSet.SubscriptionID, diskEncryptionSet.ResourceGroup, diskEncryptionSet.Name)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(desFieldPath, diskEncryptionSet, err.Error()))
+		} else if resp == nil || resp.EncryptionSetProperties == nil || resp.EncryptionSetProperties.EncryptionType != azenc.DiskEncryptionSetTypeConfidentialVMEncryptedWithCustomerKey {
+			allErrs = append(allErrs, field.Invalid(desFieldPath, diskEncryptionSet, invalidDiskEncryptionSetTypeMsg))
+		}
+	}
+
+	for idx, compute := range ic.Compute {
```
sadasu, line 129: I see that we expect multiple `compute` instances but only 1 `control` and `default`?

mresvanis: That is correct, users can define multiple [compute MachinePools](https://github.com/openshift/installer/blob/master/pkg/types/installconfig.go#L119-L131), but only one `MachinePool` for the `control-plane` or the `default` one.

pkg/asset/installconfig/azure/validation.go
mresvanis: Thank you for spotting this, I indeed left out the specifics from the unsupported case error message. Do you think [this](https://github.com/openshift/installer/pull/7469/files#diff-75606c90c997f4218689ce173a97316d61be39a8df4f5bebe08a4cd0e9f24973R229-R245) makes sense?

pkg/asset/installconfig/azure/validation.go
mresvanis: Did you have something like [this](https://github.com/openshift/installer/pull/7469/files#diff-75606c90c997f4218689ce173a97316d61be39a8df4f5bebe08a4cd0e9f24973R96-R104) in mind?

PR #7470, opened 2023-09-06
AGENT-422: agent: Accept pull secret from Data field

rna-afk:
Made changes to the pull secret config logic to accept
pull secret information from the Data field if StringData
field is not populated.

### DIFF
--- a/pkg/asset/agent/manifests/agentpullsecret.go
+++ b/pkg/asset/agent/manifests/agentpullsecret.go
@@ -112,16 +112,20 @@ func (a *AgentPullSecret) finish() error {
 	}
 
 	// Normalise the JSON formatting so that we can redact the file reliably
-	normal, err := normalizeDockerConfig(a.Config.StringData[pullSecretKey])
+	pullSecretValue := a.GetPullSecretData()
+	normal, err := normalizeDockerConfig(pullSecretValue)
 	if err != nil {
 		return err
 	}
+	a.Config.StringData = map[string]string{}
 	a.Config.StringData[pullSecretKey] = normal
+	a.Config.Data = nil
 
 	secretData, err := yaml.Marshal(a.Config)
 	if err != nil {
 		return errors.Wrap(err, "failed to marshal agent secret")
 	}
+
 	a.File = &asset.File{
 		Filename: agentPullSecretFilename,
 		Data:     secretData,
@@ -143,11 +147,11 @@ func normalizeDockerConfig(stringData string) (string, error) {
 
 func (a *AgentPullSecret) validatePullSecret() field.ErrorList {
 	if err := a.validateSecretIsNotEmpty(); err != nil {
-		return err
+		return field.ErrorList{err}
 	}
 
-	fieldPath := field.NewPath("StringData")
-	dockerConfig := a.Config.StringData[pullSecretKey]
+	fieldPath := field.NewPath("stringData")
+	dockerConfig := a.GetPullSecretData()
 	if err := validate.ImagePullSecret(dockerConfig); err != nil {
 		return field.ErrorList{field.Invalid(fieldPath, dockerConfig, err.Error())}
 	}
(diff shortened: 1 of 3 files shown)


pkg/asset/agent/manifests/agentpullsecret.go
```diff
@@ -63,6 +63,9 @@ func (a *AgentPullSecret) Generate(dependencies asset.Parents) error {
 			StringData: map[string]string{
 				pullSecretKey: installConfig.Config.PullSecret,
 			},
+			Data: map[string][]byte{
```
andfasano, line 66: This doesn't seem to be required: it's ok to generate a secret filling the `StringData` field only (when the source it's `install-config.yaml`)

pkg/asset/agent/manifests/agentpullsecret.go
```diff
@@ -112,11 +115,17 @@ func (a *AgentPullSecret) finish() error {
 	}
 
 	// Normalise the JSON formatting so that we can redact the file reliably
-	normal, err := normalizeDockerConfig(a.Config.StringData[pullSecretKey])
+	pullSecretValue, ok := a.Config.StringData[pullSecretKey]
+	if !ok {
+		pullSecretValue = string(a.Config.Data[pullSecretKey])
+	}
+	normal, err := normalizeDockerConfig(pullSecretValue)
```
andfasano, line 122: Extracting the field with the effective secret data could sound as a good target for refactoring a small func, ie something like an internal version of `GetPullSecretData` but working directly on a Secret object

pkg/asset/agent/manifests/agentpullsecret.go
```diff
@@ -112,11 +115,17 @@ func (a *AgentPullSecret) finish() error {
...
 	if err != nil {
 		return err
 	}
 	a.Config.StringData[pullSecretKey] = normal
+	a.Config.Data = map[string][]byte{}
+	a.Config.Data[pullSecretKey] = []byte(normal)
```
andfasano, line 128: Same as before: it's ok to keep just StringData

cmd/openshift-install/testdata/agent/configimage/manifests/default_manifests.txt
```diff
@@ -104,6 +104,8 @@ spec:
 status: {}
 -- expected/pull-secret.yaml --
 apiVersion: v1
+data:
+  .dockerconfigjson: ewogICJhdXRocyI6IHsKICAgICJxdWF5LmlvIjogewogICAgICAiYXV0aCI6ICJjM1Z3WlhJdGMyVmpjbVYwQ2c9PSIKICAgIH0KICB9Cn0=
```
andfasano, line 108: Regarding the integration tests, in order to match the issue reported inhttps://issues.redhat.com/browse/AGENT-422, my expectation would be to see a new one where:
* A set of ZTP manifests are provided
* The pull secret file contains the secret in the Data field
The expectation should check that the generated Secret file contains the expected StringData

pkg/asset/agent/manifests/agentpullsecret.go
```diff
@@ -143,43 +152,48 @@ func normalizeDockerConfig(stringData string) (string, error) {
...
+	return nil
 }
 
 // GetPullSecretData returns the content of the pull secret.
 func (a *AgentPullSecret) GetPullSecretData() string {
-	return a.Config.StringData[".dockerconfigjson"]
+	if _, ok := a.Config.StringData[pullSecretKey]; ok {
+		return a.Config.StringData[pullSecretKey]
+	}
+	return string(a.Config.Data[pullSecretKey])
```
zaneb, line 198: We need to decode from base64 at this point.

pkg/asset/agent/manifests/agentpullsecret.go
```diff
@@ -112,11 +115,17 @@ func (a *AgentPullSecret) finish() error {
 	}
 
 	// Normalise the JSON formatting so that we can redact the file reliably
-	normal, err := normalizeDockerConfig(a.Config.StringData[pullSecretKey])
+	pullSecretValue, ok := a.Config.StringData[pullSecretKey]
+	if !ok {
+		pullSecretValue = string(a.Config.Data[pullSecretKey])
```
zaneb, line 120: Need to decode from base64 here. Can we call a.GetPullSecretData() to avoid duplicating the implementation?

pkg/asset/agent/manifests/agentpullsecret_test.go
```diff
@@ -48,6 +48,9 @@ func TestAgentPullSecret_Generate(t *testing.T) {
 				StringData: map[string]string{
 					".dockerconfigjson": "{\n  \"auths\": {\n    \"cloud.openshift.com\": {\n      \"auth\": \"b3BlUTA=\",\n      \"email\": \"test@redhat.com\"\n    }\n  }\n}",
 				},
+				Data: map[string][]byte{
+					".dockerconfigjson": []byte("{\n  \"auths\": {\n    \"cloud.openshift.com\": {\n      \"auth\": \"b3BlUTA=\",\n      \"email\": \"test@redhat.com\"\n    }\n  }\n}"),
```
zaneb, line 52: This is wrong, the Data field is always base64-encoded.

pkg/asset/agent/manifests/agentpullsecret.go
rna-afk: Surprisingly, string function converts it from base64 to string. Tried it with the integration tests and it seemed to convert. Will look into it further to confirm it.

pkg/asset/agent/manifests/agentpullsecret_test.go
```diff
@@ -156,7 +156,7 @@ metadata:
   namespace: cluster-0
 stringData:
   .dockerconfigjson: 'foo'`,
-			expectedError: "invalid PullSecret configuration: StringData: Invalid value: \"foo\": invalid character 'o' in literal false (expecting 'a')",
+			expectedError: "invalid PullSecret configuration: Data: Invalid value: \"foo\": invalid character 'o' in literal false (expecting 'a')",
```
zaneb, line 159: This is definitely wrong because the invalid data is in the stringData field, not Data. As noted above, I think we should change the field to use in errors when data is missing altogether to `stringData`.

We also need a separate test for when the invalid data is actually in the `data` field.

pkg/asset/agent/manifests/agentpullsecret.go
```diff
@@ -143,43 +146,48 @@ func normalizeDockerConfig(stringData string) (string, error) {
 
 func (a *AgentPullSecret) validatePullSecret() field.ErrorList {
 	if err := a.validateSecretIsNotEmpty(); err != nil {
-		return err
+		return field.ErrorList{err}
 	}
 
-	fieldPath := field.NewPath("StringData")
-	dockerConfig := a.Config.StringData[pullSecretKey]
+	fieldPath := field.NewPath("Data")
```
zaneb, line 152: Actual field name is `data` not `Data` (this was wrong before).

pkg/asset/agent/manifests/agentpullsecret.go
```diff
@@ -143,43 +146,48 @@ func normalizeDockerConfig(stringData string) (string, error) {
...
-
-	if len(a.Config.StringData) == 0 {
-		allErrs = append(allErrs, field.Required(fieldPath, "the pull secret is empty"))
-		return allErrs
+	if len(a.Config.StringData) == 0 && len(a.Config.Data) == 0 {
+		return field.Required(fieldPath, "the pull secret is empty")
 	}
 
 	pullSecret, ok := a.Config.StringData[pullSecretKey]
 	if !ok {
```
zaneb, line 172: We should have conflict error if both are present (or maybe just if both are present and they have different values?).

PR #7482, opened 2023-09-11
OpenStack: fix IPv6 docs

MaysaMacedo:
Given we are using a provider network in the example, we should recommend adding the network in the router as external gateway instead.


mandre (commented):
/lgtm
/approve

PR #7484, opened 2023-09-13
OCPBUGS-19303: Changed OKD/FCOS workaround to also support Agent-based Installer

JM1:
OKD/FCOS uses FCOS as its bootimage, i.e. when booting cluster nodes the first time during installation. FCOS does not provide tools such as OpenShift Client (oc) or crio.service which Agent-based Installer uses at the rendezvous host, e.g. to launch the bootstrap control plane.

RHCOS and SCOS include these tools, but [FCOS has to pivot the root fs](https://github.com/openshift/installer/blob/master/data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template) to [okd-machine-os](https://github.com/openshift/okd-machine-os) first in order to make those tools available.

Pivoting uses 'rpm-ostree rebase' but the rendezvous host is booted the first time the node boots from a FCOS Live ISO where the root fs and /sysroot are mounted read-only. Thus 'rpm-ostree rebase' fails and necessary tools will not be available, causing the setup to stall.

Until [rpm-ostree has implemented support for rebasing Live ISOs](https://github.com/coreos/rpm-ostree/issues/4547), this patch adapts [the workaround for SNO installations](https://github.com/openshift/installer/pull/7445) to also support Agent-based Installer.

In particular, the Go conditional {{- if .BootstrapInPlace }} which is used to mark a SNO install has been replaced with a shell if-else which checks at runtime whether the system is launched from are on a Live ISO. Most code in the OpenShift ecosystem is written with RHCOS in mind and often assumes that tools like oc or crio.service are available. These assumptions can be satisfied by applying this workaround to all Live ISO boots. It will not remove functionality or overwrite configuration files in /etc and thus side effects should be minimal.

The Go conditional {{- if .BootstrapInPlace }} in the release-image-pivot.service has been dropped completely. This service is only used in OKD only, so OCP will not be impacted at all. The 'Before=' option will not cause systemd to fail if a service does not exist. So, in case bootkube.service or kubelet.service do not exist, the option will have no effect. When bootkube.service or kubelet.service do exist, it must always be ensured that release-image-pivot.service is started first because it might reboot the system or change /usr in the Live ISO use case. So it is safe to drop the Go conditional and ask systemd to always launch release-image-pivot.service before bootkube.service and kubelet.service.

### DIFF
--- a/data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
+++ b/data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
@@ -42,25 +42,72 @@ if [ ! -f /opt/openshift/.pivot-done ]; then
   record_service_stage_start "rebase-to-okd-os-image"
 {{if .IsFCOS -}}
   mnt="$(podman image mount "${MACHINE_OS_IMAGE}")"
-{{- if or (.BootstrapInPlace) (eq .Invoker "agent-installer") }}
-  # SNO setup boots into Live ISO which cannot be rebased
-  # https://github.com/coreos/rpm-ostree/issues/4547
-  mkdir /var/mnt/{upper,worker}
-  mount -t overlay overlay -o "lowerdir=/usr:$mnt/usr" /usr
-  mount -t overlay overlay -o "lowerdir=/etc:$mnt/etc,upperdir=/var/mnt/upper,workdir=/var/mnt/worker" /etc
-  systemctl daemon-reload
 
-  # Workaround for SELinux denials when launching crio.service from overlayfs
-  setenforce Permissive
+  # The bootstrap host during SNO installation and the rendezvous host of Agent-based Installer both boot into a Live
+  # ISO which cannot be rebased. Until rpm-ostree supports this live rebase [0], the following workaround will mount the
+  # proper OKD/FCOS Machine OS image over the existing mount at /usr and copy new config files to /etc.
+  # [0] https://github.com/coreos/rpm-ostree/issues/4547
+  if grep -q coreos.liveiso= /proc/cmdline; then
+    mount -t tmpfs -o size=50% none /var/mnt/
+    rsync -aHAXx "$mnt/" /var/mnt/
+    mount -t overlay overlay -o lowerdir=/usr:/var/mnt/usr /usr
+    rsync -rlt --ignore-existing /var/mnt/etc/ /etc/
 
-  systemctl start crio.service
-  # No reboot necessary because SNO setup will reboot system
-{{ else }}
-  pushd "${mnt}/bootstrap"
-    # shellcheck disable=SC1091
-    . ./pre-pivot.sh
-  popd
-{{ end -}}
+    # Agent-based Installer will launch a ephemeral control plane at the rendezvous host which will create and publish
+    # Ignition configs for the other master nodes. These Ignition configs must match what the in-cluster control plane
+    # would generate else machine config operator will fail [0]. Because the rendezvous host is booted with a FCOS Live
+    # ISO without any OKD/FCOS related changes, we have to copy the manifests from OKD Machine OS manually to the
+    # bootstrap manifests folder of the rendezvous host.
+    # [0] https://access.redhat.com/solutions/4970731
+    mkdir -p /var/opt/openshift/manifests
+    cp -av /var/mnt/manifests/*.* /var/opt/openshift/manifests/
+
+    # Load new systemd unit files and configuration such as crio.service after mounting the content of OKD/FCOS Machine
+    # OS over /usr and copying new files to /etc
+    systemctl daemon-reload
+
+    # Apply presets from OKD Machine OS
+    systemctl preset-all
+
+    # On OKD/FCOS prior to commit e859a66 [0] systemd-resolved is used by default and NetworkManager's DNS handling is
+    # disabled. In this case, CoreDNS fails to listen to 127.0.0.53:53 when Agent-based Installer boots its the
+    # rendezvous host with a Fedora CoreOS bootimage because by default FCOS' systemd-resolved already listens to this
+    # port. OKD/FCOS disables resolved's stub listener [1] but the resolved must be restarted for this setting to take
+    # effect.
+    # On OKD/FCOS since commit e859a66 [0] systemd-resolved is disabled by default and NetworkManager's DNS handling is
+    # used. However, the bootimage is vanilla FCOS and thus uses systemd-resolved by default. The latter has to be
+    # disabled after rebasing to OKD Machine OS and NetworkManager as well as the service to fix /etc/resolv.conf have
+    # to be started.
+    # [0] https://github.com/openshift/okd-machine-os/commit/e859a6643330596a8a282aeb4bf853763a2d219e
+    # [1] https://github.com/openshift/okd-machine-os/blob/28dec35d60ea07069366b22ebdcb296d429b15e9/overlay.d/99okd/etc/systemd/resolved.conf.d/okd-no-dns-stub.conf
+    if [ -e /etc/systemd/resolved.conf.d/okd-no-dns-stub.conf ]; then
+        systemctl restart systemd-resolved.service
+    else
+        systemctl disable --now systemd-resolved.service
+    fi
+
+    if systemctl list-unit-files -q fix-resolvconf.service >/dev/null; then
+        systemctl stop NetworkManager.service
+        systemctl start fix-resolvconf.service
+        systemctl start NetworkManager.service
+        nmcli general reload dns-full
+    fi
+
+    # Workaround for SELinux denials when launching crio.service from overlayfs
+    setenforce Permissive
+
+    # crio.service is not part of FCOS but of OKD Machine OS. It will loaded after systemctl daemon-reload above but has
+    # to be started manually
+    systemctl start crio.service
+
+    # No reboot necessary because setup will reboot the system automatically
+  else
+    pushd "${mnt}/bootstrap"
+      # shellcheck disable=SC1091
+      . ./pre-pivot.sh
+    popd
+  fi
+  record_service_stage_success
 {{else if .IsSCOS -}}
   chmod 0644 /etc/containers/registries.conf
   rpm-ostree rebase --experimental "ostree-unverified-registry:${MACHINE_OS_IMAGE}"
--- a/data/data/bootstrap/systemd/common/units/kubelet.service.template
+++ b/data/data/bootstrap/systemd/common/units/kubelet.service.template
@@ -1,7 +1,13 @@
 [Unit]
 Description=Kubernetes Kubelet
 Wants=rpc-statd.service crio.service release-image.service
+{{if .IsOKD -}}
+Wants=release-image-pivot.service
+{{end -}}
 After=crio.service release-image.service
+{{if .IsOKD -}}
+After=release-image-pivot.service
+{{end -}}
 
 [Service]
 Type=notify
--- a/data/data/bootstrap/systemd/common/units/release-image-pivot.service.template
+++ b/data/data/bootstrap/systemd/common/units/release-image-pivot.service.template
@@ -3,11 +3,7 @@
 Description=Pivot bootstrap to the OpenShift Release Image
 Wants=release-image.service
 After=release-image.service
-{{- if or (.BootstrapInPlace) (eq .Invoker "agent-installer") }}
 Before=bootkube.service kubelet.service
-{{ else }}
-Before=bootkube.service
-{{ end -}}
 
 [Service]
 Type=oneshot
--- a/data/data/bootstrap/systemd/common/units/release-image-pivot.service.template
+++ b/data/data/bootstrap/systemd/common/units/release-image-pivot.service.template
@@ -3,7 +3,7 @@
 Description=Pivot bootstrap to the OpenShift Release Image
 Wants=release-image.service
 After=release-image.service
-Before=bootkube.service kubelet.service
+Before=bootkube.service kubelet.service dnsmasq.service
 
 [Service]
 Type=oneshot


data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
```diff
@@ -41,26 +42,28 @@ if [ ! -f /opt/openshift/.pivot-done ]; then
 
   record_service_stage_start "rebase-to-okd-os-image"
 {{if .IsFCOS -}}
+  # get more space because 7.8gb is not enough
```
andfasano, line 45: Is this due https://github.com/openshift-metal3/dev-scripts/pull/1542? Sometimes ago we were affected by https://issues.redhat.com/browse/OCPBUGS-8036, but then it was fixed

JM1: The increased space is required because container images could not be pulled due to insufficient storage and because i copy (rsync) all files from the machine os image to a podman volume (which had to do because the podman image would be "unmounted" at one point, quasi-erasing /usr)

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
```diff
@@ -42,25 +42,46 @@ if [ ! -f /opt/openshift/.pivot-done ]; then
...
-    . ./pre-pivot.sh
-  popd
-{{ end -}}
+    # Agent-based Installer will launch a ephemeral control plane at the rendezvous host which will create and publish
+    # Ignition configs for the other master nodes. These Ignition configs must match what the in-cluster control plane
+    # would generate else machine config operator will fail [0]. Because the rendezvous host is booted with a FCOS Live
+    # ISO without any OKD/FCOS related changes, we have to copy the manifests from OKD Machine OS manually to the
+    # bootstrap manifests folder of the rendezvous host.
+    # [0] https://access.redhat.com/solutions/4970731
+    mkdir -p /etc/mcc/bootstrap/
```
andfasano, line 62: suggestion: for a slightly cleaner workaround, since the `/etc/mcc/bootstrap/` folder is going to be created subsequently by [bootkube.sh](https://github.com/openshift/installer/blob/4e4eec6f3a8630b3d2d83020fac6eaa420707085/data/data/bootstrap/files/usr/local/bin/bootkube.sh.template#L361-L365), the content of `/var/mnt/manifests` could be copied into `/var/opt/openshift/manifests` (as if they were generated by the installer)

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
```diff
@@ -42,25 +42,46 @@ if [ ! -f /opt/openshift/.pivot-done ]; then
...
+    cp -av /var/mnt/manifests/*.* /etc/mcc/bootstrap/
+
+    systemctl daemon-reload
```
andfasano, line 65: This seems to be superflous now

JM1: With `rsync -rlt --ignore-existing /var/mnt/etc/ /etc/` above and the `/usr` mount we change a bunch of files which includes systemd services. For example, crio.service is not part of FCOS but we need it for agent-based installer.

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
JM1: Thank you, done!

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
andfasano: Ok, it could be worth adding a comment then to hilight that

data/data/bootstrap/systemd/units/release-image-pivot.service.template
```diff
@@ -3,11 +3,7 @@
 Description=Pivot bootstrap to the OpenShift Release Image
 Wants=release-image.service
 After=release-image.service
-{{- if .BootstrapInPlace }}
```
JM1, line 6: I dropped this condition only as a temporary measure to test the remaining parts and because I could not find a proper go-template variable for agent-based. But we should consider the consequences before removing the condition completely. Do we always have a kubelet.service? Even in cases where we do not use agent-based and bootstrap-in-place?

andfasano:
/test okd-e2e-agent-compact-ipv4 

(useful to test the non-SNO paths)

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
JM1: Done.

data/data/bootstrap/systemd/units/release-image-pivot.service.template
JM1: `release-image-pivot.service` is only used on OKD, so OCP will not be impacted at all. 

[`Before=` will not let systemd fail if a service does not exist](https://serverfault.com/questions/1062205/does-systemd-fail-if-a-dependency-in-after-requires-doesnt-exist), so in case where `bootkube.service` or `kubelet.service` does no harm.

When `bootkube.service` or `kubelet.service` do exist, we would always want to ensure that `release-image-pivot.service` is started first because it might reboot the system or change `/usr` in the live iso use case.

That is why I think it is safe to drop the `.BootstrapInPlace` check and unconditionally ask systemd to launch `release-image-pivot.service` before `bootkube.service` and `kubelet.service`.

andfasano: There could be other scenarios where it may be useful to differentiate between abi and not-abi cases, for this reason I've prepared https://github.com/openshift/installer/pull/7508 that could be helpful to support the current scenario, and it may be handy for future ones

JM1:
cc @vrutkovs @andfasano @LorbusChris Patch is ready for review.. finally ☺️

JM1:
Sibling patch to fix the api-int.* endpoint issue for OKD/FCOS: https://github.com/openshift/installer/pull/7516

andfasano:
> cc @vrutkovs @andfasano @LorbusChris Patch is ready for review.. finally ☺️

Thanks @JM1. Given that the current patch works and looks fine to me within the ABI scope,  I'd feel more comfortable to limit it only to SNO or ABI cases (by using the approach described in #7508), at least initially.

I'm not sure if dropping the conditionals will require more extensive testing/validation for non-SNO and non-ABI cases (and I'd like to hear the feedback on that from @vrutkovs and @LorbusChris), and in any case the template code simplification may followup in a subsequent PR, limiting the current one just to the ABI scope.

JM1:
FYI Last patch is rebased on top of master.

vrutkovs (approved):
LGTM

andfasano:
Note: now requires https://github.com/openshift/installer/pull/7580

andfasano:
Looks like now this PR could be rebased, since https://github.com/openshift/installer/pull/7580 landed

andfasano:
The following tests are required to pass for this patch:

/test okd-e2e-agent-sno-ipv6
/test okd-e2e-agent-compact-ipv4

andfasano:
/hold 

agent compact and sno jobs are not passing

JM1:
Gimme a release image to debug, please.

 /test okd-e2e-agent-compact-ipv4

JM1:
Gimme a release image to debug, please.

/test okd-e2e-agent-compact-ipv4

JM1:
C'mon.. please gimme a release image to debug.

/test okd-e2e-agent-compact-ipv4

LorbusChris:
> Looks like now this PR could be rebased, since #7580 landed

https://github.com/openshift/installer/pull/7580 broke OKD IPI. That'll hopefully be fixed by https://github.com/openshift/installer/pull/7629

andfasano:
Since https://github.com/openshift-metal3/dev-scripts/pull/1597 landed, let's retry the agent jobs

/test okd-e2e-agent-compact-ipv4
/test okd-e2e-agent-sno-ipv6

JM1:
https://github.com/openshift/installer/pull/7634 is required in order to fix the okd-e2e-agent-* jobs.

data/data/bootstrap/systemd/common/units/kubelet.service.template
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=Kubernetes Kubelet
-Wants=rpc-statd.service crio.service release-image.service
-After=crio.service release-image.service
+Wants=rpc-statd.service crio.service release-image.service{{if .IsOKD}} release-image-pivot.service{{end}}
```
vrutkovs, line 3:
```suggestion
Wants=rpc-statd.service crio.service release-image.service
{{if .IsOKD}}Wants=release-image-pivot.service{{end}}
```

same functionality but cleaner to read / revert

data/data/bootstrap/systemd/common/units/kubelet.service.template
```diff
@@ -1,7 +1,7 @@
...
+After=crio.service release-image.service{{if .IsOKD}} release-image-pivot.service{{end}}
```
vrutkovs, line 4: same here

data/data/bootstrap/systemd/common/units/kubelet.service.template
JM1: Good idea, ty!

JM1:
Failed job `okd-e2e-agent-ha-dualstack` has a couple of SELinux error messages in agent-gather's [journal.export](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_installer/7484/pull-ci-openshift-installer-master-okd-e2e-agent-ha-dualstack/1726970555036340224/artifacts/e2e-agent-ha-dualstack/agent-gather/artifacts/agent-gather-master-0.tar.xz) which prevents coredns from being started at the rendezvous host. I have some ideas how to fix those, but first:

Why did `okd-e2e-agent-sno-ipv6` pass? Is SELinux disabled for `okd-e2e-agent-sno-ipv6`? And why did  `okd-e2e-agent-ha-dualstack` pass in PR #7641?

JM1:
FYI: Opened PR #7756 to gather CRI-O containers logs from rendezvous host, so that we can actually see why CoreDNS failed.

andfasano:
The compact failure was due a timeout since some operators were degraded (at least the installation reached that point!)

JM1:
/test okd-e2e-agent-compact-ipv4

(need a release image for debugging again)

JM1:
/test okd-e2e-agent-sno-ipv6

(need a working release image)

JM1:
/test okd-e2e-agent-sno-ipv6

(need a working release image)

andfasano:
Rrequires now:
* https://github.com/openshift/installer/pull/7779
* https://github.com/openshift/okd-machine-os/pull/701

aleskandro:
Hi, I did some other testing and I'm still unsure the patch openshift/okd-machine-os#701 is the root cause of the failure and "required" to let this enablement done.

Rather, a race could happen when we disable systemd-resolved and restart NetworkManager as of the workaround (rather than the usual rebase and reboot) letting the NetworkManager to set 127.0.0.53 as nameserver (i.e., the previous systemd-resolved listener). I'd suggest:

1. Adding `systemctl preset-all` after `systemctl daemon-reload` as this is what we actually expect after the rebase: see https://github.com/openshift/okd-machine-os/blob/master/Dockerfile#L35 and https://github.com/openshift/okd-machine-os/blob/master/overlay.d/99okd/usr/lib/systemd/system-preset/00-okd.preset
2. Add `nmcli general reload dns-full` after restarting NetworkManager.service (not sure this is enough though)

## UEFI issue:

In the real BM nodes I tested with, before installing the node, we have no EFI Entries:

```
BootCurrent: 0006
BootOrder: 0005,0000,0006,0007
Boot0000* Integrated NIC 1 Port 3 Partition 1   VenHw(d227c733-f75f-4341-b749-4d1759ec8538)
      dp: 01 04 14 00 33 c7 27 d2 5f f7 41 43 b7 49 4d 17 59 ec 85 38 / 7f ff 04 00
Boot0001* Hard drive C: VenHw(d6c0639f-c705-4eb9-aa4f-5802d8823de6)feff1a00000000000000000000000104f00000c802000000cc0000c8840100c800000000000000000000000000000000000000000000000000000000000000000000000000110000000000000000001c0002010c00d041030a010000000101060000000101060000007fff04005000450052004300200048003700330030005000200041006400610070007400650072002800620075007300200031004100200064006500760020003000300029000000
      dp: 01 04 14 00 9f 63 c0 d6 05 c7 b9 4e aa 4f 58 02 d8 82 3d e6 / 7f ff 04 00
    data: fe ff 1a 00 00 00 00 00 00 00 00 00 00 00 01 04 f0 00 00 c8 02 00 00 00 cc 00 00 c8 84 01 00 c8 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 11 00 00 00 00 00 00 00 00 00 1c 00 02 01 0c 00 d0 41 03 0a 01 00 00 00 01 01 06 00 00 00 01 01 06 00 00 00 7f ff 04 00 50 00 45 00 52 00 43 00 20 00 48 00 37 00 33 00 30 00 50 00 20 00 41 00 64 00 61 00 70 00 74 00 65 00 72 00 28 00 62 00 75 00 73 00 20 00 31 00 41 00 20 00 64 00 65 00 76 00 20 00 30 00 30 00 29 00 00 00
Boot0002* BRCM MBA Slot 1800 v21.6.0    BBS(128,BRCM MBA Slot 1800 v21.6.0,0x0)feff1800000000000000000000000200660380cf80000000800080cf4b0380cf00000000000000000000000000000000000000000000000000000000000000000000000000120000020000000000001c0002010c00d041030a010000000101060000030101060000007fff04004200520043004d0020004d0042004100200053006c006f0074002000310038003000300020007600320031002e0036002e0030000000
      dp: 05 01 23 00 80 00 00 00 42 52 43 4d 20 4d 42 41 20 53 6c 6f 74 20 31 38 30 30 20 76 32 31 2e 36 2e 30 00 / 7f ff 04 00
    data: fe ff 18 00 00 00 00 00 00 00 00 00 00 00 02 00 66 03 80 cf 80 00 00 00 80 00 80 cf 4b 03 80 cf 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 12 00 00 02 00 00 00 00 00 00 1c 00 02 01 0c 00 d0 41 03 0a 01 00 00 00 01 01 06 00 00 03 01 01 06 00 00 00 7f ff 04 00 42 00 52 00 43 00 4d 00 20 00 4d 00 42 00 41 00 20 00 53 00 6c 00 6f 00 74 00 20 00 31 00 38 00 30 00 30 00 20 00 76 00 32 00 31 00 2e 00 36 00 2e 00 30 00 00 00
Boot0003* Virtual Floppy        BBS(Floppy,Virtual Floppy,0x90)feff0000000014000000000000000c032f04409c01009000000000002f04409c00000000000000000000000000000000000000000000000000000000000e04a000f07570791500f0757079000000002d0002010c00d041030a00000000010106000014030506000d0003050600030003050600000003110500017fff04005600690072007400750061006c00200046006c006f007000700079000000
      dp: 05 01 17 00 01 00 90 00 56 69 72 74 75 61 6c 20 46 6c 6f 70 70 79 00 / 7f ff 04 00
    data: fe ff 00 00 00 00 14 00 00 00 00 00 00 00 0c 03 2f 04 40 9c 01 00 90 00 00 00 00 00 2f 04 40 9c 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 0e 04 a0 00 f0 75 70 79 15 00 f0 75 70 79 00 00 00 00 2d 00 02 01 0c 00 d0 41 03 0a 00 00 00 00 01 01 06 00 00 14 03 05 06 00 0d 00 03 05 06 00 03 00 03 05 06 00 00 00 03 11 05 00 01 7f ff 04 00 56 00 69 00 72 00 74 00 75 00 61 00 6c 00 20 00 46 00 6c 00 6f 00 70 00 70 00 79 00 00 00
Boot0004* Virtual CD/DVD        BBS(CDROM,Virtual CD/DVD,0x90)feff0000000014000000000000000c030904409c03009000000000000904409c00000000000000000000000000000000000000000000000000000000000e04a000ca4856441400ca485644000000002d0002010c00d041030a00000000010106000014030506000d0003050600030003050600000003110500007fff04005600690072007400750061006c002000430044002f004400560044000000
      dp: 05 01 17 00 03 00 90 00 56 69 72 74 75 61 6c 20 43 44 2f 44 56 44 00 / 7f ff 04 00
    data: fe ff 00 00 00 00 14 00 00 00 00 00 00 00 0c 03 09 04 40 9c 03 00 90 00 00 00 00 00 09 04 40 9c 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 0e 04 a0 00 ca 48 56 44 14 00 ca 48 56 44 00 00 00 00 2d 00 02 01 0c 00 d0 41 03 0a 00 00 00 00 01 01 06 00 00 14 03 05 06 00 0d 00 03 05 06 00 03 00 03 05 06 00 00 00 03 11 05 00 00 7f ff 04 00 56 00 69 00 72 00 74 00 75 00 61 00 6c 00 20 00 43 00 44 00 2f 00 44 00 56 00 44 00 00 00
Boot0005* Integrated NIC 1 Port 2 Partition 1   VenHw(3a191845-5f86-4e78-8fce-c4cff59f9daa)
      dp: 01 04 14 00 45 18 19 3a 86 5f 78 4e 8f ce c4 cf f5 9f 9d aa / 7f ff 04 00
Boot0006* Virtual CD/DVD        PciRoot(0x0)/Pci(0x14,0x0)/USB(13,0)/USB(3,0)/USB(1,0)
      dp: 02 01 0c 00 d0 41 03 0a 00 00 00 00 / 01 01 06 00 00 14 / 03 05 06 00 0d 00 / 03 05 06 00 03 00 / 03 05 06 00 01 00 / 7f ff 04 00
Boot0007* EFI Fixed Disk Boot Device 1  PciRoot(0x1)/Pci(0x0,0x0)/Pci(0x0,0x0)/Ctrl(0x0)/SCSI(1,0)/HD(2,GPT,79f4d9d7-2b3d-4436-b2d7-f59ae0e4cdfd,0x1000,0x3f800)
      dp: 02 01 0c 00 d0 41 03 0a 01 00 00 00 / 01 01 06 00 00 00 / 01 01 06 00 00 00 / 01 05 08 00 00 00 00 00 / 03 02 08 00 01 00 00 00 / 04 01 2a 00 02 00 00 00 00 10 00 00 00 00 00 00 00 f8 03 00 00 00 00 00 d7 d9 f4 79 3d 2b 36 44 b2 d7 f5 9a e0 e4 cd fd 02 02 / 7f ff 04 00
MirroredPercentageAbove4G: 0.00
MirrorMemoryBelow4GB: false
```

After the installation is done, a new entry is created and set as the first one, but it tries to use the `EFI/redhat` folder (as hard-coded in https://github.com/openshift/assisted-installer/blob/master/src/ops/ops.go#L251-L261  ?)


## MCO issue

The master MCP gets degraded while the installation tries to conclude.
```
Node master-01 is reporting: "command \"/usr/bin/rpm -qf /etc/audit/rules.d/mco-audit-quiet-containers.rules\" returned with unexpected error: error: file /etc/audit/rules.d/mco-audit-quiet-containers.rules: Permission denied\n: exit status 1", Node master-02 is reporting: "command \"/usr/bin/rpm -qf /etc/audit/rules.d/mco-audit-quiet-containers.rules\" returned with unexpected error: error: file /etc/audit/rules.d/mco-audit-quiet-containers.rules: Permission denied\n: exit status 1"
```

```
root@master-01:/etc/kubernetes/static-pod-resources# ls -l /usr/bin/rpm
-rwxr-xr-x. 2 root root 251 Jan  1  1970 /usr/bin/rpm
root@master-01:/etc/kubernetes/static-pod-resources# ls -lZ /usr/bin/rpm
-rwxr-xr-x. 2 root root system_u:object_r:rpm_exec_t:s0 251 Jan  1  1970 /usr/bin/rpm
root@master-01:/etc/kubernetes/static-pod-resources# rpm -qf /etc/audit/rules.d/mco-audit-quiet-containers.rules^C
root@master-01:/etc/kubernetes/static-pod-resources# ls -lZ /usr/libexec/rpm-ostree/wrapped/rpm
-rwxr-xr-x. 2 root root system_u:object_r:bin_t:s0 20528 Jan  1  1970 /usr/libexec/rpm-ostree/wrapped/rpm
root@master-01:/etc/kubernetes/static-pod-resources# rpm -qf /etc/audit/rules.d/mco-audit-quiet-containers.rules
error: file /etc/audit/rules.d/mco-audit-quiet-containers.rules: Permission denied
root@master-01:/etc/kubernetes/static-pod-resources# /usr/libexec/rpm-ostree/wrapped/rpm -qf /etc/audit/rules.d/mco-audit-quiet-containers.rules
file /etc/audit/rules.d/mco-audit-quiet-containers.rules is not owned by any package
```


I resolved this manually by running the following in each node after the rebase:
```
rpm-ostree usroverlay
chcon -t bin_t /usr/bin/rpm
```

I'm not sure what's the best solution for this.


With these changes (change in the selinux type label for `rpm` and manually adding the UEFI entry to boot the nodes after the rebase):

```
[...]
INFO Host: master-02, reached installation stage Joined 
INFO Host: master-00, reached installation stage Waiting for bootkube 
INFO Host: master-02, reached installation stage Done 
INFO Node master-01 has been rebooted 2 times before completing installation 
INFO Host: master-00, reached installation stage Waiting for bootkube: waiting for ETCD bootstrap to be complete 
INFO Host: master-00, reached installation stage Waiting for controller: waiting for controller pod ready event 
INFO Bootstrap configMap status is complete       
INFO cluster bootstrap is complete          
[...]
DEBUG Still waiting for the cluster to initialize: Cluster operator authentication is not available 
INFO Cluster is installed                         
INFO Install complete!                            
INFO To access the cluster as the system:admin user when using 'oc', run 
INFO     export KUBECONFIG=/tmp/installer/auth/kubeconfig 
INFO Access the OpenShift web-console here: ...
INFO Login to the console with user: "kubeadmin", and password: "..." 
bash-4.4$ 
```

JM1:
Great analysis, @andfasano, thank you ☺️

To fix the SELinux issue, we might have to [relabel like FCOS does during initrd run](https://github.com/coreos/fedora-coreos-config/blob/08b13806bb61faf85df2a93ffed9d7508ad47240/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator#L258).

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
```diff
@@ -42,25 +42,67 @@ if [ ! -f /opt/openshift/.pivot-done ]; then
...
+    # used. However, the bootimage is vanilla FCOS and thus uses systemd-resolved by default. The latter has to be
+    # disabled after rebasing to OKD Machine OS and NetworkManager as well as the service to fix /etc/resolv.conf have
+    # to be started.
+    # [0] https://github.com/openshift/okd-machine-os/commit/e859a6643330596a8a282aeb4bf853763a2d219e#diff-808ba069aeee05cbeb08aa7b8b5b4f6feb8aefe15ea9737339d07f8d7bf5d74a
+    # [1] https://github.com/openshift/okd-machine-os/blob/28dec35d60ea07069366b22ebdcb296d429b15e9/overlay.d/99okd/etc/systemd/resolved.conf.d/okd-no-dns-stub.conf
+    if [ -e /etc/systemd/resolved.conf.d/okd-no-dns-stub.conf ]; then
+        systemctl restart systemd-resolved.service
+    else
+        systemctl disable --now systemd-resolved.service
+    fi
```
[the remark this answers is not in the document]
JM1: This is for the old okd-machine-os, before your change. for easier backporting of this pr to older releases

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
```diff
@@ -42,25 +42,67 @@ if [ ! -f /opt/openshift/.pivot-done ]; then
...
+    # would generate else machine config operator will fail [0]. Because the rendezvous host is booted with a FCOS Live
+    # ISO without any OKD/FCOS related changes, we have to copy the manifests from OKD Machine OS manually to the
+    # bootstrap manifests folder of the rendezvous host.
+    # [0] https://access.redhat.com/solutions/4970731
+    mkdir -p /var/opt/openshift/manifests
+    cp -av /var/mnt/manifests/*.* /var/opt/openshift/manifests/
+
+    # Load new systemd unit files and configuration such as crio.service after mounting the content of OKD/FCOS Machine
+    # OS over /usr and copying new files to /etc
+    systemctl daemon-reload
```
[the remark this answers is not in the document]
JM1: Will this affect the services that were created by Ignition?

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
```diff
@@ -42,25 +42,67 @@ if [ ! -f /opt/openshift/.pivot-done ]; then
...
+    # [1] https://github.com/openshift/okd-machine-os/blob/28dec35d60ea07069366b22ebdcb296d429b15e9/overlay.d/99okd/etc/systemd/resolved.conf.d/okd-no-dns-stub.conf
+    if [ -e /etc/systemd/resolved.conf.d/okd-no-dns-stub.conf ]; then
+        systemctl restart systemd-resolved.service
+    else
+        systemctl disable --now systemd-resolved.service
+    fi
+
+    if systemctl list-unit-files -q fix-resolvconf.service >/dev/null; then
+        systemctl start fix-resolvconf.service
+        systemctl restart NetworkManager.service
```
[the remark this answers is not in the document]
JM1: we have to manually start fix-resolvconf.service because neither `daemon-reload` nor `preset-all` will start or stop any services.

andfasano:
> Great analysis, @andfasano, thank you ☺️
> 
> ~To fix the SELinux issue, we might have to [relabel like FCOS does during initrd run](https://github.com/coreos/fedora-coreos-config/blob/08b13806bb61faf85df2a93ffed9d7508ad47240/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-live/live-generator#L258).~ The SELinux issue did not happen at the rendezvous host but master-1, so it probably is something else. Besides that, my patch uses `setenforce Permissive" anyway..

--> @aleskandro :)

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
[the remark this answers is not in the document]
aleskandro: It should not... but it's worth a try imo.

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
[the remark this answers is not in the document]
aleskandro: Through daemon-reload and preset-all we will get the services (re-)configured and enabled.
In particular, fix-resolvconf.service will be started `before` NetworkManager.service starts. I'm worried that executing it while NM is running (before launching the restart, i.e., before NM stops) can give an uncertain behavior like filling the empty file with content given by the previous NM configuration and then NM doesn't update it again as it finds it in a good shape. See https://github.com/openshift/okd-machine-os/blob/master/overlay.d/99okd/usr/lib/systemd/system/fix-resolvconf.service#L4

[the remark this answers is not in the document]
JM1: Note that this whole script gets executed *only* at a live iso during cluster install. There will not be a reboot which would cause the new services to be launched. If we do not start `fix-resolvconf.service` manually, nobody else will. Restarting `NetworkManager.service` will not start `fix-resolvconf.service`.

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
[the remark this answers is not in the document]
JM1: Thinking about this again I noticed that `systemctl preset-all` will not have an visible effect for us because it will only enable or disable services. But we do not reboot nor soft-reboot, hence nobody will restart services except us.

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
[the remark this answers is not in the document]
aleskandro: @JM1 you're right. We should have added WantedBy=NetworkManager.service maybe to make what I want. In this case, can it be possible to run 
```
systemctl stop NetworkManager.service
systemctl start fix-resolvconf.service
systemctl start NetworkManager.service
```
to run exactly like in the case of the rebased system?

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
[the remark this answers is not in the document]
aleskandro: @JM1, enabling services with Wants/WantedBy or Requires/RequiredBy has an effect in the current system after systemctl preset-all. I still think it's worth it.

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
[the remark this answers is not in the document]
JM1: Done!

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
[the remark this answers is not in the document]
JM1: Done.

andfasano:
(https://github.com/openshift/machine-os-images/pull/33 landed, starting another round of agent jobs)

/test okd-e2e-agent-compact-ipv4
/test okd-e2e-agent-sno-ipv6
/test okd-e2e-agent-ha-dualstack

andfasano:
Retrying, since https://github.com/openshift/okd-machine-os/pull/706 landed

/test okd-e2e-agent-compact-ipv4
/test okd-e2e-agent-sno-ipv6
/test okd-e2e-agent-ha-dualstack

LorbusChris:
We'll have to fix the Samples Operator first: https://github.com/openshift/cluster-samples-operator/pull/525

JM1:
Do you believe that? okd-e2e-agent-compact-ipv4 and okd-e2e-agent-ha-dualstack have passed 😱

JM1:
Both jobs [okd-e2e-aws-ovn-upgrade](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/7484/pull-ci-openshift-installer-master-okd-e2e-aws-ovn-upgrade/1732718748986511360) and [okd-e2e-aws-ovn](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/7484/pull-ci-openshift-installer-master-okd-e2e-aws-ovn/1732718748948762624) complete deployment successfully, then run and pass e2e tests. Still they fail because some step is running into a timeout, but which one?!?

Job [okd-e2e-agent-sno-ipv6](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_installer/7484/pull-ci-openshift-installer-master-okd-e2e-agent-sno-ipv6/1732718748915208192) still fails for unknown reasons. It lacks logs to debug it properly.

andfasano:
Looks like we got a pretty solid green runs also on agent okd sno jobs!

/lgtm

PR #7486, opened 2023-09-14
OCPBUGS-18552: Truncate vlan names defined in nmstate if > 15 chars

bfournie:
Include the assisted-service change https://github.com/openshift/assisted-service/pull/5389 to ensure that vlan names don't exceed 15 chars into the agent-based installer.


andfasano:
Could be it worth at this point evaluating the possibility to vendor it (by previously refactoring properly the assisted-service related code)?

bfournie:
> Could be it worth at this point evaluating the possibility to vendor it (by previously refactoring properly the assisted-service related code)?

It would require moving the assisted-service script out of the internal dir I believe, see https://issues.redhat.com/browse/AGENT-81.

PR #7488, opened 2023-09-14
CORS-2836: Refactor Stages, encapsulate terraform

patrickdillon:
Due to Hashicorp's moving Terraform from an MPL to BSL license, the installer's usage of Terraform presents a risk that we would like to contain. This PR aims to:

1. **Contain Terraform**: moves all terraform dependencies to `pkg/terraform`
2. **Provide an alternative path**:  introduces a generic `infrastructure.Stages` interface (based on `terraform.Stages`) 

The existing code couples the Cluster Asset to terraform dependencies:
```mermaid
flowchart LR
   a[Cluster Asset] --- b[Terraform] & c[Stages]
   b --- c
```

This PR creates an interface to decouple the cluster asset from Terraform dependencies; and, potentially, allow other infrastructure providers.
```mermaid
flowchart LR
   a[Cluster Asset] --- b[Stages Infrastructure Interface] --- c[Terraform] & d[Cloud SDKs] & e[CAPI]
```

https://github.com/openshift/installer/pull/7534/commits/fc87c7503a0380e1e79b9fd40678ff39d968ed7f demonstrates how we can use this approach to use build tags to select infrastructure providers when building the binary, allowing us to produce a terraform-free binary with alternate infrastructure providers.

~~The interface design (laid out in the first commit) is still evocative of terraform because:~~
~~* it uses **stages**, which is a way we grouped resource configuration to allow destruction of bootstrap resources with terraform destroy~~
~~* it expects **tfVars** as inputs to the infrastructure provider~~

~~Both of these opinionated implementations seem easy enough to work with on behalf of other infrastructure providers, but we can explore collapsing these ideas further into the interface.~~

### DIFF
--- a/pkg/asset/cluster/cluster.go
+++ b/pkg/asset/cluster/cluster.go
@@ -3,11 +3,9 @@ package cluster
 import (
 	"context"
 	"fmt"
-	"os"
 	"path/filepath"
 	"strings"
 
-	"github.com/hashicorp/terraform-exec/tfexec"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
@@ -18,9 +16,7 @@ import (
 	"github.com/openshift/installer/pkg/asset/installconfig"
 	"github.com/openshift/installer/pkg/asset/password"
 	"github.com/openshift/installer/pkg/asset/quota"
-	"github.com/openshift/installer/pkg/metrics/timer"
-	"github.com/openshift/installer/pkg/terraform"
-	platformstages "github.com/openshift/installer/pkg/terraform/stages/platform"
+	infra "github.com/openshift/installer/pkg/infrastructure/platform"
 	typesaws "github.com/openshift/installer/pkg/types/aws"
 	typesazure "github.com/openshift/installer/pkg/types/azure"
 	typesopenstack "github.com/openshift/installer/pkg/types/openstack"
@@ -92,21 +88,6 @@ func (c *Cluster) Generate(parents asset.Parents) (err error) {
 		platform = typesazure.StackTerraformName
 	}
 
-	stages := platformstages.StagesForPlatform(platform)
-
-	terraformDir := filepath.Join(InstallDir, "terraform")
-	if err := os.Mkdir(terraformDir, 0777); err != nil {
-		return errors.Wrap(err, "could not create the terraform directory")
-	}
-
-	terraformDirPath, err := filepath.Abs(terraformDir)
-	if err != nil {
-		return errors.Wrap(err, "cannot get absolute path of terraform directory")
-	}
-
-	defer os.RemoveAll(terraformDir)
-	terraform.UnpackTerraform(terraformDirPath, stages)
-
 	logrus.Infof("Creating infrastructure resources...")
 	switch platform {
 	case typesaws.Name:
@@ -123,18 +104,16 @@ func (c *Cluster) Generate(parents asset.Parents) (err error) {
 		}
 	}
 
-	tfvarsFiles := make([]*asset.File, 0, len(terraformVariables.Files())+len(stages))
+	tfvarsFiles := []*asset.File{}
 	for _, file := range terraformVariables.Files() {
 		tfvarsFiles = append(tfvarsFiles, file)
 	}
 
-	for _, stage := range stages {
-		outputs, err := c.applyStage(platform, stage, terraformDirPath, tfvarsFiles)
-		if err != nil {
-			return errors.Wrapf(err, "failure applying terraform for %q stage", stage.Name())
-		}
-		tfvarsFiles = append(tfvarsFiles, outputs)
-		c.FileList = append(c.FileList, outputs)
+	provider := infra.ProviderForPlatform(platform)
+	files, err := provider.Provision(InstallDir, tfvarsFiles)
+	c.FileList = append(c.FileList, files...) // append state files even in case of failure
+	if err != nil {
+		return fmt.Errorf("%s: %w", asset.ClusterCreationError, err)
 	}
 
 	return nil
@@ -157,56 +136,4 @@ func (c *Cluster) Load(f asset.FileFetcher) (found bool, err error) {
 	}
 
 	return false, nil
-}
-
-func (c *Cluster) applyStage(platform string, stage terraform.Stage, terraformDir string, tfvarsFiles []*asset.File) (*asset.File, error) {
-	// Copy the terraform.tfvars to a temp directory which will contain the terraform plan.
-	tmpDir, err := os.MkdirTemp("", fmt.Sprintf("openshift-install-%s-", stage.Name()))
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to create temp dir for terraform execution")
-	}
-	defer os.RemoveAll(tmpDir)
-
-	var extraOpts []tfexec.ApplyOption
-	for _, file := range tfvarsFiles {
-		if err := os.WriteFile(filepath.Join(tmpDir, file.Filename), file.Data, 0o600); err != nil {
-			return nil, err
-		}
-		extraOpts = append(extraOpts, tfexec.VarFile(filepath.Join(tmpDir, file.Filename)))
-	}
-
-	return c.applyTerraform(tmpDir, platform, stage, terraformDir, extraOpts...)
-}
-
-func (c *Cluster) applyTerraform(tmpDir string, platform string, stage terraform.Stage, terraformDir string, opts ...tfexec.ApplyOption) (*asset.File, error) {
-	timer.StartTimer(stage.Name())
-	defer timer.StopTimer(stage.Name())
-
-	applyErr := terraform.Apply(tmpDir, platform, stage, terraformDir, opts...)
-
-	// Write the state file to the install directory even if the apply failed.
-	if data, err := os.ReadFile(filepath.Join(tmpDir, terraform.StateFilename)); err == nil {
-		c.FileList = append(c.FileList, &asset.File{
-			Filename: stage.StateFilename(),
-			Data:     data,
-		})
-	} else if !os.IsNotExist(err) {
-		logrus.Errorf("Failed to read tfstate: %v", err)
-		return nil, errors.Wrap(err, "failed to read tfstate")
-	}
-
-	if applyErr != nil {
-		return nil, errors.Wrap(applyErr, asset.ClusterCreationError)
-	}
-
-	outputs, err := terraform.Outputs(tmpDir, terraformDir)
-	if err != nil {
-		return nil, errors.Wrapf(err, "could not get outputs from stage %q", stage.Name())
-	}
-
-	outputsFile := &asset.File{
-		Filename: stage.OutputsFilename(),
-		Data:     outputs,
-	}
-	return outputsFile, nil
 }
--- a/pkg/destroy/bootstrap/bootstrap.go
+++ b/pkg/destroy/bootstrap/bootstrap.go
@@ -3,16 +3,13 @@ package bootstrap
 
 import (
 	"fmt"
-	"os"
-	"path/filepath"
 
 	"github.com/pkg/errors"
 
 	"github.com/openshift/installer/pkg/asset/cluster"
 	openstackasset "github.com/openshift/installer/pkg/asset/cluster/openstack"
 	osp "github.com/openshift/installer/pkg/destroy/openstack"
-	"github.com/openshift/installer/pkg/terraform"
-	platformstages "github.com/openshift/installer/pkg/terraform/stages/platform"
+	infra "github.com/openshift/installer/pkg/infrastructure/platform"
 	typesazure "github.com/openshift/installer/pkg/types/azure"
 	"github.com/openshift/installer/pkg/types/openstack"
 )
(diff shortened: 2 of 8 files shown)


pkg/destroy/bootstrap/bootstrap.go
```diff
@@ -45,68 +41,69 @@ func Destroy(dir string) (err error) {
...
-
-		if err := stage.Destroy(tempDir, terraformDirPath, targetVarFiles); err != nil {
-			return err
-		}
-
-		if err := copy(stateFilePathInTempDir, stateFilePathInInstallDir); err != nil {
-			return errors.Wrap(err, "failed to copy state file from the temporary directory")
-		}
-	}
+	// TODO
```
patrickdillon, line 44: once we port this over to the new interface (which I will try to do tomorrow), we should be able to test with e2e to check for regressions

I have tested locally on aws and confirmed terraform was able to provision infra

pkg/infrastructure/provider.go
```diff
@@ -0,0 +1,36 @@
...
+	// fileList: the file list for the cluster asset.
+	// returns tfvars output file and a state file
+	Provision(tfVars, fileList []*asset.File) (*asset.File, *asset.File, error)
+
+	// DestroyWithBootstrap is true if the stage should be destroyed when destroying the bootstrap resources.
+	DestroyWithBootstrap() bool
+
+	// Destroy destroys the resources created in the stage. This should only be called if the stage should be destroyed
+	// when destroying the bootstrap resources.
+	Destroy(directory string, terraformDir string, varFiles []string) error
```
jhixson74, line 32: It would be nice to abstract this further by gutting anything terraform from this interface.

jhixson74 (commented):
I don't have much to say since this is a WIP. I like the general direction this is going. It would be nice to abstract away infrastructure from terrafrom and be more generic so we can plug in any kind of "infrastructure" we want. While here it would also be nice to make the stages a little more flexible so we can plug into that at any point during the install process. Too much to ask? ;-)

patrickdillon:
Taking this off draft to run CI tests. Will update the commit structure to facilitate review ASAP. Also want to continue iterating on the interface design + https://github.com/openshift/installer/pull/7488#discussion_r1333853278

patrickdillon:
/hold

That ran ALOT of tests.

patrickdillon:
Build flag work has been moved to #7534

pkg/infrastructure/alibabacloud/alibabacloud.go
```diff
@@ -0,0 +1,10 @@
+package alibabacloud
+
+import (
+	"github.com/openshift/installer/pkg/infrastructure"
+	"github.com/openshift/installer/pkg/terraform/stages/alibabacloud"
```
patrickdillon, line 5: This level of indirection--having a separate provider package pointing to terraform--is not helpful. Instead we should remove all of these infrastructure/provider packages (we only need them with new infra providers) and just have ProviderForPlatform point directly to terraform like in: https://github.com/openshift/installer/pull/7534/commits/32a178c4e5cd53baf602b250539e2da63aace1e8

pkg/asset/cluster/cluster.go
```diff
@@ -129,12 +115,13 @@ func (c *Cluster) Generate(parents asset.Parents) (err error) {
...
 	for _, stage := range stages {
-		outputs, err := c.applyStage(platform, stage, terraformDirPath, tfvarsFiles)
+		outputs, stateFile, err := stage.Provision(tfvarsFiles, c.FileList) //TODO: make sure c.FileList ends up being used
 		if err != nil {
-			return errors.Wrapf(err, "failure applying terraform for %q stage", stage.Name())
+			errors.Wrapf(err, "provisioning infrastructure in stage %s", stage.Name())
 		}
 		tfvarsFiles = append(tfvarsFiles, outputs)
 		c.FileList = append(c.FileList, outputs)
+		c.FileList = append(c.FileList, stateFile)
```
patrickdillon, line 124: 1. Handle nil values like https://github.com/openshift/installer/pull/7534/commits/955a030c528a4b8237d916b6cd98e79bd2961cf6
2. Further consider the expectations around `outputs` & `statefile`

patrickdillon:
Not sure if the stages abstraction is helpful--although it is easy. We could make the interface more generic and not have stages... but that is simple enough to replicate by just having a single stage.

pkg/infrastructure/alibabacloud/alibabacloud.go
patrickdillon: > Is the plan to have the infrastructure package contain the non-terraform ways to layout the infrastructure? For AWS, would it contain the cloud-sdk calls?

Yes your summary/understanding is exactly right. 

I have refactored with the change from https://github.com/openshift/installer/commit/32a178c4e5cd53baf602b250539e2da63aace1e8 so I think this should be a little less convoluted now.

pkg/infrastructure/provider.go
```diff
@@ -0,0 +1,28 @@
...
+// Stage is an individual stage of infrastructure provisioning.
+type Stage interface {
+	// Name is the name of the stage.
+	Name() string
+
+	// Provision creates the infrastructure resources for the stage.
+	// tfVars: the terraform variables files
+	// fileList: the file list for the cluster asset.
+	// returns tfvars output file and a state file
+	Provision(tfVars, fileList []*asset.File) (*asset.File, *asset.File, error)
```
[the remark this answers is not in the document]
patrickdillon: I will lay this out in more detail, because this part is definitely a little confusing, but a quick answer: The design here is still using `tfVars` as the input format to the infrastructure provider, whether it's terraform or not. `tfVars` is just a json blob in the format expected by the terraform configs. The `tfvars` package does not have any terraform dependencies to generate the files. So the UX (or developer experience) is still tied to terraform, but there is not a code dependency. 

In the AWS poc, the new infra provider deserializes the TFVars into a couple of variables and it seems pretty nice to work with:

```go
	clusterConfig := &tfvars.Config{}
	clusterAWSConfig := &tfvarsaws.Config{}
	for _, file := range tfvarsFiles {
		if file.Filename == "terraform.tfvars.json" {
			if err := json.Unmarshal(file.Data, clusterConfig); err != nil {
				return nil, nil, err
			}
		}

		if file.Filename == "terraform.platform.auto.tfvars.json" {
			if err := json.Unmarshal(file.Data, clusterAWSConfig); err != nil {
				return nil, nil, err
			}
		}
	}
```

Another downside, besides the UX, might be that we're deserializing/serializing unnecessarily. I will write this up in a design doc and see if we can find room for improvement.

pkg/terraform/stages/alibabacloud/stages.go
```diff
@@ -1,39 +1,65 @@
...
+			[]providers.Provider{providers.AliCloud},
+			stages.WithCustomBootstrapDestroy(removeFromLoadBalancers),
+		),
+	}
+
+	// It would be nice to not need to repeat this for each platform but at this stage
+	// Perfect is the enemy of good
+	terraform.UnpackTerraform(terraformDir, platformStages)
+
+	cleanup := func() error {
```
zaneb, line 58: nit: could potentially create a terraform.Cleanup() function for this to avoid having to repeat the inline func for each platform. Or return it from UnpackTerraform().

zaneb (commented):
At a high level, this looks very reasonable to me.

pkg/asset/cluster/cluster.go
```diff
@@ -129,12 +115,13 @@ func (c *Cluster) Generate(parents asset.Parents) (err error) {
 	}
 
 	for _, stage := range stages {
-		outputs, err := c.applyStage(platform, stage, terraformDirPath, tfvarsFiles)
+		outputs, stateFile, err := stage.Provision(tfvarsFiles, c.FileList) //TODO: make sure c.FileList ends up being used
 		if err != nil {
-			return errors.Wrapf(err, "failure applying terraform for %q stage", stage.Name())
+			errors.Wrapf(err, "provisioning infrastructure in stage %s", stage.Name())
```
r4f4, line 120:
```suggestion
			return errors.Wrapf(err, "provisioning infrastructure in stage %s", stage.Name())
```

pkg/terraform/stages/alibabacloud/stages.go
patrickdillon: After moving stages inside the interface this is no longer an issue.

patrickdillon:
I think this is much better  and simpler to not have stages as part of the interface.

@r4f4 @rwsu @zaneb @jcpowermac @jhixson74 PTAL

r4f4:
> I think this is much better and simpler to not have stages as part of the interface.
> 
> @r4f4 @rwsu @zaneb @jcpowermac @jhixson74 PTAL

Agreed. A provider is free to do the provision in stages if it so desire.

jcpowermac:
> I think this is much better and simpler to not have stages as part of the interface.
> 
> @r4f4 @rwsu @zaneb @jcpowermac @jhixson74 PTAL

lgtm...the poc for vsphere is a single flow so this works out better...and simpler

patrickdillon:
As the direction of using CAPI in the installer has become more concrete in https://github.com/openshift/installer/pull/7579 I think it will probably be better for CAPI provisioning to not use the `Provision` interface as the process works directly with Assets (manifests, kubeconfig, ignition, etc), `BootstrapDestroy` and `ExtractHostAddresses` would be useful though. 
g
We could fit the CAPI provisioning model into the infrastructure interface, but I think the goal should be on containing terraform rather than defining a perfect interface. This interface could potentially be temporary once terraform is entirely replaced. We can consider this more in future PRs.

zaneb:
> As the direction of using CAPI in the installer has become more concrete in #7579 I think it will probably be better for CAPI provisioning to not use the `Provision` interface as the process works directly with Assets (manifests, kubeconfig, ignition, etc), `BootstrapDestroy` and `ExtractHostAddresses` would be useful though.

Does this suggest we should consider making the Assets the input to `Provision()` and let e.g. `terraform.InitializeProvider()` do the translation to Terraform vars?

jcpowermac:
> As the direction of using CAPI in the installer has become more concrete in #7579 I think it will probably be better for CAPI provisioning to not use the `Provision` interface as the process works directly with Assets (manifests, kubeconfig, ignition, etc),

Does this still allow for potential of prerequisite infrastructure tasks to be performed. For example importing the ova doesn't seem possible in capi/capv, but I certainly could be wrong.

jhixson74:
This looks good to me. I would like to know the same as @jcpowermac about ability of prerequisite infrastructure tasks to be performed.

patrickdillon:
> Does this suggest we should consider making the Assets the input to `Provision()` and let e.g. `terraform.InitializeProvider()` do the translation to Terraform vars?

Right. That is the current implementation: `Provision` accepts Assets and the provider translates them to what it needs. The approach seems unnecessarily clunky, but for terraform  and the aws sdk installer the clunkiness is easily manageable because we know we are always just getting two tfvars viles. 

Using CAPI for installs we depend on multiple kinds of manifests, so doing this translation is more cumbersome. My initial comment was triggered by the thought that it is simply unnecessary, but it's a problem that could be handled.

patrickdillon:
> > As the direction of using CAPI in the installer has become more concrete in #7579 I think it will probably be better for CAPI provisioning to not use the `Provision` interface as the process works directly with Assets (manifests, kubeconfig, ignition, etc),
> 
> Does this still allow for potential of prerequisite infrastructure tasks to be performed? For example importing the ova doesn't seem possible in capi/capv, but I certainly could be wrong.

That's more of a question for the capi-installer specific implementation but the answer is yes. For the aws install, it is necessary to create IAM roles using SDK calls in advance of the install and DNS records afterward.

> This looks good to me. I would like to know the same as @jcpowermac about ability of prerequisite infrastructure tasks to be performed.

Yes, but given the context we have from other conversations, I think you might be interested in editing bootstrap ignition. Ignition will continue to be a dependency of the cluster, so it will be generated in advance of the cluster infrastructure.

r4f4 (commented):
/lgtm

patrickdillon:
/hold
for testing

zaneb:
>> Does this suggest we should consider making the Assets the input to Provision() and let e.g. terraform.InitializeProvider() do the translation to Terraform vars?
>
> Right. That is the current implementation: Provision accepts Assets and the provider translates them to what it needs. The approach seems unnecessarily clunky, but for terraform and the aws sdk installer the clunkiness is easily manageable because we know we are always just getting two tfvars files.

This implies that any new implementations of the interface have to accept TFVars as the input to Provision() in order to co-exist with the Terraform one. What I'm asking is whether we should pass the assets that are dependencies of TFVars (which are presumably the same ones we'd need for e.g. CAPI providers) to Provision(), and have the conversion to TFVars happen behind that API instead of in front of it.

patrickdillon:
> Right. That is the current implementation: Provision accepts Assets and the provider translates them to what it needs. The approach seems unnecessarily clunky, but for terraform and the aws sdk installer the clunkiness is easily manageable because we know we are always just getting two tfvars files.
> 
> This implies that any new implementations of the interface have to accept TFVars as the input to Provision() in order to co-exist with the Terraform one. 

I just updated the PR description. The most recent version just expects generic assets:

```go
   // Provision creates the infrastructure resources for the stage.
   // dir: the path of the install dir
   // vars: cluster configuration input variables, such as terraform variables files
   // returns a slice of File assets, which will be appended to the cluster asset file list.
   Provision(dir string, vars []*asset.File) ([]*asset.File, error)
```
> What I'm asking is whether we should pass the assets that are dependencies of TFVars (which are presumably the same ones we'd need for e.g. CAPI providers) to Provision(), and have the conversion to TFVars happen behind that API instead of in front of it.

The current implementation allows for this. By keeping tfvars outside the interface we get the benefit of passing using them as inputs to other providers, as is the case for the aws sdk implementation.

patrickdillon:
/retest
Jobs failed due to failure to pull images in pre-install steps.

PR #7491, opened 2023-09-15
OCPBUGS-18830: AWS terraform bootstrap destroy will not refresh state

barbacbd:
Destroy AWS bootstrap will now use the terraform option -refresh=false.
https://developer.hashicorp.com/terraform/cloud-docs/run/modes-and-options#skipping-automatic-state-refresh Adding the option so that the terraform state is not refreshed. This should counter the requirement to bump the terraform aws provider for DescribeSecurityGroupRule changes that we have noticed since the bump to 5.4.0.

### DIFF
--- a/pkg/terraform/stages/aws/stages.go
+++ b/pkg/terraform/stages/aws/stages.go
@@ -1,9 +1,14 @@
 package aws
 
 import (
+	"github.com/hashicorp/terraform-exec/tfexec"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+
 	"github.com/openshift/installer/pkg/terraform"
 	"github.com/openshift/installer/pkg/terraform/providers"
 	"github.com/openshift/installer/pkg/terraform/stages"
+	awstypes "github.com/openshift/installer/pkg/types/aws"
 )
 
 // PlatformStages are the stages to run to provision the infrastructure in AWS.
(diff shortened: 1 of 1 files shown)


barbacbd:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.14-amd64-nightly-aws-sc2s-ipi-disconnected-private-fips-f14

pkg/terraform/stages/aws/stages.go
```diff
@@ -17,6 +21,18 @@ var PlatformStages = []terraform.Stage{
...
+		stages.WithCustomBootstrapDestroy(customBootstrapDestroy),
 	),
 }
+
+func customBootstrapDestroy(s stages.SplitStage, directory string, terraformDir string, varFiles []string) error {
+	opts := make([]tfexec.DestroyOption, 0, len(varFiles)+1)
+	for _, varFile := range varFiles {
+		opts = append(opts, tfexec.VarFile(varFile))
+	}
+	opts = append(opts, tfexec.Refresh(false))
```
sadasu, line 33: Can we add a comment + a debug message to say we are changing default behavior by specifically setting refresh=False because that was resulting in Terraform trying to describe fields (specifically security group rules) which is not an allowed operation in some regions like CS2. So, will help us in the future if we hit some unintended consequences of setting refresh=false.

r4f4 (commented):
/lgtm

PR #7492, opened 2023-09-16
OCPBUGS-19086: Check if nmstatectl executable exists in the system

kevydotvinu:
- Check if nmstatectl executable exists
- Print `install nmstate` 

Fixes OCPBUGS-19086


pawanpinjarkar (approved):
/approve

PR #7500, opened 2023-09-19
OCPBUGS-18690: ic: azure: validate NVMe-only family types

r4f4:
NVMe-only controller type VM families are not yet supported by the azure terraform provider [1]. This commit adds a check during install validation so those VM families cannot be selected, which would cause the install to fail with
```
09-07 17:55:57.613  level=error msg=Error: creating Linux Virtual Machine: (Name "jima-test-wlgrr-bootstrap" / Resource Group "jima-test-wlgrr-rg"): compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidParameter" Message="The VM size 'Standard_E112ibs_v5' cannot boot with OS image or disk. Please check that disk controller types supported by the OS image or disk is one of the supported disk controller types for the VM size 'Standard_E112ibs_v5'. Please query sku api at https://aka.ms/azure-compute-skus  to determine supported disk controller types for the VM size." Target="vmSize"
```

[1] https://github.com/hashicorp/terraform-provider-azurerm/issues/22058

### DIFF
--- a/pkg/asset/installconfig/azure/validation.go
+++ b/pkg/asset/installconfig/azure/validation.go
@@ -162,9 +162,19 @@ func validateFamily(fieldPath *field.Path, instanceType, family string) field.Er
 	windowsVMFamilies := sets.NewString(
 		"standardNVSv4Family",
 	)
+	diskNVMeVMFamilies := sets.NewString(
+		"standardEIBDSv5Family",
+		"standardEIBSv5Family",
+	)
 	allErrs := field.ErrorList{}
 	if windowsVMFamilies.Has(family) {
 		errMsg := fmt.Sprintf("%s is currently only supported on Windows", family)
+		allErrs = append(allErrs, field.Invalid(fieldPath, instanceType, errMsg))
+	}
+	// FIXME: remove when supported has been added to the provider
+	// https://github.com/hashicorp/terraform-provider-azurerm/issues/22058
+	if diskNVMeVMFamilies.Has(family) {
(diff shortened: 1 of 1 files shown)


barbacbd:
Reference: 
```A mix of this https://learn.microsoft.com/en-us/azure/virtual-machines/vm-naming-conventions#naming-convention-explanation + https://learn.microsoft.com/en-us/azure/virtual-machines/ebdsv5-ebsv5-series + the az cli.```

**Review by barbacbd — COMMENTED** [review]
/lgtm

File: pkg/asset/installconfig/azure/validation.go [hunk 1]
```diff
@@ -162,11 +162,21 @@ func validateFamily(fieldPath *field.Path, instanceType, family string) field.Er
 	windowsVMFamilies := sets.NewString(
 		"standardNVSv4Family",
 	)
+	diskNVMeVMFamilies := sets.NewString(
```
rwsu, line 165:
```suggestion
	unsupportedDiskNVMeVMFamilies := sets.NewString(
```
I would suggest calling it unsupportedDiskNVMeVMFamilies

rwsu (commented):
Looks good. I would suggest renaming the variable to make it more clear.

pkg/asset/installconfig/azure/validation.go
r4f4: I understand but then it won't match the existing pattern (e.g, `windowsVMFamilies`), so I'd rather not change that.

PR #7503, opened 2023-09-19
Power VS: Change server deletion criteria

mjturek:
Currently we only delete VMs whose names contain the InfraID. However the machine API can launch VMs that don't meet this criteria. [0]

To ensure that we delete all VMs related to a cluster, we will instead delete any VM that uses the boot image imported for the cluster.

[0] https://github.com/openshift/verification-tests/blob/b7c9c081e619c01db65aa44f6573e6f022bc30a1/features/upgrade/machines/machines.feature#L142-L156

### DIFF
--- a/pkg/destroy/powervs/power-instance.go
+++ b/pkg/destroy/powervs/power-instance.go
@@ -18,6 +18,25 @@ const (
 func (o *ClusterUninstaller) listPowerInstances() (cloudResources, error) {
 	o.Logger.Debugf("Listing virtual Power service instances (%s)", o.InfraID)
 
+	// Get the image id
+	images, err := o.listImages()
+	imageID := ""
+	if err != nil && len(images.list()) > 0 {
+		imageID = images.list()[0].id
+		if len(images.list()) > 1 {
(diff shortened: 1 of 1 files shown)


mkumatag (changes requested):
we shouldn't be doing this, I see a threat doing this. There are scenario where you are using the same boot image for multiple clusters or even image can be remove post installation(there is no hard dependency on keeping the image).

Rather we should delete the machinesets gracefully so that it knows how to take care of this during the uninstallation.

/hold

mjturek:
@mkumatag I've changed this to also check the network that's used. Basically if a server uses any resource that was created for the cluster, it will get deleted.

AshwinHIBM:
lgtm

barbacbd (commented):
Is it possible to tag/label these instances and use that as a lookup?

sadasu:
@mjturek do you still need these changes?

mjturek:
> @mjturek do you still need these changes?

I think we can close this one for now. We could not come to a consensus on the right solution here. If we decide on the correct implementation in the future, we'll open a new PR. Thanks!

PR #7504, opened 2023-09-19
Improve API and Ingress VIPs validation

MaysaMacedo:
This commit enhance VIPs validation to ensure the primary IP family of the VIPs matches the primary IP family of all the network fields.

### DIFF
--- a/pkg/types/validation/installconfig.go
+++ b/pkg/types/validation/installconfig.go
@@ -632,7 +632,7 @@ func validateAPIAndIngressVIPs(vips vips, fieldNames vipFields, vipIsRequired, r
 			allErrs = append(allErrs, field.Required(fldPath.Child(fieldNames.IngressVIPs), "must specify VIP for ingress, when VIP for API is set"))
 		}
 
-		if len(vips.API) == 1 {
+		if len(vips.API) >= 1 {
 			hasIPv4, hasIPv6, presence, _ := inferIPVersionFromInstallConfig(n)
 
 			apiVIPIPFamily := corev1.IPv4Protocol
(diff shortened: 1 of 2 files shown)


mkowalski:
This whole `validateAPIAndIngressVIPs` became an ugly monster, we should once sit and rewrite it from scratch. Nevertheless, your change is valid and a good one

/lgtm

pkg/types/validation/installconfig.go
```diff
@@ -675,20 +683,28 @@ func validateAPIAndIngressVIPs(vips vips, fieldNames vipFields, vipIsRequired, r
 			allErrs = append(allErrs, field.Required(fldPath.Child(fieldNames.APIVIPs), "must specify VIP for API, when VIP for ingress is set"))
 		}
 
-		if len(vips.Ingress) == 1 {
```
cybertron, line 678: Huh. Was this a bug? If there's only one VIP we could never have had both v4 and v6 so the checks below would always fail.

cybertron (commented):
This looks reasonable, although I will note that the VIPs don't care about any of the other networks. However, since we already have the logic here I'm okay with using it to ensure consistency across all the networks.

pkg/types/validation/installconfig.go
MaysaMacedo: Yeah, looks like it

pkg/types/validation/installconfig_test.go
```diff
@@ -286,6 +286,34 @@ func validDualStackNetworkingConfig() *types.Networking {
 	}
 }
 
+func InvalidPrimaryV6DualStackNetworkingConfig() *types.Networking {
```
MaysaMacedo, line 289: This represent the scenario where keepalived was not happy because I forgot to change the order of `serviceNetwork` to have v6 primary address.

PR #7506, opened 2023-09-20
OCPBUGS-19444: Use changes to AgentClusterInstall during loading

zaneb:
In the Load() method of AgentClusterInstall, we sometimes make changes to the input data to maintain backwards compatibility and such. However, these changes had no effect because we always used the original file read from disk as the generated output, even when it didn't match the actual config.

Regenerate the data to use in the Agent ISO at the conclusion of loading.

### DIFF
--- a/pkg/asset/agent/manifests/agentclusterinstall.go
+++ b/pkg/asset/agent/manifests/agentclusterinstall.go
@@ -235,15 +235,6 @@ func (a *AgentClusterInstall) Generate(dependencies asset.Parents) error {
 
 		a.Config = agentClusterInstall
 
-		agentClusterInstallData, err := yaml.Marshal(agentClusterInstall)
-		if err != nil {
-			return errors.Wrap(err, "failed to marshal agent installer AgentClusterInstall")
-		}
-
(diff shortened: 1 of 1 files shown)


pkg/asset/agent/manifests/agentclusterinstall.go
```diff
@@ -323,6 +312,15 @@ func (a *AgentClusterInstall) finish() error {
 		return errors.Wrapf(err, "invalid PlatformType configured")
 	}
 
+	agentClusterInstallData, err := yaml.Marshal(agentClusterInstall)
```
bfournie, line 315: Will need to pass in `agentClusterInstall` (or add it to` AgentClusterInstall struct`).

zaneb: :man_facepalming: thanks.
It's already passed in, just as `a.Config`

zaneb:
e2e jobs all timed out for some reason
/retest

pawanpinjarkar (commented):
/lgtm

PR #7508, opened 2023-09-20
AGENT-710: Use invoker for bootstrap template generation

andfasano:
Some of the scripts generated by the `create ignition-configs` command for OKD may require to discriminate the cases where SNO is being managed or agent-based installer is currently used, thus adding the `OPENSHIFT_INSTALL_INVOKER` env var in the template inputs.

### DIFF
--- a/data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
+++ b/data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
@@ -42,7 +42,7 @@ if [ ! -f /opt/openshift/.pivot-done ]; then
   record_service_stage_start "rebase-to-okd-os-image"
 {{if .IsFCOS -}}
   mnt="$(podman image mount "${MACHINE_OS_IMAGE}")"
-{{- if .BootstrapInPlace }}
+{{- if or (.BootstrapInPlace) (eq .Invoker "agent-installer") }}
   # SNO setup boots into Live ISO which cannot be rebased
   # https://github.com/coreos/rpm-ostree/issues/4547
   mkdir /var/mnt/{upper,worker}
--- a/data/data/bootstrap/systemd/units/release-image-pivot.service.template
+++ b/data/data/bootstrap/systemd/units/release-image-pivot.service.template
@@ -3,7 +3,7 @@
 Description=Pivot bootstrap to the OpenShift Release Image
 Wants=release-image.service
 After=release-image.service
-{{- if .BootstrapInPlace }}
+{{- if or (.BootstrapInPlace) (eq .Invoker "agent-installer") }}
 Before=bootkube.service kubelet.service
 {{ else }}
 Before=bootkube.service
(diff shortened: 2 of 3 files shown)


data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
```diff
@@ -42,7 +42,7 @@ if [ ! -f /opt/openshift/.pivot-done ]; then
   record_service_stage_start "rebase-to-okd-os-image"
 {{if .IsFCOS -}}
   mnt="$(podman image mount "${MACHINE_OS_IMAGE}")"
-{{- if .BootstrapInPlace }}
+{{- if or (.BootstrapInPlace) (eq .Invoker "agent-installer") }}
```
JM1, line 45: IMHO this Go conditional should be replaced with a proper shell conditional which checks whether we are on a Live ISO or not. [Example](https://github.com/openshift/installer/pull/7484).

data/data/bootstrap/systemd/units/release-image-pivot.service.template
```diff
@@ -3,7 +3,7 @@
 Description=Pivot bootstrap to the OpenShift Release Image
 Wants=release-image.service
 After=release-image.service
-{{- if .BootstrapInPlace }}
+{{- if or (.BootstrapInPlace) (eq .Invoker "agent-installer") }}
```
JM1, line 6: [Previous Discussion](https://github.com/openshift/installer/pull/7484#pullrequestreview-1631108866).

JM1 (changes requested):

data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh.template
JM1: Because all OCP code is based on RHCOS and thus often assumes that certain tools such as `oc` and services like `crio.service` are available.

andfasano: IINW it's not possible to use the shell check everywhere (for example, in the service definition). The `Invoker` attribute may provide more flexibility when generating the templates

JM1: Actually we do not need the shell check in the service definition, we can just drop the complete condition there ([previous discussion](https://github.com/openshift/installer/pull/7484#pullrequestreview-1631108866)). But I agree that your patch implements missing and useful functionality which should be merged. Sorry for derailing the discussion of this patch, [let us continue it in the initial PR](https://github.com/openshift/installer/pull/7484).

JM1: @andfasano I cannot mark this as resolved, i think you have to do it.

data/data/bootstrap/systemd/units/release-image-pivot.service.template
JM1: This is only meant as a reference, it is not supposed to block this patch.

JM1:
@andfasano: Do we want to cherry-pick this to 4.14 (and 4.13) in order to prepare the OKD ABI fix series (e.g. #7516) for older releases?

PR #7509, opened 2023-09-20
WIP: Custom DNS changes for AWS

sadasu:


pkg/terraform/stages/aws/stages.go
```diff
@@ -41,3 +52,37 @@ func customBootstrapDestroy(s stages.SplitStage, directory string, terraformDir
 		"failed to destroy bootstrap",
 	)
 }
+
+func extractAWSLBConfig(s stages.SplitStage, directory string, config *types.InstallConfig) ([]string, []string, error) {
```
barbacbd, line 56: If the user is going to supply load balancer (internal and external) information, why do we need to extract it from terraform? Shouldn't terraform skip any creation of this data, so we would already know this information and can populate any structure without the output of terraform right ?

pkg/terraform/stages/aws/stages.go
```diff
@@ -41,3 +52,37 @@ func customBootstrapDestroy(s stages.SplitStage, directory string, terraformDir
...
+	var apiLBDNSNames []string
+	if apiLBDNSNameRaw, ok := outputs["aws_external_api_lb_dns_name"]; ok {
+		apiLBDNSNames[0], ok = apiLBDNSNameRaw.(string)
+		if !ok {
+			return nil, nil, errors.New("could not read API LB DNS Name from terraform outputs")
+		}
+	}
+
+	var apiIntLBDNSNames []string
+	if apiLBDNSNameRaw, ok := outputs["aws_external_api_lb_dns_name"]; ok {
```
barbacbd, line 81: Looks like a copy/paste here, probably want the internal name.

pkg/terraform/stages/aws/stages.go
sadasu: We donot expect the user to provide LB information anymore. They will not pre-create these resources. Instead, Installer will create these resources like a regular IPI install and only skip route53 configuration.

sadasu:
/close

AWS implementation should be based on the CAPI work.

PR #7510, opened 2023-09-20
OCPBUGS-18986: Tag aws instance profiles.

barbacbd:
** Instance profiles are not always cleaned up correctly. Tags are available for these resources. Tag the resources and let the destroy functions find the resources by tag to be removed.

### DIFF
--- a/data/data/aws/bootstrap/main.tf
+++ b/data/data/aws/bootstrap/main.tf
@@ -84,6 +84,13 @@ resource "aws_iam_instance_profile" "bootstrap" {
   name = "${var.cluster_id}-bootstrap-profile"
 
   role = var.aws_master_iam_role_name != "" ? var.aws_master_iam_role_name : aws_iam_role.bootstrap[0].name
+
+  tags = merge(
+    {
+      "Name" = "${var.cluster_id}-bootstrap-profile"
+    },
+    local.tags,
+  )
 }
 
 resource "aws_iam_role" "bootstrap" {
--- a/data/data/aws/cluster/iam/main.tf
+++ b/data/data/aws/cluster/iam/main.tf
@@ -8,6 +8,13 @@ resource "aws_iam_instance_profile" "worker" {
   name = "${var.cluster_id}-worker-profile"
 
   role = var.worker_iam_role_name != "" ? var.worker_iam_role_name : aws_iam_role.worker_role[0].name
+
+  tags = merge(
+    {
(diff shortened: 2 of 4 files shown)


r4f4:
@barbacbd do you know in which version of the provider the instance profile tagging capability was added? We probably want to backport this fix to all supported OCP versions.

barbacbd:
> @barbacbd do you know in which version of the provider the instance profile tagging capability was added? We probably want to backport this fix to all supported OCP versions.

I don't but I will try to find this out so we can backport

barbacbd (commented):
The installer v4.11 uses the provider v4.5.0. This well beyond the minimum supported version for instance profile tags in the provider. We can backport to 4.11 if we want to. @r4f4 @sadasu

r4f4:
> The installer v4.11 uses the provider v4.5.0. This well beyond the minimum supported version for instance profile tags in the provider. We can backport to 4.11 if we want to. @r4f4 @sadasu

Then I think we should do it.

sadasu:
Does this fix make this method https://github.com/openshift/installer/blob/master/pkg/destroy/aws/aws.go#L368, obsolete?

sadasu:
Do we want to rename "[findUntaggableResources](https://github.com/openshift/installer/blob/master/pkg/destroy/aws/aws.go#L368)" to "findInstanceProfiles" after this fix?

r4f4:
> Do we want to rename "[findUntaggableResources](https://github.com/openshift/installer/blob/master/pkg/destroy/aws/aws.go#L368)" to "findInstanceProfiles" after this fix?

I'd say "no". `findUntaggableResources` is for resources that cannot be tagged, which in this case happened to be instance profiles but could include other types in the future.

r4f4:
That being said, we could probably remove the profile retrieval from `findUntaggableResources` if we also include tagging for the bootstrap profile. That would result in an empty `findUntaggableResources` which we could consider deleting and re-adding once bump into another untaggable resource. Or we could just leave it as-is and the profiles will be deleted either by the tag or by their name.

sadasu:
Should we add the same tag here https://github.com/openshift/installer/blob/master/data/data/aws/bootstrap/main.tf#L83?

r4f4:
> Should we add the same tag here https://github.com/openshift/installer/blob/master/data/data/aws/bootstrap/main.tf#L83?

Yeah I think so. WDYT @barbacbd ?

r4f4 (commented):
/lgtm

PR #7511, opened 2023-09-20
OCPBUGS-19092: Enable serial console for external OCI platform

pawanpinjarkar:
When creating an Agent ISO for OCI, add the kernel argument `console=ttyS0` to the ISO/PXE kargs.

CoreOS does not include a console arg by default when using metal as the platform because different hardware has different consoles and specifying one can cause booting to fail on some, but it [does](https://github.com/coreos/fedora-coreos-config/blob/stable/platforms.yaml) on many cloud platforms. Since we know when the user is definitely using OCI  and we know the correct settings for OCI, we should set them up automatically.

@pawanpinjarkar

### DIFF
--- a/cmd/openshift-install/testdata/agent/pxe/configurations/compact_external.txt
+++ b/cmd/openshift-install/testdata/agent/pxe/configurations/compact_external.txt
@@ -1,31 +1,31 @@
-# Verify a default configuration for the SNO topology for ARM architecture
+# Verify a default configuration for the compact topology for external platform. 
+# Verify kernel args 'console=ttyS0' is added for extenal oci platform.
 
 exec openshift-install agent create pxe-files --dir $WORK
 
-stderr 'Created iPXE script agent.aarch64.ipxe'
+stderr 'Created iPXE script agent.x86_64.ipxe'
 
-exists $WORK/boot-artifacts/agent.aarch64-initrd.img
-exists $WORK/boot-artifacts/agent.aarch64-rootfs.img
-exists $WORK/boot-artifacts/agent.aarch64-vmlinuz
-exists $WORK/boot-artifacts/agent.aarch64.ipxe
+exists $WORK/boot-artifacts/agent.x86_64-initrd.img
+exists $WORK/boot-artifacts/agent.x86_64-rootfs.img
+exists $WORK/boot-artifacts/agent.x86_64-vmlinuz
+exists $WORK/boot-artifacts/agent.x86_64.ipxe
 exists $WORK/auth/kubeconfig
 exists $WORK/auth/kubeadmin-password
 
-grep 'initrd --name initrd http://user-specified-pxe-infra.com/agent.aarch64-initrd.img' $WORK/boot-artifacts/agent.aarch64.ipxe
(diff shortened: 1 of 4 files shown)


pkg/asset/agent/image/agentartifacts.go
```diff
@@ -64,7 +65,18 @@ func (a *AgentArtifacts) Generate(dependencies asset.Parents) error {
...
-	a.Kargs = kargs.KernelCmdLine()
+
+	platform := agentManifests.AgentClusterInstall.Spec.PlatformType
+	var platformName, consoleArgs string
+	if agentManifests.AgentClusterInstall.Spec.ExternalPlatformSpec != nil {
+		platformName = agentManifests.AgentClusterInstall.Spec.ExternalPlatformSpec.PlatformName
+	}
+	if platform == hiveext.ExternalPlatformType && platformName == "oci" {
+		consoleArgs = " console=ttyS0"
+	}
```
zaneb, line 76: I'd like this code to be inside the `KernelCmdLine()` function. This is possible by making ACI a dependency of Kargs.

pkg/asset/agent/image/kargs.go
```diff
@@ -1,29 +1,46 @@
...
+// Generate generates the kernal args configurations for the agent ISO image and PXE assets.
 func (a *Kargs) Generate(dependencies asset.Parents) error {
+	agentManifests := &manifests.AgentManifests{}
+	dependencies.Get(agentManifests)
+
+	platformName := agentManifests.GetExternalPlatformName()
+	// Add kernel args for external oci platform
+	if agentManifests.AgentClusterInstall.Spec.ExternalPlatformSpec != nil &&
+		agentManifests.AgentClusterInstall.Spec.ExternalPlatformSpec.PlatformName == platformName {
+		logrus.Debugf("Added kernal args to enable serial console for %s %s platform", hiveext.ExternalPlatformType, platformName)
```
zaneb, line 32: s/kernal/kernel/

pkg/asset/agent/image/kargs.go
```diff
@@ -1,29 +1,46 @@
...
-// Generate generates the configurations for the agent ISO image and PXE assets.
+// Generate generates the kernal args configurations for the agent ISO image and PXE assets.
 func (a *Kargs) Generate(dependencies asset.Parents) error {
+	agentManifests := &manifests.AgentManifests{}
+	dependencies.Get(agentManifests)
+
+	platformName := agentManifests.GetExternalPlatformName()
+	// Add kernel args for external oci platform
+	if agentManifests.AgentClusterInstall.Spec.ExternalPlatformSpec != nil &&
+		agentManifests.AgentClusterInstall.Spec.ExternalPlatformSpec.PlatformName == platformName {
```
zaneb, line 31: platformName replaced the wrong bit here. This should be:

```
if agentManifests.GetExternalPlatformName() == "oci" {
```

cmd/openshift-install/testdata/agent/pxe/configurations/compact_external.txt
```diff
@@ -0,0 +1,59 @@
...
+
+-- agent-config.yaml --
+apiVersion: v1alpha1
+metadata:
+  name: ostest
+  namespace: cluster0
+rendezvousIP: 192.168.111.20
+bootArtifactsBaseURL: http://user-specified-pxe-infra.com
+
+-- expected/agent.x86_64.ipxe --
```
andfasano, line 55: Is is still really needed?

rwsu (commented):
/approve

zaneb:
Integration test failure looks... persistent: https://prow.ci.openshift.org/pr-history/?org=openshift&repo=installer&pr=7511

PR #7512, opened 2023-09-20
OCPBUGS-19093: skip agent-tui on OCI

pawanpinjarkar:
Network configuration of any kind is rarely needed in the cloud, anyway. So on OCI specifically we mostly are slowing boot down by 20s for no real reason. Disable agent-tui for the external OCI platform by simply not adding the binary to the ISO image.

### DIFF
--- a/pkg/asset/agent/image/agentartifacts.go
+++ b/pkg/asset/agent/image/agentartifacts.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/openshift/assisted-image-service/pkg/isoeditor"
+	"github.com/openshift/assisted-service/models"
 	"github.com/openshift/installer/pkg/asset"
 	config "github.com/openshift/installer/pkg/asset/agent/agentconfig"
 	"github.com/openshift/installer/pkg/asset/agent/manifests"
@@ -40,6 +41,7 @@ func (a *AgentArtifacts) Dependencies() []asset.Asset {
 		&Kargs{},
 		&BaseIso{},
 		&manifests.AgentManifests{},
+		&manifests.AgentClusterInstall{},
 		&mirror.RegistriesConf{},
 		&config.AgentConfig{},
 	}
@@ -51,29 +53,34 @@ func (a *AgentArtifacts) Generate(dependencies asset.Parents) error {
 	kargs := &Kargs{}
 	baseIso := &BaseIso{}
 	agentManifests := &manifests.AgentManifests{}
+	agentClusterInstall := &manifests.AgentClusterInstall{}
 	registriesConf := &mirror.RegistriesConf{}
 	agentconfig := &config.AgentConfig{}
 
-	dependencies.Get(ignition, kargs, baseIso, agentManifests, registriesConf, agentconfig)
+	dependencies.Get(ignition, kargs, baseIso, agentManifests, agentClusterInstall, registriesConf, agentconfig)
 
 	ignitionByte, err := json.Marshal(ignition.Config)
 	if err != nil {
 		return err
 	}
+
 	a.CPUArch = ignition.CPUArch
 	a.RendezvousIP = ignition.RendezvousIP
 	a.IgnitionByte = ignitionByte
 	a.ISOPath = baseIso.File.Filename
 	a.Kargs = kargs.KernelCmdLine()
+
 	if agentconfig.Config != nil {
 		a.BootArtifactsBaseURL = strings.Trim(agentconfig.Config.BootArtifactsBaseURL, "/")
 	}
 
-	agentTuiFiles, err := a.fetchAgentTuiFiles(agentManifests.ClusterImageSet.Spec.ReleaseImage, agentManifests.GetPullSecretData(), registriesConf.MirrorConfig)
-	if err != nil {
-		return err
+	var agentTuiFiles []string
+	if agentClusterInstall.GetExternalPlatformName() != string(models.PlatformTypeOci) {
+		agentTuiFiles, err = a.fetchAgentTuiFiles(agentManifests.ClusterImageSet.Spec.ReleaseImage, agentManifests.GetPullSecretData(), registriesConf.MirrorConfig)
+		if err != nil {
+			return err
+		}
 	}
-
 	err = a.prepareAgentArtifacts(a.ISOPath, agentTuiFiles)
 	if err != nil {
 		return err


pkg/asset/agent/image/agentartifacts.go
```diff
@@ -64,19 +65,31 @@ func (a *AgentArtifacts) Generate(dependencies asset.Parents) error {
...
 	a.IgnitionByte = ignitionByte
 	a.ISOPath = baseIso.File.Filename
+
+	platform := agentManifests.AgentClusterInstall.Spec.PlatformType
+	var platformName string
+	if agentManifests.AgentClusterInstall.Spec.ExternalPlatformSpec != nil {
+		platformName = agentManifests.AgentClusterInstall.Spec.ExternalPlatformSpec.PlatformName
+	}
+	var skipNetworkTUI bool
+	if platform == hiveext.ExternalPlatformType && platformName == "oci" {
```
zaneb, line 75: Maybe we could add an `ExternalPlatformName()` function or something to AgentClusterInstall, since this code has to be repeated here and for the kargs.

pawanpinjarkar: Addded via https://github.com/openshift/installer/pull/7511/files#diff-ed3ddff72757f2b9f8078d820e70ac801ad04c31885e033d6722ac6ae865f10e

pawanpinjarkar:
> > Network configuration of any kind is rarely needed in the cloud, anyway. So on OCI specifically we mostly are slowing boot down by 20s for no real reason. Disable agent-tui for the external OCI platform by simply not adding the binary to the ISO image.
> 
> I'm not clear about this requirement @pawanpinjarkar can you please clarify? Apart the network configuration the agent-tui performs a check about the availability of the release image, and 20s one-off over a workflow that may last several minutes does not seem a big gain.

The main issue is as we have seen in the demo, the network TUI didn't get displayed in the Oracle cloud console. So @zaneb suggested that we should not display the service message and not wait for the TUI , also because any custom network configurations are hardly needed in the cloud. But you have raised a valid point about the availability of release images. @zaneb  WDYT?

zaneb:
> Apart the network configuration the agent-tui performs a check about the availability of the release image, and 20s one-off over a workflow that may last several minutes does not seem a big gain.

It's more that you can't see it, because there's no graphical console. And it blocks systemd output on the serial console while it's running.
In fact, you make a good point about the release image that I hadn't considered, because if the release image is not available agent-tui will never exit and you will never get to the getty screen on the serial console that tells you about how the release image is not available, instead it will be frozen forever.

andfasano:
> It's more that you can't see it, because there's no graphical console. And it blocks systemd output on the serial console while it's running. In fact, you make a good point about the release image that I hadn't considered, because if the release image is not available agent-tui will never exit and you will never get to the getty screen on the serial console that tells you about how the release image is not available, instead it will be frozen forever.

Thanks, that wasn't clear initially from the bug description, but agree that the latter point it's the most relevant, to avoid getting stuck on boot. Wondering if it's worth to have a more generic disabling mechanism, or if it something that could be intercept directly on the agent-tui service level

zaneb:
> Wondering if it's worth to have a more generic disabling mechanism, or if it something that could be intercept directly on the agent-tui service level

This is an intriguing idea. I'm just not sure how to detect the existence of the graphical console:

- /dev/tty1 seems to get unconditionally created, so we can't trigger off the existence of that
- /dev/tty2 (or 3 or 4) is a possibility... do any of these exist automatically or are they only created on demand?
- /dev/fb0 (framebuffer device) maybe??

If we could get this to work then we could even start agent-tui on the serial console when it is available but make them mutually exclusive so we don't get two copies running.

zaneb:
`/sys/class/vtconsole/vtcon0`?
https://docs.kernel.org/fb/fbcon.html

zaneb:
Testing with libvirt indicates that checking for either `/dev/fb0` or `/sys/class/vtconsole/vtcon1` should work for determining when a visual console exists (on x86).
I lean toward `/dev/fb0` because it's not clear to me how many vtcon devices would be created on an architecture where e.g. _only_ a graphical console is supported by the kernel.

zaneb:
See #7526

zaneb:
/close
Addressed by #7526

pkg/asset/agent/image/agentartifacts.go
```diff
@@ -64,19 +68,22 @@ func (a *AgentArtifacts) Generate(dependencies asset.Parents) error {
...
 	a.Kargs = kargs.KernelCmdLine()
 	if agentconfig.Config != nil {
 		a.BootArtifactsBaseURL = strings.Trim(agentconfig.Config.BootArtifactsBaseURL, "/")
 	}
 
-	agentTuiFiles, err := a.fetchAgentTuiFiles(agentManifests.ClusterImageSet.Spec.ReleaseImage, agentManifests.GetPullSecretData(), registriesConf.MirrorConfig)
-	if err != nil {
-		return err
-	}
+	if agentManifests.AgentClusterInstall.Spec.PlatformType != hiveext.ExternalPlatformType && agentClusterInstall.GetExternalPlatformName() != string(models.PlatformTypeOci) {
```
zaneb, line 77: nit: this can reduced to

```
if agentClusterInstall.GetExternalPlatformName() != string(models.PlatformTypeOci) {
```

zaneb (commented):
/approve

pkg/asset/agent/image/agentartifacts.go
zaneb: actually there's a bug as well, if we're going to keep both conditions the `&&` should be `||`, otherwise this will never be true

pawanpinjarkar: right, thanks for catching this. Just one condition to check platformName is also sufficient. Updated accordingly.

pkg/asset/agent/image/agentartifacts.go
```diff
@@ -51,32 +53,37 @@ func (a *AgentArtifacts) Generate(dependencies asset.Parents) error {
...
+	if agentClusterInstall.GetExternalPlatformName() != string(models.PlatformTypeOci) {
+		agentTuiFiles, err := a.fetchAgentTuiFiles(agentManifests.ClusterImageSet.Spec.ReleaseImage, agentManifests.GetPullSecretData(), registriesConf.MirrorConfig)
+		if err != nil {
+			return err
+		}
 
-	err = a.prepareAgentArtifacts(a.ISOPath, agentTuiFiles)
-	if err != nil {
-		return err
+		err = a.prepareAgentArtifacts(a.ISOPath, agentTuiFiles)
```
andfasano, line 83: `prepareAgentArtifacts` must be executed always. It's ok to pass an empty slice to avoid adding agent-tui.

pkg/asset/agent/image/agentartifacts.go
```diff
@@ -51,32 +53,45 @@ func (a *AgentArtifacts) Generate(dependencies asset.Parents) error {
...
+		agentTuiFiles, err = a.fetchAgentTuiFiles(agentManifests.ClusterImageSet.Spec.ReleaseImage, agentManifests.GetPullSecretData(), registriesConf.MirrorConfig)
+		if err != nil {
+			return err
+		}
+
+		err = a.prepareAgentArtifacts(a.ISOPath, agentTuiFiles)
+		if err != nil {
+			return err
+		}
+	} else {
```
bfournie, line 89: Could you remove the else and just call `prepareAgentArtifacts()` once? agentTuiFiles will be empty if `PlatformTypeOci`

mhanss:
The bug has been verified using this PR, its working fine.
`Oct 16 17:01:24 master-0 systemd[1]: Get interactive user configuration at boot was skipped because of an unmet condition check (ConditionPathExists=/usr/local/bin/agent-tui).`
/qe-approved

PR #7513, opened 2023-09-20
OCPBUGS-18876: Pass CPUPartitioning via install-config overrides if set

bfournie:
If CPUPartitioning is set to install-config, set it in the install-config overrides in agentclusterinstall.yaml.


bfournie:
Testing in dev-scripts by setting in install-config.yaml
cpuPartitioningMode: AllNodes

Results in, after cluster is installed:
$ oc get configmaps -n kube-system cluster-config-v1 -o yaml | grep cpuPartitioning
    cpuPartitioningMode: AllNodes

PR #7519, opened 2023-09-21
OCPBUGS-19376: GCP default value for service account

barbacbd:
** When authenticating with VM credentials the empty string was used in terraform, but null should be used.


jianli-wei (approved):

r4f4 (commented):
/lgtm

PR #7520, opened 2023-09-21
CORS-2845: azure: Enable storage account encryption

rna-afk:
Added a field to accept the customer managed key vault id and
user assigned identity required for the encryption of storage
account.

### DIFF
--- a/data/data/azure/vnet/main.tf
+++ b/data/data/azure/vnet/main.tf
@@ -43,15 +43,52 @@ data "azurerm_resource_group" "network" {
   name = var.azure_network_resource_group_name
 }
 
+data "azurerm_key_vault" "keyvault" {
+  count = var.azure_keyvault_name != "" ? 1 : 0
+
+  name                = var.azure_keyvault_name
+  resource_group_name = var.azure_keyvault_resource_group
+}
+
+data "azurerm_key_vault_key" "keyvault_key" {
+  count = var.azure_keyvault_name != "" ? 1 : 0
+
+  name         = var.azure_keyvault_key_name
+  key_vault_id = data.azurerm_key_vault.keyvault[0].id
+}
+
+data "azurerm_user_assigned_identity" "keyvault_identity" {
+  count = var.azure_keyvault_name != "" ? 1 : 0
+
+  resource_group_name = var.azure_keyvault_resource_group
+  name                = var.azure_user_assigned_identity_key
+}
+
 resource "azurerm_storage_account" "cluster" {
   name                            = "cluster${var.random_storage_account_suffix}"
   resource_group_name             = data.azurerm_resource_group.main.name
   location                        = var.azure_region
-  account_tier                    = "Standard"
+  account_tier                    = var.azure_keyvault_name != "" ? "Premium" : "Standard"
   account_replication_type        = "LRS"
   min_tls_version                 = contains(local.environments_with_min_tls_version, var.azure_environment) ? "TLS1_2" : null
-  allow_nested_items_to_be_public = false
+  allow_nested_items_to_be_public = var.azure_keyvault_name != "" ? true : false
   tags                            = var.azure_extra_tags
+
+  dynamic "customer_managed_key" {
+    for_each = var.azure_keyvault_name != "" ? [1] : []
+    content {
+      key_vault_key_id          = data.azurerm_key_vault_key.keyvault_key[0].id
+      user_assigned_identity_id = data.azurerm_user_assigned_identity.keyvault_identity[0].id
+    }
+  }
+
+  dynamic identity {
+    for_each = var.azure_keyvault_name != "" ? [1] : []
+    content {
+      type         = "UserAssigned"
+      identity_ids = [data.azurerm_user_assigned_identity.keyvault_identity[0].id]
+    }
+  }
 }
 
 resource "azurerm_user_assigned_identity" "main" {
--- a/pkg/asset/cluster/tfvars.go
+++ b/pkg/asset/cluster/tfvars.go
@@ -181,6 +181,14 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		}
 	}
 
+	lengthBootstrapFile := int64(len(bootstrapIgn))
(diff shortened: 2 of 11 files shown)


jhixson74:
It looks like there is an issue with generating the terraform variables. Other than that, this looks straight forward and good.

rna-afk:
@gpei can someone from the QE team test this? I'm having some permission problems for creating certs.

gpei:
@rna-afk Sure, @jinyunma could you please take a look at this?

pkg/asset/cluster/tfvars.go
```diff
@@ -181,6 +181,12 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
 		}
 	}
```
jhixson74, line 183: Maybe leave a comment here describing why this is necessary.

pkg/tfvars/tfvars.go
```diff
@@ -27,13 +27,17 @@ type config struct {
 }
 
 // TFVars generates terraform.tfvar JSON for launching the cluster.
-func TFVars(clusterID string, clusterDomain string, baseDomain string, machineV4CIDRs []string, machineV6CIDRs []string, useIPv4, useIPv6 bool, bootstrapIgn string, masterIgn string, masterCount int, mastersSchedulable bool) ([]byte, error) {
+func TFVars(clusterID string, clusterDomain string, baseDomain string, machineV4CIDRs []string, machineV6CIDRs []string, useIPv4, useIPv6 bool, bootstrapIgn string, bootstrapIgnSize int64, masterIgn string, masterCount int, mastersSchedulable bool) ([]byte, error) {
 	f, err := os.CreateTemp("", "openshift-install-bootstrap-*.ign")
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to create tmp file for bootstrap ignition")
 	}
 	defer f.Close()
```
jhixson74, line 36: Here too.

jhixson74 (approved):
This looks good to me.

jhixson74:
/hold cancel 

I think we are good here.

data/data/azure/vnet/main.tf
```diff
@@ -43,15 +43,52 @@ data "azurerm_resource_group" "network" {
...
 resource "azurerm_storage_account" "cluster" {
   name                            = "cluster${var.random_storage_account_suffix}"
   resource_group_name             = data.azurerm_resource_group.main.name
   location                        = var.azure_region
-  account_tier                    = "Standard"
+  account_tier                    = var.azure_keyvault_name != "" ? "Premium" : "Standard"
   account_replication_type        = "LRS"
   min_tls_version                 = contains(local.environments_with_min_tls_version, var.azure_environment) ? "TLS1_2" : null
-  allow_nested_items_to_be_public = false
+  allow_nested_items_to_be_public = var.azure_keyvault_name != "" ? true : false
```
r4f4, line 74: Do you know why this is needed?

pkg/asset/cluster/tfvars.go
```diff
@@ -181,6 +181,12 @@ func (t *TerraformVariables) Generate(parents asset.Parents) error {
...
 
+	lengthBootstrapFile := int64(len(bootstrapIgn))
+	if installConfig.Config.Platform.Azure != nil && installConfig.Config.Platform.Azure.CustomerManagedKey != nil &&
+		installConfig.Config.Platform.Azure.CustomerManagedKey.UserAssignedIdentityKey != "" {
+		lengthBootstrapFile = (((lengthBootstrapFile / 512) + 1) * 512)
```
r4f4, line 187: If the size is 512, you're returning here 1024. Shouldn't 512 be fine (no padding needed)?

pkg/types/azure/validation/platform.go
```diff
@@ -40,6 +40,9 @@ var (
 
 	// tagKeyPrefixRegex is for verifying that the tag value does not contain restricted prefixes.
 	tagKeyPrefixRegex = regexp.MustCompile(`^(?i)(name$|kubernetes\.io|openshift\.io|microsoft|azure|windows)`)
+
+	// subscriptionIDRegex is for verifying the identity key associated with the managed key provided.
+	subscriptionIDRegex = regexp.MustCompile(`^([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})$`)
```
r4f4, line 45: Did you base this on some Azure docs? It would be good to include a link to it.
Ideally, I think we should validate we can access that subscription in the installconfig validation instead of relying on a regex.

data/data/azure/vnet/main.tf
rna-afk: Probably could remove it. @jhixson74 should we remove this?

pkg/types/azure/validation/platform.go
rna-afk: I reckon this checks for the standard UUID regex. @jhixson74 added significantly to the code here, any thoughts?

pkg/asset/cluster/tfvars.go
rna-afk: Fixed this.

data/data/azure/vnet/main.tf
jhixson74: This is required to allow blob access

pkg/types/azure/validation/platform.go
jhixson74: This is  close but should also account for upper case. You can use this one: https://github.com/openshift/installer/blob/7956adbe7f21c37f57bbc3ba3401d7b595eddc81/pkg/types/azure/validation/disk.go#L16

rna-afk: Fixed it

data/data/azure/vnet/main.tf
r4f4: Oh, so it has to be public because it's in a different subscription. That's fine however, since to access it you need the the encryption key.

r4f4 (commented):
LGTM but I have not tested it.

PR #7308, opened 2023-07-10
OCPBUGS-15421: Allow different service account for xpn installs in gcp

patrickdillon:
Rebase of #7284

PR #7523, opened 2023-09-22
[release-4.13] OCPBUGS-19670: Allow different service account for xpn installs in gcp

sadasu:
**Add service account field to the install config for control Planes **Validation to ensure service account field added is only used for xpn installs **Add Validation to ensure that the service account is available in the host project

** Machine manifests will use the provided service account for control planes.

**Pass the service account data used for creating control plane nodes during a shared vpn install.

Manual back port of https://github.com/openshift/installer/pull/7308


sadasu:
Something went wrong here.

PR #7525, opened 2023-09-22
Wait for all operators

deads2k:
Bringing back https://github.com/openshift/installer/pull/7289 with a new commit correcting the problem observed in https://github.com/openshift/installer/pull/7318

### DIFF
--- a/cmd/openshift-install/create.go
+++ b/cmd/openshift-install/create.go
@@ -3,9 +3,11 @@ package main
 import (
 	"context"
 	"crypto/x509"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


deads2k:
many of the payload runs failed with messages like

```
{  configuration has 2 errors:

  * images[2]: duplicate image name 'libvirt-installer' (previously defined by field 'base_images[libvirt-installer]')
  * images[3]: duplicate image name 'upi-installer' (previously defined by field 'base_images[upi-installer]')
}
```

is that normal?

r4f4:
> many of the payload runs failed with messages like
> 
> ```
> {  configuration has 2 errors:
> 
>   * images[2]: duplicate image name 'libvirt-installer' (previously defined by field 'base_images[libvirt-installer]')
>   * images[3]: duplicate image name 'upi-installer' (previously defined by field 'base_images[upi-installer]')
> }
> ```
> 
> is that normal?

It seems the payload job is trying to import installer images but the namespace for this PR already has its own built images. Or something like that. Same thing happened last time when I tried to run payload jobs. I'll report to the testplatform team.

deads2k:
> It seems the payload job is trying to import installer images but the namespace for this PR already has its own built images. Or something like that. Same thing happened last time when I tried to run payload jobs. I'll report to the testplatform team.

Thanks.  That's worth fixing in general, but means we may have to merge based on the data we are able to collect.

deads2k:
closing in favor of https://github.com/openshift/installer/pull/7535

PR #7526, opened 2023-09-25
OCPBUGS-19093,OCPBUGS-19688: Allow agent-tui to use serial console

zaneb:
Run agent-tui on the serial console when there is no graphical console available.

### DIFF
--- a/pkg/asset/agent/image/ignition.go
+++ b/pkg/asset/agent/image/ignition.go
@@ -248,6 +248,7 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
 func getDefaultEnabledServices() []string {
 	return []string{
 		"agent-interactive-console.service",
+		"agent-interactive-console-serial@.service",
 		"agent.service",
 		"assisted-service-db.service",
 		"assisted-service-pod.service",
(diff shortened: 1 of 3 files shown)


zaneb:
@jeffdyoung does this approach (determining whether there is a graphical console by looking for the framebuffer device, `/dev/fb0`) seem valid for ARM/Power as well?

pkg/asset/agent/image/ignition.go
```diff
@@ -248,6 +248,7 @@ func (a *Ignition) Generate(dependencies asset.Parents) error {
 func getDefaultEnabledServices() []string {
 	return []string{
 		"agent-interactive-console.service",
+		"agent-interactive-console-serial@.service",
```
andfasano, line 251: q: how does it get activated (ie who's passing the instance parameter)?

zaneb: Because it has `WantedBy=serial-getty@.service` it gets started with the same parameter as serial-getty when the latter is started.

The go API here is a little weird, but when you look at the Ignition it generates, the effect is that we're just setting enabled=true in the systemd unit entry with this filename.

jeffdyoung:
> @jeffdyoung does this approach (determining whether there is a graphical console by looking for the framebuffer device, `/dev/fb0`) seem valid for ARM/Power as well?

@zaneb I created a libvirt vm w/o graphic console and didn't see a /dev/fb0 device, so I think what you are proposing will work for ARM. 
@jaypoulz @prb112 can you confirm for power?

pkg/asset/agent/image/ignition.go
andfasano: Ok thanks!

prb112:
Based on Manoj's feedback. 

/lgtm

